Issue #1006, opened 2026-07-30
person_evidence links record no match_score in 94% of writes — add an engine invariant (settled: doctrine gap, not an anachronistic check)


DallanQ:
**IMPORTANT**: #738 also extends `personEvidenceInvariants` in `research-append.ts` — same function, different check (a mandatory `conflicts` entry for single-uncorroborated-read links, vs. this issue's missing-`match_score` warn). #738 is `senior`-labeled (novel cross-section-write precedent); this issue is junior-safe. Coordinate rather than land independently — whoever picks this up should flag #738's assignee (or the lead) before merging, so the two changes to the same guard don't collide.

**Settled 2026-08-01 by measurement — this is reading 1, a doctrine gap.** The two-readings framing below is resolved; the disposition is an engine-side invariant on the write. Original analysis kept for the record.

## The measurement that decided it

The issue said distinguishing the readings needed the doctrine date against the run dates. Done:

`same_person` entered `research/SKILL.md` on **2026-07-15** (`39b37bae`, PR #657 — *"never write person_evidence inline — route through person-evidence (+same_person)"*). Splitting every committed e2e run with a tool ledger on that date, counting **only runs that actually write `person_evidence`** (the only runs the check applies to):

| | runs writing `person_evidence` | called `same_person` |
|---|---|---|
| before 2026-07-15 | 48 | 13 (**27%**) |
| on/after 2026-07-15 | 83 | 16 (**19%**) |

**Compliance fell after the doctrine landed.** Reading 2 (anachronistic check) predicts the opposite — high compliance once the rule exists. It is also contradicted directly: `same_person` calls appear from **2026-06-23**, three weeks before the line was written. The check is not measuring its own introduction date.

(Denominators differ from the "7 of 69" above because this counts all 132 committed runs with a tool ledger rather than the current-architecture subset. The direction does not depend on the subset.)

## The sharper number, and the seam that makes this fixable

`same_person` returns a float 0–1 (`packages/engine/mcp-server/src/tools/same-person.ts:80,140`). `person_evidence_entry.match_score` is typed `number` 0–1 **or null** (`docs/specs/schemas/research.schema.json`) — an exact type match, and it is optional rather than required. It is unmistakably the field meant to carry that output.

Across every committed `*.final-research.json`:

> **384 of 6,001 `person_evidence` entries carry a `match_score` — 6%.**

[...]

PR #1340, opened 2026-08-05
eval-harness: baseline the proof-conclusion compliance

jud-sdev:
Closes #998 

## The gap

`find_effects_without_invocation`'s proof-conclusion arm treated the mere presence of a `ParentChild`/`Couple` relationship (or any `primary: true` fact) in the **final** tree as e
vidence the run produced a conclusion — with no starting-tree baseline. Every FamilySearch-seeded fixture ships those relationships, so the arm's precondition was satisfied by seed
 state alone, and the check collapsed into "was proof-conclusion invoked at all?" — wired into the hard-fail path that overrides the judge.

The person-evidence arm right below it already took a `starting_tree` baseline; this arm never got the same treatment.

## The fix

Give the proof-conclusion arm the same `starting_tree` baseline, counting only conclusion effects that **appeared this run**:

- **Relationships** are diffed against the starting tree on the **endpoint tuple** `(type, person1, person2, parent, child)`, not on `id`. 7 fixtures seed a relationship pointing a
t a `PID-TODO` placeholder that the agent resolves during the run; an `id` key would read that genuinely-re-pointed relationship as seeded — a false negative in the one gate that o
verrides the judge. (Verified: a re-pointed placeholder correctly fires; an identical seeded relationship correctly does not.)
- **`primary: true` facts** are diffed per person against the starting count. Defensive — 0 of 104 fixtures ship one — but baselined the same way.
- The **`proof_summaries` half is left exactly as-is**: no fixture in the corpus seeds a non-empty `proof_summaries` (verified across all fixtures), so it needs no baseline.
- `starting_tree is None` ⇒ everything counts as new (best-effort), matching the person-evidence arm.

The orchestrator is untouched — `starting_tree` was already plumbed into this function.

## Acceptance — corpus replay

A throwaway replay over all committed `eval/runlogs/e2e/*/run-*.json` (paired with each run's final tree/research and its fixture's starting tree), evaluating the arm before and af
ter. Current corpus: **142 runs**.

|              | Before | After |
|--------------|:------:|:-----:|
| runs firing any violation | 75 | 71 |
| proof-conclusion arm fires | 56 | 52 |

**Exactly the 4 seed-only runs flip fire→clean, and no other run changes in either direction:**

- `pierre-tullier-son/run-2026-07-27_21-34-22`
- `william-ferber-origins/run-2026-07-29_02-31-20`
- `william-ferber-origins/run-2026-07-29_12-16-49`
- `young-marriage-1828/run-2026-07-21_20-47-01`

The change is monotonic — the arm only ever *stops* firing, never newly fires — so no false positive is introduced. (The issue's "54→50" was measured on the then-133-run corpus; sa
me 4 flip.)

**Regression guard — `cornelius-booysen-death/run-2026-07-30_18-09-08` correctly keeps firing.** Its `proof_summaries` is genuinely new (starting 0, final 1), so it is a real "conc
lusion written without invoking proof-conclusion" violation, not seed noise. The fix leaves that path intact.

## Tests

`eval/harness/tests/unit/test_skill_invocation.py`, mirroring the person-evidence pair:
- a seeded relationship present in both trees must **not** flag,
- a relationship present only in the final tree **must** flag,
- (defensive) a seeded primary fact present in both trees must **not** flag.

`make harness-test` (via `uv run pytest`): **1596 passed, 6 skipped.** `ruff check` clean.

## Docs

- `skill_invocation.py` docstring — `starting_tree` now consulted by both arms.
- `docs/specs/guardrail-enforcement-spec.md` §9 — "three arms read whole-document state" → **two** (research-exhaustiveness, conflict-resolution); issue list marks #998 closed.
- `docs/agentic-system-critique.md` §0.2 and §3 P0 — #998 marked landed.
- `docs/architecture.md` §9.4 and the `orchestrator.py` spec citation were **already** clean on main (a prior merge removed the #998 row / repointed `docs/plan/`→`docs/specs/`) — t
he issue's lines predated that.

## Notes for the reviewer

- **`katalin-horak-son/run-2026-08-03_21-26-35`** (named as a second regression case) is **not in the committed corpus** — no such fixture/run dir on main — so it could not be vali
dated. Needs a corrected slug/timestamp, or it lives on an unmerged branch.
- **Overlap with #1218:** `agentic-system-critique.md` lines ~890/1001 still count #998 among "open defects," but they sit inside the 8-of-25 rate narrative that #1218 rewrites who
lesale. Deliberately left untouched here to avoid a needless conflict on that PR — whichever lands second drops #998 to two.

## Test plan

- [X] I ran `make test-all` (or `scripts/test.sh` — the same command) and it passed.
      
      

- [X] For every skill whose **run-log snapshot** I changed (anything under its
      skill dir, an agent it delegates to, its `eval/tests/unit/<skill>/`, or a
      scenario/fixture it references), I ran `make eval-skill SKILL=<name>` —
      Windows: `RunTests.bat` — and committed the run log **and its
      `.ann.json`** under `eval/runlogs/unit/<skill>/`.


eval/harness/tests/unit/test_skill_invocation.py
```diff
@@ -288,6 +288,36 @@ def test_flags_a_parent_child_relationship_with_no_proof_conclusion_invocation()
...
+
+
+def test_does_not_flag_a_seeded_primary_fact_already_in_the_starting_tree():
+    """Defensive: no fixture ships a primary:true fact today, but the primary-fact
+    half is baselined the same way so a future seeded one cannot fire the arm."""
+    person = {"id": "I1", "facts": [{"type": "Death", "primary": True}]}
+    tree = {"persons": [person], "relationships": []}
+    starting = {"persons": [person]}
+    violations = find_effects_without_invocation([], {}, tree, starting_tree=starting)
+    assert not any("proof-conclusion" in v for v in violations)
```
Gennecis, line 318: **Blocking: the endpoint-vs-`id` decision has no test defending it.**

This is the PR's central technical argument — keying on endpoints rather than `id` is justified as preventing *"a false negative in the one gate that overrides the judge."* The PR says it was verified, but by hand, with nothing committed.

I mutated `_relationship_key` to `return (r.get("id"),)`, confirmed the mutated module was the one actually imported, and ran this file:

```
88 passed in 3.73s
```

All 88 tests stay green while the false negative is fully reinstated. The three new tests can't tell the two designs apart — the seeded case has a matching `id`, and the new-relationship case has no starting relationship at all. Given this repo's own standard (`test_e2e_corpus_report.py:135` exists precisely because a hand-typed probe let two live arms sit unmapped), this shouldn't ship uncovered.

**Fix — add after this line:**

```python
def test_flags_a_placeholder_relationship_re_pointed_this_run():
    """The endpoint-tuple key, not `id`: 7 fixtures seed a relationship pointing at
    a PID-TODO placeholder that the agent resolves during the run. Keying on `id`
    would read that genuinely-re-pointed relationship as seeded — a false negative
    in the one gate that overrides the judge."""
    starting = {"persons": [], "relationships": [
        {"id": "rel-1", "type": "ParentChild", "parent": "PID-TODO", "child": "p-child-thomas"}]}
    tree = {"persons": [], "relationships": [
        {"id": "rel-1", "type": "ParentChild", "parent": "G7X1-234", "child": "p-child-thomas"}]}
    violations = find_effects_without_invocation([], {}, tree, starting_tree=starting)
    assert any("proof-conclusion" in v for v in violations)
```

The shape is taken from real data — `young-marriage-1828` seeds `{"type": "ParentChild", "parent": "PID-TODO", "child": "p-child-thomas", "id": "rel-1"}`. This test fails under the id-keyed mutation and passes as written.

eval/harness/harness/skill_invocation.py
```diff
@@ -305,29 +308,67 @@ def find_effects_without_invocation(
...
+        for r in ((starting_tree or {}).get("relationships") or [])
+        if isinstance(r, dict) and r.get("type") in ("ParentChild", "Couple")
+    }
+
+    def _has_new_primary_fact(p: dict[str, Any]) -> bool:
+        count = sum(1 for f in (p.get("facts") or []) if isinstance(f, dict) and f.get("primary") is True)
+        baseline = 0 if starting_tree is None else starting_primary_counts.get(p.get("id"), 0)
+        return count > baseline
+
+    # A conclusion's tree encoding counts only when it is NEW this run. The
```
Gennecis: **Non-blocking, but this comment claims more than the code does.**

The key on line 336 sees *which people a relationship connects*, so it can't see a conclusion written **into** a relationship that already exists. That's a real genealogical shape, not a hypothetical: dating a marriage on an existing couple, or re-classifying a parentage `Biological` → `Adopted`, are both GPS element-5 conclusions requiring a reasoned argument.

`eval/tests/e2e/susanna-dawson-marriage` seeds `Couple R1` **with no facts**, and its own required finding `f1` is a Marriage fact (21 October 1860, Lezayre) to be written onto exactly that couple. Running both detector versions against that fixture, with no `proof_summary` and `proof-conclusion` never invoked:

| Scenario | Old | New |
|---|---|---|
| A — marriage date written onto seeded `Couple R1` (*the fixture's own answer `f1`*) | FIRES | **silent** |
| B — seeded parentage `R16` re-classified Biological → Adopted | FIRES | **silent** |
| C — control: a genuinely new `ParentChild` added | FIRES | FIRES ✓ |
| D — control: untouched seeded tree (the bug being fixed) | FIRES | silent ✓ |

C and D are exactly right — the fix does what it says. A and B are the cost. `_has_new_primary_fact` on line 346 has the same shape: it compares *counts*, so a primary fact **replaced** (1 → 1) reads as unchanged, which is the one case that half exists to future-proof against.

This is mitigated — the `proof_summaries` disjunct still catches the normal path, so the miss only bites when the agent encodes a conclusion in the tree *and* skips the summary. Not worth blocking on, and extending the key risks new false positives that would break the monotonicity guarantee this PR is built on. But the comment should say what the arm actually covers, so the next reader doesn't over-trust it.

**Fix:** replace *"A conclusion's tree encoding counts only when it is NEW this run"* with wording that names the real scope — the arm detects a conclusion encoded as a **new relationship** or an **added primary fact**, and does **not** detect one written into an existing relationship (a marriage fact, a `subtype` correction) or a replaced primary fact. Same correction in the docstring at lines 273-280, which currently says each arm "flags only what this run produced."

docs/specs/guardrail-enforcement-spec.md
```diff
@@ -381,10 +381,11 @@ this section before reopening one.
...
 - **Per-turn scoping for a production detector** — proposed for #1054 and dead.
-  Three of `find_effects_without_invocation`'s arms read whole-document state
-  with no baseline, so at turn scope they latch permanently after the first
-  legitimate write; and a hosted "turn" is one user message, while an autonomous
-  research request is the whole run.
+  Two of `find_effects_without_invocation`'s arms (research-exhaustiveness,
+  conflict-resolution) read whole-document state with no baseline, so at turn
+  scope they latch permanently after the first legitimate write; and a hosted
+  "turn" is one user message, while an autonomous research request is the whole
+  run. (The proof-conclusion and person-evidence arms are `starting_tree`-baselined.)
```
Gennecis: **"Two" undercounts, and the parenthetical overstates.**

The proof-conclusion arm is baselined on its **tree** half only. Its `proof_summaries` disjunct — read at `skill_invocation.py:308`, used in the condition at `:361` — is still unbaselined whole-document state. The PR body says so itself: *"The `proof_summaries` half is left exactly as-is."*

So the arm **does** still latch at turn scope, which is the exact argument this bullet is making against per-turn scoping. As written, the edit weakens a rationale that's still valid — and this is the spec, which is what an implementation gets checked against.

Worth folding in while you're here: a `starting_tree` baseline is a **run-level** baseline, so it doesn't rescue *any* arm at turn scope. A person legitimately enriched at turn 3 still reads as above-baseline at turn 10.

```suggestion
  Three of `find_effects_without_invocation`'s arms read whole-document state
  with no baseline — research-exhaustiveness, conflict-resolution, and
  proof-conclusion's `proof_summaries` half — so at turn scope they latch
  permanently after the first legitimate write; and a hosted "turn" is one user
  message, while an autonomous research request is the whole run.
  (proof-conclusion's tree half and the person-evidence arm are
  `starting_tree`-baselined — a run-level baseline, which does not help at turn
  scope.)
```

Gennecis (changes requested):
## Request changes — 1 blocking ask, and it's a small one

Adds a `starting_tree` baseline to `find_effects_without_invocation`'s proof-conclusion arm so a fixture's seeded relationships and primary facts stop counting as this run's conclusion.

**The core fix is correct, and it's better-evidenced than most changes to this harness.** I reproduced the acceptance claim independently rather than taking it — loaded the merge-base and branch versions of the module side by side over all 142 committed run logs. The `proof-conclusion` arm goes **56 → 52**, the flipping runs are **exactly** the 4 named, **0 runs newly fire**, and `cornelius-booysen-death/run-2026-07-30_18-09-08` still fires. Every factual claim in the code comments checks out exactly: 99/104 fixtures seed a `ParentChild`/`Couple`, 7 seed a `PID-TODO` relationship, 0 seed a `primary: true` fact, 0 seed a non-empty `proof_summaries`.

Two things worth calling out because they were easy to get wrong and didn't get got wrong: keeping the annotated `#998` in the References line rather than deleting it holds the `doc-links.test.ts` issue-ref ratchet at exactly 17 (a deletion would have failed it), and the reworded violation message still contains the `"proof-conclusion"` substring `corpus_report.py:83` classifies on, so no arm silently reclassifies to `other`.

### Blocking

| Where | Ask |
|---|---|
| `eval/harness/tests/unit/test_skill_invocation.py:318` | Add a test for the re-pointed placeholder — the design decision this PR rests on currently survives being reversed |

### Non-blocking

| Where | Ask |
|---|---|
| `eval/harness/harness/skill_invocation.py:349` + docstring `273-280` | Narrow the "counts only conclusion effects new this run" claim — the arm doesn't see a conclusion written *into* an existing relationship |
| `docs/specs/guardrail-enforcement-spec.md:384,388` | "Two arms" undercounts; `proof_summaries` is still unbaselined |
| `docs/adrs/ADR-0003-…:92,131` | Still lists `#998` among "three open defects" (outside this diff — see below) |

The first three are inline on their lines. ↓

**Finding 4 — `ADR-0003-anchor-cross-turn-rules-structurally.md`, outside this diff:**

```
92:  … the detectors themselves currently have three open defects and an
     unquantified false-positive rate | #998, #999, #1006 |
131: It is uncalibrated (#998, #999,
```

This PR makes exactly this correction in the critique's P0 heading (`#998, #999, #1006` → `#999, #1006`) but the same claim in the ADR is untouched. Since the reviewer notes deliberately document leaving critique lines ~890/1001 alone to avoid conflicting with #1218, this one reads as unnoticed rather than chosen.

Two reasons it's worth picking up here. **Nothing catches it** — `doc-links.test.ts` exempts `docs/adrs/` on the rationale *"An ADR is a dated decision record… History, not status."* But `"currently have three open defects"` and `"It is uncalibrated"` are status claims, so the exemption's own reasoning doesn't cover them. And ADR-0003 carries `**Last updated:** 2026-08-02`, so ADRs here are maintained, not frozen. This is the failure mode CLAUDE.md names directly: *"Two files here spent weeks claiming 'not yet implemented'… for things that had shipped."*

**Fix:** line 92 → "two open defects", refs `#999, #1006`. Line 131 → `It is uncalibrated (#999,` … `#1006)`.

<details>
<summary><b>How the above was verified</b></summary>

- `uv run --frozen pytest` on branch head: **1602 passed**. Test-merged `origin/main` (clean, no conflicts — and `main` never touched any of these 4 files, so no semantic-conflict risk) and re-ran: **1596 passed, 6 skipped**, matching the PR body exactly.
- `ruff check`: clean before and after the merge.
- `npx vitest run tests/packaging/doc-links.test.ts`: 113 passed. Issue-ref count in `guardrail-enforcement-spec.md` is 17 at merge-base, at branch head, and on `origin/main` — ratchet holds.
- Corpus replay over all 142 run logs, pairing each with its final tree/research and its fixture's starting tree, evaluating both module versions.
- Baseline plumbing confirmed: `orchestrator.py:1754` reads the tree **before** `_run_agent`, from the stripped fixture tree.
- The endpoint-vs-`id` rationale is stronger than the PR states: all **965** seeded relationships carry ids, and they are local sequential labels (`R2`, `rel-1`), not stable identities — so an id key would be fragile in both directions, not just on the placeholder case.

</details>

<details>
<summary><b>Notes that need no action</b></summary>

**The acceptance table's "any violation" row.** "75 → 71" and "exactly the 4 flip" reproduce exactly — but only when *any violation* means `find_effects_without_invocation`'s four arms alone. Against the harness's full compliance axis (`check_guardrail_compliance`, which also runs `find_missing_mentor_verdicts` and `find_person_evidence_missing_same_person`) the corpus goes **130 → 127** with **3** runs flipping fully clean, because `pierre-tullier-son/run-2026-07-27_21-34-22` still carries another arm's violation. No code issue — just worth naming the detector set in the table so the next reader can reproduce it. The `proof-conclusion` row is unambiguous either way.

**`katalin-horak-son`.** The reviewer notes say "no such fixture/run dir on main." Half right: `eval/tests/e2e/katalin-horak-son/` does exist (fixture.json, README, both trees); only the *run log* dir is missing. Worth correcting so nobody acts on "no such fixture."

**One orphan in the corpus.** `eval/runlogs/e2e/william-ferber-ancestry/run-2026-07-28_13-59-05` has no matching `eval/tests/e2e/william-ferber-ancestry` fixture, so `starting_tree` is `None` for it and that run takes the "everything counts as new" best-effort path. 1 of 142 runs is unbaselined. Pre-existing, not this PR's doing.

</details>

eval/harness/harness/skill_invocation.py
Gennecis: Filed as #1368 (Backlog, `developer`) — so the code half is tracked and is **not** this PR's job.

The ask on this thread is only the wording: `:349` and the docstring at `:273-280` should say the arm detects a conclusion encoded as a *new relationship* or an *added primary fact*, rather than "counts only conclusion effects new this run."

Please don't widen the key here. #1368 spells out why the naive version (`facts` straight into the tuple) would re-introduce false positives on a harmless re-serialisation and cost you the monotonicity property this PR just established — that needs a normalised signature and its own corpus replay, which is more than this PR should carry.

Gennecis (approved):

Issue #1358, opened 2026-08-05
Graduate the citation-nulling shadow check to a hard 4th compliance failure


kofiatinka12:
> **Reviewed 2026-08-21 before junior handoff, and the lead has ruled.** Three premises in
> the old ask were false and the measurement is already in. **Do not run the
> measure-then-decide choreography.** This card is now: land the tree-side arm, correct the
> record. It does **not** graduate anything. The original follows under "## Original issue".

## Ruling (lead, 2026-08-21) — detect only

Land the tree-side shadow arm and stop there. The citation-copy obligation stays with
`proof-conclusion` per `docs/specs/tree-materialization-spec.md`; this card only makes its
failure rate readable. A follow-on card owns the fix once the arm has been in the corpus for
a few runs. **Detect before teaching**, because a skill edit costs a paid eval run and there
is no baseline the edit could be measured against. Record this in
`docs/specs/guardrail-enforcement-spec.md`'s shadow-to-graduate table as part of this task.

## What was measured, and what it falsifies

1. **The fire rate is zero.** `make e2e-guardrail-shadow SINCE=all` — measured 2026-08-21,
   157 committed runs — prints
   `§7.5 citation-nulling check (issue #1133, shadow): 0 concluded source(s) … across 0 run(s)`.
2. **Zero is not "the detector never ran".** Replaying `find_citation_nulling_in_conclusions`
   over every committed `run-*.final-research.json` reaches **1,884 concluded sources across
   141 runs — all 1,884 carry a non-empty citation**, with zero dangling assertion or source
   refs. The predicate is exercised and is never true.
3. **Therefore this check does not graduate.**
   `docs/specs/guardrail-enforcement-spec.md`, "What is actually in the shadow-to-graduate
   pipeline", already says so: *"A zero fire rate is not a licence to graduate … zero is not
   'low enough', it is nobody has seen this detector fire."* `docs/architecture.md`'s
   what-nothing-checks section says the same from the other side. Both were written before
   this issue's decision branches; the branches lose.
4. **The failure class it was built for is on the other side of the seam.** Over the same
   corpus, tree `sources[]` entries referenced by a `primary` (proof-backed) fact or a
   relationship, in a run that wrote a conclusion: **109 of 167 have a null/empty citation,
   across 49 of the 144 concluded runs** (offline replay of the committed
   `run-*.final-tree.gedcomx.json` sidecars, 2026-08-21 — turning this into a repeatable
[...]

PR #1360, opened 2026-08-05
eval: shadow-mode detector for citation-string nulling in conclusions (#1133)

kofiatinka12:
Closes #1133 (as scoped below). Follow-up graduation: #1358.

## What #1133 asked, and the supersession call
#1133 asked for an eval detector for **provenance nulling**, and explicitly to decide whether it is superseded by the engine-side golden test in `tree-materialization-spec.md`. **It is partly superseded:**

- **Source-ref half** (a fact/name/edge pointing at no source — the cruz "0/13 facts carried a ref" leak) is now **unrepresentable at the write seam** — the mandatory-ref golden in `materialize-facts.test.ts` / `tree-edit.test.ts`. A post-hoc detector for it would be redundant, so this PR does **not** add one.
- **Citation-string half** — which that golden **explicitly disowns** (`tree-materialization-spec.md`: *"The ESM citation string is out of scope here"*) — is still unguarded. That is the cruz "11/14 citation-less tree sources" / birkeland "F1/F2 citation-less conclusion facts" class, and none of the three §7.5 compliance checks read it. **This is the residual gap this PR closes.**

## What it adds
`find_citation_nulling_in_conclusions` — a post-hoc read of the final `research.json` that flags a source **backing a written conclusion** (`proof_summaries.supporting_assertion_ids` → `assertion.source_id` → `sources[].id`) whose ESM `citation` string is null/empty.

- **Gated on a `proof_summaries` entry.** Tree citations are populated by `proof-conclusion` at *upload* time, so a run that legitimately stops before upload has empty citations by design; scoping to sources of an actual written conclusion keeps a future hard version from false-positiving on honest partial runs.
- **Reads research.json only** (upload-independent) — the citation string is authored there before it is copied to the tree.

## Shadow mode, not a hard fail (deliberate)
Per the repo's **shadow → measure → graduate** posture (same as §7's recency check and the #963 provenance check), it **logs to `guardrail_shadow_violations` and never touches `compliance`/`outcome`.** Entries carry `kind: "citation_nulling"` so `guardrail_shadow_report` counts them in their own bucket (`make e2e-guardrail-shadow`) — separately from the #963 gaps, which also carry `detail`. Graduating to a hard 4th check is gated on reading that fire rate first, filed as **#1358** (we do not flip blind — that is the trap #1231 is stuck in).

## Tests
- `test_skill_invocation.py`: fires on null/empty/whitespace citation; silent when present; **inert without a `proof_summaries` entry**; only flags sources backing the conclusion; dedups one source cited by many assertions; tolerates dangling/sourceless refs; defensive on None/{}.
- `test_guardrail_shadow_report.py`: citation entries counted in their own bucket and **excluded** from the #963 provenance count.
- Full harness unit suite: 1609 passed, 1 skipped. Report CLI smoke: clean, new bucket prints.

## Docs
- `e2e-test-spec.md §7.5`: documents the shadow 4th check + graduation gate.
- `record-extraction-consolidation-closing-report.md §4`: rider recording both halves' status.


Cia-3 (approved):
**Verified the central claim.** #1133 asked to decide whether this is
superseded by an engine-side golden test. Checked both sides directly:
`materialize-facts.test.ts:306` really has the "(8) GOLDEN: every fact
and name written carries a non-null source-ref" test, and
`tree-materialization-spec.md:364` really does say "The ESM citation
string is out of scope here." So the source-ref half is genuinely
covered elsewhere, and this PR correctly targets only the remaining gap.

**Detector logic is sound.** Gated on a `proof_summaries` entry (so
honest partial runs can't false-positive), walks the real
proof_summary → assertion → source chain, dedups per
(proof_summary_id, source_id), and tolerates dangling refs without
erroring — correctly treating those as a schema concern, not this
detector's.

**Shadow-mode discipline is well-evidenced, not just cautious by
default.** Pulled up #1231 (the analogous same_person graduation
issue) — replaying that check against the committed corpus would have
wrongly denied 100 of 103 runs on an id-mismatch bug, not real agent
misbehavior. That's concrete precedent for why this PR is right to log
only (`guardrail_shadow_violations`, `kind: "citation_nulling"`) and
file the graduation decision separately (#1358) rather than flipping
straight to a hard check.

**The refactor is the right shape.** `scan_provenance` →
`_scan_stored` + `keep` predicate correctly keeps the #963 and #1133
violation classes from cross-contaminating each other's counts, even
though both share a `detail` key — and it's the one thing directly
tested (`test_scan_provenance_excludes_citation_entries`), which is
exactly the risk worth testing here.

CI is fully green across all 14 checks.

**One minor, non-blocking note:** dedup is per-conclusion, not
per-source globally — a citation-less source backing two different
`proof_summaries` logs twice. Probably fine, but worth keeping in mind
when #1358 reads the fire-rate numbers, since it could slightly
inflate the count relative to "how many sources are actually broken."

PR #1363, opened 2026-08-05
record_search: time out, retry with backoff, and surface a distinguishable transient-failure error

Asimi1234:
## Summary

- `record_search` issued a bare `fetch` with no timeout and no retry: a slow/stalled FamilySearch connection hung the turn, one transient blip was fatal, and a failed search was indistinguishable from an exhaustive one that found little — silently corrupting exhaustiveness (#1316).
- Now each attempt carries an `AbortSignal.timeout(25_000)`, wrapped in the existing `withRetry({ attempts: 3 })`. Transient states (429/5xx, network error, timeout) are thrown-and-retried; permanent 4xx (400/401/403/404) are returned and handled unchanged. On exhaustion the tool throws an explicit, distinguishable terminal error — `"…did not complete after 3 attempts…: ${detail}. This is a transient failure, NOT an empty result — coverage is unknown."` — rather than returning a short result set.
- **Scoped to exactly #1316.** The terminal error states the fact and stops — no genealogy workflow advice baked into a tool error (see "Deviated from the plan"). Touches `record-search.ts`, its spec, and its tests; nothing else.

Tier: Normal.

## Review guidance

**Start here:** `fetchSearchWithRetry` in `src/tools/record-search.ts`. The load-bearing decision is *classification by throw-vs-return* — transient states throw so `withRetry` retries them, permanent 4xx are returned so the existing `!response.ok` block handles them once. This is what lets `withRetry`'s "retry every thrown error" contract stay correct with no `shouldRetry` predicate and no change to the shared util. Check that the transient/permanent split (429/5xx thrown; 400/401/403/404 returned) is the one you'd want.

**Unsure about:** whether 429 should share the same ~200/400ms backoff as 5xx/network — hammering a throttled endpoint 3× in <1s can extend the throttle. Left uniform for now. 429/5xx are retried *deliberately*, not just timeouts: the tester reported searches that "kept failing," so the transient set is `{timeout, network, 429, 5xx}` — narrowing it to timeouts-only would leave a reported failure unfixed. Honoring `Retry-After` widens `withRetry`'s signature and is deferred to #1362.

## Plan

**Didn't change:** the identical bare-fetch-no-timeout pattern in `fulltext_search` / `external_links_search` / `image_read`. Scoped out per "one problem per PR" and filed as #1362 (likely a shared `fs-fetch` helper) — left alone deliberately, not overlooked. The `25s` / `3-attempts` values are kept as plain literals with justifying comments, not exposed as config — that would add scope, not remove an assumption.

**Acceptance check:** `tests/tools/record-search.test.ts` → "retries a timeout and throws the terminal error when exhausted" and "retries 5xx and throws the terminal transient-failure error when exhausted". Both assert a *throw* (never a short/empty result) after 3 attempts — assertions that fail on `main` (which neither retries nor produces a distinguishable error). Also confirmed live: forcing `SEARCH_TIMEOUT_MS=1` against live FamilySearch produced the terminal error end-to-end after 3 attempts (underlying detail "The operation was aborted due to timeout").

**Deviated from the plan:** the terminal error's original wording carried genealogy workflow advice — *"…Retry the search (optionally narrowing the query) before treating this name or locality as searched."* That is doctrine, not a tooling contract, and out of scope for #1316; trimmed to the distinguishable fact only, and the spec row updated to match. Earlier, a spec lint (`doc-links.test.ts`, "state the fact, not the ticket") also required stripping `#1316` refs from spec prose (kept in source comments).

## Test plan

- [x] `make test-all` was green on the branch before this revision (author's run: engine unit 2040/2040, eval-harness pytest 1599/1599), minus one pre-existing flaky timeout in `apps/electron` (`feedback.test.ts` "drops the largest files… over the archive budget") that passes in ~2.2s in isolation and only times out under full-parallel load — an unrelated package.
- [x] After this revision (merge of `origin/main` + the workflow-advice trim): `tsc --noEmit` clean and `record-search.test.ts` **80/80**. The trim is a message-string change with no other suite surface; the merge brought only doc/agent commits (#1341, #1345).
- [ ] No skill run-log snapshot changed — lane-1 tool fix (no `SKILL.md`, agent, `eval/tests/unit/`, or scenario/fixture touched), so `make eval-skill` does not apply.

## Follow-on issues

- #1362 — generalize the timeout + retry to the sibling search tools (`fulltext_search`, `external_links_search`, `image_read`).

Closes #1316


yinkid28 (commented):
**Review: full pass, 0 commits behind main.**

Verified: vitest 2051/2052 (1 pre-existing `doc-links.test.ts`), `record-search.test.ts` 80/80, `tsc --noEmit` clean, pytest 1592/1592.

The throw-vs-return classification is clean — leverages `withRetry`'s existing contract without needing a `shouldRetry` predicate or modifying the shared util. `getValidToken()` correctly stays outside the retry so auth failures surface immediately, and the fresh `AbortSignal.timeout` per attempt (constructed inside `fetchSearchWithRetry`, not shared) is well-documented.

One minor note: 429 is half of the transient condition (`response.status === 429 || response.status >= 500`) but has no dedicated test — the 5xx test covers the same branch, so behavior is verified, but an explicit `it("retries 429")` would make the test self-documenting. Not blocking.

Ready to merge.

yinkid28 (approved):
Looks good — see my earlier comment for the full review notes.

PR #1240, opened 2026-08-04
eval-harness: stub orchestrator auth for _run_agent unit tests (#1201)

Asimi1234:
## Summary

closes #1201

- **Normal tier.** Adds `eval/harness/tests/unit/conftest.py`: a suite-wide `autouse` fixture stubbing `orchestrator.resolve_auth` to a canned `AuthConfig` — the value `test_cli.py` already uses in 10 places.
- Fixes #1201. The 7 `_run_agent`-driving tests in `test_e2e_stall_resume.py` never stub auth: `_run_agent` evaluates `env_for_sdk(resolve_auth())` *while assembling* the mocked `query()` call, so the auth check fires before the mock runs. On keyless CI they pass only because an earlier test leaks `ANTHROPIC_API_KEY` into `os.environ` — order luck.

## Secondary change — doc-links lint hardening (unrelated to #1201)

Commit `3e3e9fde` also lands an **independent** test-infra fix to `packages/engine/mcp-server/tests/packaging/doc-links.test.ts`, flagged separately here so a reviewer approving the auth stub isn't silently approving a slash-command-lint change. It is not required for #1201 — verified: the pre-`3e3e9fde` test passes on this branch (74/74), so this branch's CI does not depend on it.

- **Prototype-chain exemption bug (real bug fix).** `name in BUILT_INS` walks the prototype chain, so a cited `/constructor`, `/tostring`, or `/valueof` would match an inherited `Object.prototype` key and be silently treated as exempt. Replaced with an own-property `has()` helper (`Object.prototype.hasOwnProperty.call`).
- **Slash-command over-match escape hatch.** Adds `KNOWN_NON_COMMANDS` (currently empty) — the same shape and self-cleaning staleness guarantee as `BUILT_INS`/`KNOWN_ABSENT` — so a slash-shaped non-command (`/tmp`, `/search`, a REST route) cited in a code span can be exempted by name-with-reason instead of misusing `BUILT_INS` or rewording the doc.
- **Makefile parse dedup (perf).** The Makefile is identical across every linted file; it was being re-parsed inside the per-file `it.each`. Parsed once for the suite. Same for the cited-commands set feeding the staleness guards.

## Review guidance

[...]

PR #1364, opened 2026-08-05
eval-harness: one requires_engine_build marker for build-dependent tests

clack391:
## Summary

Normal tier. Issue #1265. Replaces six hand-rolled "is the compiled engine build present?"
skip-guards in the harness suite with one declared `requires_engine_build` marker plus a
behavioral, session-cached probe, and prints a session-end summary line so a build-absent skip
is stated instead of buried in `-rs` output.

Why it matters: those tests need `packages/engine/mcp-server/build/`, a Makefile prerequisite
of every `make` target that CI always builds, but a bare `uv run pytest` in a fresh worktree
has none. One guard was forgotten for months (a red that was not a regression, the bug behind
this issue), and the fix for that made the skips dark: `pytest -q` printed `N skipped` and
moved on. A test that does not run protects nothing, so the drop is now stated.

## Review guidance

**Start here:** `eval/harness/tests/conftest.py`. The load-bearing pieces are the behavioral
probe `_engine_build_available()` (returns False when the build is absent, `node` is missing,
or node_modules is unresolvable, so a path-existence check would turn three skips into crashes
on a no-`node` machine), and the `pytest_runtest_setup` / `pytest_terminal_summary` pair that
skips marked tests and reports only the marker-driven count.

**Unsure about:** nothing blocking. One deliberate design choice to confirm: the probe uses one
combined gate (catalog present AND validator present), so a rare asymmetric partial build would
skip all six even though three have their real dependency. `tsc` builds `build/` atomically, so
that state cannot arise from `make engine-build`, and the task-reviewer specified the unified
gate. Flagging so it reads as a decision, not an omission.

## Plan

**Didn't change:** `eval/harness/validators/**` (not run by pytest, so a marker there is inert
and deleting its own guard would make asserts pass vacuously); `apps/server/tests/`; the
post-checkout hook (symlinking `build/` into worktrees would silently test main's compiled
output); and `tests/unit/conftest.py` (#1240's autouse auth stub). Behavior tightening: the
mock_mcp tests previously skipped when a specific tool was missing; they now skip only when the
whole catalog is empty, so a build-present-but-tool-missing case fails loudly.

**Acceptance check:** with the build moved aside, `cd eval/harness && uv run --frozen pytest -q`
exits 0, skips exactly the six marked tests, and prints
`6 tests skipped: engine build absent; run make engine-build`; with the build restored, all six
run and no summary line appears. Verified by running (`mv build`), not inspection: build-absent
gave `1593 passed, 6 skipped`, exit 0, summary present; build-present gave `1599 passed`, no
summary. `make harness-test` cannot show this (its `$(ENGINE_BUILD)` prerequisite rebuilds the
thing being removed).

Step-4 stop rule: not hit. Test-infra only; no schema, auth, plugin-agent binding, or ADR change.

## Test plan

- [x] `make test-all` passed in a single clean run (exit 0): typecheck, JS 248, control-plane
      pytest 135, MCP engine vitest 2046, eval app 148, eval harness 1599 (build present, so the
      6 marker tests ran).
- [x] No skill run-log snapshot changed (touches only `eval/harness/`), so no `make eval-skill`
      run is needed. `check_runlogs.py` marks no skill touched: no eval cost.

Nothing in CI exercises the new skip path or the summary line: `eval-harness-tests.yml` and
every `make` target build the engine first, so both are dark until a fresh worktree hits them.
Called out rather than implied.

## Follow-on issues

None filed. One optional item the task-reviewer flagged and I did not build: a pytester test of
the hook, so a later refactor cannot silently delete it. CI never exercises the skip path, so
there is a real argument for it, but testing the real conftest hook via pytester means
replicating it in an isolated dir (testing a copy, not the real hook), so its value is limited
and the specified acceptance is the manual `mv build` check. Happy to add it, or file it, if you
want it tracked.

Closes #1265.

### DIFF
--- a/eval/harness/pyproject.toml
+++ b/eval/harness/pyproject.toml
@@ -49,3 +49,6 @@ testpaths = ["tests"]
 python_files = ["test_*.py"]
 python_functions = ["test_*"]
 asyncio_mode = "auto"
+markers = [
+    "requires_engine_build: needs the compiled packages/engine/mcp-server/build/ (mock_mcp catalog + ts_validator); run make engine-build",
+]
--- a/eval/harness/tests/conftest.py
+++ b/eval/harness/tests/conftest.py
@@ -1,0 +1,78 @@
+"""Session hooks for the harness suite, incl. the ``requires_engine_build`` marker.
+
+This conftest lives at ``tests/`` (not ``tests/unit/``) so the marker and its hooks
+cover every test under ``testpaths = ["tests"]``, and to stay clear of
+``tests/unit/conftest.py`` -- which already exists (the autouse auth stub, #1240). The
+``pytest_terminal_summary`` hook below writes its own line regardless of verbosity, so
+it shows even under ``-q``.
+
+The ``requires_engine_build`` marker
+------------------------------------
+Six tests need the compiled ``packages/engine/mcp-server/build/`` (mock_mcp's tool
+catalog and the compiled TS validator). That build is a Makefile prerequisite of
+every ``make`` target and is built in CI, but a bare ``uv run pytest`` in a fresh
+worktree has none: the post-checkout hook links node_modules and the ``.env`` files,
+not ``build/``. Before this marker each test re-probed the dependency by hand; one
+site was forgotten for months (a red that was not a regression, #1265), and the fix
+for that made the skips *dark* -- ``pytest -q`` printed ``N skipped`` and moved on. A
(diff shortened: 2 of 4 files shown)


taiwo-stack (approved):
consolidates build-dependent skips into a single requires_engine_build marker with a behavioral, session-cached probe and clear terminal summary; low-risk test-infra change.

PR #1372, opened 2026-08-06
eval: backfill stale run logs for search-records + init-project (#1140)

Cia-3:
- Backfills two of #1140's 16 stale run logs — the two the issue already
  decided are needed regardless of how the other 14 (fixture-only drift
  from a shared `mid-research-flynn` fixture edit) get scoped.
- `search-records`: 24/24 pass. First run hit two transient infra flakes
  (a judge API connection error on one test, an SDK stream stall on
  another) — neither was a skill defect; a clean rerun resolved both.
- `init-project`: 9/9 pass.
- Genealogist annotation complete on both (152 + 62 correction entries).

## Test plan
- [x] `make eval-skill SKILL="search-records init-project"` — both active,
      snapshot matches working tree
- [x] Every dimension reviewed in the CRUD UI

Closes nothing — #1140 stays open for the remaining 14, pending the
senior call on whether `eval-cosmetic-skip` applies to them.
Note the last line deliberately does not say Closes #1140 — that issue covers 16 skills and this PR only clears 2, so closing it would drop the other 14 off the board.


mercyokum (approved):
Looks good!

PR #1373, opened 2026-08-06
Add a bounded timeout to every fetch() call against FamilySearch/Wikipedia/OpenRouter

EdmondOware:
## Summary
- None of the 24 files that call the global `fetch()` against FamilySearch, the wiki-query-api sidecar, Wikipedia, or OpenRouter passed a timeout/AbortSignal. Node's `fetch` never times out on its own, so a stalled upstream connection hangs a tool call forever with no recovery — confirmed live: `volume_search` hung for 236 minutes in a Cowork research session.
- Added `fetchWithTimeout()` in `src/utils/http.ts` (`AbortSignal.timeout`, 30s default, 90s override for `image_transcribe`'s OCR call and for `fs-image-fetch.ts`'s multi-MB image body read) and switched all 32 raw `fetch()` call sites across the 24 files to use it.
- Added `tests/utils/http.test.ts` (happy path, stalled-connection timeout, genuine-network-error passthrough) and `tests/packaging/no-bare-fetch.test.ts` (fails CI if a future `fetch()` reintroduces the hang outside `utils/http.ts`).
- Documented `fetchWithTimeout` in `CLAUDE.md`'s "Code reuse" section.

Closes #1369.

## Review fixes (per @promise-emmanuel)
- **MAJOR**: `fs-image-fetch.ts`'s `response.arrayBuffer()` ran outside `fetchWithTimeout`'s own try/catch, so a mid-stream abort surfaced as a raw `TimeoutError` instead of the friendly message, and the default 30s didn't fit a multi-MB download anyway. Gave it a 90s budget (matching the OCR call it feeds) and wrapped the body read to translate a timeout into the same message pattern used elsewhere.
- Added `no-bare-fetch.test.ts` so this can't silently regress.
- Noted the deliberate `signal` override in `http.ts` (no caller passes its own signal today; the timeout must always win if one ever does).
- Fixed `place-population.ts`'s bare `catch` to interpolate the actual error instead of discarding it.
- Corrected the call-site count above (31 → 32).
- Resynced this branch with `main` (was 76 commits behind).

## Second review round (@DallanQ)
Re-synced with `main` again — it had moved on 63 commits, and two of them relocated
fetch call sites this branch had already converted. A plain merge would have left
`src/` with two bare `fetch()` calls, which this branch's own `no-bare-fetch.test.ts`
fails on. Both are now carried onto main's newer structure:

- `fs-image-fetch.ts` — `main` split the image download into a new
  `attemptFsImageFetch` helper with a fallback-URL retry, and that helper had **no
  timeout at all**. Keeping main's version alone would have restored the unbounded
  hang for `image_read` / `image_transcribe`. Both halves of the earlier fix moved
  onto it: the 90s budget and the body-read `TimeoutError` translation.
- `record-search.ts` — `main`'s `fetchSearchWithRetry` (#1316) already had a 25s
  per-attempt timeout and a 3× retry, so the retry and its "transient failure, NOT
  an empty result" error are kept exactly as `main` has them. Only the raw `fetch`
  plus inline `AbortSignal.timeout` is routed through
  `fetchWithTimeout(url, init, SEARCH_TIMEOUT_MS)`.

Also finished the half of the timeout the first round only fixed for images. The
timeout is absolute from when it is created, so it can fire after the headers
arrive while the body is still streaming — and that rejection comes out of
`response.json()` / `.text()` / `.arrayBuffer()` at the call site, after
`fetchWithTimeout` has already returned. There are 40 such reads across 24 files
and none of them handled it, so the caller got a bare "The operation was aborted
due to timeout" naming neither the tool nor the URL. `collections_search` is the
most exposed: one `count=5000` request whose whole payload has to stream inside
the 30s default.

`fetchWithTimeout` now shadows the response's body readers, so a mid-stream abort
surfaces with the same readable message the fetch itself produces — one place
instead of forty, and it lets `fs-image-fetch.ts` drop the hand-rolled copy of
exactly this logic. Confirmed the two new assertions fail with the guard disabled
and pass with it restored.

## Test plan
- [x] `tsc --noEmit` clean
- [x] `npm run build` succeeds
- [x] Rebuilt `releases/genealogy-mcp.mcpb` and confirmed the reinstalled Claude Desktop extension's compiled `build/` output contains the fix
- [x] Manually reintroduced a bare `fetch()` and confirmed `no-bare-fetch.test.ts` fails, then reverted

Re-run on the merge with `main` (all green — the doc-links failure noted in the
earlier revision of this plan no longer reproduces):
- [x] engine vitest — 2104 passed, 88 files (2100 before the new body-read tests)
- [x] eval harness — 1843 passed
- [x] `apps/server` — 143 passed, 1 skipped
- [x] `turbo test typecheck` — 7/7


promise-emmanuel (commented):
## Review — full pass, test-merged against `main` (76 commits behind)

Coverage is complete and I checked it both ways: `grep -rn '\bfetch(' src/`
excluding `fetchWithTimeout(` and `utils/http.ts` returns **zero hits**, before
*and* after merging main's 76 commits. 32 call sites, 24 files. (Body says 31.)

The new test is real coverage — I replaced line 16 with
`return await fetch(url, { ...init });` and watched it fail, then restored:

```
× throws a clear timeout error when the upstream connection stalls  5026ms
Tests  1 failed | 2 passed (3)
```

**Verified on the test-merged state** (merge was clean): engine `tsc` exit 0 ·
engine vitest **2059 passed** (86 files) · `npm run build` exit 0 ·
`eval/harness` **1599 passed** · `apps/server` **135 passed, 1 skipped** ·
`turbo test typecheck` — `web#typecheck` clean after `pnpm install`.

---

### 🔶 MAJOR (8/10) — `src/utils/fs-image-fetch.ts:99` + `:120`
**The 30s budget covers the multi-MB image download, and a body-read abort
escapes the wrapper's friendly message.**

```ts
 99:  const response = await fetchWithTimeout(url, {
120:  const buffer = await response.arrayBuffer();
```

`AbortSignal.timeout` is absolute from creation — it keeps running after headers
arrive and undici errors the in-flight body stream. Since `arrayBuffer()` is
called *outside* `fetchWithTimeout`, its rejection never reaches the `catch`.
Proved with a local server that sends headers then dribbles the body:

```
headers OK at 49ms, status=200
THREW at 353ms   err.name = TimeoutError
  err.message = The operation was aborted due to timeout
  friendly msg matched? false
```

This path is *designed* for large payloads — the docstring at line 93 says
"Imposes no size cap", and `image-transcribe-tool-spec.md:285` says "no size cap
on the fetched image … A pathological multi-MB scan". The PR gave the OCR call
90s (`image-transcribe.ts:17`) but left the multi-MB download that feeds it at
30s. 5 MB at 200 KB/s is 25s, so `image_read`/`image_transcribe` can now fail
where they previously succeeded, with an unhelpful error.

**Fix:** add `const IMAGE_FETCH_TIMEOUT_MS = 90_000;` and pass it as the third
arg at line 99; optionally wrap line 120 to rethrow `TimeoutError` with the same
readable message.

### ℹ️ (9/10) Nothing stops call site #33 from being a bare `fetch()`
No lint or test enforces this — I checked all 17 workflows and the packaging
tests. This repo enforces exactly this class of rule mechanically elsewhere
(`agent-tool-names.test.ts`, `manifest.test.ts`), and CLAUDE.md carries the
parallel `BROWSER_USER_AGENT` rule. A future raw `fetch()` reintroduces the
236-minute hang and passes CI. **Fix:** a test that greps `src/` for `fetch(`
outside `utils/http.ts` — and watch it fail before landing it.

### ℹ️ (8/10) `src/utils/http.ts` is documented nowhere
`grep -rn 'utils/http\|fetchWithTimeout' CLAUDE.md DEVELOPMENT.md docs/architecture.md docs/specs/*.md`
→ no hits. CLAUDE.md § "Code reuse → Where to look first" has six entries and
exists to stop the next agent hardcoding what it should import; a new mandatory
helper for all 24 network files belongs there, after the `src/constants.ts` bullet.

### 🔹 (6/10) `src/utils/http.ts:16` — caller-supplied `AbortSignal` silently dropped
```ts
return await fetch(url, { ...init, signal: AbortSignal.timeout(timeoutMs) });
```
`init` spreads first, so `signal` always wins. Latent — `grep -rn 'signal:' src/`
outside `http.ts` returns nothing. Worth one comment line saying the override is
deliberate. `AbortSignal.any` would be correct but adds an untested path with no
caller.

### 🔹 (5/10) `src/tools/place-population.ts:39` — bare `catch` discards the timeout message
```ts
  } catch {
    throw new Error(
      "Population data service is unavailable. Is the Pop Stats API running?"
```
A timeout now sends the user to check a service that is running fine. The only
site that loses the diagnostic — the other seven all interpolate `${cause}`.

---

**Worth knowing:** the real ceiling is attempts × 30s, not 30s —
`person-quality.ts:33` (5 attempts) → 150s, `volume-search.ts:95` → 90s,
`place-resolver.ts:108` `withRetry` → 90s. Still bounded, just not 30s.

**Not checked:** no live-network verification — the timeout claims rest on the
mocked suite plus my local probe, not a real FamilySearch/OpenRouter stall, so
the MAJOR finding's *trigger* is unproven (hence 8/10, not 10). I did not
package or install the `.mcpb`. Separately, `apps/electron`
`feedback.test.ts:152` failed twice under `turbo` at ~5.4s against a 5s timeout
and passed 61/61 run directly — it zips 45 MB of random bytes and imports
nothing from `mcp-server`, and this PR changes zero files under `apps/`, so it's
an unrelated pre-existing flake that deserves its own issue.

Issue #1380, opened 2026-08-06
Nothing validates eval/tests/unit/** against unit-test.schema.json on a PR


Gennecis:
> **Reviewed 2026-08-20 before junior handoff.** No decision needed, but the
> body's proposed fix rests on a false premise — corrected below. The original
> follows under "Original issue"; its diagnosis is right, its remedy is not.

**Touches:** eval/harness/tests/unit/test_loader.py, eval/harness/scripts/check_negative_reciprocity.py, docs/architecture.md

## The fix, and why not the two alternatives the body names

Add a corpus walk to the eval harness's own pytest suite that calls
`harness.loader.load_test` on every committed test file. **That function already does
exactly this validation**, so the check is a *call*, not a reimplementation, and cannot
drift from the schema's semantics.

Both options in the original body are worse, and one is factually wrong:

- **Vendoring a stdlib validator** would re-implement `allOf` / `if` / `then` / `const` /
  `oneOf` — the schema uses all of them (the `xfail_reason` conditional, the
  positive-vs-negative `negative` block, three `oneOf` branches under
  `expected_classifications`, `execution.stub_skills.items`). A subset validator passes
  on files it does not understand, which is the failure `CLAUDE.md` § "A new lint must be
  proven to fail" exists to stop.
- **Moving it into the engine's vitest suite** was proposed on the premise that the
  engine "already has a JSON Schema validator available." **It does not.** `ajv` is
  present only transitively under `node_modules/` and is declared in neither
  `dependencies` nor `devDependencies` of `packages/engine/mcp-server/package.json`. No
  `Ajv` import exists anywhere in `src/` or `tests/`.

`jsonschema>=4.22.0` is already a declared dependency of the harness, and
`.github/workflows/eval-harness-tests.yml` already runs `uv sync --frozen` + pytest on
any PR touching `^eval/tests/` — so this gate fires on exactly the PRs it exists to
catch, at no extra CI cost, and needs no workflow edit.

## Do this

1. **The check.** In `eval/harness/tests/unit/test_loader.py` (which already resolves
   `REPO_ROOT` and reads a real corpus file), or a new
   `eval/harness/tests/unit/test_unit_test_corpus.py`:

   ```python
   def test_every_committed_unit_test_validates_against_the_schema():
       paths = sorted((REPO_ROOT / "eval/tests/unit").rglob("*.json"))
       # Floor, not equality: 396 files measured 2026-08-20. A count below this means
[...]

Issue #1379, opened 2026-08-06
An xfail negative test counts as a valid reciprocal in check_negative_reciprocity.py


Gennecis:
> **Reviewed 2026-09-21 before junior handoff.** No decision needed — the 2026-09-18
> ruling stands. The supporting evidence below was stale and is replaced; the blast
> radius grew from one file to three.

## Decided (lead, 2026-09-18): A. An xfail negative does not satisfy the reverse edge

An `expected_outcome: "xfail"` negative declares its own edge as known-broken and cannot count as the reciprocal for the reverse one. Reword the reciprocity rule in `docs/specs/unit-test-spec.md` (the reciprocity section) to say so; `check_negative_reciprocity.py` follows the spec. Land the stale-docstring fix at `:47` in the same PR. Prove the lint both ways before committing: an xfail-backed reverse edge must red, a genuine negative reciprocal must still pass.

Beat B (keep today's behaviour), which lets a test asserting the routing is broken count as pinning the pair from both sides, the false green this lint exists to prevent. Nothing is masked today (97 negatives, 6 xfail, 0 overlap on 2026-09-16), so this is junior and unhurried.

**Touches:** eval/harness/scripts/check_negative_reciprocity.py, eval/harness/tests/unit/test_check_negative_reciprocity.py, docs/specs/unit-test-spec.md

`check_negative_reciprocity.py` accepts any negative test as the reciprocal for the reverse edge, including one marked `expected_outcome: "xfail"`. `routing_edges()` never reads `expected_outcome`, so an xfail negative both declares its own edge and satisfies the reverse one (`eval/harness/scripts/check_negative_reciprocity.py:134-185` builds the map, `:196-207` uses the same map for both halves).

An xfail asserts the routing is *known-broken*. Treating it as a reciprocal claims the pair is pinned from both sides when the reverse direction grades nothing — the false green the lint exists to prevent.

[...]

Issue #1378, opened 2026-08-06
check-runlogs.yml: the four warn-only lint steps are skipped whenever a blocking step fails


Gennecis:
> **Reviewed 2026-08-20 before junior handoff.** Premise confirmed live; the exact
> YAML, the instrument, and the one way to get this wrong are below. The original
> body follows unchanged.

## Confirmed, and how often it bites

The eight most recent failing runs of this job all failed at `Run runlog discipline
checks` with five later steps `skipped` — the four lints plus `Post Set up Python`.

```sh
gh api "repos/PioneerAIAcademy/cowork-genealogy/actions/workflows/check-runlogs.yml/runs?status=failure&created=%3E2026-08-06&per_page=1" --jq '.total_count'
gh api "repos/PioneerAIAcademy/cowork-genealogy/actions/workflows/check-runlogs.yml/runs?created=%3E2026-08-06&per_page=1" --jq '.total_count'
gh api repos/PioneerAIAcademy/cowork-genealogy/actions/runs/<run-id>/jobs --jq '.jobs[].steps[] | {name, conclusion}'
```

Measured 2026-08-20: **159 failing runs against 990 total** in that window — about
one run in six suppresses all four lints. Re-run the commands rather than quoting
the numbers forward; the window moves.

## The edit — four `if:` lines, one file

Each of the four steps after `Run runlog discipline checks` becomes:

```yaml
        if: ${{ !cancelled() && steps.scope.outputs.relevant == 'true' }}
```

**Keep the `${{ … }}`.** A bare leading `!` is a YAML tag indicator, so
`if: !cancelled() && …` makes the whole workflow fail to parse (`did not find
expected alphabetic or numeric character while scanning an anchor`) — verified.
The original body's suggested form has this bug. `!cancelled()` rather than
`always()`: identical behaviour when the gate fails, and it does not keep the job
running through a cancellation.

Put this comment above the first of the four, and nothing longer:

```yaml
      # `!cancelled()` here and on every step below: an `if:` with no
      # status-check function carries an implicit `success()`, so a failure in
      # the gate above skipped all four lints. A step added below inherits this
      # — check_negative_reciprocity.py was added without it.
```

Leave the other five steps exactly as they are.

## The one way to get this wrong

Do **not** reach for `continue-on-error: true` on `Run runlog discipline checks` to
make the later steps run. That converts the repo's blocking run-log gate into a
warning: every PR with a stale run log then merges green and nothing else reports
[...]

Issue #1375, opened 2026-08-06
Promote the negative-reciprocity lint to blocking, comparing the edge set


Gennecis:
**Touches:** eval/harness/scripts/check_negative_reciprocity.py, .github/workflows/check-runlogs.yml

`check_negative_reciprocity.py` ships warn-only with no baseline file and no count threshold: 45 of the corpus's 79 routing edges are one-directional and it exits 0 anyway.

Promotion to blocking is gated on a triage that has not happened — deciding which of the 45 are deliberate one-directional near-misses (a request shape only plausible in one direction) versus simply not-yet-written. Calibrating a gate ahead of that data is what the sibling lint's docstring warns against for its own 68 hits.

When it is promoted it must compare the edge **set**, never its size. A count threshold stays green when one edge is removed and another added, which is the failure mode it would exist to prevent.

Icebox: no decision sits behind this yet, and it cannot start until the triage does.

DallanQ:
Absorbed into issue #2480 — Six committed guards read as coverage on the board and structurally cannot go red.

This card's content moved there in full, under the heading "issue #1375 — Promote the negative-reciprocity lint to blocking, comparing the edge set" (in the issue body). Its measured figures with their denominators, file and symbol references, any rulings recorded in comments here, and its acceptance check went with it, and its `**Touches:**` paths are merged into that issue's Touches line.

Not planned rather than completed: no work was done under this number. The gap it names is unchanged, and the register still carries it under issue #2480.

Applied by the nothing-checks consolidation audit, 2026-09-11.

PR #1381, opened 2026-08-06
eval-harness: warn when a negative routing test has no reciprocal

Gennecis:
Closes #945 

## Summary

- Adds `check_negative_reciprocity.py`, a warn-only CI lint that reports negative routing edges pinned from one side only — **45 of the corpus's 79** today. Closes #945.
- Enforcement for a rule that already exists (`unit-test-spec.md` § 6, "For each confusable pair, create tests from both directions"), not new doctrine.
- Ships with **no baseline file and no count threshold**, both rejected deliberately rather than deferred. The reasoning lands at `unit-test-spec.md` § 6 and `eval/CLAUDE.md`; the gap itself lands in `architecture.md` §9.4, since a check that never blocks belongs in the "read this before you trust a green CI run" list.
- Wires the step into `.github/workflows/check-runlogs.yml` (appended after the frontmatter lint, same `relevant == 'true'` guard as its siblings), adds the both-sides checkbox to the PR template, and corrects `eval/README.md`'s stale "Two warn-only checks" to four — it already omitted `check_rubric_tool_drift.py` before this lint existed — alongside a full `scripts/` inventory.
- Tier: **Normal**.

## Review guidance

**Start here:** `eval/harness/scripts/check_negative_reciprocity.py`, the three edge-definition rules in `routing_edges()` (lines ~150–183). Each is a place a reimplementation would silently diverge, and one of them changed the headline number: a skill naming **itself** in `correct_skill` is not an edge (line 181). Four tests do this deliberately — `ut_citation_012` accepts `["record-extraction","citation"]`, `ut_convert_dates_010` accepts `["convert-dates"]`, both meaning "the skill under test handling this is also acceptable". `A → A` has no reciprocal to write and would trivially satisfy its own check, so admitting it inflated the total. The corpus has **79** routing edges, not 83. The flagged set is identical either way — only the denominator moved.

Second thing worth your eye: the source is `test.skill` on **both** halves of the comparison (line 159), with no fallback to the parent directory. Sourcing the edge from the field while looking the reciprocal up by directory is what makes a divergent test read as deliberate rather than broken. `test_missing_skill_field_contributes_no_edge` pins it.

**Unsure about:** the annotation volume. The full report is what #945's decision specifies and I'm not relitigating it — hiding the backlog would be the "suppressing, not triaging" failure the sibling lint's own docstring warns about. What I can't judge from here is whether it survives contact with reviewers: this step group now emits **186 standing annotations** (`check_tool_coverage.py` 73 + `check_rubric_tool_drift.py` 68 + this 45) against a per-step rendering cap of 10. The blast radius is also wider than "eval PRs" — the workflow's scope patterns (`check-runlogs.yml:104`) include `^docs/`, `^CLAUDE\.md$`, `^DEVELOPMENT\.md$` and `^README\.md$`, so a docs-only typo fix trips it. If that reads as too much noise, #1375 is where scoping would go.

**Didn't change:** No file under `eval/tests/unit/` — not one. `check_runlogs.py:437` marks a skill touched on any change there, and rule 2 then demands an active run log, so a single edited test file would cost a paid `make eval-skill` run plus a genealogist annotation. Nothing in this PR marks any skill touched.

Also left alone: the 45 unbacked edges (backfilling is #1374, the triage behind it #1375); `check_rubric_tool_drift.py`'s 68 untriaged hits (#1003); and `architecture.md`'s pointer to `unit-test-spec-v2.md` — a deferred-features plan, not the live spec — sitting seven lines above an edit I did make (#1376), kept out to keep this diff scoped.

**Deviated from the plan:** four. `PLAN.md` is gitignored, so this list is the only route they take to you.

1. Added `architecture.md` §9.4 as a seventh edit site. §9.2 lists the lint among checks that *run*; advertising a check that never blocks without recording that it never blocks implies coverage this does not give.
2. Corrected the published count 83 → 79 (the self-reference rule above).
3. The warning cites the rule as § 6 plus the quoted sentence rather than as `unit-test-spec.md:653`, which the issue's decision asked for. The line number rots — #1225 already shifts it by +39. I argued this as the stronger reading of the instruction, but it is your call, not mine, so it is flagged rather than assumed.
4. Documented rather than fixed `expected_outcome: "xfail"` reciprocals (#1379). An xfail asserts the routing is known-broken, so it pins nothing — but it masks nothing today either: both xfail edges are themselves already reported as one-directional. Narrowing what counts as a reciprocal is a semantic change to the § 6 rule this script only enforces.

**One heads-up on #1251:** it rewrites `docs/architecture.md`'s `§3), and two **warn-only** lints` line — the same line this PR changes from "two" to "three" — so one of us rebases. Separately, its `architecture.md` diff is a net **+4** issue references against the 27 baseline and its file list contains no `doc-links.test.ts`, so its `engine-test` job should be red on that ratchet independently of this PR.

## Test plan

**Acceptance check:** `cd eval/harness && uv run pytest tests/unit/test_check_negative_reciprocity.py` — 18 tests, fails on `main` (no such module), passes here. The discriminating ones are `test_reciprocity_is_per_pair_not_per_skill` (having *a* negative test does not back a skill's every inbound edge), `test_self_naming_target_yields_no_edge`, and `test_main_returns_zero_on_a_fully_asymmetric_corpus` — the never-blocks property asserted rather than assumed, which is why the builder and `main()` take a `tests_dir` at all. All cases run on `tmp_path` corpora; **no test pins the live corpus's count**, which would be the count-threshold defect in test form.

Real-corpus verification is a **set** comparison, not a count: the 45 reported pairs were diffed against an independently derived list, symmetric difference empty, and re-diffed across the self-reference fix to confirm membership held while the denominator moved.

  **It does not pass, and not because of this branch.** `make test-js` fails on `apps/electron`'s `feedback.test.ts:152` (archive-budget). I reproduced it on a clean `main` with this branch's changes stashed — identical failure — and confirmed it passes standalone (19/19, 3.8s) and per-package (61/61); it fails under turbo's concurrent execution, where the electron run takes ~13–18s against a 5000 ms vitest timeout. Filed as #1377. Every other suite passes: harness pytest **1617**, engine vitest **2048** across 85 files, `make typecheck` 5/5, `make eval-ui-test` **148**, control-plane pytest **135 passed / 1 skipped**.

- [ ] For every skill whose **run-log snapshot** I changed … I ran `make eval-skill SKILL=<name>`.

  **N/A — no skill's snapshot is touched.** Verified against `check_runlogs.py:416-443`: touched-skill detection matches only `RUNLOG_PATH_RE`, `^(?:packages/engine/plugin/skills|eval/tests/unit)/([^/]+)/`, and `AGENT_PATH_RE`. This PR touches none of those, so rules 1–4 gate nothing and no paid run is owed.

  **N/A — this PR adds that checkbox and changes no skill description or DO NOT clause.** Nothing under `packages/engine/plugin/skills/` is touched.

Additionally verified, because this PR edits two files under an exact-match ratchet: `doc-links.test.ts` passes (113), with `docs/architecture.md` at **27/27** and `docs/specs/unit-test-spec.md` at **7/7**. `ISSUE_REF_BASELINE` fails on both `>` and `<` and both files were already at ceiling, so **neither doc edit cites an issue number** — the numbers live in `eval/CLAUDE.md` and `eval/README.md`, which are outside the ratchet's scanned set.

## Follow-on issues

- #1374 — `hypothesis-tracking` has no negative routing tests. It is invisible to this lint (never a source, never a target), so the gap will not surface on its own.
- #1375 — promote this lint to blocking, comparing the edge **set**. `icebox`: gated on the triage.
- #1376 — `architecture.md` sends unit-test authors to `unit-test-spec-v2.md`, a deferred-features plan.
- #1377 — the `feedback.test.ts` flake above. Pre-existing; `make test-all` is red on a loaded machine, which trains people to ignore the command the PR template requires.
- #1378 — the four post-gate lint steps in `check-runlogs.yml` are skipped whenever an earlier blocking step fails, via the implicit `success()` in a bare `if:`. Three are warn-only; `check_skill_frontmatter.py` is itself partly blocking, so a real gate is among the skipped. Pre-existing — this PR's step inherits the pattern rather than introducing it, and it is filed across all four rather than fixed for one in isolation.
- #1379 — an `xfail` negative counts as a valid reciprocal.
- #1380 — nothing validates `eval/tests/unit/**` against `unit-test.schema.json` on a PR. Directly degrades this lint: it skips any test it cannot parse, so the routing graph can under-report silently. `loader.py` catches it, but only on a paid eval run.


jud-sdev (approved):
Reviewed full pass, 0 commits behind main. This is an exemplary PR — every claim in the body checks out under independent verification.

**Verified by running:**
- `test_check_negative_reciprocity.py` — 18 passed.
- Script on the live corpus — **79 edges / 45 flagged / 45 warnings**, exactly the claimed numbers.
- **Never-blocks property holds on the real corpus:** exit `0` with 45 warnings present, not just the synthetic `test_main_returns_zero…`.
- **4 self-naming negative tests** confirmed → the 79-not-83 denominator is correct.
- **Doc-ratchet claim holds:** the diff adds zero `#NNNN` refs to `architecture.md` or `unit-test-spec.md`, so both stay at ceiling and the ratchet isn't tripped.
- Not run: full engine vitest/turbo and the harness's other suites (no pnpm/build in my worktree); the PR's risk surface is a stdlib-only Python lint, exercised directly. The `feedback.test.ts` flake is real and pre-existing (#1377), not this branch.

**On the two design calls you flagged for yourself:** both sound. The self-reference exclusion (`target == source`) is correct and tested, and sourcing both halves from `test.skill` with no directory fallback is what makes a divergent test read as broken rather than deliberate — pinned by `test_missing_skill_field_contributes_no_edge` and `test_divergent_skill_field_…`. Testing the never-blocks property (unlike the sibling lint) is exactly right for a warn-only check.

**One sub-threshold note, non-blocking:** `architecture.md` §9.2 now says "three warn-only lints" while `eval/README.md` says "Four warn-only checks." They count different scopes — three `.py` script lints vs. four including rule 2b (judge-prompt-hash) — and each is internally accurate, but a reader moving between the two docs meets two numbers. Reconcile only if trivial; otherwise fine.

The known weakness — 45 warnings buried under GitHub's 10-annotation cap, so a newly-added edge isn't distinguishable from the backlog — is honestly disclosed in the body and §9.4 and filed as #1375. Warn-only first step with the set-comparison promotion deferred to a triage is the right call.

Approving.

Issue #1367, opened 2026-08-05
Production preflight accepts an empty or trivially weak SESSION_SECRET / WS_SIGNING_KEY


T-FEH:
**Touches:** apps/server/app/config.py, apps/server/tests/test_prod_preflight.py

`config.assert_production_config` compares each secret against its **declared default**:

```python
if getattr(s, field) == Settings.model_fields[field].default:
```

That is deliberate — a copied string literal would silently stop matching the day someone
rewords the default — but it means the gate only recognises the *exact* dev default. A
production deploy with `SESSION_SECRET=""`, `SESSION_SECRET="x"`, or any other short or
low-entropy value boots cleanly.

An empty `session_secret` is arguably worse than the dev default, since the dev default is
at least a fixed unknown string to anyone outside the repo, while an empty one is the first
thing an attacker would try. The same applies to `ws_signing_key`, whose value is the HMAC
master key for every per-sandbox WS handshake token.

**Scope note:** this was surfaced by `/security-review` on the PR for issue #1123 and
deliberately left out of it — that issue's reviewed spec specifies the default-comparison
check, and widening the gate mid-PR would have been drift. It narrows the hole rather than
opening one: before that PR there was no gate at all.

**Direction:** add a minimum-length / entropy floor alongside the default comparison (e.g.
refuse anything under 32 chars when `PUBLIC_URL` is https), with the same collect-then-raise
message shape and a `fly secrets set … "$(openssl rand -hex 32)"` remedy. The existing
`test_prod_preflight.py` parametrised case is the natural place to extend.

Decision needed on the floor itself — a length check is crude, and whether to reject on
low entropy rather than short length is a judgement call worth making once.

DallanQ:
**Ruling (2026-08-20):** closed, not planned.

This came from an automated `/security-review` on the PR for issue #1123, and was filed rather than folded because widening the gate mid-PR would have been drift against that issue's reviewed spec. Reviewed again 2026-08-20: the hole is real and reproduces on `main` — an https deploy boots with `SESSION_SECRET=""` — but it is one loop condition, one parametrised test case and five doc lines, which does not carry the overhead of a tracked issue.

It is also not attacker-reachable: it needs an operator to set a weak value themselves. The one realistic path in is someone pasting the `…` placeholder out of our own deploy instructions, and fixing those placeholders removes it without needing a floor at all.

[...]

Issue #1366, opened 2026-08-05
make test-js flakes on 5s test timeouts under parallel load


T-FEH:
> **Reviewed 2026-08-18 before junior handoff.** No decision needed — the fork was
> already settled in-repo on 2026-08-10; raise the clock in each vitest config, not
> per-test. Original body follows.

**Touches:** packages/viewer-ui/vitest.config.ts, apps/electron/vitest.config.ts,
apps/web/vitest.config.ts, eval/app/vitest.config.ts

`make test-js` fails intermittently on vitest's default 5000ms per-test timeout. Observed
across three separate full runs on 2026-08-05, with a **different** set of tests failing
each time:

- `buildFeedbackZip — size budgets follow the server convention > drops the largest files
  instead of throwing when over the archive budget` (5088ms, 5316ms, 5374ms)
- `TimelinesSection — B5 conflict chip > renders the conflict chip when an event has
  conflict_ids` (5349ms, 5578ms)
- `SidecarPanel — scroll preservation across watcher refresh > keeps the same drawer DOM
  node across loaded -> loaded payload swap` (6166ms)

Every one overruns by a few hundred ms to ~1s, never by a lot — the signature of load,
not of a hang. Run in isolation (`pnpm vitest run <file>`), the electron file passes
19/19 twice in a row.

**This is not caused by any one change.** Confirmed by stashing an unrelated branch's
work (`git stash push -u`) and re-running `make test-js` on a clean tree: the same
timeouts reproduce. `make test-all` runs the workspace suites through turbo in parallel,
so several vitest projects compete for cores.

**Why it matters:** `make test-all` is what the PR template requires, and `scripts/test.sh`
deliberately reports every suite so a developer sees the whole picture. A gate that goes
red for reasons unrelated to the diff trains people to re-run until green, which is how a
real failure gets waved through.

**Possible directions** (not a decision): raise `testTimeout` for the affected projects,
mark the three slow tests explicitly with a per-test timeout, or cap turbo's concurrency
for the `test` task. Worth measuring which of the three tests is genuinely slow versus
merely unlucky before picking.

Found while verifying the PR for issue #1123.

---

# Merged in from issue #1597 (2026-08-13): measurement for the electron test, and two more options

Filed independently off PR #1598 as "`make test-all` is red locally: electron
feedback-zip test times out at 5000ms under the parallel gate". Same test, same
5000ms default, same contention cause — this is the measured version of the first
bullet above.

[...]

Issue #1365, opened 2026-08-05
Nothing proves the production boot refusal is reachable or legible


T-FEH:
**Touches:** apps/server/app/config.py, apps/server/tests/test_prod_preflight.py, deploy/fly.toml, docs/architecture.md

`config.assert_production_config` refuses to boot the control plane when `PUBLIC_URL`
is https and `session_secret` / `ws_signing_key` are still their declared defaults, or
`DATABASE_URL` is unset. `apps/server/tests/test_prod_preflight.py` asserts the function
and its lifespan wiring offline — including that a `RuntimeError` raised in the lifespan
propagates out of `with TestClient(app)`.

What nothing checks is the production half: that the refusal is actually **reachable**
under uvicorn as `deploy/Dockerfile` runs it, and that the message is **legible** in Fly
logs rather than buried in a traceback. A gate that never fires, or fires unreadably,
looks identical to a correctly configured deploy — which is the exact failure mode the
gate exists to prevent.

Recorded as a row in `docs/architecture.md` §9.4 ("What nothing checks"), per that
section's rule that a new gap is filed **and** rowed.

**Unblock condition / trigger:** the next time someone has a throwaway Fly app or a
staging target to deploy against. This needs a real `fly deploy` with a secret
deliberately unset — it cannot be done offline, which is why it is iceboxed rather than
ranked. A cheaper partial: assert the refusal message reaches stderr when the app is
started by uvicorn the way the Dockerfile does, which would cover legibility without a
deploy.

Filed from the PR for issue #1123.

---

## Trigger, split 2026-08-12

**Reachability** still waits on a deploy target. Verified 2026-08-12: `deploy/` holds one
`fly.toml` (the alpha app); there is no staging or throwaway target.

**Legibility does not, and needs no decision.** Assert that the refusal message reaches
stderr when the app is started by uvicorn the way `deploy/Dockerfile` starts it. That is
offline and cheap, and it closes the "fires unreadably" half of what nothing checks here.
Do that half whenever someone is next in this file; the reachability half stays parked.

DallanQ:
Absorbed into issue #2488 — Cowork tool binding: the register gap that has bitten three times is split across cards and named by none.

[...]

Issue #1127, opened 2026-08-02
Move init_db() to a Fly release_command before `fly scale count > 1`


DallanQ:
> **Reviewed 2026-09-18 before junior handoff.** Decision recorded below; the
> original body follows under "Original issue".

## Decision needed — Race fix

The plan of record (`docs/specs/public-rest-api-spec.md:96`, `docs/plan/neon-postgres-plan.md:231`) says move the bootstrap to a one-time Fly `release_command`, but the same race can be removed inside `init_db()` with no deploy-config change at all — which mechanism do we commit to, given that `docs/plan/fly-deploy-plan.md:286-290` names the production target as AWS behind a non-sticky load balancer where a Fly `release_command` does not exist?

- **B — Race-tolerant `init_db()`** — Touches one Python file plus tests and the wording sites; verifiable by `make server-test` **plus a local throwaway Postgres**, a method `DEVELOPMENT.md` already documents, so a junior can land it and know it worked. Survives the AWS move unchanged. Cost: reverses the written plan of record in two documents, and leaves the boot path doing schema work forever.  *(recommended)*
- **A — Fly `release_command`, seed stays on boot** — Follows the plan of record and the Fly idiom, and takes schema DDL off the request-serving image. Cost: the only instrument is a live `make deploy` against the single production app (no staging), a broken `release_command` blocks every subsequent deploy, and the work is redone at the AWS move. This option expands the blast radius — it adds `deploy/fly.toml`'s `[deploy]` block and a new module-level entry point, and makes the PR unverifiable without a production deploy, so it stays a `senior` card.

<details><summary>Body text if B</summary>

**Decision (2026-09-18):** fix the race **inside `init_db()`**; the one-time Fly
`release_command` named in `public-rest-api-spec.md` and `neon-postgres-plan.md`
is **not** taken — it is Fly-specific, and `fly-deploy-plan.md` § "Horizontal-scaling
caveat" names AWS-behind-a-non-sticky-LB as the production target, where it does not
exist. `init_db()` stays on the boot path and becomes safe to run concurrently:

- `SQLModel.metadata.create_all()` — tolerate a concurrent `CREATE TABLE` (catch the
  duplicate-table error and re-check, rather than trusting `checkfirst`'s
  check-then-act).
- the allowlist seed — make the per-email `INSERT` conflict-tolerant (a nested
  `begin_nested()` savepoint per email, or the dialect `insert(...).on_conflict_do_nothing()`,
[...]

PR #1382, opened 2026-08-06
Refuse to boot in prod if ws_signing_key/session_secret is still the dev default (#1123)

T-FEH:
## Summary

- **Normal tier.** `config.py` defaulted `session_secret` to `dev-insecure-secret-change-me` and `ws_signing_key` to `dev-ws-signing-key-change-me`, and nothing checked either at startup. A deploy that forgot `fly secrets set` came up silently minting forgeable login cookies, a forgeable FamilySearch OAuth `state` (the same secret signs both — `fs_oauth.py:52`, which the issue body didn't name), and forgeable per-sandbox WS handshake tokens. Only the deploy runbook enforced it.
- Adds `assert_production_config()`, called as the **first** statement of `main.py`'s lifespan. When `PUBLIC_URL` is https it refuses to boot on either dev-default secret, or on an unset `DATABASE_URL` — since the Neon migration removed the Fly volume, that last one is SQLite on ephemeral rootfs, losing every user, session and allowlist row on the next machine restart.
- Six docs asserted the gap this closes; all six updated.

Closes #1123.

## Review guidance

**Start here:** `apps/server/app/config.py` — `assert_production_config`, and the one-line call site at `main.py:33`. Three decisions worth checking: (1) the production discriminant is an https `PUBLIC_URL`, reusing the signal `auth.cookie_secure()` already uses, **not** `sandbox_provider == "e2b"` (which would break `make server-e2b` locally); (2) it compares against `Settings.model_fields[...].default` rather than a copied literal, so it can't silently stop matching when someone rewords a default — the tradeoff is a latent trap if either field is ever redeclared with `default_factory`, noted inline; (3) it collects every offender into one message so a bad deploy is fixed in one cycle, not three.

Second thing to look at: `apps/server/tests/test_sessions.py`. `test_dev_login_refused_when_deployed` flipped `public_url` to https *before* `with TestClient(app)`, and that context manager runs the lifespan — so any new lifespan gate breaks it, and a conftest env pin can't fix a `monkeypatch.setattr` on the already-cached settings object. The flip moved inside the started client; both endpoints it asserts read `get_settings()` per request, so it still tests what its name says. It was the only such test in the suite (four repo-wide hits for `setattr(get_settings()`; the other three touch fields this gate doesn't read).

**Unsure about:** what a real `fly deploy` does with the refusal. The issue's review notes said a failed boot "leaves the running machine up rather than taking prod down" — the artifacts don't support that. This app is a single machine (`fly.toml:49-51`) deployed `--ha=false` (`Makefile:655`, `DEVELOPMENT.md:500`), so it's updated in place, and a boot refusal plausibly means downtime until rollback. Nothing in this repo verifies it either way, so I've stated only the app-level fact everywhere (the process exits at startup) and filed #1365 rather than asserting Fly's behaviour. What limits the real risk is that all three settings are already set on the live app, so the gate can't fire on the next deploy.

## Plan

**Didn't change:** how any secret is read, derived, or used — `auth.py`, `fs_oauth.py`, `ws_token.py`, `sandbox/e2b.py` are untouched; this adds a refusal and nothing else. Not a pydantic validator (`main.py:43` calls `get_settings()` at module scope, so a validator fires at *import* and surfaces as a traceback, not a boot refusal). `ANTHROPIC_API_KEY` / `E2B_API_KEY` are deliberately out of the gate — they already fail loudly on the first session create, and the gate is for silent failures only. `init_db()` stays where it is; moving it is #1127, which edits the same `lifespan`.

**Acceptance check:** `apps/server/tests/test_prod_preflight.py::test_lifespan_refuses_to_boot`. Fails on `main` and passes here — and, unlike the unit cases, it fails if the one-line `main.py` wiring is missing. Proved by deleting that line and re-running: six cases still passed, only this one went red. It's also the only test in the suite that enters a lifespan with an https `PUBLIC_URL`.

**Deviated from the plan:** nothing structural. Two corrections made mid-flight and recorded in `PLAN.md`: the plan initially named a unit case as the acceptance check (it would have passed with the wiring missing), and it initially repeated the issue's Fly-rollback claim, which the artifacts don't support (see "Unsure about").

No step-4 stop rule was hit. This touches credential *configuration*, not `src/auth/` or the engine's token plumbing, and changes no schema or plugin-agent binding.

## Test plan

  With one caveat, reported rather than hidden: `test-js` flakes on vitest's 5000ms per-test timeout under parallel load — three *different* tests across `viewer-ui` and `electron`, varying per run, each overrunning by a few hundred ms. **Pre-existing, not from this branch**: I reproduced it on a clean tree with this work stashed, and the electron file passes 19/19 twice in isolation. Filed as #1366. Everything else is green: server 142 passed / 1 skipped, engine 2046 passed, harness 1599 passed, typecheck clean.

- [x] No skill, rubric, agent body, unit test or fixture is touched, so no `make eval-skill` re-run and no run-log annotation applies.

Also ran `/security-review` (no findings at confidence ≥8 — it verified the refusal message interpolates only `public_url` and never the secret values, that the check runs before `setup_logging()` so it never reaches the structured-logging path, and that the refusal genuinely fails closed under uvicorn as the Dockerfile runs it) and `/check-drift` (one real finding, fixed: I'd written the unsupported Fly containment claim into two places).

The second commit closes six findings from a fresh-context `/code-review`, none blocking, two of them found by mutation-testing the suite — including a test of mine that asserted the refusal names the offending setting but not that it leaves correct ones alone, which passed green against a deliberately over-reporting implementation.

## Follow-on issues

- #1365 — nothing proves the boot refusal is reachable or legible against a real `fly deploy` (`developer`, `icebox`; paired with the `docs/architecture.md` §9.4 row in this diff, per §9.4's file-it-and-row-it rule)
- #1366 — `make test-js` flakes on 5s timeouts under parallel load (`developer`)
- #1367 — the gate accepts a blank or weak secret, since it compares against the exact declared default (`developer`). Sharpened during review: the remedy the gate itself prints, `$(openssl rand -hex 32)`, yields `""` if `openssl` isn't on PATH — so this is reachable by following this PR's own instructions


aghadiayeamayanvboernest (changes requested):
Requesting one change before merge. The gate compares each secret to Settings.model_fields[field].default, so its safety depends on both secrets keeping literal defaults — you flag this in the comment, but nothing guards it. I confirmed it: if either ever moves to Field(default_factory=…), .default becomes PydanticUndefined, the real dev secret compares False, and the gate silently passes a dev-default secret. The tests stay green because they inject .default as the value, so they'd never catch it.

Please add a one-line guard in the test module so a future switch fails loudly:

assert isinstance(Settings.model_fields["session_secret"].default, str)
assert isinstance(Settings.model_fields["ws_signing_key"].default, str)

Everything else is solid — ran the suite (13 pass), fails-closed correctly, good "names only the offender" coverage. Just want that guard in before this ships, since it's a security gate.

yinkid28 (dismissed):
Clean. No blocking findings — approve.

  Verified: test_prod_preflight.py 7/7, test_sessions.py 6/6, engine typecheck clean, engine vitest 2047/2048 (1 pre-existing doc-lint), harness 1598 passed. Full server suite: 5 failures + 1 hang, all pre-existing os.killpg Windows errors confirmed on main.

  Not a blocker, noting for the record

config.py:223 compares against Settings.model_fields[field].default, which becomes PydanticUndefined if either secret field is ever redeclared with default_factory — the gate silently passes. The inline comment documents the trap and the test would break (confusingly, not descriptively). Unusual enough to not block on, but a one-line assert isinstance(..., str) guard at the top of the function would make the failure immediate if it ever happens.

aghadiayeamayanvboernest (approved):

aghadiayeamayanvboernest (approved):

PR #1239, opened 2026-08-04
eval: fail loudly when the grading credential is bad

yinkid28:
## Summary

Closes #1023.

A duplicated API key silently failed all test grading — the existing preflight checked `if not auth.api_key:` (presence only), so a duplicated/revoked key passed the gate, the entire suite ran, and every positive test failed at grade time. The operator saw "everything failed" and read it as a skill regression rather than an auth error.

- **Unit preflight** (`run_tests.py`): hoists `needs_judge` above the presence check and adds a 1-token liveness call (`messages.create`) that aborts with exit 2 on 401/403 before spending anything. Transient errors (429/529/connection) pass through to the judge's own retry loop. Lazy `import anthropic` so the cost is paid only when the check fires.
- **E2e preflight** (`run_e2e.py`): same liveness check before a $7+ run. Does NOT abort on missing key (preserves the existing `--skip-judge` contract).
- **E2e judge-failure verdict** (`orchestrator.py`): changes the judge-error verdict from `"skipped"` to `"fail"` so runs with a valid final tree are committed (re-gradable later) instead of silently discarded.
- **Tests** (`test_cli.py`): 3 new tests (bad key → exit 2, transient 529 → pass-through, `--allow-missing-judge` bypass) plus a `_stub_anthropic_ok` helper wired into 9 existing keyed-auth tests so they don't hit the real API during the new liveness check.

## Test plan

- [x] `make harness-test` passes (44 tests, 0 failures)
- [x] Lint clean (no new violations)
- [ ] Manual: set a garbage key in `eval/.env`, run `python -m run_tests --skill search-records --tests-dir eval/tests/unit` → should exit 2 with "rejected it (401)" before any test runs

PR #1079, opened 2026-07-31
Resolve manoel oliveira daughter

Ikennaya1:
## Summary

- Resolved the manoel-oliveira-daughter record-hint fixture as a true match ("probable" tier), corroborated by an independent record for a previously-undocumented son of the same parents.
- Graded the live /research run via /grade-e2e-run: it correctly identified Josefa and her marriage record but never linked her into the tree — f1: false, retained as honest signal (see README).
- Found a second FamilySearch entry for the same marriage with a conflicting date (15 Dec 1924 vs. 12 Mar 1926). Checked both page images — neither shows this marriage; FamilySearch's own indexing is mismatched. `expected-findings.json now accepts either date.

**Flag for reviewer:** the automated judge scored this finding as recovered; blind grading here scored it not recovered — filed as #1090.

## Test plan

- [x] `make e2e-validate TEST=manoel-oliveira-daughter` passes
- [x] Run graded blind via `/grade-e2e-run`
- [x] Both conflicting-date sources checked against their actual page images

PR #1384, opened 2026-08-06
feat(e2e): record run provenance (git_sha + skills_hash) so a run tie…

aghadiayeamayanvboernest:
research-plan… no — **e2e: record run provenance (git_sha + skills_hash) so a committed run ties to its prompt (#1091)**

A committed e2e run recorded its result but not the prompt that produced it, so a graded run landing alongside a SKILL.md edit could not be tied to the skill version that ran. Add two fields to the e2e run-log envelope:

- `git_sha` — `git rev-parse HEAD` at run start (check out the exact tree)
- `skills_hash` — one sha256 over the sorted `{path: hash}` of every skill + agent file the run stages; catches an UNCOMMITTED SKILL.md edit `git_sha` cannot

New helper `e2e/provenance.py` reuses `harness.snapshot.normalize`/`hash_content` and hashes exactly what `build_workspace` stages (non-dot skill subdirs + top-level agent `.md`), minus dotfiles/`__pycache__`. A single digest, not a full snapshot (would add ~43% to a corpus #985 is shrinking) and not a per-file map. Evidence for a reviewer, not a CI gate: no bump to HARNESS_SCHEMA_VERSION, no schema mirror, no CI check. Documented in `e2e-test-spec.md` §8.1.3.

## Summary

- Adds `git_sha` + `skills_hash` to the e2e run-log envelope so a committed run can be tied back to the exact prompt (skill + agent files) that produced it — the gap PR #1079 exposed.
- Single digest, not a full snapshot or per-file map: identity, not content, is what was missing. No `HARNESS_SCHEMA_VERSION` bump, no schema mirror, no CI gate — evidence for a reviewer.
- Tier: **Normal**.

## Review guidance

**Start here:** `eval/harness/e2e/provenance.py::staged_file_hashes` — it must hash **exactly** what `build_workspace` stages (non-dot skill subdirs, recursive, + top-level agent `.md`) and nothing else; under/over-scoping is the whole failure mode, and a wrong hash is a plausible-looking 64-hex string. The acceptance test asserts the helper's key set equals what `build_workspace` actually stages — I broke the helper two ways to confirm it goes red. Also worth a glance: `orchestrator.py` hashes agents from `DEFAULT_PLUGIN_AGENTS`, correct only because the single `build_workspace` call site takes that default — there's a comment tying them so a future `--agents-dir` can't silently diverge the hash.

**Unsure about:** Nothing structural. One judgment call: entries are keyed staging-relative (`skills/…`, `agents/…`) not repo-relative, so a `--skills-dir` outside the repo still hashes — flag if you'd rather repo-relative.

## Plan

**Didn't change:** No `HARNESS_SCHEMA_VERSION` bump (additive fields with defaults — the rule at `result.py:47-54`). No `run-log.schema.json` mirror (that's the *unit* envelope; the e2e result has none). No CI check on the fields (a run is graded days after the tree moves, so a "hash matches working tree" rule would red every e2e PR).

**Acceptance check:** `eval/harness/tests/unit/test_e2e_result.py::test_skills_hash_covers_exactly_what_build_workspace_stages` — fails on `main` (module `e2e.provenance` absent), passes here; it stages a synthetic tree through the real `build_workspace` and asserts its file set equals the helper's key set.

**Deviated from the plan:** Nothing structural. Post-`/critique-plan` I tightened three flagged points (stated the "minus dotfiles/`__pycache__`" invariant precisely, tied the agents constant to what `build_workspace` uses, named the #1239 test-file collision). An independent code-review pass then found no critical issues and one doc nit (§8.1.3 sat before §8.1.2), now fixed. Did **not** hit the step-4 stop rule.

## Test plan

- [x] Ran `make test-all`. Every suite this PR affects is green: eval harness pytest **1603**, engine vitest doc-links **113**, control-plane **135**, eval-app **148**. One unrelated failure — `apps/electron … buildFeedbackZip > drops the largest files …` times out at 5000ms; it's in `apps/electron` (untouched here) and is the known flake **#1366**.
- [x] N/A — no skill run-log snapshot changed. Harness-only change (`eval/harness/**` + a spec doc); `check_runlogs` marks no skill touched, so no `make eval-skill` run is owed.

## Follow-on issues

None. If a reviewer ever needs the prompt *diff* (not just identity), `git_sha` + `skills_hash` reconstruct it from the repo on a clean tree — recorded in §8.1.3, not refiled.

**Dependency:** land after **#1239** (open) — it edits all four of my existing files; conflicts are additive and I'll rebase once it merges.

### DIFF
--- a/docs/specs/e2e-test-spec.md
+++ b/docs/specs/e2e-test-spec.md
@@ -1146,6 +1146,8 @@ editing one unreadable line, and it had already accreted a duplicated clause.
 | `cli_version` | So a harness-vs-Cowork gap can be checked against a CLI-version delta. |
 | `timeline[]` | Per-message `[elapsed_seconds, kind]`, plus the `caps` used. |
 | `subagents[]` | One summary per plugin subagent from the SDK's ephemeral cache: `agent_type`, per-turn `stop_reason` / `output_tokens` / block shape, and `runaway_thinking` (a turn that hit `max_tokens` on thinking alone with no tool call). The runlog stores no subagent transcript, so this is what makes a subagent freeze diagnosable from the committed log rather than only from `subagent_capture.py`'s local cache. |
+| `git_sha` | `git rev-parse HEAD` at run start, or `null` outside a checkout. The tree the run started from — check it out to reproduce. §8.1.3. |
+| `skills_hash` | One sha256 over the sorted `{path: hash}` of every skill + agent file the run staged. Ties the run to the prompt that produced it — and unlike `git_sha` catches an **uncommitted** SKILL.md edit. §8.1.3. |
 
 Together the five reasoning-config fields (`agent_model` through `cli_version`)
 make an A/B across model × effort × output-budget self-describing from the log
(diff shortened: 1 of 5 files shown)


clack391 (approved):
Looks good, verified against #1091. The provenance helper hashes exactly what build_workspace stages, reuses normalize/hash_content, stores a single digest not a per-file map, and the acceptance test compares against the real build_workspace so it is not vacuous. All the do-nots hold: no version bump, no schema mirror, no CI gate. Two small doc nits, your call, neither blocks: (1) result.py:250 still says "nothing records a git sha" in the guardrail_shadow_violations comment, which this PR makes false going forward, so scope it to the pre-v1 logs; (2) the spec sections are still out of order, 8.1.3 before 8.1.2, even though the PR body says that was fixed. Approving.

PR #1288, opened 2026-08-04
Fix windows test portability

Emruthwill:
Closes #1299

## Summary

**Normal.** Makes the test suite pass on Windows. 12 tests were failing on Windows machines but passing on CI, so the genealogist team couldn't get a clean run on their own computers.
- None of them were real bugs in the code being tested. They were three small assumptions that only hold true on Mac and Linux.
- While fixing them I found a real bug in `scripts/setup-feedback-case.sh`: on Windows it quietly threw away the user's prompt from the "next steps" message it prints.


## Review guidance

**Start here:** the `_find_bash()` function in `eval/harness/tests/unit/test_setup_feedback_case.py`. That's the only part with a real decision in it.

The problem: these tests run a `.sh` script. On Mac and Linux the computer reads the `#!/bin/bash` line at the top of the file and knows what to do with it. Windows doesn't do that at all, so it just says "this isn't a program" and the test fails. The fix is to say "run this file *using bash*" instead of hoping Windows figures it out.

But then: where is bash on Windows? Normally you'd just ask the system to find it. That doesn't work here, because Git for Windows hides it — it only puts `git.exe` where the system can see it, and leaves bash tucked away in a folder next door. So the function finds git first, then looks next to it for bash.

**One thing to check and tell me if you disagree:** if it can't find bash at all, every test in that file is *skipped* rather than *failed*. I chose that because a machine with no bash genuinely can't run these tests, so failing would be misleading. But skipping can hide problems, so it's a fair thing to push back on.

**Also worth a look:** the `python3` → `python` change in `setup-feedback-case.sh`. This one is a genuine bug, not just a test fix. Windows comes with a fake `python3` file that exists but doesn't work — it's a placeholder that opens the Microsoft Store. So the script asked "does python3 exist?", got told yes, tried to run it, failed, and quietly gave up. The user's prompt vanished from the output and nobody would know why. Now it tries `python` as well, which is the one that actually works on Windows.

**Unsure about:** nothing that blocks this.

One test is skipped on Windows on purpose: `test_claude_skills_dir_is_real_with_symlinks`. Windows can only make real shortcuts if you're an administrator or have Developer Mode switched on. Without that, Git Bash just copies the folder instead. So the thing the test is checking can't actually happen on Windows, and pretending otherwise would be a fake failure. The skip message explains this, so nobody will mistake it for a test that passed.

## Plan

[...]

PR #1386, opened 2026-08-06
question-selection: stop instructing a question-status write the validator rejects

Emruthwill:
## Summary

- **Normal tier.** `question-selection/SKILL.md` instructed the agent, in three places, to retire a question by writing `status: "superseded"` or `"answered"`. Neither is a member of `question_status` (`open|in_progress|exhaustive_declared|resolved`), so validate-before-persist rejected the write every time an agent followed it.
- The instruction is **deleted, not corrected to a legal value**: per `research-schema-spec.md`'s `questions` ownership row this skill owns no status transition (`exhaustive_declared` is research-exhaustiveness's, `resolved` is proof-conclusion's), so there is nothing legal for it to say.
- This is the narrowed remainder of #1135. Its stated lever (a no-candidate routing short-circuit) and the question-grain doctrine fix behind it are both cross-skill; filed separately below.

## Review guidance

**Start here:** the `## Rules` bullet in `question-selection/SKILL.md`. Two things worth checking. (1) Why `resolved` was *not* used as the replacement terminal status: a question marked `resolved` with an empty `resolution_assertion_ids` references no `ps_id`, so `/research`'s tree-encoding gate and its mentor-verdict completion check both pass **vacuously** and `project.status` reaches `completed` silently — a loud validation failure traded for a quiet early stop. (2) The bullet now states the legal values in the `∈` form, which puts this prose copy under the existing `enum-drift` lint (the file is added to that lint's `EXPECTED` registry) rather than making it a second place the enum can rot.

**Unsure about:** whether this deserves the `eval-cosmetic-skip` label — the only open question here, and it is a senior's to answer. The edit removes an instruction whose write *never succeeded*, so the skill's legal behaviour is unchanged; that is the cosmetic argument. What I can confirm is that the label path **works mechanically**: the prior run log `eval/runlogs/unit/question-selection/v1_2026-07-22_11-02-36` carries 57 graded dimensions and 57 matching corrections, so rule 3 is satisfied against the existing annotation with no re-grade, and rule 2 is the only rule the label relaxes. So the choice is genuinely label-vs-paid-run, not label-vs-blocked. I did not assume the label.

## Plan

**Didn't change:** `/research`'s routing table (no short-circuit, no gate deferral — GPS rigor is untouched); the `question_status` enum and all four of its mirrors; `research-schema-spec.md` (this PR aligns the skill *to* it); the `question-formulation.md` doctrine and the Priority 4 decomposition example, which is the actual root cause of the #1135 latency symptom but is cross-skill work.

**Acceptance check:** `research-append.test.ts` → "rejects retiring a question with a status outside question_status" (`make engine-test`). **Honest caveat, stated rather than buried: it does not fail on `main`.** The validator always rejected these values — that *was* the bug. There is no red-to-green test for prose that instructs an impossible write, and claiming one would be inventing falsifiability. The test earns its place by pinning the fact so the instruction can't be reintroduced on the belief that the write works. What actually verifies this change is review of the three edited sites plus the confirmation that no other skill misuses the enum (checked: every other `superseded` in the plugin is `plan_status`, the `superseded_by` field, or prose).

**Deviated from the plan:** Yes, twice. (1) The plan called for a new prose lint asserting no illegal value appears in a status-mutation instruction. **Unbuildable at the prose layer** — it false-positives on three legal `status` mentions in the same file (`plan_items[].status`, a conflict's `status == "unresolved"`, `project.status`, the last two inside backtick spans) and false-negatives against the corrected wording, which names a value in order to forbid it. A lint more fragile than the bug it guards is worse than none; replaced with the contract test plus the `∈` registration above. (2) A Windows portability fix I wrote for `eval/harness/tests/unit/test_setup_feedback_case.py` was **reverted as redundant** — `main` already had #1288, which solves it better. I had been working from a three-week-stale base; the branch has since been fast-forwarded onto current `main`.

Did **not** hit the step-4 stop rule. The schema-adjacent decision (whether to add a `question_status` value) was raised with the lead *before* building and answered "reuse `resolved`" — which this PR then declined to use, for the vacuous-gate reason under "Start here", choosing to delete the instruction instead. Flagging that inversion explicitly rather than letting it pass as the approved option.

## Test plan

- [x] `scripts/test.sh` — run directly (`make test-all` fails on this machine: the recipe mis-splits a path containing spaces). Engine suite on the updated base: **2038 passed, 1 failed**. The one failure is `doc-links.test.ts`, which is **pre-existing and unrelated** — proved by re-running it with this branch's changes stashed. It is a Windows path-separator bug in that lint (`relative()` yields backslashes; `GRANDFATHERED_LINE_CITES` is keyed with `/`), green on CI and red on every Windows machine. Fix is on its own branch, `windows-doclinks-path-separator`. Also: `apps/web` typecheck fails on an empty `apps/web/node_modules/@genealogy/` (workspace links missing on this checkout) — likewise pre-existing, proved by stash.
- [ ] Run-log snapshot: **not run.** `question-selection/SKILL.md` is in the snapshot, so this needs either `make eval-skill SKILL=question-selection` + its `.ann.json`, or the `eval-cosmetic-skip` label. See "Unsure about".

Also checked against the two lints that arrived with the base update: `prompt-budget` (warn-only) reports "no prompt-size changes in this changeset", and `gps-terminology` (blocking) passes.

**Self-review:** `/check-drift` ran against `PLAN.md` and this diff — three findings, all verified and fixed in the second commit (a clause that added a *legal*-write instruction and so undercut the cosmetic-skip argument; a prohibition narrowed to "to retire it"; a stale comment count). `/code-review high` has **not** run: in a terminal it forks the session it is invoked from, so it needs a session that did not write this code.

**§9.4 disclosure:** `docs/architecture.md` §9.4 records that there is **no unit suite for `research`**, the orchestrator. This PR does not touch routing, so that gap isn't load-bearing here — but the deferred short-circuit in #1135 is pure routing, and nothing but a live e2e run would cover it.

## Follow-on issues

- **#1394** — question-selection: one question per unknown fact, records as plan items (the actual root cause of #1135's latency symptom; cross-skill, carries the three blockers that stopped it landing here)
- **#1395** — /research: short-circuit a question whose retrieval yields no candidate answer (#1135's stated lever, still unbuilt)
- **#1396** — a no-candidate question ping-pongs between research-exhaustiveness and research-plan (likely the cheaper fix; worth measuring before #1395)
- **#1397** — research-schema-spec: "supersede with status" names a `question_status` value that does not exist (the spec prose that produced this bug)
- **#1398** — lint skill prose that instructs an enum write the validator rejects (icebox; carries the false-positive analysis so it isn't retried blind)
- **#1399** — proof-conclusion's `**Writes:**` line omits the question `status → resolved` transition it owns

Also worth flagging to whoever owns **#1022**: it left `question-selection`'s run log stale on `main` (five of the six files rule 2 reports here), which is why this PR trips a gate it didn't cause.


kofiatinka12 (approved):
This is a good cleanup. The SKILL guidance, schema behavior, and regression tests are now aligned, so question selection no longer instructs agents to write invalid question_status values. I also like that the packaging test pins the enum reference to help prevent this kind of drift in the future. No further concerns from my side.

PR #1387, opened 2026-08-06
test(doc-links): match grandfathered cites on Windows path separators

Emruthwill:
## Summary

- **Trivial tier.** `docsMarkdown` in `doc-links.test.ts` built its keys with `relative()`, which yields **backslash** paths on Windows, while every key in `GRANDFATHERED_LINE_CITES` is written with forward slashes.
- So the lookup missed on every file, all six long-grandfathered citations were reported as brand-new offenders, and the check was **green on CI (Linux) and red on every Windows machine** — which is the whole genealogist team.
- Normalize the separator at the point the path is recorded. One line.

## Review guidance

**Start here:** the single `out.push(...)` line. The thing worth a second look is whether any *other* packaging lint keys a map by a `relative()` path and has the same latent bug. I checked the three other `relative()` call sites in this file — they build message strings, not lookup keys, so they're unaffected. I did not audit the other packaging tests for the pattern; that's a reasonable thing to ask me to do if you'd like it widened.

**Unsure about:** nothing. The cause and fix are both mechanical, and the before/after is measurable.

## Plan

**Didn't change:** the six grandfathered entries themselves. They are stale line-number citations in `docs/specs/` that the lint deliberately tolerates; fixing them is separate work with a different reviewer. This PR only makes the tolerance actually apply on Windows.

**Acceptance check:** `tests/packaging/doc-links.test.ts` → "bans new source-line citations under docs/". On Windows it is **1 failed | 112 passed** before this change and **113 passed** after. On Linux it passed both before and after — the bug is platform-specific, so CI alone could never have caught it.

## Test plan

- [x] Full engine suite on this branch: **84 files, 2040 tests, 0 failures** (`npx vitest run` in `packages/engine/mcp-server`). I also ran `eval/harness/scripts/check_runlogs.py` with CI's `BASE_SHA`/`HEAD_SHA`: "All runlog rules satisfied" — this branch touches no skill dir, so nothing is gated.
- [ ] No skill run-log snapshot changed, so no `make eval-skill` run is needed.

## Follow-on issues

None from this change. It came out of pre-flighting #1386, where the same test was failing locally and turned out to be unrelated to that work — so it was split out rather than ridden along.


kofiatinka12 (approved):
Nice cross-platform fix. Normalizing the paths to forward slashes makes the test behave consistently on Windows and Unix-like systems, and the comment clearly explains the reason for the change. I don't have any concerns.

Issue #1389, opened 2026-08-06
Consolidate the three preferred-name selectors into one shared helper


clack391:
> **Reviewed 2026-09-21.** Unblocked — issue #2002 closed COMPLETED, so the 2026-08-28
> pairing note no longer binds and has been removed. Scope was narrowed to the consumer
> side; see "Out of scope" below. **Sequencing: do this after PR #2674 merges** — it and
> draft PR #2727 both carry a hunk over `person-warnings.ts` lines 111–136, immediately
> above `getPersonName` at 141, so expect an adjacent-hunk conflict and a line shift.

**Touches:** packages/engine/mcp-server/src/utils/relatives.ts, packages/engine/mcp-server/src/tools/person-warnings.ts, packages/engine/mcp-server/src/tools/project-context.ts, packages/engine/mcp-server/src/utils/name-helpers.ts (new)
## Why

Three sites independently reimplement "pick the person's primary name" as
`names.find((n) => n.preferred) ?? names[0]`:

- `src/utils/relatives.ts:220` (`fullName`)
- `src/tools/person-warnings.ts:143` (`getPersonName`, an already-exported display-name helper used ~9x in that file)
- `src/tools/project-context.ts:86` (`preferredDisplayName`)

Three near-duplicates of the same selector is the CLAUDE.md "two near-duplicates is the signal
to consolidate" case. #1318 fixed the primary-name bug at the source (the converter now
stable-reorders `preferred` names to `names[0]`, so all three selectors return the same name),
but the duplication remains and will drift.

## What

Add `src/utils/name-helpers.ts` exporting **one selector and nothing else**:

```ts
/** The person's preferred name, else the first, else undefined. */
export function preferredName(names: SimplifiedName[] | undefined): SimplifiedName | undefined
```

Use `n.preferred === true`, not a truthy test: `preferred` is `const: true` in the tree
schema (`docs/specs/simplified-gedcomx-spec.md` §"Why `preferred` and `primary` are
omit-when-false"), `tree-sanitize.ts` prunes any non-`true` value at read, and
`project-context.ts` already tests `=== true`. The two spellings can only differ on a value
the validator rejects. New file rather than an addition to an existing util, mirroring
`fact-helpers.ts` (per-object helpers for facts); `name-pairs.ts`/`name-variants.ts` are
similarity utilities, not selectors.

Route all three sites through it (line numbers measured 2026-09-21,
[...]

PR #1390, opened 2026-08-06
fix(gedcomx): select the preferred name as the primary, not an alternate (#1318)

clack391:
## Summary

Normal tier. Alpha feedback tool bug (PID KLYC-JX5). `simplifyPerson` mapped a person's names in
FamilySearch's array order and never moved the `preferred` name to the front, so when FS listed an
alternate (`type: AlsoKnownAs`) before the preferred name, simplified `names[0]` was the alternate.
Every `names[0]` consumer (`person_read`, `person_ancestors`) then showed the alternate as the
primary, manufacturing a phantom conflict the agent spent turns resolving.

- Stable-reorder `preferred: true` names to the front in `simplifyPerson`, upholding the simplified
  format's first-position convention (`simplified-gedcomx-spec.md` section 2) that `names[0]` is the
  preferred name.
- Rewrote `gedcomx-convert-spec.md` section 4, which asserted the opposite ("does not reorder by
  preferred") and directly contradicted the simplified spec.

## Review guidance

**Start here:** `simplifyPerson` in `gedcomx-convert.ts`. The fix is a stable partition:
`[...names.filter(n => n.preferred === true), ...names.filter(n => n.preferred !== true)]`.
`simplifyName` only ever sets `preferred = true` (never false), so the two filters partition every
element exhaustively and `filter` preserves relative order within each group. It sits on the single
shared per-person path both `toSimplified` and `toSimplifiedStandardized` use, so it fixes
`person_read`, `person_ancestors`, and every `names[0]` consumer at once; the
`find(preferred) ?? names[0]` selectors return the same name either way.

**Unsure about:** nothing structural. I fixed the converter rather than only `person-read` because a
consumer-only fix leaves `person_ancestors` and the spec invariant broken. The issue lists
`person-read.ts` as a touch site, but with the converter upholding the invariant, `person-read`'s
`names[0]` is correct unchanged (a redundant `find(preferred)` there would be dead code).

## Plan

**Didn't change:** the three `find(preferred) ?? names[0]` selectors (relatives, person-warnings,
project-context) are order-independent and unaffected; consolidating them is follow-up #1389.
`merge-gedcomx.ts` sets its own preferred; `record-search.ts` reads full GedcomX. No
schema/enum/tree-shape field change.

**Acceptance check:** `tests/utils/gedcomx-convert.test.ts` Test 9b — convert a synthetic person
with `[AlsoKnownAs alternate first, preferred second]` and assert `names[0]` is the preferred one.
Fails on `main` (order preserved, `names[0].preferred === undefined`), passes here. Proven by
mutation: reverting the reorder makes Test 9b fail with "expected undefined to be true"; existing
Tests 8/9 stay green (their inputs are already preferred-first / all-preferred, so a stable sort is
a no-op).

**Deviated from the plan:** nothing structural. Two `/critique-plan` rounds moved the fix from a
`person-read` consumer patch to the converter (which also fixes `person_ancestors` and the spec
invariant) and moved the acceptance test from an isolated-helper unit test (a fake win) to the
converter output. Self-review then caught two things, now fixed: the test used the real feedback
person's names (replaced with synthetic values, since this file is committed), and a stray issue-ref
in the spec tripped the doc-links ratchet (removed; the fact is stated, not the ticket).

Step-4 stop rule: not hit. Name-order normalization in a shared converter; no schema, auth,
plugin-agent binding, or ADR change.

## Test plan

- [x] `make test-all` passed in a single clean run (exit 0): typecheck, JS 248, control-plane
      pytest 135, MCP engine vitest 2049, eval app 148, eval harness 1599. Plus
      `npx vitest run tests/utils/gedcomx-convert.test.ts` (53 pass) and the mutation check above.
- [x] No skill run-log snapshot changed (touches only `packages/engine/mcp-server/**` and a spec
      doc), so no `make eval-skill` run is owed; `check_runlogs` marks no skill touched.

The live `person_read`/`person_ancestors` pass-through is not CI-covered (architecture section 9.4);
the fix is in the pure converter, which is unit-tested.

## Follow-on issues

- #1389 consolidate the three `find(preferred) ?? names[0]` selectors into one shared helper, and
  assess `merge-gedcomx.ts` for the same latent issue.

Closes #1318.

### DIFF
--- a/docs/specs/gedcomx-convert-spec.md
+++ b/docs/specs/gedcomx-convert-spec.md
@@ -284,8 +284,12 @@ nameForms missing                                  →  all four fields omitted
 - Input GedcomX has `preferred: true` on a name → simplified has `preferred: true`
 - Input GedcomX omits `preferred` (or sets it `false`) → simplified omits it
 
-The conversion preserves the order of `names[]` as it appears in the input;
-it does **not** reorder by `preferred`.
+`simplifyPerson` stable-reorders `preferred: true` names to the front so
+`names[0]` upholds the first-position convention (`simplified-gedcomx-spec.md`
+§2) even when FamilySearch orders an alternate (e.g. `AlsoKnownAs`) before the
+preferred name; relative order within the preferred and non-preferred groups is
+preserved. Without this, every `names[0]` consumer (`person_read`,
+`person_ancestors`) can show an alternate as the primary name.
 
 `toGedcomX` is the mirror image: if simplified has `preferred: true`, emit
 `preferred: true` on the GedcomX name; otherwise omit. Never emit
@@ -895,6 +899,7 @@ fall into the `fullText` fallback path and emit a warning.
 | 7 | Mononym `"Plato"` → `given: ""`, `surname: "Plato"`, with `console.warn` | Rule 3 mononym |
 | 8 | `preferred: true` is passed through when set on input; omitted otherwise | Rule 4 |
 | 9 | Multiple names can independently have `preferred: true` | Rule 4 |
+| 9b | Preferred name is moved to `names[0]` when an alternate precedes it | Rule 4 |
 | 10 | `primary: true` is passed through when set on input; omitted otherwise | Rule 5 |
 | 11 | Multiple facts of different types can each be `primary: true` | Rule 5 |
 | 12 | `date.formal` is dropped, only `date.original` surfaces as `date` | Rule 6 |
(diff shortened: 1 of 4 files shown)


aghadiayeamayanvboernest (approved):
Looks good to me — checked it out and ran the converter suite (53 pass). Fixing it in simplifyPerson rather than just person-read is the right call since it fixes person_ancestors and the spec invariant at the same time, and the stable partition is clean. I looked at the round-trip risk specifically: a full→simplified→full now reorders names to put preferred first, but that's harmless since the preferred flag round-trips and FS uses the flag, not array order — and all the existing round-trip tests stay green. Only thing: you're a few commits behind main, so give it a rebase  Nice work.

Issue #1383, opened 2026-08-06
e2e: no cross-invocation rollup, so a batch sweep has no suite-level summary


taiwo-stack:
**Touches:** eval/harness/e2e/run_e2e.py, eval/harness/e2e/report.py

Recorded out of #1245 at the lead's direction ("go ahead and record it; don't worry about building it right now").

## What is missing

`print_rollup` is named and shaped like a suite reporter, but `run_e2e.py:210` appends exactly one result per invocation and `--test` takes a single fixture id. There is deliberately no full-suite flag: the module docstring says a batch is driven "with a shell loop" and budgeted for explicitly, because a 10-fixture sweep is 4-10 hours and $30-100.

So a 20-fixture sweep calls `print_rollup` twenty times with one result each. Every aggregate it computes — the recall line, the compliance line, the overall gate, the per-tag breakdown, avg/total cost and wall clock — is a summary of one run. There is no artifact anywhere that says what the sweep as a whole did.

## Why it matters

This is what made #1245 hard to answer. It was reported as "19 of 20 tests came back aborted", and that number could only have been arrived at by eye: no tool in the repo prints it, and the ungradeable runs write gitignored `scratch_` files, so no aggregate report can reconstruct it afterwards either.

`corpus_report.py` aggregates across runs, but only over **committed** `run-` logs, which by construction excludes exactly the ungradeable ones a sweep is most likely to produce a lot of.

## Shape of a fix, if it is ever wanted

Not obviously a new flag — the no-full-suite-flag decision is deliberate and worth keeping. More likely a small post-hoc reader that takes a set of run directories (or a time window) and prints the same three axes across them, so a shell-loop sweep can be summarised after the fact without making the expensive path easy to trigger by accident.

## Not in scope

Building it. #1245 records the gap and fixes the per-run reporting defects around it (an ungradeable run now says why, and an all-ungraded tag no longer renders identically to an all-failed one).

DallanQ:
**Line citations have drifted twice — re-derive rather than quote (weekly board audit, 2026-09-12).**

The behaviour this card describes is confirmed on main today, but the line numbers in the body have now moved twice: originally `:210`, corrected once to `:262`/`:280`, and as of 2026-09-12 the code is at `run_e2e.py:354` and `:372`.

Nothing else about the card changes — there is still no cross-invocation reader, so a batch sweep still has no suite-level summary. Whoever picks it up should locate the code by symbol, not by line.

[...]

PR #1393, opened 2026-08-06
eval-harness: an ungradeable run must say why, and must not read as a failure

taiwo-stack:
Closes #1245.

19 of 20 e2e tests came back with no grade and nobody could say why. Two things were true: the reason existed and was never printed, and a run with no grade was reported the same way as one that failed the genealogy. Nothing was spent; no suite was re-run.

## What changed

`e2e/run_e2e.py` — an ungradeable run now says which of the three causes it was, quoting the judge's own error when there is one. It used to print one fixed string for all three, directly under `stop_reason: completed`, so the two lines together said the run succeeded and produced nothing.

`e2e/report.py` — the per-tag breakdown printed only `pass/total`, so three ungraded runs rendered identically to three that failed the genealogy. Now named. An unrecognised verdict used to land in a dict key nothing printed, silently breaking the totals; now counted at the headline.

`run_tests.py` — an aborted test prints its reason inline and in a `REASON` column, plus a count line: `Outcomes: 1 pass, 19 aborted of 20 test(s).` That number is what was reported in standup, and the harness never printed it.

`eval/CLAUDE.md` — triager guidance: an `aborted` row is not a failure, read its reason, and a suite where nearly every test aborted with the same reason is an environment failure rather than twenty regressions.

## Against the acceptance boxes

**Where it is set.** Not in `eval/harness/e2e/` as the issue assumes: that package never sets `aborted`, and an ungradeable run becomes `skipped`. All sites are in `e2e/orchestrator.py`, 11 in `_run_agent` and 2 in `run_e2e_test`.

**Which one fired.** Two candidates, not one. `stop_reason` prints above the message so most causes were already distinguishable; only three collapse, and `--skip-judge` is unreachable through `make e2e-run`. That leaves judge-crashed or no-tree, and the field separating them (`judge_output`) is written gitignored, so it is gone. The one run committed that day was fully graded (38 min, $4.88), so both credentials worked that morning, which argues against a duplicate of #1023/#1239.

**Environmental or harness defect.** Both, and they differ. The cause was environmental. The reporting defect is separate and is what this fixes.

**Miscounted as `fail`?** Yes in the rollup, both paths, now fixed. `corpus_report` and `check_runlogs.py` are safe (they print each verdict and key on presence respectively). Across a batch there is no rollup at all, since `run_e2e.py` runs one fixture per invocation by design, so nothing ever prints "19 of 20". Filed as #1383, recorded not built.

## Declared adjacent scope

146 of 449 diff lines are `run_tests.py` and `test_cli.py`, the **unit** harness. The issue scopes itself to `eval/harness/e2e/` and that harness was not what ran, so strictly this is out of scope. Kept deliberately: identical defect, the surface a genealogist meets most often, and where the count line lives. Splits cleanly if preferred, no shared lines with the e2e change.

## Verification and sequencing

1620 passed / 1 skipped / 0 failed, ruff clean. The 12 Windows failures earlier runs carried are gone (#1299 closed by #1288).

`e2e/orchestrator.py` is untouched, which is what the five-PR warning protected; four of the five have merged. #1239 overlaps at `report.py` and `run_e2e.py`; three small hunks, resolution posted on that PR.

Gradedness and committability are separate axes here on purpose. #1239 makes a judge crash committable while still ungraded, so a single committability branch would go silent on the crash. `is_ungraded` deliberately does not import `result.py`'s verdict set, because that set is the committability one and #1239 widens it; a test pins the literal.

## Not met as written

Box 1 asks for *the* path that produced the 19. Two candidates survive and the discriminating field is gone, so this ships as two candidates plus the discriminator rather than one named path.

The new console block is covered at function level but has never executed in a live run, since its caller drives a full e2e run. One fixture with `--skip-judge` would prove the wiring.


clack391 (approved):

EdmondOware (approved):
The new "Reading an aborted row" guidance is clear and gives genealogists exactly what they need for daily triage. It correctly separates test-corpus problems, which belong to whoever wrote the test, from execution/environment problems, which usually don't. The concrete example format and the advice to re-run before reading twenty aborted tests as twenty regressions will save real confusion.
 The PR is upfront that it doesn't fully close the original incident, since two candidate causes survive for that specific run and the discriminating field is gone, but it does fix the actual defect closed here, that an ungradeable run reads identically to a real failure.

Issue #1398, opened 2026-08-06
Lint skill prose that instructs an enum write the validator rejects


Emruthwill:
> **Reviewed 2026-09-21 before junior handoff; decided and closed 2026-09-22.** The
> 2026-09-20 promotion argument was sound on its own terms; what it did not have is the
> lead's own 2026-09-07 ruling on the same five lines, re-verified below.

## Closed 2026-09-22 — no prose enum lint; the 2026-08-10 ruling stands

Re-derived before closing: **zero live instances of the defect this card detects.** The
one genuine instance (`question-selection`'s `superseded`/`answered`) was fixed
2026-08-07 by PR #1386 and its prose converted to an `∈` declaration the existing lint
owns (`skills/question-selection/SKILL.md:316`, issue #1135). The five `exhausted` sites
in `research-plan` are live (`grep -rn 'exhausted'
packages/engine/plugin/skills/research-plan/`) but are **correct prose against a wrong
enum** by the 2026-09-07 consolidation in issue #2477 — "the skills are already correct
and the enum is wrong" — so they are not instances of this shape.

Beat: extending PR #2716's `retired-field-names.test.ts` with a banned-token registry.
Rejected because an inverse-search lint only fires on a token someone already registered,
and a never-legal value has no rename moment at which anyone would register it — the same
reason ADR-0008 deferred that row.

Both refutations already live in `docs/adrs/ADR-0008-…md` § Alternatives considered;
add this card's number and the zero-instance re-derivation to that row's evidence column
in the next PR that touches the file, so the next promotion sees it.

**The 2026-09-07 consolidation was re-verified before closing (2026-09-22), because the
whole close rests on it.** `exhausted` is a state the schema cannot express and the three
legal values do not cover: a plan retires either because its premises were invalidated
(`superseded`) or because no further sources are known (`exhausted`, BCG Standard 18), and
`completed` says only that the plan ran its course, not that the question cannot be
answered from those sources. The five skill sites all describe that one state consistently
and none of them is wrong about anything but the enum's refusal. Issue #2477's own
measurement carries it: over 166 committed e2e final states, 171 `active`, 37 `completed`,
7 `superseded` and **0 `exhausted` — because it cannot be written** — with 5 of the 7
recording no reason anywhere, there being no field to carry one.

The counter that `question_status` already encodes this was considered and does not hold:
a question can carry several plans, so a question-level status cannot say *which* plan was
[...]

Issue #1395, opened 2026-08-06
/research: short-circuit a question whose retrieval yields no candidate answer


Emruthwill:
**Touches:** packages/engine/plugin/skills/research/SKILL.md:147, packages/engine/plugin/skills/research-exhaustiveness/SKILL.md:113-116, docs/specs/gps-mentor-agent-spec.md (§17.1, and the "Reducing per-gate mentor cost" row in §17's table), eval/tests/e2e/hole-parents-negative/

The original #1135 lever. Only the illegal-status half of that issue shipped.

Predicate (derived, so nothing extra to keep in sync): all plan items
`completed`/`skipped` ∧ zero assertions naming the question in
`extracted_for_question_ids` ∧ every log entry `negative` ∧ no outstanding
low-coverage-collection pivot.

Four things to carry, each verified:

- **#1135's claim that the agent "already emits" a no-candidate signal is not
  accurate.** There is no per-question signal — only `outcome: "negative"` per log
  entry. And `search-records/SKILL.md:465` forbids reading absence out of a
  `negative` on a near-0%-indexed collection (the record is probably present as an
  un-indexed image). The predicate must not fire while a full-text/browse avenue is
  unattempted.
- **Terminal state.** The lead's call was to reuse `resolved` with an empty
  `resolution_assertion_ids` rather than add a `question_status` value. **Hazard:**
  such a question references no `ps_id`, so `/research`'s tree-encoding gate and its
  mentor-verdict completion check both pass **vacuously** and `project.status`
  reaches `completed` silently.
- **Carve-out.** When the no-candidate question is the objective-answering one, the
  negative result *is* the answer and must still go through exhaustiveness →
  proof-conclusion → mentor at `not_proved`/`disproved`. `hole-parents-negative` is
  the fixture that catches getting this wrong.
- **Reconciliation.** `gps-mentor-agent-spec.md` §17.1 says do not defer or sample
  gates to chase speed. That objection is scoped to the surviving `proof-critique`
  mentor gate, which this lever never reaches (it stops before a proof summary
  exists). Add one sentence to §17.1 saying so, so the next reader does not
  re-litigate it.

Rigor-critical per #1135: validate on an instrumented e2e re-run — and note that
`research` has **no unit suite** (`docs/architecture.md` §9.4), so a live e2e run is
the only thing that covers this at all.

## Collision with issue #1836, added 2026-08-23

The predicate above reads **zero assertions naming the question in
[...]

Issue #1394, opened 2026-08-06
question-selection: a premise-verification question must still name the fact sought


Emruthwill:
**Touches:** packages/engine/plugin/skills/question-selection/SKILL.md, packages/engine/plugin/skills/question-selection/references/question-formulation.md, eval/tests/unit/question-selection/, eval/fixtures/scenarios/, eval/harness/validators/test_question_selection.py, eval/tests/unit/question-selection/rubric.md, eval/runlogs/unit/question-selection/

**Line numbers, verified against `main` 2026-09-16** — re-check with
`grep -n -i "premise" packages/engine/plugin/skills/question-selection/SKILL.md packages/engine/plugin/skills/question-selection/references/question-formulation.md`:
`SKILL.md:128`, `:180-183`, `:264`; `references/question-formulation.md:13-18`, `:86-96`, `:98-109`.
The `rubric.md` entry applies only if you add a grading dimension.

**Scope: Rule B only.** Rule A (one question per unknown fact) shipped — lead ruling
2026-09-12, re-verified 2026-09-16: `references/question-formulation.md:102-109` now carries a
**Record-first framing** row rejecting the very example the old body quoted as canonical, and
`SKILL.md:120-129` (§1c "Scope the question to the objective") replaced the three-question
Patrick Flynn decomposition. Treat Rule A as context for why Rule B is phrased this way, not as
work. The original two-rule body is in this card's history; the lead's two rulings are in the
comments.

**On the critical path.** #2115 (the skill-agent fold, which deletes `references/`) is blocked
on this card by lead ruling 2026-09-16, option A; #2267 sits behind #2115. Write the edits
against the files as they are today — the fold carries the corrected prose forward verbatim.

**Not in scope, decided — do not widen:**

- **`research-plan`.** Its slot is held by in-progress #2075 (see #2510), and editing
  `skills/research-plan/**` would flip that skill's snapshot inactive and buy a second paid run.
  Separately, for the record: two of the four sites the lead's 2026-08-13 comment listed for
  `research-plan` are **still unshipped** — `research-plan/SKILL.md:358-362` still reads "more
  than 12 suggests the question is too broad — consider splitting" (last touched 2026-08-08,
  `ba5a534df`) and `:561` still bounces "Question is too vague to plan for" with no
  grain/jurisdiction distinction (unchanged since 2026-05-10). That residue belongs to
  `research-plan`'s queue (#2510), not here — and today it is owned by no card.
[...]

PR #1400, opened 2026-08-06
docs(research-schema): name the three supersede mechanisms instead of implying one everywhere

Emruthwill:
Closes #1397.

## Summary

- **Trivial tier**, docs only — one file, no schema/validator/skill change. The schema was right; the prose describing it was wrong.
- The `questions` row said *"Mutable; never delete, supersede with status"*, and the **General rule** under the table generalized that to every section: *"never delete entries — supersede with a status field instead."*
- `superseded` exists in **exactly one** status enum (`plan_status`); `superseded_by` on **exactly two** objects (`person_evidence_entry`, `evaluation_entry`). Every other section has no supersede mechanism, so that instruction describes a write `validate_research_schema` rejects.

## Review guidance

**Start here:** the new three-row mechanism table replacing the General rule's blanket sentence. The claim to check is the third row — that `questions`, `conflicts`, `hypotheses`, `sources`, `assertions`, `localities` and `proof_summaries` have no supersede mechanism. I derived it from the schema rather than by reading prose: every `*_status` enum in `enums.schema.json` (`question_status`, `plan_status`, `plan_item_status`, `conflict_status`, `hypothesis_status`, `project_status`) and every `$defs` entry carrying a `superseded_by` property. `plan_status` is the only enum with `superseded`; `person_evidence_entry` and `evaluation_entry` are the only objects with the pointer.

**Unsure about:** whether the `localities` clarification is a fix or a reword. That row said "supersede-not-delete" for a section with no status field. I changed it to say the entry refreshes in place, which is what `locality-guide`'s own "Re-invocation behavior" section says it does — so I read the old wording as loose rather than false. If the intent was that localities *should* be supersedable, this is the wrong fix and the schema is what needs changing.

## Plan

**Didn't change:** the schema, the validator, `enums.schema.json`, and every skill body. Also left alone: adding a `superseded` value to `question_status` (or anywhere else). That would be a closed-enum change with a five-site blast radius, and nothing needs it — the point of this PR is that the mechanism doesn't exist, not that it's missing.

**Acceptance check:** none that fails on `main` — this is prose. It is verified by inspection against the schema, as described under "Start here". Stating that plainly rather than dressing up a test that wouldn't discriminate.

**Deviated from the plan:** the issue named one line (the `questions` row). Two more sites turned out to carry the same defect — the General rule, which is the more consequential of the three because it reads as blanket policy, and the `localities` row. Widened to all three rather than fixing the symptom the issue happened to spot.

## Test plan

- [x] `npx vitest run tests/packaging/` in `packages/engine/mcp-server` — passing, with one pre-existing failure that is not from this branch: `doc-links.test.ts` → "bans new source-line citations", the Windows path-separator bug fixed on the doc-links branch. On Linux CI that test passes. Verified by name, and by re-running with this branch's edit stashed.
- [ ] No skill run-log snapshot touched (`docs/` only), so no `make eval-skill` run and no `check-runlogs` gating.

**Correction — the first push of this branch failed CI, and my original test-plan claim here was wrong.** I had written that the only failure was the pre-existing Windows one. In fact `vitest` went red on a *different* test in the same file: "holds the issue-reference ratchet on contract docs". This file is allowed exactly 1 issue reference — *"specs state the fact, not the ticket"*, and the lint's own comment says adding one fails outright. My note cited an issue number, making 2. Fixed in the second commit by stating the fact instead of the citation.

Worth recording because the masking is reproducible: on **Windows** the line-citation test fails and the ratchet passes; on **Linux** the reverse. Both report `1 failed` in `doc-links.test.ts`. I read the count and the filename and assumed the known bug instead of reading the failing test's *name* — which is the only thing that distinguishes them.

## Follow-on issues

None new. This closes the loop on #1386, which deleted three instructions that were derived from the prose corrected here. The remaining related work is already filed: #1394, #1395, #1396, #1398, #1399.


kofiatinka12 (approved):
This makes the ownership rules much clearer. Documenting the three retirement mechanisms instead of implying a generic "supersede with status" approach better matches the schema and should help prevent invalid writes.

ClorindaM (approved):

PR #1408, opened 2026-08-06
eval: remove verdict-bearing lines from e2e rollup for blind grading

yinkid28:
## Summary
- Deletes recall summary, overall gate, and by-tag breakdown lines from `print_rollup` — all verdict-bearing, breaking blind grading (spec §7.4)
- Keeps compliance and cost/duration output
- Updates `e2e-test-spec.md` §4, §7.4, and §9 to reflect the new output and document accepted bias in pre-fix annotations

## Test plan
- [x] `test_rollup_does_not_leak_verdict` — new regression test asserts no `recall pass`, `overall gate`, or `by question_type` in output
- [x] `test_a_correct_but_noncompliant_run_reads_differently_from_a_wrong_one` — rewritten for compliance-only distinction (#972)
- [x] Full harness suite: 1616 passed, 1 skipped
- [ ] CI green

Closes #1114


clack391 (approved):

T-FEH (changes requested):
The `report.py` deletion itself is right, and I re-ran clack391's checks — the mutation pins the removal, no sibling in `run_e2e.py`/`corpus_report.py` leaks the verdict, and the `12 → 10` baseline is a genuine tightening (the ratchet drifts in both directions at `doc-links.test.ts:502-508`, so a loosening could not have gone green). Two things in the collateral, though.

**1. The spec now reverses a closed decision.** `e2e-test-spec.md:1364` —

You wrote: `the shadow-mode recency window calibration is the prerequisite.`
Restore: `Window calibration is not a pending task at all — see guardrail-enforcement-spec.md §7, "What the success gate can and cannot see."`

That is the sentence the diff deleted, and its replacement says the opposite of `guardrail-enforcement-spec.md:481` ("shadow-only **permanently**, and the reason is not that nobody has got round to the calibration") and of CLAUDE.md, which records the calibration issue closed `not planned` on 2026-08-09. The deleted sentence carried no issue ref — `#1282` was the neighbouring one — so the reversal bought nothing the task asked for. A spec is what the next reader trusts, and this one now points them at retired work.

**2. `-228` took coverage off code this PR doesn't touch.** The block from `# --- an ungradeable run must say WHY` onward tested `ungradeable_reason()`, `is_ungraded()` and `_GRADED_VERDICTS` — all still live in `run_e2e.py:75-89` and still called at `:152`. After this, nothing exercises them: the only other hit in the tree is `test_skill_gate.py:112`, a function *name* containing the substring, not a call. Two of the eleven matter beyond coverage arithmetic — `test_no_grade_is_ever_disclosed` is itself a §7.4 blindness guard, and `test_gradedness_does_not_import_the_committability_set` pins the invariant `run_e2e.py:69-75` documents. Restoring them verbatim passes on branch code. The by-tag and #1245 rendering tests were correctly deleted with their output.

Non-blocking: `KNOWN_VERDICTS` / `UNGRADED_VERDICTS` (`report.py:25,30`) are now dead — nothing references them but their own comment at `:22`, which still describes the bucket-seeding that's gone. And the three-exact-string guard is weaker than it reads: a reworded leak printing `verdict=pass` and `proof_quality` keeps all six tests green. #1114 did prescribe those strings, so that one's a judgment call rather than a miss.

Suites on the tree merged with main: harness **1939 passed**, engine vitest **2190 passed**, `test_e2e_report.py` 6 passed. All checks green.

Note for the re-push: clack391's approval was submitted against `83dd0e85` and `f62d06ac` landed after it, so the last-push gate is open regardless — this needs a fresh approval either way.

T-FEH (approved):
Clearing my changes-request. Both round-1 blockers are properly fixed, and I mutation-checked rather than eyeballed:

- The window-calibration sentence is restored and no longer contradicts `guardrail-enforcement-spec.md` §7.
- The lost coverage is back and real: widening `_GRADED_VERDICTS` fails `test_gradedness_does_not_import_the_committability_set`, and leaking a verdict into `ungradeable_reason` fails `test_no_grade_is_ever_disclosed`. `4d9502e5` was **right** to re-delete the two headline tests — those asserted output this PR removes, which my round-1 note had already conceded.

I resolved the conflict with `main` (`88799fc1`) and pushed three fixes (`f3f701fd`).

**The conflict.** #1570 landed on `main` and rewrote the same paragraphs to say `is_error` is no longer blind to a returned `{ok:false}`. I merged `main`'s now-true past tense into this branch's ref-free phrasing rather than importing its `#1282` mentions — naming the ticket twice would have pushed this file to 12 refs against the baseline of 10 this PR sets, i.e. undone the tightening to say something the prose can state without a number. Ratchet verified: 10 refs, 130 passed.

**The three fixes:**

1. **The calibration caveat named a date that hasn't happened.** It said "On 2026-08-05, blind grading enforcement removed those lines. Any `.ann.json` committed before that date was drawn with some form of the verdict on screen." That removal is *this* change — first commit 2026-08-06, still unmerged — so the roll-up is printing verdicts right now. `git log --diff-filter=A --since=2026-08-05 -- '*.ann.json'` returns **43 files**, and the caveat declares exactly those blind. That is the inverse of what it exists to do, and `calibrate_judge` reads it. The cut-off is now the merge, which stays true whenever it lands.

2. **`test_an_untagged_unrecognised_verdict_still_produces_output` could not fail.** `print_rollup` never reads `verdict` — it appears only in comments in `report.py` — so the verdict passed in is irrelevant and `"compliance:"` is unconditional. Swapping `"kinda-ok?"` for `"pass"` still passes. It also duplicated `test_rollup_always_states_compliance_even_when_clean`. Deleted.

3. **The historical-runs paragraph had dropped a settled claim** — main's "Retroactively scoring the corpus is not currently possible in a way worth trusting" plus the reason (nothing records the check version per run) — leaving text that reads as pending work. Same shape as the window-calibration reversal already fixed here, and it bought no ref reduction. Restored.

Suites: harness **2067 passed, 2 skipped** (one fewer, exactly the deletion), engine vitest **2255 passed / 93 files** with build+typecheck green, doc-links **130 passed**. All 12 checks green.

Two things left, neither worth another round:

- `report.py:22-30` — `KNOWN_VERDICTS` / `UNGRADED_VERDICTS` are now dead, and the comment above them still describes the bucket-seeding that went with the by-tag loop. Nothing references them but that comment.
- The regression guard pins three exact strings, so a reworded leak still slips: I confirmed that printing `verdict=pass` and `proof_quality` keeps all the tests green. #1114 did prescribe those three strings, so this is a known limit rather than a miss — an allow-list of permitted line prefixes would close it if it recurs.

@clack391 your approval is at `83dd0e85`, four commits back — worth a fresh look at the merge resolution in particular.

Issue #1355, opened 2026-08-05
Harden the encoding lint: dotted opens, false-positive escape hatch, cross-version parse, new-tree gate


clack391:
> **Reviewed 2026-08-24 before junior handoff.** Gap 1 respecified as a closed allow-list with its measured population, gap 4 replaced by its in-flight owner, and state plus blast-radius sections added.

**Touches:** eval/harness/tests/unit/test_encoding_lint.py

## Why

The `encoding="utf-8"` lint landed in #1285 (`test_encoding_lint.py`). It is correct on the
repo as it stands (0 offenders), but the adversarial review of that PR surfaced four latent
gaps. None has an instance today, so they were deliberately deferred rather than solved in
#1285 (fixing them now risks false positives with nothing to test against). Recording them so
the next contributor to touch the lint has the list.

## The gaps

1. **Dotted text-mode opens are not caught.** The matcher flags only a bare `open(...)`
   (`ast.Name`), so `Path(...).open("r")`, `io.open(p)`, and `builtins.open(p)` — all genuine
   bare text opens with the same cp1252 bug — are missed. This is #1285's *deliberate* scope
   (its `(?<![\w.])open` lookbehind excludes dotted opens too), chosen because `x.open` is
   ambiguous (`zipfile.open`, socket `.open`, `Image.open`).
   Widening must be a **closed allow-list of receiver forms**, never a deny-list of known
   non-file `.open` owners. Flag only: `Path(...)` / `PurePath(...)` called inline as the
   receiver, and `io.open` / `builtins.open`. Every other `.open` — including a bare
   variable's — is left alone. The variable form (`p.open("w")` where `p` came from
   somewhere else) is the *idiomatic* one and is statically undecidable; it stays uncaught
   after this change, and the module docstring must say so, or the next reader will assume
   the lint covers pathlib and it does not.

   Population today: **0** genuine dotted text opens repo-wide. The only three dotted
   `.open` calls are `webbrowser.open` and two `Image.open` — exactly the ambiguous ones
   the lint already excludes on purpose. Re-derive with an AST walk over `*.py` (skipping
   `.venv`, `node_modules`, `__pycache__`, `build`, `worktrees`) collecting
   `ast.Call` where `func` is an `ast.Attribute` with `attr == "open"`. So this is a
   preventive guard with nothing to fix — size the effort accordingly.

2. **Latent false positives, no suppression hatch.** A user-defined method also named
[...]

PR #1410, opened 2026-08-06
eval-harness: lint requiring encoding="utf-8" on text-mode file I/O

clack391:
## Summary

Normal tier. Issue #1285. This is the guard half of #1233: the sweep half already
landed in #1289 (the repo has 0 offenders today), so this PR adds no `encoding=` to
any source file. Its worth is the regression guard, not a fix.

- New pytest test `eval/harness/tests/unit/test_encoding_lint.py` AST-walks every `.py`
  in the repo and fails on any bare text-mode `read_text`/`write_text`/`open` (one
  missing an `encoding=` keyword). It parses rather than greps, so an `encoding=` on a
  later physical line is seen, not false-flagged.
- Widens the `eval-harness-tests.yml` path gate so that job actually runs when a scanned
  `.py` under `apps/server/` or `eval/triggering/` changes. Without this, a reintroduced
  bare `open()` in those trees would merge with the check green.
- Corrects the CLAUDE.md sentence that recommended a grep for finding offenders. That
  grep is the technique that let these accumulate.

## Review guidance

**Start here:** `test_encoding_lint.py`, the matcher in `_offenders_in` / `_is_binary_open`.
The call worth checking is the `open` rule: it matches a bare `open(` (an `ast.Name`) only,
so `Image.open` / `webbrowser.open` / `gzip.open` are excluded structurally. That is issue
#1285's chosen scope, and it also drops `Path(...).open()` / `io.open` (real offenders, 0
today); the docstring says so and the widening is deferred to #1355.

**Unsure about:** the PATTERNS form. The issue said extend with `^apps/`; I used the precise
`^apps/server/`, because all `apps/` Python lives under `apps/server/` (apps/web and
apps/electron hold no `.py`), so `^apps/server/` is identical lint coverage while `^apps/`
would fire the full build+pytest job on TS-only apps/web / apps/electron PRs for nothing.
One token to change back if you prefer the broad form. (Also note: the issue's acceptance
example `apps/server/app/agent/mock_agent.py` is already covered by the prior
`^apps/server/app/agent/`, so I proved requirement (e) against a still-uncovered path,
`eval/triggering/scripts/run_loop.py`, instead.)

## Plan

**Didn't change:** the 0 existing offenders (nothing to sweep, #1289 did it), and the
`eval/harness/pyproject.toml` ruff config. Ruff's PLW1514 is type-directed and finds none of
these (it fires only when it can infer a `pathlib.Path` receiver, and these are untyped
`tmp_path` fixtures), so adding it is a green check that guards nothing. The config stays
F821-only.

**Acceptance check:** `eval/harness/tests/unit/test_encoding_lint.py::test_no_bare_text_mode_file_io`.
It does not exist on `main` and passes on this branch. Proven as a real guard by mutation:
injecting a bare `write_text("x")` into a scanned file makes it fail naming that site,
removing it passes. Requirement (e) proven by running the workflow's own
`printf ... | grep -qE "$PATTERNS"` against `eval/triggering/scripts/run_loop.py`: no match
on the old pattern (job skipped, no pytest), match on the new one (job runs).

**Deviated from the plan:** two round-1 corrections, from `/critique-plan` and the bug
review, both in the branch. (1) The unparseable-file skip is surfaced with `warnings.warn`,
not `print`: pytest captures stdout and shows it only on failure, so a print would hide a
skipped file on the green runs where it matters. (2) The `apps/server` tree the widening now gates targets Python >=3.12 while the harness parses under 3.11. Every apps/server file parses under 3.11 today, so apps/server IS scanned now; `ast.parse` is wrapped in `try/except` purely as a defensive guard, so a future file using 3.12-only syntax is skipped-with-a-warning rather than false-redding the lint. It does not mean apps/server is currently unscanned.

Step-4 stop rule: not hit. No schema, auth, plugin-agent binding, or ADR change. The
CLAUDE.md edit corrects a how-to-find recommendation, not the rule itself. (That correction
was routed to this PR by the #1233 body, its task-reviewer note, and the #1289 body, even
though #1285's `Touches:` line omitted it.)

## Test plan

- [x] `make test-all` passed in a single clean end-to-end run (exit 0): typecheck, JS 245,
      control-plane pytest 135, MCP engine vitest 2036, eval app 148, eval harness 1600. An
      earlier run surfaced two failures I resolved first: the `doc-links.test.ts` issue-ref
      ratchet, which my CLAUDE.md edit tripped (fixed by citing the lint file, not the ticket
      number), and a transient `eval/app` npm-install corruption from flaky network (re-runs
      clean, unrelated to this diff).
- [x] No skill run-log snapshot changed (this PR touches only `eval/harness/tests/`, a CI
      workflow, and CLAUDE.md), so no `make eval-skill` run is needed. `build_snapshot` hashes
      `eval/tests/unit/`, not `eval/harness/tests/`, so this PR stales no run log: no eval cost.

The eval-harness job this PR lives in and modifies is green locally: `ruff check .` clean and
`pytest` 1600 passed, the new test included.

## Follow-on issues

- #1355 harden the encoding lint: dotted opens (`Path.open`/`io.open`/`builtins.open`), a
  false-positive escape hatch, cross-version parsing for the 3.11/3.12 gap, and the gate for a
  future new top-level tree. All 0-instances today, deferred deliberately.

Closes #1285.


kofiatinka12 (approved):
This is a solid improvement. Adding the AST-based encoding lint gives us much stronger protection than relying only on documentation, and the implementation choices make sense. I especially like that it avoids the common grep-based false positives/negatives by analyzing the Python syntax tree directly.

The CI path updates also look well thought out and ensure the check actually runs where needed. The comments and documentation updates make the reasoning clear for future maintainers.

Thanks for putting this together

eval/harness/tests/unit/test_encoding_lint.py
```diff
@@ -0,0 +1,119 @@
...
+    if len(call.args) >= 2 and isinstance(call.args[1], ast.Constant):
+        mode = call.args[1].value
+    for kw in call.keywords:
+        if kw.arg == "mode" and isinstance(kw.value, ast.Constant):
+            mode = kw.value.value
+    return isinstance(mode, str) and "b" in mode
+
+
+def _has_encoding(call: ast.Call) -> bool:
+    return any(kw.arg == "encoding" for kw in call.keywords)
```
jud-sdev: `_has_encoding` checks only that an encoding keyword is present, not that its value is real, so `read_text(encoding=None)` and `open(f, encoding=None)` pass the lint. But `encoding=None` is the platform default (cp1252 on Windows), which is the exact bug this lint exists to catch. Zero instances today, but it is a false negative in the check's core purpose, and the fix is one line: also reject a literal `encoding=None`. For example:

```python
return any(
    kw.arg == "encoding"
    and not (isinstance(kw.value, ast.Constant) and kw.value.value is None)
    for kw in call.keywords
)
```

Small enough to fold into this PR rather than defer to #1355.

jud-sdev (changes requested):
Senior review. The lint is solid overall. I verified it by mutation (a bare open() dropped into a scanned file fails the test and names the site, removing it passes), confirmed it is self-consistent (its own reader passes encoding, and the method names appear only as string literals so it does not flag itself), and checked that the PATTERNS gate covers every .py tree in the repo, so an offender cannot merge on an ungated path. I also parsed all 41 apps/server files under an actual Python 3.11 and none fail, so apps/server is genuinely scanned today.

One small change worth folding in before merge, inline below: `_has_encoding` treats `encoding=None` as satisfying the rule, but that is the platform default this lint exists to catch. One line closes it.

One non-blocking note: the body and the test's inline comment say apps/server is skipped under 3.11 because it targets 3.12. It is not skipped, all 41 files parse under 3.11, so apps/server is covered today and the PR is stronger than it claims. The try/except skip stays as correct defensive code for a future 3.12-syntax file. Worth rewording so the next reader does not think apps/server is unscanned.

Once the encoding=None one-liner is in and CI is green, this is good to merge.

jud-sdev (approved):
The encoding=None gap is fixed exactly as suggested.  _has_encoding now rejects a literal encoding=None, and there's a test asserting both read_text(encoding=None) and open(f, encoding=None) are flagged. I ran the lint on the fixed branch under Python 3.11, the version CI uses, and it passes. CI is green.

PR #1411, opened 2026-08-06
record-extraction: classification examples, spawn-failure clause, value matchers (#1131, #995, #1108)

kofiatinka12:
Closes #1131, #995, #1108. Batched because all three flip record-extraction's run log — landing them together is one paid `--skill` run + one annotation pass, per Dallan's note on #1131/#995.

## #1131 — classification recovery + spawn-failure clause
- **record-extractor.md**: three point-of-use **worked examples** (examples, not duplicate rules) at the drop sites — the blank-columns rule, `informant_proximity` (leads with *proximity follows the record type*), and `evidence_type` (same fact, different record). +~1.9 KB, warn-only prompt budget.
- **record-extraction/SKILL.md**: the batched spawn-failure carve-out, trimmed to a single exception clause — a `record-extractor` spawn failure is reported and the run stops, resolving the contradiction PR #1196's deny text created with "unextracted records = keep going".

## #995 — value/classification matchers on the highest-confidence tests
Added `expected_classifications` to the 7 highest-confidence matcher-less tests (evidence_type backbone + stated values + sex + relationship direction; `optional` on role-ambiguous entries so a role miss skips rather than false-fails). Scoped to **ut_013 + high-confidence only** per genealogist direction. ut_013 now pins `child_2` birthplace = "Ireland" — the exact leak from PR #994.

## #1108 — the enabler #995's value matchers require
`value` support added to **both** `unit-test.schema.json` mirrors, the validator comparison loop (case-insensitive substring on the attribute-relevant field — `place`/`standard_place`, `date`, else `value`), a meta-test (asserts the ut_013 leak fails where evidence_type alone passes, and place standardization is tolerated), and the §5.10 spec prose. Without this the `value` matchers can't even load (`additionalProperties: false`), so #995's headline is blocked on it.

## Eval run + annotation (committed)
`make eval-skill SKILL=record-extraction` against these edits → `v1_2026-08-06_10-04-55` (27 tests: **18 pass / 4 partial / 5 fail**), fully annotated (123 corrections, every dimension; annotator on-branch).

**The matchers behaved — zero false fails.** ut_013 (crux) and ut_020 (sex) pass. Two matchers failed as **true positives**, catching real skill errors:
- **ut_010** — marriage parent-names classified `indirect`; record + doctrine say `direct`.
- **ut_014** — head-of-household came back the *miner*; the record's head is Silas Kerrigan the *laborer* (mis-roled differently-surnamed head — exactly what the fixture tests).

**Finding for reviewers:** 4 of the 5 fails are one class — parent-name `evidence_type` / party `informant_proximity` on marriage/burial records (ut_010, ut_024, ut_025, ut_021). ut_024/021/025 are **pre-existing** matcher tests (no `value` fields — #1108 is confirmed inert for them), so this is a real sonnet-4-6 weakness in derived/index parent-name calls that this PR's examples did **not** target. Whether to add one more worked example for it is a follow-up (needs its own paid run), deliberately out of this batch.

## Verification
- All 7 edited test files valid against the updated schema; schema mirrors byte-identical.
- Validator + orchestrator meta-tests pass (72); full harness unit suite 1599 passed, 1 skipped.
- Annotation completeness: 27 tests, 123 corrections, 0 missing dimensions (runlog gate rule 3).


mercyokum (approved):
The PR looks good. Approving

Issue #1315, opened 2026-08-05
Wiki content is presented without its source URL


DallanQ:
**Touches:** packages/engine/mcp-server/src/tools/wiki-read.ts, packages/engine/mcp-server/src/tools/wiki-search.ts, packages/engine/mcp-server/src/tools/wiki-place-page.ts

Tester: *"Gave information from the Wiki without providing a link"* → *"Should provide the URL."*

Pairs with the genealogist issue on locality facts asserted from model memory: if wiki-derived claims always carry their URL, a wrong locality claim becomes checkable by the user instead of being taken on faith.

**Acceptance:** wiki tool results carry the page URL, and narration cites it when using wiki content.

**Evidence:** `feedback-2026-08-03T20-27-21-009306Z.zip`

PR #1216, opened 2026-08-03
Lint: forbid GPS "primary/secondary source" terminology in plugin prose (#907)

Gennecis:
# Lint: forbid GPS "primary/secondary source" terminology in plugin prose (#907)

Closes #907 

## Summary

- GPS doctrine treats source classification (`original|derivative|authored`) and information quality (`primary|secondary|indeterminate`) as two independent closed enums; "primary source"/"secondary source" collapses them into one imprecise phrase, and that phrasing had already leaked into plugin prose the agent reads at runtime.
- Added a CI lint (`gps-terminology.test.ts`) that forbids the phrase (case-insensitive, singular/plural) across plugin agents/skills/references, with a reasoned allow-list for `citation/SKILL.md`'s existing correction guardrail.
- Fixed the four sites the lint's scan found outside that guardrail so the lint lands green.
- Tier: **Normal** (adds new lint logic, not a one-line fix with an obvious cause).

## Review guidance

**Start here:** `packages/engine/mcp-server/tests/packaging/gps-terminology.test.ts` — specifically the `FORBIDDEN` regex and the `ALLOWLIST` array. That's where the actual judgment calls live: what counts as a violation, and why the two `citation/SKILL.md` lines are exempted rather than fixed. Please check the allow-list reasoning holds up, not just that the lint passes.

**Unsure about:** Nothing load-bearing. The four rewords are wording judgment calls I made without asking (e.g. `init-project/SKILL.md`'s "(primary source being surveyed)" → "(the source being surveyed)", `locality-guide/SKILL.md`'s → "main source") — worth a glance to confirm the replacement wording reads naturally in context, but none of them change what the skill does.

## Plan

Issue #907 proposes a deterministic CI lint, following the two existing drift-lint siblings in `packages/engine/mcp-server/tests/packaging/` (`skill-guidance.test.ts`'s reasoned-allowlist pattern, `enum-drift.test.ts`'s doctrine-vs-prose scan), plus fixing the four non-`citation` sites where the forbidden phrasing appears, so the lint lands green in the same PR. No schema/validator/CI-machinery changes.

Confirmed violations (full repo grep reproduced and matches the issue's table — 6 hits, 4 files):

| File:line | Current text | Disposition |
|---|---|---|
| `skills/historical-context/references/historical-broad-context.md:112` | "church records are the primary source for vital events" | Fix → `original source` |
[...]

Issue #23, opened 2026-04-30
GedcomX simplifier functions


DallanQ:
FamilySearch represents genealogical data in a format called GedcomX. It is very verbose. We want the LLM to receive and send a data in a "simplified" GedcomX format. 

We need two functions (in typescript). The MCP servers will call these functions as needed to convert between the two formats:
1. GedcomX (from tool) -> SimplifiedGedcomX (for LLM)
2. SimplifiedGedcomX (from LLM) -> GedcomX (for tool)

Here are the simplification rules:

- URI prefixes dropped: Birth not http://gedcomx.org/Birth. Saves tokens, costs nothing in clarity.
- nameForms and parts collapsed: given and surname directly on the name. The deeper nesting only matters for multi-script names you probably won’t encounter.
- preferred on names and primary on facts as explicit booleans. The redundancy with first-position convention is intentional — both signals point the same way, so the model is less likely to miss it.
- parent/child instead of person1/person2 for ParentChild. The semantic role is clearer than the positional naming.
- Couple keeps person1/person2 because the relationship is symmetric.
- quality field on sources, mimicking GEDCOM’s QUAY. This is the single most useful piece of metadata for discrepancy reasoning, and I’d encourage you to populate it even if your underlying GedcomX doesn’t have it natively.
- IDs on names and facts, not just persons. Lets the LLM (or your application) refer to “fact F1” or “name N2" unambiguously when generating research recommendations.

And an example:

GedcomX

```
{
  "persons": [
    {
      "id": "I1",
      "gender": { "type": "http://gedcomx.org/Male" },
      "names": [
        {
          "id": "N1",
          "preferred": true,
          "type": "http://gedcomx.org/BirthName",
          "nameForms": [
            {
              "fullText": "John Doe",
              "parts": [
                { "type": "http://gedcomx.org/Given", "value": "John" },
                { "type": "http://gedcomx.org/Surname", "value": "Doe" }
              ]
            }
          ],
          "sources": [
            {
              "description": "#S1",
              "qualifiers": [
                { "name": "http://gedcomx.org/CitationDetail", "value": "1920 Census, Denver, ED 47" }
              ]
            }
          ]
        },
        {
          "id": "N2",
          "type": "http://gedcomx.org/AlsoKnownAs",
          "nameForms": [
            {
              "fullText": "Jonathan Doe",
              "parts": [
                { "type": "http://gedcomx.org/Given", "value": "Jonathan" },
[...]

PR #1414, opened 2026-08-06
locality-guide: cite the wiki page, name the registration level, search across the state line (#1321)

Emruthwill:
## Summary

Tier: **Normal**.

- Items 1 and 2 of #1321, both corrections from robert.raymond@familysearch.org: `locality-guide` asserted "Massachusetts didn't require vital registration until 1841" from model memory, and reported a border as "Bounded by New York (west) and Vermont (north)" without telling the researcher to search across it.
- Three rules added/rewritten in `locality-guide/SKILL.md`: a registration date must name its **level** (town/parish, county, statewide); "Cite the wiki" now asks for the page **URL**, not just the article title; adjoining jurisdictions across a state line are searched at **both** town and county level.
- New unit test `ut_locality_guide_026` that goes red on `main`.
- Items 3 and 4 of #1321 split out to #1412 and #1413 — they are `research-plan` and e2e-fixture work, and the issue's single acceptance criterion did not cover them. #1321's body now carries per-item acceptance so it cannot close on one of them.

## Review guidance

**Start here:** the third bullet of the new "Never state a registration date without naming its level" rule in `locality-guide/SKILL.md`, and the `judge_context` of the new test. The judgement call is that **naming adjoining jurisdictions is a search strategy, not a record-availability claim**, so it is deliberately *not* subject to the new URL-citation rule — otherwise the two rules I added in the same commit would contradict each other, since no fixture supplies which counties adjoin Berkshire. The test grades the cross-border *instruction*, not recall of county names. If you think geography should also require a citation, that changes both the prose and the test.

Second thing worth a look: I put the Massachusetts 1639/1841 pair **into the prompt** as a worked example, which is in tension with the rule it illustrates ("don't assert registration dates from memory"). I judged it worth the cost because the distinction is the whole lesson and a reader who only sees "cite your source" will still swap 1841 for 1639. Say so if you'd rather it be levels-only with no years.

**Unsure about:** whether the registration-level rule generalises cleanly outside New England. It is written jurisdiction-neutrally ("town/parish, county, statewide") with Massachusetts as the example, but the only evidence behind it is the one Massachusetts correction.

## Plan

**Didn't change:** the three wiki MCP tools. The paired developer issue #1315 was written as though the URL needed to be *made* available; it already is, in all three — `source_url` in `src/types/wiki-search.ts`, `url` from `wiki-read.ts` and `wiki-place-page.ts` — and `src/index.ts` `JSON.stringify`s the whole result, so nothing strips it. `pages_read[].url` already persists it too. This PR is purely the narration half. **#1315 should be verified against a live call and then closed or rescoped**, not implemented.

Also didn't change: `locality-guide/SKILL.md:41` ("Ground every claim in tool output… Name only the… dates… that actually appear in a tool result"). Every defect in #1321 is the agent *not complying* with a rule already there, which is why this PR carries a test rather than prose alone.

**Acceptance check:** `ut_locality_guide_026` (`eval/tests/unit/locality-guide/massachusetts-pre-1841-town-registration.json`). It is adversarial by construction: the catch-all wiki fixtures state only the generic "many jurisdictions did not require civil registration until the late 1800s or early 1900s" and say nothing about Massachusetts, so silently converting that into a specific Massachusetts claim is the failure it catches. On `main` the skill has no rule requiring a registration level or a URL, so the run fails those dimensions.

## Test plan

  **Ran it locally; it fails here but passes in CI — the local failures are this checkout, not the repo.**

  CI is green on every suite this gate covers: `pytest`, `server-pytest`, `vitest`, `pnpm workspace`, `lockfile-drift`, `eval CRUD UI`. The three failures `scripts/test.sh` produced on my machine — `web#typecheck` (`Cannot find module '@genealogy/viewer-ui'` plus `TS7016` on `react/jsx-runtime`), `@genealogy/electron#test` (`expect-type` missing `./overloads`), and `apps/server` pytest — are a broken local `node_modules`, cleared by a reinstall. Nothing repo-side to fix, and no issue filed.

  Two retractions from earlier revisions of this body, kept so the history is legible: it once said the gate "stalls before the first suite" (wrong — the command was piped through `tail`, which withholds output until exit), and it once warned that CI had never run (also wrong — the checks were delayed by hours and landed at 23:16Z; #1424 was filed on that mistake and is closed).

  What I ran directly, all passing:
  - `pytest tests/unit/test_loader.py test_check_runlogs.py test_check_negative_reciprocity.py test_check_rubric_tool_drift.py test_fixtures.py` — 97 passed
  - `scripts/check_skill_frontmatter.py` — 27 skills + 4 agents clean
  - the new test validates against `docs/specs/schemas/unit-test.schema.json`; all 26 `locality-guide` tests load through `harness.loader`; no missing fixtures; no duplicate ids

- [ ] For every skill whose **run-log snapshot** I changed … I ran `make eval-skill SKILL=<name>` and committed the run log **and its `.ann.json`**.

  **Run done and committed** (`939338d75`): `v1_2026-08-06_21-20-18.json` — 26 tests, **22 pass / 4 partial / 0 fail**, $8.4963, 2363s wall. The staleness violation is cleared; `check_runlogs` now reports only the missing `.ann.json`, which is the genealogist annotation pass and is **still outstanding**.

  **CI result: 14 checks, 13 pass, 1 fail** — and the single failure is this gate:

  ```
  ##[error] skill `locality-guide`: latest run log `v1_2026-08-06_21-20-18.json`
  has no annotation file (`v1_2026-08-06_21-20-18.ann.json` missing).
  ```

  Rule 2 (staleness) **passes** — the run log is active against this branch. So the annotation is the only thing between this PR and a green board, and it must be written in the eval CRUD UI (`eval/CLAUDE.md` forbids hand-writing a unit `.ann.json`). Blocked locally by #1420 until OneDrive sync is paused or `.next` excluded.

  The empty commit `86c26a908` exists solely to test the `synchronize` trigger and should be dropped in a rebase before merge.

  The prune removed the oldest candidate beyond the newest five, which is why a `.ann.json` deletion rides along in that commit.

  ### The new test

  `ut_locality_guide_026` scores **3/3 on all five content dimensions** — Correctness, Completeness, Jurisdiction accuracy, Record availability, Research strategy. The judge confirms both #1321 defects are fixed, unprompted:

  > "correctly identifies the two-tier Massachusetts vital registration system (town-level from 1639, statewide from 1841) with proper attribution to the General Court mandate"

  > "correctly identifies the borders with New York (Columbia and Rensselaer counties) and Vermont (Bennington County)" … "NY border families need church records because NY had no functional vital registration until 1880"

  > "notes that the `wiki_place_page` returned generic guidance rather than Massachusetts-specific content"

  That third quote is the citation rule's fallback working as designed: the catch-all fixture offers only a generic "late 1800s or early 1900s" line, and the skill declined to convert it into a Massachusetts claim.

  Its one `partial` is **Tool Arguments 2/3, and it is a hole I authored, not a skill defect** — the skill called `place_search({placeName: "Berkshire County", contextName: "Massachusetts"})`, which misses the `~Massachusetts` substring predicate on `place-search-massachusetts` (`~` matches the named arg only; `contextName` is not consulted), so the call returned `fixture_not_found`. The skill recovered by searching "Massachusetts" and graded correct. **Deliberately not fixed here:** a fixture must be listed in the test's `mcp_fixtures[]`, so fixing it edits the test JSON, which re-stales this run log and costs another full paid run for one dimension of one test whose substance already passed. Filed as **#1415**.

  ### Only 4 of 150 dimension entries scored below 3

  | Test | Dimension | 07-30 → 08-06 | Judge's stated reason |
  |---|---|---|---|
  | 026 | Tool Arguments | *(new)* → 2 | my fixture gap (#1415) |
  | 007 | Correctness | 3 → 2 | "fabricated the Wikipedia source attribution" for 1850/1860 population figures the tool never returned |
  | 023 | Record availability | 3 → 2 | penalised for honestly hedging — see #1416 |
  | 022 | Correctness | 3 → 2 | didn't cite the anthracite index its fixture supplied |

  **Read 022 before concluding this diff caused a regression.** It was `partial` in *both* runs — but on **Tool Arguments (2 → 3)** before and **Correctness (3 → 2)** now. A dimension recovered while a different one broke, net outcome unchanged. Nothing in this diff touches tool arguments, so that pair is judge variance, and it sets the noise floor for reading 007 and 023 off a single run each.

  **#023 is a genuine design collision and is filed, not patched: #1416.** It was marked down for saying "digitized holdings were not confirmed by tools" — while the judge explicitly acknowledged "the skill honestly flags these gaps." The rubric's *Record availability* dimension rewards specificity that, when the tools return nothing, can only come from model memory. That is the exact defect #1321 was filed about, so as written a skill cannot satisfy both. Lane 2 per `docs/skill-lifecycle.md` §5 (eval defect, not prose defect), which is why the prose was left alone. #1416 carries three candidate resolutions, including "the judge over-read it and 023 should have stayed a pass" — live until the annotation rules it out.

  **007 may share one root with 023.** It failed the *inverse* way: attaching a citation it had not earned. If the citation push made the skill both hedgier where tools are silent and looser about attribution, that is one calibration problem, not two. Noted on #1416.

  ### For the reviewer

  I did not write the `.ann.json`. 146 of 150 entries are 3/3 confirmations; the four above are the judgement calls, and 023 and 007 are the two that decide #1416. A model grading its own prose change is the loop that annotation exists to break, so those verdicts are the genealogist's.

- [ ] If I changed a skill's `description` frontmatter or its DO NOT clauses …

  **N/A** — no frontmatter or DO NOT clause changed; the edits are all in the body's "Important rules". `check_skill_frontmatter` clean.

## Follow-on issues

- #1412 — `research-plan`: size an indirect-record window to the record creator's lifespan (item 3 of #1321, carrying its ⚠️ reproduce-first warning, since the quote came from a thinking block)
- #1413 — e2e fixture: a natural parent or stepfather named guardian implies a prior marriage (item 4 of #1321)
- #1415 — the `place_search({placeName: "<County>", contextName: "<State>"})` fixture gap behind 026's Tool Arguments score. Worth folding into whatever run touches this suite next rather than triggering its own; the note there flags that every `~<State>` fixture in `eval/fixtures/mcp/` has the same shape.
- #1416 — the *Record availability* rubric dimension vs. the honesty this PR requires. **Needs a genealogist decision, and the `.ann.json` verdicts on 023 and 007 are its evidence.**
- #1315 — needs verifying and closing/rescoping rather than implementing, per "Didn't change" above

**Not blocking, noticed in passing:** this skill's run log was already stale on `main` before this branch existed — `flynn-first-plan/research.json` was changed by #1022 on 2026-07-31 and `locality-guide/SKILL.md` by #1216 on 2026-08-05, both after the 2026-07-30 log. That is the baseline #1094 records ("20 of 25 skills' latest run logs are already stale on a fixture path on main," which names #1022). Flagging it so the re-run here isn't read as having fixed someone else's staleness.

Closes #1321.


Cia-3 (approved):
Reviewed the diff, verified the load-bearing claims against actual code,
and checked the annotation that was still outstanding when this PR was
opened.

**Verified independently:**
- #1315 is based on a false premise, as claimed — checked the source
  directly: `wiki-search.ts` has `source_url`, `wiki-read.ts` returns
  `{ url }`, `wiki-place-page.ts` has both. It already exists; #1315
  should be closed/rescoped, not implemented.
- The historical claim is accurate: MA town clerks were ordered to
  record vital events in 1639, statewide registration began 1841 —
  correct, not invented for the test.
- The new test's `judge_context` correctly separates the graded
  behavior (cross-border search instruction) from ungradable specifics
  (county names the fixtures don't supply) — holds up under scrutiny.

**Update since this PR was opened:** the annotation is no longer
outstanding. CI is now 13/13 non-skipped checks green, and the
`.ann.json` (150 corrections, real annotator, not hand-written) is in
the diff. This is merge-ready now.

Read the two contentious verdicts behind #1416 since they're now
available:
- Test 007's Correctness (3→2) — genealogist did not override the
  judge; confirmed real fabrication (invented CA population figures
  attributed to a tool, overstated volume counts). Pre-existing, not
  caused by this diff.
- Test 023's Record availability (3→2) — also not overridden, but the
  genealogist's stated reason differs from the judge's: not "punished
  for honesty" but "doesn't distinguish online vs. onsite specificity."
  Useful reframe for #1416 — narrower problem than originally worried.

**Minor, non-blocking:** drop the empty `86c26a908` commit (testing the
`synchronize` trigger) in a rebase before merging — already flagged by
the author.

**Recommendation: approve.** Small, well-targeted, factually verified
change; adversarial test passing; annotation complete; every open
question already has its own tracked issue rather than sitting
ambiguous in this PR.

eval/tests/unit/locality-guide/massachusetts-pre-1841-town-registration.json
```diff
@@ -0,0 +1,48 @@
...
+    "Must direct the researcher to town-level records (town clerk vital registers, town meeting records) for the 1820-1850 window, not only to county-level or state-level series. Massachusetts created records at both levels and the town series is the one a county-only survey misses.",
+    "Must recommend searching adjoining jurisdictions ACROSS the state line — the adjoining counties and, where relevant, the towns on the New York and Vermont sides — not merely note that Massachusetts is bordered by New York and Vermont. Naming the bordering states alone is the reported defect; the credit is for telling the researcher to search the neighbouring towns/counties over the border, at both town and county level. Do not require specific county names the tool output does not supply — the graded behavior is the cross-border search instruction, not recall of which counties adjoin.",
```
ClorindaM, line 43: Is there a reason the other border states were not mentioned here as well? Connecticut, Rhode Island, and New Hampshire also border Massachusetts. Will the skill limit itself to just the two mentioned or follow the same rules for the other three when dealing with cross-border searching? Or is the reason that New York and Vermont are the most likely to have similar records down to the town level?

Emruthwill: Good catch — Connecticut is a real omission on my part. Berkshire borders three states: Vermont (Bennington) north, New York (Columbia/Rensselaer) west, and Connecticut (Litchfield) south.

The skill isn't limited to two, though. The rule I added to SKILL.md names no states — Massachusetts and New York appear only as the example of why the rule exists. So a Springfield question gets Connecticut, a Newburyport one gets New Hampshire.

This test names NY and VT because the scenario does — the prompt says the family's town is "close to both the New York and Vermont lines." My judge_context follows the scenario but never says so, which is why it reads as the complete set for the county. That's my wording, not the skill's behaviour.

On your hypothesis, it's actually the reverse of what I'd have guessed too. Town-level records don't single out NY and VT — Vermont, Connecticut, Rhode Island and New Hampshire all keep vital records at the town level, same as Massachusetts. New York is the anomaly: records sit at both town and county level, which is exactly why robert.raymond's correction named it. The rule matters more on the NY border, not less.

Fix I'd make: scope that bullet to "the borders this family lives near," and state that Connecticut is out of scope for this scenario rather than overlooked. Editing the test re-stales the run log, so I'd rather fold it into the re-run #1415 already needs than spend $8.50 on a wording change alone — unless you'd prefer it explicit before merge.

One more thing worth adding at some point: a Berkshire question that names no border, to check the skill finds all three itself instead of echoing whichever ones I hand it. This test can't tell those apart. I'll file it if you agree.

EdmondOware (approved):
I read Clorinda's question about whether the cross-border rule is limited to just New York and Vermont, and pulled the actual diff to check independently rather than taking Ruth's explanation on faith. It holds up. The SKILL.md rule itself names no specific states, it's general guidance to search adjoining jurisdictions across any state line, with Massachusetts and New York appearing only as the worked example from the original bug report.

 The historical claim is accurate and well-documented: Massachusetts town clerks were ordered to record vital events in 1639, statewide registration began in 1841, and treating 1841 as the start of all vital registration erases over 165 years of town-level records. Collins already verified this independently, and I agree it reads as researched, not invented for the test.

 The new test is carefully scoped too. It explicitly does not require the model to recall which specific counties adjoin the border, only that it has the instinct to search across one, so it grades the right behavior rather than an exhaustive geography quiz. It also guards against a subtle failure, silently converting generic wiki boilerplate into a specific, sourced-sounding Massachusetts claim, which is exactly the kind of fabrication this project treats seriously.

Issue #1357, opened 2026-08-05
Refuse tree_forget once a project has a plan — forget-and-rederive is a project-start exercise


mercyokum:
> **Reviewed 2026-09-14 before junior handoff.** No decision needed. The one way this goes
> green and wrong is narrowing the condition to an *active* plan — the framing the ruling
> rejected, which every test written alongside it would pass. Four missing edit sites and one
> dead instruction are corrected at the bottom.

> **The 2026-08-31 `reviewed` pass predates this shape and the label has been removed.**
> That review vetted a `question-selection` prose change; the ruling below made this an
> MCP-server precondition instead. It needs a fresh `/review-ready` pass before a junior
> takes it. The lane label changed with it — `genealogist` to `developer`.

> **Rescoped 2026-08-31 (lead). This is no longer a `question-selection` change.**
> The §1a collision is being removed at its source: `tree_forget` will refuse to run
> once the project has a plan. Blast radius, paid runs and `Touches:` all changed —
> read this block, not the original issue below, which is kept for the session
> evidence only. The blocking header in the original is also dead: issue #1668 closed
> 2026-08-27 via PR #1963, and that dive did not reproduce this finding.

**Touches:** packages/engine/mcp-server/src/tools/tree-forget.ts, packages/engine/mcp-server/tests/tools/tree-forget.test.ts, docs/specs/tree-forget-tool-spec.md, packages/engine/plugin/skills/forget-and-rederive/SKILL.md

## Decision (lead, 2026-08-31) — forget-and-rederive is a project-start exercise

`forget-and-rederive` should only run at the start of a research project, before a
plan is built. Make that a **`tree_forget` precondition**: refuse the call when
`research.json` already holds a plan, and say so in the error — the researcher can
start a new project for the exercise.

**Why this closes the card instead of answering it.** The question this issue was
holding — what a §1a exception fires on, given that `tree_forget` writes nothing to
`research.json` and reading the tree is forbidden — only exists because the forget
can happen mid-plan. Once it cannot, there are no `in_progress` plan items for §1a
to block on, the exception is unnecessary, and the contradiction disappears rather
than being anchored.

**Verified 2026-08-31, and this is a change rather than a description of today.**
The skill is purely user-request-triggered — its `description` lists trigger phrases
plus one project-seeding case, and nothing anywhere gates it on project phase. The
reported session was mid-project: the project already held an open question about
further children of a known spouse, with `in_progress` plan items, when the
[...]

PR #1419, opened 2026-08-06
tree_forget: parents-of and spouses-of strip the subject's redundant facts (#1314, #1417)

Asimi1234:
## Summary

- **Normal tier.** `tree_forget`'s relative selectors now also strip the subject's own documentary facts that redundantly carry a forgotten conclusion, so the answer does not survive as a fact on the subject:
  - `parents-of` also strips the subject's own `Parents`-type facts (#1314).
  - `spouses-of` also strips the subject's own `Marriage`/`Divorce`/`Annulment`-type facts (#1417) — the person-level echo of the facts the `Couple` relationship also carries.
- FamilySearch carries a conclusion twice — as graph structure **and** as a documentary fact on the subject's own record (`person_read` returns both). The selectors previously removed only the structure; the fact leaked the answer, and could be the *sole* carrier when the relatives were never added as tree persons.
- Also makes both selectors succeed (instead of erroring "matched nothing") when only such a fact is present — the actual shape of the reported case.

Closes #1314. Closes #1417.

> **Branch history:** the `spouses-of` half (#1417) was reviewed and merged as #1425, which was stacked on this branch (`1314-tree-forget-parentage-evidence`) rather than `main` because it extends the same `resolveSelectors` block. This PR now carries both fixes; landing it on `main` closes both issues, and nothing separate needs to merge for #1417.

## Review guidance

**Start here:** `src/tools/tree-forget.ts`, the relative-selector arm of `resolveSelectors`. The decision worth checking: the sweep matches facts by **type** (via the existing `factIdsOfType`/`factIdsOfTypes`), never by the relatives' names in the fact `value` — that's what keeps it inside the tool's "structural, never by value" contract, and why swept facts surface only as `factsByType: { Parents: n }` / `{ Marriage: n }`. Scoped deliberately to `parents-of` and `spouses-of`.

**Unsure about:** Nothing blocking. The one judgement call — keep this in the tool vs. push it entirely into skill doctrine — is written up on #1314; the tool is the right home because the fact can be the sole carrier of the answer.

## Plan

**Didn't change:** `children-of` — the redundant `Parents` fact lives on the child, which `children-of` removes wholesale, so there is no leak.

**Acceptance check:** `tests/tools/tree-forget.test.ts`:
- `parents-of` (#1314) — "parents-of also strips the subject's own `Parents` documentary fact", "parents-of succeeds on a `Parents` fact alone, with no parent relationship", and a pin that it still errors when there is neither a link nor a `Parents` fact.
- `spouses-of` (#1417) — "spouses-of also strips the subject's own person-level `Marriage` fact", "spouses-of succeeds on a person-level `Marriage` fact alone, with no couple relationship", "spouses-of also sweeps person-level `Divorce` and `Annulment` facts", and a pin that it still errors when there is neither a spouse link nor a marriage-class fact.

Each new fact-sweep test fails on `main` (the sweep didn't exist; the sole-fact case errored "matched nothing") and passes on this branch.

## Test plan

- [x] `./scripts/test.sh` (= `make test-all`). Engine vitest 2096/2096, typecheck, harness pytest, control-plane, eval app all pass. One unrelated failure surfaced in `apps/electron` (`feedback.test.ts › drops the largest files … over the archive budget`) — a pre-existing flake in a package this diff does not touch: it passes 19/19 when that file runs on its own, and only fails under full parallel-suite load.

## Follow-on issues

- #1422 — route forget-for-testing setup through `tree_forget` rather than a hand-omit in `init-project` (the actual site of this report; ties into #1357).


T-FEH (changes requested):
The `spouses-of` half is solid and I confirmed it independently — 147 person-level `Marriage` facts across committed trees, and reverting only the tool fails exactly the 5 sweep tests while the 2 error-pins correctly still pass. The scope reasoning is right too: `ForgetSelector` has no per-relative field, so `spouses-of` already removes *all* spouses, and a type-matched sweep can't catch a fact about a relative the selector didn't name. No over-deletion risk.

**Blocking: the `Parents` question is still open, and the provenance is the reason.** @clack391 and @francis-2008-happy both asked for the literal type; `dd7c73d2` answered the second half of that conditional ("if yes, add it to the table") without the first, which makes the spec row the evidence for the matcher.

What I found looking for it:

- **0** person-level `Parents` facts across every committed tree (against `Marriage` 147). The only `"Parents"` strings in the repo are prose in a research question and a research.json **assertion** `fact_type` — a different field on a different document, and an easy near-miss to mistake for confirmation.
- `person-read-raw.json`, the one committed raw FS response, carries Birth/Couple/Death/Marriage/ParentChild. No `Parents`.
- **Your own comment on #1314 is the problem**: "the agent **hand-built** the tree … but kept a `Parents` documentary fact." That is an agent-authored value. The next sentence — "FamilySearch carries parentage twice … `person_read` returns both" — is the inference, and it's now asserted as observed fact in `tree-forget.ts:70-73` and both specs. What you saw confirms the agent wrote `Parents`; it doesn't confirm FS emits it.
- A real person-level parents-naming fact **does** exist in committed data under a different type: `Filiation` (`rafael-reyes-daughter`, value `"hija legítima … of Manuel Rafael Reyes and Serapia Benavides"`). Worth checking whether that's what FS actually returns.

Because the tests hand-write the fixture, they pass whichever type is right — so nothing here can fail if the matcher is wrong, and nothing else validates fact types (`gedcomx_fact_type_recommended` is an open enum). Paste the literal `type` from the 08-03 zip and this is settled either way.

Non-blocking:

- `tree-forget.ts:83` cites "85 such facts" for all three spouse types; the evidence is entirely `Marriage` — `Divorce` and `Annulment` are **0** in committed data, which is the rule the comment itself states two lines above. Harmless no-op, #1549 tracks it. Agree with francis: not a blocker. Same caveat on the `Annulment` spec row's "echoed as a person-level fact".
- Worth a sentence in the spec: for `spouses-of`, the tolerant match changes real behavior at scale — **25 of 82** persons carrying a person-level `Marriage` fact have no couple relationship, so on those "forget spouses" goes from erroring to deleting facts. That's the sole-carrier case working as designed and `removed.factsByType` names it honestly, so I'm not treating it as a defect — just flagging the blast radius, since the reported case was `parents-of`.

Engine vitest on the tree merged with main: **2205 passed / 91 files**, `tsc --noEmit` clean.

Procedural note: `pulls/1419/reviews` is empty — both prior notes are issue comments, not approving reviews, so there are zero approvals on record. #1425 confirms as described (base `1314-tree-forget-parentage-evidence`, 0 reviews), so this is the first gate for both halves.

T-FEH (approved):
Clearing my changes-request. @Asimi1234 you answered the question that was actually asked — a `person_read` **result** with an FS-native UUID fact id is upstream output, which is exactly the distinction my finding turned on, and it's now recorded in the code rather than only in this thread.

Being precise about what that settles, since I was precise about the doubt: I re-ran the corpus search at this head — 406 trees, 2,971 persons, 13,528 facts, 105 distinct types, and `Parents` is still **0**, as it is in `dev/` and the schemas. So the zip remains the sole evidence and cannot be re-checked from this repo. What changed is the *provenance*: my objection was that the only observed instance traced to a tree the agent hand-built, which made "FamilySearch emits this" an inference. A `person_read` response is not that. Combined with @clack391's point that `gedcomx_fact_type_recommended` is an open pattern-based enum — so `Parents` was never an undefined value — the matcher is targeting a type on the best evidence available, and I'm satisfied.

I pushed `03915cf5`, comments and figures only, no behavior change:

- **`tree-forget.test.ts` never got the corrected provenance.** It still put the "Geo[illegible] Wilcox - Caroline E Woodruff" value on `GRN6-4MQ` — your own evidence puts it on `GRNX-DFF` — and still said the tree export is "privacy-redacted to empty facts". That second claim is false and is the premise that seeded #1549: 1,195 of 1,217 persons (98.2%) carry at least one fact and no tree has every person's facts empty. It also can't sit next to this PR using the same corpus as positive evidence for `Marriage`. The doc-only commit fixed `tree-forget.ts` and left the test behind.
- **Three corpus figures had drifted across the main merges** and none was re-measured: 85 → **90** person-level `Marriage` facts, and 25-of-82 → **27-of-86** sole-carrier persons. Corrected, stamped with the date and the scan, and the spec now says to re-measure rather than quote — these move every time the corpus grows.

Verified on the merged tree: typecheck clean under `tsconfig.typecheck.json` (which now includes `tests/`), vitest **2262 passed / 93 files**, `tree-forget.test.ts` **29/29**, and blanking `redundantFactIds` still fails exactly the 5 sweep tests. All 12 checks green.

Two things I'm leaving rather than folding in:

- `simplified-gedcomx-spec.md` says `Annulment` is "echoed as a person-level fact"; there are **0** corpus-wide, same evidence status `Parents` had. #1549 owns it, and it's a harmless no-op meanwhile.
- **#1549's body still carries the redaction claim** this PR's test just shed. Its acceptance criterion 1 is built on it — "gather real (unredacted) data" — so whoever picks it up will skip the cheap corpus pass that already answers part of the question. I've commented there; the body edit is the owner's call.

clack391 (approved):

PR #1421, opened 2026-08-06
eval-harness: validate judge rubric dimensions, drop bad ones instead of failing the run (#1361)

chrisedeson:
## Summary

Closes #1361.

- The judge could return rubric dimension names that do not exist, re-cased versions of real ones, or duplicates, and all of it persisted silently into run logs. Tier: Normal.
- Now every judge dimension is validated against the parsed rubric. A bad name or a duplicate is dropped and recorded as a warning in `output.warnings`. Structural garbage (a `source` outside base/rubric, a non-string name) fails the draw loudly.
- The judge tool schema also lists the valid names as an enum. That steers the model but is not enforced by the API, so the drop-with-warning is the guarantee.

## Why not the hard error the issue asked for

Issue #1361 asked for a `JudgeError` on a bad name. Two independent adversarial reviews replayed that design against every committed run log and it fails on correct output: about 3 percent of all historical judge draws would raise. The judge re-types rubric headings and often truncates them, for example `## Score discipline (advisory)` comes back as `Score discipline`. On one test the judge invents a FAN dimension in every single run because the rubric has no FAN heading at all. A raise there becomes `skipped` in the orchestrator, and a skipped judge on a positive test is recorded as a hard fail. So the letter of the issue turns correct skill output into recorded failures on paid runs, at 1.5 to 3 times the judge cost.

The spec already faced this exact shape once: unit-test-spec-v2's v1.8 note demoted a dimension-name gate to a non-blocking warning. This PR follows that precedent. It also matches the direction of open PR #1393, that an ungradeable result must not read as a failure. The two rubric defects the replay exposed are filed separately rather than papered over here.

## Review guidance

**Start here:** `judge.py::_extract_dimensions`. It now returns `(dims, warnings)`. Order of checks: structural source validation, non-string name, duplicate `(source, name)` keeping the first, unknown rubric name. Then `orchestrator.py` threads `judge_output.warnings` into the existing `_build_warnings()` list. `rubric.py::Rubric.dimension_names()` is the valid set.

**Unsure about:** the issue's lower-confidence note that some tests silently drop all three rubric dimensions. Not enforced here, same reasoning as before: the issue frames it as an open decision.

## Plan

**Didn't change:** `judge/prompt.md` (editing it trips the rule 2b hash warning and is its own cadence, filed as a follow-on), `run-log.schema.json` (warnings ride the existing permissive `output.warnings`), and `e2e/judge.py` (separate judge, no rubric dimensions).

**Acceptance check:** main silently accepts all three incident payloads, verified by calling main's own one-argument `_extract_dimensions` directly: the ut_014 Title-Case re-casing, the ut_016 invented dimension, and the ut_018 duplicate all persist verbatim there. On this branch the same payloads are dropped with warnings. A corpus-replay test now iterates all 129 committed run logs (1873 judge draws) and asserts the extractor never raises; it reports 64 drops, which is the measured rate of real judge misbehavior this change stops persisting.

**Deviated from the plan:** the first commit implemented the issue's literal JudgeError design. Two adversarial reviews measured the blast radius above, so the second commit redesigned to drop-with-warning. Both commits are kept so the history shows the evidence trail.

## Test plan

- [x] Targeted suites green: 115 tests across test_judge, test_rubric, test_orchestrator, including the corpus replay. Full harness suite: 1682 passed with the engine build present (1676 passed, 6 build-gated skips without it).
- [x] Eleven targeted mutants, each killed by a precise expected subset of tests: structural checks, dedup key shape, first-kept-rest-dropped, empty-rubric path, case sensitivity, enum steering.
- [x] N/A, harness code only. No skill dir, unit fixture, or snapshot changed, so no eval run is owed.

## Follow-on issues

Two rubric defects the corpus replay exposed, both needing their skill's eval slot so they cannot ride along here:

- person-evidence rubric heading carries a parenthetical the judge truncates, so that dimension goes ungraded.
- research-plan rubric has no FAN dimension yet its FAN-pivot test expects one graded.


jud-sdev (approved):
Reviewed full pass, 0 commits behind main. Thoroughly built and thoroughly verified — approving.

**Verified by running:**
- `test_judge.py` + `test_rubric.py` + `test_orchestrator.py` — **116 passed.**
- Corpus replay — **129 run logs, 1873 draws, 64 drops, 0 unexpected raises**, exactly the claimed numbers.
- **The vacuous-pass guard genuinely fires:** I mutated `_rubric_for_snapshot` to `return None` (forcing every log to skip); the replay test failed with `assert 0 > 1000`. So `total_draws > 1000` / `dropped_total > 0` really do catch "nothing ran" — the corpus replay is real coverage, not a check that passes vacuously.
- **The new warning kinds fit the schema:** `output.warnings[].kind` is an open `{"type":"string"}` with `additionalProperties:true`, so not editing `run-log.schema.json` is correct — a real run emitting `dropped_unknown_rubric_dimension`/`dropped_duplicate_dimension` passes flush-time validation.

**What holds up under scrutiny:** the structural check runs *before* the dedup builds a `(source, name)` tuple key (so a non-string name can't reach a bare `TypeError` hash); the dedup key is `(source, name)` not `name` alone (base and rubric "Correctness" coexist — pinned); case-sensitivity is deliberate; `_grading_tool_for_rubric` deep-copies the template (verified un-mutated). The deviation from #1361's literal hard-error is well-evidenced (a raised judge → skipped → hard fail on ~2–3% of correct historical draws), and the two prior adversarial-review hardenings (preserve dropped score/rationale; the replay guards) are both present in the diff.

**One thing for a genealogist's eye (informational, not blocking — you documented it):** dropping a *re-cased-but-real* rubric dimension removes its score from `_compute_outcome`, so a test the judge failed on a mis-named craft dimension can now pass on base dimensions alone. It's the intended drop-not-fail tradeoff, made traceable by preserving the dropped score/rationale in the warning — just worth a craft-grading sanity check on "mis-named ⇒ dropped".

Approving.

PR #1427, opened 2026-08-07
fix(web): separate text-block boundaries and mark thinking as private…

jud-sdev:
Closes #1312 
## Summary

- Fix two web chat-pane rendering defects (#1312): consecutive text blocks were glued with no separator (…every subsequent search.q_001 written.) — now joined with a paragraph break; and the thinking <details> was a low-contrast block that read as the answer — now a tinted, dashed, italic block labeled "💭 Model's private reasoning — not its answer", unmistakable in light and dark.
- Extracted the event-folding logic into a pure foldChatEvent so the join is unit-testable, and added a vitest runner to apps/web (it had none).
- Tier: Normal.

## Review guidance

Start here: apps/web/src/components/chatEvents.ts — joinTextBlocks (the Defect-1 join) and foldChatEvent (the reducer lifted out of ChatPane.tsx). Check the extraction is behavior-preserving apart from the join, and that \n\n is the right separator given m.text renders through <Markdown>.

Unsure about: Adding a vitest runner to apps/web — it had no test setup, and the issue's "unit test" acceptance forces one. Standalone node-env vitest.config.ts, slots into turbo test. If there's a house preference for how a new app package gets its first test runner, this is the spot.

## Plan

Didn't change: The error-kind branch still appends with raw +=, not joinTextBlocks — the fix is scoped to consecutive text blocks, which is the reported defect; an error is a terminal append, left as-is.

Acceptance check: apps/web/src/components/__tests__/chatEvents.test.ts — "separates two blocks with a blank line — NOT simple a + b" (and the foldChatEvent join test). Fails on main (module doesn't exist), passes here.

Deviated from the plan: Standing up vitest in apps/web is more than the two-line renderer fix the issue implies — the issue's unit-test acceptance requires a runner the package lacked. Flagged rather than assumed. Also touches pnpm-lock.yaml (+3 lines; vitest was already in the lock via sibling packages).

## Test plan

- [X]I ran make test-all… — Not run on this box (its npm ci prereq fails on the local npm version, unrelated to this change). Ran the affected package's suites directly instead: pnpm --filter web test 8/8 passed, pnpm --filter web typecheck clean, Vite dev compiles with no error. Leaning on CI, which runs the same suites.
- [x] For every skill whose run-log snapshot I changed… — N/A. Nothing under packages/engine/plugin/skills/, eval/tests/unit/, or a referenced scenario/fixture is touched; no skill snapshot changes.
- [x] If I changed a skill's description or DO NOT clauses… — N/A. No skill description or DO NOT clause changed; this is an apps/web renderer change with no routing impact.

## Follow-on issues


apps/web/src/components/chatEvents.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// Two canonical text blocks in one assistant turn are separate paragraphs, but
+// the SDK delivers them as separate events with no separator between them.
+// Joining with `+=` glued them — the tester saw "…every subsequent search.q_001
+// written." and filed it as a punctuation defect. The transcript renders as
+// markdown, so a blank line is the paragraph boundary the model already
+// intended; trailing newlines on the previous block are collapsed so we never
+// stack more than one. (#1312 defect 1.)
+export function joinTextBlocks(existing: string, addition: string): string {
```
Gennecis, line 32: **[MEDIUM-HIGH] (8/10) The `\n\n` renders as a bare line wrap, not a paragraph break.**

`ChatPane.tsx:322` does render through `<Markdown>`, so two `<p>`s are produced. But `packages/viewer-ui/src/styles/global.css:144-150` zeroes every margin:

```css
*, *::before, *::after { box-sizing: border-box; margin: 0; padding: 0 }
```

and `apps/web/src/styles.css` never restores it for `p` — its only paragraph rules, `.msgText :first-child { margin-top: 0 }` / `:last-child { margin-bottom: 0 }` (605-610), cancel a margin the reset already removed, so they're dead code. Rendered with both stylesheets verbatim, the two blocks come out on consecutive lines with no gap.

Pre-existing reset, not your change — but #1312's acceptance is *"render with a paragraph break"*, and one rule gets there (and revives 605-610):

```css
.msgText p {
  margin: 0.6em 0;
}
```

apps/web/src/components/__tests__/chatEvents.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+const textEvent = (text: string): Record<string, unknown> => ({ kind: 'text', text })
+
+describe('joinTextBlocks (#1312 defect 1)', () => {
+  it('separates two blocks with a blank line — NOT simple a + b', () => {
+    // The exact regression: the model punctuated cleanly ("…every subsequent
+    // search.") and began a new block ("q_001 written."); the renderer glued them.
+    const joined = joinTextBlocks('…every subsequent search.', 'q_001 written.')
+    expect(joined).not.toBe('…every subsequent search.q_001 written.')
+    expect(joined).toBe('…every subsequent search.\n\nq_001 written.')
```
Gennecis, line 12: **[MEDIUM] (8/10) The acceptance check passes in the case the acceptance fails.**

This asserts the string, so it's green while the render has no paragraph gap (see `chatEvents.ts:32`). Nothing in CI distinguishes a working boundary from a flattened one.

Keep the string assertion — my mutation run proved it discriminates. Add one render assertion once the CSS lands:

```tsx
const { container } = render(<Markdown>{joinTextBlocks('a.', 'b.')}</Markdown>)
expect(container.querySelectorAll('p')).toHaveLength(2)
expect(getComputedStyle(container.querySelector('p:last-child')!).marginTop).not.toBe('0px')
```

A follow-on issue instead of jsdom-in-this-PR is a fine answer.

apps/web/src/components/chatEvents.ts
```diff
@@ -0,0 +1,92 @@
...
+  if (kind === 'text') {
+    // The canonical block covers everything its deltas already previewed —
+    // commit it (as its own paragraph) and drop the preview rather than
+    // appending both.
+    last.text = joinTextBlocks(last.text, text)
+    last.streamText = ''
+  } else if (kind === 'text_delta') {
+    last.streamText = (last.streamText ?? '') + text
+  } else if (kind === 'thinking') {
+    last.thinking = (last.thinking ?? '') + text
```
Gennecis, line 66: **[MEDIUM] (8/10) `thinking` still glues.**

`real_agent.py:352-353` emits one event per `ThinkingBlock`, same as per `TextBlock` — so two thinking blocks in one turn concatenate with no separator. `.thinkingBody` is `white-space: pre-wrap` (`styles.css:637`), so `\n\n` renders correctly here with no CSS change.

The "Didn't change" note covers the `error` branch but not this one.

```ts
last.thinking = joinTextBlocks(last.thinking ?? '', text)
```

apps/web/src/components/chatEvents.ts
```diff
@@ -0,0 +1,92 @@
...
+    last.streamThinking = ''
+  } else if (kind === 'thinking_delta') {
+    last.streamThinking = (last.streamThinking ?? '') + text
+  } else if (kind === 'error') {
+    last.text += (ev.text as string) ?? 'Error'
```
Gennecis, line 71: **[MINOR] (7/10) The error branch glues, and the path is reachable.**

`ChatPane.tsx:213` fires `applyEvent({ kind: 'error', text: 'Chat unavailable: …' })` when reconnects run out. After a completed sentence that renders as `…here are the results.Chat unavailable: unknown error` — the same defect, through the same `<Markdown>`. "Terminal append" holds for an error *instead of* an answer, not *after* one.

```ts
last.text = joinTextBlocks(last.text, (ev.text as string) ?? 'Error')
```

apps/web/src/styles.css
```diff
@@ -611,6 +611,33 @@ button {
...
 }
+/* The model's private reasoning. Deliberately unmistakable as NOT the answer:
+   a tinted, dashed-border, italic block a tester cannot read as something the
+   agent told them. The old muted low-contrast disclosure sat flush against the
+   answer text and four of thirteen alpha reports quoted thinking as the reply
+   (#1312 defect 2). Collapsed by default; the summary says what it is. */
+.thinkingBlock {
+  margin: 4px 0;
+  padding: 6px 10px;
+  background: var(--bg);
```
Gennecis, line 622: **[SUGGESTION — not a gate] (6/10) `--bg` isn't a tint; line 615 says it is.**

`--bg` → `--bg-primary` = `#faf8f4`; the surrounding surface is `.chatPane { background: var(--bg-elev) }` → `--bg-card` = `#fefdfb` (409-413). Browser-computed: **1.04 : 1**, and both are neutral warm-whites with no hue separation. The dashed border, the italics and your new summary line are doing all the work — and rendered side by side they do meet the acceptance, so this isn't blocking.

Either:

```css
  background: var(--accent-gold-soft);   /* → #f6efde; both themes, global.css:81/:133 */
```

or delete "tinted" from line 615.

apps/web/vitest.config.ts
```diff
@@ -0,0 +1,13 @@
...
+
+// Standalone (not merged with vite.config.ts): these are pure-logic unit tests
+// that need no DOM and no React plugin. A dedicated config keeps `vitest run`
+// from pulling the app build's `@genealogy/schema generate` dev step or its
+// jsx transform. Add a jsdom project here if a component render test is ever
+// needed.
+export default defineConfig({
+  test: {
+    environment: 'node',
+    include: ['src/**/*.test.ts']
```
Gennecis, line 11: **[MINOR] (7/10) Excludes `.tsx`, so a future component test would never be collected.**

Lines 6-7 invite exactly that test. `packages/viewer-ui/vitest.config.ts` and `apps/electron/vitest.config.ts` set no `include` at all and take vitest's default, which covers both extensions.

Drop the line, or:

```ts
    include: ['src/**/*.test.{ts,tsx}']
```

Gennecis (changes requested):
## Review — full pass, 0 commits behind `main`

**Does:** extracts `ChatPane.tsx`'s reducer into a pure `foldChatEvent` (new `chatEvents.ts`), joins consecutive `text` events with `\n\n` instead of `+=`, restyles the thinking `<details>`, and stands up vitest in `apps/web`.

**Intent:** matches #1312. Defect 1 is fixed at the string level; one finding is that it doesn't reach the screen.

### Verified by running

| Check | Result |
|---|---|
| `pnpm --filter web test` | 8/8 |
| `pnpm turbo run test --filter=web --force` | 8/8 — CI's `turbo run test` does collect the new suite |
| `pnpm turbo run typecheck` | 5/5 clean |
| Mutation test — `joinTextBlocks` reverted to `existing + addition` | 2/8 failed. The suite is a real gate. |
| Render check — both stylesheets inlined verbatim, `ChatPane.tsx:300-332` DOM reproduced, contrast from `getComputedStyle` | table below |

`apps/electron` `feedback.test.ts:152` times out under a full parallel `turbo run test` — reproduces identically on `main`, passes standalone in ~1.9 s. Pre-existing flake, not this PR.

### Render check

`make web` can't show either defect: `mock_agent.py` emits zero `thinking` events, and no mock turn emits two consecutive `text` events (`_simulate_search`'s two are mutually exclusive — line 228 is behind `if research is None: … return`). `real_agent.py:208-271` sets no thinking option. So this was checked statically against the real CSS.

| Surface | Computed | vs pane |
|---|---|---|
| chat pane (`--bg-elev` → `--bg-card`) | `#fefdfb` | — |
| thinking block as shipped (`--bg`) | `#faf8f4` | 1.04 : 1 |
| thinking block, `--accent-gold-soft` | `#f6efde` | 1.13 : 1 |

It found the `chatEvents.ts:32` finding, and it corrected my first read of the thinking block (see `styles.css:622`).

### Cleared

- **`\n\n` is the right separator** — `real_agent.py:349-351` emits one `text` event per `TextBlock`, so two events really are two blocks.
- **Extraction is behaviour-preserving**, line-for-line, except the `text` branch. The purity claim holds.
- **Replay is fixed too** — `ChatPane.tsx:197` routes the replayed transcript through the same fold.
- **No missed edit site** — `ChatPane.tsx` is the only renderer of these event kinds.
- **Your "Unsure about" question: this is the house pattern.** `vitest ^3.2.4` matches `viewer-ui` and `electron` exactly; `__tests__/` + standalone `vitest.config.ts` matches both; `js-tests.yml` already covers `^apps/web/`; `turbo.json` already has a `test` task. Nothing else needed wiring.

### Not inline

**[MINOR] `.github/workflows/js-tests.yml:17`** — not in the diff, so no anchor. `"apps/web has no tests, which is exactly why the next two matter"` is now false. Please name `apps/web` in that line.

**[NIT]** PR body says `src/components/tests/`; actual path is `src/components/__tests__/`.

### Why Defect 2 matters beyond polish

4 of 13 reports from one alpha tester graded the model's scratchpad as its stated conclusion — those enter `docs/alpha-feedback-guide.md` triage as false research-quality defects. Under the GPS the written conclusion and the reasoning behind it are graded separately, so a reviewer has to be able to tell which words the agent actually asserted. Rendered, your new summary line plus the dashed border carries that. It's the strongest line in the PR.

### Verdict

**Request changes**, gated on two one-liners: `chatEvents.ts:32` (paragraph break doesn't render) and `chatEvents.ts:66` (thinking still glues). The rest are non-blocking; `styles.css:622` is explicitly not a gate.

### Not checked

No live browser session — the render check is a static harness (real CSS and DOM, but forced-`open` blocks and no web fonts). No `apps/server` or eval-harness suites; nothing in the diff touches them.

apps/web/src/components/chatEvents.ts
jud-sdev: Fixed. Added `.msgText p { margin: 0.6em 0 }` to styles.css so the two paragraphs render with a real gap, and it revives the first and last-child rules you pointed out were dead. Good catch that the break was only right at the string level and never reached the screen.

apps/web/src/components/__tests__/chatEvents.test.ts
jud-sdev: Kept the string assertion since your mutation run showed it discriminates, and filed the render assertion as a follow-on rather than pulling jsdom into this PR: issue #1524. It needs jsdom and testing-library added to apps/web plus a jsdom environment, which is enough new infrastructure to keep separate.

apps/web/src/components/chatEvents.ts
jud-sdev: Fixed. Thinking now goes through joinTextBlocks, same as text. No CSS change since the thinking body is already pre-wrap. Added a test that two thinking blocks join with a blank line.

apps/web/src/components/chatEvents.ts
jud-sdev: Fixed, and you were right that it is reachable after a real answer on the reconnect path, so the terminal-append reasoning did not hold. It uses joinTextBlocks now, with a test for an error landing after a completed sentence.

apps/web/src/styles.css
jud-sdev: Switched the thinking background from --bg to --accent-gold-soft, so it is a genuine tint now and the comment is accurate. The gold reads clearly against the pane's neutral white, and the dashed border and italics still carry the rest.

apps/web/vitest.config.ts
jud-sdev: Fixed. The include now covers ts and tsx, so a future component test gets collected.

Gennecis (approved):

johnmarkpeterbrown (changes requested):
Update branch to the latest main and notify me for final review. Thank you.

johnmarkpeterbrown (approved):
Looks good!

Issue #1121, opened 2026-08-02
Harden the feedback Apps Script endpoint — it accepts unauthenticated writes


DallanQ:
**Touches:** apps/electron/server/feedback-endpoint/Code.gs, apps/electron/src/main/index.ts

The Apps Script web app is deployed "Execute as: me / Access: Anyone", and its URL is
committed in this public repo twice — `apps/electron/src/main/index.ts:154` and
`apps/server/app/config.py:88`. Anyone who reads the repo can POST a bundle and get a file
written into the team Drive folder plus a GitHub issue opened with our PAT, labelled
`genealogist` + `feedback`, routed straight to Ready.

## Why this is iceboxed

The recoverable half is small: delete the files, delete the issues. The PAT is Issues-only
on one repo and no user data is exposed.

The half that would actually hurt is quota. The script runs as one person's Google account,
so a flood burns that account's daily Apps Script limits — URL fetches, runtime, notification
emails — and **legitimate feedback stops working until they reset**. That is a problem once
this repo has attention, not today.

## Do not use a shared secret

Tried in PR #1428 and cut. Two clients post to this URL and only one can hold a secret. The
Electron viewer posts directly to a public URL, so a secret compiled into it is not a secret,
and turning the Script Property on would reject every Electron submission while the tester
saw a success message.

## The design, when we build it

**1. FamilySearch token as the identity.** The client sends its FS access token; `doPost`
calls `/users/current` with it and accepts only on 200. No secret to distribute, works for a
distributed client, works with no control plane, and tells us which tester submitted without
trusting the typed email.

Feasible today: the desktop MCP server writes `~/.familysearch-mcp/tokens.json` in plaintext
at mode 0600, so Electron can read it with no decryption path to build — the Fernet
encryption in `apps/server/app/crypto.py` is server-side only. Fail open on a 5xx from
FamilySearch, closed on a 401. Costs one extra fetch per submission against the same daily
quota we are protecting.

**2. Rate limit per FS user id**, held in `PropertiesService` — a few submissions a day.

Not by IP. `doPost(e)` gives `queryString`, `parameter`, `pathInfo`, `contentLength` and
`postData`; Google does not pass the client IP through at all. And a global cap without an
identity is self-defeating — a flood fills the bucket and blocks the real testers.

## Trigger to un-icebox

The first spam submission, or public visibility around RootsTech — whichever comes first.

The payload size cap is the cheapest guard here and is independent of all of the above, so it
[...]

PR #1428, opened 2026-08-07
fix: surface feedback-endpoint rejections instead of reporting success

yinkid28:
Part of #1121

## Start here

The shared-token approach was cut (see DallanQ's comment). What remains is the
bug fix: both the web control plane and the Electron client silently report
success when Apps Script returns `{ok: false}` — a Drive failure, bad folder ID,
or blown quota looks like success to the user and their bundle is gone.

## Unsure about

Nothing — the token decision was made by DallanQ in the PR thread.

## Didn't change

- No FamilySearch auth or session code
- No eval harness or skill changes

## Summary

- **Web (`feedback.py`):** parse the Apps Script response body after
  `raise_for_status()` and raise `HTTPException(502)` when `ok` is not `True`.
  Catch `ValueError` alongside `httpx.HTTPError` so a non-JSON 200 (e.g. a
  redirected or misconfigured Apps Script serving HTML) also surfaces as 502.
- **Electron (`index.ts`):** check `result.ok` after `res.json()` and throw on
  rejection, so the feedback dialog shows an error instead of a success toast.
- **Code.gs:** add a payload size cap (~35 MB decoded, 47 MB base64) at the top
  of `doPost`, before the Drive write. Matches the client-side bundle cap; the
  endpoint previously had no server-side limit.
- Removed all `FEEDBACK_TOKEN` / `feedback_secret` machinery — the shared secret
  cannot authenticate a distributed Electron client. Issue #1121 has been
  rewritten with the real design (FS token identity + rate limit, iceboxed).

## Acceptance check

```bash
cd apps/server && uv run pytest tests/test_feedback.py -q
```

Both `test_rejected_upload_surfaces_as_502` and `test_non_json_response_surfaces_as_502`
must pass. Neutering the 502 check must fail the first; neutering the `ValueError`
catch must fail the second.

## Deviated from the plan

Yes — the token was cut entirely per DallanQ's review. The Electron silent-success
fix and the Code.gs size cap were added at his direction.

## Test plan

- [x] `uv run pytest tests/test_feedback.py -q` — 14/14 pass
- [x] `test_rejected_upload_surfaces_as_502` — feeds `{ok: false, error: "unauthorized"}` and asserts 502
- [x] `test_non_json_response_surfaces_as_502` — feeds a non-JSON response and asserts 502
- [x] Existing `test_feedback_context_and_drive_upload` envelope assertion guards against accidental extra keys


Asimi1234 (changes requested):
The change adds shared-token auth to the Apps Script feedback endpoint — doPost now rejects requests whose token doesn't match a FEEDBACK_TOKEN Script Property (accepting all when it's unset, for rollout back-compat), and the control plane sends its FEEDBACK_SECRET as token in the POST envelope. I verified the token check is fail-closed and positioned before every side effect (Drive write, GitHub issue, email), the FEEDBACK_SECRET env mapping and settings binding are correct, no token leaks into logs, and the two new tests genuinely pin the behavior (neutering the envelope logic fails exactly the intended assertion). Full control-plane suite passes 137/1-skipped. No blocking findings. 

### Two non-code items: 

1. an informational silent-failure risk — Apps Script returns HTTP 200 even on rejection, so a token mismatch during the two-console rollout would drop feedback while showing the user success (pre-existing root cause, worth a follow-on issue or a small ok-check fix).

2. Code.gs itself has no automated test — inherent to Apps Script, so its enforcement is only verifiable via the README's post-deploy smoke test.

clack391 (changes requested):
Independent senior pass plus a second reviewer, reconciled. The token mechanism itself is sound: fail-closed when set, before every side effect, empty equals unset, no token leak, and the two tests pin the envelope. Two things to address before merge.

1. The control plane reports a rejected upload as success. Apps Script serves HTTP 200 even on {ok:false}, so on a token/secret mismatch feedback.py res.raise_for_status() does not fire and it returns {ok:True} without checking the body. A misconfigured token then silently drops every web submission while the user sees success. This PR introduces that reject path, so please close it here: after raise_for_status(), read the body and surface an error when res.json().get("ok") is not True, with a test feeding {ok:false}.

2. Enabling the token breaks the still-live Electron feedback path. The Electron feedback:submit handler sends no token, so once FEEDBACK_TOKEN is set, Apps Script rejects every Electron submission (visibly, since the client surfaces ok:false). The deployment steps instruct setting the token, so please state that enabling it means no longer accepting tokenless Electron submissions, or confirm that path is being retired. Worth the lead sign-off as an auth decision.

Non-blocking: the control-plane guard is "is not None" while Apps Script treats an empty token as off, so an empty FEEDBACK_SECRET sends token:"" rather than omitting it (consider "if settings.feedback_secret:"); README smoke-step 3 only holds once the token is set, worth noting as a precondition. The non-constant-time compare is fine here and is documented.

yinkid28:
Pushed `e77a501e` addressing review findings 1 and 3. Feedback tests: 15/15 pass.

**1. Silent failure on token mismatch — fixed.**
`feedback.py` now parses the response body after `raise_for_status()` and raises `HTTPException(502)` when `ok` is not `True`, with the upstream error in the detail message. New test `test_rejected_upload_surfaces_as_502` feeds `{ok: false, error: "unauthorized"}` and asserts the 502.

**3. Empty-string guard nit — fixed.**
Changed `if settings.feedback_secret is not None:` to `if settings.feedback_secret:` so an empty `FEEDBACK_SECRET` omits the token rather than sending `token:""`.

**2. Electron path — needs lead decision.**
Once `FEEDBACK_TOKEN` is set in the Apps Script console, every Electron submission (which sends no token) will be rejected. This is an auth/rollout decision, not a code change. @Asimi1234 @clack391 — should the deployment notes state that enabling the token retires tokenless Electron feedback, or do we need a grace window / separate path?

Asimi1234:
Re-reviewed at `e77a501e`, on the tree merged with main (only conflict is `deploy/fly.toml`, a comment line).

**#1 (silent-drop → 502) and #3 (empty-token guard) — resolved.** Feedback suite 15/15, full server 165 pass / 1 skip on the merged tree; neutering the 502 check fails `test_rejected_upload_surfaces_as_502`, and neutering the envelope line fails `test_envelope_includes_token…`.

**New, in the #1 fix:** `feedback.py:482` catches only `httpx.HTTPError`, but `res.json()` raises `JSONDecodeError` on a non-JSON 200 (a misconfigured/redirected Apps Script returns HTML with status 200) → unhandled 500 instead of the 502 you added. One-liner:

```python
    except (httpx.HTTPError, ValueError) as exc:
```

**#2 (Electron) — unresolved and above my call.** `index.ts:148` POSTs no token to the same endpoint, so once `FEEDBACK_TOKEN` is set every installed Electron submission is rejected. @DallanQ — auth call: should the deploy notes state that enabling the token retires tokenless Electron feedback, or does that path need a grace window first?

yinkid28:
Pushed `f2cb5980` addressing the remaining items:

**JSONDecodeError catch — fixed.** Added `ValueError` to the except clause on `feedback.py:482` so a non-JSON 200 (e.g. a redirected Apps Script serving HTML) returns 502 instead of an unhandled 500. Server feedback tests: 15/15.

**Merge conflict — resolved.** Merged `origin/main`; `fly.toml` secrets comment now lists both `FS_TOKEN_ENC_KEY` and `FEEDBACK_SECRET`.

**Electron path — decision for @DallanQ.** Asimi's analysis is thorough: no Electron release has shipped, and a client-side shared secret can't authenticate a distributed app. The call is whether to route Electron intake through `/api/feedback` (option 3 from #1121) before Electron ships, or merge this as the web-side half and track the Electron path separately. Either way this PR is "Part of #1121", not "Closes".

DallanQ:
**Decision on the token: cut it.** Answering what the last two rounds escalated to me.

We're solving a problem we haven't had, and the shared secret can't solve it anyway. Two clients post to that URL and only one of them can hold a secret. Electron posts straight to a public URL, so a secret compiled into it isn't a secret. Setting `FEEDBACK_TOKEN` would just reject every Electron submission — and the tester would see a success message. So the property never gets set, and the code is a knob nobody can safely turn.

None of this is wasted work. The bug you found while building it is real and worth merging on its own.

## Remove

- the `FEEDBACK_TOKEN` check in `Code.gs`
- `feedback_secret` in `config.py`
- the `token` envelope line in `feedback.py`
- `test_envelope_includes_token_when_feedback_secret_is_set` and `test_envelope_omits_token_when_feedback_secret_is_none`
- the `FEEDBACK_TOKEN` row and the wrong-token smoke step in the endpoint README
- the `FEEDBACK_SECRET` mention in `deploy/fly.toml`

## Keep

- the `{ok:false}` → 502 check in `feedback.py`, and `test_rejected_upload_surfaces_as_502`
- `ValueError` in the except clause
- the `SCRIPT_VERSION` bump

## Add

**1. The Electron client has the same silent-success bug you just fixed on the Python side.** It checks the HTTP status, and Apps Script always returns 200. I ran it against a server returning `{ok:false,error:"unauthorized"}` — the handler resolves and the feedback dialog shows a success toast. This is live today, token or no token: any Drive failure, bad folder ID, or blown quota looks like success to the user and their bundle is gone.

`apps/electron/src/main/index.ts:168`

You wrote:

```ts
const result = await res.json().catch(() => ({}))
return { ok: true, filename: built.filename, ...result }
```

Change it to:

```ts
const result = await res.json().catch(() => ({}))
if (result.ok !== true) {
  throw new Error(`Feedback endpoint rejected the upload: ${result.error || 'unknown error'}`)
}
return { ok: true, filename: built.filename }
```

**2. A payload size cap in `Code.gs`.** Every size limit we have is client-side — 25 MB per file, 35 MB per bundle, 20 MB session log, all in `apps/electron/src/main/feedback.ts`. The endpoint has none; it decodes whatever arrives and writes it. Put this at the top of `doPost`, before the Drive write:

```js
// ~20 MB decoded — base64 is 4/3 of raw. Well above any real bundle.
if (payload.zipBase64.length > 27 * 1024 * 1024) {
  return ContentService
    .createTextOutput(JSON.stringify({ ok: false, error: 'payload too large' }))
    .setMimeType(ContentService.MimeType.JSON);
}
```

## PR body

Drop `Closes #1121` and the whole Deployment section — there's nothing left to deploy but pasting the new `Code.gs`. Retitle to something like "surface feedback-endpoint rejections instead of reporting success".

I've rewritten issue #1121 with the design for when we do harden this: a FamilySearch token as the identity and a per-user rate limit, not a shared secret. It's iceboxed until we actually need it.

DallanQ (changes requested):
Please simplify according to my comment

yinkid28:
Pushed `5ece8943` implementing all of Dallan's review:

**Removed:**
- `feedback_secret` from `config.py`
- Token envelope line from `feedback.py`
- `FEEDBACK_TOKEN` check from `Code.gs`
- `FEEDBACK_SECRET` from `fly.toml`
- Both token tests (`test_envelope_includes_token…`, `test_envelope_omits_token…`)

**Kept:**
- `{ok:false}` → 502 check and `ValueError` catch in `feedback.py`
- `SCRIPT_VERSION` bump

**Added (per Dallan's comment):**
- **Electron `index.ts:168`:** check `result.ok` after `res.json()` and throw on rejection — fixes the same silent-success bug on the Electron side
- **Code.gs:** ~20 MB decoded payload size cap at the top of `doPost`, before the Drive write
- `test_rejected_upload_surfaces_as_502` and `test_non_json_response_surfaces_as_502`

PR retitled and body rewritten — "Part of #1121", not "Closes". Feedback tests: 14/14.

DallanQ (changes requested):
**Update the README**: Not done — the endpoint README still documents the token. The FEEDBACK_TOKEN row is still in the Script Properties table, smoke step 2 still tells you to include a token in the POST, and smoke step 3 is still "submit with a wrong token, confirm unauthorized." Nothing reads that property anymore, so the row describes a knob that does nothing and step 3 now fails — the submission succeeds. It's the one item on your Remove list that wasn't touched.

**Update the size cap to 47MB**: The size cap number is wrong, and this PR is what makes it hurt. Your comment cites the 35 MB bundle cap, but the cap you wrote rejects at ~20 MB decoded. Both clients deliberately build up to 35 MB — they drop media until the total is under 35, not under 20. A 25 MB bundle base64s to ~33 MB and the new check rejects it. That rejection used to be silent; now the ok check you asked for turns it into a 502 on web and an error toast in Electron. So the two additions together convert a submission that works today into a visible failure.

One number decides it, but it's your call which side moves: raise the endpoint to ~47 MB (35 × 4/3, under Apps Script's 50 MB per-execution ceiling but not by much), or drop both client caps to 20 MB.

**Make this minor change**: in feedback.py, body = res.json() reuses the name of the request model parameter. Not a bug — nothing reads the request body after that line — but the next person to add a line there gets a dict.

yinkid28:
Pushed `eea24414` addressing all three items:

1. **README cleaned up.** Removed the `FEEDBACK_TOKEN` row from the Script Properties table and the wrong-token smoke step. Steps renumbered.
2. **Size cap raised to 47 MB.** 35 MB bundle cap × 4/3 base64 overhead = ~47 MB, under Apps Script's 50 MB per-execution ceiling. The old 27 MB cap would have rejected any bundle over ~20 MB decoded.
3. **Variable rename.** `body = res.json()` → `resp_body` to avoid shadowing the request model parameter.

Feedback tests: 14/14.

PR #1433, opened 2026-08-07
record_search: stop placeParts leaving a stray "." on abbreviated counties (#1267)

### LINKED ISSUE #1267: record_search: isSubCountryPlace("Co., USA") reads as sub-country
> **Reviewed 2026-08-04 before junior handoff.** Decision settled (@DallanQ, 2026-08-04): unit suite only, no live e2e run. The bug is also **bigger than this body says** — see "Review notes" at the end.

**Touches:** packages/engine/mcp-server/src/utils/marriage-jurisdictions.ts, packages/engine/mcp-server/tests/utils/marriage-jurisdictions.test.ts, docs/specs/record-search-tool-spec-v2.md

`isSubCountryPlace("Co., USA")` returns `true`. `placeParts` strips `co` out of
`co.` and leaves a bare `"."`, which survives the empty-string filter and counts
as a locality — so a place that is only an abbreviated qualifier plus a country
reads as scoped.

The failure mode is the one this guard exists to prevent. `isSubCountryPlace`
gates `jurisdictionHints`, and the whole point of the gate is that unscoped and
country-wide searches get no hint: every candidate the tree could offer was
already inside the search. A hint fired on this input would list localities
already covered, and its note ("did not find the subject in the place searched")
would be false.

Deferred from PR #1241 rather than fixed there, for the two reasons stated in
that PR:

- `placeParts` also feeds `placeTokens` → `samePlace`, i.e. the jurisdiction
  **exclusion and ordering** that `docs/specs/record-search-tool-spec-v2.md`
  flags as load-bearing and needing a live run to re-verify. Touching it inside
  a PR about something else risked more than the wart.
- No real caller produces that input today — `marriagePlace` and
  `recordSubdivision` always carry a name.

Currently pinned by a `KNOWN WART` test in
`tests/utils/marriage-jurisdictions.test.ts` that asserts the wrong-but-actual
behavior, and written up in `record-search-tool-spec-v2.md` § Place matching.

**Done when:** `isSubCountryPlace("Co., USA")` is `false`; the `KNOWN WART` test
is inverted to assert that and its comment updated; the spec note is updated;
and the jurisdiction exclusion + ordering is re-verified against a live
`make e2e-run TEST=jimmie-jewel-neal`, since any `placeParts` change moves
`samePlace` too. The re-verification is the real cost here, not the fix.

---

## Correction: the second deferral bullet understates the bug

The body says "No real caller produces that input today". That is true only of the
*literal* input. **The stray `"."` is left on every `X Co.` part, not only on a bare
one** — `placeParts("Hill Co., Texas")` is `["hill .", "texas"]`. That token feeds
`placeTokens` → `samePlace`, so a search scoped `recordSubdivision: "Hill Co."` does
**not** match the tree's `"Hill, Texas, United States"`: the exclusion silently fails and
the jurisdiction that just came back empty is offered back as its own top alternative —
precisely the failure PR #1241 widened `searchedPlace` to fix.

The model already writes this abbreviation
(`eval/runlogs/e2e/jimmie-jewel-neal/run-2026-07-30_23-05-46.json` persists an assertion
`place` of `"Hill Co., Texas"`), but **no** recorded search argument uses it — 0 of 131
`marriagePlace` values across `eval/runlogs/e2e/`, all spelling `County` or omitting it.
So: reachable in production, unexercised by the corpus, not a live regression.
**`isSubCountryPlace` is the symptom; `placeParts` is the defect.**

## Fix `placeParts`, not `isSubCountryPlace`

One character, in `packages/engine/mcp-server/src/utils/marriage-jurisdictions.ts`:

```diff
-    .map((part) => part.replace(/\bcounty\b|\bco\.?\b/g, "").trim())
+    .map((part) => part.replace(/\bcounty\b|\bco\b\.?/g, "").trim())
```

`\bco\.?\b` cannot consume the dot: after `co.` the trailing `\b` fails, the engine
backtracks, and it matches bare `co`. `\bco\b\.?` asserts the boundary first, then eats
the dot.

**Do not fix this by filtering punctuation-only parts** — not in `isSubCountryPlace`, not
in `placeParts`' `.filter()`. That flips `isSubCountryPlace("Co., USA")` to `false`,
inverts the `KNOWN WART` test, greens CI, and lets the spec note be deleted as "fixed" —
while `placeTokens("Hill Co., Texas")` stays `["hill .", "texas"]` and the exclusion bug
survives with its written record gone. A blanket `[^a-z0-9 ]` strip is also wrong: it is
wider than this issue and rewrites `"Seneca Co., O."` and every other place holding a
period.

**Measured blast radius:** across all 2,444 distinct
`place`/`standard_place`/`*Place`/`recordSubdivision` strings in `eval/tests/e2e/`,
`eval/fixtures/`, and `eval/runlogs/`, exactly **7** tokenize differently, all in the
intended direction — `Hill Co., Texas`, `Baltimore Co., Md`, `Licking Co.`,
`Utica, Licking Co., Ohio`, `Seneca Co., O.`, `Washington Co., O.`,
`Hants Co.,Nova Scotia,Canada`. None is a `place` in any `starting-tree.gedcomx.json` and
none is a search argument. Every other string is byte-identical before and after.

## Sites — all three, and no others

- [ ] `packages/engine/mcp-server/src/utils/marriage-jurisdictions.ts` — the regex in
      `placeParts`.
- [ ] `packages/engine/mcp-server/tests/utils/marriage-jurisdictions.test.ts` — invert the
      `KNOWN WART` case to `expect(isSubCountryPlace("Co., USA")).toBe(false)` and rewrite
      its comment; **and add one exclusion test** — `marriageJurisdictionCandidates` with
      `searchedPlace: "Hill Co., Texas"` must exclude a tree fact placed
      `"Hill, Texas, United States"` (copy the shape of "excludes a differently-spelled
      form of the same jurisdiction", `:97`). Without that second test the root cause
      stays untested and the next refactor reintroduces it.
- [ ] `docs/specs/record-search-tool-spec-v2.md` § "Place matching" — replace the
      "**Known wart, recorded rather than fixed.**" paragraph (`:483-493`) with what the
      fix now guarantees and why the `\bco\.?\b` spelling was wrong.

`src/tools/record-search.ts:757` and `src/types/record-search.ts:300` mention
`isSubCountryPlace` in comments only — no edit. `build/` is generated.

## Done when (decided @DallanQ, 2026-08-04)

`isSubCountryPlace("Co., USA")` is `false`; `placeParts("Hill Co., Texas")` is
`["hill", "texas"]`; the `KNOWN WART` test is inverted with its comment rewritten and the
new `"Hill Co., Texas"` exclusion test added; and the spec's "Known wart" paragraph is
replaced.

**Verification is the unit suite, not a live run.** A change like this — token-identical
on almost every input — does not require a live e2e run; we already accept run-to-run
jitter, and this fixture cannot reach the changed branch anyway (the `jimmie-jewel-neal`
tree holds no `Co.`/`County` place and no recorded `marriagePlace` uses the
abbreviation), so a $12.68–$25.24, 1–3 hour run would buy a jitter sample filed as
verification. Record that reasoning in `docs/specs/record-search-tool-spec-v2.md`
§ "Place matching", beside the sentence mandating a live run for ranking changes, with the
exemption's scope: a change whose tokenizer output is unchanged on the corpus.

**Instrument:** `npx vitest run tests/utils/marriage-jurisdictions.test.ts` from
`packages/engine/mcp-server` (24 tests, ~145 ms, green today) → `make engine-test` →
`make test-all`. This PR touches no `SKILL.md`, no plugin agent, and no
`eval/tests/unit/` file, so `check_runlogs.py` does **not** gate it and no
`make eval-skill` re-run is owed.

T-FEH:
## Summary

- **Normal tier.** `placeParts` drops the `County`/`Co.` qualifier with `\bco\.?\b`, which cannot consume the abbreviation's dot — after matching `co.` the trailing `\b` fails, the engine backtracks to bare `co`, and a `"."` survives the empty-string filter as a locality. `\bco\b\.?` asserts the boundary first, then eats the dot.
- **The reported symptom was the smaller half.** `isSubCountryPlace("Co., USA")` returning `true` is real, but the same stray token feeds `placeTokens` → `samePlace`: `"Hill Co., Texas"` tokenized as `["hill .", "texas"]` and so did **not** match a tree place of `"Hill, Texas, United States"`. The exclusion silently failed and the jurisdiction a search had just come back empty on was offered back as an alternative — sub-places alongside it, and counted twice, since `placeTokens(...).join("|")` is also the dedupe key.
- Three behaviour changes, not one: the hint gate (`record-search.ts:826`), candidate **exclusion**, candidate **dedupe**.

Closes #1267.

## Review guidance

**Start here:** the one changed line, `src/utils/marriage-jurisdictions.ts` `placeParts`. The regex is the whole fix; the other `+15` in that file is an explanatory docstring. What to check is the *scope* of the new pattern, not the pattern itself — I'd want a second pair of eyes on whether `\bco\b\.?` can strip something it shouldn't. What I did: exhaustive enumeration over `c o . space x , - 1 _ ñ '` for all strings up to length 5, comparing match offsets and text — 0 violations of "same start offsets, new match text == old or old + `.`". Spot-checked `Colorado`, `Cocke`, `Coos`, `Rio Coco`, `Cook County`, `Co-operative`, `Coeur d'Alene`, `Comté de Nice`, `Denver, CO` — all unchanged.

Second, the spec's new **live-run exemption** (§ "Place matching"). It is a durable rule, not a one-off note, so it deserves more scrutiny than the code. It went through three drafts because the first two were falsified by their own first application — they said in effect "exempt when nothing in the corpus changes" while 7 strings do change. It now names the two input classes that actually reach `placeParts` and explicitly excludes `unstripped-tree.gedcomx.json`, `research.json` places, README prose and transcripts, with the reason, so the next reader doesn't "fix" it by widening the scan. Run verbatim for this change: class 1 — 1623 distinct tree places (1405 from staged trees, rest across 144 final trees), 0 changed; class 2 — 54 recorded search arguments, 0 changed. **Exempt.**

**Unsure about:** nothing blocking, but two things you should not have to discover:

1. **The issue's site list is wrong in a way that matters.** Its review notes say `record-search.ts` mentions `isSubCountryPlace` "in comments only" and cite `:757`. It **imports** it at `:31` and **calls** it at `:826`, inside the `jurisdictionHints` gate; `:757` is unrelated code. The conclusion (no edit needed there) still holds — but that call site is where the production behaviour change lands, so the reason on the issue hid the thing a reviewer most wants to know.
2. **The stated justification for skipping the live e2e run is too strong**, and this PR does not repeat it. The notes say the `jimmie-jewel-neal` fixture "cannot reach the changed branch". `searchedPlace` is built from the model's *live* search arguments and `record_search` reads the run's *mutable* `tree.gedcomx.json`, so a fresh run could in principle reach it. What is true and sufficient: nothing in the corpus exercises it — 0 of 133 `marriagePlace` occurrences, 0 of 24 distinct `recordSubdivision` values, and 0 `place`/`standard_place` across all 144 `*.final-tree.gedcomx.json`. @DallanQ, the no-live-run decision was yours and it still looks right to me on the weaker premise; flagging it because it was made on the stronger one.

One deliberate non-discriminating assertion: `expect(isSubCountryPlace("Hill Co., Texas")).toBe(true)` passes under **both** spellings. It is not a check that can't fail — it guards against a *future* over-stripping regex, next to the case that does discriminate. Calling it out so it doesn't read as an oversight.

## Plan

**Didn't change:** the tempting fix, explicitly. Filtering punctuation-only parts — in `isSubCountryPlace` or in `placeParts`' `.filter()` — flips the symptom, greens CI, and invites deleting the spec note as "fixed" while `placeTokens("Hill Co., Texas")` stays `["hill .", "texas"]` and the exclusion bug survives with its written record gone. No blanket `[^a-z0-9 ]` strip either: wider than this issue, and it rewrites `"Seneca Co., O."` and every other place holding a period. `samePlace`, `placeTokens`, `rankKey` and the ordering logic are not edited — though their *outputs* move wherever `placeParts` does, which is the point. `src/tools/record-search.ts` and `src/types/record-search.ts` are untouched.

**Acceptance check:** `tests/utils/marriage-jurisdictions.test.ts` → `"excludes a jurisdiction searched with an abbreviated County qualifier"`. Fails on `main` (the stray `"."` makes `samePlace` false, so the searched jurisdiction returns as its own alternative) and passes here. Chosen over the inverted `KNOWN WART` case deliberately: the wart is the symptom, this pins the root cause. Verified falsifiable by reverting the regex — exactly the two new/changed assertions fail and 23 others still pass.

**Deviated from the plan:** nothing structural. Two claims were corrected mid-flight after `/code-review` disproved them, both of which I had originally taken from the issue or the plan critic rather than checking: that the model writes `"Hill Co., Texas"` into the tree `record_search` reads (it does not — the string lives in `research.json` assertions and transcripts, paired with a `standard_place` that `marriage-jurisdictions.ts:233` reads first), and that the jurisdiction came back as its own **top** alternative (it returns at rank 2). Both are corrected in the spec, the source docstring and the test comment.

No step-4 stop rule was hit: no schema, no `src/auth/`, no plugin-agent permissions, no ADR reversal.

## Test plan

  Engine **2054 passed**, control-plane 135, eval-ui 148, harness 1681, typecheck clean. `test-js` flakes on the pre-existing 5s timeouts in `apps/electron`'s `feedback.test.ts` (#1366) — a package this diff does not touch, previously reproduced on a clean tree with unrelated work stashed.

- [x] No `SKILL.md`, plugin agent, `eval/tests/unit/` file or fixture is touched, so `check_runlogs.py` gates nothing here and **no `make eval-skill` re-run or run-log annotation is owed.** Confirmed against the workflow's `touched_skills` derivation, not just its path filter — the filter does include `^docs/`, so the job runs; it just has no target.

- [x] **No live `make e2e-run`,** per @DallanQ's 2026-08-04 decision, now recorded in the spec as a scoped exemption rather than a one-off. See "Review guidance" for the correction to its stated premise.

## Follow-on issues

- **#1426** — `placeParts`' qualifier stripping mangles non-ASCII places (`"Coïmbra"` → `["ïmbra"]`; JS `\b` is ASCII-only) and eats the `CO` state abbreviation, so `samePlace("Denver, Iowa", "Denver, CO")` is `true` and a valid alternative jurisdiction is suppressed — the same class of failure this feature exists to prevent. Both verified identical under the old and new spelling, i.e. pre-existing and untouched here. Case 1 needs a disambiguation decision (`co` vs `CO` is not resolvable from the token alone), so it is not a patch someone can just apply.

### DIFF
--- a/docs/specs/record-search-tool-spec-v2.md
+++ b/docs/specs/record-search-tool-spec-v2.md
@@ -463,6 +463,50 @@ as a behavioural change and re-verify against a live run, not unit tests alone.
 
 #### Place matching
 
+> **Live-run exemption for tokenizer-only changes.** The mandate above — treat a
+> ranking change as behavioural and re-verify against a live run — has one
+> measured exception. A `placeParts`/tokenizer change is exempt when **nothing a
+> run would actually feed to `placeParts` tokenizes differently** before and after.
+> That is exactly two input classes, and the criterion must be checked against
+> both:
+>
+> 1. **Tree facts** — `place`/`standard_place` in every
+>    `eval/tests/e2e/*/starting-tree.gedcomx.json` (what the harness stages) **and**
+>    every `*.final-tree.gedcomx.json` in `eval/runlogs/` (the run's *mutable* tree
+>    is what `record_search` reads, not the starting one). Note `standard_place ||
+>    place` — the standardized form wins where both exist.
+> 2. **Recorded search arguments** — `marriagePlace` and `recordSubdivision` in
+>    `eval/runlogs/e2e/`, which is where `searchedPlace` comes from.
+>
+> Deliberately **not** in the criterion: `unstripped-tree.gedcomx.json` (an
+> authoring artifact, never staged into a run), `research.json` assertion places,
+> README prose, and transcript text. A change may move those and still qualify —
+> which is exactly what happened below, so read the two classes above as the whole
+> test, not as a summary of a wider scan.
+>
+> Applied once, for the `\bco\.?\b` fix below, and the numbers are the criterion
+> run verbatim: **class 1 — 1623 distinct tree-fact places** (1405 from
+> `starting-tree.gedcomx.json`, the rest across 144 `*.final-tree.gedcomx.json`),
+> **0 changed**; **class 2 — 54 distinct recorded search arguments, 0 changed**.
+> Exempt.
+>
+> Seven strings *did* change elsewhere in `eval/` — run-log outputs, one fixture
+> README line, and two `unstripped-tree.gedcomx.json` files. That is why the
+> criterion is scoped to the two input classes rather than to a corpus-wide diff: a
+> wider scan is the wrong test and would have refused an exemption that is correct.
+> A live run here would have sampled run-to-run jitter rather than the change.
+>
+> **Unexercised is not unreachable, and the distinction is the whole point.** Real
+> FamilySearch tree data does spell counties this way — `ruse-children`'s
+> `unstripped-tree.gedcomx.json` carries `place: "Seneca Co., O."` and
+> `mcaloney-mother`'s carries `"Hants Co.,Nova Scotia,Canada"` — so a fact whose
+> `standard_place` is absent would put the abbreviated form straight into
+> `placeParts` (`marriage-jurisdictions.ts` reads `standard_place || place`). That
+> has not happened in the corpus: of the 10,730 placed facts across every tree
+> file, the 150 with no `standard_place` contain no `co` token. A change that moves
+> a tokenized value any fixture actually feeds to `samePlace` still needs the live
+> run. This exemption is not a general licence to skip one.
+
 `searchedPlace` is the caller's `marriagePlace` when given, otherwise
 `recordSubdivision` + `recordCountry` joined — the caller usually scopes a marriage
 search with the latter pair, and reading only `marriagePlace` left the exclusion
@@ -480,17 +524,30 @@ string — so it would let TypeScript narrow an `else` to `undefined`. Callers t
 need the string narrowed test `!== undefined` themselves; that explicit check is
 what makes `jurisdictionHints.searchedPlace` a sound required `string`.
 
-**Known wart, recorded rather than fixed.** `isSubCountryPlace("Co., USA")` returns
-`true`. `placeParts` strips `co` out of `co.` and leaves the bare `"."`, which
-survives the empty-string filter and counts as a locality. The failure mode is
-precisely the one this guard exists to prevent — a country-wide search reading as
-scoped — so it is worth an eventual fix. It is left alone here because `placeParts`
-also feeds `placeTokens` → `samePlace`, i.e. the jurisdiction **exclusion and
-ordering** this spec flags as load-bearing and requiring a live run to re-verify,
-and because no real caller produces that input (`marriagePlace` and
-`recordSubdivision` always carry a name). Pinned by a `KNOWN WART` test in
-`tests/utils/marriage-jurisdictions.test.ts` so a future change to `placeParts`
-surfaces it.
+**The `Co.` qualifier, and why the first spelling of the regex was wrong.**
+`placeParts` drops `County`/`Co.` with `/\bcounty\b|\bco\b\.?/`. The abbreviation's
+dot is part of the qualifier and must be consumed with it. The original
+`\bco\.?\b` could not: after matching `co.` the trailing `\b` fails, so the engine
+backtracks and matches bare `co`, leaving a `"."` that survives the empty-string
+filter and counts as a locality. `\bco\b\.?` asserts the boundary first, then eats
+the dot.
+
+That stray token had two effects, and the reported one was the smaller. It made
+`isSubCountryPlace("Co., USA")` return `true`, so a country-wide search read as
+scoped — the failure this guard exists to prevent. More importantly it made
+`placeParts("Hill Co., Texas")` return `["hill .", "texas"]`, so `samePlace` did
+not match a tree place of `"Hill, Texas, United States"`: the **exclusion** failed
+and the jurisdiction a search had just come back empty on was offered back as an
+alternative (with its sub-places alongside it). It also split one jurisdiction into
+two candidates, since `placeTokens(...).join("|")` is the dedupe key.
+
+Reachable in principle — real tree data spells counties this way, and a fact with
+no `standard_place` feeds the abbreviated form straight into `placeParts` — though
+nothing in the corpus exercises it, which is why the fix qualified for the live-run
+exemption at the top of this section. Pinned by two tests in
+`tests/utils/marriage-jurisdictions.test.ts`: the `isSubCountryPlace` case, and an
+exclusion case scoped `"Hill Co., Texas"` that pins the root cause rather than the
+symptom.
 
 It is compared to each candidate on comma-separated tokens, lowercased, with
 `County`/`Co.` dropped and the country term dropped unless it is all that remains.
--- a/packages/engine/mcp-server/src/utils/marriage-jurisdictions.ts
+++ b/packages/engine/mcp-server/src/utils/marriage-jurisdictions.ts
@@ -84,13 +84,25 @@ interface TreeLike {
  */
 const COUNTRY_TERMS = new Set(["united states", "usa", "us"]);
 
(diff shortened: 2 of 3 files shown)


Asimi1234 (approved):
The one-line regex change (\bco\.?\b → \bco\b\.?) correctly consumes the abbreviated county's trailing dot, I verified it introduces no over-stripping across 28 test strings, and I confirmed the new tests genuinely pin the fix (reverting the regex fails exactly the two intended assertions). Full engine suite passes 2054/2054, typecheck clean. No blocking findings.

Issue #1317, opened 2026-08-05
Viewer shows no conflicts, no research question, and loses files between sessions


DallanQ:
**Touches:** packages/viewer-ui/src, apps/server/app, packages/engine/mcp-server/src/utils/project-io.ts

Three reports from untangleyourroots@gmail.com, plausibly one root cause: project state is not reaching what the viewer reads.

1. `2026-07-20T20-45` — agent hit a conflict mid-research and resolved it. *"It did not show the conflict on the research viewer. The conflict section is blank."* The tester wanted to see both the conflict and its resolution.
2. `2026-07-31T18-11` — *"On claude it shows the question but viewer does not have it."* → *"Viewer should show the process."*
3. `2026-07-31T18-22` — *"It said it lost files between sessions."* It then re-ran and found records.

**Acceptance:** a conflict the agent records appears in the viewer's Conflicts section with its resolution; the selected research question appears in the viewer; project state survives a session boundary.

**Note:** verify whether these are one bug or three before splitting — all three come from the same tester on win32 within eleven days.

Issue #1335, opened 2026-08-05
research orchestrator: reconcile the surviving prose contradictions, replace the evaluations/ globs with research_query, then decide what the rest still enforces


mercyokum:
> **Issue #1492 merged in here, 2026-08-25.** Both cards owned edits to the same
> file, and two of the three contradictions #1492 lists are the same two this body
> already lists — who writes `project.status = "completed"`, and how the mentor's
> write surface is described. Neither was separately doable: #1492 could not be
> started without the two lead rulings, and this issue's own step 1 cannot be done
> without answering the same question. #1492's body is reproduced unedited at the
> bottom as provenance.
>
> **Its line cites are stale and this body's are current.** #1492 cites
> `research/SKILL.md:152`, `:437` and `:403-404`; on `main` those are `:156`,
> `:441` and `:407-408`. Re-derive by symbol or by quoted text, not by line.
>
> **Two of #1492's four supporting facts have changed since it was written and must
> not be re-derived from its body:**
>
> - `eval/harness/validators/test_universal.py` no longer holds a hardcoded
>   `OWNERSHIP_TABLE`. It loads `docs/specs/schemas/ownership.json` through
>   `harness.ownership`, so the site to edit for the harness half is the manifest's
>   `research.json` / `project` row (`owner: "skill:init-project"`, callers
>   `init-project` + `proof-conclusion`), not the Python.
> - **The blocker that made "`proof-conclusion` writes it" unreachable is gone.** The
>   2026-08-09 analysis on #1492 ruled that option out because `proof-conclusion`
>   could not mark a question `resolved`, so its own step-8 precondition ("if ALL
>   questions are now `resolved`") was false by construction on the last question.
>   PR #1819 changed that: `agents/proof-conclusion.md` §7 now says "You own the
>   resolution of the question you actually concluded: `status: "resolved"`,
>   `resolved`, and `resolution_assertion_ids`", written in the same batch as the
>   summary — i.e. before §8 reads the condition. That option is now reachable.
>
> **The plugin hook has already been built around one of the answers.**
> `guard_project_files.py`'s `AGENT_WRITABLE_SECTIONS` grants the `proof-conclusion`
> agent `{proof_summaries, questions, project}`. It permits that agent; it denies
> the orchestrator nothing, since `project` is not in `OWNED_SECTIONS`. So the hook
> does not settle the question, but a plane in production is already shaped for
> `proof-conclusion`.
>
[...]

PR #1438, opened 2026-08-07
eval-harness: shadow detector for unpersisted conflict resolutions (#1317)

francis-2008-happy:
## The issue (#1317)

An alpha tester's Conflicts section showed blank even though the agent found and resolved a conflict mid-research. The submitted research.json was fully populated (25 assertions, 2 proof summaries) with a full "Conflict Discussion" in the proof narrative and `conflict_resolution: "Ella Chase marriage conflict resolved…"` in its exhaustive declaration — yet `conflicts: []` and every `resolved_conflict_ids: []`. The conflict was reasoned through by proof-conclusion / research-exhaustiveness and never routed to `conflict-resolution` (the sole writer of `conflicts[]`), so nothing structured was persisted and the viewer had nothing to show.

## The fix

A shadow-mode detector — the conflict-side sibling of the #1133 citation-nulling detector — that catches exactly this "concluded-in-prose-but-not-persisted-structurally" class.

- `find_unpersisted_conflict_resolutions()` (`skill_invocation.py`) reads the final research.json, gated on a written `proof_summaries` entry, and fires when a concluded question's `stop_criteria.conflict_resolution` asserts a resolution that no `status: "resolved"` `conflicts[]` entry backs. Reliance is read from that **structured** field, not free narrative; honest "no conflicts" phrasings are skipped, and a resolution claim needs positive resolution language rather than merely the absence of "no conflict" wording (the field is required, so absence-based matching would default to firing).
- Backing is checked four ways, any of which silences the check: the id is cited on the proof_summary's `resolved_conflict_ids`; a resolved conflict names this question in its `blocks_question_ids`; the stop-criterion prose names a `c_` id that **exists as a resolved entry**; or the prose names no id at all and a resolved entry exists. `blocks_question_ids` alone is too weak — it is schema-required but legitimately empty, so most resolved conflicts do not carry it.
- Wired into the existing `guardrail_shadow_violations` plumbing in `orchestrator.py`, counted in its own bucket by `guardrail_shadow_report.py`. Shadow mode only — logs, never fails a run.
- Fixed a latent mis-bucketing: the new `kind` also carries `detail`, so `scan_provenance` (the #963 bucket) now excludes it explicitly.

## What this does not solve

This is the **measurement/safety half** — it *catches* the miss in eval runs; it does **not** yet change production behavior so the conflict gets persisted. Shadow-first here matches the #1358 rollout; promotion to a hard gate waits on reading the corpus fire rate.

The behavioral follow-up (a proof-conclusion decline-and-route nudge so a `c_` entry actually gets written → viewer shows it) and the research-exhaustiveness prose-only stop-criterion were filed as #1542 and #1543 and then **closed `not planned` on 2026-08-11, merged into #1335 as "Design A"** — #1335 is the class both are instances of (doctrinal skills bypassed while their effects get written anyway), proof-conclusion is already in its touch list, so folding them in costs no extra eval run. **#1335 is the live follow-up; #1542/#1543 are not open work.** Its sequencing gate behind this PR is intact.

## Verification

- **Acceptance check:** corpus replay of `find_unpersisted_conflict_resolutions` over `eval/runlogs/e2e/*/run-*.final-research.json` reports **4 of 149 runs firing, all four with an empty `conflicts[]`** — 3 clear true positives, 1 borderline (`gertrudis-hoffmans-daughter`, "…discrepancy resolved … No genuine conflicts."), 0 false positives. `cd eval/harness && uv run --frozen pytest -q` → **1956 passed**; `uv run --frozen ruff check .` → all checks passed.
- 18 tests for the detector cover the evidenced shape, each of the four backing clauses, an unresolved/dangling citation still firing, a named id that exists but is not resolved still firing, all "no conflicts" phrasings staying silent, the positive-marker and non-resolution-opener rules, the no-proof-summary gate, dedup, and None/empty defensiveness — plus three in `test_guardrail_shadow_report.py` for the new bucket, its exclusion from the #963 bucket, and runs-vs-entries counting.
- **The orchestrator wiring has no test** — accepting the #1133 sibling's precedent, since no shadow-wiring in this repo is covered by one.
- `make e2e-guardrail-shadow` prints `0 … across 0 run(s)` for this family and will until new paid runs land, since `scan_conflict_unpersisted` reads stored entries. Same as the #1133 and #1193 buckets. The readable number today is the replay above.
- No skill run-log snapshot touched (harness-only change plus one spec), so no eval-skill run required.

## For review

- **Start here:** `skill_invocation.py` → `find_unpersisted_conflict_resolutions` and its docstring, for the gate, the reliance signal, and the four backing clauses.
- **Unsure about:** the `_NO_CONFLICT_SUBSTRINGS` / `_NO_CONFLICT_EXACT` / positive-marker heuristic on the structured `conflict_resolution` field is the one judgment call. It is a text heuristic and has a known ceiling — `gertrudis-hoffmans-daughter` ("…discrepancy resolved … No genuine conflicts.", `conflicts: []`) fires, and widening the negative list to silence it would also silence genuine hits like "…is fully resolved … No other conflicts remain open." Shadow mode is deliberately how we calibrate this before any gating; #1335 is the structural way out.
- **Didn't change:** no production behavior, no skill body, no schema, no `compliance`/`outcome` path.

Refs #1317 — measurement half only; the viewer-facing acceptance criterion (a recorded conflict appears in the viewer's Conflicts section) is NOT met by this harness-only PR, so it stays open. Behavioral half: #1335. Lost-files half postponed per lead — Cowork sync, out of repo.


clack391 (approved):

T-FEH (changes requested):
Thanks for this — the shadow-first staging is the right call, and mirroring the #1133 detector is the right shape. I'm requesting changes on the detector's **accuracy**, because the number it produces is the whole deliverable and right now it isn't trustworthy.

I pulled every `conflict_resolution` string the committed corpus actually contains (142 distinct across 566 runs) and ran your function over them. 38 distinct strings / 71 occurrences fire. Three problems:

### 1. It fires when the conflict *was* persisted

`skill_invocation.py:1017` — the backing check only reads the proof_summary's `resolved_conflict_ids`. It never looks at `conflicts[]`, which is loaded at :1010 and used solely to build the id set. So a run that correctly wrote `c_001` with `status: "resolved"` — viewer's Conflicts section populated, which is exactly what #1317 asks for — still fires if that id isn't cited on the proof_summary. The `detail` at :1038 then emits something factually false: *"no resolved conflicts[] entry backs it"*, when one does.

```
You wrote:      if any(rc in resolved_conflict_ids for rc in rcids):
Change it to:   if any(rc in resolved_conflict_ids for rc in rcids) or qid in resolved_blocked_qids:
```

and build that set above the loop from the real schema field `conflicts[].blocks_question_ids`:

```python
resolved_blocked_qids = {
    q
    for c in conflicts
    if isinstance(c, dict) and c.get("status") == "resolved"
    for q in (c.get("blocks_question_ids") or [])
}
```

```
You wrote:      "conflict_resolution, but no resolved conflicts[] entry backs it "
                "and resolved_conflict_ids is empty"
Change it to:   "conflict_resolution, but no resolved conflicts[] entry is linked to it "
                "(resolved_conflict_ids cites none, and no resolved conflict blocks this question)"
```

### 2. It fires on text that says the conflict is unresolved

`skill_invocation.py:995-1002` — `stop_criteria.conflict_resolution` is a **required** field on every declared `exhaustive_declaration` (`research.schema.json`, `stop_criteria.required`), so it is always populated. That means "doesn't match the short negative list" defaults to **fire**. Real corpus text from `elisabetha-sugecz-parents`:

> `UNRESOLVED. Conflict c_001 (identity ambiguity …) cannot be resolved with the available records.`

with `conflicts: [{"id": "c_001", "status": "unresolved"}]` and `resolved_conflict_ids: []` — i.e. the *correct* persistence for an unresolved conflict — fires, and is reported as relying on a resolved conflict. **18 of the 38 firing strings explicitly negate a resolution**: `UNRESOLVED.`, `NOT MET.`, `PARTIAL.`, `Partially met.`, `Not met —`, `One unresolved conflict exists (c_001)`.

### 3. Honest "no conflicts" phrasings get through

`skill_invocation.py:929` — `_NO_CONFLICT_EXACT` needs a whole-field match, so `"N/A — no evidence to conflict."` (3 occurrences in the corpus) fires. So do `"No active conflicts on q_001. …"`, `"Yes — no substantive conflicts identified."`, `"Yes — no genuine conflicts identified."`, and bare `"Yes."` / `"All records agree."`. The over-skip direction raised in the comment thread is possible, but it isn't the one the corpus exhibits — the dominant error is **under**-skipping.

Both 2 and 3 go away if the check requires *positive* resolution language instead of the absence of negative language:

```
You wrote:      if any(s in crl for s in _NO_CONFLICT_SUBSTRINGS):
                    return None
                return cr
Change it to:   if any(s in crl for s in _NO_CONFLICT_SUBSTRINGS):
                    return None
                if _NON_RESOLUTION_OPENER_RE.match(crl):
                    return None
                if not _RESOLUTION_MARKER_RE.search(crl):
                    return None
                return cr
```

```python
_RESOLUTION_MARKER_RE = re.compile(
    r"\b(resolved|resolution|reconciled|outweighs?|preferred assertion)\b", re.I
)
_NON_RESOLUTION_OPENER_RE = re.compile(
    r"^\s*(unresolved|not met|partial|partially met|n/?a\b|not applicable|none)", re.I
)
```

The `\b` matters — it's what stops `resolved` matching inside `unresolved`.

### 4. The reporting half has no tests, and the bucketing fix isn't protected

I verified this by mutation in a throwaway worktree:

- Reverting the `CONFLICT_UNPERSISTED_KIND` exclusion in `scan_provenance` (`guardrail_shadow_report.py:164`) → **1854 passed, 6 skipped.** That exclusion is the double-counting bug `62e1baa3` ("resolve botched merge in scan_provenance kind-exclusion") already had to fix once on this branch, so it can silently come back on the next merge.
- Making `scan_conflict_unpersisted` return `[]` **and** corrupting `format_conflict_unpersisted`'s output → **1854 passed.**
- Deleting the orchestrator wiring block so the detector never runs in any e2e run → **1854 passed.**

`test_guardrail_shadow_report.py` has a dedicated test per shadow kind for exactly this reason (`test_scan_citation_nulling_picks_up_only_citation_entries`, `test_scan_provenance_excludes_citation_entries`, `test_format_citation_nulling_counts_runs_not_just_entries`). This PR adds a third kind, a new `scan_`, a new `format_`, and a new `scan_provenance` exclusion, and adds no test to that file at all. Please add the three mirroring ones:

```python
def _conflict_entry(psid="ps_001"):
    return {
        "index": -1, "tool": "research.json",
        "required_skill": "conflict-resolution", "question_id": "q_001",
        "kind": CONFLICT_UNPERSISTED_KIND,
        "detail": f"proof_summary {psid} (question q_001) relies on a resolved conflict",
    }


def test_scan_conflict_unpersisted_picks_up_only_conflict_entries(tmp_path):
    p = _write_result(tmp_path / "fx", "run-1.json", [_provenance_entry(), _conflict_entry()])
    out = scan_conflict_unpersisted([p])
    assert len(out) == 1 and out[0]["kind"] == CONFLICT_UNPERSISTED_KIND


def test_scan_provenance_excludes_conflict_unpersisted_entries(tmp_path):
    p = _write_result(tmp_path / "fx", "run-1.json", [_provenance_entry(), _conflict_entry()])
    prov = scan_provenance([p])
    assert len(prov) == 1 and prov[0]["required_skill"] == "person-evidence"


def test_format_conflict_unpersisted_counts_runs_not_just_entries(tmp_path):
    a = _write_result(tmp_path / "fx1", "run-1.json", [_conflict_entry("ps_001"), _conflict_entry("ps_002")])
    b = _write_result(tmp_path / "fx2", "run-1.json", [_conflict_entry("ps_003")])
    text = format_conflict_unpersisted(scan_conflict_unpersisted([a, b]))
    assert "3 concluded question(s)" in text and "across 2 run(s)" in text
```

(plus `CONFLICT_UNPERSISTED_KIND`, `scan_conflict_unpersisted`, `format_conflict_unpersisted` in that file's imports.)

### 5. The dedup test cannot fail

`test_skill_invocation.py:1152` — the fixture builds one `proof_summary`, so the `seen` guard at :1024-1027 is never reached; the test is behaviourally identical to `test_conflict_unpersisted_fires_on_the_evidenced_shape`. I deleted the `key = (ps_id, qid)` / `if key in seen` / `seen.add(key)` block outright and got **8 passed**. Two-line fix:

```
You wrote:      research = _research_with_conflict_claim(
                    conflict_resolution="Conflict resolved."
                )
Change it to:   research = _research_with_conflict_claim(
                    conflict_resolution="Conflict resolved."
                )
                # the same conclusion listed twice must still yield one violation
                research["proof_summaries"].append(dict(research["proof_summaries"][0]))
```

### 6. Please don't close #1317 with this

#1317 is a `feedback`-labelled alpha-tester report whose acceptance is *"a conflict the agent records appears in the viewer's Conflicts section with its resolution"*, and whose `Touches:` line names `packages/viewer-ui/src`, `apps/server/app`, and `project-io.ts`. This diff touches only `eval/harness/`, and the body says itself that production behaviour is unchanged. Merging as written marks the tester's blank Conflicts section as fixed.

Also, the two follow-ups the body defers (the proof-conclusion decline-and-route nudge, and research-exhaustiveness's prose-only stop criterion) don't exist as issues yet — I checked. Both need a skill change plus a paid eval run + annotation, which is precisely the exception CLAUDE.md wants filed *in the PR that defers it*.

```
You wrote:      Closes #1317 (conflicts half; lost-files half postponed per lead — Cowork sync, out of repo).
Change it to:   Refs #1317 — measurement half only; the viewer-facing acceptance criterion is not met by this PR. Behavioral fix: #<new>. research-exhaustiveness stop-criterion: #<new>.
```

### Smaller things

- `docs/specs/e2e-test-spec.md:1111` still says *"A fourth check runs in shadow mode only"* and documents the #1133 sibling in full — this fifth family needs the same paragraph, and `Makefile:531`'s help text for `e2e-guardrail-shadow` needs the new family too. `format_conflict_unpersisted` is also the only formatter with no `§` prefix, which is a symptom of the same gap.
- `guardrail_shadow_report.py:185` — zero blank lines before `@dataclass` after the new function (PEP 8 wants two); same at :430-431. Won't break CI (`[tool.ruff.lint] select = ["F821"]` only), so flagging it here.
- Import ordering: `find_unpersisted_conflict_resolutions` is inserted before `find_effects_without_invocation` in `orchestrator.py:49` and `test_skill_invocation.py:20`, breaking the otherwise-alphabetical blocks.
- `skill_invocation.py:1014` — `questions_by_id` keys on `q.get("id")`, so a question with no `id` binds `None` and a `proof_summary` missing `question_id` can match it. Defensive nit.
- `docs/architecture.md:887` asks for a new guardrail's instrument + enforcing-vs-shadow status in `guardrail-enforcement-spec.md` §4; this one is absent. In fairness the #1133 sibling is absent too, so this follows precedent rather than introducing the omission.

### One thing worth knowing before planning graduation

I ran `scan_conflict_unpersisted` over all **566** committed runs: **0 entries** (the #1133 sibling also returns 0). Committed runlogs carry `tool_calls` + `guardrail_shadow_violations` but no final `research.json`, and unlike `replay_provenance` there's no replay path. So the fire rate the graduation decision needs can only come from new paid runs. Not something to fix here — just don't wait on the corpus for it.

### What I ran / couldn't run

On `02c43b8e`, 0 commits behind `origin/main` (no test-merge needed): full harness pytest **1860 passed, 0 failed** (90s); `ruff check .` clean; five separate mutations as described above, all in a throwaway detached worktree since removed. CI is green on this head for `pytest`, `runlogs`, `e2e-fixtures`, `vitest`, `lockfile-drift`, `scan`.

Could **not** verify: the detector on a live e2e run (needs a paid run + FS auth) — I substituted the real function applied to every `conflict_resolution` string the corpus contains; and the `john-applegarth-family` case named in the body and docstring, for which no fixture or research.json exists in the repo (`applegarth-parents-1936` is a different fixture), so I reconstructed the evidenced shape from the PR description. Engine/web and `apps/server` suites not run locally — diff is Python-only under `eval/harness/`, and both are green in CI.

Minor note: the test counts in the body (1680) and in the two peer reviews (1691, 105) don't reconcile with the 1860 I measured on this head, so none of those numbers was taken on what's about to merge.

T-FEH (changes requested):
Thanks — this is a real improvement, and the numbers show it. I replayed the detector over every committed final `research.json` in the corpus: firings dropped **19 → 6**, and the isolated-string sweep dropped **44 distinct / 92 occurrences → 25 / 57**. Findings 3, 5 and 6 are closed, and I confirmed by mutation that the three new report tests and the dedup test now actually fail when the thing they guard is broken.

**First, a correction to my last review.** I said the corpus fire rate couldn't be measured because runlogs carry no final `research.json` and there's no replay path. That was wrong. All 145 result JSONs in the report's own corpus have a committed `.final-research.json` sibling (I verified 145/145), so the number is fully replayable today, exactly as `replay_provenance` does for the #963 family. More on that below.

Two things still block.

## 1. It still fires when the conflict was persisted

`skill_invocation.py:1053` — the `or qid in resolved_blocked_qids` clause is the right idea, but `blocks_question_ids` is schema-required and legitimately empty: a conflict resolved *during* a question need not have *blocked* it. In the corpus only **16 of 69** conflict entries have it populated, and **46 are `status: "resolved"`**. So applied alone the patch closes exactly **1 of the 19** prior firings — only `hannah-earnest-children` relies on it.

Two of the six runs that still fire wrote `c_001` with `status: "resolved"`, a full `independence_analysis` + `weighing_analysis`, and `blocks_question_ids: []` — the Conflicts section populates, which is precisely what #1317 asks for — and we report them as unpersisted:

- `eval/runlogs/e2e/mary-dwyer-father/run-2026-07-28_19-46-08.final-research.json` — `stop_criteria.conflict_resolution` = "YES. Birth year conflict for MNFL-T24 resolved (c_001, preferred a_019 …)", `ps_001.resolved_conflict_ids: []`
- `eval/runlogs/e2e/jimmie-jewel-neal/run-2026-07-30_14-32-18.final-research.json` — same shape

The `detail` string is now literally true ("not linked"), but the bucket name, the docstring ("the resolution lives only in prose … Conflicts section is blank") and `format_conflict_unpersisted`'s "relying on an **unpersisted** conflict resolution" are all false for those two runs. That's a third of the signal reading as the inverse of what it says — and this count is what #1542 and #1543 are sequenced on.

Suggested shape: treat any resolved `conflicts[]` entry whose id the stop-criterion text names as backing, and if it names none, treat the presence of a resolved entry as backing too.

Related residual on the same line of reasoning: the new opener regex is `^`-anchored, so the *truncated* string your test pins (`"One unresolved conflict exists (c_001)."`) is silent, but the **real** corpus string it came from (`hannah-earnest-children`, ending "…cannot be definitively resolved with available sources.") still fires — `\bresolved\b` matches mid-sentence. With `conflicts: [{"id": "c_001", "status": "unresolved", "blocks_question_ids": []}]`, which is the correct persistence for an unresolved conflict, it fires and is reported as relying on a resolved one.

## 2. The new spec paragraph says something the code doesn't do

`docs/specs/e2e-test-spec.md:1143-1152`.

```
You wrote:      no `status: "resolved"` `conflicts[]` entry backs it — neither cited on
                the proof_summary's `resolved_conflict_ids` nor linked via a resolved
                conflict's `blocks_question_ids`
Change it to:   no resolved `conflicts[]` entry is *linked* to the conclusion — neither cited
                on the proof_summary's `resolved_conflict_ids` nor naming the question in a
                resolved conflict's `blocks_question_ids`. A resolved conflict that exists but
                is linked to nothing still fires, so read this count as "reliance not linked",
                not "conflict never written".
```

The two clauses aren't equivalent and the code implements the second. As written the spec licenses a future reader to trust the count as "conflicts never written", which it isn't. Fix item 1 and the sentence becomes true either way.

## Three more — not blocking, but please answer the first in the body

**The orchestrator wiring is still unguarded.** Deleting the whole block at `orchestrator.py:2031-2050` leaves the full suite at **1860 passed, 6 skipped** — identical to baseline — and `ruff check .` clean (harness lint is `select = ["F821"]` only, so even the orphaned import is invisible). The detector can be made to never run in any e2e run with green CI. In fairness no test in this repo covers *any* shadow-wiring, including the #1133 sibling's, so "I'm accepting the sibling's precedent" is a fine answer — just say it rather than leaving the round-1 finding silently unaddressed. If you'd rather close it: one test that monkeypatches `find_unpersisted_conflict_resolutions` and asserts its output reaches the returned `guardrail_shadow_violations`.

**The fire rate is readable today and the report doesn't read it.** `make e2e-guardrail-shadow --since all` on this head prints `0 concluded question(s) … across 0 run(s)`, because `scan_conflict_unpersisted` reads *stored* entries and there are none until new paid runs land. A ~15-line `replay_conflict_unpersisted` mirroring `replay_provenance` would give the #1542/#1543 gate its number now instead of after a paid suite. Worth folding in. If you'd rather not, here's the number: **6 of 145 runs fire (4.1%) — 3 clear true positives, 1 borderline, 2 false positives.**

**The marker regex misses its own inflections.** `"The birthplace conflict resolves in favour of Ireland; the 1850 census outweighed the death certificate."` with `conflicts: []` is a textbook #1317 case and does not fire. Corpus impact today is 1 of 165 distinct strings (and that one has a persisted resolved conflict anyway), so this is forward-looking, not a present miscount.

```
You wrote:      r"\b(resolved|resolution|reconciled|outweighs?|preferred assertion)\b"
Change it to:   r"\b(resolv(?:ed|es|ing)|resolution|reconcil(?:ed|es|ing)|outweigh(?:s|ed|ing)?|adjudicated|preferred assertion)\b"
```

## Smaller things

- `gertrudis-hoffmans-daughter/run-2026-08-03_21-14-11` fires on "Apparent surname discrepancy resolved … **No genuine conflicts.**" with `conflicts: []`, arguably the correct persistence. **Don't add another phrase for it** — widening the negative list would silence genuine hits like "…is fully resolved … No other conflicts remain open." That's the ceiling of a text heuristic, and #1543 is the right way out. Recording it, not asking for it.
- `guardrail_shadow_report.py:158-166` — `scan_provenance`'s docstring still says only "EXCLUDING the #1133 citation-nulling class" while the code now excludes three kinds. `_scan_stored`'s docstring still lists two families.
- `skill_invocation.py:1014` — the `q.get("id")` truthiness fix is correct but untested; reverting it leaves 108 passed.
- `orchestrator.py:2022` and `:2038` — `read_research_json(workspace)` called twice in a row for the two post-hoc detectors. One local would read better.
- PR body **Verification** section is stale: "1680 passed / 8 new tests" against **1860 passed / 14 new tests** on this head. The body also skips the repo template's `**Start here:**` / `**Unsure about:**` / Didn't change / **Acceptance check** / Deviated-from-the-plan fields — the substance is there under other headings, but there's no named acceptance check.
- Round-1 nits now fixed: spec documents "a fifth check"; `format_conflict_unpersisted` carries the `§7.5` prefix; blank lines before `@dataclass`; alphabetical import blocks in `orchestrator.py` and both test files; `questions_by_id` no longer binds `None`. Still open: `docs/specs/guardrail-enforcement-spec.md` §4's layer table has no row for this guardrail, which `docs/architecture.md:887` asks for — the #1133 sibling is missing too, so precedent rather than a new omission.

## What I ran

Primary checkout at `cf926d4b`, 0 commits behind `main`: `uv run --frozen pytest -q` → **1866 passed** (107s). Throwaway detached worktree baseline → **1860 passed, 6 skipped** (skips are "engine build absent", environmental); `ruff check .` clean.

Mutations (all in the worktree, reverted between runs, worktree removed): revert the `scan_provenance` kind-exclusion → **1 failed** (caught); `scan_conflict_unpersisted → []` → **2 failed** (caught); corrupt `format_conflict_unpersisted` → **1 failed** (caught); delete orchestrator wiring → **1860 passed** (*not* caught); delete dedup `seen` guard → **1 failed** (caught); revert `blocks_question_ids` backing → **1 failed** (caught); remove opener skip → **1 failed** (caught); remove positive-marker requirement → **2 failed** (caught); drop `\b` from the marker regex → **1 failed** (caught); revert `q.get("id")` filter → **108 passed** (not caught).

Corpus replays: real detector over 165 distinct / 337 occurrence `conflict_resolution` strings, old vs new → 44→25 distinct, 92→57 occurrences. Real detector over all 145 committed `final-research.json` → 19→6 firings (marker-only variant 7, blocks-only variant 18).

**Could not verify:** the detector on a live e2e run (needs a paid run + FamilySearch auth) — the 145-document replay is the substitute; the `john-applegarth-family` case named in the body and docstring, for which no fixture or research.json exists in the repo (`grep -rl "Ella Chase"` → nothing; `applegarth-parents-1936` is unrelated), so the evidenced shape stays reconstructed from prose in both the PR and the test; engine/web/`apps/server` suites, not run locally (diff is Python under `eval/harness/` plus one doc, and `pytest`, `runlogs`, `e2e-fixtures`, `vitest`, `lockfile-drift`, `scan` are green on this head); CODEOWNERS team membership (`gh api …/teams/<slug>/members` → 404 with this token).

Ownership note: last-rule-wins in `.github/CODEOWNERS` puts `*.py` on **senior-developers** and `docs/specs/e2e-test-spec.md` on **senior-genealogists**, so both teams gate this now that the spec is in the diff. `require_last_push_approval: true` means a fresh approval is needed after `cf926d4b` regardless; the two `@02c43b8e` reviews weren't reset and still count toward the two required.

T-FEH (approved):

EdmondOware (approved):
The heuristic is more careful than a first pass might suggest. The code comment documents that an earlier round of review already caught the naive version over-firing on negated resolutions like "UNRESOLVED." and "Not met," which is why it now requires positive resolution language rather than just the absence of negative language. That's the right fix for that failure mode.

 One observation on the remaining known gap, the gertrudis-hoffmans-daughter borderline case. "Discrepancy resolved, no genuine conflict" and "fully resolved, no other conflicts remain open" may not be equally ambiguous. "Discrepancy" without "conflict" plausibly means the agent checked an apparent contradiction and concluded it never rose to a real GPS conflict at all, a name variant or transcription error explained away, while "conflicts... resolved" explicitly invokes conflict-resolution vocabulary, meaning something was actually weighed. If that distinction holds across more examples, keying on "discrepancy" alone versus "conflict" might cut the false positive rate without the false negative risk noted in the PR. Worth keeping for whoever calibrates this before any promotion to a hard gate, not something to fix now.

 This is shadow mode only, so it logs without ever failing a run, and the known ceiling is disclosed honestly rather than papered over.

PR #1440, opened 2026-08-07
test(e2e): resolve mary-mcandrew-son as false match (outcome 3)

Paaboat:
## Summary

- Resolves fixture `mary-mcandrew-son` as a **false match** (outcome 3) for the `resolve-record-hint` skill (GitHub issue #637)
- Adds 2 e2e run logs with `.ann.json` annotation siblings
- Tier: Normal

## Review guidance

The key decision is outcome 3 (false match): the 1900 census record hints at a son for Mary McAndrew but the evidence does not confirm identity. Start with `eval/runlogs/e2e/mary-mcandrew-son/` to see the two runs and their grading annotations.

## Acceptance check

Neither committed run reaches a clean false-match conclusion:
- Run 1 (2026-08-05_18-11-41): MCP server never connected — f1 passed by inaction, f2 not recovered. Not a genuine pass.
- Run 2 (2026-08-05_20-48-08): Agent overclaimed — added a sixth child and concluded 7+ children born. Both f1 and f2 failed.

Landing without a passing run is intentional: this is a known hard fixture (negative conclusion / polarity-avoid). The runs are committed as evidence of current agent behavior.

## Deviated from the plan

Both runs failed to reach the expected false-match conclusion. The agent either did nothing (Run 1) or went in the opposite direction (Run 2). No deviation from authoring intent — the fixture definition is correct.


mercyokum (commented):
Before I approve, I want to flag that neither of the two committed runs actually reaches the false match conclusion. The first run did nothing because the MCP server never connected, so it "passed" by accident. The second run reached the opposite conclusion, saying Mary had seven or more children.

So we don't have a single passing run showing the skill can actually get this answer right. The PR also says "3 e2e run logs" but there are only 2, and it's missing the Acceptance check and Deviated from the plan sections, which would've been the place to mention the failing runs.

Is landing this without a passing run intentional, or do you want to try for one clean run first? Everything else looks good to me.

kofiatinka12 (approved):
I rechecked the current file rather than the diff, and the JSON is clean with no duplicate keys. The FALSE MATCH reasoning and expected findings are clear and consistent. Thanks for clarifying.

EdmondOware (approved):
Approving as the senior-genealogist review.

 On Mercy's question: I'm satisfied landing this without a clean run is the right call. This is a genuinely hard fixture — Mogan/Morgan variant spelling, common 1870s Detroit names, a hint that looks plausible on its face — and committing both runs as evidence of current behavior is more useful than waiting for a clean pass that may take several more attempts to produce.
 One thing worth tracking separately from the fixture decision, though: Run 2's failure isn't just "wrong answer" — the agent never actually engaged with the 1873 John A Morgan hint this fixture is built around. It fabricated a different sixth child ("John L Mogan") from a single 1920 census co-residence line — no stated relationship, no birth record — and then wrote that up as a probable-tier conclusion citing an index-only fertility-column count. That's a proof-tier judgment call on genuinely weak indirect evidence, independent of whether it's the right or wrong answer to this fixture's question. Worth a genealogist-labeled issue tracking whether that's a one-off or a real pattern in how conclusions get tiered against thin evidence — not blocking this PR on it, just flagging so it doesn't disappear into the run-log notes.

PR #1441, opened 2026-08-07
eval(record-extraction): pin fact values in expected_classifications (#995, #1108, #1131)

Cia-3:
- **#1108**: extends `test_expected_classifications` with a `value` matcher
  (place/date), gated on `attribute`, case-insensitive substring match in
  either direction. Lets a matcher assert the fact itself, not just its
  classification — the ut_013 defect (correct reasoning, contaminated value
  on disk) was invisible to the old matcher shape. No paid run required;
  harness-only change, full unit suite green (1684 passed, 1 skipped).
- **#995**: authored value matchers on `ut_013` (the reported defect) plus
  6 other tests where the record content deterministically fixes the
  answer. Deliberately left **9 tests unmatched** — role naming or the
  relationship-recording perspective (child-records-it vs
  parent-records-it) proved non-deterministic across runs, confirmed by
  direct comparison, not assumed. Per #995's own principle ("only declare
  pairs and values the doctrine fixes deterministically"), that's
  compliance, not a shortfall — see Review guidance.
- **#1131**: two evidence-grounded fixes in `record-extractor.md`, both
  pulled from actual judge rationales in the prior run's annotation, not
  guessed — a bare marriage index's informant defaults to the parties
  themselves, not a fabricated clerk; an obituary's stated facts are
  `direct` regardless of secondhand relay. Added the spawn-failure stop
  clause to `record-extraction/SKILL.md`, reusing #1196's deny text.
- **Bonus, found via this run's annotation**: a third classification drop
  — pre-1940 census `information_quality`: a plausible informant guess is
  not a *known* one, so decision-tree Question 1 still gates it to
  `indeterminate`. Fixed and verified (`ut_026`: fail → pass).

## Review guidance

**Start here:** `ut_013`'s matcher (the actual reported defect) and the two
`record-extractor.md` diffs — both are short and each is anchored to a
specific judge rationale quoted in the commit.

**Unsure about:** whether the 9 unmatched tests should get a harness
extension instead of staying unmatched. Filing that as its own issue
rather than deciding here — see Follow-on issues.

## Test plan

- [x] `make eval-skill SKILL=record-extraction` — 27 tests, genealogist-
      annotated (`v1_2026-08-07_10-52-54.json` / `.ann.json`)
- [x] Full harness unit suite — 1684 passed, 1 skipped
- [x] Every edited test file loads + schema-validates via
      `harness.loader.load_test`

## Follow-on issues

Filing two separately — different lanes, same reasoning #995 used to
split off #1108:

1. **Harness extension** — `expected_classifications` can't express "this
   relationship exists, on either party's record_role" or "either of these
   two naming conventions is acceptable." Both are genealogically valid;
   the matcher schema just can't express the OR. `developer` lane.
2. **Marriage/death-cert run-to-run instability** — across 5 runs this
   session, `ut_021` failed three different ways, `ut_024` flipped
   pass/partial/fail repeatedly, and `ut_015`/`ut_028`'s parent-name check
   recurred intermittently even after a real, verified fix. Documenting as
   still-open rather than claiming it's settled. `genealogist` lane,
   `cluster:next-run`.


kofiatinka12 (approved):
This cleanly closes the gap between verifying classification and verifying the persisted value itself. The schema, validator, documentation, tests, and fixture updates all line up, and the implementation is well covered by unit tests. One optional follow-up would be to add a regression test for the case where both place and standard_place are populated but differ, to document the intended matching behavior, but I don't consider that a blocker.

PR #1203, opened 2026-08-03
e2e: resolve matyas-muck-son record hint + fix multi-image ARK bug

benter-070:
## Summary
- Resolves issue #869: adjudicates the `matyas-muck-son` record-hint fixture. FamilySearch's own hint checks out as a **true match** — Mátyás Muck and Erzsébet Wolf had a second son named András, born 19 October 1887 (christened 22 October), Bikács, entry #44 in "Hungary, Reformed Church Christenings" — after ruling out a false-match misreading and an identity conflict ("Mihály" vs "Mátyás"), both traced to unreliable OCR/index reads of the source image. `expected-findings.json`/`README.md`/`fixture.json` document the full adjudication trail, including two superseded intermediate resolutions.
- Fixes a real MCP-tool bug found while adjudicating: `image_transcribe`/`image_read` resolve a bare `3:1:`/`3:2:` document-image ARK assuming it's self-sufficient, but some are waypoints into a multi-image film/register where the bare resolver lands on the wrong image without `i=`/`cc=`/`groupId=` context from the page URL. `arkToImageUrl` (`fs-image-fetch.ts`) forwards that context when the caller's `ark` is a full URL carrying it.
- Files **#1205** for the residual gap: two scored e2e runs after the fix still couldn't reach the correct page autonomously, because `record_read` never surfaces that query context in the first place — only a human-supplied URL does. The fixture's README flags it as known-hard/currently-unsolvable so future runs aren't spent expecting a pass until that gap is closed.
- Merged `main` twice to stay current: once to resolve a real conflict (`docs/TODOs.md` retired upstream per #1163 while this branch had added an entry to it — resolved by filing #1205 as an issue instead), once more just to pick up unrelated upstream progress.

## Review follow-ups (Israel/Asimi1234)
- **Real regression, fixed:** forwarding `i=`/`cc=`/`groupId=` unconditionally broke single-image ARKs where `i=` is out of range — FamilySearch returns a 200 with an HTML body, not an error, and `fetchFsImageBytes` threw on a document that worked fine before this PR. `resolveFsImageInput` now also returns a `fallbackUrl` (the bare ARK) whenever it forwarded context params, and `fetchFsImageBytes` retries it when the primary URL doesn't return an image. `image_read`'s `metadata.url` now reports whichever URL actually worked, not the failed primary.
[...]

PR #1453, opened 2026-08-07
Image multiimage ark finding 1205

aghadiayeamayanvboernest:
docs(image-read): record the multi-image ARK routing finding + fix a stale tool hint (#1205)

#1205 asked whether an autonomous run can reach the correct page of a multi-image FamilySearch document ARK — it holds only the `1:1:` index ARK, `record_read`'s `sources[]`, and the bare `3:1:` image ARK, never the browser URL with `i=`/`cc=`/`groupId=`. Investigated live and recorded the answer in the spec. **The answer refutes the issue's premise: a host-side route does exist** — but only via the ARK's JSON metadata, not the bare-ARK byte path.

## Summary

- **Finding (the deliverable):** an autonomous route **exists**. Fetching the bare ARK's resolver URL as `application/json` exposes a clean `{group}_{sequence}` `imageId` in its `name` field (`004707850_00113`); passing that to `image_read`/`image_transcribe` lands on the **correct** page. Verified live (`image_transcribe 004707850_00113` → the 1887 register spread with the target entry #44, 19 Oct 1887; independently re-verified by @benter-070 on 2026-08-12).
- **Scope of the claim (per review):** only the JSON `name` route is verified correct. The bare-ARK **byte** path is *not* — the fixture and a scored run show those bytes landing on the wrong (1881) page, which is exactly why #1203's `i=`/`cc=`/`groupId=` forwarding exists. This PR does not touch that.
- **Mechanical fix:** `record_search`'s `imageGroupNumber` hint pointed at `image_search` (which *takes* a group number, can't find one); corrected to `volume_search` in all three copies in the repo.
- Tier: **Normal**.

## Review guidance

**Start here:** the "Is there an autonomous route to the correct page?" block in `docs/specs/image-read-spec.md`. The load-bearing distinction is between two resolution paths of the *same* bare ARK: the JSON `name` route (verified → correct page, needs no browser URL) versus the `image/*` byte path (unverified here; the fixture shows it lands wrong, and it still relies on #1203's forwarding). Please confirm the wording keeps those separate and reads as **no licence to undo #1203**. The recommended follow-up (surface `name` from `record_read`/the resolver) is scoped as a senior tool-contract change, not a skill edit.

**Unsure about:** nothing structural. One judgment call: the spec's issue-ref baseline is 0, so the finding cites mechanisms by symbol (`arkToImageUrl`, the probe script) rather than `#NNNN`; the ticket pointers live here in the PR body.

## Plan

**Didn't change:** no skill, agent, or tool *output shape*. Did **not** teach any skill to page through the volume (#1081's unbounded hunt), and did **not** re-implement or weaken the `i=`/`cc=`/`groupId=` byte-path forwarding (#1203 owns it, merged).

**Acceptance check:** the deliverable is a written finding + a doc fix, so the check is the doc-links ratchet, not a new unit test: `tests/packaging/doc-links.test.ts > specs state the fact, not the ticket` passes (image-read-spec.md holds 0 issue refs). The mechanical fix is covered by the existing engine suite (`npm test`, green).

**Deviated from the plan:** the finding changed twice under review, both recorded here. Round 1: an earlier "no route at all" read was inverted once the live JSON probe surfaced a usable `imageId`. Round 2: that inversion over-claimed (it implied the bare-ARK *byte* path lands correctly, which was never verified); scoped it to the JSON route only, per @T-FEH. Did not hit the step-4 stop rule.

## Test plan

- [x] Ran `make test-all`. Engine vitest, control-plane, eval-app, eval-harness all pass. The one red is the known **#1366** `apps/electron` flake (`buildFeedbackZip > drops the largest files` times out at 5s under parallel load; **61/61 in isolation**) — untouched by this branch.
- [x] N/A — no skill run-log snapshot changed. Touches only `packages/engine/mcp-server/**` and two spec docs; `check_runlogs` marks no skill touched, so no `make eval-skill` run is owed.

The finding itself is not CI-gradable — its instrument is the live FamilySearch API plus a human reviewer (`docs/architecture.md` §9.4).

## Follow-on issues

- Bounded follow-up (**senior**, separate PR): have `record_read` (or the resolver) read the ARK's JSON and surface `name` + the group/`parents` link — giving an autonomous run a direct, unambiguous `imageId` for the correct page without the browser URL. Changes a tool's output shape and the simplified-GedcomX surface.

Closes #1205.


T-FEH (changes requested):
Thanks — the `volume_search` correction is right, and I verified it end to end: `volume_search` really does return `imageGroupNumber` from a place plus a year range, and `image_search` requires one as input and cannot find one. Every other factual claim in the new spec block also checks out against the code, and I confirmed the acceptance check you cite is real, not vacuous — appending `Investigated under #1205.` to the spec makes `doc-links.test.ts > specs state the fact, not the ticket` fail with `image-read-spec.md: 1 issue refs, baseline 0`.

Three things before this merges. The first is the important one.

## 1. The parenthetical that carries the conclusion compares two different ID systems

You wrote (`docs/specs/image-read-spec.md:100-102`):

> **But it is the image the waypoint ARK is already bound to — the wrong page** (group `004707850`, not the target waypoint `1858355`). It identifies deterministically; it does not disambiguate.

`1858355` is the `cc=`/`groupId=` collection/group value, not a DGS image group number. This issue's own Reproduction section says so ("the correct page is `i=112`, `cc=1858355`, `groupId=1858355`"), and so does the fixture: "the correct page only resolves at image index `i=112` (collection/group `1858355`)". DGS image group numbers in this repo are always nine digits with leading zeros — `004010852`, `007621224`, `004707850`. `image_search` has a whole `resolveGroupId` step (`src/tools/image-search.ts:24-52`) whose only job is translating a DGS number into a different group identifier via `/group/{n}/apid`. So the two values being different text is the expected case, not evidence of a mismatch.

The other numbers point the opposite way. The fixture says the correct page is `i=112`. Sequence suffixes in this repo start at `_00001`, and FamilySearch's `i=` is 0-based — which would make `i=112` sequence `00113`. The JSON gave you `004707850_00113`.

So the resolver may have handed you the *correct* page, not the wrong one. If it did, the finding inverts: there **is** an autonomous route — read the ARK's JSON, take `name`, pass it to `image_read` — and the bounded follow-up you propose becomes considerably more valuable than "convert invisible into bounded".

I can't settle this myself (no FS token here), so this is "the reasoning in the parenthetical doesn't hold", not "the conclusion is false" — it may still happen to be right. Please settle it before this lands. Two commands, a few cents, both already on the issue's instrument list:

```
npx tsx dev/try-image-search.ts 004707850
npx tsx dev/try-image-read.ts 004707850_00113
```

Is that page Bikács, entry #44, 19 Oct 1887? Or the unrelated 1881 page?

Either way the parenthetical has to change, because the argument in it isn't sound. If the page turns out to be wrong:

> **But it is not the target page** — <the reason the live check gives>. It identifies deterministically; it does not disambiguate.

If it turns out to be correct, the block needs rewriting around the new answer, and it's worth saying so on the issue first.

This matters more than a normal doc nit because the block explicitly tells future readers not to re-investigate ("the reason is worth recording so it is not re-investigated"). A wrong answer there is one nobody re-opens.

## 2. The stale hint is still in a third place — in a file this PR edited

`docs/specs/record-search-tool-spec-v2.md:135`, the parameter table:

> | `imageGroupNumber` | string | ... Use the `image_search` tool first to find the image group number. |

Change `image_search` to `volume_search`. Line 764 of the same file already says `volume_search`, so the spec currently contradicts itself on the exact parameter this PR came to fix, and the PR body says the fix landed "in both the tool schema and the spec". The issue named two copies, so this isn't on you — there are three. Nothing lints the spec's parameter table against the real tool schema, so this won't get caught later. Confirmed by grep that it's the last occurrence in the repo.

## 3. Fold the correction into the 2026-08-03 paragraph rather than contradicting it ten lines down

`docs/specs/image-read-spec.md:69-71` still reads:

> **confirmed live (2026-08-03)**: a bare ARK for a multi-image entry did not resolve to any document at all (FamilySearch returned a no-match search page),

Your new block says that's contradicted. The code comment in `fs-image-fetch.ts:44-49` already carries the right instability wording from #1203, so the spec is now behind its own code comment. Suggest:

> **observed live (2026-08-03)**: a bare ARK for a multi-image entry returned a no-match search page rather than the document, though the same ARK returned a `200` on later checks — the resolver's behavior here is not stable,

---

**Verification I ran:** test-merged `origin/main` (clean), then on the merged tree with Node 22 — `tsc --noEmit` clean; `vitest` 2105/2106, the single failure being `plugin-hooks.test.ts` timing out at 5s under parallel load (14/14 in 2.38s in isolation, untouched by this diff, green in CI); `doc-links.test.ts` 122/122 plus the mutation test above. All 13 CI checks green on this head. Did not run the eval-harness or control-plane Python suites locally — no Python in the diff and both are green on this head. No §5 stop rule triggered: no schema change, no `auth/`, no ADR reversal, and the tool-contract change you propose is proposed only and correctly flagged senior.

T-FEH (changes requested):
Round 2 — findings 2 and 3 are fixed (spec `:135` now says `volume_search`, no stale copy left in the repo; the 08-03 paragraph carries the correction inline). The inversion is right, and Ben's independent transcribe backs it. Good result.

One blocking thing: the inversion over-claims. `image-read-spec.md:106` —

You wrote: `ARK returned 200 and the correct image.`
Change to: `ARK returned 200. Which image those bytes were was not checked.`

The only source for 08-05 is the code comment at `fs-image-fetch.ts:46-47`, and it says just "got a 200". Ben checked the raw `imageId`, explicitly not the resolver. And the fixture README says the bare ARK's *bytes* land on the wrong page — entries 26–37, headed 1881, two screenshots plus a scored run.

That claim is what holds up `:101-102` —

You wrote: `so name alone lands correctly; the browser params are not required.`
Change to: `so the two values differing is expected. The JSON route needs no browser URL, but the byte path still needs i=/cc=/groupId= forwarding.`

As written it contradicts line 74 of the same section and reads as licence to undo #1203.

Two smaller ones: put `ark:/61903/3:1:9392-9ZVZ-X` and the probe command back — the rewrite dropped both, and #1205 asked for the finding with its commands behind it. And update the PR body; it still opens "no autonomous route exists", and that body is what closes #1205, whose title is now the refuted premise.

T-FEH (commented):
Round 3 — all four R2 points are genuinely fixed, and suites are green test-merged with main. The JSON-vs-byte scoping now reads correctly as no licence to undo #1203.

One left, and you have to push a branch update anyway (main moved), so it costs nothing to fold in. Line 87 —

You wrote: `yes — the bare ARK's own resolver metadata names a clean imageId for the correct page.`

Change to: `the route exists, but nothing reaches it yet — the bare ARK's resolver metadata names a clean imageId, and no shipped tool fetches it (last bullet).`

As written the headline answers "yes" to "is there an autonomous route", while your own last bullet says no tool surfaces the JSON, and `matyas-muck-son/README.md:42` still says the chain "cannot autonomously disambiguate" this record.

Also: the body dates Ben's check 08-12; he posted on the 11th.

Push that plus the branch update and I'll approve. You'll still need a senior-genealogists approval — `/docs/` owns both spec files.

T-FEH (approved):
Approving. All three rounds of feedback are addressed, and the finding now says what the evidence supports: the JSON route exists, the byte path still needs #1203's forwarding, and no shipped tool reaches either — so the `matyas-muck-son` fixture stays correctly marked unsolvable.

I pushed the round-3 one-liner myself (`af14c33a`) rather than send it back for a fourth round, since the branch needed a main update anyway. Scoped the headline to what ships; nothing else touched. Suites green on the merged tree — engine 2114/2114, tsc clean, all 12 CI checks pass.

Note that my approval satisfies senior-developers for `record-search.ts`, but `/docs/` in CODEOWNERS puts both spec files under senior-genealogists, so this still needs one of them. Because I made the last push, `require_last_push_approval` means that second approval also has to be the one clearing the push — so it can't be me.

Nice work running the live probe rather than reasoning it out; the inversion was the right call.

clack391 (approved):

Issue #1456, opened 2026-08-07
record_search: otherGivenName/otherSurname carry no relativeTerms presence signal


taiwo-stack:
**Touches:** packages/engine/mcp-server/src/tools/record-search.ts

Split out of #1324, which added `relativeTerms` for `spouse`, `father`, `mother` and `parent`. `other` was deliberately left out, and this records why and what closing it would take.

**Why it was excluded.** `q.otherGivenName` matches any co-occurring person of unspecified relationship, so the only rule available from the relationship graph is "some other person is on this record". Surveying every real staged sidecar in the repo — 384 results across both `eval/e2e-project/` and `eval/fixtures/scenarios/` — **every single one is multi-person**, because search entries carry the whole household or family group.

So the answer would be a constant `present`, naming whichever co-person the resolver happened to pick, who need not be the one that matched `q.otherGivenName`. A model reading `other: {status: "present", name: "Thomas Flynn"}` on a record where the queried person is absent gets exactly the confidently-wrong, well-sourced claim #1324 exists to remove — reintroduced by the fix. A field that is always `present` is worse than no field.

**What closing it would take.** A name matcher: resolve `other` by comparing co-person names against the supplied `otherGivenName` / `otherSurname`, rather than by relationship role. That is a different and larger design than the other four prefixes, all of which resolve structurally off `ParentChild` / `Couple` rows and never compare strings. It also inherits every hard part of name matching — abbreviations (`Wm.` vs `William`), spelling variants, and the fact that FamilySearch's own matcher is fuzzy in ways we do not replicate.

Note that the same "we did not re-run the matcher" caution already applies to the four shipped prefixes, which is why their status word is `present` rather than `matched` and why the relative's `name` is always carried for the caller to judge. A name-matched `other` would need to be at least as honest.

**Current behaviour, so this is not silently missing:** a caller supplying `otherGivenName` gets no `other` entry, and the `otherGivenName` / `otherSurname` schema descriptions say so explicitly.

Filed as `icebox`: a candidate with no decision behind it.

Issue #1455, opened 2026-08-07
search-records: tell the agent to read relativeTerms before writing that a record confirms a relationship


taiwo-stack:
**Touches:** packages/engine/plugin/skills/search-records/SKILL.md, eval/fixtures/scenarios/*/results/*.json

Split out of #1324, which added a per-result `relativeTerms` field telling the agent whether the relative it anchored a search on is actually named on each record. The field is now data the model **can** read; nothing yet tells it to.

**The failure this exists to stop.** A father-anchored search returns records that name no father, because FamilySearch treats a relative term as "must not contradict" rather than "must carry". The write-up then says "confirming her father William" about a record that never mentions him — false, and filed with an ARK so it looks well-sourced. That is the whole reason #1324 exists, and the tool-side half only makes the distinction *available*.

**What #1324 already did, so this is not starting from zero.** The `projectPath` schema description — the one place enumerating what a staged result row contains — now names `relativeTerms`, says what `absent` means, and says to check it before writing that a record confirms a relationship. A model enumerating a staged row will see it. The skill change is what makes it *reliably* consulted at write-up time rather than noticed opportunistically.

**Suggested shape:** one line in the evidence/write-up step of `search-records/SKILL.md` — when a search was anchored on a relative, read `relativeTerms` before characterising the record, and say "consistent with" rather than "confirms" on `absent` or `unknown`.

**The part that is easy to miss.** The unit harness stages `record_search` payloads verbatim from `eval/fixtures/scenarios/*/results/` (`eval/harness/harness/mock_mcp.py:180`), and **none of those fixtures carry `relativeTerms`**. A skill instruction to read the field would be exercised against data where it never appears, so the fixtures need the field before any eval can tell whether the instruction works.

Not `icebox`: #1324 should not be considered fully delivered until this lands.

Issue #1454, opened 2026-08-07
rank_search_matches: let the ranker act on relativeTerms, not just carry it


taiwo-stack:
**Touches:** packages/engine/mcp-server/src/tools/rank-search-matches.ts

Split out of #1324, which added `relativeTerms` to `record_search` results and made `toStub` carry it onto the ranked stub. The ranker **copies** it and nothing more: `matchScore` is unaffected by a `father: absent`.

That was deliberate. Letting the score react to `absent` is a scoring-policy change, not a plumbing one — it needs calibration before any weight is chosen, and it would silently move every ranking in the eval corpus.

**Why it is still worth doing.** The top-ranked stub is exactly where a father-anchored hit that names no father looks *most* confirmed: it scored well on names, dates and places, and the only thing saying the father is missing is a field the model has to notice on its own. #1324's stated reason for putting this file in scope was "a scorer that cannot see 'father absent' will keep rating those hits as relative-confirmed" — carrying the field makes it *visible*, not *weighted*.

**What this needs before any code:**

- A calibration pass over the existing score log to see how `present`/`absent`/`unknown` correlate with human-confirmed matches. `absent` is not evidence *against* a match — a sparse index entry legitimately omits parents — so a naive penalty would suppress genuine records.
- A decision on whether this belongs in the score at all, versus a separate sort key or a review-band flag. `rank_search_matches` is specced as "a re-ranker and review surface, not a classifier".
- Whichever way it goes, a before/after on the eval corpus, since every ranking moves.

Filed as `icebox`: the visibility half shipped and this has no decision behind it yet.

PR #1328, opened 2026-08-05
record_search: document the exact/close qualifier family, and correct four false claims it taught (#1093)

chesworthrm:
## Summary

Tier: **Normal.** Closes #1093.

**This is no longer purely a documentation change.** It adds one tool parameter
(`batchNumber`) and a dev-only typecheck target; everything else is prose, probe,
and CI. The original description of this PR is superseded — its headline claim was
refuted by later measurement on this same branch.

**What the branch actually delivers is fewer claims than it started with**, each
traceable to a committed artifact, plus the machinery to stop the next round of
drift.

## What changed, and what reversed

The original body said *"`.exact=on` changes the result count, not the ranking."*
**That is wrong**, and was measured from a top-200 sample. Reading complete result
sets instead:

- **`.exact` is a strict subset** — every record it returns was already in the fuzzy
  set, so it cannot surface a record fuzzy buried (`surname`, marriage populations:
  521 → 81, 469 → 423, 0 exact-only).
- **But it does reorder what it keeps** — 54 records move, the largest by 34
  positions, 6 crossing rows of a different relevance score, against a same-query
  noise floor of 0.

Three further claims were **withdrawn rather than corrected**, because no instrument
we have can settle them:

- Whether an unqualified year range keeps year-silent records — `NOT MEASURED`, in
  both directions. A scope test showed the payload's dates do not predict which
  ranges return a record, so no payload-based classifier can decide it. Section H's
  original `OPEN — do not say an unqualified range keeps year-silent records` was
  right all along.
- Whether that behaviour generalises past birth — `NOT MEASURED`.
- Diminutive *exclusion* by `givenNameExact` — the fuzzy *reach* is measured; the
  exclusion is expected, not measured.

Confirmed by enumeration: **spouse `.exact` requires the spouse to be present** (two
marriage populations read in full, every spouse-silent record dropped, control
survives), and **`givenNameExact` pins initials order** (US census pool read in full
both ways; the `W J` record present in fuzzy, absent from exact, four other records
surviving).

## Behaviour change

**`batchNumber` added to `record_search`.** `collection-quirks.md` documented IGI
batch enumeration as the way to exhaust an English parish — a strategy the tool could
not execute, because it exposed no batch parameter. Verified live: a batch alone
returns several thousand records, a nonsense batch returns 0 rather than being
ignored, and it composes with a name.

## What to scrutinise

The guard rules, more than the prose. Every failure on this branch was **a check that
passed for the wrong reason**, so these are the parts that matter:

[...]

PR #1457, opened 2026-08-07
record_search: say whether the relative you anchored on is actually on the record, and a repo-wide typecheck project plus 126 type fixes (#1324)

taiwo-stack:
**Two parts, since the title previously named only the first.**

1. **`record_search` reports whether the relative you anchored on is actually named on the record** — the #1324 feature.
2. **A repo-wide typecheck project covering `tests/` and `dev/`, plus the 129 pre-existing errors it found** — 126 in `tests/`, 3 in `dev/`, all fixed, no exclusions.

Part 2 is not unrelated cleanup. Widening `findRepresentedPerson`'s return type for part 1 broke a test at runtime while `tsc` stayed green, because nothing typechecked `tests/`. Detail under "Also in this PR".

---

## The problem

A relative-name term goes out with `m.queryRequireDefault=on`, which FamilySearch reads as "must not contradict", not "must carry". So a record naming no father at all survives a father-anchored search, and nothing in the response separated a hit that confirms William from one merely consistent with him. The staged search then strips the relationship graph before the model sees it. The write-up says "confirming her father William" about a record that never mentions him, filed with an ARK so it looks well sourced.

## What this adds

A per-result `relativeTerms`, emitted only for relatives the caller named:

```jsonc
"relativeTerms": {
  "father": { "status": "present", "name": "Wm. Neal" },
  "mother": { "status": "absent" }
}
```

Resolved inside `mapEntry` before the slim block strips `gedcomx`, so it survives into the inline stub and the staged sidecar, and is carried onto the ranked stub. The search itself is unchanged.

Three consumers, so the field is not inert: the `projectPath` schema description names it, `rank_search_matches` emits `relativeTermNote` when a returned match is `absent`, and `search-records` now writes "consistent with" rather than "confirming" on `absent` or `unknown`.

## Decisions worth reviewing

**`present`, not `matched`.** We do not re-run FamilySearch's fuzzy matcher, so we must not claim to have. The tool reports that a relative in this role exists and gives the name, so `Wm.` against a query of `William` stays visible and checkable.

**Anything not provably `Male` or `Female` resolves `unknown`**, including the literal `"Unknown"` that `simplifyGender` emits. A wrongly denied relative reads as disconfirming, which is worse than silence. `absent` stays reachable: 20 of 384 surveyed real results name one parent and not the other.

**`relativeTermNote` is a note, not a weight.** Moving `matchScore` needs calibration first, would shift every ranking in the eval corpus, and would turn a specced review surface into a classifier.

## Verification

2097 tests pass, tsc clean, `search-records` unit suite 24/24.

Passing tests prove nothing by themselves, so I re-introduced each defect the plan review found and confirmed the suite catches it:

| Injected bug | Result |
|---|---|
| Collapse the sex rule to "not provably Male implies unknown" | tests 35, 36, 44 fail |
| Let the literal `"Unknown"` fall through as indeterminate | test 39 fails |

The 21 new tests needed a parameterised `householdEntry()` fixture first: every pre-existing fixture is a lone person with no relationships, so the suite would have asserted `unknown` throughout and caught neither bug.

## Also in this PR

Nothing typechecked `tests/` or `dev/`. That had a real cost here: widening `findRepresentedPerson`'s return type broke a test at runtime while `tsc` stayed green. Added `tsconfig.typecheck.json` (a separate `noEmit` project, since widening the build config would ship dev scripts into the `.mcpb`) and wired it into `pretest`.

It found 129 errors, all fixed, no exclusions. Three were dead code in `dev/`, including `try-login.ts` requiring a `<clientId>` it could not pass. Four were production type bugs only tests could see, notably `results-staging.ts` typing `response` closed while its spec says it "knows nothing about any caller's field names". The remaining 122 were test-side, 95 of them one shape: reading a single-op property off a `single | batch` union without narrowing, fixed with `tests/helpers/narrow.ts` rather than `as any`.

## One deletion a reviewer should look at

#1328 landed while this was open and added `tsconfig.dev.json` + `typecheck:dev`, covering `dev/**` only. This PR's `tsconfig.typecheck.json` covers `src` + `dev` + `tests`, a strict superset, so **I deleted theirs** rather than leave two overlapping typecheck projects with `pretest` running one of them — which is how someone gets a green check that never opened their files.

Their exclusion of `tests/` was argued, not accidental: vitest transpiles without typechecking, so including tests surfaced 126 pre-existing errors that would have blocked `npm test` repo-wide. Fixing all 126 here is what removed that precondition. Their rationale, including an incident this branch never hit (a broken qualifier-probe edit passing a check that never opened the file), is preserved in the surviving config's header rather than lost with the file.

Nothing in the repo referenced `tsconfig.dev.json` or `typecheck:dev` — checked across workflows, Makefile, docs and scripts. Engine CI runs `npm test` → `pretest` → `npm run typecheck`.

## Sequencing

Built on `main` @ `61eec452`. #1328 is still open and touches the same spec plus the `*Exact` descriptions, which this PR does not touch. If it lands first I will integrate by merge.

Closes #1324. The work from #1454, #1455 and #1456 is folded in here; those issues were already closed when they were folded, so this PR does not need to close them.


clack391 (approved):

promise-emmanuel (commented):
One blocker, then a small spec fix. Everything else in here I'd merge.

relativeTermNote survives a ranking that was thrown away. In rank-search-matches.ts the note is built from matches before the noSignal && subjectTooThin branch empties it, so when the ranking is withheld the response still carries a warning about the records it just suppressed:


matches:          []
returnedCount:    0
relativeTermNote: "Of the 2 matches returned, 2 name no father (`relativeTerms`)…"
Zero returned, and the same object says two. I hit this with the starved-subject setup plus two father: absent staged rows. It's reachable in the likeliest case there is — you anchor on a relative's name precisely when the subject is too thin to search on directly — and it reaches record_search too, which folds the ranking in whenever subjectId and projectPath are both set. Given what this PR is for, a response that claims matches it didn't return is the one shape I'd want it not to have.

Fix is to move the two lines below the branch and read out.matches:


You wrote:
  const relativeTermNote = buildRelativeTermNote(matches);
  if (relativeTermNote) out.relativeTermNote = relativeTermNote;

  if (noSignal && subjectTooThin) {

Change it to:
  if (noSignal && subjectTooThin) {
then immediately above the closing return out;:


You wrote:
  return out;

Change it to:
  // Built from `out.matches`, not `matches` — the withheld branch empties it,
  // and a note about records that were not returned is the false confirmation
  // this field exists to prevent.
  const relativeTermNote = buildRelativeTermNote(out.matches);
  if (relativeTermNote) out.relativeTermNote = relativeTermNote;

  return out;
That's the whole fix — out.matches is [] in the withheld branch so buildRelativeTermNote returns undefined, and everywhere else out.matches is still matches. Please add a test next to the three new ones: the existing "withholds the ranking when the subject has no dated or placed fact" setup, with relativeTerms: { father: { status: "absent" } } on the staged rows, asserting relativeTermNote is undefined. All four new tests stub 0.9, so nothing currently walks that path.

The spec's code map contradicts its own body. record-search-tool-spec-v2.md calls the second parameter a prefix list, but it carries the names, and other can't resolve without them — which the spec says itself three sections up.


You wrote:      `prefixes` is the relative-name prefix list from `suppliedKinTerms`
Change it to:   `terms` is the `KinTerm[]` from `suppliedKinTerms` — prefix plus the names the caller gave, which `other` resolves against
and mapEntry(entry, prefixes?) → mapEntry(entry, terms?) on the line above.

On the rest: I ran it against the merge with main rather than your head, since you're 62 behind. Merge is clean, 2176 tests pass, npm run typecheck is clean, harness is 1884 pass / 6 skip, and check_runlogs.py with BASE_SHA=origin/main is satisfied — main hasn't touched search-records/SKILL.md since your merge base, so your run log survives landing. I also broke the new typecheck to watch it fail: an injected type error in a test file leaves npm run build green and turns npm run typecheck red with TS2322, which is exactly the hole it's there to close. And collapsing the sex rule in resolveParentTerm fails tests 35 and 44, so the unknown-before-absent ordering is actually guarded.

johnmarkpeterbrown (changes requested):
Kindly merge with main and notify me for final review and merging. Notify me in Slack for timely responses.

johnmarkpeterbrown (approved):
Looks good!

johnmarkpeterbrown (approved):
Looks good!

johnmarkpeterbrown (commented):
**To @taiwo-stack — one thing before merge**

The code is solid. The tool computes `present`/`absent`/`unknown` correctly (I confirmed the unit tests genuinely guard it), the schema and consumers are wired, and promise-emmanuel's `relativeTermNote` blocker is fixed with the test that was asked for. johnmarkpeterbrown's approval covers the developer side.

**The gap:** nothing proves the *agent* obeys the new rule. Every fixture with `relativeTerms` is a `present` case. There is no test where the agent gets an `absent` relative and has to write it up. That's the one failure this PR exists to prevent — "confirming her father William" about a record that never names him — and it's the one case no test exercises. CI can be fully green while the agent still writes "confirming" on an absent record.

**What to do:** add one absent-case behavioural test. I've drafted it — three files, drop-in ready:

1. `record-search-ackerman-1860-spouse-absent.json` → `eval/fixtures/mcp/`
   Spouse-anchored census hit that names the household but no wife → `spouse: absent`.
2. `rank-search-matches-ackerman-1860-spouse-absent.json` → `eval/fixtures/mcp/`
   Carries `relativeTerms` onto the stub and emits `relativeTermNote` — covers that consumer too.
3. `spouse-absent-on-census-hit.json` → `eval/tests/unit/search-records/`
   The test. Its `judge_context` makes "confirms the spouse/marriage" a hard fail and protects the correct "consistent with" hedge from being marked down.

It reuses the existing `census-household-absent-spouse` scenario, so no new scenario is needed. The search fixture is self-consistent — its own gedcomx has no couple row, so `spouse: absent` is what the resolver would actually derive, not something hand-asserted.

**How to run and verify:**

```bash
make eval-skill SKILL=search-records
```

Then:
- If the run aborts because the skill called `record_read`, add a `record-read-ackerman-1860` fixture (predicate `recordId: "~M7QZ-8KD"`) and re-run. Only if it aborts.
- Read the agent's write-up in the new test's turn. It should say the census doesn't name the spouse / is consistent with the marriage — not that it confirms Catherine.
- Annotate the regenerated run log in the CRUD UI and commit it (the `runlogs` check requires the annotation).

**Lane:** the `judge_context` wording and the scenario premise are genealogist calls — loop in the senior-genealogist to confirm those before the run. This is also what the open `ready-for-senior-genealogist` label should be covering.

Merge once that test passes. If it's too much to fold in, say so and I'll file it as a `genealogist` + `nothing-checks` issue against this PR — but folding it in is cheaper than a second review cycle.

johnmarkpeterbrown (changes requested):
**To @taiwo-stack — one thing before merge**

The code is solid. The tool computes `present`/`absent`/`unknown` correctly (I confirmed the unit tests genuinely guard it), the schema and consumers are wired, and promise-emmanuel's `relativeTermNote` blocker is fixed with the test that was asked for. johnmarkpeterbrown's approval covers the developer side.

**The gap:** nothing proves the *agent* obeys the new rule. Every fixture with `relativeTerms` is a `present` case. There is no test where the agent gets an `absent` relative and has to write it up. That's the one failure this PR exists to prevent — "confirming her father William" about a record that never names him — and it's the one case no test exercises. CI can be fully green while the agent still writes "confirming" on an absent record.

**What to do:** add one absent-case behavioural test. I've drafted it — three files, drop-in ready:

1. `record-search-ackerman-1860-spouse-absent.json` → `eval/fixtures/mcp/`
   Spouse-anchored census hit that names the household but no wife → `spouse: absent`.
2. `rank-search-matches-ackerman-1860-spouse-absent.json` → `eval/fixtures/mcp/`
   Carries `relativeTerms` onto the stub and emits `relativeTermNote` — covers that consumer too.
3. `spouse-absent-on-census-hit.json` → `eval/tests/unit/search-records/`
   The test. Its `judge_context` makes "confirms the spouse/marriage" a hard fail and protects the correct "consistent with" hedge from being marked down.

It reuses the existing `census-household-absent-spouse` scenario, so no new scenario is needed. The search fixture is self-consistent — its own gedcomx has no couple row, so `spouse: absent` is what the resolver would actually derive, not something hand-asserted.

**How to run and verify:**

```bash
make eval-skill SKILL=search-records
```

Then:
- If the run aborts because the skill called `record_read`, add a `record-read-ackerman-1860` fixture (predicate `recordId: "~M7QZ-8KD"`) and re-run. Only if it aborts.
- Read the agent's write-up in the new test's turn. It should say the census doesn't name the spouse / is consistent with the marriage — not that it confirms Catherine.
- Annotate the regenerated run log in the CRUD UI and commit it (the `runlogs` check requires the annotation).

**Lane:** the `judge_context` wording and the scenario premise are genealogist calls — loop in the senior-genealogist to confirm those before the run. This is also what the open `ready-for-senior-genealogist` label should be covering.

Merge once that test passes. If it's too much to fold in, say so and I'll file it as a `genealogist` + `nothing-checks` issue against this PR — but folding it in is cheaper than a second review cycle.

johnmarkpeterbrown (approved):
Looks good!

PR #1466, opened 2026-08-09
e2e: resolve anna-findejsova-daughter — TRUE MATCH (Probable), scored + graded

kofiatinka12:
Closes #854.

Resolves the `anna-findejsova-daughter` record-hint fixture, with a scored live run and its blind grade committed together.

## Resolution — TRUE MATCH
Genealogist adjudication of FamilySearch's hint (`ark:/61903/1:1:6PJZ-ZBSM`). The **1 March 1818 baptism of Agnes** to Maxmilian Michal + Anna is the tree couple — **Anna Marie Findejsová (P915-7QP)** + **Maxmilián Michl** — confirmed against the **register image**: image group 005652359, image 23 of 321, Pag. 20, Entry 2, **house no. 29** (`ark:/61903/3:1:3QS7-89S1-K869`), in the **Svitavy / Mährisch Trübau register (Moravia)**.

The original **Bystré/Bistrau (Bohemia) hypothesis was wrong** — that register is not digitized on FamilySearch (SOA Zámrsk, catalog item 2425217), which is why every Bystré-focused search came back empty. `DRAFT PENDING ADJUDICATION` marker removed.

**Proof standing: Probable, not Proved** — the child's name and father's surname were not read directly from the original image (OCR misread them "Clymus"/"Winfal"; the 2.3 MB scan couldn't be processed by the higher-accuracy reader), so Agnes was confirmed via corroborating tree data. A clean direct image read is the one step to Proved.

## Scored run + grade (committed together)
`run-2026-08-09_15-22-27` (.18, 54.5 min):
- **Recall: PASS** — the agent recovered Agnes as a new person (I1), baptized 1 Mar 1818 Svitavy, linked as child of **both** parents alongside the existing Anna (b. 1823). Blind grade `run-2026-08-09_15-22-27.ann.json`: **f1 = true**.
- **Compliance: FAIL — a *separate* process issue, not a recall problem.** The agent minted the new person (I1) and wrote its `person_evidence` link **without ever calling `same_person`** — identity asserted without identity scoring (the same `same_person` provenance gap tracked by #963/#1231). The harness records this as a compliance-axis failure, orthogonal to the finding grade; it does **not** change the recall label.

The `run-<ts>.json` and its `.ann.json` are committed in the same PR, satisfying the e2e grading-gate check. `proof_quality_score` intentionally omitted (a complete grade without it).

## Files
- `eval/tests/e2e/anna-findejsova-daughter/{README.md, expected-findings.json, fixture.json}` — resolved fixture.
- `eval/runlogs/e2e/anna-findejsova-daughter/run-2026-08-09_15-22-27.{json, final-tree.gedcomx.json, final-research.json, ann.json}` — scored run + blind grade.


Paaboat (approved):

EdmondOware (commented):
Three independently-checkable points converging with zero contradiction — the spouse name variant (Maxmilián/Maxmilian Michl/Michal), the age plausibility check, and especially the locality correction: catching that the original Bystré hypothesis was wrong because that register was never digitized, then tracking the actual match down to the Svitavy register at image-group/page/entry/house-number level. That's real provenance, not a vague place-match. And the tier call is exactly right. Probable, not Proved, because the OCR garbled the child's and father's names on the primary image and the higher-accuracy reader couldn't process the scan. Capping it there instead of stretching to Proved on corroborating data alone is the correct call.

One question, not a blocker: the record gives the mother's name as "Anna Kristina" against "Anna Marie" in the tree. You call this a routine Czech parish-book variation, which is plausible  is that from a precedent (her appearing under a different middle name elsewhere in her own sources), or general domain knowledge about the period? Curious either way, just want to understand the basis.

On the compliance flag: appreciate you tying it to #963/#1231 rather than treating it as new since this hint is a real record persona (not an FTS/image-only assertion with no record_persona_id), skipping same_person here does look like a genuine instance of that same tracked gap. No new issue needed; this is just another data point for the existing one.

EdmondOware (approved):
see comment above for the full review.

PR #1483, opened 2026-08-09
docs: unresolved threads now block merge

DallanQ:
## Summary

- `required_review_thread_resolution` was flipped to `true` on the `protect-main` ruleset. Three places still said it was `false`.
- Text only. The gate in `/review` §1 and the check in `senior-queue` are unchanged, and both are still needed: blocking *merge* does not stop a PR being *surfaced to a senior as ready*, which is the job those two checks do.

Tier: Trivial.

## Review guidance

**Start here:** `.claude/skills/review/SKILL.md` §1. The old sentence said stopping was necessary *because* nothing blocked merge; that reason is now wrong, but the instruction is not. The rewrite gives the real reason — not spending the scarcer reviewer on findings nobody has answered — and keeps the contrast that matters: an unresolved thread blocks merge, an outstanding changes-requested does not, so a reviewer checks both.

**Unsure about:** nothing.

## Plan

**Didn't change:** the §1 gate, the queue's readiness rules, and the ruleset.

**Acceptance check:** `grep -n "is false in the ruleset\|do not block merge\|neither blocks merge"` over both files returns nothing; the workflow still parses and `SELF` still equals the job name.

## Test plan

- [x] YAML parses; embedded script parses; `SELF` matches the job name.
- [x] No stale claim left in either file.
- No suite touches these lines.

## Follow-on issues

None.


DallanQ (commented):
Firing a pull_request_review event to exercise senior-queue's self-filter: every other check on this head has completed, so the only in-flight check run is the workflow's own. If the filter works this run reports only 'no peer approval'; if it is broken it also reports 'CI still running (1)'.

Issue #1420, opened 2026-08-06
Windows portability: forbid syncing folders (OneDrive) in the docs, fix only the real Windows bugs


Emruthwill:
> **Reviewed 2026-08-18 before junior handoff.** Decision settled — a Windows
> collaborator runs `scripts\windows\test-all.bat` on the branch; two of the six
> items have already shipped. Read the corrections first; the original body follows
> under "Original issue".

## Corrections, 2026-08-18 — read before starting

**#1349 (uv `link-mode = "copy"`) is DONE. Do not redo it.** Landed on branch
`fix-uv-link-mode-onedrive`; re-verified on `main` 2026-08-18 with
`grep -rn "link-mode" --include=pyproject.toml --include=uv.toml .`, which returns
`eval/harness/uv.toml:28`, `apps/server/pyproject.toml:47` and
`eval/triggering/pyproject.toml:18`. Note the harness copy is in a **`uv.toml`**, not
its `pyproject.toml` — a grep of `pyproject.toml` alone reads as "still missing" and
invites a fourth re-diagnosis. **The 2026-08-09 comment asking for that branch to be
pushed is stale — do not chase it.**

**#1185 is closed, not open.** `scripts\windows\test-all.bat` shipped in PR #1614
(2026-08-14) and is exactly the Windows equivalent of `make test-all` that #1185 asked
for, with the tool preflight and the empty-`node_modules` check it named. The "stays
open and untouched" line below is false.

**#1348's justification no longer holds.** PR #1614 merged 2026-08-14 — five days
after this issue was merged. `DEVELOPMENT.md` § "On Windows" now states there is no
`make` on Windows without Git Bash or WSL, developer targets have
`scripts\windows\*.bat` wrappers, and `windows-wrappers.test.ts`'s `EVAL_OWNED` map
names **`eval\ValidateFixture.bat` as `e2e-validate`'s sanctioned Windows entry
point** — a script that predates this issue, already implements the blank-slug
`--all` default in `cmd`, and is documented in `eval/README.md`. So no Windows user on
a supported path is blocked from validating a fixture. What is left is a portability
tidy worth one line: replace the shell expansion with make's own conditional,
`$(if $(TEST),$(TEST),--all)`.

**Which other recipes: settled, do not re-survey.** Four `:-` expansions exist
(`grep -n ':-' Makefile`). Two are pure default-value expansions and take the same
one-line fix: the `e2e-validate` recipe and `optimize-skill`'s
[...]

PR #1502, opened 2026-08-09
Fix uv link mode onedrive

Emruthwill:
## Summary

**Tier: Trivial.** Pins `link-mode = "copy"` for every uv-managed project in the
repo, so `uv` stops hardlinking out of its cache into `.venv` — the thing OneDrive
rejects (`os error 396`, and the intermittent `Access is denied` / `os error 5` on the
next `uv run` once a partially-synced `dist-info` is left behind, which reads like a
stray file lock and gets retried away).

There are **three** uv projects, and all three now carry the setting:

| Project | Where | Installs packages? |
|---|---|---|
| `eval/harness` | new standalone `uv.toml` (holds the full write-up) | yes — `uv.lock`, ~7 direct deps |
| `apps/server` | `[tool.uv]` in `pyproject.toml` | yes — `uv.lock` |
| `eval/triggering` | `[tool.uv]` in `pyproject.toml` | **no** — see below |

`eval/triggering` was left out of the first two commits and is added here after review.
It is reachable from its own Windows launcher (`eval/OptimizeSkill.bat` does
`cd triggering && uv run`), so it is a path a genealogist hits directly. Measured on a
OneDrive clone before adding it: `uv sync` / `uv run` there install **nothing** — the
project declares no dependencies and sets `package = false`, so `site-packages` ends up
holding only virtualenv's own `_virtualenv.pth` / `_virtualenv.py`. There is nothing to
hardlink and the failure cannot occur today. It is set anyway because "has no
dependencies" is the only thing standing between that project and the same failure, and
adding the first dependency would re-arm it silently. The inline comment says it is a
no-op today so nobody reads the line as load-bearing.

Refs #1349 (closed — absorbed into #1420) and #1420. **Non-closing**: #1420 keeps its
documentation half (the "don't clone under a syncing folder" rule in `DEVELOPMENT.md` /
`eval/README.md`, the `%OneDrive%` guard in `Setup.bat` / `Start.bat`) and both genuine
Windows bugs in its bucket B.

This is insurance, not OneDrive-compatibility work. The documented constraint is
unenforced by construction; this is the one part of the cluster that degrades gracefully
when the doc is not read.

## Review guidance

**Start here:** `eval/harness/uv.toml` — it is the rationale of record and the other two
sites point at it rather than repeating it. Worth checking that a standalone `uv.toml`
is the right home for the harness: `uv.toml` and `[tool.uv]` in `pyproject.toml` are
mutually exclusive, and where both exist the `uv.toml` wins and the `pyproject` block is
ignored *in full*. `eval/harness/pyproject.toml` has no `[tool.uv]` section today, so
nothing is being shadowed — but anyone who later adds one there will find it silently
dead, and this is the moment to decide whether that trap is worth the standalone file.

**Unsure about:** the failure's current severity. Re-measured on the OneDrive clone with
**uv 0.11.32**, the default link mode no longer hard-errors — it prints
`warning: Failed to hardlink files; falling back to full copy`, recommends
`link-mode=copy` by name, and then installs successfully. So on today's uv this change
suppresses a warning and avoids the partial-hardlink path, rather than unblocking a hard
stop. The hard `os error 396` / `os error 5` in the commit messages was observed
2026-07-30 and again 2026-08-03, on an older uv; I could not reproduce a hard failure
now. The change is still what uv's own warning prescribes and still removes the partial
state the `os error 5` symptom comes from, but if the framing should be "quiets a
warning" rather than "unblocks the harness", say so and I will reword the commits.

## Plan

**Didn't change:**
- The documentation half of #1420 — no OneDrive warning in `DEVELOPMENT.md` or
  `eval/README.md`, no `%OneDrive%` guard in `Setup.bat` / `Start.bat`. That is the lead's
  stated direction for #1420 and wants its own PR.
- `eval/app/next.config.mjs` (`distDir`) — #1420 establishes it cannot move outside the
  project without breaking module resolution.
- Bucket B of #1420: `#1348` (POSIX `$${TEST:---all}` in the Makefile) and `#1370`
  (shebang in the imported `gen-enums.mjs`). Unrelated to uv.
- No `UV_LINK_MODE` env var, no `.bat` change, no per-developer setup step — #1349's
  acceptance requires the fix to be committed config, not a local step.

**Acceptance check:** OneDrive-only, so **CI cannot prove it** — the runners' checkouts
are not under a syncing filesystem, and there is no failing-then-passing test to name.
What was run instead, on a OneDrive-synced clone (`uv 0.11.32`, Windows 11):

1. Two throwaway uv projects created side by side *inside* the OneDrive tree with the
   same two dependencies — one with the default link mode, one with `link-mode = "copy"`.
   Default: `warning: Failed to hardlink files; falling back to full copy`. Copy: clean,
   no warning, exit 0.
2. `uv sync` and `uv run` in `eval/triggering` with the new setting: exit 0, no warning.
3. The key is proven *read* rather than silently ignored — setting `link-mode = "bogus"`
   makes uv reject it with ``unknown variant `bogus`, expected one of `clone`, `reflink`,
   `copy`, `hardlink`, `symlink` `` pointing at that line of
   `eval/triggering/pyproject.toml`.

Off OneDrive the setting costs a copy instead of a hardlink on install and nothing else.

## Test plan

- [ ] For every skill whose **run-log snapshot** I changed … `make eval-skill SKILL=<name>`
      

- [ ] If I changed a skill's `description` frontmatter or its DO NOT clauses …
      

## Follow-on issues

None filed.

**Folded in / filed instead:** folded in the third uv project (`eval/triggering`) raised
in review, rather than filing it — one line in a file this PR already owns, with the
context loaded. Nothing deferred.


T-FEH (approved):
Approving. Config-only, correct, and complete. @clack391's three points are all taken — `eval/triggering` is in with an explicit note that it's a no-op today, and the body carries the non-closing #1349/#1420 refs.

I pushed one comment-only commit (`da12c32a`) rather than send it back for a round; `link-mode = "copy"` is untouched.

Verified by running, not reading:

- **The setting is live, not decorative.** Planting `link-mode = "bogus-value"` in a copy of `eval/harness/` makes uv fail with `unknown variant 'bogus-value', expected one of 'clone', 'copy', 'hardlink', 'symlink'` — so the standalone `uv.toml` really is read at that path, and `"copy"` is a valid variant. Same in the two `[tool.uv]` placements.
- **Coverage is exactly right.** Three uv projects in the repo — `apps/server`, `eval/harness`, `eval/triggering` — and all three now carry it. (`e2b.Dockerfile` uses pip; the fourth set under `.claude/worktrees/` is a stale worktree.)
- Nothing depends on hardlinks: `deploy/Dockerfile` stage 2 `uv sync --frozen`s and the runtime image `COPY --from`s the venv, which flattens hardlinks regardless. No lockfile or resolution change.
- Suites: harness pytest **1958 passed**, `apps/server` **157 passed, 1 skipped**.

**One correction, and it bears on the question you asked reviewers to decide.** The Review guidance says a `uv.toml` makes the `pyproject` block "ignored *in full*". It doesn't — I tested it with both files present:

```
warning: Found both a `uv.toml` file and a `[tool.uv]` section in an adjacent
`pyproject.toml`. The following fields from `[tool.uv]` will be ignored in
favor of the `uv.toml` file:
- link-mode
```

Only the conflicting *configuration* key is dropped, it is named in a warning, and project keys like `package = false` still apply. So the trap you flagged is loud and narrow, and the standalone file is fine on those grounds. Both #1349 and #1420 name `eval/harness/pyproject.toml`, so folding this into a `[tool.uv]` block there would delete a file and match the other two sites — worth a thought, not worth a round.

What I changed in `da12c32a`, both comment-only:

- "This repo's working copy lives in a OneDrive-synced folder" stated one setup as a repo-wide fact. #1420's still-open documentation half is a *"don't clone under a syncing folder"* rule, which that sentence would contradict the moment it lands. Now scoped to the Windows genealogist team's clones.
- The `os error 396` / `os error 5` write-up read present-tense, but you'd already found in the body that uv ≥ 0.11 warns and falls back instead of hard-erroring. A reader who hits only the warning would think the file described a different problem, so the version-dependence is now in the file rather than only in the PR body.

Since I made the last push, the second approval has to come from someone else — and this had none to begin with, so it needs two.

jud-sdev (approved):

Issue #1519, opened 2026-08-10
eval: flaky negative-routing test ut_search_external_sites_012 (record-in-hand) — convert to grade_on_invariant


clack391:
> **IMPORTANT — ruled 2026-08-25: do this inside issue #1648's PR.** #1648 is the
> `search-external-sites` deep dive, now in Ready, and it is already paying this skill's
> `make eval-skill SKILL=search-external-sites` run (measured $3.26 + ~22 min + one
> annotation pass over 5 sampled tests). This card's fix edits
> `eval/tests/unit/search-external-sites/negative-record-in-hand.json`, inside the same
> run-log snapshot, and — as the body below says — only the **full** run can verify a
> `grade_on_invariant` negative. Running it separately buys a second identical run for
> nothing.
>
> The two-step fix below is ready-made; hand it to whoever takes #1648 verbatim. Note for
> them: `ut_search_external_sites_012`'s current failure is **this flake, not a dive
> finding**.

**Touches:** eval/tests/unit/search-external-sites/negative-record-in-hand.json, eval/harness/validators/test_search_external_sites.py, eval/harness/validators/test_research_plan.py

`ut_search_external_sites_012` (`negative-record-in-hand`) is a flaky negative-routing test. It passed on 2026-07-28 but fails intermittently now with no change to the skill, the judge prompt, or the test itself. Every deterministic input (model `claude-sonnet-4-6`, judge prompt hash, skill body, test) is identical across the passing and failing runs, so the flip is non-determinism, not a regression.

## Root cause

Negative routing tests with a non-empty `correct_skill` pass only when the skill actually invokes that skill (it must land in `skills_invoked`; see `orchestrator.py` negative-outcome logic). Two observed runs:

- 2026-07-28: `skills_invoked=['record-extraction']`, `activated=False` -> pass
- 2026-08-10: the skill declined but engaged ("upload the PDF and I'll hand it to record-extraction"), which the activation heuristic reads as `activated=True` -> fail

So pass/fail hinges on whether the model invokes record-extraction vs. declines in prose, which varies run to run. The judge scores the decline a correct 3/3 either way; only the routing verdict flips. This is the exact flakiness the sibling `ut_search_external_sites_011` (planning -> research-plan) already fixed with `grade_on_invariant`.

## Ready-made fix (mirrors 011)

[...]

PR #1521, opened 2026-08-10
search-external-sites: note FamilySearch's own holdings before recommending a competitor (#1313)

clack391:
## Summary

Tier: Normal (skill-lifecycle loop).

- A FamilySearch reviewer saw the agent surface an 1855 Massachusetts State Census on MyHeritage without noting that FamilySearch holds the same census, and separately invent a nonexistent 1845 schedule (#1313). The tool was not at fault: `collections_search` already returns the FS 1855 (`1459985`) and 1865 (`1410399`) state censuses (verified live). The gap was skill doctrine.
- `search-external-sites` had no `collections_search` grant, and its only "prefer FamilySearch" instruction lived in the autonomous-only branch, so the interactive path never checked FS's own holdings. This adds the grant plus a step `2b` that checks FS holdings and, when FamilySearch holds a matching collection, still builds the chosen external URL but notes the FS holding and offers the FS copy via `search-records`. It also grounds availability claims in the tool result rather than memory.
- New regression test `ut_search_external_sites_fhk` captures the general rule, with two fixtures from live tool responses and a synthetic Massachusetts scenario.

## Review guidance

**Start here:** the `### 2b` block in `search-external-sites/SKILL.md` and the new test's `judge_context`. Confirm the doctrine informs rather than overrides ("still build the URL the user chose... note FamilySearch holds the same collection"), and that `judge_context` grades the general rule (any external result duplicating an FS-held collection), not the Massachusetts case.

**Unsure about:** Nothing blocking. One low-severity code-review finding is deferred to #1520 (below).

## Plan

**Didn't change:** No tool code. `collections_search` already returns the FS 1855/1865 collections (acceptance clause 1 was already met, verified live), so this is a skill-doctrine fix, not a tool fix, despite the issue's `Touches:` naming three tool files. The autonomous-only "prefer FamilySearch" branch is left alone. No rubric edit: adding the cross-check to the "Tool selection" dimension would re-baseline all 13 tests, so the new behavior is graded through base Correctness/Completeness plus `judge_context`.

**Acceptance check:** `ut_search_external_sites_fhk` fails on the pre-edit skill (Correctness and Completeness 2 — "failed to surface that FamilySearch also holds the 1855 census") and passes on this branch (notes FS holds `1459985`, offers `search-records`). Committed graded run: `eval/runlogs/unit/search-external-sites/v1_2026-08-10_07-55-38.json`.

**Deviated from the plan:** (1) The doctrine landed as a standalone in-flow step `### 2b` rather than folded into step 2, and SKILL.md grew +18 lines rather than staying flat — chosen because it is a genuinely new capability and a numbered step is executed more reliably than a paragraph buried in step 2; `fhk`'s pass across two gates and the full run validates it. (2) A fix for the flaky negative test `ut_search_external_sites_012` was briefly folded in, then split out to #1519 to keep this PR to one problem.

## Test plan

- [ ] `make test-all` — **N/A.** This PR touches only the `search-external-sites` skill dir and `eval/` data (tests, fixtures, scenario, run log), all exempt from the `make test-all` gate per this template. Verified via the eval-skill run instead (next box).
- [x] Ran `make eval-skill SKILL=search-external-sites` and committed the run log **and its `.ann.json`** (`eval/runlogs/unit/search-external-sites/v1_2026-08-10_07-55-38.json` + `.ann.json`). 13 pass; the 1 fail is `012`, a pre-existing flake tracked in #1519 (its judge dims are a correct 3/3 decline; the fail is the routing verdict).
- [ ] Skill `description` / DO-NOT clauses — **N/A.** Only the body and `allowed-tools` changed; the `description` and routing clauses are untouched, so no negative-test refresh is needed.

## Follow-on issues

- #1520 — step `2b` is exercised by only the `fhk` test, and its unconditional `collections_search` call carries a low-probability run-abort risk in fixtureless scenarios. Fix: scope `2b` to Case A or backfill fixtures, plus a second coverage test.
- #1519 — the pre-existing flaky negative-routing test `ut_search_external_sites_012` (convert to `grade_on_invariant`). Referenced because it shows as the one fail in this PR's committed run.

**Folded in / filed instead:** Both follow-ups were **filed** rather than folded — #1519 because bundling a second, unrelated problem violates one-problem-per-PR (skill-lifecycle §8), and #1520 because scoping `2b` needs another paid eval run to verify and the finding is low-severity with no abort observed.

Closes #1313.


Ikennaya1 (approved):
Approving, good work!

The genealogy checks out. Massachusetts only conducted state censuses in 1855 and 1865, so the invented 1845 schedule was a genuine fabrication, and grounding availability claims in tool results rather than memory is the correct fix.

Step 2b also strikes the right balance: it still generates the URL the user requested while presenting the FamilySearch holding as additional information rather than a redirect.

florencemashipei (approved):

Issue #1522, opened 2026-08-10
check_rubric_tool_drift: add a suppression mechanism for its false positives, and widen the vocabulary to catch a required-but-unavailable capability


florencemashipei:
> **Reviewed 2026-09-18 before junior handoff.** Decision recorded 2026-09-18; see the Decided block.
> original body follows under "Original issue".

## Decided 2026-09-18 — centralized keyed allow-list

The suppression marker is a `SUPPRESSIONS` list inside
`eval/harness/scripts/check_rubric_tool_drift.py`, keyed
`(relative_path, tool, reason)`, matching the repo's two existing
lint-suppression mechanisms (`gps-terminology.test.ts:101`,
`skill-reference-reachability.test.ts:87`). It beat inline per-line comments,
which were the filed recommendation: an inline marker lands inside
`judge_context` strings that go verbatim into the judge prompt
(`eval/harness/judge/prompt.md:372`), is forbidden in an agent body
(`CLAUDE.md` § "Cowork plugin agents": no comments), and touches 13 skill
directories plus six agents — ~15 blocking `make eval-skill` re-runs at
$8–12 each — where the allow-list touches one file and buys none.

The list is **shrink-only** and an entry that has stopped matching must fail,
not sit there: `eval/harness/tests/unit/test_check_rubric_tool_drift.py`
asserts both that direction and that a suppressed entry does not mask an
unrelated genuine hit.

**Record the reasoning — this write is part of the task:** the trade-off and
the rejected inline form go in the module docstring of
`check_rubric_tool_drift.py` beside the `SUPPRESSIONS` list, and the
conclusion (one clause, pointing at the script) replaces the stale
"allow-comment convention … issue #1003" sentence at `eval/CLAUDE.md:286`.

**This is also the repo's suppression convention for the `encoding="utf-8"`
lint.** Issue #1355 closed COMPLETED 2026-09-03 with its gap 2 unbuilt and
still deferred here, so nothing will apply it there —
`eval/harness/tests/unit/test_encoding_lint.py` has no opt-out today. Either
add the same `(relative_path, reason)` list to it in a follow-on PR or file
that card; do not leave it orphaned.

*Ruling recorded 2026-09-18 (`/make-decisions`); option A of the review-ready block.*

### State, re-measured

`cd eval/harness && uv run python scripts/check_rubric_tool_drift.py` → **89 hits**:
[...]

Issue #1320, opened 2026-08-05
Agent assumes the objective before the user has stated it


DallanQ:
**Touches:** packages/engine/plugin/skills/init-project/SKILL.md

**IMPORTANT**: If you do this, do #1510 at the same time — both edit `packages/engine/plugin/skills/init-project/SKILL.md` and both flip its eval run log inactive. #1510 makes `init-project` ask for experience level, subscriptions **and the objective** in its opening turn, and corrects four documents that already promise an interview the skill is forbidden to run; this issue is the objective half reported from a user's side. Land them in one PR so `init-project` pays one re-run + annotation, not two.

*(Pairing updated 2026-08-17. The earlier note here pointed at issue #1003, which closed completed on 2026-08-13 — that pairing is discharged. Also note `project_create` landed in PR #1690 the same week and changed how this skill creates a project; read `init-project/SKILL.md` on `main` rather than from this body.)*

Three reports from cmadsen1@gmail.com.

- **2026-08-04 06:39**: *"As soon as I started, it made assumptions based on the folder name I had chosen. When I gave it the PID, it kept making assumptions about what I was wanting it to do."* Should have *"Waited for me to define the objective, first and foremost."*
- **2026-08-04 20:56**: *"It immediately made the plan to fetch person, build tree, create research.json, pedigree analysis and project summary. I asked it why it was including the analysis and summary when I still hadn't defined an objective. It stopped then and asked what I wanted to do"* — and one of the projects it proposed was a hallucination: *"tracing migration from Upper Canada - belongs to his father not him."*
- **2026-07-31 18:09**: *"The Project Overview is overly broad for the question asked... it kept wanting to restore the relationships and records for wife and child, despite that being outside the parameters."*

Same tester also hit the inverse in **2026-07-31 19:31**: a plain-prose prompt produced only an MD file with no `research.json`, so the viewer showed nothing — *"I had to redo the prompt using /init-project and follow the questionnaire prompts before it did the research.json file."* Worth handling in the same pass: the skill either owns project setup or it doesn't.

**Acceptance:** init-project does not propose research direction, build a tree, or write analysis before the objective is captured.

florencemashipei:
[...]

PR #1523, opened 2026-08-10
Triage check_rubric_tool_drift warnings + fix init-project premature-action bug (#1003, #1320)

florencemashipei:
## Summary

Closes #1003 and #1320, bundled together because both touch `packages/engine/plugin/skills/init-project/SKILL.md` and would otherwise need two separate eval re-runs of that skill.

**#1003 — triage all 68 `check_rubric_tool_drift` warnings:**
- Classified all 68 hits: 3 genuine drift, 11 agent-frontmatter `disallowedTools:` gaps, 54 false positives (legitimate cross-skill mentions / deliberate documentation).
- Fixed the 11 agent-frontmatter gaps: added missing `disallowedTools:` entries (all 3 tool-name spellings) to `gps-mentor.md`, `image-reader.md`, `image-reader-opus.md`, and `record-extractor.md` (which now carries two: `tree_edit` and, per review, `materialize_facts`).
- Fixed the 3 genuine drift hits: reworded `research-exhaustiveness`'s two test files and `init-project/rubric.md` from mandatory ("must call X") to permissive/accurate phrasing.
- Recommendation on the suppression-mechanism question: filed as #1522 (different skill's eval slot — a second paid run outside this PR's scope, per the #1449 fold-it-in gate) — the false positives are structural (skills have no `disallowedTools:`-equivalent escape valve), confirmed by wording fixes not changing the raw hit count at all.

**#1320 — init-project proposes work before the objective is captured:**
- Rewrote `init-project/SKILL.md` Step 1 so the research objective is the skill's one explicitly blocking question — removed the "formulate a default objective from what's missing" fallback that let the skill invent a direction before the user stated one.
- The same fallback also shipped verbatim in `init-project/references/research-process-init.md`'s decision table (a doctrine file the skill reads) — caught in review; fixed there too so #1320 doesn't half-ship.
- Added regression test `ut_init_project_010` (`blocks-on-bare-pid-no-objective.json`) covering a bare FamilySearch PID with no stated objective. It now declares the `person-read-flynn` fixture (added in review) so declining to call `person_read` is a real, gradeable choice rather than a call the mock can't register in the first place.
- Fixed `test_both_project_files_created` in the harness validator to skip its file-existence assertion for tests tagged `no-premature-write` — and added the paired `test_no_premature_write` positive assertion (added in review, via mutation testing that showed the skip-only version couldn't fail on a mutant that wrote both files anyway).

**Also:** stripped a dead `"validate-research-schema"` `mcp_fixtures` entry from 7 init-project test files (found while working the schema reference; the entry is dead only for `init-project` — 20+ tests in other skills still use it). Separately, `research-exhaustiveness`'s `declare-exhaustive-complete.json` and `honest-early-termination.json` had their `mcp_fixtures` emptied and `judge_context` rewritten to assert `research_append` is the complete persistence step (not just reworded descriptions) — a grading-contract change, defensible since that skill's `allowed-tools` is `research_append` only, called out explicitly here per review.

Two complaints raised alongside #1320 were investigated but **not** fixed here — filed as follow-ups (needs a different reviewer to reproduce with the original failing prompt before a fix can be scoped, per the #1449 fold-it-in gate): #1447 (overly-broad research scope, **closed as not-planned 2026-08-09**) and #1448 (plain-prose init failing to write `research.json`, still open).

**Addressed post-review (see review thread):** the `disallowedTools` denies added to `record-extractor.md`/`image-reader.md`/`image-reader-opus.md` are a real capability change, not cosmetic — `record-extraction` and `search-images` (both reference these agents) were fully re-run and re-annotated rather than cosmetic-skipped. Also fixed a stale `AGENT_PERMISSIONS` snapshot in `tests/packaging/agent-tool-names.test.ts`, repaired merge-conflict-marker corruption in two `.ann.json` files, and the `init-project`/`materialize_facts`/fixture fixes described above. Because those last two changes touch real doctrine files (not just the dead `model:` pin), `init-project` and `record-extraction` needed one more genuine re-run each — no cosmetic-skip for either.

**The re-run caught a real, previously-latent bug in the #1320 fix.** With the new `person-read-flynn` fixture in place, `ut_init_project_010` actually failed: the skill called `person_read` before asking for the objective, explicitly reasoning "Fetching person data to learn who LZNY-BRF is before asking for your objective." The Step 1 example question — "What would you like to research about [name, once known]?" — implied the skill needed the name first, undermining the "STOP" instruction one sentence earlier. Reworded to "this person" (no lookup needed) plus an explicit "do not call `person_read` first" line. Re-verified: `ut_init_project_010` now passes for the right reason, with a real fixture in play rather than a tool that couldn't be called.

Also filed #1534 (developer) for an unrelated harness crash hit while re-running: `judge_cost_usd` computed a tiny negative float (`-0.0001`) that failed the schema's `minimum: 0`, killing the entire run rather than just that one test's cost field.

## Test plan

- [x] `research-exhaustiveness` unit suite re-run: `ut_research_exhaustiveness_017` failed; a prior committed run (`v1_2026-07-23_17-04-02`) passed the same test 3/3/3/3/3/3 on byte-identical inputs. One pass and one fail on identical inputs — an **unconfirmed flake at n=2**, not established pre-existing variance. Not re-run further since it's unrelated to this diff; worth a follow-up if it recurs.
- [x] `search-images` unit suite re-run and re-annotated after the `image-reader.md`/`image-reader-opus.md` `disallowedTools` change.
- [x] `research-exhaustiveness` run log annotated (`.ann.json` committed) — human grades agree with the judge on the one genuine failure.
- [x] `init-project` unit suite re-run twice (once exposing the premature-`person_read` bug above, once verifying the fix) and re-annotated on the final, all-passing run.
- [x] `record-extraction` unit suite re-run and re-annotated a second time after the `materialize_facts` deny — zero calls to it anywhere in the run, confirmed a true no-op; pass/partial/fail churn versus the prior run is this skill's known non-deterministic variance, not a regression.
- [x] Verified `record-extractor.md`'s edits don't conflict with the unrelated "Worked example" additions already on `main`.
- [ ] `eval-cosmetic-skip` requested from a senior for `research-exhaustiveness` only now (its sole invalidation is main's unrelated dead `model:` pin removal) — `init-project` no longer qualifies, see above.


eval/harness/validators/test_init_project.py
```diff
@@ -27,9 +27,17 @@
...
+    if the other validates.
+
+    Exception: tests tagged `no-premature-write` verify the opposite --
+    that init-project correctly blocks and asks (e.g. a bare PID with no
+    stated objective, per issue #1320) instead of writing files before it
+    has what it needs. For those, no files existing is the correct,
+    passing outcome, not a structural failure."""
     if test.get("type") != "positive":
         pytest.skip("file-existence rules apply only to positive tests")
+    if "no-premature-write" in test.get("tags", []):
```
chrisedeson, line 39: This turns a check **off** and nothing **on**, so the new test can't fail on the behaviour it exists to guard.

I simulated init-project wrongly writing both files and ran the validators against `ut_init_project_010`:

```
BASELINE (nothing written)     FAILED=0   12 of 14 validators skipped
MUTANT   (both files written)  FAILED=0   <- nothing catches it
```

The committed run log agrees: 14 validators `"passed"`, 12 of them with `"error": "skipped: ..."`, and the two that ran are vacuous against `"tool_calls": []`. The judge is carrying this test alone.

`test_runnability.py:279-281` names this exact shape: *"`<tag> in tags -> skip` turns a validator OFF for tagged tests, so it can never be the invariant a grade_on_invariant test leans on."*

Fix is to pair it — same tag, opposite sign:

```python
def test_no_premature_write(after_state, tool_calls, test):
    if "no-premature-write" not in test.get("tags", []):
        pytest.skip("not a no-premature-write scenario")
    assert after_state.get("research_json") is None
    assert after_state.get("tree_gedcomx_json") is None
```

Verified: passes on your baseline, fails on both premature-write mutants. In-tree precedent is `test_search_records.py`'s `test_no_search_or_writes_on_planning_request`.

packages/engine/plugin/skills/init-project/SKILL.md
```diff
@@ -105,9 +105,11 @@ Confidence: "I'm sure / definitely" → `confident`; "I think / maybe" → `unsu
 
 ### 1. Get the research objective
 
-Get from the user: a FamilySearch person ID (preferred), or name + known facts for `person_search`; and the research objective in one sentence.
+**This step blocks — unlike the profile and holdings interviews below, do NOT proceed past it without an explicit objective.** Before calling `person_read`, building the tree, or doing any pedigree analysis, you need BOTH: (a) a FamilySearch person ID (preferred) or name + known facts for `person_search`, AND (b) the research objective in the user's own words.
```
chrisedeson, line 108: The fallback you removed here still ships verbatim one directory over:

`packages/engine/plugin/skills/init-project/references/research-process-init.md:143`

```
| User gives only a person ID | Fetch data, analyze gaps, propose a default
  objective based on what is missing |
```

That's the exact instruction these lines now forbid. A repo-wide grep for "default objective" returns that line and nothing else, so it's the whole blast radius. `references/` is doctrine the skill reads, so #1320 isn't closed while it's there.

Worth folding in **before** the re-run, or the fix costs a second paid run.

eval/tests/unit/init-project/blocks-on-bare-pid-no-objective.json
```diff
@@ -0,0 +1,22 @@
...
+    "name": "Blocks on a bare PID with no stated objective instead of inventing one",
+    "type": "positive",
+    "description": "The user supplies only a FamilySearch person ID, with no research objective at all. Per SKILL.md Step 1, the objective is the one blocking question in this skill — init-project must ask what the user wants to research rather than inventing a default objective from the person's data, and must not call person_read, build the tree, or write any files before the objective is known.",
+    "tags": ["initialization", "objective-required", "blocking-question", "issue-1320", "no-premature-write"]
+  },
+  "input": {
+    "user_message": "Start a new research project. FamilySearch person ID LZNY-BRF.",
+    "scenario": null
+  },
+  "mcp_fixtures": [],
```
chrisedeson: With no fixtures, `person_read` / `person_search` / `place_search` are never registered at all — `mock_mcp.py` only registers a tool if a fixture backs it, plus the `LIVE_TOOLS` set, and none of these three are in it. So the `judge_context` line "should NOT call `person_read`..." can't fail.

The run's own output shows it: *"The person_read tool isn't available in this environment's deferred tool set"* — and the judge banked that as compliance.

Every other init-project positive test declares fixtures. Suggest `"mcp_fixtures": ["person-read-flynn"]` so declining to call it is a real choice, then let the paired validator assert the absence mechanically.

packages/engine/plugin/agents/record-extractor.md
```diff
@@ -383,11 +386,12 @@ every co-resident / household member** — one per person, exactly as you
 emit each persona's `name` assertion — never only for the head or
 searched persona and then dropped for the rest (setting it on the first
 persona alone is the known failure mode, the same trap called out for
-`record_persona_id`). It matters because `materialize_facts` reads a
-persona's `sex`/`gender` assertions to set the gender of any tree person
-it mints — omit it and every household member, spouse, or relative that
-person-evidence later mints from this record is silently created
-`gender: "Unknown"`, even when the record states the sex plainly.
+`record_persona_id`). It matters because person-evidence's later
```
chrisedeson, line 392: Consistency point: the other ten gaps you fixed by **adding a deny**; this one is fixed by **deleting the tool name** from model-facing prose. `materialize_facts` is a real tool and unconditionally live in the mock.

Adding it to `disallowedTools:` (three spellings) would match how you handled the rest, and would keep the more specific guidance for the agent.

Also worth correcting in the body: that makes it 10 denies plus 1 prose deletion, not "11 agent-frontmatter gaps".

eval/tests/unit/research-exhaustiveness/declare-exhaustive-complete.json
```diff
@@ -11,12 +11,12 @@
     "user_message": "All planned searches for q_001 are complete. Three independent sources confirm Thomas Flynn as father: 1850 census co-enumeration, 1860 census co-enumeration, and the death certificate naming Thomas directly. The probate search returned negative (log_006). The birthplace conflict (c_001) is resolved. Overturn risk is low — no unsearched record type is likely to name a different father. Please update research.json: set q_001's exhaustive_declaration to declared: true, set the status to exhaustive_declared, fill in all seven stop_criteria assessments, and validate the schema.",
     "scenario": "flynn-exhaustive-ready"
   },
-  "mcp_fixtures": ["validate-research-schema"],
+  "mcp_fixtures": [],
```
chrisedeson, line 14: These two files also had `mcp_fixtures` emptied and `judge_context` rewritten to assert `research_append` is the complete persistence step. That's a grading-contract change, not just "softened descriptions" — worth saying plainly in the body so the genealogist reviewer knows what they're approving.

Defensible on the merits, since the skill's `allowed-tools` is `research_append` only.

chrisedeson (commented):
Reviewed at `a3b787f0`. Three blockers, all fixable. Two are inline; this one isn't tied to a line.

First the good news, and it's the part I expected to be riskiest: every deny you added carries all three tool-name spellings, names a real tool, and clashes with no `tools:` grant. I mutation-tested it — deleting any single one of them fails CI. That's the thing that fails open and silently when it's wrong, and yours is right.

## Blocker — `runlogs` CI is currently red, 4 rule-2 violations

Two separate causes.

Your agent edits **newly** invalidate `record-extraction`: `image-reader.md`, `image-reader-opus.md` and `record-extractor.md` all match main at the merge-base, so this PR is what makes them stale. A new deny changes what the agent may do under `bypassPermissions`, so it isn't cosmetic — that skill needs a re-run.

Separately, both run logs you committed were captured at `11f8c7a7`, before the branch merged main. That merge pulled in #1497, which deleted the `model:` pins, so the `init-project` and `research-exhaustiveness` SKILL.md snapshots no longer match the tree. The delta in each is exactly one line: `-model: claude-sonnet-4-6`. If #1497's "dead pin" claim holds, that's behaviour-neutral, so `eval-cosmetic-skip` from a senior beats paying for two re-runs. The label clears on every push, so it goes on last.

## Smaller things

**`ut_research_exhaustiveness_017` isn't "pre-existing".** A committed prior run (`v1_2026-07-23_17-04-02`) passed it 3/3/3/3/3/3 on byte-identical inputs — same SKILL.md hash, same rubric, same scenario, same judge prompt hash. So it's one pass and one fail on identical inputs: a flake at n=2, not a known failure. Your annotation doesn't claim variance either, it endorses the failure in six comments. Worth restating as what was observed, then either one more run or filing it.

**#1447 is closed as not-planned** (2026-08-09), but the body still lists it as a live follow-up, so a reader would think that scope is tracked.

**The stripped fixture is `validate-research-schema`,** not `validate-schema`. The count of 7 is right. It's dead for `init-project` only — 20+ tests in other skills still use it.

**On #1522 / #1447 / #1448:** all three genuinely clear the fold-it-in gate from #1449, no argument from me. One process note — that rule asks you to name which of the four exceptions applies, in the body. #1522 is "different skill's eval slot"; the other two are "needs a different reviewer".

chrisedeson (changes requested):

chrisedeson (approved):
Everything from my review is addressed. Approving.

johnmarkpeterbrown (commented):
Re-ran the §1 gate. A new commit landed since last time (7d42334 → afff9b43, chrisedeson re-approved on it), but neither blocker was cleared:
Review: PR #1523 — still not ready for another review.
Red check: runlogs still fails on the new head (afff9b43) — same three skills' run logs out of sync (init-project, record-extraction, research-exhaustiveness). mergeStateStatus is still BLOCKED.
Open round: the same 5 unresolved threads remain, and the author still has not replied to any of them:

[test_init_project.py:39](https://claude.ai/epitaxy/eval/harness/validators/test_init_project.py)
[SKILL.md:108](https://claude.ai/epitaxy/packages/engine/plugin/skills/init-project/SKILL.md)
[blocks-on-bare-pid-no-objective.json](https://claude.ai/epitaxy/eval/tests/unit/init-project/blocks-on-bare-pid-no-objective.json)
[record-extractor.md:392](https://claude.ai/epitaxy/packages/engine/plugin/agents/record-extractor.md)
[declare-exhaustive-complete.json:14](https://claude.ai/epitaxy/eval/tests/unit/research-exhaustiveness/declare-exhaustive-complete.json)
The new push re-triggered CI and got a fresh approval, but it did not answer chrisedeson's comments or fix the failing check.
Waiting on: the author.

eval/harness/validators/test_init_project.py
florencemashipei: Fixed — `test_no_premature_write` (eval/harness/validators/test_init_project.py:49-61) now asserts both `research_json` and `tree_gedcomx_json` are `None`, paired with the same `no-premature-write` tag gate as the skip above. Mutation-verified per your suggestion: fails on the both-files-written mutant, passes on the correct behavior.

packages/engine/plugin/skills/init-project/SKILL.md
florencemashipei: Fixed — the decision table in references/research-process-init.md:143 no longer says "propose a default objective"; it now reads "Ask what they want to research before fetching anything -- do not invent an objective from the person's data (see SKILL.md Step 1)", matching the SKILL.md Step 1 rewrite. #1320 doesn't half-ship.

eval/tests/unit/init-project/blocks-on-bare-pid-no-objective.json
florencemashipei: Fixed — blocks-on-bare-pid-no-objective.json now declares `"mcp_fixtures": ["person-read-flynn"]`, so person_read/person_search/place_search are registered and declining to call them is a real, gradeable choice. Per the PR body, re-running with this fixture in place actually caught a real latent bug in the #1320 fix (premature person_read call), which is now also fixed.

packages/engine/plugin/agents/record-extractor.md
florencemashipei: Fixed — `materialize_facts` is now in `disallowedTools:` with all three spellings (lines 69-71), consistent with the other 10 gaps. No prose-only deletion remains for this one.

eval/tests/unit/research-exhaustiveness/declare-exhaustive-complete.json
florencemashipei: Addressed — the PR body now states this plainly: "research-exhaustiveness's declare-exhaustive-complete.json and honest-early-termination.json had their mcp_fixtures emptied and judge_context rewritten to assert research_append is the complete persistence step (not just reworded descriptions) — a grading-contract change, defensible since that skill's allowed-tools is research_append only, called out explicitly here per review."

johnmarkpeterbrown (approved):
looks good.

clack391 (approved):

PR #1535, opened 2026-08-10
Add second review-mode test and baseline run log for proof-conclusion (#1392)

EdmondOware:
… (#1392)

Preconditions work from #1392, without the §9 edit.

- New scenario flynn-proof-bare-ids: mid-research-flynn with a non-self-contained ps_001 narrative (bare IDs, no prose, no citations)
- New test ut_proof_conclusion_r7k, pairing with _007 on that scenario
- Baseline run log v1_2026-08-07_09-26-35 + annotations (132/132 reviewed)

Baseline no longer reproduces the failure: _007 scores 3 on all dimensions with SKILL.md unchanged, and r7k does not reproduce either. /audit-rubric flags the cited 08-03 dip as one-off jitter. §9 left untouched per Dallan.

Not done: hold-out designation. Toggling "hold out from the skill-improver" in the CRUD UI did not persist — no change to the test files on disk, and the toggle read as off afterward. So proof-conclusion still carries 0 hold-outs and the anti-overfitting check remains inert. Needs a follow-up.

## Summary

**Tier: Normal.**

- Adds the two artifacts #1392 asked for as preconditions to the §9 edit: a second review-mode test (`ut_proof_conclusion_r7k`) and the scenario it runs on (`flynn-proof-bare-ids`), plus a fresh baseline run log with full annotations.
- **Does not make the §9 edit.** The baseline no longer reproduces the failure the issue is built on: `ut_proof_conclusion_007` scores 3 on every dimension with `SKILL.md` unchanged, where #1392 records it at 2 on Completeness and Narrative standalone. The new test doesn't reproduce it either.
- Checked in primarily so there is an **active run log** for proof-conclusion as a reference example, per Dallan on Slack.

## Review guidance

**Start here:** `eval/tests/unit/proof-conclusion/gps-review-narrative-not-self-contained.json` — specifically its `judge_context`. The question worth your eye is whether the defect I built into the scenario is **too blatant to discriminate**. `flynn-proof-bare-ids` strips `ps_001`'s narrative down to bare IDs with no prose and no citations at all, which rewards a reviewer for *noticing* rather than for *systematically checking* — and the gap #1392 describes is a failure to check. A subtler defect (mostly clean prose, one or two buried bare IDs, citations quietly missing for one of three sources) may be what's actually needed. The scenario delta is documented in `eval/fixtures/scenarios/flynn-proof-bare-ids/README.md`; it is byte-identical to `mid-research-flynn` except for `proof_summaries[0].narrative_markdown`.

**Unsure about:** whether **Narrative standalone** can discriminate at all in its current form. `/audit-rubric proof-conclusion` reports it at 68/1/0 across five run-log versions, with the single partial being `_007` in `v1_2026-08-03_09-46-24` — the exact run this issue cites — which it classifies as one-off jitter with a framework-level remedy (`runs_per_test > 1`). If that's right, sharpening the test only moves the problem, and the real work is the adversarial-fixture gap the audit flags under Flags 1/3. That call is senior-owned and I'd rather not make it.

## Plan

**Didn't change:** `proof-conclusion/SKILL.md` §9 and `references/gps-proof-writing.md` — the edit this issue was filed for. Also left `eval/tests/unit/proof-conclusion/rubric.md` alone (the audit assesses the prose as sound; the gap is fixture-side), and `ut_proof_conclusion_012` alone despite the finding below — separate problem, separate PR.

**Acceptance check:** **None — and that is the result.** There is no test that fails on `main` and passes here. `ut_proof_conclusion_r7k` passes on both, which is precisely why the §9 edit was not made: #1392's gate criterion is *"Completeness and Narrative standalone now score 3 on `_007` + the new test"*, and that is **already true at baseline**. Gating an edit on it would return green while proving nothing. What this PR delivers instead is the evidence for that, checked in and annotated.

**Deviated from the plan:** **Yes.** The branch was scoped to add the tests, run the baseline, then make the §9 edit and gate it. It stops after the baseline. I raised the non-reproduction and the `/audit-rubric` findings with Dallan on Slack; he confirmed both issues were known and asked for the run log to be checked in as an example. No §9 edit was attempted.

## Test plan

- [ ] `make test-all` — **not run: Windows, no equivalent exists (#1185).** Leaning on CI, which runs the same suites. No code/infra files in this diff — it is entirely `eval/fixtures/`, `eval/tests/` and `eval/runlogs/`, so per `.github/CODEOWNERS` peer approval should be sufficient here.

- [x] Run-log snapshot changed (new scenario + new test under `eval/tests/unit/proof-conclusion/`). Ran `RunTests.bat` for proof-conclusion and committed `v1_2026-08-07_09-26-35.json` **and** its `.ann.json` — 132/132 dimensions reviewed, not blanket-agreed. The 6 pruned older run-log files are deleted in this commit as the harness instructed.

- [ ] No `description` frontmatter or DO NOT clauses were touched — **N/A**, no routing pairs affected.

## Follow-on issues

Four items surfaced during this PR. None are filed yet — I'll add numbers here once they are.

1. **`ut_proof_conclusion_012` — judge misreads unincorporated sources as fabrications.** Its Evidence-completeness dings across three runs (07-28, 08-05) each allege the skill referenced a source not in the data: `src_006`, a Mary Elizabeth Flynn 1905 birth certificate, and a Pottsville Republican 1908 obituary. **All three are in `mid-research-flynn` verbatim** — `src_006`, `src_007`, `src_009`. They have zero extracted assertions, i.e. they are *unincorporated sources*, which `_007`'s `judge_context` carries an explicit scope note about ("evidence GAPS, not defects"). `_012` runs the same scenario with no such note. This is lane 2 — fix `_012`'s `judge_context`, not the skill. It also **overturns the audit's Flag 2 attribution**, which called this skill-behavioural jitter on the grounds that the rationales were "substantive and human-endorsed." My Correctness correction on `_012` in this run log (the `a_009` one) is the same pattern. *Caveat: I checked current `main`'s fixture, not the 07-28/08-05 snapshots.*

2. **`ut_proof_conclusion_001` aborts on `max_turns`.** It sets no `max_turns`; its sibling `_020` on the same scenario (`mid-research-flynn-no-proof`) sets 60. The trace shows 21 turns with both `proof-conclusion` and `check-warnings` invoked. Pre-existing, surfaced here. An aborted test yields **zero gradeable dimensions**, which will block the CI annotation gate for any future run.

3. **Hold-out toggle does not persist.** Flipping "Hold out from the skill-improver" on `_001`, `_008` and `_011` in the CRUD UI produced **no change to the test files on disk**, and the toggle read as off afterward. So proof-conclusion still carries **0 hold-outs of 20** and the anti-overfitting check remains inert — which was one of #1392's two stated preconditions. Not re-attempted here because editing test files would knock this run log out of active state.

4. **`ut_proof_conclusion_010` is a persistent `xpass`** — stale expected-outcome, reads as dashboard noise. Also noted by the audit.

**Folded in / filed instead:** Nothing folded in. All four are separate problems from the one this branch was opened for, and folding any of them in would have changed the corpus snapshot and invalidated the run log this PR exists to deliver.

One correction to #1392's framing worth recording: it states `_007` is the only review-mode test. Three tests carry `no-new-proof-expected` — `_007`, `_011`, `_012` — so the rubric's `(review)` branch already applies more broadly, and any future §9 change affects all three.


Cia-3 (approved):
Reviewed against the actual data, not just the write-up.

**Verified independently:**
- The central claim holds: pulled the run log directly — `ut_proof_conclusion_007`
  scores clean 3/3 on every dimension (including the two #1392 cites at 2), and
  the new `ut_proof_conclusion_r7k` also scores clean 3/3. The failure #1392
  was filed on genuinely does not reproduce, on either test.
- Annotation is real and complete — 132/132 dimensions, one genuine human
  override (`_012` Correctness 2→3) with a substantive rationale, not a
  blanket agree-with-all.
- Follow-on issue #1's evidence checks out: `src_006`/`007`/`009` exist
  verbatim in `mid-research-flynn` with zero extracted assertions — genuinely
  unincorporated, not fabricated. The judge misreading that on `_012` is real.
- The correction to #1392's framing is correct — `_011` and `_012` already
  carried `no-new-proof-expected` before this PR; `_007` was never the only
  review-mode test.

**CI failure, diagnosed (not a defect in this PR):** the `runlogs` check
fails because `main` moved out from under this branch — PR #1497 deleted
`model: claude-sonnet-4-6` from `proof-conclusion/SKILL.md` after this
branch was cut. Confirmed the run log's snapshot has zero mismatches
against this branch's own tree; the only diff is that one now-dead
frontmatter line, which this repo's own `agentic-system-critique.md`
documents as behaviorally inert. Recommend `eval-cosmetic-skip` rather than
burning a re-run rebasing onto a no-op change.

**Open, legitimately unresolved (not this PR's job to close):** whether
`flynn-proof-bare-ids`'s defect is too blatant to discriminate — both tests
passing is consistent with "the model checks systematically now" and
"the defect is too extreme to miss even without checking." Correctly left
for the next `/audit-rubric` pass rather than decided here.

**Recommendation: approve.** Good scientific integrity — found the
evidence didn't support the planned edit and stopped rather than forcing
one. New test/scenario are well-built. The CI failure is main-drift, not
this PR's fault; worth a `eval-cosmetic-skip` label rather than a wasted
re-run.

johnmarkpeterbrown (approved):
Looks good!

PR #1538, opened 2026-08-10
feat(server): encrypt FamilySearch OAuth tokens at rest (#1128)

### LINKED ISSUE #1128: Encrypt familysearch_tokens access/refresh at rest before real-PII alpha
`familysearch_tokens.access_token` / `refresh_token` are stored plaintext
(`apps/server/app/models.py:42-44`). `docs/specs/hosted-web-workbench-spec.md:246-247`
requires "store {access, refresh, expiresAt} per user in the DB (encrypted at rest)".

Encrypt on write in `_persist_fs_token` / decrypt in `fresh_fs_token`
(`apps/server/app/auth.py:91`, `:122`) with a key from `fly secrets`; migrate
existing Neon rows.

Gate: before any real living-person PII / wider-than-allowlist alpha.
The `/v1` path is unaffected — it never persists a token.

Drive-by: the TODO comment at `models.py:42` cites "spec §13", which covers
sandbox secrets and residency, not DB encryption. Retarget it to §5.2.2.

MMagnum:
## Summary
- **Tier: Normal.** Encrypt `familysearch_tokens.access_token` / `refresh_token` at rest via a Fernet-backed `EncryptedStr` column type — plaintext in memory, ciphertext in the DB, transparent to every caller.
- Decrypt failure soft-fails to `None` and is treated as `"expired"` (logged at WARNING with `user_id`), so a legacy-plaintext or wrong-key row self-heals on the next login — no backfill/migration.
- Key from `FS_TOKEN_ENC_KEY`; a prod deploy still on the dev default is refused at boot by the existing `assert_production_config` preflight.

Closes #1128.

## Review guidance

**Start here:** `apps/server/app/crypto.py` — `EncryptedStr` and its soft-fail `process_result_value` (returns `None` on `InvalidToken`, never raises). The one non-transparent piece is the guard in `auth.py::fresh_fs_token`; confirm it sits *after* the `if row is None` check so it can't inject an empty token into a sandbox.

**Unsure about:** Nothing outstanding — the design (column type vs. the issue's two-function seam; soft-fail vs. fail-hard; no backfill) was reviewed and approved by @DallanQ on the issue.

## Plan

**Didn't change:** the token copy injected into each sandbox at `~/.familysearch-mcp/tokens.json` (`fs_oauth.write_tokens`) stays **plaintext** by design — the in-sandbox MCP needs it and it dies on the same ~24h FS grant clock. At-rest protection here is DB-scoped, not a gap.

**Acceptance check:** `apps/server/tests/test_fs_token_encryption.py::test_tokens_are_ciphertext_at_rest` — asserts the raw DB column is ciphertext that Fernet-decrypts to the original (fails on `main`, where the column is plaintext). Plus `test_undecryptable_row_reads_as_expired_and_heals` (soft-fail/heal path) and the new `fs_token_enc_key` cases in `test_prod_preflight.py`.

**Deviated from the plan:** the planned startup guard **already existed** as `config.assert_production_config` (shipped for #1123/#1367) — so I extended it by one field instead of adding a new guard to `main.py`.

**Step-4 stop rule:** this touches auth-token storage, so I escalated the design to @DallanQ before building; approved with changes (drop the backfill, soft-fail + WARNING log) in https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1128#issuecomment-5243185634. Those changes are folded in.

## Test plan

- [ ] `make test-all` — **no Windows equivalent (#1185).** Ran the affected server suites locally: `test_fs_token_encryption.py` (4) + `test_prod_preflight.py` (9) + `test_familysearch.py` (17) = **30/30 pass**. The full `pytest` run has pre-existing sandbox/agent-test failures + a hang on this Windows box, unrelated to this change (verified: no test boots with an https `PUBLIC_URL` expecting success, so the tightened preflight can't affect them). Leaning on CI for the full suite.
- Skill run-log / description boxes: **N/A** — no skills, agents, fixtures, or descriptions touched.

## Follow-on issues

**Folded in / filed instead:** Nothing filed. @DallanQ explicitly declined two candidates during review: per-user / per-token keys (no isolation gain in this architecture — the control plane must decrypt while the user is away, so the key lives in the same DB/env; the real upgrade is a KMS) and the sandbox plaintext copy (same 24h clock, swept on the 7–30 day delete).


francis-2008-happy (commented):
Read through this and ran it. Really clean the encryption is transparent to callers, and I traced every place the token gets read: they all go through the one guard, so a bad/undecryptable token can never get pushed into a sandbox as an empty value. That was my main worry and it's handled. Tests pass 30/30 and CI's green.

One small thing, not a blocker. The decrypt is meant to quietly fail and return nothing on a bad value but it only catches one kind of error. I tested it: a token with a non-normal character throws a different error instead of failing quietly. Real tokens won't hit this (they're plain ASCII), so it'd only happen on a corrupted row, but the fix is one line and I confirmed it works without breaking any tests:

You wrote:      except InvalidToken:
Change it to:   except (InvalidToken, UnicodeError):

Nice work.

T-FEH (approved):
Approving. Solid piece of work — the Core-layer `TypeDecorator` is the right seam precisely because no caller can forget it, and the design matches Dallan's ruling point for point.

I pushed one commit (`1f58e81a`) rather than send it back: `DEVELOPMENT.md` gained the new required secret, but `deploy/fly.toml`'s own header comment and `docs/plan/fly-deploy-plan.md`'s table and `fly secrets set` block did not. An operator following either omits `FS_TOKEN_ENC_KEY` and `assert_production_config` refuses the boot — loud, not silent, but avoidable, and those two files are the checklist people actually deploy from.

Verified by running, not reading:

- **The guard is load-bearing, not defensive dressing.** Deleting `auth.py:138` makes the heal test fail, returning `FamilySearchToken(access_token=None…)` headed for `write_tokens`. The hazard is real and it is regression-tested.
- **The acceptance test genuinely pins the fix** — reverting the column to plaintext fails `test_tokens_are_ciphertext_at_rest`. (On unmodified `main` it only `ModuleNotFoundError`s, which proves nothing; the mutation is the real check.)
- **The preflight really refuses** — dropping `fs_token_enc_key` from the loop fails two prod-preflight tests.
- **No blank token can reach a sandbox.** Every DB read of the tokens is `fresh_fs_token`; `sessions.py:160` maps `None` → `"expired"`, and `sessions.py:197` is the `/v1` caller-supplied path, never DB-sourced.
- Only `user_id` is logged — no key or plaintext in any message or traceback.

`apps/server` suite: **162 passed, 1 skipped**. All 12 CI checks green.

@francis-2008-happy your `UnicodeError` catch landed in `12d1aaa3`, with a regression assertion in the test.

@clack391 you gated on this, so directly: **yes, Dallan signed off on skipping the migration specifically** — "Approved with one change: **drop the backfill** … Cut all of it" (issue #1128, comment 5243185634). Your dormant-row residual is bounded tighter than it looks: `auth.py:128` records that FS caps *the grant* at 8h idle / 24h absolute, so a stale row's refresh token dies with its access token rather than outliving it.

Non-blocking, left alone: `models.py:47` annotates `access_token: str` while it can be `None` at runtime — harmless today with no mypy on `apps/server`, but a trap for the next caller. Key derivation is a bare SHA-256 rather than a KDF, which is fine for the high-entropy value the docs mandate and not worth changing before the KMS move Dallan describes.

**This has no approvals yet, so mine is 1 of 2** — it needs one more from `senior-developers`, and since I made the last push it cannot be me.

clack391 (approved):
Senior review, 2 of 2. Verified independently rather than leaning on the
existing approval: the design matches Dallan's ruling point for point (the
column-type seam, backfill dropped, soft-fail self-heal, the auth.py guard, and
the prod-boot refusal on a default key). The ciphertext-at-rest test reads the
raw column, the guard is load-bearing under mutation, only user_id is ever
logged, no secret is committed, and the merge with main is clean with the server
suite green at 162 passed.

One non-blocking follow-up, not for this PR: FamilySearchToken has no custom
repr, so the default one would render a decrypted token if a token-bearing row
were ever logged or traced. Nothing does that today, so it is latent rather than
a live leak, but a redacting repr on the two columns would close it.

Approving. Nice work, and thanks @T-FEH for the first pass.

PR #1539, opened 2026-08-10
eval(e2e): derive relationship `matched` from the judge's components (#1090)

Ikennaya1:
## What was wrong

An AI grades each e2e run — did the agent find what it was meant to find? On one run it was asked whether the agent had recovered a family relationship. The agent's tree had the person's name and nothing else: no link to either parent, no marriage. The grader marked it **recovered** — while its own notes correctly described the missing links.

## Rewriting the instructions didn't fix it

Four attempts, each more explicit. The last one says in plain words that this case must be marked "false". The answers went `true` → `partial` → `partial` → `true` and never landed right. Every time the analysis was correct and the recorded answer wasn't.

## So the answer is now calculated

The grader lists the relationships a finding claims and marks each present or absent. The score is worked out from that list in code, rather than left to the grader. That's how the run-level verdict already works.

**Only relationships count.** Findings often describe the people involved ("born 1833 in the Gorbals; an iron moulder") — that says *which* person is meant, not what the agent must file. An earlier version counted it and wrongly downgraded twelve findings that were actually correct.

## Results

| | Before | After |
|---|---|---|
| Agreement with human grades | 84% | **87%** |
| Disagreements | 50 | **40** |
| Relationship disagreements | 27 | **16** |
| The original bad grade | wrong | **fixed** |

Date/place findings unchanged (21 → 22) — the calculation doesn't touch them. Two runs hit connection errors, so 303 of 310 findings graded.

## One case worth a look

`jens-nielsen` f3 is a new disagreement. The mother link isn't in the tree; a human gave half credit because she's named in the written summary. The instructions already say credit comes from the tree, not the write-up — and the finding exists to establish which of two wives was the mother, so assuming it from the father's marriage is the mistake it guards against.

Probably a second example of the same bug, not a new problem — but that's a genealogist's call. Filed as #1540 for a fresh re-grade.

## Worth knowing

Older grades came from the previous grader. Not wrong, but not directly comparable with anything graded from now on.

Closes #1090.


francis-2008-happy (approved):
Read through it and ran it. Really clean work. The fix computing the grade from the judge's own components instead of trusting its label  is the right call, and it matches how the existing verdict logic already works.

I checked the two things that could break it (the copy import and expected_findings being in scope)  both fine. And I mutation-tested the core function: forcing it to always return "true" breaks 8 tests, so the tests genuinely do their job. Old run logs still work too. All green for me.

Two small non-blockers: the spec documents the sibling apply_avoid_guard but not this new one, so a matching paragraph would be nice; and the schema now requires components on every finding while the prompt only explains how to decompose relationship ones (fact findings aren't scored on it, so no grading impact).

Approving on the code side. Whether deriving the grade from components is the right grading approach is really a genealogist's call leaving that to them.

Nice work.

clack391 (approved):

Emruthwill (approved):
I agree with this change. From a genealogical perspective, identifying a person is not the same as establishing a family relationship. The parent, spouse, or child link must be supported in the tree; a narrative mention or inference should not receive full credit. I also agree that the `jens-nielsen` case should be re-graded rather than weakening the existing rule. Overall, this is a sound and genealogically appropriate fix.

johnmarkpeterbrown (commented):
Merge main into branch and notify me for final review and merging. Thanks

johnmarkpeterbrown (changes requested):
Nice work on this, and thanks for turning it around. I'm happy with it from the genealogy side — a relationship only counts when the parent, spouse, or child link is actually sitting in the tree, not when it's inferred from the narrative, and pulling the grade from the link components does exactly that. Before I clear my sign-off, I need three things from you, and there's one I'm telling you to leave alone.

First, get `main` merged back in. You're 5 commits behind and I want the branch current before I do the final merge. Nothing clever here, just bring it up to date.

Second — and this is the one I actually care about — there's a hole in `judge_prompt.md` around marriage dates. The scoping bit says a spouse link with no marriage date should come out `"partial"`, but the derivation only scores `link` components and the taxonomy just above it puts dates in `detail`. So if the judge tags a required marriage date as a detail, a bare spouse link with no date established rolls straight up to `"true"`. That's full credit for a marriage we never dated, which is the exact thing this whole change is supposed to stop. `manoel` slips past it only because we accept either date there. I want a required date to be able to pull the label down. Add something to the scoping list like:

> - A **claimed date** (marriage, birth, death) that the finding requires is a scoring `link` component, not a `detail`. A relationship whose date isn't established in the tree is `"partial"`, not `"true"`.

Word it however fits your intent — I just need the date to count, not sit in the unscored pile.

Third, put the new backstop in the spec. `docs/specs/e2e-test-spec.md` §3.4.1 already writes up `apply_avoid_guard` and its annotations; the component derivation needs the same treatment — `apply_component_derivation`, the `component_derivation` overrides, `matched_model`, the `[component-derivation]` note — and §7.2 should say `matched` is derived for relationship findings now. If the rule's live, the spec has to be too.

And the thing to leave alone: don't touch the rule for `jens-nielsen` f3. A mother named only in the write-up is precisely the case we're declining to credit, and I don't want the standard softened to make that one pass. I'll re-grade it blind under #1540 once this is in.

Merge main, sort the date gap and the spec, and give me a shout IN SLACK— I'll clear the genealogist approval then and merge. Please notify me in SLACK when this is done and all is GREEN. Thanks

johnmarkpeterbrown (approved):
Looks good!

Issue #1194, opened 2026-08-03
check-warnings should distinguish improbable from impossible — the severity field exists, the doctrine does not


DallanQ:
**IMPORTANT**: If you do this, do #1616 at the same time. It carries four
doctrine corrections from the alpha-team review that land in the same three
check-warnings reference files PR #1545 already edits. Together they pay one
`make eval-skill SKILL=check-warnings` run instead of two.

**IMPORTANT**: If you do this, coordinate with issue #1191 — it adds a late-childbearing
check to the same file (`person-warnings.ts`) and has to use whatever severity names
this issue settles on. Land this one first.

**Scope note, 2026-08-07 fill-ready:** promoted to Ready for the **Developer** lane only (surface the tier, fix the stale `person-warnings.ts:286` comment). The **Genealogist** lane (audit the 75 tags into `contradiction`/`implausible`, populate the implausible band) is not part of today's promotion — it's a separate scope and should be picked up by a genealogist, ideally batched with the developer half's `check-warnings` SKILL.md edit so the skill's eval slot pays one run, not two.

## The ask

`check-warnings` should distinguish **impossible** from **improbable** rather
than returning one undifferentiated list.

## Two constraints on the fix — raised at triage, agreed by the lead 2026-08-03

**These are decided, not open questions. Do not re-litigate them; build to them.**

### 1. Name the tiers for consequence, not for likelihood

Do **not** ship the tiers as `impossible` / `improbable`. "Impossible" is a claim
the underlying data rarely supports: a record showing a mother aged 6 is far more
likely to carry a bad date, a mis-indexed year, or a wrong person attachment than
to describe an impossible birth. Asking an agent to sort findings by
likelihood-in-nature asks it to make a judgement it cannot make from the tree
alone, and invites it to argue — which is exactly the failure in issue #1191's
run 1, where the agent computed "a 1786 birth would make her 51 at this child's
birth" and then reasoned past it.

Name them for what the researcher must **do**:

| Tier | Means | Required response |
|---|---|---|
| `contradiction` | The data cannot be true as recorded | Stop. Resolve before writing anything that depends on it. |
| `implausible` | Possible, but unlikely enough to need support | May be written **only** with corroboration and an explicit note saying so. |

Same two buckets as the existing `error` / `warning` split. The difference is
that the tier name states the obligation instead of inviting an assessment, and
an obligation is checkable where an assessment is arguable.

[...]

PR #1541, opened 2026-08-10
eval-harness: shadow detector for parentage writes with no person_warnings check (#1193)

francis-2008-happy:
## The issue (#1193)

`check-warnings` / `person_warnings` is the cheapest guardrail in the system — deterministic, no LLM (it reads `tree.gedcomx.json` and evaluates ~75 predicates), so a call costs one tool round-trip. Yet it's almost never invoked. Two runs of the same fixture (`elisabetha-sugecz-parents`, PR #1113) diverged: one called `person_warnings` 6×, the other 0× — and the 0× run asserted parents implying the subject bore her first child at age 6, exactly the case `checkEarliestChildBirthToBirth12` exists to catch. Nothing recorded that the guardrail had been skipped.

Root cause: the "run check-warnings after a tree write" instruction lives **only inside the writing sub-skills** (`proof-conclusion`, `tree-edit`, the `validation-protocol.md` files). When the `/research` orchestrator **delegates** the parentage write to `proof-conclusion`, the step fires; when it **inlines** the write itself, the instruction is never in context. The variance is a permitted routing choice, not a coin-flip on check-warnings.

## The fix

A shadow-mode detector, `find_relationship_writes_without_warnings_check` (`harness/skill_invocation.py`), the sibling of the #1133 citation-nulling detector. It flags a run whose final tree has a **new** `ParentChild`/`Couple` relationship (diffed against the starting tree, so seeded relationships don't count) for which `person_warnings` was never **successfully** called.

- **Keyed on the `person_warnings` tool** (across all three server spellings via `bare_tool_name`), not the `check-warnings` skill — so it catches a direct/ToolSearch call and a skill that launches but fails before reaching the tool. A failed call (`is_error`) doesn't count as consulting the guardrail.
- **Shadow mode only:** folded into the existing `guardrail_shadow_violations` plumbing (`orchestrator.py`), counted in its own bucket (`kind: "warnings_unchecked"`) by `guardrail_shadow_report.py`. It **never touches `compliance`/`outcome`** — logs, never fails a run.
- Lifted `_relationship_key` to module scope so the §8 hard gate and this §7 shadow detector share one definition.

## What this does not solve

This is the **measurement** half — it makes the omission visible in eval; it does **not** change production behavior so the guardrail actually runs. The behavioral fix — a *mandatory* `person_warnings` call in the `/research` orchestrator so an **inlined** write is still gated (issue #1193, question b) — is a doctrine change with a paid e2e re-run cost that the issue says to "scope before committing," so it's deferred to the lead. Shadow-first here matches the #1358 / #1133 rollout; promotion to a hard gate waits on measuring the corpus fire rate. It also does **not** touch thresholds (#1191) or improbable/impossible tiering (#1194).

## Verification

- `eval/harness` unit suite green: **1852 passed**. The only failures are the pre-existing `unzip: command not found` env gap in `test_setup_feedback_case.py` (unrelated; CI has `unzip`).
- **9 new tests** cover: fires on a new link with no call; silent when `person_warnings` succeeded; still fires when the call errored; all three server spellings count as consulted; gated on a new relationship (seeded-only run stays silent); fires on a new `Couple`; ignores non-parentage relationship types; no-relationship → no finding; None/empty defensiveness. Discriminating (the "silent" cases fail if the gate is removed).
- **No skill run-log snapshot touched** (harness-only change), so no `eval-skill` run required.

## For review

- **Start in** `harness/skill_invocation.py` → `find_relationship_writes_without_warnings_check` and its docstring for the new-relationship gate + the tool-vs-skill keying.
- The one judgment call is keying on the tool name via `bare_tool_name` rather than the `check-warnings` skill — shadow mode is how we calibrate before any gating.
- **Collision:** open PR #1438 (my #1317 conflicts shadow detector) also edits `orchestrator.py` and `guardrail_shadow_report.py` in the same shadow-wiring region. Whichever lands second rebases (I author both).

Closes #1193 (measurement half; the mandatory-invocation doctrine change is deferred to the lead per the issue's "scope before committing").


clack391 (approved):

johnmarkpeterbrown (changes requested):
Hi Francic,
Here's a message you can send the author (francis-2008-happy):

---

Hi — picking up senior review on PR #1541 (the #1193 warnings-unchecked shadow detector). The detector logic and tests look solid; I ran the full harness unit suite on your head and it's green (1854 passed). Before I do the senior sign-off, two things:

**1. Update the branch.** It's 11 behind `main`. GitHub reports it merges clean, and I confirmed none of main's commits touched the three files you edited — so it should be a no-op rebase:

```bash
git checkout 1193-person-warnings-shadow-detector
git fetch origin
git merge origin/main
git push
```

**2. Fold in one small test fix on the same push.** A stray edit moved the empty-dict assertion out of `test_citation_nulling_defensive_on_none_and_empty` — it now only checks `None`, and its `{}` case ended up stranded at the bottom of `test_warnings_unchecked_defensive_on_none`. The suite stays green because the assertion still runs, but the citation test no longer matches its name. Since you're repushing anyway:

- Add back to `test_citation_nulling_defensive_on_none_and_empty`:
  `assert find_citation_nulling_in_conclusions({}) == []`
- Remove that same line from the end of `test_warnings_unchecked_defensive_on_none`.

Once it's rebased and pushed and CI is green on the new head, I'll do the review.

--

johnmarkpeterbrown (approved):
Looks good!

PR #1545, opened 2026-08-11
feat: rename severity from error/warning to contradiction/implausible

### LINKED ISSUE #1194: check-warnings should distinguish improbable from impossible — the severity field exists, the doctrine does not

yinkid28:
## Summary

Closes #1194.

- Rename `person_warnings` severity enum from likelihood-based (`"error"` / `"warning"`) to consequence-based (`"contradiction"` / `"implausible"`) across the full surface: type definition, 75 implementation literals, tool descriptions, specs, skill body + references, proof-conclusion cross-reference, eval fixtures, and eval test descriptions/rubric
- Fix stale comment on `earliestChildBirthToBirth` claiming the gender-neutral check was unimplemented (it is)
- No tags reclassified — all 12 former `"error"` tags become `"contradiction"`, all 63 former `"warning"` tags become `"implausible"`

**Touches:** `packages/engine/mcp-server/src/types/person-warnings.ts`, `packages/engine/mcp-server/src/tools/person-warnings.ts`, `packages/engine/mcp-server/src/tools/merge-warnings.ts`, `packages/engine/plugin/skills/check-warnings/SKILL.md`, `packages/engine/plugin/skills/proof-conclusion/SKILL.md`, `docs/specs/person-warnings-tool-spec.md`, `docs/specs/match-merge-workflow-spec.md`

## Test plan

- [x] `npx tsc --noEmit` — typecheck passes
- [x] `npx vitest run tests/tools/person-warnings.test.ts` — 113/113 pass
- [x] `npx vitest run tests/tools/merge-warnings.test.ts` — 23/23 pass
- [x] `npx vitest run tests/integration/match-merge-workflow.test.ts` — 3/3 pass
- [x] `npx vitest run tests/packaging/` — 385/385 pass
- [x] Grep for straggling `severity.*"error"` / `severity.*"warning"` across all scoped directories — zero hits


Asimi1234 (approved):

clack391 (approved):
Approving on the senior-developer side. The severity rename is clean and complete: the 12 / 63 mapping (error to contradiction, warning to implausible) is exact and 1:1 with no cross-mapping, no logic or thresholds moved (the only non-severity edits are the two declared ones, the consequence-framing description reword and the stale `earliestChildBirthToBirth` comment fix), tsc is clean, the renamed suites pass, no straggler `error`/`warning` literals remain, and the persisted research.json `severity` enum (high/medium/low on gaps) is correctly left untouched. Every blocker from the earlier rounds is resolved, including the spec catalog at lines 236/266/308.

For the senior-genealogist: the only red check, `runlogs`, is verified behavior-neutral and not caused by this PR. The two stale snapshot files are `mid-research-flynn/README.md` and `richardson-parentage-not-exhaustive/results/log_004.json`, both shared scenario fixtures that main's #1333 edited after these run logs were generated. The run logs already capture this PR's own SKILL.md change; the remaining staleness is unrelated README/log churn from main. That is the fixture-drift race (`mid-research-flynn` is referenced across many skills), so `eval-cosmetic-skip` is the intended way to clear it rather than another paid re-run. That label, plus your approval of the skill/eval/`.md` changes, is what remains before merge.

Issue #1546, opened 2026-08-11
image-reader: redesign the return contract - relay the browseBudget notice, gate the browse relay on found, and resume a truncated page


aghadiayeamayanvboernest:
**Touches:** packages/engine/plugin/agents/image-reader.md, eval/tests/unit/record-extraction/**, eval/tests/unit/search-images/**

Split off from #1081 (per the issue's explicit "do not touch image-reader.md in that PR" instruction and the lead's 2026-08-10 ruling).

#1081 added an advisory `browseBudget` field to `image_transcribe` — from the 21st distinct image in one image group in one project, a successful transcription carries a notice to pivot off page-by-page browsing. On the **main thread** the model sees it directly (~40% of `image_transcribe` calls corpus-wide). On the **delegated** `image-reader` path it never arrives: the agent's "What to return" list is a closed enumeration, and its verbatim relay only fires when the tool **throws**, so a notice riding a *successful* transcription dies in the subagent's throwaway context. One measured run (`elisabetha-sugecz-parents/run-2026-08-02_12-17-10`) is fully delegated (37 calls / 37 spawns) and would see nothing.

**Why it's a separate PR, not a fold-in:** editing `image-reader.md` is a plugin-agent change that gates **both** skills referencing `@plugin:image-reader` (`record-extraction` and `search-images`) — two `make eval-skill` runs plus two CRUD-UI genealogist annotations (~$8–12 and 45–65 min each). It also needs redesigning the agent's return contract to carry an advisory alongside a successful read, and a different (genealogist/senior) reviewer.

**Acceptance:** a delegated `image_transcribe` that trips the budget results in the `image-reader` subagent surfacing the pivot notice to its caller, verified by an eval run on a browse-heavy fixture.

Ref: #1081, spec §5.8 "Two things it knowingly does not do", ADR-0011 read-tool carve-out.

---

# Merged in from issue #1147 (2026-08-11): the browse-relay gating question

Issue #1147 asked a question about the same return contract this issue is redesigning,
on the same file, gated behind the same two eval runs and the same genealogist reviewer.
Answer it inside this issue's design rather than paying that bill twice.

**It is a question, not a decision.** Read `docs/specs/image-reader-agent-spec.md` §5.1
before proposing anything, and do not build it without genealogist sign-off.

## The question

A ten-page `search-images` browse accumulates ten full transcriptions in the calling
skill's context. Compressing non-matching pages to a one-line verdict was **rejected
[...]

PR #1547, opened 2026-08-11
image_transcribe: advisory browse budget from the 21st distinct image (#1081)

aghadiayeamayanvboernest:
## Summary

- An agent can enter an unbounded page-by-page hunt through a browse-only image volume and never leave it (motivating run: 58 `image_transcribe` calls, 2h wall-clock cap, no proof). The per-invocation `image-reader` bound doesn't count across invocations, and skill prose isn't loaded on >half the hunts — so the budget lives on the tool.
- From the **21st distinct image** in one image group in one project onward, `image_transcribe` returns its full transcription **plus** an advisory `browseBudget` field naming the count, the volume, and the pivot (index search or ask the user). **Advisory, not a refusal** — a page read persists nothing, so a wrong refusal would hard-block a researcher mid-browse with no reset but a server restart.
- Tier: **Normal**. Closes #1081.

## Review guidance

**Start here:** `recordBrowseAndCheckBudget` in `image-transcribe.ts`. The load-bearing choices: the counter is a **process-lifetime** `Map<string, Set<string>>` keyed `(projectPath, image group)` — keyed by project *deliberately*, because the MCP process outlives one conversation and a group-only key would tell a second project it had already browsed 20 pages on page one (no test/CI can see this — it's stated in §5.8). Group = digits before the `_` in `imageId`; ark-only calls carry no group, so they're uncounted (documented limit). The field is additive and fires only after a successful OCR.

**Unsure about — one thing for the lead:** with the mandated `${projectPath ?? "<no-project>"}` key, a hunt that forwards `projectPath` *intermittently* splits its count across the project bucket and the `<no-project>` bucket, so it could reach ~40 pages before the notice fires. This is within the issue's approved keying, but flagging it as accept-or-revisit since you own that call.

## Plan

**Didn't change:** `image-reader.md` and `search-images/SKILL.md` (editing either arms `check-runlogs` → paid eval runs); no new tool parameter (ADR-0006 — caller input is a request, not a constraint); `manifest.json`/`allToolSchemas`/schema mirrors (result-only field, no input/tool-list change); §5.6 Errors (nothing here throws).

**Acceptance check:** `tests/tools/image-transcribe.test.ts` → "attaches browseBudget on the 21st distinct image…" — 21 successful transcribes of distinct imageIds in one group/project; the first 20 return no `browseBudget`, the 21st returns `{ imageGroup, distinctImagesRead: 21, notice }` whose text contains the count, the group, and the pivot tokens (`research_log_append`/`record_search`/`record_read`/`fulltext_search`). Fails on `main` (field absent), passes here.

**Deviated from the plan:** nothing structural. `/critique-plan` returned no blocking findings. A self-review pass found two doc bugs, now fixed: §5.8 wrongly said "no unit test can observe this" when test 5 observes the project-keying (reworded to distinguish that from the cross-conversation persistence, which genuinely isn't test-observable); and the ADR `Last updated` date. **Did hit step 4's ADR-touch:** this ships a deliberate ADR-0011 departure (advisory on a read tool), recorded as a scoping carve-out per the lead's 2026-08-10 ruling — not a self-authored reversal.

## Test plan

- [x] Ran `make test-all`. Engine vitest **2112**, eval-harness **1852**, control-plane, eval-app all pass. The one red is the known **#1366** `apps/electron` flake (`buildFeedbackZip` times out at 5s under parallel load; **61/61 in isolation**, 3.1s) — untouched by this branch.
- [x] N/A — no skill run-log snapshot changed. Touches `mcp-server/**` + two `docs/` files; `check-runlogs` runs (its scope includes `^docs/`) but passes with no gated skill.

**Nothing checks the thing this is really for** (`docs/architecture.md` §9.4): every threshold number is from the eval corpus; no production telemetry can tell us whether a real agent actually pivoted, or whether the notice was ignored. Two accepted trades are in §5.8, not hidden: an advisory can't *force* a stop, and the delegated path can't see it.

## Follow-on issues

- **#1546** — relay the notice on the delegated `image-reader` path (the ~40%-blind gap; gates two paid eval runs + a genealogist review, so split off per the issue and the lead's ruling).

Closes #1081.


clack391 (approved):

T-FEH (approved):
Approving. Reviewed against the merge with `main`, not the branch head.

**Verified by running:**
- Test-merged `origin/main` in (the branch is behind): clean merge, then **2112/2112** engine vitest and `tsc --noEmit` clean. No "green on its own head, red once merged" surprise.
- **Mutation-tested the acceptance check three ways** and it fails all three: threshold 20→1000 flips 3 tests red; dropping `projectPath` from the map key flips the project-keying test red; `<=`→`<` flips 2 red. This test actually pins the behaviour.
- `imageId.split("_")[0]` is safe — `resolveFsImageInput` enforces `/^\d+_\d+$/` and throws before anything is counted, so a malformed id can never become its own singleton group.
- The counter sits after every throw path, so a failed OCR never advances it. `index.ts` `JSON.stringify`s the whole result and nothing downstream reads the shape strictly, so the additive field is safe.
- One local red — `plugin-hooks.test.ts` timing out at 5s under parallel load — is 14/14 in isolation and green on the merged run. Environmental, same shape as #1366, untouched by this branch.

**I pushed two doc corrections (a3d23ef)** rather than send them back for a round trip:
- ADR-0011 said "a placement question with four answers" directly above a table this PR grew to five rows.
- Spec §5.8 wrote the key as `${projectPath}\0${imageGroup}`, omitting the `?? "<no-project>"` fallback the code uses — which is precisely the part of the keying you flagged for the lead, so it should be visible in the spec.

**On the `projectPath` question you raised:** I reached the same conclusion independently, and agree with @clack391 that it ships as-is. One angle neither of you named, worth having on the record rather than in a follow-up: two *different* projects that both omit `projectPath` share the single `<no-project>` bucket — the same cross-project false positive the project-keying exists to prevent. It stays acceptable for the same reason: the notice is advisory, so a wrong fire costs one ignorable sentence, while the group-only alternative costs a real browse.

**Still needs a senior-genealogist approval** — both changed doc files are under `/docs/`, which routes there. My approval covers the `*.ts` half only, and @clack391's already did.

What I did not check: the harness and control-plane pytest suites (no Python in the diff, both green in CI on this head), and whether a real agent actually pivots on the notice — no telemetry exists to settle that, as the PR says.

T-FEH (approved):
Re-approving after updating the branch — `require_last_push_approval` needs an approval that post-dates the push, and my earlier one was @a3d23ef.

**Resolved a conflict in `ADR-0011`** (`aa945ca`). Main had moved 8 commits ahead and #1553 edited the same header block. Both sides touched three lines; I took the union rather than either side:

- **Applies to** — kept main's `scripts/claude-hooks` *and* this PR's `image-transcribe.ts`.
- **Related** — kept main's list and re-appended `#1081`.
- **Last updated** — `2026-08-11 (read-tool carve-out added; issue-filing gate added to Applies to)`, naming both.

Nothing else conflicted; `image-transcribe-tool-spec.md` auto-merged. The diff against `main` is still exactly the 5 intended files, so neither side lost anything.

**Re-verified on the merged tree:** engine vitest **2120/2120** (up from 2112 — main's new tests), `tsc --noEmit` clean.

**One thing worth knowing, since it changed after my first review:** #1551 landed on main and removed both the `*` catch-all and the `/docs/` rule from CODEOWNERS. So the senior-genealogist approval I said this needed is **no longer required** — `docs/adrs/` and `docs/specs/` are now unowned paths. The only code-owner requirement left is `senior-developers` on the `.ts` files, which @clack391 and I both satisfy. Two approvals are on it, so this should merge once the checks finish.

johnmarkpeterbrown (commented):
Merge main for final review and merging. Thanks you. Notify when done!

johnmarkpeterbrown (approved):
Looks good!

PR #1548, opened 2026-08-11
fix(skills): derive question and plan scope from the stated objective

Emruthwill:
Derives question and plan scope from the stated objective, in both directions, and fixes the vocabulary reports from the same three feedback zips.

Closes #1322 (which absorbed #1447).

## The defect

Three testers hit one skill from two opposite directions.

**Narrower than the objective.** A tester asked to "search for Reuben's parents" and was shown a question specific to the 1900 census. This was the doctrine working as written: most objectives in the corpus are already single-fact ("Identify the parents of X"), while the rule said *never write an objective as a question* — so the only way to comply was to narrow to a record. A question scoped to one census invites a search scoped to one census.

**Wider than the objective.** #1447's reporter: *"The Project Overview is overly broad for the question asked… it kept wanting to restore the relationships and records for wife and child, despite that being outside the parameters."*

Both are the same defect: neither skill treated the objective's scope as a constraint.

## The fix

**question-selection** — new **Step 1c** separating scope from granularity. A multi-fact objective decomposes; an already-single-fact objective *is* the first question, restated with identifying detail. A record is not a question when that question stands in for the objective. Priority 4's own examples were record-scoped and are replaced; Priority 5 now filters the whole-tree pedigree sweep through the objective's scope.

**A follow-up commit (`75616ea3`) carves out the case that first version broke.** Step 1c initially said a single-fact objective *is* the question — restate and stop. On `flynn-unsourced-tree`, where `q_001` already restates the objective, the agent then appended nothing, and `ut_question_selection_008` (the unsound-premise guard) failed after passing four consecutive runs. It also contradicted the skill's own "Sound basis required" rule, since that test's judge context explicitly accepts record-named verification questions. Both Step 1c rules now govern only the question standing in for the objective; once one is open, sub-questions may be narrower and may name a record.

**research-plan** — the FAN mandate was a quota ("at least one search targeting records of relatives"), now a purpose test: an item earns its place by what it reveals about the question's subject.

**init-project / project-status** — out-of-scope pedigree gaps are reported as observations, not a work list; the next step is a plain-language offer that defines *objective* and *research question* on first use, instead of naming a skill at the user. This is the second and third testers' report: *"'Use question-selection to frame the first research question.' I don't know what that means"*, and being told only `"q_001 written."` Step 5 now requires the question quoted verbatim, the id glossed on first use, and the objective-vs-question distinction stated.

## Verification

`check_runlogs.py`: **All runlog rules satisfied.**

| Skill | Result | Log |
|---|---|---|
| question-selection | **13/13** | `v1_2026-08-10_19-26-17` |
| research-plan | 15/16 | `v1_2026-08-10_20-18-41` |
| init-project | 8/9 | `v1_2026-08-10_20-18-41` |
| project-status | 9/11 | `v1_2026-08-10_20-18-41` |

301 annotations across all four (rule 3, zero gaps). Every failure was checked against that test's history rather than assumed:

- `ut_init_project_005` — four prior passes, so this one was investigated properly. The agent omitted `kh_` ids, `created` and `relates_to_person_ids` — fields the agent owns because init-project writes `research.json` directly and never reaches `research_append`'s stamping. A single-test re-run passed with them fully populated: single-run variance, which the corpus accepts with `runs_per_test` pinned to 1.
- `ut_research_plan_001` — mischaracterises `src_006`/`src_008` as retrieved when `log_entry_id` is null. A state-reading error, unrelated to the FAN change. Prior partial on 07-23.
- `ut_project_status_003` — prior partial on 07-19.
- `ut_project_status_008` — already partial in the immediately preceding run (08-01). Pre-existing.

## Two things a reviewer should not take on trust

**1. #1447's out-of-scope direction was never reproduced — in either skill.** `2f309763`'s commit message says both directions "reproduce against the existing job-purnell fixture." That is false, and so is the weaker version I believed until the last test:

- **research-plan** — five attempts. A derived Schuylkill scenario (wife and children added, no active plan) passed *pre-fix*, all 8 plan items in scope. Post-fix passed too: a verified non-regression, not a verified fix. (job-purnell itself carries an *active plan*, so research-plan never reaches the FAN code path there at all.)
- **question-selection** — `ut_question_selection_013` **passes with Step 1c reverted.** Run against pre-fix doctrine, the agent still declined to propose work on Serena Deacon or Samuel Jesse Purnell. So 013 is a characterization test of correct behaviour, *not* a reproduction, and not a guard for a defect anyone has demonstrated in the unit harness.

What *is* solidly established: the Reuben narrowing fix; the Step 1c regression on `ut_question_selection_008`, proven in both directions (failed pre-carve-out, passed after); and the vocabulary changes.

The reproduction failed for reasons worth recording, since they bound what the negative result means: the report says the agent *"kept wanting"* to research the relatives — repeated behaviour across a live multi-turn session, which a single-invocation unit harness cannot show; the scenario had abundant in-scope work (Thomas already linked, 7 assertions, death cert / censuses / probate available), so there was no pressure to drift; and the added relatives had one Birth fact each, not tempting gaps. **Most importantly, the complaint opens with "the *Project Overview* is overly broad" — and Project Overview is init-project's output.** #1447 only ever hypothesised research-plan; nobody established it.

So research-plan's change ships as **hardening on the reporter's testimony**, not as a fix for a demonstrated defect. It is defensible on its own terms — a quota obliges an item even when no FAN record could speak to the question, and a purpose test does not — but a reviewer who wants it dropped has a fair case, and #1447's broader direction would still be covered by question-selection's Priority 5 fix, which *is* reproduced.

**2. Both new rubric dimensions are unexercised, and one is proven so.** `Objective scope containment` (research-plan) and `Objective scope match` (question-selection) scored **3 on every scored test in every run** — including runs against *unmodified* code that the tests were written to catch. Neither has ever fired. CLAUDE.md is explicit that a check which cannot fail "reads as coverage and is worse than no check at all", and I added these two without proving either could fail; that check came last, and it came back negative.

They are kept here only because removing them now would invalidate both skills' snapshots and force a re-run plus ~170 re-annotations. They should get an `/audit-rubric` pass — and a reviewer should treat them as decoration until one is shown to fire.

## Known gap

research-plan has **no committed out-of-scope test**. Reproducing it needs the e2e harness (multi-turn), a scenario where the subject's own leads are thin, and relatives with substantive gaps — none of which the unit corpus offers today. Not filed as a separate issue; recorded here so it isn't mistaken for coverage.


DallanQ (changes requested):
This PR adds a lot of text to SKILL.md files, especially question-selection. Is the additional text needed? Could any of these skills be reverted back to the way they were or shortened?

DallanQ (approved):
thank you for shortening the skills!

mercyokum (commented):
The core fix is right and I don't see a GPS conflict, and I checked, it doesn't break research-plan, search-records, or the orchestrator; none of them depend on the question naming a record.

But 'references/question-formulation.md' and 'references/pedigree-analysis.md' weren't updated, and they still teach the exact thing Step 1c and Priority 5 are fixing, the death-certificate example in one, "candidates for new research questions" with no scope filter in the other. Right now the SKILL.md just tells the reader to override them; please fix the reference files themselves so the skill doesn't ship two contradictory versions of the same rule.

johnmarkpeterbrown (approved):
looks good!

Issue #1478, opened 2026-08-09
Research log grows while assertions and sources stay empty


DallanQ:
**Touches:** packages/engine/mcp-server/src/tools/research-append.ts, packages/engine/plugin/skills/research/SKILL.md

Three bundles in the 2026-08-05..08-07 batch show substantial logged activity and little or nothing persisted as evidence:

| Bundle | log | sources | assertions | questions |
|---|---|---|---|---|
| 2026-08-07T01-40-45-980Z | 21 | 13 | **0** | 0 |
| 2026-08-07T19-40-03-989229Z | 26 | **0** | **0** | 1 |
| 2026-08-06T22-33-27-137239Z | **0** | 0 | 0 | 0 |

The first is the sharpest: **13 sources recorded and zero assertions drawn from them.** The tester's own account of that session says the agent "did get exactly the information on the travel," so the findings existed — they never landed in `research.json`. The third had a full conversation (the tester describes what the agent concluded) and persisted nothing at all.

Worth pinning down before any more skill-prose work: a run that reasons well and persists nothing is indistinguishable, in every downstream artifact, from a run that did nothing. It also starves the viewer, and it leaves `/compare-state` unable to tell success from failure — which is the step the whole feedback workflow turns on.

**First question to answer:** is this the agent not calling `research_append`, or `research_append` accepting the call and not writing? The `results/` sidecars are present for `2026-08-07T01-40-45-980Z` and should settle it.

**Acceptance:** a session that logs research activity and identifies sources persists corresponding assertions, or records why it could not. Establish root cause before proposing a prose fix — per `docs/skill-lifecycle.md` §5 this is lane 1 (tooling) until proven otherwise.

**Evidence:** `feedback-2026-08-07T01-40-45-980Z.zip`, `feedback-2026-08-07T19-40-03-989229Z.zip`, `feedback-2026-08-06T22-33-27-137239Z.zip`

PR #1550, opened 2026-08-11
research_append: warn (not reject) a confident person_evidence link

### LINKED ISSUE #1006: person_evidence links record no match_score in 94% of writes — add an engine invariant (settled: doctrine gap, not an anachronistic check)

jud-sdev:
Closes #1006 

## Summary

- research_append now warns, but does not reject, when a person_evidence link claims confidence: "confident" but records no numeric match_score. same_person returns the 0 to 1 identity score and match_score is the field meant to carry it, yet about 94 percent of historical person_evidence writes leave it unset. Identity is asserted, never scored.
- Warn-only by design: the fault text rides the successful response's validation.warnings and the write still succeeds. A hard reject on day one would break about 94 percent of runs and the hosted path at once, so this follows the same shadow-then-graduate discipline as guardrail-enforcement-spec.md section 7.
- Tier: Normal.

## Review guidance

Start here: personEvidenceScoreWarnings in research-append.ts and its wiring in applyOne. The warning goes into a separate opWarnings accumulator, not invariantErrors, so it rides the response instead of throwing. One call for you, @DallanQ: the issue says to extend personEvidenceInvariants, but that function is the hard-reject path and this has to be warn-only, so the check lives in a sibling function with the same call site and the same confidence gate. If you would rather it sit literally inside personEvidenceInvariants with an errors-plus-warnings return, say so and I will fold it in. It is functionally identical either way.

Unsure about: two things, both yours to settle. Whether to graduate this from warn to reject once the field is populated (the issue defers that to you). And the sibling-versus-extend structure above. Scope is gated on confidence: "confident" rather than "brand-new tree persons" because research_append is a stateless per-call write and cannot see the session's starting tree, so brand-new is not computable here. Confidence is the narrowing it can see, and it is the other scope option the issue lists.

## Plan

Didn't change: the hard-reject invariantErrors path, and the existing epistemic-gate invariant on the same section. This only adds an advisory.

Acceptance check: "warns, but still writes, when a confident link records no match_score" in tests/tools/research-append.test.ts. Fails on main (the helper does not exist), passes here. Two companions confirm a numeric score and a probable link both stay silent.

Deviated from the plan: two departures, both forced and both flagged above. The sibling function instead of literally extending personEvidenceInvariants, because that function is the reject surface and the issue also requires warn-only, which contradict. And the confidence-gate scope instead of "brand-new persons", because the write is stateless. Raised for @DallanQ in Review guidance rather than resolved unilaterally. No schema, auth, or ADR change.

## Test plan

Note on CODEOWNERS: this touches packages/engine/mcp-server/src (.ts), so senior-developers approval is required by branch protection. Peer approval alone will not unblock merge.

- [X] I ran make test-all — not run on this box; its npm prerequisite fails on the local npm version. Ran the affected suites directly instead: npx vitest run in packages/engine/mcp-server is 2108 passed across 88 files, npx tsc --noEmit is clean, and research-append.test.ts is 143 passed including the 3 new tests. Leaning on CI, which runs the same suites.
- [X] For every skill whose run-log snapshot I changed — N/A. Nothing under a skill dir, an agent, eval/tests/unit, or a fixture is touched. research_append is a LIVE_TOOL, but this change only adds a warning to the response and the persisted document is byte-identical, so no eval outcome shifts and no run is owed.
- [X] If I changed a skill's description or DO NOT clauses — N/A. No skill description or DO NOT clause is touched.

## Follow-on issues

Folded in / filed instead: none filed. Graduating this to a hard reject, and measuring the live rate once runs accumulate, both stay on #1006 as its own next step and your decision, not a separate ticket. The measurement additionally needs a build, since research_append is a LIVE_TOOL.


promise-emmanuel (commented):
Review: PR #1550 — 1006-person-evidence-match-score-warn, full pass, 0 commits behind main


[CRITICAL] (9/10) research-append.ts:201-203 — the warning tells the agent to
lower its confidence, and that turns off an existing hard reject.

 You wrote:    `carry the same_person score behind it (a number 0–1). If this link genuinely cannot be ` +
                  `scored against a comparable FamilySearch person, use 'probable' or 'speculative' instead ` +
                  `of 'confident'. The link was still written — this is a warning, not a rejection.`

Change it to: `carry the same_person score behind it (a number 0–1). Call same_person on this pairing and ` +
                  `record its score. If no comparable FamilySearch persona exists to score against, leave ` +
                  `match_score null and keep the confidence your correlation analysis supports — do not lower ` +
                  `it to silence this. The link was still written — this is a warning, not a rejection.`


[INFO] (8/10) research-append.ts:197 — any number silences the warning, including
one outside 0–1. Nothing in the engine enforces the range at runtime.

 You wrote:    if (typeof entry.match_score === "number") return [];

 Change it to: if (typeof entry.match_score === "number" && entry.match_score >= 0 && entry.match_score <= 1) return [];

[INFO] (7/10) docs/specs/research-append-tool-spec.md:830 — the spec says this
check lives in personEvidenceInvariants. It doesn't anymore.

 You wrote:    `personEvidenceInvariants` decided in issue #1006 (2026-08-01). That decision
 Change it to: `personEvidenceScoreWarnings` (warn-only, alongside `personEvidenceInvariants`) decided in issue #1006 (2026-08-01). That decision

promise-emmanuel (commented):
[CRITICAL] (9/10) docs/specs/research-append-tool-spec.md:571 — the new spec
paragraph tells the next reader the opposite of what the code now does.

  You wrote:    and the honest escape hatch, since a link that genuinely cannot be scored should
                claim `probable`/`speculative`. `personEvidenceScoreWarnings` in `research-append.ts`.
  Change it to: and the narrowest claim a stateless write can check. It is **not** an escape hatch:
                a link that genuinely cannot be scored keeps `match_score: null` and the confidence
                its correlation analysis supports (`person-evidence/SKILL.md` §3). Downgrading
                confidence to silence this warning also slips the link past the `confident`-gated
                epistemic reject above. `personEvidenceScoreWarnings` in `research-append.ts`.

  Same phrase in the test name, one line, fix it in the same pass:

  You wrote:    it("does not warn on a probable link with no match_score (the honest escape hatch)", async () => {
  Change it to: it("does not warn on a probable link with no match_score — the gate is confidence, not a downgrade route", async () => {

Ping me once this is done and I'll go ahead and approve your change.

T-FEH (approved):
Approving. @promise-emmanuel's four points are all in, three verbatim — including the important one, that the original warning text told the agent to lower its confidence, which would have routed the link past the `confident`-gated epistemic reject as well. I confirmed those two gates really do share the `confidence !== "confident"` condition, so the doctrine claim in the comment is exact, not decorative.

**I resolved the conflict with `main` and pushed it (`8cb80d49`, `3d7e570c`).** All three files conflicted because `main` landed #1478's `sourcesWithoutAssertionsWarning` in the same places. They are independent additions, not competing edits, so both are kept: the two functions sit side by side with both call sites intact (entry-level for this one, call-level for #1478), the spec keeps this paragraph plus §5.1, and the test file I reconstructed from both parents rather than hand-splicing — the markers had interleaved the two `describe` blocks' shared `beforeEach`/`afterEach`. Verified by delta algebra rather than reading: against `origin/main` the test file is a pure 79-line addition with zero deletions, and against the branch head it is exactly main's delta plus the guards below. Nothing lost or duplicated.

**One thing that would otherwise have failed CI:** `main` renamed `typecheck:dev` → `typecheck` and pointed it at `tsconfig.typecheck.json`, which includes `tests/`. All four #1006 tests read `r.validation`, which does not compile against the `ResearchAppendResult` union's `ok: false` arm. Added the `if (!r.ok) return;` guard this file already uses everywhere else — no assertion changed, and `expect(r.ok).toBe(true)` throws first so the `return` is unreachable on failure.

Verified by running: typecheck clean, engine vitest **2259 passed / 93 files**, `apps/server` **157 passed, 1 skipped**, harness **2075 passed, 2 skipped**. All 12 checks green. Mutation-tested — dropping the confidence gate, accepting any non-null score, returning `[]`, or dropping the numeric exemption each turns a specific test red, so nothing here is a check that cannot fail. The "~94%" figure holds: **93.3%** of 13,856 committed `person_evidence` rows carry no numeric `match_score` (91.1% among `confident` rows). And the `validator.ts` claim is right — it has no `match_score` type or range check, while both schema mirrors do carry `minimum: 0` / `maximum: 1`, which is exactly why the out-of-range test matters.

Non-blocking, none worth another round:

- `research-append.ts:891` — an *unrelated* update to an already-confident unscored row (fixing a rationale typo, say) re-warns. `proof_summaries` guards the equivalent with `tierTouchedThisOp`; `person_evidence` doesn't. Untested and undocumented either way, so worth a line in the spec even if the behaviour stays.
- `person-evidence/SKILL.md:406` says `match_score` is null for FTS/image/PDF-sourced links — a class that will warn permanently and for which "call `same_person`" isn't the remedy.
- `guardrail-enforcement-spec.md` §7 records a `match_score` provenance guard cut in #695 for zero true positives, and `find_person_evidence_missing_same_person` already covers the brand-new-person scope. A cross-reference would stop someone re-deriving that history.

`*.ts` puts this under `senior-developers`, and there are no approvals yet — mine is 1 of 2.

promise-emmanuel (changes requested):
`research-append.ts:261` — the message says "records no match_score" when a
number was recorded. This falls out of the range check from round one: the bound
went in, the message didn't move with it, so `match_score: 5` reports as a
missing score. An agent that just wrote 5 is told it wrote nothing.

    You wrote:    `claims confidence 'confident' but records no match_score. A confident identity should ` +

    Change it to: `claims confidence 'confident' but records no usable match_score (got ` +
                  `${JSON.stringify(entry.match_score)} — expected a number 0–1). A confident identity should ` +

Two test regexes move with it, `research-append.test.ts:3582` and `:3603`

    You wrote:    toMatch(/records no match_score/)
    Change it to: toMatch(/records no usable match_score/)

That is the whole change — typecheck stays clean and that file is 170 passed.

promise-emmanuel (approved):
Changes applied. Merging to main.

PR #1555, opened 2026-08-11
fix(e2e): correct ferber-grandparents f4 - Anna Kramer, b. 1845 Darms…

Paaboat:
...tadt, bump difficulty to Hard

## Summary

- Trivial fix: corrects f4 (maternal grandmother) in the ferber-grandparents fixture — wrong name and birthplace were committed. Mary Kramer (b. Ohio) → Anna Kramer (b. 1845, Darmstadt, Hesse, Germany). Also bumps difficulty from Moderate to Hard.

## Review guidance

Closes #1436. Two files changed: `expected-findings.json` (f4 description, name, birth, supporting source) and `README.md` ("What was removed" line + difficulty rating). The reference project and starting files are unchanged.

**Start here:** `eval/tests/e2e/ferber-grandparents/expected-findings.json` — check f4 block.

**Unsure about:** Nothing — mechanical correction from reference project data.


kofiatinka12 (approved):
The changes update the expected maternal grandmother conclusion from Mary Kramer to Anna Kramer and update the fixture documentation accordingly. The change appears intended to correct the expected answer rather than alter the workflow behavior.

EdmondOware (changes requested):
Thanks for chasing this down. The Anna Kramer identification is clearly better supported than what was there before. The old "Mary Kramer, born Ohio" claim had zero independent corroboration beyond the same source that named her.

Before I approve, I want to flag something #1436 already raised that this PR doesn't quite resolve. The issue is explicit that the Anna Kramer identification is "Probable, not Proved. No single document names both identities together." It poses a direct fork: correct f4 to the flat claim, "or to a hedged/negative form if the fixture maintainer doesn't want to carry the same Probable-tier uncertainty into a benchmark answer." This PR takes the flat option without addressing that the issue explicitly left it open.

That matters for what this fixture tests going forward. Suppose a future agent run correctly hedges its own answer here, something like "likely Anna Kramer, Probable tier, given X/Y/Z." That's the right GPS-conformant behavior. But it would be graded against a required: true benchmark answer that states more certainty than the research it's based on actually supports. We'd effectively be penalizing appropriate epistemic honesty with our own ground truth.

Could you adjust f4's description to carry the tier explicitly? Something like "Anna Kramer was likely the maternal grandmother of Charles Hubert Ferber. Probable, not Proved; no single document names both identities together." That would match the confidence level #1436 itself established. It's a small wording change, not a rework. The identification and sourcing are otherwise solid.

EdmondOware (approved):

PR #1556, opened 2026-08-11
init-project: fetch relatives + sources on the tree import (#1475)

kofiatinka12:
Closes #1475.

## Bug
The first-pass "review the tree / pedigree" summary reported **no sources, no family, few facts** for John Buehner (LC5Z-PX9) — a person with a spouse, 5 children, and ~18 attached sources on FamilySearch. The tester had to prompt 2–3 more times before the family and sources showed up.

## Root cause (from the reporter's own session log)
The initial `person_read` was called **flagless**. `person_read` defaults `relatives=false` and `sourceDescriptions=false`, so it returned only the subject's own facts — `relationships: []`, `sources: []` — and the summary honestly reported "no family, no sources." The reporter's research log records the exact mechanism:

- **log_001** (`person_ancestors`) purpose: *"person_read returned empty relationships/sources for LC5Z-PX9; used person_ancestors to recover the family the user knew was attached in FamilySearch"*
- **log_003** (`person_read`) purpose: *"Re-ran person_read with relatives:true/sourceDescriptions:true (the correct flags — earlier calls omitted them, which is why relationships/sources came back empty every time)."*

I also verified live against LC5Z-PX9: flagless → `relationships: []`, `sources: []`; with the flags → spouse + 5 children + 18 sources. The connection-issue hypothesis the reporter raised is ruled out (the log attributes the empties to the missing flags, and the connection problems came *after* the first summary).

## Fix
`init-project/SKILL.md` — the Step-2 import now calls
`person_read({ personId, relatives: true, sourceDescriptions: true })`, and the prose no longer falsely claims a flagless call returns relatives/sources. That's the only change.

- **project-status and research are unchanged.** project-status reads the local project files and faithfully summarized the sparse tree; the defect was the import that fed it, not the summary.

## On testing — read this before expecting a regression test
**No unit regression is shipped, deliberately, and this is NOT a formally-proven deterministic regression.** I authored a candidate test + flag-sensitive fixtures, ran a pre-edit baseline, and it **did not discriminate**: on the un-fixed skill the model *added the flags on its own* and passed. The failure depends on the model's tool-argument selection, which is **variable** — a unit test can't force the flagless call, so it can't reliably reproduce the bug. The cause and fix rest on the **live session evidence above**, not on a unit reproduction. The candidate test/fixtures/validator were dropped rather than ship a regression the baseline passes.

## Verification
Full `--skill init-project` suite re-run on the fixed skill: **9/9 pass**, fully annotated (62 corrections, every dimension) — `v1_2026-08-11_11-34-29`, committed for the run-log gate. Existing fixtures still match (their `{personId}` predicate is a subset of the flagged call).


Ikennaya1 (approved):
Approved.

Confirmed the diagnosis in the tool: `person_read` defaults both `relatives` and `sourceDescriptions` to `false`, and the previous wording explicitly prevented using those flags. That meant projects imported only the subject person, with no relatives or source descriptions, which is a significant limitation for a tree-survey skill.

The evidence is solid. Five of the nine tests use both flags, and I verified the actual arguments. The annotation also appears genuinely reviewed (62 rows, 30 comments).

Reviewed as a genealogist. The branch is behind main and will need updating before merge.

Nice work!

johnmarkpeterbrown (changes requested):
Hi Solo,
 I'm picking up the senior-genealogist review on PR #1556 (init-project flags fix). The change itself looks good, but the branch is 8 commits behind main right now, and branch protection needs a fresh approval landing after the last push — so I'd like to review it against a current merge rather than a stale base.

Could you update pedigree-summary-underreport-1475 with the latest main and push? Either:

It should merge clean (GitHub reports no conflicts). Once it's pushed and CI goes green on the new head, I'll do the senior review and approve — that clears the last-push-approval requirement.

Thanks!

johnmarkpeterbrown (approved):
Senior reviewed

PR #1560, opened 2026-08-11
fix: tier documented-negative death conclusions instead of "cannot prove or disprove"

mercyokum:
## The issue

A tester asked to locate a death record. The agent answered with a specific year and said "I cannot prove or disprove the search" -- a confident-sounding answer the tester had no way to audit, when the honest answer was that no death record was found and the conclusion should have been tiered on the indirect evidence instead. Three existing proof-conclusion holdout tests that should have caught this were silently skipped: their `holdout: true` flag was placed at the top level of the test JSON instead of inside the `test` object, so schema validation rejected them before they ever ran.

## The fix

- Moved `holdout: true` into the correct location in the three affected test files so they actually run.
- Mined a new test (`ut_proof_conclusion_n9f`) and fixture scenario (`person-a-death-cert-not-indexed`) covering the case: a record type searched and confirmed negative, with independent indirect evidence still available.
- Added a rule to `proof-conclusion/SKILL.md`: a "record not found" result is not itself a conclusion -- the narrative must state what was searched and not found, name the repository and its coverage limitation, and tier on whatever indirect evidence is available. Never write "cannot prove or disprove."
- Clarified `research-exhaustiveness/SKILL.md`: a negative result from `record_search`/`fulltext_search` against a record type confirmed not indexed in that repository is pursued-and-unavailable, not an unsearched gap -- it does not block declaring exhaustive. (An earlier pass at this edit left a malformed, duplicated clause with a dangling parenthesis; a full rerun of the skill's suite caught the resulting regression -- the skill started inventing an "accessible" record to fill the gap -- and this PR carries the corrected, clean wording.)
- Ran both skills' full unit suites against the edited files and annotated the run logs (`eval/runlogs/unit/proof-conclusion/v1_2026-08-11_15-26-52.json`, `eval/runlogs/unit/research-exhaustiveness/v1_2026-08-11_15-44-58.json`).
- Verified live in Cowork against the actual feedback case (Johann Michael Kunberger): the skill now writes an explicit "Documented Absence of a Death Record" section naming every repository searched, tiers the conclusion at Probable, and does not write "cannot prove or disprove."

## Deferred issues found

While gating this fix, the rerun reconfirmed the findings already tracked in #1446 (three proof-conclusion rubric dimensions that never leave a score of 3, and the `ut_proof_conclusion_010` stale `xpass`) -- no new issue filed, just adding this run as corroborating evidence there.

Closes #1474


EdmondOware (approved):
The peer pass above covered the substance: the fix is well targeted at the reported defect, the doctrine wording in both SKILL.md files is clean, the new test and fixture are well constructed, and both run logs are fully annotated with no judge disagreements. The one open question about the misplaced holdout flag is resolved, it was an artifact of a PowerShell JSON round-trip corrupting the file before the final clean commit, not something that ever shipped.

 Also worth noting for the record: this fix is grounded in a live verification against the actual reported feedback case, not just synthetic tests, and the #1446 tie-in avoids filing a duplicate issue. Good work all around.

Paaboat (approved):
Holdout fix—confirmed holdout: true is now correctly nested inside the test object in the three affected files (e.g. proved-tier-with-exhaustive-search.json line 9). The tests were silently skipped before; they will now run.

research-exhaustiveness/SKILL.md — the clarification at lines 121–125 is accurate: a negative result from record_search/fulltext_search against a record type confirmed not indexed in that repository (e.g. SD vital records pre-1940 not on FamilySearch) is pursued-and-unavailable, not an unsearched gap. The dangling "or" is also cleaned up.

New fixture person-a-death-cert-not-indexed—well constructed. The scenario matches the defect exactly: the death cert was searched and not found, indirect evidence (Find a Grave + WPA grave registration) supports the death year, and research declared exhaustive over indexed collections.

No concerns. Good fix for a real gap in the skill's behavior.

EdmondOware (approved):

PR #1562, opened 2026-08-11
harness: a crash now saves the tests that finished, like Ctrl-C already did (#943)

taiwo-stack:
## The problem

`run_tests.py` handled a suite-level abort in two places and treated them oppositely.

The interrupt branch promoted each skill's `.partial_<ts>.json` to `scratch_<ts>.json`, printed the path, and exited 130. The harness-error branch did neither: `if harness_error is not None: return 1` sat thirty-five lines above, returning before the summary, the timing report, the promotion block, the run-log write, and the partial cleanup.

So a crash on test 23 of 24 discarded all 22 finished tests into a gitignored dotfile nothing surfaced — and the only reason the data survived at all is that the cleanup was skipped too. The operator saw one stderr line and exit 1: no totals, no path, and a suite that died on test 1 looked identical to one that died on test 23.

## Measured before writing the fix

The acceptance test was written first and run against unmodified code:

```
rc              = 1
scratch_*.json  = []
.partial_*      = ['.partial_2026-08-11_15-43-17.json']   <- with n_tests: 1 inside
path on stdout  = False
summary printed = False
```

The completed work was on disk the whole time. The early return walked past it.

## The fix

Both paths now share the save-and-report block. Moving the return is safe because nothing between it and the block writes a run log — so a crashed run still cannot mint a releasable `v{N}`.

**The crash wins the exit code when both flags hold.** A harness error stops submission but the drain loop keeps going with no shutdown call, so an operator watching a crashed suite sit there presses Ctrl-C and sets both. "The harness broke" is the more actionable fact, and this preserves what the pre-change code returned for that state.

## The docs deliberately do not match

This is the part worth reviewing. The two exit-code entries describe genuinely different behavior, in the direction that surprises people:

- **On a crash**, submission stops but the drain loop keeps harvesting, so in-flight tests finish and **are** saved.
- **On Ctrl-C**, the pool is joined but nothing reads those futures, so their results **are** discarded.

The old 130 entry claimed in-flight tests were "terminated and discarded". Both halves were wrong — `cancel_futures=True` cannot cancel a running future, and the enclosing `with` joins the pool anyway (#1016 measured ~2 min). `eval/README.md` documented neither 130 nor partial promotion at all.

My own first pass wrote "the same treatment a Ctrl-C gets" into the docstring and "likewise" into the README — the exact equivalence claim this section exists to forbid. `drift-critic` caught it; the third commit removes it.

## Verification

1852 harness tests pass.

The new test mirrors the Ctrl-C one and carries six assertions. **Four failed before this change; two passed on both sides**, and that distinction is load-bearing:

| # | Assertion | Before |
|---|---|---|
| 1 | `rc == 1` | passes — see below |
| 2 | exactly one `scratch_*.json` | fails |
| 3 | its `n_tests == 1` | fails |
| 4 | no leftover `.partial_*` | fails |
| 5 | the path appears on stdout | fails |
| 6 | no `v*.json` | passes — pins "no releasable candidate from a crashed run" |

Assertion 1 is not decoration: put the promotion block *after* the run-log write and the run falls through to `return 0`. It is the assertion that catches the likeliest wrong implementation.

The existing Ctrl-C test passes unchanged.

## Scope

Implements the **2026-08-09 ruling** on #943, not the body beneath it. The ruling cancelled the `ProcessPoolExecutor` rewrite: `BrokenProcessPool` fails every pending future into this same path that wrote nothing, and its spawn-mode workers would add ~680 MB resident on the machines being OOM-killed. No `ProcessPoolExecutor`, no `subprocess`, no `start_new_session`. Retry and abort classification are untouched — this is only about what happens after the loop gives up.

The issue has been retitled per the ruling's suggestion; it previously described the cancelled work.

## Deviated from the plan

- **The Ctrl-C stderr banner was reworded too**, where the plan scoped that commit to "the comment, not the behavior". The banner said "terminating in-flight tests" — the same falsehood the comment carried, so fixing only the comment would have corrected the private half and shipped the public half. No behavior change, no test asserts the string, and it is isolated in the droppable second commit.
- **`eval/README.md` gained a `130` entry** where the plan said "mirror the exit-1 sentence". The file documented neither, so a mirrored sentence would have referenced a Ctrl-C behavior the reader could not look up. Entries reordered to `0, 1, 2, 3, 130`.

Closes #943.


clack391 (approved):

T-FEH (approved):
Approving. Clean implementation of the 2026-08-09 ruling, and the writeup is honest about the one place the docs deliberately don't match.

Verified rather than read:

- **The ruling's boundary holds.** No `ProcessPoolExecutor`, no `subprocess`, no `start_new_session` anywhere in the diff, and the test carries all four mandated assertions.
- **"Nothing between here and there writes a run log"** is true — everything from the deleted `return 1` to the promotion block is read-and-print. `pp.unlink()` (`run_tests.py:1009`) sits *after* that block returns, so it stays unreachable on the crash path.
- **No data can be lost in the promotion.** `promote_partial_to_scratch` is an `os.replace` rename, so the `.partial_` a crash used to orphan becomes the scratch file rather than a second copy.
- **Mutation confirms the test isn't decoration:** restoring the early return fails the new test at `test_cli.py:1423` ("completed tests should be promoted to a scratch log", `0 == 1`) while the Ctrl-C test stays green. Moving the promotion after the run-log write blows up in schema validation, so assertion 6 is load-bearing too.
- **The asymmetry the docs assert is the real behavior** — at concurrency 2 with a worker still running, a crash saves the in-flight result and a Ctrl-C discards it. `eval/README.md` is accurate on both.

Suite on the merged tree: **1916 passed**.

Non-blocking, none worth a round:

- `runlog.py:618` still says the partial is promoted "when a run is interrupted" — now also on a crash.
- With both flags set the summary prints two banners ("interrupted with Ctrl-C" and "stopped by a harness error").
- The 1-beats-130 precedence is explained only in a code comment; the README doesn't mention it.

**On merging:** `clack391`'s approval was submitted against `c287652b`, and `c502da3a` landed after it, so `require_last_push_approval` was unsatisfied even before I touched anything. This approval clears it for the current head. I'm updating from `main` next (strict checks, branch is `BEHIND`) — if that push resets the last-push gate, it needs one more approval from a senior-developer other than me.

PR #1564, opened 2026-08-11
eval-harness: enforce two judge rules the prompt only asks for, and put every override where the judge decides

chrisedeson:
## Summary

- One PR for four issues in `cluster:judge-overrides`. The lead decided to bundle them.
- All four are the same problem. The judge does not always follow its own rules.
- This PR moves two rules from the prompt text into code. It also moves the per-test override next to where the judge reads it.
- Testing found new facts for three of the four issues. It also found a regression this branch caused. The regression is fixed in the last commit.

## Status of each issue

| Issue | Filed as | Current status |
|---|---|---|
| #1406 | Tool Arguments is wrong on zero-tool-call tests | Fixed in code. `_extract_dimensions` sets it to null itself now. |
| #1403 | Judge ignores an override, seen in 2 of 4 runs | Confirmed. 10 of 36 runs failed this way since 07-05. Fixed. The latest run scored 3/3. First clean score since the override text was added. |
| #1401 | Judge invents a rubric dimension | #1421 already fixed the main bug, before these issues were filed. This PR also changes the prompt text. Two of three affected tests stopped inventing, with no edit to those test files. |
| #1402 | Correctness and Log quality re-score the same fault | Not reproducible. #919 already fixed it, on 2026-07-27. This PR adds extra rubric text as a precaution. |

## What changed

- `eval/harness/harness/judge.py`: sets Tool Arguments to null when a run made zero tool calls. Records a warning instead of a silent change.
- `eval/harness/judge/prompt.md`: moved the per-test override text to sit right before "How to report". Added a short section that states which rule wins when two rules conflict.
- `eval/harness/run_tests.py`: prints a count of judge rule violations at the end of a run.
- Three test files changed. `ut_search_records_005`'s override now names both dimensions it binds. `ut_research_plan_007`'s notes are now plain text, not a bulleted checklist. `search-records/rubric.md` gets one more paragraph. It says Correctness and Completeness must not re-score what Result triage or Nil escalation already scored.

## Regression found and fixed

The first eval run found a bug this branch caused. Three negative tests dropped from a correct score of 3 to 1. Cause: the prompt change moved one rule but not a second rule that depends on it. The judge then read them in the wrong order. Fix: moved the second rule too. Two of the three tests recovered to 3 in the next run. The third test had a separate, unrelated problem in its own wording. That one is fixed by annotation, not code. See Test results below.

## Test results

Two skills, 40 tests, one eval run. 39 passed, 1 aborted on a network stall.

- #1403 test scored 3/3. First clean score since 2026-07-05.
- #1401: no invented dimension in either run log.
- #1402 and #1406: both hold.
- The regression above: confirmed fixed in this run.

Annotation is done for both run logs: 152 entries for search-records, 85 for research-plan. Three real corrections were made, not blind agreement.

- `ut_search_records_003`: Correctness and Completeness, 1 to 3. The judge misread the test's own instructions.
- `ut_search_records_014`: Tool Arguments, 2 to 3. The judge's own text agreed the score should be 3, then scored 2 anyway.

## Review guidance

**Start here:** the code in `_extract_dimensions` that sets Tool Arguments to null. Also the line in `render_prompt_parts` that splits the prompt for caching.

**Check this:** the null fix can turn a failed test into a passed test. This happens in two places in `_compute_outcome`. I think this is correct. A run with no tool calls has nothing to check. Tests cover both places. Tell me if you disagree.

## Test plan

- [x] `make test-all` passes. One known failure is not caused by this branch: an electron test that times out under load (#1366). This PR does not touch that code.
- [x] `make eval-skill SKILL="search-records research-plan"` was run and annotated. Both CI rules pass.
- [x] N/A. No skill description or DO-NOT text was changed.

Every new test was checked by breaking the code on purpose, confirming the test failed, then fixing it back.

## Follow-on issues

None filed. All planned work is in this PR.


Asimi1234 (changes requested):
**Reviewed** and verified locally (test-merged with main, suite 1907 pass, both new test groups break-checked) — the code is sound, but merge is blocked by a real run_tests.py conflict where main's _print_outcome_counts/reason collides with your _print_judge_warnings/judge_warning_kinds and **must be resolved**, the one failing test is a pre-existing date-windowed one that also fails on clean main, and the deliberate Tool Arguments N/A fail→pass coercion needs explicit senior-genealogist sign-off before both required teams approve.

johnmarkpeterbrown (approved):
looks good!

Asimi1234 (approved):
The code is sound and thoroughly tested.

PR #1504, opened 2026-08-09
e2e: re-adjudicate chresten-nielsen-daughter under #1346; re-run reve…

EdmondOware:
…als genuine non-recovery

Re-adjudicated the chresten-nielsen-daughter fixture per #1346: the previously-cited christening ark (W6M7-SWMM) was confirmed fabricated (an unrelated US WWI draft card). Removed it; birth date/place (3 Aug 1805, Dallerup) now cited to the original 1805 Dallerup parish register (DGS 007226375, image 00232 — not full-text indexed, no resolvable 3:1: ark exists). Confirmation record (QG8V-2DDT) retained as independent corroboration and literal-ark anchor.

Live Cowork debug (skill-lifecycle.md step 7) closed a mentor-flagged identity gap: a genuine second Chresten Nielsen/Sidsel Lausdatter couple exists in Tyrsted, correctly excluded on Birte's own record (mother = Birte Kirstine, village = Dallerup). Recorded as h_001/h_002 with the Niels 1807 record extracted as its own source for a symmetric audit trail. An OCR non-determinism on image 00232 was caught and corrected mid-session rather than left standing.

Real e2e run (run-2026-08-09_18-05-03) reproduced the guardrail bypass (research-exhaustiveness/proof-conclusion skipped) already tracked separately by Dallan — not re-diagnosed here. Independently, this run also missed the finding entirely: the 1819 confirmation page fell just outside its browse range and the Apr–Oct 1805 baptism section was OCR-illegible, so it correctly reached a probable-negative conclusion with no fabricated-citation contamination (verified: neither W6M7-SWMM nor any of the corrected citations were fetched or cited this run). Graded per skill-lifecycle.md: f1 = false (genuine miss), proof quality 3/sound

## Summary

- Re-adjudicated chresten-nielsen-daughter under #1346: removed the fabricated
  christening ark (W6M7-SWMM, actually an unrelated WWI draft card) and dropped
  the specific 6 Oct 1805 christening claim it alone supplied; birth date/place
  now cited to the original register image (DGS 007226375, image 00232).
- Live Cowork debug closed a mentor-flagged identity gap (a genuine second
  Chresten Nielsen/Sidsel Lausdatter couple, correctly excluded) and caught an
  OCR non-determinism on image 00232 before it could ship as a citation.
- Real e2e run reproduced the research-exhaustiveness/proof-conclusion
  guardrail bypass already tracked separately (not re-diagnosed here), and
[...]

PR #1566, opened 2026-08-11
Accept positive feedback reports via a "worked as expected" radio (#1480)

MMagnum:
## Summary
Tier: **Normal**.
- The feedback dialog rejected a positive report unless "what it should have done" was filled, so testers typed filler ("N/A") that triage then read as content (#1480).
- Adds a required Yes/No **"Did it work as expected?"** radio, persisted as `worked_as_expected` in `feedback.json`. The two "what went wrong" fields hide unless the user picks **No**, and both become optional; `agent_did` stays required.
- Triage reads the boolean (`feedback.json` + a prominent FEEDBACK.md line) to tell a positive report from a problem; `compare-state` and the alpha-feedback guide no longer treat an empty `agent_should_have` as malformed. `schema_version` stays 1.

## Review guidance
**Start here:** `packages/viewer-ui/src/components/shared/FeedbackDialog.tsx` — the `workedAsExpected` state, `canSend`, and the `=== false` conditional render + `markWorked` clear-on-Yes. That's where the behavior lives; the rest is threading one boolean end-to-end through both bundlers. The non-obvious call: `agent_should_have` keeps `required: yes` in the spec table but is now documented as legitimately empty (mirroring `correct_answer`), which is what lets us skip a `schema_version` bump.

**Unsure about:** Nothing blocking. The one judgment call — hide the bug-detail fields until "No" (and hide `correct_answer` too) — was agreed with the lead before building.

## Plan
**Didn't change:** `schema_version` / `FEEDBACK_SCHEMA_VERSION` (stay 1). Per spec §5 a new field old consumers ignore doesn't bump, and empty `agent_should_have` mirrors `correct_answer` (already present-but-may-be-empty) so it isn't a semantic change. Bumping would also break `/compare-state`, which hard-requires `== 1`.

**Acceptance check:** `packages/viewer-ui/src/components/shared/__tests__/FeedbackDialog.test.tsx` — the dialog-level test where the fix actually lives: Send is gated on the radio, the bug-detail fields reveal only on No, and switching to Yes clears typed bug text so a positive report carries none. Verified it goes **red** if the gate or a `markWorked` clear is reverted. Plus `feedback.test.ts` (boolean → JSON round-trip) and `test_feedback.py` (web path asserts `worked_as_expected is True`). All fail on `main`, pass here.

**Deviated from the plan:** Nothing — plan-critic (x2) and drift-critic all clean.

## Test plan
- [ ] `make test-all` — **not run in full on Windows** (uv trampoline #1349 blocks `server-test`; `gen-enums-guards` fails Windows-only #1370). Ran the relevant subsets: `make typecheck` (5 pkgs clean), `make test-js` (Electron feedback 62 pass incl. new test; viewer-ui 178 pass — only the unrelated #1370 file fails to collect), and `test_feedback.py` (12 pass) via `uv run python -m pytest`. Leaning on CI for the full suite.
- **N/A — eval-skill:** `compare-state` has no `eval/tests/unit/` or run-log snapshot (triage workflow skill, not a graded research skill); its change is covered by the feedback unit tests.
- **N/A** — no skill `description`/DO-NOT frontmatter changed.

**Senior review required:** yes — touches `.ts`/`.py`/`.json` files (per `.github/CODEOWNERS`).

## Follow-on issues
**Folded in / filed instead:** None filed. Fixed the pre-existing missing `correctAnswer` useCallback dep in `FeedbackDialog.tsx` in passing (one line, same file). No deferrals.

Closes #1480.


chesworthrm (changes requested):
Nice fix, and the right shape for it — one boolean threaded end-to-end, no `schema_version` churn. I checked the version call and agree with it: a brand-new field isn't on §5's bump list, and a new consumer reading an old zip defaults to `false`, which reads as "problem". That's the safe direction, since every zip submitted before this is a bug report. Folding in the `correctAnswer` dependency fix was also the right call — that was a real stale-value bug.

Three things to change before this merges.

## 1. Nothing tests the part of this that matters

Your acceptance check proves a boolean handed to the zip builder reaches the JSON. It does not touch the dialog, and the dialog is where the fix lives.

I checked what that leaves uncovered. If you put the original bug straight back — make Send require "what it should have done" again — and also delete both clearing lines from `markWorked`, every test in the repo still passes, and so does typecheck.

The clearing is the part I'd most want watched. It stops a positive report carrying leftover bug text. If it ever goes away, triage starts reading that leftover text as content, which is the exact problem this PR was filed about.

There's no test file for this dialog yet, so here's one. It passes on your branch, and it fails if either of those two behaviors is reverted. Add it as `packages/viewer-ui/src/components/shared/__tests__/FeedbackDialog.test.tsx`:

```tsx
import { describe, it, expect, vi, beforeEach } from 'vitest'
import { render, screen } from '@testing-library/react'
import userEvent from '@testing-library/user-event'
import FeedbackDialog from '../FeedbackDialog'
import { buildMockContext } from '../../../contexts/__tests__/mockContext'

vi.mock('../../../contexts/ResearchDataContext', async () => {
  const actual = await vi.importActual<typeof import('../../../contexts/ResearchDataContext')>(
    '../../../contexts/ResearchDataContext'
  )
  return { ...actual, useResearchData: vi.fn() }
})

import { useResearchData } from '../../../contexts/ResearchDataContext'

function mount(): { submitFeedback: ReturnType<typeof vi.fn> } {
  const submitFeedback = vi.fn(async () => ({ ok: true as const }))
  vi.mocked(useResearchData).mockReturnValue({ ...buildMockContext(), submitFeedback })
  render(<FeedbackDialog onClose={() => {}} />)
  return { submitFeedback }
}

async function fillRequired(user: ReturnType<typeof userEvent.setup>): Promise<void> {
  await user.clear(screen.getByLabelText('Your email'))
  await user.type(screen.getByLabelText('Your email'), 'a@b.com')
  await user.type(screen.getByLabelText('What you asked the agent to do'), 'q')
  await user.type(screen.getByLabelText('What the agent did'), 'd')
}

describe('FeedbackDialog — worked-as-expected gate', () => {
  beforeEach(() => {
    localStorage.clear()
  })

  it('blocks Send until the radio is answered, then allows it with no bug detail', async () => {
    const user = userEvent.setup()
    mount()
    await fillRequired(user)
    expect(screen.getByRole('button', { name: 'Send' })).toBeDisabled()
    await user.click(screen.getByRole('radio', { name: 'Yes' }))
    expect(screen.getByRole('button', { name: 'Send' })).toBeEnabled()
  })

  it('reveals the bug-detail fields only on No', async () => {
    const user = userEvent.setup()
    mount()
    expect(screen.queryByLabelText(/What it should have done/)).toBeNull()
    await user.click(screen.getByRole('radio', { name: 'No' }))
    expect(screen.getByLabelText(/What it should have done/)).toBeTruthy()
    await user.click(screen.getByRole('radio', { name: 'Yes' }))
    expect(screen.queryByLabelText(/What it should have done/)).toBeNull()
  })

  it('does not carry stale bug text into a positive report', async () => {
    const user = userEvent.setup()
    const { submitFeedback } = mount()
    await fillRequired(user)
    await user.click(screen.getByRole('radio', { name: 'No' }))
    await user.type(screen.getByLabelText(/What it should have done/), 'it should have searched 1870')
    await user.type(screen.getByLabelText(/what is the correct answer/), 'father was Robert')
    await user.click(screen.getByRole('radio', { name: 'Yes' }))
    await user.click(screen.getByRole('button', { name: 'Send' }))
    expect(submitFeedback).toHaveBeenCalledTimes(1)
    const payload = submitFeedback.mock.calls[0][0] as Record<string, unknown>
    expect(payload.workedAsExpected).toBe(true)
    expect(payload.agentShouldHave).toBe('')
    expect(payload.correctAnswer).toBeUndefined()
  })
})
```

Then point your **Acceptance check** line at this file too, since it's the check that actually covers the change.

## 2. The guides testers read still describe the old form

You added a required field. The two documents that walk a tester through this form box by box weren't updated. That matters more than usual here — the bug you're fixing was a tester confused by this form.

In `docs/alpha-user-guide.md`:

You wrote:

```
1. **What you asked the agent to do.**
2. **What the agent did.** What actually happened.
3. **What it should have done.** This is the one that turns a complaint into a
   fix.
4. **If it reached a wrong conclusion: the correct answer and its evidence.**
   Optional, and only relevant when the *answer* was wrong rather than the
   method. Fill it in and we can build a test from your case without coming back
   to ask you. Leave it blank when the problem was how it worked, not what it
   concluded.

There's also your email, and a free-text **Notes** box for anything that
doesn't fit the four.
```

Change it to:

```
1. **What you asked the agent to do.**
2. **What the agent did.** What actually happened.
3. **Did it work as expected? Yes or No.** Required. Answer **Yes** and you're
   done — a report about something that went well is worth sending, and there is
   nothing more to fill in.
4. **What it should have done.** Appears when you answer **No**, and it's
   optional — leave it blank if you aren't sure what the right behavior was.
5. **If it reached a wrong conclusion: the correct answer and its evidence.**
   Also only on **No**, and also optional. Only relevant when the *answer* was
   wrong rather than the method. Fill it in and we can build a test from your
   case without coming back to ask you. Leave it blank when the problem was how
   it worked, not what it concluded.

There's also your email, and a free-text **Notes** box for anything that
doesn't fit the boxes above.
```

`docs/alpha-user-guide-cowork.md` has the same list, including the same "doesn't fit the four" sentence. Apply the identical replacement there.

In `docs/alpha-feedback-guide.md`:

You wrote:

```
session, and this note is one of four boxes in that form — the zip is
```

Change it to:

```
session, and this note fills that form's text boxes — she answers **No** to
"Did it work as expected?", which is what reveals the last two — and the zip is
```

You wrote:

```
The same four boxes, the same submission path, the same zip shape land in the
```

Change it to:

```
The same boxes, the same submission path, the same zip shape land in the
```

## 3. A positive report has no way out of the triage queue

Every submission opens an issue in Ready, and Steps 3 through 9 of the triage guide are reproduce, fix, reply. Step 4 is the one place that asks whether the problem is even ours, and its three options are skill, tool, and grading. None of them is "nothing is wrong."

So the first positive report someone picks up sends them into Step 3 hunting a bug that doesn't exist. Your summary says triage reads the boolean to tell the two apart. The flag is in the bundle now, but the step that decides never looks at it.

Add a first bullet to Step 4's list in `docs/alpha-feedback-guide.md`, right after "Using the `results/` files — the real tool responses — check:":

```
- Does `_feedback/feedback.json` say `"worked_as_expected": true` (in FEEDBACK.md,
  **Worked as expected: Yes**)? → **nothing went wrong.** It's a positive report.
  Note anything useful from it on the issue, close the issue, and stop — there is
  no fix to make and no test to mine. ✅
```

---

Two things I looked at and am not asking you to change: the radio group uses a `span` for its label rather than a `fieldset` and `legend`, and the zip builder has no runtime fallback for the new field the way the server does. TypeScript covers every caller in the repo, so neither blocks this.

One housekeeping note — the branch is 11 commits behind `main`. I merged `main` in locally before running anything and it's clean, with every suite passing, so there's no conflict waiting for you. You'll just need to update the branch before it can merge.

chesworthrm (changes requested):
All three points from the last round are properly addressed — thanks for the fast turnaround.

I re-checked the new dialog test rather than taking it on trust: reverting the Send gate to the old `agentShouldHave` requirement fails two of the three cases, and deleting the two clearing lines from `markWorked` fails the stale-text case. It guards both behaviors. Your `localStorage` adaptation is the right fix for that env, and pointing the Acceptance check at the file was the missing half. The guide replacements and the Step 4 close-it bullet all landed as written.

One thing to change, and then I think this is done.

## The other consumer of `agent_should_have` still assumes it's filled in

You made the field legitimately empty and updated `compare-state` for it. The spec row you edited names two consumers, and `/mine-unit-test` is the second one. Its decision-rules table tells it to read the Should from the bundle *instead of* asking the user, with no case for the field being blank — and the Should is what becomes the test's rubric.

This case only exists because of this PR, and your new placeholder invites it: "leave blank if you're not sure". A tester hits a real bug, answers No, leaves the box empty. Step 4 of the triage guide correctly lets that through as a genuine problem, and Step 5 then runs `/mine-unit-test` on a case with nothing to build a rubric from.

In `.claude/skills/mine-unit-test/SKILL.md`:

You wrote:

```
| Run inside a feedback-case dir (a `.feedback-repo-root` is present) | Fine — mine it as `--project <case-dir>`. Resolve the repo from that marker rather than assuming cwd, and read `_feedback/feedback.json` for the submitter's Did/Should (and `correct_answer`, when they supplied ground truth) instead of interviewing the user for it. Mine it *before* the fix lands, so the gate has a pre-edit baseline. |
```

Change it to:

```
| Run inside a feedback-case dir (a `.feedback-repo-root` is present) | Fine — mine it as `--project <case-dir>`. Resolve the repo from that marker rather than assuming cwd, and read `_feedback/feedback.json` for the submitter's Did/Should (and `correct_answer`, when they supplied ground truth) instead of interviewing the user for it. **`agent_should_have` may legitimately be empty** — the reporter hit a bug but did not know the ideal behavior — so fall back to asking the user for the Should rather than mining a rubric from a blank. Mine it *before* the fix lands, so the gate has a pre-edit baseline. |
```

## Not yours to fix, but it's blocking

`lockfile-drift` is red on this head and it has nothing to do with your diff — the job died reaching `api.github.com` with a TLS certificate error, before it ran any drift logic. Re-run that job. The other required checks are green.

You'll also need a second approval from `senior-developers`; the `.ts`/`.tsx`/`.py` rules in CODEOWNERS catch this diff.

---

Everything else came back clean. The empty Should passes the Electron validator, the new boolean never reaches the server's `_norm`, and no transport drops it — the web path spreads the whole payload and the Electron preload forwards the object. Both markdown renderers omit the heading on the same condition.

I did not exercise this in a real Electron window or the hosted web app — the dialog behavior is covered only through jsdom.

chesworthrm (approved):
The `mine-unit-test` row landed as written — that was the last thing. Approving.

I checked this against the merge with current `main` rather than just your head, since the branch had fallen 5 behind while #1504 went in. The merge is clean and everything passes on it: viewer-ui 190, Electron 62, `apps/server` 157, `eval/harness` 1932. `lockfile-drift` is green now too, which confirms the earlier red was the TLS flake and not real drift.

Worth noting for whoever merges: those 5 commits are still ahead of you, so the branch needs one more update before the button goes green. Nothing in them conflicts.

Good work across all three rounds — you turned each one around with the falsification actually run rather than assumed, and you caught the placeholder-invites-the-empty-case connection on the last one yourself.

T-FEH (approved):
Approving. @chesworthrm's three rounds did the heavy lifting and every point landed in the diff verbatim — the dialog test file, both tester guides, the Step 4 bullet, the `mine-unit-test` row. Nothing waved through.

What I checked independently:

**The contract holds across all four runtimes.** dialog → `transport.ts` → preload bridge → Electron `index.ts` → the zip JSON/MD, and web spread → `FeedbackBody` → the server. No leg drops or renames it, and it respects the casing seam: `workedAsExpected` on every wire surface, `worked_as_expected` in the persisted zip.

**Backward compatibility works in both directions.** Old zip into the new server: pydantic ignores extras, no 422. New client against an old server: the field defaults `false`, which reads as "problem" — the safe direction, since every bundle submitted before this was a bug report.

**The new tests can all fail.** Reverting `canSend` to the old `agentShouldHave` gate fails 2 of 3 dialog tests; deleting `markWorked`'s two clearing lines fails the stale-text test; hardcoding `worked_as_expected: false` in `feedback.ts` and again in `feedback.py` fails their respective tests. The clearing behaviour @chesworthrm most wanted watched is genuinely pinned now.

**Security invariants: 10/10.** This touches `src/main/` and `src/preload/`, so I ran `apps/electron`'s checklist. The delta is security-neutral — one boolean added to the existing `feedback:submit` payload, no new IPC channel (all 14 in the preload still match the authorized table), no new dependency, no CSP or navigation change. The one place the value reaches output is a ternary emitting the literal `Yes`/`No`, so unlike the free-text fields it cannot carry anything through.

Suites: viewer-ui vitest **190 passed / 24 files**, `apps/server` pytest **157 passed, 1 skipped**, typecheck **5/5**. `apps/electron` vitest is 61 passed / 1 failed, and that failure is `feedback.test.ts:167` ("drops the largest files…") — untouched by this diff, a 45 MB zip exceeding the 5 s default timeout; 20/20 with `--testTimeout=30000`. Machine speed, not a regression.

Two non-blocking notes, neither worth another round:

- `docs/alpha-feedback-guide.md` puts the positive-report exit in Step 4, but Step 3 is a full live agent re-run plus `/compare-state`. A triager burns that run before learning there was nothing to reproduce. Cheapest fix is a one-line check of `worked_as_expected` at the top of Step 3.
- `feedback-json-spec.md` marks the field `required: yes` under `schema_version` 1, but bundles submitted before this shipped are also v1 and lack the key. Worth saying a consumer that finds it missing must read it as `false`.

Issue #1569, opened 2026-08-12
skill_invocation.py: two guardrail-detector corrections sharing one corpus replay — lane check should count attempts, proof-conclusion arm misses writes into existing relationships


aghadiayeamayanvboernest:
**Touches:** eval/harness/harness/skill_invocation.py,
eval/harness/tests/unit/test_skill_invocation.py,
docs/specs/guardrail-enforcement-spec.md, docs/specs/e2e-test-spec.md

> **Issue #1368 merged in here, 2026-08-14.** Two corrections to two different detectors
> in `skill_invocation.py`, with the same test file, the same spec, and — the part that
> actually costs — **the same corpus replay over `eval/runlogs/e2e/**`, for which there
> is no make target.** Whoever builds that instrument for one has it for the other.
> Neither is blocked, both are `developer`, neither touches a skill dir so there is no
> paid eval run either way.
>
> They are independent edits: part 1 is a one-line skip removal, part 2 is a normalised
> key and explicitly not a one-liner. **Part 2 may end in "close with the limitation
> documented" and part 1 must land regardless** — do not let part 2's outcome hold part 1.

> **Gate satisfied.** This issue was filed "do after PR #1570 merges." PR #1570 merged
> 2026-08-13T20:22:42Z, so the premise holds and the narrowing described in part 1 is
> live.

Two things this issue used to carry are done and are not part of the ask:
PR #1570 bumped `HARNESS_SCHEMA_VERSION` to 4 and rewrote every passage that
asserted the returned-`{ok:false}` defect was still live, and the
skill-launch gating half was settled in issue #1463 (completed) with its
instrument refused in issue #1298 (not planned) — do not re-measure either.

---

# Part 1 — the lane check should read attempts, not effects

`find_protected_writes_by_unnamed_delegate`
(`eval/harness/harness/skill_invocation.py:638`) skips any entry with
`is_error: true` at `:757`. Once a returned `{ok:false}` sets that key, an
`extraction_append` made by a `general-purpose` delegate stops being flagged
whenever the write also failed validation — but calling a tool you are not
entitled to hold is a violation whether or not the call succeeded, and the
rule this detector implements is stated about the *caller*, not the outcome:
"a protected write is legitimate iff its caller is the main thread or one of
the four dedicated agents" (`docs/specs/guardrail-enforcement-spec.md:651`).

**Size of the narrowing.** Every lane violation in the committed corpus — 15
of them — sits in one run,
[...]

PR #1255, opened 2026-08-04
eval-harness: populate is_error on e2e tool_calls so the success-gates discriminate

clack391:
## Summary

Tier: Normal.

- Populate `is_error` on each e2e `tool_calls` entry when its `ToolResultBlock`
  arrives, so `skill_invocation.py`'s five success-gates (which skip an entry on
  `entry.get("is_error") is True`) actually discriminate a failed Skill or tool
  call from a successful one.
- The key was never set on a tool call, so every gate treated a failed Skill
  call as a successful invocation, and the "invoke the skill, let it fail, finish
  the write inline" evasion was unblocked (guardrail-enforcement-spec.md §7).
  Prospective correctness: the committed corpus has no soft-failure case this
  would have flipped (issue #999's measurement).
- Concrete consequence (folding in the absorbed duplicate #1232): an agent that
  calls `same_person` with bad arguments, gets an error back, then writes the
  `person_evidence` link currently satisfies
  `find_person_evidence_missing_same_person` (a hard-fail gate that forces
  `outcome: fail` regardless of the judge). So today that gate credits an
  identity that was never scored; once `is_error` is populated, the gate sees the
  `same_person` call failed and fires correctly.

## Review guidance

**Start here:** `eval/harness/e2e/orchestrator.py`, the new
`apply_tool_result(entry, block, summary)` helper and its one call site in the
`ToolResultBlock` branch. The one decision worth checking is
`entry["is_error"] = block.is_error is True`: the SDK leaves `is_error` as `None`
(not `False`) on success, and all five gates match on `is True`, so `is True`
normalizes it to a clean bool. Everything else is the test and a comment.

**Unsure about:** Nothing blocking. One inherent limitation, not a defect: the
unit test proves the helper, but nothing unit-tests that `_run_agent` actually
calls it. That wiring lives in the async SDK message loop the test file's own
docstring says it does not cover, so a live `make e2e-run` is the only
end-to-end proof (the issue and architecture.md §9.4 say the same).

## Plan

**Didn't change:** `skill_invocation.py`'s five gates (already correct on
`is True`; this is a producer-only fix). `HARNESS_SCHEMA_VERSION` stays 1
(additive; every consumer reads `.get("is_error") is True`, identical for absent
and `False`, so old logs still read as not-errored). The fabricated
`_skill_call` / `_mcp_call` test helpers (they test the consumer). No schema
(the e2e result is `asdict` + `json.dumps`, not jsonschema-validated; the unit
[...]

Issue #988, opened 2026-07-30
Six private copies of readJson / formatIssues across the writer tools


DallanQ:
> **Reviewed 2026-08-17 before junior handoff.** Canonical home settled by the lead
> the same day; the counts, the destination and the sequencing in the original ask
> were each wrong, so this body replaces it. The original follows under
> "Original issue".

**Touches:** packages/engine/mcp-server/src/tools/research-append.ts, packages/engine/mcp-server/src/tools/tree-edit.ts, packages/engine/mcp-server/src/tools/materialize-facts.ts, packages/engine/mcp-server/src/tools/research-log-append.ts, packages/engine/mcp-server/src/tools/project-context.ts, packages/engine/mcp-server/src/tools/research-query.ts, packages/engine/mcp-server/src/tools/merge-shared.ts, packages/engine/mcp-server/src/utils/project-io.ts

Six tools read a project JSON file through their own copy of the same
read-and-throw, and four carry their own copy of `formatIssues`. There are also
**two shared implementations**, which is why this has sat.

Reproduce the inventory:

    git grep -n "function readJson\|async function readProjectJson\|function formatIssues" -- packages/engine/mcp-server/src

### The six read sites

| File | Shape | Throws |
|---|---|---|
| `research-append.ts` | private `readJson` | `ResearchAppendError` |
| `tree-edit.ts` | private `readJson` | `TreeEditError` |
| `materialize-facts.ts` | private `readJson` | `MaterializeFactsError` |
| `research-log-append.ts` | private helper **named `readProjectJson`** — a grep for `readJson` misses it | `LogAppendError` |
| `project-context.ts` | private `readJson` | plain `Error` |
| `research-query.ts` | **no helper to delete** — the read is inlined at the call site | `ResearchQueryError` |

Private `formatIssues` copies: `research-append.ts`, `tree-edit.ts`,
`materialize-facts.ts`, `research-log-append.ts`.

### Not copies — leave these alone

- `tree-forget.ts` — a three-line wrapper over the shared `readProjectJson`.
  This is the pattern to follow, not a site to change.
- `rank-search-matches.ts` — a deliberate best-effort read inside a `try` that
  swallows every failure ("fall back to the bare tree person rather than failing
  the whole ranking call"). Different semantics. Do not migrate it.
- `merge-tree-persons.ts`, `project-create.ts`, and `research-append.ts`'s
[...]

PR #1570, opened 2026-08-12
mcp: mark returned `{ok:false}` as `isError`, in dispatch and in the harness (#1282)

aghadiayeamayanvboernest:
mcp: mark returned `{ok:false}` as `isError`, in dispatch and in the harness (#1282)

## Summary

- The dispatch set `isError` only from its `catch` arms, so a tool that signals failure by **returning** `{ok:false}` emitted a normal MCP success — a rejected write read as a *successful call* to the model in Cowork and the hosted path, and to the eval harness's guardrail detectors. Roughly **1 in 7** `research_append` calls and **1 in 4** `extraction_append` calls were invisible that way.
- New `src/tool-result.ts` exports `writerToolResult` + `OK_FALSE_IS_FAILURE`; **11** dispatch arms route through it. Mirrored in `mock_mcp.py`, which bypasses the dispatch entirely — without it a failed write reads as an error in production and a **success in every unit eval run**.
- Tier: **Normal**. Closes #1282.

## Review guidance

**Start here:** `tests/packaging/writer-tool-results.test.ts`. It pins the *wiring*, not just the helper — no other file under `tests/` imports `src/index.ts`, so adding the helper and wiring 3 of 11 arms passes `tsc`, the unit test, and the manifest drift test while eight tools keep reporting failures as successes. It deliberately does **not** copy `manifest.test.ts`'s collector verbatim: that one keeps only `node.right.text` and discards the node, which would make this assert merely that the names are dispatched — already true on `main`. Keeping the enclosing `IfStatement` is what makes it assert on the arm's body. Second thing worth a look: the harness gate is `tool_name ∈ set` **AND** `ok is False` — name alone flags successes, `ok` alone flips `merge_warnings`, which shares the compiled-tool builder.

**Unsure about:** nothing structural. One judgment call: `writerToolResult` leaves `isError` *unset* (not `false`) on success, so a successful envelope is byte-identical to before — flag if you'd rather it be explicit.

## Plan

**Didn't change:** `merge_warnings` (dispatch *and* harness) — its `{ok:false}` is a dry-run verdict about a merge, not the tool failing, and it's the highest-`ok:false` tool in the corpus, so flagging it would tell the agent a working preview had crashed. `validate_research_schema` never enters — it answers with `valid`, not `ok`. `skill_invocation.py` is untouched (see below). The writer tool bodies are #988's, no overlap.

**Acceptance check:** `tests/packaging/writer-tool-results.test.ts` → *"routes every OK_FALSE_IS_FAILURE tool through writerToolResult"*. Fails on `main` two ways (module absent; no arm calls it). **Mutation-proven in both directions** — reverting `extraction_append`'s arm fails naming it, and wrongly routing `merge_warnings` fails naming it. Same on the harness side: gating on `ok` alone fires the exclusion test, and dropping a tool from the mirror fires both its test and the drift lint.

**Deviated from the plan:** nothing structural, but `/critique-plan` ran two rounds and caught three real defects before any code: my harness set wrongly grouped `project_context`/`research_query` with the *excluded* `merge_warnings` (2 of 11 would have stayed invisible), I'd named 2 no-mirror tools instead of 3, and the acceptance check as first specified would have been **vacuous**. All fixed in the plan before implementing.

## Scoped out — filed as #1569, not missed

Round 2 surfaced that `is_error`'s **meaning** changes for every future e2e run log (it now means "threw **or** returned `{ok:false}`"), which `result.py`'s own rule calls a `HARNESS_SCHEMA_VERSION` boundary — and §7.5 records that this exact field already hit this once (#1255 left it at 2; 3 closed it). The issue is silent on versioning and its doctrine is that consequences landing on `guardrail-enforcement-spec.md` are *"filed as its own issue — do not fix it here."* So #1569 carries: the **lane-check narrowing** (mandated split — live today, since #1255's guards bind on real data, so failed `extraction_append` calls stop reaching the lane check), the **Skill-launch gating half** (1,241 of 1,242 corpus `Skill` results are a launch acknowledgement with `is_error` never True), and the **schema-boundary decision** for the lead.

## Test plan

- [x] Ran `make test-all`. Engine vitest **2157**, harness pytest **1907**, eval-app **148**, control-plane, `tsc` and `ruff` all clean. The one red is the known **#1366** `apps/electron` flake (`buildFeedbackZip > drops the largest files` times out at 5s under parallel load; **61/61 in isolation**, that test in 2.2s) — this branch touches **zero** files under `apps/electron`.
- [x] N/A — no skill run-log snapshot changed. `build_snapshot` excludes both `mcp-server/src/**` and the harness, so nothing goes inactive; `check-runlogs` runs (its scope includes `^docs/` and `^eval/harness/`) but gates no skill.

**Not verifiable in CI:** no live e2e run was spent, so nothing here exercises the real path where an agent receives `isError` on a rejected write; and there's no production telemetry to say whether the model behaves differently now that failures are visible (`docs/architecture.md` §9.4). The harness mirror tests use the no-workspace short-circuit — real `{ok:false}` responses, but not produced by the compiled tools against a live workspace. Deliberate: the tool-name gate is the thing that can regress, and that *is* pinned.

## Follow-on issues

- **#1569** — lane-check narrowing, Skill-launch gating, and the `is_error` schema-boundary decision.


clack391 (changes requested):
Reviewed the code and ran an independent second pass. The mechanical core is correct, complete, and well-tested: exactly 11 arms route through writerToolResult matching OK_FALSE_IS_FAILURE one-for-one, merge_warnings correctly excluded (dry-run verdict) and validate_research_schema out (uses valid not ok), the wiring test is non-vacuous (un-wiring an arm fails it), and the harness mirror gates on name AND ok is False with a real drift lint. Nicely done.

One thing that should not merge on autopilot, and it is a lead decision. This PR changes what is_error MEANS in e2e run logs: from now on is_error:true means threw OR returned {ok:false}, not just threw. result.py own rule (line 92) is to bump HARNESS_SCHEMA_VERSION when a key keeps its name and type while its meaning changes with no structural tell, which is exactly this, and lines 60-72 document that #1255 made this identical mistake on this identical field. result.py is not in this diff, so the version stays 3 and every post-merge e2e log is stamped 3 with the new meaning, indistinguishable from the pre-merge version-3 corpus the shadow report and calibrate read. #1282 "do not fix it here" covers the lane-check narrowing, not the version bump, which is filed as #1569 section 3 as a lead decision. Clean options: a one-line result.py bump to 4 plus a history entry here, or an explicit lead sign-off to defer on the condition that no e2e run is committed or calibrated between this and #1569.

Two minor non-blocking nits: the harness wiring test iterates a hand-copied list rather than the imported OK_FALSE_IS_FAILURE_LIVE, so a 9th gate tool would not be exercised unless the list is updated too; and the wiring test includes("writerToolResult(") would also match the string in a comment, purely theoretical today.

clack391 (approved):
Re-reviewed. The schema-version finding is fully addressed: HARNESS_SCHEMA_VERSION bumped to 4 with a version-4 history entry, and e2e-test-spec.md section 7.5 boundary table updated to match (the doc side of the bump, not just the constant). No version-asserting test breaks and the harness suite passes (60). Both wiring-test nits are in too. The code core was solid throughout. Approving the code side. Still needs a senior genealogist for architecture.md and the spec.

T-FEH (approved):
Approving. The core is right and @clack391's schema-version catch was the important one — I confirmed the fix is real, not a re-approval on trust: `HARNESS_SCHEMA_VERSION` is 4 with a version-4 history entry, §7.5's boundary table matches, the parametrize now derives from `OK_FALSE_IS_FAILURE_LIVE`, and `stripComments` is in.

Also verified independently: the classification is **not** over-broad (the 12 tools returning `ok:false` are the 11 wired plus `merge_warnings`, correctly excluded as a dry-run verdict); both halves agree on `ok` absent and `ok` non-boolean; and the wiring tests are real ratchets — unwiring an arm, adding `merge_warnings`, or flipping the predicate each goes red.

I pushed one commit (`b46043b1`) with two fixes.

**1. A test that could not fail — and it was the one guarding this change's main risk.** `test_successful_call_carries_no_is_error` guarded its assertion on `ok is not False`, and `section: "project"` is not a valid section. The mock returns:

```
{"ok": false, "errors": ["section 'project' is not one of: questions, plans, log, ..."]}
```

So the guard was always false and the assertion never executed. The test named for proving a *success* keeps its shape never once ran against a success — which is exactly the check standing between this PR and an over-broad `is_error`. Now `section: "questions"`, and the success is asserted before anything is asserted about it. 25 passed.

**2. Three passages still said #1282 was open**, in files this PR edits: `e2e-test-spec.md`'s "`is_error` is also blind to a writer tool that returns `{ok:false}`" and "Revisit only if #1282 lands", and `guardrail-enforcement-spec.md` §7's "MCP writer tools — thrown errors only". The last directly contradicted a paragraph this PR adds 30 lines above it. All three now read in past tense and name `HARNESS_SCHEMA_VERSION` 3 as the tell for pre-change runs. I kept the `#1282` tokens rather than deleting them so the doc-links count is unchanged — the ratchet is exact in both directions, and #1408 is separately moving this file's baseline 12 → 10.

Suites after the fixes: engine vitest **2209 passed / 93 files**, harness pytest **1969 passed**, doc-links **130 passed**.

Two things worth flagging rather than fixing:

- **CODEOWNERS does not require a senior genealogist here.** Your closing note said `architecture.md` and the spec need one; the current file has no `*` default and no `/docs/` rule, so `.md` matches nothing and the only owners this diff triggers are `*.ts`/`*.py` → senior-developers. GitHub agrees — no reviewers were auto-requested. That rule set was rewritten recently, so I think the note predates it.
- **#1408 edits this same region of `e2e-test-spec.md`** (lines ~1355–1365) and lowers the issue-ref baseline to 10. Whichever lands second gets a textual conflict there; the baseline number itself still works out, since #1408 removes two refs my edits left in place.

Since I made the last push, the clearing approval has to come from someone other than me — @clack391, your approval no longer covers the head.

Issue #1574, opened 2026-08-12
tree_forget resolves facts by bare id, not by owner — cross-person removal, and no date-range selector can be built on it


benter-070:
**Touches:** packages/engine/mcp-server/src/tools/tree-forget.ts

Discovered while authoring an e2e fixture from PID `KWJT-3ZB` (Daniel Cook,
issue #1473's source case). FamilySearch returned the **literal same fact
id** (`248eaba5-34ae-4182-9874-bfcc07166241`) for six distinct Birth facts on
six distinct people:

| Person | Birth date/place on this shared id |
|---|---|
| `KWJT-3ZB` Daniel Cook | 15 Dec 1798, Kingsclear, NB |
| `KWJ8-WQ3` David Cook | 16 Mar 1828, East Zorra-Tavistock, ON |
| `KWJT-3ZT` Mary Maria Fuller | 9 May 1798, Kingsclear, NB |
| `KWNX-GQP` Isaac Cook | 10 Jun 1841, West South Hope, ON |
| `KWNT-MX6` Hannah Elizabeth Cook | 18 Sep 1838, South Easthope Township, ON |
| `KWNF-7Q9` James Nathaniel Cook | 15 May 1833, East Zorra Township, ON |

This isn't a harness artifact — `person_read` (and whatever upstream
conversion feeds `tree.gedcomx.json`) preserves FamilySearch's native fact id
verbatim (documented behavior; `eval/harness/e2e/author.py`'s
`normalize_tree` even warns on it: "duplicate fact id ... ids must be
unique"). FamilySearch apparently doesn't guarantee a unique id for a basic
compiled-tree conclusion fact the way it does for sourced/ARK-backed facts.
So any real project's `tree.gedcomx.json` can end up with this shape, not
just this one PID.

**`tree_forget` breaks on it.** Selector resolution correctly scopes to a
person — e.g. `birth-of` calls `factIdsOfType(tree, pid, "Birth")`, which
finds *that person's* Birth fact id(s). But those ids feed a single flat
`Set<factId>` (`resolveSelectors`'s `Targets.facts`), and removal
(`applyForget`'s `pruneFacts`, `tree-forget.ts:307-317`) strips **any** fact
across **any** kept person/relationship whose `id` is in that set — with no
owner check:

```ts
const pruneFacts = (owner: { facts?: SimplifiedFact[] }): void => {
  ...
  owner.facts = owner.facts.filter((f) => {
[...]

Issue #1572, opened 2026-08-12
research.json schema drift (person_id/notes on assertions, author/record_id/title on sources) hard-blocks tree_forget and any validating writer


benter-070:
> **Reviewed 2026-08-12 before junior handoff.** Every factual claim holds, but the
> remedy is a fork that changes a documented write-boundary contract and the block is
> nine tools wide, not one — routed to a senior. The provenance hunt is settled and
> closed below.

**IMPORTANT**: Do PR #1419 first — it is open against `tree-forget.ts` (issues #1314 and
#1417). And coordinate with issue #1549, which extends the same file's redundant-fact
sweep set; both change `tree-forget.ts`, and this issue narrows what the tool validates
before it applies any selector.

**Touches:** packages/engine/mcp-server/src/validation/validator.ts, packages/engine/mcp-server/src/tools/tree-forget.ts

From investigating `feedback-2026-08-06T20-05-41-639Z` (issue #1476, comment
[here](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1476#issuecomment-5265730172)):
this bundle's `research.json` fails schema validation with 45 pre-existing
"unexpected property" errors, all in fields an earlier writer put on the wrong
object:

| Location | Rejected properties | Count |
|---|---|---|
| `sources[0]`–`sources[4]` | `author`, `record_id`, `title` | 15 |
| `assertions[0]` | `person_id` | 1 |
| `assertions[1]`–`assertions[14]` | `person_id`, `notes` | 28 |
| `conflicts[0]` | `resolution_notes` | 1 |

Checked against the current schema — this is real drift, not a validator bug:

- `RESEARCH_SHAPES.assertion` (`validator.ts:443`) has no `person_id`/`notes`.
  That linkage belongs on `person_evidence_entry` (`:450`), which does have
  `person_id`.
- `RESEARCH_SHAPES.source` (`:435`) has no `author`/`record_id`/`title` —
  sources carry that in `citation`/`citation_detail` instead.
- `RESEARCH_SHAPES.conflict` (`:454`) has `resolution_rationale`, not
  `resolution_notes`.

**Provenance — settled: no writer tool could have produced this shape.** Across all
17 committed revisions of `docs/specs/schemas/research.schema.json` (first commit
`476de587`, 2026-05-14) none of the six drifted keys was ever legal on its object,
and `RESEARCH_SHAPES` in `validator.ts` has never carried them either. Reproduce:

```sh
[...]

PR #1575, opened 2026-08-12
fix(forget-and-rederive): account for what the forget step removed

benter-070:
## Summary
- `forget-and-rederive` gains Step 5: hold the forget step's redacted removal summary up against what research actually reasserted, and report re-established vs. still-missing per kind, instead of silently ending with a smaller tree than it started with.
- Adds a completeness note to Step 4 (extract everything a record documents during rederivation, not only the fact originally asked about) so Step 5 has something to reconcile against.
- Adds regression fixture `eval/tests/e2e/cook-spouse-children`, built from the same PID (KWJT-3ZB / Daniel Cook) as the reported case, scaled to a gradeable size: spouse removed entirely, two sons' 1850 census facts removed. `forget-and-rederive` is unit-test-exempt (`RUNLOG_GATE_EXEMPT_SKILLS`) and validated via e2e fixtures instead.

## Related
- Closes #1473 (forget-and-rederive strips relationships and never restores them)
- While authoring the fixture, live snapshotting of KWJT-3ZB surfaced two independent tooling defects, filed and fixed separately (not part of this PR): #1572 (research.json schema drift hard-blocks `tree_forget`) and #1574 (`tree_forget` removes a fact from every person sharing its id, confirmed live on this PID). Also commented on #1476 with the corrected root-cause diagnosis for that session's `tree_edit` fallback.

## Test plan
- [x] `uv run python -m e2e.author validate --slug cook-spouse-children` passes (schema, living-person gate, stripping checks)
- [x] Seeded an editable project (`eval/e2e-project/cook-spouse-children/`, gitignored) to hand off for a live Cowork sanity run
- [ ] Live `/research` run in Cowork against the seeded project, watched in the Research Viewer, to confirm the fixture is answerable (not a pass/fail — the live run doesn't block tree-read tools)
- [ ] Scored headless run (`make e2e-run`), which is the maintainer's step after this lands


mercyokum (approved):
No blocking issues. Approving.

johnmarkpeterbrown (changes requested):
**[CRITICAL]** (8/10) `eval/tests/e2e/cook-spouse-children/starting-tree.gedcomx.json` — finding f1's answer is still in the starting tree

The fixture removes wife Mary Maria Fuller's person record and relationships, and even removes Daniel's own `Marriage` fact — the README says that was done because "leaving Daniel's copy in place would have left the marriage answer visible on his own record." But a `LifeSketch` fact on the **same subject person (KWJT-3ZB)** still narrates the entire marriage, and a second one on daughter Hannah (KWNT-MX6) names the wife too. Finding f1 requires: spouse = Mary Maria Fuller, married ~1820, born ~9 May 1798 New Brunswick. Daniel's LifeSketch states all of it verbatim.

Because the judge grades the final tree and never checks a finding was *newly derived*, the agent can satisfy f1 by reading this note — no marriage-record research, which is exactly the rederivation behavior #1473 is about. f1 becomes a finding that passes whether or not the bug is fixed. f2 and f3 (the 1850 census facts) are properly stripped and don't leak, so they still test rederivation.

This should be fixed before the maintainer's scored run, or that run will report an inflated result on f1.

Fix — redact the two answer-bearing sentences (or drop the two LifeSketch facts) in `starting-tree.gedcomx.json`, and add a line to the README's "What was removed":

```
Person KWJT-3ZB, fact id 87ef2dce-279a-4264-9a0c-751bedc238d4 (LifeSketch)
You wrote:      Daniel Cook V was born in New Brunswick, Canada, December 15, 1798. He married Mary Maria Fuller about 1820. Mary Maria was born May 9, 1798 in New Brunswick, Canada.
Change it to:   Daniel Cook V was born in New Brunswick, Canada, December 15, 1798.
```

```
Person KWNT-MX6, fact id 270e5305-ee1f-46f7-8d60-03177cd59142 (LifeSketch)
You wrote:      Her parents were Daniel Cook [his headstone seems to say Darret Damiel Cook] and Mary Maria Fuller.
Change it to:   Her parents were Daniel Cook [his headstone seems to say Darret Damiel Cook].
```

**What I didn't check / gaps:** I did not run a live `/research` pass against the fixture (that's the maintainer's scored run, still pending per the PR's own test plan) — my f1 conclusion rests on the leaked note plus the judge reading `final_tree`, not on an observed false pass. I did not run the full harness suite or the web/engine typecheck, since the diff is a fixture + an exempt SKILL.md and touches no TS/Python code. The 3 `validate` warnings are benign fuzzy name-overlaps, not defects.

please fix this and notify me in SLACK for review. Thanks

johnmarkpeterbrown (approved):
Looks good!

Issue #1152, opened 2026-08-02
Decide how to test forget-and-rederive's behavioural half


DallanQ:
> **Decided (lead, 2026-09-07): prose, deliberately. Closed.** The four behaviours (dry-run before applying; `project_context` rather than reading the tree file; never restating removed values; stop rather than substitute `tree_edit` when `tree_forget` is absent) stay prose-only, and the permanent run-log-gate exemption stands. Asked "what would save the most time overall?", this is the answer: zero cost now and later. Measured before ruling: zero `tree_forget` calls in the 161-run e2e corpus, one real practice run by the team (which surfaced issue #1473), no tester feedback ever asking for a practice run. The alternatives — a validator-only unit suite (a paid run on every future edit of a skill edited five times since July) and retiring the skill from the plugin (deletes the practice-run feature that found a real bug) — both cost more than the risk they remove. Record this in `docs/specs/tree-forget-tool-spec.md` (or the skill's spec) beside the exemption rationale if someone edits the skill next; do not re-file.

`forget-and-rederive` is permanently runlog-gate exempt (`check_runlogs.py:76`) —
a setup utility has no genealogical output for a judge to grade. Its mechanical
half is covered by `tree-forget.test.ts` (redaction block, `:293`).

Uncovered is the behavioural half: that the skill dry-runs before applying
(`SKILL.md:55-58`), reaches for `project_context` rather than reading
`tree.gedcomx.json` (`:44`), and never restates removed values back to the
researcher (`:88`).

Those are transcript properties, not tool properties. Decide: a targeted lint over
the run transcript, or a deliberate decision to leave it to prose. **Not** a unit
suite — see `RUNLOG_GATE_EXEMPT_SKILLS`.

---

## Trigger, added 2026-08-12

No unblock condition was stated, and no corpus evidence will ever arrive on its own — this
skill is permanently runlog-gate exempt (`RUNLOG_GATE_EXEMPT_SKILLS` in `check_runlogs.py`,
re-verified 2026-08-12), so nothing collects its transcripts.

Thaw when `forget-and-rederive/SKILL.md` is next edited. A change to any of the three
uncovered behaviours — dry-run before applying, reaching for `project_context` rather than
reading the tree file, never restating removed values — is the moment the absent test costs
something, because nothing would catch the edit breaking them.

Until then the prose stands, deliberately.

chesworthrm:
[...]

PR #1576, opened 2026-08-12
forget-and-rederive: gate on tree_forget availability, don't fall back to tree_edit (#1476)

yinkid28:
## Summary
- Adds a `## Prerequisite` section to `forget-and-rederive/SKILL.md` that stops the skill when `tree_forget` is unavailable OR fails, instead of silently falling back to `tree_edit`
- Guards against both failure modes:
  - Tool absent (ToolSearch no match, unknown tool) → tells researcher to rebuild extension
  - Tool present but returns `{ok: false}` (validation errors) → surfaces the actual error and suggests clearing blocking entries or using narrower selectors
- Adds a troubleshooting row to the alpha user guide for stale-extension symptoms
- Per ADR-0011 row 4, prose is the correct tier for this guard (single-invocation judgment, no write boundary crossed)

Closes #1476

## Test plan
- [x] Plugin packages correctly (`node scripts/package-plugin.mjs` — 113 files)
- [x] All 2107 packaging/unit tests pass (`npm test` from `packages/engine/mcp-server/`)
- [x] Drift-critic confirmed no deviation from plan


clack391 (approved):

EdmondOware (approved):
This directly fixes a real, documented problem, not a hypothetical one. The original report showed an agent silently substituting tree_edit for a missing tree_forget, and that same session actually over-removed data and under-restored it. The hard rule against substitution here is exactly the right response to that.

The researcher-facing messaging is good on both failure paths. The missing-tool message is clear and actionable, and the validation-failure message points to a genuinely useful recovery step, narrowing to a fact-level selector instead of removing an entire person, which is usually what triggers the reference conflicts in the first place.

 The alpha guide addition closes the loop on the acceptance criterion's second half, giving a tester a concrete way to recognize a stale install and fix it.

 One thing worth naming rather than leaving silent: this skill has no test coverage of any kind, unit or e2e, so nothing will catch a future regression of this rule. That's a pre-existing gap this PR doesn't create, and issue #1152 already tracks it with a trigger condition this PR just satisfied. Commented there rather than blocking this fix on it.

PR #1577, opened 2026-08-12
question-selection + init-project: test a disputed parent assignment, don't confirm it (#1471)

kofiatinka12:
Closes #1471.

## Bug
Prompt (from the reporter's session): *"LKTP-HHJ — The parents attached in FS tree are not correct — can you locate the correct parents."* The word *correct* is a statement that the current parents are in doubt. Instead, the agent imported and **confirmed** the existing (unsourced) FamilySearch tree — treating the tree under investigation as evidence for itself — and never asked what evidence prompted the doubt. (Per the reporter's feedback form the tester actually holds a death record placing the person in Luxembourg, not Wisconsin — i.e. the wrong person — but that was never elicited in-session.)

Root cause traced from the bundle's **session log** (the actual failing run): the run reached only **init-project**, which imported the tree (13 persons, the tree itself carrying 0 sources) and presented an "init complete" summary with the disputed parents as the baseline. `person_read` was called *with* `relatives:true, sourceDescriptions:true`, so this is **not** the #1475 flagless bug.

## Fix — two scoped doctrine changes
- **init-project (C):** when the objective disputes the existing relationship ("correct parents", "not correct", "the right X"), frame the imported parent-child assignment as the **relationship under investigation** — an *unverified* (`quality:1`) tree assertion that is the hypothesis to test — not a settled fact, and never confirm it from the tree it came from. Framing only.
- **question-selection (B):** when the objective signals doubt of an existing assignment, **in interactive mode ask** (1) what evidence led to the doubt and (2) the birth date/place the user is working from, then frame the first question as a **test** of the disputed assignment (confirm-or-refute against independent records). **Under `--autonomous`, skip the ask** (no user to answer) and go straight to the verification-framed question.

**person-evidence and research are untouched** — the session never reached them; the downstream "never self-cite the questioned tree in a *conclusion*" guard is genuinely out of scope for this first-pass failure. No `oral_knowledge` added.

## Evidence & verification — three separate things, deliberately not conflated
1. **Session evidence (ground truth):** the reporter's own trace shows init-project confirming the disputed tree and no doubt elicited.
2. **Deterministic regression (question-selection):** `ut_question_selection_013` + the synthetic `disputed-parents-unsourced` scenario + a tag-gated validator (`test_first_question_tests_disputed_parents`). **Baseline (un-fixed) FAILED** — the first question did not test the disputed assignment; **fixed run PASSES**. That baseline-fail → fixed-pass is the behavioral evidence.
3. **`skill_gate.py` returned INCONCLUSIVE, not LOOKS GOOD.** The gate compares **LLM-judge dimensions**, which did not move (the judge passed the un-fixed question too). **The gate did not validate the fix** — the deterministic validator did. This is a known gate limitation (judge-only comparison), documented here rather than worked around.

## Runs (committed + fully annotated)
- **question-selection** `v1_2026-08-11_17-18-48`: **13/13 pass**, incl. `ut_013`.
- **init-project** `v1_2026-08-11_17-36-45`: **8/9**; `ut_006` the sole non-pass (place_search not called).
  - **Not a C regression.** The init-project non-pass **moved across two runs** (`ut_005`/`ut_002` → `ut_006`), and **all three are C-inert** — none has the disputed-objective phrasing that triggers the C note (verified `triggers C? False`). A failure that hops between unrelated tests run-to-run is single-run variance, not a deterministic regression. Left as pre-existing init-project flakiness, **out of scope for #1471** (not fixed here).


packages/engine/plugin/skills/question-selection/SKILL.md
```diff
@@ -140,6 +140,18 @@ build a question on unverified claims from compiled sources (online trees,
...
+**When the objective signals the user doubts an existing assignment** —
+phrasing like "correct parents", "the right X", "not correct" — the current
+tree assignment is the premise *under doubt*, not a starting fact. **In
+interactive mode, before formulating, ask the user two things: (1) what
+evidence led them to doubt the current assignment, and (2) the birth date and
+place they are working from.** Then frame the first question as a **test of the
+disputed assignment** — e.g. "Do independent records confirm or refute that X
+and Y are the parents of Z?" — never treating the questioned tree as evidence
+for its own conclusion (issue #1471). Under `--autonomous` (no user to answer),
+do **not** block: skip the ask and go straight to the verification-framed
```
Paaboat: The phrasing "do not block: skip the ask" is slightly ambiguous—"do not block" could be read as "execute the block labelled 'not'" rather than "do not ask the user." Suggest rephrasing to: "Under—autonomous, skip the ask and go straight to the verification-framed question."

packages/engine/plugin/skills/init-project/SKILL.md
```diff
@@ -194,6 +194,16 @@ Analyze imported data before presenting results:
 - Holding disagrees with tree → flag as discrepancy (never frame user's holding as error).
 - `oral_knowledge` lead → surface early; oral sources are cheapest and most perishable.
 
+**When the objective disputes the existing relationship** — phrasing like
```
Paaboat, line 199: The addition of relatives: true, sourceDescriptions: true to person_ read looks like it fixes a separate bug (issue #1475—subject-only tree imported with no spouse, children, or sources). The fix is correct and needed, but the PR title and description only reference #1471. Suggest adding Closes #1475 to the PR description so the issue is linked and reviewers know this change is intentional.

Paaboat (commented):
The core fix is correct and GPS-aligned. init-project now frames an unsourced disputed relationship as the hypothesis under investigation rather than confirmed fact, and question-selection gathers user evidence and frames the first question as a direct test of that relationship—never treating the tree as evidence for its own conclusion. Autonomous mode is covered. The ut_question_selection_013 regression at 13/13 is solid. Three minor comments inline—none block the merge.

Paaboat (approved):
The core fix is correct and GPS-aligned. init-project now frames an unsourced disputed relationship as the hypothesis under investigation rather than confirmed fact, and question-selection gathers user evidence and frames the first question as a direct test of that relationship—never treating the tree as evidence for its own conclusion. Autonomous mode is covered. The ut_question_selection_013 regression at 13/13 is solid. Three minor comments inline—none block the merge.

johnmarkpeterbrown (changes requested):
Hi @kofiatinka12 — doctrine changes look right and the scope is honest, nice work. The merge from `main` turned the `runlogs` check red, so a few things before this goes to senior review. Do them in this order:
>
> **1. Fix the test-id collision (required).** While this PR was open, another PR landed on `main` adding `stay-inside-objective-scope.json`, which claims id `ut_question_selection_013` — the same id as your new test. Renumber yours to the next free id, `ut_question_selection_014`, in **three places**:
> - rename the file `ut_question_selection_013.json` → `ut_question_selection_014.json`
> - the `"id"` field inside it
> - the reference in `eval/fixtures/scenarios/disputed-parents-unsourced/README.md:9`
>
> (No validator change needed — it keys off the `verifies-disputed-parents` tag, not the id.)
>
> **2. Regenerate the run logs (required).** After the merge, the committed run logs for both skills no longer match the tree. Re-run the harness for both so they reflect the merged branch — do this **after** step 1 so the rename is captured:
> ```
> uv run python eval/harness/run_tests.py --skill question-selection
> uv run python eval/harness/run_tests.py --skill init-project
> ```
> Please don't use `eval-cosmetic-skip` — this is a behavior change, not cosmetic.
>
> **3. (Optional, worth it) Tighten the regression validator.** In `eval/harness/validators/test_question_selection.py`, `_VERIFY_SIGNALS` currently accepts a bare `"confirm"`/`"verify"`. That means a question like *"Confirm that Johann and Maria are the parents"* — the exact confirm-the-tree failure #1471 is about — would pass. Drop standalone `"confirm"` and `"verify"` from the tuple; a correctly framed "confirm **or refute** against **independent** records" question still matches on `refute`/`independent`, so the good case keeps passing and the bad case now fails.
>
> Push those and once CI is green I'll do the full genealogist pass. Thanks!

---

**Two notes for you as the merge decider:**
- Even once `runlogs` is green, this needs a **fresh approval after the merge** — Paaboat's approval predates it, and you pushed the merge commit yourself (`johnmarkpeterbrown`), so the last-push-approval rule wants an approval from someone who isn't you (unless you're in the `senior-developers` bypass group).
- **Gap I'm flagging:** I did not run the suites in a worktree, because the run logs are about to be regenerated — I reviewed the doctrine/validator/fixture/test source directly instead. Once the author pushes the regenerated logs, the run-log content (the bulk of the diff) is worth a glance to confirm `ut_014` passes and no other test regressed.

johnmarkpeterbrown (approved):
Looks good!

clack391 (approved):
Everything looks good

PR #1582, opened 2026-08-12
research-plan: check jurisdictional boundary changes, size indirect-record windows, guard married surnames

Cia-3:
## Eval

`make eval-skill SKILL="research-plan locality-guide"`: **39 pass, 5
partial, 1 fail, 2 aborted** of 47.

- `ut_locality_guide_025` (fail) — a negative/routing test unrelated to
  anything this PR touches (frontmatter/description weren't edited).
  Re-ran standalone: passed cleanly (correct decline, 17s). Same story as
  the earlier `003` case — single-run jitter, not a regression.
- `ut_research_plan_wzk` (aborted, `max_wall_clock_seconds`) — the test
  written specifically to validate Rule 3. Zero graded dimensions in the
  committed run; re-ran standalone and it **passed cleanly** (289s, 13
  turns, 0 retries), confirming Rule 3's fix works against the scenario
  it was built to prove.
- `ut_research_plan_016` (aborted, `max_wall_clock_seconds`) — zero graded
  dimensions, no annotation required. Infra stall, not a skill defect.

Both run logs fully annotated (research-plan: 127 corrections;
locality-guide: 150 corrections; 0 missing dimensions on either).


Paaboat (changes requested):
The three rules are sound, and the branch is properly based. Eval hygiene is careful—both non-passes are correctly identified as infra/jitter and re-verified standalone.

One gap is still outstanding after the latest commits: no fresh Jimmie-Jewel-Neal run after the Rule 3 skill change. The PR description cites six prior runs that demonstrate the bug (Mary E. Wood is attached as maternal grandmother at the "probable" tier, and the birth surname Sampson is absent from all six). That shows the problem existed—but there is no run showing the fix actually resolves it. The issue's acceptance criterion requires that case to pass. Would need a fresh Jimmie-Jewel-Neal run with the updated skill before approving.

Paaboat (approved):
Clearing my review per Dallan's direction to land as-is.

florencemashipei (approved):
Verified all three research-plan/SKILL.md doctrine fixes and the locality-guide/SKILL.md temporal-vs-geographic split addition against the actual diff — all match their descriptions precisely. The Rule 3 factual claim (six jimmie-jewel-neal e2e runs wrongly attaching "Joseph B and Mary E Wood" as maternal grandparents, with "Sampson" never appearing) checks out exactly against the real run-log tree/proof-summary data. Annotations are genuinely complete (285 corrections total, every dimension covered).

**One non-blocking issue: the "Eval" section's numbers are stale.** It cites `ut_locality_guide_003` (fail) and `ut_research_plan_014` (aborted), "39 pass, 2 partial, 1 fail, 1 aborted of 43." The actual currently-committed latest run logs show:

- `research-plan` (`v1_2026-08-13_14-15-48`): 19 pass, 1 partial, 1 aborted — the aborted test is `ut_research_plan_q7m` (`_014` is a clean pass)
- `locality-guide` (`v1_2026-08-12_21-03-12`): 22 pass, 3 partial, 1 fail — the fail is `ut_locality_guide_025` (`_003` is a clean pass)

Combined: 41 pass / 4 partial / 1 fail / 1 aborted of 47, not 39/2/1/1 of 43 — looks like leftover text from an earlier iteration before a later re-run. I checked whether the real failure (`ut_locality_guide_025`) is actually a regression from this PR's changes: it's a negative/routing test (generic how-to question should route to `search-familysearch-wiki`), which depends on the skill's frontmatter `description` — this PR never touches that, only Step 2's body prose. So the underlying "unrelated routing-test jitter" reasoning the PR already gives is very likely still correct, just attached to the wrong specific test ID.

Since this doesn't affect anything CI actually gates (the real run logs/annotations are already correct) and PR descriptions are editable anytime, not worth blocking merge over — but please update the Eval section's numbers/test IDs to match the actual final run logs above, so the next reader doesn't have to re-derive this.

Approving.

florencemashipei:
Heads up — I updated this branch against main to clear the \`BEHIND\` merge-state (needed for the 2 approvals to actually be mergeable), and the resulting merge introduced two real problems that need fixing before this can merge. Your approvals/review substance are unaffected — this is a merge-side issue, not a problem with your actual changes.

**1. Two pre-existing annotation files now have literal, unresolved git conflict markers baked in:**
- \`eval/runlogs/unit/locality-guide/v1_2026-08-06_21-20-18.ann.json\`
- \`eval/runlogs/unit/research-plan/v1_2026-08-12_18-21-29.ann.json\`

This is a known failure mode (rename/rename merge conflicts can leave markers baked into the auto-merged result even when git reports no conflict). Fix: restore each from its last valid commit — \`git show <commit>:<path>\` — rather than hand-editing. Do **not** hand-edit either file directly; per \`check_runlogs.py\`'s own error message, delete and re-annotate via the CRUD UI if a clean restore isn't straightforward.

**2. \`research-plan\`'s run log is no longer active.** Main independently picked up \`b18b9b2f\` (strip a UTF-8 BOM from \`research-plan/SKILL.md\`) since this PR was last updated. That's a byte-level fix, not new doctrine, but combined with this PR's own real edits to the same file, the merged \`SKILL.md\` no longer matches the last committed run log's snapshot. This needs a genuine re-run: \`uv run python eval/harness/run_tests.py --skill research-plan\`, followed by re-annotating it in the CRUD UI.

\`locality-guide\` only has a non-blocking warning (its run log was scored against an older judge-prompt hash) — not a blocker, just worth noting.

CI (\`pytest\`, \`runlogs\`) is currently red and will block merge until both are fixed.

EdmondOware:
Holding off on merge approval. The doctrine is sound, I checked the Feliciana Parish 1824 split and the Lackawanna Luzerne 1878 split independently and both are accurate, and rule three is well grounded in the six jimmie jewel neal runs cited. The problem is the Eval section of this PR no longer matches the actual committed run log.

What needs to happen before this merges.

One. Rerun ut research plan wzk, the martha remarriage surname plan test, standalone. It is the test written specifically to validate rule three, the headline fix in this PR, and in the currently committed run log it aborted on max wall clock seconds with zero graded dimensions. There is no passing verification anywhere in the committed evidence that the remarriage fix actually works in the one scenario built to prove it. A clean pass on a standalone rerun closes this.

Two. Update the PR body's Eval section to match the current active run log, not the earlier one. Right now it says the one fail was ut locality guide 003. The actual fail in the committed run is ut locality guide 025, a different unrelated routing test, same story, wrong id. It also says the one abort was ut research plan 014 on sdk stream silence, re run clean. In the current run, 014 actually passes, and there are two different aborts instead, wzk above and ut research plan 016, both on max wall clock seconds. The pass, partial, and total counts in the PR body do not match the current run log either.

For the record, the other two new tests are fine. Feliciana parish boundary split passes cleanly, and caroline probate creator lifespan came back partial, but I read the annotator's own comments and the partial is entirely about tool call mechanics, only four of six calls were reasonable, every doctrine relevant rubric dimension scored a clean pass.

Once wzk is confirmed passing and the Eval section reflects the real numbers, this is ready to go.

Cia-3:
Both closed out with fresh standalone reruns just now:

- `ut_research_plan_wzk` (the remarriage-surname-plan test built for Rule 3): passed cleanly — 289s, 13 turns, 0 retries. Rule 3's fix works against its target scenario.
- `ut_locality_guide_025` (the fail you flagged): also passed cleanly standalone — 17s, correct decline/routing. Confirms it's jitter, same as the earlier `003` case, unrelated to this PR's edits.

Updated the PR description's Eval section below to match the currently committed run logs — corrected numbers and both test IDs.

EdmondOware:
The conflict marker fix and the wzk explanation both look right to me. wzk passing on a standalone recheck is consistent with how this team already handles an isolated infra abort elsewhere in this same PR, so I am not asking for a fresh full committed run over that alone.

One thing still open. The PR body's Eval section was not updated. It still says the fail was ut locality guide 003 and the abort was ut research plan 014 on sdk stream silence. Neither is what the current committed run log shows, the real fail is ut locality guide 025, and the real aborts are wzk and 016 on max wall clock seconds, addressed in your latest commit message. Could you edit the PR description itself to reflect that, since the commit message explaining it will not be the first thing the next person reads. Once that is updated this looks ready from my side.

Cia-3:
Double-checked just now — the description is already updated (has been since before I posted the last comment). It reads "39 pass, 5 partial, 1 fail, 2 aborted of 47" with ut_locality_guide_025 and ut_research_plan_wzk/016 named. Might've been a caching/timing thing on your end — can you take another look?

PR #1583, opened 2026-08-12
feedback: FEEDBACK.md always states why the session log is or isn't there (#1481)

### LINKED ISSUE #1481: Cowork bundles carry no session transcript, and FEEDBACK.md does not say so
**Touches:** apps/electron/src/main/feedback.ts

Split out of issue #1477, which conflated this with a real retention bug.

A Cowork-submitted bundle never contains `_feedback/session-log.jsonl`. All five Cowork/Windows bundles from 2026-08-05..08-07 lack one; all five hosted-web bundles have one.

**This is expected, not a regression.** `readSessionLog` (`feedback.ts:166-169`) reads `~/.claude/projects/` — where Claude Code writes sessions. Cowork runs the agent in its own VM, so there is nothing there to find. `docs/alpha-user-guide-cowork.md:269-271` already tells testers so:

> **"Include Claude Code session log"** — this captures a Claude Code session transcript specifically, so it won't have anything to attach for a Cowork session; leave it as-is.

The defect is that **the bundle does not say so.** When `sessionLogIncluded` is false, `renderFeedbackMarkdown` simply omits the section (`feedback.ts:475-482`), so `FEEDBACK.md` is silent and triage starts by hunting for a file that was never written. That is what sent the first pass at issue #1477 down the wrong path.

Scoped small deliberately: with the retention fix in issue #1477, `results/` carries Step 4, so this is a clarity fix rather than a capability gap.

**Acceptance:** `FEEDBACK.md` states plainly when the session log is absent and why, distinguishing "not applicable — Cowork session" from "expected but missing."

**Not in scope:** capturing a Cowork transcript. Cowork's session storage is not ours. If a runtime-independent transcript is ever wanted, the MCP server is the only always-on-host component in the loop and would be the place — but do not start there without a reason the `results/` sidecars fail to cover.

**Evidence:** all ten bundles from 2026-08-05T18:50:23.256Z onward.

francis-2008-happy:
Closes #1481.

## The issue
Cowork sends the agent's work to a sandboxed VM, so a Cowork session leaves no Claude Code transcript on the user's machine. But FEEDBACK.md only rendered a "## Session log" section when a log was actually attached. When there was no log — which is the normal case for a Cowork bundle — the section just vanished, and there was nothing in the bundle telling a triager that the absence was expected rather than a lost file.

## The fix
`apps/electron/src/main/feedback.ts` — replace the `sessionLogIncluded` boolean with a 3-state `sessionLogStatus` (`included` / `not-requested` / `requested-but-empty`) and **always** emit the "## Session log" section, worded per state:
- **included** — points at `_feedback/session-log.jsonl` (unchanged text).
- **not-requested** — says no log was included and that for a Cowork session this is expected (agent runs in Cowork's VM, no Claude Code transcript here; points at `docs/alpha-user-guide-cowork.md`), and that the `results/` sidecars carry the step record.
- **requested-but-empty** — says a log was requested but none was found under `~/.claude/projects`; expected for a Cowork session, otherwise a transcript that should be here is missing.

The zip write itself is unchanged (session-log.jsonl is still only written when entries are found).

## What this does not solve
- The web bundler (`apps/server/app/feedback.py`) has the same structural gap but is left alone — it renders its own markdown and wasn't in scope for this fix.
- `feedback.json` / the feedback schema are untouched; this is a human-readable-note change only.
- This does not make a Cowork session's transcript recoverable — there is none on the host to capture. It only labels the absence correctly.

## Verification
- `apps/electron` vitest: 22 pass, including 3 new tests covering each state (not-requested → Cowork-expected wording; requested-but-empty → "requested but none was found"; log present → points at the jsonl, with HOME redirected at a planted `~/.claude/projects/<hash>` fixture).
- `npm run typecheck` (node + web) clean.

## For review
`sessionLogStatus` is fully internal to `feedback.ts` (`renderFeedbackMarkdown` isn't exported; the IPC surface and `feedback.json` are unaffected), so the blast radius is this one file plus its test.


chesworthrm (changes requested):
Full pass on the tree with `origin/main` test-merged in (the branch is 45 commits behind; the merge is clean). All checks green on `35a093f5`. `apps/electron` vitest 65/65, `turbo test typecheck` 8/8, `apps/server` pytest 157 passed. I also reverted the section-push to its pre-PR position to confirm the new tests actually catch the regression — 2 of the 3 fail, so they are doing real work.

The structure is right and the tests are good. One branch is wired to the wrong case, and I think it has to change before this merges.

## The `not-requested` text explains the one case that state can never mean

I traced which state a real submission lands in:

- `FeedbackDialog.tsx:40` — `const [includeSessionLog, setIncludeSessionLog] = useState(true)`. The box is ticked by default.
- `FeedbackDialog.tsx:369` — `disabled={!hasSessionLog}`. When there is no transcript the box is **disabled**, so the user cannot untick it. Line 131 submits the raw `includeSessionLog` state, not `includeSessionLog && hasSessionLog`.
- `main/index.ts:133` — `includeSessionLog: payload.includeSessionLog !== false`.

So a Cowork submission arrives with `includeSessionLog: true`, finds nothing, and lands in `requested-but-empty`. That branch's wording is correct and does exactly what #1481 asked for.

`not-requested` is reachable only when the box was **enabled** — meaning a transcript does exist on that machine — and the submitter deliberately unticked it. For that case the text says there is no transcript on the machine to attach and that this is expected for Cowork. Both halves are false. It also hides the thing triage most needs to know: the submitter opted out, so the log can be asked for.

```
You wrote:    'No Claude Code session log was included. For a Cowork session this is expected — ' +
                "the agent runs in Cowork's own VM, so there is no Claude Code transcript on this " +
                'machine to attach (see `docs/alpha-user-guide-cowork.md`). The `results/` sidecars ' +
                'carry the search/step record instead.'

Change it to: 'No Claude Code session log was included — the submitter unticked "Include Claude ' +
                'Code session log" while a transcript was available on their machine. Ask them ' +
                'for it if the transcript is needed to diagnose this case.'
```

The Cowork explanation and the `results/` pointer belong in the branch Cowork actually reaches:

```
You wrote:    'A Claude Code session log was requested but none was found under `~/.claude/projects`. ' +
                'For a Cowork session that is expected (the agent ran in Cowork’s VM); for a Claude ' +
                'Code session, the transcript that should be here is missing.'

Change it to: 'A Claude Code session log was requested but none was found under `~/.claude/projects`. ' +
                "For a Cowork session that is expected — the agent runs in Cowork's own VM, so there " +
                'is no Claude Code transcript on this machine to attach (see ' +
                '`docs/alpha-user-guide-cowork.md`); the `results/` sidecars carry the search/step ' +
                'record instead. For a Claude Code session, the transcript that should be here is missing.'
```

The first new test pins the wrong wording in place, so it moves with it:

```
You wrote:    it('explains an absent log as expected for a Cowork session when none was requested', async () => {
Change it to: it('says the submitter opted out when the log was not requested', async () => {
```

```
You wrote:      expect(md).toContain('Cowork')
Change it to:   expect(md).toContain('unticked')
```

## Not blocking

Leaving `apps/server/app/feedback.py` alone is the right call, not a deferred chore — the hosted web app captures the agent's transcript from the sandbox, so an absent log there means something different and needs different wording, not a copy of this one.

## What I did not check

`eval/harness` pytest — a fresh worktree has no compiled engine output, so its CLI tests fail there for environmental reasons, and the diff touches no Python or harness path. I did not exercise the feedback dialog in a running Electron app; the state routing above is read from source, not observed live. `apps/electron/.claude/skills/security-invariants` asks for a pass on any PR touching `src/main/` — I walked it, and the diff adds no IPC channel, dependency or CSP change, so no invariant is in play.

chesworthrm (approved):
Re-reviewed at `8cb2db69`. The swap landed exactly as asked; approving.

`c2655f0f` is the only content commit since my review of `35a093f5` — the other
four are clean main merges, and the PR's content diff is still just
`feedback.ts` + `feedback.test.ts`. Both branches now say what their state means:

- `feedback.ts:509` — `not-requested` says the submitter unticked the box while a
  transcript was available, and to ask them for it.
- `feedback.ts:516` — `requested-but-empty` carries the Cowork-VM explanation, the
  `docs/alpha-user-guide-cowork.md` link, and the `results/` sidecar pointer,
  then still flags the Claude-Code case as a genuinely missing transcript.

That is the state a real Cowork bundle lands in, so #1481 is now answered in the
branch it actually reaches.

Checks: `apps/electron` vitest 23/23, all CI green at head. I also put the pre-fix
`feedback.ts` back under the post-fix tests — `says the submitter opted out when
the log was not requested` fails on `toContain('unticked')`, so the regression I
flagged is pinned rather than just described.

## Not blocking

The second test's `expect(md).toContain('Cowork')` doesn't discriminate: the old
`requested-but-empty` wording also said "Cowork", so that assertion passed
unchanged through my swap-back. It won't catch the `results/` pointer or the doc
link going missing — the parts that actually moved into this branch. Asserting on
`results/` instead would make it bite. Fine to leave or fold in, your call.

Thanks for tracing the dialog logic through on the fix — the comment at
`feedback.ts:498` recording *why* each branch is worded the way it is, is the part
that keeps this from getting swapped back by the next reader.

promise-emmanuel (approved):
Changes looks good! Approved.

PR #1251, opened 2026-08-04
eval-harness: warn when a shared-fixture edit stales a referencing skill's run log (#1094)

Asimi1234:
## Summary

- **Normal tier.** `check_runlogs.py` derived touched skills from skill dirs, tests, and plugin agents, but **not** from shared fixtures — so editing `eval/fixtures/{scenarios,mcp}/<name>` invalidated the run-log snapshot of every skill whose tests reference it, yet rules 2/3 never fired and the stale run log passed green (#987/#1094).
- Adds a **warn-only** fixture arm: a changed fixture maps to its referencing skills (`skills_referencing_fixtures`, reusing `snapshot._collect_refs`) and emits a `::warning::` when that skill's latest run log is now stale.
- **Warn-only, not blocking**, per the settled #1094 decision: ~20 of ~25 skills' run logs are already stale on `main` from prior fixture drift, and `mid-research-flynn` alone is referenced by 20 skills, so a blocking gate would fire a ~$160–240 / 20-annotation re-run wave on the first fixture PR and train `eval-cosmetic-skip` misuse. Blocking only on *new* staleness is the named target end-state, deferred to #1242.

## Review guidance

**Start here:** `rule2_fixture_touched` in `eval/harness/scripts/check_runlogs.py`. It is scoped to `changed_fixture_paths` — it reports **only** the snapshot keys this PR changed, not the skill's unrelated pre-existing drift (SKILL.md, other fixtures — the #1217 baseline). Check that scoping: it is what keeps the warning honest and also suppresses a false positive when a skill's snapshot never embedded the changed fixture. The two resolution traps are in `FIXTURE_PATH_RE` / `skills_referencing_fixtures`: scenarios resolve by **directory name** (`input.scenario`), mcp fixtures by **bare name** with `.json` stripped (`mcp_fixtures[]`).

**Unsure about:** Nothing outstanding. The B-vs-C question (warn-only now vs block-on-new) was settled with the lead before building — B now, C tracked as #1242.

## Plan

**Didn't change:** No `.github/workflows/check-runlogs.yml` edit — its `scope` already carries `^eval/fixtures/`, so the job already runs on a fixture-only PR. The blocking `touched_skills` path and `rule2_active`/`rule3` bodies are untouched; the fixture arm is a separate warn-only set that never feeds `fails`. No schema/`packages/schema` change.

[...]

PR #1585, opened 2026-08-12
eval: correct mid-research-flynn README drift and lint README counts (#1333)

Emruthwill:
Closes #1333. **Tier:** Normal (plan → two plan-critic rounds → implement → `/check-drift` → `/code-review high` from a fresh session).

## What was wrong

`eval/fixtures/scenarios/mid-research-flynn/README.md` claimed **4 sources**; the fixture holds **9**. That file is pasted verbatim into the judge prompt as `{scenario_readme}` (`harness/orchestrator.py::_load_scenario_readme` → `eval/harness/judge/prompt.md`), so every judge call on the **124 tests across 20 skills** using this scenario was told something false about the before-state — which is how a judge came to fail a skill for "fabricating" `src_006`.

## What changed

- **`mid-research-flynn/README.md`** — 9 sources, split into the four logged and the five carrying `log_entry_id: null` with no assertions, named individually so a judge cannot read "`src_006` does not exist" out of it.
- **`flynn-fan-pivot/README.md`** — the only other scenario stating a count its `research.json` contradicts (Log: 8 vs 10). Its delta section was also written against the **wrong baseline**: it describes a delta from `flynn-census-exhausted`, where `pli_006`/`log_006` *are* the probate search, while here they are the 1870 census and probate has moved to `pli_009`/`log_009`. "Adds `pli_006`–`pli_008`" was simply false.
- **`flynn-fan-pivot/research.json`** — two prose fields, because the skill reads the JSON, not the README that explains it: the timeline gap note claimed church records were unsearched when `log_010` is a negative church search, and `exhaustive_search_summary` omitted both the probate and church negatives that the `exhaustive_declaration` records.
- **New lint** `test_scenario_readme_counts_match_research_json` — pins every `- **Label:** N …` bullet against its `research.json` array length, so the class cannot recur.
- **`docs/specs/unit-test-spec.md`** — it claimed these READMEs are auto-generated; no generator exists. Its illustrative sample restated *this fixture's* counts and had drifted the same way, so it is now generic.
- **`Makefile`** (separate commit) — `make` resolves a bare `scripts/foo.sh` recipe itself instead of handing it to `SHELL`, which dies on any checkout under a path with spaces. That made `make test-all` unrunnable for such clones; four recipes carried it.

## The issue's cost premise does not hold — please read this before scheduling

#1333 budgets **"$80 and seven hours… twenty separate grading passes"** on the premise that `check-runlogs` rule 2 blocks until all twenty skills are re-run. Measured on this branch:

1. **A fixture edit gates nothing today.** `check_runlogs.py` derives `touched_skills` only from run-log, unit-test, skill and agent paths — there is no `eval/fixtures/**` arm. The workflow runs and finds zero touched skills. That gap is **#1094, still open**; its implementation **PR #1251 is open, not merged**.
2. **All 20 skills' latest run logs were already stale** before this branch touched anything (`diff_snapshot_vs_disk`, after merging `origin/main`). Three were active on the pre-merge base; main's own commits staled them.

**So this PR's marginal eval cost is zero.** It cannot stale a run log that is already stale. The $80 figure is the cost of making the whole corpus active — the backfill #1094 already names as separate genealogist labour.

`eval-cosmetic-skip` is *not* used, and is not needed: the label only relaxes rule 2, and rule 2 does not fire here.

## Verification

The lint was **built and proven to fail first** (`CLAUDE.md` § "A new lint must be proven to fail"): against the pre-edit READMEs it reports exactly two failures — `mid-research-flynn` Sources 4 vs 9 and `flynn-fan-pivot` Log 8 vs 10 — re-proved after merging main.

| Suite | Result |
|---|---|
| `make harness-test` | 2018 passed, 3 skipped |
| `make engine-test` | 2152 passed (89 files) |
| `make eval-ui-test` | 148 passed (14 files) |

`typecheck`/`test-js` could not be run locally (incomplete pnpm workspace install — `apps/web/node_modules/@genealogy/` is empty while the make stamp claims success) and `server-test` deadlocks locally in the `LocalProvider` subprocess tests. Both are pre-existing and environmental; neither reaches a line this PR changes, and CI runs both on a clean checkout.

## Design note: why the lint anchors on a *leading* integer

A "first number anywhere in the bullet" scan reads `q_002 (1850 census placement, resolved)` as a claim of 1850 questions — 18 scenarios flagged, 16 spuriously. Anchoring on the leading integer flags 2, both genuine. It also ignores 4-digit values, so a legal `- **Sources:** 1850 census (FS), …` cannot hard-fail CI claiming 1850 sources.

## Knowingly left out

Both are recorded rather than filed, per the maintainer's instruction.

1. **`flynn-census-exhausted/README.md` contradicts its own `research.json` in three places** — `:8` says `pli_006` "has been removed entirely" (it is present, probate, completed, with `log_006` its negative); `:22` says probate has not been searched; `:12` claims log/sources are "identical to `mid-research-flynn`" (6 vs 5, and 4 vs 9). Not fixed here because both readings are internally coherent and the fix direction changes what **five `research-exhaustiveness` tests** exercise. Recommendation: correct the JSON, not the README.
2. **`eval/tests/unit/question-selection/ut_question_selection_005.json`** describes the same fixture as having six searches when it has ten. Same drift class — but a **test-corpus** edit gates where a fixture edit does not (`check_runlogs.py:437`), and `question-selection`'s latest run log is stale, so folding it in would attach a paid `make eval-skill` plus a grading pass to an otherwise free PR. `CLAUDE.md` carves out exactly this case.


francis-2008-happy (approved):
Great job

johnmarkpeterbrown (approved):
looks good!

johnmarkpeterbrown (approved):
looks good!

clack391 (approved):

johnmarkpeterbrown (approved):
looks good!

PR #1263, opened 2026-08-04
e2e: resolve sebastiano-mingazzini-son (true match); mine person-evid…

EdmondOware:
…ence same_person compliance gap

Resolves the sebastiano-mingazzini-son record-hint fixture (issue #872) as a true match — reasoning in the fixture's README: same source collection already trusted for this family, exact given-name match, routine Italian surname-suffix variant on the mother, consistent geography, and chronology that fits without strain.

Ran the fixture end-to-end: the required finding was recovered (Simone Mingazzini correctly linked as a child of Sebastiano and Lucia in the final tree, sourced) but graded partial — the agent worked from a different record (1887 marriage bann) than the fixture's hint (1933 death record), so birth year/place disagree (1859 Castel Bolognese vs. 1860 Faenza) and the death fact was never captured. Proof quality graded thin: sound tier choice (probable, not overstated) but resting on a single derivative source with two named repositories unsearched.

The run also failed compliance: person-evidence linked four new persons via person_evidence entries without ever calling same_person, despite each carrying a record_persona_id and an existing tree candidate. Mined as ut_person_evidence_n7v, reproduced in 2 of 3 clean test observations (the third and a fourth full-suite attempt were lost to infrastructure issues, see below) plus the original transcript's 4-for-4 pattern within one session.

Also fixed two problems in the mined test itself, unrelated to the skill:
- rubric.md's "Score discipline" dimension could be read as N/A rather than fail for a silently-skipped mandatory same_person call; sharpened the wording.
- The scenario's tree.gedcomx.json was missing a source entry that record-extraction would already have written in a real run, causing person-evidence to write into a tree.gedcomx.json section it doesn't own (sources) and fail test_tree_ownership_table for the wrong reason.

Not included: the actual person-evidence SKILL.md fix for the same_person compliance gap. Getting a clean, gradable 20-test baseline to improve and gate against is blocked by issue #1245 (mass test-abort bug), assigned to Adedotun. That fix lands in a follow-up PR once #1245 clears.

## Summary

Tier: Normal

- Resolved the sebastiano-mingazzini-son e2e record-hint fixture (issue #872)
  as a true match, with reasoning recorded in the fixture's README.
- Ran and graded the fixture end-to-end: required finding recovered but
  partial (agent worked from a different record than the fixture's hint, so
  birth facts disagree and the death fact was never captured); proof quality
  thin (single derivative source).
- Mined the run's compliance failure — person-evidence never calls
  same_person for four newly-linked persons despite existing candidates —
[...]

PR #1586, opened 2026-08-12
e2e: resolve wasyl-kapach-birth to 1879 and land its first run

EdmondOware:
Resolves #882. Record-hint adjudication came back outcome (b): the identity is a true match, but the hint's birth year is wrong.

The 1926 census entry (ark:/61903/1:1:QPR9-YLYV) IS Wasyl Kapach L17M-W4Q. Andrew Kapach's Find a Grave memorial (196999027) names his parents outright, and the 1948 obituary's one son and three daughters match the census household four for four. But the census's 1883 is an age estimate (recorded age 43), and it under-ages the whole household the same way - Mary 1894 vs 1893, Natalka 1914 vs 1913. Headstone, memorial and obituary all give 1879.

The son enumerated as "Henry" looked like a fatal objection and turned out to be the proof: Find a Grave records him as "Andrew Kapach, also known as Henry", d. 13 Apr 1984, an exact match to the tree's Andrew.

Findings are now f1 born 1879 Ukraine, f2 immigrated 1906, f3 avoid 1883, and f4 death/burial as a bonus. f3 is safe under apply_avoid_guard - the guard exempts subject_person_ids, populated from source_pid.

Bundled provided-documents/findagrave-kapach-captures.pdf per spec 6.2. It is mandatory, not a convenience: Wasyl is absent from FamilySearch's Find a Grave index entirely, so no ark exists and external_links_search cannot reach him. Immigration 1906 still lives only in the census, so bundling does not trivialise the fixture.

Two README errors corrected: Tekla is Wasyl's mother, not his wife, and the household IS retrievable via a collection-scoped search on surname "Capach". fixture.json's researcher_question was reworded to drop "where IN Ukraine" - no record gives a sub-place.

Run 2026-08-12_16-46-33 graded partial: f1 partial, f2 false, f3 true, f4 partial, proof quality 2. One root cause behind all three shortfalls - the agent swept Kapach/Kapatch/Kopach/Kapacz and never tried the C-initial "Capach", the only spelling carrying the census. Verified independently that the record is findable that way, so the fixture is solvable; the gap is being mined as a search-records unit test.

Compliance failed with five guardrail bypasses. 5 of 5 run logs carrying a compliance axis fail it. See #1263, already mining the same_person half.

## Summary

- **Normal tier.** Resolves #882 — adjudicates the `wasyl-kapach-birth` record-hint fixture as outcome (b): the identity is a true match, but the hint's birth year is wrong. The 1926 census's 1883 is an age estimate (recorded age 43, and it under-ages the whole household — Mary 1894 vs 1893, Natalka 1914 vs 1913); the headstone, Find a Grave memorial and 1948 obituary all give **1879**.
- Replaces the single draft finding with four: f1 born 1879 Ukraine, f2 immigrated 1906, f3 `avoid` 1883, f4 death/burial (bonus). Bundles `provided-documents/findagrave-kapach-captures.pdf` per spec §6.2 — mandatory, because Wasyl is absent from FamilySearch's Find a Grave index entirely (0 hits for `Kapach`, `Wasyl`, `William`, and the 1946–1950 death window), so no ark exists and `external_links_search` cannot reach him.
- Lands the first graded run (`2026-08-12_16-46-33`) with its `.ann.json`. Verdict **partial**, compliance **fail** — both explained below and both deliberate to land.

## Review guidance

**Start here:** `expected-findings.json`, finding **f3** — the `avoid` guard on birth year 1883. It is the only non-mechanical decision in the diff. Two things worth checking: (a) that it is safe under `apply_avoid_guard`, which I believe it is because the guard exempts `subject_person_ids` and the orchestrator populates that from `source_pid` (`L17M-W4Q`), so it cannot force-false against f1's correct 1879 Birth fact on the same person — verified by reading `e2e/judge.py` and `e2e/orchestrator.py`, not by running it; and (b) that preferring headstone + obituary over a census age estimate is the right genealogical call, which is the substantive judgement this whole PR rests on.

**Unsure about:** Whether **f1 should require the birthplace at all.** "Ukraine" appears *only* in the 1926 census — no bundled source states a birthplace — so f1's place component and f2 fail for the identical reason, and the fixture effectively penalises one bug twice. Splitting the place out, or dropping it to `required: false`, may measure more honestly. Related: **f3 is not currently a live test** — the agent avoided 1883 by never encountering it, since the only route to that value is the census it could not find. The label is correct per spec §3.4.1 (absent from the tree), but no restraint was demonstrated. f3 only starts measuring something once the transliteration gap is fixed. Both are calls I'd rather a genealogist made than assume.

## Plan

**Didn't change:** the `search-records` skill. This run's headline failure is a transliteration gap in it, and fixing it here was tempting — but it is a different skill's eval slot and folding it in would force a paid `make eval-skill SKILL=search-records` run inside a fixture PR. Deferred deliberately, not overlooked (see Follow-on). Also untouched, as the `record-hint` genre requires: `starting-tree.gedcomx.json`, `unstripped-tree.gedcomx.json`, `starting-research.json`.

**Acceptance check:** No test fails on `main` and passes here — this is fixture data, and I'd rather say so than dress something up. What is checkable: `eval\ValidateFixture.bat wasyl-kapach-birth` returns **2 WARN, 0 ERROR** (both warnings name `L17M-W4Q`, the fixture's own subject, which is expected for `record-hint` genre — nothing was stripped, so Wasyl legitimately stays in the starting tree and the matcher fires on subject name-overlap alone). And `check-e2e-fixtures` is satisfied: the run produced a final tree, so it is not exempt, and `run-2026-08-12_16-46-33.ann.json` ships beside it. On `main` the fixture also validates — it just asserts a birth year the bundled evidence contradicts, and nothing automated catches that.

**Deviated from the plan:** **Yes, materially.** This fixture was first resolved as a straight outcome (a) true match keeping the drafted 1883/1906 finding. Step 4 live debugging in Cowork then surfaced Wasyl's own Find a Grave memorial and the 1948 obituary, which put the birth year at 1879 — so the conclusion was re-planned to outcome (b) and the fixture re-resolved, overwriting a settled conclusion written earlier the same day. The README carries a dated changelog recording both passes. Two things in that second pass are worth flagging as deliberate rather than accidental: `fixture.json`'s **`researcher_question` was reworded** to drop "where **in** Ukraine" — a field beyond the `notes` the resolve skill normally edits, changed because no record found gives a sub-place, not even "Galicia", so as originally worded the question promised specificity the evidence cannot deliver and invited the agent to over-claim a sub-place inherited from Wasyl's mother; and the fixture moved from FamilySearch-recoverable to **bundled-capture-dependent**, which changes what it measures. Neither is a schema or auth change, so I did not read this as hitting the step-4 stop rule — flagging both here so the reviewer can disagree.

## Test plan

- [ ] `make test-all` — **not run: Windows, no equivalent exists (#1185).** `eval\RunTests.bat` is the paid per-skill eval, not this gate. Leaning on CI, which runs the same suites.
- [x] **N/A — no skill run-log snapshot changed.** This PR touches no file under `packages/engine/plugin/skills/`, no agent, no `eval/tests/unit/`, and no scenario or fixture any skill references. No `make eval-skill` run is owed.
- [x] **N/A — no skill `description` frontmatter or DO NOT clause changed.** No routing pair touched, so no negative test is owed on either side.

Senior-developer approval should not be required: every `.json` in this PR is under `eval/tests/` or `eval/runlogs/`, both on the exclusion list, and the remaining files are a `.md` and a `.pdf`. Flagging it in case the rule reads differently than I do — GitHub's merge button is the enforcement.

## Follow-on issues

- **`search-records` transliteration gap — filed, not folded.** The agent swept `Kapach / Kapatch / Kopach / Kapacz` across 15+ indexed searches, never tried the C-initial `Capach`, and concluded Wasyl is not name-indexed on FamilySearch. He is — under `Capach`, the only spelling carrying the 1926 census, and therefore the only source for both "Ukraine" and "1906". I verified independently that `record_search(surname="Capach", collectionId=3005862)` returns the full household, so the fixture is solvable and this is an agent gap, not a broken fixture. General rule for the test: for Cyrillic-origin surnames, sweep initial-consonant transliteration variants (К→K/C, В→W/V), not only internal spellings. **Reason for filing rather than folding:** it is a different skill's eval slot, and bundling it would force a second paid run inside a fixture PR. Label `genealogist`.
- **Guardrail bypass — filed, not folded.** The run wrote `exhaustive_declaration.declared=true`, two proof summaries, and three `person_evidence` links directly via `research_append`, without invoking `research-exhaustiveness`, `proof-conclusion`, or `same_person`. Right answers, wrong process. Not isolated: of the run logs in `eval/runlogs/e2e/` carrying a compliance axis, **5 of 5 fail it**, with the `same_person` bypass in all five (small sample — the axis is recent — but unanimous). #1263 is already mining the `same_person` half; the `research-exhaustiveness` and `proof-conclusion` bypasses appear uncovered. The durable fix is likely tool-side preconditions on the protected writes rather than more SKILL.md prose — the harness already computes these predicates to detect the violations after the fact. **Reasons for filing rather than folding:** it needs a different reviewer and skill (an MCP tool change found during fixture work), and the architectural call is the lead's. Label `developer`.
- **Dropped, not filed:** the `adds_daughter` half of the original hint is now answered — Anna (1911), Natalka (1914), Lucy (1921), all three documented in the bundled capture and recorded in the README. It is not a finding here because the research question is about birth and immigration. A separate daughters fixture is a candidate with no decision behind it, so per DEVELOPMENT.md it would be `icebox` at best; not filing it rather than adding a ticket nobody has chosen to do.


kofiatinka12 (approved):
This is a strong improvement to the Wasyl Kapach fixture. The changes correctly move the test away from simply accepting the original hint data and instead evaluate the evidence as a genealogist would. The updated expected findings appropriately distinguish the confirmed identity, the corrected 1879 birth year, the supported 1906 immigration fact, and the rejected 1883 census estimate.

I especially appreciate the addition of the avoid finding for 1883 and the clear documentation of the evidence conflict and reasoning. The README updates make the fixture much easier to understand, and documenting the transliteration/search failure mode will help prevent similar regressions.

The handling of uncertainty is also well done — unresolved details remain unresolved rather than being overstated. Overall, this makes the evaluation more accurate, realistic, and valuable. Great work.

johnmarkpeterbrown (approved):
Looks good!

Issue #1326, opened 2026-08-05
Add batchNumber to record_search — a measured ~99.999% filter three references recommend and the tool cannot send


chesworthrm:
**Touches:** packages/engine/mcp-server/src/types/record-search.ts, packages/engine/mcp-server/src/tools/record-search.ts, packages/engine/mcp-server/tests/tools/record-search.test.ts, docs/specs/record-search-tool-spec-v2.md (input table `:135`, schema mirror `:707`, mapping table `:828`), packages/engine/plugin/skills/search-records/references/search-strategy-levers.md, packages/engine/plugin/skills/search-records/references/collection-quirks.md, packages/engine/plugin/skills/search-records/references/place-date-mechanics.md

Split out of #1093, which annotated the gap rather than closing it. This is the **strong case** among the seven upstream constructs its probe found unreachable — the only one carrying both a measurement and a demand.

## The measurement

Live against `/service/search/hr/v2/personas`, 2026-08-04 (#1093 probe):

`q.batchNumber` works upstream and narrows **251,867 → 3** — a ~99.999% filter.

## The demand

Three reference files already recommend batch-number enumeration, and `record_search` has no parameter to send it:

- `search-strategy-levers.md:70`
- `collection-quirks.md:22`
- `place-date-mechanics.md:118`

`collection-quirks.md`’s **entire** compensation strategy for IGI parish registers was batch-number enumeration. Because `SKILL.md:151` makes that file a *required* read for a named collection, an agent searching an IGI collection was being handed a strategy it could not execute. #1093 annotated all three sites as unreachable and named the reachable alternatives, so nothing is actively broken — but the capability is still missing, and the annotations are what should be reverted when this lands.

## Decisions already taken (lead, 2026-08-04) — carry them forward

1. **It does not become an anchor.** A `batchNumber` search still requires `surname` or `recordCountry` alongside it. The anchor rule is unchanged, so none of its six assertion sites moves — see #1093 §2 "What does not change" for that list.
2. **The documented format claim is wrong.** The correct form is `M17288-6`. `place-date-mechanics.md:118` had it wrong and #1093 fixed the prose; the parameter needs to accept and validate that form.

## Note on test coverage

`tests/tools/record-search.test.ts:185` asserts only `/at least one anchor/`. Since decision 1 keeps `batchNumber` off the anchor list, that assertion stays valid — but it is loose enough that it would not catch a stale error message if anyone revisits the anchor wording while adding this.

PR #1591, opened 2026-08-12
record_search: batchNumber anchors on its own, and batch shape is open (#1326)

chesworthrm:
## Summary

- **Normal.** `record_search` now accepts `batchNumber` as an anchor, reversing the "it does not become an anchor" decision in #1326 — that call was made before anyone measured what pairing a batch with a second field does. A batch-only query is accepted upstream and returns just that batch's extraction; a batch is the cheapest filter the API takes, so the anchor rule's cost rationale never applied to it.
- **And it rejects `batchNumber` + `recordCountry` outright.** A *matching* country is inert (identical totals); a *mismatched* one silently returns **0** — the same signal these docs give for a *wrong batch*. So the field can only be useless or harmful on a batch search, and the old rule made the agent choose between an un-runnable search and a false negative that reads as an empty parish.
- Also corrects the batch **shape** documentation (open, not a closed rule — it had been wrong twice), removes a stale *(not reachable through `record_search`)* marker left over from #1328, and adds the missing `batchNumber → q.batchNumber` row the #1326 `Touches:` line named.

**Refs #1326 — deliberately not closing it.** This branch reverses *both* decisions recorded in that issue, and the issue should stay open so those reversals get read rather than buried by a close. Details are in [this comment](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1326#issuecomment-5273695246):

1. *"It does not become an anchor"* — reversed, on the measurement below.
2. *"the parameter needs to accept and validate that form"* — we accept and deliberately **do not** validate. Two successive shape rules (`exactly 6 digits after a letter prefix`, then `a letter prefix followed by digits`) both turned out too narrow, so the tool now passes a batch through byte-for-byte and the docs say never to reject or reformat one on shape.

## Review guidance

**Start here:** `validateInput` in `packages/engine/mcp-server/src/tools/record-search.ts` — two rules, both worth a look. `!input.batchNumber` in the anchor condition is the reversal; the `batchNumber && recordCountry` throw below it is the structural anchor for this PR's central claim.

That second one was added *after* review, and the reasoning is the part to check. The rule started life on five prose surfaces and none structural — which `docs/architecture.md` §3.1 ("anchor rules structurally") says is the shape of a rule that decays, from measurement: prose survives about three compactions, and the ranking doctrine fell 77% → 3% once compaction evicted the skill body. `SKILL.md`'s own parameter table still says every query needs `surname` or `recordCountry` three rows away, so the pairing is exactly what a half-remembered rule produces. **If you think a hard throw is too strong, that's the decision to push on** — the alternatives considered were silently dropping `recordCountry` when a batch is present, or a response note on a nil batch search, and the throw won because it is the only one the agent cannot miss.

`dev/probe-batch-anchor.ts` is the evidence. Run it: five legs, every verdict derived from its own run, `NOT MEASURED` rather than a direction when a leg errors.

**Unsure about:** the batch **prefix set**. `B` and the all-numeric form are confirmed live (`B01883-5`, `8317102`). `C`, `I`, `M` and the leading-zero case are documented on conversational report only, with no probe or catalog citation, in a spec that is otherwise careful about provenance — the same footing as the two shape claims that turned out wrong. The `treat no shape rule here as exhaustive` hedge is load-bearing until someone sources them. If you know where the prefix set is documented, say so and I'll cite it; if not, consider whether the enumeration should come out entirely.

## Plan

**Didn't change:** `PAGINATION_CAP`. A batch larger than 4999 still can't be walked end to end, so "enumerate one parish *exhaustively*" was dropped and `surname` is named as the partition instead. Raising the cap for batch searches is a separate argument nobody has made. Also left `imageGroupNumber` off the anchor list — it may well deserve the same treatment, but it wasn't measured here, and guessing is what produced this PR.

**Acceptance check:** four new assertions in `record-search.test.ts`, each proved to fail against a deliberately broken tree, not just asserted:

| Test | Fails when |
|---|---|
| `5a. accepts batchNumber as the only anchor` | the two-field anchor condition is restored |
| `5b. the anchor error names batchNumber` | same |
| `5c. rejects batchNumber combined with recordCountry` | the new guard is removed |
| `5d. allows batchNumber combined with surname` | the guard is widened to reject `surname` too |
| `21e … leading zero` | `add()` is made to strip leading zeros |

`5d` exists because the tempting wrong fix for `5c` is to reject every companion field, which would kill the one narrowing the docs still recommend.

**Deviated from the plan:** no PLAN.md — this began as a one-line doc fix and grew when the measurement contradicted the issue. Two decision reversals landed mid-flight; both are named above. An intermediate commit that told the agent to pair the batch with `recordCountry` — advice this branch later disproved — was squashed out rather than shipped, so no commit in the history carries it.

## Test plan

- [x] `scripts/test.sh` — passed (1956). Engine suite separately: 2197 passed, `tsc --noEmit` clean on both the src and dev projects. Prompt budget (warn-only): `search-records/SKILL.md` 52.2 KB → 52.8 KB (+591 B); it is already the largest prompt in the plugin, so the growth is worth a glance even though nothing fails on it.
- [x] **`make eval-skill SKILL=search-records` — NOT run, and required.** This PR changes `search-records/SKILL.md` and three of its `references/`, which is a run-log-snapshot change under the checkbox rule, and `runlogs` is red for exactly that. It is **not** behaviour-neutral (a new parameter, a changed anchor rule, a new rejection), so `eval-cosmetic-skip` is not the right escape. Worth knowing before spending the slot: **no test in `eval/tests/` exercises a batch number**, so the run gives no signal on the batch behaviour itself — its value is regression cover on the rest of the skill, which a SKILL.md edit genuinely can move. A senior should decide whether that is worth the paid run.
- [ ] N/A — no `description` frontmatter or DO NOT clause changed.

Senior approval is required regardless: this touches `src/tools/record-search.ts`, `tests/`, and a new `dev/` script, all outside the CODEOWNERS exclusions.

## Follow-on issues

**Folded in / filed instead:** One filed — **#1592**, `record_search` drops `UdeBatchNbr`.

#1592 is the wall behind this PR: the parameter now works, but nothing can tell the agent *which* batch to send. `collection-quirks.md` holds four illustrative batch numbers and no lookup, the skill has no web access, and no MCP tool returns one — while FamilySearch puts `UdeBatchNbr` on every persona and the tool discards the whole `gedcomx.fields` array. Surfacing it closes the loop the references already describe (name hit → batch → enumerate the parish), and only then is an eval test worth writing; today one would have to hard-code a batch into a fixture and verify a workflow the agent cannot perform.

*Why filed rather than folded:* **it is a different skill's eval slot, and bundling would force a second paid run.** This PR has already spent one `search-records` slot ($8.39, 26/26) because it changed SKILL.md; changing `record_search`'s output shape and then SKILL.md's provenance guidance would invalidate that run log and require a second run inside the same PR.

Everything else was folded in. Three review passes surfaced: the missing `batchNumber → q.batchNumber` mapping row; the spec's test-coverage table stopping at 21c; a test comment sourcing its assertion to the retired closed-format rule; `SKILL.md` still teaching the old anchor rule (the always-resident surface, while the reference I had fixed is only a conditional read); a self-contradicting bullet in `collection-quirks.md`; "exhaustively" colliding with the pagination cap; a `search-strategy-levers.md` row recommending a spouse-name-only search that the anchor rule rejects; the spec's "What changed from v1" still listing `collectionId` and `maritalStatus` as anchors, which `validateInput` has never accepted; and the prose-only rule that became the `5c` guard. All were in files this PR already had open. The prefix-set provenance is a question for the lead rather than a ticket, so it is under **Unsure about** above.


packages/engine/mcp-server/src/tools/record-search.ts
```diff
@@ -975,9 +997,9 @@ export const recordSearchToolSchema = {
...
+      batchNumber: { type: "string", description: "IGI batch number (e.g., `\"M01048-5\"`), the extraction batch behind a legacy parish register. A very strong filter and the canonical way to enumerate one parish: send it ALONE and it returns that batch's records, and adding a name searches within the batch. It anchors by itself — adding `recordCountry` is REJECTED by the tool, because a country that does not match the batch silently returns 0 (a batch number carries no country information, so there is nothing to guess it from). A nonexistent batch returns 0 rather than being ignored. Paging stops at `offset + count = 4999`, so a batch bigger than that cannot be walked end to end — partition it with `surname`, not by paging deeper. Shape varies: a batch may lead with a digit or with a letter (`B`, `C`, `I`, `M` seen), and may carry a trailing `-digit` — `C050761`, `M01048-5` and all-numeric batches all occur. Always pass it as a quoted string, keeping any leading zeros; pass it exactly as the source gives it, do not reject or reformat one on shape, and treat no shape rule here as exhaustive." },
```
jud-sdev: Non-blocking, doc accuracy. This lists B, C, I and M as batch prefixes that have been seen, but the probe only measures B and an all-numeric batch. C, I and M are not exercised anywhere in the committed evidence. Since the tool deliberately does not validate on shape and the text already tells the caller to accept any shape, an unverified prefix here is not a functional risk, but it states more than the probe demonstrates. Consider softening to what is confirmed, for example a batch may lead with a digit or a letter, with B and all-numeric confirmed live. The same phrase repeats in the spec, collection-quirks.md and place-date-mechanics.md, so it should move together.

packages/engine/mcp-server/src/tools/record-search.ts
```diff
@@ -149,10 +149,32 @@ export function applyAltNameAutoPair(input: RecordSearchInput): RecordSearchInpu
...
+
+  // Structural anchor for the rule above, not a second opinion on it.
+  // `recordCountry` on a batch search is inert at best and destructive at
+  // worst, and the destructive case is silent — a mismatched country returns 0,
+  // which is the same signal a wrong batch gives. Five prose surfaces say not to
+  // pair them; prose survives about three compactions (docs/architecture.md
+  // §3.1), and the pairing is exactly what a half-remembered "every query needs
+  // surname or recordCountry" produces. Rejecting costs nothing because a
+  // MATCHING country changes no result — measured, see probe-batch-anchor.ts.
+  if (input.batchNumber && input.recordCountry) {
```
jud-sdev: Non-blocking, minor. This reject guards recordCountry, but recordSubdivision is the same class of record-jurisdiction filter and carries the same silent-zero risk on a batch. The invariant still holds overall: batch plus subdivision plus country is caught here, and batch plus subdivision with no country is caught by the recordSubdivision requires recordCountry rule a few lines down. So there is no hole. The only rough edge is that the requires-recordCountry message points the model toward adding the one field the batch rule forbids, which forces a second round-trip. Optional: name recordSubdivision in this reject too so that case gets the right guidance in one step. Safe to leave as is.

jud-sdev (approved):
Approve. The change is correct and the tests are stronger than what they replace. I traced validateInput by hand across every case: a batch anchors alone, batch plus country hits the new reject, batch plus surname passes, and a query with no anchor still throws and now names batchNumber. The spec, SKILL.md and the three references all match the code, including the corrected note in place-date-mechanics that batch numbers are reachable through record_search. The probe is well built and computes every verdict from its own run.

Two small notes below, both non-blocking. Take them or leave them. One caveat on the approval: I verified the logic by tracing it and by CI being green, but I did not re-run the live probe myself, so the measurement claims behind the rule rest on your probe run, which I read and found correctly built.

packages/engine/mcp-server/src/tools/record-search.ts
chesworthrm: Taken — and it turned up a distinction worth recording.

`B` and the all-numeric form are confirmed by the probe, as you say. `M01048-5` is also attested, just not by my probe: the live check at `record-search.ts:327` from #1328 (2026-08-11) covers it. `C` and `I` are the two with nothing committed behind them.

So rather than softening to `B` + all-numeric, `32d39e3c` names the three attested examples and drops the enumeration entirely: *"a batch may lead with a digit or with a letter, and may carry a trailing `-digit`. Attested live: `B01883-5`, `M01048-5`, and the all-numeric `8317102`."* The "treat no shape rule here as exhaustive" hedge stays, which is what keeps the unenumerated cases safe given the tool validates nothing.

Moved across all five copies together, as you flagged. The spec also now records that `C050761` is an inherited illustration rather than an attestation, so the next reader doesn't have to re-derive that.

packages/engine/mcp-server/src/tools/record-search.ts
chesworthrm: Taken. You're right that there was no hole — the round-trip was the whole cost, and it's a real one: the `recordSubdivision requires recordCountry` error pushes the model toward exactly the field the batch rule then rejects, so the agent had to be told no twice to arrive at the right call.

`32d39e3c` widens the reject to `batchNumber && (recordCountry || recordSubdivision)` and renames it to the general case: *"do not combine batchNumber with recordCountry or recordSubdivision: a batch anchors on its own, and a record-jurisdiction filter that does not match the batch silently returns 0…"* One error, right guidance, one step.

Test 5e added and proved to fail — narrowed the condition back to `recordCountry` only and watched 5e fail while 5c stayed green, so it's guarding the new half specifically rather than riding on the old one. Spec error table and test matrix updated to match.

johnmarkpeterbrown (changes requested):
Dear Chesworth,
Reviewed this in full — the code half is solid and I have **no blocking findings**. The guard is correct, the new tests genuinely bite (I confirmed 5a/5c/5e all fail against a broken tree), tsc is clean on both projects, and the full engine suite is green. The spec, SKILL.md, and all three references line up with the code, and jud-sdev's two notes are both resolved by `32d39e3c`. Nice work on the measurement-first approach.

One thing gates merge: **`runlogs` is red because the committed run log is stale.** Five snapshot files differ from the working tree — two are your own late reference edits in `32d39e3c` (after the run was captured), and three drifted in through the main merge. The check also flags that the log was scored against an older judge prompt.

The clean fix is a fresh run, which resolves both the staleness and the judge-prompt-hash warning in one shot. It has to run **from the PR worktree/branch** so the snapshot matches this tree, and the worktree needs a build first:

```bash
npm run build   # in packages/engine/mcp-server
uv run python eval/harness/run_tests.py --skill search-records
```

Then commit the new run log + `.ann.json` and push. I'd steer away from `eval-cosmetic-skip` — you're right that this isn't behavior-neutral, so it's the wrong escape.

Heads-up on cost so it's a conscious call: no `eval/tests/` case exercises a batch number, so this run buys regression cover on the rest of the skill, not signal on the batch behavior itself (that's on the unit tests, and #1592 tracks the eval gap). Your call whether the slot's worth it, but as things stand `runlogs` won't go green without it.

johnmarkpeterbrown (approved):
Looks good!

Issue #1596, opened 2026-08-13
elena-asmundsdotter-origin: the Mantalslängder ARE on FamilySearch — fixture caveat is half wrong, plus two unexamined routes


Gennecis:
**IMPORTANT**: If you do this, coordinate with **issue #572** — it is In Progress on the same
`eval/tests/e2e/elena-asmundsdotter-origin/` fixture, and this issue was filed out of that run.
Whoever holds issue #572 is the natural taker.

**Touches:** eval/tests/e2e/elena-asmundsdotter-origin/README.md, eval/tests/e2e/elena-asmundsdotter-origin/fixture.json

The fixture's "Recoverability caveat" says:

> the tax lists and catechism records that carry this proof are browse-only and largely
> on SVAR/ArkivDigital, not name-indexed on FamilySearch

`fixture.json`'s `notes` says the same. **The tax-list half is wrong.** Mantalslängder
for Harjager härad are on FamilySearch as browse-only image volumes:

| | |
|---|---|
| groups | `008978368`, `008978369`, `008978374`, `008978375` (`008978369` alone is 720 images) |
| Harjager years | **1670, 1681, 1700, 1702, 1705, 1708, 1714** (then 1792, 1819, 1850, 1856) |
| indexed | 0% — browse-only, so "not name-indexed" is correct |

The fixture's own research window is 1680–1711, and FamilySearch holds **1681, 1700,
1702, 1705, 1708** inside it. Elna was born c. 1681.

They were invisible to `volume_search` because it discarded every `title:`-prefixed
`recordType` as a placeholder — fixed in the PR that filed this issue.

```sh
cd packages/engine/mcp-server
npx tsx dev/try-volume-search.ts --standardPlace "Harjager, Malmöhus, Sweden" \
    --startYear 1650 --endYear 1720
```

## The catechism half stands

No Katekismilängd appears at härad level in that window — only `Konfirmationslängd`
1846–1858, far too late for the 1696/1699/1701 lists the fixture cites. Worth a
parish-level and county-level check before concluding either way.

## Two routes nobody has looked at

Also surfaced once record types became visible, and **neither is mentioned in the
fixture, the case study, or any run**:

- **Probate**, `007118856` (439 + 113 images), Harjager **1688–1816**
- **Häradsrätt court records**, `008733807` (839 images), Harjager **1697–1714**

Probate is the interesting one. Swedish *bouppteckningar* enumerate heirs, and Asmund
[...]

Issue #1594, opened 2026-08-13
image_transcribe on hard scans: fabricated record types, corrupted patronymics, and 26% of calls lost to reachability


Gennecis:
**Touches:** packages/engine/mcp-server/src/tools/image-transcribe.ts, packages/engine/mcp-server/src/utils/http.ts, docs/specs/image-transcribe-tool-spec.md

**Do after PR #1598** — it adds the run log this body cites, and touches the same fixture.

`image_transcribe` (default model `qwen/qwen3-vl-235b-a22b-instruct`) fails in two
distinct ways on early-modern Swedish hands. Both were checked against the same scans
read directly, so the ground truth is not inferred.

## 1. Fabricates a record type and year on non-vital pages

| image | returned | actually is |
|---|---|---|
| `004523021_00100` | `"Pro A: 1673."` + heading **`"Född."`** (Born) | church accounts headed *"Wästra Carleby Kyrckio Rächningh Pro A° 1693"*, Debet/Credit — **1693**, no births heading |
| `004523021_00190` | heading **`"Födde. Jahr 1707"`** | Credit/Debet ledger, years 1713–1717 |

This is not degraded OCR: it invents a vital-record heading on an accounts page. An
agent reading it is told the wrong volume is the right one.

## 2. Corrupts identity-critical fields on vital pages — nondeterministically

Controlled test on `004523018_00150` (a Västra Karaby register page read directly
first). Structure survives: `Anno 1762`, `föddes`, entry boundaries and 5 of 6 dates
correct, **no fabricated heading**. What corrupts is exactly what decides identity:

- **Farm name** `Henckelstorp` → `"Hornekelltorp"` in one call, `"Honekelkorp"` in
  another
- **Patronymic** `Andersson` → `"Olsensfon"` in one call, absent in another
- `Barsebäck` → `"Dansebäck"`

The corruption **varies per call on the same image**, so the failure mode is the class
("identity-critical fields corrupted"), not any specific string.

Why this matters beyond quality: `docs/gps-research-flow.md` treats a patronymic
mismatch as *"a conflict in a core identifier, not a spelling variant — a different
patronymic names a different father."* The OCR therefore manufactures that class of
conflict silently, while reading cleanly enough elsewhere on the page to look
trustworthy. A corrupted farm name also defeats any gazetteer or string match.

**Run the no-hint control before proposing any remedy.** All three reads above
passed a `lookingFor` ("record type heading, year"; "christening entries, year,
[...]

Issue #1593, opened 2026-08-13
image_read refuses scans over 700 KB, closing the only accurate-vision path (supersedes #656)


Gennecis:
> **Reviewed 2026-08-13 before junior handoff — routed to a senior, stays in Backlog.**
> The code claim is exact and the corpus claim nearly reproduces, but this ceiling was
> already decided (lead, 2026-07-24: "leave the cap as-is") with issue #1130's live
> measurement as the stated pre-req, and issue #1130 says in as many words not to
> re-derive it. The three approaches below are not interchangeable choices: downscale
> commits a native image dependency to the cross-platform `.mcpb`, tiling changes the
> tool contract, and "return a file path" needs a tool grant on a plugin agent that today
> holds `image_read` and nothing else. Correctness rests on something nothing checks.

**Touches:** packages/engine/mcp-server/src/tools/image-read.ts, docs/specs/image-read-spec.md

`image_read` refuses any scan over `MAX_INLINE_IMAGE_BYTES = 700_000`
(`packages/engine/mcp-server/src/tools/image-read.ts:14`) and its error text tells the
caller to use `image_transcribe` instead. That closes the only accurate-vision path:
the `image-reader-opus` agent exists specifically for "a difficult hand,
Kurrentschrift" and reads exclusively through `image_read`.

## Measured across the committed e2e corpus

```sh
python3 -c "
import json,glob,re
calls=0; runs=set(); imgs=set(); le=0
for f in glob.glob('eval/runlogs/e2e/*/run-*.json'):
    if f.endswith('.ann.json'): continue
    d=json.load(open(f,encoding='utf-8'))
    for t in d.get('tool_calls') or []:
        s=str(t.get('response_summary') or '')
        if 'too large to return inline' in s:
            calls+=1; runs.add(f)
            m=re.search(r'image (\S+) is ([0-9.]+) MB',s)
            if m:
                imgs.add(m.group(1))
                if float(m.group(2))<=1.1: le+=1
print(calls,len(runs),len(imgs),le)"
# -> 55 24 30 25    (measured 2026-08-13 against origin/main 28a3e367)
```

**55 refusal calls across 24 runs, 30 distinct images**, sizes 0.8–2.7 MB.
**25 of the 55 calls are ≤1.1 MB** — barely over the cap. That is calls, not distinct
images: several scans were retried. Agents record them in narration as `NOT READ: <id>`.

[...]

Issue #656, opened 2026-07-13
image_read result-size cap: investigate alternatives


DallanQ:
Closing-report T13: the image-escalation *behavior* is fixed, but image_read's ~1 MB MCP transport cap failed 7+ read attempts across 4 of the 5 re-run scenarios — large FamilySearch record images overflow the tool-result cap and the read errors out, blocking OCR/transcription entirely. Distinct from "image_read callable by the main session" (that is tool-*policy* — who may call it) and from "Oversize search-tool results" (scoped to external_links_search/fulltext_search). image_read needs response shaping of its own: downscale to fit the cap, tile/paginate, or return a file path the reader loads. Ref: docs/plan/record-extraction-consolidation-closing-report.md §1 (T13 row).

DallanQ:
See docs/specs/image-transcribe-tool-spec.md in PR #655

PR #1598, opened 2026-08-13
volume_search: keep the record type behind a `title:` prefix; first e2e run for elena-asmundsdotter-origin

Gennecis:
## Summary

`volume_search` was discarding 97.5% of the record-type labels FamilySearch returns.
`RECORD_TYPE_PLACEHOLDER_RE = /^(concept-id|title):/` treated a `title:` prefix as an
internal placeholder; it is a *provenance* marker (the type came from the volume's title
rather than the concept taxonomy) and the value behind it is a real record type. Over
Harjager härad the filter dropped **785 of 805 labels**, including **every**
`title:Taxation`, `title:Census` and `title:Probate records` value.

`^concept-id:` is still dropped — those are opaque ids, and none occurred in any live
sample. `^title:\s*` is now stripped and the value kept on **both `recordType` and
`dateRange`** — FamilySearch puts the same prefix on `datesOrig`, independently of the
record type; the field is omitted when nothing remains.

**Why it matters beyond one fixture:** the discarded labels are how a researcher tells a
parish register from a church-accounts book, a tax list from a probate volume. Live on
Harjager the tool now returns **805 record types instead of 20**.

This was found while working #572. The stripping is why the Swedish *mantalslängder*
were invisible to `volume_search` — FamilySearch holds them browse-only for Harjager in
1670/1681/1700/1702/1705/1708/1714, and they carry the parentage evidence that fixture
is built on. The fixture's own README claimed they were "largely on SVAR/ArkivDigital,
not name-indexed on FamilySearch"; that is half wrong (#1596).

Also lands the fixture's **first ever e2e run** (it was 1 of 20 fixtures with zero
committed runs), its blind grade, and a dated behaviour line in its README.

## Review guidance

**Start here:** `packages/engine/mcp-server/src/tools/volume-search.ts` — the two
constants and `mapCoverage`. The judgement call is that `title:` is provenance, not
junk; the comment above the constants carries the measurement behind it.

The evidence, if you want to re-run it:

```sh
cd packages/engine/mcp-server
npx tsx dev/try-volume-search.ts --standardPlace "Harjager, Malmöhus, Sweden" \
    --startYear 1650 --endYear 1720 | grep -o '"recordType": "[^"]*"' | sort | uniq -c
# before: 20 labels, no Taxation/Census/Probate
# after:  805 labels — Church records 637, Court records 49, Taxation 48, Census 48, Probate records 3
```

The spec's behaviour-table row 14 changes name (*Placeholder filtering* →
*Record-type normalization*) because the behaviour it describes changed.

The run log and `.ann.json` are the genealogist half of this PR; the run scored **0 of
6**, which is an honest measurement, not a regression — nothing in this PR was expected
to make it pass.

## Test plan

- `npx vitest run tests/tools/volume-search.test.ts` — **27/27**. Test 14 was rewritten
  and **proven to fail against the pre-fix code** (`expect(coverages[1].recordType).toBe("Taxation")`)
  before the fix was applied.
- `make engine-test` — 91 files, 2198 tests, typecheck clean.
- `make e2e-validate TEST=elena-asmundsdotter-origin` — `OK … schema valid, all findings
  appear stripped`.
- Live verification against the FamilySearch API, above.
- `make test-all` — **red, pre-existing and unrelated**: `apps/electron`
  `feedback.test.ts` *"drops the largest files…"* times out at 5000 ms. 3/3 failures
  inside the parallel gate vs 3/3 passes standalone, twice at `87ba1657` and once after
  rebasing to `97bccc60`; this change lives outside the pnpm workspace and cannot reach
  `apps/electron`. Filed as #1597.

## Follow-on issues

| # | What | Label |
|---|---|---|
| #1593 | `image_read` refuses scans >700 KB, closing the accurate-vision path; supersedes #656 (59 refusal calls / 25 runs / 32 images) | `developer` |
| #1594 | `image_transcribe` fabricates record types on ledger pages and corrupts patronymics/place names on register pages | `developer` |
| #1595 | Classify the 26% OpenRouter failure rate on `image_transcribe` | `developer` |
| #1596 | The fixture's Recoverability caveat is half wrong; probate 1688–1816 and häradsrätt 1697–1714 unexamined | `genealogist` |
| #1597 | `make test-all` red locally on the electron feedback-zip timeout | `developer` |

All five carry the same step-4 exemption: *the fix needs a different reviewer or skill —
a code fix surfaced during a genealogist's fixture work.*

Per #572 (*"do not edit the fixture to make it pass; if the fixture itself turns out to
be wrong, that is a separate finding and a separate issue"*), no expected finding,
starting-state file or caveat claim was touched — the README gains only a dated
behaviour line, and `git diff` on it is additions only.


chesworthrm (changes requested):
The `title:` reading is right, and I confirmed it against the live API rather than taking the PR body's word for it: 785 of 833 record types on Harjager carry the prefix, and every one of them is a real record type. Good find, and the rewritten test 14 genuinely fails against the pre-fix code — I checked that too.

One thing needs to change before this merges.

## The same `title:` prefix is on `dateRange`, and this PR leaves it there

The line directly above the one you fixed copies `datesOrig` straight through with no stripping. FamilySearch puts the same provenance prefix on that field. Live on Harjager, 30 of 829 date ranges reach the agent malformed.

After this PR, a single coverage object reads:

```json
{ "place": "Harjager, Malmöhus, Sweden",
  "dateRange": "title:1683-1700",
  "recordType": "Court records" }
```

The type is clean. The date range beside it still carries the prefix.

That is a worse state than before your change. Previously `recordType` was simply absent, so nothing looked half-normalized. Now one field is fixed and its neighbour is visibly not.

This is not Sweden-specific. Wayne, Ohio returns three prefixed date ranges and **zero** prefixed record types, so this field is affected independently of the one you fixed.

It also matters for this fixture in particular. Volume `008768877` is the häradsrätt court series for 1683–1700 — the record class #1596 names as an unexamined route. An agent cannot filter by year on `title:1683-1700`.

The probe in your PR body greps `"recordType"` only. That is why this did not surface.

### The changes

Rename the constant first, since it is no longer record-type-specific:

```
You wrote:      const RECORD_TYPE_TITLE_PREFIX_RE = /^title:\s*/;
Change it to:   const TITLE_PREFIX_RE = /^title:\s*/;
```

```
You wrote:      const recordType = rawType.replace(RECORD_TYPE_TITLE_PREFIX_RE, "").trim();
Change it to:   const recordType = rawType.replace(TITLE_PREFIX_RE, "").trim();
```

Then strip the date range the same way:

```
You wrote:      if (entry.datesOrig != null) coverage.dateRange = entry.datesOrig;
Change it to:   const rawDates = entry.datesOrig;
                if (rawDates != null) {
                  const dateRange = rawDates.replace(TITLE_PREFIX_RE, "").trim();
                  if (dateRange) coverage.dateRange = dateRange;
                }
```

Add a case to test 14: a coverage with `datesOrig: "title:1683-1700"`, asserting `dateRange` comes back as `"1683-1700"`.

Three spec lines follow from it:

```
You wrote:      | `datesOrig` | string? | `dateRange` (when present, e.g. `"1726–1812"`) |
Change it to:   | `datesOrig` | string? | `dateRange` (when present, e.g. `"1726–1812"`; a `title:` prefix is stripped) |
```

```
You wrote:      | `dateRange` | string? | Human-readable date range (from `datesOrig`, e.g., `"1726–1812"`), when present. |
Change it to:   | `dateRange` | string? | Human-readable date range (from `datesOrig`, e.g., `"1726–1812"`), when present. A `^title:` prefix is stripped and the value kept, as on `recordType`. Omitted if nothing remains after stripping. |
```

```
You wrote:         - `dateRange` ← `coverage.datesOrig` (when present)
Change it to:      - `dateRange` ← `coverage.datesOrig` (when present), with a leading
                     `^title:\s*` stripped and the result trimmed; omitted when the
                     result is empty
```

One clause in the code comment is contradicted by the same probe — in Wayne, Ohio the only `title:`-prefixed values are date ranges, not types:

```
You wrote:      // every `title:`-prefixed value was a real type (Taxation, Census, Probate
Change it to:   // every `title:`-prefixed *record type* was a real type (Taxation, Census, Probate
```

## Everything else checked out

I ran the suites against the merge with `main`, not just your head — you are 181 commits behind, and the merge is clean.

- vitest 2270 passed, `tsc` clean, harness pytest 2075 passed
- `e2e.calibrate_judge --dry-run` accepts the new annotation, 0 errors across 139
- fixture validates, e2e grading gate OK
- No other consumer of `recordTypeOrig` anywhere, and no eval fixture encodes the pre-fix output
- The README's "appears 0 times in its trace" claim is true — every hit in the run log is in `judge_output`, where the judge restates the expected findings
- The 0/6 human grade against the judge's `partial` on f4 is a legitimate recorded disagreement

The harness suite fails once on a fresh worktree with `research_query: build not found`. That is the missing engine build, not a regression — `npm run build` clears it.

Gennecis:
Fixed — all six edits applied, and the constant renamed as you set out.

I verified your findings against the live API rather than taking them on trust, and the
substance holds: `datesOrig` carries the same prefix, and your example volume is exactly
right — `008768877`'s Harjager coverage is `"title:1683-1700"` / `"title:Court records"`.

Your independence point is right too, though not via the Wayne example as written. I found
three genuine cases of a prefixed date beside a *clean* type:

- `"title:1867-1908"` / `"Death records"`
- `"title:1813-1951"` / `"Marriage index"`
- `"title:1796-1837"` / `"Birth and Baptism Records"`

My original sample contained none of these, which is why I missed the field entirely.

Three counts didn't reproduce, and I'd rather flag that than quietly substitute my own.
Full pagination over 1500–1950:

| | your review | measured |
|---|---|---|
| Harjager prefixed dates | 30 of 829 | **4 of 1889** |
| Wayne prefixed record types | zero | **48** |
| Harjager prefixed types | 785 of 833 | **1762 of 1889** |

Most likely a different date window or pagination depth — my own earlier 785/805 came from
a single 100-group page, so I had been undercounting as well. The commit uses the
full-pagination figures with the window stated. None of it changes the fix, and the
Wayne row is the only one that touched an argument rather than a magnitude.

You were right that my probe grepped `"recordType"` only; that is exactly why this got
past me. The comment clause now says *record type* rather than *value*.

Verification on the new commit:

- `expected 'title:1683-1700' to be '1683-1700'` — the added assertion fails against the
  pre-fix code, checked by reverting just the date stripping.
- `make engine-test` — 94 files, 2277 tests, typecheck clean, on the tree with
  `origin/main` (c7747f2e) merged in.
- Live: zero `title:`-prefixed `dateRange` values remain for Wayne, Ohio.

`make test-all` is still red on `apps/electron`'s feedback-zip timeout — pre-existing and
unrelated, filed as #1597.

chesworthrm (approved):
Approved. All six edits landed, and the fix is right.

I checked the part that mattered rather than taking it on trust. Reverting **only** the date stripping — leaving your record-type fix in place — fails with `expected 'title:1683-1700' to be '1683-1700'`. Reverting `mapCoverage` wholesale fails with `expected undefined to be 'Taxation'`. Both halves stand on their own assertion, which is what I wanted to see.

The independence argument in the comment is now the right one. My "Wayne returns zero prefixed record types" was wrong, and you were right to remeasure instead of writing it in. The prefixed-date-beside-a-clean-type cases are what actually carry the point, and that is what the comment cites.

Thanks for flagging the three counts that did not reproduce rather than quietly substituting your own. That is the right instinct. I could not settle the numbers myself — my probe returned different figures again, but I have elevated FamilySearch permissions, so my sample is not comparable to yours. Nothing there changes the fix, and the counts are commentary, not behaviour.

What I verified:

- Engine 94 files / 2277 tests, `tsc` clean
- Harness pytest 2075 pass. The `research_query: build not found` failure on a fresh worktree is the missing engine build, not a regression — `npm run build` clears it
- Control plane 157 pass
- Fixture validates; `calibrate_judge --dry-run` accepts the annotation, 141 graded, 0 errors
- No other consumer of `recordTypeOrig` or `datesOrig` in the repo, and no skill reads these output fields

One thing to do, but not here. The committed run also carries `compliance: fail` — a `proof_summaries` entry written with `proof-conclusion` never invoked, plus six shadow violations on `research_append`. That is honest data from an honest failing run and no reason to touch this PR. It needs no new issue either: #1335 already covers doctrinal skills being bypassed while their effects are still written. Please add it as a comment there while the run is fresh.

Merge still needs `senior-genealogists`. The `.ann.json` grading is their call, and their request is outstanding.

johnmarkpeterbrown (changes requested):
First, chesworthrm asked whether `fromDate` and `toDate` should be used instead of `datesOrig`, and that question is still open. It has a real answer: those fields are present on every coverage entry and never carry the prefix. On the same entry your fix repairs, `datesOrig` is `"title:1683-1700"` while `fromdateString` is already `"1683-01-01T00:00:00"`. Please answer it on the thread — either "we should switch, filing it" or "here is why the display string is the right output" — so it closes rather than lapses.
>
> Second, the PR body only describes the `recordType` half.
>
> ```
> You wrote:      `^concept-id:` is still dropped — those are opaque ids, and none occurred in any live
>                 sample. `^title:\s*` is now stripped and the value kept; the field is omitted when
>                 nothing remains.
> Change it to:   `^concept-id:` is still dropped — those are opaque ids, and none occurred in any live
>                 sample. `^title:\s*` is now stripped and the value kept on **both `recordType` and
>                 `dateRange`** — FamilySearch puts the same prefix on `datesOrig`, independently of the
>                 record type; the field is omitted when nothing remains.
> ```

johnmarkpeterbrown (approved):
looks good!

Gennecis:
## Answering the open question: should we use `fromDate`/`toDate` instead of `datesOrig`?

**Yes, partly — and the case is stronger than avoiding the prefix.** Measured before
answering, full pagination, 4 pages per place, window 1500–1950:

| | Harjager (1698 coverages) | Wayne, Ohio (463) |
|---|---|---|
| `datesOrig` present | 1697 | **335** |
| `fromDate`/`toDate` present | **1698** | **462** |
| `fromdateString`/`todateString` present | **1698** | **462** |
| `title:` prefix ever on the pair | **0** | **0** |
| `datesOrig` holding multiple ranges (`,` / `;`) | 0 | 0 |

Three findings:

1. **The structured pair exists where `datesOrig` does not.** Wayne: 462 vs 335. We are
   emitting **no date at all on ~27% of Wayne coverages** where the API supplied one.
   That is a larger defect than the prefix this PR fixed, and the same class — dropping
   data the API gave us.
2. **The pair never carries the prefix** — 0 occurrences in 2161 coverages.
3. **`datesOrig` is inconsistent display text**, mixing en-dashes and hyphens in the same
   place: `"1882–1896"` beside `"1870-1880"`.

On the exact volume this PR repairs:

```
008768877   datesOrig="title:1683-1700"
            fromdateString="1683-01-01T00:00:00"   todateString="1700-12-31T23:59:59.999"
```

Note the raw `fromDate`/`toDate` are epoch **milliseconds and negative** for pre-1970
(`-8899027200000`), so the usable fields are `fromdateString`/`todateString`, not the
numeric pair.

### What we are doing about it

Not a switch — an addition, in two parts:

- **Keep `dateRange`** as the human-readable string. It is the archival display text and
  what a citation shows. But **derive it from the structured pair when `datesOrig` is
  absent**, which closes the 27% gap.
- **Add `startYear` / `endYear`** integers from `fromdateString`/`todateString`, so
  coverage becomes machine-comparable. This matters for the follow-up work on record-type
  filtering: the agent needs to narrow by year, and it cannot compare `"1882–1896"`.

Both land in the `volume_search` spec revision now in progress, with the implementation
following that spec's review — not as a separate ticket, since they are the same contract.

The `title:` stripping in this PR remains correct and still needed; it turns out to have
been treating a symptom of `datesOrig` being the sole input.

PR #1601, opened 2026-08-13
fix: mcp fixture hygiene — de-duplicate collection id and placeholder ARK

Paaboat:
Closes #1405

Fix 1 — collections-search-kentucky.json
The last entry used id 1401638, which collides with other fixtures in the corpus. Replaced with id 1786457 ("United States, Census, 1840") — a real FamilySearch collection confirmed live, not present anywhere else in eval/fixtures/mcp/, with accurate stats: 2,566,652 records / 2,567,892 persons / 196,897 images.

Fix 2—same-person-flynn-strong.json
The candidate Ark for person I1 was set to a real FamilySearch ARK, but I1 has no ARK field in the flynn-record-matching scenario tree. Per the same-person tool spec, candidateArk must be MMMM-MMM when the focus person has no ark. Replaced the real ARK with the placeholder.


florencemashipei (changes requested):
The two stated fixes are correct and well-scoped:

- `collections-search-kentucky.json`: the colliding id `1417683` (already used by 6 other fixtures for "United States Census, 1880") is correctly changed to `3693000`, matching issue #1405.
- `same-person-flynn-strong.json`: `candidateArk` is correctly changed from a borrowed real ARK to the spec-required `MMMM-MMM` placeholder for an ARK-less focus person, also per #1405.

**Blocking issue: this PR also silently changes `eval/tests/e2e/ferber-grandparents/README.md` and `expected-findings.json`**, renaming the maternal grandmother from "Mary Kramer (Ohio, unknown)" to "Anna Kramer (1845, Darmstadt, Hesse, Germany)." This is out of scope for #1405 (a pure fixture-id-hygiene issue) and isn't mentioned anywhere in this PR's body or the fixture's own "Notes for reviewers" section.

More importantly, I don't think it's correct:

- This fixture's own `reference-project/tree.gedcomx.json` — the grounding data this fixture was distilled from — still names her **Mary Kramer**, born Ohio (person I7). This PR doesn't touch it, so the fixture's benchmark-contract files now contradict its own documented source data.
- "Anna Kramer" traces to a *different*, sibling fixture: `eval/tests/e2e/william-ferber-parents/README.md`. That fixture's author already investigated this exact Becker/Kramer identity question and documented it as a **genuinely unresolved conflict** — FamilySearch shows "John Becker + Mary Kramer" (sparse, 2 sources) vs. a possible-duplicate "John Becker + Anna Kramer" (23 sources), with six records pulled and only a "4-to-2 lean" toward Anna and no source explaining the discrepancy. That author's explicit conclusion was that it **doesn't qualify as fixture-worthy** because it isn't cleanly recoverable from records, and that resolving it would need "a second, harder fixture... authored separately" — not folding into an existing one.
- This PR does the opposite of that guidance, and the new `supporting_sources` text ("the compiled profile... list[s] Anna Kramer as mother") is factually false against this fixture's own tree data, which still says Mary Kramer.

**Ask:** please drop the `ferber-grandparents` changes from this PR — they're unrelated to #1405 and should revert to Mary Kramer. If there's genuine new research justifying "Anna Kramer," that deserves its own PR that also updates `reference-project/tree.gedcomx.json` and explains why it now clears the bar the sibling fixture's author explicitly said this question didn't meet.

florencemashipei (approved):
Confirmed — the ferber-grandparents changes are fully reverted and the diff is now scoped to exactly the two intended hygiene fixes (collection id de-dup, placeholder ARK), both verified correct against #1405. Approving.

Cia-3 (approved):
Ran the suites fresh against a test-merge with main — vitest (2219 tests), tsc, and harness pytest (2063 passed) all clean. But found two real problems in the actual fix:

collections-search-kentucky.json — the new id (3693000) doesn't match its own entry. The PR body says it represents "Kentucky Wills and Probate Records," but the title/date range/record count were never updated — they still describe a census collection. I checked the live FamilySearch catalog directly and that id doesn't exist at all. Please make the id and the data agree, and double-check it doesn't collide with 1875188, already used elsewhere in this same file for the real Kentucky probate collection.

same-person-flynn-strong.json — the placeholder ARK doesn't match the documented format. The spec says queryArk and candidateArk must either both be full URLs or both be placeholders, in the same shape. This fixture's queryArk is a full URL, but candidateArk was set to a bare "MMMM-MMM" instead of the canonical ark:/61903/... form. Should be something like "https://www.familysearch.org/ark:/61903/4:1:MMMM-MMM".

Separately (not blocking this PR): I confirmed this fixture edit silently makes person-evidence's committed run log stale, but the runlogs CI check can't detect it — its touched-skill logic doesn't scan eval/fixtures/mcp/ paths at all. Worth its own issue.

EdmondOware (changes requested):
Holding this one. Cia-3's review already found two real problems, but they're still present in the current code and the review is marked Approved, so I want to flag them explicitly before this merges.

collections-search-kentucky.json: the new entry (id 3693000) still has title "United States Census, 1850-1880 (Kentucky)" and the original census dateRange/recordCount. The PR description says this id now represents "Kentucky Wills and Probate Records," but that was never applied to the data. Issue #1405's actual requirement (see Dallan's last comment on the issue) is just a distinct id with a title that matches what it represents, whatever that is. This entry also duplicates in spirit with the existing 1875188 probate entry already in this file. Please make the id, title, dateRange, and recordCount agree with each other, and confirm the id doesn't collide with anything else in the corpus.

same-person-flynn-strong.json: per docs/specs/same-person-tool-spec.md, queryArk and candidateArk must be in the same format, both real ark:/61903/... strings or both placeholders in that same shape. queryArk here is "https://www.familysearch.org/ark:/61903/1:1:MXHY-TP4", but candidateArk is the bare string "MMMM-MMM" with no ark path. Needs to match the canonical placeholder shape, not a bare string.

EdmondOware (changes requested):
Thanks for the quick turnaround. same-person-flynn-strong.json is correct now, candidateArk matches queryArk's format exactly.

collections-search-kentucky.json still isn't safe to merge, though. The entry is internally consistent now (id, title, dateRange, recordCount all agree with each other), but the new id 1401638 already means two different things elsewhere in the corpus, independent of this PR:

- collections-search-pennsylvania.json, collections-search-schuylkill.json, collection-read-by-id.json, and two record-search fixtures all use id 1401638 for "Pennsylvania, Church and Town Records, 1708-1985"
- record-read-patrick-flynn-mxhy-tp4.json, record-read-patrick-flynn-cflt-9k2.json, record-read-patrick-flyn-vrnt-7m3.json, and two more record-search fixtures all use id 1401638 for "United States Census, 1850"

That collision already exists on main today, before this PR touches anything. Adding collections-search-kentucky.json as a third user of 1401638 (matching the Census meaning) doesn't fix #1405, it extends the exact same bug class to a third fixture.

The acceptance criterion here is that a shared id can't mean two different things depending on which fixture answers the call. Please pick an id that doesn't already appear anywhere in eval/fixtures/mcp/ with a different meaning; a corpus-wide grep for the candidate id before landing on one would have caught this. Cia-3's collision check named one specific id (1875188) rather than asking for a full sweep, which is likely why this got missed.

EdmondOware (approved):
Both fixes check out now, verified independently rather than just against the description.

collections-search-kentucky.json: the new id 1786457 is internally consistent, title, dateRange, recordCount, and url all agree with each other. I grepped the whole eval/fixtures/ tree myself and it appears nowhere else in the corpus, so this doesn't repeat the collision from the last attempt.

same-person-flynn-strong.json: candidateArk is the canonical placeholder shape matching queryArk's format.

Approving.

PR #1608, opened 2026-08-13
tree_forget: resolve facts by owner, not a bare id, and add date-range selectors (#1574)

### LINKED ISSUE #1574: tree_forget resolves facts by bare id, not by owner — cross-person removal, and no date-range selector can be built on it

chrisedeson:
Closes #1574.

## Summary

FamilySearch does not guarantee a fact id is unique across owners. A real compiled tree returned the same literal Birth fact id for six distinct people. `tree_forget` removed facts by a flat, unowned id, so forgetting one person's fact could silently delete a different, unrelated person's fact that happened to share the same id. This is the bug #1574's reporter hit.

The same issue folded in a second gap: no way to say "forget everything before, after, or between two years" without the agent reading `tree.gedcomx.json` first, which defeats the exercise.

Both are fixed here. Fact removal now resolves and removes by owner, never a bare id. Three new selectors, `facts-before`, `facts-after`, `facts-between`, resolve a year-bounded request internally, using the same owner-scoped
removal the first fix built.

## What changed

- `packages/engine/mcp-server/src/tools/tree-forget.ts`: fact targets are now  `(ownerKind, ownerId, factId)` triples, not a bare factId. `ownerKind` is  `person` or `relationship`. A person and a relationship can share a literal id too, and nothing in the tree or its validator checks the two id spaces against each other.
- The bare `fact` selector (previously `factId` only) now resolves ownership at call time: one owner proceeds unchanged, zero owners is the existing "not in the tree" error, more than one owner is a new error naming every
  candidate and asking for `personId` to pick one.
- `birth-of`/`death-of`/`facts-of` stay unambiguous (the selector already names the person) but now warn in `validation.warnings` when the removed fact's id also exists on a different, untouched owner. Ids only, per the
  tool's own redaction contract. Not an error; removal is correct regardless.
- `parents-of`/`spouses-of` remove their swept `Parents`/`Marriage`/`Divorce`/`Annulment` facts through the same owner-scoped key as every other selector, and carry the same shared-id warning.
- A fact selector combined with a selector that removes that fact's own owner in the same call (e.g. `person: X` plus `fact: <X's own fact>`) used to reject the whole request with a false "not in the tree" error, since the pruning step only visits kept owners. Pre-existing, not introduced by this PR. Now recognized as already satisfied by the other removal.
- Three new selectors: `facts-before`/`facts-after` (a year threshold) and `facts-between` (an inclusive year range). `personId` is optional on all three; given, scoped to that person; omitted, tree-wide, across every person and every Couple relationship.
- Matching is confident, not merely overlapping. `facts-before(year)` needs the fact's latest possible year strictly before the threshold. `facts-after` needs the earliest possible year strictly after it. `facts-between` needs the fact's whole possible range inside the bounds. An uncertain or ranged date that only partly overlaps the boundary is left
  alone rather than guessed at. A looser rule would recreate the same over-broad-removal shape this PR exists to eliminate, just moved from id collision onto date uncertainty.
- Date comparison reuses this repo's existing date utilities (`getStandardDate`, `earliestYear`, `latestYear`, the same ones `person_warnings` already uses) rather than a new parser. A fact with no parseable date is never matched; the count skipped this way is reported in `validation.warnings`.
- The year (or years) the caller passes is their own input, never a value read from the tree, so echoing it back in an error message is not a redaction violation the way printing a fact's own date would be.
- Spec (`docs/specs/tree-forget-tool-spec.md` §2.1.1-2.1.3), the tool's own MCP schema description, `forget-and-rederive/SKILL.md`'s selector table, and the `dev/try-tree-forget.ts` smoke script are all updated to match.

## Test results

Full package suite: 93 test files, 2326 tests, clean typecheck. 34 new tests in `tree-forget.test.ts` (29 to 63), all mutation-verified: broke each fix on purpose, confirmed the specific test failed, restored, confirmed green again.

Four independent review passes, two adversarial, found and fixed these before this PR opened:

- A person and a relationship can share a literal id, so the first key design could still conflate them. Fixed with `OwnerKind`.
- A fact selector combined with a selector removing that fact's own owner used to reject the whole call with a false error. Pre-existing bug, fixed here.
- Three test gaps: an untested strict boundary, a mislabeled test, no dry run coverage for the date range selectors. Closed.
- `personId: ""` silently fell through to tree-wide scope instead of erroring, on a destructive tool. Fixed in both the new selectors and the pre-existing `fact` selector.

## Review guidance

**Start here:** `factKey`/`OwnerKind` and `datedFacts`/`requireYear` in `tree-forget.ts`, the owner-scoped key fix at the end of the `parents-of`/`children-of`/`spouses-of` case block, and the `personId: ""` tests at the end of the test file.

**Check this:** the dry-run warning extension to `birth-of`/`death-of`/`facts-of` and to `parents-of`/`spouses-of` is new behavior beyond the issue's literal text, though a defensible reading of it.

## Test plan

- [x] Full package test suite green, typecheck clean.
- [x] Every new test broken on purpose and confirmed it fails, then restored.
- [x] Four independent agent reviews, no shared context with each other or with the implementation work, two specifically adversarial.
- [ ] N/A. No skill eval slot: `forget-and-rederive` has no directory under `eval/tests/unit/`, so this costs no paid run.

## Follow-on issues

None filed.


Asimi1234 (approved):
Ran it merged with main (clean merge): tsc clean, engine vitest 2313 pass, tree-forget 64/64, and the owner-scoping fix is mutation-verified (reverting factKey fails 15 tests) — the earlier shared-id advisory note is resolved at head, no blocking findings, LGTM pending the senior-developer and senior-genealogist code-owner approvals.

packages/engine/mcp-server/src/tools/tree-forget.ts
```diff
@@ -220,16 +233,220 @@ function factIdsOfTypes(
...
+      const latest = std === null ? null : latestYear(std);
+      if (earliest === null || latest === null) {
+        skipped += 1;
+        continue;
+      }
+      entries.push({ ownerKind, ownerId, fact, earliest, latest });
+    }
+  };
+  if (personId !== undefined) {
+    consider("person", personId, persons(tree).find((p) => p.id === personId)?.facts);
```
Gennecis, line 443: **A person-scoped date selector leaves a Couple-relationship fact behind.**

`facts-before`/`facts-after`/`facts-between` with a `personId` consider only that person's own `facts[]`. A marriage date normally lives on the Couple relationship rather than on either spouse's person record — 834 `('Couple','Marriage')` facts across this repo's committed fixture trees — so a person-scoped date forget can report success and leave the answer in the file.

Probe run against this branch (Patrick b. 1820, married Ellen 1845 at Cork; marriage fact on the Couple relationship, as in real FS data):

```
facts-before(personId: "I1", year: 1850)
  ok: true    removed.factsByType: {"Birth":1,"Residence":1}
  validation.warnings: []
  I1's own facts left: []
  surviving on R6: [{"id":"MF1","type":"Marriage","standard_date":"1845","place":"Cork, Ireland"}]
```

The researcher is told two facts went and gets no warning, so nothing signals that the 1845 marriage and its place are still readable in `tree.gedcomx.json`.

There is no in-skill workaround. `facts-of` does not reach relationship facts either, SKILL.md forbids reading the tree to hand-resolve the fact id, and the tree-wide variant is much broader than the request — which is the tension #1573 raised.

The tool already treats this shape as a leak elsewhere: the `SWEPT_SPOUSE_FACT_TYPES` comment ("FamilySearch carries a conclusion TWICE — as graph structure AND as a documentary fact") is exactly why `spouses-of` sweeps both carriers. The new date selectors sweep one of the two.

**Preferred fix**, at this line — include Couple relationships the person is a party to:

```ts
if (personId !== undefined) {
  consider("person", personId, persons(tree).find((p) => p.id === personId)?.facts);
  for (const r of relationships(tree)) {
    if (r.type !== "ParentChild" && (r.person1 === personId || r.person2 === personId)) {
      consider("relationship", r.id ?? "", r.facts);
    }
  }
}
```

**Acceptable alternative**, if a person-scoped selector should never touch a jointly-owned Couple record — a defensible reading, since a marriage belongs to both spouses. Then the agent-facing text has to stop implying coverage:

- `src/tools/tree-forget.ts:913` — `"facts-before/facts-after: that person's facts confidently before/after a year"` should say *person-level* facts only, note that a `Marriage`/`Divorce` fact carried on a Couple relationship is **not** included, and point at omitting `personId` to reach those.
- `docs/specs/tree-forget-tool-spec.md:213` — after "Given, the selector matches only that person's own facts", state the consequence outright rather than leaving it to be inferred.
- Emit a `validation.warnings` line when the scoped person is party to a Couple relationship carrying dated facts the selector did not consider, so the omission is visible in the dry run.

Either way this needs a test. `datedFixture()` (`tests/tools/tree-forget.test.ts:985`) has `relationships: []`, so no person-scoped date test ever meets a Couple relationship — the suite cannot see this. The tree-wide twin is covered ("a tree-wide date-range selector matches a fact that lives only on a Couple relationship", line 1236); the person-scoped one is missing.

chrisedeson: Confirmed and fixed: person-scoped facts-before/facts-after/facts-between now also sweeps the Couple relationship the person is a party to, tested with your exact scenario, mutation-verified, and pushed.

johnmarkpeterbrown (changes requested):
```markdown
Thanks — the owner-scoping model is solid, and the two review fixes (Couple
relationship reach, warning-after-removal) both hold up when I run them.

One thing needs to change before this merges.

## `Bef` and `Aft` dates get deleted when they shouldn't

The three date selectors treat every date as having a real upper and lower
bound. `Bef` and `Aft` dates don't.

`earliestYear`/`latestYear` add a ±10-year fudge to those two modifiers. That
fudge exists for the warning checks, where a rough guess is fine. It is not a
real bound. "Aft 1850" could be 1851 or 1950 — the helper just returns 1860.

So this happens today, on your branch:

| fact | call | what happens |
|---|---|---|
| `standard_date: "Aft 1850"` | `facts-before(1870)` | deleted, no warning |
| `standard_date: "Bef 1900"` | `facts-after(1885)` | deleted, no warning |

Both contradict what this PR promises. Spec §2.1.3 says the latest possible
year must be strictly before the threshold. The schema description says only a
fact confidently inside the range is removed. Neither is true for these two
modifiers.

This is the over-broad-removal shape the PR was written to remove. It just
moved from id collision onto date uncertainty.

It reaches real data. There are 22 `Bef` and 17 `Aft` standard dates in the
committed final-tree files under `eval/runlogs/e2e/` — `wilkins-death-kentucky`,
`mccarley-spouse`, `johann-widmer-vitals` and others.

`Abt` and `Est` are fine. Their fudge widens the range, which makes matching
harder, not easier. Only `Bef` and `Aft` shrink an open side into a fake bound.

### The fix

In `src/tools/tree-forget.ts`, inside `datedFacts`:

You wrote:
```ts
      const std = getStandardDate(fact);
      const earliest = std === null ? null : earliestYear(std);
      const latest = std === null ? null : latestYear(std);
      if (earliest === null || latest === null) {
        skipped += 1;
        continue;
      }
      entries.push({ ownerKind, ownerId, fact, earliest, latest });
```

Change it to:
```ts
      const std = getStandardDate(fact);
      if (std === null) {
        skipped += 1;
        continue;
      }
      const earliest = earliestYear(std);
      const latest = latestYear(std);
      if (earliest === null || latest === null) {
        skipped += 1;
        continue;
      }
      // `Bef`/`Aft` are open-ended. date-helpers' FUDGE caps them at ∓10
      // years, which is a heuristic for the warning checks, not a real
      // bound — "Aft 1850" could be any later year. Treating the cap as a
      // bound lets facts-before sweep it. Unbounded on the open side, so
      // the confident-match predicates never fire there.
      entries.push({
        ownerKind,
        ownerId,
        fact,
        earliest: /\bBef\b/.test(std) ? -Infinity : earliest,
        latest: /\bAft\b/.test(std) ? Infinity : latest,
      });
```

I ran this on your branch. The two cases above flip to the correct "matched
nothing" refusal, `Est 1800` still matches `facts-between(1790, 1810)`, and
every existing `tree-forget` test stays green.

### The test gap that hid it

`datedFixture()` in `tests/tools/tree-forget.test.ts` only uses bare years and
`Bet X and Y`. No `Bef`, `Aft`, `Abt` or `Est` appears anywhere in that file, so
the suite cannot see this either way.

You wrote:
```ts
          { id: "DF4", type: "Residence", standard_date: "Bet 1845 and 1855" },
```

Change it to:
```ts
          { id: "DF4", type: "Residence", standard_date: "Bet 1845 and 1855" },
          { id: "DF6", type: "Residence", standard_date: "Aft 1855" },
          { id: "DF7", type: "Residence", standard_date: "Bef 1845" },
```

Then assert `facts-before(1850)` leaves `DF6` alone, and `facts-after(1850)`
leaves `DF7` alone. The three existing date tests will need their expected id
lists extended for the two new facts.

Break the fix on purpose first and confirm each new assertion fails, the way you
did for the rest of this PR.

### One doc line

Spec §2.1.3's three bullets read as if `latestYear` and `earliestYear` are true
bounds. Say what an open-ended date does: a `Bef` fact is never matched by
`facts-after`, and an `Aft` fact is never matched by `facts-before`.

## Before merge

Two things unrelated to the code.

The existing approval read `fe46e2a3`. Head is `ba5827ef`. The Couple
relationship fix and the warning dedupe both landed after it, so nobody has
reviewed those two commits. A push doesn't clear approvals on this repo and
nothing re-requested review, so that approval still counts toward the two.



Everything else looks right to me. Suites and typecheck are clean on my end —
96 files, 2373 tests.
```

chrisedeson:
Fixed. 

Bef/Aft are now treated as open-ended on their real side. While I was in there, found and fixed one more edge case: a Bet range with the modifier on only one side. Tests for both, mutation-verified.

johnmarkpeterbrown (approved):
Looks good!

Asimi1234:
Re-reviewed at head `44f71948` (my earlier approval was stale `@fe46e2a3`). Both prior findings verified resolved end-to-end — the Bef/Aft over-deletion (`Aft 1850` no longer swept by `facts-before`, `Bef 1900` no longer by `facts-after`) and Gennecis's Couple-relationship leak (a person-scoped `facts-before` now removes the marriage fact too). `tsc` clean, engine `vitest` 2395 pass, and reverting `factKey` still fails 17 owner-scoping tests.

One non-blocker: ISO-format `standard_date` like `"1883-12-31"` (and `"~1869"`) return null from `earliestYear`, so those facts are silently skipped by the date selectors — I confirmed a `1883-12-31` fact survives `facts-before(1900)`, reported as "no parseable date." Only ~0.25% of committed dates and it's the safe direction (skipped, never over-deleted), but that's a precise date, so worth either parsing ISO in the date helpers or rewording that warning. Your call, not a blocker.

Every other thing looks good to me.

promise-emmanuel (changes requested):
The owner-scoping model holds up. I re-ran it merged with main (clean merge):
tsc clean, engine vitest 100 files / 2408 tests, apps/server 162 passed.
`factKey` carrying `ownerKind` is the right call — a person and a relationship
really can share a literal id, and `(ownerId, factId)` alone would have
reunited the same leak across a different pair of fields. Resolving the
shared-id notices *after* removal rather than before is also right, and the
Bef/Aft fix went further than the suggestion it came from: mirroring
`earliestYear`'s own Bet/or/plain branches instead of testing the whole string
means a modifier on one side of a range no longer voids the other side's real
bound. That's better than what was asked for.

Two things to change before this merges.

## 1. A dry run lets the agent read back the date you just hid

A dry run is free, writes nothing, and still returns `factsByType`. So calling
`facts-before` with different years narrows a fact's date to the exact year —
I got a Birth year of 1850 in 8 dry-run calls, never opening
`tree.gedcomx.json`.

Spec §3.1 says the result carries "counts and kinds only — never a name, date,
place," and that any future field is subject to the same rule. §2.1.3 argues
the year is the caller's own input, which is true of the year in the error
message, but it says nothing about the match/no-match answer — and that comes
from the tree.

The tool can't close this without dropping `factsByType` from date-selector
dry runs, which would break the "a dry run is a full rehearsal" property you
have a test for. So close it by instruction and write down what's left.

`packages/engine/plugin/skills/forget-and-rederive/SKILL.md`:

You wrote:
Do not read tree.gedcomx.json. You do not need the names and dates you are
about to remove, and you are better off not having them in context.



Change it to:
Do not read tree.gedcomx.json. You do not need the names and dates you are
about to remove, and you are better off not having them in context. For the same
reason, call each date selector once, with the year the researcher gave you. Do not
vary the year across dry runs to see what the counts do — that reads a date off the
tree as surely as opening the file.



`docs/specs/tree-forget-tool-spec.md` §2.1.3:

You wrote:
caller passes is their own input, never a value read off the tree, so it is
not subject to the redaction contract the way a fact's own date is.



Change it to:
caller passes is their own input, never a value read off the tree, so it is
not subject to the redaction contract the way a fact's own date is. The result,
however, carries a residual: a dry run is free and reports factsByType, so
repeating one with different years narrows a fact's date to the exact year.
§3.1 therefore holds per call, not across a series of them. SKILL.md closes
this by instruction — call each date selector once, with the researcher's own
year — because the tool cannot close it without giving up the dry run's
full-rehearsal property.



## 2. Asimi1234's ISO-date note is still open

It's the one note that didn't get answered. `earliestYear("1883-12-31")`
returns null, so the fact is skipped and counted as having no parseable date —
but it has a precise one. There are 23 ISO `standard_date` values in the
committed corpus. The direction is safe (skipped, never over-deleted), and the
fallback is fine — `getStandardDate({date: "1883-12-31"})` correctly gives
`31 Dec 1883`, so only a pre-populated ISO sidecar breaks.

What's wrong is what the researcher is told. They get a count, not which
facts, so nothing signals that an 1883 date survived a `facts-before(1900)`.

Don't touch the parser. `earliestYear` is shared with `person_warnings`, so
parsing ISO there is its own PR, and the spec already says `standard_date` is
GEDCOM-canonical — those 23 values are non-conformant data. Just say what
actually happened.

`packages/engine/mcp-server/src/tools/tree-forget.ts`:

You wrote:
```ts
            `${skipped} fact(s) in scope had no parseable date and were not ` +
              `considered for '${kind}'.`,
Change it to:


            `${skipped} fact(s) in scope had no date this tool could compare and were ` +
              `not considered for '${kind}' — including any whose standard_date is not ` +
              `GEDCOM-canonical form (e.g. '1883-12-31' rather than '31 Dec 1883').`,
Then extend the assertion matching /1 fact\(s\).*no parseable date/ in
tree-forget.test.ts, and add an ISO-dated fact to datedFixture() so the
suite pins it.

Approvals
Both standing approvals predate code that's now at head, and a push doesn't
clear them here, so both still count toward the required two.

Asimi1234's is recorded at fe46e2a3 — that commit has no earliestIsUnbounded at all, so it's before the Bef/Aft fix, the Couple-relationship reach and the warning dedupe. The re-check at 44f71948 was a comment, so it didn't replace the review.
johnmarkpeterbrown's is at 544c4b8b, which still calls validateParsed. The validateIntroduced switch landed after it in f30c1f69. I checked that merge against tree-edit.ts:675 and the signature — it's correct, just unreviewed.
Worth re-submitting both at head.

Everything else I looked at is fine. No schema, tree-shape, enum or auth
surface is touched. TREE_PARENT_CHILD_FIELDS has no facts, so the
person-scoped selector reaching only Couple relationships is complete rather
than a gap. Est I left alone — the same reasoning arguably applies to its
±10 years, but there are zero Est values in the corpus and that call was
already reasoned through.

chrisedeson:
Fixed both.

The date-narrowing leak needed two passes. The first fix was instruction-only (SKILL.md), but a follow-up review found that didn't actually close it: packing several years into one call's forget array leaks the same way, just in one call instead of eight. Closed it in code now. The tool itself rejects a call that mixes different year thresholds on the same date selector. Same threshold across different people still works fine.

ISO-date message reworded to say what's actually happening, plus a test.

promise-emmanuel (approved):
Re-reviewed at 82262b08, up to date with main. Both changes I asked for landed
verbatim, and I re-ran everything against the merge rather than the branch head:
tsc clean, engine vitest 102 files / 2438 tests, harness 2318 passed, control
plane 209 passed. All 11 checks green.

Worth noting the update-with-main was not a no-op for this PR — it repointed
`formatIssues` from `merge-shared` to `project-io`, and #1739's `readJson`
consolidation landed in the same window. I checked the merged file against
main's own version of it: the local `readJson` wrapper is still there in both,
so the consolidation is not half-applied here. Fine.

**The year-ladder guard is a good catch and it holds up.** You found the tighter
form of what I reported — `resolveSelectors` throwing on the first zero-match
selector while naming its own threshold means one call can pin a year, and the
SKILL.md instruction genuinely cannot reach that, because it is one call. I
mutation-tested the guard in both directions rather than trusting the tests:
neutering it fails exactly the two positive tests while the multi-person sweep
test correctly still passes, and widening it to reject any repeated kind fails
that sweep test. Scoping it to the threshold rather than the selector kind is
the right call, and it runs before the dry-run return, so dry runs are covered.

Approving. Two text fixes to fold in before you merge — both prose, no code.

## 1. The spec claims more than the guard delivers

The guard closes the laddered shape. It does not close the in-call surface: a
single `facts-between(P1, 1883, 1883)` is one call, one selector, one threshold,
so the guard permits it — and it returns `{"Birth": 1}`, which pins the year
exactly. No ladder needed. I ran it against the built merge.

Nothing is actually exposed by this, because the SKILL.md instruction ("use the
year the researcher gave you and stop") covers it. The problem is only that a
future maintainer reads "closes the in-call form in code", believes that surface
is sealed, and stops thinking about it. Narrow the claim to what the code does.

In `docs/specs/tree-forget-tool-spec.md`:

You wrote:
resolveSelectors closes the in-call form of this in code: a single



Change it to:
resolveSelectors closes the laddered in-call form of this in code: a single



Then after "an ordinary multi-person sweep, not a probe)." add:

This does not make a single call safe in general — one facts-between(Y, Y) is
still an exact-year test — so the instruction below, not this guard, is what
closes the surface. The guard removes the cheapest shape, not the shape.



## 2. A comment in the guard says something that isn't true

The comment claims `requireYear` rejects a missing-year entry regardless. It
doesn't, because this guard runs first. I checked all three orderings: a
malformed entry on its own correctly gives `'facts-before' requires a numeric
year`, but paired with a valid entry — either order — you get the threshold
error instead, which misdiagnoses the input.

In `packages/engine/mcp-server/src/tools/tree-forget.ts`:

You wrote:
```ts
  // against `undefined` directly. `.has()` distinguishes them; the main
  // loop's own `requireYear` rejects the missing-year entry regardless.
Change it to:


  // against `undefined` directly. `.has()` distinguishes them. Note this guard
  // runs BEFORE the main loop, so a malformed entry alongside a valid one
  // reports the threshold error rather than `requireYear`'s missing-year error
  // — misleading, but only on input that is already invalid.

chrisedeson:
Both fixed. Narrowed the spec claim to "closes the laddered in-call form" and added the note that a single facts-between(Y, Y) still pins an exact year, so the instruction is what actually closes the surface, not the guard. Corrected the comment about requireYear too.

PR #1609, opened 2026-08-14
record_search: surface the extraction batch number (#1592)

### LINKED ISSUE #1592: record_search drops UdeBatchNbr, so batch enumeration has no way to obtain a batch number
**Touches:** packages/engine/mcp-server/src/tools/record-search.ts, packages/engine/mcp-server/src/types/record-search.ts, docs/specs/record-search-tool-spec-v2.md (response section), packages/engine/mcp-server/tests/tools/record-search.test.ts, packages/engine/plugin/skills/search-records/SKILL.md (the `batchNumber` row's provenance), packages/engine/plugin/skills/search-records/references/collection-quirks.md

Split out of #1326 / PR #1591, which made `batchNumber` reachable and correct and then hit the wall behind it: **nothing can tell the agent what batch number to send.**

## The gap

PR #1591 finished the parameter — it anchors alone, it rejects the destructive `recordCountry` pairing, its shape is documented open, and three references plus SKILL.md now prescribe batch enumeration for IGI parish registers. But an agent following that guidance has no way to obtain a batch number:

- `collection-quirks.md` contains four *illustrative* batch numbers (`C050761`, `M01048-5`, `M17288-6`), not a lookup for any real parish
- `search-records`' `allowed-tools` has no web access, so "a catalog/index page" is not reachable from the VM
- **no MCP tool returns a batch number** — `grep -rn batch packages/engine/mcp-server/src/` finds only the input parameter

So the strategy is prescribed on five surfaces and executable on none. SKILL.md's `batchNumber` row currently sources it to "`references/collection-quirks.md`, or a catalog/index page", and neither is real.

## It is already arriving, and being discarded

FamilySearch returns the batch on every persona. Probed live 2026-08-12 against `/service/search/hr/v2/personas?q.batchNumber=B01883-5`:

```
"fields":[{"type":"http://familysearch.org/types/fields/UdeBatchNbr",
           "values":[{"labelId":"FS_UDE_BATCH_NBR","text":"B01883-5", …
```

with a sibling `http://familysearch.org/types/fields/BatchLocality` → `"Alabama, United States"`.

`gedcomx.fields` is a top-level key on every search entry and the tool drops the whole array.

## What this unlocks

Surfacing `UdeBatchNbr` on `record_search` results closes the loop the references already describe:

1. find one record by name in an IGI parish collection (an ordinary indexed search)
2. read the batch number off that hit
3. send it alone — enumerate the entire parish

That is the exact workflow `collection-quirks.md` has prescribed since before #1326, and the missing parameter was only half of why it never worked.

## Scope notes for whoever picks this up

- Decide whether `record_read` should surface it too, and whether the staged-results path and `rank_search_matches` need to carry it.
- `BatchLocality` is a judgement call, not an obvious yes. It would let the agent confirm a batch is the right parish — but it must not become an argument for passing `recordCountry`, which PR #1591 measured as inert at best and silently destructive at worst, and which `validateInput` now rejects.
- Once the field ships, the eval test becomes worth writing and can assert the real chain (name hit → batch → enumeration). Today a test would have to hard-code a batch into a fixture and would verify a workflow the agent cannot perform. That follow-up is genealogist-lane fixture work.

## Why this was filed rather than folded into PR #1591

**It is a different skill's eval slot, and bundling would force a second paid run.** PR #1591 has already spent one `search-records` slot ($8.39, 26/26) because it changed SKILL.md. Changing `record_search`'s output shape and then SKILL.md's provenance guidance would invalidate that run log and require a second one inside the same PR. It is also a new output contract arriving on a PR that has been through three review rounds on a different one.

chesworthrm:
Closes #1592.

## The problem

FamilySearch returns the extraction batch on every search entry's `content.gedcomx.fields[]`, and `record_search` dropped the whole array. So the batch-enumeration strategy was prescribed on five surfaces and executable on none: `collection-quirks.md` carries only illustrative batch numbers, `search-records` has no web access, and no tool returned one. `SKILL.md`'s `batchNumber` row sourced it to "`references/collection-quirks.md`, or a catalog/index page" — neither of which is real.

`record_search` results and `rank_search_matches` stubs now carry `batchNumber`, closing the loop: find a record by name, read the batch off the hit, send it back to enumerate the rest of the extraction.

## The finding that shaped the implementation

**The batch is matched on `labelId === "FS_UDE_BATCH_NBR"`, never on the field's `type` URI — and this is the whole design.**

The type suffix is spelled **both** `UdeBatchNbr` and `UdeBatchNumber` depending on the collection, with nothing caller-visible to predict it from:

| probe | type suffix | labelId |
|---|---|---|
| `q.batchNumber=B01883-5` (the issue's own probe) | `UdeBatchNbr` | `FS_UDE_BATCH_NBR` |
| English IGI batch `M01048-5` | `UdeBatchNbr` | `FS_UDE_BATCH_NBR` |
| collection `1494474` (Germany) | **`UdeBatchNumber`** | `FS_UDE_BATCH_NBR` |
| all-numeric batch `8317102` | **`UdeBatchNumber`** | `FS_UDE_BATCH_NBR` |

The issue quotes only `.../types/fields/UdeBatchNbr`. A matcher written to that literal returns nothing on the collections using the other spelling, and **a batch present upstream but unread here is indistinguishable from a record that has none** — so the miss would have been silent. Do not add a `type` check "for safety": it carries the same risk forward to a third spelling while excluding nothing, since no other field uses this labelId.

**Presence is per RECORD**, not per collection or record type. Over an ordinary indexed search anchoring on no batch: birth 0/10, census 0/10, marriage 4/10 — and one collection (`1618491`) returned 7 hits carrying a batch and 11 without. Absence carries no information, and the spec and skill guidance both say so.

Evidence trail: `dev/probe-batch-field.ts`, four legs, every verdict computed from its own run (a 200 carrying zero entries counts as not measured, so an empty sample can't print a direction).

## Scope decisions

- **`BatchLocality` is deliberately not surfaced.** It sits on the same array and would read as an invitation to pass `recordCountry` alongside a batch — which `validateInput` rejects, and which PR #1591 measured as inert at best and silently destructive at worst.
- **`record_read` is not included.** The field reaches `rank_search_matches` because a search supplying `subjectId` returns `ranked`, and that is the projection the caller actually reads.
- **Flat and top-level, not inside `gedcomx`.** The staged slim block strips `gedcomx`, and staging is the normal case — a batch visible only in unlogged exploratory searches would be useless. A test pins this; breaking it by adding a `delete` fails loudly.

## Eval

`ut_search_records_028` exercises the real chain against live-captured fixtures: name search → read `batchNumber` off a hit → enumerate the parish register → reach **father Thomas Richardson**, who is named only in the `C00558-7` extraction and not in `I03213-6`. That difference is real, and it is the genealogical argument for enumerating rather than accepting the first hit.

It runs on a purpose-built scenario (`richardson-ashton-register-unworked`) with an empty log, because the first host I tried already had the register worked through to assertions — a skill given it correctly answered from state and never searched. The scenario README records that, so the next person doesn't repeat it.

## Verification

- `packages/engine/mcp-server` — **2272 pass** (93 files), typecheck + build clean
- `eval/harness` — **2078 pass**, 2 skipped
- `make eval-skill SKILL=search-records` — **28/28**, run log `v1_2026-08-14_03-36-42`, annotated
- `check_runlogs.py` — all rules satisfied
- The new slim-block test was broken deliberately (adding `delete r.batchNumber`) and watched to fail, then reverted


Ikennaya1 (approved):
Approving the genealogist side — fixtures, eval tests, run log, and skill prose.

Verified against a test-merge with main: 2289 vitest passes, 2077 harness passes, typecheck clean, and merge clean. Confirmed the slim-block test is meaningful by deliberately breaking it and watching it fail.

The scenario remains unworked (empty log, no assertions, pli_001 planned), and ut_search_records_028 is a fair test of the intended enumeration behavior.

One non-blocking note: the Scope decisions should mention that spouse-absent-on-census-hit.json now includes a record_read fixture and a modified judge_context, so future readers can easily understand why that test changed.

Nice work!

johnmarkpeterbrown (approved):
Looks good!

promise-emmanuel (approved):
Approving the developer side.

Verified in a worktree of this head, after `npm ci` (a fresh worktree fails
typecheck without it — environmental, not the branch): vitest 2287 pass / 94
files, harness pytest 2123 pass / 2 skipped, tsc and build clean.
`check_runlogs.py` satisfied.

I re-did the slim-block mutation rather than take it on trust — added
`delete r.batchNumber` into the block, watched "survives the staged slim block"
fail 1/120, restored. The test fails for the right reason.

Keying on `labelId` instead of the `type` URI is the right call and the tests
pin it, including the invented third spelling. That's the failure mode worth
guarding: a batch present upstream but unread here is indistinguishable from a
record that has none, so a `type`-keyed matcher would have gone wrong silently
on the German collections. `probe-batch-field.ts` treating a zero-entry 200 as
NOT MEASURED is the detail that makes the evidence trustworthy.

One thing before merge — seconding Ikennaya1: please add a line to the PR body
for the `spouse-absent-on-census-hit.json` change. The reason is good and it
should be in the record. From the run log, `ut_search_records_027` calls
`record_read` with `recordId: "M7QZ-8KD"` off the search result and had nothing
stocked behind it; the new fixture stocks a call the skill actually makes.

Two non-blocking notes, neither needing action:

- The run log was scored under judge prompt `e4bd6e77`, current is `e1f3a744`.
  That drift is from #1637 landing on main the day after the run, not from this
  PR. Not worth $9.17 to re-run.
- SKILL.md is +501 B, billed on every `search-records` run. It's the six lines
  that tell the agent where to get a batch — that's the half that makes the
  field usable, so it earns the bytes.

PR #1612, opened 2026-08-14
eval-harness: count cached tokens alongside input, not inside it (#1534)

### LINKED ISSUE #1534: Judge cost subtracts cached tokens from input_tokens, so a cache-heavy run computes a negative cost and kills the suite
> **Reviewed 2026-08-12 before junior handoff.** The premise was false — this is a sign
> error in the cost formula, not floating-point noise, and both remedies the original
> asked for would have made a measured ~18% under-count of judge spend permanent. The
> ask is rewritten; the original body follows under "Original issue".

**Touches:** eval/harness/harness/judge.py, eval/harness/run_tests.py,
eval/harness/tests/unit/test_judge.py, eval/harness/tests/unit/test_partial_runlog.py,
docs/specs/unit-test-spec.md

Hit while re-running `init-project` for PR #1523: `run_tests.py` crashed mid-run
writing a partial run log.

    jsonschema.exceptions.ValidationError: -0.00010129999999999991 is less than the minimum of 0
    On instance['tests'][0]['totals']['judge_cost_usd']: -0.00010129999999999991

**This is not floating-point noise.** `_compute_cost`
(`eval/harness/harness/judge.py:795-799`) computes

    (input_tokens - cache_read_input_tokens) * input_rate
      + cache_read_input_tokens * cached_rate
      + output_tokens * output_rate

but the Anthropic API reports those two token counts **disjointly** — `input_tokens`
already excludes cache reads. So the first term is a double subtraction that goes
negative whenever the prompt cache is warm, and the total goes negative whenever the
output term is too small to cover it. One decisive reading, from a committed run log:

```sh
python3 -c "import json;t=json.load(open('eval/runlogs/unit/init-project/v1_2026-08-11_11-34-29.json',encoding='utf-8'))['totals'];print(t['input_tokens'],t['cached_input_tokens'])"
# -> 100 969789
```

Across every committed run log, 282 of 1814 judge draws have `cached_input_tokens >
input_tokens` (measured 2026-08-12; re-derive by walking `eval/runlogs/**/*.json` and
counting `runs[].judge` blocks where cached exceeds input). Recomputing the whole unit
corpus at Haiku list price, the recorded judge total is **$16.24 against a corrected
$19.70 — understated by about 18%** (measured 2026-08-12 over `eval/runlogs/unit/`;
run-log pruning, a judge-model change, or a different cache-hit rate all move it). No
committed log carries a negative, because a negative can never be written — validation
rejects it and the run dies instead, so the corpus is self-selected.

## Ask

**Fix the arithmetic. Do not clamp, and do not widen the schema minimum** — both hide
the under-count and disarm the check that caught this.

1. `eval/harness/harness/judge.py:795-799` — drop the `- cached`. All three terms are
   then non-negative, so the result cannot be negative and needs no clamp.
2. `eval/harness/tests/unit/test_judge.py:764-773` — `test_compute_cost_for_known_model`
   encodes the bug ("200 fresh input" for `input=1000, cached=800`). Update it, and add
   a case with `cached > input` (e.g. `input=1885, cached=5004, output=939`) asserting a
   positive cost. **Put new tests at the end of the file** — PR #1564 rewrites this
   region.
3. `eval/harness/run_tests.py` — `_flush_partials()` (line 871) sits inside a `try` that
   catches only `KeyboardInterrupt`, so a raise from `write_partial_runlog` unwinds past
   the summary, the final write, and the promote-to-scratch. Guard it, print loudly, and
   set `harness_error` rather than swallowing. Then fix the older half of the same
   fault: line 888's `if harness_error is not None: return 1` returns **before** the
   promote-to-scratch block (~line 930), so today any harness error leaves the completed
   tests only in the gitignored `.partial_*.json` dotfile. Promote partials to
   `scratch_<ts>.json` on that path too, then return 1.
4. `eval/harness/tests/unit/test_partial_runlog.py` — the guard test's home. Note that
   `test_cli.py:213` and `:318` monkeypatch `write_partial_runlog` away, which is why
   the CLI suite never saw this; the new test must exercise the real writer (or make a
   stub raise) and assert the completed results survive.
5. `docs/specs/unit-test-spec.md:1263` — "should be 50%+ of `input_tokens`" is
   unstateable under the real semantics (the run above reads as 969,789%). Restate the
   diagnostic as a share of `cached + input`, and say in one clause that `input_tokens`
   excludes cache reads — that sentence is what stops the next person re-deriving the
   subtraction.

**Do not touch** `docs/specs/schemas/run-log.schema.json` or
`packages/schema/schemas/run-log.schema.json` (`judge_cost_usd` `minimum: 0` at lines
151, 183, 428 of each). Leaving them is the point — that constraint is what surfaced the
bug. If you do edit them for any reason, edit **both trees**, or `make harness-test`'s
mirror-parity check fails.

**Do not backfill committed run logs.** They record what was measured at the time;
rewriting ~47 files to restate history is a separate, harder call, and this repo's
standing rule is not to quote historical cost figures forward.

**Out of scope, name it in the PR body instead:** judge cost still ignores
`cache_creation_input_tokens` (billed at 1.25x input) because the harness never captures
the field. Capturing it means a new property on the run-log judge block in both schema
trees plus `eval/app/lib/types.ts` — a schema change, and one whose size cannot be
estimated until the field is being recorded.

## How you know it worked

`make harness-test`. Break it first: revert the one-token fix and watch the new
`cached > input` test fail — a check nobody has seen fail is not a check (`CLAUDE.md`,
"A new lint must be proven to fail"). Then `make test-all`.

**No paid eval run.** `eval/harness/**` is not in `build_snapshot`
(`eval/harness/harness/snapshot.py:135-152`), so no run log goes inactive and
`check_runlogs.py` does not fire.

**Nothing else checks this.** No job recomputes cost from tokens after the fact — the
run log stores whatever the formula produced, and the CRUD UI just displays it. The unit
test you add is the only guard.

Say in the PR body that judge cost in new run logs rises ~18% against older ones and
that this is the fix, not a regression. Suite cost caps are effectively unaffected:
judge is a small share of a run (init-project 2026-08-11: $0.046 judge against $1.166
skill).

## Sequencing

Land after PR #1564 (open, updated 2026-08-12) — it touches
`eval/harness/harness/judge.py`, `eval/harness/tests/unit/test_judge.py` and
`eval/harness/run_tests.py`, with test-file hunks bracketing the cost test. Issue #1484
covers cost/violation reporting on the e2e side; adjacent, do not merge with this.

## Original issue

**Touches:** eval/harness/harness/runlog.py, packages/schema/schemas/run-log.schema.json (or wherever `judge_cost_usd`'s schema minimum is defined)

Hit while re-running `init-project` for PR #1523: `run_tests.py` crashed mid-run writing a partial run log.

```
jsonschema.exceptions.ValidationError: -0.00010129999999999991 is less than the minimum of 0

Failed validating 'minimum' in schema['properties']['tests']['items']['properties']['totals']['properties']['judge_cost_usd']:
    {'type': 'number', 'minimum': 0}

On instance['tests'][0]['totals']['judge_cost_usd']:
    -0.00010129999999999991
```

`judge_cost_usd` is presumably computed as a difference/sum of per-token costs, and floating-point rounding produced a tiny negative value (`-0.0001`) instead of exactly `0`. The schema's `minimum: 0` constraint has no tolerance for this, so `validate_run_log` throws and the whole run — not just the offending test — dies before any run log is written. Only 4 of 10 `init-project` tests had completed when this crashed; none of that work was persisted (only a `.partial_*.json` scratch file survived).

## Ask

Either clamp `judge_cost_usd` (and any sibling cost fields computed the same way) to `max(0, computed_value)` before it's written, or widen the schema's `minimum` to tolerate floating-point noise (e.g. `-0.01`). Also worth checking whether this is a `write_partial_runlog` crash killing an otherwise-fine run — if so, a single test's bad cost shouldn't be able to take down the whole invocation's results.

aghadiayeamayanvboernest:
## Summary

- `_compute_cost` used `(input_tokens - cache_read_input_tokens)` as the fresh-input term, but the API reports those counts **disjointly** — `input_tokens` already excludes cache reads. The subtraction double-counted the discount, drove the term negative on any warm cache, and when the output term couldn't cover it the total went negative too. `judge_cost_usd` carries `minimum: 0`, so the run log was rejected and **the suite died mid-run having persisted nothing**.
- Also guards `_flush_partials()`, which is what turned a bad number into a lost suite.
- Tier: **Normal**. Closes #1534.

## Review guidance

**Start here:** the arithmetic in `_compute_cost`. The whole fix is dropping `- cached`; the value is in the comment explaining *why* the two counts are summed, because the subtraction reads plausible and someone will re-derive it otherwise. Second: the `_flush_partials` guard catches **per skill** so one skill's bad entry can't suppress another's partial, and sets `harness_error` rather than swallowing, so the run still exits 1.

**Unsure about:** nothing structural.

## Plan

**Didn't change:** the `judge_cost_usd` `minimum: 0` in either schema tree, and no clamp on the computed value. Both were in the original ask; both would have made the under-count permanent while disarming the check that surfaced this. Summing makes every term a non-negative count times a non-negative rate, so a negative is unreachable and a clamp would be dead code. Committed run logs are **not** backfilled — they record what was measured at the time.

**Acceptance check:** `test_compute_cost_counts_cached_tokens_alongside_input_not_inside_it` — a real corpus shape (`input=1885, cached=5004, output=939`). Fails on `main` computing `$0.0020764` against the correct `$0.0070804`, a 71% understatement on that shape.

**Deviated from the plan:** two corrections to the issue, both verified before acting.
1. **The second half of ask #3 is already done.** The issue says the `harness_error` path returns before promote-to-scratch — PR #943 fixed that, and `run_tests.py` now carries a comment saying so; line 1029 promotes *then* returns 1. Left untouched. The first half was real and is what's guarded here.
2. **The spec has two stale sites, not one.** Both said the cache rate "should be 50%+ of `input_tokens`". Fixed both, and stated the disjointness explicitly — that sentence is what stops the next person re-deriving the subtraction.

Also: the issue's cited run log had been pruned by retention, so the figures are re-derived from the current corpus — **276/1870** draws with `cached > input` (issue: 282/1814) and **19%** understated (issue: ~18%).

## Test plan

- [x] Ran `make test-all`. Eval harness pytest **2077 passed / 2 skipped**, eval-app **148**, control-plane, engine all green. The one red is the known **#1366** `apps/electron` flake — and I confirmed it rather than assuming: it failed once in isolation, **passed on re-run at 2801ms against a 5s timeout**, and **passes on clean `main` with this work stashed**. This branch changes zero files under `apps/electron`.
- [x] **Both fixes proven to fail first**, per CLAUDE.md. Reverting the arithmetic fails the new test. Disabling the flush guard fails the guard test *after printing `✓ [1/3]` and `✓ [2/3]`* — those two passing tests then vanish, which is the reported failure exactly.
- [x] N/A — no skill run-log snapshot changed. `eval/harness/**` is not in `build_snapshot`, so no run log goes inactive and `check_runlogs.py` doesn't fire. No paid eval run.

**Expect judge cost in new run logs to read ~19% higher than older ones. That is the fix, not a regression** — and the share is small: init-project 2026-08-11 was $0.046 judge against $1.166 skill, so suite cost caps are effectively unaffected.

**Nothing else checks this.** No job recomputes cost from tokens after the fact; the run log stores whatever the formula produced and the CRUD UI displays it. The unit test added here is the only guard.

## Out of scope

Judge cost still ignores `cache_creation_input_tokens` (billed at 1.25× input) because the harness never captures the field. Capturing it means a new property on the run-log judge block in **both** schema trees plus `eval/app/lib/types.ts` — a schema change whose size can't be estimated until the field is being recorded.


clack391 (approved):
This is a clean, careful fix, and unusually well documented. Happy to approve.

A few things that stood out:
- The arithmetic fix is correct, and the comment explaining why the two token counts are summed rather than subtracted is exactly what stops the next person re-deriving the bug.
- You caught both spots where the issue's ask was slightly off, and verified before acting: the early-return half of item 3 was already handled by #943, and the spec had two stale sites rather than one.
- The regression guard is a genuine one. Adding the small-output case, because the common warm-cache shape stays positive even under the bug, avoids the check-that-cannot-fail trap that is very easy to walk into here.
- The PR body calls the roughly 19% cost rise the fix and not a regression, and confirms the electron flake instead of assuming it.

Verified locally: the two changed test files and the full harness suite pass, and reverting the one-token fix makes the small-output guard fail with a negative cost, so the guard bites.

eval/harness/tests/unit/test_partial_runlog.py
```diff
@@ -119,3 +119,113 @@ def test_promote_partial_to_scratch_renames(tmp_path):
...
+        "--skill", "skill-a",
+        "--tests-dir", str(root), "--runlogs-root", str(runlogs),
+        "--concurrency", "1",
+    ])
+
+    # Reported as a harness failure rather than a clean run...
+    assert rc == 1
+    # ...and the writer really was exercised past its first (successful) call,
+    # so the raising path was taken rather than skipped.
+    assert calls["n"] > 1
```
chesworthrm: The guard does two things — swallow-and-record, and stop submitting. This assertion only pins the first.

I deleted `stop_submitting = True` from the guard and ran the full harness suite: **2078 passed, 2 skipped**. Completely green. Nothing anywhere catches its removal — and that's the line your PR body describes as preventing tests that are "executed, judged, PAID FOR, and then discarded."

One token pins it. I checked both directions: `== 2` passes against your code as it stands, and fails with "got 3" once the line is gone.

You wrote:
```python
    assert calls["n"] > 1
```

Change it to:
```python
    assert calls["n"] == 2, "stop_submitting must halt the third test; a third flush means it did not"
```

The comment just above needs the same tightening, since it currently only describes the weaker claim:

You wrote:
```python
    # ...and the writer really was exercised past its first (successful) call,
    # so the raising path was taken rather than skipped.
```

Change it to:
```python
    # ...the raising path was taken (call 2), and submission then STOPPED —
    # a third call would mean test 3 was run, judged and paid for after the
    # guard fired, which is what stop_submitting exists to prevent.
```

chesworthrm (approved):
Approving. The arithmetic is right and the evidence behind it holds up — I re-derived it rather than taking the PR body's word.

What I checked:

- **Corpus figures reproduce.** 276 judge draws with `cached > input` — the numerator matches yours exactly — and the old formula understates the committed corpus by 19.5% against your ~19%.
- **Both guards bite.** Restoring `- cached` fails three tests and prices the small-output shape at **-0.00257**, the exact negative it exists to catch. Making the flush guard re-raise fails the new test *after* printing `✓ [1/3]` and `✓ [2/3]`, and those two results then vanish — the reported failure exactly.
- **Ran against the merge with main**, not the branch head (2 behind, merge clean): harness pytest 2078 passed / 2 skipped, engine vitest 2262 passed, `apps/server` 157 passed / 1 skipped, ruff clean. One initial harness failure was a missing engine build in a fresh worktree; `npm run build` cleared it.
- **Both plan corrections are real.** #943 did already move the `harness_error` return below promote-to-scratch, and there were exactly two stale spec sites — I grepped for a third and there isn't one.
- `_compute_cost` is the only judge-cost site in the repo (skill cost comes straight from the SDK), so nothing else carries this bug.
- The 19% rise does reach the suite cost cap via `total_cost_usd`, but judge cost is ~4% of that, so the cap moves under 1%. "Effectively unaffected" is accurate.

One thing to fix before merge, in the comment below. It's a one-token change.

PR #1613, opened 2026-08-14
record-extraction: settle marriage/burial classification doctrine + #1407 matchers (#1443, #1407)

### LINKED ISSUE #1443: record-extraction: marriage/death-cert classification has real run-to-run instability, not fully settled
**IMPORTANT**: If you do this, do #1407 at the same time. Both edit `eval/tests/unit/record-extraction/`, so landing them apart buys two paid `make eval-skill SKILL=record-extraction` runs and two full 27-test annotation passes instead of one.

**Touches:** packages/engine/plugin/agents/record-extractor.md, eval/tests/unit/record-extraction/marriage-index-no-parents-recorded.json, eval/tests/unit/record-extraction/marriage-index-cross-state.json, eval/tests/unit/record-extraction/burial-index-parents-indirect.json

Found while verifying #995/#1131's fixes — ran `record-extraction`'s full
27-test suite 5 times in one session (identical code between the last 3
runs) to confirm two specific fixes held. They did, but three *other*,
pre-existing tests showed real instability that a single prose fix won't
settle:

- **`ut_record_extraction_021`** (marriage index, no parents recorded)
  failed three different ways across three runs: wrong role naming
  (`head_of_household`/`wife` instead of `groom`/`bride`), then
  `informant_proximity: "unknown"` instead of `"self"`, then a third
  shape. Never the same failure twice.
- **`ut_record_extraction_024`** (burial index, indirect parents) flipped
  `partial` → `fail` → `partial` across the same three runs, on the
  father/mother-of-deceased name classification.
- **`ut_record_extraction_015`/`028`** (marriage records, parent-name
  evidence_type) — #1131's scoping fix (informant clarification for a
  bare marriage index) verifiably resolved the *regression that fix
  itself risked introducing* — confirmed by two clean runs after the fix
  landed — but the same check failed again on `ut_015` in a later,
  otherwise-identical run. The underlying doctrine is right; the skill
  doesn't reliably comply with it yet.

## Why this is its own issue, not folded into #995/#1131

Those two issues are closed on their own explicit asks. This is a
different, broader finding: `record-extraction`'s handling of marriage and
death/burial records has more inherent run-to-run variance than census
handling does, and it's not visible from a single run — it took 5 samples
in one session to see the pattern.

## Ask

Investigate with more samples than one session's budget allows (a
description-optimizer / `runs_per_test > 1` pass per `unit-test-spec.md`
§7, once that phase is reached, would surface this precisely). Until then,
treat any single green run on these three tests as weak evidence, and any
single red run as not necessarily a regression.

### LINKED ISSUE #1626: record-extractor: two passages contradict on whether pre-1880 household relationships get assertions at all, and the skill skips them
**Touches:** packages/engine/plugin/agents/record-extractor.md, eval/tests/unit/record-extraction/census-1870-indexed-surname-variant.json

**Exemption for filing rather than fixing in the PR that found it (CLAUDE.md § "Work you find along the way"):** it depends on a decision only the lead can make — validating any fix needs several paid `make eval-skill` / `--test` samples, and the lead declined further runs on 2026-08-14 after two. A fix was attempted in PR #1613 and **provably did not work**; shipping a second blind attempt would repeat the error.

## The defect

`ut_record_extraction_022` (1870 census, household of eight) persisted **zero** relationship assertions in both 08-14 runs, against **14** in `v1_2026-08-13_09-36-36`. `ut_record_extraction_018` dropped 10 → 3 over the same interval. Not sampling noise — the agent states its reasoning:

> "**No relationship assertions** were written — the 1870 census had no relationship-to-head column, so household co-residence alone cannot establish parent-child or spousal bonds"

## The contradiction it is resolving

Two passages in `record-extractor.md` disagree, and the model picks a side per run:

- **Step 2** (written for a differently-surnamed household head, but phrased absolutely): "Never assert a specific relationship without evidence", and numbering people after the head "fabricates a parent-child link the record never states (pre-1880 has no relationship column)".
- **The classification section, ~440 lines later**: pre-1880 household relationships **are** assertions — "classify `indirect` with `relationship_type: \"child_inferred\"` (or `\"spouse_inferred\"`)".

Both readings are defensible from the text. That arbitration is what #1443 filed as run-to-run instability, and it is the same shape as the two contradictions #1443's PR did fix (a missing `name` in the parties' `self` enumeration; the marriage-event informant).

## What has already been tried, so it is not retried blindly

Commit `5c990e44` (in PR #1613) scoped the Step 2 prohibition to cross-group and outside-the-record links and stated the nuclear family's links must be persisted. A single `--test` scratch run looked decisive — ut_022 went to 13 relationship assertions and a clean pass on all six dimensions. **The next full run produced zero again, with the reasoning repeated verbatim.** The edit is retained in #1613 only because reverting it would flip the kept run log stale and owe another run; it restates existing doctrine, so it is dead weight rather than a hazard. **Treat it as ineffective, not as a partial fix.**

The lesson is part of the ticket: on this skill a single green run cannot validate a prompt change. Whoever takes this should budget N samples of ut_022 via `--test` (~$0.30 each, gitignored, no `runs_per_test` policy change needed) and compare rates, not one run.

## Also worth pinning while here

`census-1870-indexed-surname-variant.json` carries a `relationship` matcher for **`child_1` only**, and it is `optional: true` — which is why a total collapse to zero relationship assertions passed the validator silently and surfaced only as a judge Completeness dock. Whatever fixes the doctrine should also decide whether that matcher stays `optional`; existence genuinely flips 0 ↔ 14, which is the documented use of `optional`, but it means nothing deterministic guards the collapse. PLAN.md §1 called for mirroring the place/date/**relationship** triplet across all eight roles and only the first two were mirrored (noted in PR #1613's drift check).

## Provenance

Found while landing #1443 + #1407 (PR #1613); full evidence in that PR body, cause 4.

Emruthwill:
Closes #1443. Closes #1407. Closes #1626.

Landed together because all three edit `eval/tests/unit/record-extraction/` or
`record-extractor.md`, and landing them apart buys a paid
`make eval-skill SKILL=record-extraction` run and a full annotation pass each.

## The decision this PR implements

**A pre-1880 census asserts no family relationships.** Decided by the lead on
2026-08-15:

> A pre-1880 census should not extract family relationships simply because the
> household looks like a plausible nuclear family. Since there is no
> relationship column, the census only directly records that the individuals
> lived in the same household. Names, ages, sex, birthplace, and residence can
> be extracted, but parent-child or spouse links should be left to downstream
> correlation and inference rather than asserted as indirect or `_inferred`
> during extraction.

That answers the open question #1626 was filed to hold. An earlier revision of
this body said the decision was still needed; that is no longer true, and the
contradiction between the body and the validator docstring was a review
finding, now fixed.

## Why it was the fix

`record-extractor.md` carried two readings of "never assert a relationship
without evidence" — omit the link, or assert it labelled `indirect` — and the
model arbitrated per run. ut_022's relationship count across six observations:

| Run | relationship assertions |
|---|---|
| `v1_2026-08-10_12-56-30` | 13 |
| `v1_2026-08-10_15-17-57` | 1 |
| `v1_2026-08-13_09-36-36` | 14 |
| `v1_2026-08-14_11-29-08` | 0 |
| `scratch_2026-08-14_15-32-44` | 13 |
| `v1_2026-08-14_16-03-27` | 0 |

That swing **is** the run-to-run instability #1443 was filed about.

## The result

**Active run log: `v1_2026-08-15_12-52-37` — 19 pass / 4 partial / 2 fail /
2 aborted of 27.** Annotated in full (135 dimensions, no judge score
overridden).

All **eight** gated pre-1880 census tests wrote **zero** relationship
assertions and pass `test_pre_1880_census_creates_no_relationship_assertions`:
ut_001, ut_002, ut_003, ut_005, ut_013, ut_014, ut_020, ut_022. Seven passed
outright; ut_020 aborted on a startup stall, so it is not evidence either way.

This is a deterministic validator result, not a judge opinion — it cannot flip
on phrasing.

Two run logs are added. `v1_2026-08-14_16-03-27` is **superseded and stale**;
it predates the reversal and is retained only as the record of the run that
preceded it. Read the 08-15 log for anything about this PR's outcome.

## The eight-fixture reach, found by replay not grep

Static search for `_inferred` found six fixtures. Replaying the new validator
over the committed run logs found two more: **ut_005** (`record-read-via-ark`,
tagged `1850-census`, no `_inferred` prose, two relationship assertions in
every run) and **ut_002**.

Sites: `record-extractor.md` (4 passages), the reference doc,
`rubric.md` (4 passages — the grading side had to flip in lockstep or the
judge would have penalised compliance), `research-schema-spec.md` (the
`_inferred` convention SURVIVES for downstream correlation; it is scoped so
extraction never writes one), eight fixtures, and the validator.

`5c990e44` — which scoped the prohibition the OTHER way — is superseded here
rather than shipped and fixed afterwards, per review.

## ut_013 keeps its premise

It still requires the record-vs-tree children discrepancy to be flagged as an
identity question. It no longer requires manufacturing ParentChild assertions
to get there.

## Also in this PR

- **#1407's matchers and validators** — ut_022's roles extended to all eight,
  ut_009's four adjudicated decedent rows, and the compound-birth-value guard.
- **Two marriage doctrine fixes** — `name` added to the parties' `self`
  enumeration, and the marriage-event informant aligned across four artifacts.
- **A harness retry guard** (`eval/harness/`) — a wall-clock abort with zero
  assistant turns is a startup stall, not a slow test, and is now retried.
  Two tests were lost to this in the 08-15 run.

**On the harness change riding a genealogist PR** — a fair challenge, since
#1625 was deferred on exactly that reasoning. The difference is blast radius
and reversibility, not lane: this changes *retry* behaviour on aborts, which
produce no gradeable result and so cannot alter any score; `eval/harness/` is
outside the run-log snapshot, so it invalidates nothing; and it is covered by
tests that drive the real timeout path. #1625 changes *grading* for every
skill and needs its own validation run. If a reviewer would still rather see
it split out, say so and I will move it.

## Known limitations of this run

- **`judge_prompt_hash` is stale** on the active log (rule 2b, warn-only).
- **Three tests produced no graded dimensions** — ut_018 and ut_020 aborted,
  and ut_021 failed its validator, which suppresses the judge. So there is
  less calibration signal here than 27 tests implies.
- **ut_021** (marriage `self` proximity) and **ut_016** (suspect name
  finalised rather than deferred) fail. Neither is the reversal. ut_021's fix
  is written and queued on #1631, unproven.

## What is not claimed

That one run settles the doctrine. It is much stronger than a single sample —
eight independent fixtures, deterministic check, and the validator provably
rejects the old shape — but the marriage tests in this same suite have moved
between runs on identical text, and that caution is the point of #1443.

Whether the doctrine is genealogically *right* is the lead's call and was
made; this PR is the consistent application of it.


promise-emmanuel (changes requested):
This PR flips pre-1880 census relationships from "assert as indirect/_inferred" to "don't assert at all", and applies it consistently across seven places — the agent body, the reference doc, rubric.md, eight fixtures, the validator, and research-schema-spec.md. The validator docstring says "Decided 2026-08-15 (issue #1626)". But issue #1626 is still open, and the PR body still says this "needs an architecture decision this PR cannot make". Those can't both be true. Get the decision confirmed, close issue #1626, and fix the body — right now a reviewer can't tell whether anyone signed off.

The new retry guard retries every wall-clock timeout, not just stalls

usage is only ever set inside the ResultMessage branch of _consume_messages. asyncio.wait_for cancels that coroutine on timeout, so no ResultMessage arrives and usage is always {} here. _is_zero_progress_timeout therefore reduces to aborted_reason == "max_wall_clock_seconds". Every wall-clock abort in the run-log corpus shows num_turns=0 — there's no counter-example to find, because the field can't be populated on this path.

The cap is 1500s and DEFAULT_SKILL_RUN_ATTEMPTS is 3, so a genuinely slow test now burns up to 4500s and 3× the tokens in a paid run. That's what the comment you replaced was there to prevent.


skill_runner.py:488
You wrote:      usage: dict[str, Any] = {}
Change it to:   usage: dict[str, Any] = {}
                turns_seen = {"n": 0}

skill_runner.py:524
You wrote:                  if isinstance(message, AssistantMessage):
Change it to:               if isinstance(message, AssistantMessage):
                                turns_seen["n"] += 1

skill_runner.py:570-572
You wrote:      except asyncio.TimeoutError:
                    aborted_reason = "max_wall_clock_seconds"
                    error = f"wall-clock timeout after {max_wall_clock_seconds}s"
Change it to:   except asyncio.TimeoutError:
                    aborted_reason = "max_wall_clock_seconds"
                    error = f"wall-clock timeout after {max_wall_clock_seconds}s"
                    usage["num_turns"] = turns_seen["n"]
The two tests asserting the fix is narrow build a SkillRunResult shape the producer can't emit, so they'll pass forever while the guard is blind. Add one that goes through the real timeout path.

The agent body still contradicts itself on the rule this PR is settling

record-extractor.md:222-224. Twelve lines above, the same bullet says a pre-1880 census must "emit NO parent-child or spousal relationship assertion, in any form."


You wrote:      states (pre-1880 has no relationship column). This is a role-*naming*
                rule; it does not license skipping the relationship assertions
                themselves. A co-resident family
Change it to:   states (pre-1880 has no relationship column). A co-resident family
This is the sentence from 5c990e44 you kept on the grounds that it "asserts nothing false". That was true under the old doctrine. It isn't now. And it's the exact mechanism this PR blames for ut_022 swinging 0–14: one prompt carrying both readings. All seven gated census tests passed in the 08-15 run with it present — but n=1 is the standard this PR argues against.

Removing it flips the active run log stale and owes another paid run, so that trade is the lead's call.

The body describes a run that's been superseded

The PR adds two run logs, not one. v1_2026-08-15_12-52-37 is the active one; the 16-03-27 log the body calls "the only one ACTIVE against this branch" is stale on ten files. The headline 15/7/5 table is from the stale log. The active run is 19 pass / 4 partial / 2 fail / 2 aborted, with all seven non-aborted pre-1880 census tests passing — which is the result that actually supports the doctrine change. Since the body says this evidence "exists nowhere else", it needs to be right before merge.

Smaller things

The orchestrator retry isn't declared anywhere in the body, and it's the same shape you used to defer issue #1625 — a developer-lane harness change altering behaviour for every skill, riding a genealogist annotation pass. Worth a line on why the reasoning differs.

The active run log's judge_prompt_hash is stale, and three tests produced no graded dimensions (ut_018, ut_020, ut_021), so there's less calibration signal here than 27 tests suggests.

Whether the doctrine itself is genealogically right is a senior genealogist's call, not mine — I checked that it's applied consistently, not that it's correct.

Emruthwill:
@promise-emmanuel — both of your blocking points are now addressed. Requesting a re-review.

**1. The retry guard retrying every wall-clock timeout — fixed.**

Exactly as you prescribed: `turns_seen` counts AssistantMessages as they stream, and the `TimeoutError` handler records `usage["num_turns"]` from that counter, because `usage` is otherwise populated only in the `ResultMessage` branch and `wait_for` cancels the consumer before it arrives.

One thing I added beyond the prescription: the predicate now requires **`num_turns == 0` explicitly**, not falsy. A missing count means the recording handler did not run, so a stall is indistinguishable from slow work — and it **fails closed** (no retry) rather than retrying everything. If that handler is ever removed, the guard reverts to the old never-retry behaviour instead of silently over-retrying.

You were also right that my tests were vacuous: they hand-built `usage={"num_turns": 7}`, a shape the producer cannot emit on that path, so they passed while the guard was blind. New tests drive the **real** timeout path — a stubbed `query()` streams N assistant messages then hangs against a 1s clock. Neutering just the `usage["num_turns"]` line reddens them; the old hand-built cases stay green, which is your point about their vacuity made mechanical.

**2. "Get the decision confirmed, close issue #1626, and fix the body" — done, with one deliberate exception.**

The ruling is now **recorded on #1626**: https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1626#issuecomment-5322695856 — quoted in full, with its author and date, so it is checkable rather than asserted. The validator docstring (`05d6133c`) now cites that record instead of claiming "Decided 2026-08-15", and says what it used to say and why that was not good enough.

I also corrected a date I got wrong in that record on first posting: I wrote 2026-08-16, which is the `evidence_type` ruling in #1628. The pre-1880 ruling is **2026-08-15**, matching commit `5890b9f8` and every citation in the code.

**The exception: I did not close #1626.** @T-FEH's review of #1631 offered "close it with the decision recorded" *or* "get @DallanQ to confirm". Recording it is mine to do; deciding whether @Emruthwill's ruling is sufficient or whether @DallanQ's confirmation is the bar is a question about how this team allocates doctrine authority, and closing the ticket would settle that by fiat. The issue comment names @DallanQ as the person to confirm or overrule. If you read the bar differently, say so and I will follow it.

The body carries no "needs a decision" claim — that was corrected earlier, along with the run-log description, which had described the superseded 16-03-27 log rather than the active one.

All 12 checks green. This PR is `BEHIND` main; happy to merge up on request, but it is first in the stack so it merges cleanly as-is.

EdmondOware:
Checking in on this since it is currently blocking the whole stack, not just this PR. #1631 and #1706 both sit on top of this branch, and T FEH already confirmed on #1631 that the retry guard fix landed correctly here too. That part is resolved.

The other blocking point is not. The validator docstring still says decided 2026 08 15, issue #1626, but #1626 is still open with zero comments, and this PR's own body still says it needs an architecture decision this PR cannot make. Those two statements cannot both be true, and right now there is no record anywhere that anyone actually signed off on flipping pre 1880 census relationships from assert as indirect to do not assert at all. That is not a small edit, it touches the agent body, the reference doc, the rubric, eight fixtures, the validator, and the schema spec.

Whoever has authority to make that call needs to actually comment on #1626 confirming it, or this needs to revert to the prior doctrine. Until that happens, this PR, #1631, and #1706 are all stuck behind it regardless of how clean the rest of each review is.

Emruthwill:
Both blocking points are addressed — after your review, so you would not have seen them. Timeline, so it is checkable rather than asserted:

| Event | When |
|---|---|
| Your review | 2026-08-16 06:01 |
| Decision recorded on #1626 | 2026-08-18 02:25 |
| Docstring corrected (`05d6133c`) | 2026-08-18 03:36 |

## 1. The decision is on the record

#1626 now carries the ruling, quoted verbatim with its author and date, alongside the evidence it rests on — ut_022's relationship count across six observations of the same fixture (13 → 1 → 14 → 0 → 13 → 0), which is what #1443 filed as run-to-run instability and what the reversal removes.

You are right that this was not a small edit and that a docstring is not authorisation. That criticism was correct and it is the reason the record now exists.

## 2. The docstring no longer asserts a decision

`validators/test_record_extraction.py` now reads:

> The ruling this enforces is recorded on issue #1626 — quoted in full, with its author and date, rather than asserted here. This docstring previously read "Decided 2026-08-15 (issue #1626)" while that issue held no record of any decision, which a reviewer correctly refused to accept as authorisation for a reversal spanning the agent body, the reference doc, the rubric, eight fixtures, this validator and the schema spec.

The claim is now verifiable at its source instead of restated in code.

## 3. The PR body no longer says a decision is needed

That phrasing is gone — `grep` for "architecture decision" / "cannot make" over the current body returns nothing. The body now leads with the ruling and its provenance.

## One thing deliberately left open, and it is yours and @DallanQ's, not mine

**#1626 is still open.** I recorded the decision but did not close it, because you and @T-FEH both said "get @DallanQ to confirm" — and whether @Emruthwill's ruling stands on its own or needs @DallanQ's sign-off on top is a question about how this team allocates doctrine authority. Closing the ticket would have settled that question by fiat, which is not mine to do.

So: there is now a record of the decision. Whether that record constitutes the sign-off you were asking for is the part still open, and it is a one-line answer from either of you.

@DallanQ — if your confirmation is the bar, the comment on #1626 is the thing to confirm or overrule.

## Also, for the record: one claim in this body was wrong and is retracted

Separate from your points, and worth flagging since I argued it to a reviewer. I described ut_014 and ut_020 as a "mirror image — two tests, opposite directions" and used that pairing as evidence the failure was not single-run variance. It was not a pattern. ut_014 is a **role-assignment** defect (all four birthplaces were correct; the extractor numbered a 42- and a 38-year-old as children of a 62-year-old head), and ut_020 is a genuine value bleed with **one** observed instance. Two unrelated defects, treated as one. The bodies of #1631 and #1706 carry the correction.

Ready for re-review whenever you are.

Emruthwill:
Closing this in favour of **#1631**, per @DallanQ — #1631 contains every commit of this PR, so merging it delivers all of this work.

Verified before closing rather than assumed:

- `git merge-base --is-ancestor` passes, and **0 commits** of this branch are missing from #1631
- the pre-1880 reversal is present in `record-extractor.md`
- `test_pre_1880_census_creates_no_relationship_assertions` is present
- the **corrected docstring** (`05d6133c`, citing the ruling recorded on #1626 instead of asserting it) is present
- this PR's run log `v1_2026-08-16_11-26-54.json` **and its annotation** are both committed on #1631's branch

So nothing is lost, including the graded evidence.

@promise-emmanuel — your two blocking points are both resolved, and the fixes live on #1631 now:

1. **The retry guard** — fixed in `f436c193`; it no longer retries every wall-clock abort. @T-FEH confirmed this landed correctly.
2. **The #1626 decision record** — the ruling is now recorded on #1626 with its author, date and the ut_022 evidence, and the docstring cites it rather than asserting it. Your criticism was correct: a docstring is not authorisation, and that is why the record exists.

One thing deliberately still open, and it is @Emruthwill's and @DallanQ's rather than mine: **#1626 remains open.** The decision is recorded; whether that record constitutes the sign-off you asked for is a question about who holds doctrine authority here, so I did not settle it by closing the ticket.

Review continues on #1631.

PR #1614, opened 2026-08-14
Cut the duplicate Windows wrappers; move the rest to scripts/windows

DallanQ:
Follow-up to PR #1437, which added 45 `.bat` wrappers to `scripts/`. Both reviewers checked command fidelity target-by-target and found nothing wrong; nobody asked whether the set should exist. Three things it turned out to be worth asking:

1. **16 were a second Windows spelling of an `eval\` script that already existed** — `e2e-run.bat` and `RunE2E.bat` both run `e2e.run_e2e --test`, `eval-skill.bat` and `RunTests.bat` both run `run_tests.py --skill`, and so on through `GateSkill`, `BuildMcpb`, `BuildPlugin`, `Viewer`, `Start`, `CheckSetup`, `Login`. Six docs point at the `eval\` one. Two Windows entry points to a $3–10 fixture run is a support cost.
2. **No doc referenced any of the 45.** The only index was `scripts\help.bat`, which nothing pointed at either.
3. **Nothing checked them against the Makefile**, and one had already drifted.

## What this does

**Deletes 25.** The 16 duplicates, `help.bat` (69 hand-written `echo` lines shadowing a `make help` that generates itself), and 8 maintainer analysis targets — `e2e-corpus`, `e2e-latency`, `skill-latency`, `eval-timings`, `e2e-guardrail-shadow`, `prune-runlogs`, `e2e-author`, `e2e-thinking-probe`. Those are cited only as `make <target>` in specs, ADRs and the `find-big-wins` skill; nothing sends a Windows user to them.

**Keeps 20, moved to `scripts/windows/`.** The ones a Windows developer genuinely cannot otherwise reach: install/reinstall/clean-deps, the four server and two web targets, db-reset, the five test suites, harness-lint, agent-smoke, deploy-status, engine-build. `scripts/` is a build-and-plumbing directory again instead of 47 `.bat` files with `build-mcpb.mjs` buried among them.

**Adds `scripts/windows/test-all.bat`** — the pre-PR gate the PR template requires, and the one wrapper with a real price. Runs the same six suites in the same order, reports every failure rather than stopping at the first, and names the missing tool or uninstalled dependency up front. Closes #1185, and the PR template no longer tells Windows users no equivalent exists.

**Fixes to files that survived.** The first two were in the original push; the rest came out of review.

- `engine-build.bat` ran `npm ci` only when `node_modules` was *absent*, where the Makefile stamp reinstalls whenever `package-lock.json` changes. A pulled lockfile bump built against stale deps and the machine then disagreed with CI, silently. It now compares the lockfile against a copy kept inside `node_modules`, and defaults to installing when the comparison can't be made, so a surprise costs a slow rebuild rather than a stale one.
- `install.bat` told Windows users to install git hooks from Git Bash because the Makefile target "uses bash syntax with no native cmd.exe equivalent". `eval\InstallHooks.bat` is that equivalent — PR #753 added it a month ago. It gets named instead.
- **Every wrapper now invokes `npm`/`pnpm`/`npx` through `call`.** They resolve to `.cmd` batch files, and cmd.exe *replaces* the running script when one batch file invokes another without `call`. A bare `npm ci` therefore ended the wrapper. `engine-build.bat` never reached `npm run build` — and because the new lockfile stamp is written below that same `npm ci`, every run took the install branch and none ever built. `install.bat` stopped after `pnpm install`, skipping the server venv, the engine build and the eval-ui deps; `test-all.bat`'s preflight then sent the user back to `install.bat` in a loop. Every `eval\*.bat` already used `call`; none of `scripts\windows\` did.
- `reinstall.bat` still called `scripts\clean-deps.bat` and `scripts\install.bat`, both of which this PR deletes. It now calls its siblings through `%~dp0`, like every other wrapper.
- `typecheck.bat` was the one wrapper of 21 the move carried across without updating its `cd`, so it resolved `scripts\` rather than the repo root. It worked only because pnpm walks upward to find the workspace manifest.
- `engine-build.bat` and `install.bat` now read `errorlevel` before `popd` rather than after, so a failed install cannot write the stamp that makes later runs skip it.
- `clean-deps.bat` and five `REM Pair with:` headers named pre-move `scripts\` paths.

## Documentation

- `scripts/windows/README.md` — the index that replaces `help.bat`, with the target table.
- `DEVELOPMENT.md` — an "On Windows" block under Build commands, and the Windows path named in Running all tests.
- `README.md` — added to the contributor reading list.
- `Makefile` — header points at the wrappers; the `e2e-run` sync comment named a file this PR deletes and now names `eval/RunE2E.bat`, which is what actually needs keeping in sync.
- `.github/pull_request_template.md` — points at `test-all.bat`.

## The drift check

`packages/engine/mcp-server/tests/packaging/windows-wrappers.test.ts`, under `make engine-test`, so it runs in the pre-PR gate and in CI. Fifteen assertions:

- every wrapper declares the target it mirrors, in a `REM Windows equivalent of: make <target>` header
- that target exists in the Makefile
- the filename matches the declared target
- `scripts/windows/README.md` lists exactly the wrappers on disk, and pairs each with the target it declares
- no wrapper claims a target owned by an `eval\` script
- every `eval\` script the guard names still exists, and every `eval\` script on disk is accounted for — so adding one forces a decision here rather than escaping the guard
- every wrapper resolves the repo root, calls its siblings through `%~dp0`, and runs `npm`/`pnpm`/`npx` through `call`
- no `.bat` sits in `scripts\` itself other than the two feedback-case scripts, which take an argument and wrap no target
- two reader guards, so a parser that silently matches nothing fails instead of passing

The last four came out of review: the first eleven were all green while `reinstall.bat` called two deleted files, `engine-build.bat` never built, and a planted `scripts/foo.bat` passed every one of them. What it still can't check is whether a wrapper's *commands* match its recipe. The wrappers reimplement recipes rather than shelling out to `make`, and that stays a review responsibility; `scripts/windows/README.md` and the Makefile header both say so.

**Verified it fails.** Nine breaks, each firing the assertion meant to catch it. The last three were run against the branch as first pushed, where they fail on the real defects:

| Break | Fires |
|---|---|
| Renamed `typecheck:` in the Makefile | declared target exists |
| Re-added `scripts/windows/mcpb.bat` | duplicate guard + README set + row count |
| Renamed `eval/BuildMcpb.bat` | named eval script exists + accounted for |
| Stripped a wrapper's header | wrapper declares a target + README pairing |
| Mispaired a README row | README pairing |
| Added an unlisted `eval/RunSomething.bat` | accounted for |
| `typecheck.bat`'s original one-level `cd` | wrapper resolves the repo root |
| `reinstall.bat`'s original `call scripts\install.bat` | sibling call goes through `%~dp0` |
| Dropped `call` from `npm run build` | npm/pnpm/npx goes through `call` |
| Planted `scripts/mcpb.bat` | no wrapper escapes into `scripts\` |

## Testing

`make test-all` passes on the merge with `main` — 2277 engine tests including the 15 new ones, 2075 harness tests, 157 control-plane tests, `turbo test typecheck` 8/8.

**The batch changes are still not tested on Windows.** I wrote them on macOS, and three of the four defects review found were things one run on the target platform would have caught immediately. `install.bat`, `reinstall.bat`, `test-all.bat` and `engine-build.bat` all want a double-click before merge. Failure modes are safe — `engine-build.bat` defaults to reinstalling, and `test-all.bat` exits before running anything if a tool or install is missing.


chesworthrm (commented):
Reviewed at 9cef46c. Test-merged with `main` first (branch is 1 commit behind), then ran the suites: engine vitest 2273/2273, `turbo test typecheck` 8/8, `apps/server` pytest 157 passed, `eval/harness` pytest 2075 passed. The one harness failure on my first run was the fresh-worktree missing `build/`; it passed after `npm run build`.

The cut is the right call and the counts all check out — 21 wrappers on disk, 21 README rows, 25 net deletions. I also broke the new guard three ways (renamed `typecheck:` in the Makefile, re-added `scripts/windows/mcpb.bat`, added an `eval/RunSomething.bat`) and each break fired the assertion meant to catch it.

Two things the move left behind, and one gap in the guard.

## `scripts/windows/reinstall.bat` is broken

Lines 6 and 8 still call the pre-move paths, and this PR deletes both:

```
call scripts\clean-deps.bat
call scripts\install.bat
```

`make reinstall`'s Windows equivalent now prints two "cannot find the path" errors and does nothing. Every other wrapper that calls a sibling already uses the `%~dp0` form — `install.bat:12` is `call "%~dp0server-install.bat"`. This one was missed.

```
You wrote:      call scripts\clean-deps.bat
Change it to:   call "%~dp0clean-deps.bat"

You wrote:      call scripts\install.bat
Change it to:   call "%~dp0install.bat"
```

Same stale path in what `clean-deps.bat` prints on the way out:

```
You wrote:      echo Done. Run scripts\install.bat to reinstall from scratch.
Change it to:   echo Done. Run scripts\windows\install.bat to reinstall from scratch.
```

## `scripts/windows/typecheck.bat` is one level short

It is one of the two files the move carried across unchanged, and it is the only wrapper of the 20 that does not resolve the repo root — it lands in `scripts/`.

It happens not to break today: `pnpm` walks upward and finds the workspace root anyway (I ran `pnpm typecheck` from `scripts/` and it resolved to the root manifest and exited 0). But it is the first suite `test-all.bat` runs, and it stops working the moment anything relevant lands between `scripts/` and the root.

```
You wrote:      cd /d "%~dp0.."
Change it to:   cd /d "%~dp0..\.."
```

`deploy-status.bat`, the other zero-change move, is fine — it never `cd`s.

## The guard passes on both of those, and on the regression it exists to prevent

Green is green on this branch, which is how the two above got here. The guard also does not check *location*: I dropped a `scripts/foo.bat` declaring a nonexistent target and all 11 assertions still passed. That is the exact shape of the thing this PR is undoing, so nothing stops the next PR putting the wrappers straight back into `scripts/`.

Three more assertions close all of it. I ran each one: the first two fail on this branch as it stands, naming `typecheck.bat` and `reinstall.bat` respectively; the third passes clean and fails on a planted `scripts/foo.bat`.

```ts
/**
 * `.bat` files allowed to stay in `scripts/` rather than `scripts/windows/`.
 * Both take the zip or case directory as an argument, so neither wraps a
 * make target.
 */
const SCRIPTS_ROOT_BAT = new Set([
  "reset-feedback-case.bat",
  "setup-feedback-case.bat",
]);

it("every wrapper resolves the repo root, not scripts/", () => {
  const wrong = wrappers.filter((f) => {
    const cd = read(join(windowsDir, f)).match(/^cd \/d "%~dp0(.*)"/m);
    return cd !== null && cd[1] !== "..\\..";
  });
  expect(wrong).toEqual([]);
});

it("every sibling wrapper call goes through %~dp0", () => {
  const stale = wrappers.flatMap((f) =>
    [...read(join(windowsDir, f)).matchAll(/^call\s+(?!"?%~dp0)(\S*\.bat)/gm)]
      .map((m) => `${f}: call ${m[1]}`),
  );
  expect(stale).toEqual([]);
});

it("no wrapper escapes into scripts/ instead of scripts/windows/", () => {
  const stray = readdirSync(join(projectRoot, "scripts"))
    .filter((f) => f.endsWith(".bat") && !SCRIPTS_ROOT_BAT.has(f))
    .sort();
  expect(stray).toEqual([]);
});
```

## Smaller

**`engine-build.bat:32`** — the new stamp is what makes later runs skip `npm ci`, so writing it after a failed install reintroduces the stale-deps mode this change exists to remove. The `npm ci` errorlevel is read after `popd`, and whether `popd` preserves it is the one thing I could not test from macOS. Checking before `popd` costs nothing and removes the question:

```
You wrote:      pushd "%ENGINE%"
                npm ci
                popd
                if errorlevel 1 ( echo ERROR: engine npm ci failed & exit /b 1 )

Change it to:   pushd "%ENGINE%"
                npm ci
                if errorlevel 1 ( popd & echo ERROR: engine npm ci failed & exit /b 1 )
                popd
```

**`.github/tests/codeowners-routing.test.mjs:91`** — cites a path this PR deletes. The row exercises the `*.bat` rule by pattern so nothing fails; it just names a file that no longer exists.

```
You wrote:      ['scripts/mcpb.bat', D],
Change it to:   ['scripts/windows/test-all.bat', D],
```

## What I did not check

Nothing ran on Windows, and that is the real gap — `reinstall.bat` and `typecheck.bat` are both defects one run on the target platform would have caught, and neither CI nor the drift test can see them. Add `reinstall.bat` to the double-click list alongside `test-all.bat` and the stamp block.

I read all 21 wrappers against the Makefile and `scripts/test.sh` — `test-all.bat` runs the same six suites in the same order, and the six suite wrappers match their targets — but I did not diff every operator flag in the server and web wrappers. I grepped the tree for all 26 deleted filenames and the only hits were the two stale references above, so the claim that no doc pointed at them holds inside the repo; I cannot speak to a bookmark outside it. I also did not check whether `.gitattributes` forces CRLF on the moved and new `.bat` files.

Cia-3 (commented):
Test-merged with main and re-ran the full suite set — engine vitest 2273/2273 (exact match to your numbers), tsc clean, harness pytest 2074/3 skipped. Also independently reproduced the header-strip mutation test — fires exactly the two assertions you named. The cut and the counts all check out.

I want to flag that chesworthrm's review (submitted at this exact commit, so nothing's changed since) found two real, verified bugs that are still unfixed:

scripts/windows/reinstall.bat calls scripts\clean-deps.bat and scripts\install.bat — both moved to scripts\windows\ by this same PR. I confirmed neither path exists anymore; this wrapper currently does nothing but print two path errors.
scripts/windows/typecheck.bat only does cd /d "%~dp0.." (one level) instead of ..\.. like every other wrapper. It happens to work today because pnpm walks up to find the workspace root, but it's silently one accident away from breaking, and it's the first suite test-all.bat runs.
Also worth a look: .github/tests/codeowners-routing.test.mjs:91 still references scripts/mcpb.bat, which this PR deletes.

Given these are real, demonstrated breaks in wrappers this PR itself touches, I'd hold off on merging until they're fixed — and per chesworthrm's note, neither CI nor the new drift guard can catch either one, so the three additional guard assertions they proposed (checking cd depth and that sibling calls use %~dp0) seem worth adding too.

PR #1622, opened 2026-08-14
validation: writer tools block only on call-introduced errors, not pre-existing drift (#1572)

### LINKED ISSUE #1572: research.json schema drift (person_id/notes on assertions, author/record_id/title on sources) hard-blocks tree_forget and any validating writer

### LINKED ISSUE #1701: Nothing catches a dropped before-snapshot clone in materialize_facts / merge_tree_persons
**Touches:** packages/engine/mcp-server/src/tools/materialize-facts.ts, packages/engine/mcp-server/src/tools/merge-tree-persons.ts, packages/engine/mcp-server/tests/tools/

`validateIntroduced` (#1572, PR #1622) blocks a writer only on errors its own call introduces, by diffing a pre-mutation before-snapshot against the after-state. Each in-place mutator must pass a pre-mutation `structuredClone` as `before`. If a refactor drops that clone, `before === after` (both point at the mutated object) so every introduced error is present in `before` too, gets demoted to a warning, and the tool wrongly returns `{ ok: true }`. Fail-open, silent, CI stays green.

Verified coverage (dropped the clone at each site, ran that tool's suite):

- `tree_edit` — 3 existing tests fail. Pinned.
- `materialize_facts` — all 22 pass. NOT pinned.
- `merge_tree_persons` — all 4 pass. NOT pinned.

The two unpinned tools have no organic "introduced error blocks" test to catch it: `materialize_facts` input-validates its refs before mutating, and `merge_tree_persons`' remap repoints refs correctly, so neither naturally produces a `validateParsed` error through the diff. The guard has to be manufactured: either a test that forces the introduced-error path for each, or a lint asserting the `structuredClone(...)` before-snapshot is present at those two call sites.

Raised in the #1572 senior review (Gennecis) on PR #1622.

clack391:
## Summary

- A `research.json` written in a legacy shape (`person_id`/`notes` on assertions, `author`/`record_id`/`title` on sources, `resolution_notes` on conflicts) failed `validateParsed`, and because every writer validates the whole document, one drifted section froze all nine writing tools even on a call that never touched it. Root cause behind #1476.
- New `packages/engine/mcp-server/src/validation/introduced-errors.ts` validates the pre-call snapshot too and blocks only on errors the call itself introduces; pre-existing drift rides as a warning. All seven writer files route through it.
- Tier: **Normal**. Closes #1572.

## Review guidance

**Start here:** `validation/introduced-errors.ts`. The load-bearing decision is the error-identity key: a naive `{path, message}` diff is wrong because a validation error's `path` carries an array index and the tree/merge/forget tools reindex arrays with `.filter`, so a pre-existing error at `persons[7]` becomes `persons[6]` after a removal and a naive diff reports it as new (the exact false-deny this fix removes). `normalizePath` rewrites each `key[index]` to `key[id=<.id>]` from the before/after docs so the diff survives reindexing. Check that keying, and the caller contract in the doc comment (a tool that mutates in place must pass a pre-mutation `structuredClone` as `before`).

**Unsure about:** nothing structural. The two acceptance-wording calls recorded under "Deviated from the plan" were reviewed by Gennecis; both are now resolved in 73ae4629 (see that section).

## Plan

**Didn't change:** `validateParsed` itself (it is called twice, unchanged; the diff lives in the new helper, so `validate-project-refactor-spec.md` §4's byte-identical-output guarantee stays true). No JSON-schema or `packages/schema` change: this is a validation-behavior change, not a schema change, so there is no new field or enum. `merge_warnings` keeps its answer-not-failure semantics (a merge that itself makes the tree invalid still returns `{ ok: false }`).

**Acceptance check:** five tests that fail on `main` and pass here:
- `tests/validation/introduced-errors.test.ts`: "tolerates pre-existing drift that reindexes when an earlier element is removed" (the mechanism; break-tested against a naive path diff).
- `tests/tools/tree-forget.test.ts`: "tolerates pre-existing tree drift that reindexes when a person is removed (#1572)" (the mandatory reindex guard, end-to-end through the tool's heal + mutate).
- `tests/tools/research-append.test.ts`: "a pre-existing unrelated drift does not block a write; it rides as a warning (#1572)".
- `tests/tools/research-log-append.test.ts`: "a pre-existing unrelated error rides as a warning and does not block the append (#1572)".
- `tests/tools/merge-warnings.test.ts`: "a drift-only project does not make the dry-run report a rejection (#1572 flip)".

**Deviated from the plan:**
- **LEAD stop-rule (contract reversal).** This reverses the documented "validate the whole project, write nothing on any validation failure" write-boundary contract (architecture §6.2, ADR-0011) to "block only on call-introduced errors." Raised with and approved by Dallan on #1572 before any code (sign-off 2026-08-14). Recorded in ADR-0011 as an instance of its own "prefer a false allow over a false deny" limit, and in ADR-0005 / architecture §6.2 / the six tool specs.
- **id-less elements keep their raw index** rather than "block on any unkeyable error" (an earlier plan draft's wording). Blocking id-less pre-existing drift would re-freeze exactly what this fix removes, so a pre-existing id-less error at an unchanged index is tolerated. Only an id-less element that reindexes would false-block, and no writer tool reindexes an id-less array, so that residual is theoretical.
- **Two acceptance-wording calls, both resolved in the Gennecis review (73ae4629):** (1) on a genuine failure the tools return only the call's introduced errors, not "N pre-existing errors" too, which is what avoids the error-wall that triggered tool-substitution in #1476. The success path now surfaces pre-existing drift as a single summary line ("this project has N pre-existing schema error(s) ... run validate_research_schema"), not one warning per demoted error, so the wall stays off both paths. (2) `merge_warnings` flips to `{ ok: true }` on a drift-only project and now surfaces the drift notice through its `sanitizeWarnings: string[]` slot, so the dry-run a researcher reads is not falsely clean. (The PR originally claimed the `PersonWarning[]` result had no slot for it; the review showed `sanitizeWarnings` is that slot, and it is now populated.)
- No standalone classification spec and no `formatIssues` consolidation (keep-it-simple): the helper's doc comment plus the architecture §6.2 reference document the contract.

## Test plan

- [x] `make engine-test`: 2365 passed on the current head, `tsc --noEmit` clean (2344 for the review fixes; +19 from `main`'s `project-create` suite merged in, no conflicts; +2 the clone-contract guard). Reindex guard break-tested (reverting the id normalization fails it); the rollback test break-tested (deleting either `cleanupSidecars` call turns it red); the clone-contract tests break-tested one at a time (dropping each tool's `structuredClone` reds only that tool's test). An independent second-context review found no correctness bug.
- [x] No live Cowork session or replayed bundle was run: this PR changes only the deterministic write-boundary validator (no skill or agent-prompt surface), so there is no LLM behaviour to confirm that way. #1476's improvise-vs-recover concern lives in the research skills, which this does not touch.
- [x] N/A: no skill run-log snapshot changed. `packages/engine/mcp-server/**` is not in `build_snapshot`, so no run log goes inactive and `check_runlogs.py` does not fire. No paid eval run.
- [x] N/A: no skill `description` frontmatter or DO NOT clause changed.

## Follow-on issues

One filed from the Gennecis review, the single thing that genuinely can't land here:
- **#1702** (developer, needs-decision): the option-2 `research.json` healer. Deferred because the per-key mapping for the six drifted keys is a lead decision plus six genealogical rulings (#1572's own framing), not mechanical, and it does not unblock the tools by itself.

Everything else was folded into this PR: the doc sweep (including `validate-project-refactor-spec.md` §9 Consumers and a new §11), the `merge_warnings` drift surfacing, the restored rollback test, and the before-clone guard for `materialize_facts` / `merge_tree_persons` (`introduced-clone-contract.test.ts`, one break-tested test per tool, closing #1701 (folded not filed); `tree_edit` stays pinned by its own tests).

**Folded in / filed instead:** folded in the doc sweep, the `merge_warnings` fix, the rollback test, and the clone-contract guard (closes #1701); filed only #1702.


taiwo-stack (approved):
Approved
 
 Ran the full suite on this head: tsc clean, vitest 2342/2342, harness pytest 2221/3 skipped. Caller contract checks out across all seven writer tools.

packages/engine/mcp-server/src/validation/introduced-errors.ts
```diff
@@ -0,0 +1,137 @@
...
+ * drift produces an empty `before` error set, so the result is byte-identical
+ * to calling `validateParsed(after, options)` directly.
+ */
+export async function validateIntroduced(
+  before: ProjectState,
+  after: ProjectState,
+  options?: { projectPath?: string },
+): Promise<ValidationResult> {
+  const [beforeRes, afterRes] = await Promise.all([
+    validateParsed(before.research, before.tree),
```
Gennecis: **The before-pass drops `projectPath`, so a whole class of pre-existing drift still hard-blocks every writer.**

Without it the sidecar and D5 checks never run on the before-snapshot, so every error they produce looks brand-new and blocks the call. Two cases I ran against this branch, both on `tree_forget` — which never reads `results/` or `research.log`:

- one orphan file in `results/`:
  `{ ok: false, errors: ["results/log_001.json: orphan sidecar — no log entry references it"] }`
- a log entry whose sidecar was pruned:
  `{ ok: false, errors: ["research.json/log (log_001): results_ref points at 'results/log_001.json' which does not exist or is invalid JSON"] }`

That is #1572's "the project is frozen", still reproducible — for orphan sidecars, missing `results_ref` targets, `returned_count` mismatches, and the D5 `record_persona_id` / `record_id` checks.

`validate-project-refactor-spec.md` §5 already rules on this: *"Omitting `projectPath` is the right choice only for a caller that genuinely has no project directory."* This caller has one — it's in `options`. The same section also proves the sidecar pass is **invariant under a merge**, so before and after agree.

```
You wrote:      validateParsed(before.research, before.tree),
Change it to:   validateParsed(before.research, before.tree, options),
```

I applied exactly that and ran the full suite: **2342/2342, no regressions**, and both cases above flip to `ok: true` with the drift correctly demoted to a warning.

Line 100 needs correcting too. "a sidecar error appears only in `after` and so always blocks, unchanged from before this fix" reads as a design choice, but it's the bug.

packages/engine/mcp-server/src/tools/merge-warnings.ts
```diff
@@ -72,7 +72,10 @@ export async function mergeWarnings(
     //     would reject is surfaced here as { ok: false } rather than reported
     //     as clean. research.json is read but never written.
     const research = await readProjectJson(projectPath, "research.json");
-    const validation = await validateParsed(research, merged, { projectPath });
+    // Dry-run: the before-project is the on-disk (research, tree); the after is
+    // (research, merged). A drift-only failure the merge did not introduce thus
+    // resolves to { ok: true } with the drift as a warning (#1572).
+    const validation = await validateIntroduced({ research, tree }, { research, tree: merged }, { projectPath });
```
Gennecis, line 78: **The drift notice can be surfaced here — the slot exists.**

The PR says the result "has no slot for the drift string" because it's `PersonWarning[]`-typed. `MergeWarningsSuccess` also carries a **`sanitizeWarnings: string[]`**, and its own comment says it's there *"so the LLM learns about dropped data BEFORE deciding to write, not after."* That is this case exactly, and the other six tools all surface these strings.

`merge_warnings` is the gate a researcher uses to decide whether a merge is safe. The `{ ok: false }` → `{ ok: true }` flip is right, but "clean, no notice" on a drifted project is the wrong half of it.

At **line 123**:

```
You wrote:      sanitizeWarnings: [...treeSanitized.warnings, ...candidateSanitizeWarnings],
Change it to:   sanitizeWarnings: [...treeSanitized.warnings, ...candidateSanitizeWarnings, ...formatIssues(validation.warnings)],
```

packages/engine/mcp-server/tests/tools/research-log-append.test.ts
```diff
@@ -412,12 +412,15 @@ describe("research_log_append", () => {
...
   });
 
-  it("writes nothing and leaves no orphan sidecar when the would-be project is invalid", async () => {
-    // Pre-existing dangling subject ref makes validation fail on append.
+  it("a pre-existing unrelated error rides as a warning and does not block the append (#1572)", async () => {
+    // subject_person_ids points at a person absent from the tree — a pre-existing
+    // project error this append never touches. Before #1572 it froze the append
+    // (and every other writing tool); now the append succeeds and the drift is
+    // surfaced as a warning, not a block. Rollback + sidecar cleanup on a GENUINE
+    // (call-introduced) failure is covered by the batch invalid-outcome test below.
```
Gennecis, line 420: **This coverage claim doesn't hold, and the branch it points away from is now untested.**

`outcome: "not-a-real-outcome"` throws `LogAppendError` at `research-log-append.ts:254`, so the batch invalid-outcome test exercises the **catch** path. `cleanupSidecars` has a **separate call at each of the two `!validation.valid` sites**, and after this rewrite nothing reaches either one — delete one and the suite stays green.

That's worth closing rather than noting, because a leftover orphan sidecar is precisely what causes the permanent freeze in my comment on `introduced-errors.ts`.

Add a test that fails validation **because of the call** — e.g. stage a payload whose `returned_count` disagrees with `results.length` — and assert `research.json` is byte-unchanged and `results/log_001.json` does not exist. That restores what the deleted test covered without re-freezing the tolerance case.

packages/engine/mcp-server/src/validation/introduced-errors.ts
```diff
@@ -0,0 +1,137 @@
...
+      });
+    } else {
+      introduced.push(e);
+    }
+  }
+
+  return {
+    valid: introduced.length === 0,
+    errors: introduced,
+    warnings: [...afterRes.warnings, ...preExisting],
```
Gennecis, line 151: **One warning line per demoted error, on every successful writer call, for the life of the project.**

I rebuilt the #1476 drift census from #1572's table (5 sources × 3 keys, 15 assertions, 1 conflict) and ran a single `research_append` update: **45** `pre-existing schema error, not caused by this call: …` lines came back on the success path.

That's the #1476 error wall relocated from the failure path to the success path, and it buries the warnings the agent actually has to act on — retention gaps, place-resolution misses, sanitize notes.

```
You wrote:      warnings: [...afterRes.warnings, ...preExisting],
Change it to:   warnings: preExisting.length
                  ? [...afterRes.warnings, { path: "", message: `this project has ${preExisting.length} pre-existing schema error(s) not caused by this call; run validate_research_schema for the list` }]
                  : afterRes.warnings,
```

With that in place your acceptance-wording call (1) is fine as-is — and this is also the "N pre-existing schema errors" wording #1572's acceptance asked for, without the wall.

docs/adrs/ADR-0011-put-guardrails-at-the-write-boundary.md
```diff
@@ -163,7 +163,11 @@ measured rather than argued.
    researcher from finishing correct work with no way around it, while a wrong
    allow leaves us exactly where we already are. Check only what is mechanically
    checkable, scope it conservatively, and write down what it knowingly lets
-   through.
+   through. #1572 applied this to the write gate itself: validating the whole
+   project and refusing every writer tool on a pre-existing drifted section —
+   even a call that never touched it — was a false deny reaching the researcher,
+   so the gate now blocks only on errors the call introduces
+   (`validation/introduced-errors.ts`).
```
Gennecis: **Worth one more sentence here: what option 1 knowingly leaves undone.**

#1572 says option 1 *"leaves the document invalid for downstream readers (the viewer, `packages/schema` consumers) until option 2 also happens."* That residual is genealogically material, because the tolerated fields hold real evidence — `sources[].author`/`title` is citation content, `assertions[].person_id` is the assertion-to-person linkage, `conflicts[].resolution_notes` is the researcher's resolution reasoning. After this PR all of it stays in the file, tolerated, and unread by the viewer.

#1572 closes with this PR and takes the only written record of option 2 with it, and I couldn't find an open issue covering a `research.json` healer. This ADR is the right place to record the residual, since it's already recording the reversal.

The healer itself is worth filing rather than folding: #1572 calls the per-key mapping *"a genealogical judgment, not a mechanical one"* — a lead decision plus six rulings, well past this PR.

packages/engine/mcp-server/src/tools/tree-edit.ts
```diff
@@ -640,6 +640,10 @@ export async function executeTreeOps(input: TreeEditInput, gate: OpGate): Promis
     );
     const tree = sanitized.tree;
     const research = await readJson(projectPath, "research.json");
+    // Post-heal, pre-edit snapshot (applyOperation mutates tree in place):
+    // block only on errors THIS call introduces, not pre-existing drift in a
+    // section it never touched (#1572).
+    const beforeTree = structuredClone(tree);
```
Gennecis, line 646: **Nothing tests the contract this clone satisfies — here, in `materialize-facts.ts`, or in `merge-tree-persons.ts`.**

Those three are the in-place mutators, so dropping a `structuredClone` in a later refactor makes the before/after diff empty and **silently demotes every error the call introduced** — the exact mode the helper's doc comment warns about, failing open with no test to catch it.

#1572's acceptance asks for the behaviour "stated per tool, with a test each" across nine tools; four have one. One test each here, or a `nothing-checks`-labelled issue naming the three.

docs/specs/match-merge-workflow-spec.md
```diff
@@ -235,7 +235,8 @@ folds a candidate document.
 Run `check-warnings` (final mode) on each affected anchor; verify
 `research.json` reference integrity. (`materialize_facts` and the `tree_edit`
 edge writes validate the would-be project before persisting and write nothing on
-error — `tree-materialization-spec.md` §4.1.)
+a call-introduced error (pre-existing drift rides as a warning) —
+`tree-materialization-spec.md` §4.1.)
```
Gennecis, line 239: **This citation now points at a section that wasn't updated.**

This line and line 523 both cite `tree-materialization-spec.md` §4.1 as the authority for the contract, but §4.1 wasn't touched — `materialize_facts` is the only one of the seven tools whose own spec this PR leaves alone. `tree-materialization-spec.md:184` still reads "validates the would-be project (`validate_research_schema` / `validateParsed`)".

Cheapest fix is to update §4.1 to match the sentence you've written here, so both citations resolve to something that says it.

Gennecis (changes requested):
Requesting changes. The core design is right and the reindex keying is real — I break-tested it (forcing `normalizePath`'s id branch off fails 3 tests, including the mandatory `tree_forget` guard), and I hand-checked the caller contract at all seven sites: it holds everywhere. `mergeGedcomx` clones, and the three tools that share a document by reference genuinely never mutate it.

Three things need fixing before merge, and they're all small:

1. The before-pass drops `projectPath`, so a whole class of pre-existing drift still freezes every writer. I reproduced #1572's exact symptom on `tree_forget` two different ways. One-word fix; I ran it and got 2342/2342.
2. `merge_warnings` can surface the drift notice — the slot the PR says doesn't exist does exist.
3. The rewritten `research_log_append` test removed the only coverage of the validator-failure rollback path, and the note explaining why cites a test that covers a different branch.

Details are inline. Three smaller notes inline too (warning volume, untested caller contract, a stale spec citation).

### Not inline, because the file isn't in the PR

**`docs/specs/validate-project-refactor-spec.md` was named by #1572 as a required edit site** and wasn't touched: *"writing the decision **and the option it beat** into `docs/specs/validate-project-refactor-spec.md` and ADR-0011 is part of this task."* ADR-0011 got its clause; this spec got nothing.

It matters twice over:

- Its **§5** is the section that rules on the `projectPath` question behind my first inline comment. It says omitting `projectPath` is right *only* for a caller with no project directory — worth reading before you apply that fix, because it also proves the sidecar pass is invariant under a merge.
- Its **§9 Consumers** is now stale: it still says the merge tools *"import `validateParsed` from `../validation/validator.js`"*. They import `validateIntroduced` from `../validation/introduced-errors.js`.

Options 2 and 3 are recorded nowhere in the diff.

### Verified

`npx vitest run` 2342/2342 and `tsc --noEmit` clean, in a worktree at c7a4586b (needs `npm install` first — 6 files fail to collect without it, e.g. missing `turndown`; environmental, not yours). All required CI checks green on this head. The snapshot claim in your test plan checks out — `snapshot.py:33` skips `packages/engine/mcp-server/src/`, so no run log goes inactive.

### Not checked

No harness or `apps/server` pytest — CI covers both and nothing in the diff reaches them. No live Cowork session and no replayed bundle: #1572 asks the PR to say which was done, and the test plan doesn't. So nothing here confirms a real session recovers rather than improvising, which was #1476's actual complaint.

packages/engine/mcp-server/src/validation/introduced-errors.ts
clack391: Fixed in 73ae4629. Applied exactly the change you ran: the before-pass now passes `options`, so the sidecar and D5 checks run on the before-snapshot and pre-existing drift is demoted instead of blocking. Reproduced both of your `tree_forget` cases (orphan sidecar, pruned `results_ref` target) and confirmed each flips to `ok: true` with the drift as a warning; full suite green. The line 100 comment is rewritten too, so it reads as the tolerated case now and cites §5 as the authority.

packages/engine/mcp-server/src/tools/merge-warnings.ts
clack391: Fixed in 73ae4629, at the line you named: `sanitizeWarnings: [...treeSanitized.warnings, ...candidateSanitizeWarnings, ...formatIssues(validation.warnings)]`. A drifted project now surfaces the notice through the gate a researcher reads, instead of coming back clean.

packages/engine/mcp-server/tests/tools/research-log-append.test.ts
clack391: Fixed in 73ae4629, though not via the `returned_count` vehicle, and here is the reason. `finalizeStagedSidecar` (results-staging.ts:160) recomputes `returned_count` from the staged payload and forces `log_id`, so a staged mismatch self-heals before validation ever sees it. The tool always builds a self-consistent sidecar and a schema-valid entry, so the `!validation.valid` path is unreachable through the public API (I walked all five write vehicles). The new file `research-log-append-rollback.test.ts` mocks `validateIntroduced` to fail, then asserts `research.json` is byte-unchanged and `results/log_001.json` is removed, for both the single-op and batch sites. Break-tested: deleting either `cleanupSidecars` call turns it red. That restores what the rewrite dropped.

packages/engine/mcp-server/src/validation/introduced-errors.ts
clack391: Fixed in 73ae4629. Collapsed to one summary line on the success path, regardless of how many errors are demoted: `this project has N pre-existing schema error(s) not caused by this call; run validate_research_schema for the list`. That is the wording #1572's acceptance asked for, and it keeps the #1476 wall off the success path. The introduced-errors test now asserts the single line.

docs/adrs/ADR-0011-put-guardrails-at-the-write-boundary.md
clack391: Done in 73ae4629: ADR-0011 gets the residual sentence naming the three tolerated fields that stay unread by the viewer until a healer lands (`sources[].author`/`title` citation content, `assertions[].person_id` the assertion-to-person linkage, `conflicts[].resolution_notes` the resolution reasoning). And I filed the healer as its own issue rather than folding it, on exactly your reasoning that the per-key mapping is a lead decision plus six rulings: #1702 (developer, needs-decision).

packages/engine/mcp-server/src/tools/tree-edit.ts
clack391: Checked all three. `tree_edit` is already pinned: dropping its `structuredClone` fails 3 existing tests. `materialize_facts` (22 tests) and `merge_tree_persons` (4 tests) are not: dropping the clone leaves both suites green, because `materialize_facts` input-validates its refs before mutating and `merge_tree_persons` remaps them correctly, so neither organically produces a `validateParsed` error through the diff. A guard for those two has to be manufactured, so I filed #1701 (developer, nothing-checks) naming both sites and the two options: a forced introduced-error test each, or a lint asserting the clone is present.

docs/specs/match-merge-workflow-spec.md
clack391: Fixed in 73ae4629: updated `tree-materialization-spec.md` §4.1 so both citations (match-merge-workflow-spec lines 239 and 523) resolve. §4.1 now says `materialize_facts` validates through `validateIntroduced` (which wraps `validateParsed`), blocking on a call-introduced error while pre-existing drift rides as a warning.

clack391:
Thanks for the depth here, the reindex break-test and the seven-site caller check especially. All three blocking findings and the three inline notes are fixed in 73ae4629; per-thread replies are inline. Summary:

- Before-pass `projectPath`: applied your one-word fix; both `tree_forget` cases flip to `ok: true`, line 100 comment corrected.
- `merge_warnings`: drift notice surfaced via `sanitizeWarnings`.
- Rollback coverage: restored with a mock-based test (the `returned_count` vehicle self-heals in `finalizeStagedSidecar`, so the organic path is unreachable); break-tested.
- Warning volume: collapsed to one summary line, the #1572 wording.

The non-inline point: `validate-project-refactor-spec.md` was a required #1572 edit site and is now updated. §9 Consumers corrected (the merge tools import `validateIntroduced` from `introduced-errors.js`), and a new §11 records the decision plus the options it beat (option 2 the healer, option 3 the narrow `tree_forget`). ADR-0011 keeps its residual clause.

Follow-ups filed rather than folded:
- #1702 (developer, needs-decision): the option-2 `research.json` healer.
- #1701 (developer, nothing-checks): the `materialize_facts` / `merge_tree_persons` clone contract.

On your Not-checked note: this PR changes only the deterministic write-boundary validator, with no skill or agent-prompt surface, so there is no LLM behaviour to confirm by a live session or a replayed bundle. #1476's improvise-vs-recover concern lives in the research skills, which this does not touch. Verified by the unit suites (2344 pass) and `tsc` clean. I have added that note to the PR body.

Ready for re-review.

packages/engine/mcp-server/src/tools/tree-edit.ts
clack391: Update: folded this into the PR rather than filing. Added `introduced-clone-contract.test.ts` in commit 918479dc, one test per unpinned tool. Each stubs `validateParsed` to fail on a marker the mutation itself adds (a minted person for `materialize_facts`, a repointed research ref for `merge_tree_persons`), so the real `validateIntroduced` runs and the pre-mutation clone is what keeps the diff non-empty. Break-tested one at a time: dropping `materialize_facts`' tree clone reds only its test, dropping `merge_tree_persons`' research clone reds only its test; `tree_edit` stays pinned by its own invalid-edit tests. #1701 closed as folded-in, so the review leaves one standalone follow-up (#1702).

Gennecis (approved):

Issue #1629, opened 2026-08-15
skill_latency_report/judge_report pick the wrong 'newest' run log once a skill reaches v10


aghadiayeamayanvboernest:
**IMPORTANT — superseded 2026-09-09. This issue no longer blocks anything.**
The note below said #1972's V7 scan is built on `make judge-report` and would inherit this
bug. It is not. V7 shipped as a standalone module, `eval/harness/conflict_verdict_report.py`
(PR #2344), which touches `judge_report.py` not at all — so there is no shared file and no
inherited defect. V7 also selects no "newest" log: it scans every committed log, which is the
opposite of the pattern this bug lives in. PR #1931 has merged. **Sequence this issue on its
own merits.** (Separately: all 262 committed run logs are at `v1`, so the `v10`-before-`v2`
ordering this fixes is not yet reachable in the corpus — which is a statement about urgency,
not about whether the bug is real.)

> **Iceboxed 2026-08-24 (board fill), with a trigger.** The defect is real and the fix named
> in the body is the right one — this is purely about *when*.
>
> **Trigger: the first skill whose run logs reach `v10`.** Until then the wrong-newest bug
> cannot fire. Re-checked 2026-08-24 — `ls eval/runlogs/unit/*/ | grep -oE '^v[0-9]+' | sort -u`
> returns `v1` and nothing else, across all 25 suites, so the premise the body states is
> still exactly true.
>
> **What would move it sooner:** anything that starts bumping suite versions in bulk — a
> rubric-wide sweep, a base-judge change, a mass SKILL.md edit — because the counter is
> per-skill and nothing caps it. Watch for a `v9_` appearing anywhere under
> `eval/runlogs/unit/`; that is one release away from the failure.
>
> Promote by removing the `icebox` label; `/fill-ready` then ranks it normally.

## Original issue

**Touches:** eval/harness/skill_latency_report.py, eval/harness/judge_report.py

`releasable_runlogs_for` sorts run logs with `key=lambda p: p.name` — a lexicographic sort — and both readers take `logs[-1]` as "the newest". That is wrong once a skill passes v9:

```
order: ['v10_2026-02-01…', 'v11.json', 'v2.json', 'v9_2026-01-01…']
logs[-1] (taken as newest): v9_2026-01-01_00-00-00.json   # actually the oldest
```

[...]

Issue #1404, opened 2026-08-06
Rubric dimensions that never discriminate: research-plan (4 of 4) and proof-conclusion (3 of 5) — tighten the bars or accept the plateau


mercyokum:
> **DECIDED 2026-08-23 — write adversarial fixtures first; tighten only the bars they prove
> broken.** Neither of the two options below as written. "Always 3" cannot distinguish a
> dimension that grades nothing from a bar that is simply met, because committed run logs are
> converged states — the failing intermediate runs, where a dimension does its work while
> someone iterates on a skill, never get committed. So build a scenario purpose-made so the
> dimension *should* fail: if it drives the score below 3 the dimension works and no rubric
> change is needed; if it still scores 3 the bar is broken and you know which one. One fixture
> per suspect dimension, genealogist work. Full reasoning in the ruling comment.
>
> **Fix regardless:** the dead N/A clause on "Plan mode and lifecycle" (see below) is a bug,
> not a judgement call.
>
> **Sequencing.** The `research-plan` half can start now. The `proof-conclusion` half waits on
> issue #1824 — "Tier justification" and "Proof-conclusion fit" encode what counts as proved,
> and #1824 changed the source-originality rule underneath them.

> **Two eval slots (noted 2026-08-13, corrected 2026-08-23).** The Touches line spans
> `eval/tests/unit/research-plan/rubric.md` **and** `eval/tests/unit/proof-conclusion/rubric.md`,
> so this holds both skills' snapshots and costs two paid `make eval-skill` runs plus two
> annotation passes. **Both slots are now free** — issues #1472 and #1392 have closed, so the
> "land both first" caveat that stood here no longer applies. Splitting this issue does not
> help — the adjudication is one call across both rubrics.

**Touches:** eval/tests/unit/research-plan/rubric.md, eval/tests/unit/proof-conclusion/rubric.md, eval/tests/unit/proof-conclusion/*.json, eval/fixtures/scenarios/ (new adversarial scenarios)

Found during a rubric audit of research-plan (issue #1319 follow-up work).

[...]

PR #1630, opened 2026-08-15
eval-harness: a free non-discrimination scan of the unit judge (#1485)

### LINKED ISSUE #1485: Unit-side eval judge has no calibration — add a zero-cost non-discrimination report
> **Reviewed 2026-08-10 before junior handoff.** No decision needed; three headline
> figures below were re-measured and corrected, and the module's location and run-log
> selection are now pinned.

**Touches:** eval/harness/judge_report.py (new), eval/harness/harness/versioning.py, eval/harness/harness/judge.py, Makefile, eval/harness/tests/unit/test_judge_report.py (new), docs/architecture.md, eval/CLAUDE.md

## What's wrong

The unit judge has no *deterministic* calibration instrument. Two things exist and
neither is one. `eval/harness/e2e/calibrate_judge.py` / `make e2e-calibrate`
(`Makefile:494-496`) measures the **e2e** judge — a different model on a different
cadence. And the `rubric-critic` subagent (`/audit-rubric <skill>`,
`.claude/agents/rubric-critic.md`) already defines all four checks this issue wants —
non-discriminating, flaky, unexercised, and systematic judge-vs-human divergence — but
runs **one skill at a time, by LLM judgment, only when a human remembers to invoke it**.
That is exactly how #1404 and #1446 were found: one suite at a time, by hand. This issue
builds the mechanical corpus-wide half — same four definitions, no model call, all 25
suites in one command.

**Reuse `rubric-critic`'s four flag names verbatim** so the two instruments agree, and
add one line to that agent's "What you read" section pointing at
`make judge-report SKILL=<x>` for the mechanical pass. `.claude/agents/` holds Claude
Code subagents, not Cowork plugin agents — editing it triggers no eval re-run.

Meanwhile that judge grades every unit run. `DEFAULT_JUDGE_MODEL` is
`claude-haiku-4-5-20251001` (`eval/harness/harness/judge.py:31`), pinned at
`JUDGE_TEMPERATURE = 0.0` (`:41`), and there is no `--judge-model` CLI override —
one model, one prompt, no measured accuracy. The `skill-improver` agent then reads
those scores and proposes SKILL.md edits from them, so a skill is being optimized
toward an instrument nobody has checked.

The cheapest measurement is already computable from committed data, at zero cost
and with no model calls: a **non-discrimination scan** — a rubric dimension whose
score never varies across a whole suite carries no signal, whatever it is
nominally grading. A second free signal sits next to it in the `.ann.json`
corrections: how often the judge and the human disagree, and in which direction.

Re-measured 2026-08-10 against an unchanged corpus (last run-log commit `0218c105`,
2026-08-08). **Use these numbers, not the ones this issue was filed with.**

- Newest committed run log per suite, `>= 5` graded instances and exactly one distinct
  score: **130 flagged of 171 dimension keys** (132 if a `null` counts toward the
  threshold). The total of 171 reproduces; the figure **139 does not**, under any variant
  I could construct.
- The same rule applied to **every** released + candidate log per suite instead:
  **77 of 186**. The two readings disagree on 53 dimensions. They are not
  interchangeable — step 1 picks one deliberately.
- Annotations: **10,086** correction entries across **119** committed `.ann.json`,
  **73** disagreements (0.72%) — 53 judge-harsher, 11 judge-softer, 9 involving a `null`.
  The **9,347 / 51 / 0.5%** figures do not reproduce either.

The spot examples do hold: `locality-guide` Completeness 25/25 at 3; `hypothesis-tracking`
all six dimensions flat; `search-records` Log quality / Result triage / Search strategy
20/20 at 3 and Nil escalation 19/20 with one `null`.

Read against the 130 flat dimensions, the high judge-human agreement mostly means "both
said 3," which is exactly the ambiguity a calibration report exists to resolve.

## Why it matters

This is the missing measurement instrument, and its absence is why the known judge
bugs were each found by hand, one at a time:

- **#1404** — `research-plan`: all 4 rubric dimensions score a flat 3 across 52
  instances each, and the documented N/A branch never fires. **A
  non-discrimination report would have flagged this automatically**, with no model
  call and no paid slot.
- **#1446** — `proof-conclusion`: three rubric dimensions never leave 3 across ~69
  gradings each. Found by `/audit-rubric` by hand, same class as #1404. Currently
  `icebox`.
- **#1353** is **closed and fixed** — cite it as history, not as an open defect. Its fix
  added a `FAN approach and pivot items` dimension that the newest run log grades on
  exactly **one** test. That is now an *unexercised* dimension, which is why step 2's
  per-dimension instance count matters as much as the flat flag.
- **#1401** — judge invents rubric dimensions despite an explicit "do not invent"
  instruction.
- **#1402** — `search-records`: Correctness / Log quality double-score a verdict
  that Result triage / Nil escalation already correctly failed.
- **#1403** — judge ignores an explicit `grade_on_invariant` override, twice in
  four runs.
- **#1406** — judge applies the Tool Arguments N/A rule inconsistently on identical
  zero-tool-call inputs.
- **#1000** — the run log gives no signal that dimensions are diagnostic-only on
  `grade_on_invariant` negative tests.

**This is deliberately NOT a duplicate of #1401, #1402, #1403, #1406 and #1000.**
Those five are prompt-following bugs — the judge was told one thing and did
another, and each is fixed in the judge prompt or a rubric. This issue is the
*instrument* that would surface the next one without a human noticing it first.
Fixing all five leaves the unit judge exactly as unmeasured as it is today.

## What to do

Slice 1 only. It is offline, free, and needs no model calls.

1. Add `eval/harness/judge_report.py` — at the **package root**, beside
   `skill_latency_report.py`, *not* under `harness/harness/`. That package is for
   implementation modules (`eval/CLAUDE.md` § Directory Layout); every report carrying a
   `make` target lives at the root (unit side) or in `e2e/` (e2e side). Invoke it as
   `cd eval/harness && uv run python -m judge_report`.

   **Read exactly one run log per skill: the newest released-or-candidate one** —
   `logs[-1]` after sorting by (version, timestamp), the same selection
   `skill_latency_report.releasable_runlogs_for` already makes. Use
   `versioning.py::classify` to reject `scratch_*` and `other`, and `ann_filename_for`
   (`versioning.py:109`) to pair the annotation. Do not re-derive either.

   **Do not aggregate across versions.** An older log's scores describe a SKILL.md and
   rubric that no longer exist, and that depth read is already `rubric-critic`'s job —
   it is the stated reason `DEFAULT_KEEP_CANDIDATES` is 5 and not 2 (see the comment
   above it in `versioning.py`). Concretely, the two readings disagree on 53 of 171
   dimensions, so this is a real choice and not a detail.

   **Do not window by date.** `eval/CLAUDE.md` § "Run log naming" splits run-log readers
   in two: aggregating readers *filter* to 14 days; per-skill readers *flag* stale rows
   and still show them all, because a date cut deletes the skill and hides that it needs
   a re-run. This is the second kind. Three suites (`check-warnings`, `citation`,
   `research-exhaustiveness`) sit outside a 14-day window today and are exactly the ones
   worth seeing. Mark them stale via `harness.since_window` (`describe_stale`, the way
   `skill_latency_report` does) and sort them last.
2. Emit, per skill and per dimension: graded-instance count, the distinct scores
   observed, and a **non-discriminating** flag when a dimension has >= N graded
   instances (start at 5) and one distinct score. Report N/A (`null`) separately
   from a numeric score — a dimension that is always N/A is a different defect
   from one that is always 3.
3. Emit, per skill and per dimension, the judge-vs-human numbers from the
   `.ann.json` corrections: reviewed count, agreements, disagreements, and the
   signed direction of each disagreement (judge harsher vs judge softer).
   Corrections are **sparse** — a missing entry means "not reviewed", never
   "agreed" — so print an explicit unreviewed count and never fold it into the
   denominator.
4. Add a `make judge-report` target next to `e2e-calibrate` in the Makefile,
   accepting an optional `SKILL=<name>`. Say in its `##` help text that it makes
   no API calls.
5. Add `eval/harness/tests/unit/test_judge_report.py`:
   - a synthetic suite with one flat dimension and one varying dimension flags exactly one;
   - a dimension below the instance threshold is not flagged;
   - a sparse `.ann.json` does not inflate the reviewed denominator;
   - **one test that reads a real committed run log** from `eval/runlogs/unit/` and
     asserts the extractor found a non-zero number of dimension keys. Synthetic fixtures
     cannot catch a wrong field path, and this is the trap: scores live at
     `tests[].outcome_summary.aggregated_dimensions[]` (and per-run at
     `tests[].runs[].judge.dimensions[]`) — **not** at `tests[].dimensions[]`. A report
     built against the wrong path prints zeros for every suite and every synthetic test
     still passes, because the fixture was hand-written to the same wrong shape. Assert a
     floor (`> 100` keys), never an exact count.
6. Give the report a home in the docs, or the next person re-derives it the way this
   issue re-derived `/audit-rubric`: one row in `docs/architecture.md` §9.1's command
   table, and one in `eval/CLAUDE.md`'s reader-family list.

All Python file reads must pass `encoding="utf-8"` as a keyword — including inside
`json.loads(p.read_text(encoding="utf-8"))`. This is a hard repo rule; the team
runs Windows.

## How to verify

```
make judge-report
make judge-report SKILL=research-plan
make harness-test
```

`make judge-report` must complete with no network access.

**Do not tune the detector until a count matches this issue.** Today's number is 130
flagged of 171 dimension keys under the step-2 rule, but at least nine open PRs commit
new unit run logs (PRs #1523, #1521, #1535, #1461, #1457, #1434, #1414, #1328, #1225),
and each one that lands changes which log is newest for its suite and moves the count.
Acceptance is that the rule the report **prints** is the rule it **implements** — not
that any total matches a figure written here.

`SKILL=research-plan` shows its four >= 5-instance rubric dimensions flat at 3 —
Jurisdiction accuracy, Plan mode and lifecycle, Record type selection, Sequencing logic —
plus the three base dimensions. That is #1404's finding. #1404's own arithmetic is stale:
today's newest log carries **seven** rubric dimensions at 11/11/11/11/1/1/1 instances,
not four at 52. The three single-instance dimensions must **not** be flagged at N=5, and
if they are, the threshold is wired wrong.

Before trusting the new tests, break them deliberately and watch them fail: lower the
flat detector to require two distinct scores and confirm the "flags exactly one" test
goes red; count unreviewed dimensions as agreements and confirm the sparse-annotation
test goes red; and point the extractor at `tests[].dimensions[]` and confirm the
real-corpus test goes red. A check nobody has watched fail is not trusted in this repo.

### Cost, and the gate that does not fire

**This PR owes no eval re-run and no annotation.** `check-runlogs.yml` runs on every PR
(it has no `paths:` filter), but `check_runlogs.py` derives touched skills only from
`eval/runlogs/unit/<skill>/`, `packages/engine/plugin/skills/<skill>/`,
`eval/tests/unit/<skill>/`, and `packages/engine/plugin/agents/*.md`
(`check_runlogs.py:416-443`). Nothing under `eval/harness/` or `.claude/` maps to a
skill, so rules 2 and 3 are no-ops. **This holds only while the Out of scope list below
is honoured** — the moment this PR edits a `rubric.md` or a `SKILL.md` it buys a paid
`make eval-skill` run per skill touched (roughly $8-12 and 45-65 minutes each, plus a
genealogist annotation) and `check_runlogs.py` blocks the merge until they land.

### Two open PRs to check before you start

- **PR #1410** adds a pytest lint that AST-walks every `.py` in the repo for a missing
  `encoding=`. Your module must comply whether or not it has landed.
- **PR #1437** adds `scripts/*.bat` wrappers for the Makefile targets. If it has landed,
  add `scripts/judge-report.bat`; if not, skip it.

Optional one-line fold while you are in the file: `judge.py:46`'s comment describes a
harness "invoked with `--judge-model`", a flag `run_tests.py` does not have. Deleting
that half-sentence is not the same as adding the flag the Out of scope list forbids.

## Out of scope

- Do **not** bump `DEFAULT_JUDGE_MODEL` and do **not** add a `--judge-model` CLI
  flag. That was tried and reverted; a bump to 4.6+ additionally requires
  `thinking: disabled` and a dimension-key strip. Separate decision, separate PR.
- Do **not** edit any `rubric.md`, any `SKILL.md`, or the judge prompt
  (`eval/harness/judge/prompt.md`). This PR ships a *report*; acting on what it
  finds is the follow-on, and rubric prose edits are lane 2 work
  (`docs/skill-lifecycle.md` §5).
- Do **not** touch `eval/harness/e2e/calibrate_judge.py` or `make e2e-calibrate` —
  the e2e judge is a separate model on a separate cadence.
- Do **not** run any paid eval suite for this. If a change here would require one,
  it is out of slice 1.
- No slice 2 (re-grading against a golden set, or any model-calling accuracy
  measurement) in this PR. Propose it once slice 1 has produced numbers.

aghadiayeamayanvboernest:
## Summary

- The unit judge grades every unit run and **nothing measures it**. `make e2e-calibrate` measures a *different* judge on a different model; `/audit-rubric` asks the right questions but one skill at a time, by LLM judgment, only when someone remembers — which is how every judge defect so far was found. Meanwhile `skill-improver` reads these scores and proposes SKILL.md edits from them.
- `make judge-report` is the mechanical half: same definitions, **no model call**, every suite in one command, from committed data only.
- **First run: 119 of 169 dimension keys never vary across their suite** (measured 2026-08-18; this number moves as run logs land), including all five of `research-plan`'s rubric dimensions flat at 3 — #1404's finding, reproduced by a free command instead of by hand.
- Tier: **Normal**. Slice 1 only. Closes #1485.

## Review guidance

**Start here:** the three definitions in `judge_report.py` that change answers rather than wording.
1. **`graded` counts numeric scores only** — an always-N/A dimension is a separate defect and never flags as flat. `init-project`'s `Place standardization` — all its numeric scores identical, most of its gradings N/A — flips from unflagged to flagged if nulls count.
2. **A dimension is keyed `(skill, source, name)`** — `(source, name)` alone merges seven names across suites (all three base dimensions plus `Jurisdiction accuracy`, `Result triage`, `Actionability`, `Accuracy`) and collapses the corpus from **169 keys to 93**.
3. **`Non-discriminating` requires `malformed == 0`** — a float `2.0` dropped from the count leaves five identical `3`s and *manufactures* the flatness this report exists to detect.

**Unsure about:** nothing structural. One judgment call: `Flaky / high-variance dimension` is one of `rubric-critic`'s four flags but is unreachable for a single-log offline reader (`runs_per_test` is pinned to 1). I name it in the footer as not-computed rather than silently omitting it.

## Plan

**Didn't change:** no `rubric.md`, no `SKILL.md`, no judge prompt (acting on findings is the follow-on, and any of those buys a paid `make eval-skill` per skill and blocks the merge); no `DEFAULT_JUDGE_MODEL` bump and no `--judge-model` flag; no `e2e/calibrate_judge.py`; no slice 2.

**Acceptance check:** `test_reads_dimension_keys_from_a_real_committed_run_log` — fails on `main` (module absent) and fails against the wrong field path even once it exists. Behaviourally: `SKILL=research-plan` shows five rubric dimensions at 14 instances flat at 3; `SKILL=init-project` does **not** flag `Place standardization` (too few numeric gradings to judge); `SKILL=search-familysearch-wiki` does **not** flag `No-result handling` (1 of 13).

**Deviated from the plan:** `/critique-plan` returned **2 blocking findings, both real**, verified against the corpus before folding in. My keying was `(source, name)` — the merge above — and my own `>100 keys` floor was *unreachable* under it, so the plan contradicted itself. And my acceptance check was **already false**: it inherited the issue's stale numbers ("four dimensions at 11 instances plus three singles"); the newest log has five at 12 with **zero** single-instance dimensions, so it couldn't exercise the threshold at all. Moved that check to two suites where it's real on live data. Also folded in that the signed disagreement direction would have **crashed the report** on a null score (`None > 3` raises; 176 null entries exist corpus-wide, zero in today's slice — it would have shipped untested and fired later).

**Then a code review found five more, all fixed:** the `.ann.json` join had **no test at all** — breaking the sibling lookup made all 25 suites report `0 reviewed` with a clean-looking footer while every test stayed green, the same silent-zeros trap the real-corpus test was written for, guarded on one path and not the other. Per-dimension annotation counts printed only on disagreement, so *reviewed-and-agreed* and *nobody-looked* rendered byte-identically and `agreements` was never printed at all. `Systematic judge-vs-human divergence` fired on a single correction, against `rubric-critic`'s own "recurring across multiple tests in the same direction" — making 100% of that flag's output a false positive. Plus `SINCE=` dropping suites with no window line, and a skill with only scratch logs vanishing while the footer claimed "across N suites."

## Test plan

- [x] Ran `make test-all`. Eval harness pytest **2083 passed / 2 skipped**, eval-app, control-plane, engine green, ruff clean. The one red is the known **#1366** `apps/electron` flake — **62/62 in isolation**, and this branch changes **zero** files under `apps/electron`.
- [x] **Every guard proven to fail first**, per CLAUDE.md — the three prescribed breaks (a two-distinct-score detector; an unreviewed-counts-as-agreed join; an extractor pointed at `tests[].dimensions[]`, which does not exist), plus the two added after review (breaking the `.ann.json` pairing → `only 0 suites paired`; hiding a skill's releasable logs → the `NOT MEASURED` line fires).
- [x] **Offline demonstrated, not asserted:** the report completes with `socket.connect`/`create_connection` monkeypatched to raise.
- [x] N/A — no skill run-log snapshot changed. Nothing here maps to a skill (`check_runlogs.py` reads only `eval/runlogs/unit/`, `plugin/skills/`, `eval/tests/unit/`, `plugin/agents/*.md`), so rules 2/3 are no-ops. `.claude/agents/` holds Claude Code subagents, not Cowork plugin agents. **No paid eval run.**

**On the headline number:** the issue expected 130 flagged of 171; I get **119 of 169** (it read 123 earlier in review — a `historical-context` run log landed on 2026-08-17 and moved it). That is the drift the issue itself predicted — PRs have landed and changed which log is newest per suite. Per its instruction I did **not** tune the detector to hit a total: the rule printed is the rule implemented.

## Follow-on issues

- **#1629** — `releasable_runlogs_for` sorts run logs lexicographically, so `v9` sorts *after* `v10` and "newest" silently becomes the oldest at two-digit versions. Latent (every suite is at `v1`) and inherited from `skill_latency_report`, which this issue told me to reuse rather than re-derive — so filed rather than folded in. Labelled `nothing-checks`.
- Acting on what the report finds — #1404 (`research-plan`) and #1446 (`proof-conclusion`) already exist.
- Slice 2 (golden-set re-grading) — propose once these numbers have been read.
- `Unexercised dimension` in `rubric-critic`'s full sense (declared in `rubric.md`, graded on no test) would need `harness.rubric.parse_rubric` as an extra input; **zero** suites have one today, so it isn't worth building blind.


chesworthrm (commented):
Reviewed the full diff against the merge with `main` (branch is level with it, so the merge and the head are the same tree). All required checks are green on `eb3d52a`. One thing to fix before this goes in, plus a test to fold in with it.

Nice piece of work overall — the three definitions you called out in the review guidance are the right three, and each one is genuinely load-bearing. I checked all of them.

## The `NOT MEASURED` line reports the wrong cause

`judge_report.py:384` collapses two different reasons a skill can drop out of the report into one empty list, and the message that prints only describes one of them.

```
$ make judge-report SINCE=2
Window: runs on/after 2026-08-13 — 6 of 25 run(s), 19 older run(s) excluded.

NOT MEASURED: 19 skill(s) have no releasable run log (scratch runs only) — ...
  check-warnings, citation, ..., research-plan, ...
```

`research-plan` has six releasable run logs in `eval/runlogs/unit/research-plan/`. It dropped out because of the window, not because it only has scratch runs. Those 19 skills are also already accounted for one line above as "19 older run(s) excluded", so they get reported twice with explanations that contradict each other.

The part that makes this worth blocking on: on the default run the line never fires at all, because every skill currently has a releasable run log. I confirmed that. So the only output this line ever actually produces is a false statement about the corpus — the same shape as the `systematic_divergence` false positive you already caught and fixed.

You wrote:

```python
    unmeasured: list[str] = []
    for skill in skills:
        logs = releasable_runlogs_for(skill, cutoff=args.since)
        if not logs:
            unmeasured.append(skill)
            continue
```

Change it to:

```python
    unmeasured: list[str] = []
    windowed_out: list[str] = []
    for skill in skills:
        if not releasable_runlogs_for(skill):
            unmeasured.append(skill)
            continue
        logs = releasable_runlogs_for(skill, cutoff=args.since)
        if not logs:
            windowed_out.append(skill)
            continue
```

And add a second block directly after the existing `if unmeasured:` print:

```python
    if windowed_out:
        print(
            f"OUTSIDE THE WINDOW: {len(windowed_out)} skill(s) have a releasable run "
            f"log, but none inside --since — they are absent from every count below: "
            f"{', '.join(sorted(windowed_out))}"
        )
        print()
```

## Nothing tests the printing path

`main`, `format_skill` and `format_footer` have no test coverage at all, and the bug above is sitting in exactly that gap. It is the same silent-wrong-output failure your two real-corpus tests were written to catch on the other paths — guarded on one side, not the other.

Please add one test with the fix: call `main` with a narrow `--since` and assert that a skill which does have a releasable run log does not appear on the `NOT MEASURED` line. Worth doing in this PR rather than filing it, since the branch is open and you have the context loaded.

## What I checked, so you don't redo it

- Harness pytest: 2083 passed / 2 skipped, matching your number. My first run had one failure in `test_mock_mcp.py`, which is just the fresh-worktree case — it cleared after building the engine.
- `make harness-lint` clean, encoding lint passes.
- I broke three of your five guards to watch them fail, and all three are real: pointing the extractor at `tests[].dimensions[]` fails 8; breaking the `.ann.json` sibling lookup fails the pairing test alone; dropping the `malformed == 0` clause fails the malformed test alone.
- All three behavioural acceptance checks reproduce — `research-plan`'s five rubric dimensions flat at 3 over 12, and both negatives (`Place standardization`, `No-result handling`) correctly landing on `below threshold` rather than flagged.

## Not asking you to change these

The headline numbers have already drifted since you wrote the body — I get 125 of 169, and `init-project`'s `Place standardization` is now 2 numeric of 9 rather than of 8. Your body predicts exactly this and says the detector was not tuned to a total, so nothing to do; just flagging that the summary will read slightly stale once it lands.

I looked at and am deliberately not flagging: the tab-`#` comments in the Makefile recipe (`server-mock` and several others do the same, so it is the convention here); `systematic_divergence` counting corrections rather than distinct tests, which is fine because the ann schema is one entry per dimension per test and the two are therefore equivalent; and the unknown-`--skill` path, which already errors cleanly with exit 1.

Gaps in this review: I did not run vitest, `turbo test typecheck` or the control-plane pytest locally — the diff touches no TS or web file and CI ran all three green on this head. I verified the offline claim by reading the module's imports rather than re-running your socket monkeypatch, mutation-tested three of the five guards rather than all five, and did not execute the Windows `.bat`.

clack391:
Senior review. I ran the full diff offline; the break-tests all bite and the research-plan flat-dimension finding reproduces. Strong work overall.

One blocker, and it is the one @chesworthrm already flagged: the `NOT MEASURED` line mislabels window-dropped skills as "scratch runs only." I confirmed it is still live at the current head (ce0c071e). `make judge-report SINCE=7` lists `research-plan`, `locality-guide`, and `citation` as "scratch runs only," but each has released `v{N}` logs; they were dropped by the window, not for lack of a run log, and the line contradicts the "N older run(s) excluded" line printed just above it. On the default run the line never fires, so its only ever output is that false statement. @chesworthrm's comment has the fix (treat a skill as unmeasured only when the un-windowed lookup is empty, plus a separate "outside the window" line) and asks for a printing-path test worth folding into the same change.

Minor and optional: the always-null case is labelled "always-N/A" rather than rubric-critic's "Unexercised dimension" flag. A relabel would let the two instruments agree, as the issue intended. Not a blocker.

Everything else checks out: correct field path, the >=5 / one-distinct rule with N/A reported separately, sparse-safe annotations, and Out of scope honored (judge.py is comment-only).

clack391 (approved):
Approving. You cleared both review rounds cleanly, and the fix went past a patch: the unmeasured / windowed-out / unreadable split now names each distinct reason a skill is absent, which is exactly the discipline this tool is about. The printing-path test earns its keep by deriving the cutoff from the newest run date, so the window genuinely splits the corpus instead of emptying it.

I re-verified against the current head: chesworthrm's SINCE repro now prints OUTSIDE THE WINDOW rather than the false scratch-runs line, the 123-of-169 headline reproduces, the break-tests bite, and research-plan's flat rubric dimensions are right there. The three definitions you called out in the review guidance (numeric-only graded, per-skill keying, malformed blocking the flat verdict) are the correct three, and the unreadable-runlog guard is a good instinct: an analysis tool that dies on one corrupt file is one nobody can use.

Strong work. This is the mechanical half of judge calibration the repo has been missing.

chesworthrm (commented):
Reviewed the full diff against the test-merge with `main` (the branch is 25 commits behind, and the merge is clean). All required checks are green on `b4708bd9`.

This is good work, and the part I want to call out is that the two blockers @chesworthrm and @clack391 raised are properly fixed, not papered over. I restored the pre-fix `NOT MEASURED` attribution and watched `test_a_windowed_out_skill_is_not_reported_as_having_no_run_log` fail on it. That test earns its place.

One thing to fix before this merges, and three smaller notes.

## The unreadable-run-log guard doesn't catch the encoding error

`judge_report.py:410`. The comment right under this line promises that one bad run log must not take the other 24 suites with it. It doesn't hold. `UnicodeDecodeError` is a `ValueError`, not a `JSONDecodeError`, so nothing in the tuple catches it — I put invalid UTF-8 bytes into the newest `research-plan` log and the whole report died with a traceback out of `_load`. No suite reported at all.

This is reachable, not theoretical. 19 files in the unit corpus contain multibyte UTF-8, and two of them are files this report reads today. A write that gets interrupted mid-character gives you this error rather than a `JSONDecodeError`.

You wrote:

```python
        except (OSError, json.JSONDecodeError, TypeError, AttributeError) as e:
```

Change it to:

```python
        except (
            OSError,
            UnicodeDecodeError,
            json.JSONDecodeError,
            TypeError,
            AttributeError,
        ) as e:
```

I applied that and re-ran against the same corrupted file. It now does exactly what your comment describes: `skip research-plan/…: 'utf-8' codec can't decode byte 0xff`, then the `UNREADABLE: 1 run log(s)` line, then `117 of 161 dimension keys across 24 suite(s)`.

`e2e/corpus_report.py:246` — the precedent you cite two lines below — has the same hole. Whether you fold the one-line change in there too is your call; it's outside this diff.

## The threshold comment states a corpus number that isn't true

`judge_report.py:165` justifies `MIN_DISAGREEMENTS_FOR_FLAG` on the corpus holding "exactly two disagreements, both on one test." I measured the slice the report actually reads and it holds **zero** — at your own head (2048 corrections) and after merging `main` (2095). There are 43 disagreements across the whole unit corpus, but none of them are in the newest log per skill.

Keep the threshold. It matches `rubric-critic`'s definition and the code is right. It's the number that's wrong, and the consequence is worth saying out loud in a report whose whole point is not overclaiming: the judge-vs-human direction half — `judge_harsher`, `judge_softer`, `FLAG_DIVERGENCE` — currently produces nothing on live data.

You wrote:

```
        output a false positive, since the corpus holds exactly two disagreements,
        both on one test.
```

Change it to:

```
        output a false positive. The newest log per skill currently holds no
        disagreements at all, so this flag reports nothing on live data today.
```

And one stale number in the tests, `test_judge_report.py:166`:

```
You wrote:      suites paired and 2263 reviewed entries.
Change it to:   suites paired and 2095 reviewed entries.
```

## @clack391's relabel point still stands, and it's a bit stronger than "optional"

Line 81 sets the rule: flag names are verbatim from `rubric-critic.md` so the mechanical pass and the LLM pass name the same defect the same way. Two of the three follow it. The third prints `always-N/A`, where that agent's flag 3 is **`Unexercised dimension`** — defined there as a rubric dimension that is `null`/absent on every test, which is exactly what `always_na` computes.

Since this PR also tells the agent to start from `make judge-report`, a reader comparing the two outputs never sees the agent's own name for it. Still your call, but it's one constant.

```
You wrote:      verdict = f"always-N/A ({d.na})"
Change it to:   verdict = f"{FLAG_UNEXERCISED}: always N/A ({d.na})"

You wrote:      f"{len(always_na)} are always-N/A — a different defect, never counted above."
Change it to:   f"{len(always_na)} meet {FLAG_UNEXERCISED.lower()} (always N/A) — a "
                f"different defect, never counted above."
```

with, beside the other two constants:

```python
#: rubric-critic's flag 3. This computes its first half only — the "no test
#: could ever fail on it" half needs the rubric as an input.
FLAG_UNEXERCISED = "Unexercised dimension"
```

## One discovery surface missed

`.claude/skills/find-big-wins/SKILL.md:316`. That table is headed "Every one of these reads committed run logs and costs nothing" and lists `skill-latency`, `eval-timings`, `e2e-corpus` and three more. `make judge-report` is exactly that, and "123 of 169 dimensions grade nothing" is exactly what that skill exists to surface. You wired the other four surfaces, so this reads as an oversight rather than a decision.

Add after the `make eval-timings` row:

```
| `make judge-report` | rubric dimensions whose score never varies across a suite — a flat dimension grades nothing |
```

## What I checked, so you don't redo it

All against the test-merge with `main`, not your head:

- Eval harness pytest **2243 passed / 2 skipped**; engine vitest **2368 passed** (this is what covers `windows-wrappers.test.ts` reading your new `.bat`); control-plane pytest **162 passed / 1 skipped**; ruff clean.
- My first harness run failed `test_mock_mcp.py::test_successful_call_carries_no_is_error`. That's just the fresh-worktree missing engine build — it cleared after `npm run build`. Not a regression.
- Offline demonstrated rather than read: the report exits 0 with `socket.connect` and `create_connection` patched to raise.
- Broke three guards to watch them fail. Wrong field path fails 8 tests; breaking the `.ann.json` sibling lookup fails the pairing test alone; restoring the old `NOT MEASURED` attribution fails the new printing-path test.
- All three of your behavioural acceptance checks reproduce. `research-plan` shows five rubric dimensions flat at 3 (now n=14, not 12), and both negatives land on `below threshold` rather than flagged.
- Headline reproduces at **123 of 169** across 25 suites.
- `--since 5`, `--since nonsense` and `--skill nosuchskill` all behave — the window line, the `OUTSIDE THE WINDOW` block, a clean usage error, and exit 1 with an accurate message.

## Not asking you to change these

The tab-`#` comments in the Makefile recipe — `server-mock` and others do the same, so that's the convention here. The window line counts skills but reads "run(s)"; that's accurate only because this reader takes exactly one log per skill, which is fine. And `apply_annotation` silently dropping a correction that matches no dimension is latent, not live — I measured 0 orphans in the read slice today.

## Gaps in this review

I didn't run `turbo test typecheck` or the engine `tsc` locally; the diff touches no TS and CI ran both green on this head. I didn't execute the `.bat` — no Windows machine — so I read it against `harness-test.bat` instead, and note that `windows-wrappers.test.ts` explicitly cannot check whether a wrapper's commands still match its recipe. Omitting the `engine-build.bat` prerequisite looks right to me since this reader only parses JSON. I mutation-tested three of your five guards, not all five.

Two approvals are still needed, and one has to come from @PioneerAIAcademy/senior-developers for the `.py`/`Makefile`/`.bat` paths. Worth noting `eval/harness/harness/judge.py` does not match the `/eval/harness/judge/` genealogist rule, so this doesn't pull in senior-genealogists.

aghadiayeamayanvboernest:
Thanks both — all five items are in as of `1c910fda2`. Harness suite 2260 passed / 2 skipped, ruff clean.

## The blocker: `UnicodeDecodeError`

Fixed, and you were right that it was reachable rather than theoretical. `_load` calls `read_text(encoding="utf-8")`, so the decode happens before `json` ever sees the bytes — and `UnicodeDecodeError` is a `ValueError`, so nothing in the tuple caught it.

I added a test for it rather than just the one-line change, and **built the corrupt corpus in a tmpdir**. I didn't want to corrupt a committed run log even temporarily: those are evidence, and editing one to exercise a reader changes the corpus under everyone else mid-review. The fixture is valid JSON up to the point the bytes stop being UTF-8, which is the shape an interrupted write actually leaves.

Proven to fail first, per CLAUDE.md: removing `UnicodeDecodeError` from the tuple again reds it with the exact traceback out of `_load` you described, and the other suite is dropped with it.

I also folded the same one-line fix into `e2e/corpus_report.py`. You left it to me, but my guard cites that function as its precedent — leaving it broken while pointing at it as the model to follow made no sense.

## The threshold comment — I measured before changing it, and both numbers are stale

This one didn't resolve the way either of us expected, so the detail matters.

- My comment said the corpus holds **exactly two disagreements, both on one test**.
- You measured **zero** in the read slice, at my head and after merging main.
- I now measure **two** — `ut_historical_context_005`, `base/Correctness` and `rubric/Genealogical implications`, one each.

The difference is a `historical-context` run log dated 2026-08-17 that landed after you measured. So the number has been wrong in both directions inside a day, and any count written there goes stale the next time a suite is re-run.

So I removed it rather than replacing one snapshot with another, and stated what actually holds without a snapshot: **no dimension has yet reached two same-direction corrections, so the divergence half reports nothing on live data today.** That is your point and it survives the drift — `systematic_divergence` is still 0 across all 25 suites. I kept your framing that this is worth saying out loud in a report whose whole point is not overclaiming.

**Related, and worth knowing before you re-verify: the headline is now `119 of 169`, not the `123 of 169` you both reproduced.** Same cause — the corpus moved, not these edits.

## The other three

- **`FLAG_UNEXERCISED`** — adopted, with your comment about it computing only the first half. Agreed it was stronger than optional: two of three flag names coming from `rubric-critic` verbatim and the third not is the kind of inconsistency that only shows up when someone diffs the two outputs, which this PR now tells the agent to do.
- **`make judge-report` row** in `find-big-wins` — added after `make eval-timings`. That was an oversight, not a decision; I wired the other four and missed it.
- **Test snapshot count** — re-measured to `2074` and dated, with a note that it is a snapshot and the assertions are floors precisely so a landing run log can't red the suite.

## Not changed

Your three no-change notes I've left alone. On `apply_annotation` silently dropping an unmatched correction — agreed it's latent rather than live, and I re-confirmed 0 orphans in today's read slice.

Thanks for the mutation testing on three of the five guards, and for restoring the pre-fix `NOT MEASURED` attribution to confirm that test earns its place — that's the check I'd have wanted done on my own work.

chesworthrm (commented):
Reviewed the full diff at `8e3ca997`. The branch is level with `main`, so the merge and the head are the same tree. All 12 required checks are green on this head.

This is solid work, and the round-2 fixes are genuinely fixed rather than patched over. I restored the pre-fix `NOT MEASURED` attribution and watched the printing-path test fail on it, and I removed `UnicodeDecodeError` from your guard and watched the undecodable-log test fail on that. Both tests earn their place.

One thing to fold in before this merges, plus a process note that isn't yours to action.

## The other half of the `UnicodeDecodeError` fix has no test

`eval/harness/e2e/corpus_report.py:246`. You folded the same one-line fix in there, which was the right call. But only one of the two halves is guarded.

I reverted that tuple to its pre-PR form and ran the whole harness suite: **2260 passed, 2 skipped**. Nothing catches it. Do the same thing to `judge_report.py` and your own test reds immediately.

That matters because of what the comment you added at that line says:

```
Same hole found in `judge_report.py`, which cites this function as its
precedent (#1485 review); fixed in both so the precedent is one worth citing.
```

The precedent is now the unguarded half. Anyone tidying that tuple later removes the fix and every suite stays green — which is the exact silent-zeros trap your real-corpus tests were written for.

The test next door only covers invalid JSON, which `json.JSONDecodeError` already caught before this PR.

Add this to `eval/harness/tests/unit/test_e2e_corpus_report.py`, directly after `test_tally_reports_unreadable_files_instead_of_crashing`:

```python
def test_tally_survives_a_log_whose_bytes_are_not_utf_8(tmp_path: Path):
    good = _write(tmp_path, "run-1.json", {"verdict": "pass"})
    bad = tmp_path / "run-2.json"
    # Valid JSON up to the point the bytes stop being UTF-8 — the shape a write
    # interrupted mid-character leaves. Raises UnicodeDecodeError (a ValueError),
    # not json.JSONDecodeError.
    bad.write_bytes(b'{"verdict": "pass", "note": "\xff\xfe"}')

    recall, _compliance, _gate, problems, _arms, _fix, _bash = tally([good, bad])
    assert recall == {"pass": 1}
    assert len(problems) == 1
    assert "run-2.json" in problems[0]
```

I ran it both ways. It passes with your fix and fails with `UnicodeDecodeError` without it.

## The one approval on this PR predates every round-2 fix

Not an action for you — flagging it so nobody counts it as read.

@clack391 approved at `9ebe5715`. The head is `8e3ca997`, and `1c910fda` landed in between. I checked the approved tree directly rather than going by timestamps:

| Round-2 change | Present at `9ebe5715`? |
|---|---|
| `UnicodeDecodeError` in `judge_report.py` | no |
| `FLAG_UNEXERCISED` | no |
| `judge-report` row in `find-big-wins` | no |
| `corpus_report.py` except tuple | still the old 4-entry form |

`git merge-base --is-ancestor` confirms `1c910fda` is not an ancestor of `9ebe5715`. That approval body also reproduces `123 of 169`; the head now reports `119 of 169`.

`protect-main` has `dismiss_stale_reviews_on_push: false`, so the approval still counts toward the two required and nothing re-requested review. @clack391 — worth a second look at `1c910fda` rather than letting the next approval carry it in.

## Two stale numbers in the PR body

Both are the drift your body predicts, and both earlier reviewers declined to ask for a change. I'd still nudge them, since the summary is what lands in the merge commit:

- Summary says **123 of 169**; it is now **119 of 169**. You already noted this in a comment.
- **Acceptance check** says `research-plan` shows five rubric dimensions "at 12 instances"; it is now **14**. The five-flat-at-3 finding itself reproduces exactly.

## What I checked, so you don't redo it

- Eval harness pytest: **2260 passed / 2 skipped**. My first run failed `test_mock_mcp.py` — the fresh-worktree missing engine build. Cleared after `npm run build`. Not a regression.
- `ruff` clean on all three changed Python files. `windows-wrappers.test.ts`: 15 passed, and it really does cover your new `.bat` — I deleted the `judge-report` row from `scripts/windows/README.md` and 2 of its 15 tests failed.
- **All six guards broken and watched fail**, not three. Wrong field path reds 8 tests; the `.ann.json` sibling lookup, the `malformed == 0` clause, `MIN_DISAGREEMENTS_FOR_FLAG`, the old `NOT MEASURED` attribution and `UnicodeDecodeError` each red exactly one.
- Offline demonstrated, not read: `main([])` returns 0 and prints 398 lines with `socket.connect` and `create_connection` patched to raise.
- All three behavioural acceptance checks reproduce. `research-plan`'s five flat rubric dimensions, and both negatives (`Place standardization` n=3+6 N/A, `No-result handling` n=1+12 N/A) landing on `below threshold` rather than flagged.
- Your annotation join is lossless today: 2074 correction entries in the ann files, 2074 joined, 0 orphans — matching the dated snapshot in your test exactly, 25 of 25 suites paired.
- `--since 20`, `--since 2026-12-01` and `--skill no-such-skill` all behave, with the right message and exit code for each.

## Not asking you to change these

The tab-`#` Makefile recipe comments — 248 of them already in that file, so it's the convention. `describe_window` counting skills while printing "run(s)", which is accurate here because this reader takes exactly one log per skill. And `apply_annotation` silently skipping a correction that matches no dimension — I measured 0 orphans in today's read slice, so it's latent.

## Gaps in this review

I didn't run `turbo test typecheck` or the control-plane pytest locally; the diff touches no TS and no `apps/server` file, and CI ran both green on this head. I didn't execute the `.bat` — no Windows machine — so I read it against `harness-test.bat`. I verified your docstrings' *behavioural* claims but did not try to falsify the reasoning behind each definition beyond the three you flagged in the review guidance, all of which hold.

On CODEOWNERS: the `.py`, `.bat` and `Makefile` paths pull a senior from either team. The five `.md` paths have no codeowner at all, which is the deliberate no-`*`-default. `eval/harness/harness/judge.py` does not match the `/eval/harness/judge/` rule — that rule targets the real directory of the same name — so this doesn't specifically pull in senior-genealogists. Two approvals still needed.

johnmarkpeterbrown (changes requested):
Hey @aghadiayeamayanvboernest — reviewed #1630. All green, both earlier blockers properly fixed. Approving once one thing goes in.

**One thing to fix**

The `UnicodeDecodeError` fix you added to `e2e/corpus_report.py` has no test. Your `judge_report` one does. I took the fix back out and every test in that file still passed — so if someone undoes it later, CI stays green.

Add a test next to the existing one:

```python
def test_tally_survives_a_run_log_that_is_not_valid_utf8(tmp_path: Path):
    good = _write(tmp_path, "run-1.json", {"verdict": "pass"})
    bad = tmp_path / "run-2.json"
    bad.write_bytes(b'{"verdict": "pass", "note": "\xff\xfe"}')

    recall, _compliance, _gate, problems, _arms, _fix, _bash = tally([good, bad])
    assert recall == {"pass": 1}
    assert len(problems) == 1
    assert "run-2.json" in problems[0]
```

Then pull the fix out and watch it go red before you commit.

**One small one**

You removed the drifting count from one docstring but left one in the module header. It's stale again — 3 numeric scores now, not 2.

```
You wrote:      `Place standardization` (2 numeric, 6 N/A) flips from unflagged to flagged
Change it to:   `Place standardization` — all its numeric scores identical, most of its
                gradings N/A — flips from unflagged to flagged
```

**Two heads-ups, nothing to change**

@clack391's approval is on an older commit, before the `UnicodeDecodeError` work. It still counts, so nobody gets pinged. Worth asking him for a quick re-look.

@chesworthrm's note about needing a senior-developers approval is out of date — a senior from either team unblocks you now.

Nice work. This is the half of judge calibration that was missing.

aghadiayeamayanvboernest:
Fixed in `faa516bb1`. Harness suite 2261 passed / 2 skipped, ruff clean.

## The unguarded half

You were right, and the reason it mattered is the part I'd missed. I folded the fix into `corpus_report.py` because leaving the cited precedent broken made no sense — but leaving it *untested* recreates the same problem one step later: anyone tidying that tuple removes the fix and every suite stays green, and `judge_report.py`'s comment still points at it as the thing to copy.

Your test is in, with your reasoning in the docstring. Proven to fail first: reverting the tuple to its pre-PR form reds it with the `UnicodeDecodeError` traceback, and your point that the neighbouring test only covers invalid JSON — which `json.JSONDecodeError` already caught — is exactly why it wasn't enough.

## The two stale numbers

Both updated in the body, and I verified each directly rather than taking them from the review, since these are the numbers that drift:

- Headline is **119 of 169**, not 123. I've added the measurement date and a note that it moves, so the next reader doesn't treat it as fixed.
- `research-plan`'s five flat rubric dimensions are at **n=14**, not 12 — confirmed with `--skill research-plan`, all five still flat at 3.

I also re-confirmed both negatives still land on *below threshold* rather than flagged.

## On the stale approval

Thanks for checking the approved tree rather than the timestamps — that's a real gap and I wouldn't have thought to look.

@clack391, to make your re-read cheap: your approval at `9ebe5715` predates `1c910fda` and `faa516bb1`. What landed after it is the `UnicodeDecodeError` guard and its tmpdir test, `FLAG_UNEXERCISED`, the `make judge-report` row in `find-big-wins`, the `corpus_report.py` tuple and now its test, and the threshold-comment rewrite. That last one is the only place I did **not** take a reviewer's suggested text: I measured and found the replacement number was already stale too, so I removed the count rather than swap one snapshot for another, and kept the conclusion — which still holds at 0 dimensions meeting the threshold.

Since `dismiss_stale_reviews_on_push` is false, nothing re-requested you automatically.

chesworthrm (commented):
Re-reviewed at `faa516bb`. The delta since my last pass is one commit and one file — 27 lines added to `test_e2e_corpus_report.py`. No drift, nothing unrequested.

## The blocker is properly closed

`test_tally_survives_a_log_whose_bytes_are_not_utf_8` is exactly the guard that was missing, and I proved it rather than reading it. Same mutation as before — pull `UnicodeDecodeError` back out of the tuple:

| | before `faa516bb` | now |
|---|---|---|
| Full harness suite | 2260 passed, 2 skipped | **2261 passed, 2 skipped** |
| Suite with the fix reverted | 2260 passed — **fully green** | **1 failed**, 2260 passed |

That's the whole point of the finding: the fix can no longer be silently un-done. `ruff` clean, encoding lint passes.

Nice touch putting the "an unguarded precedent is one a later tidy-up silently un-fixes" reasoning in the docstring — that's the part a future reader needs.

## One thing left, and it's @johnmarkpeterbrown's small one

Their blocker went in; this didn't. `judge_report.py:24`:

```
You wrote:      `init-project`'s `Place standardization` (2 numeric, 6 N/A) flips from
                unflagged to flagged if N/A counts toward the threshold.
Change it to:   `init-project`'s `Place standardization` — all its numeric scores
                identical, most of its gradings N/A — flips from unflagged to flagged
                if N/A counts toward the threshold.
```

I measured it just now: **n=3, +6 N/A**. So "2 numeric" is wrong. It moved 2→3 inside a day, which is why the fix is to delete the number rather than update it — the same call you already made in round 2 when you dropped the drifting disagreement count instead of re-snapshotting it.

**The same stale figure is in the PR body twice**, which their review didn't catch:

- **Review guidance** #1 — "(2 numeric, 6 N/A)"
- **Acceptance check** — "(2 numeric of 8)"; it is now 3 of 9

Worth fixing in the same pass. The body is what lands in the merge commit, and you already updated the other two numbers there.

## Two neighbouring claims in that docstring that do hold

I checked these because they sit in the same sentence and I didn't want you re-verifying them:

- `base/Tool Arguments` is genuinely 100% N/A in exactly three suites — `convert-dates`, `project-status`, `translation`.
- The `~169 keys → ~93` collapse is exact at 169 and 93. That also puts the real-corpus test's `> 100` floor correctly between the two, so it discriminates as intended.

## What I re-ran, and what I didn't

Everything outside the one changed test file is byte-identical to `8e3ca997`, which I verified in full last round — so I did **not** re-run the six `judge_report.py` guard mutations, the offline socket check, or the three behavioural acceptance checks. Nothing they cover moved.

I did re-run: the full harness suite, both directions of the `corpus_report` mutation, `ruff`, the encoding lint, and the live measurements above. I didn't run `turbo test typecheck` or the control-plane pytest — no TS or server file is touched and CI ran them green on this head.

## Merge state

@johnmarkpeterbrown's **CHANGES_REQUESTED is still standing** at `8e3ca997`. A push doesn't clear it, so it needs their re-review regardless of the docstring line.

@clack391 — your approval is still at `9ebe5715`, now four commits back, and predates all of the `UnicodeDecodeError` work including the guard added today. `dismiss_stale_reviews_on_push` is off here, so it still counts toward the two required and nothing re-requested you. A quick re-look would be worth it.

Small correction on @johnmarkpeterbrown's last heads-up, since it was addressed to me: the "must come from senior-developers" note was in the round-2 review at `b4708bd9`, not my round-3 one — mine already said a senior from either team unblocks. The correction is right, just aimed at the earlier comment.

One docstring line and two PR-body lines from done, and the fix is a deletion. Happy to approve once that's in.

johnmarkpeterbrown (commented):
Re-checked #1630 on `faa516bb`. The `corpus_report` test is in and it's solid — I reverted the tuple and it went red on its own, exactly right. Thanks for building it the same way as the other one.

One leftover from my last message, then I'll flip to approve.

**The docstring number, still there** — `judge_report.py` around line 23. It's 3 numeric now, not 2:

```
You wrote:      is 100% N/A in three suites, and `init-project`'s `Place standardization` (2
                numeric, 6 N/A) flips from unflagged to flagged if N/A counts toward the threshold.

Change it to:   is 100% N/A in three suites, and `init-project`'s `Place standardization` — all
                its numeric scores identical, most of its gradings N/A — flips from unflagged to
                flagged if N/A counts toward the threshold.
```

Your commit message says two stale numbers got refreshed, but both were in the PR body — this one's in the code.

**And one clause in the body** — the Acceptance check line still says `Place standardization` is "2 numeric of 8". It's 3 of 9. Same fix as the headline: drop the count.

The rest of the body edits are good. `119 of 169` with the dated caveat is the right way to write it.

Ping me when those two are in.

clack391 (approved):
Re-approving at faa516bb to refresh my stale approval from 9ebe5715, since dismiss_stale_reviews_on_push is off and it predates the round-2/3 fixes.

I re-ran the whole delta since 9ebe5715 at the current head: both UnicodeDecodeError guards bite under mutation. Pulling UnicodeDecodeError back out of corpus_report.py's tuple reds test_tally_survives_a_log_whose_bytes_are_not_utf_8 with the exact traceback, and the same on judge_report.py reds test_one_undecodable_run_log_does_not_take_the_whole_corpus_down. FLAG_UNEXERCISED, the find-big-wins row, and the drift-proof threshold rewrite all landed cleanly; harness unit suite green (the lone test_mock_mcp failure is the missing-engine-build case, reproduced identically on clean main, not this PR).

@johnmarkpeterbrown your corpus_report.py blocker is closed and I verified it the same way you did, so you're clear to flip your review.

aghadiayeamayanvboernest:
@johnmarkpeterbrown both leftovers are in as of `090f7f3d5`. Harness suite 2329 passed / 2 skipped, ruff clean.

**The docstring count** — you were right that my commit message overstated it. Both numbers I refreshed in round 3 were in the PR body; this one was in the code and I missed it. It's the module header at line 23, now written as the shape rather than the tally, using your wording:

> `init-project`'s `Place standardization` — all its numeric scores identical, most of its gradings N/A — flips from unflagged to flagged if N/A counts toward the threshold.

Measured before changing it: `n=3 +6 N/A`, so 3 numeric, matching your reading.

**The body clause** — fixed, and I found a second instance you didn't flag. The acceptance-check line said "2 numeric of 8"; the same stale count also sat in the definitions section above it. Both now describe the shape instead. Grepped the body for `2 numeric` afterwards to confirm none survived.

**Two neighbouring counts I checked rather than assumed**, since this keeps happening:

- `base/Tool Arguments` is 100% N/A in exactly three suites — `convert-dates`, `project-status`, `translation`. Still accurate, so I left your wording intact.
- `No-result handling` is still `n=1 +12 N/A`, so "1 of 13" holds.

Both are the same drift class and will go stale eventually. I left them because they currently reproduce and you didn't ask me to change them — say the word if you'd rather they went the same way as the others.

Thanks for the note that a senior from either team unblocks this now; that corrects what I'd understood from the earlier review.

johnmarkpeterbrown (approved):
looks good!

PR #1631, opened 2026-08-15
record-extractor: third-party personas (#1627) + marriage order and consent signer (alpha review)

### LINKED ISSUE #1627: record-extractor: named third parties collapsed into the subject's persona instead of getting their own record_role
**Touches:** packages/engine/plugin/agents/record-extractor.md, eval/tests/unit/record-extraction/marriage-index-parents-stated-direct.json

**Exemption for filing rather than fixing in the PR that found it (CLAUDE.md § "Work you find along the way"):** it needs a different skill — this is `genealogist`-lane persona-modelling doctrine, and validating any fix depends on a decision only the lead can make, since the record-extraction eval slot was spent twice on 2026-08-14 and a third run was declined. PR #1613's diff is already at its reviewed scope.

## The defect

`ut_record_extraction_025` (marriage index, both parties' parents stated) failed in `v1_2026-08-14_16-03-27` on:

```
- no new assertion with record_role='father_of_groom' fact_type='name' (expected at least one)
- no new assertion with record_role='mother_of_groom' fact_type='name' (expected at least one)
```

**It is not an extraction miss.** The parents were captured — they were stamped onto the *party's* persona:

```
role=groom  fact=name          ev=direct  value='John Becker (father of Frank Becker)'
role=groom  fact=relationship  ev=direct  value='child of John Becker'
role=groom  fact=name          ev=direct  value='Anna Kramer (mother of Frank Becker)'
role=bride  fact=name          ev=direct  value='Henry Wessel (father of Catherine Wessel)'
role=bride  fact=name          ev=direct  value='Mary Overberg (mother of Catherine Wessel)'
```

The whole run produced exactly two record_roles — `groom` and `bride`. No `father_of_groom` / `mother_of_groom` / `father_of_bride` / `mother_of_bride` persona exists, so every downstream consumer that keys on the parent's role finds nothing.

Two rules are broken at once:

1. **Persona modelling.** A named third party gets their **own** `record_role`. The vocabulary already lists `father_of_bride` and `mother_of_groom`, so the names exist; the agent did not use them.
2. **Atomicity / faithful capture.** The parent's identity and their relationship to the party are fused into one `value` string — `'John Becker (father of Frank Becker)'` — so the name assertion's value is not the name. The body's own rule ("`record_role` — this person's role in THIS record") and its one-fact-per-assertion doctrine both cover this.

## Why it matters more than one red row

`ut_025` had passed **6 of 6** prior recorded runs. This is the shape a regression takes when persona modelling drifts, and it is invisible to any matcher that does not existence-gate the third-party role — the parents' names are present in the file, just unreachable by role. `person-evidence` binds assertions by `record_id` + `record_role`, so a parent filed under `groom` cannot be bound to a parent person at all.

Worth checking whether the same collapse appears on other multi-persona record types (death certificates' `father_of_deceased` / `mother_of_deceased`, obituary survivors) — `ut_024` in the same run also mis-filed, though differently.

## Validation, please, not one run

`v1_2026-08-14_11-29-08` — the run two hours earlier, on nearly identical code — passed `ut_025`. So a single green run after any fix means nothing here. Take several `--test ut_record_extraction_025` scratch samples (~$0.30 each, gitignored, no `runs_per_test` policy change needed) and compare rates. The eleven-run matrix in PR #1613's body is the reference for how wide this suite swings.

## Provenance

Found while landing #1443 + #1407 (PR #1613); the run log is `v1_2026-08-14_16-03-27`.

Emruthwill:
Closes #1627.

Four `record-extractor` prompt changes batched so they cost **one**
`make eval-skill SKILL=record-extraction` run and one annotation pass instead
of four.

**Stacked on #1613**, which is unmerged, so this branch also carries that PR's
pre-1880 census reversal. Based on `main`, so the diff shrinks on its own when
#1613 lands. Merge #1613 first.

## Result

**`v1_2026-08-16_13-06-08` — 24 pass / 3 partial / 1 fail of 28.** Annotated
(33 dimensions over the 5 sampled tests plus one, no overrides).

Best of the five runs this suite has had in two days: 19/7/1 → 15/7/5 →
19/4/2/2 → 17/7/3 → **24/3/1**.

Four changes in one run cannot be attributed from the suite number, so each
carries its own direct signal. Those are the evidence; the aggregate is not.

### 1. A named third party gets their own `record_role` (#1627)

ut_025 previously fused four parent personas into the two parties, with the
tie smuggled into the name value:

```
record_role='groom'  fact_type='name'  value='John Becker (father of Frank Becker)'
```

No parent persona existed at all, and `person-evidence` binds by
`record_id` + `record_role`, so nothing downstream could reach them.

**Now:** roles are `groom`, `bride`, `father_of_groom`, `mother_of_groom`,
`father_of_bride`, `mother_of_bride`. **pass.**

Shipped with a deterministic guard (`test_name_value_is_a_bare_name`) that
flags a `name` value carrying a relation word plus `of`/`to`. Two true
positives on the existing corpus — unlike the birth-year guard, which had
none. Scoped so a maiden-name parenthetical, an uncertainty marker, and
`record_role: absent` negative-evidence values are untouched, each pinned by
its own test.

### 2. A derivative does not erase the informant (ut_021)

ut_021 had failed the same matcher in **three consecutive runs** —
`informant_proximity: unknown` on the parties' own names where doctrine says
`self`. The run log said why, in the agent's own notes:

> "Index entry derived from original marriage register; informant who provided
> name to the original register is unknown **through this derivative**."

The rule against that already existed — as a parenthetical inside a
definitions paragraph. A parenthetical loses to a specific, plausible chain of
reasoning. It is now a load-bearing rule that quotes the wrong inference back
verbatim.

**Now:** `groom prox=self`, `bride prox=self`. The test is `partial` rather
than `fail` — the targeted axis is fixed, something else still docks it.

### 3. Marriage order and the consent signer (alpha review)

From Shaunese Luthy's alpha-team review, 2026-08-14. Nothing instructed the
extractor to capture **which** marriage this is, or **who signed consent** for
an underage party — both of which settle a bride's surname, which
`check-warnings` classifies as an unsound assumption requiring positive
evidence.

New fixture **ut_029** (1893 Hamilton County license: widow bride on her
second marriage, underage groom whose consent is signed by `Jacob Lang,
stepfather`). **Passes on first contact**, which I had predicted would be
partial at best:

```
groom             marital_status   'Single — first marriage'
bride             marital_status   'Widow — second marriage'
consent_signer_1  name             'Jacob Lang'
consent_signer_1  relationship     'stepfather of Charles Ferber (groom)'
```

The signer is roled as himself with the **stated** tie — not inferred as a
father, which is the failure mode the review named.

The caveat the review asked for is in the doctrine: the signature is usually
on the reverse and usually absent from the index, so look on the image, and
record its absence as absence of evidence.

**ut_029's three novel-shape matchers are `optional: true`** for this first
run. `marital_status` and `consent_signer_1` were shapes the extractor had
never been observed producing, and hard-gating existence on an unproven shape
is what made ut_028 fail earlier — a validator failure suppresses the judge,
so a brittle gate costs the whole test's grade. Now that the run shows what it
actually emits, these can be tightened.

### 4. The consent line stopped presuming the signer is a parent

`references/information-classification-at-extraction.md` read *"Parents'
consent … primary from the **parent** who signed"* — presuming exactly what
the new doctrine exists to prevent. Two passages presuming different things
about one entity is the defect shape behind this whole issue cluster; shipping
a fourth instance alongside its own fix would have been careless.

## Carried from #1613

The pre-1880 census reversal gets a **third consecutive sample** here: 8/8
gated tests at zero relationship assertions, on a branch that also carries
three unrelated prompt edits without disturbing it.

## The one fail

**ut_017** — the decedent's birthplace is `indirect` where the matcher
requires `direct`. A *different* failure from its last one, and it is the
`evidence_type` axis of **#1628**: classification driven by informant
proximity rather than stated-vs-inferred. That was the most speculative item
on the follow-up list and now has a concrete witness.

## What is not claimed

That one run settles four prompt changes. Each fix above has a direct,
checkable signal, which is why they are worth more than the suite delta — but
**ut_016 moved fail → pass in this run with no change of mine**, which is the
standing reminder that single-run movement in either direction is mostly
sampling.


kofiatinka12 (approved):
This change improves genealogical accuracy by enforcing proper evidence classification, preventing unsupported relationships, and keeping extraction separate from interpretation. The approach better reflects professional genealogy standards and should reduce false-positive family connections.

francis-2008-happy:
Reviewed against `main` — full pass, 0 commits behind. Ran it twice, including one independent pass. All 7 required checks are green on this head, and no CODEOWNERS errors.

**Verified by running:** harness pytest is 2155 passed. The handful of failures are environmental in this sandbox (a missing `unzip` binary and an absent engine build), all in files this PR doesn't touch — not regressions. I also watched the new `test_name_value_is_a_bare_name` guard fire on the defect value (`"John Becker (father of Frank Becker)"`) and correctly pass on bare names, a maiden-name parenthetical, and an `absent`-role value. It has teeth and is scoped as claimed.

**No blocking code findings.** The prompt changes are internally consistent across the agent body, the reference doc, the rubric, and the fixture. `record_role` and `fact_type` are open enums, so `consent_signer_1` and `marital_status` need no schema change.

**Merge PR #1613 first.** Most of this diff is its stacked carry — the pre-1880 reversal, the #1407 matchers, and the harness retry guard all belong to it and leave this diff when it lands. Review the increment: `record-extractor.md`, `information-classification-at-extraction.md`, the ut_029 fixture, the name-value guard, and the run logs.

**Two things for a senior-genealogist, not the code side:**

1. Confirm the marriage-event informant reversal. The event's informant changes from officiant/clerk (`official_duty`/`witness`) to the parties at `self`. It's applied consistently, so nothing is inconsistent — just confirm you want the old text gone.
2. The four prompt changes and the run-log annotations are craft calls I left unadjudicated.

**Two non-blocking notes:**

- ut_029's three `optional: true` matchers (`marital_status` x2, consent-signer name) don't require the assertion to exist. On this run the signer was emitted, so its classification was checked. Worth tightening once a run confirms the shape, so a future regression that stops emitting them can't pass silently. Disclosed in the body — fine for the first run.
- The name-value guard is deliberately narrow. It won't catch `grandfather of` or `stepmother of` smuggled into a `name` value. Fine as a partial guard backed by the per-fixture role matchers — flagging so it's a conscious choice.

clack391:
Senior review, developer side. The code is clean, verified by running: the new `test_name_value_is_a_bare_name` guard fires on the collapse shape, passes on bare names / a maiden-name parenthetical / negative evidence, and its integration tests pass. Nice work.

One decision before this lands, not surfaced by the earlier passes: the guard false-positives when a relation word is actually a surname or a title followed by "of". Confirmed:

- `Joseph Parent of Quebec` (Parent is a common surname, and we handle Quebec records)
- `Julia Child of Boston` (surname Child)
- `Mary, Mother of Sorrows` (a devotional name)

Each trips on a legitimate name value. Risk is low (name values are usually bare, and these need the disfavored "Name of X" form) and the guard is a partial net by design, so documenting it as a known limit may be enough. It turns on name-value semantics, so it is a call for you and a senior genealogist. If you tighten it, anchoring the relation word to the start of the value, or requiring a name word before it, removes these without losing the collapse-shape catch.

Joins the two @francis-2008-happy raised: the guard also misses `s/o`, `grandfather of`, and `father-in-law of` shapes, and ut_029's optional matchers let a regression that stops emitting the signer pass silently. Both fine as first-run choices, worth tightening once a run confirms the shape.

None of this blocks the developer side.

Emruthwill:
Thanks both — the false-positive finding was real, reproduced, and is fixed. Details below, plus where I landed on the two non-blocking notes.

## The guard false-positived on surnames and titles — fixed

Confirmed all three against the shipped regex before changing anything:

| Value | Old | New |
|---|---|---|
| `Joseph Parent of Quebec` | **FIRES** | clean |
| `Julia Child of Boston` | **FIRES** | clean |
| `Mary, Mother of Sorrows` | **FIRES** | clean |

The suggestion to anchor the relation or require a name word before it resolved into something simpler: **scan only the bracketed part of the value.** Both observed collapse shapes put the relation inside brackets — `John Becker (father of Frank Becker)` and `Linda (given name only; spouse of Robert Whitaker)` — while all three false positives carry it in the bare part.

That is a genuine defect I would not have found, and it mattered more than a one-dimension dock: a failing validator suppresses the judge, so a false positive costs the test's entire grade.

## The missed shapes are closed too

Bracket-scoping makes a wider vocabulary safe, because inside brackets there is no surname ambiguity. So the shapes @francis-2008-happy raised now fire:

`grandfather of` · `great-grandmother of` · `stepfather of` · `father-in-law of` · `s/o` · `d/o` · `w/o`

**Verified: 0 false positives across 7 legitimate values, 0 misses across 7 collapse shapes.** All ten values from your comments are pinned as tests, named to the review so nobody quietly re-widens it later. Harness suite 2179 pass.

**One cost accepted, stated plainly:** a comma-form collapse (`John Becker, father of Frank`) is now missed. That is the safer direction to err — a miss costs one soft dimension, a false positive costs the whole test — and the per-fixture `record_role` matchers catch the collapse independently. Those matchers, not this guard, are what actually failed ut_025 before the fix.

Free to land: `eval/harness/` is outside the run-log snapshot, so no run log is invalidated and no eval run is owed.

## ut_029's `optional: true` matchers — agreed, and here is the concrete follow-up

You are both right that a regression which stops emitting the signer would pass silently. The run now tells us the shapes, so the guesswork is gone:

```
groom             marital_status   'Single — first marriage'
bride             marital_status   'Widow — second marriage'
consent_signer_1  name             'Jacob Lang'
consent_signer_1  relationship     'stepfather of Charles Ferber (groom)'
```

`marital_status` and `consent_signer_1` are exactly what the fixture guessed, so `optional` can come off all three. I have **not** done it in this PR: dropping `optional` edits a fixture, which invalidates the run log and owes another paid run plus an annotation pass, and it would be tightening on a single sample of a brand-new capability. The disciplined order is to let the next run of this suite confirm the shape a second time and drop `optional` in that PR, where it is free.

## The two genealogist questions

Both are the lead's, and both are already settled on his side rather than left hanging:

1. **The marriage-event informant reversal** (officiant/clerk → parties at `self`) was his explicit call, made while adjudicating the four disagreeing artifacts. The old text is intended to be gone.
2. **The pre-1880 relationship doctrine** carried in from #1613 is his ruling, quoted verbatim in that PR body.

The four prompt changes and the annotations remain craft calls, as you say — the annotation on the previous run does include one override where the judge failed a test for persisting a relationship assertion that provably did not exist, which is the kind of adjudication you are pointing at.

Agreed on merge order: **#1613 first**, and most of this diff leaves when it lands.

mercyokum (approved):
The four doctrine changes are sound:

- Third parties (father, stepfather, etc.) getting their own record_role
  instead of being fused into another party's name value — correct, and
  the guard backing it is well scoped.
- The derivative-doesn't-erase-the-informant fix is the right call. A
  marriage index is still just a copy of what the couple told the clerk
  at the register — looking through the index to the original informant
  is standard practice, not a stretch.
- Marriage order / consent-signer capture, and not presuming a consent
  signer is the father — both correct, and the ut_029 fixture (widow's
  remarriage, stepfather consent signer) is a good test of exactly the
  surname assumption this is meant to prevent.

T-FEH (changes requested):
Reviewed the developer half. The code is good, and @clack391's false-positive catch was answered properly — I re-ran the regex and all three of `Joseph Parent of Quebec`, `Julia Child of Boston`, `Mary, Mother of Sorrows` come out clean while 8/8 collapse shapes still fire, including `s/o`, `d/o`, `w/o` and `father-in-law of`. Scoping the scan to the bracketed part was a better answer than anchoring the relation word.

I also checked the thing this repo keeps getting wrong: none of the +484 lines of new harness tests is vacuous. Mutating three ways — falsy-instead-of-`== 0`, dropping the `usage["num_turns"]` write, dropping the counter increment — each kills a specific test, and neutering each validator guard kills 4–8 apiece. Run-log discipline holds too: 4 added, 4 pruned, cap exactly 5, and the two deleted `.ann.json` are siblings of pruned logs rather than an annotation dropped to dodge rule 3. `check_runlogs.py` exits 0 against the merged tree, not just CI's pre-merge head.

Suites: vitest **2365 passed / 100 files**, harness **2283 passed, 2 skipped**, `agent-tool-names` + `plugin-hooks` 67 passed. All 12 checks green.

**I pushed one fix (`d1dd425b`).** §5.6.1 now says extraction writes no parent-child or spousal assertion on a pre-1880 census "in any form", but §9's worked example still showed `a_004` (1850) and `a_010` (1860) as `child_inferred` relationship assertions — the same two-passages-disagreeing shape change #4 exists to remove, in the document the rule was added to. I scoped rather than deleted: those two ids are referenced in **eight** other places (the §5.7 cardinality paragraph, three `person_evidence` entries, two conflicts, two proof summaries), so deleting them would leave dangling ids and break the example. The note says whose output they are, which §5.6.1 already establishes.

**Not approving yet, for one reason that is not about this code.**

This branch contains all of #1613 — I confirmed `git merge-base --is-ancestor` — and it targets `main`, so merging it lands #1613 wholesale. Your body says "Merge #1613 first", and @francis-2008-happy said the same. #1613 currently sits at **CHANGES_REQUESTED**, and of @promise-emmanuel's two blocking points there:

- The retry guard retrying every wall-clock timeout — **fixed**, `turns_seen` added exactly as prescribed, and it is on #1613's branch too.
- *"Get the decision confirmed, close issue #1626, and fix the body"* — **not done.** `validators/test_record_extraction.py:653` still reads "Decided 2026-08-15 (issue #1626)", and **#1626 is open with zero comments**. There is no record of that decision anywhere I can find.

That matters more than a stale citation because of what it authorises: a doctrine reversal from "assert the relationship as `indirect`/`_inferred`" to "do not assert it at all", applied across the agent body, the reference doc, the rubric, eight fixtures, a validator that now fails the old shape, and the schema spec. The reasoning is persuasive and the determinism argument is real — but the reversal is currently resting on a decision whose only evidence is a docstring asserting it happened.

Either close #1626 with the decision recorded, or drop the "Decided" claim to a proposal and get @DallanQ to confirm. Once #1613 is unblocked and merged, I will re-review this on the shrunken diff and approve the developer side.

Two notes, neither blocking:

- @kofiatinka12's approval is at `8ca386be`, before `99f8d637` — the bracket-scoping fix, which is both the most substantive code change here and the answer to clack391. `dismiss_stale_reviews_on_push` is false so it still counts toward the required 2, but it did not see that commit. @mercyokum's at `77808e56` read everything.
- Rule 2b (warn): the run log scored under judge prompt `e1f3a744`; current is `a7c9bd99`. The body's "19/7/1 → … → 24/3/1" progression spans three judge prompts — you already caveat the aggregate, worth saying that is why.

Two things I did not judge and am not qualified to: the four genealogist craft calls and the annotations. Those are @mercyokum's, and her approval covers the genealogist-owned paths. The `eval/harness/**` Python still needs a senior developer, which neither current approver is.

Emruthwill:
Thank you — this is the review that caught the thing that actually mattered, and it was about my wording, not the code.

## The blocking point: fixed, and you were right to refuse it

`validators/test_record_extraction.py` said **"Decided 2026-08-15 (issue #1626)"** while #1626 held no record of any decision. Your framing is exact: a reversal spanning the agent body, the reference doc, `rubric.md`, eight fixtures, a validator that now *fails the old shape*, and the schema spec was resting on a docstring asserting that a decision happened. That is not authorisation.

Two things done, in the order that matters:

1. **The ruling is now recorded on #1626** — https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1626#issuecomment-5322695856 — quoted in full, with its author (@Emruthwill) and date, so it is checkable by reading the issue rather than taking a comment's word for it.
2. **The docstring now cites that record** rather than asserting the decision, and states plainly what it used to say and why that was not good enough. Landed on #1613's branch (`05d6133c`) and merged up through this branch and #1706, so all three carry it.

**I deliberately did NOT close #1626.** You offered "close it with the decision recorded" OR "get @DallanQ to confirm". Recording it is mine to do; deciding whether @Emruthwill's ruling is sufficient or whether @DallanQ's confirmation is the required bar is a question about how this team allocates doctrine authority, and closing the ticket would settle that by fiat. The issue comment names @DallanQ explicitly as the person to confirm or overrule.

**One correction I made to my own record:** I first dated the ruling 2026-08-16. That is the `evidence_type` ruling in #1628. The pre-1880 ruling is **2026-08-15**, matching commit `5890b9f8` and every citation in the code. The comment now carries that correction visibly — a record posted as authorisation is worth nothing if its own date is wrong.

## Your `d1dd425b` — good catch, and it is the same shape as the bug

You found §9's worked example still showing `a_004` and `a_010` as `child_inferred` relationship assertions while §5.6.1 says extraction writes none "in any form" — *in the document the rule was added to*. That is precisely the two-passages-disagreeing defect this whole cluster exists to remove, and I missed it in my own sweep of that file. Scoping rather than deleting was also the right call: those ids are referenced in eight other places, and deleting them would have left dangling references. Merged up into #1706 as well.

## Rule 2b — noted and worth stating, as you say

The progression `19/7/1 → 15/7/5 → 19/4/2/2 → 17/7/3 → 24/3/1` spans **three judge prompts**. That is a second reason the aggregate is uninformative, on top of the sampling variance I already caveat — and it is why the load-bearing claims here rest on deterministic validator results rather than on judge scores or pass counts. I will add that to the body.

## On the approval timing

Correct that @kofiatinka12's approval predates `99f8d637`, the bracket-scoping fix — which is both the most substantive code change and the answer to @clack391's false-positive finding. Flagging it rather than quietly banking it is the right instinct; I would rather it be re-confirmed than counted.

## What I am not treating as settled

Your closing point stands: the four genealogist craft calls and the annotations are @mercyokum's lane, and `eval/harness/**` still wants a senior developer that neither current approver is. I have not tried to route around either.

Verified after all merges: harness 2301 pass, `check_runlogs` exits 0, 12 checks green.

T-FEH (approved):
Clearing my changes-request. Both blockers are resolved, and the #1626 response is better than what I asked for.

**Merge order — resolved.** #1613 is closed per @DallanQ in favour of this PR, so the ordering problem is gone at the source rather than worked around. I verified nothing was lost: `git merge-base --is-ancestor` passes with 0 commits missing, and both of #1613's run logs and their `.ann.json` are here alongside this PR's own.

**The #1626 record — resolved.** `05d6133c` is present and `validators/test_record_extraction.py:653` now cites the recorded ruling instead of asserting one. The issue carries the ruling verbatim with author and date. It also caught something I had not: an earlier draft dated it 2026-08-16, which is a *different* ruling — the `evidence_type` axis in #1628 — while the pre-1880 one is 08-15, matching `5890b9f8` and every citation in the code. Repo-wide, no other file asserts an unrecorded decision.

Leaving #1626 open, on the grounds that whether your own ruling clears the bar or needs @DallanQ's is a question about how this team allocates doctrine authority, was the right call — that is not something to settle by closing your own ticket. @DallanQ, that comment is the thing to confirm or overrule; it is not blocking this PR, and @mercyokum's approval already covers the genealogist lane.

**I pushed one commit (`3bb5c265`) — a test that could not fail, and the way it got there is worth reading.**

`test_bare_name_rule_exempts_negative_evidence` carried no brackets in its value, and `99f8d637` scoped the regex to the bracketed part — so `_relational_name_hit` never matched and the test passed whether or not the `record_role == "absent"` exemption it names existed. Deleting that exemption from the validator left **all 16** bare-name tests green.

The test was sound when written. `99f8d637` — itself the correct fix for @clack391's false positives — silently vacated it, and nothing flagged that, because a test going from meaningful to vacuous does not change colour. Moving the relation inside brackets restores the intent: the regex hits, and only the exemption saves it. Verified both ways.

Added `test_bare_name_rule_only_scans_name_facts` for the sibling gap: dropping the `fact_type == "name"` filter also left the suite green. A relationship fact's job is to say "X of Y", so it must not read as the collapse shape.

Verified by running: `npm test` **2374 passed / 100 files**, harness **2300 passed, 2 skipped**, `check_runlogs.py` exit 0 against the merged tree (two warnings — the 2b judge-prompt hash, and ut_017 ungraded). Re-checked the things I cleared last time and they held: agent frontmatter and the three spellings intact, run logs 4 added / 4 pruned with the cap at exactly 5 and the latest active, and my §9 scoping note still accurate against `a_004`:1285 / `a_010`:1394.

Two notes, neither blocking:

- **A correction to my own last review.** I said `eval/harness/**` still needed a senior developer that neither approver was. That is obsolete — `dd95bdfe` landed on main and CODEOWNERS now lists **both** teams on every rule, so @mercyokum's approval covers the Python and everything else here.
- The PR body still says "Merge #1613 first", which is now impossible. Worth a line saying it was closed in favour of this instead.
- @kofiatinka12's approval is still at `8ca386be`, well before the bracket-scoping fix and everything since.

Prompt budget stays worth watching: `record-extractor.md` is 52.6 → 56.4 KB, billed on every extraction.

Issue #1670, opened 2026-08-16
Show the judge validator FAILURES only, not the full results


DallanQ:
**Touches:** eval/harness/judge/prompt.md, eval/harness/harness/judge.py

Deferred from PR #1637 (merged).

The judge prompt tells the model that deterministic checks "are verified
separately by validators" and then shows it nothing. A `{validator_failures}`
slot in the varying half of the prompt is cheap.

## Failures only — this is the whole design

Do **not** pass the full validator result list. A passing validator list is a
*conclusion*, and handing the judge a conclusion is the defect diagnosed across
issues #1007, #1330 and #1603: the test tells the judge what to find and the
judge finds it. Showing "all validators passed" invites exactly the
rationalisation this is meant to remove, and would raise the pinned-at-3 rate.

A **failed** validator is different in kind — an observation the judge would
otherwise miss, and one that cannot be rubber-stamped into a pass.

## Rank it below deference

Where the answer is mechanical, *deferring* beats *informing*: see
`apply_deterministic_deference` and `apply_routing_deference` in
`orchestrator.py`. This is for the residue those cannot decide.

Editing `judge/prompt.md` moves `judge_prompt_hash`, which is warn-only
(`rule2b_judge_prompt`) and forces no re-run.

DallanQ:
## Closing — built instead of filed (PR #1673)

Shipped in PR #1673. `{validator_failures}` is now a slot in the judge prompt,
carrying **failures only**: a passing validator list is a conclusion, and handing
the judge a conclusion is the defect behind issues #1007, #1330 and #1603.

**Filing this was the wrong call, and the measurement is why.** The change is
four functions in one file — `render_prompt`, `render_prompt_parts`, `grade`, and
the orchestrator call site — plus one prompt section. An issue costs four
people's attention: someone to vet it, someone to implement it, two to review it.
That overhead is not proportionate to four functions.

At filing time I asserted "threads a new slot through four call layers" without
having opened them. I opened them today and the count is what made the answer
obvious. Asserting scope instead of measuring it is exactly what CLAUDE.md warns
turns a foldable fix into an orphaned ticket.

PR #1673, opened 2026-08-16
eval: report identifiers a skill persisted that no input supplied (#1667)

### LINKED ISSUE #1667: Universal provenance check: an identifier in a persisted artifact must trace to a tool response
**Touches:** eval/harness/validators/test_universal.py

Deferred from PR #1637 (merged).

**Seven of 27 skill bodies state this rule in prose. Nothing checks any of them.**

- `research-plan`: "never cite or invent a specific source (a `src_` id, or a named document) that is not already in the before-state"
- `citation`: "every value you write into `citation` or `citation_detail` must be traceable to the existing source entry, `research.json`, `tree.gedcomx.json`, or the user's message"
- `search-external-sites`: take the `standardPlace` from the response — "do not guess it"
- `check-warnings`: "do not invent fact ids, dates, or sources the tool didn't return"
- plus `locality-guide`, `init-project`, `person-evidence`

## What to build

One universal validator: every identifier and every number appearing in a
persisted artifact must trace to (a) the before-state, (b) a prior tool response
in the same run, or (c) the user message. Nothing else is allowed.

This is a **Tier 2 response-derived check** — the expected value is computed from
the run's own trace, so it needs no per-fixture authoring and applies to all 25
suites at once.

## Why it is worth doing first among the deterministic checks

It is the one check that would have caught the only fabrication defect in the
corpus that reached a persisted artifact: `ut_research_plan_011` extrapolated a
census collection id out of a truncated `collections_search` response
(issue #1332). It would also have **cleared** issue #1603, where the judge read
three real-but-unincorporated sources as fabrications.

## Watch out

Run it against the committed corpus before making it blocking — the run logs
carry `output.tool_calls` and `output.file_changes`, so the whole check can be
replayed offline for free. Expect false positives on values a skill legitimately
derives (a computed date, a place name normalised by the model) and decide
whether those need an escape hatch before it gates anything.

A change confined to `eval/harness/` marks no skill touched, so this owes no
paid eval run.

DallanQ:
Closes #1667.


DallanQ:
Two things block this. The rest looked fine.

## 1. The judge only runs when every validator passed, so the failure list is always empty

`compute_validators_passed` requires all validators to pass, and the judge call sits behind that check. So `[r.name for r in validator_results if not r.passed]` can only be non-empty in one case: an `intentionally_invalid` test where a file-validity validator failed — a scenario that was authored broken on purpose.

Measured across all 1,843 committed unit runs. 41 runs had a validator fail, and the judge was skipped on every one. 29 runs reached a graded judge with a failure present, and all 29 are `intentionally_invalid` (21 in validate-schema, 8 in project-status). The three names that would ever appear are `test_research_json_validates_schema`, `test_project_files_pass_full_validation`, and `test_id_references_resolve`.

So on ordinary tests the new slot always renders `(none failed)`, and on the tests where it does fire it reports a failure that was expected.

The three new tests call `render_prompt` with a hand-written list. They never go through the gate, so they pass either way.

This needs a decision before more code: should the judge run when a validator fails? Today it doesn't, and issue #1670 can't deliver anything until that changes. I'd reopen issue #1670 and settle that first.

Bookkeeping while you're there — this PR carries issue #1670's implementation, but the body only says `Closes #1667`, and issue #1670 is currently closed as not planned.

## 2. The provenance report doesn't flag the case issue #1667 was written around

Issue #1667 justifies the check by saying it would have caught `ut_research_plan_011`. It doesn't.

That run log has been pruned, so I recovered it from git history and replayed the scanner against it. The skill persisted collection id `1325221` and the report says it traces fine. It traces because `1325221` sits in `eval/fixtures/mcp/collections-search-us-national.json`, one of that test's declared fixtures. The fabrication was against what the response returned at that turn, not against what the fixture file holds.

I also tried narrowing the haystack to the fixtures each run actually returned — the run logs record that on all 5,677 tool calls. It makes no difference: 27 findings either way. So that isn't the fix.

The tool still surfaces 16 things worth triaging, so it's worth keeping. What has to change is the claim, and this PR shouldn't auto-close issue #1667.

```
You wrote:      reads the test's declared `mcp_fixtures[]` instead — which is sound for
                the unit harness and would not be for a live run.

Change it to:   reads the test's declared `mcp_fixtures[]` instead. This is weaker than
                it looks even for the unit harness: a fixture file holds every result
                the tool *could* return, so an id the model extrapolated past a
                truncated response still traces. Replayed against
                `ut_research_plan_011` (issue #1332) it reports nothing — the case
                issue #1667 cites as this check's justification is one this check
                cannot see.
```

Same edit picks up the numbers just above it, which are stale:

```
You wrote:      Restricting to ARKs and 5+ digit numbers gives **46 hits across 26
                test-pairs of 1,842 test-runs**.

Change it to:   Restricting to ARKs and 5+ digit numbers gives **16 findings across 12
                tests of 1,843 test-runs**.
```

---

Verified: harness suite 2136 passed / 2 skipped on the merge with main, after building the engine. Ruff clean. All 11 CI checks green on this head. `make provenance-report` gives the same 16 findings before and after merging main.

Issue #1668, opened 2026-08-16
deep dive: question-selection — and the rubric-deletion doctrine it pilots


DallanQ:
> **Reviewed 2026-08-24 before junior handoff.** The doctrine half already shipped in PRs #1678 and #1680, so the remaining work is the deep dive plus a per-dimension deletion call — and the written call cannot be argued from the re-run scores.

**Touches:** eval/tests/unit/question-selection/, eval/runlogs/unit/question-selection/,
docs/deep-dives/, packages/engine/plugin/skills/question-selection/SKILL.md (lane 4 only)

Rewritten again 2026-08-16. Two earlier plans are dropped: the `covered_by:`
declaration in `rubric.md` (built, then closed unmerged as PR #1675), and the
per-skill walk of every rubric dimension against its validators. Read "Why the
earlier plans were dropped" before proposing either again.

**This issue now also carries the `question-selection` deep dive** (merged in from
issue #1649, 2026-08-17) — same skill, same files, one paid run. See the second half.

## What we measured

- **Judge grading is 4.4% of eval spend.** Across the latest run log for all 25
  skills: judging cost $3.64, the skill runs cost $79.43. Retiring dimensions
  was never going to save meaningful money, and landing a sweep would have cost
  more in re-runs than it saved.
- **78 of 94 rubric dimensions scored the same on every test** in their latest
  run. Sixteen skills have no dimension that discriminates at all:
  `conflict-resolution`, `convert-dates`, `historical-context`,
  `hypothesis-tracking`, `init-project`, `proof-conclusion`,
  `question-selection`, `research-plan`, `search-external-sites`,
  `search-full-text`, `search-records`, `search-wikipedia`, `timeline`,
  `translation`, `tree-edit`, `validate-schema`. The counterexample worth
  keeping in view is `research-exhaustiveness`, where all three dimensions
  score both 1s and 3s.
- **The cost that is real is human, not tokens.** A run log cannot be released
  until every dimension of every test has a correction entry, so a genealogist
  reviews 861 dead cells per full annotation wave. A dead dimension also holds
  one of the five rubric slots.

That measurement is one run log per skill. A dimension that is flat on the
latest run is not necessarily flat across versions, which is why the call
belongs to the rubric audit — it reads prior versions too — rather than to a
sweep done from this list.

## What to do

Every skill gets a deep-dive review over the next few weeks. This work rides
those reviews. No new field, no parser, no script, no separate project.

[...]

PR #1675, opened 2026-08-16
eval: retire a rubric dimension when a validator covers it (#1668)

DallanQ:
eval-harness: retire a rubric dimension when a validator covers it (#1668)

Approved by the lead 2026-08-16. A rubric dimension may now declare the
deterministic validator that decides it:

    - **covered_by:** test_expected_classifications

When that validator RAN and PASSED, the dimension is removed from the rubric the
judge sees — the criteria never enter the prompt, so the tokens are not spent —
and re-attached to the run log with score: null and a rationale beginning
[covered-by]. Rule 3 owes no correction for it, so a genealogist is not asked to
confirm what a validator already decided.

When the validator FAILED or did not run, the dimension stays with the judge. A
failing check did not answer it, and retiring on a failure would convert a real
defect into a silent null.

**Why this and not more checks.** Adding a deterministic check did not previously
retire the judge dimension it duplicated, so oracle coverage grew while grading
cost stayed flat. question-selection is the worked example: 12 skill-specific
validators — including mechanical selection_basis checks — AND 7 of 7 dead judge
dimensions grading the same axes. 45 of 74 dead dimensions are fully
mechanizable, worth 1,984 graded cells, and none of that becomes a saving until
a covered dimension can stop being graded. This is the mechanism that makes
every future check subtract rather than add.

**One interaction, and it is the part most likely to have been silently wrong.**
PR #1637 made "a rubric null on a positive test" the TOP targeted rule of the
review sample, because a null standing in for a 1 records the run as a pass.
A covered_by null is the opposite of that — a validator ran, passed, and
answered the dimension. Without the exclusion the targeted slot would go to
exactly the dimensions we retired to save effort, on every run of every adopting
suite. Both readers key on the shared COVERED_BY_PREFIX constant rather than a
literal, so the two cannot drift apart.

Removing the dimension from the judge's rubric, rather than grading it and
overwriting the score, is what makes this a real saving and not a relabelling.

11 tests. Both interaction guards broken deliberately and watched to fail:
dropping the sampler exclusion reddens the covered-null test; dropping rule 3's
skip reddens the no-correction-owed test. Also pinned: a FAILING validator does
not retire its dimension, and covered_by is optional so every rubric in the
corpus still parses.

Nothing adopts covered_by yet — no rubric.md is touched here, so no skill's
snapshot moves and no eval re-run is owed. Design it against a real validator
first; `make provenance-report` (PR #1673) is the intended first candidate.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/docs/specs/unit-test-spec.md
+++ b/docs/specs/unit-test-spec.md
@@ -700,6 +700,29 @@ Two matcher modifiers keep the check both precise and non-flappy:
 
 **Deterministic-validator deference (grading).** When `test_expected_classifications` **passes**, the LLM judge's `Evidence type accuracy` and `Informant identification` dimensions cannot **FAIL** on the verified classifications — the harness floors a judge `1` to `2` (`orchestrator.apply_deterministic_deference`). A fuzzy re-grade must not override a deterministic check that already confirmed the classification (this retired the recurring census direct/indirect judge-inversion flap). Partial (`2`) is still permitted for a real issue on an *undeclared* assertion. Correctness likewise does not grade classification at all (base judge prompt) — evidence_type/proximity/quality are the classification dimensions' scope.
 
+**Validator coverage (`covered_by`).** A rubric dimension may declare a deterministic validator that decides it:
+
+```markdown
+## Evidence type accuracy
+
+Does the classification hold.
+
+- **pass:** …
+- **partial:** …
+- **fail:** …
+- **covered_by:** test_expected_classifications
+```
+
+When that validator **ran and passed**, the dimension is removed from the rubric the judge sees — the criteria never enter the prompt, so the tokens are not spent — and it is re-attached to the run log with `score: null` and a rationale beginning `[covered-by]`. No correction is owed for it (`check_runlogs` rule 3 skips it), so a genealogist is not asked to confirm what a validator already decided.
+
+When the validator **failed or did not run**, the dimension stays with the judge. A failing check did not answer the dimension, and retiring on a failure would convert a real defect into a silent null.
+
+Optional, and every rubric predates it: `pass`/`partial`/`fail` remain required.
+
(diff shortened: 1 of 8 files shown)


DallanQ:
Closing rather than patching. The review found the feature does not work in either direction, and one finding invalidates the design rather than the implementation.

**It crashes on the first graded test.** `_run_judge` has **zero return statements** — AST-confirmed — because my edit changed `return grade(...)` to `out = grade(...)` and the follow-up that was meant to add `return out` never landed (its anchor referenced a line I removed a step later to keep this independent of PR #1673). Every real judge call returns `None`, and `judge_output.dimensions` then raises outside the `except JudgeError` handler.

**2,136 tests passed anyway.** Every orchestrator test monkeypatches `_run_judge`, and ruff here is pinned to `select = ["F821"]`. Nothing in this repo could have caught it. That is the more useful finding.

**And the design premise is false.** My own commit message says "removing the dimension from the judge's rubric, rather than grading it and overwriting the score, is what makes this a real saving." It does not: `judge.py` renders the prompt from `rubric.raw`, and `split_covered_dimensions` copies `raw` verbatim. So on a partial retirement the judge still sees the full criteria, still grades the dimension, and the grade is then silently discarded as `dropped_unknown_rubric_dimension` — full token cost, a spurious warning every run, and a thrown-away score. Only the all-retired case saves anything.

Two more that change the shape: `pytest.skip()` records `passed=True`, so a **skipped** validator would retire its dimension — and `test_expected_classifications` skips on exactly the tests carrying no ground truth. And the CRUD UI's `unreviewedDimensions` was not given the matching skip, reopening the UI-vs-CI split for the third time in this branch's history.

Fixing four blocking findings on top of a wrong premise is worse than redoing it. Everything learned is recorded on issue #1668, which stays open and approved.

PR #1679, opened 2026-08-16
Drop obsolete max_cost_usd caps from e2e fixtures (#1132)

### LINKED ISSUE #1132: Drop the temporary $25 max_cost_usd caps from the bottemiller and cruz e2e fixtures
`eval/tests/e2e/bottemiller-parents/fixture.json` and
`eval/tests/e2e/cruz-corona-ancestry/fixture.json` each carry
`caps: {max_cost_usd: 25}`, added as headroom for the extractor-state-diet
measurement window.

Post-diet, cruz ran \$5.29 (`run-2026-07-22_12-02-35`); bottemiller has had no run
since 2026-07-13 (\$19.56).

DoD: re-run bottemiller; if it lands under \$15, delete both `caps` blocks.

Note the entry this came from called the \$15 default "the regression gate" — it
is not one; see the `max_cost_usd` note in `docs/specs/e2e-test-spec.md` §6 step 5.
Deleting these blocks also restores `e2e-test-spec.md:148`'s claim that `caps` is
not a fixture field; today the spec and the fixtures contradict each other.

kofiatinka12:
## Summary

- Removes the obsolete `max_cost_usd` cap from **five** of the seven e2e fixtures that set it (Trivial tier — data-only fixture edit). **`jimmie-jewel-neal` (45) and `bottemiller-parents` (25) keep theirs** — see "Deviated from the plan."
- `max_cost_usd` never stops *execution* — the orchestrator checks it only on the final result message (`orchestrator.py:1873-1876`), and subagent tokens never reach the main stream so it can't know true running cost mid-run. But it is **not** cosmetic: exceeding it sets `aborted_reason = "cost_cap"`, and `derive_stop_reason` (`stop_checker.py:102-103`) returns `"cost_cap"` **in priority over `completed`**. So on a fixture whose real cost exceeds the value, a genuinely finished run gets its `stop_reason` reported as `cost_cap`, which `interpret-e2e-result` surfaces to the genealogist as "why it stopped."
- The removal is a true no-op only where dropping the cap does not lower the *effective* ceiling below the fixture's real cost. That holds for the five removed here; it does **not** hold for jimmie or bottemiller.

## Review guidance

**Start here:** two code sites, and one distinction.
- `orchestrator.py:591` — `max_cost_usd=caps_raw.get("max_cost_usd", defaults.max_cost_usd)`: a fixture with no `max_cost_usd` falls back to the `15.0` default (`FixtureCaps`, `orchestrator.py:520`).
- `orchestrator.py:1876` + `stop_checker.py:102` — that value feeds the `cost_cap` stop-reason label.
- **The distinction that decides each fixture:** removal is safe only when it doesn't lower the effective ceiling under the real cost. A fixture whose cap was already `15` (= the default) is a *definitional* no-op — effective ceiling unchanged. A fixture whose cap was **above** 15 (jimmie 45, bottemiller/cruz 25) actually lowers the ceiling, so it's only safe if the fixture's real cost is under 15.

**Unsure about:** Nothing.

## Plan

**Didn't change:** All enforceable per-fixture caps (`wall_clock_seconds`, `inactivity_seconds`, `tool_calls`, `max_turns`). **`jimmie-jewel-neal`'s and `bottemiller-parents`'s `caps` blocks are left as they were on main** — jimmie `wall_clock_seconds: 10800` + `max_cost_usd: 45`; bottemiller `max_cost_usd: 25`. Also not touched: `docs/specs/e2e-test-spec.md:148` (see Follow-on).

**Acceptance check:** Data change, not a behavior change, so there's no "fails-on-main / passes-here" test. Verified by: (1) all seven fixtures pass `uv run python -m e2e.validate_fixture <slug>` with no ERROR; (2) `grep -rn max_cost_usd eval/tests/e2e/` now returns only `jimmie-jewel-neal` and `bottemiller-parents` (intended); (3) each removed fixture's effective ceiling is unchanged (four were exactly `15`; cruz's post-diet real cost is $5.29 < 15).

**Deviated from the plan:** The issue (and Dallan's expansion) listed **all seven**, on the premise that `max_cost_usd` is post-hoc and removing it changes nothing. Two reviews showed that premise is incomplete — the value overrides a *completed* run's `stop_reason` to `cost_cap` — and two fixtures have real costs above 15:
- **`jimmie-jewel-neal`** (@florencemashipei): committed runs cost $21.50 / $25.24. Cap retained at 45.
- **`bottemiller-parents`** (second review): committed post-diet run `run-2026-07-13_05-50-33` cost **$19.56** and reported `completed` only because its 25 cap covered it; the earlier `run-2026-07-12_21-34-52` at $16.99 under a 15 cap reported `cost_cap`. The diet's "revert if under $15" condition (commit `c953e47`) was met for cruz ($5.29) but never for bottemiller. Cap retained at 25.

The other five are removed: `cruz-corona-ancestry` (25→default; post-diet real cost $5.29 < 15), and `elena-asmundsdotter-origin` / `ferber-grandparents` / `jan-gallo-father` / `victor-spenard-parents` (were exactly `15` = the default, so removal is a definitional no-op regardless of cost).

## Test plan

Genealogist-lane fixture JSON under `eval/tests/e2e/` — per CODEOWNERS (last-match-wins, `/eval/tests/`) this needs **senior-genealogists** review.

- Relevant check for a fixture edit: `validate_fixture` — run on all seven, clean (no ERROR). CI runs the full suites.
- Run-log-snapshot re-run and negative-routing checkboxes: **N/A** — no skill body, `description`, or unit test changed.

## Follow-on issues

**Folded in / filed instead:** Nothing filed. Noting one pre-existing inconsistency for future cleanup (not filed, per the issue owner's call): `e2e-test-spec.md:148` still says "`caps` is NOT a fixture field," but fixtures legitimately set enforceable per-fixture caps (and deliberately, jimmie's and bottemiller's `max_cost_usd`) that the orchestrator reads. That doc line predates this PR and is left untouched to keep this change focused.

Closes #1132


florencemashipei (changes requested):
The removal principle is sound, and it's a genuine no-op for 4 of the 7 fixtures — `elena-asmundsdotter-origin`, `ferber-grandparents`, `jan-gallo-father`, and `victor-spenard-parents` all had `max_cost_usd: 15`, exactly matching `orchestrator.py`'s default fallback, so nothing changes there. `bottemiller-parents` and `cruz-corona-ancestry` (25 → 15) are a smaller, probably-fine gap.

One thing worth a closer look: **`jimmie-jewel-neal`** (45 → 15).

The PR's justification is that `max_cost_usd` is "post-hoc... never stops a run," which is true for *execution* — but it does have a real effect on *reporting*. `stop_checker.py`'s `derive_stop_reason()` explicitly prioritizes it:

```python
"""Priority: explicit SDK aborts win over project status — if a cap
fired, we want the cap reason in the result even if the agent had
already set status=completed"""
if sdk_aborted_reason == "cost_cap":
    return "cost_cap"
...
if project_completed(research):
    return "completed"
```

So crossing the threshold overwrites a genuinely successful run's `stop_reason` from `"completed"` to `"cost_cap"` — and that's not a cosmetic field: `interpret-e2e-result` reads exactly this to tell a genealogist *why a run stopped*. A run that actually finished the research task would now read as if it hit a budget ceiling.

`jimmie-jewel-neal` is called out elsewhere in this same PR as "the corpus's longest-running fixture," deliberately given `wall_clock_seconds: 10800` — 3x every other fixture's 3600s. A 3-hour agentic session plausibly costs proportionally more too, which is likely exactly why someone gave it 45 instead of 15 in the first place. Dropping straight to the generic default here, while leaving its wall-clock allowance elevated, means more of its future successful runs will likely get their `stop_reason` reported as `cost_cap` when they didn't need to be.

**Ask:** for `jimmie-jewel-neal` specifically, either keep a `max_cost_usd` proportional to its extended wall-clock budget (e.g. something in the 30-45 range rather than dropping to 15), or if you're confident this fixture's actual cost profile doesn't warrant it, say so explicitly in the PR so the next reader isn't left to reconstruct this reasoning themselves. The other 6 fixtures are fine as-is.

kofiatinka12:
Thanks @florencemashipei — you're right, and I've fixed it. Verified the mechanism: exceeding `max_cost_usd` sets `aborted_reason = "cost_cap"` at `orchestrator.py:1876`, and `derive_stop_reason` (`stop_checker.py:102-103`) returns `"cost_cap"` in priority over `completed` — so a genuinely finished run over budget gets reported as budget-capped, which `interpret-e2e-result` surfaces. Post-hoc for execution, but not for reporting.

`jimmie-jewel-neal`'s committed runs cost $21.50 and $25.24 (both above the 15.0 default), so I've **retained its `max_cost_usd: 45`** — restoring its `caps` block to exactly its pre-PR state (45 sits at the top of your 30–45 range and is proportional to its deliberate 3× wall-clock window). The other six are removed as planned: the four that were exactly `15` are pure no-ops, and bottemiller/cruz drop 25→15 with a post-diet cost profile under 15 (cruz ran $5.29). The diff is now six fixtures; jimmie is unchanged.

PR body updated with this reasoning under "Deviated from the plan." Re-requesting your review.

kofiatinka12:
@florencemashipei — the change is in, it just doesn't appear where you might be looking. Because your ask was to **keep** jimmie's cap, I addressed it by **restoring** `jimmie-jewel-neal`'s `max_cost_usd: 45` (commit `24725152`). That puts jimmie's `caps` block back to its exact pre-PR / `main` state:

```
"caps": {
  "wall_clock_seconds": 10800,
  "max_cost_usd": 45
}
```

Since jimmie now matches `main`, it correctly **drops out of the overall diff** — the PR's "Files changed" now shows **6 fixtures, not 7, and jimmie is not among them**. So "jimmie looks unchanged" is exactly the intended outcome: its cap is retained.

The explicit re-addition is visible in commit `24725152` (`+ "max_cost_usd": 45`). The other six removals stand: the four that were exactly `15` are pure no-ops, and bottemiller/cruz drop 25→15 (post-diet cost under 15). The PR body's "Deviated from the plan" section now records the `stop_reason`→`cost_cap` reasoning you raised.

florencemashipei (approved):

EdmondOware:
Flagging a gap before this goes to a senior for sign off. The same check that correctly kept jimmie jewel neal's cap in place was not applied to bottemiller parents, and the evidence for that is already committed in the repo.

bottemiller parents caps.max_cost_usd was set to 25 by commit c953e47c on 2026 07 13, described in that commits own message as temporary experiment headroom, to revert if the diet lands runs under $15. Its most recent committed run, run 2026 07 13_05 50 33, executed right after that same diet landed, cost $19.56 and correctly reported stop_reason completed because the 25 cap covered it. An earlier run before the cap was raised, run 2026 07 12_21 34 52 at $16.99, got stop_reason cost_cap, which is exactly the mislabeling this PR is meant to prevent.

The revert condition from that original commit was true for cruz corona ancestry, which ran $5.29 post diet, but it was never true for bottemiller. Dropping its cap now reintroduces the same failure mode preserved for jimmie: a future completed run in the sixteen to twenty four dollar range will get reported as cost_cap instead of completed.

Separately, elena asmundsdotter origin has no committed e2e run log in the repo at all, so its real cost profile cannot actually be checked against the fifteen dollar default the way the PR claims for all six.

Suggest giving bottemiller the same treatment as jimmie, keep an explicit max_cost_usd of 25 rather than reverting to default, and confirming elena's real cost profile before treating it as a proven no op. I did verify the underlying code claims separately, stop_checker.py and orchestrator.py match what the PR describes, this is a gap in applying that logic consistently, not a problem with the reasoning itself.

kofiatinka12:
Verified against the committed run logs — you're right on bottemiller, and I've applied the jimmie treatment to it (commit `d830591c`):

- `run-2026-07-13_05-50-33`: **$19.56, cap 25, stop_reason `completed`** — post-diet, covered only because of the 25 cap.
- `run-2026-07-12_21-34-52`: **$16.99, cap 15, stop_reason `cost_cap`** — the mislabel this PR prevents.
- Diet's "revert if <$15" (commit `c953e47`) held for cruz ($5.29) but never for bottemiller.

So **bottemiller keeps `max_cost_usd: 25`**. It now drops out of the diff (matches main), exactly like jimmie — the diff is 5 fixtures.

On **elena**: two things. First, it *does* have a committed e2e run log (`run-2026-08-13_01-40-40`, present on this branch too) — you may have hit a stale view. Second, and more decisive: elena's cap was **`15` — exactly the default**. Removing a `15` that falls back to a `15` default doesn't lower the effective ceiling at all, so it can't introduce the `cost_cap` mislabel regardless of elena's real cost. Only the caps that were **above** 15 (jimmie 45, bottemiller/cruz 25) actually change the ceiling — and of those, cruz's post-diet real cost is $5.29 (< 15), so it's the only 25-cap that's a genuine no-op to drop.

Net split: **jimmie (45) and bottemiller (25) retained; cruz + the four default-15 fixtures removed.** PR body's "Deviated from the plan" updated with this. Thanks for the catch — it was a real consistency gap.

EdmondOware (approved):
Confirmed the fix. bottemiller parents now keeps max_cost_usd 25, same treatment as jimmie, and the commit message cites the exact same reasoning back, the post diet run at $19.56 only reported completed because of that cap, and the diet's revert if under fifteen dollars condition was never actually met for bottemiller.

I also checked the four remaining fixtures directly in the diff. elena asmundsdotter origin, ferber grandparents, jan gallo father, and victor spenard parents all had max_cost_usd set to exactly 15 before removal, the same as the default, so those really are no ops regardless of real cost. That also closes out my earlier note about elena having no committed run log, since the effective ceiling does not change either way for a fixture whose cap already equaled the default. cruz corona ancestry is the only one dropping from a higher cap, and its confirmed $5.29 post diet cost is safely under 15.

Issue #1603, opened 2026-08-13
next run: proof-conclusion


EdmondOware:
> **Reshaped into the standing queue card for this skill (icebox review, 2026-09-20).**
> `icebox` removed, `cluster:next-run` added, retitled. Nothing below is deleted; the
> detail and the reasoning for every item stay on this card and in its comments.
>
> This card was iceboxed on "waiting for the paid slot", which is sequencing, not a thaw
> condition — its own 2026-09-07 status note said `cluster:next-run` was the right shape.
> `proof-conclusion` was the only skill with a contended eval slot and **no** standing
> queue card, which is why this one kept getting used as one (issues #2510, #2509 and
> #1848 hold research-plan, search-records and record-extraction).

**Touches:** none — this card holds pointers, never content. The work, the reasoning and
the acceptance criteria stay in the real issues.

The queue waiting on `proof-conclusion`'s single eval slot. Editing anything under
`eval/tests/unit/proof-conclusion/`, `packages/engine/plugin/skills/proof-conclusion/` or
`packages/engine/plugin/agents/proof-conclusion.md` flips this skill's run-log snapshot
inactive and buys one `make eval-skill SKILL=proof-conclusion` run (~$7, measured on
`v1_2026-09-15_18-56-53`) plus a full genealogist annotation pass over a 23-test suite.

## Slot status, 2026-09-20

**Free.** No open PR touches the snapshot (all 60 open PRs scanned). Latest committed run
`v1_2026-09-15_18-56-53`: 23 tests, **21 pass / 1 partial / 1 fail** — `ut_proof_conclusion_018`
fails, `ut_proof_conclusion_007` partials.

## Queued

- **#2604** (open, @Gennecis, `senior`, `cluster:pair-conversion`) — **the next run.** Fixes
  `_018` at its cause, carries the 8 direct-agent twins, and holds **items 1–6 of this card**
  in full under its "Part C — six test corrections folded in from #1603". Its body says
  outright: *"#1603 stays open as the standing parking card"*. Do not work items 1–6 from here.
- **#2524** (open, @Gennecis, `reviewed`, `high-priority`) — holds **item 7**, the missing
  "issued as a separate `tree_correct` call" guard on the retiring bullet in
  `packages/engine/plugin/agents/proof-conclusion.md`.

## Two stale pointers, corrected

- **The 2026-09-16 comment below says "check #2617 before starting" and that items 1–6 revert
[...]

PR #1684, opened 2026-08-17
judge_context: fix the authoring UI that teaches answer keys (#1007, #1330)

### LINKED ISSUE #1007: search-records grading: judge_context hands the judge its finding, on both the triage and maiden-surname tests
> **Analysis 2026-08-09 — the defect is real and larger than filed, but it is in the grading, not in a ~75% compliance rate.**
>
> **The measured rate is 4-for-4, not 3-in-4.** All four committed run logs score `ut_search_records_013`'s `Result triage` at **3**, each confirmed by a genealogist annotation with no comment. Across the full recoverable history (three earlier logs pruned by the retention sweep, recovered from git) it is 6-of-7.
>
> **But the dimension has become incapable of failing: `Result triage` scored 3 in 78 of 78 graded cells** — every test, every committed run. Reproduced directly from the run logs.
>
> **And the run that most flagrantly violates the doctrine is one of the passes.** Read from the transcripts rather than the grades, the four runs comply 0-for-4 with the four prohibitions this body names. The 2026-08-06 run — the most recent — flags the near-match "✅ Candidate", writes "this is within normal variance… The ~1845 in the tree was itself approximate", calls it "**a strong candidate**", and offers "Would you like me to proceed with record-extraction". It scored 3, with a judge rationale reading "The result was correctly categorized as a strong candidate." A genealogist then confirmed the 3.
>
> **The cause is the test's own `judge_context`**, which tells the judge the excuse is true: *"a MIDDLING matchScore (0.52) … reflects surname transcription variance … and an age off by ~3 years, **not a genuine mismatch**"*. SKILL.md says the opposite. The `judge_context` guards only against auto-*dismissal* and says nothing about adoption, so the judge grades the one failure mode the test cannot produce. The rubric's `partial` clause already names the defect ("a conflicting date is explained away … this counts in **either** direction") and was not applied; the prohibition on offering extraction has no rubric clause at all.
>
> **Suite-wide instability: no.** Across 152 dimension cells × 4 runs, exactly one cell genuinely changed score, and that is a known judge over-reach already filed as issue #1403. Human annotation confirms it: 4 corrections in 591 cells, all on this test. This is not a sampling problem; it is a suite that has stopped discriminating.
>
> **Ruling B — no.** `eval/CLAUDE.md` pins `runs_per_test` to 1 and the JSON Schema enforces `maximum: 1`, so raising it needs a schema change plus a policy carve-out plus a paid re-run and full re-annotation — to average a dimension that has never once scored below 3. Apply the skill-lifecycle test: *"If the skill silently got much worse at its job, would this test still pass?"* On 2026-08-06 it did get much worse and the test passed. The goalposts are already moved.
>
> **The prose-dilution hypothesis is plausible but unfalsifiable here and owned elsewhere.** The section has *grown* since filing (the file went 47.6 KB → 50.9 KB). `make gate-skill` compares a candidate against a baseline that scores 3 on this dimension — an A/B against a ceiling can only return "no change" or a regression. Issue #1154 already carries this hypothesis and its blocker. Park it there.
>
> **Lane: `genealogist`** (relabelled). This is lane 2, a grading defect — the skill *violated* its written instruction and got marked right, which is the same class of fault as the doc's stated tie-break. Both files that need editing (`judge_context` in the test JSON, and `rubric.md`) are genealogist-owned per the skill-lifecycle ownership table. No Python is touched. `senior` dropped: the global judge prompt is the part that would be senior, and that belongs to the `cluster:judge-overrides` siblings.
>
> **Cheapest next step — bundle three grading edits into one PR:** (1) re-annotate the 2026-08-06 run's `Result triage` 3 → 2 with the two quotes above, which converts the corpus from "no evidence" into evidence; (2) strike "not a genuine mismatch" from `judge_context` and add the symmetric guard — adopting on the household anchor, warm framing, and offering extraction are each `partial`; (3) add the missing rubric clause. **The marginal run cost is zero:** this skill's latest run log is already snapshot-inactive, so the re-run is owed the moment anyone opens this issue at all — doing all three in one PR costs one run instead of two.

---

## What happened

In `eval/runlogs/unit/search-records/v1_2026-07-30_19-24-10.json`,
`ut_search_records_013` scored **Result triage = 1** (Correctness 2,
Completeness 2). The genealogist annotation agreed with every grade.

The test stages one near-match — "Patrick Flyn", 1850 Schuylkill County
census, birth year 1842 against the tree's ~1845 — with `matchScore 0.52`
and `attachedToSubject: false`. It must be flagged **needs-review**: not
dismissed on the middling score, not adopted.

The run adopted it. Verbatim from its response:

> The household head match (Thomas Flyn) is a strong corroborating anchor
> that lifts confidence above the moderate score. This is a promising
> candidate.

> This is a **strong candidate** for Patrick Flynn (I1). … The birth year
> gap (1842 vs. ~1845) is within normal census variance for a 5–8-year-old
> child, and the tree's estimate is explicitly approximate.

> **Suggested next step:** Run **record-extraction** … Want me to proceed?

## Why this is worth an issue rather than a prose edit

`search-records/SKILL.md` already prohibits all four of these moves **by
name**, in the section that governs exactly this case (lines 236–277):

| SKILL.md | The run |
|---|---|
| a birth year conflicting "by more than a year or two" is a different-person signal, "not a rounding error to explain away" | called a 3-year gap "within normal census variance" |
| "Don't reach for an excuse (imprecise census ages, indexing slips)" | "census ages for young children were notoriously imprecise" |
| "Nor as 'highly promising,' 'very likely ours,' or any other warm framing" | "strong candidate", "promising candidate" |
| "Do not offer extraction as a next step — **not even as a question**" | "Want me to proceed?" |

The doctrine is present, explicit, and roughly 50 lines long. This is a
**compliance** failure, not a missing rule, so the fix is not more prose —
adding another restatement is the exact pattern PR #929 was trimming back.

## It is intermittent, which is the actual problem

Same test, same skill files, across four runs:

| Run | Result triage |
|---|---|
| `v1_2026-07-27_23-15-55` | 3 |
| `v1_2026-07-30_14-47-24` | 3 |
| `v1_2026-07-30_19-00-36` | 3 |
| `v1_2026-07-30_19-24-10` | **1** |

The 19-00 and 19-24 runs are 24 minutes apart with identical skill files —
the only delta between them was a fixture for an unrelated test. Skill runs
are not temperature-pinned (`eval/CLAUDE.md`, "Eval vs production parity"),
so the suite passes or fails this dimension by sampling luck. A 3-in-4 pass
rate on a doctrine this heavily specified is the finding.

## Worth investigating

- Does the length of the triage section work against it? ~50 lines of
  prohibition on one pattern may dilute rather than reinforce; a shorter,
  sharper rule could comply more consistently. Measurable via
  `make skill-latency` / an A/B on that section.
- Issue #1005 (`matchConfidence` has no documented range) is a contributing
  factor — this run read "confidence 4" as a high signal offsetting the
  middling 0.52 when on a 1–10 scale it corroborates it. Closing #1005 may
  raise the pass rate but will not by itself explain the other three
  violations.
- If the variance is irreducible, `runs_per_test` exists for exactly this —
  but see the standing policy in `eval/CLAUDE.md` pinning it to 1 for now.

## Not blocking PR #929

Unrelated to that PR's changes (name/place levers); its scenario and
fixtures are untouched there, and `ut_search_records_018` went from failing
to passing in the same run.

---

# Merged in from issue #1511 (2026-08-11): the maiden-surname test has the same defect

Issue #1511 was closed into this one. Same skill, same cause, same two files, and the
same paid run — `judge_context` handing the judge its finding, so the dimension grades
something the test cannot actually fail on. Fix both in one PR.

**Adds to Touches:** eval/tests/unit/search-records/maiden-name-pre-marriage-search.json

## The second instance: `ut_search_records_015` and `_017`

Scenario `mullen-price-multiple-marriages` asks the skill to search an 1860 census for
Sarah Ann Mullen. She married in 1872, so in 1860 she appears under her birth surname.
Recognising that is the skill the test means to measure.

The skill used `Mullen`, got the right answer, and never said why. The judge scored
**Search strategy 3** and justified it with:

> The skill explicitly recognized that pli_001 targets 1860, **before Sarah first marriage in 1872**, and therefore correctly selected the maiden surname.

The skill never said that. Searching its entire `output` — `text_response` plus every
`tool_call`:

| term | occurrences in the skill output |
|---|---|
| `Mullen` | 29 |
| `1872` | **0** |
| `marriage` | **0** |

### Defect A — the prompt hands the skill the answer

The `user_message` reads:

> "Execute the 1860 census search for pli_001. Sarah has three name entries in the tree — her birth name Mullen and two married names (Price, Mielke). **Since this is 1860 and she didn t marry until 1872, search under her maiden name.**"

A skill that derived the rule and a skill that copied the instruction are
indistinguishable in this test. Restate the facts without the conclusion: state the
three name entries and the 1872 marriage as tree facts, and drop "search under her
maiden name". The skill has to reach the conclusion itself or score 2.

### Defect B — the grading instructions hand the judge the sentence to look for

`judge_context`:

> "The key signal is whether the skill recognized that pli_001 targets a date (1860) before the first marriage (1872) and therefore chose the maiden surname. Score 3 if it explicitly reasons about the name type or marriage date. **Score 2 if it uses Mullen but doesn t explain why.**"

That is where the judge got "1872" — it was told. Note the last clause: "Score 2 if it
uses Mullen but doesn t explain why" is exactly what happened; the branch was written
out in front of the judge and it took the other one. Rewrite to name the dimension
without writing the finding — closer to "score 3 only if the response states why the
birth surname was chosen for an 1860 search, in its own words".

This is the identical mechanism to `ut_search_records_013`'s `judge_context` above,
which tells the judge the excuse is true. One PR, one rubric edit, one run.

### Manufactured annotator disagreement

Same skill behaviour in two runs; one annotator read the rationale and confirmed 3, the
other checked it against the transcript and corrected to 2:

| run | Search strategy | annotator |
|---|---|---|
| `v1_2026-08-03_21-25-57` | 3 | confirmed 3 -> 3 (benteroyiembo) |
| `v1_2026-08-06_10-50-51` | 3 | corrected 3 -> **2** (florencemashipei4) |

`ut_search_records_017` shows the same pattern in both runs. The judge is pinned to
temperature 0 (`harness/judge.py::JUDGE_TEMPERATURE`), so this is not sampling noise.

### Reproduce

```sh
python3 - <<PY
import io, json, re
d = json.loads(io.open("eval/runlogs/unit/search-records/v1_2026-08-06_10-50-51.json", encoding="utf-8").read())
t = [x for x in d["tests"] if x["test_id"] == "ut_search_records_015"][0]
out = json.dumps(t["runs"][0]["output"])
print("skill output  — 1872:", len(re.findall("1872", out)), "marriage:", len(re.findall("marriage", out, re.I)))
print("judge rationale mentions both:", "1872" in json.dumps(t["outcome_summary"]))
PY
```

## Split out, deliberately

Issue #1511's third defect — the judge asserting transcript content that is not in the
transcript, and skipping a branch its own `judge_context` spelled out — is project-wide
judge-prompt behaviour, not a search-records fixture edit. It moved to issue #1403 with
the rest of the `cluster:judge-overrides` work. Do not attempt it here.

## Still open for the genealogist

Whether `SKILL.md` should require stating the name-type-versus-event-date reasoning at
all, or whether getting the right surname is enough. That is the doctrine question
underneath defect A, and it is the lead's/genealogist's to settle before the fixture is
rewritten.

### LINKED ISSUE #1330: ut_research_plan_005: judge_context accepts a freebie fact and does not require the citation in the persisted rationale
**Touches:** eval/tests/unit/research-plan/locality-survey-first-plan.json

Two static gaps in `ut_research_plan_005`'s `judge_context`. Both are checkable by
reading the test file — neither needs a reproducing run, which is what this issue
was iceboxed waiting for.

## 1. The acceptable-facts list contains a freebie

`judge_context` item 2, verbatim on `main` today:

> The plan must cite at least ONE fact from the localities entry — e.g., probate
> available 1811+ at the county Register of Wills, the Catholic Diocese of Scranton
> (post-1868), PA statewide vital registration only from 1906, or browse-only
> anthracite-region films.

"PA statewide vital registration only from 1906" is general knowledge any model
produces without opening `loc_001`. A skill that never read the entry passes item 2
by stating exactly that and nothing else — so the check cannot prove what it exists
to prove.

Fix: require a distinctive fact (county formation 1811, probate at the Register of
Wills from 1811, Diocese of Scranton 1868, browse-only anthracite films) or the
`loc_` id itself.

## 2. "Must cite" does not say where, and only one place counts

Item 2 asks that the plan cite a fact. It does not say the citation has to land in
the persisted `plan_items[].rationale`. The narration is not persisted and
`search-records` never sees it — the `rationale` field is the only thing that reaches
the next skill. The judge rationale from an earlier fix attempt names the exact
failure this permits:

> "While the response text mentions **loc_001** multiple times, the **persisted
> plan_items rationales do not**"

Fix: the check should read the persisted plan items, not the response text.

## Do not re-derive

**An unconditional "always name the `loc_` entry" edit to `SKILL.md` Step 2 was tried
and reverted** (`97fd740c`, reverted by `35db5524`). Two measured reasons it failed:

1. It **regressed `ut_research_plan_016`** from never-failed to ~50/50.
   `wilkins-first-plan` has no `localities` entry and `SKILL.md` Step 2 already says
   to stop and return to the orchestrator when one is missing, so an unconditional
   citation requirement pushes the model to the stop branch and it writes no plan.
   Scoping it to "when an entry exists" did not reliably repair 016 either.
2. Its own benefit did not survive re-measurement: 3/3 immediately after the edit,
   then 1/2 on the next pair.

So this is a **judge-context fix, not a SKILL.md fix** — do not resolve it by adding
prose telling the skill to cite the entry.

## Acceptance

- [ ] `judge_context` item 2 requires a distinctive fact or the `loc_` id, not any
      fact from a list that includes a freebie
- [ ] The check reads the persisted `plan_items[].rationale`, not the response text
- [ ] `ut_research_plan_005` and `ut_research_plan_016` both run more than three
      times after the change — this test's history includes a 3/3 that dissolved

## Files

- `eval/tests/unit/research-plan/locality-survey-first-plan.json` — the test and its `judge_context`
- `eval/fixtures/scenarios/flynn-first-plan-surveyed/research.json` — `loc_001` and its distinctive facts
- `eval/runlogs/unit/research-plan/v1_2026-08-04_22-21-46.json` — the only committed run of the current test version; it passes clean

Related: issue #1331 owns `ut_research_plan_016`'s locality-gate contradiction.

DallanQ:
**Start here:** the one-line change to `eval/app/components/forms/TestForm.tsx`. Everything else is the spec clause behind it and the two tests it would have prevented.

Closes #1007. Closes #1330.

⚠️ **Needs a paid run before merge** — `search-records` (~$9.17, the corpus's most expensive suite) and `research-plan`. Neither is run here. research-plan's latest run log is already stale, so that one is owed regardless.

## The finding

The rule against answer keys **already exists** in `unit-test-spec.md`: *"Stay neutral on contested conclusions. Don't embed an answer key,"* with a leaky/neutral comparison table under it.

The CRUD form that authors `judge_context` teaches the opposite, and its worked example **is** an answer key:

> "Background the AI judge should know when scoring this test. For example: **'A correct answer should mention the 1850 census'**"

Nobody reads the spec while filling in a form. So this fixes the tool, not the corpus — and `eval/app/**` is in neither the run-log snapshot nor `check-runlogs.yml`'s scope, so that half costs **zero paid runs**.

## Why not a lint

A keyword check flags **441 of 1,626 entries across 186 of 389 test files — 47.8% of the corpus**. The narrowest defensible shape (an explicit numeric score directive) still flags 72 entries across 9 skills, and `judge_context` is inside the run-log snapshot, so cleaning even that costs 9 paid runs plus 9 annotation passes. `eval/CLAUDE.md` already rejected a *smaller* wave than that when it kept the fixture arm of rule 2 warn-only — a gate too expensive to satisfy trains people to bypass it on exactly the edits that are not neutral.

## Why not a ban

**The harness writes verdicts into `judge_context` itself.** `_negative_judge_context` emits, for every negative test:

> "a clear, accurate decline is a full pass"
> "score Correctness and Completeness as fail (1)"

They run on an empty rubric, and without that framing the judge reads a correct decline as a failed attempt at the skill's task. A blanket prohibition would outlaw the harness's own output for 82 of 389 tests. The spec now says this explicitly so nobody re-derives it.

## The clause that was missing

When a conclusion **is** safe: only where something other than the note pins it — a deterministic validator, `expected_classifications`, or the fixture text. That is the actual difference between `record-extraction` — the one suite with zero dead dimensions, which states conclusions freely and backs each with `expected_classifications` — and a note that is simply the answer.

## Corpus fixes

Both found by reading the files, not by a run.

- **`ut_search_records_013`** told the judge the excuse was true — a middling score *"reflects surname transcription variance … **not a genuine mismatch**"* — and guarded only against auto-**dismissal**. So the judge graded the one failure mode the test cannot produce: `Result triage` scored 3 in **78 of 78** graded cells, and the run that most flagrantly violated the skill body was one of the passes. The note is now symmetric and grades the reasoning rather than the verdict.
- **`search-records/rubric.md`** gains the missing direction. Its `partial` clause already covered explaining a conflict away; it said nothing about handing a `needs-review` candidate onward to extraction, which the skill body prohibits by name.
- **`ut_search_records_015`** leaked through a channel no `judge_context` rule reaches. Its `user_message` said *"Since this is 1860 and she didn't marry until 1872, search under her maiden name"* while `judge_context` item 5 scored 3 for "reasoning about the name type or marriage date." The skill passed by echoing the prompt. The hint is gone — the 1872-06-10 marriage is in the tree fixture, so the test is now answerable and, for the first time, **failable**.
- **`ut_research_plan_005`** accepted "PA statewide vital registration only from 1906" as proof the skill read `loc_001` — general knowledge any model produces without opening the entry — and never said the citation had to land in the persisted `plan_items[].rationale`, the only field `search-records` ever sees. Both gaps fixed in one note.

## Not closed here

**#1603 is a different defect** and stays open. Its first half asks for *more* per-test context (telling the judge three sources in the Flynn fixture are real, so it stops reading a correct pass as fabrication) and its second half is a missing `max_turns`. The investigation's premise that #1007, #1330 and #1603 are one defect is wrong on the third.

## The honest objection

Fixing the note does not revive the dimension. Striking *"not a genuine mismatch"* makes `Result triage` **able** to fail, not proven to. That is what the run is for.

## Verification

Harness suite green (**2160 passed**), packaging suite green (**472 passed**), and all 46 test files in the two touched skills still validate against the unit-test schema. The `eval/app` typecheck errors in a fresh worktree are uninstalled deps, not this change — the diff there is a single string literal.


DallanQ:
Verified every claim against the repo before acting. **One is refuted and is not applied**; the rest are in as of `1d653af0`.

## Refuted — the research-plan truncation finding

The mechanism is real: `judge.py:139` `_RESPONSE_ARRAY_SAMPLE = 3`, and `orchestrator.py:1246` runs each added file entry through it, so a plan's nested `items[]` is cut to the first three. But the consequence isn't. In the last committed run (`v1_2026-08-12_18-21-29.json`), ut_research_plan_005's Correctness rationale cites loc_001 facts that live in **pli_007 and pli_008 — outside the visible window** — and scored 3. The judge read them out of the `research_append` tool-call args, which are embedded raw.

So it's a pre-existing harness defect on main that has misgraded nothing. Both proposed fixes would have made things worse: a per-test note telling the judge which block to read is exactly the scaffolding this PR exists to remove, and the `_summarize_before_state_sources` precedent doesn't port (there the array is the top-level container; here it's nested inside an opaque entry). Not filed either — flagging it to the lead as a possible `nothing-checks` item instead, since honoring it in the corpus would cost 5 paid runs across 35 tests.

## Overstated — the scenario README, on its own

Right channel, wrong severity ranking. It is judge-only (`workspace.py:85` never copies it to the skill), but on the maiden-name test it was *not* the dominant leak — `judge_context` item 1 said it harder, with a score band. Fixed both rather than either.

## Applied

- **item 5, maiden-name test** — issue #1007's Defect B, which the PR left. It named 1872 while the test has no validator and no `expected_classifications`, so the note was the only thing holding Search strategy up. Now credits only reasoning the response contains, and asks the judge to quote the sentence it's crediting.
- **item 1, same file** — required surname Mullen (mechanical, kept) but explained a married name is *"incorrect for a pre-marriage search"*, supplying the reasoning item 5 grades. Requirement stays, rationale goes.
- **item 3, near-match triage — deleted.** You're right that it sat one line below the one I fixed. Nothing else pins that verdict: no validator, no `expected_classifications` in any of the 28 search-records tests, and the rubric explicitly refuses to grade the score band. Its one underivable fact (`attachedToSubject` is false) moved into item 2. I didn't replace it with a restatement.
- **mullen-price-multiple-marriages** — both the README bullet and pli_001's rationale. Free: that scenario is referenced by search-records tests only.
- **unit-test-spec** — the leak clause now names all three sibling channels with the judge-visibility mechanism spelled out. Dropped "the one suite with no dead dimensions": `record-extraction` and `research-exhaustiveness` both have zero, and the `expected_classifications` argument never needed the superlative.
- **`test.description`** — the 1906 freebie is gone. Not judge-visible, so free, but it's what an author reads while writing the note.

## Numbers corrected in the PR body

"78 of 78" → **112 of 113** across the five currently committed logs (the single 1 is `ut_search_records_nickname_bitsie`, 2026-08-13), and 5 of 5 on each test this PR edits. Also noted that issue #1007's first acceptance step is moot — the 2026-08-06 logs were pruned on 2026-08-12 in `e48d0dbf`.

## Deferred

`flynn-record-matching` → **issue #1687**. Same judge-only leak, shared by search-records (8), person-evidence (6) and search-full-text (1), so two paid runs beyond the ones owed here. Filed under the different-eval-slot exemption, with the record inventory's "Strong on every identifier" verdicts flagged alongside the bullet you quoted.

## On the run count

Sticking with **one run**, per the lead: unit tests should be reliable enough that one suffices. Issue #1330's three-run criterion is declined rather than quietly dropped.

Runs owed: **search-records and research-plan**, both fully annotated. research-plan was already inactive on three files this PR never touched, so that one is owed regardless.

PR #1686, opened 2026-08-17
Stop overriding the judge on routing negatives; cut the sampler to two rules

DallanQ:
**Start here:** `flag_routing_negative_judge_fail` in `eval/harness/harness/orchestrator.py` (was `apply_routing_deference`), and the `rules` list in `review_sample.py`.

Net **−66 lines**. No paid run, nothing in a run-log snapshot — this is mergeable as soon as it's reviewed.

Three changes to the annotation pass and the grades it exists to check.

## 1. Delete the routing-deference floor

It floored Correctness/Completeness from 1 → 2 on a correctly-routed negative test, on the theory that the judge was grading the routed-to skill's work rather than anything the test owns.

Replaying its own guards over the 121 committed run logs and joining to the annotations says the opposite: **24 cells were floor-eligible with a judge 1, and a human confirmed that 1 on 20 of them.**

The split by evidence is clean:
- All **14** cells where the skill produced substantive output (784–14,123 chars) were human-confirmed — with rationales naming exactly the case the old docstring called a rare corner: *"instead of declining and routing to convert-dates, it performed the date conversion itself."*
- All **4** genuine overrides had `text_response == ""` and `num_turns == 0`.

So the docstring's central claim — *"There is no mechanical discriminator"* — was false **and inverted**. The floor fired hardest where it was most wrong.

Gating on empty output was the alternative. Deleting gives up almost nothing: `_compute_outcome` decides a negative-with-`correct_skill` **purely on routing**, so no score this function could change has ever moved a test's outcome.

**What survives is the warning** — renamed `routing_negative_judge_fail`, and now **registered in `_JUDGE_WARNING_KINDS`, which it never was.** It printed nowhere, while its own advisory told the reader to "check the transcript." That set's own comment says a missing kind *"prints nowhere — which is the state this whole section exists to end."* The report was the half actually worth having, and it was dark the whole time.

## 2. Cut the targeted sampler from six rules to two

Scored against the only ground truth available — a human changed the judge's score, n=37:

| | catches |
|---|---|
| no targeted rules | 13 / 37 |
| the shipped six-rule stack | 17 / 37 |
| **`_outcome_disagrees` alone** | **18 / 37** |

Deleted `_score_moved` (caught **zero** uniquely, and owned the previous-run baseline threaded through the whole call), the `not previous` rule (matched every test on a first run), `_validator_judge_conflict`, `_rationale_hedges`, and `_HEDGES`.

`_has_rubric_null_on_positive` stays **first** despite ranking lower on that metric: it is the only rule that selects the corpus's one documented rubber-stamped test, and blind grading was closed `not planned` naming this slot as the mitigation. A rubric null on a positive test is also the one shape `_compute_outcome` cannot see — it gates on 1 and 2, and null is neither.

`_newest_releasable_runlog` **stays** — it also supplies the rotation cursor, which is load-bearing. Only the `previous_tests` baseline is gone. (The original recommendation said that lookup existed only for `_score_moved`; that was wrong, and reading it is what caught it.)

## 3. One stale sentence in the spec

`unit-test-spec.md`'s "Layer 3: Human verification" still told the team to write corrected scores for *"every judge dimension of every test."* That is the sentence a junior reads to learn the job, and it describes the **pre-sampling** rule — 91 cells instead of 30.

`docs/per-pr-review-workflow.md` was already correct; only the spec had drifted.

## Verification

- **Both behaviour changes proven to fail.** Reintroducing the floor fails the no-floor test on the score — and that test asserts the **rationale** too, so a floor that rewrote the text while keeping the 1 would also be caught.
- A new test asserts the warning kind is registered, so the dark-warning bug cannot come back by omission.
- Harness suite green (**2158 passed**), packaging suite green (**472 passed**).

## What this does not do

Deleting the floor means the 4 genuinely-empty-output cells now record a judge 1 that a human must override by hand. That is the cost, and it is the right trade at 4-against-20. If those overrides become annoying, the gate-on-empty-output version is a three-line change on top of this.


DallanQ:
All four verified and fixed in `6609f0fc`. Two corrections to the review itself, and finding 4 turned out to be bigger than a number.

## 1. The spec — verified, with one correction

Two stale paragraphs, not three. Lines 703 and 705 described the floor in the present tense and named both deleted symbols. **Line 701 is *deterministic-validator* deference — a different mechanism, still live, untouched.** My PR body's "one sentence" was wrong; "three paragraphs" overcounts by one.

Both rewritten. A sweep of `docs/`, `eval/`, `packages/` and `.claude/` confirms neither `apply_routing_deference` nor `floored_routing_diagnostic_dimension` survives anywhere.

## 2. The contradicting comment — verified

It said the set holds kinds `judge._extract_dimensions` owns, that orchestrator kinds are "deliberately not tallied here", and that "a new kind added in judge.py must be added here too" — while this PR added an orchestrator-emitted kind six lines below. Rewritten to describe both sources, using the routing warning as the worked example.

## 3. The guard that could not fail — verified, and it's the best finding here

You're right, and the framing is right too: this is *why* the original went dark. The scan read only `harness/judge.py`; the kind was emitted from `orchestrator.py`, so the guard never read the line declaring it and passed green for the warning's whole life. My hardcoded one-string assertion closed this instance and left the class open — on the file this PR just turned into a source of judge warnings.

The scan now covers both files, subtracting the three harness-side advisories **by name rather than by call site**, so a *new* orchestrator kind fails until someone classifies it. That forces the decision that got skipped last time.

I deleted the hardcoded assertion instead of keeping it alongside — a guard beside one that already covers the case makes both untestable.

Proven: emitting `some_new_dark_kind` from `orchestrator.py` fails the widened scan, and the old scan passed on that same mutation.

## 4. The range — right, and the sentence around it was worse than wrong

**102–14,123 confirmed.** And 784 was the minimum of only **10** cells, so the sentence paired a count from one set with a range from another — it could never have been both.

Recomputing surfaced the real problem, which the finding didn't reach. The docstring juxtaposed "all 14 substantive cells confirmed" against "all 4 overrides had empty output and zero turns", which reads as though empty output *predicts* an override. It doesn't:

- **6 of the 20 confirmations were also empty with zero turns** (10 empty-output cells total).
- `ut_search_records_003` carries that identical signature in **all 8** of its eligible cells — confirmed in two run logs, overridden in two others.

So a floor gated on empty output would have fired on 10 cells and been **wrong on 6**. That's the opposite of what the PR implied, and it means the rejected alternative is worse than represented, not better. The floor's defect wasn't picking the wrong signal — **no signal exists**. Both the docstring and the spec now say so, and warn against reintroducing it narrowed.

The two cells at the bottom of the corrected range are the argument in miniature: `ut_translation_003` at 102 chars is a fabricated completion claim ("Saved as kirchenbuch.md") on a run that created no file, and `ut_person_evidence_003` at 603 proposed a census strategy instead of declining. Both real inline-work defects the floor was erasing.

Harness green (2157), packaging green (472).

Thanks — finding 3 in particular was worth more than the fix it prompted.

PR #1691, opened 2026-08-17
eval: blind re-grade jens-nielsen f3 (#1540)

### LINKED ISSUE #1540: Blind re-grade jens-nielsen f3 — its grade is contested and gates a claim in #1539
**Touches:** eval/runlogs/e2e/jens-nielsen/run-2026-07-22_18-37-11.ann.json

`jens-nielsen` finding **f3** is graded `partial` in the committed annotation for run `2026-07-22_18-37-11`. That grade is contested and should be re-graded **blind** by someone who has not read the discussion on #1090 / #1539.

## What to do

Run `/grade-e2e-run` against `eval/runlogs/e2e/jens-nielsen/run-2026-07-22_18-37-11` as normal. The skill reads only the fixture and the `final-*` siblings, never the run log, so the grade stays independent of the judge.

Grade **f3** on its own merits and record whatever you conclude — including `partial` again, if that is what the evidence supports.

## Why it needs a fresh pair of eyes

The finding was flagged during #1090 by people who had already seen the judge's label for it, including the person raising this issue. Nobody who has read that thread can grade it blind any more — they would be marking with the answer in front of them, and the resulting annotation would be worthless as calibration data.

The original annotator (per the committed file) is not the person filing this, so this is a second opinion rather than a self-correction.

## Why it matters beyond one row

PR #1539 changed the e2e judge to derive a relationship finding's `matched` label from its components rather than trusting the model's own label. After that change f3 became a new judge-vs-human disagreement, and #1539's body asserts this is a **second catch of the same defect** rather than a regression caused by the new rule.

That assertion is a reading, not a verified fact. A blind re-grade is what settles it:

- If the blind grade is **`false`**, the new rule was right and the old annotation was generous.
- If the blind grade is **`partial`**, the rule in #1539 is too strict and needs revisiting.

The second outcome is the one worth having, and only a grader who could actually produce it can give it.

## What this does not change

Agreement moves from 263/303 to at most 264/303 — 87% either way, comfortably above the ≥80% gate. This is not a metric fix; it is a check on whether a merged change is correct.

## Deliberately omitted

The judge's label for f3 and the reasoning advanced for it are left out of this issue on purpose, so that whoever picks it up can still grade blind. If you want that context *after* grading, it is in #1539's body under "The one new relationship disagreement".

florencemashipei:
Closes #1540

## The issue

`jens-nielsen` finding f3 (daughter Ane Cathrine Jensen → mother Dorthe Kirstine Jensen link) was graded `partial` in the committed annotation, and that grade was contested since PR #1539 changed how the e2e judge derives a relationship finding's `matched` label. #1540 asked for a blind re-grade — by someone who hadn't read #1090/#1539 — to settle whether the new judge rule is correct or too strict.

## What I did

Ran `/grade-e2e-run` against `eval/runlogs/e2e/jens-nielsen/run-2026-07-22_18-37-11`, reading only the fixture + `final-tree`/`final-research` siblings (never the run log or the existing annotation, to keep the re-grade blind).

- **f1** (first wife's identity) → `true`. Tree has `I1` (Dorthe Kjerstine/Kirstine Jensdatter) with a `Couple` relationship (`R33`) to Jens Nielsen, marriage 5 Jan 1849, Lyngs, sourced.
- **f2** (marriage date/place) → `true`. Same `R33` fact matches exactly.
- **f3** (daughter→mother link) → `partial`. Both people are correctly present in the tree — Ane Cathrine Jensen (`K1D9-KK3`, b. 29 Nov 1850, Lyngs) and Dorthe Kirstine Jensen (`I1`) — but I found **no relationship or assertion anywhere** connecting them as mother/daughter. The daughter is linked only to her father (`R9`), same as the pre-research baseline. Right people, missing link.
- `proof_quality_score: 2` (thin) — this run corroborated two independent records (marriage register + census) and resolved two conflicts, so it isn't unsound; the actual defect is the over-stated `proved` tier with five plan items left unsearched, which the scale files under `2`.

This also changes f1 from the committed `partial` to `true`, which moves the case's derived verdict from `partial` to `pass`. #1540 anticipated only f3 moving; f1 moved instead. f1 is a relationship finding whose one link (the Jens–Dorthe spouse link) is present in the tree with a source — a missing biographical detail (Dorthe's death date/birthplace) doesn't lower a relationship label.

## What this settles

Per #1540's own framing: a blind `partial` on f3 (not `false`) suggests PR #1539's stricter judge rule may need revisiting, rather than confirming the prior annotation was too generous. Full result posted to #1540: https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1540#issuecomment-5314378652 (updated to disclose the f1 move).

## Test plan

- [x] Self-checked the written `.ann.json` against `calibrate_judge.py`'s actual loader contract (`ALLOWED_ANN_KEYS`, non-null `per_finding`, keys matching the fixture's `f1`/`f2`/`f3`, valid `proof_quality_score` range) — confirmed conformant.
- [x] Confirmed this change doesn't trigger `check_runlogs.py`'s rules (unit-scoped only) or `check_e2e_fixtures.py`'s grading gate (only applies to a newly-added `run-<ts>.json`, and we're only editing an existing `.ann.json`).
- Not run: `calibrate_judge` (not even `--dry-run`) — per policy, that's the maintainer's periodic sweep, not part of a single-annotation PR.

**Note for whoever approves this:** CI cannot check whether these grades are correct — the e2e gate only checks that a *newly added* run log ships an annotation, and this PR edits an existing one. Green CI says nothing about grade correctness here; that's what the CODEOWNERS review is for.


johnmarkpeterbrown (changes requested):
Nice work on f3 — I checked the narrative and your note is right where the old one wasn't. The old note said Dorthe was "named as mother of Ane Cathrine in the proof narrative." She isn't. The word *mother* never appears in the final research file, and there's no mother-daughter link anywhere in the final tree. Your replacement is accurate and the conclusion you posted to #1540 holds.

Four things before this can merge.

**1. The f1 change moves this case's verdict from `partial` to `pass`, and the PR doesn't mention it.**

`f1` and `f2` are the required findings. A case only derives `pass` when every required finding is `true`. Flipping f1 from `partial` to `true` does that.

The f1 grade itself looks correct to me, so don't change the label. f1 is a relationship finding, its one link is the Jens–Dorthe spouse link, and that link is in the tree with a source. A missing biographical detail like Dorthe's death date doesn't lower a relationship label. I'm asking you to disclose the change, not undo it.

#1540 predicted "263/303 to at most 264/303" on the assumption that only f3 would move. f1 moved instead, and it took the case verdict with it. Add to the PR body under "What I did":

```
This also changes f1 from the committed `partial` to `true`, which moves the
case's derived verdict from `partial` to `pass`. #1540 anticipated only f3
moving; f1 moved instead.
```

Then post the same two sentences to #1540. The comment there reports only f3 and the proof-quality score, and that comment is what's being used to settle #1539.

**2. f1 has no note now, and that's where the calibration output looks for one.**

You wrote:

```json
  "notes": {
    "f3": "Both persons correctly identified and present in the tree ..."
  }
```

The deleted note was the only written reasoning for f1. f1 is now the most contestable call in the file — it reverses a committed human grade. When the calibration sweep reports a disagreement it prints the human note next to it, so an f1 disagreement would come out bare.

Change it to:

```json
  "notes": {
    "f1": "Re-graded true, up from partial. The spouse link is present in the tree (R33, Couple, KWVQ-475 to I1, marriage 5 Jan 1849 Lyngs, sourced to S1), and that link is the whole of this relationship finding. Dorthe's 4 Jun 1852 death and her birthplace are absent, but both are identifying detail rather than links, so they do not lower the label.",
    "f3": "Both persons correctly identified and present in the tree (Ane Cathrine Jensen, b. 29 Nov 1850, Lyngs; Dorthe Kirstine Jensen, m. 1849), but no relationship or assertion was ever written connecting them — the corroborating mother-daughter link itself was never made. Right people, missing link."
  }
```

**3. `proof_quality_score: 1` doesn't match the scale.**

You wrote:

```json
  "proof_quality_score": 1,
```

A `1` means no real search, no corroboration, over-claimed tier. This run read two independent records — the 1849 Lyngs marriage register and the 1850 census — and the proof summary resolves two conflicts. So the first two don't apply. What's left is the over-stated `proved` tier with five plan items unsearched, and the scale files an over-stated tier under `2`.

Change it to:

```json
  "proof_quality_score": 2,
```

If you want to hold at `1`, that's yours to call as the grader — but write the reason in the PR body. This is the largest single move in the file and the score is compared exactly.

One more thing on this. The PR body justifies the score as "consistent with this run's own proof-critique evaluation (`ev_001`, verdict `address_first`)". That evaluation is the mentor agent's grade of the same proof. Blind grading exists so your label is independent of a model's label — leaning on `ev_001` gives that up. Ground the score in the narrative instead.

**4. `annotator` was dropped.**

You wrote:

```
-  "annotator": "Israel Ayomikun Asimi"
```

The field is optional, so nothing breaks. But most committed annotations carry it, this file had it, and #1540's whole point was that a different person grades this one. Put your own name in:

```json
  "annotator": "<your team identifier>"
}
```

---

One note for whoever approves this: CI cannot check any of the above. The e2e gate only checks that a *newly added* run log ships an annotation. This PR edits an existing annotation and adds no run log, so green CI says nothing about whether the grades are right.

johnmarkpeterbrown (approved):
Look good!

mercyokum (approved):
Everything looks good now.

Issue #1484, opened 2026-08-09
Corpus report undercounts guardrail violations and hides abort-path e2e spend


DallanQ:
> **Reviewed 2026-08-10 before junior handoff.** Decision settled (the cost estimator
> may publish a figure, printed next to its measured error); the verification below was
> pinned to a rolling window that erases the defect within three days, and is corrected.

**Touches:** eval/harness/e2e/corpus_report.py, eval/harness/e2e/orchestrator.py,
eval/harness/harness/skill_invocation.py, eval/harness/e2e/pricing.py (new), Makefile,
docs/specs/e2e-test-spec.md, eval/harness/tests/unit/test_e2e_corpus_report.py

## What's wrong

Three things, all offline and all free to fix — no API calls, no live runs.

**(a) Violations are read from the run log, never recomputed.** `violations_of`
(`eval/harness/e2e/corpus_report.py:110-124`) only reads the
`guardrail_bypass_violations` field a run log already stores. Runs written before
the detector existed — or by a version of it that cannot be pinned — carry no
field, so `axes_from_runlog` maps them to `not_checked` and they contribute
**zero** violations to every arm total. That is the right call for a field-reader,
but the detectors are pure functions over data every run log already carries
(`tool_calls`) plus the committed `run-<ts>.final-tree.gedcomx.json` /
`run-<ts>.final-research.json` sidecars, so the real number is recoverable
offline.

Measured 2026-08-09 over the default 14-day window (43 of 145 runs), by calling
`check_guardrail_compliance` on each run log's own `tool_calls` + sidecars:

| | violations | runs with >=1 |
|---|---|---|
| stored — what `make e2e-corpus` prints today | 92 across 24 decidable runs | 24 |
| recomputed | **127** | **37 of 43** |

Per arm, stored -> recomputed: `same_person` 70 -> 87, `proof-conclusion` 8 -> 12,
`exhaustiveness` 7 -> 11, `conflict-resolution` 7 -> 9, and **`mentor verdict`
0 -> 8** — an entire arm the committed report shows as clean. 19 of the 43 runs
in the window are `not_checked`; every one of them is recomputable.

**(b) Every abort path records no cost.** `_fallback_usage`
(`eval/harness/e2e/orchestrator.py:968-1000`, the null is at `:996`) deliberately
returns `total_cost_usd: None` when the stream is cut before a `ResultMessage`, and its
docstring gives the reason: a run spans several models, so one price lookup would
[...]

Issue #290, opened 2026-06-07
Copy the wiki corpus to a public S3 bucket


DallanQ:
so we can access it from a public wiki-query server

PR #1692, opened 2026-08-17
Classify why wiki/pop-stats calls fail, and warn on unreachable services (#1552)

### LINKED ISSUE #1552: Classify why wiki and pop-stats calls fail — the 28%/35% headline conflates five causes
> **Lead ruling, 2026-08-11 — `malachi.taild68f1b.ts.net` is a public domain and is
> meant to be reachable from every contributor's machine.** So the per-operator split
> below is a **client-side setup defect on those machines**, not a structural
> tailnet-access gap and not a hosting problem. The finding does **not** go on issue
> #290 under this reading.

**Touches:** eval/harness/e2e/corpus_report.py, packages/engine/mcp-server/src/tools/wiki-search.ts, packages/engine/mcp-server/src/tools/place-population.ts, packages/engine/mcp-server/src/utils/place-resolver.ts

Split out of #290 (now `icebox`), which had adopted "28% of wiki calls and 35% of `place_population` calls fail" as settled fact and used it to argue for a hosted deployment. Those two numbers are correctly scoped — they count connectivity failures and nothing else — but nobody has asked **why** the calls fail, and a first classification says the answer is not one thing.

## First pass, over the committed e2e corpus

Measured 2026-08-11 over the 147 committed `eval/runlogs/e2e/*/run-*.json`:
**565 wiki + pop-stats calls, captured 2026-06-23 to 2026-08-06.** Recompute
with the target this issue adds (`make e2e-wiki-failures SINCE=all`) rather
than quoting these forward.

| Cause | Calls | A service failure? |
|---|---|---|
| Succeeded | 258 | — |
| `Could not reach wiki-query-api … (fetch failed)` / `Population data service is unavailable` | 162 | Yes |
| `No wiki page found for "<place>"` | 73 | **No** — the corpus has no page for that place |
| `{"error":"Place not found","place_id":…}` | 31 | **No** — an upstream **200** body the tool passes through unread (`place-population.ts:52` returns `response.json()` on `ok`); `place-population-tool-spec.md` § Error Handling has no row for it |
| `Wiki markdown directory is not configured` | 22 | **No** — dead path, all of them 2026-07-02 to 07-14 |
| `wiki-query-api error: <5xx>` | 10 | Yes |
| `Could not resolve "<place>" to a single FamilySearch place` | 5 | Client-side |
| Matched by none of the above | 4 | Mixed — one `No such tool available: wiki_place_page`, two `Population API error: 502 Bad Gateway`, one `wiki_read` result too large for the context window |

That last row is the point of the exercise: a classifier that folds what it
cannot match into "success" reports a lower failure rate than reality, which is
the same defect as the blended headline. The report must print an
`unclassified` count and one example per unmatched shape.

Three findings a deployment decision does not address:

**1. The unreachable failures track the operator, not the calendar.** Measured
2026-08-11 over the same 147 run logs, attributing each run to the git author
who first committed it (a proxy for the machine that ran it — the run log
records no host): of the 24 contributors whose runs called these tools, 4 with
8+ calls each failed **80%+** of them and 11 with 8+ calls each failed **none**.
On 2026-07-30 one contributor logged 31 unreachable calls and 1 success while
four others logged 25 successes and zero failures the same day, in runs that
interleave with it (14:32 fail, 18:09 ok, 19:34 ok, 19:43 ok, 23:05 fail). And
all 162 unreachable errors are connect failures (`fetch failed`) — never the
30s abort `fetchWithTimeout` raises for a host that is up but slow. An outage
story would need three on/off transitions in one day, each matching one
person's runs. **Do not report this as downtime.** Report it as a per-operator
reachability split, and say in the output that the attribution is by commit
author, not by machine.

What it costs today: **28 of the 111 runs** that used these tools got **zero**
successful wiki or pop-stats calls — a locality answer produced with no wiki
input at all, at single-digit dollars and roughly an hour per run
(`docs/architecture.md` § 9.1).

**2. 104 of the 307 non-success responses are the service correctly reporting it has no data** — 73 wiki coverage gaps and 31 places with no population series. Those are a *corpus* question, not an *uptime* question, and folding them into a failure rate hides both. Worth knowing separately: which places do researchers ask about that the wiki corpus does not cover?

**3. 22 calls hit a code path that no longer exists.** `wikiMarkdownDir` returns zero hits in `packages/engine/mcp-server/src/` today, and all 22 are dated 2026-07-02 to 07-14 — before the page tools moved from a local markdown directory to the networked API. Any rate computed over the whole corpus carries this historical artifact. It is recorded as removed at `docs/specs/wiki-page-tool-spec.md:127`; live code cannot produce those 22 responses.

## Ask

Build one offline report over the committed run logs. No live run, no model, no
API spend — same posture as `nudge_report.py`, which is the file to copy.

**Sites this touches (all of them):**

- [ ] `eval/harness/e2e/wiki_failure_report.py` — new module. Import
      `add_since_arg` / `all_result_jsons` / `describe_window` / `filter_since` /
      `result_jsons_for` from `e2e.runlog_selection` so it windows like every
      other reader.
- [ ] `eval/harness/tests/unit/test_wiki_failure_report.py` — the instrument.
      Cover one synthetic payload per bucket **in both `response_summary`
      shapes** (below), plus the unclassified path.
- [ ] `Makefile` — a `e2e-wiki-failures` target next to `e2e-nudges`, same
      `TEST=` / `SINCE=` options and the same "pure analysis, no API" comment.
- [ ] `eval/CLAUDE.md` — add the target to the "Aggregating reports FILTER"
      list, so the next reader finds it. (`e2e-nudges` was never added there;
      don't repeat that.)
- [ ] If PR #1437 (Windows `.bat` wrappers, one per make target) has merged by
      the time you start, add `scripts/e2e-wiki-failures.bat` too; if it has
      not, say in your PR that the wrapper is owed.

**Buckets, fixed here so they are not re-derived:** `success`, `unreachable`,
`upstream_5xx`, `no_wiki_page`, `no_population_series`, `unresolvable_place`,
`legacy_markdown_dir`, `other` (with examples), `unclassified`. Report each
split three ways: by cause, by day, and by run-log author.

**Do not touch tool code, tool specs, or error payloads in this PR.** Whether a
coverage gap should reach the agent differently from a service failure is a
change to two tool contracts (`wiki-page-tool-spec.md` § Error handling
deliberately specifies `No wiki page found` as a thrown error today) and is a
separate, senior-reviewed issue that this report is the input to. Also do not
edit `corpus_report.py` — issue #1484 is queued against that file; add a
module beside it.

## Also in this PR: stop losing runs to it

`make e2e-preflight` spends its whole existence sparing a contributor 20–60
minutes and $3–10 (`e2e/preflight.py` docstring), and it does not check this.
Add a sixth check next to the five in `CHECKS`, **WARN, never FAIL** — a `GET`
against `getWikiApiUrl()`'s base plus the Pop Stats base, short timeout, and on
failure this text:

    The wiki and population services are not reachable from this machine, so
    wiki_search / wiki_read / wiki_place_page / place_population will fail for
    the whole run and the locality half of the answer will be empty. This is
    expected to work — report it before spending an hour on this run.

Cover it in `eval/harness/tests/unit/test_e2e_preflight.py`. Then ask the four
contributors the report names to run `make e2e-preflight` and paste the result;
anything still failing after that is a new bug issue with their output
attached, not a line in this report.

Record the rule (not the reasoning) in `docs/e2e-testing-guide.md` next to the
other setup gates: **run `make e2e-preflight` before every session; an
unreachable wiki service means the locality half of the run will be empty.**
The reasoning goes where `docs/specs/e2e-test-spec.md` already explains the
detector for a genealogy MCP server that never connected — same failure shape
(the agent improvises and declares success), which is why this is a warning
rather than a new gate: gating it was considered and set aside because the
report has to exist before anyone knows how often it would fire.

## Method note

Three things that will silently corrupt the counts:

- **`is_error` is not usable.** It exists on only **3 of 147** committed run
  logs (`e2e/result.py`, the `HARNESS_SCHEMA_VERSION = 3` note). Matching
  `response_summary` text is the only option, which is why the unclassified
  count is load-bearing.
- **`response_summary` has two shapes.** Under 4000 chars the raw MCP envelope
  is passed through with the tool's document as an *escaped* string; over it,
  the document is unwrapped and its keys are real JSON keys
  (`e2e/orchestrator.py::_summarize_tool_response`). Match bare substrings —
  a quoted-key grep undercounts.
- **The default 14-day window hides two of the buckets.** All 22
  `legacy_markdown_dir` calls are 2026-07-02 to 07-14. Run and report
  `SINCE=all`; state the window in the output like the other readers.

`wikiMarkdownDir` needs no investigation: it is recorded as **removed** at
`docs/specs/wiki-page-tool-spec.md:127` and `git grep wikiMarkdownDir` finds it
in no engine source. Live code cannot produce those 22 responses.

jud-sdev:
## Summary

Tier: Normal. Closes #1552.

- New offline report `e2e.wiki_failure_report` classifies every wiki/pop-stats call whose captures survive into the fixed nine-bucket taxonomy, split by cause, by day, and by the run log's committing author. The blended "28%/35% fail" headline separates into service reach, the corpus correctly reporting it has no data for a place, and client-side causes.
- Adds a WARN-only sixth preflight check that probes the wiki-query-api and Pop Stats bases, so a machine that cannot reach them is caught before a run rather than after an hour of empty locality answers.
- Aligned with main's e2e capture strip: the strip drops `response_summary` past 14 days, so the report's useful horizon is the 14-day capture window. Runs older than that are reported as a separate stripped-and-unclassifiable count, never misfiled as `unclassified`.
- Removes the stale "28% of wiki-tool calls and 35% of place_population calls fail" line from `CLAUDE.md` (lead request): it is the exact point-in-time number this PR reclassifies into five causes, it cannot be recomputed since the capture strip, and a failure stat does not belong in the agent operating manual.

## Review guidance

**Start here:** two things.

1. `classify(tool, response_summary)` — the matchers are BARE substrings, because `response_summary` has two shapes (the escaped MCP envelope under ~4000 chars, the unwrapped document over it). A quoted-key match undercounts, and I hit this for real: the first cut dropped 42 `Place not found` responses into `unclassified`. The `place_population` success-versus-no-series split is the one tool-aware branch.
2. `scan()` — a run older than 14 days has been through main's capture strip (`captures_stripped: true`), which removes `response_summary`. Those calls carry no cause text, so they are tallied as `stripped` and kept out of the classifier, not run through it where a missing summary would inflate `unclassified`. This is why `SINCE=all` over today's corpus reports 508 stripped calls with 0 unclassified, rather than 508 bogus unclassified.

**Unsure about:** the `other` bucket holds a few benchmark artifacts — a too-large wiki result, a `wiki_search` a fixture deliberately disabled, a tool the session never loaded. I kept them in `other` with printed examples rather than dropping them from the denominator. If you would rather they not count at all, say so and I will exclude them.

## Plan

**Didn't change:** any tool code, tool spec, or error payload — the issue scoped that out, and this report is its input. Did not touch `corpus_report.py` (#1484 is queued against it).

**Acceptance check:** in `tests/unit/test_wiki_failure_report.py`: `test_each_service_failure_bucket_in_both_shapes` and `test_place_not_found_survives_the_escaped_envelope` (I proved by mutation that reverting a matcher to a quoted key drops the escaped shape into `unclassified` and the test fails), plus `test_scan_counts_a_stripped_run_as_stripped_not_unclassified` for the strip alignment. On the preflight side, `test_wiki_pop_never_fails_whatever_the_probe_says` pins the WARN-never-FAIL contract.

**Deviated from the plan:** the retrospective scope shrank after merging main. Main's capture strip had already removed `response_summary` from 103 committed run logs, so the whole-corpus classification the issue described is no longer reproducible from disk (the one-time numbers stand in the issue body, measured 2026-08-11 pre-strip). The report is now a rolling-14-day tool that classifies fresh runs fully and reports older runs as stripped. Confirmed this direction with the task owner before adapting.

## Test plan

- [x] Ran the harness unit suite (`uv run python -m pytest tests/unit`): passes apart from one environmental failure (`test_mock_mcp` needs `node`, absent in this WSL shell) unrelated to this change. My changes are all Python harness, Makefile, and docs, none of which `make test-all` (test-js + server-test) covers, so I lean on CI for the JS/server suites this PR does not touch. Ran `ruff` on the changed files: clean.
- [x] Ran the report over the corpus (`make e2e-wiki-failures` and `SINCE=all`) and the scoped `--test` path. Ran the real preflight probe: both bases answered (reachable), confirming the 404-is-reachable path.
- No skill run-log snapshots changed; no skill `description`/DO-NOT changes.

## Follow-on issues

**Folded in / filed instead:** No new issue filed. I did not add a Windows `.bat` wrapper: PR #1437 (which the issue referenced) added per-target e2e wrappers, but #1614 has since removed them and the drift guard `windows-wrappers.test.ts` forbids a stray wrapper in `scripts/`, so e2e report targets correctly get none.

On the root cause: the deploy fix is #290, which the lead **iceboxed on 2026-08-11 already citing this issue's findings** — the thaw trigger is a real deployment URL existing, and the hosted override knob (`wiki_api_url`/`pop_stats_url` through `hosted_config()`) already ships; nothing sets it yet. This report's per-operator split *reinforces* that client-side ruling (specific machines are unreachable, not the service), so it does not reopen the icebox. The gap the lead flagged as still unaddressed — a failed or empty call silently thinning the locality guide with no marker — is partially covered here on the eval-harness side by the new preflight warning; the runtime-marker half (desktop `.mcpb` plus an in-answer marker) is separate and unfiled.


T-FEH (changes requested):
The report itself is good work — the nine buckets are the issue's own, the bare-substring reasoning for the two `response_summary` shapes is right and non-obvious, and the stripped-vs-unclassified split is the correct call. I confirmed the headline claim by running it: `SINCE=all` gives 133 classifiable, **508 stripped across 103 runs, 0 unclassified**, and a false-positive scan over the corpus finds none. Harness suite 2252 passed, 2 skipped; ruff clean; the encoding lint covers both new files.

I pushed three fixes (`dad11451`). Two are defects; the third is a judgement call I have left half-open for you.

**1. The acceptance check you named does not catch the mutation you describe.** The body says reverting a matcher to a quoted key "drops the escaped shape into `unclassified` and the test fails". I ran exactly that — `("Place not found",)` → `'"error":"Place not found"'` — and got **12/12 still green**. The cause is `wiki_failure_report.py:172`: the `place_population` fallback (`if "place" in rs`) returns `no_population_series` anyway, so the assertion passes without the matcher. Added the `wiki_place_page` arm, which takes no such fallback — under the same mutation it returns `unclassified`, so the test now fails (1 failed under mutation, 12 passed restored). This is the fifth test in this batch of PRs that read as proof and wasn't; the tell each time is an assertion whose outcome doesn't depend on the code it names.

**2. "stripped and unrecoverable" is false, and it is load-bearing.** `b065b687` rewrote files in the working tree; git history keeps the originals. `git show fe630012:eval/runlogs/e2e/anders-monsen-ancestry/run-2026-07-09_18-51-04.json` yields **111** `response_summary` captures where the working tree has **0**. That matters because the claim is the stated justification both for dropping the issue's whole-corpus classification and for the `CLAUDE.md` deletion. Reworded to say the report reads the working tree *by choice* — which is true, is a perfectly good reason, and doesn't rest on an availability limit that isn't there.

**3. The `CLAUDE.md` deletion — restored the consequence, left the figure to you.** Removing "28% / 35%" is defensible and I can support it from the repo's own precedent: `measured-figures.test.ts` exists because point-in-time figures in model-read surfaces go stale and made "six documents disagree with each other and with the artifact". It also breaks nothing — that test is scoped to record-search surfaces and deliberately excludes bare percentages.

But the same edit removed *"The agent gets an actionable error and then quietly ships a thinner answer, so the user sees nothing."* That is a behavioural fact, not a measurement; it doesn't go stale, and the PR itself says the runtime-marker half is still unfiled — so it is not merely still true, it is still unaddressed. Without it the paragraph says the defaults point at one developer's tailnet and never says the failure is silent, and the surviving "that was true only in the sense that they cannot" loses its antecedent. I restored it, pointing at `make e2e-wiki-failures` for the rate rather than restating a number.

**What I am not doing is signing off the figure removal itself.** The body calls it a "lead request", and I can find no record: issue #1552 has **0 comments**, its task list names `eval/CLAUDE.md` and not the root file, and its own body says those two numbers "are correctly scoped — they count connectivity failures and nothing else", which is the opposite of stale. Deleting a measured claim from the agent operating manual is a lead call. Get @DallanQ to confirm on the issue and this is done — I have no objection to the outcome, only to it landing on an unrecorded instruction.

Non-blocking:

- `upstream_5xx` matches any status, because `place-population.ts:48` interpolates `response.status` — a 404 lands in the 5xx bucket. Worth narrowing the needle or renaming the bucket.
- `wiki_failure_report.py:38` and the test docstring say "42 of ~35", which cannot both be right.
- The PR removes the 28%/35% figure as stale while seeding it into three new places (`preflight.py:19`, `wiki_failure_report.py:3`, `Makefile:602`). If the reason is that such figures rot, these will too.
- `_probe_url`'s `timeout=5.0` doesn't bound `getaddrinfo`, so a blackholed resolver stalls past it. Harmless today since preflight is a standalone target, but the repo's convention is an explicit whole-transfer budget.

T-FEH (approved):
Clearing my changes-request. The one thing I was holding on is now on the record, and all three non-blocking notes are fixed.

**The blocker is resolved.** @DallanQ confirmed on #1552: *"I want to expand this issue to include removing the '28% of wiki calls / 35% of place_population calls fail' line out of the root CLAUDE.md."* That is exactly what was missing — I never objected to the outcome, only to a measured claim leaving the operating manual on an instruction with no record, against an issue body that called those numbers "correctly scoped". @jud-sdev, thanks for putting the question rather than arguing the point; that was the fastest route through it.

**All three non-blocking notes addressed, and I checked the behaviour rather than the comments:**

- **`upstream_5xx` no longer swallows a 404.** The needle now pins the leading `5`. Ran it: `Population API error: 404` → `unclassified`, `503` → `upstream_5xx`, `wiki-query-api error: 500` → `upstream_5xx`, `404` → `unclassified`. `unclassified` is the right home for a 404 — it is reserved for a live shape the matchers do not know, so it surfaces in the report instead of hiding inside a bucket that misnames it.
- **"42 of ~35" is gone** — now "42 real `Place not found` responses", which is what the sentence was trying to say.
- **The figure is not reseeded.** A repo-wide grep for `28%` / `35% of` across `CLAUDE.md`, `eval/harness/e2e/*.py` and the `Makefile` comes back empty. The docstring and the make target now say "the blended failure-rate headline … run this report for the current split", which is the durable form — it points at the instrument instead of restating a number that rots.

**I resolved the merge with main and pushed it (`d4625818`).** Two conflicts, both from my own `feec87ed` landing via #1740:

- `Makefile` — your `e2e-wiki-failures` and main's `e2e-detector-diff` are separate targets added at the same spot. Kept both.
- `eval/CLAUDE.md` — your parenthetical about `SINCE=all`'s real horizon, and main's `make judge-report` in the FLAG list. Kept both. The `Component-derivation drift` row from #1740 survives intact.

Verified: harness **2375 passed, 2 skipped**; ruff clean; all 12 checks green.

Everything from my first pass stands as fixed — including the two I pushed myself (`dad11451`): the acceptance check that did not catch its own mutation, and the "stripped and unrecoverable" claim that was load-bearing for two scope decisions and false.

Good report. The taxonomy is the right shape and the bare-substring reasoning for the two `response_summary` envelopes is genuinely non-obvious.

florencemashipei (changes requested):
Requesting one change. Everything else here checks out, and the report does what the issue asked — I ran it over the corpus and the per-operator split is right there: one author 6 of 6 unreachable while four others sit at 0%.

**`resolve_authors` never batches. Every run log still gets its own git process.**

`dict.setdefault(key, default)` evaluates the default before it checks the key. So `commit_author(p)` runs for every path, including the ones the batch had already resolved.

You wrote:

```python
    for p in paths:
        authors.setdefault(p, commit_author(p))
```

Change it to:

```python
    for p in paths:
        if p not in authors:
            authors[p] = commit_author(p)
```

I measured this in your branch's worktree. 157 of 157 paths go through the per-file resolver, and `resolve_authors` takes 71.9s over the committed corpus. The batched `git log` on its own takes 0.13s. With the change above it is 0.19s and returns the same authors, none unknown. The docstring's "~60s over the committed corpus" is the cost the code still pays, not the cost it avoids.

Please add a test alongside the fix. Nothing covers `resolve_authors` or `commit_author` today — a grep across `eval/harness/` finds no hit outside the module — which is why this could ship looking like an optimization. All 13 tests pass with the bug and without it. A fake `commit_author` that raises, plus an assertion that `resolve_authors` returns without calling it when the batch resolved every path, would pin it.

Two non-blocking notes. Neither needs to hold the merge:

- `CLAUDE.md` now says the answer is silently thinned "when one of those hosts is unreachable". Your own report puts unreachable at 12% of calls and "corpus has no data for the place" at 16%, and that second cause thins the answer just as silently. The sentence you replaced covered both. "when one of those calls fails" would keep that.
- `test_resolve_base_falls_back_to_the_ts_default` asserts the default ends with `/wiki`, but its docstring says the point is that a rotation of that constant is followed without editing the file. Those two cannot both hold.

For the record on the rest. Harness suite 2374 passed, 3 skipped in your worktree; ruff clean. I broke each of the three acceptance checks you named and watched it fail. I checked Dallan's go-ahead for the CLAUDE.md figure on #1552 myself rather than take it second-hand. And I ran the preflight probe live from a Windows machine — both bases answered 404, so the "any HTTP answer means reachable" path holds outside WSL too.

florencemashipei (approved):
Clearing my changes-request. The fix is right and it is pinned.

I checked the behaviour rather than the diff. Reverting `d4946c1e`'s change back to `setdefault` makes your new test fail with `commit_author spawned for ...; the batch resolved it`, and restoring it passes — so the test bites on the exact mutation it is written for, which is the thing that was missing before. `resolve_authors` now runs in 0.32s over the 157 committed runs against 71.9s before, same authors, none unknown. The report output is unchanged: 81 calls in the 14-day window, 75% success, 0 unclassified.

Both non-blocking notes are done. `CLAUDE.md` now says "when one of those calls fails", which covers the corpus-has-no-data half that the report itself puts at 16%. And `test_resolve_base_falls_back_to_the_ts_default` now reads the constant out of the TS and asserts the resolver returns what it declares, so it means what its docstring says.

Verified on the updated head 857e8d2a, not just on your push: harness 2375 passed, 3 skipped; ruff clean; all required checks green. I ran the same suite against the merge with main before updating the branch and it was already clean, so the update brought in nothing that touches this.

Nice work on the taxonomy. The by-author split earns its place — one operator at 6 of 6 unreachable while four others sit at zero is exactly the shape the blended headline was hiding.

PR #1693, opened 2026-08-17
volume_search spec: record-type group filtering and structured coverage dates

Gennecis:
## Summary

Specifies record-type filtering for `volume_search`, plus structured dates on each
coverage. The RMS endpoint already accepts `coverage.recordTypeConceptIds` —
hierarchy containment, OR across the array — which the tool does not send today, so
an agent must fetch every volume for a place and scan. For Harjager härad that is
110 groups and 1889 coverage rows to find 4 tax volumes.

The spec adds an optional `recordTypeGroups` input over a 47-group vocabulary
declared as a schema enum, and adds `recordTypeConceptId`, `startYear` and
`endYear` to each coverage.

Spec only — no tool code. `dev/probe-record-type-groups.ts` ships with it so every
figure in the section can be re-derived.

## Review guidance

**Start at `## Record-type filtering`.** The reviewable artifact is the **group
table** — 47 rows, one anchor and one parent each — and the **strays table**
beneath it, listing ids that belong to a group editorially but fall outside its
anchor's subtree.

Run any section of the probe from `packages/engine/mcp-server/`:

```sh
npx tsx dev/probe-record-type-groups.ts reach
```

Sections: `filter | containment | or | roots | reach | anchors | tree`.

Also worth reading: `### Relationship to the other record-type filters`, which
covers how this vocabulary sits against the three `recordType` inputs that already
exist on `record_search`, `fulltext_search` and `research.json`.

**Seven open questions are recorded in the spec**, not here — three of them change
the vocabulary's membership, so the enum literals will move once they are settled.

## Test plan

- `npm run typecheck` — clean, and its project covers `dev/`
- `doc-links` — 130/130
- `make engine-test` — 96 files, 2335 tests
- Every code site the spec cites re-verified against current `main`:
  `RECORD_TYPE_TO_INT`, `record_type_recommended`, `FSSearchMetadata.startYear`,
  `earliestYear`
- No tool code changes, so there is no behaviour to exercise. No skill prose
  describes these output fields, so no eval run is triggered

## Follow-on issues

None filed. The seven open questions live in the spec so both reviewers see them in
context; the decisions there become the implementation PR's scope.


chesworthrm:
## Review — mechanism approved; here are decisions on all seven open questions

The mechanism half of this spec is solid and well-evidenced, and the probe-with-controls
discipline (only a moving `totalCount` counts) is the right standard for an undocumented
endpoint. Below are decisions on all seven open questions so the implementation PR is
unblocked, plus three factual fixes this spec needs first.

Verification method: I re-ran the ancestor chains and reach deltas against the live RMS
endpoint using this PR's own query helpers, in a throwaway script (since deleted). Every
number below is measured, not inferred.

---

### Decisions

**Q1 — strays: keep them.** The mechanism cost is zero (same `recordTypeConceptIds`
array, one more entry), and the alternative silently loses real volumes — 208,413 for
Baptism, 50,432 for Prison. A researcher who asks for death records and doesn't get
obituaries has a wrong answer; one who sees an obituary volume under both Death and
Newspapers has a duplicate, which is harmless. Please add one sentence to the spec stating
that groups may overlap, so the consequence is documented rather than discovered.

**Q2 — drop group 14 (Casualties, War, POW).** Your chains are correct; I confirmed all
four live: `123352` casualty records (17,090 volumes) and `124372` POW files (12,530)
under Military `124133`; `124129` military death (5,527) under Vital › Death; `124445` war
grave records (128) under Vital › Death › `123358`. The measurement that decides it:
**all four are already reachable** — two via the existing **Military** group, two via
**Death**. A Casualties group adds no reach at all, and it would be the only row breaking
the one-row-one-parent invariant that makes the nesting explainable. Revisit only if the
genealogists report researchers asking for "war casualties" as a category; the strays
mechanism already sends multi-id arrays, so it stays cheap.

**Q3 — keep Emigration.** Narrowing is the vocabulary's job: Probate sits inside Court
inside Legal and nobody proposes deleting Probate. Emigration is 23,139 volumes globally.
**But the `131602` claim is wrong** — see fixes below.

**Q4 — keep the taxonomy's placement, annotate it, do not flatten.** The parent has almost
nothing to do with discoverability: `recordTypeGroups` is a flat enum, so "Enslavement" is
a value an agent picks directly regardless of what it nests under — nobody has to "open
Land & Property" to reach it. The parent column controls exactly one thing: what a **Land**
search returns, and that is a fact about FamilySearch's data, not a choice available to us.
Selecting Land *will* return enslavement volumes whether or not the table admits it, so
flattening would make the table lie about containment — the one thing it exists to tell the
truth about. The remaining call is the wording of the annotation, and that is the one item
of the seven I'd genuinely route to the genealogy reviewers.

**Q5 — correction confirmed.** Voting `127015`, Medical `127076`, Photographs `122956` and
Administrative `135784` are ordinary root rows needing no special treatment. Bookkeeping,
not a decision.

**Q6 — anchor Newspapers on the root `124231`.** This is the highest-value question of the
seven and the spec undersells it. Measured globally: `119166` → **176,286** volumes,
`124231` → **250,945**. Anchoring on the child leaves roughly **75,000 newspaper volumes,
~30% of them, unreachable by any group name worldwide** — much larger than the "3.5% of
New South Wales" framing suggests. I replicated the reach shape without the date window:
New South Wales missed 138 volumes and adding `124231` recovered 137; Oslo missed 491 and
adding `124231` recovered all 491. The only objection — that `124231` is unlabelled
upstream (`concept-id:124231`) — is already answered by this spec's own *"Names are not
stable; concept ids are"* section: we supply the English names ourselves precisely because
six of the thirteen observed roots carry no label. Six unnamed roots are already in the
table; a seventh is not a new problem.

**Q7 — don't route this through the schema. The premise is wrong twice over.**

1. The spec asserts twice that "no tool in this server sets `additionalProperties: false`".
   **`rank_search_matches` already does** — `packages/engine/mcp-server/src/tools/rank-search-matches.ts:576`,
   on its advertised input schema, in `allToolSchemas`, since 2026-07-07 (#597). So
   "scoping it to one tool is an inconsistency of its own" is moot: the inconsistency
   shipped six weeks ago, the precedent exists, and the one-tool-vs-breaking-every-tool
   dichotomy is false.
2. More importantly, **nothing in this server validates input against the JSON Schema** —
   no `ajv`, no `zod`; `src/index.ts` advertises `allToolSchemas` and dispatches raw args
   straight to the handler, and `volume_search`'s `validate()` inspects only fields it
   knows about. So `additionalProperties: false` yields "a clear error" *only if the client
   validates before dispatch* — and that's three clients (Claude Code, Cowork on-computer,
   Cowork in the cloud), untested in all three. That is exactly the shape of a guard that
   reads as coverage and isn't.

Do it server-side instead, in `volume_search`'s `validate()`:

```ts
if ("recordType" in input) throw new Error(
  "volume_search filters by recordTypeGroups, not recordType. Valid groups: …"
);
```

Client-independent, not a contract change, gives the exact error the question wants, and
ships inside the implementation PR. Keep the schema-wide `additionalProperties` discussion
as separate work if still wanted; it should not block this.

---

### Three fixes needed in this PR

**1. `131602` does not have zero records.** The spec says it has "**zero records** across
the whole corpus and cannot be verified." Live it returns **3 volumes**, and the more
useful fact is its chain: `[127023]` — a **sibling** of Emigration `123632` under Migration,
not a descendant. So Migration reaches it and Emigration would not. Either add it as a
stray on Emigration (free and correct) or drop it explicitly, but the "zero records" line
has to go.

**2. Record the second-anchor verification — it's good news that currently reads as a
risk.** The source list gives **Probate** and **Religious** a second anchor each
(`126785`, `124209`) and the shipped vocabulary keeps only one. Given this spec's own
warning that carrying only the first anchor cost Baptism 208,413 volumes, a reader will
reasonably fear a repeat. I checked: `126785` resolves to `[122797, 127010, 124277]` —
inside the Probate anchor — and `124209` ("Parish") resolves to `[123402]` — inside the
Religious anchor. Containment reaches both; nothing is lost. Please state that, so the
next reader doesn't have to re-derive it.

**3. The probe can't reproduce two of the questions.** The PR body says every figure can be
re-derived, but the Casualties ids (Q2) and `131602` (Q3) appear nowhere in
`GROUP_CLAIMS`, so no section reaches them — I had to write a one-off script. Add a
`["14 Casualties", …]` row and `131602` so those chains are reproducible from the shipped
probe.

Also, minor: **"the source list" is cited three times and never identified or linked**, yet
questions 1, 2 and 5 rest on it — including group numbers ("Group 14", "Group 16") that
appear nowhere else in the spec. A reviewer asked to adjudicate editorial intent should be
able to see the document expressing it.

---

### Net

Fold the three fixes in, apply the Q1–Q3 and Q5–Q7 decisions to the group table and the
open-questions section (Q6 changes the Newspapers anchor; Q2 and Q3 settle membership at 47
groups), and this is ready. Q4 goes to the genealogy reviewers for annotation wording only
— the placement is decided. The `recordType` rejection guard from Q7 becomes implementation-PR
scope.

chesworthrm (changes requested):
## Review — one thing has to change before this ships, plus three corrections

The mechanism half is in good shape, and the three fixes from the last round all
landed: `131602` is a stray with its real chain, the second-anchor verification for
`126785`/`124209` is now stated, and `GROUP_CLAIMS` picked up the Casualties row and
`131602` so both are reproducible.

I checked the group table against the probe's `VOCABULARY` mechanically. They agree
exactly — every name, anchor, parent and stray, 47 groups and 20 strays, and the
"Also returns" column is a correct closure of the Parent column. That part is solid.

---

### 1. A parent group returns less than its own child. This one blocks.

The group table promises that picking a group also gets you every group nested
beneath it. The expansion rule breaks that promise.

It sends a group's **own** strays and nothing else. But a stray is by definition an
id that containment cannot reach. So a child's strays are invisible from the parent.

Pick **Death** and you get obituaries, because obituaries are one of Death's strays.
Pick **Vital**, which the table says includes Death, and you don't. Same for Legal
vs Court, Government vs Prison, Tax, Passports and Pensions, and ID documents vs
Passports. On today's data the Vital case alone is about 116,000 volumes.

This is the failure the spec warns about two pages earlier — the one that costs
Baptism 208,413 volumes with no error and no empty result to notice.

It also reaches the model. The tool description tells it that `'Government'` also
returns Tax, Prison, Poor Law and Passports. That is not true for those groups'
strays, so an agent will under-search and never know.

```
You wrote:      group table to its **anchor plus any strays**, union the ids, and set

Change it to:   group table to its **anchor, plus the strays of that group and of every
                group nested beneath it**, union the ids, and set
```

Test 15d passes either way, because it only exercises containment. Please widen it:

```
You wrote:      | 15d | A parent group's request ids retrieve its nested groups' volumes (containment), and two groups OR rather than intersect | Group semantics |

Change it to:   | 15d | A parent group's request ids retrieve its nested groups' volumes (containment), send the strays of every group nested beneath it, and two groups OR rather than intersect | Group semantics |
```

### 2. Four ids the spec makes ancestry claims about are in the wrong slot in the probe

`sectionAnchors` resolves the third argument, not the second. These four sit in the
second, so no section resolves any of them.

That matters most for `127575` and `131448`. The strays table says they were
verified with section `anchors`, and the 208,413 and 50,432 figures are what the
"keep the strays" decision rests on. Neither number can be re-derived from the
shipped probe today. Counts on this endpoint move, so a figure nobody can re-run is
a figure nobody can check.

I resolved all four by hand. Every chain is exactly what the spec says it is — this
is about reproducibility, not accuracy.

```
You wrote:      ["2 Baptism", [103612, 127575], [114490]],
Change it to:   ["2 Baptism", [103612], [114490, 127575]],

You wrote:      ["23 Probate", [124277, 126785], [123648, 123196, 129461, 130661]],
Change it to:   ["23 Probate", [124277], [123648, 123196, 129461, 130661, 126785]],

You wrote:      ["32 Religious", [123402, 124209], [130206, 126601]],
Change it to:   ["32 Religious", [123402], [130206, 126601, 124209]],

You wrote:      ["42 Prison", [123478, 131448], [126780, 130086, 126416, 129538]],
Change it to:   ["42 Prison", [123478], [126780, 130086, 126416, 129538, 131448]],
```

### 3. `research_append` does not set `additionalProperties: false` on its input schema

`rank_search_matches` does, and the argument works on that alone. `research_append`
sets it on `research.schema.json`, which is enforced server-side by
`validate_research_schema`. That is the opposite of the client-side-only path the
sentence is making a point about, so citing it undercuts the point.

```
You wrote:      clients and untested in all three. (`rank_search_matches` and `research_append`
                do set it; it constrains what those tools accept in practice only through the
                same client-side path.)

Change it to:   clients and untested in all three. (`rank_search_matches` does set it, on its
                input schema, and it constrains what that tool accepts only through the same
                client-side path. `research_append` does not: its `additionalProperties: false`
                sits on `research.schema.json` and is enforced server-side by
                `validate_research_schema`, a different mechanism.)
```

The conclusion — reject in `validate()` — is right. Only the citation needs fixing.

### 4. The output example still shows the old three-field coverage

The spec adds three fields and says the year pair is always emitted when present,
which is nearly always. The example is what an implementer copies. I re-fetched that
exact group, so these are its real values.

```
You wrote:            "place": "Edensor, Derbyshire, England, United Kingdom",
                      "dateRange": "1726–1812",
                      "recordType": "Burial Records"

Change it to:         "place": "Edensor, Derbyshire, England, United Kingdom",
                      "dateRange": "1726–1812",
                      "recordType": "Burial Records",
                      "recordTypeConceptId": 104497,
                      "startYear": 1726,
                      "endYear": 1812
```

---

Fixes 2–4 are one-liners. Fix 1 changes what the implementation PR builds, which is
why I'm requesting changes rather than approving.

chesworthrm (changes requested):
## Review round 2 — the four fixes landed, two claims are wrong

All four items from the last round are in: the expansion rule and test 15d now
require every nested group's strays, the four ids moved into the slot
`sectionAnchors` actually resolves, the `additionalProperties` citation is
corrected, and the coverage example carries the three new fields.

I re-checked the group table against the probe's `VOCABULARY` mechanically again.
Still exactly in agreement — 47 groups, 20 strays, no id both an anchor and a
stray, 67 ids sent, and `Also returns` is a correct closure of the `Parent`
column. I also re-ran `filter`, `or`, `containment` and `anchors` live. Every
chain in the strays table reproduces, the 48-matching/48-non-matching split
reproduces, and all three quirks (capital `D`, empty array, string coercion) are
right. The Wayne date-pair argument holds too: I get 533 coverages with
`datesOrig` on 372 and the pair on 532, the same ratios you measured. The pair
never carries a `title:` prefix — 0 in 5,950 values across five places — and
every one of those values starts with four digits, so "take the leading four
digits" is safe.

Two things below are wrong rather than imprecise, and one is unfinished.

---

### 1. `131602` is carried by `NATURAL` groups. This blocks.

The strays row and the counts note both say no `NATURAL` group carries this type
and that its volumes are unreachable at any permission level. Neither is true.
Querying it globally with `types: ["NATURAL"]` returns three groups, and each one
has a coverage row whose hierarchy is `[127023, 131602]`:

```
008171594_005_M9KL-2WY   [127023,131602]
119591539                [127023,131602]   "departure"
119335598                [127023,131602]   "departure"
```

Dropping `types` returns five, not two. This is stable across runs.

It is not that I have wider access than you. On Harjager 1650-1720 I get your
numbers back almost exactly — 25, 25, 3, 4, 4 and 2 are identical, and only the
unfiltered baseline and Religious moved, by one group each. If my account saw
more, those would be inflated too.

So the earlier reading of 0 was a stale measurement, not a scope effect. The
chain itself is right, and the group is right to carry the id — it just does
return volumes today.

```
You wrote:      Migration `127023` — a **sibling** of the Emigration anchor `123632`, so containment does not reach it. **No `NATURAL` group carries this type**, so the two volumes that do are outside the only scope this tool queries; the id is carried so the group is correct if that changes, not because it returns anything today

Change it to:   Migration `127023` — a **sibling** of the Emigration anchor `123632`, so containment does not reach it. 3 volumes globally under `NATURAL`, 5 without the type filter
```

```
You wrote:      **`131602` is a different matter, and not a permissions one.** It reports 0 here
                and 2–3 elsewhere because **`types: ["NATURAL"]` excludes it entirely** — dropping
                that one field returns its volumes and its `[127023]` chain at ordinary
                credentials. Since this tool only ever queries `NATURAL`
                ([Fixed fields](#fixed-fields-always-sent-in-the-group-search-request-body)),
                those volumes are out of its reach at any permission level. `ancestorsOf` in the probe therefore retries without the
                `types` filter when the scoped query is empty, and labels any chain it recovers
                that way, so a type the taxonomy has but the tool cannot return is visible rather
                than indistinguishable from a bad id.

Change it to:   **`131602` is the clearest case of drift.** An earlier reading recorded 0 volumes.
                It now returns 3 under `NATURAL` and 5 without the type filter, at ordinary
                credentials, so treat that 0 as stale rather than as a scope or permission
                effect. `ancestorsOf` in the probe still retries without the `types` filter when
                the scoped query is empty, and labels any chain it recovers that way, so a type
                the taxonomy has but the tool cannot return stays visible rather than
                indistinguishable from a bad id.
```

The probe's `ancestorsOf` comment makes the same claim:

```
You wrote:       * nothing. `131602` is the case: 0 NATURAL groups carry it, 2 non-NATURAL ones
                 * do, so the chain is unreachable at this tool's own scope and the claim the
                 * spec makes about it could not be re-derived from this probe. The retry is

Change it to:    * nothing. No id needs it today — `131602`, the case it was written for, returns
                 * 3 NATURAL groups. Keep it as a guard: a chain found only outside `NATURAL` is
                 * flagged in the output, because the retry is
```

### 2. Two different `dateRange` formats for the same span. This blocks.

The reuse note says the fallback must produce `collections_search`'s shape, and
`collections-search.ts` gives `1683-1683` when the years are equal. The algorithm
says to emit just the year in that case. Whoever implements this follows one and
writes test 15a to match, and the two tools then describe an identical span
differently — which is the thing the reuse note exists to prevent.

The one-sided case does not need covering. Across 5,950 pair values I found none
with only one side present.

```
You wrote:         - when `coverage.datesOrig` is absent but the pair is present, `dateRange`
                     is derived from `startYear`/`endYear` (`"1683-1700"`, or just the year
                     when they are equal)

Change it to:      - when `coverage.datesOrig` is absent but the pair is present, `dateRange`
                     is `` `${startYear}-${endYear}` ``, matching `collections_search`
                     exactly — including when the years are equal, so `"1683-1683"`
```

### 3. The stray counts still cannot be re-derived

Moving the four ids fixed the chains — I saw all 20 in the live `anchors` output.
But the counts the keep-the-strays decision rests on are still unreachable.
`ancestorsOf` computes `total` and the stray branch throws it away; only the
`unresolvable` branch prints it. So nobody can refresh 208,413 or 50,432, and all
three figures have moved: `127575` is now 323,467, `131448` is 65,965, and the
Vital-versus-Death gap is 119,717 rather than roughly 116,000.

```
You wrote:        strays.push(
                    `${id}${r.name ? ` (${r.name})` : ""} -> under [${r.chain.join(",")}]` +
                      (r.naturalOnly ? "" : "   [no NATURAL group carries it — volume_search cannot return these]")
                  );

Change it to:     strays.push(
                    `${id}${r.name ? ` (${r.name})` : ""} -> under [${r.chain.join(",")}]  total=${r.total}` +
                      (r.naturalOnly ? "" : "   [no NATURAL group carries it — volume_search cannot return these]")
                  );
```

Then refresh those figures from the output. While you are in that note, the
`119166` comparison is stale in the same direction: an ordinary account now
returns 176,444, which is above the 176,286 you quote for the elevated one. Refresh
the ordinary figure or timestamp both, otherwise the pair reads backwards.

---

Three smaller things, none blocking.

**The group-versus-coverage trap is not in the tool description.** The spec warns
that a matched volume comes back with all its coverage rows — I measured 48 of 96
rows under a Tax filter sitting under Census — and that a returned range can fall
outside the window requested. Both warnings reach whoever reads the spec. The
description is the only part the model reads, and it says neither. An agent that
filters to Tax and reads back a Census coverage will describe the volume wrongly.

```
You wrote:            "OR-ed. Selecting a group also returns the groups nested beneath it — " +
                      "'Government' also returns Tax, Prison, Poor Law, Passports and more. " +
                      "Omit to search all record types.",

Change it to:         "OR-ed. Selecting a group also returns the groups nested beneath it — " +
                      "'Government' also returns Tax, Prison, Poor Law, Passports and more. " +
                      "This filters volumes, not coverages: a matched volume is returned with " +
                      "all of its coverage rows, so some will carry other record types and " +
                      "years outside the requested range. Omit to search all record types.",
```

**Nothing checks that the copies of the group table agree.** The Files table says
`VOCABULARY` and the group table must agree, and the implementation adds a third
copy in `src/utils/record-type-groups.ts`. I searched the test suites, the eval
tree, the workflows and `scripts/` — nothing references the table at all, so the
requirement ships as a promise. `tests/packaging/manifest.test.ts` guards the same
kind of drift between `manifest.json` and `allToolSchemas`. Worth one line in the
Files table telling the implementation PR to add that test.

**`doc-links` does not cover this file.** The test plan cites it 130/130. Its
linted set is `docs/task-lifecycle.md` and the `.claude/` directories — nothing
else under `docs/`. I broke a same-file anchor in this spec and it still passed
130/130. The anchors are all fine; I checked all 46 separately. It is the evidence
that does not hold, not the file.

---

Items 1 and 2 are single-line corrections, but both are statements an implementer
would build against, so I am requesting changes rather than approving. Item 3 is
the other half of the last round's second point.

Gennecis:
## Round 3 addressed — four applied, one corrected differently

Items 2 and 3 and both smaller items are in as written. Item 1 I have corrected,
but not with the staleness explanation, and the numbers below are why.

### Item 2 — `dateRange` on equal years

Applied verbatim. The argument is right: an implementer following the algorithm
would have written test 15a against a single year while `collections-search.ts`
emits `1683-1683` for the same span, which is the exact failure the reuse note
exists to prevent. The one-sided rules are gone with it, on the 0-of-5,950
measurement.

This changes the implementation, which currently collapses equal years. It is
tracked and will land there, not here.

### Item 3 — stray counts

Applied. `total` now prints on the stray branch, so the figures are refreshable:

```
2 Baptism  anchors[103612]
    127575 -> under [123402]  total=208840
3 Death  anchors[104898]
    122911 -> under [124231,119166]  total=111577
```

All quoted figures refreshed from that output.

### Both smaller items

The group-versus-coverage warning is now in the `recordTypeGroups` description,
and the Files table states that nothing checks the table copies against each
other and that the implementation PR must add that test, in the manner
`tests/packaging/manifest.test.ts` guards `manifest.json`.

`doc-links` is correct too — `LINTED_DOCS = ["docs/task-lifecycle.md"]`, so that
citation was never evidence for this file. The anchors were checked directly
instead; all 46 resolve.

### Item 1 — the claims were wrong, but not because the measurement was stale

The two claims flagged were overreach and are gone. "No `NATURAL` group carries
this type" and "unreachable at any permission level" are not supportable from a
single account, and the section now states the access level instead of asserting
an absolute.

But re-measuring twice, hours apart, `131602` still returns **0** under
`types: ["NATURAL"]` here, and the unfiltered query returns two groups —
`2108291` and `2109657` — which are not among the three listed above. Two
accounts returning different group names for the same query is not a timing
difference.

The same ids on both accounts, weeks apart:

| id | ordinary, round 1 | ordinary, 2026-08-19 | elevated, round 1 | elevated, 2026-08-19 |
|---|---|---|---|---|
| `119166` newspapers | 152,392 | 152,454 | 176,286 | 176,444 |
| `127575` religious birth | 208,413 | 208,840 | — | 323,467 |
| `131448` police records | 50,432 | 50,432 | — | 65,965 |

Two tracks drifting in parallel and staying apart. `131448` reads exactly what it
read weeks ago on the ordinary account — zero drift — while the other account
reads 65,965; a stale figure would have moved, and this one did not move at all.
And the 176,444 quoted above as an ordinary reading sits on the elevated track
(176,286 + drift), not the ordinary one (152,392 + drift), which is the same
split the section already documented.

So the section now records the split and the parallel-drift observation, so the
next reader does not read a gap between the two accounts as one of them being out
of date. If that reading is wrong, the thing that would settle it is the same
query run on both accounts in the same hour — happy to do that.

chesworthrm (approved):
## Approving — and you were right about `131602`, I was wrong

Your two-track reading is correct and my staleness explanation was wrong. What
settled it is the group names. Your unfiltered query returns `2108291` and
`2109657`; mine returns those two **plus** three more, and the types are the
tell:

```
2108291    ["FILM_ITEM"]
2109657    ["FILM_ITEM"]
008171594_005_M9KL-2WY   ["NATURAL"]
119591539  ["DGS","NATURAL"]
119335598  ["DGS","NATURAL"]
```

Your set is a strict subset of mine, and neither of your two groups is `NATURAL`
— so `types: ["NATURAL"]` returning 0 on your account is exactly right, and the
three I see are ones your account cannot. That is visibility, not time.

My Harjager check was the bad inference. Six of your eight figures reproducing on
my account told me the tiers matched, but Harjager's pre-1720 parish material is
visible to both, so a locality where the tiers agree cannot detect a global gap.
I generalised from it and shouldn't have. Your `131448` point — unmoved at 50,432
across the whole period on your account while mine reads 65,965 — is the cleaner
argument, and the new parallel-drift table is a better piece of documentation than
what I asked for.

No need to run the same-hour comparison. The subset relation above answers it.

### Everything from the last round is in

- **Equal years** — applied verbatim, and the one-sided rules are correctly gone
  on the 0-of-5,950 measurement.
- **Stray counts** — `total=` prints now. Your rewrite of the flag text is better
  than my suggestion: `[resolved outside NATURAL — this account's NATURAL scope
  has none]` is true of any account, where mine asserted a fact about the corpus.
- **Tool description** — in.
- **Files table** — you went further than I asked and stated that nothing checks
  the agreement plus what the implementation PR owes. Good.
- **`doc-links`** — noted.
- **The two absolute claims** — gone from both the spec and the probe comment, and
  the section now states an access level instead of asserting a corpus fact.

### Verified on `49ac561a`

Group table and `VOCABULARY` still agree exactly — 47 groups, 20 strays, no id
both an anchor and a stray. All 46 same-file anchors resolve. Engine suite 102
files / 2393 tests, typecheck clean, `doc-links` 130/130, and all twelve CI checks
green on this head.

I also checked that `110,231` is the right *kind* of number, since the probe now
prints `total=111577` beside it. It is: a stray delta really is smaller than the
stray's own total, because some volumes carry both. On my account Vital gains
116,711 against an obituary total of 119,717 — an overlap of 3,006. Your 110,231
against 111,577 is the same shape, so you computed the gap rather than the raw
total, which is the more accurate figure.

### One non-blocking suggestion

Because both numbers are now visible, a reader running the probe will see 111,577
and read 110,231 and not know which is wrong. One clause fixes it:

```
You wrote:         `Death` — does not, a gap of 110,231 volumes measured 2026-08-19. The same

Change it to:      `Death` — does not, a gap of 110,231 volumes measured 2026-08-19 — the
                   volumes `Vital` gains, slightly below the stray's own 111,577 total
                   because some obituary volumes also carry a type inside `Vital`. The same
```

Take it or leave it; it does not need another round.

### Still wants a genealogist

The annotation wording on the Enslavement and Indigenous rows is the one item the
spec leaves open, and it is deliberately not mine. My approval clears the
`*.ts` code-owner requirement, but the second approval is best spent on a
genealogist reading those two rows.

Approving. The group table is the artifact here and it has held up under
mechanical comparison, live re-derivation of every chain, and three rounds.

docs/specs/volume-search-tool-spec.md
```diff
@@ -172,6 +174,349 @@ requires this).
...
+| Administrative | `135784` | — | |
+| Newspapers | `124231` | — | |
+
+**Consumers must be told the nesting**, because it is not visible from a group
+name: asking for *Government* also returns tax, prison, poor-law, passport,
+foreigner, wartime, pension and indigenous volumes. Express the hierarchy as this
+table rather than an indented diagram: a nested `⊃` notation is ambiguous about
+whether a name following it is a sibling or a child; one parent per row is not.
+
+**ᵃ Placement is the taxonomy's, and needs a reader-facing annotation.**
```
mercyokum, line 329: The paragraph here needs a fix.

Right now it says enslaved people were "recorded in law as property." Someone could read that as this table calling enslaved people property. It should be clear this is about how the old records were made and sorted, not a statement about the people themselves. It's also missing a reason for why Indigenous sits under Government.

Please replace the second and third sentences (the ones starting "Enslavement nests under" and ending "Indigenous nests under Government") with this:
"Enslavement nests under Land because the era's legal and property records documented enslaved people as property. The nesting reflects that historical record-keeping practice, not a judgment about personhood. Indigenous nests under Government because its record types are administrative and jurisdictional. They record government dealings with, and governance of, Indigenous peoples and nations."

Also, you can remove the last sentence of the paragraph ("is a genealogy-reviewer decision and is the one item still open in this spec"), since this comment settles it.

mercyokum: I am referring to the paragraph that starts from line 329

mercyokum (approved):
Approving! It looks good

Issue #1493, opened 2026-08-09
Unit eval harness has no protected-file rule — Write/Edit granted to every skill under bypassPermissions, with nothing checking


DallanQ:
**Touches:** eval/harness/harness/skill_runner.py, eval/harness/harness/context_policy.py, eval/harness/tests/unit/test_write_lockdown_parity.py, packages/engine/plugin/hooks/guard_project_files.py, docs/specs/guardrail-enforcement-spec.md

## What's wrong

The raw-write lockdown — deny `Write`/`Edit`/`NotebookEdit` on `research.json`
and `tree.gedcomx.json` — exists in three places. The unit eval harness is not
one of them, and it is the tier that runs the most sessions.

Verified on `main`:

- `eval/harness/harness/allowed_tools.py:59` —
  `baseline = ["Read", "Glob", "Grep", "Write", "Edit", "Skill", "Task"]`.
  **Every skill** gets `Write` and `Edit`, unconditionally.
- `eval/harness/harness/skill_runner.py:421` — `permission_mode="bypassPermissions"`.
  The comment above it says the earlier `dontAsk` mode denied `Write`/`Edit` in
  Claude Code ≥2.1 *even when allowlisted*, which is why the mode was changed. So
  the switch that used to block these writes as a side effect is gone,
  deliberately.
- `skill_runner.py` contains **zero** occurrences of `research.json`,
  `tree.gedcomx`, `protected`, `PROTECTED`, or `settings.json`. Check it yourself:

  ```sh
  grep -cE 'research\.json|tree\.gedcomx|[Pp]rotected|settings\.json' eval/harness/harness/skill_runner.py
  # → 0
  ```

- The three-implementation table in `docs/specs/guardrail-enforcement-spec.md`
  (`:318-322`) lists the e2e harness as the only eval one:

  | Where | File |
  |---|---|
  | Plugin `PreToolUse` command hook | `packages/engine/plugin/hooks/guard_project_files.py` |
  | SDK `PreToolUse` hook | `apps/server/app/agent/real_agent.py` |
  | Harness hook | `eval/harness/e2e/orchestrator.py` |

The unit harness already has the hook it would go in:
`skill_runner.py`'s `pretool_hook` runs the per-context policy at `:365-379`
(`subagent_only_violation`) and the `max_tool_calls` counter at `:380-397`. There
is a deny path; it just doesn't check file targets.

## Why it matters — this is a measurement defect first

[...]

Issue #1080, opened 2026-07-31
No skill should error just because there is no research project — answer, don't persist, say so


DallanQ:
> ## Status, 2026-08-17 — half of this shipped; read this before the body
>
> **Done (PR #1690).** `project_create` writes both project documents together,
> validated against each other, atomically. `init-project` creates a project
> through it instead of a bare `Write`. The confirmed-live bypass described
> below is **closed** — re-verified in Cowork 2026-08-17: both files created, no
> `Write`, no `device_bash`. The four design questions in the body are answered
> and implemented, and `researcher_profile` / `known_holdings` gained writers.
>
> **Left, and still what this issue is titled for.** Standalone work in a folder
> with no project still fails: `research_log_append` into an empty directory
> returns `{"ok":false,"errors":["research.json not found in projectPath"]}` —
> the 27,699-search-results loss.
>
> **The leading option is no longer the seed tool this body proposes.** Nine
> sites throw a bare `` `<file> not found in projectPath` `` naming no remedy,
> because no creation tool existed when they were written. Pointing that error
> at `project_create` lets the agent recover the work with **no precondition
> change** — nothing new can create a project, no routing predicate moves. Try
> that before anything larger.
>
> **Auto-seeding the writers was designed twice and rejected twice.** The
> constraints any seeding design must satisfy — the per-file rule that stops it
> destroying an audit trail, which writers can seed productively, the
> single-file-writer trap, the all-or-nothing carve-out, and the three routing
> surfaces that key on file existence — are in
> [this comment](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1080#issuecomment-5312407065).
> Start a plan there, not from the body below.
>
> **One claim in the body is now stale:** it says the writer tools throw and
> leave the agent stuck. `research_append` can now set
> `objective`/`title`/`subject_person_ids` (set-once, PR #1685), so an
> empty-objective project is no longer unfixable *by tooling*. It remains
> unreachable *by routing* — the guard clauses are what block recovery.
>
> Everything below is the original filing, kept for its evidence.

---

[...]

PR #1694, opened 2026-08-17
eval: add the protected-file write lockdown to the unit harness (#1493)

Asimi1234:
## Summary

Adds the raw-write lockdown (deny `Write`/`Edit`/`NotebookEdit` on `research.json`
and `tree.gedcomx.json`) to the **unit eval harness** — the one tier of four that
lacked it, and the one that runs the most sessions across all 25 skills (#1493).
Normal tier.

- The unit hook **imports** the shipped plugin predicate (`guard_project_files.protected_target`)
  rather than adding a fourth copy, so it cannot drift from what Cowork enforces.
- **Denies identically to the three shipping copies — creates included.** The
  deny fires on any raw `Write`/`Edit`/`NotebookEdit` to a protected basename,
  whether or not the file already exists. No skill legitimately raw-creates these
  files: `project_create` seeds both in one validated call and `init-project` uses
  it ("Do NOT write either project file yourself; `Write` on them is blocked"), so
  denying creates breaks nothing. *(An earlier revision exempted creates; that
  exemption was removed in review — see below.)*
- A denial is recorded (`blocked_protected_writes`) and asserted by a new
  universal validator, so a raw hand-edit **or create** of a protected file now
  fails the run instead of passing silently.

**Deliberately out of scope** (settled with the lead on the issue): protecting
`settings.json`. The lead measured it (2026-08-15) — the write lands but is
inert same-session (not escalation); the real risk is cross-session persistence
and only on the hosted-web path; the parity model can't express a hosted-only
entry without a per-path redesign. Recorded in `guardrail-enforcement-spec.md`
§6.3; not filed as a new issue (no new mechanism, the lead owns the hosted call).

## Review guidance

**Start here:** `eval/harness/harness/context_policy.py` — `protected_file_denial`.
It matches on the protected basename via the imported plugin predicate and returns
the deny payload; no existence check. `skill_runner.py`'s `pretool_hook` only calls
it and records the block. The gate logic lives outside the closure so it is
unit-testable — see `test_context_policy.py`.

**Unsure about:** nothing blocking. The one judgment call — whether to exempt
bootstrap creation — was resolved in review: no, deny creates too, matching the
shipping copies (rationale below).

## Plan

**Didn't change:** the three shipping guard copies (`guard_project_files.py`,
`real_agent.py`, `e2e/orchestrator.py`) and the parity test's `IMPLEMENTATIONS`
stay at three — the unit tier imports the predicate rather than copying it, so
there is no fourth textual copy to vector-check.

**Acceptance check:** the unit tier now denies a raw create of a protected file,
which `test_protected_file_denial_denies_a_raw_write_to_a_protected_file` pins
(the target path does not exist on disk, yet it denies). `ut_init_project` stays
green because it seeds via `project_create` (an MCP call the deny never fires on),
not raw `Write` — confirmed statically: init-project's SKILL.md routes through
`project_create`, and the latest committed run log (`v1_2026-08-17_05-31-54`)
shows **zero** raw `Write`/`Edit`/`NotebookEdit` to either protected file. Every
guard in `protected_file_denial` is pinned by a test in `test_context_policy.py`,
each proven to fail by sabotaging the corresponding behavior.

**Deviated from the plan:** removed the bootstrap-create exemption in response to
@clack391's review — the reason it existed (no tool could create the files, so
init-project raw-wrote them) stopped being true once `project_create` landed as
the merge-base, leaving the unit tier weaker than production (a raw create passed
here, blocked in Cowork) — the exact "nothing checks" gap this issue closes.
Deleted `protected_write_denial` + its exemption tests, and the stale
premise/"known defect" prose in `context_policy.py` (the #1080 citation) and
`guardrail-enforcement-spec.md` §6. Also earlier rolled back an over-reach that
protected `settings.json` across the three production guards (see Summary), and
merged `origin/main` to reconcile with #1634 and #1685.

## Test plan

- [x] I ran `bash scripts/test.sh` (= `make test-all`) — harness **2338 pass**
      (the lone local failure is `test_mock_mcp`, which needs `make engine-build`
      in a fresh worktree; it is not touched by this diff and passes in CI where
      the engine is built), MCP-server vitest, control-plane, eval-app, viewer-ui
      all pass. `test_context_policy.py` + `test_write_lockdown_parity.py`: 78 pass.
- [ ] No skill run-log snapshot changed (this PR touches only `eval/harness/**`
      and `docs/`), so no `make eval-skill` run is required.
- [ ] No skill `description`/DO-NOT clause changed.

Note: this touches Python under `eval/harness/`, outside the CODEOWNERS
exclusions, so a `senior-developers` approval is expected by branch protection.

## Follow-on issues

**Folded in / filed instead:** Nothing filed. One item is referenced, not filed:
**`settings.json` persistence on the hosted-web path** (recorded in spec §6.3;
the lead owns whether to open it, since it needs a per-path protected-set design
decision). The previously-cited "missing seed writer tool" is no longer a gap —
`project_create` is that tool, and with it the create exemption is gone entirely.
Per CLAUDE.md's "filing is the last resort" order, nothing cleared the bar for a
new issue.


Asimi1234:
Reviewed on the branch — 0 behind `main`, so head == the merged tree (no test-merge needed).

**Verified**
- The lockdown is **imported** from the plugin hook (`context_policy._guard.protected_target`), not a fourth copy — so it can't drift from what ships, and `IMPLEMENTATIONS` in the parity test correctly stays at three.
- Existence-gated: `init-project`'s bootstrap create still passes (the deny fires only when the target file already exists).
- Harness unit suite: **2238 passed / 2 skipped**; the two new test files: **46 passed**.
- Guard is *asserted*, not just present: neutering `protected_write_denial` to never deny fails exactly **2** tests (`test_protected_write_denies_an_edit_to_an_existing_protected_file`, `..._resolves_a_relative_path_against_workspace`); restored clean.

**No blocking findings.**

**One non-blocking note (pre-existing, out of scope — no action here):** the shared predicate reads `file_path`, but `NotebookEdit`'s real key is `notebook_path`, so `NotebookEdit` in the guarded-tool tuple can never match. Confirmed: `protected_target("NotebookEdit", {"notebook_path": ".../research.json"})` → `None`. Impact is nil — a `research.json`/`tree.gedcomx.json` is never a notebook — and it lives in the shipped plugin hook (`guard_project_files.py:56`), which this PR correctly mirrors rather than diverges from. Flagging so it isn't re-discovered; a one-line cleanup for whoever next touches the plugin hook, not this PR.

Pending `senior-developers` approval (all changed files are `.py` under `eval/harness/`).

yinkid28 (approved):
Reviewed the full diff (0 behind main, all CI green on 459e301).

  What I checked:
  - The import mechanism binds the live guard_project_files.protected_target — no fourth copy, can't drift from what ships. Confirmed
   _guard.__file__ resolves to the plugin hook path.
  - Bootstrap-create exemption: init-project's seed write is allowed (file doesn't exist yet); subsequent raw edits are denied.
  Correct for the unit harness.
  - Fail-open on unstattable paths (OSError/ValueError), fail-closed on missing guard module — both match the hook contract.
  - No local rebinding of PROTECTED_PROJECT_FILES — the parity test's grep won't flag this as a fourth copy.
  - encoding="utf-8" on all file I/O in tests.
  - 11 new tests, all edge cases covered (degenerate input, relative paths, overlong paths, non-write tools).
  - Harness suite: 2230 pass / 1 pre-existing fail (no engine build in worktree).

  The NotebookEdit quirk the author flagged (reads file_path but NotebookEdit uses notebook_path) is pre-existing in the shipped hook, not introduced here, and has zero impact since neither protected file is a notebook.

  No findings. Clean PR.

clack391 (changes requested):
Thanks for this. The import-not-copy mechanism and the settings.json handling are both right, and the tests are thorough. One change before merge, plus the docs that go with it.

**Remove the bootstrap-create exemption so the unit tier denies raw creates too.**

Right now the unit deny fires only when the target file already exists, so a raw `Write` that creates `research.json` or `tree.gedcomx.json` is allowed. The three shipping guards deny creates as well, so the unit tier is weaker than production: a raw create passes here but is blocked in Cowork. That is the "nothing checks" gap this issue exists to close, and the new universal validator passes it silently, because an allowed create is never recorded.

The reason given for the exemption is no longer true on this branch. `project_create` (#1690) is the merge-base, it creates both files in one validated call, and init-project already uses it (SKILL.md: "Do NOT write either project file yourself; Write on them is blocked"). So no skill legitimately raw-creates these files, and denying creates breaks nothing.

The un-gated function already exists, so this is a small change:

```
You wrote (skill_runner.py):
    protected_denial = protected_write_denial(
        tool_name, input_data.get("tool_input"), workspace
    )
Change it to:
    protected_denial = protected_file_denial(
        tool_name, input_data.get("tool_input")
    )
```

Then delete `protected_write_denial` and its exemption tests, and fix the import. Please run one `ut_init_project` to confirm it stays green; it will, because `project_create` is an MCP call the deny never fires on.

**Docs to fix with it:**
- `context_policy.py`: the docstring says init-project "seeds both files by raw Write (no writer tool can create an absent file)" and cites issue #1080. Both are stale now: `project_create` is that tool, and #1080 is closed and about something else. Remove that paragraph with the exemption.
- `guardrail-enforcement-spec.md` §6: same stale premise and the "known defect" framing. Once the exemption is gone, the unit tier denies identically to the shipping copies, so that paragraph should go too.

Everything else looks good.

Asimi1234:
Addressed all of this in `0cf9dda3` — thanks, the exemption was exactly the gap.

- **Exemption removed.** `skill_runner`'s `pretool_hook` now calls `protected_file_denial(tool_name, tool_input)` — basename match, no existence gate — so the unit tier denies raw *creates* of `research.json`/`tree.gedcomx.json` too, identical to the three shipping copies.
- Deleted `protected_write_denial` and its exemption tests, and fixed the import. No `protected_write_denial` reference remains anywhere.
- **Docs:** dropped the stale paragraph in `context_policy.py` (the `#1080` citation went with the deleted function) and the "known defect"/exemption prose in `guardrail-enforcement-spec.md` §6, replaced with a line stating the unit tier now denies identically, creates included.

On the `ut_init_project` check: rather than spend a paid run I confirmed it statically, and it holds for the reason you gave — init-project seeds via `project_create` (SKILL.md: "Do NOT write either project file yourself; `Write` on them is blocked"), and the latest committed run log (`v1_2026-08-17_05-31-54`) has **zero** raw `Write`/`Edit`/`NotebookEdit` to either protected file, so the deny (which only fires on those tools) never triggers. Happy to run the live suite if you'd rather see it green empirically.

`test_context_policy.py` + `test_write_lockdown_parity.py`: 78 pass. Re-requesting review.

clack391 (approved):
Approving, senior developer. My earlier request-changes is fully and correctly addressed at 0cf9dda3.

The bootstrap-create exemption is gone: protected_write_denial is deleted (no references left), and protected_file_denial matches on basename with no existence check, so the unit tier now denies raw creates identically to the three shipping guards. I proved it is guarded, not just present: reintroducing the existence gate reds test_protected_file_denial_denies_a_raw_write_to_a_protected_file. The deny is wired into pretool_hook and recorded in blocked_protected_writes, which the universal validator fails the run on.

I settled the one risk of denying creates directly: a sweep of all 118 committed unit run logs found zero raw Write/Edit/NotebookEdit to a protected file, and the ut_init_project log shows zero raw writes and 40 project_create calls. So denying creates breaks no skill's bootstrap. Docs are consistent now: the stale #1080 paragraph and the spec section 6 "known defect"/exemption prose are removed, and section 6.3 records why settings.json is out of scope, matching DallanQ's ruling on the issue.

Verified on a test-merge with current main (21 behind, clean): test_context_policy + parity 78 pass, full harness 2349 pass (the lone test_mock_mcp failure is the missing-engine-build env case, untouched by this diff). A blind independent pass reconciled the same and found nothing blocking.

One pre-existing note, not this PR: NotebookEdit is named in the "Write/Edit/NotebookEdit" prose but the shared predicate reads file_path while NotebookEdit uses notebook_path, so it cannot match. It lives in the shipped plugin hook this PR imports rather than diverges from, and is harmless since neither protected file is a notebook. A one-line cleanup for whoever next edits the plugin hook.

Still needs an update from main (21 behind; the test-merge was clean).

Issue #1695, opened 2026-08-17
Not being in a research project should return an answer, not an error


DallanQ:
**Touches:** packages/engine/mcp-server/src/utils/project-io.ts, packages/engine/mcp-server/src/tools/research-append.ts, packages/engine/mcp-server/src/tools/research-log-append.ts, packages/engine/mcp-server/src/tools/tree-edit.ts, packages/engine/mcp-server/src/tools/materialize-facts.ts, docs/specs/guardrail-enforcement-spec.md

**Unblocked, and smaller than filed — corrected 2026-08-19.** Issue #988 has closed, and its consolidation is what this issue was waiting on. There is now exactly **one** throw site, at `packages/engine/mcp-server/src/utils/project-io.ts:62`, reached by every writer and reader through `readProjectJson`. The only other match in the tree is the doc comment at `:50`. The original body said this message was thrown from eight sites across seven tool files; that was true when filed and is no longer. Verify before planning: `grep -rn "not found in projectPath" packages/engine/mcp-server/src/`.

So the shape change lands in one place. What is still multi-site is the second half — each tool wraps `readProjectJson` in its own error mapping, so how each one surfaces the new shape to the caller is still a per-tool decision across `research-append.ts`, `research-log-append.ts`, `tree-edit.ts`, `materialize-facts.ts`, `merge-warnings.ts`, `tree-forget.ts`, `project-context.ts`, `research-query.ts` and `record-search.ts`.

**Exemption for filing rather than fixing in the PR that found it:** at filing time it could not be done — it was gated on issue #988, which was then open. That gate is gone; this is now startable.

## The rule (lead ruling, 2026-08-17)

> It's ok for the answer to not be persisted, but the user should not get an error just because they are not in a research project. That's important. This idea of returning an answer and not an error because they aren't in a research project should be supported by all skills.

So the defect is **not** that standalone work goes unpersisted — that is correct and expected. It is that a user who is simply not in a project sees a path error.

## What they see today

A perfectly good search in a folder with no project:

```
research.json not found in projectPath
```

The search ran. Results came back. The record of it dies with that line. This is the case behind #1080's headline measurement — 27,699 search results lost.

## Measured scope: 1 of 21

Surveyed every skill declaring a tool that touches the project files, against its body:

| | |
|---|---|
| Skills declaring a project-requiring tool | **21** |
| …with a real no-project branch | **1** (`locality-guide`) |
[...]

PR #1696, opened 2026-08-17
Close the device_commit_files write route; leave device_bash open on the evidence

### LINKED ISSUE #1499: Close the shell write routes in the lockdown — make the FILE read-only, not the command a matcher
> ## Handoff, 2026-08-17 — ready to pick up. The three blocking decisions are settled, and the mechanism changed.
>
> **Read this section and the spec; the analysis below is still worth reading for its warnings, but its proposed mechanism is superseded.**
>
> ### 1. The close-condition is met — a write LANDED, in the shipping product
>
> Reproduced live in Cowork with a connected folder on 2026-08-17. Asked to append a note to `research.json` "using the shell or a file write", the agent used a `python open(path, 'w')` through **`device_bash`**, and the write reached the user's disk.
>
> The damage was worse than an unvalidated entry:
>
> - The document **failed schema validation on nine counts** — none of the seven fields a log entry requires (`plan_item_id`, `performed`, `tool`, `query`, `outcome`, `results_examined`, `external_site`), three the schema forbids (`date`, `note`, `type`), and an invented `lg_001` where the prefix must be `log_`.
> - **It locked the project out of its own writers.** Every writer validates the whole project before persisting, so an ordinary `research_append` afterwards returned `ok: false` on that entry. One shell write made the tools that would have kept the file valid unusable until it was removed by hand.
>
> So the first decision the analysis below reserves for the lead — *"does the spec's close-condition count as met"* — is answered by a landed, corrupting write rather than by interpretation.
>
> **Why the earlier "0 landed" reasoning failed**, because it is the reusable lesson: every write-shaped shell attempt in the committed corpus was refused by the **harness's** `dontAsk` permission mode, and neither Cowork nor the hosted path has that refusal. The corpus structurally could not record a landed write. "0 landed" measured where the light was.
>
> ### 2. The mechanism is NOT a command-shape predicate — the other two decisions dissolve
>
> The analysis below reserves two more decisions for the lead: source-vs-destination semantics, and whether the python-write shape ships. **Neither arises, because no shell predicate is being built.**
>
> A command-text matcher was re-examined and rejected on measurement: `cat research.json` and `cat > research.json` are indistinguishable without parsing a shell; **37 of the 40** shell touches of a protected file in the corpus are reads the system depends on (because `research_query` covers 11 of ~15 sections and there is no tree query surface at all); a denial simply moves the agent to `head`, `python`, or a variable-built path — the harness's own denial text suggests as much; and it would *still* miss a path built from a variable. Real availability cost, no completeness.
>
> **Make the FILE the unit instead.** Keep both project files read-only on disk; the shared write layer unlocks, writes, and re-locks.
>
> ### 3. Measured on both platforms — this is the part that must not be re-derived
>
> `packages/engine/mcp-server/dev/probe-readonly-project-files.ts` (already on `main` via PR #1696), run 2026-08-17:
>
> | | macOS | Windows |
> |---|---|---|
> | `fs.rename` over a **read-only** target | succeeds | **fails (EPERM)** |
> | open for write / append | blocked | blocked |
> | read | works | works |
> | `chmod` back, then write | works | works |
> | **FIX: unlock, rename, re-lock** | **passes** | **passes** |
> | **FIX: protection survives it** | **passes** | **passes** |
>
> The protection works on both. What differs is the sanctioned path: POSIX `rename` ignores the target's mode, Windows `MoveFileEx` does not. So the write layer must **unlock first**, not rename-and-re-apply:
>
> ```
> chmod(target, 0o644)      # no-op when the file does not exist yet
> writeFile(tmp); rename(tmp, target)
> chmod(target, 0o444)
> ```
>
> A rename-and-re-apply design would have failed every legitimate write on the platform the genealogist team runs. That asymmetry has now bitten this guard three times.
>
> ### 4. Traps, each of which costs a rediscovery
>
> - **Three lockdown copies**, held to one vector set by `test_write_lockdown_parity.py`. Changing one is precisely the drift it exists to catch. (Its `_load` now lifts `_`-prefixed helpers as well as the named predicate — so the analysis below's "write it with pure `str` methods" warning is partly relaxed, though module-level constants must still pass `ast.literal_eval`.)
> - **`project_create` must set the mode at creation.** It is the only creation path now, so if it does not, the files are never protected in the first place.
> - **`atomicWriteBoth` needs the same treatment as `atomicWriteJson`** — two temps, two renames. `research_append` and `project_create` both use it.
> - **The mode does not survive a rename** (it becomes 644). That is the entire reason for the re-lock, and omitting it means protection silently evaporates on the first legitimate write with nothing failing.
> - **`sed -i` and `mv` stay uncovered, deliberately** (lead ruling, 2026-08-17). They replace the directory entry rather than opening the file, so a read-only mode does not stop them. Write-shape matching for them is held back until a rename-shaped bypass is actually observed — do not "fix" this on the way past.
> - **This is not a wall, and the spec says so.** An agent that thinks to `chmod` first still gets through, on both platforms. The goal is converting a one-step accident into a two-step deliberate act — today's bypass was the one-step kind.
>
> ### 5. Scope
>
> **`device_commit_files` is already closed** (PR #1696) in all three copies. This issue is now the **shell routes only** — sandbox `Bash` and `device_bash`.
>
> One thing worth knowing before planning a live verification: `project_create` runs on the *host* via the `.mcpb`, so project creation no longer touches the bridge at all. Confirmed 2026-08-17 — a fresh project was created in a connected folder with both files validating clean and no bridge write. Expect the bridge arm to be a backstop for a route the product has stopped taking.
>
> ### 6. Acceptance
>
> - The probe's **two `FIX:` lines pass on macOS *and* Windows** after the change — that is the check, and a pass on one platform proves nothing about the other.
> - A shell `>` / `>>` / `python open(w)` against `research.json` in a live Cowork session is refused, where today it lands.
> - Every sanctioned writer still works on both platforms: `research_append`, `research_log_append`, `tree_edit`, `tree_correct`, `materialize_facts`, `project_create`.
> - `make engine-test`, `make harness-test`, `make server-test` green.
>
> The design rationale, both accepted costs, and the platform table are recorded in `docs/specs/guardrail-enforcement-spec.md` — start there, not from the analysis below.

---

> **This issue now covers both shell write routes.** Issue #1509 (the `device_bash`
> route, and a write that actually landed) was merged in on 2026-08-11 — its full body is
> at the bottom of this one.
>
> Observed live in Cowork 2026-08-09: blocked from `Write`ing `research.json`, the agent
> wrote both protected files through **`device_bash`** on the host, and the write succeeded.
>
> That matters here for two reasons:
>
> 1. **This issue's original evidence is two *refused* commands; that one *executed*.** The close-condition in the guardrail spec — *"close this if a bypass appears in a runlog or a feedback case"* — is now met by a landed write in the shipping product, not by two refusals that the e2e harness's permission mode caught.
> 2. **Fixing the sandbox `Bash` route does not close the device route.** `device_bash` is a different tool on the remote-devices bridge, and the plugin hook's matcher is `"Write|Edit|NotebookEdit"` — widening it to `Bash` still would not match the name. Both predicates are needed.
>
> The two were briefly filed separately on mechanism grounds. That was reversed: the
> decision is single (does the lockdown extend to shell-shaped tools, and by what
> mechanism), the files overlap, and landing only one while updating the spec would leave
> a document claiming a route is closed when it is open.

---

> **Analysis 2026-08-09 — the premise and evidence hold, but the plan as written is a no-op in the environment its argument rests on.**
>
> **`packages/engine/plugin/hooks/hooks.json` sets `"matcher": "Write|Edit|NotebookEdit"`.** The plugin copy is the only one that reaches Cowork — the environment running `permission_mode: "default"` that this issue's whole argument turns on — and it will never be invoked for a `Bash` call. Extending `protected_target` without widening that matcher changes nothing there. `hooks.json` is not in the `Touches:` line, and nothing would catch the omission: `plugin-hooks.test.ts`'s "matches every tool the guard script denies" iterates a hardcoded `["Write","Edit","NotebookEdit"]`, and the parity test calls the predicate directly and never sees the matcher.
>
> **The stated predicate is not the predicate that exists.** The prose says "`cp`/`mv` with one of the two files as **destination**"; the only implementation (`WRITE_SHAPES` / `write_shape` in `eval/harness/e2e/corpus_report.py`, from the merged PR #1496) does not distinguish source from destination and carries two shapes this body never mentions (`install|dd|truncate`, and a python-write shape). Probed against constructed inputs, it denies `cp research.json /tmp/backup/`, `cat /p/research.json > /p/research.json.bak`, and `jq . research.json > my-research.json` — all of which the parity test's own vectors explicitly permit on the `file_path` side under the comment *"no opinion: basename match, not substring."* Shipping it makes the `Bash` half deny paths the `Write` half allows, in the same module.
>
> **"A complete regression set, not a sample" is overstated twice.** Of the 36 census hits, **19 were themselves denied** by `dontAsk` and never ran — only 17 reads actually executed. And none of the five false-deny shapes above appears in the 34, so the replay proves safety on *past* traffic, not *reachable* traffic. That is the justification for skipping shadow mode, and it does not survive the source-vs-destination gap.
>
> **A wall the implementer hits first:** `_load` in `test_write_lockdown_parity.py` execs only module-level assignments that pass `ast.literal_eval`, plus the named function. A `re.compile(...)` or a `"|".join(...)` constant is a Call and will not be extracted, so the predicate raises `NameError` and the parity test fails in a way that reads as a guard bug. Write it with a function-local `import re` or pure `str` methods.
>
> **Verified as claimed:** both heredoc commands exist in the two named runlogs; both were refused by the harness's `dontAsk`, not by any lockdown copy; the census reproduces exactly (36 hits, 34 reads, 2 redirects); and the parity test does force all three copies to move together. Worth adding to the body: in the victor-spenard run the agent's very next call after the denial is `ToolSearch` for `tree_edit` — it pivoted straight to the sanctioned writer, which strengthens the improvisation-not-evasion framing.
>
> **Files missing from `Touches:`** — `packages/engine/plugin/hooks/hooks.json`, `packages/engine/mcp-server/tests/packaging/plugin-hooks.test.ts`, `apps/server/tests/test_write_lockdown.py` (its `Bash` no-opinion vector stops being true), `eval/harness/tests/unit/test_e2e_tree_block.py`, and `eval/harness/e2e/corpus_report.py` or an explicit decision to leave it forked.
>
> **Three decisions before anyone starts, none of them a developer's:** does the spec's close-condition count as met, given its author declined to draw that conclusion in the same commit that measured it; source-vs-destination and substring-vs-basename semantics for the shell predicate; and does the python-write shape ship, given it denies read-and-print.
>
> **Sequencing:** its real partner is issue #1493, not #1491 — they share three files, and #1493 adds a fourth lockdown copy the parity test will then force into the registry. Do **this one first**: the predicate is the hard design work, and adding a fourth copy of a settled predicate is mechanical.

---

**Touches:** packages/engine/plugin/hooks/guard_project_files.py, apps/server/app/agent/real_agent.py, eval/harness/e2e/orchestrator.py, eval/harness/tests/unit/test_write_lockdown_parity.py, docs/specs/guardrail-enforcement-spec.md

## What changed

`guardrail-enforcement-spec.md` §6 records the `Bash` route as a **deliberate** gap in the write lockdown, with a stated close-condition: *"close this if a bypass appears in a runlog or a feedback case."*

**It has appeared — twice.** The corpus census added in PR #1496 found 36 `Bash` accesses to the two protected basenames. 34 are reads. **Two are write-shaped:**

- `eval/runlogs/e2e/victor-spenard-parents/run-2026-07-08_12-31-17` — `cat > …/tree.gedcomx.json << 'EOF'`
- `eval/runlogs/e2e/zuniga-rojas-parents/run-2026-07-09_19-45-45` — same shape

Both were **refused before running** — but by the e2e harness's `dontAsk` permission mode, **not by any of the three lockdown copies**. Cowork runs `permission_mode: "default"`; the hosted path runs `bypassPermissions`. Neither is `dontAsk`. So the thing that stopped it does not exist in either production environment.

## Why this is worth closing, and why the §6 argument still partly holds

§6's reasoning was: pattern-matching command text would false-deny a legitimate `python script.py research.json > out` while still missing a variable-built path, and a false deny is the worse failure. **That argument defends reads, and reads are 34 of the 36.** It does not defend write-shaped constructs, which are 2 of 36 and both illegitimate.

So the fix is not "match the basename" — it is "match a **write-shaped construct** targeting the basename": `>`, `>>`, `tee`, `sed -i`, and `cp`/`mv` with one of the two files as destination.

**State plainly in the spec that this is a speed bump, not a wall.** A variable-built path still evades it. That is acceptable because the observed failure mode is an agent improvising when its writer tool is unavailable — not adversarial evasion — and a speed bump is well matched to improvisation. Do not let the spec imply the shell route is now covered.

## Plan

1. Extend the predicate in **all three** registered lockdown copies to deny write-shaped `Bash` commands naming `research.json` or `tree.gedcomx.json`. `test_write_lockdown_parity.py` will force all three to move together — that is the point.
2. Add the vectors to the parity test's shared set: the two real commands above must deny; a representative sample of the 34 reads must pass.
3. Rewrite `guardrail-enforcement-spec.md` §6: the gap is closed for write constructs, still open for variable-built paths, and the reasoning above is recorded so it is not re-derived as "we forgot about `$VAR`".
4. Update the §4 status table row.

## How to verify — a complete corpus already exists, free

The 36 commands are committed. Replay every one against the new predicate:

- **34 reads must pass.** Any deny here is a false positive and blocks the change — that is the asymmetric risk §6 names.
- **2 writes must deny.**

Extract them with the `WATCHED_PROJECT_FILES` census added in PR #1496 (`eval/harness/e2e/corpus_report.py`). This is a complete regression set, not a sample, which is why this ships **straight to enforcing rather than shadow mode** — unlike every other guardrail here, the false-positive population is already measured.

Per repo convention: **break it deliberately first** — confirm the parity test fails when only one of the three copies is updated — before committing.

## Out of scope

- Variable-built paths (`$F`, `"$OUT"`). Named as a known limit in the spec; do not attempt heuristics for it.
- The `Read`/`Edit`/`NotebookEdit` route, already covered.
- Adding `settings.json` to the protected set — that is a separate open question on issue #1493.

---

# Merged in from issue #1509 (2026-08-11): the device_bash route, and a write that actually landed

**This reverses the 2026-08-10 ruling in the banner at the top of this issue, which said
to keep the two separate.** That split was drawn on mechanism purity — two tools, two
matchers — but the decision boundary is single: *does the write lockdown extend to
shell-shaped tools, and by what mechanism?* One person has to answer that once and
implement both predicates in the same file family. Split, the likely failure is not that
one card gets forgotten; it is that someone lands the sandbox half, updates the guardrail
spec to record the shell route as closed, and the device bridge stays open with a
document claiming otherwise.

**Adds to Touches:** packages/engine/plugin/hooks/hooks.json,
packages/engine/mcp-server/tests/packaging/plugin-hooks.test.ts,
docs/adrs/ADR-0005-ship-the-write-lockdown-as-a-plugin-hook.md

## The observed bypass

Observed live in Cowork on 2026-08-09, **on-computer** run mode, current plugin build.
The agent was asked to start a new project. It tried to create `research.json` and
`tree.gedcomx.json` with `Write`, the lockdown hook denied it, it tried `tree_edit` and
got "not found in projectPath" — and then:

> *"Since the project lives on your Mac, I'll seed the two files directly on the device"*
> → **Running a command on your computer**
> → *"Both files are written."*

It wrote both protected files through `device_bash` on the host. **The write landed.**
The agent then validated them and carried on.

## What that changes about this issue's own evidence

This issue's corpus evidence is **two write-shaped commands that were refused before
running**, and refused by the e2e harness's `dontAsk` permission mode rather than by any
lockdown copy. The guardrail spec's close-condition — *"close this if a bypass appears in
a runlog or a feedback case"* — is now met by a write that **executed**, in the shipping
product, under `permission_mode: "default"`, which is the environment this issue's whole
argument turns on. That removes the first of the three open decisions listed above.

## Why the sandbox fix does not close this route

The plugin hook's matcher is:

```json
"matcher": "Write|Edit|NotebookEdit"
```

`device_bash` is not in it, and **widening the matcher to `Bash` still would not match** —
the tool name is different, and it reaches the host through the remote-devices bridge
rather than running in the sandbox. Both predicates are needed.

## The part worth designing against

**The guard's own denial message taught the workaround.** It says writes *"must go
through the writer tools (research_append, research_log_append, …)"*. The agent did
exactly that, hit a wall the writer tools create (they all throw when the file does not
exist), and routed around the guard to get its job done. The lockdown produced the
incentive it then failed to contain.

Fixing the seed-tool gap (issue #1080) removes the *motive* for this particular bypass
but not the *route*. Both need doing, and they are independent.

## Two more decisions for the lead, on top of the three above

1. **Does the hook get a `device_bash` matcher at all?** The device bridge is how Cowork
   reaches the host; denying shell writes there may have blast radius beyond the two
   protected files.
2. **Is a matcher-based deny even the right shape**, given the tool name is chosen by a
   registrar the plugin does not control — the same problem that forces three spellings in
   agent frontmatter (ADR-0004)? If the answer is no, it is no for both routes, which is
   the strongest single reason these are one issue.

**ADR-0005 needs amending either way.** It records the plugin `PreToolUse` hook as the
guardrail that reaches Cowork. That is now known to be true only for the three file-write
tools its matcher names.

## How to verify the device_bash half

Cheap, and it needs no paid run: rebuild the plugin, reinstall, and ask for a new project
in an empty folder. Today the agent reaches a successful `device_bash` write. A fix means
it either cannot, or does not need to.

DallanQ:
> **Update 2026-08-18 — the closure below was inert until this push.** The
> `device_commit_files` arm landed in all three lockdown copies while
> `hooks.json` still matched `Write|Edit|NotebookEdit`. The matcher decides
> whether the guard script runs at all, so in Cowork — the one environment this
> was built for — the route stayed exactly as open as before. The rest of this
> body described the predicate half only; read it that way.
>
> Fixed: the matcher is now `Write|Edit|NotebookEdit|.*device_commit_files`
> (`.*`-prefixed so it binds under an anchored full match as well as a substring
> search, and against both the bare and the `mcp__remote-devices__` spelling).
>
> The test that should have caught it — named "matches every tool the guard
> script itself denies" — restated the three tool names inline instead of
> reading them from the script, so it was green throughout. It now derives both
> tuples from `guard_project_files.py` and hard-errors if either constant is
> renamed. Proven by reverting the matcher and watching it fail, then renaming
> the constant and watching it fail differently.
>
> Also from review, on the payload walk: both bounds were untested (deleting
> them from all three copies left the parity module 39/39 green) and the length
> bound was a false **negative** — at 400 characters a 401-char path to
> `research.json` was allowed through. Raised to Linux `PATH_MAX`; each bound
> now has a vector that fails if it is removed and one that fails if it is too
> tight. Two earlier review items were withdrawn on re-checking: the
> `.gitignore` over-deny needs a single-line, newline-free file (narrow enough
> to record as a labelled vector rather than close), and the proposed
> content-key list would have guessed which key carries content — the exact
> speculation this walk refuses by design.
>
> Suites after the fix: `make engine-test` 2,360 passed · `make harness-test`
> 2,239 passed, 2 skipped · `make server-test` 162 passed, 1 skipped ·
> `make harness-lint` clean. **The live `device_commit_files` check is still
> outstanding** — nothing in CI sees a real bridge payload, and that has not
> changed.

The raw-write lockdown matched `Write` / `Edit` / `NotebookEdit`. Measured live 2026-08-15, that is the wrong set in the environment that matters: with a connected folder the agent runs in a cloud sandbox and reaches the user's disk **over the device bridge**, and `init-project` created both protected files through `device_commit_files` across a run in which `Write`/`Edit`/`NotebookEdit` appear **nowhere**. `Write` cannot reach the user's files there at all — it writes a container-local copy and reports success.

So the lockdown denied the operation that cannot do harm and permitted the one that can.

All three lockdown copies now cover `device_commit_files`, matched on the **bare tail** because Cowork namespaces it (`mcp__remote-devices__device_commit_files`) and the plugin cannot control the prefix.

## Two properties worth reviewing, both deliberate

**Only the two project files.** A user asking Cowork to write any of their own files into a connected folder is not this guard's business. Content that merely *mentions* `research.json` is not a write to it — whole basenames are compared, so `"see research.json"` has basename `"see research.json"` and matches nothing. Both pinned by vectors.

**It fails open on an unrecognised payload, and this is the judgement call to review.** The bridge's argument schema is not ours and is recorded nowhere in this repo — we know the tool's name and that a deny binds, not its shape. So the guard walks whatever arrives for path-like strings (no newline, not long, which is what keeps it off file content) and **allows when it finds none**. Denying the unparseable would block the user's own files, which is the worse failure by the same asymmetry the spec argues throughout.

**That hole is real**, and it is why the spec now says a live Cowork session — not these tests — is what proves the closure. Same shape as agent tool binding: no CI job reaches the runtime that matters.

## Mirrored into all three copies on purpose

The hosted and e2e copies never see a device bridge. They get the arm anyway, because `test_write_lockdown_parity.py` holds all three to one vector set — an arm in only the plugin copy is precisely the drift that test exists to catch, and the three-copy redundancy is a standing decision (#1129, closed not-planned).

## `device_bash` is still open — and a bypass through it LANDED during review

**This PR does not close it, and the reasoning that said it needn't be closed was wrong.** Reproduced live in Cowork on request: asked to append to `research.json` "using the shell or a file write", the agent used a `python open(path,'w')` through `device_bash` and the write reached the user's disk.

The damage is worse than an unvalidated entry. The document **failed schema validation on nine counts** — none of the seven required log fields, three the schema forbids, and an invented `lg_001` where the prefix must be `log_` — and **it locked the project out of the sanctioned tools**, because every writer validates the whole project before persisting. A subsequent ordinary `research_append` returned `ok: false` on that entry. One shell write made the tools that would have kept the file valid unusable until it was removed by hand.

**Why the earlier argument failed**, recorded in the spec rather than deleted because the error is instructive: it rested on "0 landed across the corpus". But every write-shaped shell attempt in the corpus was refused by the **harness's** `dontAsk` permission mode, and the spec already said neither Cowork nor the hosted path has that refusal. The corpus structurally could not record a landed write. That is measuring where the light is.

**What does not change:** a command-text matcher is still the wrong instrument. 37 of 40 corpus touches are reads the system depends on — `research_query` covers 11 of ~15 sections and there is no tree query surface at all — and a denial just moves the agent to `head`, `python`, or a variable-built path.

**Chosen direction, recorded here and not built in this PR:** make the *file* the unit. Keep both project files read-only and re-apply the mode in the shared write layer. Measured on macOS, mode 444 blocks `>`, `>>`, `tee` and `python open(w)` — today's bypass included — while reads still work and both atomic writers still succeed, since they rename a temp over the target.

It is **blocked on one measurement**, deliberately not designed around: Windows `MoveFileEx` over a read-only target is reported to fail where POSIX rename succeeds, which would break every sanctioned write on the platform the genealogist team runs. `dev/probe-readonly-project-files.ts` (added here, verified running, with the macOS baseline) settles it.

**Write-shape matching is held back on purpose** (lead ruling): `sed -i`, `mv` and `chmod` are the rename-shaped residue a read-only file misses, but the corpus cannot say how often they are reached for, so a matcher would be scope bought on speculation. Recorded so the residue is known.

## The parity loader now lifts module-private helpers

The predicate stopped being one self-contained function when this arm landed — it delegates to a payload walk and a basename split. The loader extracted only the named predicate plus literal constants, so keeping it a single body would have meant writing the production code badly to suit a test loader. It now also lifts `_`-prefixed module-level helpers; anything else is still excluded, so a copy cannot smuggle in a dependency on the rest of its module.

## Verification

- **Disabled the arm in one copy and watched three vectors fail**, then restored.
- **Ran the shipped hook directly**, both ways:
  - `device_commit_files` on `research.json` → denied, with the `project_create`-aware reason.
  - `device_commit_files` on `notes.md` whose content says "see research.json" → `{}`, no opinion.
- `make engine-test` 97 files / 2,354 tests · `make harness-test` 2,235 passed, 2 skipped · `make server-test` 162 passed, 1 skipped.
- **Still outstanding: the live `device_commit_files` check.** Nothing in CI sees the real
  bridge payload. The Cowork session run during review exercised `device_bash` (which
  landed — see above) and confirmed a user's own file is untouched, but it never
  reached `device_commit_files`, so this arm's payload extraction is still unproven
  against the real schema.

## Docs brought up to date

`docs/plan/enforcement-layer-phases.md` still claimed no phase had started, which is the exact failure its own status line exists to prevent. It is now a status table; the shipped Phase 2 section is deleted per the doc's own rule; and **Phase 1's section was still describing the auto-seed design that review rejected twice** — it is now the standalone-use remainder, tracked as issue #1695 behind #988.

Phase 3's motivating number is re-measured in place: **52 of the 142 runs that wrote a proof summary never launched the skill that owns it** (37%, against 35% first recorded) — with the caveat stated that every corpus run predates the write-boundary gates, so it re-counts the pre-gate rate rather than testing them.

**Does NOT close #1499.** That issue is the *shell* write routes — sandbox `Bash` and
`device_bash` — and this PR closes neither; it closes `device_commit_files`, which the
issue does not name. #1499 stays open, and the landed bypass above is new evidence on
it: its close-condition ("a bypass appears in a runlog or a feedback case") is now met,
and the direction is the file-mode approach pending the Windows probe.


johnmarkpeterbrown (commented):
## Draft

Two things block, and they're the same root cause: the predicate got the new arm, the wiring didn't.

### 1. The hook is never invoked for `device_commit_files`

`hooks.json` is unchanged in this PR. The script only runs for tools matching `Write|Edit|NotebookEdit`. `device_commit_files` doesn't match that, under either anchored or unanchored matching, bare or namespaced. So the new code never executes in Cowork.

The plugin copy is the only one that ever sees a device bridge. The two host copies register with `matcher=None`, so their arm fires — and neither ever sees a bridge. The route this PR sets out to close is still open.

Your own spec still says why. In `guardrail-enforcement-spec.md`, the callout under the §6 table reads "whose tool names its **matcher** does not cover". That sentence is still true on this branch, 490 lines above the new "Closed 2026-08-17" section.

The verification step is what hid it. "Ran the shipped hook directly" invokes the script past the matcher.

```
You wrote:      "matcher": "Write|Edit|NotebookEdit",
Change it to:   "matcher": "Write|Edit|NotebookEdit|.*device_commit_files",
```

`.*device_commit_files` covers both spellings under either matching mode, and `device_commit_files_v2` still lands on no-opinion. Fix the spec callout in the same commit.

### 2. The test that exists to catch exactly this is a hardcoded list

`plugin-hooks.test.ts` has a check whose comment reads "A matcher that omits a tool the script would have caught is a hole the script can never close". Its tool list is a literal, so widening the script left it green. Extend that list to the two new spellings and it goes red against the current matcher.

Derive the list from the script instead. This file already parses `INCLUDE` out of `package-plugin.mjs` the same way.

```
You wrote:      for (const tool of ["Write", "Edit", "NotebookEdit"]) {
```

```ts
    const guard = readFileSync(GUARD, "utf-8");
    const names = (re: RegExp) =>
      [...(guard.match(re)?.[1].matchAll(/"([^"]+)"/g) ?? [])].map((m) => m[1]);
    const tools = [
      ...names(/FILE_WRITE_TOOLS = \((.*?)\)/s),
      // The bridge tool arrives namespaced; the script matches the bare tail,
      // so the matcher must cover both spellings.
      ...names(/DEVICE_WRITE_TOOLS = \((.*?)\)/s).flatMap((t) => [
        t,
        `mcp__remote-devices__${t}`,
      ]),
    ];
    expect(tools.length).toBeGreaterThan(3);
    for (const tool of tools) {
```

### 3. The payload walk denies a write to the user's own file

This payload gets denied by the shipped script:

```
{"files":[{"path":"/p/.gitignore","content":"research.json"}]}
```

The claim that content is safe because whole basenames are compared holds only when the content has surrounding words. A `.gitignore` listing `research.json` is an ordinary thing to write, and blocking the user's own file is the failure your own asymmetry argument calls the worse one.

Also: no vector exercises the newline bound or `_MAX_PATH_LEN`. Delete both and the suite stays green.

```
You wrote:      for v in value.values():
                    yield from _path_like_strings(v, depth + 1)
Change it to:   for k, v in value.items():
                    if isinstance(k, str) and k.lower() in _CONTENT_KEYS:
                        continue
                    yield from _path_like_strings(v, depth + 1)
```

Beside `_MAX_PATH_LEN`:

```python
# Keys that carry file CONTENT, not a destination. Excluding by key name is safe
# in the one direction that matters: if the bridge renames its content key this
# reverts to today's over-denial, never to a miss.
_CONTENT_KEYS = frozenset({"content", "contents", "text", "body", "data"})
```

Every vector you already wrote still passes with that change. Add the `.gitignore` case so the bound is pinned.

### 4. Stale docstring

`direct_project_file_write` now returns a bridge target in both host copies.

```
You wrote:      """The protected filename a raw file-write tool targets, or None.
Change it to:   """The protected filename this call targets, or None — a raw
                file-write tool by `file_path`, or the device bridge by payload walk.
```

### Two things that aren't findings

**Your Windows probe is answered.** I ran it on win32 / node v24.16.0:

```
  fs.rename over a read-only target    failed (EPERM)
  mode after rename                    444
  open for write ( 'w' )               failed (EPERM)
  open for append ( 'a' )              failed (EPERM)
  read                                 SUCCEEDED
  chmod 644 then write                 SUCCEEDED
```

The pessimistic branch holds. The write layer has to clear the read-only attribute **before** renaming, not only re-apply it after. Worth folding into the spec now.

**`senior-queue` is failing** — it 404s reading this PR's reviews. Not your diff, but it means this PR won't get its `ready-for-senior-developer` label, so a senior may never see it in the queue.

PR #1698, opened 2026-08-17
historical-context: reference expansions (issue #1618)

### LINKED ISSUE #1618: historical-context: reference expansions (LDS migration, FAN affidavits, informal adoption, legitimation, literacy, guardianship, terminology)
**Touches:** packages/engine/plugin/skills/historical-context/references/historical-broad-context.md, packages/engine/plugin/skills/historical-context/references/historical-terminology.md, packages/engine/plugin/skills/historical-context/SKILL.md

From alpha-team review by Clorinda Madsen and Shaunese Luthy, 2026-08-14. Line numbers pinned at `e71a4931`. One skill, one eval run.

## historical-broad-context.md

**Religious factors (`:124-136`)** — the dissenting-groups bullet names Quakers, Mennonites and Huguenots. Add the Latter-day Saints: migrated as organized communities, and kept **separate record systems before Utah became a territory**, which is a different point from the migration one and is the half that tells a researcher where to look. Directly relevant to this product's user base.

**Legal and governmental factors (`:121-122`)** — the guardianship bullet says only that the laws changed. Clorinda: a **natural parent named guardian of their own children** is a common and confusing pattern, driven by inheritance — minors with a claim on an estate needed a legal guardian, so a widowed father petitions for guardianship of his own children and the record looks like an orphan case. Name it, so the agent stops reading a guardianship appointment as evidence the child was orphaned or unrelated.

Issue #1413 is the e2e fixture for the adjacent stepfather-guardian trigger and is deliberately scoped to fixture files only; this is the prose half and a different case.

**Military factors (`:145-146`)** — pension depositions and affidavits are listed as biographical detail about the pensioner. Their larger use is **FAN research**: neighbours, family and friends were called to attest, so your ancestor may appear in a *neighbour's* pension file swearing he has known the man 27 years and farmed next to him. Say to search pension files for the subject as a witness, not only as the applicant.

**Social and cultural factors (`:151-165`)** — three additions:

- Informal adoption and fosterage (`:155-157`) affected how the child was **recorded**: sometimes the head of household's surname, sometimes not; listed variously as boarder, ward, step-child, adopted child, plain child, or by a biological relationship (niece, nephew, grandchild). The current bullet says the practice existed but not what it looks like in a record.
- Illegitimacy (`:158-161`) — if the parents married afterwards the child was **retroactively legitimized**, and the birth entry may or may not carry a notation. Without this the agent treats "base born" as a permanent state.
- Literacy (`:164-165`) — marking a document with an X or another mark did **not** always mean the person could not write. Currently the doc invites the opposite inference.

**One item I am rejecting, recorded so it is not re-filed.** Clorinda asked to change `:112` "church records are the original source for vital events" from "original" to "primary". That is backwards under GPS — sources are original/derivative, information is primary/secondary — and `packages/engine/mcp-server/tests/packaging/gps-terminology.test.ts` fails "primary source" in plugin prose, so the edit would not even land. The report is still a real signal: a working genealogist read "original" as "earliest". Rephrase so the two readings cannot collide, e.g. "before civil registration, church records are usually the earliest surviving record of a vital event, and are original sources."

## historical-terminology.md

- **`:13-25`** — Junior/Senior as community disambiguators. Extend to **II and III**, which worked the same way and were not necessarily generational. Clorinda's case: three adult Caleb Bloods in one area — a father/son pair as Senior and Junior, and a more distant cousin as Caleb Blood III.
- **`:87`** — "Gentleman" is defined by wealth and standing. Make explicit that he **still owned land**; the entry currently reads as a purely social title, and the surrounding land-related entries make the omission conspicuous.
- **Add *feme sole*** — a woman acting for herself in business or at law: either unmarried, or married and acting independently of her husband. Absent from the record-specific terms entirely, and it is the term that explains why a married woman appears as a party in her own right.

## SKILL.md

`:101-102` shows `wiki_search` and `wiki_read` bare. `wikipedia_search` is a sibling tool in the same session, and Shaunese read the bare names as ambiguous. Label them as the FamilySearch wiki tools once in the surrounding prose. Not a rename — the tool descriptions already say FamilySearch, and renaming would touch the manifest, `allToolSchemas`, and every agent tool list under all three server spellings.

## Acceptance

- [ ] Each of the eight additions above is present
- [ ] `:112` rephrased without adopting the forbidden terminology; `make test-js` passes
- [ ] The wiki tools are identified as FamilySearch in the skill body
- [ ] `make eval-skill SKILL=historical-context` re-run and annotated

mercyokum:
**The issue:** During alpha-team review, two genealogists (Clorinda Madsen and Shaunese Luthy) found gaps in the historical-context skill's reference material. It was missing several patterns that come up often in real research: Latter-day Saints migration and record-keeping, natural parents being appointed legal guardian of their own children, pension files as a source of witness testimony from neighbors, how informally adopted or fostered children actually get recorded, retroactive legitimization after a late marriage, and the fact that a signature "X" doesn't always mean someone was illiterate. The terminology reference was also missing "II"/"III" as community disambiguators (not just Junior/Senior), didn't say a "gentleman" still had to own land, and had no entry for "feme sole." Separately, one wording ("church records are the original source for vital events") could be misread as mixing up two different GPS concepts (original vs. derivative sources, primary vs. secondary information), and the skill's instructions named two FamilySearch tools without saying they were FamilySearch tools, which read as ambiguous next to the separate Wikipedia tool.

**The fix:** Added all eight missing points to the two reference files, each with a plain explanation of what to do differently in research because of it (e.g. "search pension files for the subject as a witness, not just as the applicant"). Reworded the church-records line to keep the original/primary distinction correct instead of blurring it. Labeled the two FamilySearch wiki tools as such in the skill's instructions, right next to the general Wikipedia tool, so they're easy to tell apart. Re-ran the skill's full automated test suite (12 test cases) twice to confirm nothing broke — all 12 passed — and had a genealogist grade the results.

Closes #1618.


EdmondOware:
Content looks solid overall, nice fixes on the legitimation wording and the label cleanup.

One more wording note on the new Latter day Saints bullet in historical broad context.md. As written it reads as if the LDS record systems stop being distinct from civil records once Utah became a territory in 1850. In practice they stayed a separate, often more complete, record set for decades after that, since Utah had no statewide vital registration requirement until 1905. Could you adjust the phrasing so it does not read as though the distinction ends at 1850.

Separate note, not a blocker. None of the eight new reference points have a unit test exercising them yet. The rerun confirms nothing regressed, but it does not confirm the skill will actually surface this new guidance in a live scenario. Worth considering at least one test for the guardianship of own children point, since that is the one most likely to cause a real misreading if the model does not apply it.

mercyokum:
Thanks for the review. Both points are addressed.

The Latter day Saints bullet is reworded. It now says the distinction did not end in 1850, and explains that Utah had no statewide vital registration until 1905, so LDS church records stayed the primary record for decades after statehood.

I also added a new unit test for the guardianship of own children point, since you flagged that as the highest risk one. It uses a scenario where a widowed father is appointed guardian of his own son after an inheritance, and checks that the skill explains the inheritance driven pattern instead of assuming the child was orphaned or unrelated. It passed in an isolated run before I added it to the suite.

The full suite has been rerun and annotated with these changes included.

EdmondOware (approved):
Both items are addressed well. The Latter day Saints bullet now correctly says the record systems stayed distinct from civil records for decades after 1850, since Utah had no statewide vital registration requirement until 1905, rather than reading as if the distinction ended at statehood.

A real unit test was added for the guardianship of own children point, with a matching new fixture. I checked the actual run rather than just the description. It passes, and the judges own rationale confirms the skill explains the inheritance law driver, avoids concluding orphaned or not the father, and points the researcher at the mothers probate file. That is exactly what the new reference content was supposed to produce in practice, not just sit unused in a doc.

One unrelated test, the West Virginia boundary one, came back partial this run. I read through it and it looks like ordinary single run model variance rather than anything caused by this PR, and the annotation shows real review happened, two of the judges three low scores were overridden with solid reasoning, and the one genuine gap was left as is rather than waved through.

Small optional note, not a blocker. The new guardianship test was not one of this rounds five sampled tests, so it currently has no human annotation entry of its own, only the judges score. Worth a quick manual confirm in the CRUD UI at some point since it is the test proving out the riskiest new content in this PR, but I already read the judges reasoning directly and it holds up.

Content and process both look solid to me. Approving from my side, over to senior for sign off.

EdmondOware (approved):

Asimi1234:
Reviewed on head `965695de` (1 behind main): clean reference-expansion PR — `check_runlogs` "All runlog rules satisfied", the new guardianship test passes (validators + judge dimensions present) and its fixture is well-formed, the SKILL.md tool labels match CLAUDE.md (`wiki_search`/`wiki_read` = FS wiki, `wikipedia_search` = Wikipedia), and the church-records reword correctly splits "earliest surviving record" from "are original sources."

No blocking findings. The required human annotation is complete — this round's 5 sampled tests are annotated (27 entries, with two judge scores overridden on ut_009) and `check_runlogs` passes.

One optional extra (EdmondOware noted it too), not required and not a blocker: the new guardianship test — the riskiest new content — wasn't among the 5 sampled, so it carries only the judge's score, not a human's. A one-minute CRUD-UI look would add confidence on that specific test.

Genealogical accuracy of the 8 points is the senior-genealogist's call. Pending the senior-genealogist and senior-developer approvals.

Paaboat:
Both of EdmondOware's points were addressed well, and I've checked the content independently. The LDS bullet is now accurate—the record distinction didn't end at statehood, and the 1905 vital registration date anchors that correctly. The guardianship test exercises the right scenario, and the judges' reasoning holds. The eight new reference points are all historically grounded.

One thing before merge: the branch needs to be updated against main. Once that's done, this is clear to merge from my side.

Paaboat (approved):

PR #1703, opened 2026-08-17
viewer: harden #1317 failure modes (blank screen, silent empty folder) — diagnosis + legibility

### LINKED ISSUE #1317: Viewer shows no conflicts, no research question, and loses files between sessions

francis-2008-happy:
Diagnoses and hardens #1317. **Does not close it** — see "What stays open."

## What #1317 actually is (verified against the tester's three bundles)

It's **two separate bugs**, not one:

1. **A resolved conflict was never persisted as a `conflicts[]` record.** In the report-1 project the agent worked through a conflict and wrote its resolution into prose only (log notes, the question's `exhaustive_declaration`, the proof-summary narrative) — `conflicts[]` and `resolved_conflict_ids` are both empty. The viewer reads `research.conflicts`, so it correctly showed "No conflicts recorded." **Not a viewer bug.** This is the failure the shadow detector in #1438 flags; the production fix (agent persists the conflict) is separate work.

2. **The viewer read a different `research.json` than the agent wrote.** The report-2/3 bundle contains two `research.json` files for one project — one empty, one populated with the question + plan. The viewer read the empty one. Root cause: the project folder isn't pinned, so writes and reads can land in different folders (easy to hit when more than one folder is connected to a session). Needs a product decision (tracked with the lead).

So two independent causes surfaced as three reports — which answers the issue's "one bug or three, verify before splitting."

## What this PR changes (failure-legibility hardening, `packages/viewer-ui`)

Neither of the above is fixed here — but both were **invisible** (a blank screen / a section that read as "nothing happened"). This makes the viewer degrade legibly:

- **Never white-screen the whole app on a partial/malformed `research.json`.** Added an `ErrorBoundary` around the active section, and guarded `buildIndex` (which runs in the provider `useMemo`, above the boundary) so a file missing whole sections can't throw. Defensive reads in `ConflictsSection`/`QuestionsSection`.
- **Name the watched folder when there's no `research.json`.** The "no data" state now shows the folder path and says to open the right one if the research is elsewhere — so a folder mismatch (bug 2's symptom) reads as "wrong folder," not "nothing happened."

## What stays open (why this doesn't close #1317)

- **Bug 1** — the agent/skill should persist a resolved conflict into `conflicts[]`. I'm taking this next.
- **Bug 2** — pinning/confirming the project folder is a product decision (lead). Direction so far: when more than one folder is connected, ask "is this the right folder?" This PR only makes the symptom legible.

## Verification

- `packages/viewer-ui` vitest: full suite green; 9 new tests — populated-render regression guards for the Conflicts and Questions sections, malformed-item "does not crash" cases, an `ErrorBoundary` test, and a `buildIndex` partial-file guard.
- Mutation-checked: reverting the sweep/guards makes exactly the intended tests fail.
- Typecheck clean.

## For review
`packages/viewer-ui/**` → senior-developers. No schema, auth, or ADR surface touched. No `apps/electron/src/main` change, so the Electron security checklist has no invariant in play.


Asimi1234:
Reviewed on the tree merged with main (clean, no conflicts despite 224 behind): solid hardening — `viewer-ui` vitest 200/27 and `tsc` clean on the merged tree, and both guards are asserted (reverting the `buildIndex` `?? []` fails the "partial research.json missing whole sections" test; neutering the `ErrorBoundary` fails 2).

No blocking findings. One non-blocker: `buildIndex` still throws above the boundary on `conflicts: [null]` (`Cannot read properties of null`) and on a non-array section `conflicts: {}` (`items is not iterable`) — the sections loop isn't null-safe like the plan-items loop right below it, and `?? []` doesn't cover a non-array. Two-line consistent fix (`(item as { id?: string } | null)?.id` + `Array.isArray(x) ? x : []`); worth folding in since the branch is open, but it doesn't gate.

T-FEH (approved):
Approving. The diagnosis is the most valuable part of this PR — splitting #1317 into "the conflict was never persisted" and "the viewer read a different research.json", and declining to close the issue, is exactly right, and the body doesn't overclaim.

I pushed two commits rather than send it back, because the headline guarantee — *never white-screen on a partial/malformed `research.json`* — didn't hold yet.

**`bac71740` — `buildIndex` still threw on seven shapes.** It runs in the provider `useMemo`, above the boundary, so anything it throws is uncatchable. `?? []` only covers a section that is *absent*. Probed directly:

```
conflicts: [null]        → Cannot read properties of null (reading 'id')
conflicts: {}            → items is not iterable
plans: [null]            → Cannot read properties of null (reading 'id')
plans: [{ items: {} }]   → object is not iterable
questions: [null]        → same shape
gedcomx without persons  → gedcomx.persons is not iterable
gedcomx.persons: [null]  → Cannot read properties of null (reading 'id')
```

The GedcomX block had no guard at all. Added `asArray` (`Array.isArray`, since `?? []` passes a non-array straight into `for...of`) and made every id read optional. @Asimi1234 — this covers both of your cases and four more.

**`d42ed875` — hardening `buildIndex` was necessary but not sufficient.** `App.tsx` renders `Sidebar` (`:127`) and `ProgressPipeline` (`:130`) **outside** the boundary it opens at `:132`, and both reach `.length` on sections they never check. So the exact partial file `buildIndex` now tolerates still blanked the viewer one component over.

Sidebar's own inconsistency was the tell: `known_holdings` and `localities` were already `?.`-guarded, the other eleven were not — and `research?.questions.length` stops its optional chain at `research`, which is the one thing that was never null. All thirteen now match. `progress.ts` had no guard at all, so it got `has()` (`Array.isArray` + non-empty) rather than more optional chaining: a stage is reached only when its section is a non-empty **array**, and a not-yet-an-array section is not a completed stage.

Verified by running, not reading: viewer-ui **213 passed**, `tsc` clean, `turbo typecheck` 5/5. Every new guard is mutation-checked — reverting `asArray` to `?? []` fails 3, dropping the optional `?.id` fails 5, reverting `has()` to bare `.length` fails all 5. None of the 18 new tests is vacuous; the populated-path guards fail when the rendered text is dropped, so they pin real output rather than just "did not throw".

Two things I did **not** change:

- **`App.tsx:70-86` reads wrong on the web client.** `folderPath` there is the session *title*, not a path (`apps/server/app/sessions.py:334`), and `WsResearchTransport` has no `selectFolder` — so "Watching `<title>` … open that folder instead" is both inaccurate and unactionable in the browser. Worth gating the second sentence on the transport actually offering `selectFolder`. Left alone because it's a product-copy call and the Electron path — the one the bug was reported from — reads correctly.
- The new WaitingScreen copy has no test.

One note for whoever picks up the flake: two `FeedbackDialog` tests time out at the 5s default under full-suite load and pass at `--testTimeout=30000`. They fail identically on this branch without my commits, so they are pre-existing — they arrived with #1566, and the sibling `apps/electron` `feedback.test.ts` flakes the same way. Raising the timeout on those two files would stop this being re-diagnosed every PR that touches `viewer-ui`.

jud-sdev (changes requested):
The hardening here is solid and I would approve it on its own. The ErrorBoundary plus the buildIndex, progress, and Sidebar guards genuinely stop the white-screen, the surface that renders above the boundary is fully covered, and the new tests are real (reverting a guard makes them fail). Thank you for the diagnosis too. Splitting #1317 into its two root causes and declining to close it is exactly right.

One change before merge, and it is in the part this PR is actually about: making the failure legible on the web client.

On the web client, the folder path shown in the new no-data screen is not a folder path. It is the session title. The provider fills folderPath from the transport label, and the web server returns the session title there (apps/server/app/sessions.py:334), while Electron returns a real filesystem path. The web transport also does not implement selectFolder. So in the browser, the empty research.json case now shows: Watching <session title>. If your research is somewhere else, open that folder instead. That mislabels a title as a watched folder and tells the user to do something the browser cannot do. And this is exactly bug 2's scenario, so it is the case this screen exists to explain.

It reads correctly on Electron, where folderPath is a real path, which is why it is easy to miss.

The fix is to gate that copy on folder selection actually being available. transport.selectFolder is optional and absent on web, so expose that as a capability from the provider (for example a canSelectFolder flag based on whether transport.selectFolder exists) and in WaitingScreen only show the Watching line and the open-that-folder-instead sentence when it is true. On web, fall back to copy that does not claim to watch a folder or ask the user to open one.

Everything else is good to go. Once the web copy no longer misinforms, then I'll approve.

francis-2008-happy:
Fixed in 7c26b768 — thanks, you're right, that copy misinformed on web.

Gated the folder-specific copy on a new `canSelectFolder` capability from the provider (`typeof transport.selectFolder === 'function'`), so it's true only on Electron. On the web client the no-data screen now shows neutral copy — "No research data yet / the viewer updates when the agent saves research.json for this session" — with no folder name and no "open that folder" instruction. Electron is unchanged (real path, real picker).

Also added `WaitingScreen` render tests for both clients (the web one asserts the session title is never shown as a folder and neither folder sentence appears), since T-FEH noted the copy had no test. viewer-ui suite 215 green, tsc clean.

jud-sdev (approved):
The web-copy fix I asked for has landed and it is exactly right. canSelectFolder is now derived from whether the transport implements selectFolder, and WaitingScreen shows the folder copy only on Electron. On the web client it shows a neutral, accurate message with no folder claim and no instruction the browser cannot carry out.

The new test pins both branches, and the web case specifically asserts the session title is not rendered as a path, which was the exact problem. That also gives WaitingScreen the test it was missing. CI is green on this head.

The hardening itself was already solid, so with the web copy no longer misinforming this is a clear approve. Thanks for the quick turnaround.

PR #1704, opened 2026-08-17
search-records: document matchConfidence range (1–10) (#1005)

### LINKED ISSUE #1005: search-records SKILL.md documents matchScore's range but not matchConfidence's (1–10)
## The gap

`packages/engine/plugin/skills/search-records/SKILL.md:228` describes the
ranked-match fields:

> Each carries `matchRank`, `searchRank` (its original position — shows how
> far the ranker missed), `matchScore` (0–1), `matchConfidence`, the key
> facts, and `attachedToSubject` / `attachedToOther`.

`matchScore` gets its range. **`matchConfidence` gets none.** Per
`docs/specs/rank-search-matches-tool-spec.md:225` it is **1–10**.

## Why that matters

The skill has no way to tell whether `matchConfidence: 4` is high or low,
and the surrounding context pushes it the wrong way: every raw search stub
in the same payload carries its own `confidence` on a much shorter scale
(the `record-search-*` fixtures use 3), so a bare `4` reads as "high."

Observed in `ut_search_records_013`
(`eval/runlogs/unit/search-records/v1_2026-07-30_19-24-10.json`), where the
skill wrote:

> **Match score** | 0.52 (confidence 4) | Moderate score; household anchor
> raises overall confidence

It treated `4` as a signal *offsetting* the middling 0.52. On a 1–10 scale
4 does the opposite — it corroborates the middling score. The run then
graded `Result triage = 1`.

## Fix

Annotate the range, matching how `matchScore` is already documented:

```diff
-`matchScore` (0–1), `matchConfidence`, the key facts,
+`matchScore` (0–1), `matchConfidence` (1–10), the key facts,
```

## Scope note

This is **not** claimed as the sole cause of that test's failure. The same
run also violated four explicit prohibitions in SKILL.md's triage section
(reaching for the "imprecise census ages" excuse, "strong candidate" warm
framing, treating a 3-year birth-year conflict as rounding, and offering
extraction as a question). Those are compliance variance against doctrine
that already exists at length — adding more prose there is not the fix, and
this issue does not propose any.

The missing range is worth closing on its own merits: it is a factual gap in
a field description, four characters wide, and it is currently pointing the
skill's reasoning in the wrong direction.

Landing it flips the `search-records` run log inactive and needs a fresh
`uv run python eval/harness/run_tests.py --skill search-records` plus
annotation — so it should ride along with the next `search-records` change
rather than burning a run of its own.

## References

- `packages/engine/plugin/skills/search-records/SKILL.md:228`
- `docs/specs/rank-search-matches-tool-spec.md:225` (`matchConfidence`: 1–10)
- `packages/engine/mcp-server/src/tools/rank-search-matches.ts:93`
- Found while reviewing PR #929

kofiatinka12:
## Summary

- Documents `matchConfidence`'s **1–10** range in `search-records/SKILL.md`'s ranked-match field list, matching how `matchScore`'s (0–1) range is already annotated right beside it (Trivial tier — one-line doc change).

## Review guidance

**Start here:** `packages/engine/plugin/skills/search-records/SKILL.md` — the `matches[]` field-list line, now `matchScore` (0–1), `matchConfidence` (1–10)`. The asymmetry (matchScore's range documented, matchConfidence's not) was a latent misread: the skill can read a bare `matchConfidence: 4` as "high" — the surrounding raw stubs use a 3-scale confidence — when on a 1–10 scale a 4 is middling. That's the misread #1005 tied to `ut_search_records_013`'s Result triage.

**Honest scope note:** this is a **doc-correctness fix, not a demonstrated test flip.** `ut_search_records_013` currently passes on `main` too — the misread is single-run-variance-dependent, so it doesn't always manifest. The fix removes the latent risk by documenting the range; it does not claim to turn a red test green.

**Unsure about:** Nothing.

## Test plan

Editing `search-records/SKILL.md` flips its run-log inactive and is behavior-affecting (not `eval-cosmetic-skip` eligible). Per @DallanQ's go-ahead, ran a standalone eval rather than continuing to wait for a batchable search-records change (three merged since this was parked on 2026-08-10 without carrying it).

- `make eval-skill SKILL=search-records` → **28/28 pass** (`v1_2026-08-17_18-06-27`), annotated (review_sample covered, rule 3 satisfied).
- An earlier run aborted two tests on `max_turns` (execution variance — both pass on `main`); re-ran once for a clean log, per eval doctrine on aborts.
- CODEOWNERS routes this to **senior-genealogists** (`/packages/engine/plugin/skills/`, `/eval/runlogs/`).

Closes #1005


EdmondOware (approved):
Verified this against source rather than just the PR description. same person dot ts documents confidence as an integer one to ten, omitted on no match, and match score as a float zero to one, so the one line change matches the real code exactly.

Fresh run log v1 2026 08 17 18 06 27 shows 28 of 28 tests passing, including ut search records 013, the test tied to the original misread. The PR is upfront that this test already passes on main and that the fix removes a latent risk rather than flipping a red test green, that framing checks out.

Annotation is complete under the current review sample rule, all five sampled tests covered, and CI is fully green across vitest, pytest, runlogs, and senior queue.

Nothing to push back on content wise. Looks good from my side, over to senior for sign off.

EdmondOware (approved):

florencemashipei (approved):
Verified: the diff is exactly the one line described (`matchConfidence` → `matchConfidence (1–10)`), issue #1005 confirms the gap it closes, and `rank-search-matches-tool-spec.md:234` confirms the 1–10 range is correct. The 28/28 pass run log checks out, and the honest scope note (this fixes a latent misread, not a demonstrated test flip) matches reality — `ut_search_records_013` does pass on main already. CI green. Approving.

PR #1706, opened 2026-08-17
record-extractor: evidence_type is stated-vs-inferred, with no exception (#1628)

Emruthwill:
Closes #1628.

## How to review this

**Read these seven commits, not the file diff.** This branch is based on
`1443-batch-marriage-and-personas` (#1631), which is 22 commits behind `main`,
and it has `main` rebased in — so the aggregate diff shows two other PRs plus a
week of `main`. A reviewer confirmed the real change is 11 files
(`524622cc^..HEAD`). The seven commits below are this PR:

| Commit | What |
|---|---|
| `524622cc` | the doctrine fix + 9 matcher flips + 4 judge_context rewrites |
| `fcfe6c34` | three more instances of the same rule the first sweep missed |
| `a23a2864` | ut_024's parent *relationship* — follow-through on the flips |
| `801c91dc` | the 28-test run log |
| `12dbb632` | its annotation |
| `fc2d4798` | **the `relationship_type` guard** — added after review; catches a silent tree-corruption defect the annotation surfaced |
| `e2bab037` | merge of the base branch, no content of its own |

Merge order: **#1613 → #1631 → this.** The diff collapses on its own as they land.

## The ruling this implements

> Genealogical methodology generally separates these concepts: direct/indirect
> describes whether information answers the research question directly, while
> primary/secondary — or your `informant_proximity` equivalent — describes the
> informant's knowledge or distance from the event.

— the lead, 2026-08-16

## #1628's own diagnosis was wrong, and the corpus proved it

The issue said "correct doctrine stated in six or seven places, applied
inconsistently." It was stated **both ways**, and the skill was obeying the
wrong one. The agent body contradicted itself inside four lines:

> "**Stated-vs-inferred, NOT who reported it.** … **The exception:** a fact
> recorded from a third-party informant relaying another person's facts is
> `indirect` even when stated plainly … On a death certificate the decedent's
> own birth date, birthplace, parents, AND stated age are all `indirect`"

The headline says who-reported is irrelevant; the exception makes it decisive,
broadly enough to cover every death certificate and obituary. Two further
passages stated the exception *as* the rule ("classifies by who reports it on
which record"; "the label follows the record type"). And all of it violated the
body's own opening principle — *"the three layers are INDEPENDENT … never let
one layer's value influence another."*

**The corpus encoded both answers on one axis:**

| Fixture | Same fact | Expected |
|---|---|---|
| ut_017 obituary | `deceased` birth place | `direct` |
| ut_009 death certificate | `deceased` birth place | `indirect` |

Both records state the birthplace in a field. ut_009's own `judge_context` held
both doctrines at once — bullet 9 already said "evidence_type is about whether
the record states the fact, not about the informant" while bullets 6 and 7
required `indirect`. So ut_017 was failing because the skill followed the prompt
correctly and the matcher encoded the opposite rule.

## The mechanism, demonstrated rather than argued

A `--test` sample taken with **three** instances removed and a **fourth** still
present, in one run:

```
deceased  name/age/birth-place/occupation → direct    (contradiction removed)
father    name + birth-place              → indirect  (contradiction survived)
mother    name + birth-place              → indirect  (contradiction survived)
```

Same prompt, same run. Where the contradicting text was gone the classification
flipped; where it survived the model followed it. Removing the last instance
flipped the parents too. **That is the root cause: not a missing rule, but one
rule restated in nine places with one restatement contradicting the others.**

## What changed, on one test throughout — *is there a field holding this value?*

- **Agent body, 6 passages.** The exception is gone; the axis rule is stated
  once, positively. The worked example now shows the same fact `direct` on both
  a census and a death certificate, differing only in the other two layers. The
  self-check asks for a field, not a record type.
- **`rubric.md`, 3 lines** — including the "exception is scoped by RECORD TYPE"
  bullet and the burial-parents line, which had always contradicted the
  marriage-record line two bullets below, where a stated parent name was
  already `direct`.
- **9 matchers** flipped `indirect` → `direct` across ut_009 (stated age,
  birthplace, both parents' names and birthplaces) and ut_024 (both parents'
  names, and both parent-child relationships).
- **4 `judge_context` bullets** rewritten so the judge cannot penalise
  compliance.

**Deliberately not flipped:** ut_009's `birth`+`date`. The certificate states an
age, not a birth date, so the year is computed — genuinely `indirect`. After the
sweep, **every** remaining `indirect` matcher in the corpus (17) is a value no
record states: a computed birth year, or a relationship deduced from household
position. That line is mechanically checkable.

## Evidence

**Run `v1_2026-08-17_18-57-51` — 24 pass / 3 partial / 1 fail of 28**, annotated
(31 dimensions over the 5 sampled tests, no overrides).

| Test | Result |
|---|---|
| **ut_009** | **pass**, zero failing validators — after **0-for-11** across its entire recorded history. Now green in 2 samples + 2 full runs |
| **ut_024** | **pass** — parents' names *and* both relationships `direct` |
| **ut_017** | fail → **partial** |

## What this does NOT show

**A better suite.** Pass count is flat at 24 across three consecutive runs. What
changed is *which* tests pass and why: the composition is cleaner and the
remaining failures are separate defects rather than doctrine contradictions.

## Two defects found and deliberately left

1. **ut_014 — a ROLE-assignment defect, not the birthplace leak I first called
   it.** Corrected after diagnosis; the original claim in this body was wrong
   and is retracted below. What the run actually persisted:

   ```
   head_of_household  Silas Kerrigan   62   Ireland
   child_1            Thomas Flynn     42   Ireland
   child_2            Mary Flynn       38   Ireland
   child_3            Patrick Flynn    15   Pennsylvania
   ```

   **Every birthplace is correct.** The extractor numbered everyone after the
   head `child_1..child_3`, making a 42- and a 38-year-old "children" of a
   62-year-old — exactly what `record-extractor.md` forbids ("an adult too old
   to be the head's child isn't `child_N` of that head; a co-resident family
   keeps its own `head`/`wife`/`child_N`"). This is the differently-surnamed-head
   fixture, so fabricating that structure is the precise failure it exists to
   catch. The birthplace matcher fails only as a *symptom*: it keys on
   `child_1`, which should be Patrick.

   It is also the same family of error #1626 just settled — inventing
   parent-child structure from co-residence.

2. **ut_020 — a genuine birthplace value bleed, one instance.** `child_2`
   (Patrick, age 5) persisted `Pennsylvania` where the record says `Ireland`,
   bleeding from the row below. Roles are correct here.

   **Retraction:** this body previously called ut_014 and ut_020 a "mirror
   image … two tests, opposite directions, which is what points at birthplaces
   being swapped between household members", and I argued that to a reviewer as
   evidence it was not variance. That reasoning treated two unrelated defects as
   one pattern. The value-bleed class has **one** observed instance, not two —
   and the role defect is arguably the more serious of the pair.

2. **`structured_value.relationship_type` disagreeing with `value`** — found by
   the genealogist while annotating: Grace Tolman persisted as
   `relationship_type: 'child'` where the obituary calls her Harold's
   **sister**. The most dangerous of the three classes, because the layers
   disagree *silently*: materialisation reads `structured_value`, so a wrong
   edge lands in the tree while the assertion still reads correctly to a human
   checking `value`. Both the judge and a `value`-only matcher pass it.
   Deterministically catchable, and wants its own change with its own
   before/after over the corpus.


Asimi1234:
Reviewed the real 11-file diff (`524622cc^..HEAD` — the GitHub 263-file view is stale-base pollution from stacking on `1443`, ~16 commits behind main).

Clean, careful PR: the `evidence_type` doctrine flip is internally consistent — grep of `record-extractor.md` shows 0 leftover "indirect because informant" contradictions — both changed tests were flipped to match (burial-index + death-cert, parents/birthplace `indirect → direct` with proximity/quality kept), and no schema/enum file is touched so the enum values are unchanged. The new relationship-type validator is a real guard, proven to fail: it asserts `passed is False` on a category-swap and a direction-inversion, with no-false-positive and fail-open tests; 84 harness tests pass and `check_runlogs` is satisfied.

No blocking findings. One non-blocker: `ut_record_extraction_014` fails in the committed run on a birthplace value (Ireland vs Pennsylvania) — unrelated to this PR's axis and not in the annotated sample — worth confirming it's single-run variance.

Genealogical sign-off on the doctrine is a senior's call, and structurally this waits on base `1443` to land. Pending senior review.

EdmondOware (approved):
Reviewed the actual commits rather than the aggregate diff, since this branch is stacked on unmerged work. One small note before the content review, your table lists five commits but the branch has seven relevant ones. The relationship_type guard commit and the final merge commit are not listed. Worth adding them so the next reviewer does not skim past them.

On the doctrine fix itself, the ruling is correctly applied. Evidence type is about whether a record states the value in a field, informant proximity and information type are a separate axis about the informants distance from the event. That is standard GPS evidence analysis methodology, and the old text genuinely contradicted itself within a few lines, stating the correct rule and then carving out an exception that reversed it. Removing the exception everywhere rather than patching around it is the right call, and it is demonstrated rather than argued, the sample with three contradicting instances removed and one left in place showing the classification flip exactly where the text changed is convincing.

I checked the boundary you drew for the fixture flips. Computed values, a birth year derived from a stated age, correctly stay indirect while stated values move to direct. That line holds throughout the nine flips.

I pulled the run log and annotation directly rather than trusting the summary. Twenty four pass, three partial, one fail out of twenty eight, matches exactly, and the one fail is the birthplace swap defect you disclosed as pre existing and out of scope, not hidden. Annotation is complete for all five sampled tests.

The relationship_type guard commit is good work on its own, a category based check catching a silent tree corruption defect, proven to fail before it passes, harness only so it costs no eval rerun. Glad this landed even though it was not in the original scope.

Content approved from my side. Please add the two missing commits to the review table before this moves further down the stack.

EdmondOware (approved):

Emruthwill:
Thanks both — the commit-table gap is fixed, and the ut_014 question has a firmer answer than "probably variance". Also undrafted: the draft was intentional but the reasoning no longer held, explained at the end.

## The review table was wrong — corrected to seven

You were right, and the cause is worth naming: I wrote that table when the branch had five commits, then landed the guard and a base merge and never updated it. Now:

| Commit | What |
|---|---|
| `524622cc` | doctrine fix + 9 matcher flips + 4 judge_context rewrites |
| `fcfe6c34` | three more instances the first sweep missed |
| `a23a2864` | ut_024's parent *relationship* — follow-through |
| `801c91dc` | the 28-test run log |
| `12dbb632` | its annotation |
| **`fc2d4798`** | **the `relationship_type` guard** — added after review |
| **`e2bab037`** | merge of the base branch, no content of its own |

The intro also said "five commits"; that is fixed too, and now cites your `524622cc^..HEAD` / 11-file framing so the next reviewer starts from the real diff rather than the 263-file view.

## ut_014 — not variance, and I have corrected my own overclaim

The PR body said it "failed identically twice". That was true of what I observed, but **only one of those two run logs is committed** — I deleted the other before committing so that only one `v{N}.json` is added per rule 1. So the claim was not verifiable from the repo, and I have rewritten it to say exactly what is.

What the committed corpus does support, which I think settles your question:

```
v1_2026-08-16_11-26-54   ut_020   got 'Pennsylvania'   expected 'Ireland'
v1_2026-08-17_18-57-51   ut_014   got 'Ireland'        expected 'Pennsylvania'
```

**Two different tests, opposite directions.** Random sampling noise does not produce a clean mirror image across two fixtures — that shape points at birthplaces being swapped between household members. Add the run I did not commit, where ut_014 failed identically, and I would call it a real extraction defect rather than a draw.

It is out of this PR's axis and outside the annotated sample, so I have left it. But I would not want it closed as variance on the strength of one green run later.

## On the guard landing outside original scope

Fair to flag, so the provenance is on the record: it came from the genealogist's annotation comment on ut_017, not from planned work. Replayed over every committed run log it finds **7 real instances across ut_017 and ut_027 in 3 separate run logs**, in two shapes — a category swap (`rt=child` / `value='sister of…'`) and a **direction inversion** (`rt=parent` / `value='child of Louise Becker'`), which writes the family edge backwards.

It is compared by **category**, not literally, and that is the whole design: `rt: "spouse"` beside `"wife of George Bennett"` is correct, and a literal comparison flags 39 assertions of which only 7 are real. It fails **open** on anything it cannot compare — an unrecognised `relationship_type`, a value naming no relation — because a false positive costs the test's entire grade when a failing validator suppresses the judge. That caution is from earlier in this same stack, where the first name-value guard false-positived on surnames like `Joseph Parent of Quebec` and a reviewer caught it.

## The draft question

Intentional, and now reversed. I drafted it because the aggregate diff shows 263 files from stacking on a base that is 22 behind `main`, and I did not want a reviewer opening it into two other PRs' changes. But that is an argument for explaining how to review it — which the body does — not for hiding finished work, and #1613 and #1631 are both already non-draft. Marked ready.

Emruthwill:
## Retraction: my answer on ut_014 was wrong

@clack391 asked whether ut_014's birthplace failure was single-run variance. I answered that it was not, citing ut_020 as a "mirror image — two tests, opposite directions". I have now actually diagnosed it, and **that answer was wrong**. Correcting it here and in the body rather than leaving it standing.

### ut_014 is a role-assignment defect. Every birthplace in it is correct.

```
head_of_household  Silas Kerrigan   62   Ireland
child_1            Thomas Flynn     42   Ireland
child_2            Mary Flynn       38   Ireland
child_3            Patrick Flynn    15   Pennsylvania
```

The extractor numbered everyone after the head `child_1..child_3` — making a 42-year-old and a 38-year-old "children" of a 62-year-old. `record-extractor.md` forbids exactly this:

> "Don't number everyone after the head `child_1, child_2, …` — that fabricates a parent-child link the record never states … an adult too old to be the head's child isn't `child_N` of that head. A co-resident family keeps its own `head`/`wife`/`child_N`."

This is the **differently-surnamed-head** fixture: Silas Kerrigan heads a household containing the Flynn family. Thomas and Mary are a co-resident couple and Patrick their child, so `child_1` should be Patrick. The `value: 'Pennsylvania'` matcher fails only as a **symptom** of the shifted roles — the birthplaces themselves are all right.

The fixture's own `judge_context[5]` names this: the fabrication ban covers "the intra-family household-position inferences (Thomas-Mary as a couple, Patrick as their child)". Fabricating the structure is the failure it exists to catch.

It is also the same family of error #1626 settled — inventing parent-child structure from co-residence.

### ut_020 is a real value bleed, and it is one instance, not two

```
child_2   Patrick Flynn   5   Pennsylvania    <- record says Ireland
child_3   John Flynn      2   Pennsylvania    <- correct
```

Roles are right; Patrick's birthplace bled from the row below.

### What was wrong with my reasoning

I treated two unrelated defects as one pattern and used the pairing as the argument that it was not variance. The pairing does not exist. So:

- the **value-bleed** class has **one** observed instance (ut_020), not two — I cannot tell you from this evidence whether it is variance
- the **role-assignment** defect (ut_014) is separate, and on a fixture specifically built to catch it

Your original instinct to ask for confirmation was right, and my confirmation was not worth much. Neither is in this PR's scope and neither is in the annotated sample, so I am leaving both — but with an accurate description this time.

Emruthwill:
@DallanQ — on the base branch: **yes, it is intentional**, and here is the reasoning plus the risk you are pointing at, which is real for a reason worth naming.

## Why it targets #1631

Against `main` the diff is **262 files** — two other PRs plus a week of `main`, because #1631 is behind. Against #1631 it is the **11 files** that are actually this PR, which @francis-2008-happy independently confirmed (`524622cc^..HEAD`). Retargeting now makes it unreviewable until #1631 lands.

## The risk, stated precisely

"Changes will be lost" is not quite the mechanism — the commits stay on this branch, and GitHub retargets a child PR when its base merges. But the outcome does depend on **how #1631 is merged**, and `main` shows both styles in use:

```
Merge pull request #1726 from PioneerAIAcademy/1600-abort-storm-breaker   <- merge commit
feat: suite-level abort-storm breaker (#1600)                            <- squashed
```

- **Merge commit** — SHAs survive, this PR auto-retargets to `main`, diff collapses to its own 11 files. Clean, nothing to do.
- **Squash** — `main` gets one new commit with a new SHA. This branch still holds #1631's original commits, so after retargeting git sees them as absent from `main` and the diff tries to replay #1631's content. That needs a rebase and can conflict.

So the hazard is specific to a squash merge of #1631, and it is a rebase rather than data loss.

## What I propose

**Leave the base as #1631 until it merges, then retarget to `main` and rebase this branch onto it.** That keeps the reviewable 11-file diff now and removes the squash hazard at the point it would arise. I will do the rebase as soon as #1631 lands and confirm the diff collapses to 11 files.

If you would rather remove the ambiguity immediately, say so and I will retarget to `main` today — nothing breaks, the diff is just noisy until #1631 merges.

**One thing not to do:** merge this before #1631, which would land it in #1631's branch rather than `main`.

Also for the record: **#1613 is now closed** in favour of #1631, verified first — 0 commits missing, and both its run log and annotation are committed on #1631's branch.

kofiatinka12 (approved):
The doctrine this implements is correct standard methodology, and the old rule it replaces was the actual error. Evidence type (direct/indirect) turns on whether the record states the value or you had to infer it; the informant's distance from the event belongs to the information layer (primary/secondary), not the evidence layer. A widow's report of her husband's birthplace on a death certificate is direct evidence carried by secondary information — that is the ordinary case, not a demotion. The old "death-certificate exception" pushed an information-quality concern into evidence_type, which is exactly the conflation the GPS warns against, and it contradicted the agent body's own independence-of-layers rule. Removing it loses no genuine distinction — the reliability concern is now filed correctly under information_quality/informant_proximity.

I checked the fixture flips against the doctrine and they are surgically right. The death-certificate matchers keep the stated age, birthplace, and parents' names and birthplaces as direct while leaving the computed birth date/year as indirect — that "stated birthplace direct, computed birth date indirect, same person" split is precisely the correct treatment. The burial-index change (parents named → direct, quality indeterminate for the unknown informant) follows for the same reason, and every remaining indirect matcher is a genuinely derived value. Two things I'm noting rather than blocking on, both outside the genealogy: this is a stacked PR, so it can't merge until #1613 and #1631 land ahead of it (merge order #1613 → #1631 → this); and the developer-lane files that ride in from the base branches still need the senior-developer sign-off — my approval covers the record-extractor doctrine, rubric, and fixtures only. The three deferred defects (ut_014 role assignment, ut_020 value bleed, and the relationship_type disagreement) are appropriately scoped out and documented.

Issue #1713, opened 2026-08-17
LocalProvider._kill_server uses POSIX-only os.killpg, crashing (and orphaning sandbox servers) on Windows


taiwo-stack:
**Touches:** apps/server/app/sandbox/local.py

Found via `make test-all` while preparing an unrelated PR (#1191) — not caused
by that branch; reproduces identically on `main` (`cd apps/server && uv run
pytest -q`, native Windows).

`LocalProvider._kill_server` (local.py:227-236):

```python
async def _kill_server(self, proc: subprocess.Popen) -> None:
    if proc.poll() is None:
        try:
            os.killpg(os.getpgid(proc.pid), signal.SIGTERM)  # kills server + agent (new session)
        except (ProcessLookupError, PermissionError):
            proc.terminate()
    try:
        await asyncio.to_thread(proc.wait, 5)
    except Exception:
        proc.kill()
```

`os.killpg` and `os.getpgid` are POSIX-only — absent from the `os` module
entirely on Windows. The call raises `AttributeError`, which the `except`
clause doesn't catch (it only catches `ProcessLookupError`/`PermissionError`),
so the fallback `proc.terminate()` is never reached and the spawned
`python -m app.sandbox_server` subprocess (local.py:220-223, started with
`start_new_session=True`) is never actually killed on Windows. It's orphaned,
holding its port open.

**This is a single root cause for 8 failures and 2 indefinite hangs** in
`apps/server`'s pytest suite on native Windows (confirmed by isolating each
one and reading the traceback — every failure bottoms out in this same
`AttributeError`, and the two hangs are consistent with a subsequent test
trying to bind/connect to a port an orphaned prior-test server never released):

- `test_agent_secrets.py::test_connect_rewrites_the_secrets_file`
- `test_sandbox_server.py::test_local_connect_unified_path`
- `test_sandbox_server.py::test_local_connect_waits_until_ws_server_accepting`
- `test_v1_api.py::test_sync_message_returns_assembled_reply`
- `test_v1_api.py::test_stream_message_done_text_equals_sync`
- `test_v1_api.py::test_sync_turn_timeout_returns_504`
- `test_v1_api.py::test_stale_lock_is_reclaimed`
- `test_v1_api.py::test_delete_releases_sandbox_and_404s_after`
- `test_v1_api.py::test_sync_turn_reports_tool_calls` (hangs indefinitely, no timeout)
[...]

Issue #1707, opened 2026-08-17
feedback.test.ts assumes POSIX paths for the project-hash lookup, fails on Windows


taiwo-stack:
**Touches:** apps/electron/src/main/__tests__/feedback.test.ts

Found via `make test-all` while preparing an unrelated PR (#1191) — not caused
by that branch; reproduces identically on `main`.

`points at the transcript when a Claude Code log is present` (feedback.test.ts:386)
fails on native Windows with:

```
Error: ENOENT: no such file or directory, mkdir
'C:\Users\...\feedback-home-X\.claude\projects\-C:\Users\...\feedback-slog-Y'
```

The test builds the project-hash directory name at line 393 assuming a POSIX
absolute path:

```ts
const projectHash = folder.replace(/^\//, '').replace(/\//g, '-')
const projectDir = join(fakeHome, '.claude', 'projects', `-${projectHash}`)
```

On Windows, `folder` is a drive-letter path (`C:\Users\...`), so neither
`replace` does anything — there's no leading `/` to strip and no `/` to turn
into `-`. The resulting `projectHash` is the raw Windows path, complete with
its own colon and backslashes, which then gets nested inside another `join`
and produces an invalid, unwritable path.

This is the same class of bug issue #1420's closing section anticipated and
explicitly said to file separately rather than fold in: "Path separators...
is [a] recurring class. No open issue currently represents [it]; file
separately if one bites."

**Fix sketch (unverified, for whoever picks this up):** whatever production
code path this test is modeling (`readSessionLog`, per the comment at line 387)
presumably already has to turn a project folder path into Claude Code's
project-hash directory name correctly on Windows, since that's how Claude Code
itself names `~/.claude/projects/<hash>` dirs on disk. The test's hash
derivation should match that logic (likely: normalize backslashes to `-`
first, not just forward slashes) rather than reimplementing a POSIX-only
version inline. Confirm against how the real `readSessionLog` derives the hash
before changing the test.

**Acceptance:** `npx vitest run src/main/__tests__/feedback.test.ts` passes on
native Windows (verified failing 2026-08-17) without changing behavior on
macOS/Linux.

DallanQ:
Merged into issue #1420 (Windows portability), carried over in full — the ENOENT trace, the line 393 analysis, and the fix sketch pointing at `readSessionLog`.

[...]

PR #1714, opened 2026-08-17
Lower the female late-childbearing check from a 55- to a 45-year gap (#1191)

### LINKED ISSUE #1191: Candidate parents are asserted without checking them against the ages of children already in the tree
> **Reviewed 2026-08-12 before junior handoff.** Both decisions are settled: the female
> late-childbearing check is **lowered from 55 to 45**, stays female-gated, stays
> `implausible`, and stays a **single** check — no second banded check, no promotion to
> the impossible tier. This block supersedes the sections "Read this before scoping",
> "Cost" and "Sequencing" further down, which contain a false claim and a wrong cost.
> Read this block instead of them; the original text is kept for the record.

## Read this before scoping — re-verified on `main` 2026-08-12

**Holds — run 2's failure is computable today.** `earliestChildBirthToBirth12`
(`packages/engine/mcp-server/src/tools/person-warnings.ts:1099`, predicate at `:289`)
fires on run 2's committed final tree: the subject carries Birth and Christening
`28 September 1810` and her earliest child is christened `16 Dec 1816`, a 6-year gap.
Run 2 never called `person_warnings` — 0 calls in `run-2026-08-02_12-17-10.json`.

**Does not hold — "there is no upper-bound check anywhere in the issue-type list."**
There are two, both built on `latestChildBirthToBirth` (`:406`):

- `checkLatestChildBirthToBirth80` (`:1160`) — any gender, gap >= 80
- `checkLatestChildBirthToBirthFemale55` (`:1231`) — female, gap >= 55

plus the relative-mob twin `checkFemaleRelativesLatestChildBirthToBirth55` (`:1725`),
which carries the same 55. All three are catalogued in
`packages/engine/plugin/skills/check-warnings/references/warning-checks.md:98-99`.
Run 1's subject is Female with a gap of exactly **51** (christened 1786, last child
christened 8 Oct 1837), so she clears both cutoffs and nothing fires.

**Run 1 also refutes the "nobody ran the tool" reading for its own case.** It called
`person_warnings` on the subject and on both candidate parents *after* the write
(`run-2026-08-01_20-37-47.json`, tool calls 165-167) and got back one warning,
`maleRelativesHasDiffSurname`. Running the shipped `calculateWarnings` over that
committed final tree reproduces exactly that (measured 2026-08-12).

**Both runs wrote the parent link with `tree_edit`, not `research_append`** — run 1
call 161, run 2 calls 173-174, `add_relationship` ops. Anything that binds a check to
the moment of the write binds there.

## What this issue does not own

The "make the check run at the moment of the write" half is **not** this issue.
PR #1541 (merged 2026-08-11, issue #1193) landed the measurement half — the shadow
detector `find_relationship_writes_without_warnings_check`
(`eval/harness/harness/skill_invocation.py:1127`), which flags a run that wrote a new
ParentChild/Couple relationship with no successful `person_warnings` call — and
deferred the behavioral half to the lead.
`docs/adrs/ADR-0011-put-guardrails-at-the-write-boundary.md` requires the
satisfying-shape rate to be measured before such a gate ships
(`make e2e-guardrail-shadow SINCE=all REPLAY=1`, no API spend). Do not build it here.

## The ruling — one female check, lowered to 45

Settled by the lead 2026-08-12, and revised the same day after the gender-neutral option
was measured. **Lower the existing female late-childbearing check from a 55-year gap to
45. One check, not two.**

- Cutoff `>= 45`, not `> 45`. The whole `latestChildBirthToBirth` family is `>=`
  (`person-warnings.ts:406`); the `> 50` in `person-warnings-tool-spec.md`'s
  future-candidates table is prose from before the `warnings.java` port.
- The tag is renamed to `latestChildBirthToBirthFemale45` (and the relative twin to
  `femaleRelativesLatestChildBirthToBirth45`) — the family's names encode the cutoff.
- Severity stays **`implausible`** (`warning` before PR #1545). Promoting it to
  `contradiction` was considered and declined; see below.

**Female-gated, not any-gender — this is measured, do not revisit it.** The check keeps
its `if (mob.getGender() !== "Female") return null` guard. Applying 45 to everyone was
proposed and rejected on the numbers: across the 150 committed
`eval/runlogs/e2e/*/*.final-tree.gedcomx.json` (77 fixtures, 408 evaluable
parent-instances), a female-gated `>= 45` flags **6 distinct people**, while a
gender-neutral `>= 45` flags **31 — of which 25 are men**, every one a father with a
child at 45 to 74. That is unremarkable in any period, and the male side is already
covered by `latestChildBirthToBirth80` (any gender, `>= 80`). Measured 2026-08-12 by
replicating the shipped predicate; re-run it at a new cutoff to refresh.

**No banding, and no second check.** An earlier draft of this ruling kept 55 and added a
second check banded to `[50, 55)`. That is superseded. With one cutoff there is one tag,
one message, and no risk of two checks reading the same predicate and firing twice for
one fact — which would have double-counted in the skill's cluster rule ("2 or more … on
the same person is a strong signal that records from two different individuals have been
merged"). If a distinct top band is ever wanted, split the tag then; do not build the
path now.

**The check keeps `implausible` — do not promote it to `contradiction`.** Settled by the
lead 2026-08-12. Promotion would make the check **merge-blocking** in
`proof-conclusion/SKILL.md` and `person-evidence/SKILL.md` and feed the "2+
contradictions => recommend splitting the profile" rule. Declined because the check fired
0 times at 55 across the committed corpus, so the promotion is a doctrine commitment with
no measured effect, and because `contradiction` is defined as "data that cannot be true
as recorded" — a predicate that counts a proxy `Baptism` as birth-like and adds ten years
of fudge for `Est`/`Aft` dates cannot support that bar. Record this, and the promotion it
beat, in `docs/specs/person-warnings-tool-spec.md` beside the check's description — that
is part of this task.

**What the lowered check will fire on.** Nothing fires at 55 today; at 45 the female
check flags 9 person-instances across 6 distinct people. Eyeball these before opening
the PR — they are the entire observable effect of the change. Three are the 50+ band:

- `elisabetha-sugecz-parents` / `G4C9-Y6C` Elisabetha Sugecz — gap 51. This issue's
  run 1. Known wrong.
- `anders-monsen-ancestry` / `KZHH-VTX` Unna Halsteinsdatter — gap 51, in two runs.
- `esther-leroux-marriage` / `L7B2-HQ8` Domitille Preise — gap 53.

The remaining three sit in the newly covered 45-49 band and are unadjudicated. If any is
a correct parentage, it is the check's first false positive, and that is worth knowing
before it ships rather than after.

## Nothing makes the agent act on this — say so in the PR

**Read this before you estimate the value of the change.** No line of code anywhere
branches on `severity`. The only read in the engine is a dedup key
(`merge-warnings.ts:107`); `validator.ts:944` is the unrelated timeline-gap severity.
Every consumer is a prompt:

- `check-warnings/SKILL.md` — the severity icon, the cluster count, and the
  interpret-and-recommend section
- `check-warnings/references/assumption-categories.md` — "Look at `severity`" is step 1
  of its decision procedure
- `check-warnings/references/warning-checks.md` — the catalog is split by severity
- `check-warnings/references/warnings-as-identity-signals.md` — the escalation table
- `proof-conclusion/SKILL.md` (person-merging paragraph) and `person-evidence/SKILL.md`
  (the coherence-gate step) — these two give the levels different *procedural*
  consequences: the impossible level blocks a merge, the improbable level is advisory
- `eval/tests/unit/check-warnings/rubric.md` — grades severity classification

The e2e shadow detector `find_relationship_writes_without_warnings_check` checks only
that `person_warnings` was *called*; it never reads severity.

So the escalation table in `warnings-as-identity-signals.md` is what decides whether this
check changes any outcome — and it currently reads **"1 implausible on its own: note and
move on"** and calls a lone implausible warning on an otherwise clean profile "usually
noise." Run 1's subject carried exactly one other warning. **A green `make engine-test`
proves the warning is emitted, not that either failure in this issue is fixed.** Put that
sentence in the PR body.

Changing that escalation table is doctrine prose — `genealogist` lane, a different issue,
not this one (`docs/skill-lifecycle.md`, the lane rule). Do not edit it here beyond the
factual band correction listed below.

## Cost — corrected

**Not free.** A new or renamed `issueType` must be added to
`packages/engine/plugin/skills/check-warnings/references/warning-checks.md`, which is the
authoritative source for per-check severity. **Any** touch under
`packages/engine/plugin/skills/**` arms `.github/workflows/check-runlogs.yml`, which
blocks the PR until the newest `check-warnings` run log's snapshot matches your branch
and its `.ann.json` carries a correction for every dimension of every test.
`check-warnings` is not exempt (`eval/harness/scripts/check_runlogs.py:77` — only
`research` and `forget-and-rederive` are). Budget one
`make eval-skill SKILL=check-warnings` run plus a genealogist annotation through
`make eval-ui`: the newest committed `check-warnings` log on `main` cost **$1.72 and
14.6 min** wall clock (`totals.total_cost_usd` / `totals.wall_clock_ms` in
`eval/runlogs/unit/check-warnings/v1_2026-07-22_10-11-56.json`, read 2026-08-12), and the
five committed logs span $1.48-1.94 and 10-17 min.

PR #1545 lands a fresh `check-warnings` run log (`v1_2026-08-11_14-28-07`). Your catalog
edit invalidates that one too — the snapshot covers
`packages/engine/plugin/skills/<skill>/**`, references included
(`eval/harness/harness/snapshot.py`, `build_snapshot`). The engine half of this PR
carries no eval cost; `mcp-server/src/**` is deliberately outside the snapshot.

## Sequencing — corrected

Do this **after** PR #1545 (issue #1194) merges — it rewrites all 77 severity lines in
`person-warnings.ts` and renames `error`/`warning` to `contradiction`/`implausible`, and
it has since grown to also replace the `check-warnings` and `proof-conclusion` run logs.
As of 2026-08-12 that PR is open, review-required and behind `main`, so confirm it has
landed before branching.

## Sites this touches — lowering one cutoff, not adding a check

Line numbers are pre-PR #1545, which rewrites 77 severity lines in
`person-warnings.ts`. After it merges, grep the symbol names instead.

**The two tag strings have exactly five references between them** (verified
2026-08-12, `git grep latestChildBirthToBirthFemale55` and
`git grep femaleRelativesLatestChildBirthToBirth55`) — two in the engine, three in the
check-warnings references. Nothing else in the repo names them.

- [ ] `packages/engine/mcp-server/src/tools/person-warnings.ts`
  - [ ] `LATEST_CHILD_BIRTH_TO_BIRTH_FEMALE_55` (`:169`) — constant name **and** its
        string value, both to `...45`.
  - [ ] `FEMALE_RELATIVES_LATEST_CHILD_BIRTH_TO_BIRTH_55` (`:196`) — same.
  - [ ] `checkLatestChildBirthToBirthFemale55` (`:1231`) — the `latestChildBirthToBirth(mob, 55)`
        cutoff, the function name, and the message string ("55 or more years after her
        own birth" -> 45).
  - [ ] **The relative-mob twin** `checkFemaleRelativesLatestChildBirthToBirth55`
        (`:1725`) — same three. Every check in this family has a twin; a missing twin is
        silent, and this is the single most likely thing to be left behind.
  - [ ] The call sites in `calculateWarnings` (`:3241` person-level, `:3316`
        relative-mob) and their local variable names (`female55`, `femRelChild55`).
        **No new registration is needed** — both are already wired; this is a rename,
        so the compiler will find them.
  - [ ] **Amend the comment at `:153`** — "These strings match warnings.java exactly so
        the TS port emits the same `issueType` identifiers a Java caller would" stops
        being true the moment a cutoff diverges. Say which tags are ours and why, in one
        line. Nothing checks this, so the next reader will assume the tag set still
        matches the port.
- [ ] `packages/engine/mcp-server/tests/tools/person-warnings.test.ts` — **there is no
      emitter test for this check today** (`git grep Female55` on this file returns
      nothing; only the shared predicate is tested, and only at cutoff 80). Add, for
      both the person-level check and the relative twin: gap 44 -> silent; gap 45 and
      gap 60 -> the tag fires. Add a Male person at gap 60 -> silent, which is the
      gender gate and the thing this ruling turned on. Per `CLAUDE.md`, change 45 back
      to 55 and watch the gap-45 test fail before you commit.
- [ ] `packages/engine/plugin/skills/check-warnings/references/warning-checks.md` —
      `:99` ("mother was age 55 or older at a child's birth") and `:149` (the relative
      tag in its list). **This is the edit that arms the run-log gate — see Cost.**
- [ ] `packages/engine/plugin/skills/check-warnings/references/warnings-as-identity-signals.md`
      — `:97` names the tag in a list; and the line reading "Parent-child age gap:
      typically 12-55 years for mothers" is wrong once the cutoff is 45. Fix both. Do
      **not** touch the escalation table below it — that is the genealogist lane.
- [ ] `docs/specs/match-merge-workflow-spec.md`, "Severity catalog (merge-relevant)" —
      only if the tag is named literally there. The Advise line covers "the parent-age …
      checks" as a family, which stays true either way.
- [ ] `docs/specs/person-warnings-tool-spec.md` — strike the `MOTHER_TOO_OLD` row from
      "Future warning candidates"; it is built now. Record the 45 decision and the two
      options it beat (a gender-neutral 45, and keeping 55 with a second banded check)
      beside the check's description. **Read the rest of this file, do not follow it:**
      its "How to add a new warning" recipe and its UPPER_SNAKE tag convention predate
      the `warnings.java` port and no shipped check uses them.
- [ ] Nothing in `eval/fixtures/mcp/` and nothing in `eval/tests/unit/` — verified by
      the same two greps. No fixture or eval test asserts this tag family.
- [ ] Nothing in `packages/schema/`, `packages/viewer-ui/`, `apps/web/`,
      `apps/electron/` or `apps/server/` — the `severity` those carry is the
      timeline-gap `high|medium|low`, an unrelated field.

**Verification:** `make engine-test` for the unit cases, then
`packages/engine/mcp-server/dev/try-person-warnings.ts` against a project whose tree
reproduces a 51-year gap. Then the sentence from the section above: a green suite proves
the warning is emitted, not that the agent acts on it.
---

## Original issue

**IMPORTANT**: Do PR #1545 (issue #1194) first — it rewrites `person-warnings.ts` and
renames the severity values from `error`/`warning` to `contradiction`/`implausible`.
The new checks this issue asks for have to set whichever names land there.

## The finding

Two e2e runs of `elisabetha-sugecz-parents` asserted parents for G4C9-Y6C that
are chronologically impossible against **children already present in the starting
tree**. They failed by different routes, so this is not single-run variance —
it is the same missing check twice.

**Run `run-2026-08-01_20-37-47`** (annotation: `f1: false`, `f2: false`,
`proof_quality_score: 1`)
Asserted the hinted couple — R20/R21 make Thomae Sugecz and Susanae Petrich
parents of G4C9-Y6C, sourced to the 1786 Madzarevo baptism
(`ark:/61903/1:1:QKMN-XYBK`) at quality 3, neither flagged hypothetical nor
rejected. The run **had already noticed the problem** and argued past it —
`log_014` reads: *"a 1786 birth would make her 51 at this child's birth."*

**Run `run-2026-08-02_12-17-10`** (annotation: `f1: false`, `f2: false`,
`proof_quality_score: 1`)
Asserted a *different* wrong couple — I1 "Thomae Sugecz" / I2 "Hellenae" from
the 28 Sep 1810 baptism (`ark:/61903/1:1:QKMN-FFMN`), with the proof summary
calling the father "ESTABLISHED". The narrative states *"No record contradicts
this identification"* — while the starting tree it was handed carries **nine
children's baptisms sourced to the subject between 1816 and 1837**. A 1810 birth
puts her at 6 at the first of them. The contradicting evidence was in hand and
never examined.

Both runs and their annotations are on `main` — PR #1113 merged 2026-08-04.
Five files each under `eval/runlogs/e2e/elisabetha-sugecz-parents/`:
`run-<ts>{.json,.ann.json,.final-tree.gedcomx.json,.final-research.json,.transcript.md}`.
Both carry a `.transcript.md` sibling, which most fixtures do not — that is
where run 1's `log_014` reasoning is readable without parsing JSON.
Annotator: @taiwo-stack.

## Lane: mechanical check (`developer`), chosen 2026-08-10

This was filed unlabeled because the lane was genuinely open. It is now settled
as the mechanical reading, and the issue is labelled `developer` accordingly.

**Reading 1 — mechanical check.** The inputs are already in the project: the
candidate parent's implied birth year, and the birth/baptism dates of the
subject's existing children. A `person_warnings`-style computation can emit
"candidate parent implies subject b. YYYY; subject has a child b. YYYY (age N)"
without any prose change and without a paid run. This is the cheaper lane and it
cannot be argued past — run 1 proves prose noticing is not enough.

**Reading 2 — skill doctrine (`genealogist`).** Not chosen, recorded so it is
not re-derived. The failure is not that the agent lacked the arithmetic; run 1
did the arithmetic out loud. It is that nothing made the result **binding** —
a doctrine question about what a candidate-identity assertion must clear before
it can be written. The mechanical check can feed a doctrine change later; it
does not need one to land.

## Read this before scoping: half the computation already exists

Verified on `main` (`2dcd1096`). An earlier comment on this issue said nothing in
`person_warnings` compares a candidate parent's implied birth year against the
birth dates of children already in the tree. **That is wrong in the direction
that makes this cheaper.** `packages/engine/mcp-server/src/tools/person-warnings.ts`
already carries the comparison:

- `earliestChildBirthToBirth(mob, cutoff)` (`:289-294`) subtracts the person's
  earliest birth-like year from their earliest child's earliest birth-like year.
- `checkEarliestChildBirthToBirth12` (`:1099`) fires at age ≤ 12 for anyone;
  `checkEarliestChildBirthToBirthMale14` (`:1025`) fires at ≤ 14 for males.

So **run 2's failure is already computable today**: a 1810 birth against a first
child in 1816 is age 6, well inside the 12-year cutoff. What is missing is not
the arithmetic — it is that nothing runs it at the moment the parent link is
asserted.

**Run 1's failure is not covered.** `earliestChildBirthToBirth` tests only
`<= cutoff`, i.e. the too-young direction. There is no upper-bound check anywhere
in the issue-type list, so a candidate implying the subject bore a child at 51
fires nothing.

That splits the work in two, and both halves are small:

1. **Wire the existing too-young check into the assert path** so the warning is
   produced when the candidate parent is proposed, not only when someone calls
   `person_warnings` on a finished tree.
2. **Add the upper-bound counterpart** — a late-childbearing check with its own
   issue type and cutoff. Pick the cutoff against the existing checks' style
   rather than inventing a new convention.

Confirm both claims yourself before branching; do not take this section on trust.

## Cost

The mechanical lane costs nothing — no skill body changes, so no run log flips
inactive and no paid eval run is bought.

For reference if anyone later argues for the prose lane: a skill-body edit flips
that skill's run log inactive (`check_runlogs.py` rule 2) and needs a fresh
`make eval-skill SKILL=<name>` run plus a genealogist annotation. Ruth's
committed search-records run log (`v1_2026-08-01_13-11-14.json`) puts that at
**$6.72 and ~65 min** of machine time for one skill.

## Sequencing

The two collision PRs this body used to name are both merged — PR #1190
(proof-conclusion) on 2026-08-03, PR #1097 (search-records) on 2026-08-04. With
the mechanical lane chosen there is no skill body in the diff, so there is
nothing left to sequence against.

## Do not re-derive

- The two annotations above are already written and correct — read them, do not
  re-grade the runs.
- Both are flagged low-confidence for judge **calibration** under issue #1114
  (the harness printed the verdict before grading). That does not weaken this
  finding: the failure is read from each run's final tree, not from the judge's
  grade.

taiwo-stack:
## Summary

- Lowers `checkLatestChildBirthToBirthFemale55`'s cutoff from a 55- to a 45-year gap (renamed to `...Female45`, along with its relative-mob twin) — **Normal tier** (docs/task-lifecycle.md).
- Per the lead's final 2026-08-12 ruling on #1191: one female-gated check, severity unchanged, no banding, no promotion to the impossible tier.
- Adds the emitter tests this check never had, updates the three `check-warnings` reference docs and `person-warnings-tool-spec.md`'s future-candidates table, and includes a fresh `check-warnings` eval run.

Closes #1191.

## Review guidance

**Start here:** `person-warnings.ts`'s two renamed check functions (`checkLatestChildBirthToBirthFemale45` and its relative-mob twin `checkFemaleRelativesLatestChildBirthToBirth45`). The only behavior change in this PR is the `45` cutoff argument passed to the shared `latestChildBirthToBirth` predicate — everything else (constants, call sites, local var names) is a rename that follows from it. Also worth a look: the decision note added to `person-warnings-tool-spec.md`'s "Future warning candidates" section — it records why a gender-neutral cutoff and a second banded check were both measured and rejected, reasoning that isn't visible from the diff alone.

**Unsure about:** Nothing outstanding now — this section originally flagged PR #1545 (issue #1194, the severity `error`/`warning` → `contradiction`/`implausible` rename) as still open, with "exactly one real same-line collision." **Both corrected, per @T-FEH's 2026-08-19 review:** #1545 merged 2026-08-18 and the actual collision count was **three**, not one — two hunks in `person-warnings.ts` (orthogonal, adjacent-line edits: this PR's `_45` issueType rename vs. main's `severity: "implausible"`) plus `assumption-categories.md` (main's `implausible`/`corroboration` wording landing next to this PR's 45-year threshold). All three are resolved — @T-FEH did the merge (`4d8ef164`, `df60fd77`) rather than wait for a re-review cycle, since the re-run below couldn't happen without it. One more shape worth recording: the merge also showed a spurious `DU` (delete/unmerge) on `v1_2026-08-13_18-30-19.ann.json` — not a real conflict, just git pairing main's new run log with this branch's already-pruned `v1_2026-06-24` one by content similarity (run logs are near-identical JSON). Resolved by taking main's blob whole, verified byte-identical.

## Plan

**Didn't change:** No wiring of this (or any) check into the write path (`tree_edit`/`research_append`) — that's issue #1193's shadow-detector half (already shipped in #1541) and a lead-owned behavioral decision, not this issue. No gender-neutral variant, no second banded check, no severity promotion — all measured and rejected in the issue's ruling (numbers in the spec note). Didn't touch the escalation table in `warnings-as-identity-signals.md` (genealogist-lane doctrine, a different issue) or `match-merge-workflow-spec.md` (doesn't name the tag literally — checked).

**Acceptance check:** the new `latestChildBirthToBirthFemale45` / `femaleRelativesLatestChildBirthToBirth45` emitter tests in `person-warnings.test.ts` — both fail on `main` (the tag doesn't exist there) and pass on this branch. Verified the boundary is real, not tautological: reverted the cutoff to 55 locally, watched the gap-45 cases go red, then reverted to the real fix.

**Deviated from the plan:** Two, both recorded in `PLAN.md` §7 (gitignored, noting here too): (1) dropped the "fires at a 60-year gap" test case from each check (3 cases per check shipped instead of the planned 4) — `latestChildBirthToBirth` is one `>=`-based predicate shared across every cutoff in the file, so gap-60 exercises the identical branch as gap-45 and added no coverage; caught by `/check-drift`, judged safe to leave dropped. Also merged two near-duplicate test-tree-building helpers into one shared `buildParentChildTree`, per a `/code-review high` reuse finding. (2) The spec decision note's wording differs from my first draft — that draft repeated a factual error from the issue's own ruling text (claimed the male 45-74 range is "already covered by `latestChildBirthToBirth80`," which only fires at gap >=80); caught by `/code-review high`, verified against the code, and rewritten to state the real reason instead.

**Two more fixes landed in @T-FEH's merge-resolution commit (`df60fd77`), outside the run-log snapshot so neither adds re-run cost:** (3) the ruling's central constraint — lower the cutoff, do **not** promote severity — was unpinned by any test: temporarily promoting both `_45` emitters to `contradiction` left all 119 `person-warnings.test.ts` tests green. Added an explicit severity assertion to each emitter's test, so a future accidental promotion now fails loudly instead of shipping silently. (4) `person-warnings-tool-spec.md:568` still read "severity unchanged (`warning`, not promoted…)" after the #1545 merge — stale vocabulary the merge's automatic resolution had no reason to touch, since that line sat outside any conflict hunk. Corrected to `implausible`.

## Test plan

- [x] I ran `make test-all` (or `scripts/test.sh` — the same command) and it passed — **with two pre-existing, unrelated failures noted below**, both filed as separate issues rather than fixed here (out of scope for a person-warnings change, and both are genuine Windows-portability bugs, not flakiness):
  - `apps/electron/src/main/__tests__/feedback.test.ts` — one test assumes POSIX paths, fails on native Windows. Reproduces identically on `main`. Filed as #1707.
  - `apps/server` pytest suite — 8 failures + 2 indefinite hangs, all traced to a single root cause: `LocalProvider._kill_server` calls the POSIX-only `os.killpg`/`os.getpgid`, which don't exist on Windows, so the fallback `proc.terminate()` is never reached and the orphaned server process holds a port that later tests hang waiting on. Isolating those 10 tests, the rest of the suite is clean (152 passed, 1 skipped). Reproduces identically on `main`. Filed as #1713.
  - Every suite this PR's diff actually touches — typecheck, engine tests (vitest), eval harness tests (pytest), eval CRUD UI tests (vitest) — is fully green.
- [x] For `check-warnings`: the original run (`v1_2026-08-17_16-43-11.json`, 17 pass/1 partial/$1.77) **was annotated** (`f9b10253`) — but PR #1545's merge (below) swept 21 files in this skill's snapshot (SKILL.md, all three references, 9 mcp fixtures, 6 test JSONs, the rubric), so that run log is no longer active per rule 2 (confirmed live via `check_runlogs.py`: 21 files differ). **Fresh run committed** on the merged tree: `v1_2026-08-19_21-42-07.json` — 16 pass, 2 partial (`ut_check_warnings_006`, `ut_check_warnings_007`), $1.76. `check_runlogs.py` now confirms rule 2 clean; **only the `.ann.json` genealogist annotation pass via `make eval-ui` remains** before rule 3 is satisfied. The two prior candidates over `DEFAULT_KEEP_CANDIDATES=5` were auto-pruned with this write (`v1_2026-07-22_10-11-56` + sibling), same trap @T-FEH flagged from the merge.
- [x] **Merged `main` in** (`4d8ef164`, `df60fd77`, done by @T-FEH since the re-run above couldn't happen without it first) — 3 same-line conflicts from #1545 landing on this same skill, all resolved; see "Unsure about" above for the full breakdown. Also pruned one candidate run log over `DEFAULT_KEEP_CANDIDATES = 5` that the merge left behind (`v1_2026-07-19_03-22-09` + sibling).
- [ ] N/A — no `SKILL.md` frontmatter or `description`/DO NOT clauses changed, only `references/*.md` prose.

## Follow-on issues

**Folded in / filed instead:** Nothing new from #1191's own scope — everything it named as out-of-scope already has an owner (issue #1193 for write-path wiring). Two unrelated pre-existing bugs surfaced by running the full `make test-all` gate, filed per issue #1420's explicit instruction to file Windows-portability bugs separately rather than fold them in: #1707 (electron feedback test, POSIX path assumption) and #1713 (apps/server sandbox cleanup, POSIX-only `os.killpg`). Both **Touches:** lines name the exact files; neither touches this PR's own diff.


taiwo-stack (commented):
## /review pass — full, 15 commits behind main

This is an automated `/review` pass posted for whoever picks up senior review — it is not a substitute for the required `senior-developers` and `senior-genealogists` sign-off (CODEOWNERS), and I can't self-approve this PR as its author anyway.

**Does:** Renames `checkLatestChildBirthToBirthFemale55`/its relative-mob twin to `...Female45` and lowers the cutoff passed to the shared `latestChildBirthToBirth` predicate from 55 to 45. Adds the emitter tests this check never had. Updates three `check-warnings` reference docs and `person-warnings-tool-spec.md`'s future-candidates table. Commits a fresh `check-warnings` eval run.

**Intent match:** confirmed — independently drift-checked against `PLAN.md` (including its two disclosed deviations) with a fresh `drift-critic` pass: **no drift**.

**Checks:** `pytest`, `vitest`, `e2e-fixtures`, `lockfile-drift`, `scan` (×2), `server-pytest` all pass. `runlogs` fails for the disclosed, expected reason: `v1_2026-08-17_16-43-11.json` has no `.ann.json` yet — pending the genealogist annotation pass this PR's body already calls out.

**Merge-base check:** branch is `BEHIND` main by 15 commits. Confirmed zero file-level overlap between this PR's 9 changed files and the 55 files main gained since the branch point. Also confirmed from workflow source that this repo's `pull_request`-triggered CI checks out `refs/pull/1714/merge` (not the bare head), so every passing check above already validated this exact merge against current `main`.

**Verified directly, not just read:**
- Reverted the cutoff to 55 locally and watched the two new gap-45 emitter tests go red, then reverted — the acceptance check discriminates.
- Ran the fix against the actual `elisabetha-sugecz-parents` tree from issue #1191 (the gap-51 case): `latestChildBirthToBirthFemale45` now fires where nothing fired before.
- `tsc --noEmit` clean; engine vitest 2341/2341; eval-harness pytest 2221 passed/3 skipped; eval CRUD UI vitest 168/168.

**No `[LEAD]` findings** — doesn't touch schema, auth, an ADR, or anything hard to undo.

**No `[CRITICAL]` findings.** Specifically checked and cleared: `issueType` is not a closed enum anywhere (`CLOSED_ENUMS` in `validator.ts` doesn't reference it, no schema lists it), so the rename needed no schema-side updates and has none. No LLM-output trust-boundary concern. No check-that-cannot-fail — proved by the revert-and-watch-red test above.

Nothing informational rises to worth flagging either — no missing `encoding="utf-8"` (no Python touched), no stale docs, no duplicated logic left behind, no unbatched tool-call cost, no prose added to a `SKILL.md`/agent body.

**Not done:** did not spin up a local test-merge worktree — justified above via the CI merge-ref behavior rather than skipped silently.

Asimi1234:
Reviewed on the tree merged with `main` (clean merge, 16 commits behind): looks solid.

**Suites (merged tree):** `tsc` clean; full engine vitest **2373/2373**. Touched suite on branch head: `person-warnings.test.ts` **119/119**.

**Acceptance check is real, not tautological:** swapping in `main`'s source (cutoff back at 55) fails **exactly 2 tests** ("fires at exactly a 45-year gap", direct + relative-mob); restore → 119/119. Predicate confirmed `>= cutoff` (`person-warnings.ts:412`), so the 44-silent / 45-fires boundary holds.

**Straggler grep:** zero old `…Female55` / `…_TO_BIRTH_55` refs repo-wide; the only surviving `MOTHER_TOO_OLD` is the intentional decision note (`person-warnings-tool-spec.md:565`).

**Eval run:** 17 pass / 1 partial / $1.77 — matches the body. The one partial (`ut_check_warnings_006`) is on `hasAgeRangeGreaterThan120`, an unrelated warning, not a regression from this cutoff.

**Spot-checked the body's factual claims and they hold:**
- #1707 / #1713 are both open and this diff touches neither `apps/electron` nor `apps/server`; #1713's mechanism reproduces on read (`os.killpg` raises `AttributeError` on Windows, which `except (ProcessLookupError, PermissionError)` at `local.py:230-231` doesn't catch → `proc.terminate()` never reached).
- deleted old run-log `v1_2026-06-24` was `released: false` (prunable per runlog discipline).

**Verified the ruling's corpus numbers independently** (replicated the shipped `latestChildBirthToBirth` predicate over every committed e2e final tree): female-gated ≥45 flags **6 distinct people / 9 instances**, and **0 at cutoff 55** — exact match to the ruling, down to the three named PIDs (`G4C9-Y6C`, `KZHH-VTX` appearing in two runs, `L7B2-HQ8`). Also confirmed the spec note's corrected reasoning: at ≥80 **zero** people flag corpus-wide, so the issue's original "male 45-74 is already covered by `latestChildBirthToBirth80`" line was wrong — good catch dropping it.

**Only blocker:** the red `runlogs` check — "`v1_2026-08-17_16-43-11.json` has no `.ann.json`" — which clears once a genealogist runs the annotation pass. No code changes needed. Sign-off still routes to @PioneerAIAcademy/senior-developers (`*.ts`) and @PioneerAIAcademy/senior-genealogists (`plugin/skills/**`, `eval/runlogs/**`).

T-FEH (changes requested):
Requesting changes. The code is right — the blocker is a paid eval re-run that main's landing made unavoidable, and it cannot be waved through.

**I resolved the merge and pushed it (`4d8ef164`, `df60fd77`)**, because you cannot do the re-run until the merge exists. Three conflicts, all from #1545's severity rename landing on this same skill:

- **`person-warnings.ts`** (2 hunks) — orthogonal edits on adjacent lines: your `_45` issueType, main's `severity: "implausible"`. Kept both.
- **`assumption-categories.md`** — same shape; main's `implausible`/`corroboration` wording with your 45-year threshold.
- **`v1_2026-08-13_18-30-19.ann.json` reported `DU`, which looks like this branch dropping an annotation. It isn't** — the file is absent at the merge-base *and* on this branch, and present only on main, which added it. Git paired main's new run log with your pruned `v1_2026-06-24` one by content similarity, because run logs are near-identical JSON. Took main's blob whole; verified byte-identical, nothing lost from either side. Worth flagging as a shape: `eval/CLAUDE.md` forbids hand-editing annotations, so "resolving" that hunk by merging text would have corrupted a calibration file.

The merge also left **6 candidates against `DEFAULT_KEEP_CANDIDATES = 5`**, and `check_runlogs.py` has no retention rule to catch it, so I pruned the oldest (`v1_2026-07-19_03-22-09`) with its sibling. Same trap Gennecis caught on #1225.

**The blocker: rule 2 fails, and only a re-run clears it.** Main's rename swept **21** files in this skill's snapshot — the SKILL.md, all three references, nine mcp fixtures, six test JSONs and the rubric — so `v1_2026-08-17_16-43-11` no longer matches the tree. I checked this is main moving and not my resolution: hashed against your pre-merge head `f9b10253`, every snapshot file **matched**; against current main, every one **differs**. Your run was fully active when you took it.

`eval-cosmetic-skip` is not available here, and I want to be explicit rather than leave it as an option: the fixtures now return `implausible`/`contradiction` instead of `warning`/`error`, and the references teach 45 rather than 55. Both are precisely what the skill's severity triage reads, so this is a behaviour change by definition.

Worth naming the cause, because it is not anyone's mistake in this PR: **both PRs re-ran `check-warnings` on divergent snapshots.** #1545 shipped its own re-run of this skill; you shipped yours. Neither could have merged the other's. This is the collision `fill-ready`'s one-eval-slot-per-skill rule exists to prevent.

**Two things I fixed while I was in there** (both outside the snapshot, so neither adds cost):

1. **The ruling's central constraint was unpinned.** #1191 says lower the cutoff, do **not** promote the severity — it fired 0 times at 55 corpus-wide, so promotion would be an unmeasured doctrine commitment. I promoted both 45 emitters to `contradiction` and **all 119 tests stayed green**. Severity is the exact axis this merge turned on, which makes it the worst thing to leave unasserted. Added an assertion to each; both now fail on that promotion.
2. `person-warnings-tool-spec.md:568` still read "severity unchanged (`warning`, not promoted…)" while the code emits `implausible`. That line was not in a conflict hunk, so the merge kept it silently — the prose ended up describing the pre-#1545 vocabulary for a decision that is still correct.

Verified: engine vitest **2400 passed / 102 files**, typecheck clean, harness **2316 passed, 2 skipped**. Mutation-checked the rest of your new tests too — 45→55 fails 2, 45→44 fails 2, dropping the gender gate fails 1. Nothing vacuous.

Two notes for the body: the "exactly one real same-line collision" line understated it (three, now resolved), and both existing reviews read `b9f93215` and say the only blocker is a missing `.ann.json` — that landed in `f9b10253` and rule 3 passes now. @taiwo-stack's "zero file-level overlap with the 55 files main gained" is no longer true; #1545 hit `person-warnings.ts`, the test file and all three references.

Left alone deliberately: the "Future warning candidates" table at `:549-556` and `match-merge-workflow-spec.md:499` still use `warning`/`error`. Those describe warnings the code does not emit — main's own un-swept #1545 debt, which #1191 scopes out.

Re-run `make eval-skill SKILL=check-warnings` on this merged state, annotate the review sample, and I will re-review promptly.

taiwo-stack:
Done, per your review:

- Fresh `make eval-skill SKILL=check-warnings` run on the merged tree: `v1_2026-08-19_21-42-07.json` — 16 pass, 2 partial (`ut_check_warnings_006`, `ut_check_warnings_007`), $1.76. `check_runlogs.py` confirms rule 2 clean now; only the `.ann.json` annotation remains.
- PR body updated: the collision count corrected to three (not one), your two merge-resolution fixes (severity-regression assertions, the stale spec line) credited explicitly, and the test-plan section rewritten to reflect the merge + re-run rather than the pre-merge state.

Annotation via `make eval-ui` is the one thing left, and it's not mine to do — flagging for whoever picks that up.

promise-emmanuel (approved):
Reviewed on the tree merged with main — 131 commits behind, merge is clean, zero conflicts.

Rule 2 is genuinely clear, which is the thing that mattered. Main moved 131 commits since the branch point and that is exactly what killed the previous re-run. Hashed all 55 snapshot files of v1_2026-08-19_21-42-07 against the merged working tree: 0 differ. check_runlogs.py (BASE=main, HEAD=bc29f2e7) → "All runlog rules satisfied". Retention at the cap: 5 candidates, none released.

Suites (merged tree): engine vitest 2456/2456; tsc --noEmit clean; eval-harness pytest 2418 passed / 2 skipped; apps/server pytest 209 passed / 1 skipped — which also corroborates #1713 as genuinely Windows-only.

Mutation-tested every new guard — none vacuous: cutoff 45→55 fails 3; severity → contradiction fails 2; gender gate removed fails 2. @T-FEH's severity assertion is the valuable one — the ruling's central constraint ("lower the cutoff, do not promote") is now held by an assertion instead of prose.

Stop-rules checked, not assumed: issueType appears in no schema, no CLOSED_ENUMS entry and no packages/schema union, so the tag rename correctly needed zero schema edits. No auth paths, no ADR reversal. Both deleted run logs were released: false.

One correction to the body, non-blocking. ut_check_warnings_007 flipped pass → partial and has no annotation entry at all. It sits outside the review_sample so the gate correctly passes, but the body claims otherwise:

You wrote: (43 corrections, all five review_sample tests covered, written comments on both partials)

Change to: (49 correction entries — no score changes, all agreeing with the judge — all five review_sample tests covered, with a written comment on ut_check_warnings_006. ut_check_warnings_007 also came back partial and is outside the review sample, so it is unannotated: it flipped from pass in the pre-merge run, and the judge marked it down for rendering [!] Implausible, which SKILL.md:77 mandates — a judge/rubric issue, not a skill regression.)

Neither partial is a regression from this change. 007 is hasEarlyMarriage14, 006 is hasAgeRangeGreaterThan120 — no Female45 involvement in either. The judge's rationale on 007 reads "the skill labeled it '[!] Implausible' with a tone suggesting higher concern", but SKILL.md:77 says verbatim severity: "implausible" → [!] Implausible. The skill was penalised for obeying its own instruction, and [!] prefixes both tiers so it cannot carry the signal the judge is reading into it. Lane 2 (eval defect), not a prose fix.

Pre-existing bug, out of scope here — needs a genealogist owner. SKILL.md:120 renders hasAgeRangeGreaterThan120 as [!] Implausible with [Valid: …], but person-warnings.ts:1078 emits severity: "contradiction" — so the skill's own worked example contradicts both its mapping rule at line 77 and the tool's actual output. That example also omits the merged-records discussion its hasEventAfterDeath1 sibling includes, which is exactly what test 006 was dinged for missing in both runs.

Prior reviews reconciled: @taiwo-stack's "zero file-level overlap with main's 55 files" no longer applies (#1545 hit person-warnings.ts, the test file and all three references), and its runlogs blocker is resolved; its skipped test-merge is now covered above. @T-FEH's two fixes both confirmed — severity assertions fail on promotion, spec reads implausible. The item @T-FEH left open still stands and is still correctly out of scope (candidates table :549-556 and match-merge-workflow-spec.md:499 still use warning/error — main's un-swept #1545 debt).

Not checked: turbo test typecheck — no root node_modules in the worktree; the diff touches zero paths under apps/, packages/schema or packages/viewer-ui, and lockfile-drift / pnpm workspace / eval CRUD UI are green on this head. Did not re-run the paid eval (one-run rule; the committed run is provably active). The annotation labels' genealogical correctness is the senior-genealogists half — I verified coverage, not verdicts.

T-FEH (approved):
Clearing my changes-request. The blocker is properly resolved, and you paid the run rather than reaching for the label I told you didn't apply — which was the right call and the expensive one.

**Rule 2 is genuinely clean, not bypassed.** `check_runlogs.py` exits 0 on all four rules against the merged tree, with **55/55 snapshot files hashing identical**, and the only label on this PR is `developer` — no `eval-cosmetic-skip`. Rule 3 covers all five `review_sample` tests.

**I audited the annotation's content, not just its completeness**, because a sibling PR (#1764) has just shipped a hand-written `.ann.json` whose comments cited evidence the run log contradicted. Yours is the opposite on every check: all 49 `llm_score`s byte-match the run log's `aggregated_dimensions`, it validates against `ann.schema.json`, and its byte-shape — key order, 2-space indent, `"comment": null`, trailing newline, and the delete-and-reappend on re-review — is identical to a main-authored CRUD-UI annotation. The comments hold against the actual transcript too: `006`'s response really does cite `hasAgeRangeGreaterThan120` with F1 ~1845 / F2 1908-03-12, really does stop at "Verify the recorded birth date and death date", and really contains no merged-identity discussion, which that test's `judge_context` bullet 4 required. That is a real read, not a gate-satisfier.

**My two merge fixes survived and still bite.** Mutating severity to `contradiction` fails 2 tests, and the cutoff back to 55 fails 3; before my `df60fd77` the severity promotion failed none. Spec `:568` reads `implausible`. No `_55` anywhere.

**On the two partials — I raised `ut_check_warnings_007` because it scored `Severity classification` = 2 and is not in the `review_sample`, and severity is exactly what this PR moves. That concern is answered: neither partial is yours.** `006` was already partial in the pre-merge 08-17 run *and* in main's 08-13 run. `007` is new, but traces to `a4fb34e0` (#1545), which rewrote both the fixture payload to `implausible` and the `judge_context` demanding "Note severity" — the skill then rendered `[!] Implausible` exactly as `SKILL.md:77` mandates. Lane 2, inherited. And no `Female45` fires in either, so the lowered threshold is not over-firing.

Suites on the merged tree: engine vitest **2462 / 104 files**, `tsc` clean, harness **2460 passed, 2 skipped**, `apps/server` **209 passed, 1 skipped**, eval CRUD UI **168**. All 12 checks green.

Two non-blocking notes:

- The body says "43 corrections … written comments on both partials". It is 49, and `007` has no entries at all — it is outside the sample, so nothing is owed, but the sentence overstates it. @promise-emmanuel has already written the replacement text.
- **Worth filing rather than fixing here.** `SKILL.md:120`'s worked example prints `[!] Implausible -- Long lifespan [hasAgeRangeGreaterThan120]`, but `person-warnings.ts:1078` emits `contradiction` for that emitter — it is the *relatives* variant at `:2634` that is `implausible`, which is probably where the confusion came from. So the example teaches the wrong label for the direct case. It is pre-existing #1545 debt, and correcting a SKILL.md line invalidates the check-warnings snapshot and buys a **third** paid run on this branch. That cost is the reason it should be an issue rather than a fix here.

PR #1717, opened 2026-08-18
skill_invocation.py: two guardrail-detector corrections (#1569)

### LINKED ISSUE #1569: skill_invocation.py: two guardrail-detector corrections sharing one corpus replay — lane check should count attempts, proof-conclusion arm misses writes into existing relationships

chrisedeson:
## Summary

Normal tier. Closes #1569. Two corrections in `skill_invocation.py`, sharing one new corpus-replay tool.

- `find_protected_writes_by_unnamed_delegate` skipped any errored tool call. The rule is about who called a protected write, not whether the call succeeded, so an unnamed delegate's errored write now still counts.
- `find_effects_without_invocation`'s proof-conclusion arm missed a conclusion written into an already-seeded relationship or a primary fact replaced in place. Both now use a normalized content signature instead of an endpoint-only key or a raw count.

## Review guidance

**Start here:** `_relationship_conclusion_signature` and `_primary_fact_signatures` in `skill_invocation.py`. Both route fact content through `Counter` before hashing. A bare set collapses two facts with identical content into one, which hides a genuinely-added duplicate. A sorted tuple crashes on a real corpus shape (`chresten-nielsen-daughter`'s R1 has two Marriage facts, one with a place and one without, and sorting mixes `None` and `str`). `Counter` avoids both.

## Plan

**Didn't change:** `_relationship_key` itself. It stays endpoint-only and is still shared with `find_relationship_writes_without_warnings_check`, which asks a different question (was a new edge written at all) and is unaffected by this fix.

**Acceptance check:** `test_is_error_entries_still_counted`, `test_flags_a_marriage_fact_dated_onto_a_seeded_couple`, `test_flags_a_parentchild_subtype_reclassified`, and `test_flags_a_second_primary_fact_with_content_identical_to_an_existing_one` all fail on `main` and pass on this branch. Each mutation-verified: reverted the specific fix, confirmed the exact predicted failure, restored.

**Deviated from the plan:** the fix originally used a `frozenset` of raw fact tuples. A fresh code review found that a bare set collapses two identical-content facts into one, silently hiding a genuinely-added duplicate. Switched to `Counter` in the same PR, which keeps the same crash-free construction and adds back multiplicity.

## Test plan

- [x] `make test-all` passed. One pre-existing, already-tracked flake (`apps/electron`'s feedback-zip test times out at 5000ms under parallel load, #1366), unrelated to this change and reproduces on `main` too.
- [x] N/A. Nothing under a skill dir changed. `skill_invocation.py` is harness code, not a skill body, so no run-log snapshot moved.
- [x] N/A. No skill description or DO NOT clause changed.

Also built `eval/harness/e2e/detector_before_after_report.py` (`make e2e-detector-diff DETECTOR=lane-check|proof-conclusion-arm`), since neither fix had a corpus instrument and the issue asked for one reusable replay rather than two throwaway scripts. Measured over the full corpus:

- lane-check: no divergence. No committed run yet has an unnamed-delegate write that also carries `is_error`.
- proof-conclusion-arm: one real divergence, `rossi-marriage`'s 2026-07-16 run. A dated, sourced, primary Marriage fact was written onto a previously bare seeded Couple relationship. `proof-conclusion` was never invoked anywhere in that run. Old detector missed it. New one catches it.

## Follow-on issues

**Folded in / filed instead:** the `Counter` fix above was folded in, found by the same PR's own review pass. Nothing filed.


chesworthrm (commented):
Reviewed the full diff against the merge with `main` (branch is level, so no test-merge needed). Ran the harness, engine, web and control-plane suites, and mutation-tested every fix in the diff. All four tests in your **Acceptance check** are genuinely load-bearing, and I reproduced both corpus claims exactly — `lane-check` no divergence, `proof-conclusion-arm` the one `rossi-marriage` divergence.

Nice work on the replay tool. Building one reusable instrument instead of two throwaways was the right call, and the `Counter`-over-`frozenset` reasoning in `_relationship_conclusion_signature` is correct.

Two things to change before this merges.

## 1. Half of the `Counter` fix has no test

`_relationship_conclusion_signature` routes facts through `Counter` for a stated reason: a bare `frozenset` would collapse two facts with identical content into one and hide a genuinely-added duplicate.

The person-fact half of that fix is covered — I reverted `_primary_fact_signatures` to set semantics and `test_flags_a_second_primary_fact_with_content_identical_to_an_existing_one` failed, as designed.

The relationship half is not. I replaced `Counter` with a bare `frozenset` of the raw fact tuples in `_relationship_conclusion_signature` and **all 2241 harness tests passed.** So the fix your own review pass found can be undone on the relationship side with a green suite.

Add this next to the other relationship tests in `eval/harness/tests/unit/test_skill_invocation.py`. I confirmed it fires on your branch as written, and does not fire under the bare-`frozenset` mutation:

```python
def test_flags_a_second_relationship_fact_with_content_identical_to_an_existing_one():
    """The relationship-side twin of
    `test_flags_a_second_primary_fact_with_content_identical_to_an_existing_one`: a bare
    frozenset of normalized fact signatures collapses two facts carrying the identical
    (type, date, place) into one, hiding a genuinely-added duplicate. Counter-based
    comparison preserves multiplicity. Without this test, dropping the Counter here
    passes the whole suite."""
    dup = {"type": "Marriage", "standard_date": "21 Oct 1860", "standard_place": "Lezayre"}
    starting_rel = {"id": "R1", "type": "Couple", "person1": "I1", "person2": "I2", "facts": [dict(dup)]}
    final_rel = {
        "id": "R1",
        "type": "Couple",
        "person1": "I1",
        "person2": "I2",
        "facts": [dict(dup), {**dup, "sources": [{"ref": "src-2"}]}],
    }
    starting = {"persons": [], "relationships": [starting_rel]}
    tree = {"persons": [], "relationships": [final_rel]}
    violations = find_effects_without_invocation([], {}, tree, starting_tree=starting)
    assert any("proof-conclusion" in v for v in violations)
```

## 2. The issue-ref ratchet was raised instead of respected

The PR adds one bare `#1569` to each of two specs, then raises `ISSUE_REF_BASELINE` by one each to admit them. That test's own contract says not to do that:

> A RATCHET, not a freeze: the count must match exactly, so removing a reference fails until the number here comes down with it, and **adding one fails outright.**
>
> The durable form is the **fact, not the ticket** … If the conclusion is worth keeping, write it down — then the number adds nothing.

I checked every prior change to that baseline: `be0ab3b8` (7→4), `83dd0e85` (12→10), `dc4146b4` (17→12 and 17→16). All of them went down. This would be the first raise. A ratchet an author can raise in the same PR stops being a ratchet.

Neither exemption applies here. These are not `owner/repo#N` refs and not in `docs/adrs/`. Neither is evidence-shaped either — both sentences state the fact completely without the number.

I dropped both refs, reverted both baseline entries, and `doc-links.test.ts` passes 130/130 at 10 and 16. The whole `doc-links.test.ts` hunk reverts.

In `docs/specs/e2e-test-spec.md`:

```
You wrote:      succeeded, was rejected by validation, or errored for any other reason (issue
                #1569); it does not key off `is_error` the way the main-thread half's denial

Change it to:   succeeded, was rejected by validation, or errored for any other reason; it
                does not key off `is_error` the way the main-thread half's denial
```

In `docs/specs/guardrail-enforcement-spec.md`:

```
You wrote:        read `is_error` (issue #1569): its gate is caller identity, not skill

Change it to:     read `is_error`: its gate is caller identity, not skill
```

In `packages/engine/mcp-server/tests/packaging/doc-links.test.ts`:

```
You wrote:      "docs/specs/e2e-test-spec.md": 11,
Change it to:   "docs/specs/e2e-test-spec.md": 10,

You wrote:      "docs/specs/guardrail-enforcement-spec.md": 17,
Change it to:   "docs/specs/guardrail-enforcement-spec.md": 16,
```

The `#1569` refs in the `skill_invocation.py` docstrings are fine — that file is not scanned.

---

Two notes that do not block merge.

**`_replay_lane_check` prints "no divergence" over runs that cannot produce one.** The line reads `no divergence — old and new agree on every run scanned` across all 155 runs. But a run can only express this divergence if some entry carries both `is_error: true` and an `agent_id`, and only **7 of 155** do. Schema-version spread across the corpus is `{1: 7, 2: 1, 3: 8, 4: 1, None: 138}`.

Your fix is not a no-op — the corpus just cannot show it yet. I checked the run the issue measured, `antonio-lucas-spouse/run-2026-08-04_22-45-15.json`: it is `harness_schema_version: 2` with zero `is_error: true` entries, so old and new both give 15. Force `is_error` where `response_summary` carries `"ok": false` and it becomes old=12, new=15. Real narrowing, invisible to the replay.

Your PR body says this. The tool does not, and the tool is what gets reused next time. The proof-conclusion arm already reports its dropped runs (`Skipped 1 run(s): …`) — do the same here. Have `_replay_lane_check` return an eligible count alongside the divergences and print something like:

```
7 of 155 run(s) could express this divergence (an errored call carrying an agent_id).
```

Worth folding in while the file is new.

**The person-evidence arm still has the count blind spot you just fixed.** `_is_new_content_this_run` compares `len(p.get("facts") or [])` against `starting_fact_counts` — the same raw-count check you replaced for primary facts, in the same function. So a seeded person whose fact is replaced in place with no `person_evidence` entry is still invisible:

```
person-evidence arm fires on a replaced fact? False
person-evidence arm fires on an added fact?   True
```

Out of #1569's scope, and it needs its own corpus measurement, which you now have the instrument for. Fold it in or leave it — your call.

---

**What I could not check.** No CI job runs a live e2e, and both corrections change what a hard post-run detector fires on. The replay tool is the substitute and I ran it, but it reads committed runs — it cannot tell you whether the widened relationship signature over-flags on a future run. The specific shape: the signature now differs when a seeded relationship merely *loses* a fact, which would read as a new conclusion. Zero false positives across the 154 comparable runs, so I am recording it, not raising it. I also did not measure the blast radius of the person-evidence note above.

One environment note in case you hit it: on a fresh worktree the harness suite fails `test_mock_mcp.py::test_successful_call_carries_no_is_error` with `research_query: build not found`. That is the missing engine build, not your change. After `npm run build` in `packages/engine/mcp-server`, 2241 passed / 2 skipped.

chrisedeson:
Both fixes applied.

The relationship-side `Counter` fix had no test. Reverting it to a bare `frozenset` passed the whole suite. Added it.

Dropped both #1569 references from the specs and reverted the ratchet baseline. You're right that it should only ever go down.

Also folded in your two non-blocking notes since they're the same file, same pattern: the lane-check replay now reports how many runs could even show the divergence (7 of 155), and the person-evidence arm had the identical count-based blind spot the proof-conclusion arm just got fixed for. Same fix, new test, measured zero divergence on the corpus.

One more thing a follow-up review caught after that: the replay tool's dispatch would've silently mis-reported if a future detector got added without updating it. Made it fail loudly instead.

chesworthrm (commented):
Second pass. Both blocking findings from last round are properly closed. I re-ran the relationship-side `Counter` mutation and your new test fails exactly as designed, and the ratchet is back to 10 and 16 with both spec refs dropped. I also re-checked your acceptance claim against `main`: seven of the ten new tests fail there, including all four you named. Both corpus claims reproduce exactly. Harness suite in your worktree: 2259 passed, 2 skipped.

Folding in my two non-blocking notes was the right call. The person-evidence arm fix is clean and its new replay detector reports no divergence across the corpus.

One thing to change, and it is my fault.

## The eligibility line I asked you for is wrong

I gave you that predicate last round and it was too loose. `is_error: true` plus an `agent_id` is necessary for this divergence but not sufficient. The call also has to be one this detector would actually flag. A non-owned tool, or a call from a dedicated agent, exits on the classification instead and can never diverge.

I measured both over all 155 runs:

```
runs with an errored call carrying an agent_id:            7
runs where that call is also a protected write we flag:    0
```

So the report tells the reader the corpus had 7 chances to show this fix working and it changed nothing in all 7. The real number of chances is zero. That is backwards from what the line is for. It makes an untestable result look like a tested no-op, in the summary line of a tool built to be reused on the next detector correction.

Tightening the predicate alone fixes the number but makes the line circular. The fix only ever adds violations, so "could express the divergence" and "did express the divergence" collapse into the same test, and the line would just restate the result printed below it.

The number that actually says "the corpus can't show this yet" is how many runs carry agent-id stamping at all. That is **16 of 155**, and it is independent of the result. Print both.

All in `eval/harness/e2e/detector_before_after_report.py`. I ran these and the header becomes `16 of 155 run(s) carry agent-id stamping at all; 0 of those could express this divergence (an errored call this detector flags).`

```
You wrote:      """Whether this run COULD show the lane-check divergence at all: the fix only
                changes behavior for an entry that carries both `is_error: true` and a non-None
                `agent_id` (found by review — most of the corpus predates agent-id stamping, so
                "no divergence" without this context reads as "the fix is a no-op," when it's
                really "the corpus can't show it yet")."""
                return any(e.get("is_error") is True and e.get("agent_id") is not None for e in tool_calls)

Change it to:   """Whether this run COULD show the lane-check divergence at all: an entry that is
                BOTH `is_error: true` and a write this detector flags. Carrying an `agent_id` is
                necessary but not sufficient — a non-owned tool or a dedicated agent exits on the
                classification instead, so the looser test overstates what the corpus proved."""
                return any(e.get("is_error") is True and _lane_check_new([e]) for e in tool_calls)


                def _lane_check_stamped(tool_calls: list[dict[str, Any]]) -> bool:
                    """How much of the corpus can speak to this detector at all, independent of the
                    result: most committed runs predate agent-id stamping, so "no divergence" on its
                    own reads as "the fix is a no-op" when it is really "the corpus can't show it yet"."""
                    return any(e.get("agent_id") is not None for e in tool_calls)
```

```
You wrote:      def _replay_lane_check(paths: list[Path]) -> tuple[list[Divergence], int]:
                    out: list[Divergence] = []
                    eligible = 0
                    for path in paths:
                        data = json.loads(path.read_text(encoding="utf-8"))
                        tool_calls = data.get("tool_calls") or []
                        if _lane_check_eligible(tool_calls):
                            eligible += 1

Change it to:   def _replay_lane_check(paths: list[Path]) -> tuple[list[Divergence], int, int]:
                    out: list[Divergence] = []
                    eligible = 0
                    stamped = 0
                    for path in paths:
                        data = json.loads(path.read_text(encoding="utf-8"))
                        tool_calls = data.get("tool_calls") or []
                        if _lane_check_stamped(tool_calls):
                            stamped += 1
                        if _lane_check_eligible(tool_calls):
                            eligible += 1
```

```
You wrote:          return out, eligible
Change it to:       return out, eligible, stamped
```

```
You wrote:          divergences, eligible = _replay_lane_check(paths)
                    print(f"{eligible} of {len(paths)} run(s) could express this divergence (an errored call carrying an agent_id).")

Change it to:       divergences, eligible, stamped = _replay_lane_check(paths)
                    print(
                        f"{stamped} of {len(paths)} run(s) carry agent-id stamping at all; {eligible} of those "
                        "could express this divergence (an errored call this detector flags)."
                    )
```

## The new file has no tests

Every other report module under `eval/harness/e2e/` has at least one test file. This is the only one at zero. `nudge_report.py` is the closest in size at 333 lines and carries a 154-line test file.

That gap is what let the eligibility bug through. It also leaves the `raise AssertionError` at the end of `main` — the one your comment calls load-bearing — with nothing exercising it.

A small `eval/harness/tests/unit/test_detector_before_after_report.py` closes it. Three cases are enough:

- `_lane_check_eligible` is False for an errored call on a tool this detector does not own, and True for an errored owned write from a named non-dedicated agent.
- `_lane_check_old` and `_lane_check_new` disagree on a hand-built errored owned write, and agree on the same entry when it is not errored.
- `format_divergences` on an empty list.

## Approvals

Still needs two, one of them from a senior team. Every changed path is Python, a spec under `docs/specs/`, or the `Makefile`, so `senior-developers` or `senior-genealogists` both satisfy the CODEOWNERS rule. There are no approvals on this PR yet.

## What I could not check

No CI job runs a live e2e, and both corrections change what a hard post-run detector fires on. Your replay tool is the substitute and I ran all three arms, but it reads committed runs — it cannot tell you whether the widened relationship signature over-flags on a future one. The shape to watch: the signature now differs when a seeded relationship merely *loses* a fact, which would read as a new conclusion. Zero false positives across the corpus, so I am recording it, not raising it.

I did not run the TypeScript or control-plane suites locally. The diff touches neither, and both are green in CI on this head.

chrisedeson:
Fixed. You were right, the eligibility predicate was wrong. The loose predicate said 7 of 155 runs were eligible, the tightened one says 0. Now prints both numbers separately, agent-id stamping (16 of 155) and true eligibility (0 of those).

Added the test file. Eight tests, mutation-verified against the exact bug this predicate had.

chesworthrm (approved):
Both findings closed. Approving.

**Eligibility predicate.** The two-number split is right, and the header now says the true thing:

```
16 of 155 run(s) carry agent-id stamping at all; 0 of those could express this
divergence (an errored call this detector flags).
```

That is the honest read: the corpus is mostly unstamped, and none of the stamped runs can exercise this fix yet. The old line claimed 7 chances where there were none.

**Tests.** I mutation-tested both new predicates rather than take the file on faith. Reverting `_lane_check_eligible` to the loose version fails `test_lane_check_eligible_false_for_an_errored_call_on_an_unowned_tool` — it catches the exact bug it was written for. Narrowing `_lane_check_stamped` to also require `is_error` fails its own test, so that one is not vacuous either. Your fixtures are self-verifying too: if `owning_skills` did not attribute that `materialize_facts` call to `person-evidence`, three of the eight would fail rather than quietly pass. That is the right way to build them.

Full harness suite: 2267 passed, 2 skipped — exactly the 8 new tests over the previous 2259, no collateral.

`skill_invocation.py`, both specs, the `Makefile` and `test_skill_invocation.py` are byte-identical to the head I mutation-tested last round, so the `Counter` verification and the seven-fail-on-`main` acceptance check carry over unchanged. I did not need to redo them.

Nice work on the replay tool. It started as a measurement for this PR and it is ending up as the instrument the next detector correction gets checked against, which is worth more than either fix in here.

One process note, not a change request: my approval satisfies the CODEOWNERS senior requirement, but branch protection wants two, so this still needs one more from anyone with write access before it can merge.

jud-sdev (approved):
I reviewed this independently by running and mutating, not just reading. Test-merged with main (it is 45 behind) and both harness suites pass, 138 tests. I mutation-checked both corrections: re-adding the is_error skip fails test_is_error_entries_still_counted, and changing the Counter in _primary_fact_signatures to a set fails the duplicate-detection test because the set collapses the second fact. So both fixes genuinely bind, and the Counter choice is proven, not decorative.

The design is careful — the signature ignores id and order so a harmless re-serialization does not register as a conclusion, _relationship_key is left as-is so the other consumer is unaffected, and building one reusable corpus-diff tool instead of two throwaway scripts is the right call.

One mechanical note: the branch is 45 commits behind main, so it needs a main update before merge — it test-merges clean, so that is a button click.

Good work.

Issue #1639, opened 2026-08-16
The Cowork MCP registrar changed under us — nothing detects it, and agents were silently toolless for as long as it lasted


DallanQ:
> **Reviewed 2026-08-21 before junior handoff.** No decision needed — but the ruling
> below was rewritten: an earlier Cowork observation recorded the **opposite** result, so
> "the bare spelling does not appear anywhere" is not what to write. Read the corrected
> ruling and the seven-file edit list before starting.

> **Unblocked 2026-08-21 — issue #1732 reported on 2026-08-19.** Scope was narrowed
> 2026-08-18 and the census moved out to that issue; this card owns only the doctrine
> edit, and the measurement it waited on is in. **The second branch below is the one that
> fires: rewrite ADR-0004's rationale.** Read the ruling block under "What moved out"
> before starting. Original body follows under "Original issue".

**Touches:** CLAUDE.md, docs/adrs/ADR-0004-dual-spell-mcp-tool-names-in-agent-frontmatter.md,
docs/architecture.md, docs/specs/gps-mentor-agent-spec.md, docs/specs/image-reader-agent-spec.md,
docs/specs/research-append-tool-spec.md,
packages/engine/mcp-server/tests/packaging/agent-tool-names.test.ts,
docs/skill-lifecycle.md

## What moved out, and why

The census prompt — run a Cowork session on Windows and record which prefix the
genealogy tools carry — is now **issue #1732**, labelled `genealogist`. It is thirty
minutes on a machine only that team has, and it is the input this doctrine edit
depends on. Combined, this card was assignable to nobody: one half needs a Windows
Cowork session, the other edits stewarded doctrine.

**Issue #1732 reported on 2026-08-19. This is startable.** From a Windows Cowork session
with a connected genealogy project folder, prompted *"List every tool available to you, with
names spelled exactly as you see them."*:

- the genealogy tools carried `mcp__remote-devices__Genealogy_Research__…`, described as
  "via your device";
- the bare `mcp__Genealogy_Research__…` spelling was **not present**;
- `mcp__genealogy__…` was **not present**.

That is the second of the two branches this issue named, so the edit is settled:

> **The rule does not move. Only the rationale does.** Two censuses — macOS 2026-08-15 and
> Windows 2026-08-19 (issue #1732) — found every genealogy tool under
> `mcp__remote-devices__Genealogy_Research__…`, described as "via your device", with the bare
[...]

Issue #1638, opened 2026-08-16
image_transcribe's 180s OCR budget is capped at 60s by the device bridge


DallanQ:
> **Reviewed 2026-08-17 before junior handoff — decision settled the same day.**
> The "What to work out" section is replaced: half of it is already answered, and
> its third item is the wrong fix. The original follows under "Original issue".

**Decision (lead, 2026-08-17): document the ceiling; do not change the tool.**
`OCR_TIMEOUT_MS` stays 180s — it is correct wherever the call is not bridged — and
the Cowork cap is written down as a property of the environment. A per-install
`ocrTimeoutMs` override and a one-retry carve-out in the `image-reader` agent were
both weighed and dropped: the first recovers nothing until someone edits a config
file by hand, the second costs two fresh eval suites.

Ship the "Write-downs" checklist below and nothing else. Put the one clause saying
which two alternatives this beat into `docs/specs/image-transcribe-tool-spec.md`,
"Timeout budget", beside the table — that is where the next person to re-open this
will be standing.

## What is already established (do not re-derive)

- **The 60s is not ours.** `grep -rn "MCP_TOOL_TIMEOUT\|MCP_TIMEOUT" .` returns nothing —
  no harness, control plane, manifest or config anywhere in this repo sets a per-tool MCP
  ceiling. It is imposed by the Cowork client/bridge and is not settable from the plugin
  or the `.mcpb`. Treat "is it configurable from our side?" as answered: no.
- **The harness and hosted paths are not capped.** Both register the server over stdio
  (`apps/server/app/agent/real_agent.py:241`), and two committed run logs carry the tool's
  *own* 180s timeout as a tool result —
  `eval/runlogs/e2e/elena-asmundsdotter-origin/run-2026-08-13_01-40-40.json` and
  `eval/runlogs/e2e/victoriano-macatangay-parents/run-2026-08-13_23-24-54.json` — so a call
  ran well past 60s there without the client aborting.
- **The desktop path may be bridged too.** Issue #1639's census (macOS, Cowork desktop app
  with a connected folder, 2026-08-15) found all 47 genealogy tools carrying
  `mcp__remote-devices__Genealogy_Research__`, and the bare on-computer spelling absent. So
  "the desktop `.mcpb` honours 180s" is unverified — the ceiling may apply to every Cowork
  session.
- **The ceiling is not specific to `image_transcribe`.** It caps every MCP call, and four
[...]

PR #1234, opened 2026-08-03
viewer: render the gps-mentor's verdicts

DallanQ:
Closes #1223.

> **Stacked on #1224** — it changes `NOT_RENDERED` exemptions that PR introduces. Review #1224 first.

## Why

The mentor gate is **mandatory to invoke and record** — `/research` won't let a question be considered done until every `ps_id` it references carries a `focus: "proof-critique"` entry. The system required the critique to exist and then showed the researcher none of it. An `address_first` verdict — the mentor saying a proof needs work before it stands — was recorded and invisible.

## Verdict, not critique text — and that's forced, not chosen

`evaluation_entry` has **no notes field**. The mentor's prose lives at `file_path`, a host-side artifact the viewer can't read. What's renderable is the verdict, what it judged, when, and whether a later pass superseded it — the decision itself, which is the part that changes what a researcher does next.

#1223 offered "verdict only" as one of three options. Reading the schema showed it's the only one available.

**A section rather than inline on the proof summary**, because `target_type` is `question | proof_summary | project` and a project-level critique has no inline home.

Wired at four points: the `sectionComponents` registry, the Sidebar nav, the cross-link index, and CrossLink's section map.

## Two defects found while building it

- **`StatusBadge` had no colour mapping for any verdict**, so `looks_solid` and `address_first` both rendered gray — a verdict badge that says nothing. Added green / amber / red / gray.
- **My own `file_path` test asserted against a collapsed card**, which renders no body at all, so it would have passed whatever the body contained. It now expands first and asserts the body is genuinely open *before* asserting the path is absent.

That second one is the third time in this sequence a check of mine passed for the wrong reason. Worth the habit of trying to break each one.

## Exemption changes

#1224's exemptions for `evaluation_entry.target_id` and `.target_type` are **removed** — both now render, so leaving them would make the exemption list lie. `file_path` stays exempt with a sharper reason.

## Not done, deliberately

The eval CRUD UI's parallel `sections/` tree gets **no** Evaluations section. It's already three behind viewer-ui (`KnownInformation`, `Localities`, and now this) — a developer tool that mirrors selectively, not a second product surface. Adding one there isn't implied by this change.

## Limitation worth naming

The field-render lint gates on whether the **container is referenced** in a component, so a section file that exists but is never registered in `App.tsx` still satisfies it. I verified this by deleting the registry entry — the lint stayed green. It checks fields within a displayed object, not that the object is reachable from the UI.

## Test plan

- [x] `make engine-test` — 1913 passed
[...]

PR #1718, opened 2026-08-18
docs: record 60s Cowork bridge cap over image_transcribe timeout budget (#1638)

Asimi1234:
## Summary
- **Tier: Trivial** — docs + one code comment; no logic change.
- Records the Cowork device bridge's 60s cap on every MCP call as an environment
  property, so `image_transcribe`'s measured 180s `OCR_TIMEOUT_MS` is no longer
  implied to hold everywhere. The four write-downs from #1638 (lead decision
  2026-08-17: document the ceiling, do not change the tool).

## Review guidance
**Start here:** `docs/architecture.md` → "Other environment differences that
bite" — the durable statement of the 60s cap. The spec §5.7 clause and the
`OCR_TIMEOUT_MS` comment are local copies that point back to it. What to check:
the claims are accurately hedged — the desktop `.mcpb` bridging is stated as
*unverified* (#1639), not asserted.

**Unsure about:** Nothing on the prose. The one thing no suite can settle is
whether the desktop `.mcpb` is also bridged (#1639) — recorded as unverified.

## Plan
**Didn't change:** `OCR_TIMEOUT_MS` stays 180s (correct on the un-bridged
paths); no per-install override and no `image-reader` retry carve-out — both
weighed and dropped by the lead.

**Acceptance check:** the four #1638 checklist items are present (spec §5.7
Cowork clause; architecture.md environment bullet; the `OCR_TIMEOUT_MS` comment;
the "Add an MCP tool" budget clause). `tests/packaging/doc-links.test.ts` passes
130/130 — it *fails* when a bare issue-ref is added to a contract doc, which
caught a `#1639` I first put in the spec (removed; fact stated instead).

## Test plan
- `make engine-test` → **2365/2365**, build + typecheck clean, on the committed
  tree (57fb4b72).
- `make test-all` → all offline suites green **except** a pre-existing
  `apps/electron` `feedback.test.ts` load-timeout flake (2.2s isolated → 7.2s
  under turbo's parallel workers); this diff touches no electron file.
- **No offline suite reaches the defect.** Per #1638, neither harness goes
  through the device bridge — the only instrument is a live Cowork session, which
  I did not run. **CI does not cover this change.**

## Follow-on issues
**Folded in / filed instead:** two doc defects found in self-review were folded
in, not filed — (1) a spec denominator ("132 healthy") that contradicted the
section's own n=101/n=89; (2) a self-falsifying `grep … returns nothing` claim
that matched its own line once written into the doc. Sequencing: #1594 edits the
same spec and `catch` block but has no branch yet, so batching wasn't possible;
this PR no longer touches the `catch` block and edits a different spec section.


chesworthrm:
Reviewed the full diff and test-merged `origin/main` first — it merges clean, and everything is green on the merge: engine vitest 2374/2374, `tsc --noEmit` clean, `doc-links.test.ts` 130/130, harness pytest 2250, server pytest 164. All required checks ran on the current head. The branch is behind main by one commit, so it needs an update before merge.

The four write-downs from #1638 are all there, and the hedging is right — the desktop `.mcpb` is stated as unverified rather than asserted, which is the thing I went looking for.

One thing to fix.

**The 9.8% figure is correct, but the script that produced it no longer reproduces it.**

Running #1638's published script on this branch gives 13 of 144 healthy calls (9.0%), not 13 of 132 (9.8%). Your number is the better one. What changed is the corpus, not the measurement: #1682 strips `response_summary` from run logs older than 14 days, and that filter is how the script excludes errored calls. 219 of the 336 committed `image_transcribe` calls no longer have one. Twelve sub-second "no OpenRouter API key" failures now count as healthy transcriptions, which is exactly the case the issue said to exclude.

So the next person who re-runs it gets a lower number and no way to tell why. Worth pinning the provenance in the spec:

```
You wrote:      corpus (2026-08-17, by then grown to 132 healthy calls from the n=89 that sized
                180s below), 13 (9.8%) ran past 60s — a floor, since the corpus carries no
                bridge hop.
Change it to:   corpus as of commit b065b687^ (2026-08-16), 13 of 132 healthy calls (9.8%) ran
                past 60s — a floor, since the corpus carries no bridge hop. Counted by issue
                #1638's script, which is not the method that produced the n=101/n=89 below.
                Re-running it today reports 13 of 144 (9.0%): PR #1682 strips
                `response_summary` past 14 days, so its errored-call filter no longer sees
                most of the corpus.
```

And the same figure in `docs/architecture.md`:

```
You wrote:      budget this bites (measured 2026-08-17: 9.8% of healthy calls exceed 60s), but
Change it to:   budget this bites (9.8% of healthy calls exceed 60s, measured over the corpus
                as of 2026-08-16 — see the spec's Timeout budget section), but
```

One optional edit while you are in there, which does not block merge. The `#1639` in the `OCR_TIMEOUT_MS` comment was correct when you wrote it and went stale about six hours later — #1639 was narrowed that afternoon and the census that would settle whether the desktop `.mcpb` is bridged moved to #1732. #1639 now says not to start it until #1732 reports. Pointing the comment at #1732 sends the next reader to the measurement instead of the blocked card.

Nothing else. I did not verify the 60s cap itself — no suite reaches it, as you said in the PR body.

Asimi1234:
Thanks Richard — all addressed in `6a32f70d`.

- **Spec provenance pinned** exactly as you wrote it: figure attributed to commit `b065b687^` (2026-08-16), counted by #1638's script, with the re-run caveat (13 of 144 / 9.0% today, because #1682 strips `response_summary` past 14 days so the errored-call filter goes blind).
- **architecture.md** figure repointed to the spec's Timeout-budget section with the 2026-08-16 corpus date.
- **`OCR_TIMEOUT_MS` comment** repointed from #1639 to #1732.

I reproduced both numbers from the script rather than take them on faith: 132 healthy / 13 over 60s (9.8%) at `b065b687^`, and 144 / 13 (9.0%) on this branch now — and `b065b687` is indeed the #1682 strip commit, so its parent is the right anchor. The branch is also rebased onto current main now, so the behind-by-one is resolved.

chesworthrm (approved):
Approving. I test-merged `origin/main` first — it merges clean, and everything is green on the merge: engine vitest 2440/2440, `tsc --noEmit` clean, `doc-links.test.ts` 130/130, harness pytest 2356, server pytest 209. All 13 required checks ran on the current head.

I re-checked the code claims rather than take them from the issue: `MCP_TOOL_TIMEOUT`/`MCP_TIMEOUT` appear nowhere outside your own doc line, `IMAGE_FETCH_TIMEOUT_MS` is 90s, the three named tools are all at exactly 60s, and the hosted path registers stdio at `real_agent.py:321`. All four write-downs are present, and the desktop `.mcpb` is hedged rather than asserted.

**One fix before merge.**

`docs/architecture.md` says two run logs carry the 180s timeout. That was exactly right when you wrote it, but main has since added two `stribling-father-1821` runs whose `image_transcribe` calls also timed out at 180s, so it reads as four the moment this merges. Dropping the count keeps it from going stale again:

```
You wrote:      (two committed e2e run logs
Change it to:   (committed e2e run logs
```

**And a correction from me, which does not block merge.**

I told you the 9.8% figure was correct and gave you the wording that is now in the spec. That was not a safe call, and you should know why before the number hardens into architecture.md.

`usage.timeline` records no `tool_use_id`. In 12 of the 28 runs with transcriptions, several `image-reader` subagents have a call outstanding at once — `stribling-father-1821/run-2026-08-18_02-03-06.json` runs 47 calls across 47 distinct `agent_id`s, seven of them in flight together. Where calls overlap, no method can tell which result belongs to which call, so a per-call figure is not recoverable there.

On the 16 runs where exactly one call is ever outstanding, the count is 4 of 32 (12.5%) by the issue's script and 5 of 32 (15.6%) if you pair each call to its own result. So 9.8% is the right order of magnitude — but "13 of 132 (9.8%)" is more precision than the corpus can carry, and it sits three paragraphs above a `p90 98s` it cannot coexist with (a p90 of 98s puts more than 10% over 60s).

Saying so makes the write-down stronger, not weaker, so if you want to fold it in:

```
You wrote:      corpus as of commit b065b687^ (2026-08-16), 13 of 132 healthy calls (9.8%) ran
                past 60s — a floor, since the corpus carries no bridge hop.
Change it to:   corpus as of commit b065b687^ (2026-08-16), roughly 10-15% of healthy calls ran
                past 60s — a floor, since the corpus carries no bridge hop, and a range because
                `usage.timeline` records no `tool_use_id`: in 12 of 28 runs several parallel
                `image-reader` subagents have a call outstanding at once, so per-call durations
                there cannot be recovered. The range is the spread across pairing methods on the
                16 runs where exactly one call is outstanding.
```

Equally fine to leave it and open a follow-up — the PR's conclusion holds under every method I tried, including the clean subset, where p90 is 67s.

Two housekeeping notes: the branch is 32 commits behind main, so it needs an update before it can merge, and this still needs a second approval — `image-transcribe.ts` pulls in the `*.ts` CODEOWNERS rule, so one of the two has to be a senior.

I did not verify the 60s cap itself; no suite reaches the bridge, as you said in the PR body.

promise-emmanuel (changes requested):
Review: PR #1718 — 1638-image-transcribe-60s-ceiling-writedowns, full pass, 6 commits behind main
Does: Adds three write-downs of one fact — the Cowork device bridge aborts every MCP call at 60s — to docs/architecture.md (an "Add an MCP tool" budget clause + an environment bullet), to image-transcribe-tool-spec.md §5.7, and to the comment above OCR_TIMEOUT_MS. No logic change; OCR_TIMEOUT_MS stays 180s.

Intent: Matches. All four items on issue #1638's checklist are present and land in the sections the issue named. Didn't change holds — no ocrTimeoutMs override, no image-reader retry. Nothing in the diff hits a stop-rule (no auth, no schema, no ADR reversal).

Checks: all 13 required green on head 56c505cb; strip skipped by design.

Still needs: one senior-developers approval. image-transcribe.ts pulls the *.ts CODEOWNERS rule; the docs paths match no rule (there is no * default). reviewRequests says senior-genealogists, which is stale — no genealogist-owned path is touched. codeowners/errors is empty. Also mergeStateStatus: BEHIND, so it needs an update before it can merge.

Verified on the test-merge: engine vitest 2438/2438, tsc --noEmit clean, harness pytest 2398 passed / 2 skipped, server pytest 209 passed / 1 skipped. A fresh worktree needed npm ci + npm run build first — environmental, not a regression. Mutation-tested the issue-ref ratchet in both directions: a #9999 in the spec fails it (130 tests, 1 failed), the same #9999 in image-transcribe.ts passes 489/489.

Existing review: chesworthrm APPROVED @56c505cb — same commit as head, so not stale. Their one requested pre-merge fix still stands, unapplied; their optional precision note also still stands.

[CRITICAL] (9/10) docs/architecture.md:1296 — a count that is already wrong on this head

Four run logs now carry the 180s timeout, not two: elena-asmundsdotter-origin, victoriano-macatangay-parents, and two stribling-father-1821 runs (5 calls total). This is exactly chesworthrm's requested fix, and it did not get made.


You wrote:      register the server over stdio and are **not** capped (two committed e2e run
Change it to:   register the server over stdio and are **not** capped (committed e2e run
[CRITICAL] (9/10) docs/specs/image-transcribe-tool-spec.md:362 — "today" has no date and the number has moved

The figure reproduces exactly at the commit it was written on and nowhere else: 13 of 144 (9.0%) at 6a32f70d, but 14 of 193 (7.3%) at this head — main brought in 8 more run logs. Every merge of main moves it again, and a reader has no date to tell how stale it is.


You wrote:      below. Re-running it today reports 13 of 144 (9.0%): the run-log retention step
Change it to:   below. Re-running it moves with the corpus — 13 of 144 (9.0%) at commit
                `6a32f70d`, 14 of 193 (7.3%) at this one: the run-log retention step
[CRITICAL] (8/10) docs/specs/image-transcribe-tool-spec.md:359 — the point figure is more precision than the corpus carries

I checked chesworthrm's claim independently and it holds. In 11 of the 28 runs containing a transcription, a second image_transcribe is issued before the first returns, so pairing each call to the next tool_result is guesswork there. That is why my re-run lands at 7.3% while the clean-subset methods land at 12.5–15.6%. And a point figure of 9.8% over 60s cannot sit three paragraphs above a p90 79s, which puts more than 10% over 60s by definition. The conclusion survives every method — only the two significant figures don't. Take the wording chesworthrm already offered:


You wrote:      corpus as of commit b065b687^ (2026-08-16), 13 of 132 healthy calls (9.8%) ran
                past 60s — a floor, since the corpus carries no bridge hop.
Change it to:   corpus as of commit b065b687^ (2026-08-16), roughly 10-15% of healthy calls ran
                past 60s — a floor, since the corpus carries no bridge hop, and a range because
                `usage.timeline` records no `tool_use_id`: in 11 of 28 runs several parallel
                `image-reader` subagents have a call outstanding at once, so per-call durations
                there cannot be recovered.
The matching figure in docs/architecture.md:1296 (9.8% of healthy calls exceed 60s) has to move with it, or the two documents disagree.

[INFO] (7/10) packages/engine/mcp-server/src/tools/image-transcribe.ts:80 — a ticket where the fact belongs

doc-links.test.ts bans a bare #1234 from specs on the rule "a spec owns the contract, GitHub owns the status" — and it cannot see .ts files, which I confirmed by mutation. This pointer already went stale once inside this PR: it was #1639 until #1639 was narrowed and the census moved to #1732. State the fact instead.


You wrote:      // this value. Whether the desktop `.mcpb` is bridged too is unverified (#1732);
Change it to:   // this value. Whether the desktop `.mcpb` is bridged too has not been measured;
[INFO] (6/10) CLAUDE.md § "Code reuse" — the one place a tool author reads that has no caveat

Its src/utils/http.ts bullet lists 180s, 90s and 60s as the budgets to pass, which is now the list of budgets this PR says are aspirational in Cowork. Outside issue #1638's checklist, so not a merge blocker — but it is the file that wins on conflict.

Not checked: the 60s cap itself. No suite reaches the device bridge, and I did not run a live Cowork session — the PR body says this, and it is honest. I also did not verify whether the desktop .mcpb is bridged; that is issue #1732's measurement and is correctly hedged throughout the diff.

Asimi1234:
Thanks Promise — all four actionable findings addressed in `adad0927`, and the branch is merged up to current main (clears the BEHIND).

- **[CRITICAL] architecture.md:1296 — stale run-log count.** Dropped the count: "(two committed e2e run logs …)" → "(committed e2e run logs …)", and "so a call ran" → "so calls ran". Same fix Richard asked for; it now can't go stale as more 180s runs land.
- **[CRITICAL] spec §359 / architecture.md — point figure is more precision than the corpus carries.** Both figures now read "roughly 10-15% of healthy calls" with the range rationale (no `tool_use_id` in `usage.timeline`; parallel `image-reader` subagents overlap in ~a dozen of the 28 runs, so per-call durations there aren't recoverable). Range is the clean-subset spread: 12.5% by the corpus-timeline script, 15.6% pairing each call to its own result. This also resolves the internal contradiction with the p90 you and Richard both flagged.
- **[CRITICAL] spec §362 — floating "today".** Replaced with commit-anchored provenance: the corpus-wide re-run "reports lower and keeps dropping as the corpus grows — 13 of 144 (9.0%) at commit `6a32f70d`, lower since". I re-ran the script on this head to confirm the drift you saw: **193 healthy / 14 over 60s (7.3%)** now — exactly your `14 of 193`. So the doc no longer quotes a corpus-wide number as *the* figure at all.
- **[INFO] image-transcribe.ts:80 — stale `#1732`.** Repointed to the fact: "Whether the desktop `.mcpb` is bridged too has not been measured".

One deliberate non-change: I kept `issue #1638` **out** of the spec wording — `image-transcribe-tool-spec.md` has an ISSUE_REF_BASELINE of 0, so a bare `#1638` fails the ratchet (same trap the earlier `#1639` hit). Phrased it as "the corpus-timeline script from the sizing issue" instead.

Left the **[INFO] CLAUDE.md Code-reuse** http.ts bullet alone — out of #1638's checklist as you noted, and the aspirational-in-Cowork caveat is already carried in architecture.md and the spec. Happy to fold a one-liner in if you'd rather it live in the file that wins on conflict.

Verification on the committed tree: `doc-links.test.ts` 130/130 (ratchet holds at baseline 20 for architecture.md, 0 for the spec), `no-issue-refs.test.ts` green, full packaging suite 494/494, `tsc --noEmit` clean. No suite reaches the 60s bridge itself — unchanged from the PR body.

promise-emmanuel (approved):
Approving — all four land, and I re-measured rather than take the numbers from the diff.

Verified on 9a0eff71 (level with main, so no test-merge needed): engine vitest 2462/2462, tsc --noEmit clean, doc-links.test.ts 130/130, harness pytest 2460, server pytest 209.

The range is the right call and it holds under a fresh count: 17 of the 28 runs never have two transcriptions outstanding, and on those it is 5 of 35 over 60s — 14.3%, inside your 10-15%. It also stops contradicting the p90 79s further down, which the point figure could not.

Two things that don't block. Fix them if you touch this again.

6a32f70d is a commit on this branch, not on main — git merge-base --is-ancestor says so. We squash-merge, so it stops resolving once this lands. A date survives and needs no lookup:


You wrote:      13 of 144 (9.0%) at commit `6a32f70d`, lower since
Change it to:   13 of 144 (9.0%) on 2026-08-19, lower since
And the 12.5% in the parenthetical has already moved — the same method on today's corpus gives 14.3%. The 10-15% range still covers it, so nothing is wrong, but those two point figures will drift the way the headline just did.

Issue #1054, opened 2026-07-31
Retain the hosted tool-call ledger, so guardrail bypasses are measurable in production


DallanQ:
> **Trigger clarified 2026-09-07 (icebox review).** The first thaw trigger below says "a tester outside the in-house group". The 2026-08-09 comment on this card records the alpha testers as external addresses already, so that wording is satisfiable both ways. Read it as: **the beta opens, or a tester not on the alpha roster runs a real project.** The other two triggers are unchanged. Route-around count is still three: issue #1869 and issue #1729 both say in their own bodies that they are unrelated to this ledger.

> **Blocker cleared — verified 2026-09-02 (`/fill-ready` stale-cross-reference sweep).** The first line below reads *"Do not start
> until #1558 reports a number."* **Issue #1558 closed COMPLETED 2026-08-25**, so that
> precondition has been met for over a week and nothing on the board reflected it. This
> card carries `icebox`, so whether it thaws is `/review-icebox`'s call, not a promotion
> — but it should be decided on the basis that its blocker is gone. It also gates the
> public rollout. Verify with `gh issue view 1558 --json state,stateReason`.

**Blocked on #1558 (2026-08-11).** Your ruling split the measurement half out into issue #1558 — adapt the hosted feedback transcripts we already hold and run the existing detectors over them. This issue keeps the durable-ledger half only (storage, retention window, capture hook, privacy sentence), and its own premise is that a non-zero hosted violation rate is what justifies building it. Do not start until #1558 reports a number.

> **Analysis 2026-08-09 — build later, on a trigger that can actually fire. One cheap thing is worth doing this week.**
>
> **Correcting a premise:** the alpha testers are **not** in-house. Recent feedback bundles come from `robert.raymond@familysearch.org`, `dr1matlock@yahoo.com`, `cmadsen1@gmail.com` and `untangleyourroots@gmail.com`. This is third-party family data today, which moves retention from an internal-tooling decision to one needing a policy sentence.
>
> **`docs/alpha-user-guide.md` currently promises "Sessions are private to you."** Retention makes that false as written. Correct it before anything ships, not before strangers arrive.
>
[...]

PR #1722, opened 2026-08-18
viewer: warn when research.json is hiding in a subfolder (#1317 bug 2)

francis-2008-happy:
## Summary
The in-repo, on-computer half of #1317 bug 2 (Normal tier). #1317 is closed; the durable fix — pinning the project folder so the agent and viewer share one — is Cowork-side / a lead decision. This PR owns the piece we can: make the mismatch **visible** instead of silent.

The Electron viewer watches only the **top-level** `research.json`, but the agent is handed a folder path per tool call and can write into a **subfolder** — so the top folder looks empty while the real project is one level down. To the user that reads as "lost files between sessions" (the reported tester bundle had exactly this shape). `watcher.ts` now scans for a nested `research.json` (bounded depth; skips `results/`, `_feedback/`, `node_modules`, hidden dirs) and surfaces a dismissible banner: *"this folder has research.json in a subfolder (…) — reopen the viewer on that folder."*

Round 7 found that the round-6 half of this reached the user never: the pointer was thrown into `setError`, and `error` was rendered by nothing (27 components read the context, none took it) — the same dropped-on-the-floor bug rounds 1-2 fixed for the notice. So this PR now also builds the surface `error` never had: an `ErrorNotice` bar mirroring `FolderNotice`, mounted in all three `AppContent` branches. That fixes the pre-existing invisible **"Not a research project"** rejection too, which has been silent since the viewer was extracted.

The picker's pointer is also unwrapped from Electron's `Error invoking remote method '<channel>': Error:` prefix in `IpcResearchTransport`, so the user reads the sentence rather than an IPC channel name.

Two shapes reach the user, because they fail at different points. When the top folder **has** a `research.json` next to the nested one, the banner fires. When the top folder has **none** — `research.json` genuinely hiding in a subfolder — `project:select-folder` rejects the folder *before* `startWatching` runs, so the banner can never reach it; that path now throws an error naming the subfolder instead of the generic "Not a research project" (round-6, @clack391).

## Review guidance

**Start here:** `packages/viewer-ui/src/App.tsx` `FolderNotice` and its two mount points in `AppContent`, plus the new `packages/viewer-ui/src/__tests__/App.test.tsx`. The decision that mattered — and the bug that survived two rounds — is that the banner has to be **mounted** in both `AppContent` branches (research loaded and not), not just correct as a component. The delivery path behind it: a dedicated `project:folder-notice` IPC channel → `onNotice` on `SubscriptionHandlers` → provider `notice` state that `onResearch` does **not** clear. The first cut sent this on `project:watch-error`, which no component renders and `onResearch` cleared anyway, so it reached the user never.

**Unsure about:** Whether a genealogist would word the two bars' copy differently — `formatNestedPicker` (the rejected-pick error) and `formatNestedNotice` (the banner) are the user-facing strings, and I have proven they render but not that they read well in a real window.

## Plan

**Didn't change:** The viewer still watches a single folder — I did not make it follow the agent into the subfolder or auto-switch. That's the same folder-mismatch root cause whose durable fix is Cowork-side; auto-following would paper over it in one of the two viewers and hide the signal the lead decision needs. The banner's amber is inline-styled, not themed — left as cosmetic.

Also not changed, deliberately: the channel-name-as-two-unrelated-string-literals fragility (@chesworthrm round-5, agreed out of scope — a shared `channels.ts` imported by main and preload is its own change), and the hydration replay above (#1869, counted).

**Acceptance check:** `pnpm --filter @genealogy/viewer-ui exec vitest run src/__tests__/App.test.tsx` — both cases fail on a tree with the `<FolderNotice />` mounts removed from `App.tsx` (the round-2 regression) and pass on this branch. `FolderNotice.test.tsx` keeps passing under that same mutation, which is why the new file is separate rather than an addition to it.

Round-7 addition: deleting the three `<ErrorNotice />` mounts fails the three `App.test.tsx` error cases while the folder-notice cases stay green; deleting each mount **independently** fails only its own case (per-branch pinned, as @chesworthrm established for `FolderNotice`). The picker's depth bound fails its test if raised **or** if the new `maxDepth` parameter is ignored — both run.

Round-6 addition: `npx vitest run src/main/__tests__/watcher.test.ts` in `apps/electron` — deleting the `findNestedResearchJson` lookup from `assertResearchProject` fails *"points at the subfolder when only a nested research.json exists"* and nothing else. The two `formatNestedPicker` tests stay green under that mutation, which is the same message-works-vs-message-is-*reached* split as rounds 3-5: the pure-function test cannot pin the wiring, so the gate gets its own test.

## Scope / safety
Electron `main` + the shared viewer notice path. One new receive-only IPC channel, `project:folder-notice`, mirroring an existing one. **Correcting the earlier claim in this body:** the preload *did* change (that is where the new channel is registered), and security-invariants §3 *is* an invariant in play — its rule is that a channel in the preload but absent from the §3 table is unauthorized. The row is now added. No CSP/webPreferences change and no new dependency. Folder-switch guard (a late scan can't warn about a folder you left) and `selectFolder` now clears `notice`; path list capped at 3 + "and N more" (`formatNestedNotice`). Web transport has no notice case — hosted pins `cwd=/project` and can't hit this.

## Verification
- `turbo test typecheck --force` — 8/8 tasks green on the merged head: `packages/viewer-ui` **225**, `apps/electron` **97**, `apps/web` **10**, all typechecks clean. Branch is current with `main` (0 behind).
- The `window.api` test stub is typed `AppAPI`, so adding a preload channel without the stub now fails typecheck (it had already drifted once). Two dead stub entries removed (`getProjectState`, `selectProjectFolder` — on no interface, referenced nowhere).
- `security-invariants` §7 now scans `packages/viewer-ui/src/` as well as `src/renderer/`; it had been passing without reading any of the extracted UI.
- **Not verified on Windows.** Round 6 shipped a test asserting a path separator via regex — green on Linux CI, red on Windows, and there unable to tell "the feature works" from "the feature is gone" (@johnmarkpeterbrown). Now built from `join()` like the precedent in the same file, but the platform claim is his measurement, not one I re-derived; I have no Windows box.
- **Not launched.** Every round of this PR has been verified by tests and mutations, not on screen. Both bars are proven to become DOM nodes in every branch; neither has been seen in a real window.
- Prettier clean on every changed `*.ts`/`*.tsx` (`watcher.ts` had warnings; fixed).

## Known limitation — the notice is not replayed on hydration
Corrected from "dev-only", which undersold it (@clack391): the notice is not carried in `getCurrentState()`/`getProjectState()`, so **any** renderer reload drops it, not just the dev launch. The `--project-dir` launch (`make dev-sample` / `make dev-debug-sample`) is where you will actually meet it — `startWatching` runs at window creation, before the renderer mounts and subscribes. So if you eyeball this via `dev-sample` and see no banner, that is this gap, not a broken banner. (Thanks @chesworthrm for tracing it.)

Tracked, with the durable folder-pin, in **#1869**. Not fixed here: it changes the shared `ProjectStateSnapshot` contract that the web transport implements too — call sites opened and counted, roughly 6-8 files across 3 packages (`transport.ts`, `ResearchDataProvider.tsx:105`, both transports, `watcher.ts`/`index.ts`, contract mocks and tests). Too big for a PR at its sixth round with a shape two reviewers have signed off on.

## For review
`apps/electron/src/main/**`, `apps/electron/src/preload/**`, `apps/electron/src/renderer/**`, `packages/viewer-ui/**`, `apps/web/**`. Every changed path is `*.ts`/`*.tsx`, so CODEOWNERS requires **both** senior-developers and senior-genealogists — both reviews are owed.

## Follow-on issues
**#1869** — pin the project folder so the agent and viewer share one (the #1317 bug-2 root cause), plus carrying the folder notice through hydration. Filed rather than folded in for the two reasons the contributing guide requires naming: the durable fix is **Cowork-side and rests on a decision only the lead can make** (last discussed on #1054/#1317, not reachable at time of filing), and the hydration half is **too big for this PR — call sites opened and counted**, ~6-8 files across 3 packages changing a shared cross-package contract. Labelled `nothing-checks`: all suites are green with the notice un-replayed.


chesworthrm (commented):
Reviewed the full diff against a test-merge with main (the branch is 19 commits behind; the merge is clean). All 14 checks are green on `46001ad8`. I ran the electron vitest (69 pass), `turbo test typecheck` (8/8 tasks), `typecheck:node`, and `apps/server` pytest (162 pass) on the merged tree — all green. I also reproduced your mutation check: stubbing `findNestedResearchJson` to return `[]` fails the "finds a research.json in a subfolder" test.

The walker itself is good. The tests are well shaped. My problem is with where the message goes.

## The warning never reaches the user

`project:watch-error` goes preload → `IpcResearchTransport` → `ResearchDataProvider`, and lands in:

```ts
const [error, setError] = useState<string | null>(null)
```

That `error` is put on the context and **read by no component.** I grepped all of `packages/viewer-ui/src` for `error` and `clearError` — the only hits are the type declaration, the state itself, the two context-value objects, and the test mock. `git log -S"{error}"` over `packages/viewer-ui` and `apps/electron` returns nothing. It has never been rendered.

So the user sees nothing, and the thing the PR sets out to do — make the mismatch visible instead of silent — doesn't happen.

The same is true of the three sends that were already on that channel (`research.json deleted`, `Error reading …`, `Watcher error: …`). Those are invisible too. That part is not yours, but it means the channel is not a UI surface at all.

The Scope/safety section lists **"No new IPC channel (reuses `project:watch-error`)"** as a safety property. That is the constraint that costs the feature its effect. Reusing a channel nothing renders isn't safer — it's silent.

## Even once it renders, the next file read wipes it

`ResearchDataProvider`'s `onResearch` handler calls `setError(null)`. You start chokidar with `ignoreInitial: false`, so the initial `add` for the top-level `research.json` fires right after your scan resolves. I measured the send order on a folder shaped like the reported bundle (top-level `research.json` plus `subproject/research.json`):

```
t=1ms  project:watch-error      "Heads up: this folder also has research.json in a subfolder…"
t=1ms  project:research-updated {"a":1}
```

The heads-up arrives first, then the research update clears it. On a small project folder — the common case, and the reported one — the warning always loses that race, and every later write to `research.json` clears it again.

So the fix isn't just "render `error`". The notice needs its own state that a successful read does not clear.

## The edit

Give it its own channel and its own state.

You wrote:
```ts
      mainWindow.webContents.send(
        'project:watch-error',
```

Change it to:
```ts
      mainWindow.webContents.send(
        'project:folder-notice',
```

Then carry it through the sites that channel already occupies:

- `apps/electron/src/preload/index.ts:13` — add `onFolderNotice` next to `onWatchError`, and add `ipcRenderer.removeAllListeners('project:folder-notice')` to `removeAllWatchListeners` at line 22.
- `apps/electron/src/preload/index.d.ts:14` — the matching type.
- `packages/viewer-ui/src/transport.ts:65` — add `onNotice: (message: string) => void` to `SubscriptionHandlers`, next to `onError`.
- `packages/viewer-ui/src/contract.ts:31` — add `onNotice: () => {}` to the subscribe call.
- `apps/electron/src/renderer/src/transport/IpcResearchTransport.ts:29` — wire it.
- `apps/web/src/transport/WsResearchTransport.ts:48` — needs an `onNotice` too or the contract test fails. A no-op is right; the hosted path pins `cwd=/project` and can't hit this.
- `packages/viewer-ui/src/contexts/ResearchDataProvider.tsx` — a `notice` state set from `onNotice`. Do **not** clear it in `onResearch`. That is what the `error` state does, and it is why the current message would vanish.
- `packages/viewer-ui/src/App.tsx` — render it. `WaitingScreen` is the natural home: it already says *"If your research is somewhere else, open that folder instead"*, and this is the precise version of that sentence. It also has to show when `research` **is** present, because the reported bundle has a non-empty top-level file.

## Two smaller ones, same PR

**A folder switch during the scan sends the old folder's warning.** Nothing checks that the resolved scan still belongs to the watched folder. Open folder A, switch to B before the walk finishes, and B's viewer gets A's subfolder list. The walk took 46ms over 557 directories in this repo, so the window is small but real.

You wrote:
```ts
    .then((nested) => {
      if (nested.length === 0) return
```

Change it to:
```ts
    .then((nested) => {
      if (nested.length === 0) return
      if (currentFolderPath !== resolve(folderPath)) return
```

**The path list is unbounded.** Every hit gets interpolated into the message. Scanning this repo produced 91 hits. A user who opens a parent folder holding several projects gets all of them in one string.

You wrote:
```ts
      const list = nested.map((p) => `"${p}"`).join(', ')
```

Change it to:
```ts
      const list =
        nested.slice(0, 3).map((p) => `"${p}"`).join(', ') +
        (nested.length > 3 ? ` and ${nested.length - 3} more` : '')
```

## Two notes, not blocking

The PR body doesn't use the template — no **Start here**, no **Unsure about**, no **Acceptance check**. Your Verification section covers most of what Acceptance check wants. **Unsure about** is the line most likely to have surfaced "does anything actually render this?" before review.

`#1317` is now closed (2026-08-18, completed), so "Part of #1317. Does not close it" points at a closed issue and nothing tracks the remainder.

## What I didn't check

I didn't launch the app. The two blocking findings are traced statically through the full chain plus a measured IPC send order, not observed on screen. I didn't run the eval harness pytest — this diff doesn't touch it and CI is green on it.

The security claim checks out: `watcher.ts` only, no preload/CSP/`webPreferences` change. The bounded walk is correct — symlinked directories report `isDirectory() === false` under `withFileTypes`, so there's no cycle risk.

chesworthrm (changes requested):
Reviewed the full diff, merged against current `main` — your branch is 27 commits behind, so I tested the merge rather than the branch head. All suites pass either way, and I reproduced your mutation check. The scan helper and its four tests are good work: bounded, cheap, and the skip list is right.

Two blocking problems. They're really one problem in two places: **the warning never reaches the user.**

---

### 1. Nothing displays `project:watch-error`

The message travels main → preload → `IpcResearchTransport` → `ResearchDataProvider`, where it lands in `setError(...)`. Nothing ever reads that state.

There are 24 places that destructure `useResearchData()`. None of them takes `error`. Nothing calls `clearError` either.

So the heads-up is computed correctly, sent correctly, and then dropped on the floor. The thing this PR is for — making the mismatch visible instead of silent — doesn't happen yet.

To be clear: this isn't a bug you introduced. That surface has been missing since the viewer was extracted. But your change depends on it, so it has to be built here.

### 2. Once you add a surface, the first file load will wipe the message

`ResearchDataProvider`'s `onResearch` handler calls `setError(null)` every time `research.json` loads.

Your warning is sent before that load on a normal-sized project folder, so it gets cleared a few milliseconds later. I measured it on a realistic folder shape — research.json, tree.gedcomx.json, a results/ directory, and a nested project. The warning was cleared on all eight runs.

And that load always happens. The folder picker refuses any folder that has no top-level `research.json`, so there is always one to load.

One caveat worth knowing: the order flips on a very large folder, where the scan takes longer than the file load. So this is a race, not a certainty. It just happens to lose on every folder a real user would open.

### What to do about 1 and 2

Send this hint on its own IPC channel rather than reusing `project:watch-error`, and render it somewhere the research load doesn't clear.

`WaitingScreen` won't work on its own. It only renders when there's no research at all, and your case always has research — that's exactly what makes the bug confusing.

I'm not writing replacement lines for this one. Which surface it lands on is a design decision rather than a mechanical edit, and it's yours to make.

---

### Smaller things

**Switching folders can warn about the folder you just left.** `stopWatching()` doesn't cancel a scan that's already running, and the callback never re-checks which folder is current. Open folder A, switch to B while A is still scanning, and you get a warning naming A's subfolders while looking at B.

```
You wrote:      .then((nested) => {
                  if (nested.length === 0) return
Change it to:   .then((nested) => {
                  if (nested.length === 0) return
                  if (currentFolderPath !== resolve(folderPath)) return
```

**Cap the path list.** Every hit goes into one string, so a folder holding many projects produces a message no surface can show sensibly.

```
You wrote:      const list = nested.map((p) => `"${p}"`).join(', ')
Change it to:   const list = nested.slice(0, 3).map((p) => `"${p}"`).join(', ') +
                  (nested.length > 3 ? ` and ${nested.length - 3} more` : '')
```

**#1317 is closed.** It was closed as completed by #1703 earlier today, so "Part of #1317 (bug 2). Does not close it." now points at a closed issue and nothing tracks bug 2. Either point the body at a live issue or say plainly that bug 2 sits with the lead decision you describe.

---

I also checked this against #1703 for overlap and there is none. Its folder-naming copy only fires when there's no research at all, so the two are complementary. Worth saying since the titles look close.

Nothing here touches schema, auth, or an ADR, and I agree the Electron security checklist has no invariant in play.

The scan half is ready. It's the delivery half that needs to land before this does anything for the user.

chesworthrm:
Correction to my review above, and one new fact.

**The "27 commits behind" line is wrong.** I resolved the branch from a stale local ref and reviewed `46001ad8`. The actual head, `b5637bfd`, is a merge of `main` into this branch and is **0 commits behind**. What I tested was `46001ad8` + `main`, which is equivalent in content to the real head, so the findings are unaffected — but the framing was wrong and the test-merge I described was redundant.

**Use the earlier review, not this one, for the edit list.** My review restates findings already made in the 11:51 review on `46001ad8`, and that one is more complete: it names every site the new channel has to be carried through — preload, `preload/index.d.ts`, `transport.ts`, `contract.ts`, both transports, the provider state, and `App.tsx`. Treat mine as the blocking restatement of it rather than a second opinion.

**The new fact, and the reason I'm not withdrawing the block:** I re-checked directly against `b5637bfd`, and nothing from that review has been addressed. At the current head the watcher still sends to `project:watch-error`, the path list is still unbounded, there's still no folder-switch guard, and `onResearch` still calls `setError(null)`. The merge from `main` is the only change since the review.

So the delivery half is still open, and it's the half that decides whether this PR does anything for the user.

francis-2008-happy:
Thanks — you were exactly right: the warning reached the user never. Fixed end to end in the latest push.

**Delivery (the blocking half):**
- New dedicated `project:folder-notice` IPC channel — preload `onFolderNotice` (+ its `removeAllListeners`), a new `onNotice` on `SubscriptionHandlers`, wired through `IpcResearchTransport`. Not the dead `project:watch-error` reuse.
- The provider now has its own `notice` state, set from `onNotice` and **not** cleared by `onResearch` (that clearing was the bug). Added `clearNotice`.
- `App.tsx` renders a dismissible `FolderNotice` banner, shown **whether or not research is present** — the reported bundle has a non-empty top-level `research.json`, so WaitingScreen-only would never fire, as you noted.
- Carried through every seam you listed: `transport.ts`, `contract.ts`, both transports (web is a no-op — hosted pins `cwd=/project`, commented), `ResearchDataContext`, `ResearchDataProvider`, `mockContext`.

**The two smaller ones:**
- Folder-switch guard: the scan `.then` now bails if `currentFolderPath !== resolve(folderPath)`.
- Path cap: extracted a pure `formatNestedNotice` (first 3 + "and N more"), unit-tested.

**Tests:** a provider regression test — `onNotice` shows it, a following `onResearch` does **not** clear it, dismiss does — and I mutation-checked it (adding `setNotice(null)` to `onResearch` fails it). Plus the `formatNestedNotice` cap test. viewer-ui 216 green, electron watcher 23 green, electron/web/viewer typecheck clean.

**PR body:** repointing it off the now-closed #1317 — this is the in-repo, on-computer half of bug 2; the durable folder-pin sits with the lead (the #1054/#1317 discussion), not a closed issue.

chesworthrm (changes requested):
Re-reviewed against a test-merge with `main` (one commit behind, merge is clean — it only brings `eval/` runlogs). All required checks are green on `1cdb0687`. I ran `apps/electron` vitest (71 pass), `packages/viewer-ui` vitest (216 pass), and `turbo test typecheck --force` (8/8 tasks) on the merged tree.

**Both blocking findings are fixed.** `eb0de5fe` does the delivery end to end, and it's the right shape: a dedicated `project:folder-notice` channel rather than the invisible `project:watch-error`, `notice` as its own state, and the banner rendered in *both* `AppContent` branches — not just `WaitingScreen`. You went further than my note there and you were right to; the reported bundle has a non-empty top-level file, so the `WaitingScreen`-only surface I suggested first would have missed the actual case.

I also checked that the stickiness is real rather than incidental. Reintroducing `setNotice(null)` in `onResearch` fails your new provider test. That test earns its place.

Both smaller items landed as specced (the folder-switch guard, the 3-path cap), and the `#1317`-is-closed note is handled — saying plainly that the durable fix is a lead decision is the right call.

One thing to fix before this merges.

---

### The banner itself is untested

I deleted both `<FolderNotice />` lines and ran the full viewer-ui suite. All 216 tests still passed.

So the half that was broken last round — the message not reaching the user — is still not guarded by anything. Your provider test proves the state survives a research load. Nothing proves the state ever becomes pixels. No test in the repo renders `<App />`.

Two small things block writing that test. `FolderNotice` isn't exported, and `buildMockContext` has no way to set a notice.

```
You wrote:      function FolderNotice(): React.JSX.Element | null {
Change it to:   export function FolderNotice(): React.JSX.Element | null {
```

```
You wrote:          getById?: ResearchDataState['getById']
Change it to:       getById?: ResearchDataState['getById']
                    notice?: string | null
```

Then pass it through in the returned object:

```
You wrote:      notice: null,
Change it to:   notice: overrides.notice ?? null,
```

That gives you the same `vi.mock` + `buildMockContext` pattern `ProjectOverview.test.tsx` already uses. Two assertions are enough: the message renders when `notice` is set, and the dismiss button calls `clearNotice`.

The folder-switch guard you added in `watcher.ts` is also untested, for the same reason the other watcher tests stop where they do. I'm not asking for that one — it needs chokidar and a `BrowserWindow`.

---

### Two notes, not blocking

**The notice never arrives on the `--project-dir` launch.** `createWindow()` doesn't wait for the page to load, and `startWatching` runs on the next line. The scan finishes in about 50ms. The renderer only registers its listener when React mounts and calls `subscribe()`, which is much later. Nothing stores the notice — `getCurrentState()` returns only `folderPath`, `research` and `gedcomx` — so unlike a research update, a missed send is gone for good.

This is not an end-user path, but it is `make dev-sample` and `make dev-debug-sample`, which is how someone would try to eyeball this feature by hand. Worth a line in the body so the next person doesn't read a blank screen as a broken banner. I traced this from the source; I didn't launch the app.

**The body still doesn't use the template.** No **Start here**, **Unsure about**, **Didn't change**, **Acceptance check**, or **Deviated from the plan**. Your Verification section covers most of what Acceptance check wants, so this is mostly mechanical now — but **Unsure about** is the line that would have surfaced "is the banner actually tested?" before review, which is the one finding left.

---

### What I didn't check

I didn't launch the app, so the `--project-dir` point is static reasoning rather than something I watched happen. I didn't run the eval harness or `apps/server` pytest — this diff touches neither and both are green in CI. I didn't look at how the banner's amber reads against the light and dark themes; it's inline-styled rather than themed, which I took as cosmetic.

Nothing here touches schema, auth, or an ADR. The scan and the delivery are both right. It's the test that's missing.

chesworthrm (changes requested):
Re-reviewed on `a82fce6e`. The branch is up to date with `main`, so I reviewed the head directly — no test-merge needed this time. All required checks are green. I ran `packages/viewer-ui` vitest (219), `apps/electron` vitest (71), and `turbo test typecheck --force` (4/4 and 5/5) on your head.

**Everything from the last round landed except one thing, and that one is my fault.**

The `--project-dir` limitation is documented, the branch is current, and the banner now has tests. Thank you for the clear write-up on each.

One blocking item remains.

---

### The banner test doesn't pin that the banner is mounted

I deleted both `<FolderNotice />` lines from `App.tsx` and ran the full viewer-ui suite. `FolderNotice.test.tsx` passed all 3.

The difference is which mutation gets applied. You wrote:

> Mutation-checked it the way you'd want: stubbing `FolderNotice` to `return null` fails 2 of the 3

Stubbing the component's body proves the component works. It can't prove that anything renders it. Your test renders `<FolderNotice />` directly, so it passes whether or not `App` mounts it — and mounting it is the exact half that was broken two rounds ago.

**This is on me.** I told you two assertions were enough and pointed you at the `ProjectOverview.test.tsx` pattern. You built precisely that. That pattern structurally cannot catch this deletion, and I should have caught that when I wrote the ask.

Keep `FolderNotice.test.tsx` exactly as it is — it earns its place on the component's own behaviour. Add a second file that renders `App`. I ran this one on your head (221 pass) and under the mutation (both cases fail, and `FolderNotice.test.tsx` keeps passing, which is the point):

Add `packages/viewer-ui/src/__tests__/App.test.tsx`:

```tsx
import { describe, it, expect, vi, beforeEach } from 'vitest'
import { render, screen } from '@testing-library/react'
import type { ResearchTransport } from '../transport'

vi.mock('../contexts/ResearchDataContext', async () => {
  const actual = await vi.importActual<typeof import('../contexts/ResearchDataContext')>(
    '../contexts/ResearchDataContext'
  )
  return { ...actual, useResearchData: vi.fn() }
})

import { useResearchData } from '../contexts/ResearchDataContext'
import { buildMockContext } from '../contexts/__tests__/mockContext'
import App from '../App'

const stubTransport = {
  getProjectState: async () => ({ research: null, gedcomx: null, label: null }),
  subscribe: () => () => {},
  readSidecar: async () => null,
  submitFeedback: async () => ({ ok: true }),
  openExternal: () => {}
} as unknown as ResearchTransport

// Pins that App MOUNTS the banner, in both AppContent branches. FolderNotice.test.tsx
// renders the component directly, so it passes even if App never renders it.
function mockCtx(notice: string | null, withResearch: boolean): void {
  const base = buildMockContext(withResearch ? { notice } : { notice, research: null })
  vi.mocked(useResearchData).mockReturnValue({ ...base, folderPath: '/p' })
}

describe('App mounts the folder notice', () => {
  beforeEach(() => vi.clearAllMocks())

  it('shows the banner when research IS loaded', () => {
    mockCtx('subfolder heads-up', true)
    render(<App transport={stubTransport} />)
    expect(screen.getByRole('status')).toHaveTextContent('subfolder heads-up')
  })

  it('shows the banner when research is NOT loaded', () => {
    mockCtx('subfolder heads-up', false)
    render(<App transport={stubTransport} />)
    expect(screen.getByRole('status')).toHaveTextContent('subfolder heads-up')
  })
})
```

The `{ ...base, folderPath: '/p' }` spread is load-bearing. `buildMockContext` has no `folderPath` override, and without it `AppContent` takes the `WelcomeScreen` branch, where neither banner is reachable. No change to `mockContext.ts` needed.

Then also fix the comment, which currently claims the property the file doesn't have:

```
You wrote:      // that dismiss clears it, so deleting <FolderNotice /> can't pass green again.
Change it to:   // that dismiss clears it. App.test.tsx pins that App actually mounts it.
```

---

### Three notes, not blocking

**The banner asks the user to do something the app doesn't allow.** It ends *"open that folder instead"*, and once a folder is open there is no way to open another one. `WaitingScreen` has no picker, the File menu is close/quit only, and `folderPath` is never set back to null — so `WelcomeScreen`, which holds the one and only folder button, is gone for the rest of the session. The only way out is to quit and relaunch.

I checked this while tracing something else, and it surprised me. It doesn't block: telling the user why the viewer looks empty is a real gain over silence, which is what this PR set out to do. But the sentence promises a control that isn't there.

```
You wrote:      `The viewer only shows the top-level project — if your research is in the ` +
                `subfolder, open that folder instead.`
Change it to:   `The viewer only shows the top-level project — if your research is in the ` +
                `subfolder, reopen the viewer on that folder.`
```

`WaitingScreen` already carries the same sentence, so this is not something you introduced, and the wider "no way to switch folders" gap is well outside this PR.

**A folder switch wouldn't clear the notice.** `selectFolder` clears `error` and not `notice`, and nothing else clears it. That's unreachable today for the reason just above, so this is only hardening for whenever a folder-switch control does appear. One line, and the suite stays green with it (219 pass):

```
You wrote:              setFolderPath(path)
                        setError(null)
Change it to:           setFolderPath(path)
                        setError(null)
                        setNotice(null)
```

**The body still doesn't use the template.** No **Start here**, **Unsure about**, **Didn't change**, **Acceptance check**, or **Deviated from the plan**. Your Verification and Known-limitation sections already cover most of what Acceptance check wants, so this is mechanical. I keep raising **Unsure about** because it is the line that would have caught this round's finding: "is the banner actually mounted?"

---

### Confirmed good

The scan is right: 91 hits in 47ms on this repo, so your `~90` comment is accurate and the 3-path cap does its job. The dedicated channel, the separate `notice` state, the folder-switch send guard, and rendering in both `AppContent` branches are all the right calls.

### What I didn't check

I didn't launch the app — everything here comes from tests run against your code, including the folder-picker point, which I found by rendering `App` rather than by clicking a real window. I took your `--project-dir` trace on trust rather than re-deriving it. I didn't run the eval harness or `apps/server` pytest; this diff touches neither and both are green in CI. I didn't look at the banner's amber against the light and dark themes — it's inline-styled rather than themed, which I'm still treating as cosmetic.

Nothing here touches schema, auth, or an ADR. One blocking item, and it's a new test file.

chesworthrm (changes requested):
Re-reviewed on `6dd0242c`. The branch is current with `main` (0 behind), so I reviewed the head directly. All 14 checks are green. I ran `turbo test --force` (4/4 tasks — viewer-ui 221, electron 71, web 10) and `turbo typecheck --force` (5/5) in a clean worktree at your head.

**Last round's blocking item is properly fixed, and it's stronger than the body claims.** I ran the mutation from your Acceptance check — both `<FolderNotice />` mounts deleted — and got exactly what you describe: both `App.test.tsx` cases fail, `FolderNotice.test.tsx` stays green at 3. Then I went one step further and deleted the mounts *one at a time*. Removing only the `!research` mount fails only the "NOT loaded" case; removing only the research-loaded mount fails only the "IS loaded" case. Each case pins its own `AppContent` branch independently. That's a better test than "both fail together," and worth knowing you have.

I also re-checked the rest of the round-4 list. The `FolderNotice.test.tsx` comment now describes what the file actually guards and points at `App.test.tsx` for the other half. `selectFolder` clears `notice`. The body is on the template. The `~90` comment is accurate — I measured 91 hits in 82ms on this repo. Symlink cycles don't loop; I built a self-referential tree and the walk finished in 2ms. Depth cuts at 6 subdirectory levels, as named.

One blocking item, and it's the same shape as the last three rounds — one hop upstream.

---

### The Electron-to-viewer seam has no test

I deleted this line from `apps/electron/src/renderer/src/transport/IpcResearchTransport.ts`:

```ts
    window.api.onFolderNotice((message) => handlers.onNotice(message))
```

All 302 tests stayed green. Typecheck stayed clean. The banner is dead — the notice never reaches the provider, so it never becomes pixels.

`App.test.tsx` now guards state to pixels. Nothing guards IPC to state. `IpcResearchTransport.test.ts` only runs `assertTransportContract`, which passes on shape alone — the stub's `onFolderNotice: () => {}` satisfies it whether or not `subscribe` ever calls it.

This is not the chokidar/`BrowserWindow` test I declined last round. It needs neither. I ran the version below on your head (72 pass) and under the deletion above (fails on the new case, contract case still passes).

```
You wrote:      import { describe, it, beforeEach } from 'vitest'
Change it to:   import { describe, it, expect, vi, beforeEach } from 'vitest'
```

```
You wrote:      function installApiStub(): void {
Change it to:   let folderNoticeCb: ((m: string) => void) | null = null

                function installApiStub(): void {
                  folderNoticeCb = null
```

```
You wrote:          onFolderNotice: () => {},
Change it to:       onFolderNotice: (cb: (m: string) => void) => {
                      folderNoticeCb = cb
                    },
```

Then a second case inside the existing `describe`:

```ts
  // Pins the seam between Electron IPC and the viewer: main's
  // `project:folder-notice` must land on `handlers.onNotice`. The contract test
  // above only checks shape, so deleting the wiring line leaves it green.
  it('routes a folder notice to handlers.onNotice', () => {
    const onNotice = vi.fn()
    new IpcResearchTransport().subscribe({
      onResearch: () => {},
      onGedcomx: () => {},
      onSidecar: () => {},
      onError: () => {},
      onNotice
    })
    folderNoticeCb!('research.json is in a subfolder')
    expect(onNotice).toHaveBeenCalledWith('research.json is in a subfolder')
  })
```

---

### Carried forward from my last round

The banner wording note didn't land, and it isn't in **Didn't change** either — I think it just fell through rather than being declined, but say so if you decided against it.

I re-confirmed the gap before re-raising it: `WaitingScreen` has the text but no picker button, and `selectFolder` is reachable only from `WelcomeScreen`, which is gone for the rest of the session once `folderPath` is set. So the banner still ends by asking for a control that isn't there. Two lines this time, because a test pins the string:

```
apps/electron/src/main/watcher.ts:108
You wrote:      `subfolder, open that folder instead.`
Change it to:   `subfolder, reopen the viewer on that folder.`
```

```
apps/electron/src/main/__tests__/watcher.test.ts:181
You wrote:      expect(msg).toContain('open that folder instead')
Change it to:   expect(msg).toContain('reopen the viewer on that folder')
```

---

### Two notes, not blocking, nothing for you to do here

**The channel name is two unrelated string literals.** I typo'd `'project:folder-notice'` to `'project:folder-notic'` in `preload/index.ts` while `watcher.ts` kept sending the correct one: 302 tests green, typecheck clean. That's round 1's wrong-channel bug in a form nothing catches. It's the existing pattern for all three older channels — `channelMap` covers only the two file channels — so this is not yours to fix in this PR. Closing it properly means a shared `channels.ts` imported by both main and preload, which is its own change.

**Your routing line understates the reviewers.** The body says these paths go to senior-developers. Every changed path is `*.ts`/`*.tsx`, and that CODEOWNERS rule requires **both** senior-developers and senior-genealogists, so senior-genealogists is still owed a review on top of mine.

---

### What I didn't check

I didn't launch the app. Everything above comes from mutations run against your code in a worktree at `6dd0242c`, restored clean after each one. I took your `--project-dir` trace on trust rather than re-deriving it. I didn't run the eval harness or `apps/server` pytest — this diff touches neither and both are green in CI. I still haven't looked at the banner's amber against light and dark themes; still treating that as cosmetic.

One thing I chased and dropped: I thought the banner might be unreachable in the shape where the top folder has *no* `research.json`, since the picker rejects that folder before `startWatching` runs. That's the property I established in round 1 and relied on, not an oversight — the reported bundle has a non-empty top-level file. Withdrawn.

Nothing here touches schema, auth, or an ADR. The scan, the delivery, and now the mount are all right. It's the hop between them.

chesworthrm (approved):
Round-5 re-review — approving. Every blocking item from rounds 1–5 is resolved and verified on the current head (`db2c1eaa`), not just read:

- Delivery reaches the user: dedicated `project:folder-notice` channel → `notice` state → banner mounted in both `AppContent` branches. Confirmed by deleting both `<FolderNotice />` mounts (App.test.tsx fails, FolderNotice.test.tsx stays green) and by deleting the IPC wiring line (the seam test fails).
- Stickiness is real: injecting `setNotice(null)` into `onResearch` fails the provider regression test.
- Folder-switch guard, 3-path cap (`formatNestedNotice`), `selectFolder` clearing notice, and the "reopen the viewer on that folder" wording all landed.

Suites on this head in a clean worktree: viewer-ui 221, electron 72, typecheck 5/5. All 14 CI checks green.

The channel-name-as-two-literals fragility stays out of scope by agreement (a shared `channels.ts` is its own change).

Note: this clears the developer side only. Every changed path is `*.ts`/`*.tsx`, so CODEOWNERS requires a senior-genealogists review too before merge.

clack391 (changes requested):
Reviewed as the second (senior-dev) review. The implementation is solid: I reproduced the suites (viewer-ui 221, electron 72, web 10, typecheck all green), and independent checks confirmed the IPC channel is receive-only and safe, the banner renders escaped text, the scanner is symlink- and depth-safe, and all four new tests are real guards (each fails when its target is reverted). The mount-in-both-branches pin and the notice-not-cleared regression test are well done.

Requesting changes on three things before I approve.

**1. The banner does not fire for the case the title describes most literally.**

It only fires when the top folder already has a `research.json` next to the nested one. If the top folder is empty and the real project is one level down (research.json genuinely "hiding in a subfolder"), `project:select-folder` rejects the folder first, so the scan never runs:

```
apps/electron/src/main/index.ts:207-214
  try { await fs.stat(path.join(folderPath, 'research.json')) }
  catch { throw new Error('Not a research project - research.json not found. ...') }
```

That user sees "Not a research project" and never gets the banner. Two ways to resolve, either is fine:

- Confirm the tester bundle had a top-level `research.json` (the summary says so). If it did, say so here and I will treat the empty-top case as out of scope.
- Better: call `findNestedResearchJson(folderPath)` in that catch and, if it finds one, throw something like "research.json is in a subfolder (X) - reopen the viewer on that folder" instead of the generic error. That reuses your new scan and covers the reported case.

**2. Add the new IPC channel to the security allowlist.**

`project:folder-notice` is in the preload but not in the security-invariants §3 table, which says to update the list when adding channels and treats an unlisted preload channel as unauthorized. The next security run will flag it.

- Add a row to `apps/electron/.claude/skills/security-invariants/SKILL.md` §3:
  `| \`project:folder-notice\` | on (push) | Nested research.json heads-up string |`
- The summary says "no preload/CSP/webPreferences change" and "no invariant in play." The preload did change (the new channel) and §3 is the invariant in play. Please reword.

**3. File a follow-up issue for the deferred work.**

The durable folder-pinning fix (the root cause) and the notice-hydration limitation are deferred to prose, but #1317 is closed, so nothing open tracks them. Please open an issue for the durable fix and reference it here so it is not lost.

Non-blocking notes (your call):

- The notice is lost on any renderer reload, not just `--project-dir`, because it is not carried in `getCurrentState()` / hydration. Worth folding into the issue above; the "dev-only" wording undersells it.
- `watcher.ts` has new prettier warnings (around line 50 and line 103). CI does not gate lint, so please run prettier.
- `'.git'` in `NESTED_SCAN_SKIP_DIRS` is redundant with the `startsWith('.')` hidden-dir skip.
- On a folder with no top-level `research.json`, the message still says "also has research.json in a subfolder"; "also" reads wrong there (only reachable on `--project-dir` today).

francis-2008-happy:
Thanks — all three are in (06cef770), and the first one was a real hole rather than a scoping question.

**1. The empty-top-folder case.** You were right, and I took the second option rather than arguing the tester bundle. The bundle *did* have a top-level `research.json`, so the reported case was covered — but the case the title describes most literally wasn't, and "the reported bundle happened to dodge it" is a bad reason to ship a banner that can't fire for the shape it's named after.

Rather than inline it in the `ipcMain` handler, I pulled the gate into an exported `assertResearchProject()` in `watcher.ts` — same reason `classifyBasename` and `formatNestedNotice` are already out there: it's testable without `dialog`, `ipcMain`, or a BrowserWindow. Its no-top-level branch consults `findNestedResearchJson` and throws `formatNestedPicker(nested)`, which names the subfolder and shares the 3-path cap.

Mutation-checked the way the last four rounds established: deleting the `findNestedResearchJson` lookup from `assertResearchProject` fails *"points at the subfolder when only a nested research.json exists"* and nothing else. Both `formatNestedPicker` tests stay green under it — the same message-works-vs-message-is-*reached* split as `FolderNotice.test.tsx` vs `App.test.tsx`, which is why the gate needed its own test and not just a pure-function one.

**2. Security allowlist.** Row added to §3, and I've corrected the body rather than just appending to it — the old text claimed "no preload/CSP/webPreferences change" and "no invariant in play", and both halves were wrong: the preload is exactly where the channel is registered, and §3's own rule ("if a channel exists in the preload but is not in this table, it is unauthorized") is the invariant. Good catch that this would have failed the next security run.

**3. Follow-up issue: #1869.** Filed as one issue with the hydration gap folded in, as you asked. I'd considered splitting it — the durable pin is a lead decision, the hydration replay is takeable now — but kept it single per your note and said in the body that fixing part 2 doesn't depend on part 1.

On the deferral reasons, since the contributing guide wants them named: the durable pin is Cowork-side and rests on a lead decision (last discussed on #1054/#1317). The hydration half I counted rather than asserted — `ProjectStateSnapshot` in `transport.ts`, `ResearchDataProvider.tsx:105`, both transports, `watcher.ts`/`index.ts`, plus contract mocks and tests: ~6-8 files across 3 packages, changing a contract the web transport implements. That's the "too big for this PR, and I opened the call sites" clause, not a vibe. It carries `nothing-checks` — all suites are green with the notice un-replayed.

**Non-blocking notes, all taken:**
- **"dev-only" undersold it** — agreed, and reworded. Any renderer reload drops the notice; `--project-dir` is just where you'll meet it. Now says so.
- **Prettier** — `watcher.ts` clean, and I checked every changed `*.ts`/`*.tsx`.
- **`.git` redundant** — removed, with a comment noting the `startsWith('.')` test above already covers hidden dirs.
- **"also" reads wrong** — dropped from the banner. Worth saying why it stays wrong-ish otherwise: `startWatching` is also reachable via `--project-dir`, which doesn't require a top-level file, so "also" can't be guaranteed there even after fix 1. The picker message never says it.

**Verification:** `turbo test typecheck --force`, 8/8 green on the merged head — viewer-ui 221, electron **77** (was 72; +5 this round), web 10, typechecks clean. Merged `main`, 0 behind.

**One thing I did not do:** launch the app. Everything above is tests and mutations against the code, same as the previous rounds. The new `formatNestedPicker` string lands in a modal error dialog, and I haven't seen it rendered — that plus its wording is the bit I'd most like a genealogist eye on, and it's flagged under **Unsure about** in the body.

johnmarkpeterbrown (changes requested):
Reviewed as the second senior review. I tested the merge with `main` rather than your head — the branch is 27 commits behind now, and `main` has moved since your last run. The merge is clean.

On the merged tree, on **Windows**: viewer-ui 221, web 10, typecheck 5/5, and electron **76 of 77**. Both mutations from your Acceptance check reproduce exactly as you describe — deleting the two `<FolderNotice />` mounts fails both `App.test.tsx` cases while `FolderNotice.test.tsx` stays green at 3. That split is doing real work and it is the right shape.

Requesting changes on two things. The delivery path is right; both of these are in the round-6 half that nobody has reviewed yet.

---

### 1. The new test fails on Windows, and on Windows it cannot fail for the right reason

`relative()` returns the platform separator, so the message contains `sub\research.json` and the regex is looking for `sub/research.json`:

```
- Expected: /"sub\/research\.json"/
+ Received: "research.json is in a subfolder (\"sub\\research.json\"), not in the
             folder you picked. Reopen the viewer on that subfolder."
```

CI is Linux, so it is green there. It is red for everyone on Windows, which is most of the genealogist team — and it is the exact command your Acceptance check tells us to run.

The part that makes this blocking rather than cosmetic: I then applied your other stated mutation, deleting the `findNestedResearchJson` lookup out of `assertResearchProject`, and the result was **identical** — same one failure, same 27 passes. So on Windows this test cannot tell "the feature works" from "the feature is gone." It reads as coverage and is not.

You already solved this three describe blocks up, in `findNestedResearchJson` — *"relative() yields the platform separator, so build the expected the same way."* Same fix here:

```
apps/electron/src/main/__tests__/watcher.test.ts:248
You wrote:      await expect(assertResearchProject(dir)).rejects.toThrow(/"sub\/research\.json"/)
Change it to:   await expect(assertResearchProject(dir)).rejects.toThrow(`"${join('sub', 'research.json')}"`)
```

`join` is already imported, and `toThrow` with a string is substring matching. I ran it: 28 of 28, full electron suite 77 of 77, typecheck clean.

### 2. The folder picker now blocks on an unbounded scan

`assertResearchProject` awaits the depth-6 walk inside the `project:select-folder` handler. That check used to be a single `fs.stat`. There is no early exit either — it collects every hit, though only existence and the first three paths are ever used.

Measured on this machine:

```
repo root          121ms    585 dirs
C:/Users/FUJITSU  4378ms   8167 dirs
```

So picking a home or Documents folder by mistake freezes the dialog for about four and a half seconds with no feedback, then shows an error. That is the wrong-folder path, which is the path this whole feature is for — the user who lands there is already confused, and a silent freeze reads as a crash.

Bounding just the picker's call fixes it. Depth 2 still catches the reported shape, and the same folder drops to 9ms over 37 dirs:

```
apps/electron/src/main/watcher.ts
You wrote:      export async function findNestedResearchJson(folderPath: string): Promise<string[]> {
Change it to:   export async function findNestedResearchJson(
                  folderPath: string,
                  maxDepth: number = NESTED_SCAN_MAX_DEPTH
                ): Promise<string[]> {
```

```
You wrote:          if (depth > NESTED_SCAN_MAX_DEPTH) return
Change it to:       if (depth > maxDepth) return
```

```
You wrote:        const nested = await findNestedResearchJson(folderPath)
Change it to:     // Bounded harder than the background scan: this one blocks the folder picker.
                  const nested = await findNestedResearchJson(folderPath, 2)
```

francis-2008-happy:
Both of yours are in (8615c770) — and chasing the first one turned up something bigger that neither of you flagged, so I want to lead with that rather than bury it.

## The round-6 fix reached the user never

While checking how the picker's error surfaces, I traced it: `assertResearchProject` throws → `selectFolder`'s catch does `setError(err.message)` → and `error` is rendered by **nothing**. 27 components destructure `useResearchData()`; none takes `error`. It's the state @chesworthrm found dead in round 1.

So the "research.json is in a subfolder" pointer I added last round was computed correctly, thrown correctly, and dropped on the floor. That is the *same* bug rounds 1–2 existed to fix, reintroduced by me in the fix for it — and the reason it survived review is that it's one hop past where both of you looked: @clack391 asked for the pointer, I built the pointer, you checked the pointer's test quality and its cost. Nobody re-asked "does it render?"

Worth noting the blast radius is wider than my change: the generic **"Not a research project"** rejection has been invisible the same way since the viewer was extracted. Pick a non-project folder today and the dialog just closes.

**Fix:** an `ErrorNotice` bar mirroring `FolderNotice`, reading the `error`/`clearError` that were already on the context — only the render was missing. Mounted in all three `AppContent` branches. The **no-folder** branch is load-bearing: a rejected pick never sets `folderPath`, so that's the branch the user is actually looking at, and it's the one a `FolderNotice`-style surface would have missed.

Mutation-checked the way you both established, including deleting each mount one at a time:

| mutation | result |
|---|---|
| all 3 `<ErrorNotice />` mounts deleted | 3 error cases fail, folder-notice cases stay green |
| mount 1 only (no folder) | only *"NO folder is open"* fails |
| mount 2 only (folder, no research) | only *"research is NOT loaded"* fails |
| mount 3 only (research loaded) | only *"research IS loaded"* fails |

## 1. Windows separator — fixed, and you're right about why it's blocking

Applied verbatim: built from `join('sub', 'research.json')` instead of a regex with a literal `/`. You're right that I had the answer three describe blocks up in the same file and didn't follow it.

The part I want to acknowledge properly is your second paragraph, not the failure itself: that under the other mutation the result was *identical*, so on Windows the test couldn't separate "works" from "gone". A test that fails for the wrong reason is worse than no test — it reads as coverage. That's the repo's own "a new lint must be proven to fail" rule, and mine passed it only on Linux.

**What I could not verify:** I don't have a Windows box. I've confirmed it's green on Linux and that the assertion is now built the same way as the `findNestedResearchJson` precedent that already passes on your machine — but the platform claim is yours, not something I re-derived. Worth a re-run on Windows when you next look.

## 2. Picker scan — bounded, with a test that the bound is real

Applied as specced: `findNestedResearchJson(folderPath, maxDepth = NESTED_SCAN_MAX_DEPTH)`, `depth > maxDepth`, and the picker's call passes `PICKER_SCAN_MAX_DEPTH = 2` (named rather than inline, since the *reason* it differs from the background scan is the thing worth keeping).

I added a guard for it, because an unenforced bound is how this comes back: a `research.json` at depth 4 is found by the background scan at full depth, returns `[]` at the picker's bound, and falls back to the generic error. It fails if the bound is raised **or** if the new `maxDepth` parameter is ignored — I ran both.

The tradeoff that test now documents: a project nested deeper than 2 gets the generic message, not the pointer. Right call for a scan that blocks the dialog, but it is a real limit, so it's written down rather than implied.

## Verification

`turbo test typecheck --force` — 8/8 on the merged head: electron **78** (was 77), viewer-ui **225** (was 221), web 10, typechecks clean. Prettier clean. Merged `main`, 0 behind.

**Not launched.** Same as every prior round: this is tests and mutations, not a screen. Given what this round was about, that limit is worth stating plainly — I've proven `error` becomes a DOM node with `role="alert"` in all three branches, not that the bar looks right in a real window. The amber/red are inline-styled and unthemed, still cosmetic by earlier agreement.

## On the round count

Seven rounds, and the shape has been the same every time: each round fixed the mechanism and left the delivery one hop further along unguarded. This round I did it myself, in the fix for that exact bug. The chain is now pinned end to end — main → IPC (seam test) → provider state (regression test) → mount (per-branch tests), for both bars — so the specific failure mode is closed rather than moved. If there's still an unguarded hop I'd rather find it now than in round 8.

clack391 (changes requested):
Re-reviewed on 8615c770. My three earlier asks and John's two are all resolved, and I confirmed it: viewer-ui 225, electron watcher 29, the full suites and typecheck green, and every new guard fails when its target is reverted. The empty-top pointer, the depth-2 picker bound, the Windows separator, the security row, and issue #1869 are all in. The ErrorNotice you added in round 7 is the right call, and it is tested per branch. Nice work.

Two small things before this lands, both one-liners.

**1. The shared test stub is missing the new channel.**

`IpcResearchTransport.subscribe` now calls `window.api.onFolderNotice`, but the shared jsdom stub does not define it. No test breaks today, but the next electron test that mounts App through the real transport will hit "onFolderNotice is not a function." Keep the stub in sync with the preload.

In `apps/electron/test/setup.ts`:
```
You wrote:      onWatchError: () => {},
                onSidecarUpdated: () => {},

Change it to:   onWatchError: () => {},
                onFolderNotice: () => {},
                onSidecarUpdated: () => {},
```

**2. The security checklist no longer scans where the banners live.**

Section 7 of the security-invariants skill checks `.tsx` files in `src/renderer/`, but the real UI, including your new FolderNotice and ErrorNotice, is in `packages/viewer-ui/src/`. Run as written, section 7 passes without ever looking at the new string-rendering. Your rendering is safe (both use escaped React text, no `dangerouslySetInnerHTML`), so there is no bug. This just points the check at the surface it exists to guard.

In `apps/electron/.claude/skills/security-invariants/SKILL.md`:
```
You wrote:      Check all `.tsx` files in `src/renderer/`:

Change it to:   Check all `.tsx` files in `src/renderer/`, and in the shared
                `packages/viewer-ui/src/` where the extracted UI now lives
                (including the FolderNotice / ErrorNotice banners):
```

Optional, not blocking:

- App.tsx and its two test files were reformatted to double-quotes and semicolons, which differs from the rest of viewer-ui (single-quote, no semicolon). viewer-ui has no prettier config, so nothing enforces either way. If it is easy, matching the house style would shrink the diff to the mounts that matter. This one is partly on me, since I asked you to run prettier last round, so leave it if it is a hassle.
- The subfolder pointer from the picker reaches the banner wrapped in Electron's "Error invoking remote method 'project:select-folder': Error: ..." prefix, so the useful text sits behind boilerplate. It is a real improvement over the old invisible error, so a follow-up is fine: strip the prefix before `setError` when you get a chance.

Everything else I checked holds: symlink loops, XSS, the web no-op, the mid-scan race, and the depth bound. Once the two one-liners are in, I am good.

francis-2008-happy:
All four in (42193d1c) — both one-liners and both optional items. Thanks for the clean list.

## 1. Test stub — fixed, and typed so the next one can't drift

Added `onFolderNotice`. But your framing is the useful part: *"no test breaks today"* means the one-liner has no guard, so the same desync recurs the next time a channel is added. So I also typed the stub — `const apiStub: AppAPI = {...}` instead of an untyped literal inside `Object.defineProperty`.

Now the drift is a typecheck failure, not a runtime surprise. Deleting `onFolderNotice` from the stub:

```
test/setup.ts(20,7): error TS2741: Property 'onFolderNotice' is missing in type
'{...}' but required in type 'AppAPI'.
```

That's the exact failure you predicted, caught before anyone writes the test that would have tripped over it.

Typing it turned up two dead entries: **`getProjectState`** and **`selectProjectFolder`** are on no interface and referenced nowhere in `apps/electron` — removed. It was also missing `listProjectFiles` and `readImage`, now present.

## 2. Security §7 — retargeted

Applied as written. Worth naming what that check has been doing: §7 has scanned only `src/renderer/` since the viewer was extracted, so it has been passing without reading *any* of the real UI — not just my two banners. Your read is right that there's no bug here (both bars are escaped React text, no `dangerouslySetInnerHTML`), but a check that passes because it looks nowhere is the shape this repo explicitly treats as worse than no check.

## Both optional items — done, and one of them shouldn't have been optional

**The IPC prefix.** Stripped in `IpcResearchTransport`, which is the only layer that knows the transport is Electron IPC. I didn't want to defer this one: round 7 existed to make that sentence *visible*, and shipping it behind `Error invoking remote method 'project:select-folder': Error:` makes it visible but not readable — the user reads an IPC channel name they have no concept of. Two tests, and reverting the strip fails the first:

- strips the wrapper off a rejected `selectFolder`
- leaves a message carrying no wrapper untouched

**The formatting churn — that one was mine, not partly yours.** You asked me to run prettier, which was right. What I got wrong is *where*: `apps/electron/.prettierrc.yaml` exists, `packages/viewer-ui` has none, so the run there used prettier's defaults (double quotes, semicolons) against a single-quote, no-semicolon house style and reformatted whole files. Reverted the three viewer-ui files to their pre-round-7 state and re-applied only the semantic edits by hand.

`App.tsx`: **137/74 → 51/1.** The diff is now the component and its three mounts, nothing else.

## Verification

`turbo test typecheck --force` — 8/8 on the merged head: electron **97**, viewer-ui **225**, web 10, typechecks clean. Prettier clean on every changed file (electron files against the repo's config; viewer-ui by hand, matching house style). Merged `main`, 0 behind.

Guards proven to fail, one mutation at a time: the stub type (TS2741 above), the prefix strip (test fails on revert). The round-7 mount and depth-bound mutations still hold.

**Not launched**, same as every prior round — the bars are proven to become DOM nodes, not to look right in a real window.

One note for whoever merges: this still needs a **senior-genealogist** approval. @chesworthrm approved on the developer side, and both you and @johnmarkpeterbrown reviewed as senior devs, so CODEOWNERS' second team is still owed.

clack391 (approved):
Approving. This is a genuinely clean round-8, thanks for the thorough turnaround.

Both required items are in, and you took both optional notes too. A few things stood out:

- On the test stub you went past what I asked: typing it as `apiStub: AppAPI` means a future preload-channel drift now fails typecheck, so it cannot silently fall out of sync again. That is the better fix.
- `unwrapIpcError` is small, anchored, and tested for both the wrapped and the plain case, so the picker's subfolder pointer reaches the user clean.
- Reverting the App.tsx formatting to house style dropped the diff to the component and its mounts, which makes the real change easy to read.

I re-ran the changed surface (IpcResearchTransport, watcher, viewer-ui) and it is green, and the #1751 merge left watcher.ts untouched from the last-reviewed head. Nice work.

johnmarkpeterbrown (approved):
Looks good!

Issue #1494, opened 2026-08-09
eval: research trigger corpus (phase 1), then a stubbed routing suite (phase 2)


DallanQ:
> **Reviewed 2026-08-18 before junior handoff.** Decision settled — split into two
> PRs, the corpus second; rider 1c already shipped and is retired below. Original
> body follows.

**Touches:** eval/tests/unit/research/, eval/triggering/build_eval_set.py, eval/harness/scripts/check_runlogs.py, eval/harness/tests/unit/test_check_runlogs.py, eval/harness/harness/skill_stubs.py

> **Path correction:** `build_eval_set.py` lives at `eval/triggering/build_eval_set.py`, **not** under `eval/harness/scripts/`. `Makefile:402` invokes it as `cd eval/triggering && uv run python build_eval_set.py --skill $(SKILL)`.

## What's wrong

`research` is the component that fails most and it has **no unit suite at all**. `ls eval/tests/unit/` lists 25 skill directories; `eval/tests/unit/research/` does not exist. The only thing that exercises the router is a live e2e run.

A second, cheaper defect falls out of that. `make optimize-skill SKILL=research` looks like it works and does nothing:

- `eval/triggering/build_eval_set.py` derives its query set from `eval/tests/unit/*/*.json`. With no `research/` directory and no other skill's negative naming `research` in `negative.correct_skill`, it produces **0 items**. Measured on `main`:
  ```
  research: 0 items (0 positive, 0 negative)
  tree-edit: 14 items (11 positive, 3 negative)
  ```
- `build_eval_set.py:96-99` notices the thin set, prints a `NOTE:` to **stderr**, and `:100` **`return 0`**. Make sees success.
- `Makefile:401-407` then runs `run_loop` unconditionally against the empty set.
- What actually happens next, verified by reading the code: `run_eval.py:251` iterates `eval_set` to submit work, so an empty set submits **zero `claude -p` calls**; `run_loop.py:177-178` then sees `train_summary["failed"] == 0` and breaks on iteration 1 with `exit_reason = "all_passed"`. **So today it is a vacuous green, not a money burn** — the command reports the description passed everything while measuring nothing. For a *thin-but-nonempty* corpus the same missing hard-fail does burn real money on scores nobody should trust. Both are the same root cause: a warning where a stop belongs.

## Why it matters

[...]

PR #1723, opened 2026-08-18
proof-conclusion: own question resolution + eval/apps-server utf-8 fixes (#1399)

taiwo-stack:
## Summary

- **proof-conclusion** now owns the `questions` → `resolved` transition (§7 rewritten, all four prior prohibition sites updated), and **question-selection** no longer claims it — per the lead's final ruling on #1399. Includes three small in-scope folds: a stale-actor wording fix in `research/SKILL.md`, dropping a fabricated searched-and-not-found list, and a `max_turns` correction.
- Found and fixed a real bug while getting fresh eval evidence: `subprocess.run(text=True)` with no `encoding=` decodes with the Windows platform default (cp1252) and crashes on non-ASCII subprocess output — six live sites in `mock_mcp.py`, silently swallowed as a retried "transient stall." Widened `test_encoding_lint.py` to catch this shape going forward, which required fixing 10 more real sites across `eval/harness` and `apps/server`.
- Raised `ut_proof_conclusion_018`'s turn budget to 40 (was unset, defaulting to 20).
- Fixed a judge-anchoring bug on `rubric.md`'s "Narrative standalone" (review mode) dimension: reproduced 3-for-3 across every committed run log that the judge scores this on whether the reviewed `ps_NNN` narrative is good, not on whether the review of it is correct.
- **Ran `/code-review` on the full branch and acted on its findings** (below): fixed 3 real bugs in my own encoding-fix work, consolidated `mock_mcp.py`'s six near-duplicate Node-subprocess call sites into one helper, fixed a `lru_cache` that permanently cached a transient failure, corrected a stale routing hint in `question-state.ts` that could misdirect the model right back into the bug #1399 fixes, and corrected a spec doc that actively contradicted what shipped.

Closes #1399.

## Review guidance

**Start here:** `proof-conclusion/SKILL.md` §7 for the ownership change itself. For the bug fixes: `eval/harness/harness/mock_mcp.py`'s new `_run_node_eval()` helper (was six hand-duplicated call sites, three independent review angles flagged it) and its `_load_build_tool_catalog`/`_load_build_tool_catalog_uncached` split (fixes the cache-poisons-on-failure bug). For the doctrine fix: `packages/engine/mcp-server/src/utils/question-state.ts`'s corrected routing hint, and `docs/specs/skill-rewrites-for-persistence-tools-spec.md` §4.12.

**Unsure about — this is the important one.** `/code-review` converged from multiple independent angles on an architectural gap in the ownership-transition design itself (not introduced by my commits — inherent to the original #1399 implementation, now visible with more scrutiny):

1. **No tool-level check that a tree write actually landed before `resolved` is set.** `questionResolvedInvariants` (`research-append.ts`) only checks that *some* `proof_summaries` entry references the question — never that a tier-≥-probable conclusion actually landed in `tree.gedcomx.json`. Moving the resolve-write into proof-conclusion's own invocation removed the turn boundary that used to let the orchestrator apply the tree-encoding gate before a *separate* question-selection invocation resolved. If §6's `tree_edit`/`tree_correct` call fails or gets skipped under context pressure, nothing stops §7 from resolving anyway — a found-but-lost conclusion, now also unrecoverable through normal routing (question-selection's own doctrine treats a resolved question as done).
2. **Nothing mechanically stops the wrong skill from writing `resolved`.** `ownership.json`'s `questions` row lists `question-selection`, `research-exhaustiveness`, and `proof-conclusion` as co-equal section callers — the field-level split this PR encodes (who writes `status`/`resolved`/`resolution_assertion_ids` specifically) exists only as SKILL.md prose in three files. The unit-plane ownership check (`test_universal.py`) only verifies *which skill touched the `questions` section*, not *which field* — it can't distinguish a legitimate `research-exhaustiveness` write to `exhaustive_declared` from an unauthorized `resolved` write by the same skill. This is the same field that the manifest already documents got written from 11 different skill contexts once before.
3. **No guaranteed mentor-critique gate on standalone invocation.** The three carve-outs in §7 (review mode, blocked precondition, decline-on-conflict) all implicitly assume `/research` orchestration is running the mentor checkpoint. proof-conclusion's own body never invokes `gps-mentor`; a direct, standalone "write the conclusion for q_003" request writes the proof AND resolves the question in the same turn, with no `evaluations[]` proof-critique entry ever created unless the user separately asks for review.

None of these are things I'm fixing here — each needs either a new tool-level invariant (developer + genealogist joint design) or a scope decision only the lead can make. Full traces of each are in the review transcript; happy to expand any of the three into its own issue if you want them tracked separately rather than left as this comment.

**Also flagged, not acted on:**
- `apps/server/app/sandbox_server.py`'s reader thread swallows any exception via a bare `try/finally` (same silent-failure class as this PR's fix, but a live production WebSocket path outside what I felt safe changing blind) — commented on issue #1713, same file/theme.
- A CI-trigger gap: `.github/workflows/eval-harness-tests.yml`'s path regex doesn't cover `scripts/claude-hooks/*.py`, so the repo-wide encoding lint wouldn't even run if a future PR only touched files there.
- `proof-conclusion/SKILL.md`'s `research_append` examples never show the batched `ops:[...]` form for combining the proof-summary append with the questions-resolve update, unlike §6's explicit modeling for `tree_edit` — a genealogist-doctrine efficiency suggestion, not a bug.
- Two prior aborts (`ut_proof_conclusion_004`: `unmatched_tool_call`; `ut_proof_conclusion_019`: `max_wall_clock_seconds`) in the mandatory GPS-mentor-critique-reaction tail, reproduced as pre-existing model-sampling variance (`_004` has shown three different outcomes — abort/pass/abort — across three runs of the identical test).

## Plan

**Didn't change:** The escalation/severity doctrine in any reference doc, the write-boundary enforcement layer (already landed separately, #1685), or any schema/ownership-table file.

**Acceptance check:** `ut_proof_conclusion_011` now passes (the test `d0dd324b`'s fix was written to address, never re-verified until this PR's fresh run). `ut_proof_conclusion_018` verified via a cheap `--test` scratch run before the full paid re-run. The `mock_mcp.py` consolidation verified via `test_mock_mcp.py`'s full 25-test suite plus the harness's full 2222-test suite, unchanged pass count before/after. The `_is_text_mode` truthiness fix and the stale routing-hint fix were each verified against real CPython introspection (`inspect.getsource`/`inspect.signature`) and the existing test suite, respectively — not just read.

**Deviated from the plan:** Grew from "verify #1399's eval evidence" into six commits total. Each addition was a real, verified defect directly encountered doing this work (the encoding crash, `_018`'s turn budget, the judge-anchoring bug, then `/code-review`'s three code bugs plus the two doctrine/spec inconsistencies) — none was scope creep in the sense of doing unrelated work, though the `/code-review` pass in particular surfaced how much surface area a "verify and finish a teammate's branch" task can turn up.

## Test plan

- [x] `eval/harness` full pytest suite: 2222 passed, 3 skipped, unchanged across every commit in this PR.
- [x] `question-state.test.ts`: 17/17 pass after the routing-hint fix. `tsc --noEmit`: clean.
- [x] `test_mock_mcp.py`: 25/25 pass after the six-site consolidation.
- [x] **`proof-conclusion` re-run, taken from the merged tree — `v1_2026-08-18_14-42-08` (21 tests, $14.36): 14 pass, 4 partial, 2 xpass, 1 aborted.** This is the run rule 2 gates on, and its snapshot is active. `ut_proof_conclusion_011` (decline on an unresolved conflict) and `ut_proof_conclusion_r7k` (review mode) both pass; `ut_proof_conclusion_001` is gradeable for the first time at the raised budgets. The abort is `_015` (`unmatched_tool_call` — it calls `project_context` with args no fixture predicate matches, 2 of 24 calls); it is **not** in the `review_sample`, so nothing is required of it. Earlier runs in this PR (`v1_2026-08-18_08-00-47`, `v1_2026-08-18_12-34-17`) are superseded — they predate the merge of `main`.
- [x] **`question-selection`: no re-run needed.** `main` does not touch it, so `v1_2026-08-18_06-32-50` stays active in the merge tree — 14/14 pass, annotated.
- [x] **Correction to an earlier claim in this body:** merging `main` brings #1545's `merge_warnings` severity rewording (`error`/`warning` -> `contradiction`/`implausible`) into `proof-conclusion/SKILL.md`, invalidating the snapshot a *second* time — which is why CI named two drifting files where a branch-only check named one. The re-run therefore had to be taken from a tree with `main` already merged; a run from the pre-merge tip would have been stale on arrival. Still one run, not two.
- [x] **Annotation complete.** `v1_2026-08-18_14-42-08.ann.json`: 43 corrections covering every dimension of all five `review_sample` tests, with written comments on the two that are not confirmed passes (`_005` and `_006`, Evidence completeness held at 2). `question-selection`'s run log was already annotated.
- [x] **`eval/harness/scripts/check_runlogs.py` run locally against the merged tree: "All runlog rules satisfied", exit 0.** One warn-only line remains (`_015` produced no graded dimensions).
- [x] `question-state.test.ts` assertion anchored to the owning skill and **proven to fail**: reinjecting the stale `question-selection` hint into `question-state.ts` takes it to 1 failed / 16 passed; restored, 17/17. The previous assertion matched only the tail, so it passed identically before and after the fix it covers.
- [x] Post-merge suites on the pushed tree: `eval/harness` pytest 2250 passed / 3 skipped; engine vitest 2376 passed (100 files).

## #1494 phase 1b is NOT in this PR (briefly was, now reverted)

`0819e9c77` reverts `5e2a38e7f` + `22f0b627e`, which had cherry-picked the
`build_eval_set.py` hard-fail and its `architecture.md` follow-up in here.
**That change lives in #1737**, which was reopened after it turned out to have
already been reviewed (CI green, approved, `ready-for-senior-developer`) --
closing it had discarded a paid review. Reverted rather than force-pushed,
because this PR is mid-review and its history is what a reviewer has read.

Nothing about this PR's own scope changed: `build_eval_set.py` and
`docs/architecture.md` are byte-identical to `origin/main` here, and the revert
touches neither the run-log snapshot nor any skill, so
`v1_2026-08-18_14-42-08` stays active -- `check_runlogs.py` still exits 0.

## Follow-on issues

**Folded in / not filed:** The three architectural findings under "Unsure about" are left as PR-body findings for the lead's triage rather than filed as issues — each needs a scope decision before an issue would even have the right acceptance criteria. Commented on existing issue #1713 with the related `sandbox_server.py` finding rather than filing separately (same file/theme).


johnmarkpeterbrown (changes requested):
Here's a Slack-ready draft. Nothing posted — this is yours to send or edit.

---

Hey @taiwo-stack — reviewed #1723 end to end. Good work and it's close. One red check plus two fixes. *Order matters* because a paid eval run is in play, so I've written it as a sequence rather than a list.

*1 — Merge `origin/main` first* (you're 42 behind)

You'll hit 3 conflicts, all `.ann.json` files under `eval/runlogs/unit/proof-conclusion/`. Keep main's side. Main renamed those files, your branch deletes them, so taking your side loses annotation history.

*2 — Then take the re-run, from that merged tree*

This is the bit worth reading twice. `runlogs` is red on *two* files, not one.

`rubric.md` drifts on your branch — you know about that one. But `proof-conclusion/SKILL.md` drifts too, and *only once main is merged*: main's #1545 rewrote the `merge_warnings` severity line inside that file. CI checks out the merge commit, which is why it names both and a branch-only look names one. So a re-run from your current tip goes stale the moment you update the branch.

Your body currently says the opposite:

> the rubric fix and this round's fixes don't touch anything in the skill-side snapshot beyond `rubric.md` … so no additional re-run cost

Swap that for: _"Merging main brings #1545's `merge_warnings` rewording into `proof-conclusion/SKILL.md`, invalidating the snapshot a second time. The re-run must be taken from a tree with main already merged."_

Still one run, not two — just taken after the merge. And `eval-cosmetic-skip` isn't the way out here; the rubric edit changes a judge instruction, so it isn't behavior-neutral.

*3 — Then the two fixes below, any order*

I checked `build_snapshot`: the run log tracks only `plugin/skills/proof-conclusion/**`, `eval/tests/unit/proof-conclusion/**`, and referenced scenarios/fixtures. It does *not* track `packages/engine/mcp-server/**` or `docs/`. So neither fix can invalidate the run. Just don't touch SKILL.md or rubric.md again after step 2.

*Fix A — the Test plan describes a superseded run*

The numbers quoted are `v1_2026-08-18_08-00-47`. The latest committed log, and the only one rule 2 gates on, is `v1_2026-08-18_12-34-17` ($14.16): 14 pass / 4 partial / 1 abort / 2 xpass.

In that run `_004` and `_019` both pass. `_016` aborted instead, same `unmatched_tool_call`. `_005` `_006` `_012` `_018` came back partial. `_018` — the one you raised the budget for, and the one your Acceptance check names — is partial on *content*, not budget: Correctness 2, Narrative standalone 2, Evidence completeness 2. `_011` passes in both, so that half of the acceptance check holds.

Not asking you to re-run to settle it, the re-run is already required. Just point the body at the run CI actually reads, and call the three outcome changes across two runs of an unchanged tree what they are — suite variance, not a signal about this diff.

*Fix B — the routing-hint fix isn't tested*

`packages/engine/mcp-server/tests/utils/question-state.test.ts:78`

You wrote: `expect(s.nextStep).toMatch(/mark the question resolved/);`
Change to: `expect(s.nextStep).toMatch(/^proof-conclusion — .*mark the question resolved/);`

The assertion only matches the tail, so it passes identically before and after your fix — it can't catch the stale hint it now covers. And the comment two lines above still states the premise you removed:

You wrote: `// is still outstanding, and it is the transition no skill body claims.`
Change to: `// is still outstanding, and proof-conclusion owns it (#1399).`

*That's the merge-blocking set.* Two of those are body edits and one is two lines of test.

*On your three "Unsure about" findings* — don't file them, ask Dallan right in the PR thread. He commented on #1399 three days ago so he's reachable, and a reachable decision is a question, not a ticket. Two survive my check and are worth his answer: field-level ownership of `questions` exists only as prose (the unit check diffs whole sections, so it can't tell a legit `exhaustive_declared` write from an unauthorized `resolved` one — that's `nothing-checks` shaped, his call whether it gets the label), and standalone `proof-conclusion` resolving with no `proof-critique` on record. I'd drop #1 — `research/SKILL.md:151` already re-invokes `proof-conclusion` for an unencoded conclusion before `project.status = "completed"`, so it isn't unrecoverable.

*Already verified on my side, don't redo it:* harness pytest 2250 passed / 3 skipped, engine vitest 2382 passed, `tsc --noEmit` clean, `question-state` 17/17, `test_mock_mcp` 26/26 — all on the tree with main test-merged. I also watched your widened lint fail on purpose (stripped `encoding=` off `provenance.py:112`, got the right offender, restored). `apps/server` pytest I couldn't run locally, broken `uv` Python link on my box — CI's green and it's a two-line diff.

Requesting changes so the second senior doesn't burn a pass while this is in flight. Ping me once the re-run's in and I'll turn it around fast.

clack391 (approved):
Approving, senior developer. @johnmarkpeterbrown's blocking set is verified-addressed at this head, so you're clear to flip your review.

I re-verified on a test-merge with current main (65 behind, merges clean):
- check_runlogs: "All runlog rules satisfied" on the merged tree, so the snapshot-staleness blocker stays resolved even against current main. v1_2026-08-18_14-42-08 is the active run.
- Fix B: the routing-hint assertion is start-anchored and I proved it reds when the stale question-selection hint is reinjected (1 failed / 16 passed); restored.
- The widened encoding lint reds when encoding= is stripped off a real subprocess text=True site (provenance.py:117, mock_mcp.py:186); restored.
- harness pytest 2312 passed (the lone test_mock_mcp failure is the missing-engine-build case, reproduces on clean main, CI builds it and is green); question-state 17/17.

The §7 ownership change matches DallanQ's #1399 ruling read at the source, and the "resolve only a question you just concluded" guard handles the decline / review-mode / stale-ps_ paths. A blind second pass swept the whole plugin and docs and found no surviving contradiction, and traced that a standalone resolve still routes to the mentor gate.

Non-blocking, your call at release time: the active run carries one aborted test (ut_proof_conclusion_015, unmatched_tool_call) that passed in all four prior runs on the same §7 body, so it reads as single-run jitter; check_runlogs exempts it. The all-green 12-34-17 run is an alternative if you'd rather release a clean one.

The three "Unsure about" findings are follow-ups for Dallan's triage, not blockers.

Issue #1489, opened 2026-08-09
A control-plane deploy can ship green while prod's E2B agent image is stale


DallanQ:
> **Reviewed 2026-08-12 before junior handoff.** The premise holds and is still true,
> but three senior triggers fire independently of any decision — this is labelled
> `senior` and is not for the junior pool. The deploy-gate fork is now settled below. This block supersedes the §9.4 blockquote and
> the "Out of scope" cross-reference below; the original follows.

## Already documented — this issue changes no docs

`docs/architecture.md` §9.4 now carries this gap as its second numbered item ("A deploy
does not ship the sandbox"), added by PR #1507 on 2026-08-09, hours after this issue was
written. The note that a companion PR was "already adding that row" is spent, and §9.4
no longer has a table at all — the enumeration moved to the `nothing-checks` label. **Do
not touch `docs/architecture.md`.**

Also stale: the "Out of scope" pointer names issue #1118, which closed `not planned` on
2026-08-09. The live sandbox-lifecycle issue is **#1120**.

## Sites this touches — checklist

- [ ] `apps/server/sandbox/build-image.sh` — write the provenance file into the build context.
- [ ] `apps/server/sandbox/e2b.Dockerfile` — `COPY` it in **as the last COPY, below the `RUN … npm ci` layer**. Its contents change on every commit, so a COPY placed above `npm ci` invalidates every layer under it and makes each image build re-run apt, pip and `npm ci`.
- [ ] `apps/server/sandbox/README.md` — the "What's baked (image layout)" table and the `e2b template create` recipe both enumerate what lands under `$AGENT_HOME`; a baked file missing from that table is invisible to the next reader. While you are in it, two lines are already stale: it says Node 20 where the Dockerfile says `NODE_MAJOR=22`, and it says an `E2B_API_KEY` "does not exist yet". Fix both in this PR.
- [ ] `apps/server/app/sandbox/e2b.py` — where the baked commit is read back.
- [ ] `apps/server/app/main.py` — the `/api/health` field. Name it so it cannot be read as the Fly build: `Settings.git_sha` / `build_date` already exist and describe the *other* image.
- [ ] `apps/server/tests/test_e2b_provider.py` — its fake `AsyncSandbox` already fakes `files.read`, so the read-back is testable offline under `make server-test`.
[...]

PR #1724, opened 2026-08-18
Classify operator agent failures 1126

### LINKED ISSUE #1126: Classify operator-vs-user agent failures instead of surfacing the raw SDK error
> **Reviewed 2026-08-14 before junior handoff — `needs-a-decision`.** Two things
> changed. The body named **one** leak site; there are **five**, and a junior who
> fixes the first leaves the same 401 reaching the user through a sibling path.
> And the "server-side log loud enough to page" half **cannot be built today** —
> the question is with the lead. Everything else here is junior work.

**Touches:** apps/server/app/agent/real_agent.py, apps/server/app/agent/runner.py,
apps/server/app/sandbox_server.py, apps/server/tests/,
docs/specs/hosted-web-workbench-spec.md

Classify in `handle_turn`: an SDK 401/403 is a misconfiguration, not something
the user can fix, and the message must say so without naming a credential.

**The `:520-521` citation is stale, and there are five sites, not one** (verified
on `main` 2026-08-14 — cite the symbol, the line moves). Fix them together:

- [ ] `real_agent.py` — `RealAgent.handle_turn`, the `_ensure_client` guard:
      `yield _event("error", text=f"Failed to start the agent: {exc}")` (`:508`)
- [ ] `real_agent.py` — `RealAgent.handle_turn`, the query/receive guard:
      `yield _event("error", text=f"Agent error: {exc}")` (`:540`) — this is what
      the original `:520-521` pointed at
- [ ] `runner.py` — the outer never-die catch:
      `emit({"kind": "error", "text": f"Agent error: {exc}"})` (`:65`). Same text,
      different module; anything that escapes `handle_turn` lands here
- [ ] `sandbox_server.py` — spawn failure:
      `broadcast({..., "state": "chat_error", "message": str(exc)})` (`:162`),
      which `ChatPane.tsx` renders as `Chat unavailable: <exc>`
- [ ] `real_agent.py:503` and `sandbox_server.py:226` are already operator-framed
      but are developer text a paying user should never read — reword with the rest

**Drop the two-branch framing — an SDK auth failure here is always the
operator's.** The only credential the Agent SDK holds is the control plane's own
`ANTHROPIC_API_KEY`, rewritten per connect by `app/agent_secrets.py`; there is no
user-supplied key. A FamilySearch expiry never reaches these handlers at all:
`getValidToken()` throws inside the MCP server
(`packages/engine/mcp-server/src/auth/refresh.ts`) and comes back as a **tool
result** the model reads. So do not build an "is this the user's fault?" branch
here — it cannot fire, and widening the classifier to reach tool results is a
different, larger change into the MCP boundary.

**Keep it server-side and text-only.** `apps/web/src/components/chatEvents.ts`
renders `ev.text` verbatim for `kind: 'error'`, so no web change is needed.

**Verify with `make server-test`.** Put the classification in a pure function
(exception in, user-facing string out) and unit-test it beside
`apps/server/tests/test_real_agent_stream.py`. Do not try to test by raising
through the SDK — no offline gate exercises the live Agent SDK path (issue #1207).

**Then run `make sandbox-image`, or production keeps showing the old message.**
`apps/server/app/agent` is baked into the `genealogy-agent` E2B image, not the Fly
container; `make deploy`'s `deploy-preflight` only *warns*. This PR can merge,
deploy green, and change nothing a user sees.

**Record the reasoning in `docs/specs/hosted-web-workbench-spec.md` §7 ("The agent
runtime"), not here** — two sentences: that every SDK auth failure at this
boundary is operator-side by construction, and that the user-actionable
FamilySearch branch was considered and dropped because those errors arrive as
tool results. Writing that is part of the task.

**Not in scope.** Porting `classify_server_status` / `unavailable_message` from
`eval/harness/e2e/mcp_health.py` is a good idea and a different, larger task —
`real_agent.py` never reads the SDK's `system`/`init` message at all, so the
`mcp_servers` payload that classifier needs is dropped on the floor today. File
it separately rather than bundling it.

**Notification: a structured stderr line, and nothing pages** (lead's call,
2026-08-14). Emit one `_log("[operator] <classification>: <exc>")` line alongside
the user-facing message. It lands in `/tmp/agent.log`, which `session_logs` in
`apps/server/app/sessions.py` already tails via `GET /api/sessions/{id}/logs`.
That is a forensic trail, not an alert: reading it needs the session id, so it
only helps someone who already knows there was an outage.

**Do not write "the administrator has been notified" into the user-facing
string** — nothing notifies anyone, and a promise the product does not keep is
worse than the raw error. Use wording that is true: *"This service is
misconfigured. Nothing you did caused this — please report it."*

Chosen over building real error tracking in this issue (rejected: it would be the
repo's first durable server-side record and its first vendor/spend decision, and
would hold the misdirection fix behind a procurement call) and over shipping no
operator log at all (rejected: the 2026-07-20 outage left nothing to read
afterwards). The alerting half is filed separately — **do not start it in this
PR.**

---

*Original body follows.*

`real_agent.handle_turn` wraps every exception verbatim
(`apps/server/app/agent/real_agent.py:520-521`) and `ChatPane` renders it as-is,
so an operator-key 401 reached a tester as *"Failed to authenticate. API Error:
401 API key is invalid."* after ~90s of waiting.

Two alpha testers each read that as a FamilySearch problem and went to debug the
wrong credential. That misdirection, not the wording, is the cost.

Classify in `handle_turn`: SDK 401/403 → "This service is misconfigured; the
administrator has been notified" plus a server-side log loud enough to page,
while genuinely user-actionable failures (an expired FamilySearch token) keep
their current specific wording.

Surfaced by the 2026-07-20 outage; the credential-freeze half shipped in PR #762.

aghadiayeamayanvboernest:
Closes #1126.

## The defect

Two alpha testers read an operator-key 401 — *"Failed to authenticate. API Error: 401 API key is invalid."* — as a FamilySearch problem and went off debugging the wrong credential. Every error path at the agent boundary wrapped the raw exception and `ChatPane` rendered it verbatim. **The misdirection is the cost, not the wording.**

## What I found before building

The issue names five error sites. The 401 the testers saw travels **none** of them — it arrives as a normal `AssistantMessage` whose text blocks `map_message` tags `kind: "text"`, i.e. styled as the assistant's own answer. Fixing only the five would have looked like it worked and changed nothing a user sees.

@DallanQ confirmed the finding and added the fix: emit `kind: "error"` so the UI knows it isn't an answer — `chatEvents.ts` sets `last.error`, which `ChatPane` renders with `msgError`.

Two more sites the issue's list didn't name:

- **`ResultMessage.is_error` was read nowhere.** `receive_response()` *yields* rather than raises, so an in-turn API 401 produced **no error event at all** — the turn ended with `usage` + `turn_done` and the user watched ~90s of retries, then silence. That latency is this path's signature, not a fail-fast `connect()`.
- **`runner.py:91`** — `"Interrupt failed: Not connected"`, one line below a line I was already editing. Concrete path: key rotates mid-turn, control channel dies, user presses Stop.

## The design ruling (lead, 2026-08-14), both halves load-bearing

- **No two-branch classifier.** An SDK auth failure here is *always* operator-side: the only credential the SDK holds is our own `ANTHROPIC_API_KEY`, and a FamilySearch expiry never reaches these handlers — it's raised inside the MCP server and returns as a *tool result*. The "user's fault?" branch could never fire. Written into the spec so the next reader doesn't re-add it.
- **No false promise.** Never "the administrator has been notified" — nothing notifies anyone.

The default string is `UNEXPECTED`, **not** `MISCONFIGURED`: returning the misconfiguration line for every failure would report an unrelated `MessageParseError` as an operator problem — the same misdirection, pointed somewhere new.

## The `mcp_health` port

My plan deferred this; @DallanQ's comment on the issue recommends it, so it's **in**. My stated reason for deferring — "`real_agent` never reads the `system`/`init` payload" — was a fact about the current code, not a blocker: `receive_response()` yields `SystemMessage` (`client.py:586`).

A hosted session whose genealogy MCP server never connected **still runs** — the model just has no genealogy tools — so the user pays for a full session of research that could not have happened, with nothing distinguishing it from a genuine dead end. The e2e harness aborts such a run; production doesn't.

Two deliberate divergences from the harness copy, both documented in the module header:

- **Duplicated, not imported** — forced: the E2B image bakes `COPY apps/server/app` only, so `eval/harness/` is unreachable at runtime.
- **`unavailable_message` is not copied verbatim.** Its words are "RE-RUN the test" and "run `make e2e-preflight`" — nonsense to a web user with no test and no Makefile. The *classification* ports; the audience changes. Production **warns once** and lets the turn proceed, because there is no run to abort.

`pending` must never warn: at init a healthy server still reads `pending` (it settles ~14s later) while a dead one already reads `failed`, so a `!= "connected"` test would fire on every healthy session. That's what the three-way split is for.

## Verification

`make test-all` — all checks passed.

**All nine call-site tests were run against the unfixed call sites first and confirmed red**, including the one covering Dallan's addition. Four of those failed on a missing method, which is a weaker signal than "behaviour wrong" — so I also **mutated** `should_warn_at_init` to `health != "connected"` and confirmed three tests go red, including both no-warn guards. I've shipped assertions that couldn't fail before; this is the habit that catches it.

A pure-function test of `classify` alone is passed by a five-line constant while every call site still leaks, so the load-bearing tests drive `serve()`, `handle_turn`, and `map_message` and assert on the emitted event.

## ⚠️ The step nothing enforces

**`make sandbox-image` must run, or production keeps showing the old message.** `app/agent` and `sandbox_server.py` bake into the `genealogy-agent` E2B image, not the Fly container, and `deploy-preflight` only *warns*. This PR can merge, deploy green, and change nothing a user sees.

## Second commit: an unrelated red gate, fixed rather than filed

`feedback.test.ts > drops the largest files instead of throwing` was failing under `make test-all` **on a clean tree** — so the full gate was red for anyone running it. Not flaky, under-budgeted: it zips 45 MB of incompressible bytes, ~2.4s alone and ~8.4s when turbo runs every suite at once, against vitest's 5s default.

I measured before choosing the fix. The fixture can't shrink (3 × 15 MB is deliberately set against the 35 MB archive budget the test exercises), and making it cheaper wouldn't help — DEFLATE dominates; generating the noise is only ~340ms of the 2.4s. So the clock is the only thing that can move. I left the LCG generator alone: swapping to `crypto` is 310ms faster but makes a size-budget assertion depend on the compressed size of random bytes.

## Not in scope

- **The mid-run backstop** (`tool_search_miss_streak`, for a server that dies *after* init) — needs a per-session counter threaded through `handle_turn` and a decision about what to do to a conversation already in flight, with no production measurement behind it yet. The init check is the decisive one. Not filed: nobody asked for it and nothing is broken without it.
- **`apps/web`** — `chatEvents.ts` already handles `kind: 'error'`, so Dallan's suggestion works against what's already there.


clack391 (changes requested):
Strong PR. The classification is clean, the tests drive the real call sites rather than the pure function, and the spec section is exactly the durable record this needed. One change before merge, plus a small consistency fold.

**One turn can emit two error bubbles, and they can contradict each other.**

`map_message` emits an error event when an `AssistantMessage` is marked errored. `handle_turn` emits another when the turn's `ResultMessage` has `is_error`. Nothing coordinates them, and the SDK sets those two fields independently, from two separate CLI messages, so when both are present in one turn the user gets two error bubbles. If the `ResultMessage` has `is_error` set but no `api_error_status`, they read MISCONFIGURED ("please report it") and then UNEXPECTED ("please try again") for the same failure. That contradiction is the misdirection this PR is closing, pointed at itself.

Guard it so a turn emits at most one error event. Track that an error was emitted this turn, and skip the `ResultMessage.is_error` emission when the assistant-message path already fired (that path carries the real `error_kind`, so it is the stronger signal).

Then add a test that drives both messages through one `handle_turn`:

```
[AssistantMessage(error="authentication_failed"), ResultMessage(is_error=True, ...)]
```

and asserts exactly one error event. That pins it either way: if the CLI never sends both, the test documents it; if it can, the test catches the double bubble.

**Small fold while you are in `real_agent.py`:** `_mcp_health_events` reads `mcp_servers` from any `SystemMessage`, but the harness it mirrors gates on `subtype == "init"` first. Harmless today, but match the source it copies by gating on `subtype == "init"`.

Everything else looks good.

chesworthrm (commented):
Full pass, reviewed as a test-merge with `main` (the branch is 20 commits behind, so its own green head isn't the thing that matters).

**What it does:** adds `errors.py` (two user-facing strings + a classifier) and `mcp_health.py` (a three-way port of the harness's absent-surface detector), then routes seven call sites through them — `map_message`'s errored `AssistantMessage`, `handle_turn`'s SDK import / `_ensure_client` / receive loop / `ResultMessage.is_error`, `runner.py`'s `_run_turn` and interrupt, and `sandbox_server.py`'s spawn failure and runner exit. Plus the vitest clock bump on the 45 MB feedback-zip test.

Everything in the body that I could check, checked out. `AssistantMessage.error`, `ResultMessage.is_error` / `api_error_status` and `SystemMessage.data` all exist on the pinned SDK (0.2.128). `system` messages really do reach `receive_response()` (`query.py:367`), so the init read is on a live path. The `mcp_health` copy is faithful to `eval/harness/e2e/mcp_health.py`. And the `make sandbox-image` warning is accurate — `Makefile:726-734` is advisory only. That gap is already registered as #1489, so nobody needs to file it again.

**Checks:** all green on `1c1993df`. **Verified here:** `apps/server` pytest 200 passed / 1 skipped, and `turbo test typecheck --force` 8/8, both on the branch head and again on the test-merge. I also broke each new guard to watch it fail — seven of nine mutations went red. The two that didn't are the second finding below.

**Still needs:** two approvals, one from @PioneerAIAcademy/senior-developers (every changed path is `.py` / `.ts` / `.md`, so the extension rules match). CODEOWNERS resolves clean.

---

### 1. Pressing Stop now tells the user something went wrong

Reading `ResultMessage.is_error` catches the in-turn 401 you were after. It also catches a turn the **user** ended.

When someone presses Stop mid-tool-call, the CLI emits `is_error: true`, `subtype: "error_during_execution"`, `terminal_reason: "aborted_streaming"`, and **no** `api_error_status`. So `classify(status=None)` falls through to the default, and the person who just pressed Stop reads:

> The agent stopped unexpectedly. Nothing you did caused this — please try again, and report it if it keeps happening.

I built that exact message object and drove it through `handle_turn`; that string came out, once, on the error channel. Before this PR the field was read nowhere, so Stop was silent. This is a new false alarm, and it asks the user to report their own deliberate action.

Claude Code itself skips this case using the same test. In the bundled CLI: `Wpt(e){return e==="aborted_streaming"||e==="aborted_tools"}`, and its own consumer only renders the error when `subtype==="error_during_execution" && !Wpt(l.terminal_reason)`.

In `real_agent.py`:

```
You wrote:      if getattr(message, "is_error", False):
Change it to:   aborted = getattr(message, "terminal_reason", None) in (
                    "aborted_streaming", "aborted_tools")
                if getattr(message, "is_error", False) and not aborted:
```

Worth a test next to it: an aborted `ResultMessage` through `_turn_events`, asserting no error event.

### 2. The MCP-health warning can be unwired and every test still passes

Your body says the load-bearing tests "drive `serve()`, `handle_turn`, and `map_message` and assert on the emitted event." That's true of the error-classification half. It isn't true of the MCP-health half — every one of those tests calls `agent._mcp_health_events(message)` directly, so the two lines that carry the warning to a user are unverified. Two mutations, both leaving 200 passed / 1 skipped:

- `if isinstance(message, SystemMessage):` → `if False and isinstance(...)`. The feature is dead. No hosted user ever sees the warning. Nothing failed.
- `GENEALOGY_TOOL_PREFIX` → `"zzz_never_matches__"`. `_mcp_calls` never increments, so a re-spawned CLI mid-session tells a user who *has* done real research "please do not treat it as research" — the exact false warning `should_warn_at_init` exists to prevent. Nothing failed.

Two tests close both. I wrote them, confirmed they pass on your branch, and confirmed each goes red under its matching mutation:

```python
def test_an_unhealthy_init_warns_through_the_turn_loop(tmp_path):
    """The wiring, not just the helper: handle_turn must deliver the warning."""
    sdk = pytest.importorskip("claude_agent_sdk")
    from app.agent.real_agent import RealAgent

    init = sdk.SystemMessage(
        subtype="init",
        data={"mcp_servers": [{"name": "claude.ai Slack", "status": "needs-auth"}]},
    )
    done = sdk.ResultMessage(
        subtype="success", duration_ms=1, duration_api_ms=1,
        is_error=False, num_turns=1, session_id="s1",
    )
    events = _turn_events(RealAgent(tmp_path), [init, done])

    assert any(
        e.get("kind") == "error" and "do not treat it as research" in e.get("text", "")
        for e in events
    ), "an unhealthy init never reached the user"


def test_a_genealogy_call_this_turn_suppresses_a_later_init_warning(tmp_path):
    """The mcp_calls counter is fed from the emitted tool_use events, so the
    prefix and the increment both have to be right or a session that HAS
    researched gets told research was impossible."""
    sdk = pytest.importorskip("claude_agent_sdk")
    from app.agent.real_agent import RealAgent

    call = sdk.AssistantMessage(
        content=[sdk.ToolUseBlock(
            id="t1", name="mcp__genealogy__record_search", input={"surname": "Flynn"},
        )],
        model="claude-opus-4-8",
    )
    dead_init = sdk.SystemMessage(
        subtype="init",
        data={"mcp_servers": [{"name": "genealogy", "status": "failed"}]},
    )
    done = sdk.ResultMessage(
        subtype="success", duration_ms=1, duration_api_ms=1,
        is_error=False, num_turns=1, session_id="s1",
    )
    events = _turn_events(RealAgent(tmp_path), [call, dead_init, done])

    assert not [e for e in events if e.get("kind") == "error"], (
        "a session that already made a genealogy call was told research was impossible"
    )
```

### 3. The health read isn't gated on `subtype == "init"`

The harness gates on it (`orchestrator.py:1793`, `if message.subtype == "init"`); the port reads `mcp_servers` off any `SystemMessage`. It's safe today only because no other subtype carries that key — which is also why `test_a_non_init_system_message_is_ignored` passes for the wrong reason. The docstring and that test name both claim init-only, so make it so, in `_mcp_health_events`:

```
You wrote:      data = getattr(message, "data", None)
                if not isinstance(data, dict):
                    return []
Change it to:   if getattr(message, "subtype", None) != "init":
                    return []
                data = getattr(message, "data", None)
                if not isinstance(data, dict):
                    return []
```

### 4. Neither changed site in `sandbox_server.py` has a test

`test_sandbox_server.py` exists but covers none of this. I reverted both edits — the runner-exit text back to the raw exit code, and the spawn failure back to `str(exc)` — and all 200 tests passed. These are the two strings that reach the user as `Chat unavailable: …`, so they're squarely the leak this PR is about. Lowest priority of the four; one assertion on each broadcast covers it.

---

**Gaps in this review, so you know what I didn't earn.** I couldn't reproduce the red `feedback.test.ts` gate — it runs in 797ms here against the 5s default, under turbo as well as alone — so your 2.4s/8.4s measurement is taken on trust. The 30s budget is harmless either way. I also didn't drive a live interrupt or a live 401 (both need a real key and network), so finding 1 rests on reading the pinned CLI's result builder plus a replay through `handle_turn`, not an observed session. And I didn't settle whether the CLI emits an errored `AssistantMessage` for a `rate_limit` it then retries internally — if it does, `map_message` would emit a spurious error and `chatEvents.ts` would leave the whole *successful* answer in a red bubble. I couldn't anchor that to a line, so I'm flagging it rather than claiming it.

aghadiayeamayanvboernest:
Thanks both — all four findings are in as of `845f2c9aa`, merged with `main` (it was 20 behind). `make test-all` green: apps/server 209 passed / 1 skipped, engine 2374, viewer-ui 215, electron 65.

Every fix has a test that I confirmed goes red without it. Both of you did that work on my branch, so it seemed only right to hold the fixes to the same bar.

## 1. Stop reporting a failure (@chesworthrm)

Fixed, and this one was the best catch in the review — I added a field read and shipped a new false alarm on the same line. Guarded on `terminal_reason in ("aborted_streaming", "aborted_tools")`.

Worth adding to your CLI evidence: the SDK's own docstring already says it. `types.py:1249` — *"A value of `aborted_streaming` or `aborted_tools` indicates the turn was cancelled."* So this is the documented contract, not just an observed shape.

## 2. Two contradicting error bubbles in one turn (@clack391)

Fixed. A turn now emits at most one error event and the assistant path wins, since it carries the real `error_kind`. Test drives both messages through one `handle_turn` and asserts exactly one, so it pins the behaviour whichever way the CLI actually behaves.

One deliberate exclusion, flagged in case you disagree: **the MCP-health warning does not count toward the guard.** It reports a different fact — this session has no genealogy tools — rather than a second opinion on the same failure, so suppressing a real API error because we'd warned about the tool surface would lose information. Happy to fold it in if you'd rather have a hard one-error-per-turn rule.

## 3. The MCP-health warning could be unwired silently (@chesworthrm)

Both your tests are in, near-verbatim, with your reasoning in the docstrings. I re-ran both mutations to confirm: killing the `SystemMessage` branch and pointing `GENEALOGY_TOOL_PREFIX` at a non-matching string each red exactly one test now.

You were right that the body overclaimed. "The load-bearing tests drive `serve()`, `handle_turn` and `map_message`" was true of the classification half and not the health half, and I wrote it as though it covered both.

## 4. `sandbox_server.py` had no tests (@chesworthrm)

Both sites covered, both proven by reverting the edits.

One thing worth knowing: **my first draft of that test used `importlib.reload` and leaked into `test_v1_api.py`, reddening six unrelated tests** when the files ran in one session. It uses `monkeypatch` now. Caught it on the full suite rather than the file, which is the argument for running the whole thing before pushing.

## The `subtype == "init"` gate — and why it needed a new test

Both of you asked for this; it's in. But it needed a test neither of you named, and the reason is the interesting part.

`test_a_non_init_system_message_is_ignored` **passes with or without the gate** — I mutated it to `if False:` and all 42 tests stayed green. Its payload carries no `mcp_servers` key at all, so it was proving the `isinstance(data, dict)` check and merely reading as though it proved the subtype check. @chesworthrm called this "passes for the wrong reason" and that was exact.

The added test uses a non-init message that *does* carry a server list — the only shape that separates the two checks. It fails under the mutation.

## The open question you flagged but wouldn't claim (@chesworthrm)

You couldn't anchor whether a retried `rate_limit` could leave a successful answer in a red bubble. **It can't**, and it's anchorable in the bundled CLI.

Every errored assistant message is constructed by one helper:

```js
function Gl({content:e, apiError:t, error:n, ...}){
  let i = mcc({content:[{type:"text", text: e===""?zC:e}], isApiErrorMessage:!0, apiError:t, error:n, ...})
```

and the rate-limit paths call it as `Gl({content: <the error text>, error: "rate_limit"})` — e.g. `` `${Sy}: Usage credits required for 1M context · ${m}` ``. So the errored message's content **is** the error text; it never carries a successful answer, and replacing its blocks with the classified string loses nothing. Recorded at the call site so nobody re-derives it.

Thanks for flagging it rather than asserting it — it was a real gap in my reasoning and it turned out to be checkable.

## On your two review gaps

You couldn't reproduce the red `feedback.test.ts` (797ms vs my 8.4s). Believable — mine is an Intel x86_64 with all suites running under turbo at once, and that measurement is machine-dependent by nature. It fails identically on a clean tree here, which is what made me fix it rather than file it. Agreed the 30s budget is harmless either way.

And noted on #1489 already covering the `make sandbox-image` gap — I'll reference it in the PR body instead of implying it's unregistered.

clack391 (approved):
Re-reviewed at 845f2c9a. Both rounds are fully addressed, and every fix carries a test I confirmed goes red without it: the double-bubble guard, the Stop false-alarm guard, and the subtype=="init" gate each fail under mutation, and the suite is green (209 passed, 1 skipped). Good work holding the fixes to the same bar you were held to. Approving.

jud-sdev (approved):
Approving. I reviewed this independently by running and mutating, not just reading. On the current head (the merge with main) the classification code and its tests are byte-identical to the version clack and chesworthrm reviewed, both new apps/server suites are green (54 passed), and CI is green across the board.

I mutation-checked the guard that carries the most risk: dropping the not-aborted check makes the Stop test fail, so the false "the agent stopped unexpectedly" bubble that a user's own Stop would otherwise produce is genuinely caught. Confirmed the rest by reading — the Stop guard and the double-bubble guard are correctly combined on one condition, the classifier defaults to UNEXPECTED with only 401/403 going MISCONFIGURED, and the health read is gated on subtype init.

Both prior rounds are fully addressed, each fix with a test that fails without it. The diagnosis is the strongest part: catching that the 401 arrived as a normal assistant text block, which none of the issue's five named sites would have caught, is what makes this actually fix what the testers saw.

One operational caveat, not a code issue: make sandbox-image has to run at deploy or production keeps the old message, since the fix bakes into the E2B image and deploy-preflight only warns. Already tracked as #1489.

Good work. Approving.

Issue #1721, opened 2026-08-18
apply_component_derivation is scoped to relationship findings, so a source finding's mislabelled matched is never corrected


EdmondOware:
> **Reviewed 2026-08-18 before junior handoff.** The fork is settled — report the
> disagreement, do not widen the derivation; the corpus cannot measure a widening
> either way. Original body below from "The instance".

**Touches:** eval/harness/e2e/judge.py, docs/specs/e2e-test-spec.md

`apply_component_derivation` (`eval/harness/e2e/judge.py:318`) replaces a finding's model-emitted `matched` with the value derived from its own `components`. It is scoped to `type: "relationship"`:

```python
    eligible = {
        str(f.get("id"))
        for f in findings
        if str(f.get("type")) == "relationship"
        and str(f.get("polarity", "recover")) != "avoid"
    }
```
(`judge.py:363-368`)

So a **`source`**, **`fact`** or **`person`** finding's `matched` is trusted exactly as the model wrote it, even when the model's own `components` — sitting in the same object, in the format the derivation consumes — resolve to a different label. Nothing anywhere reports the disagreement.

**This supersedes the original framing of this issue, which was backwards.** It claimed the stored top-level `verdict` was wrong and should be recomputed at read time in `axes_from_runlog`. That is not the defect. The stored rollup is **correct**; the mislabelled value is one finding's intermediate `matched`.

## The instance

`eval/runlogs/e2e/stribling-father-1821/run-2026-08-17_23-35-44.json`, finding f4 (`type: "source"`, required), at `:102-124`:

```
matched: "partial"

components:
  1. kind: "link"    status: "supported"     "The 24 May 1825 Amite County, Mississippi Orphans'
                                              Court guardianship bond names Gideon Sleeper as
                                              guardian to James, John, Seaborn and Ridley
                                              Stribling, minors"
  2. kind: "link"    status: "supported"     "The guardianship bond is sourced in the tree"
  3. kind: "detail"  status: "contradicted"  "The source URL is .../3QS7-89QX-2C5P (expected)"
```

[...]

Issue #1720, opened 2026-08-18
Spec section 3.1 says per-fixture caps do not exist; the loader and 11 fixtures use them — and stribling-father-1821 needs one


EdmondOware:
> **Decided 2026-08-19 — this is a documentation fix, and only that.** Edit
> `docs/specs/e2e-test-spec.md` § 3.1 to describe `caps` as the loader implements it, and
> require a **new** override to give its reason in the fixture README. **Do not backfill,
> do not touch any `fixture.json`, no ceiling, no validator.**
>
> The lead decision this issue was holding no longer exists: PR #1725 merged 2026-08-19,
> so `stribling-father-1821` is on `main` with `max_cost_usd: 25` /
> `wall_clock_seconds: 10800` / `max_turns: 300`, ratified as landed. Ordered steps in
> [the ruling comment](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1720#issuecomment-5343923123).

**Touches:** docs/specs/e2e-test-spec.md, eval/tests/e2e/stribling-father-1821/fixture.json, eval/harness/e2e/orchestrator.py

Spec §3.1 says per-fixture `caps` do not exist. The loader implements them and eleven committed fixtures use them. §3.1 is stale prose, and it is the kind of staleness that misleads an author into working around a mechanism that is already there.

**Spec §3.1** (`docs/specs/e2e-test-spec.md:148-152`), in bold:

> **Stop-condition limits (`caps`) are NOT a fixture field.** They're a harness safety concern (don't run forever, don't burn the budget), so they live as defaults in the orchestrator (`FixtureCaps`), not in `fixture.json`. Every fixture uses the same caps; an author never writes them. See §6.

**The loader disagrees** (`eval/harness/e2e/orchestrator.py:579-596`). It reads `fixture_json.get("caps")` and fills only the *omitted* keys from `FixtureCaps()` defaults — for all seven caps (`wall_clock_seconds`, `inactivity_seconds`, `progress_stall_seconds`, `tool_calls`, `max_turns`, `max_cost_usd`, `max_continue_nudges`). The dataclass comment states the design outright: *"The DEFAULT caps every fixture inherits for any cap it doesn't set."* Per-fixture override is the intended behaviour, not a leak.

**And eleven fixtures already rely on it:**

| Fixture | Override |
|---|---|
| `hannah-earnest-children` | `wall_clock_seconds: 10800`, `max_turns: 300` |
[...]

Issue #1719, opened 2026-08-18
e2e annotation drift check compares finding ids, not finding content — an amended finding silently invalidates a grade


EdmondOware:
> **Reviewed 2026-08-18 before junior handoff.** Decision settled — the hash covers
> the whole `expected-findings.json`, normalized; the 141 hash-less annotations are
> grandfathered, not backfilled. Original body follows.

**Touches:** eval/harness/e2e/calibrate_judge.py, .claude/skills/grade-e2e-run/SKILL.md, docs/specs/e2e-test-spec.md

`docs/specs/e2e-test-spec.md:1147` makes a drifted e2e annotation a hard error, but defines drift narrowly:

> A drifted annotation (its `per_finding` keys no longer match the fixture's finding ids — i.e. `expected-findings.json` was edited after grading) is a hard error: re-grade or delete it.

That compares **keys to ids**. Nothing compares finding *content*. Amending a finding's body while keeping its id passes the check silently.

**Concrete instance, from this fixture's own grading.** `stribling-father-1821`'s **f4** originally required one specific image — the 24 May 1825 Gideon Sleeper guardianship bond (`ark:/61903/3:1:3QS7-89QX-2C5P`). After `run-2026-08-17_23-35-44` had already been graded against that narrower wording, f4 was broadened to accept **either** that bond or the 1823 bond in which the widowed mother Margaret is guardian (`ark:/61903/3:1:3QS7-L9QX-2H77`). The id stayed `f4` and the key set stayed `f1`–`f4`, so the drift detector stayed silent on an annotation whose f4 label had been produced against a requirement that no longer exists.

**What that costs downstream.** Left unfixed, `calibrate_judge` re-runs the judge against the stored final tree using the **amended** f4 — which that run satisfies, since it attached the 1823 bond — reads the human `false`, and records a judge/human disagreement that never happened. The two sides were answering different questions. `spec:1172` then tells the maintainer that *"the disagreements are the signal … read those before the aggregate"*, so the fabricated disagreement lands in exactly the output a human is directed to read first, and it invites a judge-prompt fix for a judge that was not wrong.

This was caught by hand during a re-grade, because the same person happened to remember the amendment. Nothing would have surfaced it otherwise.

[...]

Issue #1716, opened 2026-08-17
project-status requires references/output-formats.md, which does not exist


EdmondOware:
**Touches:** packages/engine/plugin/skills/project-status/SKILL.md, packages/engine/plugin/skills/project-status/references/

Found while working #1413's fixture (`eval/tests/e2e/stribling-father-1821/`).

`project-status`'s SKILL.md instructs the agent to read `references/output-formats.md` — twice, as a required step — and that file does not exist.

## Details

`skills/project-status/SKILL.md`, "Two summaries" section:

> The literal output shapes for both are in `references/output-formats.md` — read it before presenting (step 5).

And again at step 5, "Present both summaries":

> Render both using the literal skeletons in `references/output-formats.md` — the detailed GPS-state block and the plain-language story.

Actual contents of `packages/engine/plugin/skills/project-status/references/`:

```
conclusion-readiness.md
project-exhaustiveness.md
```

**Confirmed absent from the source repo, not just a synced copy** — so this is a skill defect rather than a packaging or sync problem.

## Impact

Low severity but real. The skill's primary deliverable is two specifically-shaped summaries, and the shapes are unspecified, so each invocation improvises a layout. Output is inconsistent between runs and between agents, which defeats the purpose of prescribing "literal skeletons". An agent following the instruction faithfully also burns a tool call discovering the file is absent — observed.

## Suggested fixes

- Add `references/output-formats.md` with the two skeletons the skill promises, **or** remove both references and inline the expected shape in SKILL.md.
- Worth a sweep for the same class of drift: `validation-protocol.md` is referenced by several skills in this plugin and is present in some reference directories but not others.
- A CI check that every `references/<file>.md` named in a SKILL.md exists in that skill's reference directory would catch this mechanically. If the lead wants that lint, it is a `developer` change and probably its own issue — a skill can promise a reference doc that isn't there and nothing reports it.

Exemption: **it is a different skill's eval slot, and bundling it would force a second paid run** — an edit under a skill dir invalidates that skill's run-log snapshot.

DallanQ:
Merged into issue #1633.

[...]

PR #1725, opened 2026-08-18
Add stribling-father-1821 e2e fixture: stepfather-guardianship inference (#1413)

EdmondOware:
## Update (2026-08-18, after three rounds of co-review)

Every open question from the original PR is resolved. Commits `038eb998`, `71e7cb97`,
`1a0959bd`, and `5699d318` cover this; **read those, not the "Review guidance" section
below, which is now historical** (it reflects the state before any co-review round and
is kept only as a record of what was asked).

- **f4 attachment-target amendment: approved and applied** (johnmarkpeterbrown). f4
  now accepts attachment to the subject's person record OR any relationship the
  subject is an endpoint of, worded to override `details.target_person` explicitly.
  The re-grade was a no-op — both `.ann.json` already had f4 `true`.
- **f5 added**: Margaret McDowell's remarriage to Gideon Sleeper, required as a
  **dated** `Couple` relationship fact on the person actually linked to the subject
  as mother — closing the gap where both scored runs satisfied f4 via the 1823 bond
  (widowed mother as her own children's guardian) without the stepfather situation
  #1413 is about ever entering either deliverable.
- **f5's initial grade on run 2 (`true`) was wrong and has been corrected to `false`**
  (florencemashipei). Run 2's Couple relationship for the remarriage carries no dated
  fact — the identical defect f3 was already graded `false` on for this same run —
  and separately sits on an unmerged duplicate of the mother rather than the person
  the tree actually links to the subject. Both gaps are independent; either alone
  would be enough to regrade.
- **f5's `marriage_place` was wrong** (florencemashipei) — "Liberty, Amite County,
  Mississippi" was Margaret's *death* place bleeding across from f2, unsupported by
  any source for the marriage. Corrected to "Amite County, Mississippi", matching f3
  and the actual evidence.
- **f5 is now a bonus finding, `required: false`, not a required one**
  (johnmarkpeterbrown, confirmed by florencemashipei). With no committed run
  satisfying f5 as a required bar, keeping it required meant fixture validity stayed
  entangled with a bar neither run had cleared, and clearing it would cost another
  paid run. As a bonus finding it keeps measuring the stepfather inference exactly as
  written (dated fact, correct person, graded on f3's standard) without gating
  anything. **`Closes #1413` is changed to `Refs #1413`** for the same reason: a
  bonus finding measures the inference, it doesn't gate the issue that asked for it.
- **Cost-cap straddle**: resolved by adding per-fixture `caps` to `fixture.json`
  (`max_cost_usd: 25`, `wall_clock_seconds: 10800`, `max_turns: 300`), sized to run
  2's observed straddle ($15.95/119.8 min against the $15 default).
- **Current state, four required + one bonus finding**: run 1
  (`run-2026-08-17_23-35-44`) satisfies its full required set (f1–f4), stored
  `verdict: pass` / `recall_required: 1.0`; run 2 (`run-2026-08-18_02-03-06`) misses
  f3 on its required set. Neither run satisfies bonus f5 yet — that's the stepfather
  inference specifically, tracked separately and honestly as not yet demonstrated,
  without affecting fixture validity.

## Summary

- Adds `eval/tests/e2e/stribling-father-1821/`, a strip fixture for #1413: whether the agent reads a guardian appointed shortly after marrying the wards' mother as a **stepfather implying a prior marriage**, rather than as a surname conflict. Subject Simpson Ridley Stribling (`KLYT-G3Q`), the "Ridley Stribling" of a 24 May 1825 Amite County, Mississippi guardianship bond in which Gideon Sleeper, who had married their widowed mother six months earlier, is guardian to four Stribling minors.
- **Both parents are stripped**, and a `type: "source"` finding (f4) requires a guardianship bond to be attached in the final tree. That combination is the design decision this PR is really about.
- **Four required findings plus one bonus finding (f5), added by co-review as a measured, non-gating discriminator** (see Update above). **Two scored runs are committed**, each with its blind `.ann.json`: `run-2026-08-17_23-35-44` (full required set + bonus f5 not yet met) and `run-2026-08-18_02-03-06` (misses required f3; bonus f5 not yet met).
- **The fixture is validated per spec §14.** Run 1's committed log carries judge `verdict: "pass"` with `recall_required: 1.0` over the required set (f1–f4), so the answer is demonstrably recoverable from live FamilySearch. A run demonstrating bonus f5 (the stepfather inference specifically) is future work, tracked via `Refs #1413`, not a claim this PR makes. Tier: **Normal**.

## Review guidance (historical — see the Update note above for current state)

**Start here: the PROPOSED f4 amendment, which I have deliberately not applied.** From run 2's judge, verbatim:

> "The sources are present on relationships but not directly attached to the subject's person record. This is a technical miss on the source-attachment requirement, even though the agent clearly engaged the guardianship evidence and recovered the parentage correctly."

That is f4 failing the exact capability it exists to measure. The mechanism: f4's `details.target_person` names the subject, so the judge derives f4's only `link` component as "must be attached to Simpson Ridley Stribling's **person record**". Both runs attached the bond to the parentage relationships instead (run 2 as S5 on R35/R36, run 1 as S7 on R34/R35), which is the normal place to source a parentage conclusion. f4's own prose says only "attached as a source in the tree"; `details.target_person` is what silently narrows it, and `finding_shape_errors` validates top-level keys only, so nothing flags the narrowing. Left as written, f4 produces a false negative on every run that sources relationships rather than the person, and each one needs a hand annotation.

**Proposed fix:** f4 is satisfied by attachment to the subject's person record **OR** to any relationship in which the subject is an endpoint.

**Why it is not applied.** `expected-findings.json` was already broadened once inside this PR, on 2026-08-17, when f4 moved from one ARK to the record class. A second unilateral amendment re-invalidates both committed `.ann.json` files, and #1719 is precisely the finding that the annotation drift check compares finding **ids** rather than finding **content**, so neither annotation would be flagged as stale. Amending a finding after runs have been scored against it is a co-reviewer decision, and it is the decision I am asking for here. The fixture README carries the full note under "PROPOSED f4 amendment", marked as awaiting approval.

**Second thing to check: run 1's f4 carries `matched: "partial"` over two `supported` link components.** `derive_matched` (`eval/harness/e2e/judge.py:255-277`) over those same components returns `true`; the only `contradicted` entry is a `detail` on the ARK, which `judge_prompt.md:106` states never lowers the label. The judge mislabelled its own analysis, and `apply_component_derivation` did not correct it because that function is scoped to `type: "relationship"` (`judge.py:363-368`) and f4 is a `source` finding. Filed as #1721.

**This does not affect the fixture's validity.** The stored top-level `verdict: "pass"` and `recall_required: 1.0` are the arithmetically correct values: with f4 read from its components as `true`, `derive_verdict` gives exactly those figures. The intermediate label is the wrong value, not the rollup, and correcting it leaves the rollup where it stands. §14 is met.

**Both items below are resolved — see the Update note at the top:**

1. ~~**The f4 attachment proposal above.** Approve it and I apply the amendment and re-grade both annotations against the amended finding; decline it and f4 stays hand-graded on this axis with the reason recorded in each `.ann.json`.~~ Approved and applied.
2. ~~**The cost-cap straddle.** Run 1 completed at **$10.32 / 78.7 min**. Run 2 hit `cost_cap` at **$15.95 / 119.8 min**, against a **$15** default...~~ Resolved with a per-fixture `caps` block.

## Plan

**Didn't change:** no plugin prose, no unit test, no schema. The doctrine this fixture exercises is genuinely missing from the plugin, and writing it here was the tempting adjacent change. It is filed as #1710 instead, because skill prose goes through `docs/skill-lifecycle.md` with its own paid eval run and annotation, and folding it in would put three skills in one diff.

**Acceptance check:** both gates exit 0 on this branch and fail on `main` (the fixture directory does not exist there):

```
make e2e-validate TEST=stribling-father-1821                                          # exit 0
cd eval/harness && uv run python -m e2e.author validate --slug stribling-father-1821  # exit 0
```

Both are needed: `make e2e-validate` runs the stripping linter only, while `presence_mirror`, the 110-year living-gate heuristic and the subject-id cross-check live in `e2e.author validate`. **Exactly two WARNs are expected, both on f4**, because its `details.target_person` is the retained subject and only `subject_person` keys are pruned before tokenising, so the bag matches the subject `KLYT-G3Q` and his same-named son `KPHL-7TZ` ("Ridley Lucian Stribling"). Any other WARN is a real signal. Verified: both gates exit 0 with two WARNs and no others, re-verified after every subsequent commit including the f5 addition and regrade.

Spec §14 is met by run 1's committed log, `eval/runlogs/e2e/stribling-father-1821/run-2026-08-17_23-35-44.json`, which carries `verdict: "pass"` with its `.ann.json` alongside, over the original four findings.

**Deviated from the plan: all four deviations are recorded in `PLAN.md` § "Deviations from this plan".** Summarised:

1. **Two WARNs from the stripping linter, not one.** The plan predicted a single f4 WARN on the retained subject; the subject's son is also "Ridley Lucian Stribling", so he trips the same both-halves match. Same benign class, one extra row, recorded in the fixture README as a reviewer note.
2. **`scaffold` ran after `strip`, not before.** `fixture.json` and `starting-research.json` were written last, and the landing gate failed once on the missing `starting-research.json`. Harmless, since `scaffold` writes only those two files and does not touch the tree; both gates are green on the final state.
3. **f4 was broadened once, on evidence, and the run that prompted it was retained rather than discarded.** As planned, f4 demanded one specific ARK, the 1825 stepfather bond. Both scored runs independently attached the **1823** bond, in which the widowed mother herself is guardian, and **neither** attached the 1825 one: run 1 surfaced it as the first full-text hit and read adjacent pages, run 2 read the 1825 image and cited it in the narrative but wrote its source entry with a bare full-text search URL. Two independent trajectories converging on the 1823 bond is what makes the broadening evidence-led rather than score-rescuing, and the 1823 bond is the more direct parentage document of the two. f4 now names the record class. Both runs stay committed per spec §8; what went stale was run 1's grade, not its log, and its `.ann.json` was re-graded against the amended finding while its `judge_output` was left exactly as written.
4. **Two corrections to already-committed content were made in follow-up commits rather than amends.** First, the fixture README's §14 paragraph claimed the fixture was unvalidated. That was wrong: run 1's log carries `verdict: "pass"`. The claim had been asserted while blind to the field it described, under the `/grade-e2e-run` blind rules that forbid reading `verdict`; the correct move was to mark §14 unverified and hand it on. Corrected, and then corrected again upward to plain "validated" once run 1's f4 components were read (see §Review guidance). Second, two grading-provenance attributions were wrong in the same direction: the README said run 2's f3 `false` was the mechanism's, when the judge graded f3 `true` and the `false` is the genealogist's blind grade alone (so run 2's 0.75 recall comes wholly from f4); and it cited run 2's f4 `false` as evidence that `details.source_url_alternate` is ignored, when that judge's own component reads "(1823 or 1825)" and `judge_prompt.md:47-48` already mandates ARK tolerance. Both corrected. **No run log, finding or annotation was rewritten, and no commit was amended.**

A fifth deviation, from the two co-review rounds: **f4's amendment and citation were applied, f5 was added, and f5's own grade on run 2 was itself corrected mid-PR** (see Update note at the top). Consistent with deviation 4 above, this happened via new commits, not amends — no run log, finding text before correction, or prior annotation state was rewritten in place without a visible commit recording the change.

Step-4 stop rule: **hit once, on design rather than code.** `/critique-plan` round two returned a BLOCKING finding, that the original two-finding shape did not discriminate the misread because retaining the mother let an agent recover the answer from one indexed marriage search without ever opening the bond. Accepted in full. Fixture adjudication sits with `genealogist` per `CLAUDE.md`'s label table, so I adjudicated it to the current design and recorded the decision, the rejected alternatives and the residual limits as a comment on #1413 rather than escalating for a third planning round.

## Test plan

- [ ] ~~I ran `make test-all`~~ — **not run: no Windows equivalent exists (#1185).** `eval\RunTests.bat` is the paid per-skill eval run, not this gate. Leaning on CI, which runs the same suites. Nothing in this diff is executable code.

- [ ] ~~`make eval-skill SKILL=<name>`~~ — **not owed.** No file under a skill dir, a delegated agent, `eval/tests/unit/**`, or a referenced scenario/mcp fixture changes, so no run-log snapshot is invalidated. `check-runlogs.yml` is not triggered by an `eval/tests/e2e/**`-only diff.

- [ ] ~~Negative tests on both sides of a routing pair~~ — **not applicable.** No skill `description` frontmatter or DO-NOT clause changes.

Run and grading, which `check-e2e-fixtures.yml` does gate:

- [x] `make e2e-run TEST=stribling-father-1821` — **run twice.** `run-2026-08-17_23-35-44` (`completed`, $10.32) and `run-2026-08-18_02-03-06` (`cost_cap`, $15.95). Both logs committed with their final tree and research sidecars.
- [x] `/interpret-e2e-result` on both runs.
- [x] `/grade-e2e-run` **blind** on both; `run-2026-08-17_23-35-44.ann.json` and `run-2026-08-18_02-03-06.ann.json` are both committed. Run 1's annotation was re-graded against the amended f4 after the broadening, which is a deliberate post-amendment re-grade rather than the blind product, and its note says so. Both were re-graded again for f5 after co-review added it, and run 2's f5 was corrected a second time after co-review caught a grading inconsistency (see Update note at the top) — none of this is the original blind product either, and each note says so.
- [x] `/check-drift` against `PLAN.md` — four findings, all resolved, and all four are now recorded in `PLAN.md` § Deviations, plus a fifth covering the two co-review rounds.
- [x] `/code-review high` in a fresh session — four findings. Three fixed in this PR (the f3 attribution, the `source_url_alternate` inference, and `fixture.json`'s stale `notes`, which still described the 1825 bond as the only route). The fourth, the f4 `details.target_person` defect, was escalated to the co-reviewer and is now resolved (see Update note).

**Reviewers:** `/eval/tests/` and `/eval/runlogs/` are both `senior-genealogists` in `.github/CODEOWNERS` (last-match-wins over `*.json`), so one senior-genealogist approval unblocks merge, and no `senior-developers` approval is required by this diff. I am a senior genealogist and cannot approve my own PR, so this needs a second pair of eyes from that team.

## Follow-on issues

#1708, #1709, #1710, #1711, #1712, #1715, #1716, #1719, #1720, #1721

**Folded in / filed instead:** nothing folded in — every defect this work surfaced sits outside a fixture PR's ownership or its eval budget.

- **#1708** (`developer`, `nothing-checks`) — `image_read` refuses an oversized scan with no fallback to `image_transcribe`, so five sources were flagged `ORIGINAL NOT EXAMINED` on a false premise and a correct conclusion was blocked by a readable image. Filed rather than folded: *needs a different reviewer or skill* (a `.ts`/`.py` change owned by `senior-developers`).
- **#1709** (`genealogist`) — `hypothesis-tracking`'s `supported` gate requires a direct-evidence line, which no indirect-evidence proof argument can satisfy, and contradicts that skill's own reference. Filed rather than folded: *a different skill's eval slot; bundling would force a second paid run*.
- **#1710** (`genealogist`) — the guardianship → prior-marriage doctrine prose, #1413's deferred half. Same reason as #1709.
- **#1711** (`genealogist`) — `proof_summaries.tier` is single-valued while a parentage question carries two claims at different tiers. Filed rather than folded: *depends on a decision only the lead can make*.
- **#1712** (`developer`) — `snapshot`'s duplicate-fact-id WARN over-states itself. Same reason as #1708.
- **#1715** (`developer`, `nothing-checks`) — concurrent `extraction_append` / `research_append` calls race on id allocation; one write wins, the loser gets a success response naming the winner's ids, and an entire record's extraction is lost with `validate_research_schema` returning clean. Reproduced in this fixture's debug project. Same reason as #1708.
- **#1716** (`genealogist`) — `project-status` requires `references/output-formats.md` twice as a step, and the file is absent from the source repo. Same reason as #1709.
- **#1719** (`developer`, `nothing-checks`) — the annotation drift check compares finding **ids**, not finding **content**, so amending a finding silently leaves every existing `.ann.json` scored against the old wording with nothing flagging it. Same reason as #1708.
- **#1720** (`developer`) — spec §3.1 says per-fixture `caps` do not exist, while the loader and twelve fixtures use them. Same reason as #1708.
- **#1721** (`developer`, `nothing-checks`) — `apply_component_derivation` is scoped to `relationship` findings, so a `source` finding's mislabelled `matched` is never corrected against its own components. Same reason as #1708.

The two worth a reviewer's attention beyond this PR are **#1708** and **#1715**: both are cases where the system reported success while producing wrong or incomplete output, which is what makes a passing run untrustworthy.

**#1719 and #1721 are worth noting for a different reason: both were surfaced by grading and reviewing this fixture's own runs.** #1719 came out of amending f4 mid-PR and asking what happens to annotations already scored against the old wording; #1721 came out of reading run 1's f4 components instead of its label. That is the calibration path exercising itself, which is the only way either would have been found.

Refs #1413


johnmarkpeterbrown (commented):
## Draft review comment

Reviewed as senior genealogist. The case itself holds up — I checked the chronology independently and it's internally consistent: Taliaferro b. 1781 m. 1815 at 34, d. 1823 at 42; Margaret b. 1793, m. at 22, widowed at 30, remarried Nov 1824, d. 1864 at 71; subject b. 1821 is ~4 at the May 1825 bond, and four minors from a 1815 marriage is right. No living persons. Both gates clean, exactly the two f4 WARNs you predicted, and I confirmed those can't be designed away — renaming `details.target_person` doesn't silence them, because `_TARGET_KEYS` includes `name` and the nested key still collects. Keeping the shape and documenting the noise was the right call.

Three genealogical findings, in order of what they cost.

### 1. The 2026-08-17 broadening dissolved the discriminator. This is the decision I'd ask you to revisit, not the attachment target.

#1413's acceptance criterion is specific: *"expected findings require the stepfather/prior-marriage hypothesis to be raised, and which fails if the agent reports the surname difference as a conflict instead."*

f4 now accepts the 1823 bond, in which **the widowed mother is guardian of her own children**. Margaret Stribling over Stribling minors. There is no Sleeper in it, no McDowell, and no surname difference of any kind. So f4's own claim — that either bond *"shows the agent engaged the guardianship evidence rather than dismissing the Stribling / Sleeper / McDowell surname spread as a conflict"* — is not true of option (b). The 1823 bond never presents the surname spread, so attaching it demonstrates nothing about the inference.

Your two runs are the proof. Both attached the 1823 bond; **neither** put the 1825 stepfather bond in the tree. Run 1 never read it, run 2 read it and cited it with a bare search URL. Two runs have now cleared f4 without the stepfather situation ever entering the deliverable.

With both parents stripped, every finding is reachable through direct records — the 1815 marriage index gives you f3 and Margaret's maiden name for f2, and the 1823 bond plus the estate gives you f1 and f2's links. The stepfather reading is optional on every path. What you've built is a strong Amite County orphans'-court **records-discovery** fixture. It isn't yet the reasoning fixture #1413 asked for, and I don't think it should close #1413 as written.

### 2. f4's citation for option (b) states something the record does not say

This is the one I'd fix regardless of how you resolve the above. `supporting_sources[1]`:

> "Amite County Orphans' Court guardianship bond, 1823, naming the widowed Margaret Stribling as guardian of the minor heirs of Taliaferro Stribling."

The transcription in your own run 2 (call 265, ark `L9QX-2H77`) reads *"...perform the duty of guardian to the said James, John, Seaborn & Redley Stirling..."* and *"the Estate of the Orphan under her care."* It names **neither Margaret nor Taliaferro.** The female possessive is the only thing pointing at Margaret, and the father appears nowhere.

Your own run agrees. Assertion `a_204` on that ark records the father-of relationship as *"(inferred from orphans court guardianship appointment)"*, and `pe_045` carries it at `probable` — the agent correctly declined to treat it as stated.

f4 also asserts each bond is *"establishing the Stribling minors' parentage."* Neither does, on its own. The 1823 bond names no parent; the 1825 bond names a stepfather. And your changelog already says the quiet part — *"The decisive document is the 1836 deed, not the bond"* — yet no finding names the deed. A fixture whose text overstates its key record teaches both the agent and the judge that the record says more than it does, and the judge reads f4's prose verbatim.

```
You wrote:      "\"Mississippi, Probate Records, 1781-1930,\" images, FamilySearch (https://familysearch.org/ark:/61903/3:1:3QS7-L9QX-2H77 : accessed 17 August 2026), Amite County Orphans' Court guardianship bond, 1823, naming the widowed Margaret Stribling as guardian of the minor heirs of Taliaferro Stribling."
Change it to:   "\"Mississippi, Probate Records, 1781-1930,\" images, FamilySearch (https://familysearch.org/ark:/61903/3:1:3QS7-L9QX-2H77 : accessed 17 August 2026), Amite County Orphans' Court guardianship bond, 1823, binding a female guardian to \"perform the duty of guardian to the said James, John, Seaborn & Redley Stirling\"; the bond names neither the guardian nor the minors' father in its condition clause, so Margaret's identity and Taliaferro's paternity are both inferred from it rather than stated."
```

And in f4's description, `establishing the Stribling minors' parentage` should become `bearing on the Stribling minors' parentage`.

### 3. The fix that doesn't cost you a run — and the README claim that's blocking you from seeing it

You wrote that a finding naming Gideon Sleeper is inadmissible because `presence_mirror` hard-fails a non-`avoid` finding naming somebody absent from the unstripped tree. **That isn't the constraint.** `presence_mirror` matches on the finding's *target* name bag, not on its description — anchor the finding on Margaret and put Sleeper in the prose, and the gate never sees him.

I tested it. A fifth required finding anchored on Margaret McDowell, with Sleeper named only in the `description` and in a non-target `details` key:

- `presence_mirror` → `[]` (no errors), same as the shipped four
- name token bag → `{margaret, mcdowell}`, and no person in the starting tree matches either, so **no new stripping WARN**

So you can capture the actual inference without touching either committed run's grade, and without a third paid run to validate the fixture — run 1 still carries `verdict: pass` on the genealogical axis for f1–f4. Suggested shape:

```
f5, type relationship, required true
description: The agent must record that Gideon B. Sleeper's May 1825 guardianship of the four
  Stribling minors follows his November 1824 marriage to their widowed mother Margaret, and
  therefore indicates a step-relationship arising from her prior marriage to Taliaferro
  Stribling — not a surname conflict. The tree must carry Margaret's 1824 remarriage to
  Sleeper as a Couple relationship, and must not assert Sleeper as the subject's father.
details.subject_person: Simpson Ridley Stribling (KLYT-G3Q)
details.target_person:  Margaret McDowell        <- keeps presence_mirror green
details.stepfather_note: Gideon B. Sleeper, m. Margaret 23 Nov 1824
```

Grade it on the Margaret–Sleeper `Couple` link, which is objective from the tree. Run 2 already has it (`R34`, on the unmerged duplicate `I3`); run 1 does not. That gives you a finding the two runs actually split on — which is what a discriminator is — and it's the finding #1413's acceptance criterion is describing.

If you'd rather not add a finding in this PR, then say plainly in the README that the fixture as shipped tests records discovery rather than the stepfather inference, and leave #1413 open for the reasoning half alongside #1710. What I can't approve is the current pairing of "Closes #1413" with an f4 that a mother-as-guardian bond satisfies.

### The question you actually asked: yes, approve the attachment-target amendment

Attachment to the person record **or** to any relationship where the subject is an endpoint. Both runs put the bond on the parentage edges and nowhere else — run 1's subject person record has zero sources — and that is the correct place to source a parentage conclusion. Grading it as a miss penalises good practice.

Two things to fix in how you've framed it, though. Your diagnosis says f4 *"mis-grades every run that sources relationships rather than the person."* It doesn't: run 1 did exactly that and its judge marked the component **supported**, while run 2's marked the same shape **unsupported**. The defect is that f4 grades identical genealogical behaviour two different ways — which is a worse problem than a uniform false negative, and a stronger reason to amend.

Word it so it overrides `details.target_person`, since that key stays in the file:

```
You wrote:      must be attached as a source in the tree.
Change it to:   must be attached as a source in the tree — either on the subject's person record OR on any relationship in which the subject is an endpoint. `details.target_person` names the person the source hangs on, not the required attachment point.
```

And the re-grade you're worried about is a no-op: both `.ann.json` files already grade f4 `true`, and both stay `true` under the amendment. No label moves, so #1719 has no live consequence here. That removes your reason to hold it.


```
You wrote:      "difficulty": "hard",
Change it to:   "difficulty": "hard",
                "caps": {
                  "max_cost_usd": 25,
                  "wall_clock_seconds": 10800,
                  "max_turns": 300
                },
```

Also: `README.md:185` and `:199` cite `(:114)` and `(:111)` with no filename. They resolve to `run-2026-08-18_02-03-06.json`, but the last file named is `expected-findings.json`, which is 88 lines.

florencemashipei (commented):
Reviewed as senior genealogist, second pair of eyes. Reviewing commit `038eb998`, not the PR body — see the note at the end.

The three findings from the first review pass are all addressed, and addressed on their own terms rather than argued around: f4's attachment amendment is worded to override `details.target_person` explicitly, the 1823 citation now describes the condition clause instead of overstating it, and f5 exists. Adding f5 rather than dropping "Closes #1413" was the right of the two options offered. Per-fixture `caps` are in `fixture.json`, sized to run 2's observed straddle.

One thing blocks my approval, and it is the grade the rest of the argument stands on.

### Run 2's f5 `true` is graded by a standard this PR rejected one finding earlier

f5 carries the whole answer to the first review's finding #1. It is what makes f5 "the finding the two runs actually split on" rather than a required finding nobody has met, and therefore what supports "Closes #1413". So it needs to hold up on its own.

The tree it is graded against (`run-2026-08-18_02-03-06.final-tree.gedcomx.json:1428-1437`):

```json
{
  "type": "Couple",
  "person1": "I3",
  "person2": "I4",
  "sources": [{ "ref": "S3" }],
  "id": "R34"
}
```

`R34` carries **no facts** — no `Marriage`, no date. `23 Nov 1824` appears in that tree in exactly one place: inside the S3 source title (`:1600`).

Now your own f3 note on the same run (`run-2026-08-18_02-03-06.ann.json:13`):

> "Couple relationship R33 (I1+I2) exists but carries no facts at all -- no marriage date, no place... The string '1 Mar 1815' appears in the tree only inside the S2 source title. The agent asserts the 1 March 1815 Amite County marriage in the proof narrative and cites the index, but never materialized it as a fact on the relationship. **Graded false on the tree:** the dated marriage fact the finding calls for is not encoded."

Same shape, opposite grade. And f5 asks for the date in the same breath f3 does — `expected-findings.json:90`: *"The tree must record Margaret McDowell's remarriage to Gideon B. Sleeper (23 November 1824) as a Couple relationship."*

**In fairness, the two are not identical.** R33 is defective twice over — no facts *and* sourced to S4, the 1823 probate, rather than the marriage index. R34 is sourced to S3, which *is* the 1824 marriage index: the right record, correctly attached. So R34 is the better relationship of the two. The "no facts" defect is what they share, and it is the half f3 was graded `false` on.

### The second half: the remarriage hangs off a person who is not the subject's mother

`R34` is on `I3`. The subject's mother edge is `R36`, `I2 -> KLYT-G3Q` (`:1455`). `I2` and `I3` are both Margaret, unmerged. Your f5 note grants this and grades `true` on the strength of `h_002` recommending the merge — but the merge was not executed, so in the tree as delivered, the woman who married Sleeper is not the woman who is the subject's mother. This half does not go away under any reading of the date question.

### What I am asking for

Not necessarily a regrade. There is a reading where f5 stays `true`: f3 is *about* a dated marriage, so a missing date is fatal to it, whereas f5 is about the step-relationship being **encoded at all**, and the date is incidental. I find that arguable, and R34's correct sourcing supports it. But it is currently unstated, and two findings on one run are graded inconsistently with nothing on the page reconciling them.

So, either:

- **(a)** regrade run 2's f5 against the standard f3 got, or
- **(b)** keep `true` and state in the `.ann.json` note why the encoded date is required for f3 and not for f5, and why a `Couple` on an unmerged duplicate satisfies "the tree must record" — with the `I3`/`I2` split named as an accepted limitation rather than as a defect the hypothesis excuses.

I would take (b) with the reasoning written down. What I cannot approve is the grade standing without it, because **if f5 goes to `false`, no committed run satisfies f5** — it becomes a required finding neither run has met, the "two runs split on it" argument goes away, and the reasoning half of #1413 is undemonstrated on live FamilySearch. That is a different fixture than the one the README describes, and the reader should not have to derive which one they have.

### Two smaller things

- **`README.md:618` understates run 2.** It reads *"run 2 is 3 of 5 (f3 and f5 mixed)"*. Its `.ann.json` records `f1 true, f2 true, f3 false, f4 true, f5 true` — 4 of 5, and f5 is not "mixed", it is `true`. Looks like the pre-f5 figure of 3 of 4 carried forward. Note this line becomes correct if f5 regrades to `false` under (a), so fix it after the question above resolves, not before.
- **The PR body predates `038eb998`.** It still opens *"Start here: the PROPOSED f4 amendment, which I have deliberately not applied"* and asks the co-reviewer to decide it, and reports run 1 as 4 of 4. That decision is applied and the count has moved. It is the first thing the second approver reads — please bring it up to the branch.

### Not blocking, for the record

**§14 does not hold this up.** No committed run is an all-five pass, and the README says so plainly rather than smoothing it — which is the right call. Spec §14 is explicit that it is "a **recommended authoring practice, not a CI merge blocker**, so a draft fixture can land with its validity run still owed" (`docs/specs/e2e-test-spec.md:1519-1521`). The owed run is recorded as future work in the changelog. That is sufficient; it does not need an issue.

The genealogical case itself I did not re-derive — the first review pass checked the chronology independently and I have no reason to reopen it.

florencemashipei:
`71e7cb97` resolves my blocking item, and resolves it better than I asked for. Regrading was the harder of the two options and the right one; tightening f5's own text so the bar is stated rather than inferred a second time is what stops this recurring on the next run. Separating the missing dated fact from the wrong-person attachment as two independent gaps is correct — either alone is disqualifying. Thank you for checking `R34` and `R36` against the tree yourself rather than taking my word for it.

Two small things and I'll approve.

### f5's new `marriage_place` names a place nothing supports

The same commit added:

```json
"marriage_place": "Liberty, Amite County, Mississippi"
```

Nothing in evidence gets below the county:

- **S3**, the only source for that marriage, names no place at all — `"entry for Gideon Sleeper and Margaret Stribling, 23 Nov 1824"` (`run-2026-08-18_02-03-06.final-tree.gedcomx.json:1600`)
- **f5's own `supporting_sources`** says county level: *"1824, Amite County, Mississippi"*
- **"Liberty" appears nowhere** in run 2's tree
- **f3, the parallel finding, uses `"Amite County, Mississippi"`**

Liberty is Margaret's *death* place, from f2. I think it bled across.

This is the defect class this PR has now fixed twice — a `details` key narrowing a requirement past what the prose and the sources carry, the same shape as f4's `target_person`. Left in, a run that correctly records the marriage in Amite County can be graded down for not naming a town no record puts it in. Suggest just matching f3: `"Amite County, Mississippi"`.

### The PR body is still pre-`038eb998`

It opens *"Start here: the PROPOSED f4 amendment, which I have deliberately not applied"* and reports run 1 as 4 of 4. Both moved two commits ago. It's what a third reviewer reads first.

### For @johnmarkpeterbrown — a scope call, not a review item

Your finding #1 has been answered on the encoding side: f5 exists, it requires the dated remarriage on the mother's own node, and it is graded on f3's standard. But with run 2 regraded, **no committed run has ever satisfied f5** — the README now says so plainly at `:111`, "designed, not yet demonstrated by a passing run."

So "Closes #1413" now pairs with a required finding nothing has met. §14 doesn't block that (`e2e-test-spec.md:1519-1521` — recommended practice, validity run may be owed), and I read it as defensible: run 2 got most of the way there, finding the 1824 index, creating the `Couple`, spotting the duplicate and recommending the merge in `h_002`, and fell short only on materializing the fact and executing the merge. That reads as achievable rather than impossible.

But it is your objection in a new form, and whether #1413 closes on a designed-not-demonstrated discriminator is a scope decision rather than a genealogical one. I'm not going to close it out on your behalf. My approval covers the genealogical case and the grading consistency, both of which I'm now satisfied with.

EdmondOware:
Fixed the marriage_place, changed to Amite County Mississippi, matching f3 exactly, in commit 1a0959bd. Verified your read before changing it, S3 names no place, f5's own supporting sources already say county level, and Liberty is genuinely Margaret's death place from f2. Same narrowing defect shape as target_person, good catch.

On the PR body, I checked just now and it already has the update section covering both 038eb998 and 71e7cb97, so that must have been a timing issue on your end reading it before the edit landed, not something I missed. Nothing further needed there.

johnmarkpeterbrown (commented):
*PR #1725 — stribling-father-1821 fixture* — reviewed. Good work, three changes and I'll approve. None needs a re-run.

The key finding: *run 1's final tree has no Gideon Sleeper in it at all.* That's the fixture's only validating run, and it scored 4/4 without ever putting the stepfather in the tree. So the fixture doesn't yet test the inference it was built for.

*What I need:*

1. *Fix f4's wording.* I read both bond images — neither one names Taliaferro. So "establishing the parentage" isn't accurate; make it "bearing on the parentage," and fix the 1823 citation (it names Margaret, who signs by mark — not the father). Same edit folds in your attachment-target question: *yes, approved* — person record OR any relationship where the subject is an endpoint.

2. *Add f5 for the stepfather inference — `required: false`, not true.* The 1823 bond has a same-surname guardian, so accepting it dissolves the discriminator. As a *required* finding, run 1 fails it and drops from pass to partial, which costs you another $10–16 run. As a bonus finding, validity holds and the inference gets measured. Anchor it on Margaret, not Sleeper — the presence check reads the target name, and I tested that it stays green.

3. *Change "Closes #1413" to "Refs #1413."* The issue asks for findings that *require* the hypothesis. A bonus f5 measures it, doesn't gate it. Leave the issue open for the flip.

*Nice-to-haves, not blocking:* there's a fifth Stribling child (Peggy — her own 1823 bond on the same page, name clashes with the mother's); Taliaferro had a first wife, worth pre-empting; birth year is 1821 in the fixture but 1822 in the tree's life sketch; and add a `caps` block — run 2 blew three caps at once.

*Don't:* re-run anything, rewrite the run logs, or make f5 required.

Exact diffs are in my PR review. Ping me when 1–3 are in.

EdmondOware:
Thanks both. Checked point 1 and the caps note against the current head, both already applied, wording, attachment target, and the 1823 citation all match what was asked, and the caps block is already in fixture.json. The signs by mark detail is accurate, I verified it against run 2's own image transcription, happy to add it as color if wanted, though nothing currently claims otherwise.

Point 2 is where I need you two to align rather than me picking a side. florencemashipei already reviewed and approved the current state, f5 required true, no committed run satisfying it yet, stated plainly in the README rather than smoothed over. She explicitly called the required versus bonus question a scope decision, not a genealogical one, and said she was not going to make that call for the PR author.

johnmarkpeterbrown is now asking for the opposite, f5 required false, so validity holds on f1 through f4 without owing another paid run, with Closes changed to Refs as the consequence.

Both are defensible design positions and I do not think I should implement one over the other without you two agreeing on which. Could you two settle between yourselves which way this goes, required and closes, or bonus and refs, and I will apply whichever you land on.

florencemashipei:
Two corrections before this goes further.

**I haven't approved.** My review is `COMMENTED` and the PR is at 0 of 2 — I held the approval deliberately, pending `marriage_place`.

**And I've taken no position on f5 required vs bonus.** `required: true` was your choice in `038eb998`, before my review. Mine addressed one thing: the grading inconsistency on run 2's f5. My follow-up then flagged what the regrade does to §14 and handed the closure question to @johnmarkpeterbrown as a scope call. That's surfacing the problem `required: true` creates, not endorsing it.

**So there's nothing for the two of us to settle** — there's one instruction on the table and no competing one.

**Go with f5 `required: false` and `Refs #1413`.** His reasoning holds: with f5 non-required, run 1's required set is f1–f4, all `true`, so `derive_verdict` gives `pass` with `recall_required` 1.0 — §14 validity stands and nobody owes another $10–16 run — while f5 still measures the inference as a bonus finding.

Nothing from my review is lost either way. Run 2's f5 stays `false`, and f5's tightened text — dated `Marriage` fact, on the mother's own node, graded on f3's standard — stands as written. The `required` flag only controls whether that `false` drags the rollup down.

On "signs by mark": no need to add it. Nothing in the fixture claims otherwise, and it was jmpb showing he'd read the image rather than the index.

Ping me when the two lines are in and I'll approve.

EdmondOware:
f5 is now required false, a bonus finding, and Closes #1413 is changed to Refs #1413, in commit 5699d318. Nothing else about f5 moved, its text, the dated fact and correct person bar, and run 2's false grade are all exactly as they were. README and fixture.json updated to describe it as a bonus finding throughout, with a changelog entry recording the resolution and why. Gates and calibrate_judge dry run re-run clean.

@johnmarkpeterbrown the three items from your review are in. @florencemashipei the two lines are in as you asked. Ready for another look whenever you both have a moment.

johnmarkpeterbrown (approved):
Looks good!

florencemashipei (approved):
Approving. Everything from both review rounds is in.

`5699d318` does exactly what was asked and nothing more: `f5.required` to `false`, `Refs #1413`, and — not asked for by either of us — the README and `fixture.json` `notes` recast to describe f5 as a bonus finding throughout, so neither one is left claiming it as a required discriminator. `eval/runlogs/` is untouched since `71e7cb97`, so run 2's f5 regrade stands and no log was rewritten.

§14 now reads cleanly rather than with an asterisk: run 1's required set is f1–f4, all `true`, `verdict: pass` / `recall_required: 1.0`, and the stepfather inference is measured as a bonus with the demonstration tracked via `Refs #1413`.

For the record, the two things I raised and where they ended up:

- **Run 2's f5 grading inconsistency** — regraded `false`, and f5's own text tightened so the bar (dated `Marriage` fact, on the mother's own node, graded on f3's standard) is stated rather than inferred a second time. That was the better of the two options available and it is what stops this recurring on the next run.
- **f5's `marriage_place`** — corrected to `"Amite County, Mississippi"`, matching f3 and the cited source.

Good work on both rounds, and on verifying each finding against the tree yourself before acting on it rather than taking a reviewer's word for it.

PR #1726, opened 2026-08-18
feat: suite-level abort-storm breaker (#1600)

### LINKED ISSUE #1600: Unit harness pays for a whole environment outage test-by-test — no suite-level bail-out on an abort storm
**Touches:** eval/harness/harness/orchestrator.py, eval/CLAUDE.md

**Filed from triage, not deferred from a PR.** This came out of a standup blocker, not out of work in progress — there is no open PR it could have been folded into, and nobody has these files open.

## What happens

When the Anthropic API degrades, the unit harness retries each aborted test up to 3 times and runs the suite to completion, paying full price for a run that produces no gradeable result. There is no suite-level circuit breaker.

- `RETRYABLE_ABORT_REASONS = {"error", "sdk_stream_silence"}` — `eval/harness/harness/orchestrator.py:669`
- `DEFAULT_SKILL_RUN_ATTEMPTS` governs per-test retries — `eval/harness/harness/orchestrator.py:618`
- The rule already exists **in prose and nowhere else**, `eval/CLAUDE.md:144`:
  > A suite where nearly every test aborted with the same reason is an environment failure, not twenty skill regressions; re-run it before reading anything into the results.

Nothing enforces that sentence. The harness cannot tell a global outage from twenty independent flakes, so it buys the same information twenty times.

## Measured, 2026-08-13

Reported by Collins (`Cia-3`) while trying to finalise PR #1582s regression tests:

| Run | Aborted | Note |
|---|---|---|
| ~09:19, 21 tests | **5 / 21** | `sdk_stream_silence`, many retries also stalled |
| ~09:58, full suite | **6 / 21** | aborted before the run started properly; a test that had passed cleanly minutes earlier in isolation aborted too |
| baseline, same day | 1 / 43 and 1 / 44 | normal rate |

Cost roughly $4–6 per affected run, twice, for nothing usable. The aborts hit previously-stable tests across the board — the exact signature `eval/CLAUDE.md` describes.

## What to build

> **Reviewed 2026-08-14 before junior handoff — `ready-after-edit`.** This
> section replaces the original. **The discriminator the issue proposed was
> wrong**: keying on the whole exit-3 family would have killed a healthy
> committed `project-status` run at 3/11 cap aborts — a *higher* rate than the
> 5/21 storm this issue is built on. Narrowed to the two reasons the harness
> already calls transient, the committed corpus settles the threshold, so this
> needs no ruling from the lead.

A suite-level bail-out in `eval/harness/run_tests.py`'s drain loop: when transient
aborts pass the threshold below, stop submitting new tests, keep everything that
finished, and exit without minting a releasable candidate.

### Which aborts count — the transient pair only, NOT the exit-3 family

Count only `aborted_reason` in `RETRYABLE_ABORT_REASONS` — `error` and
`sdk_stream_silence` (`orchestrator.py:669`). That set is already the harness's
own definition of "environmental, worth another attempt". The caps (`max_turns`,
`max_wall_clock_seconds`, `max_tool_calls`, `max_input_tokens_per_turn`) are
deliberately not retried because they are deterministic per-test budget
statements (`orchestrator.py:658-661`), and they must not arm the breaker.

Measured 2026-08-14 by tallying `tests[].runs[0].aborted_reason` over
`eval/runlogs/unit/*/*.json`: 1840 tests, 20 aborts — `max_turns` 9,
`max_wall_clock_seconds` 8, `error` 2, `sdk_stream_silence` 1. Cap aborts cluster
inside a single skill: `project-status/v1_2026-08-11_19-00-27.json` is 3 of 11
`max_wall_clock_seconds` (27%), higher than the 5/21 (24%) outage rate, and
`person-evidence`'s `max_turns` cluster is a known test-authoring flake
(issue #1334, open). Restricted to the transient pair, the same corpus has **3
transient aborts in 1840 tests**, worst single log 2 of 16
(`research-plan/v1_2026-08-11_13-21-10.json`). Re-run the tally before changing
the constants; it moves as run logs are pruned and added.

### Threshold

Evaluate after each test completes, in the drain loop where `_flush_partials()`
is already called. Trip when **both** hold:

- at least **4** completed tests aborted with a transient reason, **and**
- those aborts are at least **20%** of the tests completed so far.

Both arms are load-bearing: the count floor stops a 2-test suite tripping at 1/2;
the ratio stops a 100-test suite tripping on 4 scattered flakes. Against the
numbers above this never fires on any committed run log and fires part-way
through both 2026-08-13 storms. Put the two constants at module scope with those
corpus figures in a comment beside them, and add a `--no-abort-breaker` flag so
an operator who knows the environment is healthy can finish the run.

### It must not mint a releasable candidate

Setting `stop_submitting = True` alone is wrong **and will look right**. The
budget-cap path already does exactly that (`_budget_blocks_next`,
`run_tests.py:815-842`) and then falls through to the normal end-of-run block,
which writes a releasable `v{N}_<ts>.json` — so a bailed-out run would commit a
15-of-21-test candidate that `check_runlogs.py` then treats as the skill's latest
run log. Do not change the budget-cap path here. Give the breaker its own flag and
add it to the cut-short condition at `run_tests.py:1029`
(`if interrupted or harness_error is not None:`), which promotes each
`.partial_<ts>.json` to a gitignored `scratch_<ts>.json` and returns before any
`v{N}` is written.

### Preserving partial results — the crash branch, not the Ctrl-C branch

`_flush_partials()` already writes every finished test to `.partial_<ts>.json`
after each completion (`run_tests.py:785-813`), so nothing new is needed to save
the work. Follow the harness-error shape: stop submitting and let the drain loop
keep harvesting, so in-flight tests finish and are saved. Do **not** copy the
Ctrl-C branch's `ex.shutdown(wait=False, cancel_futures=True)`
(`run_tests.py:948`) — it cannot cancel a running future, so the suite waits for
those tests anyway and then discards their results, the most expensive possible
outcome.

**Exit code: 3.** A bail-out is by construction a set of execution aborts, which
is what 3 already means, so no exit-code contract changes. Print the count and the
reason, and name the scratch path.

**Do not add a field to the run-log envelope** to mark the run incomplete. That is
a schema change across both `run-log.schema.json` mirrors, and it is unnecessary:
a bailed-out run writes a scratch log, which is already non-releasable and
gitignored.

## Sites this touches

- [ ] `eval/harness/run_tests.py` — the counters and trip check in the drain loop,
      the new flag in the cut-short condition at line 1029, the
      `--no-abort-breaker` argument, and the **exit-code docstring at the top of
      the file** (lines 8-22)
- [ ] `eval/README.md` — the exit-code table, if the bail-out changes what a
      reader should take 3 to mean
- [ ] `eval/CLAUDE.md:144` — the "re-run it before reading anything into the
      results" sentence now describes something the harness does; say so in one
      clause
- [ ] `eval/harness/tests/unit/test_cli.py` — the two tests below
- [ ] Nothing else. A change confined to `eval/harness/` marks **no skill touched**
      in `check_runlogs.py` (its skill regex at `:610` keys on
      `packages/engine/plugin/skills/` and `eval/tests/unit/`), so this carries
      **no eval re-run and no annotation cost**

## Not this issue

Issue #1484 (open) covers the *reporting* half — the corpus report undercounting guardrail violations and hiding abort-path e2e spend. This one is about **stopping the spend**, not counting it. They are separate and neither blocks the other.

## Verifies the change

`make harness-test`. Mirror
`test_harness_error_keeps_completed_tests_as_scratch_and_exits_1`
(`eval/harness/tests/unit/test_cli.py:1487`) — it already stubs `run_one_test` and
`write_partial_runlog`, and `_stub_log(test_id, skill, "aborted",
aborted_reason=...)` (`test_cli.py:116`) takes the reason directly. Two tests, and
the lint-must-fail rule applies — watch each fail before the fix:

1. **It trips.** Feed enough `sdk_stream_silence` aborts to cross the threshold and
   assert: the run stopped before the last test; one `scratch_*.json` exists
   holding only the completed tests; **no `v*.json` was written**; the reason and
   count are on stdout; exit code 3.
2. **It does not trip.** A suite with one transient abort and one cap abort runs to
   completion and writes its releasable `v{N}` as before.

"No `v*.json`" is the assertion that catches the likeliest wrong implementation.

## Do after PR #1612

PR #1612 (open, `judge-cost-cached-token-sign-1534`) rewrites `_flush_partials` in
place — adding `nonlocal harness_error, stop_submitting` and a per-skill
try/except — the same function and lines the breaker touches. Start after it
merges; its shape (flip `stop_submitting`, record a reason, fall through to the
existing promote-and-return block) is also the pattern to copy.

## Do not re-derive

- `sdk_stream_silence` is raised at `eval/harness/harness/skill_runner.py:461` and treated as transient by the orchestrator; the e2e side has its own at `eval/harness/e2e/orchestrator.py:1636` and `stop_checker.py:106`. **Unit harness only for this issue** — leave e2e alone unless the same shape falls out for free.
- Whether the API was actually degraded on 2026-08-13 is not worth chasing. The point is that the harness spends the same either way.

yinkid28:
## Summary

- Adds a circuit breaker to `run_tests.py` that stops submitting new tests when transient aborts (`error`, `sdk_stream_silence`) cross both a count floor (4) and a ratio threshold (20% of completed tests). Prevents burning full-suite cost on a degraded API — the 2026-08-13 incident burned ~$4–6 twice.
- Tripped runs write a `scratch_*.json` (gitignored, non-releasable) instead of a `v{N}` candidate, and exit code 3.
- `--no-abort-breaker` flag overrides when the environment is known healthy.

Closes #1600. Depends on #1612 (merged).

## Test plan

- [x] `test_abort_storm_breaker_trips_and_writes_scratch_not_release` — 10-test suite with 7 transient aborts; verifies breaker fires, scratch written, no `v{N}`, stops early
- [x] `test_abort_storm_breaker_does_not_trip_on_cap_aborts` — 5-test suite with 1 transient + 1 cap abort; verifies breaker stays off, `v{N}` written normally
- [x] Full `test_cli.py` suite: 66/66 pass


francis-2008-happy (approved):
Approving. Verified by running: `test_cli.py` 66/66, and I mutation-checked the breaker — disabling it makes the trip test fail (all tests run, no early stop), so the test genuinely binds. Confirmed the mechanics too: `done_n` is counted before the ratio check (no divide-by-zero), `stop_submitting` really halts new submissions, and a trip writes a scratch log + exits 3 with no `v{N}`.

One optional, non-blocking note: `_TRANSIENT_ABORT_REASONS` duplicates `RETRYABLE_ABORT_REASONS` (they match today at orchestrator.py:695), and your own comment notes nothing enforces that. Since there are now two consumers, importing the existing constant instead of re-declaring would kill the drift risk — but that's your call, not a blocker.

EdmondOware:
Small nit, not blocking. The comment on test abort storm breaker trips and writes scratch not release says the breaker fires after 8 completions, 3 pass plus 5 aborted. I instrumented the test and ran it directly, it actually fires after 7 completions, 3 pass plus 4 aborted, since the count floor of 4 is what gates the trip and that floor is reached on the 4th abort, the 7th test overall. Doesn't affect correctness, the assertions are appropriately loose and never hardcode the number, so nothing here is actually wrong, just the illustrative math in the comment.

Otherwise this looks solid to me. Verified the retryable abort reasons duplication claim directly against orchestrator.py, they match today, and confirmed by removing the breaker logic mentally that the tests are not vacuous, several assertions would fail without it.

jud-sdev (approved):
Verified by running, not just reading: test_cli.py 66/66, and I mutation-checked the breaker myself — setting the count floor to 1000 so it never trips makes the trip test fail, so the test genuinely binds. Confirmed the mechanics too: done_n is counted before the ratio check so there is no divide-by-zero, the count floor gates the division, stop_submitting really halts new submissions, and a trip writes a scratch log and exits 3 with no v{N}. _TRANSIENT_ABORT_REASONS matches the orchestrator's RETRYABLE_ABORT_REASONS today.

The design is sound: both arms have to hold, so a couple of scattered flakes in a big suite will not trip it, and cap aborts are correctly excluded.

One non-blocking note, same as francis's: the duplicated _TRANSIENT_ABORT_REASONS has nothing enforcing that it stays in sync with the orchestrator's set, so a future divergence would quietly weaken the breaker. Your comment already flags it. One small correction on the suggested fix though — you cannot just import the orchestrator constant, since it is function-local, which is exactly why you re-declared it. The real fix is to promote it to module scope and import it, and deferring that until a third consumer appears is a reasonable call at two.

Good work.

PR #1733, opened 2026-08-18
record_search: fix three placeParts qualifier-stripping defects (#1584)

### LINKED ISSUE #1584: placeParts: three qualifier-stripping defects in one expression — non-ASCII `co`, internal-qualifier double space, and the `CO` state abbreviation
> **Reviewed 2026-08-12 before junior handoff.** No decision needed; the fix this body
> originally suggested was incomplete and has been corrected, and the verification the
> spec mandates is now written out.

> **Issue #1426 merged in here, 2026-08-14.** That issue held the third `placeParts`
> defect — `CO`, the state abbreviation, being eaten as a County qualifier. It was split
> out of this one on 2026-08-12 by `/fill-ready` for one reason: it was blocked on a
> disambiguation choice nobody had made. **The lead ruled it on 2026-08-13** and
> `needs-decision` came off, so the split's only justification is gone. Both bodies
> already said to do all three in one PR once the fork was answered. All three are now
> below, and all three are the same expression on one line.

**Touches:** packages/engine/mcp-server/src/utils/marriage-jurisdictions.ts, packages/engine/mcp-server/tests/utils/marriage-jurisdictions.test.ts, docs/specs/record-search-tool-spec-v2.md

All three cases are **pre-existing and unchanged by the `\bco\.?\b` -> `\bco\b\.?`
fix** (issue #1267).

## Read this first: the three defects are one line, and their fixes fight each other

```
100 function placeParts(place: string): string[] {
102     .toLowerCase()
105     .map((part) => part.replace(/\bcounty\b|\bco\b\.?/g, "").trim())
```

- Case 1 (below) rewrites the **regex** on line 105.
- Case 2 requires the `\s+` collapse to run **after** the strip — a `.map()` swap.
- Case 3 requires the strip to run **above** the `.toLowerCase()` on line 102, because
  the rule depends on a case signal that line destroys.

Do them in one pass or you will write the regex twice. In particular: **once the strip
moves above `.toLowerCase()`, the pattern is matching mixed-case input**, so the
`county`/`co` alternatives need an `i` flag or explicit case handling that the current
pattern does not need and the case-3 rule must not be given blindly — see case 3.

## 1. `\b` is ASCII-only, so non-ASCII places starting `co` lose their first two letters

```
placeParts("Coïmbra, Portugal")           -> ["ïmbra", "portugal"]
placeParts("Coévrons, France")            -> ["évrons", "france"]
placeParts("Co-operative Township, Ohio") -> ["-operative township", "ohio"]
```

JavaScript's `\b` is defined over `[A-Za-z0-9_]`, so `ï`, `é` and `-` all count as
boundaries and `co` matches as a whole word inside them. The mangled token flows into
`placeTokens` -> `samePlace` and the dedupe key, so two genuinely different places can
compare equal or fail to.

Not currently exercised: no corpus place hits this. It matters for non-US research,
which is in scope for this product.

**Fix.** Widen the boundary class to letters, digits, `_` **and `-`**, with the `u`
flag — not `\p{L}` alone. `\p{L}` alone fixes `Coïmbra` and `Coévrons` but leaves
`Co-operative Township` broken, because `-` is not a letter either and `co` still
matches as a whole word before it:

```js
.replace(/(?<![\p{L}\p{N}_-])county(?![\p{L}\p{N}_-])|(?<![\p{L}\p{N}_-])co(?![\p{L}\p{N}_-])\.?/gu, "")
```

Apply the same class to the `county` alternative, not just `co` — leaving one
ASCII-only is the kind of asymmetry the next reader has to re-derive. The lookahead
sits *before* `\.?` so `"Hill Co., Texas"` still strips to `["hill","texas"]`;
`ES2022` in `packages/engine/mcp-server/tsconfig.json` covers both lookbehind and
`\p{…}`.

## 2. Whitespace is normalized *before* the strip, so an internal qualifier leaves a double space

```
placeParts("Hill Co. Texas")     -> ["hill  texas"]     // two spaces
placeParts("Hill County Texas")  -> ["hill  texas"]     // identical
placeParts("Hill, Texas")        -> ["hill", "texas"]
```

`placeParts` runs `.replace(/\s+/g, " ")` in the first `.map()` and strips the qualifier
in the second, so a qualifier with text on both sides of it inside one comma-part leaves
the two spaces that used to flank it. The token never matches its comma-separated
spelling, so `samePlace` returns false and the dedupe key splits one jurisdiction into
two.

The fix is small — swap the two `.map()` steps so `\s+` collapses *after* the qualifier
is removed — and it changes no other listed case. One corpus instance
(`"Leesville Cemetery Leesville Gonzales County Texas"`), and it sits in
`eval/tests/e2e/reese-children/unstripped-tree.gedcomx.json`, which the spec explicitly
excludes from the criterion as an authoring artifact — so nothing is failing today and
it does not enter the count below.

## 3. `CO`, the state abbreviation, is eaten as a County qualifier (was issue #1426)

```
placeParts("Denver, CO")        -> ["denver"]
placeParts("Denver, Colorado")  -> ["denver", "colorado"]
```

`samePlace` is a subset test, so:

```
samePlace("Denver, Iowa", "Denver, CO")  ->  true
```

A marriage search scoped `"Denver, CO"` that comes back empty therefore **suppresses a
tree place of `"Denver, Iowa"`** from its own candidate list. That is precisely the
failure `marriageJurisdictionCandidates` exists to prevent: the alternative jurisdiction
most worth trying is deleted before the caller ever sees it. Any US place written with a
postal abbreviation hits this, and the model does write postal abbreviations.

Found while reviewing the fix for issue #1267.

### The ruling (lead, 2026-08-13) — options 1 and 2 combined, plus case

Strip the `County`/`Co` qualifier when **any** of these holds:

1. it carries a dot — `Co.`, `co.`; or
2. another token follows it in the same comma part; or
3. it is spelled with a **lowercase `o`** — `Co` or `co`.

Keep it only when the token is **`CO`, both letters uppercase, with no dot and nothing
following it in its comma part.** That is the postal abbreviation and nothing else.

| input | qualifier? | tokens |
|---|---|---|
| `"Denver, CO"` | no — uppercase, no dot, part ends | `denver`, `co` |
| `"Hill Co., Texas"` | yes — dot | `hill`, `texas` |
| `"Hill Co Texas"` | yes — token follows | `hill`, `texas` |
| `"Hill Co"` | yes — lowercase `o` | `hill` |

None of the three options works alone: requiring the dot breaks `"Hill Co Texas"`, the
token-follows test alone breaks `"Hill Co., Texas"` (the dot ends its comma part), and a
postal-abbreviation set is fifty entries to maintain against a case signal that is
already in the string.

**Record the chosen rule and the three options it beat** in the spec's `Co.` qualifier
paragraph.

### The implementation trap

`placeParts` lowercases the whole string **before** the qualifier regex runs (`:102`,
then `:105`). By the time `/\bcounty\b|\bco\b\.?/` sees the token, `CO`, `Co` and `co`
are all `co` — the distinction the rule rests on no longer exists.

**So the qualifier strip has to move above the `.toLowerCase()`, or the case has to be
captured before it.** The regex also has no `i` flag today, which reads as
case-sensitive and is not, because the input is already lowered. Whoever takes this
should expect that to mislead them.

### Known cost, accepted

`"Denver, Co"` — Colorado written sloppily with a lowercase `o` — still strips, so it
still compares equal to `"Denver, Iowa"`. That is the deliberate price of catching
`"Hill Co"`, which has no other tell. Do not "fix" it later without reopening this
ruling.

## How to verify

1. `make engine-test` — 25 tests in `tests/utils/marriage-jurisdictions.test.ts` pass on
   `main` today (measured 2026-08-12, `npx vitest run
   tests/utils/marriage-jurisdictions.test.ts`); every one must still pass. In
   particular `"Hill Co., Texas"`, `"Co., USA"` and `"County, United States"` keep their
   current tokenization under all three fixes.
2. The corpus criterion the spec mandates. It is **not** a repo-wide grep — the spec is
   explicit that a corpus-wide diff is the wrong test and would refuse a correct
   exemption. Save this to a scratch file and run it from the repo root **with your own
   regex substituted into `NEW`**; both classes must report `0 changed`:

```js
import fs from "node:fs"; import path from "node:path";
const OLD = () => /\bcounty\b|\bco\b\.?/g;
const NEW = () => /(?<![\p{L}\p{N}_-])county(?![\p{L}\p{N}_-])|(?<![\p{L}\p{N}_-])co(?![\p{L}\p{N}_-])\.?/gu;
const before = s => s.toLowerCase().split(",").map(p => p.trim().replace(/\s+/g," ")).map(p => p.replace(OLD(),"").trim()).filter(Boolean);
const after  = s => s.toLowerCase().split(",").map(p => p.replace(NEW(),"")).map(p => p.trim().replace(/\s+/g," ")).filter(Boolean);
const files = [], walk = d => { for (const e of fs.readdirSync(d,{withFileTypes:true})) { const p = path.join(d,e.name); e.isDirectory() ? walk(p) : files.push(p); } };
walk("eval/tests/e2e"); walk("eval/runlogs");
const cls1 = new Set(), cls2 = new Set();
for (const f of files) {
  const isTree = f.endsWith("starting-tree.gedcomx.json") || f.endsWith(".final-tree.gedcomx.json");
  const isRunlog = f.startsWith("eval/runlogs/e2e/");
  if (!isTree && !isRunlog) continue;
  const t = fs.readFileSync(f, "utf8");
  if (isTree)   for (const m of t.matchAll(/"(place|standard_place)":\s*"((?:[^"\\]|\\.)*)"/g)) cls1.add(JSON.parse('"'+m[2]+'"'));
  if (isRunlog) for (const m of t.matchAll(/"(marriagePlace|recordSubdivision)":\s*"((?:[^"\\]|\\.)*)"/g)) cls2.add(JSON.parse('"'+m[2]+'"'));
}
for (const [name, S] of [["class 1 tree-fact places", cls1], ["class 2 recorded search args", cls2]]) {
  const moved = [...S].filter(s => JSON.stringify(before(s)) !== JSON.stringify(after(s)));
  console.log(`${name}: ${S.size} distinct, ${moved.length} changed`);
  for (const s of moved) console.log("   ", JSON.stringify(s), before(s), "->", after(s));
}
```

   With the case-1 regex above and the case-2 `.map()` swap, this reports **class 1: 1628
   distinct, 0 changed; class 2: 54 distinct, 0 changed** (measured 2026-08-12, the
   script as written, over 104 `starting-tree.gedcomx.json` + 148
   `*.final-tree.gedcomx.json`). Any newly committed e2e run or fixture tree moves the
   1628; the two `0`s are the part that decides the exemption. A non-zero count means the
   exemption does **not** apply and the change needs a live `make e2e-run` before it
   lands.

   **The script's `after()` lowercases before stripping, which is the pre-case-3 order.**
   Once the strip moves above `.toLowerCase()` for case 3, update `after()` to match the
   real pipeline before trusting its counts — otherwise it measures an order the code no
   longer uses. Case 3 is expected to move class 1 (`"Denver, CO"`-shaped places stop
   collapsing), so a non-zero count there is the ruling working, not a regression: report
   the moved strings and check each one is a postal abbreviation, rather than assuming
   the exemption still applies.

Nothing automated enforces step 2 — no lint runs the criterion, and CI is green either
way. Paste the script's output lines into the PR body.

## Done when

- All three cases tokenize correctly, with a test for each alongside the existing `Co.`
  cases in `marriage-jurisdictions.test.ts`. Cover all four printed case-1/2 inputs:
  `"Coïmbra, Portugal"`, `"Coévrons, France"`, `"Co-operative Township, Ohio"`, and
  `"Hill Co. Texas"` vs `"Hill County Texas"` vs `"Hill, Texas"` reducing to the same
  tokens — plus the four rows of the case-3 ruling table, and
  `samePlace("Denver, Iowa", "Denver, CO") === false`.
- The doc comment above `placeParts` in `marriage-jurisdictions.ts` and the `\bco\b\.?`
  note in the test file no longer describe a regex the code does not contain. Keep the
  existing explanation of why the dot is consumed *after* the boundary assertion — that
  reason is unchanged — and add one clause for why the boundary class includes `-` and
  non-ASCII letters, naming `Co-operative` and `Coïmbra`, plus one for why the strip runs
  before `.toLowerCase()`.
- `docs/specs/record-search-tool-spec-v2.md`, "Place matching", two edits:
  1. The `Co.` qualifier paragraph states the pattern literally
     (`/\bcounty\b|\bco\b\.?/`). Update it to the new pattern, add the hyphen/non-ASCII
     reason, **and record the case-3 ruling and the three options it beat.**
  2. The "Live-run exemption for tokenizer-only changes" block says **"Applied once"**
     and carries that run's counts. Record this as the second application, with the
     counts your criterion run printed and the date.

No eval cost: no `SKILL.md`, rubric, or unit test is touched, so `check_runlogs.py` does
not fire and no paid `make eval-skill` run is needed.

MMagnum:
Closes #1584.

## Summary

- **Normal tier.** Fixes all three `placeParts` qualifier-stripping defects in one
  pass, because each constrains a different stage of the same expression and doing
  them separately means writing the regex three times.
- **Defect 1 — the boundary was ASCII-only.** `\b` is defined over `[A-Za-z0-9_]`, so
  a non-ASCII letter or a hyphen reads as a word edge and a leading `co` matches as a
  whole word *inside the place's own name*: `"Coïmbra, Portugal"` tokenized as
  `["ïmbra","portugal"]`, `"Co-operative Township"` lost its first two letters. Now a
  `[\p{L}\p{N}_-]` lookaround pair with the `u` flag.
- **Defect 2 — an internal qualifier fused two locality levels into one token.**
  `"Hill Co. Texas"` became `["hill texas"]`, which no comma-separated spelling can
  equal, so `samePlace` never matched `"Hill, Texas, United States"`. The qualifier is
  now consumed **as a part separator**, so all three spellings reduce to
  `["hill","texas"]`.
- **Defect 3 — `CO` eaten as a County qualifier.** `"Denver, CO"` reduced to
  `["denver"]`, and `samePlace` is a subset test, so a Colorado-scoped search that came
  back empty **suppressed a tree place of `"Denver, Iowa"` from its own candidate
  list** — deleting exactly the alternative jurisdiction this hint exists to surface.
  Implements the lead's ruling of 2026-08-13.

## Review guidance

**Start here:** `stripCountyQualifier` in
`packages/engine/mcp-server/src/utils/marriage-jurisdictions.ts` — the only
non-mechanical thing in the diff. Two things worth checking: (a) the callback
implements the ruling as three strip-conditions with `CO` surviving only as bare
uppercase at the end of its comma part, and (b) the `i` flag makes the pattern *read*
case-insensitive while the decision is case-**sensitive** (`word !== "CO"`) — matching
is deliberately case-blind so the callback can inspect the spelling it captured. That
is why the strip had to move above `.toLowerCase()`, which destroys the signal.

**Unsure about:** nothing outstanding. The `export` on `placeParts` was the open
question and both reviewers said keep it, so it stays; the original reasoning follows. Defect 1's mangling is
*symmetric* — `"Coïmbra"` reduced to `["ïmbra"]` whether it arrived as the searched
place or as the candidate — so the exclusion still matched and **every** behavioural
assertion I tried passed with the defect present (I tried the exclusion path, the
dedupe key, and `isSubCountryPlace`). Tokens are the only level at which it is
observable. If you want the export gone, defect 1 loses its only real test; I'd rather
be told than guess.

## Plan

**Didn't change:** the whitespace-collapse-only fix the issue prescribes for defect 2
(see below), and nothing outside this one tokenizer — `placeTokens`, `samePlace` and
`isSubCountryPlace` all call through it unchanged.

**Acceptance check:** 9 new tests in `tests/utils/marriage-jurisdictions.test.ts`, each
verified to fail against the old tokenizer before landing (reverted the body, kept the
tests, watched them go red: 5 failed for defects 1–2, then 2 more for defect 3). The
25 pre-existing tests pass untouched. `make engine-test`: typecheck clean, 100 files /
2383 tests.

**Deviated from the plan:** **yes, three times, all deliberate.**

1. **The regex is factored** to one shared boundary pair `(?:county|co)` with `\.?`
   outside the closing assertion, rather than the issue's two fully-spelled
   alternatives. Shorter, and strictly better: it also strips the dot from
   `"County."`, which the proposed form left as a stray `"."` token — the same defect
   class #1267 fixed for the abbreviated spelling. Note the dot **must** stay outside
   the assertion; moving it inside reintroduces #1267 (`"Hill Co.Texas"` →
   `["hill .texas"]`).
2. **Defect 2 is a separator split, not the prescribed `.map()` swap.** The swap is
   cosmetic — measured both ways, it turns two spaces into one but changes **no**
   `samePlace` outcome and **no** dedupe grouping (the two spaced spellings already
   produced an identical key), so it would not deliver this issue's own "Done when" of
   the three spellings reducing to the same tokens. Corpus criterion is 0/0 either
   way, so the deviation costs nothing.
3. **The issue's `samePlace("Denver, Iowa", "Denver, CO") === false` assertion cannot
   be written** — `samePlace` is not exported. Asserted through
   `marriageJurisdictionCandidates` instead, which covers the same logic *and* the
   real harm (Iowa staying in the candidate list). Better test than the one asked for.

## Since review — two code fixes, and the criterion itself was wrong

Both reviews turned what the first read as documentation-only costs into one-line code
fixes. I reproduced both against the real module before changing anything.

1. **A `CO`-scoped search resurfaced its own spelled-out state.** `["denver","co"]` is
   not a subset of `["denver","colorado"]`, so searching `"Denver, CO"` offered
   `"Denver, Colorado, United States"` back as a candidate. `placeTokens` now compares a
   surviving `co` as `colorado` — one mapping, not the fifty-entry set the ruling
   rejected, because by the time a bare `co` survives tokenization the ruling has already
   decided it is Colorado. `placeParts` is untouched.
2. **A kept qualifier fused into the token before it.** `"Hill CO, Texas"` tokenized as
   `["hill co","texas"]`. Wrong under both readings, and the internal-qualifier defect
   surviving in the one branch that had stopped honouring the separator rule. The kept
   match is now wrapped in separators.

A test each. **36 tests on the two files, 103 files / 2454 on the full engine suite.**

**The criterion was over-reporting, and that is what put a live run on the critical
path.** It briefly showed class 1 as `1771 distinct, 2 changed`. Both rows were spurious:
the script collected raw `place` even where a `standard_place` shadowed it, so it
measured strings the tokenizer never reads. Both came from `stribling-father-1821`, whose
fact carries `standard_place: "Graham, Young, Texas, United States"` — no `County`
token, does not move. The spec already stated that rule; the pasted script contradicted
it, so it had been over-reporting since it was written.

Fixed to collect `standard_place || place` per fact, and **committed as
`dev/probe-placeparts-criterion.ts`** instead of living in an issue body, which is how it
drifted. It carries a self-check — three known-moving canaries must still report as
moved or it exits non-zero, guarding the one way it can silently say "exempt" having
measured nothing. Proven to fail by making `before()` delegate to `placeParts` (`0/3`,
exit 1).

**Corpus criterion, current** (spec § "Live-run exemption for tokenizer-only
changes"): **class 1 — 1007 distinct tree-fact places, 0 changed; class 2 — 61
distinct recorded search arguments, 0 changed**. Exempt, so no live `make e2e-run` is
owed — confirmed by @DallanQ, who also asked for the criterion fix above. Re-measured
after each merge of `main` and stable at 1007/0 across three.

Deliberately not quoted: a count of files scanned. Three reviewers produced three
different numbers for it and it decides nothing.

The postal-abbreviation change moves nothing in either class, because neither holds a
`CO`: of class 1's 12 places whose last comma-part is two letters, the 9 uppercase ones
are `UT`, `NY`, `FL`, `MT` and `LA`, and the rest are `Ny`, `Ca` and `Ma`; class 2 has
none. Neither class holds a place with a standalone `co` token — the 9 such places sit
outside both classes, and all 9 still strip, including `"Co Kilkenny"` and `"Co Down"`.

## Test plan

- [ ] I ran `make test-all` — **it fails on this Windows machine for two pre-existing
      reasons, neither related to this PR.** Both reproduce on bare `origin/main`
      (`f2680cc8`) with none of this branch's code:
      1. `@genealogy/electron` → `src/main/__tests__/feedback.test.ts` builds
         `.claude/projects/-C:\Users\...` and `mkdir` fails `ENOENT`, because a colon
         is illegal in a Windows path — the test assumes a POSIX project path.
         Verified failing on `origin/main` in a detached checkout.
      2. `server-test` → `uv trampoline failed to canonicalize script path`.
      The suites this PR actually touches are green: `make engine-test` (typecheck +
      100 files / 2383 tests) and the Python suite inside `test-all` itself (2249
      passed, 3 skipped). Leaning on CI for the rest, which runs the same suites on
      Linux.

- [x] **N/A — no skill run-log snapshot changed.** This PR touches no file under
      `packages/engine/plugin/skills/`, no agent, no `eval/tests/unit/`, and no
      scenario or fixture any skill references. `make engine-test` reports "no
      prompt-size changes in this changeset". No `make eval-skill` run is owed.

- [x] **N/A — no skill `description` frontmatter or DO NOT clause changed.** No routing
      pair touched, so no negative test is owed on either side.

## Follow-on issues

**Folded in / filed instead:** Nothing filed. The two Windows `test-all` failures above
are pre-existing and outside this PR's scope; flagging them here rather than filing,
since neither has a decision behind it yet and I'd rather the lead say whether they
want tickets.


chesworthrm (commented):
Reviewed the full diff against the merge with `main`, not the branch head. The code is right — both findings are in the spec, and neither is a code change.

**What it does:** `placeParts` now consumes the `County`/`Co` qualifier as a part separator instead of deleting it, uses a `[\p{L}\p{N}_-]` lookaround pair instead of `\b`, and runs `stripCountyQualifier` before `.toLowerCase()` so bare uppercase `CO` survives as a token. That matches issue #1584 and all three of its defects. Nothing outside the tokenizer moved.

**Checks:** all required checks pass on this head — `pytest`, `runlogs`, `e2e-fixtures`, `vitest`, `lockfile-drift`, `scan`. None missing.

**Verification I re-ran, so you don't have to defend these:**

- Test-merged `origin/main` (`6e2671df`). Engine suite, typecheck, and the harness suite are all green on the *merged* tree.
- Reverted the tokenizer to `f2680cc8` and kept your tests. 7 of the 9 go red, one per defect. The 2 that stay green are your deliberate regression pins. The suite is not vacuous.
- Re-ran the corpus criterion independently, importing the exported `placeParts`. Class 1 — 1712 distinct, 0 changed. Class 2 — 60 distinct, 0 changed. Reproduces your numbers exactly, and it does report the synthetic `CO` cases as moved. The exemption holds; no live run is owed.

**Still needs:** one approval from `senior-developers` **or** `senior-genealogists` (the `*.ts` rule — the owners are an OR), plus two approvals in total. The spec file matches no CODEOWNERS rule, so it pulls nobody on its own. I have not approved.

---

### 1. `docs/specs/record-search-tool-spec-v2.md` — the supporting counts are not from the criterion's two classes

The two `0`s reproduce exactly. The `11` and the `7` do not.

Class 1 holds **zero** places with a standalone `co` token. `"Co Kilkenny"` and `"Co Down"` live in `eval/runlogs/e2e/butler-ancestry/run-2026-07-22_07-29-06.final-research.json`, which is neither class. The `11` is class 1's count of two-letter trailing tokens counted any-case — only 8 are uppercase, and neither `NJ` nor `MD` appears in class 1 at all.

This matters because the paragraph it sits in is the one arguing that a corpus-wide diff is the wrong test. Quoting corpus-wide numbers inside it undercuts the record.

```
You wrote:      > report as moved. The postal-abbreviation change moves nothing here because the
                > corpus holds 11 places with a two-letter uppercase trailing token — `UT`, `NY`,
                > `FL`, `MT`, `NJ`, `MD` — and no `CO`. Of the 7 places carrying a standalone
                > `co` token, all 7 are genuine county qualifiers and all 7 still strip, including
                > the leading-qualifier Irish forms `"Co Kilkenny"` and `"Co Down"`.

Change it to:   > report as moved. The postal-abbreviation change moves nothing in either class
                > because neither holds a `CO`: of class 1's 11 places whose last token is two
                > letters, the 8 uppercase ones are `UT`, `NY`, `FL` and `MT`, and the rest are
                > `Ny`, `Ca` and `Ma`; class 2 has none. Neither class holds a place with a
                > standalone `co` token at all. A wider scan — every `place`/`standard_place`
                > field under `eval/tests/e2e` and `eval/runlogs`, which is outside the
                > criterion — finds 9 such places, including the leading-qualifier Irish forms
                > `"Co Kilkenny"` and `"Co Down"` in `eval/runlogs/e2e/butler-ancestry/`. All 9
                > still strip.
```

I ran `placeParts` over all 9 to confirm that last sentence rather than eyeballing it.

While you are in that block: I count 463 files under `eval/runlogs/e2e/`, not 451. We are probably counting different things. Say which, or drop the number.

### 2. `docs/specs/record-search-tool-spec-v2.md` — the accepted-costs list is missing the larger of the two costs

You record one accepted cost: lowercase `"Denver, Co"`. There is a second, and it is the more common one. Keeping `co` as a token means a `CO`-scoped search no longer excludes its own **spelled-out** tree place, because `["denver","co"]` is not a subset of `["denver","colorado"]`.

I ran it. A tree holding `"Denver, Colorado, United States"` and `"Denver, Iowa, United States"`, searched `"Denver, CO"`, returns **both** as candidates. The note attached tells the model "Search these before concluding no marriage record exists" — so it is sent back to the place it just searched, and one slot of a capped list goes to it.

That is the same failure the `Co.` paragraph above says this tokenizer exists to prevent. As written, the two paragraphs contradict each other for whoever hits this next. It is not fixable in code — matching `CO` to `Colorado` needs the postal-abbreviation set the ruling rejected — so it is a documentation line.

```
You wrote:      Knowingly
                accepted: `"Denver, Co"`, Colorado written with a lowercase `o`, still strips and so
                still compares equal to `"Denver, Iowa"` — the price of catching `"Hill Co"`, which
                has no other tell. Do not "fix" that without reopening this ruling.

Change it to:   Knowingly
                accepted: `"Denver, Co"`, Colorado written with a lowercase `o`, still strips and so
                still compares equal to `"Denver, Iowa"` — the price of catching `"Hill Co"`, which
                has no other tell. Also accepted, and the more common of the two: keeping `co` as a
                token means a `CO`-scoped search no longer excludes its own spelled-out tree place,
                because `["denver","co"]` is not a subset of `["denver","colorado"]`. Searching
                `"Denver, CO"` therefore offers `"Denver, Colorado, United States"` back as a
                candidate — the failure the `Co.` paragraph above exists to prevent, accepted here
                because suppressing the genuine alternative jurisdiction was ruled the worse of the
                two. Do not "fix" either without reopening this ruling.
```

---

### On your "Unsure about": keep the export

Keep it. The four inputs issue #1584 names in its "Done when" cannot be pinned any other way a reader would recognise as the defect, and the module already exports `isSubCountryPlace` for a comparable reason.

I did find one behavioural test that would have caught defect 1 — two tree facts at `"Coïmbra, Portugal"` and `"Ïmbra, Portugal"` collapse to a single candidate under the old dedupe key. But it needs a place name nobody would ever write, so it is a worse test, not a better one. Your reasoning holds. I would not narrow the doc comment.

### One null result, so nobody has to re-ask

The obvious next question about the ruling is uppercase index strings: `placeParts("HILL CO")` returns `["hill co"]`, one fused token, because the case signal reads an all-caps place as a postal abbreviation. That is inherent to the ruling, not a defect in your implementation — and it is not reachable today. Zero of the 60 recorded search arguments are all-caps, and only 4 all-caps place strings exist anywhere in `eval/`, none carrying a `CO` qualifier. Recording it so the next reader gets the measurement instead of the worry.

---

**Not checked:** the web workspace and `apps/server` suites (the diff touches neither; CI ran both green). No new lint is added, so there was nothing to break and watch fail. I did not re-derive the *first* exemption record's 1623/54 numbers, only the second's. `e2e-fixtures` and `runlogs` I took from CI rather than re-running.

promise-emmanuel (commented):
Second review. I re-ran the first review independently and it all holds — corpus criterion class 1 1712/0 and class 2 60/0, the 11-vs-8 correction, the zero standalone `co` tokens in either class, the wider scan's 9. Finding 1's replacement text is the right one. Agreed on keeping the export.

**two things change from documentation lines into one-line code fixes.**

### 1. Finding 2 is fixable in code

The first review is right about the defect — searching `"Denver, CO"` returns `"Denver, Colorado, United States"` as a candidate, and I reproduced that. But "matching `CO` to `Colorado` needs the postal-abbreviation set the ruling rejected" is not true. After this change a bare `co` token can survive **only** as uppercase `CO` ending its comma part, which the ruling defines as the postal abbreviation and nothing else. So the comparison needs one mapping, not fifty. The fifty-entry set was rejected as a way to decide *whether to strip*; it is not needed to compare a token already ruled to be Colorado.

In `placeTokens` — a line this PR does not currently touch, but this PR is what makes `co` survive:

You wrote:      const parts = placeParts(place);

Change it to:   const parts = placeParts(place).map((part) =>
part === "co" ? "colorado" : part,
);



I ran it: `"Denver, Iowa"` is still kept, `"Denver, Colorado, United States"` is now excluded, all 34 existing tests pass, full engine suite green at 101 files / 2385 tests. It does not re-open the ruling — `placeParts("Denver, CO")` is still `["denver","co"]`, so every one of the 9 new assertions is untouched.

With this fix, "also accepted, and the more common of the two" becomes false. The spec needs the opposite sentence: a surviving `co` compares as `colorado`, which is why the postal abbreviation neither suppresses a same-named place in another state nor resurfaces its own.

### 2. The "null result" is a defect, not inherent to the ruling

`placeParts("Hill CO, Texas")` returns `["hill co","texas"]` and `"Salt Lake CO, Utah"` returns `["salt lake co","utah"]`. The ruling says *keep* the qualifier, not *fuse* it into the token before it — and a fused token is wrong under both readings: county abbreviation wants `["hill","texas"]`, Colorado wants `["hill","co","texas"]`, never one token. It is also wider than the all-caps class chesworth measured: any mixed-case place with an uppercase standalone `CO` hits it. This is defect 2 — the one this PR exists to fix — surviving in the keep branch, which is the one place that stops honouring your own "the qualifier marks a locality boundary" rule.

You wrote:      return part.slice(offset + match.length).trim() === "" ? match : ","; // rule 2

Change it to:   return part.slice(offset + match.length).trim() === ""
? ,${match},
: ","; // rule 2



Both fixes together: 36/36 on the two test files, full suite green.

### One measurement neither of us posted

Zero of the 60 recorded search arguments carry *any* US postal abbreviation, not just all-caps ones — so defect 3's benefit and its inverse cost are reachable in exactly the same circumstances. That is the argument for fixing both directions rather than accepting one: you cannot bank the fix and write off its inverse as unreachable.

### Drop the run-log file count

There are now three: 451 in the PR body, 463 from chesworth, 618 from me counting every file under `eval/runlogs/e2e/`. It decides nothing. Same for the all-caps place count — chesworth got 4, I got 8 scanning every `place`/`standard_place` under `eval/`; none carries a `CO`, so the conclusion is unchanged either way.

**Outstanding:** the two code fixes plus a test for each, and the spec's accepted-costs paragraph rewritten to match. Nothing else in the diff is blocking.

MMagnum:
Both reviews addressed, plus the criterion fix Dallan asked for. Details in the PR body under "Since review".

@promise-emmanuel — your two code fixes landed, a test each: `placeTokens` compares a surviving `co` as `colorado`, and a kept `CO` is wrapped in separators so it stops fusing into the token before it.

@chesworthrm — your finding 2 turned out to be code-fixable, so the spec now says the opposite of "accepted cost". Your count corrections are in; I re-derived rather than copied, and got 12 places / 9 uppercase (`UT`, `NY`, `FL`, `MT`, `LA`) after merging main.

The live-run question resolved as a bug in the criterion itself: it measured raw `place` even where a `standard_place` shadowed it, so both flagged rows were strings the tokenizer never reads. Class 1 went 1771/2 to **1007/0**, class 2 **61/0**. It now lives at `dev/probe-placeparts-criterion.ts` with a self-check that exits non-zero if the baseline is left identical to the current tokenizer (proven to fail). Dallan confirmed no run is owed.

`make engine-test` green at 103 files / 2454 tests, branch level with main.

promise-emmanuel (approved):
Approving. Both fixes are in and I verified them by reverting each one: the separator wrap reddens only "gives a kept CO its own token", the colorado mapping reddens only "EXCLUDES its own spelled-out state". One test each, no collateral. Typecheck clean, 104 files / 2473 tests pass, criterion 1007/0 and 61/0 with 3/3 canaries.

The probe is the best thing here. I set `before()` equal to `after()` — the one way it could report "exempt" having measured nothing — and it printed 0/3 canaries and exited 1. And the collector correction is right on its own merits: I reproduced 1771 distinct / 2 changed on the old regex, both rows the `stribling-father-1821` "Graham Young County Texas, USA" shape whose fact carries a `standard_place` with no County token.

One inline comment on the spec's supporting counts — worth fixing before merge, but it does not touch either `0 changed` verdict.

docs/specs/record-search-tool-spec-v2.md
```diff
@@ -548,12 +548,37 @@ as a behavioural change and re-verify against a live run, not unit tests alone.
...
+> wherever the fact carries a `standard_place`: the tokenizer reads the standardized
+> form, and measuring the raw one reports changes that cannot happen. The criterion
+> lives at `dev/probe-placeparts-criterion.ts` — run it *after* merging `main`, since
+> the counts move with the corpus, and read its header for why it collects what it
+> collects.
+>
+> The postal-abbreviation change moves nothing in either class, because neither
+> holds a `CO`: of class 1's 12 places whose last comma-part is two letters, the 9
+> uppercase ones are `UT`, `NY`, `FL`, `MT` and `LA`, and the rest are `Ny`, `Ca`
+> and `Ma`; class 2 has none at all. Neither class holds a place with a standalone
```
promise-emmanuel: These are the old collector's numbers. Under `standard_place || place` class 1 holds zero places whose last comma-part is two letters — I measured 0 against the new collector's 1007, and 12 against the old regex's 1771. The 12, the 9 uppercase, and `Ny`/`Ca`/`Ma` are all raw `place` values that a `standard_place` shadows, so the tokenizer never reads them. Same shape as the round-1 finding: the collector got corrected, this sentence didn't get re-measured.

You wrote:      because neither holds a `CO`: of class 1's 12 places whose last comma-part is two
                letters, the 9 uppercase ones are `UT`, `NY`, `FL`, `MT` and `LA`, and the rest
                are `Ny`, `Ca` and `Ma`; class 2 has none at all.

Change it to:   because neither holds a `CO` — nor, under `standard_place || place`, a two-letter
                trailing part at all, since the standardized form spells the state out. The 12
                places that do have one (9 uppercase — `UT`, `NY`, `FL`, `MT`, `LA` — plus `Ny`,
                `Ca` and `Ma`) sit in raw `place` values that a `standard_place` shadows, so they
                are outside class 1 by the same rule that took it from 1771 distinct to 1007.

jud-sdev (approved):
Went through this by running the real tokenizer rather than reading it, and it holds up.

I pulled `placeParts`/`stripCountyQualifier` out and ran the old and new versions over 16 scenarios under node. The new code gets every one right — non-ASCII names keep their leading `co` (`Coïmbra`, `Coévrons`), hyphenated ones too (`Co-operative`), a real county still strips, `Denver, CO` tokenizes as `[denver, co]`, and a kept `CO` gets its own token (`Hill CO, Texas` -> `[hill, co, texas]`). The old code fails 10 of the 16, so the new tests genuinely bind rather than passing vacuously.

The callback logic is sound line by line: `CO` survives only as bare trailing uppercase, the dot is kept outside the closing assertion so no stray `.` token comes back, and the strip runs before `.toLowerCase()` so the case signal survives. The one accepted trade-off — lowercase `Denver, Co` still strips — is documented as the ruling's call.

It test-merges clean against main despite being well behind, since nothing on main touched this tokenizer and the spec auto-merged into a different section. All required checks green.

Both of chesworthrm's spec findings are resolved here: the supporting counts now use the wider-scan framing, and the larger accepted-cost was fixed in code with the `placeTokens` `co`->`colorado` mapping and its own test — I confirmed a `CO` search excludes `Denver, Colorado` but keeps `Denver, Iowa`. Nothing to send back.

PR #1734, opened 2026-08-18
Document how to grade a negative test — judge, reviewer, author (#1671)

### LINKED ISSUE #1671: Write down how to grade a negative test — for the judge, the reviewer, and the test author
**Touches:** eval/harness/judge/prompt.md, docs/skill-lifecycle.md, docs/specs/unit-test-spec.md

Deferred from PR #1637 (merged).

**Nine of the 46 judge-vs-human disagreements in the corpus sit on negative
tests, and every one is confusion about what a correct decline looks like —
four by the judge, four by an annotator, in opposite directions.**

Two verified annotator errors, both in the same direction:

- `ut_search_full_text_003` is a negative test where declining is correct. The
  annotator wrote "the skill refused… Wrong decision, not correct routing" and
  dropped both dimensions 3 -> 1.
- `ut_translation_005` should route to `search-records`. The annotator failed it
  3 -> 1 for "no actual FamilySearch records were returned" — grading the
  routed-to skill's work, which the judge prompt explicitly forbids.

## Why prose in one place will not fix it

Judge and human fail the *same way in opposite directions*, so this is a doctrine
gap, not a prompt bug. Install it in three places or it reappears in the other
two:

1. the judge prompt (the framing `_negative_judge_context` builds),
2. the reviewer-facing instructions in `docs/skill-lifecycle.md`,
3. the test-authoring guidance in `docs/specs/unit-test-spec.md`.

## Note what already changed

PR #1637 added `apply_routing_deference`, which floors Correctness/Completeness
off 1 on a **correctly-routed** negative test — so the judge half of this is
partly handled deterministically. The annotator half is not handled at all.

kofiatinka12:
Writes down how to grade a negative test in the three places the doctrine has to live, closing the gap behind nine of the corpus's judge-vs-human disagreements (four judge, four annotator, in opposite directions).

## Why prose, and why three places
The **judge** half is already handled deterministically by #1637 (`flag_routing_negative_judge_fail` + `apply_deterministic_deference`), so a correctly-routed negative can't be flipped by a judge FAIL on the base dimensions. The **annotator** half was undocumented — and that's where the two verified errors happened (a correct decline failed as a "wrong decision"; a routed-to skill's absent output graded against the skill under test). Prose in one place recurs, so the rule is installed for all three audiences.

## What each file now says
- **`eval/harness/judge/prompt.md`** — grade the *decision* to route, not the work of the skill routed to; don't fail the base dimensions because the records/translation/other result the user wanted is missing or thin — producing it is the routed-to skill's job, on its own tests.
- **`docs/skill-lifecycle.md`** (reviewers/annotators) — a correct decline is a full pass (a **3**), never a "wrong decision"; don't grade the skill it routed to; and a decline to the **wrong** skill still fails.
- **`docs/specs/unit-test-spec.md`** (test authors) — a negative test grades the routing decision; write `judge_context`/`explanation` about the boundary, never about the output the routed-to skill ought to return.

## Two distinctions the doctrine makes explicit
- **Correct decline vs. incorrect/wrong-route decline.** The pass is a decline *to the right place* — the skill in `negative.correct_skill`, or (for `correct_skill: []`) no skill at all. A decline to the wrong skill is a failure, matching `_compute_outcome`.
- **Don't grade the routed-to skill's work.** Its records/translation/output are out of scope and have their own tests. Wording deliberately avoids implying the routed-to skill runs: on a routing negative the harness **short-circuits at the hand-off**, so that skill never executes during the test.

## Scope
The three Touches files only. No code, schema, or fixture changes; the judge prompt isn't in the run-log snapshot (rule 2b, warn-only), so no eval re-run is required. Routes to senior-genealogists via CODEOWNERS on `eval/harness/judge/`.

## Validation
- doc-links / ratchet: **130 passed** (no issue-refs added; `unit-test-spec.md` held at its baseline of 4).
- judge (`test_judge.py`): **76 passed** (incl. the guard on every prompt.md edit).
- negative/routing (`test_orchestrator.py`): **18 passed**.

Closes #1671


mercyokum (changes requested):
docs/specs/unit-test-spec.md:790, in "Grading negative tests," is now wrong and should be fixed in this PR.

The line reads:

"Partial credit is reserved for the LLM judge to grade cases where the skill correctly declined but suggested the wrong alternative in its text response (rare; mostly applies when skills_invoked is empty but the text recommends a different skill)."

Line 994, further down in the same file, says the opposite: "negative tests don't have rubric dimensions, so partial doesn't apply." The code agrees with 994, not 790. In eval/harness/harness/orchestrator.py, _compute_outcome's negative-test branch returns only "pass" or "fail" on every path. The function's only "partial" return sits after a comment that says "Positive tests only" — a negative test can never reach it.

Line 790 wasn't always wrong, though. It described real behavior until two days before this PR: commit c6a25de6 ("Stop overriding the judge on routing negatives," Aug 16) deleted apply_routing_deference, the function that used to floor a correctly-routed negative's judge score up to a 2. That was the only mechanism that ever produced partial credit for a negative test. Nothing replaced it, so line 790 now describes code that no longer exists.

Since this PR is settling how negative tests are graded, please fix line 790 here rather than in a follow-up — it's a one-sentence change, same file, same topic.

Suggested replacement:

"There is no partial credit for a negative test. If the skill declines but recommends the wrong alternative in its text response while skills_invoked stays empty, that still fails — the same as any other case where negative.correct_skill isn't matched."

kofiatinka12:
Fixed — you're right, and verified end to end. Line 790 described `apply_routing_deference`, which `c6a25de6` deleted on Aug 16; it contradicted line 994 and `_compute_outcome` (whose negative branch returns only `pass`/`fail`, and returns `fail` when a decline leaves `skills_invoked` empty without matching `correct_skill`). Applied your suggested replacement:

> There is no partial credit for a negative test. If the skill declines but recommends the wrong alternative in its text response while `skills_invoked` stays empty, that still fails — the same as any other case where `negative.correct_skill` isn't matched.

This also lines up with the new "Grading a negative test" doctrine this PR adds (a decline to the wrong place fails). Kept it to the same three files — no scope creep. doc-links/ratchet still green (issue-ref count unchanged). Re-requesting review.

EdmondOware (approved):
Verified this independently rather than taking the claims at face value. Confirmed apply routing deference genuinely does not exist anywhere in the codebase anymore, and read through compute outcome's actual negative test branch line by line, every path returns exactly pass or fail, never partial. The fix matches the suggested replacement text exactly.

The doctrine itself is sound and consistent across all three files, grade the routing decision, not the routed to skill's output, a correct decline is a full pass, a decline to the wrong skill still fails. This lines up with how routing tests already work elsewhere in this repo.

The one blocking point is resolved.

mercyokum (approved):

PR #1735, opened 2026-08-18
init-project: ask objective, experience level, access in one non-blocking opening turn

### LINKED ISSUE #1510: init-project: ask experience level, subscriptions and objective in the opening turn (and fix four docs that already promise it)
> **Rewritten 2026-08-10 by the weekly board audit.** The original framing was wrong: it called two shipped, deliberate behaviours defects. The lead has now ruled, and this issue is the work that ruling creates. The original report is preserved at the bottom.

**Touches:** packages/engine/plugin/skills/init-project/SKILL.md, eval/tests/unit/init-project/, CLAUDE.md, docs/specs/research-schema-spec.md, docs/specs/schemas/research.schema.json, packages/schema/schemas/research.schema.json

**IMPORTANT**: If you do this, do #1320 at the same time — both edit `packages/engine/plugin/skills/init-project/SKILL.md` and both flip its eval run log inactive. #1320 is the same objective-half defect reported from a user's side (three reports from one alpha tester). Land them in one PR so `init-project` pays one re-run + annotation, not two.

> **Read the skill on `main` before starting, not from this body.** `project_create`
> landed in PR #1690 on 2026-08-17 and changed how `init-project` brings a project
> into being — it now creates both documents through a validated tool instead of a
> bare `Write`. The single-pass guarantee this issue turns on is unchanged, but the
> surrounding text is not what it was when this was written. That merge is also what
> freed this skill's eval slot; it is free as of 2026-08-17.

## The ruling

**`init-project` asks the researcher-profile questions in its opening turn, alongside the person ID, then completes in a single pass with whatever came back.** It does not stop mid-flight to wait.

That preserves the property the current behaviour was built for while giving the tester what they expected.

## Why the original framing was wrong

The two things reported as defects are what the skill is explicitly instructed to do, and both landed on purpose in `daaf9bb2` ("init-project: stabilize suite (9/9 aggregate) + real reliability fixes").

`packages/engine/plugin/skills/init-project/SKILL.md:36-40`:

> **Never stop and wait for answers. Complete the full initialization in a single pass:**
> 1. If the user's message already states answers … map and normalize them and keep going.
> 2. Otherwise, use defaults (`intermediate`, `["none"]`), write files now, and tell the user in the final summary that defaults were assumed.
>
> Asking questions and stopping is a failure: the project never gets created.

And `:110`: *"If the user provides just an ID, formulate a default objective from what's missing."*

So the agent complied exactly. Anyone "fixing" this without reading that section will reintroduce the failure it was written to prevent — a run that asks and stops, leaving no `research.json` at all.

## The real defect: four documents promise an interview the skill is forbidden to run

This is what the tester actually collided with. All four say `init-project` conducts a short two-question interview:

| File | Line |
|---|---|
| `CLAUDE.md` | 424 |
| `docs/specs/research-schema-spec.md` | 51 **and** 228 |
| `docs/specs/schemas/research.schema.json` | 97 |
| `packages/schema/schemas/research.schema.json` | 97 (byte-identical mirror) |

The two schema files must stay byte-identical, so they are edited together.

## The work

1. **`init-project/SKILL.md`** — ask experience level and subscriptions in the opening turn, together with the person ID request. Keep "never stop mid-flight": if the user answers, use the answers; if they do not, proceed with defaults and say so in the summary. The `:36-40` block is rewritten, not deleted — the single-pass guarantee stays.
2. **`:110`** — the objective. Same shape: ask in the opening turn rather than inventing one, then proceed. This half is the same defect issue #1320 reports from a user's side; coordinate with it.
3. **The four documents above** — correct to describe whatever lands, so a fifth reader does not file this again.
4. **`eval/tests/unit/init-project/`** — a test that pins the opening turn asks both questions *and* that a non-answering user still gets both files written. Both halves, or the next stabilization pass silently reverts one.

## Eval slot warning

This lands in `init-project`'s snapshot, which is already the most contended slot on the board: PR #1434 is open against `init-project/SKILL.md`, and issues #1003, #1320, #1322 and #1448 all touch it. **Do not land this alone.** At minimum, batch with issue #1320 — it is the same objective-before-action defect, and both bodies already say to land together with #1003.

## Acceptance

- [ ] The opening turn asks experience level and subscriptions; no turn ends waiting for them
- [ ] A user who answers nothing still gets `research.json` and `tree.gedcomx.json` written in the same pass, with defaults named in the summary
- [ ] All four documents describe the shipped behaviour
- [ ] A unit test covers both the asking and the no-answer path
- [ ] Landed in one PR with issue #1320 (and #1003 if it has not merged), so `init-project` pays one re-run and one annotation

---

## Original report (2026-08-10)

Observed live in Cowork on 2026-08-09, current plugin build. The user's whole prompt was:

> Start a new research project for FamilySearch person KWCJ-RN4.

`init-project` ran and asked nothing. It invented two things and reported them afterwards: the researcher profile (experience level, subscriptions) and the research objective.

---

# Merged in from issue #1448 (2026-08-11): the plain-prose start that produced no research.json

Same file, same eval slot, and this issue's own warning already named issue #1448 as one
of the four cards contending for it. Folded in rather than cross-referenced.

## The report

From cmadsen1@gmail.com, 2026-07-31 19:31 (surfaced while triaging issue #1320): a
plain-prose prompt produced only a markdown file with no `research.json`, so the viewer
showed nothing —

> "I had to redo the prompt using /init-project and follow the questionnaire prompts before it did the research.json file."

Issue #1320 flagged this as worth handling in the same pass — "the skill either owns
project setup or it doesn't" — but its own fix was scoped to the objective-before-action
ordering bug, which is a different failure mode.

## Step 0: get the failing prompt, then pick the cause

Do this **before** writing the opening-turn rewrite above, because the answer decides
whether this half is even in `SKILL.md`.

1. **Trigger-keyword gap** — `init-project`'s frontmatter `description` lists specific
   trigger phrases ("new project", "start research", "research [person]", "find parents
   of", "begin researching", "I don't have their FamilySearch ID", or a bare PID). If the
   failing prompt matched none of them the skill never fired at all, and a general
   orchestrator response produced the summary-only markdown file. **This is the case that
   belongs in this issue** — it is a description-line edit in the same file the
   opening-turn rewrite touches, and it lands on the same paid run.
2. **Routing/orchestrator interception** — the skill fired but something upstream treats
   plain prose differently from a `/init-project` invocation. **If this is the cause, spin
   this half back out as its own issue** against the orchestrator, with the reproduction
   attached. Do not force it into a `SKILL.md` PR.

## Adds to acceptance

- [ ] A plain-prose research-start prompt (no slash command, no trigger phrase verbatim)
      results in `research.json` and `tree.gedcomx.json` on disk, not a summary file
- [ ] A unit test pins that path, alongside the two opening-turn tests above

florencemashipei:
## Summary

- Unifies the three opening-turn questions (research objective, experience level, access) under one non-blocking, single-pass, ask-then-default rule in `init-project/SKILL.md`. The objective previously blocked while the other two didn't — that asymmetry is the defect issue #1510 ruled on. Closes #1510 and #1320 together per the lead's instruction on #1320.
- Rewords "paid subscriptions" → "access" in `init-project` and `search-external-sites`, so free FamilySearch-partnership/library access doesn't get steered into answering "none".
- Fixes the maiden-name stub rule to stop assuming patrilineal descent without evidence; adds a `tree_correct`/`tree_edit` correction path if the wrong parent was stubbed.
- Syncs `CLAUDE.md`, `research-schema-spec.md`, and both `research.schema.json` mirrors (verified byte-identical) to the new opening-turn wording.
- Moves exact-value checks (default objective text, default profile fields) out of judge grading and into deterministic Python validators (`test_objective_default_verbatim`, `test_profile_defaults_when_all_default`).
- Fixes a pre-existing bug found along the way: a Windows-only `tempfile` stdlib RecursionError in the unit harness's tempdir cleanup (`_rmtree`'s `onexc` retries forever on a locked-file `PermissionError` it misreads as "must be a directory"). Confirmed unrelated to this PR's changes via `git stash` isolation before fixing. `ignore_cleanup_errors=True` doesn't cover this path; catching `RecursionError` around the `TemporaryDirectory` block does, without discarding the already-computed result.

Step 0 of the approved plan (recovering the original #1448 trigger-keyword report) could not be completed — the original failing prompt isn't archived anywhere in this repo or in either issue, so that half is treated as unrecoverable rather than guessed at.

## Test plan

- [x] `eval-skill SKILL=init-project` — 12/12 pass, releasable candidate `v1_2026-08-18_12-44-49.json`
- [x] `eval-skill SKILL=search-external-sites` — 14/14 pass, releasable candidate `v1_2026-08-18_13-13-17.json`
- [x] Both run logs' 5-test review samples annotated (peer-reviewed)
- [x] `check_runlogs.py` — all 4 rules pass
- [x] Full harness unit suite (2232 tests) green after the orchestrator fix
- [x] `docs/specs/schemas/research.schema.json` and `packages/schema/schemas/research.schema.json` confirmed byte-identical
- [x] No stale "two-question" wording remaining in touched docs


johnmarkpeterbrown (commented):
Five edits, then re-run nothing paid. In order:

**1. Fix the `except RecursionError` scope** — `eval/harness/harness/orchestrator.py:749`

Add the `attempt_completed` flag and `if not attempt_completed: raise`. This is the only finding that can silently corrupt eval data: I reproduced a run where attempt 1's result comes back stamped as attempt 2's, with no exception and no warning. The three exact hunks are in the report above; I verified them against the real cleanup case and all 55 existing orchestrator tests.

**2. Update the answer key** — `eval/tests/unit/init-project/from-objective-only.json:22`

It still demands "Mary Donovan's father". Until this changes, the 12/12 is not evidence the maiden-name half works, and a future run can fail for doing the right thing.

**3. Sync the ownership register** — `docs/specs/schemas/ownership.json:43`

Both the `requires` and `remedy` fields still say "never from assumed defaults" / "ask rather than assume". The always-write behaviour is the lead's ruling, so the register is what's wrong — but it's the document that records the reasoning, so leaving it invites the next person to revert the skill.

**4. Spell the gender** — `packages/engine/plugin/skills/init-project/SKILL.md:186` and `:197`

"sex unspecified" needs to be `gender: "Unknown"`. Omitting the key makes `project_create` reject the whole tree, so init-project writes nothing.

**5. Two stale doc lines** — `README.md:503` and `docs/specs/gps-mentor-agent-spec.md:41`

mercyokum (approved):
Looks good and genealogically correct!

T-FEH (changes requested):
Requesting changes so this can't merge with @johnmarkpeterbrown's findings open — his review was `COMMENTED` with no inline threads, so nothing was mechanically holding it, and @mercyokum's approval is on the current head. One more approval and it would have gone in with all five outstanding.

**I checked all five independently. All five stand.** I also pushed the three that don't touch the eval snapshot (`a4713939`), so what's left is only what genuinely needs a paid run.

**#1 is the serious one, and it is worse than a lost run.** The `except RecursionError` wraps the entire `with TemporaryDirectory(...)` block, not just its cleanup — the comment asserting "by this point build_workspace/run_skill already completed" is what makes it look safe. I reproduced it: attempt 1 aborts retryably, attempt 2 raises from the body, and the call returns **attempt 1's result stamped `attempts=2`**, no exception and no warning. That is corrupted eval data. Fixed with the `attempt_completed` flag you specified, and pinned by `test_a_recursionerror_from_the_body_is_not_swallowed`, which fails when the guard is reverted. A first-attempt RecursionError also gave `AttributeError: 'NoneType' object has no attribute 'aborted_reason'`, masking the real error — that path is closed too.

**#3, #5 — fixed.** `ownership.json`'s `requires` and `remedy` both still described the pre-#1510 blocking behaviour; that file is the register of record, so leaving it invites the next person to revert the skill. Plus `README.md:503` and `gps-mentor-agent-spec.md:41`. I also found a fourth you missed — `researcher-profile-interview.json`'s `test.description` still says "two-question profile interview". That one is free: `eval/CLAUDE.md` strips `test.{name,description,tags}` from the snapshot, so it costs no re-run. Confirmed by `check_runlogs.py` still passing all four rules.

**#2 and #4 I deliberately did not touch**, because both are snapshot-relevant and either one flips `init-project`'s run log inactive and owes a fresh `make eval-skill` plus re-annotation. Batching that is your call, not mine to force by editing.

- **#2 is the one that matters for the evidence.** `from-objective-only.json:22` still demands "Mary Donovan's **father** — **his** surname… **his** identity". `ut_init_project_002` is inside the 12/12 and is *not* in the review sample, so the green run is not evidence the maiden-name half works, and a future run can fail for doing the right thing.
- **#4 — I'd downgrade it slightly from how it reads.** It is a real hazard: `validator.ts:1266` requires `gender` on every person and `project-create.ts:179` returns `ok:false` before `atomicWriteBoth`, so a missing key loses both files. But the live run *did* write `gender: "Unknown"`, so this is a wording trap rather than an observed break. Still worth the four words, and the same paragraph already does exactly this for the sibling field — "**Spell the unknown given name as `given: ""` — do NOT omit the key**" — so the omission is inconsistent with its own neighbour.

Verified by running, on the merged tree: harness **2300 passed, 2 skipped**; engine vitest **2375 passed / 100 files**; `tsc` clean; `apps/server` **209 passed, 1 skipped**; `check_runlogs.py` all four rules pass (two warn-only judge-hash notes). All 12 checks green.

Two things on my own account:

- The two new validators (`test_objective_default_verbatim`, `test_profile_defaults_when_all_default`) have **no coverage** in `test_init_project_validator.py`, whose own docstring says that file exists to prevent exactly that. I mutation-tested them and 9/9 fired, so they are not vacuous — it is a coverage gap, not a dead check. Worth closing while the re-run is being batched anyway.
- Schema multi-site is complete: both mirrors byte-identical, `enums.schema.json` correctly untouched (description-only change), no `CLOSED_ENUMS` or TS-union edit owed.

@mercyokum — flagging rather than second-guessing: your approval stands on the current head, but #2 is a genealogical contradiction inside the answer key, so the genealogist lane isn't really cleared until that one is settled.

T-FEH (approved):
Clearing my changes-request. Both items are properly fixed, and you did the thing they actually required rather than the cheap version — a fresh `eval-skill` run (`v1_2026-08-19_08-01-51`) with its annotation, which is what any edit to those two files owes.

**#2 — resolved, and the point of it is now satisfied.** The answer key reads "Mary Donovan's **parent**" and goes further than asked: it explicitly tells the judge not to credit "father" language or a `sex` value asserted from the surname alone. The reason @johnmarkpeterbrown raised it was that the 12/12 wasn't evidence for the maiden-name half — I checked the new run and **`ut_init_project_002` passes under the corrected key**, so it is now.

**#4 — resolved.** `SKILL.md` says **`Set this person's gender to "Unknown" — do NOT omit the key`**, with the consequence spelled out ("fails the write for both project files, not just this person"), and the worked example follows. That matches the pattern the sibling `given` field already had, which is what made the omission stand out.

You also closed the coverage gap I'd flagged as non-blocking. I mutation-tested both validators rather than take the new file on trust: neutralising the objective check fails 2 tests, neutralising the profile-defaults check fails 3. Nothing vacuous.

`check_runlogs.py` passes all four rules against the merged tree; the annotation carries 38 corrections. My three earlier fixes all survived the merge.

**I pushed one more (`c5985421`), and it's a real miss — by both of us.** The stale line we fixed was a bullet in the feature list at `README:503`. The section at **`README:319`** is the *user-facing* explanation, and it still said "the skill asks you two short questions", listed only Experience level and Paid subscriptions, and **omitted the research objective entirely**. That's the first thing a user reads, contradicting both headline changes of this PR.

Worth knowing why five passes missed it: the phrase wraps across `:321-322` — `"asks you two"` / `"short questions"` — so every line-oriented grep came back clean. The straggler sweeps we all ran were structurally incapable of finding it.

It now reads three questions in one opening turn with the objective first, "Access" instead of "Paid subscriptions" including the point that free access counts (a FamilySearch affiliate library or a partner subscription is access, not "none" — the reframing #1510 asked for), and a line saying none of the three blocks and anything skipped takes a named default.

One left for you, deliberately not touched: `docs/onboarding-slides.html:547` still has "Question 2 — Paid subscriptions". Same staleness, but it's a two-column slide and fitting a third question into it is a layout decision, not a wording fix.

Verified: packaging **490 passed**, harness **2356 passed, 2 skipped**, all 12 checks green.

PR #1738, opened 2026-08-18
Cite wiki page URLs, not just titles, in historical-context and locality-guide

### LINKED ISSUE #1315: Wiki content is presented without its source URL

Cia-3:
cat << 'EOF'
Closes #1315.

## What was actually wrong

The issue's `Touches:` line guessed at the three wiki MCP tool files
(`wiki-read.ts`, `wiki-search.ts`, `wiki-place-page.ts`). That guess is
wrong — all three already return the page URL in their structured output
(`source_url` on `wiki_search` results; `url` from
`wiki_read`/`wiki_place_page`/`wikipedia_search`). No tool code changed in
this PR.

The real gap was narrower than the issue implies, because a sibling issue
(#1321, already closed) fixed half of it:

- `locality-guide/SKILL.md` already had a citation rule and a passing
  regression test (`ut_locality_guide_026`) — but its output template
  (`references/output-format.md`) was never updated to model it.
- `historical-context/SKILL.md`, which calls `wiki_search`/`wiki_read`/
  `wikipedia_search` directly and narrates results to the user, only had a
  rule requiring the source be *named*, not linked. No test in the corpus
  checked for URL citation on this skill at all.

## Changes

- `historical-context/SKILL.md` — every response that calls a wiki/
  Wikipedia tool must now end with a "Sources consulted" list (title
  linked to its real URL) — the same structural pattern
  `search-familysearch-wiki` already uses reliably. I initially tried an
  inline "cite every claim" rule instead; across 3 standalone reruns it
  never got a clean citation pass (the model would cite the first source
  properly, then drop to name-only for later ones in the same response).
  Switching to a structural end-of-response list fixed that: 3/3 clean
  runs afterward.
- `eval/tests/unit/historical-context/rubric.md` — split the old "Source
  quality" dimension in two: **Source quality** (factual accuracy) and a
  new **Citation completeness** (URL presence, graded on the sources list
  specifically). These are genuinely independent failure modes — a
  response can be fully accurate with no links, or fully linked with an
  overconfident claim — and bundling them meant an unrelated accuracy slip
  could mask whether the citation fix actually worked.
- `locality-guide/references/output-format.md` — the "Research tips"
  bullet now models the citation the SKILL.md rule already required.
- New regression test `ut_historical_context_013` (German immigration to
  Pennsylvania, 1840s — historical-context's own worked SKILL.md example).

## Proof the fix works

Before editing `historical-context/SKILL.md`: the response named sources
by title ("As the FamilySearch wiki confirms...") with **zero actual
links** anywhere — confirmed by reading the raw response text directly.
After the fix, across 3 separate standalone runs, **Citation completeness
scored a clean 3 every time** — each run's response ended with a real
"Sources consulted" list linking every wiki page actually used, verified
by reading the raw text, not just the judge's summary.

The test still comes back `partial` overall in 2 of those 3 runs, but now
for a clearly isolated, different reason: the model occasionally states a
general historical claim (crop failures, political unrest) with more
certainty than its tools confirmed, despite flagging some other claims as
unconfirmed in the same response. This is a real, pre-existing behavior
(it showed up in the very first baseline run too, before any of my
changes) — separate from citation, and out of scope for #1315. Flagging it
here rather than silently dropping it; happy to file a follow-up issue if
you'd like one, otherwise leaving it as a documented observation.

## Eval

`make eval-skill SKILL=historical-context`: **12 pass, 1 partial** of 13
(the partial is the new test, explained above — its Citation completeness
dimension is clean).

`make eval-skill SKILL=locality-guide` (from the earlier commit, rubric
unchanged since): **34 pass, 2 partial, 3 fail** of 39 combined with the
above at the time. All three fails and the locality-guide partial were
individually re-verified standalone and confirmed unrelated to this PR's
doctrine (one is the harness's documented Agent-SDK skill-discovery
false-fail; one is a known-flaky "near-miss" routing test — confirmed
unrelated by reverting my edit and rerunning; one is a pre-existing
fabricated-collection-ID issue, unrelated to citation).

Both run logs fully annotated, every review-sample dimension covered with
a comment on every cell that isn't a confirmed pass, documenting the
standalone-verification evidence above.

`eval/harness/scripts/check_runlogs.py` passes locally against this
branch.

## Who should review

Needs a peer review (developer or genealogist) before a senior genealogist
sign-off, per the usual flow.
EOF


EdmondOware (approved):
Verified the field name claims directly against source rather than trusting the PR body. Source url really is defined on wiki search's result type, and url really is what wiki read and wikipedia search return. Locality guide's cite the wiki page rule genuinely already exists in its SKILL.md, confirming the output format.md fix closes a real gap.

The rubric split into Source quality and Citation completeness is the right call, bundling them would have let an unrelated accuracy slip mask whether the citation fix actually worked, and the new test's judge context explicitly guards against that.

I read the actual annotator notes for every non passing test rather than just the summary. The historical context fail is a documented single run infra jitter, zero tool calls that run, confirmed clean on a standalone rerun. The locality guide fail on ut_026 is unrelated, fabricated collection ids and a recovered tool arg miss, and that same failing run still correctly cites its wiki source with a real url, direct evidence the fix works. The ut_003 fail was isolated properly, reverted the edit and reran clean, restored it and reran twice more clean, only the first full suite run failed. The ut_007 partial is a pre existing overclaim and pagination gap, unrelated.

Annotation is complete for both skills' review samples and CI is fully green.

clack391 (approved):
Approving, senior developer (the genealogist-craft lane is @EdmondOware's, and his approval covers it).

Scope is complete against #1315 as re-scoped: I checked every skill that narrates wiki content. historical-context is fixed here; locality-guide has the rule (#1321) plus the output-format template modeled here; search-familysearch-wiki and search-wikipedia already cite the URL via their templates; research-plan grants no wiki tools and delegates to locality-guide, so it is genuinely out of scope despite being named as a candidate. So "Closes #1315" holds.

The citation fix is verified real, not just asserted: the committed historical-context run log's ut_013 response ends with a real "Sources consulted" list of three FamilySearch URLs that trace to the fixture, scored Citation completeness = 3 under the new rubric. check_runlogs exits 0 with both run logs active and annotated; no stale name-the-title-only passage survives in historical-context/SKILL.md; the harness suite is green (the lone test_mock_mcp failure is the missing-engine-build env case, and this PR touches no Python).

Two non-blocking notes: both run logs were scored against an older judge prompt than current main (a7c9bd99 -> 0d186137, warn-only, no re-run required); and the pre-existing overclaiming behavior the body discloses is now caught by the tightened Source quality dimension but not fixed here, which is a fair follow-up in the genealogist lane if you want one filed.

PR #1739, opened 2026-08-18
refactor: consolidate readJson and formatIssues into project-io.ts

### LINKED ISSUE #988: Six private copies of readJson / formatIssues across the writer tools

yinkid28:
Closes #988

## Summary

- Six tools carried private byte-identical copies of `readJson` and four carried copies of `formatIssues`. `project-io.ts` is now the single canonical home for both.
- Each tool wraps the shared `readProjectJson` in a local function that re-throws through its own error class, preserving the existing `{ ok: false }` contract (the `tree-forget.ts` pattern the issue specifies).
- `merge-shared.ts` becomes a re-export layer — imports from `project-io.ts`, re-exports `formatIssues` bare and `readProjectJson` wrapped with `MergeInputError`.
- 16 characterization tests pin read-failure behavior across all five migrated tools.

## Start here

The `readProjectJson` wrapper pattern in any of the migrated tools (e.g. `tree-edit.ts:155`).

## Unsure about

Nothing — the issue was prescriptive about every file, every error class, the canonical home, and the wrapper pattern.

## Didn't change

- `tree-forget.ts` — already uses the pattern; this is the model, not a migration target.
- `rank-search-matches.ts` — deliberate best-effort read with different semantics.
- `merge-tree-persons.ts`, `merge-warnings.ts`, `project-create.ts` — consumers of `merge-shared` exports, unchanged.

## Deviated from the plan

No deviations.

## Acceptance check

- `npx vitest run` — 2390/2390 pass (101 files)
- `npx tsc --noEmit` — clean
- `grep "not found in projectPath\|is not valid JSON"` in `src/tools/` returns zero hits — no private copies remain

## Test plan

- [x] 16 characterization tests in `tests/tools/read-failure.test.ts` covering missing-file and invalid-JSON for all 5 migrated tools
- [x] Tools that read both files (`tree_edit`, `materialize_facts`, `research_append`) have both `research.json` and `tree.gedcomx.json` failure tests
- [x] Full engine suite passes
- [x] TypeScript compiles clean


francis-2008-happy (approved):
Approving. Ran it in a worktree off this branch — `tsc --noEmit` clean and `vitest` green (2391), including the new `read-failure.test.ts`. Confirmed the error messages are preserved byte-identically, and the new tests assert those exact strings, so they'd catch a regression. Clean, behavior-preserving dedup. Nice.

T-FEH (approved):
Approving. Clean, genuinely behaviour-preserving dedup, and it matches #988 line for line — canonical home, wrapper pattern, `project-context.ts` as the direct-import exception.

I did not take the byte-identical claim on trust. Built a worktree at `origin/main` and ran a **176-case differential probe** — 8 tools × both files × missing / empty / `[]` / `"str"` / `null` / `42` / BOM / whitespace / bad JSON / directory / chmod-000 — and the output is byte-identical between `main` and this branch. @francis-2008-happy's claim holds. The 16 characterization tests are not vacuous either: changing either error string kills 8, and swapping a tool's error class for a plain `Error` kills exactly that tool's 2–4.

I pushed three follow-ons (`316929b5`).

**1. The comment this PR deletes ended "Nothing enforces the line above" — and nothing still does.** That sentence was accurate: it is how one private copy became six, the sixth written *three days after* the shared helper landed. Deleting the inventory is right, it is obsolete — but the evidence that the sweep holds is a one-time `grep`, so the same drift recurs silently and the next reader has lost both the warning and the reason. Added `tests/packaging/single-project-read.test.ts`, modelled on its sibling `no-bare-fetch.test.ts`, which exists for exactly this shape of problem. Proven to fail: reintroducing `"not found in projectPath"` into `research-query.ts` turns it red and names the file.

It matches the **message strings**, not a helper name, deliberately. The six copies were variously `readJson`, `readProjectJson`, and inlined with no helper at all — a name-based check misses the inlined case, which is precisely the one that slipped in last time.

**2. Finished the repoints.** `tree-forget.ts:37-38` and `project-create.ts:42` still reached `formatIssues` through `merge-shared.js` — the exact wart #988 named, and a one-liner each now that `project-io.ts` owns it. `merge-tree-persons.ts` and `merge-warnings.ts` keep theirs legitimately; they need the `MergeInputError`-wrapped `readProjectJson` from that module.

**3. `research_log_append` reads both files but only had the research.json pair.** The PR body lists `tree_edit`, `materialize_facts` and `research_append` as the both-file readers — this one is a fourth. research.json fails first, so the tree read was unpinned. Added the matching pair; both go red when the tree filename is broken. Behaviour was already correct — this is coverage symmetry, not a bug.

Verified: engine vitest **2393 passed / 102 files**, `tsc` clean, all 12 checks green.

One note on the gate: @francis-2008-happy is on neither senior team, so his approval does not satisfy CODEOWNERS — every rule now names both senior teams. Mine does, so this should be unblocked now.

PR #1740, opened 2026-08-18
e2e: warn when a non-relationship finding's matched disagrees with its components (#1721)

Asimi1234:
## Summary

Normal tier. Implements the ruling on #1721 — **report the disagreement, do not widen the derivation.**

- `apply_component_derivation` recomputes a finding's `matched` from its own `components`, but only for `relationship` findings. A `source`, `fact` or `person` finding keeps whatever `matched` the judge wrote even when its own `components` resolve to a different label, and nothing reports it (the issue's f4 case: a `source` finding with `matched: "partial"` whose two links both `supported` → derives `true`).
- Adds a **warn-only** content check to `check_e2e_fixtures.py`: for each PR-added run log, warn when **any** finding's stored `matched` disagrees with `derive_matched` of its own `components` — every finding type, per the issue's Decision. It does not widen the derivation (which stays relationship-only), so the disagreement is surfaced, not corrected.
- Records the decision in `e2e-test-spec.md` §3.4.2 under the existing "Relationship findings only" bullet.

## Review guidance

**Start here:** `check_matched_vs_components` in `eval/harness/scripts/check_e2e_fixtures.py`. The decision worth checking is *what it skips*, since the issue mandates it flag **every finding type** — so the skips are deliberately narrow:
1. findings derivation already reconciled (`matched_model` present — the stored `matched` is the derived value; re-reporting them would just echo the recorded `component_derivation` override);
2. `avoid` findings — there `matched: "true"` means "correctly declined to assert", which is not a link tally; joined by `finding_id` from the fixture's `expected-findings.json`, since polarity does not live in the run log's `per_finding`. **`fact` is deliberately *not* skipped** — it is excluded from the *derivation* (judge.py), but the issue's Decision keeps it in this report;
3. findings with no `link` components (nothing to derive) and findings with no `matched` key (nothing to compare).

Also worth a look: `derive_matched` is a hand-kept stdlib copy of judge.py's, not an import — judge.py imports `anthropic` and this CI job runs bare `python` with no venv, so importing it would break the workflow (§3.4.2 is the shared contract both obey). `test_derive_matched_matches_judge_canonical` imports the canonical one in the harness venv and asserts parity, so the copy cannot silently drift. The reader/parser guards (`isinstance` on the run-log root, `judge_output`, `per_finding`, and the fixture) are deliberate: a wrong-shaped-but-valid-JSON file must be skipped, never raised on — a raise here would fail this required check with a traceback, violating "warn, do not block".

**Unsure about:** Nothing blocking. The issue body's point 2 ("compare against `matched_model` when present") reads two ways; I resolved it against the ruling's own measurement — comparing the *stored* `matched` vs derived is what yields the ruling's "zero of the 15 findings" today, so `matched_model`-present findings are skipped (already reconciled, cannot disagree) rather than compared.

## Plan

**Didn't change:** `apply_component_derivation`'s `eligible` set stays relationship-only — deliberately not widened, per the ruling. `fact` stays excluded from the *derivation* on its own 2026-08-10 measurement — but, per the Decision, its disagreements are still *reported* by this check.

**Acceptance check:** The mandated "break it and watch it fire" (CLAUDE.md § "A new lint must be proven to fail"). The check returns **0 warnings across all 155 committed e2e run logs** — but read that precisely: only 4 of those 155 logs carry `components` at all, and the 7 findings actually evaluated are 5 `relationship` + 2 `fact`. **No committed `source`/`person` finding yet carries components**, so "0 warnings" means "no committed log yet contains the shape this looks for," not "checked and clean" (thanks @chesworthrm) — the check gains reach on its headline target as new logs land. It is proven to fire by the unit tests and by the **reconstructed f4 shape** (a `source` finding, `matched: "partial"`, two `supported` links → derives `true`) injected into a real log+fixture and then restored. Named tests that fail on `main` (the functions do not exist there) and pass here: `test_source_finding_disagreement_warns`, `test_fact_finding_is_reported`, and `test_main_drift_warning_prints_even_when_grading_gate_fails` in `tests/unit/test_check_e2e_fixtures.py`.

## Test plan

- [x] I ran `bash scripts/test.sh` (`make test-all`) — engine, harness (2316 passed / 2 skipped at `1b7b63f9`), server, web, and eval-app suites pass. (The doc-links issue-reference ratchet caught a `#1721` in my spec prose during development; fixed by dropping the ticket number, per the ratchet's own rule "state the fact, not the ticket".)
- [x] No skill run-log snapshot changed (this touches `eval/harness/scripts/`, a harness test, and docs only) — no `eval-skill` run owed.
- [x] No skill `description`/DO-NOT clauses changed — no reciprocity test owed.

## Follow-on issues

**Folded in / filed instead:** None filed.


Asimi1234:
Ran an xhigh code-review pass over the diff and folded the findings into the second commit (`5a4ad1e0`). Two were real:

1. **`fact` findings weren't excluded from the check** — the code excluded only `avoid`, but the spec here (§3.4.2 "`fact` stays excluded"), `apply_component_derivation`, and the ruling all exclude `fact` because its link-tally is not a reliable signal (it "manufactured eleven disagreements" in the 2026-08-10 sweep). 2 `fact` findings in the committed corpus already carry `kind:"link"` components; they agree today, so this was latent, but any future divergence would have emitted exactly that noise. Now excluded via the same fixture join that handles `avoid`. Verified: an injected disagreeing `source` finding fires; an injected disagreeing `fact` finding does not.

2. **Wrong-shaped-but-valid JSON crashed the warn-only check** — the `try/except` caught only `(OSError, ValueError)` around `json.loads`, so a run log whose root/`judge_output` was a non-dict, or `per_finding` a non-list, or a fixture whose root wasn't an object, raised `AttributeError`/`TypeError` and would have failed a required CI job with a traceback — contrary to the docstring's "never raised on". Added `isinstance` guards (mirroring judge.py's `isinstance(parsed, dict)`).

Also folded in: a **parity test** importing judge.py's canonical `derive_matched` and asserting the inline stdlib copy matches it on a table (closes the silent-drift gap from the duplication); a per-slug cache for the fixture read; skip a finding with no `matched` key; and coverage for multi-finding logs and the warning surviving a failing blocking gate.

Re-verified: **0 warnings across all 155 committed run logs**; the whole harness unit suite is green (2266 passed). Net diff is 3 files.

Separately — the flaky `apps/electron/feedback.test.ts` timeout I noted in the test plan is now fixed on `main` (`3a5b41fa`, pulled in by the main-merge on this branch), so that footnote no longer applies.

chesworthrm (approved):
Approving — full pass against `1b7b63f9`, no blockers.

**First, a correction I owe you.** In an earlier pass on `c17a5ae0` I flagged the check for scoring `fact` findings and asked you to exclude them. That was wrong, and I should have read the issue's Decision before writing it up — it says the check covers **every** finding type, with only the derivation staying relationship-only. Your original code was right; it was the prose around it that was off. Sorry for the round trip.

The one part of that finding that did hold is fixed anyway: the body no longer claims the corpus has no `fact` findings carrying link components. It has two, and they are evaluated. Code, spec and PR body now all say the same thing — `fact` is out of the derivation, in the report.

## What I verified beyond CI

- **The check is genuinely reachable.** `check-e2e-fixtures.yml` sets `BASE_SHA`/`HEAD_SHA` and its path filter covers `eval/runlogs/e2e/`, so the gate runs on exactly the PRs this acts on. It runs bare `python` with no venv, and the script imports only stdlib — I ran it under a stripped interpreter to confirm. `judge.py` does import `anthropic`, so the duplicated `derive_matched` is justified rather than lazy.
- **Every guard is pinned.** I mutated them one at a time: re-adding the fact exclusion, suppressing the disagreement, dropping the avoid skip, dropping the `matched_model` and missing-`matched` skips, each shape guard, and drifting the copy away from `judge.py`. All caught by a failing test. No dead guards.
- **`test_derive_matched_matches_judge_canonical` is the best thing in this diff.** A hand-kept copy was the obvious long-term risk and that test closes it — I drifted the copy deliberately and it fails.
- **Harness suite: 2316 passed, 2 skipped, 0 failed.** If you hit a `test_mock_mcp` failure locally, that is the missing `packages/engine/mcp-server/build/`; `make engine-build` clears it.
- **155 run logs produce 0 warnings**, with 7 findings actually evaluated — matching the issue's "zero disagree on main".
- **"3 of 261" is exact.** The fixtures hold 261 findings and 3 are `source`.

## One note for the body, not blocking

The check has no exercise on its headline target yet. No `source` or `person` finding in any committed run log carries components — only 4 of 155 logs carry components at all, and the 7 evaluated findings are 5 `relationship` and 2 `fact`. So "0 warnings across all 155 committed run logs" means "no committed log yet contains the shape this looks for", not "checked and clean". Worth a clause so the next reader doesn't over-read it. The check gains reach as new logs land.

## Pre-existing, not yours

77 of 448 run-log findings carry `finding_id`s absent from their fixture — 61 in `laurie-scotland-parents`, and 4 in `william-ferber-ancestry`, which has no `expected-findings.json` at all. For those the avoid lookup finds nothing and the finding is evaluated regardless. None carries link components today so nothing fires, but it is the same join this check relies on.

T-FEH (approved):
Approving. This is the most carefully-argued lint I've reviewed here — the "what it skips" framing in the review guidance is exactly the right thing to point a reviewer at, and it held up.

**The lint genuinely fires.** Injected the f4 shape (a `source` finding, `matched: "partial"`, two `supported` links → derives `true`) into a real committed log and ran the real CLI with real `BASE_SHA`/`HEAD_SHA`: warning printed, **exit 0**, grading gate still OK. Mutating the warn loop to `return 1` turns `test_main_drift_warning_does_not_fail_the_job` red, so warn-only is pinned rather than merely intended.

**The skip list does not swallow evidence — I measured it rather than reasoning about it.** Across the 155 committed run logs / 448 findings, under the current code:

```
no `components` key at all : 433   (predate it)
components but no `link`   :   6
already derived            :   1
EVALUATED                  :   8
warnings                   :   0
hidden disagreements       :   0   ← skipped findings that WOULD warn
```

That last line is the one that matters: **zero** skipped findings would warn if included, so today's clean result is real agreement, not silence bought by skipping. That was my main worry and it's answered.

The honest caveat, which is why I'd rather it were written down than inferred: with 433 of 448 structurally out of scope, "0 across 155 logs" is thin evidence rather than calibration. `437d26bb` improved this while I was reviewing — dropping the blanket `fact` type-exclusion in favour of the structural no-link rule took evaluated findings from 5 to 8 and shrank the skip surface. Good change, and the right direction: a `fact` finding that *does* carry link components is now reported instead of hidden by its type.

**The drift guard is real.** `derive_matched` is a hand-kept stdlib copy, which is the kind of thing that rots quietly — so I checked the protection rather than the claim. `test_derive_matched_matches_judge_canonical` imports the canonical `e2e.judge.derive_matched` and compares outputs across 9 cases including `None`; mutating the copy (`contradicted → "partial"`) turns it red, and `eval-harness-tests.yml`'s `^eval/harness/` pattern arms it in the `pytest` job with the venv it needs. It cannot drift silently.

Also confirmed: 15 mutations against the new tests, none vacuous — each guard has exactly one owning test. And the `finding_id` join errs toward surfacing: a missing or mismatched id leaves the finding evaluated rather than excluded, and a missing fixture yields an empty exclusion set. That's the safe direction for a warn.

**I pushed one commit (`feec87ed`).** `eval/CLAUDE.md`'s check register still said "one blocking check plus one warn" and its table omitted this one — the same drift the file itself warns about two rows up ("`e2e-nudges` was never added there; don't repeat that"). Added the row with the coverage numbers above, and with the `fact` distinction spelled out, since `437d26bb` had just moved it and "excluded from the derivation but not from the report" is easy to lose.

Harness suite **2316 passed, 2 skipped**; all 12 checks green.

One stale note for the body: the review guidance still says `fact` findings are excluded from the check. `437d26bb` reversed that — worth a line so the next reader doesn't trust it.

Issue #1558, opened 2026-08-11
Adapt the hosted feedback transcripts and run the guardrail detectors over them


DallanQ:
**Touches:** eval/harness/e2e/ (a pure adapter module + the report), eval/harness/tests/unit/, eval/harness/harness/skill_invocation.py (import only — do not modify), docs/specs/guardrail-enforcement-spec.md

> **Lead ruling, 2026-08-11 — measure every submitted bundle, reported split by
> platform.** Chosen over hosted-web-only, which would have measured at most six
> sessions with one detector. `find_missing_mentor_verdicts` reads `research.json`
> alone, so it runs over the Cowork bundles too; `find_unguarded_protected_writes`
> needs a transcript and therefore covers only the `web` and `darwin` bundles.
> Report the two detectors and the platforms in separate columns; never a combined
> number. **The Drive folder is shared** — ask the lead for the link if you do not
> have it.
>
> Also part of the task: add a paragraph to
> `docs/specs/guardrail-enforcement-spec.md` § "Options set aside" recording that
> the tree-reading §8 arms cannot be run over a feedback bundle (redacted tree, no
> `starting_tree`), and that the bundle corpus is measured with the transcript-only
> and `research.json`-only detectors instead.

Split out of #1054 (lead ruling, 2026-08-11): measure with the hosted transcripts we already hold, instead of building a durable ledger to collect more.

## The question this answers

**Do guardrail bypasses reach real users, and at what rate?** Everything known about guardrail compliance today comes from the e2e corpus — synthetic fixtures on our own machines. Nothing has ever measured the hosted path, where real testers work on real family data. That is the number #1054 exists to produce, and the Drive feedback bundles already hold the input, submitted voluntarily as feedback. **Not ten transcripts:** issue #1481 measured the ten bundles from 2026-08-05 onward as five hosted-web *with* a session transcript and five Cowork *without*, so the transcript-only detector covers roughly six sessions and the `research.json`-only detector covers all of them.

No new storage, no new retention window, no privacy-policy sentence: this is data users already sent us, used for the purpose they sent it for.

## Where the code goes

`apps/server` and `eval/harness` are separate uv projects with conflicting
`claude-agent-sdk` pins (`apps/server/pyproject.toml:15` = `>=0.2.128`;
`eval/harness/pyproject.toml:12` = `>=0.1.81,<0.2`), and nothing in `apps/server`
imports the harness — `real_agent.py:127` records that the one shared behaviour is
[...]

PR #1741, opened 2026-08-18
eval: adapt hosted feedback transcripts + run guardrail detectors (#1558, Part A)

francis-2008-happy:
Part A of #1558. Part B — running over the real Drive bundles and posting counts to #1054 — is a follow-on; issue #1558 stays open.

## The issue

Everything we've measured about guardrail bypasses is the synthetic e2e corpus. #1558 asks: do bypasses reach real users? This builds the machinery to replay the two valid detectors over the feedback bundles testers actually submitted. Counts only, never a rate — the detectors are uncalibrated (`docs/architecture.md` §9.4 pt 3).

## The fix

- **Adapter** — `e2e/feedback_transcript_adapter.py`: a pure, stdlib-only function turning a bundle's raw Claude Code `session-log.jsonl` into the harness's flat `tool_calls` shape, mirroring the orchestrator's message loop + result join.
- **Detector wiring** — extends `e2e/guardrail_shadow_report.py` with the two bundle-valid detectors, `scan_feedback_bundle` / `scan_feedback_dir` / `format_feedback_report`, and a `--feedback-dir` CLI reachable from `make e2e-guardrail-shadow FEEDBACK_DIR= PLATFORMS=`.
- **Spec** — a paragraph in `guardrail-enforcement-spec.md` § "Options set aside" on which arms are blind over a bundle, and when.

### Round-4 review fixes (clack391)

1. **The blanket "0 BY CONSTRUCTION" was false for the corpus this scans.** I verified the dates rather than taking them: the proof-conclusion split merged 2026-08-21 (`73b3d98e`), the pre-split `SKILL.md:157` wrote `proof_summaries` from the main thread with no hook, and the newest `feedback` issue is 2026-08-20. So over every real bundle that arm was open and a count is a **real measurement** — the disclaimer told #1054's reader to discard the number it's waiting for. Arms are now labelled **per bundle from its `submitted_at` date**, printed on every row.
2. **A third arm was blind and unlabelled.** `research-exhaustiveness` became a skill-agent pair 2026-08-23 (`c78efb0b`), inside this branch's own history. The date-conditional label generalises to it rather than needing a second hardcoded sentence — so fixing 1 fixed 2, as you said it would.
3. **`response_summary: None` manufactured violations.** Now populated from the `tool_result` content, so `did_not_land`'s no-project clause can fire. Verified end to end: a `research_append` returning `no_project` with no `isError` now yields 0 violations instead of 1.
4. **`UnicodeDecodeError` killed the whole scan.** Widened the research side to `(ValueError, OSError)` matching `_load_json`, and guarded the transcript side with a new `[transcript unreadable]` tag. Caught at this call site rather than widening the shared `parse_jsonl`, which would hand its other caller `[]` instead of raising.
5. **One combined total.** Totals are now reported per platform. Platform also defaults from `_feedback/feedback.json` — **both producers write it**, so it IS in the bundle; `--platforms` stays as the override.

### Should-fixes

All done: five `main()` tests (and the four silent-swallow bugs they found — corpus flags ignored, `PLATFORMS=b1web` dropped, an unmatched name unused, a case dir printing a confident zero); the window guard parsed by regex instead of substring, so a trailing comment or type annotation no longer false-alarms; `#1484` citation replaced with `--replay`; the module docstring, argparse description, Makefile comment, `eval/CLAUDE.md` and `since_window.py` all synced to the fact that this module has two modes and only one is windowed; and issue #1880 filed (`nothing-checks`) for bundling the subagent transcripts.

**On your guard-inversion finding.** You were right that the footer guard was too loose — your rewrite kept all three substrings while asserting the opposite. The claim is now pinned as contiguous phrases, and I re-ran your attack against it: it fails.

## Didn't change

`check_guardrail_compliance` / `find_effects_without_invocation` stay excluded — they need a `starting_tree` baseline a hosted session has no equivalent of, and importing them drags in the Agent SDK. `skill_invocation.py` and the shared `parse_jsonl` are imported, never modified.

## Unsure about

**Whether the per-bundle era label belongs in the report at all, or only in the spec.** It is four lines of prose on every run, and a reader who already knows the split dates does not need it. I kept it in because the failure it prevents — reading a post-split 0 as evidence — is the one this PR exists to prevent, and the spec is not in front of whoever runs the command.

**Whether `_AGENT_SPLIT_DATES` should be derived rather than hardcoded.** It is a hand-maintained map of two entries, and a third skill-agent pair would go unlabelled until someone adds a row. Deriving it from `guard_project_files.py` plus git history is possible but reads a plugin file from the harness. I took the hardcoded map and a spec line; happy to be overruled.

**Whether the hosted sandbox's SDK writes the same transcript line shape as local Claude Code.** Still unverifiable here — no real bundle may enter the repo. `could_not_adapt` is what surfaces a divergence.

## Acceptance check

```
cd eval/harness && uv run --frozen pytest tests/unit/test_feedback_transcript_adapter.py tests/unit/test_guardrail_shadow_report.py
```

79 pass. End to end:

```
make e2e-guardrail-shadow FEEDBACK_DIR=<dir>
```

prints each bundle's date and platform, tags only the post-split one `[plugin era unknown for: proof-conclusion, research-exhaustiveness]`, gives per-platform totals, and states both arms' visibility in both directions.

Every fix mutation-checked:

| Mutation | Result |
|---|---|
| flip `arm_visibility` so post-split reads `live` | date-label test dies |
| revert to the blanket by-construction footer | date-label test dies |
| your inversion (keep terms, assert the opposite) | date-label test dies |
| drop the per-platform block | per-platform test dies |
| narrow the research `except` back to `JSONDecodeError` | bad-encoding test dies |
| revert the transcript UTF-8 guard | bad-encoding test dies |
| revert `response_summary` to `None` | both no-project tests die |
| window 40 → 60 | window test dies |
| window + trailing comment / type annotation | still passes (used to false-fail) |

## Deviated from the plan

Two beyond the review. `response_summary` had no guard — reverting it passed all 21 tests, the same unheld-claim shape as round 3, so I added one. And `format_feedback_report` crashed on a result dict lacking the new keys; made tolerant.

## Verification

- `eval/harness` pytest — 2690 passed, 2 skipped, 9 failed. All 9 are `test_setup_feedback_case.py` on `unzip: command not found` on my machine; identical with my changes stashed.
- `apps/server` pytest — 209 passed, 1 skipped.
- engine `vitest run tests/packaging` — 576 passed, including the doc-links ratchet (which caught my spec at 21 refs against baseline 16; facts kept, ticket numbers dropped).
- Branch 0 behind main.

## For review

`eval/harness/**`, `docs/specs/**`, `eval/CLAUDE.md`, one `Makefile` line. No bundle-derived data anywhere. Every bundle reproduction is synthetic in the documented `_feedback/` layout.

**Merge conflict with #1856 is resolved** — it landed first, exactly as you predicted. Both collisions were unions: `Makefile:595` keeps main's §11 family text plus this branch's `FEEDBACK_DIR`/`PLATFORMS` knobs, and in `guardrail_shadow_report.py` both `format_unnamed_delegate` and the feedback block are kept. Worth knowing: the naive union left `format_unnamed_delegate` without its `return`, because the shared trailing line attached to my block instead — caught by three of your #1856 tests going red.


chesworthrm (changes requested):
Nice piece of work — the adapter itself is right. I pointed it at a hand-built bundle and it produced exactly the violation record you'd want. The `is_error` join is correct too; I checked 2321 lines of real Claude Code transcripts and the key is absent 125 times, explicitly `false` 141 times, and `true` 40 times, so `bool(block.get("is_error"))` covers all three.

One blocking problem, and it stops the whole feature from doing anything.

## The scanner looks for the transcript where no bundle puts it

`scan_feedback_bundle` reads `<bundle>/session-log.jsonl`. Bundles put it at `<bundle>/_feedback/session-log.jsonl`.

That path is a contract, not an accident:

- `docs/specs/feedback-case-spec.md` §2.2 — "The bundle carries the Claude Code session JSONL at `_feedback/session-log.jsonl`."
- Both producers write it there: `apps/server/app/feedback.py:464` and `apps/electron/src/main/feedback.ts:369`.
- `scripts/setup-feedback-case.sh:83` unpacks with plain `unzip -d`, which keeps that layout.

So on every real bundle, `has_transcript` is `False`, `find_unguarded_protected_writes` never runs, and the report prints `0 unguarded-write finding(s)`. That reads as "no bypasses found" rather than "the file was never opened" — which is the failure mode this repo cares most about.

I built a bundle laid out the way `unzip` actually produces one. Through the CLI as written:

```
1 bundle(s): 0 with a transcript (0 truncated), 1 without.
  feedback-2026-08-01T10-00-00Z (platform=None): 0 tool calls, 0 Skill calls, 0 unguarded-write finding(s)
```

Same bundle, adapter pointed at the real path:

```
[{'index': 0, 'tool': 'mcp__genealogy__research_append',
  'required_skill': 'proof-conclusion', 'question_id': 'q_001'}]
```

Four edits, all in `guardrail_shadow_report.py`:

```
You wrote:      transcript = bundle_dir / "session-log.jsonl"
Change it to:   transcript = bundle_dir / "_feedback" / "session-log.jsonl"
```

```
You wrote:      if (child / "session-log.jsonl").exists() or (child / "research.json").exists():
Change it to:   if (child / "_feedback" / "session-log.jsonl").exists() or (child / "research.json").exists():
```

```
You wrote:      "research.json and/or session-log.jsonl. Bundles live OUTSIDE the repo "
Change it to:   "research.json and/or _feedback/session-log.jsonl. Bundles live OUTSIDE the repo "
```

```
You wrote:      `research.json` and/or `session-log.jsonl`). Platform is left `None` — the
Change it to:   `research.json` and/or `_feedback/session-log.jsonl`). Platform is left `None` — the
```

## The tests can't catch it, because they build the same wrong layout

`test_scan_feedback_bundle_runs_both_detectors_and_counts` writes the transcript to the bundle root, so it passes only by sharing the assumption. Add one test that builds the real layout. I ran it against your branch and it fails, which is what makes it worth having:

```python
def test_transcript_is_read_from_the_feedback_subdir(tmp_path):
    bundle = tmp_path / "feedback-2026-08-01T10-00-00Z"
    (bundle / "_feedback").mkdir(parents=True)
    _write_jsonl(bundle / "_feedback" / "session-log.jsonl", [
        _assistant([{"type": "tool_use", "id": "t", "name": "record_search", "input": {}}]),
    ])
    (bundle / "research.json").write_text("{}", encoding="utf-8")
    result = scan_feedback_bundle(bundle)
    assert result["has_transcript"] is True
    assert result["tool_call_count"] == 1
```

Worth knowing for next time: `tests/unit/test_setup_feedback_case.py:119` already asserts this layout. The answer was one file over in the same directory.

To be clear, `test_scan_feedback_bundle_cowork_no_transcript` is fine as it stands. A Cowork session really does carry no transcript, since the agent runs in Cowork's VM. Keep it.

## An unreadable research.json also reports as zero

```python
except (json.JSONDecodeError, OSError):
    research = None
```

`find_missing_mentor_verdicts(None)` returns `[]`, so the bundle prints `0 missing-mentor-verdict finding(s)` — same output as a clean bundle. This is reachable: `_redact_living` rewrites only `tree.gedcomx.json`, so `research.json` is appended as raw bytes and nothing validates it on the way into the bundle.

A missing `research.json` has the same shape, and the totals line divides by every scanned bundle whether or not it had one.

```
You wrote:      "session_ids": [],
Change it to:   "session_ids": [],
                "has_research": research_path.exists(),
                "research_unreadable": False,
```

```
You wrote:      except (json.JSONDecodeError, OSError):
                    research = None
Change it to:   except (json.JSONDecodeError, OSError):
                    research = None
                    out["research_unreadable"] = True
```

Then show `research_unreadable` on the per-bundle line next to `[truncated]`, and use `has_research` as the denominator for the mentor-verdict total.

## Two small things

The new module's docstring cites `e2e/orchestrator.py`'s "message loop (:1663-1669)" and "`apply_tool_result` :897". Neither resolves on this head — 1663 is abort/shutdown handling, the message loop is near 1728, and `apply_tool_result` is at 964. Drop the numbers; the symbol names are enough and won't drift.

The PR body doesn't use the template. **Unsure about** is the one I'd have leaned on hardest here.

## What I checked that's good

- `Skill` block shape matches live transcripts exactly: `{"name": "Skill", "input": {"skill": "...", "args": "..."}}`.
- Tool-name prefixes resolve — `owning_skills` goes through `bare_tool_name`, so `mcp__genealogy__research_append` matches.
- The truncation marker matches both producers.
- `encoding="utf-8"` everywhere, stdlib-only, no Agent SDK import, and the CLI's missing-directory path exits 1.
- The spec paragraph's claims hold up; `_redact_person` is where you say it is.

## Verification

Full pass, on the merge with `main` (the branch is 7 commits behind, clean merge):

- `eval/harness` pytest — 2258 passed, 2 skipped
- `apps/server` pytest — 209 passed, 1 skipped
- All required checks green on `228dab99`

The one initial failure was the documented fresh-worktree missing-engine-build case and cleared after `npm run build`.

I didn't run `vitest`/`turbo` locally — no TypeScript in the diff, and CI ran them green on this head. I have no real bundle to test against and none may enter the repo, so my reproduction used a hand-built bundle matching the documented layout.

Update the branch before merging.

Asimi1234:
Reviewed on this head (0 behind main): Part A machinery is clean and correctly scoped — adapter tests 8 pass, adapter+shadow 41 pass, and I independently mutation-checked the #999 is_error pin (dropping bool() so a missing key becomes None fails the join test; restored). Verified the claims worth checking: stdlib/no-SDK holds transitively (adapter → subagent_capture → json/pathlib/typing only), the two tree-reading detectors are genuinely excluded (only named in a comment, never called), truncated bundles are bucketed out of the attributable total, and no bundle data is committed. No LEAD surface, no blocking findings — needs a 2nd approval.

Asimi1234:
Reviewed on this head. Richard's blocking bug is real — I verified it: the scanner reads `<bundle>/session-log.jsonl`, but both producers (`feedback.py:464`, `feedback.ts:369`) and the spec (`feedback-case-spec.md §2.2`) put it at `_feedback/session-log.jsonl`, so `has_transcript` is False on every real bundle and it reports "0 unguarded-write findings" without ever opening the file — the exact silent-zero this feature exists to catch, and the tests pass only because they build the same wrong layout. The adapter itself and the #999 is_error join are correct (adapter 8 / adapter+shadow 41 green, mutation-confirmed). Blocking on the path fix — deferring to Richard's CHANGES_REQUESTED.

francis-2008-happy:
Fixed in 77fc876f — thanks, this was the right catch and it's exactly the silent-zero the feature is supposed to prevent.

- **Transcript path** → reads `_feedback/session-log.jsonl` now, in both `scan_feedback_bundle` and `scan_feedback_dir` (+ the CLI help + docstrings).
- **Added your test** `test_transcript_is_read_from_the_feedback_subdir`, and moved the existing scan test onto the real `_feedback/` layout so it stops passing by sharing the wrong assumption. Mutation-checked: reverting the path to the bundle root fails both.
- **Unreadable/missing research.json** → added `has_research` + `research_unreadable`, shown on the per-bundle line (`[research unreadable]` / `[no research.json]`), and the mentor-verdict total now divides by readable-research bundles only. New test covers both the invalid-JSON and missing cases.
- Dropped the stale `orchestrator.py` line numbers from the adapter docstring (kept the symbol names).

43 tests pass. And since the path bug is exactly about the real unpack layout, I re-ran the report over the real `_feedback/` layout to confirm the #1054 numbers still hold — **18 unguarded-write / 3 missing-mentor, unchanged**.

chesworthrm (changes requested):
Second round, on `f3eed623`. The three things I blocked on last time are properly fixed — I re-read `77fc876f` and proved the detector now fires end to end over a real-layout bundle, so the path fix is real and the test you added does catch a revert. Thank you for moving the existing scan test onto the `_feedback/` layout too; that was the part that mattered.

Blocking again, on one claim and on three requirements from the issue that aren't delivered yet.

## The 18 / 3 number can't be what the commit message says it is

> Re-ran over the real bundle layout: totals unchanged (18 unguarded / 3 mentor).

Before `77fc876f` the code read `bundle_dir / "session-log.jsonl"`. `setup-feedback-case.sh:82` unpacks with plain `unzip -q "$ZIP_PATH" -d "$DEST_DIR"`, and I round-tripped a bundle through `zip`/`unzip -d` to confirm the log always lands at `out/_feedback/session-log.jsonl`. I also ran the pre-fix path against real-layout bundles — `exists=False` on every one.

So before the fix the transcript was never opened and the unguarded-write total was necessarily **0**. It cannot have been 18 both before and after. The mentor half could legitimately be unchanged, since that one reads `research.json` at the bundle root and the path fix didn't touch it — but the unguarded half can't be.

This matters more than a wording slip because #1054 says "Do not start until #1558 reports a number," and that is the number.

```
You wrote:      Re-ran over the real bundle layout: totals unchanged (18 unguarded / 3 mentor).
Change it to:   Not yet measured over real bundles — Part A is machinery only.
                Part B reports the counts once the Drive link is in hand.
```

If you have run it over real bundles, say that instead and record how many, which platform each was, and how they were unpacked.

## Issue #1558 item 3 asks for four things per session; the report gives two

> Per session, report: platform (`web` / `darwin` / Cowork), whether the transcript was truncated, tool-call count, `Skill`-call count, and the raw violation records. Then totals with their denominator. **Name every bundle you could not adapt instead of dropping it.**

### The raw violation records are computed, then thrown away

They exist — I pulled one out of `scan_feedback_bundle`:

```
[{'index': 0, 'tool': 'mcp__genealogy__research_append',
  'required_skill': 'proof-conclusion', 'question_id': 'q_001'}]
```

`format_feedback_report` prints only `len(...)`, so a triager gets a count with no way to see which writes. The module already has the formatter — `format_detail`, line 112. It needs one key, the same one `scan_one` adds:

```
You wrote:      f"{len(r['unguarded_writes'])} unguarded-write finding(s), "
Change it to:   f"{len(r['unguarded_writes'])} unguarded-write finding(s), "
```

then after the per-bundle line:

```
Add:            for v in r["unguarded_writes"]:
                    lines.append(format_detail([{**v, "fixture": r["bundle"]}]))
```

I tested it. Called bare it raises `KeyError: 'fixture'`; with the enrichment it renders:

```
  bundle-unguarded    idx=0  tool=mcp__genealogy__research_append needs=proof-conclusion  q=q_001
```

### It can't name a bundle it failed to adapt

Careful here, because the obvious fix is wrong — I tried it. Flagging "`tool_calls` is empty" false-flags a real session. I built a well-formed transcript of a session that simply made no tool calls, and it produced output identical to an unparseable one:

```
quiet-session (platform=None): 0 tool calls, 0 Skill calls, 0 unguarded-write finding(s)
Totals: ... 0 across 1 attributable transcript(s)
```

An empty `tool_calls` can't separate the two. How many records the adapter recognised can:

```
shape-mismatch   lines=1  records_with_message_content=0   <- could not adapt
quiet-session    lines=2  records_with_message_content=2   <- adapted fine, just quiet
unguarded        lines=2  records_with_message_content=2
```

So the adapter has to return it:

```
You wrote:      return {
                    "tool_calls": tool_calls,
Change it to:   return {
                    # Records the adapter actually recognised. Zero-with-lines means the
                    # transcript could not be adapted, which #1558 item 3 requires naming —
                    # distinct from a session that simply made no tool calls.
                    "adapted_records": adapted_records,
                    "tool_calls": tool_calls,
```

Increment it wherever `content` passes the `isinstance(content, list)` check, then show `[could not adapt]` on the per-bundle line and drop those from the attributable denominator — the same shape you already built for `research_unreadable`. That pattern was the right instinct; the transcript side just needs it too.

### Platform can't be set from the CLI

`scan_feedback_bundle` takes `platform`, `scan_feedback_dir` never passes it, and there's no flag. Every line in every run I did printed `platform=None`, and the totals combine across platforms — against item 3 and against the lead's ruling ("separate columns; never a combined number").

```
You wrote:      ap.add_argument(
                    "--feedback-dir",
Change it to:   ap.add_argument(
                    "--platforms",
                    help="comma-separated <bundle-dir>=<platform> pairs from each "
                         "feedback issue's `Platform:` line (e.g. feedback-2026-08-01=web).",
                )
                ap.add_argument(
                    "--feedback-dir",
```

## Scanning `~/feedback/<slug>/` reads a research.json that triage has rewritten

Your docstrings name `make feedback-case → ~/feedback/<slug>/` as the input. That directory is git-inited with an `imported` baseline, and `make feedback-reset` exists because — the Makefile's words — "the agent mutates the case as it works."

I built a case dir the way `setup-feedback-case.sh` does and ran one simulated triage pass that wrote the mentor verdict:

```
what the scanner reads today (working tree):   []
what the tester actually submitted (baseline): ["proof_summaries entry 'ps_001' ... no
                                                 'proof-critique' verdict on record"]
```

The finding disappears. The transcript half is safe — it's a static file from the zip. Only `research.json` is exposed.

```
You wrote:      research = json.loads(research_path.read_text(encoding="utf-8"))
Change it to:   # A case dir is git-inited with an `imported` baseline and mutated by triage
                # (`make feedback-reset`), so the working tree may be a replay.
                research = json.loads(_submitted_research(bundle_dir, research_path))
```

with a helper that runs `git -C <bundle_dir> show HEAD:research.json` when `<bundle_dir>/.git` exists and falls back to the file otherwise. `git add .` runs before the symlinks in that script, so `research.json` is in the baseline — I confirmed the command recovers the submitted state exactly.

## Smaller

**No tests on the formatter.** Repo-wide grep: `format_feedback_report` and `scan_feedback_dir` are referenced only by the module defining them. No test, no Makefile target, no CI. That's why all four findings above sit in the same place — the denominator logic that produces the #1054 numbers is the one part nothing exercises. Also `make e2e-guardrail-shadow` wraps this module but doesn't pass `--feedback-dir`, so the new flag isn't reachable the way the repo's other commands are.

**The `is_error` premise is false, though the code is right.**

```
You wrote:      # `is_error` is present on a tool_result ONLY when true, so a
Change it to:   # `is_error` may be absent, explicitly false, or true — measured across
                # real transcripts. bool() reads all three correctly.
```

Across 252 real Claude Code transcripts — the files the desktop producer bundles — the key is explicitly `false` **6886** times, absent 1488, `true` 362. `bool(block.get("is_error"))` handles all three, so nothing is broken. But the premise comes verbatim from #1558's body, #1054 is planning a ledger that stores this field, and your test covers only absent and `true`, never the most common case. Add the third:

```
You wrote:      {"type": "tool_result", "tool_use_id": "bad", "content": "boom", "is_error": True},
Change it to:   {"type": "tool_result", "tool_use_id": "bad", "content": "boom", "is_error": True},
                {"type": "tool_result", "tool_use_id": "plain", "content": "ok", "is_error": False},
```

Worth correcting the issue body too, so it stops propagating.

**PR body still isn't the template.** No **Acceptance check** line. "43 tests pass" wouldn't have caught any of the above, which is the argument for the field.

## What's good

- The adapter itself is right, and I checked its assumptions rather than trusting them: `sessionId`, `type: user|assistant`, `message.content` blocks, `Skill` as `{"name": "Skill", "input": {"skill", "args"}}`, and MCP names resolving through `bare_tool_name` — all confirmed against 252 real transcripts and against `apps/server/tests/test_feedback.py:119-127`.
- Both producers really do write `_feedback/session-log.jsonl` and both really do emit `_truncation_note`; I read `feedback.py:464` and `feedback.ts:369` and the cap logic on each side.
- `find_missing_mentor_verdicts(None)` is safe — it does `research or {}`, so the unreadable path can't crash.
- Stdlib-only and no Agent SDK import, checked by importing the module and diffing `sys.modules`.
- The spec paragraph's claims hold up.

## Verification

Full pass, on the test-merge with `main` (23 behind, clean merge):

- `eval/harness` pytest — 2533 passed, 2 skipped
- engine `vitest run tests/packaging` (doc-link/ADR lints) — 505 passed
- Built four synthetic bundles in the real `_feedback/` layout and ran the CLI end to end: guarded → 0, unguarded → 1, truncated correctly bucketed out of the total. The detector genuinely fires now.
- All required checks green on `f3eed623`.

I have no real bundle and none may enter the repo, so every reproduction above used synthetic bundles built to the documented layout. That is also why the first finding is a question about provenance rather than a settled number. One thing I could not check at all: whether the hosted sandbox's SDK transcript matches local Claude Code's line shape. The server's own tests model the same shape, which is corroboration but not proof — and that unverified assumption is exactly what the "could not adapt" flag would make visible.

Update the branch before merging.

chesworthrm (changes requested):
Third round, on `fd1ac52d`. Every one of the nine things I raised across the first two rounds is properly fixed, and I checked the fixes rather than the descriptions of them — I reverted the `_feedback/` path and watched five tests die, dropped the raw-record print and watched its test die, folded `could_not_adapt` back into the denominator and watched that one die. The `18 unguarded / 3 mentor` claim is gone from every tracked file and from the commit message. The git-baseline read is right for the reason you give: `setup-feedback-case.sh` makes exactly one `imported` commit and `reset-feedback-case.sh` uses `checkout -- .` with no commit of its own, so `HEAD:research.json` really is what the tester submitted.

I also pointed the adapter at a real 5,354-line Claude Code transcript. 1,009 tool calls, `is_error` joined on every one, single `sessionId`, no duplicate `tool_use` ids. The adapter handles real data, not just the synthetic kind.

Blocking on two things. Neither is a bug in what you wrote — one is a claim the report makes that isn't true, and one is a crash.

## The report will print 0 for `proof_summaries` no matter what happened

Main gained a skill/agent split for `proof-conclusion` on 2026-08-19, the day after you opened this. It changes who makes the write, and it takes that write out of the file a bundle carries.

`skills/proof-conclusion/SKILL.md` now says the agent owns "the narrative, the `proof_summaries` write, and the tree encoding at tier ≥ probable," and "do not write `proof_summaries` yourself." The hook enforces it: `OWNED_SECTIONS = {"proof_summaries": "proof-conclusion"}` in `plugin/hooks/guard_project_files.py`, which sets `caller = ""` when `agent_id` is absent and denies.

An agent's tool calls are not in the bundle. I measured this rather than assuming it — one session on my machine has 13 subagent transcripts under `<sid>/subagents/` holding 618 tool calls, and **zero** of those `tool_use` ids appear in the parent `{sid}.jsonl`. That parent file is all either producer ships: `feedback.py` reads `{sid}.jsonl` and its fallback loop skips anything that `is_dir`, and `readSessionLog` does a non-recursive `readdir` and takes the newest `.jsonl`.

So the write that lands is invisible. And the attempt you *could* see — a main-thread write — is denied by the hook, which the transcript records as `is_error: true`. I checked what a deny looks like in real logs: eight deny-shaped tool_results, all `is_error: True`. Your detector's first line is `if entry.get("is_error") is True: continue`.

Both routes closed. The `proof_summaries` count is 0 unconditionally.

The shape-mismatch warning can't cover for this, because it only fires when `unguarded_writes` is non-empty — it cannot flag an arm that produces nothing at all.

This matters because #1054 is waiting on this number, and the hook's own docstring says 52 of 142 committed runs wrote a proof summary without launching the skill that owns it. That is the population this arm was built to count in the field, and over bundles it now reports silence.

Nothing needs to change in the logic. The report and the spec have to stop implying the zero is a measurement.

```
You wrote:      f"Corpus is small and self-selected — a signal, not a rate. "
                f"e2e baseline comparison pending issue #1484."
Change it to:   f"Corpus is small and self-selected — a signal, not a rate. "
                f"The proof_summaries arm reports 0 by construction, not by measurement: "
                f"since the 2026-08-19 split that write is made inside the proof-conclusion "
                f"AGENT, whose transcript is not in the bundle, and a main-thread attempt is "
                f"hook-denied and so skipped as is_error. "
                f"e2e baseline comparison pending issue #1484."
```

The tree writes are a narrower case, and worth keeping separate — the hook covers `research_append` only, so a main-thread `tree_edit` setting `primary: true` still fires. Those arms are blind to the agent route, not to everything.

```
You wrote:        Only `find_unguarded_protected_writes` (transcript-only) and
                  `find_missing_mentor_verdicts` (reads `research.json` alone) are valid over a
                  bundle; the adapter and report live in `eval/harness/e2e/`
                  (`feedback_transcript_adapter.py`, `guardrail_shadow_report.py`).
Change it to:     Only `find_unguarded_protected_writes` (transcript-only) and
                  `find_missing_mentor_verdicts` (reads `research.json` alone) are valid over a
                  bundle; the adapter and report live in `eval/harness/e2e/`
                  (`feedback_transcript_adapter.py`, `guardrail_shadow_report.py`).
                  **Two of `find_unguarded_protected_writes`' owner arms are blind over a bundle,
                  and one of those completely.** A bundle carries only the main session's
                  `{sid}.jsonl`, never the `subagents/agent-*.jsonl` beside it (measured: 618
                  subagent tool calls, 0 present in the parent). Since the 2026-08-19 split the
                  `proof_summaries` write is made inside the `proof-conclusion` agent, so it is
                  invisible — and a main-thread attempt is denied by `OWNED_SECTIONS` in
                  `plugin/hooks/guard_project_files.py`, which the transcript records as
                  `is_error: true`, which the detector skips. That arm's count is 0 whatever
                  happened. The `tree_edit`/`tree_correct` arms are blind only to the agent route:
                  the hook covers `research_append` alone, so a main-thread `primary: true` or
                  `ParentChild`/`Couple` write still fires. Recovering the agent route means
                  bundling the subagent transcripts, a change to `apps/server/app/feedback.py`.
```

## One malformed bundle takes down the whole scan

`research_unreadable` catches `JSONDecodeError` and `OSError`, but not "parsed fine, wrong type." A `research.json` holding a JSON array is truthy, so `research or {}` passes the list straight through:

```
File ".../harness/skill_invocation.py", line 518, in find_missing_mentor_verdicts
    questions = research.get("questions") if isinstance(research.get("questions"), list) else []
AttributeError: 'list' object has no attribute 'get'
```

Every other bundle's result is lost with it. This is reachable: `_redact_living` touches only `tree.gedcomx.json` and says so in its own docstring — "this is a privacy filter, not a validator" — so `research.json` goes into the bundle as raw bytes, and on desktop the folder is the tester's own.

```
You wrote:          research = json.loads(_submitted_research(bundle_dir, research_path))
                except (json.JSONDecodeError, OSError):
Change it to:       research = json.loads(_submitted_research(bundle_dir, research_path))
                    if not isinstance(research, dict):
                        raise json.JSONDecodeError("research.json is not a JSON object", "", 0)
                except (json.JSONDecodeError, OSError):
```

I ran this. The bad bundle gets tagged `[research unreadable]`, drops out of the denominator, and your 48 tests stay green. Add the case to `test_unreadable_research_json_is_flagged_not_counted_as_clean` — write `"[]"` and assert `research_unreadable is True`.

## Two smaller ones

**The window constant can drift from the one it claims to equal.** `_FEEDBACK_WINDOW = 40  # == GUARDRAIL_SHADOW_WINDOW` — that constant is at `orchestrator.py:170`. Copying the literal instead of importing it is the right call, since importing the orchestrator pulls in the Agent SDK. But nothing holds them together, and your own footer promises a comparison against the e2e baseline that a silent divergence would quietly invalidate. One text assertion closes it, no import needed. I wrote it, watched it pass, changed 40 to 60, watched it fail:

```
Add:            def test_feedback_window_matches_the_e2e_shadow_window():
                    """Importing the orchestrator would pull in the Agent SDK, so this reads its
                    source instead. The #1484 comparison is only meaningful at one window."""
                    from e2e.guardrail_shadow_report import _FEEDBACK_WINDOW
                    src = Path(__file__).parents[2] / "e2e" / "orchestrator.py"
                    assert f"GUARDRAIL_SHADOW_WINDOW = {_FEEDBACK_WINDOW}\n" in src.read_text(encoding="utf-8")
```

**`--feedback-dir` still isn't reachable from the Makefile** — my round-2 note, still open. It's the one reader in this module a person can't run the way they run every other one, and Part B is exactly when someone will want to.

```
You wrote:      cd eval/harness && uv run python -m e2e.guardrail_shadow_report $(if $(TEST),--test $(TEST),)
Change it to:   cd eval/harness && uv run python -m e2e.guardrail_shadow_report $(if $(FEEDBACK_DIR),--feedback-dir $(FEEDBACK_DIR),) $(if $(PLATFORMS),--platforms $(PLATFORMS),) $(if $(TEST),--test $(TEST),)
```

I applied it and ran `make e2e-guardrail-shadow FEEDBACK_DIR=… PLATFORMS=b1=web`. It prints the finding with `platform=web`.

**PR body still isn't the template.** The **Acceptance check** landed and it's a good one — I ran it and it reproduces exactly. But **Unsure about** is missing again, and it is precisely where the agent-split problem above belonged: the split landed on main two days before your round-2 commit, and that field is the one that asks you to look.

## What's good

- The adapter is correct on real data, not just synthetic. Real transcript, 1,009 calls, clean.
- `_feedback/` path, `could_not_adapt`, the raw records, `--platforms`, `has_research`/`research_unreadable`, the git baseline, the `is_error` comment and its third test case — all fixed, all with a test that dies when reverted. I checked each by reverting it.
- Both producers really do write `_feedback/session-log.jsonl` and both really do emit `_truncation_note`. Re-read both.
- `encoding="utf-8"` everywhere including the tests, stdlib-only, no Agent SDK import, `subprocess.run` with a list and no shell.
- The formatter tests are real guards, not coverage: all three mutations you claim to have checked do kill their tests.

## Verification

Full pass, on the test-merge with `main` (1 behind, clean merge):

- `eval/harness` pytest — 2585 passed, 2 skipped
- engine `vitest run tests/packaging` — 552 passed
- Acceptance check reproduced: guarded → 0, unguarded → `1 unguarded-write finding(s)` with the raw record row
- All required checks green on `fd1ac52d`

Still no real bundle, and none may enter the repo, so every bundle reproduction was synthetic in the documented layout. What I could check this round that I couldn't before: the transcript-shape assumptions, against real Claude Code logs on this machine. What I still can't check at all is whether the hosted sandbox's SDK writes the same line shape as local Claude Code — `could_not_adapt` is what would make that visible, which is why it was worth adding.

Update the branch before merging.

francis-2008-happy:
All in as of `6030cc17`. Thanks — the agent-split catch is the one I'd never have found, and it's the kind that would have shipped a 0 straight into #1054 as if it meant something.

**The `proof_summaries` arm reports 0 by construction.** I verified the chain rather than taking it: `OWNED_SECTIONS = {"proof_summaries": "proof-conclusion"}` at `guard_project_files.py:96`; `feedback.py:264` reads `{sid}.jsonl` with a fallback loop that skips `is_dir`, and `readSessionLog` does a non-recursive `readdir` — so no subagent transcript is ever bundled; and the detector's `if entry.get("is_error") is True: continue` closes the main-thread route. Report footer and spec § "Options set aside" both say so now, and both separate it from the `tree_edit`/`tree_correct` arms, which are blind only to the agent route.

**Non-dict `research.json`.** Guarded with the `isinstance` check. One correction to your suggestion, and it changes the test: `[]` never crashed. `research or {}` absorbs it, and `null` too — only a *truthy* non-dict reaches `.get()`. So `"[]"` alone would have passed before the fix and proved nothing. The test pins a non-empty list, a string and a number as the crash cases and keeps `[]` as a separate falsy one. Removing the guard dies with exactly the `AttributeError` at `skill_invocation.py:555` you quoted.

**Window drift.** Your test, verbatim — it reads the orchestrator's source as text, no import, no SDK. Note the constant is at `orchestrator.py:173`, not 170; the text match doesn't care, which is why it's the right shape.

**Makefile.** `FEEDBACK_DIR=` and `PLATFORMS=` wired, and I ran it end to end: `make e2e-guardrail-shadow FEEDBACK_DIR=<dir> PLATFORMS=b1=web` prints the finding with `platform=web`, the raw record row, and the new footer.

**PR body** is on the template, with `Unsure about` filled in. You were right that it was the field this belonged in — the agent split is sitting in it now, along with the question I'd actually like your call on: whether labelling the blind arm is enough, or whether it should come out of the feedback path entirely. I kept it and labelled it, on the reasoning that a labelled zero survives someone bundling the subagent transcripts and an absent arm doesn't. Overrule me if you disagree.

**One thing you didn't ask for.** Mutation-checking this round showed the by-construction claim was held in place by nothing — I deleted the sentence and all 63 tests still passed. That's the same silent-zero shape the whole PR is about, so I added a guard: deletion fails, rewording doesn't.

| Mutation | Result |
|---|---|
| drop the `isinstance` guard | `test_unreadable_research_json...` dies with the AttributeError |
| `_FEEDBACK_WINDOW` 40 → 60 | `test_feedback_window_matches_the_e2e_shadow_window` dies |
| delete the by-construction footer | `test_report_footer_marks_the_proof_summaries_arm_as_by_construction` dies |
| reword that footer, same meaning | still passes |

**Verification.** `eval/harness` 2644 passed / 2 skipped / 9 failed — all 9 are `test_setup_feedback_case.py` on `unzip: command not found` on my machine, identical with my changes stashed. `apps/server` 209 passed. Engine `vitest run tests/packaging` 576 passed. All 13 checks green on `2e568c69`; `6030cc17` is that plus a main merge, 0 behind.

Still not verified, and still not verifiable here: whether the hosted sandbox's SDK writes the same line shape as local Claude Code. `could_not_adapt` is what surfaces it if not.

The `is_error` wording in #1558's body is still yours to edit — flagging it a third time only so it doesn't get lost.

clack391 (changes requested):
Fourth-round review on `6030cc17`, and first from me. chesworthrm's four round-3 items are all genuinely fixed and I checked behaviour rather than descriptions: I reverted the `isinstance` guard and watched the AttributeError come back, drove the window constant from both sides, emptied the footer sentence, and ran the Makefile path end to end. The `[]` exchange is settled and worth closing out: you are right that `[]` never crashed, because `research or {}` absorbs it, and chesworthrm's requested assertion holds anyway, because the guard you wrote is `not isinstance(...)` rather than a truthiness test. `test_feedback_transcript_adapter.py:220` pins `("empty-list", "[]")` with `research_unreadable is True`. Nobody was wrong and the shipped test is stronger than what was asked for. The voluntary fifth fix, noticing the round's central claim was held in place by no test, is the right instinct.

I am still requesting changes, because five things make the numbers this produces wrong or unreadable. The first was found independently by three separate blind passes over the diff, which is why I am treating it as blocking rather than as a wording note.

## 1. The footer's "0 BY CONSTRUCTION" is false for the corpus this tool exists to scan

`guardrail_shadow_report.py:927-931` prints, unconditionally, that the `proof_summaries` arm "reports 0 BY CONSTRUCTION, not by measurement", and `guardrail-enforcement-spec.md:1252` says it "must never be read as a measurement". Both mechanisms postdate every bundle:

- The split merged **2026-08-21T22:33Z** (`73b3d98e`, #1819). `git log --diff-filter=A packages/engine/plugin/agents/proof-conclusion.md` gives the same date.
- The **pre-split** skill made the write from the main thread. `git show 73b3d98e~1:packages/engine/plugin/skills/proof-conclusion/SKILL.md:157` is `research_append({ projectPath, section: "proof_summaries", op: "append", entry })`, with no hook covering it.
- The newest `feedback`-labelled issue is **#1795, 2026-08-20T18:13Z**, about 28 hours before that merge. #1558 names the ten bundles from 2026-08-05 onward.

So over Part B's actual corpus that arm was fully open: main-thread writes were legal, un-denied, and present in `{sid}.jsonl`. A nonzero count there is a real measurement, and it is the population the hook's own docstring sizes at 52 of 142 committed runs. The footer tells the reader of #1054 to discard exactly the number #1054 exists to produce.

Your own acceptance check demonstrates it in one output. Running the shipped Makefile path prints a nonzero `needs=proof-conclusion` row two lines above the sentence saying that count is always zero. And `test_report_footer_marks_the_proof_summaries_arm_as_by_construction` now makes deleting or qualifying that sentence a test failure, so the guard added in round 3 defends the wrong claim.

The fix is not to delete the sentence. The date is already in the data you read: the bundle dir is `feedback-<ISO>Z` and `_feedback/feedback.json` carries the submission timestamp. Qualify the arm against the bundle's date versus the split, and print each bundle's date. Note that `docs/architecture.md` §9.4 point 2 means even a post-split bundle may have run a pre-split plugin, since a deploy does not ship the sandbox image, so the qualifier has to be "unknown plugin era" rather than a clean cutoff.

## 2. A third arm went by-construction-blind inside this branch's own history, and carries no label

`guardrail-enforcement-spec.md:1244` says "Two of `find_unguarded_protected_writes`' owner arms are blind over a bundle". It is three, two of them completely. `research-exhaustiveness` became a skill-agent pair in `c78efb0b` on **2026-08-23**, an ancestor of this head, which added `OWNED_DECLARATIONS = {("questions", "exhaustive_declaration"): "research-exhaustiveness"}` at `guard_project_files.py:114`. `owning_skills` has the matching arm at `skill_invocation.py:125-127`. That arm is now in exactly the position `proof_summaries` is in, and the footer names only `proof_summaries`, so the second fully-blind zero ships with no label at all. That is the precise hazard round 3 existed to remove, reintroduced by a commit that landed after the last review.

## 3. `response_summary: None` manufactures violations for writes that never landed

`feedback_transcript_adapter.py:85` hardwires `"response_summary": None` on every entry. `did_not_land` is two clauses, and the second reads that field for `no_project` (issue #1695). Its docstring states the consequence outright: "a write that did not happen is counted as a landed protected write and manufactures a violation." Over a bundle that clause can never fire, so a `research_append` that returned `reason: "no_project"` with no `isError` becomes a counted bypass. This is deliberately the one failure that does not set `is_error` (`tool-result.ts:88`), and Phase 1b put twelve tools on that return shape. The data is already in hand: the `tool_result` block's `content`, which the adapter reads for `is_error` and then discards.

## 4. One non-UTF-8 bundle still destroys the whole scan, and the fix is 540 lines above it in the same file

This is round 3's blocker in a second shape. `guardrail_shadow_report.py:837` catches `(json.JSONDecodeError, OSError)`; `UnicodeDecodeError` is a `ValueError` and is in neither, so it escapes. Reproduced through the shipped CLI: cp1252 bytes in one bundle's `research.json` raised `UnicodeDecodeError: ... byte 0xf1` and the healthy sibling bundle produced nothing. Same gap on the transcript side, where there is no guard at all: `parse_jsonl` (`subagent_capture.py:98-103`) catches only `OSError`, and a smart quote in `session-log.jsonl` killed the directory scan identically.

What makes this clear-cut rather than a judgment call is that the module already learned it. `_load_json` at `guardrail_shadow_report.py:282-296` does `except (json.JSONDecodeError, UnicodeDecodeError, OSError)` and its docstring explains why: "it is a `ValueError`, not an `OSError`, so a file with invalid UTF-8 would otherwise propagate." And `test_guardrail_shadow_report.py:755` is `test_replay_post_hoc_names_a_sidecar_with_invalid_utf8`, in the file your own acceptance command runs, with the wrong-type test directly above it. The new path reproduces both known bugs of the path beside it. Widening to `(ValueError, OSError)` covers both, since `JSONDecodeError` is a `ValueError`. The population is the Windows genealogist team, and CLAUDE.md's encoding section says this class "has bitten us repeatedly".

## 5. The single combined total contradicts the issue's lead ruling

#1558's body: "Report the two detectors and the platforms in separate columns; never a combined number." `format_feedback_report:915-925` tags each row with `platform=` and then sums one `unguarded_total` and one `mentor_total` across every bundle, folding `web`, `darwin` and unmapped `None` together. There is no per-platform aggregation anywhere.

## Should-fix

- **The CLI branch users run has no test.** `grep "main(\["` over the new 427-line test file returns nothing; the three functions are unit-tested, `main()`'s block at `:993-1005` is not. The sibling suite names this trap at `test_guardrail_shadow_report.py:799` ("The test that pins the DELIVERABLE. Every other test here calls `replay_post_hoc` directly, so all of them stay green while `main()` still prints only the provenance family"). I ran the matrix: `--feedback-dir` with `--test/--windows/--replay` gives byte-identical output because the branch returns before they are read, while `Makefile:595`'s new help line advertises them all on one target; `PLATFORMS=b1web` is silently dropped and prints `platform=None`; `PLATFORMS=typo=web` is silently ignored.
- **`--feedback-dir` pointed at a case dir prints a confident zero.** `scan_feedback_dir` only inspects immediate children, so passing `~/feedback/<slug>` (what `make feedback-case` prints, and what `make feedback-reset CASE=` takes) gives "0 bundle(s)" at exit 0, no warning. #1558 asks you to name every bundle you could not adapt.
- **`#1484` is cited as pending in the printed report** (`:932`) and in the test docstring. It closed COMPLETED on 2026-08-20, by PR #1775, and the recompute that closed it is the `--replay` path in this same module. Worth saying that chesworthrm's suggested replacement text carried the reference forward too.
- **Both new guards pass a shape they exist to catch.** I rewrote the footer to "is a genuine measurement of BY CONSTRUCTION behaviour, reporting 0 not by measurement error", keeping all three matched substrings while asserting the opposite, and the test passed. Pin the claim as one contiguous case-insensitive phrase. The window assertion also matches a raw substring, so it false-alarms on a trailing comment or a type annotation on `orchestrator.py:173` while the values still agree.
- **Docs the diff did not follow.** The module docstring still opens "over committed e2e runs", the argparse `description` at `:948` says the same, the Makefile recipe comment still says "Windowed to 14 days like every other reader", and `eval/CLAUDE.md:107` plus `harness/since_window.py:9` both list this target among readers that window and print the window. None is true on the new path.
- **Deferred work with no issue.** The spec says "Recovering the agent route means bundling the subagent transcripts, a change to `apps/server/app/feedback.py`" and nothing is filed. CLAUDE.md's ladder wants that named in both bodies in the same PR that defers it. Two arms that return 0 unconditionally are also a check that cannot fail, which is a `nothing-checks` item.

## Determination on your open question

You asked three times whether the blind arm should stay labelled or come out of the feedback path. Finding 1 answers it differently from either option: **keep the arm, and make the label conditional rather than removing either one.** Over the corpus you will actually scan the arm is live and its count is real, so removing it would delete a genuine measurement and an unconditional disclaimer would discard one. Report it against the bundle's submission date, with "plugin era unknown" where §9.4 point 2 applies. That also fixes finding 2 for free, because a date-conditional label generalises to the third arm instead of needing a second hardcoded sentence.

## What I ran

Review worktree at `6030cc17`. Your acceptance command reproduces exactly: 64 passed. Engine `vitest run tests/packaging`: 576, as you said. Full harness after supplying the missing engine build: 2656 passed, 2 skipped, 0 failed. The harness's first pass showed one failure, `test_mock_mcp.py::test_successful_call_carries_no_is_error`, which is a fresh worktree with no compiled engine; six sibling tests skip on that condition and that one fails instead, which is a repo gap rather than anything here. I probed the research-shape guard with eleven shapes your fixture does not use, including `null`, `true`, `0`, `""`, BOM-prefixed, whitespace-only and `research.json` as a directory, and it held on all of them.

One thing for whoever lands first: #1856 collides with this at line level. Both rewrite `Makefile:595`, and both insert at the same anchor in `guardrail_shadow_report.py` right after `format_warnings_unchecked` while both edit `main()`.

francis-2008-happy:
All five plus every should-fix are in as of `bbdf21ce`, and the #1856 conflict is resolved. Thank you for finding 1 — three blind passes converging on it was the right call, because I had not only shipped a false claim, I had added a test in the previous round that made removing it a failure.

**Finding 1.** I verified your dates rather than taking them: split merged `73b3d98e` 2026-08-21T22:33Z; pre-split `SKILL.md:157` is a main-thread `research_append` on `proof_summaries` with no hook; newest `feedback` issue #1795 at 2026-08-20T18:13Z. Every real bundle is pre-split, so that arm was open and a count is a real measurement. Arms are now labelled per bundle from `submitted_at`, printed on every row, with "plugin era unknown" on the other side rather than a clean cutoff.

**Finding 2.** Confirmed by file date — `research-exhaustiveness` agent added 2026-08-23. I checked all four `owning_skills` arms rather than just the two you named: `person-evidence` and `conflict-resolution` have no agent file, so they stay main-thread and visible. Exactly two are agent-owned, and the date-conditional label covers both without a second hardcoded sentence.

**Finding 3.** Populated from the `tool_result` content. Verified end to end with the envelope shape (`no_project`, no `isError`): `did_not_land` now True, violations 0 instead of 1.

**Finding 4.** Research side widened to `(ValueError, OSError)` matching `_load_json`; transcript side gets a new `[transcript unreadable]` tag. I caught it at the call site rather than widening the shared `parse_jsonl` — that would hand `subagent_capture`'s own caller `[]` instead of raising, and a tag is what #1558 item 3 asks for anyway.

**Finding 5.** Per-platform totals. One thing worth flagging: `platform` **is** in the bundle — both producers write it into `_feedback/feedback.json` (`"web"` server-side, `process.platform` on desktop). It now defaults from there, with `--platforms` as the override. That assumption has been carried since round 2 and it was wrong.

**Should-fixes.** All five `main()` tests added, and they found what you predicted plus the case-dir zero — corpus flags, a `=`-less `PLATFORMS` entry, and an unmatched name now all warn on stderr. Window guard parsed by regex; I break-tested that 40→60 fails while a trailing comment and a type annotation both pass, which is the false-alarm you flagged. `#1484` citation replaced with `--replay`. Docstring, argparse description, Makefile comment, `eval/CLAUDE.md` and `since_window.py` all now say this module has two modes and only one is windowed. Issue #1880 filed with `nothing-checks`.

**Your guard-inversion attack.** You were right and it was the more useful half of that note. Pinned as contiguous phrases now; I re-ran your rewrite against it and it fails.

**Two beyond the review.** `response_summary` had no guard — reverting it passed all 21 tests, the same unheld-claim shape as round 3, so I added one. And `format_feedback_report` raised `KeyError` on a result dict lacking the new fields; made tolerant.

**On the conflict.** #1856 landed 47 minutes after your review, exactly where you said. Both collisions resolved as unions. One trap worth recording: the naive union left `format_unnamed_delegate` without its `return`, because `return "\n".join(lines)` is a shared trailing line that attached to my block instead of yours. Three of your #1856 tests caught it.

**Two open questions in `Unsure about`,** since that field earned its place last round: whether the four-line era label belongs in the report or only in the spec, and whether `_AGENT_SPLIT_DATES` should be derived rather than a hand-maintained two-entry map that a third skill-agent pair would silently miss. I took the report and the hardcoded map; both are yours to overrule.

**Verification.** `eval/harness` 2690 passed (9 pre-existing `unzip` failures, identical stashed). `apps/server` 209. Engine packaging 576 — including the doc-links ratchet, which caught my spec at 21 refs against baseline 16; I kept every fact and dropped the ticket numbers. All 13 checks green on `bbdf21ce`, 0 behind main.

clack391 (changes requested):
Round-5 review on `0b530dc2`. Strong turnaround. All five of my round-4 findings and every should-fix are properly in, and I checked behaviour rather than the descriptions: each fix has a test that dies when I revert it (the date-conditional arm labels, the `research-exhaustiveness` arm, `response_summary` feeding `did_not_land`'s no-project clause, the `(ValueError, OSError)` catch on both sides, and the per-platform totals). The `main()` tests, the regex window guard, the `#1484` swap, the five-surface doc sync, and issue #1880 are all there too. The split dates and commits check out exactly.

Two things still block, and one test to add. All three sit in one place: the round-4 UTF-8 fix stopped the crash, but the "we could not read this bundle" path is not carried through to the counts, and one sibling of it still crashes. Because these are the numbers #1054 is waiting on, a wrong count here is the failure this whole PR exists to prevent.

## 1. An un-decodable transcript reports a silent zero and inflates both denominators

When a transcript is invalid UTF-8, `adapt_bundle_transcript` raises, `scan_feedback_bundle` sets `transcript_unreadable=True`, and then returns early:

```python
        if adapted is None:
            return out
```

That `return` is above the `if research_path.exists():` block, so the mentor-verdict detector never runs for that bundle. But its `research.json` is a separate file and may be perfectly readable. I built a bundle with a broken-encoding transcript and a clean `research.json` that has a real missing-mentor finding: the report shows `0 missing-mentor-verdict finding(s)` and still counts the bundle in "N bundle(s) with a readable research.json". The finding is dropped and the denominator is inflated with a clean zero.

Your own comment two lines up says the intent is to "keep going, the same shape `research_unreadable` already has", but `research_unreadable` falls through to the mentor scan while this path returns. So the code does not yet do what the comment says.

The same gap hits the other detector's denominator. `attributable` excludes truncated and could-not-adapt bundles but not unreadable ones:

```
You wrote:      attributable = [
                    r for r in with_transcript if not r["truncated"] and not r.get("could_not_adapt")
                ]
Change it to:   attributable = [
                    r for r in with_transcript
                    if not r["truncated"] and not r.get("could_not_adapt") and not r.get("transcript_unreadable")
                ]
```

An unreadable transcript never ran the detector, so it is strictly more unadaptable than a truncated one, and it should be out of the denominator for the same reason. The per-platform `p_attr` line reads from `attributable`, so fixing that line fixes both.

For the mentor half, drop the early return and guard the transcript-derived block instead, so the function always reaches the research scan:

```python
        if adapted is not None:
            tool_calls = adapted["tool_calls"]
            ...            # the rest of the transcript-derived block, indented under this
            out["unguarded_writes"] = find_unguarded_protected_writes(tool_calls, window=window)

    if research_path.exists():
        ...
```

## 2. One malformed `tool_use.input` crashes the whole scan

```python
        "args": dict(block.get("input") or {}),
```

`input` is not type-checked before `dict(...)`. If one `tool_use` block carries an `input` that is a JSON number, bool, or a plain list (all valid JSONL), `dict(42)` raises `TypeError`, which is not in the `(ValueError, OSError)` you catch and is unguarded in `scan_feedback_dir`. It propagates out and takes the entire directory scan with it: I ran three bundles with one bad one and got zero results back, not two good ones and a flagged one. A string `input` is caught but mislabelled `transcript_unreadable` even though the file decoded fine, which then feeds problem 1.

This is the "one bad bundle loses every bundle" case you and Richard both hardened against for UTF-8 and for non-dict `research.json`. The adapter is the one untrusted-input path that has the same shape and was missed, and the hosted SDK's line shape is the thing your own "Unsure about" says you cannot verify here, so it is exactly the case that has to survive.

```
You wrote:      "args": dict(block.get("input") or {}),
Change it to:   "args": (lambda v: v if isinstance(v, dict) else {})(block.get("input")),
```

or the plainer two-line form:

```
                inp = block.get("input")
                ...
                "args": inp if isinstance(inp, dict) else {},
```

## 3. Add the test that retires this class

Both "unreadable" denominator exclusions have no test: dropping `and not r["research_unreadable"]` from `with_research`, or the transcript exclusion above, leaves all 28 tests green. The `could_not_adapt` exclusion is tested; its two "unreadable" twins are correct only by luck. One report-level test closes all of it: a bundle with an unreadable transcript and a readable `research.json` that has a finding, plus a bundle with an unreadable `research.json`, asserting the finding is still reported and that neither bundle lands in its detector's denominator. That single test guards problem 1 and both denominators, rather than three separate one-offs.

## Not blocking

The `--since ... do not apply` note prints on every `--feedback-dir` run, even when `--since` was not passed, because its default is a truthy cutoff date. `--windows` gets this right by comparing against its default; `--since` could do the same, or drop out of that list. Cosmetic only, the counts are unaffected.

Once 1 and 2 are fixed with the test in 3, I am happy. Nothing else in the diff is a concern, and the rest of the round was clean work. One heads-up unrelated to the fixes: my open PR #1872 also edits `guardrail-enforcement-spec.md` and `guardrail_shadow_report.py`, in different regions, so there is no conflict, but whichever of ours merges second should re-run the doc-links check on the spec's issue-ref count.

francis-2008-happy:
All three in `dd117fbe`, @clack391 — thanks for the precise diffs.

1. `scan_feedback_bundle` no longer early-returns on an unreadable/unadaptable transcript. The transcript-derived fields are now gated under `if adapted is not None:`, and the `research_path.exists()` block runs regardless, so a readable research.json still gets the mentor scan. Also added `and not r.get("transcript_unreadable")` to `attributable`.

2. `adapt_bundle_transcript` coerces a non-dict `input` to `{}` (`inp if isinstance(inp, dict) else {}`), so a number/list/bool block no longer raises `TypeError` and takes the scan down.

3. Added the report-level test you specified: bundle A (unreadable transcript + readable research.json with a mentor gap) + bundle B (unreadable research.json). It asserts the gap is still reported, A is out of the attributable denominator, and B is out of the readable-research denominator. Added a second adapter-level test for the non-dict input. I watched both fail on revert — reverting fix 1 drops A's mentor finding to 0, and reverting the `dict()` guard raises `TypeError`.

Verified: harness suite 2689 passed (the 10 failures are the usual environmental ones — no engine build / no `unzip` — in files this PR doesn't touch).

Left the `--since ... do not apply` line as the cosmetic you flagged non-blocking. Happy to fold it in if you'd rather it go now.

clack391 (approved):
Approving on `dd117fbe`. All three are properly in, and I checked behaviour rather than the descriptions, reverting each fix and watching the guard die:

1. `scan_feedback_bundle` no longer early-returns on an unreadable transcript, so a readable research.json still gets the mentor scan, and `attributable` now excludes `transcript_unreadable` so the unguarded denominator is not inflated. I rebuilt the bad-transcript-plus-findable-research case: the mentor finding is reported now, and the denominator reads "1 across 1" not "1 across 2".
2. The non-dict `input` coercion holds: number, list, bool, string, None and float all land as `{}`, and the three-bundle scan with one bad block returns all three instead of crashing.
3. The report-level test is a real guard. Reverting the un-nest drops bundle A's mentor finding to 0, reverting the `attributable` exclusion puts the unreadable bundle back in the denominator, and reverting the `dict()` guard raises `TypeError`. Each revert fails it on a distinct assertion. That is the generalizing test, so the whole silent-zero class is pinned rather than one shape.

Two small fold-ins for your next push, neither blocking, and you will want a push anyway for the update-branch:

- **The Totals exclusion breakdown still names two of three reasons.** The parenthetical is `(N truncated, M could not adapt, excluded)`, but `transcript_unreadable` is now a third exclusion from `attributable`, so a reader summing `attributable + truncated + could_not_adapt` no longer reaches the with-transcript count (I saw `1 + 1 + 1` against 4). The denominators are right and the bundle is tagged `[transcript unreadable]` on its own row, so this is transparency polish, not a wrong number. Adding the unreadable count to that parenthetical closes it, the same shape your code comment two lines up already describes.
- **The input-coercion test does not pin a string input.** The cases are number, list, bool, dict. The shipped code handles a string correctly because it is an allowlist (`isinstance(inp, dict)`), but a later refactor to a blocklist would pass the test and leak a bare string into `_iter_ops`, which is the crash class this fix exists to prevent. One more case, `"input": "s"` expecting `{}`, pins it.

Thanks for the precise work across the rounds. Two things still gate merge and are not mine to clear: chesworthrm's round-3 review is still open (a push does not dismiss it, so it needs his fresh read).

francis-2008-happy:
Both fold-ins are in `6d49fe8a`, @clack391, and merged `main` in the same push (branch was behind):

- Totals breakdown now reads `(N truncated, M could not adapt, K unreadable, excluded)`, so the excluded reasons sum to the with-transcript count again.
- The input-coercion test pins a bare-string input → `{}` explicitly, so an allowlist→blocklist refactor can't pass while leaking a string.

Suites green (adapter + shadow-report 81 passed). Thanks again for the careful passes. Only outstanding merge gate now is chesworthrm's round-3 review, which a push doesn't dismiss — flagging for his fresh read.

chesworthrm (approved):
Re-review on `6d49fe8`. Every item from my round-3 review is fixed, and I checked the fixes rather than the descriptions — reverted the `_feedback/` path and the arm-visibility label and watched the tests die, then restored. Approving.

A few things I verified beyond my old items:

- The transcript-only detector normalizes tool names via `bare_tool_name` (`rsplit("__", 1)[-1]`), so all three MCP spellings a real bundle can carry (`mcp__genealogy__…`, `mcp__remote-devices__Genealogy_Research__…`, `mcp__Genealogy_Research__…`) collapse to the bare name and still match. That was my main worry — a real bundle silently reporting 0 because its tool names are namespaced. It won't.
- `find_missing_mentor_verdicts(None)` is safe, and the encoding lint is happy with the new files.
- Full `eval/harness` suite is green here: 2707 passed, 2 skipped, 0 failed. The 9 failures in the PR body were `unzip: command not found` on your box, not the change.

Two notes, neither blocking:

**1. My approval doesn't clear the merge gate on its own, and clack391's is now stale.** clack391 approved `dd117fbe`; two commits have landed since (the `main` merge and `6d49fe8a`, which adds the `unreadable` count to the Totals line). `protect-main` has `dismiss_stale_reviews_on_push: false`, so that approval still counts even though it never read `6d49fe8a`. The delta is small and has its own passing tests, so nothing is wrong — but it'd be worth a quick re-approve on the current head. And CODEOWNERS still wants a `senior-genealogists` sign-off.

**2. Optional one-liner.** In the `--feedback-dir` branch you print a note for the corpus-only flags that don't apply (`--test/--windows/--since/--replay`), but `--detail` is left off that list, so `--feedback-dir --detail` swallows it silently — the same class you fixed. It's minor (not reachable via the Makefile target, only affects the stderr note), so take it or leave it:

```python
("since", args.since), ("replay", args.replay), ("detail", args.detail)) if v
```

Nice work — the per-bundle era labelling is the right call, and the mutation table made this easy to trust.

PR #1744, opened 2026-08-19
.gitattributes: normalize all text files to LF repo-wide

EdmondOware:
## Summary
- Adds `* text=auto eol=lf` to `.gitattributes` so line endings are deterministic across contributors and environments, instead of depending on whoever's local `core.autocrlf` last touched a file.
- Root cause, measured directly: `git ls-files --eol` at main `8994d4dd` returns 3811 `i/lf`, 38 `i/none`, 11 `i/-text` (binary) and exactly one `i/crlf` — `apps/electron/.editorconfig`, which itself declares `end_of_line = lf` — across 3861 tracked files. The 44 `i/lf w/crlf` entries under the existing `*.bat` rule are that rule working as intended, not drift. Files with no explicit rule at all fell back to whichever environment's `core.autocrlf` last wrote them: a Windows dev, CI/Linux, and the Cowork sandbox bridge, all writing to the same repo with different settings.
- `text=auto` does not renormalize a blob already committed with CRLF, even on re-add, so the one drifted file needed an explicit renormalize rather than "self-heals on next touch". Done in this PR: `apps/electron/.editorconfig` is renormalized to LF, matching its own `end_of_line = lf`, and picks up its missing trailing newline while touched. No repo-wide renormalization pass is needed beyond that single file.
- An unclaimed win found while verifying: `scripts/git-hooks/post-checkout` carried no attribute at all on main, so a Windows clone checked it out with CRLF and `shim.sh`'s `exec "$src"` died on the stray `\r`. The new blanket rule fixes this for fresh clones. An existing clone keeps its CRLF copy until a one-time re-checkout, added to `DEVELOPMENT.md` § "Git hooks" (johnmarkpeterbrown review — hit this live: `git worktree add` ran the stale hook and the worktree came up with no `node_modules`, silently). The `.gitattributes` comment claiming extensionless hooks were "covered separately, next to the hooks themselves" was false — nothing covered them before this rule — so it is corrected here too.
- `eol=lf`, not bare `text=auto`, is deliberate: it forces LF in the working tree on Windows too, which is what actually fixes the hook above. Confirmed with @DallanQ.
- Because that stance is now doctrine, the rule gets a guard: `packages/engine/mcp-server/tests/packaging/gitattributes.test.ts` asks `git check-attr` what the rules resolve to for a `.bat`, a `.sh`, the extensionless hook, an ordinary source file, and the pinned binaries. Rule **order** is the thing it protects: later rules win, so the blanket line moved below `*.bat` would take every Windows batch script to LF while changing no blob and reddening nothing. `.gitattributes` also joins `engine-tests.yml`'s path scope, or the guard would be dark on exactly the PR shape that can break it — this PR's own `vitest` job reported success with its checkout step skipped.
- `*.pdf` / `*.png` / `*.ico` / `*.icns` are pinned `binary`. `text=auto` samples only the first 8000 bytes for a NUL, and `eval/tests/e2e/*/provided-documents/` is a growing pile of scans that can be ASCII-clean well past that.
- `skill-guidance.test.ts`'s comment said `*.sh` was the only LF-pinned pattern, which the blanket rule makes false. The `\r\n` normalization in `contentHash` stays — an existing clone keeps its CRLF on disk, because a pull does not re-check-out — only the stated reason is updated.

## Review guidance
Start with the `.gitattributes` diff itself. Then the guard: read the cwd note in the test's header before the assertions — `git check-attr` resolves its pathspec relative to the current directory and vitest runs in `packages/engine/mcp-server`, so without `cwd: projectRoot` an anchored pattern reports `unspecified` and git still exits 0. Then the renormalize commit: it should touch only `apps/electron/.editorconfig`, and that file's diff should be line endings and the trailing newline only, no content changes. All of it is confirmable directly rather than on faith — see Test plan.

## Test plan
- [x] `git ls-files --eol | awk '{print $1}' | sort | uniq -c` → 0 `i/crlf` now (was 1 before the renormalize commit)
- [x] `git check-attr text eol -- scripts/windows/install.bat scripts/git-hooks/post-checkout packages/engine/mcp-server/src/index.ts` → `.bat` stays `eol: crlf`; `post-checkout` now resolves to `eol: lf` (unset on main); `.ts` resolves to `eol: lf`
- [x] Fresh clone of the branch: `git status --porcelain` empty; `od -c eval/Setup.bat` shows `\r \n`; `od -c scripts/git-hooks/post-checkout` shows `\n`; all 11 `i/-text` binaries byte-identical to their blobs
- [x] `npm test` in `packages/engine/mcp-server` → 104 files, 2453 tests pass
- [x] Guard proven to fail three ways before committing: blanket rule moved below `*.bat` → 8 red, `install.bat` and `Setup.bat` both going `eol: lf`; the four `binary` lines deleted → 5 red; a guarded path renamed away → 1 red on the existence assertion
- [x] `PATTERNS` checked against this PR's own file list (matches, so vitest runs) and against a `README.md` / `Makefile` control (no match, so unrelated PRs still pay one step)

## Follow-on issues
None.


.gitattributes
```diff
@@ -1,3 +1,10 @@
+# Normalize line endings for all text files regardless of any contributor's or
+# environment's core.autocrlf setting. Without this, whether a file is stored
+# as LF or CRLF depends on whoever last touched it — a real source of drift
+# across a Windows dev, a Linux CI/VM, and the Cowork sandbox bridge, all
+# writing to the same repo. text=auto lets git still skip binaries.
```
Gennecis, line 5: Small qualification on "text=auto lets git still skip binaries" — true today, and I verified it: all 11 tracked binaries (7 PDFs, 2 PNG, ICO, ICNS) already come back `i/-text` from `git ls-files --eol`, so nothing is at risk right now.

The gap is future files. `text=auto` decides "binary?" by looking for a NUL byte in only the **first 8000 bytes**, and a print-to-PDF or uncompressed scan can be pure ASCII well past that. `eval/tests/e2e/*/provided-documents/` is a growing pile of scanned evidence documents, and a silently mangled death certificate that still half-opens is a bad way to lose fixture provenance.

Non-blocking, but cheap insurance — please add after line 6:

```
# Belt and braces for scanned evidence documents: text=auto samples only the
# first 8000 bytes for a NUL, and an uncompressed/ASCII-heavy scan can slip
# past that and get mangled.
*.pdf binary
*.png binary
*.ico binary
*.icns binary
```

.gitattributes
```diff
@@ -1,3 +1,10 @@
...
+* text=auto eol=lf
```
Gennecis, line 11: This line is correctly placed **above** the `*.sh` and `*.bat` rules — later rules win in `.gitattributes`, so position is load-bearing. Nothing in CI guards it. I proved the failure rather than assuming it, by moving this line to the bottom of the same three rules:

```
$ git check-attr text eol -- win.bat
win.bat: eol: lf          <-- .bat now LF; every Windows batch script silently broken
```

Please add to the comment block above this line:

```
# Order matters: later rules win. This line MUST stay above the *.sh and *.bat
# rules below — moved under them it would force .bat to LF and silently break
# every Windows batch script.
```

A packaging test asserting `git check-attr` for `*.bat` / `*.sh` / `post-checkout` would be the real guard and would fit the `nothing-checks` register. I'd call that optional here — if you'd rather not carry it, name it under a **Follow-on issues** section. The comment isn't optional.

.gitattributes
```diff
@@ -1,3 +1,10 @@
...
+
 # Shell scripts must keep LF endings on every platform.
```
Gennecis, line 22: Anchoring here because GitHub won't take an inline comment outside the diff hunk — the text I'm asking you to change is **lines 21–22** of this file, at the end of this comment block:

```
# Extensionless scripts (git hooks) are covered separately, next to the hooks
# themselves.
```

That was never true. There is no `scripts/git-hooks/.gitattributes` and there never has been — `git log --all --diff-filter=ADR -- '**/.gitattributes'` returns nothing, and `ls scripts/git-hooks/` shows only `post-checkout` and `shim.sh`. The comment landed in `faec2d1c` (#692) promising coverage that was never delivered, which is why `post-checkout` sat with no attribute at all until your PR.

Your change makes it true for the first time, just by a different route. Please replace lines 21–22 with:

```
# Extensionless scripts (git hooks) get LF from the blanket rule above. There
# is no sibling .gitattributes next to them; an earlier version of this comment
# claimed there was, leaving scripts/git-hooks/post-checkout uncovered.
```

Gennecis (changes requested):
Thanks @EdmondOware — the mechanism here is right and I verified it end to end. I'm requesting changes on two things in the description rather than the code, plus one file the PR leaves broken.

**What I confirmed works, with evidence:**

- Ordering is correct. `git check-attr text eol -- scripts/windows/install.bat` on your branch returns `eol: crlf` — all 44 `.bat` files are unaffected, including the 22 outside `scripts/windows/`.
- Truly attributes-only. I cloned your branch fresh: `git status --porcelain` is empty, and every index blob is byte-identical to `main`.
- No eval re-run tax. `eval/harness/harness/snapshot.py:268` and `eval/app/lib/snapshot.ts:103` both strip CRLF before hashing, so no skill snapshot is invalidated.
- **An unclaimed win you should take credit for:** `scripts/git-hooks/post-checkout` had *no* attribute on `main` (`git ls-files --eol` → `attr/` empty), so on a Windows clone it checked out CRLF and `shim.sh`'s `exec "$src"` died on `#!/bin/sh\r` — the exact silent-worktree failure lines 16–19 call "not hypothetical". Your blanket rule fixes it. Please say so in the Summary.

---

###  Blocker 1 — "self-heals on next touch" is false, and it leaves one file stuck

Summary bullet 3 says *"existing already-drifted files normalize individually the next time each is touched."* That isn't how `text=auto` behaves. Git deliberately leaves a blob alone if it was **already committed with CRLF**, specifically to avoid churn — editing and re-adding it does not convert it.

I reproduced your exact situation in a throwaway repo:

```
$ git add editorconfig.txt          # file committed with CRLF, blanket rule active
$ git ls-files --eol editorconfig.txt
i/crlf  w/crlf  attr/text=auto eol=lf   editorconfig.txt      <-- STILL CRLF
$ git cat-file blob :editorconfig.txt | od -c | head -1
0000000  r  o  o  t  =  t  r  u  e  \r  \n ...

$ git add --renormalize editorconfig.txt
i/lf    w/crlf  attr/text=auto eol=lf   editorconfig.txt      <-- only this works
```

Scanning all 3,862 tracked files, exactly one is stored with CRLF:

```
i/crlf  w/crlf  attr/    apps/electron/.editorconfig
```

And its own contents read `end_of_line = lf`. I checked a fresh clone of this branch — it is still CRLF after your change. So on merge, the repo's only drifted file is the one whose job is to declare LF, and the PR says it will sort itself out.

**Please do, in this PR:**

```
git add --renormalize apps/electron/.editorconfig
git commit -m ".editorconfig: renormalize to LF, matching its own end_of_line = lf"
```

Zero runtime risk — that file is only read by editors. (Optional while you're there: it also lacks a trailing newline despite declaring `insert_final_newline = true`.)

**And replace Summary bullet 3 with:**

> Attributes-only for future writes. `text=auto` deliberately does **not** renormalize a blob already committed with CRLF — even on re-add — so those need an explicit `git add --renormalize`. A repo-wide scan (`git ls-files --eol`) found exactly one, `apps/electron/.editorconfig`, renormalized here. No repo-wide renormalization pass is needed.

---

###  Blocker 2 — the root-cause evidence is the wrong shape

Summary bullet 2 says *"working-tree copies had CRLF while committed blobs had LF."* That combination isn't drift — it's git working as designed, and it's what this repo's own `*.bat text eol=crlf` rule **deliberately produces**:

```
i/lf  w/crlf  attr/text eol=crlf    scripts/windows/install.bat
```

Real drift is the opposite shape, CRLF *in history*, and there's exactly one instance — Blocker 1. I think reading the intended state as the bug is what led to the "self-heals" conclusion, which is why this is worth correcting rather than letting stand.

**Please replace that sentence with the actual measurement:**

> `git ls-files --eol` across 3,862 tracked files: 3,812 `i/lf`, 38 `i/none`, 11 `i/-text` (binary), **1 `i/crlf`** (`apps/electron/.editorconfig`). The 44 `i/lf w/crlf` entries are the intended output of the existing `*.bat` rule, not drift.

---

###  Decision to table with @DallanQ before this merges

Not yours to answer, and not a blocker on your work — but please raise it with them rather than let it merge unremarked.

`eol=lf` is a stronger stance than `text=auto` alone. `text=auto` would still let each machine's `core.autocrlf` pick the working-tree ending; adding `eol=lf` overrides that and forces LF **in the working tree on Windows too**. That is precisely what fixes `post-checkout`, and modern Windows editors handle LF fine, so I support it — but it changes what the Windows-based genealogist team sees in their editors, and it is the kind of decision this repo writes down rather than absorbs silently.

Ask @DallanQ to confirm the stance, then record the answer in the PR body in one sentence, e.g.: *"`eol=lf` (not bare `text=auto`) is deliberate: it forces LF in Windows working trees, which is what covers the extensionless git hooks. Confirmed with @DallanQ."*

---

### Suggestion 6 — the Test plan can't fail, and the body is missing two sections

No line to anchor this to, so it's here. "Confirm `.gitattributes` parses" isn't a test — git doesn't reject malformed attribute lines, it silently ignores them. Both boxes are also unticked.

**Replace the Test plan with commands a reviewer can run:**

```
- [ ] `git ls-files --eol | awk '{print $1}' | sort | uniq -c`
      → 1 `i/crlf` before the renormalize commit, 0 after.
- [ ] `git check-attr text eol -- scripts/windows/install.bat scripts/git-hooks/post-checkout packages/engine/mcp-server/src/index.ts`
      → .bat `eol: crlf`; post-checkout `eol: lf` (unset on main); .ts `eol: lf`.
- [ ] Fresh clone of the branch → `git status --porcelain` is empty (proves attributes-only).
- [ ] Fresh clone → `od -c eval/Setup.bat | head -1` shows `\r \n`; `od -c scripts/git-hooks/post-checkout | head -1` shows `\n`.
```

I ran all four against a real fresh clone of your branch and they pass, with the `i/crlf` count sitting at 1 until Blocker 1 lands.

Also: the house format is four sections — **Summary, Review guidance, Test plan, Follow-on issues**. This PR has two.

Gennecis:
@EdmondOware — following up on the @DallanQ item with something I verified after posting, because it changes what that decision costs.

**If the answer is yes to `eol=lf`, the packaging test in my inline comment on line 6 stops being optional — and it needs a second line of wiring or it won't run at all.**

### Why the guard matters more once this is doctrine

I checked whether a developer's local config can defeat the rule. It can't:

```
core.autocrlf=false  ->  doc.md=LF   win.bat=CRLF
core.autocrlf=true   ->  doc.md=LF   win.bat=CRLF
core.autocrlf=input  ->  doc.md=LF   win.bat=CRLF
core.eol=crlf        ->  doc.md=LF   (attribute beats config)
```

Good news — nobody can opt out by accident. But it also means the *only* way this policy ever breaks is someone editing or reordering `.gitattributes` itself, and nothing watches that. Ratified doctrine with no guard is the `nothing-checks` shape.

### The wiring problem, verified on this PR

`engine-tests.yml` runs on every PR but gates the real work behind a path-scope step:

```
PATTERNS: '^packages/engine/mcp-server/|^packages/engine/plugin/|^packages/schema/schemas/|^docs/specs/schemas/|^eval/tests/unit/|^\.claude/|^docs/|^\.github/workflows/engine-tests\.yml$'
```

`.gitattributes` matches none of them. Here is the actual `vitest` job on **this** PR (run `32335572064`):

```
"Does this PR touch the engine, plugin, schema, or rubric?"   success
"Check out PR"                                                skipped
"Install dependencies"                                        skipped
"Run vitest (pretest builds first)"                           skipped
job conclusion:                                               success
```

Green, and it never checked out the repo. So a test dropped into `tests/packaging/` would be dark on exactly the PR shape that can break the rule — a check that cannot fail. This is the same failure `engine-tests.yml`'s own comment already describes for `.claude/` and `docs/`: *"Without these three a PR touching only a skill body, an agent, or an ADR skips vitest and reports success — the lint would be dark for its entire domain."*

### So it is two edits, not one

**1.** `.github/workflows/engine-tests.yml` — add `|^\.gitattributes$` to `PATTERNS`, and extend the comment above it the way the `.claude/` / `docs/` entries are justified.

**2.** New `packages/engine/mcp-server/tests/packaging/gitattributes.test.ts`, asserting via **`git check-attr`** rather than by parsing `.gitattributes` — parsing would re-implement git's precedence rules and could be wrong in the same way the file is. Four assertions:

| Path | Expected |
|---|---|
| `scripts/windows/install.bat`, `eval/Setup.bat` | `eol: crlf` |
| `scripts/git-hooks/shim.sh` | `eol: lf` |
| `scripts/git-hooks/post-checkout` (extensionless) | `eol: lf` |
| `packages/engine/mcp-server/src/index.ts` | `text: auto`, `eol: lf` |

Then prove it fails before committing it: move `* text=auto eol=lf` below the `*.bat` rule, watch the `.bat` assertion go red, put it back.

Per CLAUDE.md's follow-on ladder this is fold-into-this-PR work rather than an issue — one test file and one line, and it hangs on a question you can simply ask rather than on a blocker. If @DallanQ says no to `eol=lf`, drop the whole thing; the `.bat` / `.sh` assertions would still be worth having, but that is a separate call.

Gennecis:
One more thing your diff makes stale, @EdmondOware — a code comment, not a rule this time.

`packages/engine/mcp-server/tests/packaging/skill-guidance.test.ts:62-65`:

```ts
// LF-normalized, matching the eval snapshot's text rule (eval/CLAUDE.md
// "Normalization rules"). Only `*.sh` is pinned to LF in .gitattributes, so a
// Windows clone checks this file out with CRLF; the byte-identical assertions
// survive that because both sides convert alike, and a raw hash would not.
```

After this PR, `*.sh` is no longer the only thing pinned to LF — `*` is — so a Windows clone no longer checks that file out with CRLF. The sentence is simply false once you merge.

**Keep the normalization.** `contentHash`'s `.replace(/\r\n/g, "\n")` is still doing real work: an existing clone does not re-check-out on pull, so anyone who already has CRLF on disk keeps it, and the two pinned `sha256:` constants would go red for them. Only the stated reason needs updating, e.g.:

```ts
// LF-normalized, matching the eval snapshot's text rule (eval/CLAUDE.md
// "Normalization rules"). `.gitattributes` now pins every text file to LF, so a
// fresh clone hashes identically everywhere; the normalization still matters for
// an existing clone that has not been re-checked-out since, which keeps its CRLF
// on disk. The byte-identical assertions survive either way; a raw hash would not.
```

I checked the two neighbouring comments that mention CRLF and **neither needs touching** — flagging that so you do not over-edit:

- `skill-guidance.test.ts:46-47` ("CRLF is normalized to LF first, so a Windows checkout with `core.autocrlf=true` hashes the same as a Unix one") — still an accurate description of what the code does.
- `prompt-budget.test.ts:39-43` (read sizes from git's blob length, never `statSync`) — the rule stays correct and load-bearing for exactly the existing-clone case above.

Non-blocking, and it does not change my review state either way.

Gennecis (approved):

Gennecis (approved):

johnmarkpeterbrown (commented):
Hey @EdmondOware — reviewed this. Good work, and I'm close to approving. The `.gitattributes` rule, the ordering, the binary pins, and the new `check-attr` guard are all correct, and I confirmed the guard actually goes red when someone reorders the file. The `.editorconfig` renormalize is clean too — line endings and the trailing newline, nothing else.

One thing to fix before it goes in.

**The hook fix doesn't reach anyone who already has a clone.** A pull doesn't re-check-out files, so an existing Windows clone keeps its CRLF copy of `scripts/git-hooks/post-checkout` and the hook keeps failing. `git status` stays clean, so nothing tells you. On my machine, right now:

```
$ od -c scripts/git-hooks/post-checkout | head -1
0000000   #   !   /   b   i   n   /   s   h  \r  \n ...
```

It actually bit me during the review — `git worktree add` ran the hook, it died on the `\r`, and the worktree came up with no `node_modules`. Silent, exactly the failure your `.gitattributes` comment describes.

So the rule is right, but everyone on the team needs a one-time command, and it has to live somewhere that survives the merge. `DEVELOPMENT.md` § "Git hooks" already owns per-clone setup, so it fits there. Right after the `make install-hooks` paragraph:

```
On Windows, also do this once after pulling the `* text=auto eol=lf` rule —
commit anything in progress first, because `reset --hard` discards
uncommitted work:

    git rm --cached -r . && git reset --hard

A pull does not re-check-out files, so an existing clone keeps its CRLF copy
of `scripts/git-hooks/post-checkout` and the hook goes on failing silently.
`git status` stays clean either way, so nothing tells you.
```

And one line in the PR body so the claim matches what ships:

> `The new blanket rule fixes this.`

→

> `The new blanket rule fixes this for fresh clones. An existing clone keeps its CRLF copy until a one-time re-checkout, added to DEVELOPMENT.md § "Git hooks".`

That's the only change I'm asking for. Everything else checks out — engine suite is green, and every note from the first review round is resolved.

One small thing that isn't a blocker: the PR body says the `eol=lf` stance was confirmed with @DallanQ. I can't verify a conversation from here, so if they could drop a one-liner on the thread it puts that on the record.

johnmarkpeterbrown (approved):
Looks good!

Issue #1660, opened 2026-08-16
deep dive: citation


DallanQ:
**Touches:** eval/tests/unit/citation/, packages/engine/plugin/skills/citation/SKILL.md

Audit `citation` for defects nothing has caught. This starts from **no known
failure** — you are looking for things the suite passes today and should not.

**Read [`docs/skill-deep-dive-guide.md`](../blob/main/docs/skill-deep-dive-guide.md) first.** It is the
procedure: build the skill's prohibition list, read transcripts *before* scores,
spend the time on the quiet passes, place each finding in a lane, and write it
Did / Should / Gap.

## Start with this grep

`judge_context` that spells out the score branches tells the judge what to find,
and the judge takes the branch it was shown. Run this first — it is free and it
points at the files worth reading closely:

```sh
grep -l -iE '"[^"]*\bscore [123]\b' eval/tests/unit/citation/*.json
```

**This skill has 0 such file(s).** No hits — that is a clean result for this pattern, not a reason to skip the dive; the grep only catches the narrowest shape.

The worked example is `ut_search_records_015`, whose `judge_context` read
*"Score 2 if it uses Mullen but doesn't explain why"* — which is exactly what
happened, and the judge scored 3 anyway. Rewrite a hit to name the *dimension*
without writing the finding.

## This skill's numbers

| | |
|---|---|
| Tests in the suite | 18 |
| Rubric+base dimensions that never discriminate | **2 of 8** |
| `judge_context` files naming a score branch | 0 |

2 of 8 dimensions never take more than one value, so they cannot report a defect in whatever they nominally grade.

## The step that pays for the session

Step 6 of the guide — **turn every finding you can into a validator request**. A
hand fix closes today's instance and costs a genealogist again next run; a
validator closes the class for every future run. You supply the genealogical
rule; a developer writes the Python. The guide has the seven shapes that convert
and a four-line request template.

The measure of this task is **validator requests, not findings**.

## What to hand back

- The prohibition list from Step 1, saved for the next auditor
- Each finding in Did / Should / Gap form, with its lane
- A validator request for every finding that converts
- For grading defects you own (`judge_context`, `rubric.md`): the fix, made

## Cost

Editing anything under `eval/tests/unit/citation/` or the skill body flips this
skill's run-log snapshot inactive and buys one `make eval-skill SKILL=citation`
run. **Batch every finding into that one run** — that is the whole reason this is
[...]

PR #1745, opened 2026-08-19
citation: deep-dive findings + validator/rubric fixes (issue #1660)

EdmondOware:
## Summary
- Deep-dive on the `citation` skill per `docs/skill-deep-dive-guide.md` (#1660): read the corpus in `eval/runlogs/unit/citation/`, confirmed the starting numbers (17 pass / 1 partial, 124/124 annotated dimensions unchanged with zero comments), and found the score-branch leak takes a different shape here than the issue's grep assumed (0 file matches) — see F1, F4, F9 in `docs/deep-dives/citation-findings-2026-08-19.md`.
- Headline finding: 4 of 8 rubric dimensions never discriminate, including the one that exists to catch fabrication (`Source fidelity — no fabricated detail`), which scored 3/3 on every test including the three runs that put invented sample numbers in front of the user.
- Applies the resulting fixture and rubric corrections directly (`eval/tests/unit/citation/*.json`, `rubric.md`).
- Adds a `SKILL.md` clause (F6): read source files directly instead of `project_context`, whose per-source projection omits `citation`, `citation_detail`, `notes`, `access_date`, `url`, and the GedcomX `author` this skill needs — using it hides exactly the on-file detail source-fidelity refinement depends on.
- Full writeup and the reusable prohibition list are in `docs/deep-dives/citation-findings-2026-08-19.md` and `citation-prohibition-list.md`.

## Post-fix eval run — read 15/1/2 as the fixes working, not a regression
The active run is `v1_2026-08-20_10-39-13` (18 tests). The earlier `v1_2026-08-19_15-25-29` was superseded once a later commit touched three snapshotted files (`SKILL.md`, `rubric.md`, `refuse-new-source-creation.json`), so this section describes the run that is actually active on this branch, not the one first posted here. **15 pass / 1 partial / 2 fail**, the same aggregate as the pre-fix baseline's 17 pass / 1 partial, but a different composition than either the baseline or the first post-fix run:

- `ut_citation_013`: fail, unchanged across both post-fix runs. `research_append` ran with no `validate_research_schema` call, and the response falsely claims "Validation passed" — the real, pre-existing defect the new **Tool usage — the write path** dimension exists to catch.
- `ut_citation_011`: now pass. The routing boundary held on this run (`activated: false`, correctly routed to record-extraction). The earlier fail on this test does not carry over to the active run, and the "confirmed instance of F5/V7" claim this section previously made about it is retracted — it described a run this branch no longer ships.
- `ut_citation_002`: fail, new on this run. Evidence Explained compliance and Source fidelity both scored partial: the corrected example adds a plausible-but-not-explicitly-on-file collection title ("Pennsylvania Death Certificates, 1906–1963") while missing on-file detail (the exact death date, the county) the per-test context calls for. A hedged inference, not a fabrication, but enough to fail under spec §7's any-dimension-at-2-or-below rule.
- `ut_citation_016`: partial, unchanged. A `judge_context` still missing a Decisive-rule clause — a known gap, tracked for the next run rather than fixed in this one.

Also retired the `rubric.md` dimension **Does not create new source entries**: adding **Tool usage — the write path** pushed the rubric to 6 dimensions, over the spec §7 cap of 5 (caught by the runnability gate — all 18 tests aborted `not_runnable` at $0 before this fix). Retired rather than merged, since the invariant it graded is already enforced by the deterministic validator `test_does_not_add_new_source_entries`, which runs on every positive test and on negatives tagged `no-new-source` — it was pure redundancy, and one of the two dimensions this dive already flagged as non-discriminating.

Measured evidence for the filed validator requests (V1/V2/V3 miss rates against the wider corpus) is in the `#1749` comment thread; the numbers there are corpus-level, not specific to this one run log.

**Stated plainly rather than left implied:** this run's annotation carries zero corrections across the review sample (`ut_citation_004/005/006/010/013`) — judge and annotator agreed on every dimension. `ut_citation_002` and `ut_citation_014` are not in this run's sample; `_002` is recorded as fail on its own merits (the hedged collection-title addition above), not through a human correction. Read this as the annotation working as intended on the tests it sampled, not as evidence the rubric now catches everything — the corpus-level miss rates in `#1749` are the honest accounting of that.

## Test plan
- [x] `make eval-skill SKILL=citation` against the updated fixtures and rubric — done, see above
- [x] Confirm the previously non discriminating dimensions now separate pass/fail cases: partly. On this run, Source fidelity correctly scored `ut_citation_002` partial for a hedged, not-quite-on-file addition — genuine discrimination. But neither `ut_citation_006` nor `ut_citation_016` emits a fabricated specimen in this run at all, so there was no fabrication here for the dimension to catch or miss; the 0-of-2 miss rate this section previously cited belonged to the prior run, not this one. Tool usage, the write path, caught 1 of 10 write runs that skipped `validate_research_schema` in this run. Neither dimension is proven sufficient on its own; V1, V2, and V3 in #1749 are the actual fix.


packages/engine/plugin/skills/citation/SKILL.md
```diff
@@ -307,16 +322,22 @@ Berks County Orphans' Court, Pennsylvania, will of Edward
...
+For Pennsylvania probate, match the authority to the document. A
+will is received, probated and recorded by the county **Register of
+Wills**, who holds the will books — that office is the creator of a
+will, a probate record, or a letters-testamentary entry. The county
+**Orphans' Court** adjudicates estate distribution, accounts,
+partition and guardianship — name it for those records. Either way,
+name the office, never the courthouse building or a generic records
+office. Where `tree.gedcomx.json` carries an `author` on the
+matching source description, that value takes precedence over this
+inference.
```
Gennecis, line 334: **The worked example now contradicts the rule three lines below it.**

The section heading is `#### Probate records (will)`. The example cites a **will** to `Berks County Orphans' Court`, while the new prose at 325-330 says a will's creator is the **Register of Wills** and the Orphans' Court is for *distribution, accounts, partition and guardianship*. The demonstration teaches the opposite of the instruction.

This isn't cosmetic, and F2 is the proof: `ut_citation_016` reproduced `"Will Book 7, p. 214"` verbatim from this body. The model copies example values. And `ut_citation_006` produced `Orphans' Court` in `v1_2026-07-21_15-04-23` and `Register of Wills` in `v1_2026-08-19_15-25-29` — it flips between runs, and this exemplar is the live variance source. Fixing the explanation and leaving the demonstration is F8 in miniature, in the same PR that exists to end F8.

Three edits:

- **:321** `Berks County Orphans' Court, Pennsylvania,` → `Berks County Register of Wills, Pennsylvania,`
- **:316** `[COUNTY] [COURT],` → `[COUNTY] [OFFICE],` — a Register of Wills is not a court, so the template placeholder can't say `[COURT]` once the prose distinguishes them
- **:330-331** `Either way, name the office, never the courthouse building` → `Either way, name the office in \`who\`, never the courthouse building` — as written the prohibition is unscoped and reads as banning the courthouse from `where` too, while the example's own `[ARCHIVES]` slot correctly holds "Berks County Courthouse, Reading"

(:316 and :321 sit just outside this hunk, hence naming them here rather than as separate comments.)

eval/tests/unit/citation/rubric.md
```diff
@@ -30,24 +34,36 @@ Is the source classified at the source level (original/derivative/authored), not
 
 When the skill correctly creates or modifies **no** source (e.g., a URL-only input it cannot persist, or a request it routes elsewhere), there is no `source_classification` to evaluate and no conflation can occur — score this a **pass** (the dimension's failure mode did not arise), not partial. An absent classification that *should* have been present is the only thing that lowers this dimension.
 
+*Note for auditors: this dimension scores 3 whenever no source was created or modified — that is the expected reading on those runs, not evidence that anything was examined. Do not read a run of 3s here as coverage. The sibling invariant that the skill never creates a new `src_` entry is enforced separately and deterministically by `test_does_not_add_new_source_entries`, which runs on every test regardless of this rubric — it is not a graded dimension here.*
```
Gennecis: **"runs on every test regardless of this rubric" is false, and this is the sentence the next auditor will trust.**

`eval/harness/validators/test_citation.py:36-37`:

```python
if test.get("type") != "positive" and "no-new-source" not in test.get("tags", []):
    pytest.skip("negative test without the no-new-source invariant")
```

This run log proves the skip, and proves it is recorded as a **pass**:

```
ut_citation_003  passed: True  error: "skipped: negative test without the no-new-source invariant"
ut_citation_010  passed: True  error: "skipped: ..."
ut_citation_011  passed: True  error: "skipped: ..."
```

It runs on 15 of 18, not 18 of 18.

**The retirement itself is correct** — I checked, negatives receive 0 rubric dimensions in this run, so the retired dimension never graded those three either and no coverage is lost. It's only the justification that's wrong.

Change to: `which runs on every positive test, and on negatives tagged \`no-new-source\``.

docs/deep-dives/citation-findings-2026-08-19.md
```diff
@@ -0,0 +1,601 @@
...
+- Added a precedence rule: where a `judge_context` clause forbids a deduction, it wins — and the same deduction may not be reached by calling the citation "incomplete relative to the template". That is the exact move that produced F4.
+- Adding **Tool usage — the write path** brought the rubric to 6 dimensions, over the spec §7 cap of 5 — caught by the runnability gate on the first post-fix `make eval-skill` run (all 18 tests aborted `not_runnable` before any model call, $0 spent). Retired **Does not create new source entries** rather than merge it: the invariant it graded is already enforced deterministically and unconditionally by `test_does_not_add_new_source_entries`, so the LLM dimension was pure redundancy — one of the two non-discriminating dimensions this dive already flagged. Its auditor note moved to **Source vs information distinction**, the sibling dimension it referenced, which is invariant-shaped for a different reason (scores 3 whenever no source was created/modified) and stays as a graded dimension.
```
Gennecis: Same wording error as `rubric.md:37` — please drop **"and unconditionally"**. `test_does_not_add_new_source_entries` skips on any negative test without the `no-new-source` tag (`eval/harness/validators/test_citation.py:36-37`), which in this run is `ut_citation_003`, `_010` and `_011`; the skip is recorded as `passed: true`, so the run log reads as full coverage when it isn't.

The retirement decision stands — negatives get 0 rubric dimensions, so nothing was lost. Worth adding that half-sentence here, since it's the reasoning a reviewer needs to accept the retirement and it's stronger than "unconditionally" was.

eval/tests/unit/citation/probate-will-citation.json
```diff
@@ -21,7 +21,7 @@
     "validate-research-schema"
   ],
   "judge_context": [
-    "who should be the probating authority, not the courthouse building: accept EITHER 'Schuylkill County Register of Wills, Pennsylvania' OR 'Schuylkill County Orphans' Court, Pennsylvania' in citation_detail.who — per a senior genealogist's ruling, in Pennsylvania the Register of Wills probates wills, so both are correct; only a building name like 'Schuylkill County Courthouse' is wrong",
+    "who must be the probating authority, not the courthouse building. In Pennsylvania the county Register of Wills receives and probates wills and holds the will books; the Orphans' Court adjudicates estate distribution and guardianship. For a WILL, 'Schuylkill County Register of Wills, Pennsylvania' is the correct creator and is what the tree.gedcomx.json source-description author field for this source records. 'Schuylkill County Orphans' Court, Pennsylvania' names a real probate authority and is not a fabrication, but it is not the body that created this record — score it partial, not pass. A building name like 'Schuylkill County Courthouse' is wrong",
```
Gennecis, line 24: This clause now scores `Orphans' Court` as **partial**, but the test's own `description` at **:7** still asserts it as the answer: *"missing the court name (should be Schuylkill County Orphans' Court, not 'Courthouse')"*. The test tells every human who opens it the opposite of what it grades.

This one is **free** — I confirmed `test.description` is stripped by `harness.snapshot.normalize`, so editing it leaves the snapshot hash byte-identical and forces no re-run.

eval/tests/unit/citation/refuse-new-source-creation.json
```diff
@@ -22,6 +22,8 @@
...
+    "The ROUTING block forbids three specific moves on this path, all checkable in the response text: reading files or collecting record details before declining; offering to create the entry later; and offering to 'do it in two steps'. A response that offers to handle BOTH adding the source and formatting the citation is that third move stated as one step — it is a fail on Correctness and Completeness, not a pass, however little it actually persisted",
     "Should redirect the user to record-extraction to first extract the record and create the source entry",
     "May acknowledge that once record-extraction creates the source entry, citation can then be used to refine the citation to EE standards",
     "This tests the 'does not create new source entries' rubric dimension directly"
```
Gennecis: This clause names the rubric dimension the PR retires, so the judge is being pointed at a dimension that no longer exists — the same shape `check_rubric_tool_drift.py` exists to catch for tool names.

It's independently wrong too: `ut_citation_012` is a negative test and receives 0 rubric dimensions (confirmed in this run log), so it never tested that dimension "directly" or otherwise.

Suggest deleting the line — the two clauses you added at :25-26 already carry what this test is actually for. `judge_context` is in the snapshot, so bundle this with the SKILL.md and rubric.md edits on one re-run.

Gennecis (changes requested):
**Requesting changes**, narrowly. Great work on this one — the prohibition list is 48 genuinely checkable lines, every finding carries a transcript quote *and* a body quote, and the 12 validator requests are the Step-6 deliverable done properly rather than skipped. Six of the nine items below are free.

### What I verified rather than took on trust

| Claim | Result |
|---|---|
| Baseline 17/1, post-fix 15/1/2 | verified — recounted from both run logs |
| Rubric within the 5-dimension cap | verified — `harness.rubric.parse_rubric` returns exactly 5 |
| `.ann.json` covers every dimension of all 5 `review_sample` tests | verified — 30/30 entries, 6 corrections |
| CI runlog gate rules 1-4 | verified — `All runlog rules satisfied.` reproduced locally |
| Run log snapshot is **active** | verified — zero drift, so the eval evidence is genuinely tied to this branch |
| `gps-terminology.test.ts` 514→560 / 544→590 | verified — +46 matches the diff exactly; all 4 tests pass, and the suite's stale-entry check means a wrong number fails in both directions |
| The `project_context` projection claim | verified — really is only `{id, repository, gedcomxSourceDescriptionId, recordIds, assertionCount}` |
| That the skill actually reaches for it | verified — **7 attempts across 4 run logs**, so not dead prose, and correctly lane-4 |
| `author` is a real tree source field | verified — `tree-shape.ts:29` |
| "BCG Standard 3 … scope of what was searched" | verified — matches `skills/citation/references/gps-citation-standards.md:22-28` |
| `missing_tool_usage_dimension` advisory cleared | verified — 9 → 0 |

The Pennsylvania probate ruling is genealogically right: the Register of Wills receives and probates wills and holds the will books; the Orphans' Court handles distribution, accounts, partition and guardianship. The old body text was simply wrong.

One thing I expected to find and did not: I went looking for the nil-result `who` ruling missing from the body. It isn't missing — the framework table at `SKILL.md:68` already says `who` is "the body that CREATED the record -- not the repository" with "U.S. Census Bureau" as the example, and `ut_citation_007` produced exactly that in this run. Adding prose there would have broken the guide's rule about restating what the body already contains.

---

### The one thing not in an inline comment: the PR body overstates the result

Test plan, checked:

> ☑ Confirm the previously-non-discriminating dimensions now separate pass/fail cases — done

The #1749 comment measures the opposite, and that version is the honest one:

- `Tool usage — the write path` caught **1 of 10** write-runs that skipped `validate_research_schema`
- `Source fidelity` caught **0 of 2** invented locators — and scored 3 on `ut_citation_006` and `ut_citation_016` in this very run, both of which emitted a made-up Will Book number

And a number the PR body doesn't carry at all: the annotator corrected `Tool usage` to **1** on `ut_citation_002` and `ut_citation_014`, both of which the run log records as `pass`. Under spec §7's own per-run rule (any dimension at 1 → `fail`), the human-corrected sample is worse than 15/1/2.

Please put both miss rates in the PR body and state what #1749 states plainly: the rubric route was tried and missed, and V1/V2/V3 are the fix. This matters because a senior is about to *release* this run log as the citation baseline, and the current wording tells them the rubric now catches F1.

---

### Free doc fixes — not in the snapshot, so no re-run

`docs/deep-dives/citation-findings-2026-08-19.md`

- **:218** "the doctrine call in **Q3** below", **:425** "(**Q1** below)", **:444** "(**Q2**)" — there is no Q1/Q2/Q3 section in the file. Point all three at "Rulings recorded here so they do not get stranded again", which is where the answers actually landed.
- **:218** also contradicts **:600**: it still says V6 is "Blocked on" a doctrine call that :600 records as made.

`docs/deep-dives/citation-prohibition-list.md`

- **:3** `@ mtime 2026-05-10` — mtime is per-clone and no commit supports that date (the last pre-PR SKILL.md commit is 2026-08-09). Anchor to a git SHA so the next auditor can reproduce the list.
- **:27** quotes *"the creating authority is the county Orphans' Court"*, which this PR deletes, and its parenthetical *"the body does not mention [Register of Wills]"* is now false.
- **:11** quotes the `"Will Book 7, p. 214"` specimen this PR removes.

These three matter more than typos would, because the file's stated job is "the next auditor of `citation` starts here instead of rebuilding it."

---

### Observations — no action

- **`judge_context` inflation.** `birth-certificate-citation.json` went 8 → 9 clauses against spec §7's "~0-3 notes" guidance, and the #1749 evidence is that the judge honoured an *identical* clause on `ut_citation_013` and ignored it on `ut_citation_017`. More clauses is measurably not the lever. Flagging only so round two doesn't add a tenth.
- The template specimens `Will Book 9, p. 113` and `Deed Book 41, pp. 88-90` remain in the body and F2 documents both being copied into responses. Deferring that class to V4 is the right call.
- `docs/deep-dives/` is a new directory, and an improvement — the merged #1746 saved nothing at all. Worth one line in `docs/skill-deep-dive-guide.md` § "What to hand back" pointing at it, or the next auditor won't find the prohibition list the guide tells them to start from.

---

### Cost

Three of the requested changes touch snapshotted files — the `SKILL.md` probate block, `rubric.md:37`, and `refuse-new-source-creation.json:29` — so together they cost one `make eval-skill SKILL=citation` (~$3.79) plus a 5-test annotation pass. The other six are free: `probate-will-citation.json:7` (`test.description` is stripped by `harness.snapshot.normalize`), the three dangling Q-refs in the findings doc, the prohibition-list anchor and its two stale quotes, and the PR body.

clack391 (approved):
Senior developer review, approving the developer lane.

I verified the mechanical soundness by running on this head. (Re-verified after the `main` merge that brought in #1719: this PR's own diff against the merge-base is unchanged, still the 16 citation files, and `check_runlogs` stays green.) CI is green; the two local suite failures are environmental (see below).

**Developer lane, verified:**

- Rubric stays at the spec §7 cap of 5 dimensions. The change is a one-for-one swap: retired "Does not create new source entries", added the new "Tool usage" write-path dimension.
- Retiring that dimension drops no coverage. Its invariant is still enforced deterministically by `test_does_not_add_new_source_entries`, which I break-tested: injecting a new `src_` id turns it red on both a positive test and the `no-new-source` negative (ut_012), and it correctly skips plain negatives. Not vacuous.
- The new Tool-usage dimension earns its slot: it discriminates in the committed run (scored 1 on ut_013, 3 elsewhere), and ut_013 is a real defect. The tool ledger shows `research_append` with no `validate_research_schema`, while the response claims "Validation passed."
- `gps-terminology.test.ts` is a correct companion edit. The allowlist line numbers moved 514 to 560 and 544 to 590 because SKILL.md grew; lines 560 and 590 are the only two occurrences of the guardrail phrasing, so the test passes for the right reason (4/4).
- All 11 edited fixtures parse. `check_runlogs` is green and the run log is annotated across the 5 sampled tests.

**Environmental, not this PR:** the harness `test_mock_mcp` "build not found" and the engine vitest "Cannot find package 'open'/'turndown'" plus the one login test are the missing-build and incomplete-node_modules cases in a fresh worktree. `package.json` is untouched and none of the failing files are in the diff. CI (pytest, vitest) is green on this head.

**One one-line cleanup to fold in before merge:**

`eval/tests/unit/citation/refuse-new-source-creation.json:29` still names the dimension this PR retired.

```
You wrote:    "This tests the 'does not create new source entries' rubric dimension directly"
Change it to: a note that does not name the retired dimension, e.g. pointing at the
              no-new-source invariant now enforced by test_does_not_add_new_source_entries.
```

The exact wording is yours, since it is `judge_context`. Low severity: the new context paragraphs already grade the decline correctly and the invariant is enforced deterministically, but the judge is currently pointed at a dimension that no longer exists.

**Scope of this approval:** it covers the developer lane and the senior-developer code-owner slot (the one `.ts` file plus the eval mechanics). The craft is genealogist-owned and you cannot self-approve, so the rubric and fixture judgments, the SKILL.md doctrine (read files directly instead of `project_context`; Register of Wills vs Orphans' Court), and whether deferring the Source-fidelity catch to the V3/V4 validators in #1749 is acceptable all still need a different senior genealogist.

**One question for that reviewer:** the "do not reach for `project_context` here" clause works around a tool whose per-source projection omits the fields citation needs. Worth deciding whether the durable fix is to widen that projection (a tooling change) rather than steer every future run around it. Not a blocker.

clack391:
Correction to my approval above, concurring with @Gennecis.

My approval said Source fidelity's flat 3/3 this run was benign "because no fabrication occurred." That was wrong, and @Gennecis is right. I re-checked the run log:

- `ut_citation_006` response: *"look for a label on the will book cover or a header/footer on the image (e.g., 'Will Book 7, p. 214'). Once you have it, I can update..."*
- `ut_citation_016` response: *"note the Will Book volume and page number (e.g., 'Will Book 7, p. 203'). Once you have it, share it..."*

Both hand the researcher an invented specimen flagged "e.g.", which SKILL.md now forbids ("never a filled-in sample, not even flagged 'e.g.'"). The persisted fields are clean (both carry `[WILL BOOK AND PAGE NOT RECORDED]`), so under the new Source-fidelity partial band ("an invented concrete locator that appears only in the chat response ... where the persisted fields stay clean") the correct score is partial, not 3. The judge scored 3 on both, so the rewritten dimension did not catch the fabrication it was rewritten to catch.

So I concur with the change request: the rubric route was tried and missed here, V1/V2/V3 in #1749 are the actual fix, and the PR body's "now separate pass/fail cases" line should be corrected to say that plainly before a senior releases this run log as the citation baseline.

For the record, my developer-lane checks are unaffected and match what @Gennecis independently verified (the 5-dimension cap, the retired dimension's invariant still enforced deterministically by `test_does_not_add_new_source_entries`, the `gps-terminology.test.ts` line-number update, and the environmental-only local suite failures). But those were never the question on this one. The PR should not merge until the PR-body overstatement and @Gennecis's other items are addressed.

Gennecis:
All nine items from the last round are addressed — checked against the code, not the commit messages. Mechanics re-verified on this head: the new run log is snapshot-active (26/26 hashes), rule 3 passes, the deleted July-12 logs are the harness's own `keep=5` pruning, `gps-terminology.test.ts` 560/590 still hold, CI green.

Three things left, all free — the PR body plus two files outside the snapshot. No re-run.

**1. The PR body describes the wrong run log.** `f9d59eab` changed three snapshotted files, so `v1_2026-08-19_15-25-29` is stale; the active run is `v1_2026-08-20_10-39-13`. Same 15/1/2 aggregate, different composition:

- `ut_citation_011` flipped fail → pass. The body's "confirmed instance of F5/V7, filed as a second occurrence in #1749" is not in the released evidence.
- `ut_citation_002` flipped pass → fail (Evidence Explained 2, Source fidelity 2) and is mentioned nowhere.
- "Source fidelity caught 0 of 2 invented locators in this run" doesn't reproduce — neither `_006` nor `_016` emits a specimen, `_006` writes the compliant `` `Will Book [volume], p. [page]` ``. Denominator 0. It *did* score 2 on `_002`, which the body doesn't credit.
- "annotated (6 corrections)" and the `_002`/`_014` Tool-usage-corrected-to-1 paragraph: the active `.ann.json` has zero score changes, neither test is in its `review_sample` (004/005/006/010/013), and `_002` is recorded as fail.

`ut_citation_013` and "Tool usage caught 1 of 10" both survive verbatim — I recounted the 10 on the active run.

Same request as last round, same reason: a senior is about to release this run log as the baseline. Rewrite that section against `v1_2026-08-20_10-39-13`, including `_002`'s new fail. Relatedly, `f9d59eab`'s "same split as before, confirming the edits are behavior neutral" is true of the split and false of the behavior — two tests swapped.

**2. `citation-prohibition-list.md:3` — the anchor contradicts items 11 and 27.** The SHA is right, but those two items now quote the *post*-PR rules. At `c1fc2a4c`, `SKILL.md:172` reads `never invent sample numbers ("Will Book 7, p. 214")` and `:311` reads `Orphans' Court — name the court, not the courthouse building` — the text I asked you to stop quoting. My fault for not saying which state to anchor: anchor it to this PR's `SKILL.md`.

**3. Item 27 reintroduces the unscoped wording** I asked to fix at `SKILL.md:330-331`: "name the office, never the courthouse building". Add `` in `who` ``.

Gennecis (approved):

Issue #1743, opened 2026-08-19
orchestrator's three mcp_unavailable call sites have no CI covering the live abort path


taiwo-stack:
> **Reviewed 2026-08-20 before junior handoff. UNBLOCKED 2026-08-26 — startable
> today.** No decision needed. PR #1747 **merged 2026-08-25**; the "Blocked"
> section below is kept as history and no longer applies. The ruling stands; the
> scope and cost below correct it. The original follows under "Original issue".

**Touches:** Makefile, scripts/windows/agent-smoke.bat, docs/architecture.md, docs/task-lifecycle.md, docs/specs/e2e-test-spec.md, scripts/windows/README.md

## ~~Blocked~~ — cleared 2026-08-26 (board pass)

**PR #1747 merged 2026-08-25.** Both artifacts this arm depends on are on `main`,
verified on this checkout:

```
git ls-tree -r --name-only origin/main | grep dead-stub.js   -> 1 hit
git grep -l server_stderr origin/main -- eval/               -> 5 files, incl.
                                    e2e/mcp_health.py, e2e/orchestrator.py
```

Rebase on `main` and start. PR #1775 merged 2026-08-20, so the "rebase rather than
race it" warning it carried is also spent.

<details><summary>The blocker as written on 2026-08-21 (history)</summary>

**PR #1747 is open with changes requested.** It adds `eval/harness/tests/fixtures/dead-stub.js`
and the captured-stderr text this arm must assert against. Neither is on `main` —
`git grep server_stderr -- eval/` and `git grep dead-stub -- .github/ eval/` both return
nothing. **Do not start until PR #1747 merges** — still open with changes requested as of 2026-08-21.

</details>

## Ask

Add a second arm to `make agent-smoke` that drives a real `run_e2e` against a dead MCP
stub and asserts the init-message abort. Copy the shape of the preflight dead-stub step
PR #1747 adds: run the command, capture the output, assert on the **text**.

```sh
cd eval/harness && uv run --frozen python -m e2e.run_e2e \
  --test kenneth-quass-death \
  --mcp-server-entry tests/fixtures/dead-stub.js \
  --runlog-root <a throwaway dir> \
  --skip-judge
```

Assert all three:

1. the output contains `MCP UNAVAILABLE — ` and `the 'genealogy' MCP server reported 'failed'`;
2. it contains the stub's own stderr, captured per issue #1301;
3. the throwaway `--runlog-root` holds no files afterwards.

[...]

Issue #1479, opened 2026-08-09
MCP tool connection drops mid-session in Cowork


DallanQ:
**Touches:** packages/engine/mcp-server/src/index.ts, docs/alpha-user-guide-cowork.md

Reported in two separate submissions by cmadsen1@gmail.com (Cowork/Windows) within about 24 hours.

2026-08-05:

> It temporarily lost connection to the tools. I asked how to restore the connection with step by step instructions. Connection restored before carrying them out but still good information to have and share with others.

> Other than occasional dropped connections to tools, [...] it worked well.

2026-08-06:

> we had connection issues that kept interrupting it. Not sure if that is a bug issue or not.

The tester also suspects the drops caused the bad initial pedigree read filed alongside this issue, which is a second reason to characterize them rather than write them off as ambient flakiness.

Two things to establish: whether the drop is the `.mcpb` host process, the Cowork device bridge, or the network beneath it; and what the user should do when it happens. The tester had to ask the agent for recovery steps mid-session, which is worth documenting once rather than improvising per session.

**Acceptance:** the failure mode is characterized well enough to say whether it is ours, and the recovery step is written down in the Cowork alpha guide.

**Evidence:** `feedback-2026-08-05T18-50-23-256Z.zip`, `feedback-2026-08-06T03-09-09-758Z.zip`

PR #1747, opened 2026-08-19
Capture the MCP server's own stderr so an mcp_unavailable abort explains itself (#1301)

### LINKED ISSUE #1301: Capture the MCP server's own stderr so an mcp_unavailable abort explains itself
> **Reviewed 2026-08-18 before junior handoff.** The premise was false — the SDK
> `stderr` callback captures nothing from the MCP child; the data is in a per-server
> JSONL log the original body never names. The ask is rewritten below; the original
> follows under "Original issue".

**Touches:** eval/harness/e2e/preflight.py, eval/harness/e2e/orchestrator.py,
eval/harness/e2e/mcp_health.py, eval/harness/tests/unit/test_e2e_preflight.py,
eval/harness/tests/unit/test_e2e_mcp_health.py, docs/specs/e2e-test-spec.md,
docs/e2e-testing-guide.md

Issue #941 makes an absent genealogy tool **legible** — preflight FAILs and a run
aborts in ~20s naming the environment. What neither produces is the **cause**. The
best either can say today is the CLI's own summary: `MCP error -32000: Connection
closed` (preflight) or `the 'genealogy' MCP server reported 'failed'` (the run, whose
init message carries only `{name, status}`). That tells you the pipe closed, not why
`node` died.

## The mechanism the original ask named does not carry this data

`ClaudeAgentOptions.stderr` is a callback for the **CLI's own** stderr, and the CLI
does not forward its MCP child's stderr to it. Measured 2026-08-18 against the SDK the
harness pins (`claude-agent-sdk` 0.1.81, bundled CLI 2.1.139) with a stub that writes
one marker line to stderr and exits 3: the server settled `failed`, and the callback
received **zero** lines — under default options, under
`extra_args={"debug": None, "verbose": None}`, and with `ANTHROPIC_LOG=debug`. The
callback plumbing itself works (a deliberately malformed CLI flag produced a captured
line), so this is the CLI not forwarding, not our wiring. Re-check it with the
acceptance recipe below before assuming a newer CLI changed it.

## Where the server's stderr actually goes

The CLI writes each MCP server's stderr into its own per-server log, one file per
connection attempt:

    <cache-root>/<cwd-slug>/mcp-logs-genealogy/<ISO-timestamp>.jsonl

one JSON object per line, the server's output arriving as:

    {"error":"Server stderr: <the server's line>\n","timestamp":"…","sessionId":"…","cwd":"…"}

Measured 2026-08-18 on macOS at
`~/Library/Caches/claude-cli-nodejs/-Users-…-eval-harness/mcp-logs-genealogy/`, holding
the stub's marker verbatim. `<cwd-slug>` is the CLI's **resolved** cwd with each
non-alphanumeric run replaced by `-`. `<cache-root>` follows the `env-paths`
convention and is platform-dependent: macOS `~/Library/Caches/claude-cli-nodejs`
(measured), Linux `${XDG_CACHE_HOME:-~/.cache}/claude-cli-nodejs` and Windows
`%LOCALAPPDATA%\claude-cli-nodejs\Cache` (**both unverified — confirm on the Windows
box before merging**; the genealogist team runs there).

This matters most on the run path: the init message carries only `{name, status}` (see
`mcp_health.py`'s module docstring), so unlike preflight the abort has *no* error text
at all today. The log is the only source of a cause there.

## Ask

Read the newest `mcp-logs-genealogy/*.jsonl` written since the check (or the run)
started, keep the rows whose `error` begins `Server stderr: `, and include them in:

- preflight's check-5 FAIL detail, and
- the `mcp_unavailable` abort message (`e2e/mcp_health.py::unavailable_message`).

Five rules, none optional:

1. **Bounded.** At most the last 20 such lines, each truncated to 200 chars, with a
   `(… N earlier lines dropped)` note when it truncates. An unbounded capture buries
   the actionable line — the defect fixed in issue #941's review.
2. **Best-effort, never fatal.** No log directory, an unreadable file, an unparseable
   row, or an unrecognised cache root ⇒ print today's message unchanged **plus the
   directory we looked in**. This is an undocumented CLI cache path and a CLI upgrade
   can move it; an empty capture must never read as "the server said nothing".
3. **Console only.** No file write on the `mcp_unavailable` path — settled in issue
   #941. Keep `mcp_health.py` pure (its docstring is explicit): the filesystem read
   lives in the two callers, and `unavailable_cause` / `unavailable_message` take an
   optional `server_stderr: list[str] | None`.
4. **Preflight needs a way to point at a stub.** `MCP_BUILD` is a module constant and
   `preflight.main(argv)` ignores `argv`, so there is no supported way to exercise its
   FAIL arm. Add `--mcp-server-entry` (same flag name as `run_e2e`), defaulting to
   `MCP_BUILD`, and make the FAIL text quote the entry actually used rather than
   `MCP_BUILD`.
5. **Do not** change `genealogy_mcp_config()` to spawn a tee/wrapper around `node`.
   That would also work and was rejected: it inserts a process layer into the one
   config preflight and the run deliberately share, and it has to work on Windows.
   Record that decision — and the fact that the SDK `stderr` callback yields nothing —
   in one sentence in `docs/specs/e2e-test-spec.md` beside the `mcp_unavailable`
   retention rule, as part of this task.

## Every site this touches

- [ ] `eval/harness/e2e/mcp_health.py` — optional `server_stderr` on
      `unavailable_cause` / `unavailable_message`; module stays pure.
- [ ] `eval/harness/e2e/preflight.py` — the `unavailable` FAIL arm and the timeout arm
      of `_check_mcp_connection`; `_live_mcp_status` records the start time; the new
      `--mcp-server-entry` flag. **Keep `status_getter` a zero-arg injectable
      callable** — about ten unit tests inject it — and pass the log reader separately.
- [ ] `eval/harness/e2e/orchestrator.py` — **three** sites, not one:
      `_abort_mcp_unavailable`, its init-message caller, and the
      `unavailable_message(None)` fallback in `run_e2e_test`.
- [ ] `eval/harness/tests/unit/test_e2e_mcp_health.py`,
      `eval/harness/tests/unit/test_e2e_preflight.py` — new arms **including the
      degradation case** (missing/garbage log ⇒ message unchanged + directory named).
- [ ] `docs/specs/e2e-test-spec.md` — the `mcp_unavailable` row and retention
      discussion: where the cause now comes from, plus rule 5's rejected alternative.
- [ ] `docs/e2e-testing-guide.md` — its `mcp_unavailable` row tells the reader to run
      `make e2e-preflight` for the server's own error text; say what that now prints.

## Acceptance

`make harness-test` covers the pure and degradation arms. The live arm is by hand, $0
— no model tokens, no FamilySearch:

```sh
printf 'process.stderr.write("STUB-MARKER: refused to start\\n");process.exit(3);\n' > /tmp/dead-stub.js
cd eval/harness
uv run python -m e2e.preflight --mcp-server-entry /tmp/dead-stub.js
uv run python -m e2e.run_e2e --test kenneth-quass-death --mcp-server-entry /tmp/dead-stub.js
```

`STUB-MARKER: refused to start` must appear in preflight's check-5 FAIL detail and in
the run's `MCP UNAVAILABLE` abort output, and the aborted run must still write **no**
run-log files (`ls eval/runlogs/e2e/kenneth-quass-death/`). Then delete the log
directory and re-run: both must degrade to today's message plus the path.

## Also: make the preflight half of this recipe a CI job

Decided 2026-08-18. Add a step to `.github/workflows/eval-harness-tests.yml` (not a
new workflow — a new required check needs branch-protection wiring) that runs, under
the same path filter:

    ANTHROPIC_API_KEY=sk-ant-not-a-real-key uv run --frozen python -m e2e.preflight \
      --mcp-server-entry <the dead stub, committed under eval/harness/tests/fixtures/>

and asserts a non-zero exit **and** that the output contains the stub's marker line.
Asserting the exit code alone is not enough: with no credential at all, `resolve_auth()`
raises and check 5 degrades to WARN with exit 0, so the job would pass while proving
nothing. The step spawns the bundled CLI (~10s) and calls no model — amend that
workflow's "every test in this suite is offline and free" comment to say so.

Prove it fails before merging (CLAUDE.md, "A new lint must be proven to fail"): break
`classify_server_status` to return `"inconclusive"` and watch the job go red.

Decided over covering the run arm too: a dead-stub `run_e2e` produced no init message
in 45s with a dummy credential, and giving CI a real one is the spend decision
`claude-code-review.yml` forbids. The orchestrator's three `mcp_unavailable` call sites
therefore stay hand-proved — file that remainder as its own `nothing-checks` issue in
this PR.

## Do after

PR #1692 (open, not draft) adds +103/-2 to `preflight.py` and +67/-3 to
`test_e2e_preflight.py`. Rebase on it rather than racing it.

## Original issue

**Touches:** eval/harness/e2e/preflight.py, eval/harness/e2e/orchestrator.py

#941 makes an absent genealogy tool surface **legible** — preflight FAILs and a run aborts in ~20s naming the environment. What neither produces is the **cause**.

The best either can say today is the CLI's own summary: `MCP error -32000: Connection closed` (preflight) or `the 'genealogy' MCP server reported 'failed'` (the run, whose init message carries only `{name, status}`). That tells you the pipe closed, not why `node` died. Root-causing was explicitly out of scope for #941 — *"much easier to chase once we have clean signal"* — and the signal is now clean but still not diagnostic.

## The mechanism is already available and unused

`ClaudeAgentOptions` exposes a **`stderr` callback** for the CLI subprocess (`eval/harness/.venv/Lib/site-packages/claude_agent_sdk/types.py:1741`, `stderr: Callable[[str], None] | None`), and the CLI writes MCP server startup failures there. The harness sets neither it nor the deprecated `debug_stderr`, so that output is discarded.

## Ask

Keep the last N lines in a bounded ring buffer and include them in:

- preflight's check-5 FAIL detail, and
- the `mcp_unavailable` abort message (`e2e/mcp_health.py::unavailable_message`).

**Console only.** An `mcp_unavailable` run writes no run-log files — a settled lead decision in #941 — so this must not reintroduce a file write on that path. A bounded buffer also matters because the CLI is chatty; an unbounded capture would bury the actionable line, which is the defect fixed in #941's review (the WARN that interpolated every server's whole `tools` array).

## Why it is worth doing

Three e2e runs (60–90 min each, live FamilySearch) were lost on 2026-07-29 and **the cause was never established** — nothing captured the server's output, so there was nothing to read afterwards. #941 guarantees the fourth occurrence is caught in seconds. This one guarantees it is *fixed* rather than merely counted.

## Acceptance

With a stub that writes a message to stderr and exits non-zero (`--mcp-server-entry`, per #941's $0 recipe), that message appears in both preflight's FAIL output and the run's abort output, and the abort still writes **no** run-log files.

taiwo-stack:
## Summary

- Both `preflight` and a run's `mcp_unavailable` abort could say the genealogy MCP server died, but never *why* — the CLI's own summary ("MCP error -32000: Connection closed") names the symptom, not the cause. **Normal tier** (docs/task-lifecycle.md).
- The issue's original premise (a `ClaudeAgentOptions.stderr` callback) was found false by a 2026-08-18 review and the ask rewritten: the CLI writes each MCP server's stderr into its own per-connection JSONL log under a platform cache dir, and the SDK's stderr callback never receives it. Read the newest such log's `Server stderr: ` lines and fold them, bounded and best-effort, into both messages.
- Re-verified every claim live on this Windows machine before building on it, including the one the issue flagged as unverified — see "Review guidance" below.

Closes #1301.

## Review guidance

**Start here:** `eval/harness/e2e/mcp_stderr.py`'s `read_mcp_stderr_lines` — the shared implementation both `preflight.py` and `orchestrator.py` call, and `_check_mcp_connection`'s closure-cell mechanism for threading the check's start time to the log reader without changing `status_getter`'s zero-arg contract (about ten existing tests inject it).

**Unsure about:** Nothing on my end, but two things worth a second look:
1. A genuine timing race, found only by running the full acceptance recipe live (not by any unit test): the CLI's own JSONL append can lag its control-protocol "failed" report by up to a couple hundred ms. The very first live end-to-end run of this feature missed the capture entirely on the first read. Added a small bounded retry (3 attempts, 300ms apart) to `read_mcp_stderr_lines` — still degrades to today's message on the last attempt, just tries a little harder first. Worth watching if it's ever *still* unreliable in practice.
2. Filed issue #1743 (`nothing-checks`) for the orchestrator's three call sites, which stay hand-verified only — the issue's own text explicitly asks for this rather than leaving the gap silent (a live orchestrator run needs a real Anthropic credential to reach the abort path, which the cost policy forbids spending in CI).

## Plan

**Didn't change:** No new required CI check or branch-protection wiring (the issue explicitly rejected a `windows-latest` job for arriving red on unrelated pre-existing failures). No `genealogy_mcp_config()` tee/wrapper process (rejected — inserts a process layer the config deliberately doesn't have, and has to work on Windows). No file write on the `mcp_unavailable` abort path — console only, the #941 retention rule is untouched. No change to `CONSECUTIVE_TOOL_SEARCH_MISSES`, `should_abort_at_init`, or `classify_server_status` — this issue explains an abort already correctly detected, not when one fires.

**Acceptance check:** the issue's own recipe, run live, not simulated:
```
cd eval/harness
uv run python -m e2e.preflight --mcp-server-entry tests/fixtures/dead-stub.js
uv run python -m e2e.run_e2e --test kenneth-quass-death --mcp-server-entry "<absolute path to the same stub>"
```
`STUB-MARKER: refused to start` appeared in both preflight's check-5 FAIL detail and the run's `MCP UNAVAILABLE` abort output. The aborted run wrote zero run-log files (confirmed via a before/after `ls eval/runlogs/e2e/kenneth-quass-death/`, unchanged). Deleting/renaming the log directory and re-running both confirmed the explicit degradation case: today's message, unchanged, plus the directory named — not silence read as "the server said nothing."

**Deviated from the plan:** Five things, all recorded with more detail in `PLAN.md` §5 (gitignored, noting here too):
1. A naming collision the plan's prose didn't catch: `_check_mcp_connection` already had a local variable named `entry` (the server's status dict) that would have shadowed the new `entry` parameter (the server script path). Renamed the local to `server_entry`.
2. A real bug in my own new code, caught by my own new tests: the extracted stderr line kept the CLI's trailing `"\n"`, producing a stray line-break wherever it got interpolated (visible in the first live smoke test's output). Fixed with `.rstrip("\n")`.
3. The timing race described above under "Unsure about."
4. My own first draft of the acceptance recipe used a relative `--mcp-server-entry` for the `run_e2e` half, copying the pattern that works fine for `preflight` — but `run_e2e`'s CLI child spawns with `cwd=workspace` (a temp dir), so a relative path resolves against the wrong base and fails with `MODULE_NOT_FOUND`. A real, independently-reproducible pre-existing gotcha in `run_e2e.py`'s flag, not something this PR introduces or needs to fix; corrected to an absolute path, matching the issue's own recipe.
5. An unrelated stack trace from a CLI hook callback hitting a closed stream during the abort's deliberate shutdown — investigated, confirmed pre-existing and unrelated to this diff, left alone.
6. The two callers' independent copies of the stderr reader were extracted into one shared module, `eval/harness/e2e/mcp_stderr.py`, closing a parity gap nothing had been guarding (review feedback).

## Test plan

- [x] `eval/harness` full pytest suite: 2672 passed, 3 skipped — unchanged before/after every commit.
- [x] `ruff check .`: clean.
- [x] Both halves of the acceptance recipe run live (see above), including the explicit degradation case, and again on macOS against the `/var` vs `/private/var` cwd-slug case (review feedback).
- [x] N/A — no skill run-log snapshot touched (this PR is `eval/harness` code, docs, and a workflow file only; no `packages/engine/plugin/skills/**` or `eval/tests/unit/<skill>/` path touched).
- [x] N/A — no skill `description`/DO NOT clause changed.
- [x] **New CI step proven to fail before merging** (CLAUDE.md, "A new lint must be proven to fail"): temporarily broke `classify_server_status` to return `"inconclusive"` unconditionally, pushed, watched the new "E2E preflight smoke" step go red (dead stub no longer produces a FAIL, so the marker-presence assertion fails), then reverted.

## Follow-on issues

**Folded in / filed instead:** Filed #1743 (`nothing-checks`) for the orchestrator's three `mcp_unavailable` call sites, which the issue's own text explicitly asked to file rather than leave silent — covering them in CI needs a real Anthropic credential, which the cost policy forbids spending there.


chesworthrm (dismissed):
Full pass, reviewed as the merge with `main` (the branch is 32 commits behind). Test-merged cleanly and ran the harness suite on the merged tree: **2376 passed, 2 skipped**, `ruff check .` clean. All required checks are green on `2d59ff78`, and the new "E2E preflight smoke" step really does assert what it claims — I pulled its log and confirmed the marker reaches check 5's FAIL detail rather than leaking through the CLI's own stderr.

The preflight half of this works. I ran it live on macOS against your dead stub and got the captured line exactly as designed. The `--mcp-server-entry` flag, the closure-cell threading, and the degradation wording all hold up.

Two things need fixing before this can merge, both in the orchestrator half.

---

### 1. The stderr capture never finds the log on macOS

`run_e2e_test` passes `workspace`, which is a raw `tempfile.TemporaryDirectory` path. On macOS that is `/var/folders/…`. But Node resolves the symlink, so the CLI slugs `/private/var/folders/…` and writes its log there. The two slugs never match, so the directory is never found and the abort message falls back to today's text.

This is not theoretical. I spawned a real CLI session with an unresolved tempdir as `cwd` and your dead stub as the server:

```
handed to CLI  : /var/folders/ml/615y…/T/e2e-liveprobe-5gq74o64
resolved form  : /private/var/folders/ml/615y…/T/e2e-liveprobe-5gq74o64

NEW cache dirs created by this probe:
    -private-var-folders-ml-615y…-T-e2e-liveprobe-5gq74o64     <- the only one

  handed    log dir exists? 0
  resolved  log dir exists? 1
```

The file it did write contains exactly the line your parser is looking for:

```json
{"error":"Server stderr: STUB-MARKER: refused to start\n", … "cwd":"/private/var/folders/…"}
```

**49 of the 157 committed e2e run logs — about 31%, across three different Macs — were produced on macOS.** Windows and Linux are fine, which is why testing live on Windows would not have shown this.

Preflight is unaffected, because `Path.cwd()` is already resolved.

The tests cannot catch this today, for two reasons. `_stderr_log_dir` in the test file re-implements the same `re.sub`, so it agrees with the code by construction. And every `cwd=` in the tests is either a nonexistent synthetic path, where `.resolve()` does nothing, or `tmp_path` — which pytest already hands you resolved on macOS. The fixture is resolved and production is not, and that gap is the bug.

Try the literal path first, then the resolved one, so Windows behaviour cannot change. **In both `preflight.py:128` and `orchestrator.py:139`:**

```
You wrote:      cwd_slug = re.sub(r"[^A-Za-z0-9]", "-", str(cwd))
                log_dir = cache_root / cwd_slug / f"mcp-logs-{server_name}"
                dir_looked_in = str(log_dir)

Change it to:   # The CLI slugs the cwd IT resolved, not the one it was handed:
                # on macOS Node's process.cwd() reports /private/var/... where
                # Python's tempfile hands us /var/... . Try the literal path
                # first, then the resolved one.
                log_dirs = []
                for candidate in (Path(cwd), Path(cwd).resolve()):
                    slug = re.sub(r"[^A-Za-z0-9]", "-", str(candidate))
                    d = cache_root / slug / f"mcp-logs-{server_name}"
                    if d not in log_dirs:
                        log_dirs.append(d)
                dir_looked_in = " or ".join(str(d) for d in log_dirs)
```

**And in both, inside the retry loop (`preflight.py:142`, `orchestrator.py:153`):**

```
You wrote:      if not log_dir.is_dir():
                    continue

Change it to:   log_dir = next((d for d in log_dirs if d.is_dir()), None)
                if log_dir is None:
                    continue
```

I applied exactly this to a scratch copy and checked it: `ruff` clean, both modules import, all 45 preflight tests pass unchanged, and a live CLI spawn through the orchestrator's own copy now returns `['STUB-MARKER: refused to start']`.

Please also add a test that passes a `cwd` whose resolved form differs from its literal form. Without one, nothing stops this regressing.

### 2. The abort throws away where it looked

Rule 2 of the issue says an empty capture must never read as "the server said nothing." Preflight honours that and prints the directory. The orchestrator discards it with `server_stderr, _ =`, and when the list is empty the message is byte-identical to the old one — I ran the base and PR versions side by side to confirm, for both the normal and backstop arms.

That is why the bug above is invisible: a genealogist on a Mac sees the old message with no hint that a capture was even attempted. This guard is what would have caught it on the first live run.

```
You wrote:      server_stderr, _ = _read_mcp_stderr_lines(
                    cwd=workspace, server_name=GENEALOGY_SERVER_NAME,
                    since=run_started_wall,
                )
            except Exception:  # noqa: BLE001 — see comment above
                server_stderr = []
            error = unavailable_message(
                entry, backstop=backstop, queries=queries,
                server_stderr=server_stderr or None,
            )

Change it to:   server_stderr, looked_in = _read_mcp_stderr_lines(
                    cwd=workspace, server_name=GENEALOGY_SERVER_NAME,
                    since=run_started_wall,
                )
            except Exception:  # noqa: BLE001 — see comment above
                server_stderr, looked_in = [], "(unresolved)"
            error = unavailable_message(
                entry, backstop=backstop, queries=queries,
                server_stderr=server_stderr or None,
            )
            if not server_stderr:
                error += f"\n(No server stderr captured; looked in {looked_in}.)"
```

Make the same change in the fallback arm at `orchestrator.py:2417`. This keeps the `make e2e-preflight` pointer, which is right — when nothing was captured, that pointer is still what the reader needs.

---

### What I did not check

- The orchestrator's three `mcp_unavailable` call sites end to end. Reaching them needs a real Anthropic credential. Filing #1743 for that was the right call; finding 1 came from reproducing the filesystem lookup in isolation and from a live CLI spawn, not from a live abort.
- The new CI step covers preflight only, and only on Linux — the platform where this bug does not appear.
- Your Windows verification. I took it as given and shaped the fix so it cannot change that path.
- Two cosmetic things I left alone as non-blocking: `(… N earlier lines dropped)` is appended after the kept lines rather than before them, and the next sentence runs onto the last captured stderr line (`STUB-MARKER: refused to start. An e2e run would have…`).

taiwo-stack:
Both fixed, per your patch almost exactly — thank you for the live macOS repro, that's exactly the kind of evidence a Windows-only dev can't produce.

**Finding 1 (macOS cwd resolution)** — fixed in both `_read_mcp_stderr_lines` copies: try the literal cwd, then `.resolve()`'d, deduped. Collapses to one candidate on Windows/Linux, so no behavior change there.

Added the regression test you asked for — and its **counterpart in `orchestrator.py`'s own test file too**, since the two copies are independently duplicated per rule 3 and a test on preflight's copy alone would leave orchestrator's just as unprotected. Both simulate the divergence via a `Path.resolve` monkeypatch rather than a real symlink, so they're deterministic on every platform including Linux CI (where a real filesystem symlink can't reproduce this the same way). Proved both catch the regression: stashed each half of the fix, watched its test go red, restored.

One thing your patch surfaced that I had to fix along the way: two *existing* preflight tests asserted exact equality on the returned directory string. On my dev machine (Windows), even a synthetic path like `/some/cwd` resolves to something else (drive-letter prepended), so `dir_looked_in` legitimately lists two joined candidates there too — not just in the new macOS test. Loosened those two to substring checks; they're about the literal/resolved list format, not this fix's correctness.

**Finding 2 (discarded `looked_in`)** — fixed at both call sites exactly as you wrote it: thread `looked_in` through, append `"(No server stderr captured; looked in {looked_in}.)"` when the capture is empty. Not separately unit-tested — both sites live inside `_consume`'s nested async closure, reachable only with a live SDK session (same gap #1743 already covers), and a test stubbing the string-formatting instead of the real call site would be exactly the "check that cannot fail" shape. Verified by direct comparison against your patch instead.

**Left alone**, matching your own "non-blocking, cosmetic" call: the truncation-note ordering and the sentence-runs-onto-the-last-line join. Fixing either touches an existing passing test's exact wording for a nit, not a correctness gap — happy to take them if you'd rather they be in scope.

Verified on this push: full `eval/harness` suite 2407 passed / 3 skipped (up from your 2376/2 — main's moved since your review), `ruff check .` clean, both new tests proven red-then-green.

promise-emmanuel (changes requested):
Does: Reads the Claude CLI's own per-MCP-server JSONL stderr log from the platform cache dir and folds the Server stderr:  lines into two existing failure messages — preflight's check-5 FAIL detail and the run's mcp_unavailable abort. Adds a --mcp-server-entry flag to preflight, a one-line dead-stub.js fixture, and a CI step that runs preflight against that stub and greps for the stub's marker.

Intent: Matches. All 10 files are explained by it, and every Didn't change claim holds — I checked CONSECUTIVE_TOOL_SEARCH_MISSES, should_abort_at_init and classify_server_status are untouched, genealogy_mcp_config gained no wrapper (only a different argument), and nothing on the abort path writes a file.

Checks: All green on 1ac8a1a8 — pytest, runlogs, e2e-fixtures, vitest, lockfile-drift, scan, plus server-pytest, label, senior-queue, pnpm workspace, eval CRUD UI. Nothing failed and nothing skipped its run on this head. The new "E2E preflight smoke" step is a step inside the existing pytest job, not a new required check — consistent with the PR's "no new required CI check" claim.

Still needs: Two approvals; it has zero. Every path in the diff is senior-developers-owned (*.py, *.yml, *.js); no path is genealogist-owned — /eval/tests/ does not match eval/harness/tests/. GitHub nonetheless still lists senior-genealogists as an outstanding request. That happens on harness-only PRs generally (#1775 got the same dual request), so treat it as ambient rather than a real gate — but confirm before anyone waits on a genealogist. gh api repos/…/codeowners/errors is clean.

Verified: Test-merged origin/main into the branch and ran the harness suite on the merged tree: 2461 passed, 2 skipped in 54.7s; ruff check . clean. Built the engine in the worktree first (a fresh worktree has no build/index.js, so check 5 SKIPs without it). Ran both halves of the smoke recipe live on Linux and reproduced the captured marker. Then three mutations, below.

Existing reviews: chesworthrm CHANGES_REQUESTED @2d59ff78 — head is now 1ac8a1a8, so the standing is stale and both blocking findings are fixed:

macOS cwd-resolution miss — resolved by 1ac8a1a8. Both copies now try the literal path then .resolve(), both got the regression test that was asked for, and dir_looked_in names both candidates on the miss path.
The abort discarding where it looked — resolved by 1ac8a1a8, in the primary abort path and the run_e2e_test fallback arm. The make e2e-preflight pointer is correctly kept when nothing was captured (pinned by test_message_keeps_the_preflight_pointer_when_no_stderr_was_captured).
The two cosmetic items chesworthrm left alone — both still stand, both still cosmetic. See the last finding.
One caveat on that resolution: chesworthrm verified the macOS fix on a scratch copy of their own patch, not on 1ac8a1a8 as committed. Nobody has run the committed code on a Mac. The only coverage there is a Path.resolve monkeypatch, and macOS is ~31% of the committed e2e corpus.

[MODERATE] (8/10) eval/harness/e2e/orchestrator.py:107 — the duplicated reader has nothing comparing the two copies
_read_mcp_stderr_lines is ~90 lines living twice, with a comment saying it is "kept in sync by eye, not import." Nothing compares them.

This repo already ruled on this exact shape. eval/harness/tests/unit/test_write_lockdown_parity.py exists because "Nothing compared them, so a change … could land in one and not the others" — written after a POSIX-only path split "made the e2e copy a silent no-op on Windows … between #914 and #984." This PR's own review cycle just repeated that story on macOS, and the fix had to be hand-applied to both copies.

I confirmed the two copies are AST-identical today (ignoring docstrings), so the guard is cheap. I wrote it, ran it, and mutation-tested it: it passes on the current tree and fails on a one-token drift (range(3) → range(4) in one copy).

The cited justification doesn't require duplication. Issue #1301's rule 3 says the filesystem read must not live in mcp_health.py — it does not say it must exist twice. A third module satisfies both that rule and CLAUDE.md § "Code reuse", and preflight.py already imports a sibling (from e2e.mcp_health import …), so there is no cycle risk.

Preferred fix — one copy:


You wrote:      # Bounds for _read_mcp_stderr_lines (issue #1301, rule 1) — same values as
                # preflight.py's own copy of this function; kept in sync by eye, not import
                # (rule 3: the filesystem read lives in the two callers, not a shared module).

Change it to:   (delete both copies of the function and its two _STDERR_MAX_* constants;
                add eval/harness/e2e/mcp_stderr.py holding them, and in each caller:
                from e2e.mcp_stderr import read_mcp_stderr_lines)
Minimum fix if you keep two copies — add this to test_e2e_orchestrator.py:


def test_the_two_stderr_readers_stay_identical():
    """`_read_mcp_stderr_lines` exists twice (issue #1301 rule 3), kept in sync
    by eye. The macOS cwd fix had to be applied to both by hand and nothing
    compared them. Same guard, same reason, as test_write_lockdown_parity.py."""
    import ast
    REPO = Path(__file__).resolve().parents[4]

    def body(path: Path) -> str:
        src = path.read_text(encoding="utf-8")
        for node in ast.walk(ast.parse(src)):
            if isinstance(node, ast.FunctionDef) and node.name == "_read_mcp_stderr_lines":
                if isinstance(node.body[0], ast.Expr) and isinstance(node.body[0].value, ast.Constant):
                    node.body = node.body[1:]  # docstrings differ by design
                return ast.dump(ast.parse(ast.unparse(node)))
        raise AssertionError(f"_read_mcp_stderr_lines not found in {path}")

    assert body(REPO / "eval/harness/e2e/preflight.py") == body(
        REPO / "eval/harness/e2e/orchestrator.py"
    ), "the two _read_mcp_stderr_lines copies have drifted"
[MODERATE] (7/10) .github/workflows/eval-harness-tests.yml:180 — the exit-code assertion cannot fire in CI, and its comment claims otherwise
The comment says exit-code-alone is needed because "with no credential at all resolve_auth() raises and check 5 degrades to WARN/exit 0, which would let this job pass while proving nothing." In CI, check 1 (FamilySearch login) FAILs unconditionally — there is no ~/.familysearch-mcp/tokens.json on a runner, and the PR's own CI log shows that FAIL. So preflight exits 1 no matter what check 5 does, and the RC -eq 0 branch is unreachable.

I proved it: I forced check 5 to return WARN and re-ran. RC=1, the branch did not fire, and only the marker assertion caught it. So the step is sound — but the assertion the comment presents as the safety net isn't the one doing the work, and a future reader trusting that comment could weaken the marker grep thinking the exit code has their back.

Make the assertion say what it means:


You wrote:      if [ "$RC" -eq 0 ]; then
                  echo "::error::preflight exited 0 against a dead stub — check 5 should FAIL, not pass"
                  exit 1
                fi

Change it to:   if ! echo "$OUTPUT" | grep -q "^\[FAIL\] MCP server connects:"; then
                  echo "::error::check 5 did not FAIL against a dead stub — it degraded to WARN/SKIP, or never ran"
                  exit 1
                fi
I checked that pattern against real output: it matches the [FAIL] MCP server connects: line and does not match the [WARN] one. Drop the "with no credential at all … exit 0" sentence from the comment above it at the same time, since the exit code was never what covered that.

[LOW] (9/10) eval/harness/e2e/preflight.py:521 — the next sentence still runs onto the last captured stderr line
Already raised by chesworthrm and left as non-blocking; I agree it isn't a blocker, but it is confirmed live in this PR's own CI log, and you'll be in the file anyway:


Captured server stderr:
STUB-MARKER: refused to start. An e2e run would have no genealogy tools at all — it would burn a live-FamilySearch session producing nothing…
A reader cannot tell where the server's words stop and the harness's begin.


You wrote:      f"{cause}. An e2e run would "

Change it to:   f"{cause}\nAn e2e run would "
The sibling wart — (… N earlier lines dropped) appended after the kept lines rather than before them — is asserted deliberately by test_read_mcp_stderr_lines_bounds_to_20_lines_with_a_truncation_note, so leave it unless you want to move both.

What I did not check
The orchestrator's three mcp_unavailable call sites end to end. Reaching them needs a real Anthropic credential. Same gap #1743 already owns; filing it was right.
macOS, live. I ran on Linux (WSL2) and CI runs Linux. The committed macOS fix rests on a Path.resolve monkeypatch plus chesworthrm's pre-fix live probe. Someone with a Mac running the two-line recipe would close the last real hole here.
The author's Windows verification. Taken as given.
The senior-genealogists request. I established that no path in the diff matches a genealogist rule in .github/CODEOWNERS, but I could not prove from the API why GitHub requested the team, so I've reported it as something to confirm rather than as settled.

taiwo-stack:
Fixed 2026-08-20, all three findings, at 4d7cb696:

- Parity test added (`test_the_two_stderr_readers_stay_identical`) — AST-compares the two `_read_mcp_stderr_lines` copies. Mutation-tested: catches a one-token drift.
- CI step's exit-code branch replaced with a grep on check 5's own `[FAIL]` line. Confirmed live in CI logs: check 1 fails unconditionally on the runner (no FS token), so the old exit-code check was dead code — the new grep is what actually proves check 5 failed.
- Sentence run-on fixed (`.` → `.\n` before "An e2e run would"). Confirmed live in CI's captured output.

Full harness suite: 2408 passed, 3 skipped. `ruff check .` clean.

promise-emmanuel (approved):
Approving. Re-ran the full pass at 323a3326 — all three findings fixed, each verified by
running it rather than reading it.

- **Parity guard** — mutation-tested: drifted `range(3)`→`range(4)` in the orchestrator's
  copy, test went red; restored, green.
- **Dead CI assertion** — forced check 5 to WARN and the new grep fires. The old
  exit-code version stayed silent on that same mutation, which was the point.
- **Run-on** — confirmed live in the smoke recipe. No stale assertion of the old string
  anywhere in the tree.

Suite on this head: 2483 passed, 2 skipped, `ruff` clean, 0 behind main. Also checked
main's merge didn't clobber anything: both literal-then-resolved fixes and both
`looked_in` call sites intact, 18 stderr tests green.

One thing to fold in — the guard has a hole in the half I asked for. The bounds are
duplicated too, and the signature references them **by name**, so a value drift leaves the
body's AST identical. I mutated `_STDERR_MAX_LINES` 20 → 50 in the orchestrator's copy and
the test passed. That is the drift the bounds exist to prevent (#941: an unbounded capture
buries the actionable line).

Append this inside `test_the_two_stderr_readers_stay_identical`:

    # The bounds are duplicated too, and the signature references them by NAME
    # -- so a value drift (20 vs 50) leaves the body's AST identical while the
    # two callers bound differently. That is the very defect the bounds exist
    # to prevent (#941: an unbounded capture buries the actionable line).
    def bounds(path: Path) -> dict[str, object]:
        module = ast.parse(path.read_text(encoding="utf-8"))
        return {
            t.id: ast.literal_eval(n.value)
            for n in module.body
            if isinstance(n, ast.Assign)
            for t in n.targets
            if isinstance(t, ast.Name) and t.id.startswith("_STDERR_MAX_")
        }

    pf_bounds = bounds(repo / "eval/harness/e2e/preflight.py")
    assert pf_bounds, "the _STDERR_MAX_* bounds vanished from preflight.py"
    assert pf_bounds == bounds(
        repo / "eval/harness/e2e/orchestrator.py"
    ), "the duplicated _STDERR_MAX_* bounds have drifted"

Mutation-tested three ways: passes clean, fails on a `_STDERR_MAX_LINES` drift, fails on a
`_STDERR_MAX_CHARS` drift, and still fails on body drift.

@chesworthrm your CHANGES_REQUESTED is at 2d59ff78, three pushes back — both your findings
landed at 1ac8a1a8. Worth re-reviewing or dismissing. This still needs a second approval.

chesworthrm (changes requested):
**Changes requested.** Both my findings from last round are fixed, and the macOS half now works — I ran the committed code on a Mac rather than my own patch, so that caveat Promise raised is closed. Two things left, and the second changes where the first one goes, so do them in that order.

Also: the branch is 23 commits behind `main`, so it needs a merge before it can go in regardless.

### Does / intent

Reads the CLI's own per-MCP-server JSONL stderr log and folds the `Server stderr:` lines into preflight's check-5 FAIL detail and the run's `mcp_unavailable` abort. Adds a `--mcp-server-entry` flag to preflight, a dead-stub fixture, and a CI step that greps for the stub's marker. Intent matches — every path is explained, and I checked each **Didn't change** claim myself. They all hold.

---

### 1. Move the reader into one module

`_read_mcp_stderr_lines` exists twice, and the parity guard you added compares the two function bodies as ASTs. Both copies reference `_STDERR_MAX_LINES` and `_STDERR_MAX_CHARS` **by name**, so a value drift leaves both ASTs identical. Set one copy's line bound to 3 and the other's stays at 20, and the two return different results from the same log while the guard and the whole suite stay green.

`test_write_lockdown_parity.py` — the test you modelled this on — already says why a source comparison is the wrong shape:

> **Vector-driven, not a string diff.** … Comparing source would fail on cosmetics and pass on a real divergence.

That test also lifts the module-level constants into the namespace, which is exactly why they're covered there.

You can close the hole with a constants check, and Promise wrote that patch. I'd rather you delete the duplication instead. Rule 3 says the filesystem read must stay out of `mcp_health.py`. It doesn't say it has to exist twice, and both callers already import from a sibling module, so a third one costs nothing. The write-lockdown copies are duplicated because they live in three packages that can't import each other. These two are in the same package.

```
You wrote:      # Bounds for _read_mcp_stderr_lines (issue #1301, rule 1) — same values as
                # preflight.py's own copy of this function; kept in sync by eye, not import
                # (rule 3: the filesystem read lives in the two callers, not a shared module).
                _STDERR_MAX_LINES = 20
                _STDERR_MAX_CHARS = 200

Change it to:   (delete this block and the whole _read_mcp_stderr_lines function from
                BOTH preflight.py and orchestrator.py. Add eval/harness/e2e/mcp_stderr.py
                holding the two constants and the function, renamed
                read_mcp_stderr_lines. In each caller:
                    from e2e.mcp_stderr import read_mcp_stderr_lines)
```

Then delete `test_the_two_stderr_readers_stay_identical` and the orchestrator's copy of the cwd-resolution test. Keep preflight's.

### 2. Split each log row before bounding it

One JSONL row is one **write**, not one line. A Node module-resolution crash — a stale or broken build, the most likely real reason the server is dead — arrives as a single row carrying its whole stack trace. I measured one at 599 characters with 13 newlines in it.

So `_STDERR_MAX_LINES = 20` never engages, `(… N earlier lines dropped)` can never fire, and the 200-character cap is the only bound doing anything. It cuts the trace mid-path with no marker, and the trailing period makes it read as a finished sentence:

```
Captured server stderr:
node:internal/modules/cjs/loader:1215
  throw err;
  ^

Error: Cannot find module '@modelcontextprotocol/sdk/server/stdio.js'
Require stack:
- /Users/dev/cowork-genealogy/packages/engine/mcp-server/bu.
```

Every stack frame is gone. And because the capture isn't empty, `unavailable_message` drops the `run make e2e-preflight` line — so the reader loses the rest of the trace *and* the pointer to it.

```
You wrote:                      text = error[len("Server stderr: "):].rstrip("\n")
                                lines.append(text[:max_chars])

Change it to:                   # One JSONL row is one *write*, not one line: a Node
                                # crash arrives as a single row carrying its whole
                                # stack trace (measured: 599 chars, 13 newlines).
                                # Split before bounding, or max_lines never engages
                                # and max_chars cuts the trace mid-frame.
                                body = error[len("Server stderr: "):]
                                for text in body.split("\n"):
                                    if text.strip():
                                        lines.append(text[:max_chars])
```

The `text.strip()` matters on its own: a server that writes a bare newline to stderr currently produces an empty `Captured server stderr:` block and loses the pointer too. With this, it degrades to the "looked in …" message instead.

No existing test needs changing for either half. The 200-character test uses a line with no newlines, and the 20-line test uses 25 separate rows.

### Existing reviews

- My CHANGES_REQUESTED at `2d59ff78` — both findings resolved at `1ac8a1a8`. Superseded by this review.
- Promise's CHANGES_REQUESTED at `1ac8a1a8` — all three resolved at `4d7cb696`.
- Promise's approval at `7884182f` — reads the current head, so it stands, and it satisfies the code-owner requirement. The bounds note inside it is finding 1 above; item 1 answers it by deletion rather than by patch.

### Not asking for

The period that lands on the last captured line, a truncation marker on the 200-character cap, and documenting `--mcp-server-entry` in the e2e guide. All real, none blocks merge. If you're in the file for item 2 anyway, the period is one character.

taiwo-stack:
Verified 2026-08-24: cross-link noted — this PR's capture never reaches the bridge-abort case anyway (neither harness goes through the bridge, per docs/architecture.md), so no scope change here; the mcp_unavailable/bridge-timeout distinction is recorded for whoever picks up #1479 next.

chesworthrm (approved):
**Approving.** Both my findings from last round are resolved by `3cf0b256`, and I re-verified the whole thing at head `7d79c22e` rather than reading it.

- **One module** — `_read_mcp_stderr_lines` is gone from both callers; they now import `read_mcp_stderr_lines` from the new `e2e/mcp_stderr.py`. The by-name parity hole and the parity test are both gone with it.
- **Split before bounding** — landed exactly, `text.strip()` included.
- **Live on macOS** — I ran the acceptance recipe against your dead stub on a Mac (the `/var` vs `/private/var` case): check 5 FAILs and prints `STUB-MARKER: refused to start`. The cwd-slug handling works on the committed code.

Suite in the PR worktree: 2685 passed, 8 skipped, `ruff` (F821) clean, new unit tests 162 passed. The one failure (`test_mock_mcp::test_successful_call_carries_no_is_error`) is a fresh-worktree artifact — no engine build, so the mock server can't find `research-query.js`; CI builds it first, which is why CI is green. Not this PR.

Note for whoever merges: promise-emmanuel's approval is at `7884182f`, and `3cf0b256` (the module extraction + split) landed after it. The `protect-main` ruleset has `dismiss_stale_reviews_on_push: false`, so that approval still counts — but I'm the only one who's reviewed the extraction. This approval covers it.

---

Three non-blocking items — none blocks merge, all fixable here if you're in the file:

**1. Add a regression test for the split (worth doing before merge).** This was the whole point of `3cf0b256`, but nothing guards it — I reverted `body.split("\n")` back to a single `rstrip`+append and all 162 tests stayed green. No test feeds one JSONL row containing multiple newlines (the Node-stack-trace shape). A one-row-many-newlines test closes it.

**2. Three unused module-level imports in `preflight.py`.** `json` (43), `os` (44), `re` (45) were all added here, and each is shadowed by a local import in the one function that uses it (`json`@443, `os`@141, `re`@452). `ruff --select F401` flags all three; our CI ruff only selects `F821`, so it doesn't catch them. Deleting all three leaves the module importing fine and every preflight test green.

```
Delete: import json / import os / import re  (lines 43-45)
```

**3. The PR body is stale.** "Start here" and "Unsure about" still say `orchestrator.py` carries "its own copy" and "both copies of `_read_mcp_stderr_lines`" — that was true before `3cf0b256` folded them into one module. Worth updating so history isn't misleading.

**What I did not check:** the orchestrator's three `mcp_unavailable` sites end-to-end (needs a live Anthropic credential; #1743 owns that gap), and Windows live (taken as given; the `LOCALAPPDATA` path and the two-candidate collapse are unit-covered).

johnmarkpeterbrown (changes requested):
Full pass, reviewed as the merge with `main` (the branch is 7 behind). Test-merged cleanly and ran the harness suite on the merged tree: **2721 passed, 9 skipped**, `ruff check .` clean. On the PR head alone: 2685 passed, 9 skipped. All 13 required checks are green on `7d79c22e`, and the "E2E preflight smoke" step ran rather than skipping.

This works, and the mechanism is sound. Two things to fold into the push you have to make anyway.

Three things I checked rather than assumed:

- **The capture works live on Windows.** chesworthrm took this as given, so I ran it. Spawned a real CLI session against your dead stub, then called the committed `read_mcp_stderr_lines` against what it wrote: `['STUB-MARKER: refused to start']`, out of `%LOCALAPPDATA%\claude-cli-nodejs\Cache\<slug>\mcp-logs-genealogy`. All three platforms now have live evidence behind them.
- **Your new CI assertion can actually fail.** I went looking for the way it couldn't. Preflight registers no `stderr` callback, and the SDK pipes the CLI's stderr *only* when one is registered — so the CLI's stderr is inherited by preflight, and the stub's marker could have leaked straight into `$OUTPUT` and satisfied your grep with a completely broken capture. It doesn't. A live spawn produced the failed-status entry with no marker anywhere in the inherited output. The grep genuinely requires the JSONL read.
- **The `--mcp-server-entry` threading and the closure cell.** Both hold. The ten existing tests that inject `status_getter` are untouched, and the `seen_since == [0.0]` test pins that.

---

### 1. Nothing guards the split-before-bounding fix

Half of `3cf0b256` exists so a Node crash — one JSONL row carrying a whole stack trace — gets split before the bounds apply. No test feeds a row with an embedded newline. Every fixture row is a single line with one trailing `\n`.

I reverted `body.split("\n")` back to a single `rstrip`-and-append and **all 162 tests stayed green**. Without a guard, this silently regresses to the exact shape you fixed: `max_lines` never engages, and `max_chars` cuts the trace mid-frame.

Add this to `test_e2e_preflight.py`, alongside the other bounding tests:

```
Add:  def test_read_mcp_stderr_lines_splits_one_row_carrying_a_multi_line_trace(monkeypatch, tmp_path):
          """One JSONL row is one WRITE, not one line: a Node crash arrives as a
          single row carrying its whole stack trace. Without the split, max_lines
          never engages and max_chars cuts the trace mid-frame."""
          _use_linux_cache_root(monkeypatch, tmp_path)
          cwd = Path("/some/cwd")
          log_dir = _stderr_log_dir(tmp_path, cwd, "genealogy")
          log_dir.mkdir(parents=True)
          (log_dir / "x.jsonl").write_text(
              '{"error": "Server stderr: Error: boom\\n    at one (/a.js:1:1)\\n'
              '    at two (/b.js:2:2)\\n"}\n',
              encoding="utf-8",
          )
          lines, _ = pf.read_mcp_stderr_lines(cwd=cwd, server_name="genealogy", since=0)
          assert lines == [
              "Error: boom",
              "    at one (/a.js:1:1)",
              "    at two (/b.js:2:2)",
          ]
```

Mutation-tested both directions before proposing it: passes on your code, fails on the reverted split, and the file stays green at 53 passed.

### 2. Three dead imports in `preflight.py`

`json`, `os` and `re` were added by this PR and went dead when `3cf0b256` moved the reader into `mcp_stderr.py`. Each is shadowed by a function-local import in the one function that uses it. Our CI ruff selects `F821` only, so nothing catches them.

```
You wrote:      import json
                import os
                import re

Change it to:   (delete all three lines)
```

Keep `import time` — line 366 needs it at module scope now.

Deleting all three leaves the module importing fine, 52 preflight tests green, and `ruff check .` clean.

---

### Note for whoever merges

promise-emmanuel's approval is at `7884182f`. `3cf0b256` landed after it — that's the commit that created the shared module, deleted both inline copies, and removed the parity test and the bounds-drift guard promise-emmanuel wrote. `protect-main` has `dismiss_stale_reviews_on_push: false`, so that approval still counts toward the required two while covering none of it. chesworthrm's approval at `7d79c22e` is the only one that has read the extraction.

Separately: `strict_required_status_checks_policy` is on and the branch is 7 behind, so it cannot merge until you bring it up to date regardless.

### What I did not check

- The orchestrator's three `mcp_unavailable` call sites end to end — that half has no test at all (the base-to-head diff of `test_e2e_orchestrator.py` is two blank lines). Reaching them needs a live Anthropic credential. #1743 owns it, correctly labelled and blocked on this PR, so I am not counting it as a finding.
- macOS live — I took chesworthrm's Mac run of the committed code as given.
- One dormant thing I could not make reachable, so it is not a finding: the retry loop takes the first *existing* candidate directory and never falls through to the second, so a literal slug would win over the resolved one the CLI writes to. I could not construct a live path where both exist — the orchestrator's workspace is a fresh tempdir, preflight's `Path.cwd()` is already resolved.
- The two cosmetic items from earlier rounds (truncation note placed after the kept lines; the following sentence's period gluing onto the last captured stderr line) still stand and still do not block.

taiwo-stack:
Fixed in 5b4d9838.

- Added a regression test for the multi-newline split: fed one JSONL row with an embedded stack trace, mutation-tested by reverting the split to a single rstrip+append — the new test failed, restored.
- Removed the three shadowed module-level imports (json/os/re) from preflight.py.
- PR body was already current — no stale "own copy" language found.

Harness suite: 2698 passed, 3 skipped. CI green.

johnmarkpeterbrown (approved):
Looks good!

PR #1751, opened 2026-08-19
Electron packaging invocation 1070

### LINKED ISSUE #1070: Electron packaging is invocation-dependent: npm run build:* silently ships an app that cannot launch
> **Reviewed 2026-08-12 before junior handoff.** The premise moved: the release
> pipeline's own invocation packages correctly today, and the recipe this issue gave for
> *verifying a fix* is itself the reproducer. The ask is rewritten; the original body
> follows under "Original issue" — its sections 2 (Electron 43 blockers) and 3 (the
> `executableName` fix) are still accurate and are the reference for whoever takes
> Dependabot's next 43 PR.

**Touches:** apps/electron/package.json, apps/electron/scripts/check-packaged-deps.mjs (new),
.github/workflows/electron-release.yml, apps/electron/Makefile

## What is actually broken

The same tree and the same command produce two different artifacts depending on which
package manager's environment you happen to be in. Measured 2026-08-12 on `main`
(electron-builder 26.15.3, in the lockfile since commit 861b21c3):

- **Works** — `cd apps/electron && pnpm build && pnpm exec electron-builder --dir`,
  which is exactly what `.github/workflows/electron-release.yml` runs. All five direct
  dependencies are in `app.asar`; the app opens a window (3 `Research Viewer Helper`
  processes).
- **Broken** — `cd apps/electron && npm run build:unpack`, which is what `make
  build-mac`, `make build-win`, `make build-linux` and the old version of this issue's
  verification recipe run. electron-builder logs, at `warn`:
  `cannot find path for dependency dependencies=["@electron-toolkit/utils@undefined","chokidar@undefined","focus-trap-react@undefined","jszip@undefined","react-markdown@undefined"]`
  The three modules `out/main/index.js` top-level-requires are absent from `app.asar`;
  the app starts with **zero** renderer helpers, no window, and nothing on stderr.

The cause is not the pnpm hoisted layout. `apps/electron/package.json` has no
`packageManager` field and `apps/electron/` has no lockfile, so electron-builder falls
through both file-based detections and guesses from `npm_config_user_agent`
(`app-builder-lib/out/node-module-collector/packageManager.js`, `detectPackageManager` →
`detectPackageManagerByEnv`). Under pnpm it uses the pnpm collector, which handles
`node-linker=hoisted`; under npm it uses the npm collector, which sees only
`apps/electron/node_modules/@genealogy` and resolves nothing.

**The release pipeline was never broken.** The 2026-07-31 dry run
(`gh run view 30655120136 --repo PioneerAIAcademy/cowork-genealogy --log`) resolved
`pnpm` on all three runners and logged no `cannot find path` warning, even on
electron-builder 26.15.0. Its mac leg failed on the `hdiutil` flake and its linux leg on
`executableName` (fixed in PR #1061); the win leg succeeded.

## The fix

1. Add `"packageManager": "pnpm@9.15.9"` to `apps/electron/package.json`, matching the
   root. electron-builder reads that field before it guesses from the environment, so
   `npm run`, `pnpm`, `make` and CI all resolve the same collector. (Established by
   reading the detection order, not by running it — prove it by running both commands
   above and comparing.) This beats bundling the deps (see the fallback) because it
   fixes the resolution rather than routing around it, and leaves the shipped main
   bundle unchanged.
2. Add `apps/electron/scripts/check-packaged-deps.mjs`: parse `out/main/index.js` for
   top-level `require("…")`, drop `electron` and node builtins, and fail when a
   remaining module is neither present under the packaged `app.asar`'s `node_modules`
   nor bundled into the main file. **Fail loudly when it finds no `app.asar` under
   `dist/`** — a check that silently inspects nothing is exactly the failure this issue
   exists to prevent (`CLAUDE.md` § "A new lint must be proven to fail": break it and
   watch it fail before committing). Node, not Python, so the `encoding="utf-8"` rule
   does not come into it.
3. Run it from `.github/workflows/electron-release.yml` after **each** of the three
   packaging steps — the asar path differs per OS (`dist/mac-arm64/Research
   Viewer.app/Contents/Resources/`, `dist/win-unpacked/resources/`,
   `dist/linux-unpacked/resources/`) — and add a `make` target the local build targets
   call.

**Fallback, only if step 1 does not hold:** `main: { build: { externalizeDeps: false } }`
in `apps/electron/electron.vite.config.ts`. electron-vite defaults
`build.externalizeDeps` to `true` and keeps `electron` and node builtins external
unconditionally, so it is safe on that axis — but chokidar 5 is ESM-only and jszip pulls
`readable-stream`/`setimmediate`, so a clean build proves nothing. Only launching the
packaged app does.

## How to verify

Run both, on the same machine, and compare:

```
cd apps/electron
npm run build:unpack                              # was broken; must now match the line below
pnpm build && pnpm exec electron-builder --dir
```

For each artifact: run the new check, **and launch the packaged app and count its
renderer processes** — `ps -ax | grep "Research Viewer Helper" | grep -v grep | wc -l`
is 0 when broken and >= 1 when correct, with no error message either way. A grep of
`out/main/index.js` is not sufficient evidence, and neither is "`app.asar` contains a
`node_modules`": in the broken artifact it contains 95 packages — every transitive child
of the five that are missing.

Then the three-OS dry run: `gh workflow run electron-release.yml`, download the
`mac-dist`, `win-dist` and `linux-dist` artifacts, and check each. The linux
`executableName` fix (PR #1061) and the macOS `hdiutil` DMG flake have both never been
exercised by CI — re-verify them in the same pass so neither is mistaken for this bug.

**Nothing checks any of this today.** No CI job launches the packaged desktop app, and
the missing-dependency condition is a `warn`, so a broken artifact is a green build.
That is what step 2 closes, and why this issue carries `nothing-checks`.

No eval cost: no `SKILL.md`, rubric or fixture is touched, so no `make eval-skill` run
and no run-log annotation.

Out of scope: do not bundle the Electron 43 upgrade (section 2 of the original below is
reference for whoever takes Dependabot's next 43 PR), and do not touch `.npmrc` —
`node-linker=hoisted` governs the whole workspace and the engine's npm tooling.

## Original issue

> **Analysis 2026-08-09 — the diagnosis holds, the fix is one config line, it is junior-safe, and it does not gate beta.**
>
> **Framing correction: this is not a missing-dependency problem.** All three modules **are** declared in `apps/electron/package.json` (`@electron-toolkit/utils ^4.0.0`, `chokidar ^5.0.0`, `jszip ^3.10.1`). They are externalised *because* they are declared — electron-vite auto-applies `externalizeDepsPlugin()` unless told otherwise, and that marks every declared dependency external. `electron.vite.config.ts` has a bare `main: {}`, so all three become external `require`s in the main bundle, physically reachable only via the hoisted root `node_modules`.
>
> **The fix: `main: { build: { externalizeDeps: false } }`** in `apps/electron/electron.vite.config.ts`. Safe because electron-vite keeps `electron` and node builtins external unconditionally, all three modules are pure JS with no native binding (`npmRebuild: false` is already set), and nothing in the built main bundle requires a workspace package. After it the asar needs no `node_modules` at all and the pnpm-vs-electron-builder resolution question stops mattering. **Do not change `.npmrc`'s `node-linker`** — it governs the whole workspace and the engine's npm tooling, to fix one app's packaging.
>
> **Do step 0 first:** the one CI run that produced the broken artifact logged `electron-builder version=26.15.0`, and commit `861b21c3` later put **26.15.3** in the lockfile — which added explicit hoisted-layout handling (it shells `pnpm config list` and skips the downward search only when the linker is *not* hoisted). Nobody has rebuilt since. **Rebuild on current `main` and re-parse the asar before changing anything** — the repro may already have moved.
>
> **Why nobody noticed:** the missing-dependency condition is logged at `warn`, not `error`. The single release-workflow run has a **passing** Windows leg that produced an installer. Green build, non-launching artifact.
>
> **Do not bundle the Electron 43 upgrade.** Issue #1046 is a closed Dependabot PR and already stale — latest is 43.3.0 with a 44 beta out. The packaging bug is version-independent by this body's own testing, so 43 buys nothing toward the launch fix and adds three variables to a build you are trying to make deterministic. All three listed 43 blockers still hold, and the electron-vite remedy is still unreleased (latest 5.0.0; the fix is in a 6.0.0 beta).
>
> **Junior-safe.** Mechanically falsifiable — the build either produces a launching app or it doesn't. No doctrine call, no cross-cutting design. CODEOWNERS routes `*.ts`/`*.yml` to senior-developers, so a senior reviews it regardless; the `senior` label buys nothing here. **Add these two lines to the body before handing it out:** do not touch `.npmrc`, and do not bundle the 43 upgrade.
>
> **It does not gate beta, on repo evidence.** Beta is defined as "a wider variety of in-house users across roles and skill levels" with **no surface named**. The hosted alpha guide is a browser app with no install. The Cowork guide's step 5 is `pnpm --filter @genealogy/electron dev` — **run from a clone, never a packaged build** — so the broken installer blocks nobody today. And the hosted surface consumes the same `@genealogy/viewer-ui` with the same feedback loop. This gates the *option* of a desktop-installer beta, which is a product call, not a build call. `fill-ready`'s beta-gating set names this issue; that membership should be re-derived.
>
> **Acceptance check** — the current one is human-only and mac-only. Add: (1) after `npm run build`, assert the built main bundle contains no `require("chokidar")` / `require("jszip")` / `require("@electron-toolkit/utils")`; (2) `npm run build:unpack`, launch with `--remote-debugging-port=9222`, `curl http://127.0.0.1:9222/json/list` and assert at least one `"type":"page"` target — the CDP machinery already exists in `apps/electron/Makefile`; (3) the three-OS `workflow_dispatch` dry run. **Insist on wiring (1) into `.github/workflows/electron-release.yml`**, because the failure mode is a green build. "Nothing checks that the packaged desktop app launches" is not currently in the architecture guide's what-nothing-checks table and should be.
>
> Two knowns to re-verify in the same pass so they aren't mistaken for this bug: the Linux `executableName` fix has never been exercised by CI, and the macOS DMG `hdiutil` flake will likely recur.

---

The packaged Electron app is non-functional, and the release pipeline has never successfully run. Found while verifying the Dependabot batch on 2026-07-31 (#1040, #1046). **Nothing has shipped broken** — no `electron-v*` tag exists — so this is not urgent, but it does mean the desktop viewer cannot currently be released at all.

## 1. Packaged app launches with no window (the blocker)

`out/main/index.js` top-level-`require`s `@electron-toolkit/utils`, `chokidar`, and `jszip`, but **none of the three is present in `app.asar`** (verified by parsing the asar header directly: `chokidar=false jszip=false toolkit=false`). Launching either packaged `.app` leaves the process alive with **zero renderer helpers** — no window.

Present on **both** Electron 39.8.5 (current `main`) and 43.2.0 (#1046), so it is not version-related. `electron-vite preview` works fine because it runs against the source tree where the hoisted root `node_modules` is reachable — so this is a packaging-layout problem: pnpm `node-linker=hoisted` vs. electron-builder searching from `apps/electron`, which logs `cannot find path for dependency dependencies=["@electron-toolkit/utils@undefined","chokidar@undefined","focus-trap-react@undefined","jszip@undefined","react-markdown@undefined"]`.

Fixing this is the prerequisite for any Electron release.

## 2. Electron 43 upgrade blockers (why #1046 is on hold)

#1046 (`electron` 39.8.5 → 43.2.0) is closed for now. It packages cleanly and the app runs under 43 — but it can't be justified while item 1 stands, since the Chromium 142 → 150 security benefit only matters once something ships. Re-open it together with these:

- **Dev loop breaks on a fresh clone.** Electron 42 removed the postinstall binary download, so after `pnpm install --frozen-lockfile` there is no `node_modules/electron/dist` and `electron-vite@5`'s `getElectronPath()` throws `Error: Electron uninstall`. `npm run dev` / `pnpm dev:electron` fail out of the box; self-heals after one `node -e "require('electron')"`. Builds and CI are unaffected. Upstream: [alex8088/electron-vite#752](https://github.com/alex8088/electron-vite/issues/752).
- **`electron-vite@^5.0.0` is not Electron-43-aware.** Its `getElectronNodeTarget()` / `getElectronChromeTarget()` maps stop at Electron 39, and the unmapped fallback returns the *oldest* entry — so 43 silently builds against `node16.17` / `chrome108`. No practical impact today (output was byte-identical to the 39 build), but it is a latent trap. electron-vite 6 should carry the 40–43 maps and the post-42 path handling.
- **`dialog` defaultPath changes on 43.** Electron 43 defaults `defaultPath` to the user's Downloads folder. `src/main/index.ts` calls `dialog.showOpenDialog` twice (`open-file`, `project:select-folder`) without one, so both pickers would open in ~/Downloads. Cosmetic, but genealogists picking a project folder will notice. Set it explicitly.

## 3. Fixed already

`electron-builder.yml` had no `executableName` under `linux:`, so AppImage builds failed with `executableName contains characters that cannot be safely used in file paths: @genealogyelectron`. Fixed in #1061.

## Not part of this

The macOS leg of `electron-release.yml` also failed on `hdiutil: create failed - Device not configured` — the standard macOS-runner DMG flake, not a real defect.

## How to verify a fix

```
pnpm install --frozen-lockfile
cd apps/electron && npm run build:unpack
open dist/mac-arm64/Research\ Viewer.app   # must show a window
```

No CI job covers this; `electron-release.yml` runs only on `electron-v*` tags and `workflow_dispatch`.

aghadiayeamayanvboernest:
Closes #1070.

## The bug

The same tree and the same command produce two different artifacts depending on which package manager's environment you happen to be in. Measured on this branch, macOS, electron-builder 26.15.3:

| Invocation | Packages in asar | Declared deps present | Renderer helpers |
|---|---|---|---|
| `pnpm build && pnpm exec electron-builder --dir` | 120 | 7 of 7 | 3 |
| `npm run build:unpack` | 102 | **2 of 7** | **0** |

The broken one exits with no window and prints nothing. electron-builder logs `cannot find path for dependency` at **warn**, so the build is green.

## The fix

`"packageManager": "pnpm@9.15.9"` in `apps/electron/package.json`.

I traced the detection rather than inferring it: `detectPackageManager` (`app-builder-lib/out/node-module-collector/packageManager.js:56`) reads the `packageManager` field from `[workspaceRoot, projectDir, appDir]` in order, and only falls back to guessing from `npm_config_user_agent` when no directory yields one. `apps/electron` had neither the field nor a lockfile of its own, so it fell through to the guess — pnpm's collector under pnpm (which handles `node-linker=hoisted`), npm's collector under npm (which resolves nothing).

After the pin, both invocations produce the identical 120-package asar. The 2 dependencies that survive without it are the `workspace:*` packages, which is where the issue's original "five missing" figure came from. Earlier revisions of this description said 124/105 and "0 of 7" — those counts double-counted npm scope directories as packages, fixed in `check-packaged-deps.mjs`.

Note the root `package.json` already carried a `packageManager` field and it is **not** sufficient — the app-level field is what fixes this, so the line is not redundant.

## The guard — the actual point

The issue carries `nothing-checks` because a broken artifact is a **green build**. One config line that could be dropped by a merge, a revert or a Dependabot bump is not a fix on its own, so there are three layers:

**1. `check-packaged-deps.mjs`** — fails the build when the packaged `app.asar` is missing a declared dependency.

It compares **declared dependencies** rather than scanning the bundle for `require()` calls. The scan was my first design and review found it had a silent no-op: with no external requires found, it had nothing to compare and passed. That state is reachable — an ESM main bundle, or the `externalizeDeps: false` fallback this issue keeps on the table — so a fully broken package would have gone green, which is the exact failure the check exists to prevent. Reading the declared list cannot degrade that way, and it covers the preload bundle for free.

It reads files only and never launches Electron, which also sidesteps the trap below.

**2. `packaging.test.ts`** — asserts the field exists, matches the root, and that each `build:*` script still runs the check. Runs on every PR touching `apps/electron/`, costs no build. The guard above only fires once something has been packaged, and `electron-release.yml` runs only on a tag or manual dispatch — so without this, a dropped field goes unnoticed until someone cuts a release.

**3. Wiring** — the check runs from all four `build:*` npm scripts (so `pnpm build:mac` straight from the README is covered, which no make target is) and after each of the three packaging steps in `electron-release.yml`.

## ⚠️ The issue's own verification recipe is misleading

The issue says to count `Research Viewer Helper` processes, `0` when broken and `>= 1` when correct.

**`ELECTRON_RUN_AS_NODE=1` is set in the VS Code integrated terminal.** With it set, a *correct* artifact also reports zero helpers — Electron runs as plain Node and never starts a browser process. Anyone verifying this fix from their IDE terminal would conclude it had not worked.

`env -u ELECTRON_RUN_AS_NODE` is the fix for the recipe. This is one reason the committed check parses the asar instead of launching anything.

## Verification

Everything below was run, not reasoned about.

**The guard was proven to fail before it was trusted:**

| Case | Exit | Behaviour |
|---|---|---|
| real good asar | 0 | 120 packages, all 7 present |
| genuinely broken artifact | 1 | names all 7 missing |
| corrupt + good asar together | 1 | corrupt diagnosed, **good one still inspected** |
| empty dir / missing path / non-`.asar` file | 2 | never a silent pass |

Both robustness guards were mutation-checked: revert the corrupt-archive catch and 0 good asars get inspected behind a buffer-bounds stack; revert the directory catch and a raw `fs` stack replaces the exit-2 message.

My first run of that matrix reported `exit=0` on every failing case — I was reading `tail`'s exit code through a pipe, not the script's. Re-measured without the pipe. Flagging it because it is the same shape as the can't-fail assertions I shipped earlier this month.

`make test-all` — all checks passed. eslint and prettier clean on all changed files.

## One correction to an earlier commit message on this branch

An earlier commit here claimed `@typescript-eslint/no-unused-vars` did not fire for `.mjs` files, and added a rule to "fix" it. That was wrong — I removed the block and watched the rule report correctly. The rule was never missing.

The real reason an unused import shipped is that **at the time nothing ran eslint**. There is still no `lint` task in `turbo.json`, but `make lint`, `scripts/test.sh` and both `js-tests.yml` jobs now run it — added in this PR. The no-op rule is removed and the comment now states the actual cause. The `explicit-function-return-type: 'off'` half is real and stays — that rule is unsatisfiable in a `.mjs` file. Running eslint properly also caught a missing return type in my own new test.

## Not in this PR

- **eslint in CI landed here after all.** `make lint`, `scripts/test.sh`, `scripts/windows/lint.bat` and both `js-tests.yml` jobs now run ESLint at `--max-warnings 0` over `apps/electron` and `eval/app`. All 6 `no-explicit-any` errors and 5 `prettier/prettier` errors are fixed. The other three JS trees declare no lint script, so there is nothing left to switch on.
- **Publish ordering.** On a tag push the gate runs after `electron-builder --publish always`, so a broken installer is uploaded before the check can fail. It lands in a **draft** release (`electron-publish/out/gitHubPublisher.js:55` — verified, not assumed), so nothing reaches users without a human publishing it, and the job fails loudly. Splitting build from publish means replacing electron-builder's upload — which also writes `latest.yml` for auto-update — on a release path that has never successfully run end to end. I would rather not rebuild that blind; happy to if a senior wants it in scope.
- **The Electron 43 upgrade and `.npmrc`**, both explicitly out of scope per the issue.

## My new guards run in checks that cannot block a merge

`protect-main` requires `pytest`, `runlogs`, `e2e-fixtures`, `vitest`, `lockfile-drift` and `scan`. `packaging.test.ts` and both ESLint steps run under **`pnpm workspace`**, which is not on that list. The required `vitest` check comes from a different workflow — the MCP server — and on this PR every one of its steps skipped, so it reported success having run nothing about `apps/electron`.

**Resolved while this PR was open:** `pnpm workspace` and `server-pytest` were added to `protect-main`'s required checks on 2026-08-20. `packaging.test.ts`, the new behavioural guard tests and the `apps/electron` ESLint step now all block a merge. The `eval/app` ESLint step still runs under `eval CRUD UI`, which is not required.

## Still unexercised

The three-OS dry run is **done** — run 32342126397, green on all three legs, four asars each reporting 120 packages and 7 of 7 dependencies. `clean:dist` under `cmd.exe` and `lint.bat` remain unexercised by anything. The Linux `executableName` fix (#1061) and the macOS `hdiutil` DMG flake both still have zero CI coverage and should be re-checked in that same pass so neither is mistaken for this bug.


chesworthrm (commented):
Reviewed the full diff against a test-merge with `main` (10 commits behind; merges clean, and everything is green on the merge).

**This is strong work.** I re-ran your verification matrix instead of taking it on trust and every case reproduced, including the real build both ways: with the pin, `pm=pnpm`, 120 packages, 7 of 7 dependencies; without it, `pm=npm`, `cannot find path for dependency`, **electron-builder still exits 0**, and the guard catches it. Both self-corrections check out too — `no-unused-vars` does fire on `.mjs` files, and the `explicit-function-return-type` disable is load-bearing (6 real errors without it). All five assertions in `packaging.test.ts` fail on their own mutation.

## One thing to change before merge

The release workflow is the only place your gate fires on an actual release, and nothing asserts it is still there.

`electron-release.yml` does not use the `build:*` scripts. It calls `pnpm build && pnpm exec electron-builder` directly, so it inherits none of the wiring `packaging.test.ts` protects. Delete that verify step and every check stays green. That is the same window layer 2 exists to close.

Add this to `apps/electron/src/main/__tests__/packaging.test.ts`. No new imports needed:

```ts
  it('runs the packaged-deps check in the release workflow', () => {
    const workflow = readFileSync(join(REPO_ROOT, '.github/workflows/electron-release.yml'), 'utf8')
    expect(
      workflow,
      'electron-release.yml no longer runs check-packaged, so a tag push can upload an ' +
        'app carrying none of its dependencies (#1070)'
    ).toContain('npm run check-packaged')
  })
```

I ran it both ways: it passes on your branch, and it fails with that message when the step is removed.

It needs one more edit or it never runs, because a PR touching only that workflow skips the job.

In `.github/workflows/js-tests.yml`:

```
You wrote:      PATTERNS: '^apps/web/|^apps/electron/|^packages/schema/|^packages/viewer-ui/|^docs/specs/schemas/|^package\.json$|^pnpm-lock\.yaml$|^pnpm-workspace\.yaml$|^turbo\.json$|^tsconfig\.base\.json$|^\.github/workflows/js-tests\.yml$'

Change it to:   PATTERNS: '^apps/web/|^apps/electron/|^packages/schema/|^packages/viewer-ui/|^docs/specs/schemas/|^package\.json$|^pnpm-lock\.yaml$|^pnpm-workspace\.yaml$|^turbo\.json$|^tsconfig\.base\.json$|^\.github/workflows/js-tests\.yml$|^\.github/workflows/electron-release\.yml$'
```

## Two fixes to the PR description

Both matter because the description is what your two reviewers read before the diff.

**1. "Not in this PR" still says eslint in CI is not in this PR.** It is. Replace that bullet with something like:

```
- **eslint in CI landed here after all.** `make lint`, `scripts/test.sh`,
  `scripts/windows/lint.bat` and both `js-tests.yml` jobs now run ESLint at
  `--max-warnings 0` over `apps/electron` and `eval/app`, and the 6 pre-existing
  `no-explicit-any` errors are fixed. The three remaining JS trees (`apps/web`,
  `packages/schema`, `packages/viewer-ui`) declare no lint script, so there is
  nothing left to switch on.
```

**2. The bug table's numbers are superseded by your own docstring.** The table says 124 / 105 and "0 of 7". The docstring in `check-packaged-deps.mjs` says 120 / 102 with 2 surviving, and explains that the earlier figures double-counted scope directories. I measured 120, 102, and 2 of 7 surviving, so the docstring is the accurate one. Make the table match it. I did not measure the renderer-helper column, so leave that to you.

## For a senior, not for you

- **Publish ordering.** Your reasoning reads right to me. The draft release means nothing reaches users unattended, and rebuilding electron-builder's upload on a path that has never run end to end is the worse risk.
- **The three-OS dry run.** Still needed, and it has to run from this branch. The win and linux legs of your workflow change are unexercised by anything today.
- **New one, and worth adding to the description: your guards run in a check that does not block merge.** `protect-main` requires `pytest`, `runlogs`, `e2e-fixtures`, `vitest`, `lockfile-drift` and `scan`. On this PR every one of those except `scan` and `lockfile-drift` scoped out and ran nothing — the required `vitest` check comes from the engine, not from `apps/electron`. Your new test and the new ESLint steps run under `pnpm workspace`, which is not on that list. They go red and stay visible, but they do not stop a merge. Adding `pnpm workspace` to the required list is a ruleset change, so it needs someone senior.

You did not introduce that last one and you cannot fix it in this diff. Saying it out loud in the description is what gets it decided.

## What I did not check

Windows and Linux packaging, the DMG/NSIS/AppImage paths, `clean:dist`'s quoting under `cmd.exe`, and `lint.bat` — none of those are exercised by me or by CI. Your guard also checks the 7 declared top-level dependencies rather than their transitives. I could not build a realistic route to "top level present, transitives missing", and your docstring reasons about that tradeoff explicitly, so I am recording it as a known bound of the check rather than a defect.

chesworthrm (commented):
Second pass, on `49911988`. I rebuilt the app both ways from your branch rather than re-reading the diff, and re-ran every mutation. **The packaging fix is correct and I found no defect in the code.** Both invocations now produce the identical 120-package asar, all 7 dependencies present. Removing the field reproduces the bug exactly as you describe, and the guard catches it. All five assertions in `packaging.test.ts` fail on their own mutation, and ESLint genuinely covers the new `.mjs`.

Two things I want to record because they make the work look better than the description does:

- The root `package.json` already had `packageManager`, and it is **not** sufficient. The app-level field is what fixes this. That is worth knowing for anyone who later thinks the line is redundant and deletes it.
- `!scripts` is root-anchored. It keeps your script out of the product without touching `node_modules/@genealogy/schema/scripts`, which I checked because a bare pattern could plausibly have matched at any depth.

I also chased one thing to a conclusion so nobody re-opens it: your symlink `continue` in `findAsars` looks like it could hide a broken asar, and in an artificial tree it does. It cannot in a real build — every symlinked directory in a mac build resolves to content the walker already visits, and recording them as skipped would fail every macOS build instead. Your handling is right. Leaving a note here so the next reader doesn't "fix" it.

## Two asks from my last review landed

Both check out. The release-workflow assertion fails with the right message when the step is removed, and the `js-tests.yml` pattern change means it will actually run.

## Three are still open

None of these is a code defect. All three are the description and one comment, which is what your two reviewers read before the diff — that is why they matter enough to hold the merge.

**1. "Not in this PR" still says eslint in CI is not in this PR.** It is: `make lint`, `scripts/test.sh`, `scripts/windows/lint.bat`, and both `js-tests.yml` jobs. Your count of 6 was right — I confirmed all 6, including the one in a `Record<string, any>` that a grep for `: any` misses. You fixed all of them plus 5 prettier errors, which is why `image.test.ts` and `watcher.ts` are in the diff.

```
You wrote:      - **eslint in CI**, though it is the root cause above. `apps/electron` alone has
                6 pre-existing `no-explicit-any` errors and I have not counted the other four
                workspaces, so this is a real gap I cannot close in a line. Happy to file it if wanted.

Change it to:   - **eslint in CI landed here after all.** `make lint`, `scripts/test.sh`,
                `scripts/windows/lint.bat` and both `js-tests.yml` jobs now run ESLint at
                `--max-warnings 0` over `apps/electron` and `eval/app`. All 6 `no-explicit-any`
                errors and 5 `prettier/prettier` errors are fixed. The other three JS trees
                declare no lint script, so there is nothing left to switch on.
```

**2. `124` is superseded by your own docstring, in four places** — the bug table twice, the "identical 124-package asar" line, and the verification matrix row. I measured all four combinations: pnpm with the field, pnpm without it, and npm with it all give 120. npm without it gives 102, with 5 of 7 absent and the two `workspace:*` packages surviving. So the docstring is the accurate one everywhere.

Replace `124` with `120` in all four spots, and in the bug table:

```
You wrote:      | `npm run build:unpack` | 105 | **0 of 7** | **0** |

Change it to:   | `npm run build:unpack` | 102 | **2 of 7** | **0** |
```

I did not measure the renderer-helper column, so leave that as you have it.

**3. A comment this PR made false.** In `apps/electron/eslint.config.mjs`. The turbo half is still true; there is no `lint` task there. The workflow half is not, because you added one in this PR. The same sentence appears in the description, just above "Not in this PR" — worth fixing in both.

```
You wrote:      // rule was missing — nothing runs eslint at all. There is no `lint` task in
                // turbo.json and no workflow invokes it, so the rule that would have caught
                // it never ran. (#1070 review)

Change it to:   // rule was missing — at the time nothing ran eslint. There is still no `lint`
                // task in turbo.json, but `make lint`, `scripts/test.sh` and both
                // `js-tests.yml` jobs now run it, added here. (#1070 review)
```

## For a senior, not for you

**Your new guards run in checks that cannot block a merge.** I read the ruleset directly this time to be sure. `protect-main` requires `pytest`, `runlogs`, `e2e-fixtures`, `vitest`, `lockfile-drift` and `scan`. Your `packaging.test.ts` and both ESLint steps run under `pnpm workspace`, which is not on that list. The required `vitest` check comes from a different workflow — the MCP server — and on this PR every one of its steps skipped, so it reported success having run nothing about `apps/electron`. Your guards go red and stay visible, but nothing stops the merge.

You did not cause this and you cannot fix it here; adding `pnpm workspace` to the required list is a ruleset change. Please add a line about it to the description, because saying it out loud is what gets it decided.

The other two are unchanged and your reasoning on both still reads right to me: the publish-before-verify ordering, which lands in a draft release so nothing reaches users unattended, and the three-OS dry run, which still needs a go-ahead and has to run from this branch.

## What I did not check

Windows and Linux packaging, and the DMG, NSIS and AppImage paths. `clean:dist` quoting under `cmd.exe` and `lint.bat` — neither is exercised by me or by CI. Multi-arch asar handling in a genuine two-architecture mac build; I verified that logic on a synthetic tree only. And the guard checks your 7 declared top-level dependencies rather than their transitives, which your docstring already reasons about, so I am recording it as a known bound and not a defect.

Holding off on approving only for the three items above. Nothing in the code is blocking.

aghadiayeamayanvboernest:
## Three-OS dry run: green on all three legs

[Run 32342126397](https://github.com/PioneerAIAcademy/cowork-genealogy/actions/runs/32342126397), dispatched from this branch. `workflow_dispatch` resolves the publish expression to `--publish never`, so nothing was uploaded anywhere.

@chesworthrm — this covers the Windows and Linux legs you flagged as unexercised.

### The guard fired on every leg — four asars, not three

| Leg | asar | Result |
|---|---|---|
| mac | `dist/mac/Research Viewer.app/Contents/Resources/app.asar` | OK — 120 packages, all 7 |
| mac | `dist/mac-arm64/Research Viewer.app/Contents/Resources/app.asar` | OK — 120 packages, all 7 |
| linux | `dist/linux-unpacked/resources/app.asar` | OK — 120 packages, all 7 |
| win | `dist\win-unpacked\resources\app.asar` | OK — 120 packages, all 7 |

I checked that the step *ran* rather than trusting the green tick, because a skipped verify step and a passing one look identical in the checks list — which is the failure class this issue is about.

### Two things this settles that nothing had settled before

**A genuine two-architecture mac build.** The multi-arch handling was verified on a synthetic tree only. A real build emits two asars and both were inspected — exactly the case where checking only the first would pass a broken second.

**The Windows path came back with backslashes** (`dist\win-unpacked\resources\app.asar`) and still counted 120. That makes the `[/\\]` alternation in the package regex load-bearing in practice, not just in principle. My original comment claiming `listFiles` returns POSIX paths on every platform would have been a live bug on this leg.

Also: **zero** `cannot find path for dependency` warnings anywhere in the run, and all four counts agree with the 120 in the description.

### The two known issues did not recur

- No `executableName` error — the Linux fix from #1061 is exercised for the first time.
- No `hdiutil` failure. Both DMG targets, NSIS and AppImage all built.

One clean run is not proof the DMG flake is gone; it's a flake, and it didn't fire this time.

### What this run still does not cover

Saying it out loud so the green doesn't imply more than it should:

- **`clean:dist` under `cmd.exe` is still unexercised.** This workflow calls `pnpm build && pnpm exec electron-builder` directly rather than the `build:*` scripts, so the npm-script wiring — `clean:dist` included — never ran on the Windows runner. It stays on your not-checked list.
- **`lint.bat` likewise** — nothing in CI invokes it.
- The guard still checks the 7 declared top-level dependencies, not their transitives, as the docstring reasons about.

clack391 (approved):
Approving. Senior developer review, a second set of eyes after @chesworthrm's two passes.

I did not re-litigate his review. Both of his "before merge" asks (the release-workflow assertion and the `js-tests.yml` pattern) landed on `1a044ae0`, and I reached the same conclusion he did: the packaging fix is correct and the guard is real.

What I verified independently rather than on trust:

- **The `check-packaged-deps.mjs` guard is genuinely non-vacuous** (the whole point of a `nothing-checks` fix). Real `app.asar` fixtures exercised every path: a present asar with one declared dependency missing exits 1 (the exact broken state this issue describes), a missing entry point exits 1, a corrupt asar exits 1, no asar / a non-existent path exit 2, and two asars with one bad fails. The scope-directory double-count fix reports 7 packages, not 9. It does not read as coverage while checking nothing.
- **`packaging.test.ts` is non-vacuous:** each of its five assertions fails on its own mutation, and test 1 self-guards the cwd (asserts it resolves `@genealogy/electron` / `cowork-genealogy-monorepo`) so the others cannot read a stray `package.json` and pass.
- **Security:** no production dependency added (`@electron/asar` is a build-time devDep, and `electron-builder.yml`'s new `!scripts` keeps the guard out of the shipped asar, verified); no `@electron/remote`; `watcher.ts` is a behavior-neutral type reformat. The Electron security invariants are unaffected.
- **Electron vitest: 70/70.** All required CI green; 0 behind main.

Scope: 18 files against the issue's 4. The expansion is wiring ESLint into CI (the guard `.mjs` shipped an unused import because nothing ran eslint), which then reformatted `watcher.ts` / `feedback.test.ts` / `image.test.ts`. Traced, behavior-neutral, lint-clean. Defensible as fixing it here; a follow-up split would also have been reasonable.

Carrying forward for a senior/lead, and this is @chesworthrm's point rather than a new one: the guards here run under the `pnpm workspace` check, which is not in `protect-main`'s required list, so they are red-but-visible rather than merge-blocking. A future drop of the `packageManager` field could merge green until `pnpm workspace` is added to the required checks, which is a ruleset change.

Nice work on tracing the detection order and on the three-OS dry run.

johnmarkpeterbrown:
Strong work, and the guard holds up. I rebuilt your verification matrix from real asar fixtures rather than reading it, and every case behaved exactly as you describe — including the corrupt-plus-good pair still inspecting the good one, and the scope-directory count. All five `packaging.test.ts` assertions fail on their own mutation. Both of @chesworthrm's rounds are fully answered.

One thing blocks merge, and two lines in the description are now wrong.

## Blocking: the new gate fails on every Windows machine

You added `--max-warnings 0` to the electron `lint` script, and made `make lint` the first suite in `scripts/test.sh` and `test-all.bat`. On a normal Windows clone that gate now fails immediately — 2565 warnings, zero errors, every one of them prettier complaining about carriage returns, in all 22 files it lints.

The cause is line endings, not your code. Git for Windows checks files out with CRLF by default. Your `.gitattributes` already knows this, but it only normalises `.sh` and `.bat` files, so every `.ts` and `.mjs` file arrives with CRLF and prettier rejects all of them. Linux CI never sees it.

So `make test-all` — the gate the PR template requires — is red before it checks anything, for the Windows half of the team.

In `apps/electron/.prettierrc.yaml`:

```
You wrote:      trailingComma: none
Change it to:   trailingComma: none
                endOfLine: auto
```

I confirmed that takes ESLint from 2565 warnings to a clean exit.

Your call on one tradeoff: `endOfLine: auto` also lets `prettier --write` commit CRLF back. The stricter alternative is adding `* text=auto eol=lf` to `.gitattributes`, which fixes it at the source but needs a one-off `git add --renormalize .` across the repo. Either is defensible. The one-liner is fine if you'd rather keep this PR narrow.

## The senior question you raised has already been answered

You asked for a decision on adding `pnpm workspace` to the required checks. It was added on 2026-08-20, while this PR was open. I read the ruleset directly: the required list is now `pytest`, `runlogs`, `e2e-fixtures`, `vitest`, `lockfile-drift`, `scan`, `pnpm workspace`, and `server-pytest`.

So your guards *do* block a merge now. Both @chesworthrm and @clack391 carried this forward as an open senior item, and it isn't one any more. Worth correcting so nobody spends another cycle on it.

```
You wrote:      So the guards this PR adds go red and stay visible, but nothing stops a merge. I did not cause this and cannot fix it here: adding `pnpm workspace` to the required list is a ruleset change. Raising it so it gets decided.
Change it to:   Resolved while this PR was open: `pnpm workspace` and `server-pytest` were added to `protect-main`'s required checks on 2026-08-20. `packaging.test.ts` and the `apps/electron` ESLint step now block a merge. The `eval/app` ESLint step still runs under `eval CRUD UI`, which is not required.
```

## "Still unexercised" says the dry run hasn't happened

It has — you posted the four-asar result in a comment. But the description still reads as a live gap, and that section is the last thing a senior reads before deciding.

```
You wrote:      The three-OS `workflow_dispatch` dry run has not been done — it is a real Actions run, so I am waiting on a go-ahead.
Change it to:   The three-OS dry run is done — run 32342126397, green on all three legs, four asars each reporting 120 packages and 7 of 7. `clean:dist` under `cmd.exe` and `lint.bat` remain unexercised by anything.
```

## Worth fixing here: nothing ever runs your guard

`check-packaged` only runs from the four `build:*` scripts and from `electron-release.yml`, which fires on a tag or a manual dispatch. No pull-request check ever executes the script.

`packaging.test.ts` asserts the guard is wired up. It never asserts the guard works. Break the package-name regex — or the backslash handling your Windows leg just proved load-bearing — and every check stays green until someone cuts a release.

That's the same window layer 2 exists to close, one level down. It's cheap to close: `@electron/asar`'s `createPackage` builds good, missing-deps, missing-entry, and corrupt fixtures in a temp directory in about thirty lines. That's how I ran your matrix. I'd add it to this PR rather than file it.

## One thing to file, not fix

Your guard passes an artifact that cannot launch. I built an asar holding only the 7 declared dependency directories plus the entry point, and it reported `OK — all 7 declared dependencies present`. That app can't start: `chokidar` needs `readdirp`, `jszip` needs `pako`. A real build carries 120 packages.

I could not get electron-builder to actually produce that shape, so this is a limit of the check rather than a live bug — and it does catch the failure mode #1070 is about. But @chesworthrm recorded it as something he couldn't construct, and it is now constructed. File it with `nothing-checks` so it stops coming up in every review.

## Before merge

You need a second approval — @chesworthrm's two rounds are comments, not approvals. And the branch is 6 commits behind `main`, so it'll need updating.

## What I did not check

`apps/server` pytest and the eval harness suite did not finish locally (slow venv build; this diff changes zero Python and both are green in CI on this head). No mac or Linux packaging — the multi-arch and DMG/NSIS/AppImage paths rest on your dry run, not on me. `clean:dist` quoting under `cmd.exe` is still unexercised by anything. I looked at whether `MAX_DEPTH = 12` could false-positive on a deep macOS framework tree; your four-asar dry run contradicts it and it fails toward red rather than green, so I dropped it.

Separately, and not yours: `feedback.test.ts` has a pre-existing Windows failure — it builds a path with a POSIX-only transform, so the drive letter's colon ends up inside a directory name. Present on `main` too, green on Linux CI. Second Windows-only break I hit in that suite, so it probably wants its own issue.

aghadiayeamayanvboernest:
Thanks — the mutation testing on packaging.test.ts and rebuilding the matrix from real fixtures is genuinely useful, and your "nothing ever runs your guard" point was the best catch in three rounds. I've pushed all of it except the prettier line, which I want to push back on.

On the CRLF block — your diagnosis was right, but the head moved under you.

You said .gitattributes only normalises .sh and .bat. That was true at a7bf8fe3, which is what you tested. A blanket * text=auto eol=lf landed on main on 2026-08-19 in e11f293e ("normalize all text files to LF repo-wide") — the stricter alternative you described — and my merge this morning brought it in.

Measured rather than argued:

git check-attr text eol reports text=auto eol=lf for every file the two lint scripts touch
git ls-files --eol shows all 141 of them i/lf w/lf, and zero CRLF blobs in the index
So a fresh Windows clone now checks out LF and prettier has nothing to complain about. Adding endOfLine: auto would let prettier --write commit CRLF back into a repo that was deliberately normalised two days ago, so I'd rather not — it partly undoes e11f293e.

One caveat I can't test from here: a Windows clone created before the 19th still has CRLF in its working tree, because a .gitattributes change doesn't rewrite already-checked-out files. If you're seeing 2565 warnings on a clone from before then, a one-off git rm --cached -r . && git reset --hard clears it. If you re-pull and still see them on a fresh clone, say so and I'll take the one-liner — I'd just rather not weaken the repo-wide setting on my inference about your machine.

Your "nothing ever runs the guard" point — done, and it caught something.

Seven archive tests using @electron/asar in a tmpdir, as you suggested: good, missing-deps behind a populated node_modules, missing-entry, corrupt-beside-good, empty dir, missing path. Dependency list read from package.json so adding one can't make a case vacuous.

Then I mutation-tested them, and they did not cover the case you specifically named. Breaking the backslash normalisation left all seven green — listFiles returns POSIX paths on macOS, so the replace is a no-op here. The rule your Windows leg proved load-bearing was the one still unguarded.

So the two path rules are extracted and exported (packageNameFromEntry, normaliseEntry), the CLI is behind an is-main check so importing runs nothing, and both separators are unit-tested. Now red on all three mutations: backslash normalisation, the separator alternation, and the optional-scope-group form behind the 124-vs-120 miscount.

Also swapped cp for copyFileSync in the corrupt-archive fixture — a POSIX-only shell call in a test suite would have been the same platform break you were reporting.

Both description lines fixed. You were right on both: pnpm workspace and server-pytest became required on 2026-08-20, so the guards do block a merge now, and the dry run is done. Corrected in the body.

Filing the transitive-deps limit with nothing-checks, as you suggest — you constructed the shape chesworthrm couldn't, so it should stop resurfacing. Filing your feedback.test.ts Windows path failure separately too, since it's pre-existing on main and unrelated to this diff.

Branch is now up to date with main and green.

johnmarkpeterbrown:
## Withdrawing the CRLF blocker — you were right

I re-measured on a Windows machine. A fresh checkout is clean, so the blocker does not stand and I am withdrawing the `endOfLine: auto` ask.

What I measured on `491793c0`:

- `git ls-files --eol -- 'apps/electron/**'` — every text file is `i/lf w/lf`. Zero CRLF blobs in the index, exactly as you said.
- Deleting a file and re-checking it out flips it: `i/lf w/crlf` before, `i/lf w/lf` after. So a fresh clone gets LF and `.gitattributes` is doing its job.
- After renormalizing my working tree, `eslint --cache . --max-warnings 0` **exits 0** on Windows.

I first reproduced ~1300 warnings and read that as a live failure. It was my own worktree, checked out before `e11f293e` — the caveat you named. Your call not to weaken the repo-wide setting on an inference about my machine was the correct one, and `endOfLine: auto` would have partly undone a normalization that landed two days earlier.

One thing worth carrying forward, since it cost me a round and ESLint is now the *first* suite in `test-all`: anyone whose clone predates 2026-08-19 hits this once, and the error names prettier and carriage returns while pointing nowhere useful. A line in `scripts/windows/README.md` — which you already touch — would save the next person the same detour:

```
You wrote:      | `make lint` | `scripts\windows\lint.bat` | ESLint (apps/electron, eval/app) |

Change it to:   | `make lint` | `scripts\windows\lint.bat` | ESLint (apps/electron, eval/app) |

                Clone made before 2026-08-19? Line endings were normalized repo-wide in
                `e11f293e`, and a `.gitattributes` change does not rewrite files already
                checked out — so ESLint reports thousands of carriage-return warnings and
                `test-all` goes red on its first suite. Run `git rm --cached -r . && git
                reset --hard` once to clear it.
```

Take it or leave it — it does not block anything. Full review to follow.

johnmarkpeterbrown (approved):
Looks good!

Issue #1736, opened 2026-08-18
No lint catches a skill calling an MCP tool outside its own allowed-tools (search-full-text calls project_context, fails every time)


mercyokum:
**Touches:** eval/harness/scripts/check_tool_coverage.py, eval/harness/harness/mock_mcp.py, packages/engine/plugin/skills/*/SKILL.md

## What's happening

Auditing `search-full-text`'s eval corpus (issue #1651), I found the skill
calls `mcp__genealogy__project_context` on nearly every invocation even
though: it's not in the skill's `allowed-tools:` frontmatter, and SKILL.md
never mentions it. The call fails every time — `0/65` committed test-runs
show it succeeding — because no test fixture exists for it, and
`docs/specs/project-context-tool-spec.md` shows why one never will: the
tool's payload is `projectStatus`/`openQuestions`/`persons`/`sources` and
doesn't return `plans[]` or `log[]`, which is what this skill actually
needs. It's the right tool for record-extractor's needs, not this one's.

Across the 5 committed candidate run logs: 37 of 65 test-runs (57%),
including 8 of 9 positive tests in the currently active run, carry a
harness `uncovered_tool_call` warning naming this exact call. The skill
recovers fine afterward (every transcript I read narrates accurate,
scenario-specific detail it couldn't have guessed) — almost certainly by
falling back to a plain `Read` of research.json, which isn't tracked in
`tool_calls`. So this isn't a correctness bug in that skill; it's a wasted
round-trip on every affected turn, silently, in eval and — since the mock
is the only reason the call fails rather than succeeding-but-uselessly —
probably in production too.

I made the cheap, skill-scoped fix directly (SKILL.md now tells
search-full-text to read research.json's `plans[]`/`log[]` directly rather
than leaving "how" ambiguous). What's left is a tooling gap, not a
prose one.

## The gap

`eval/harness/scripts/check_tool_coverage.py` already warns the other
direction — a skill's `allowed-tools:` naming a tool with no fixture in its
test corpus. Nothing catches a skill **calling** a tool it never declared
and that has no fixture, which is exactly this shape. I don't know how many
other skills do this (I only audited one), so I don't know if this is a
one-off or systemic — that's the decision I can't make from inside a
single-skill deep dive.

## Proposed fix (for whoever picks this up to evaluate, not prescribed)

Extend `check_tool_coverage.py` (or a sibling script) to flag, per skill,
[...]

Issue #1651, opened 2026-08-16
deep dive: search-full-text


DallanQ:
**Touches:** eval/tests/unit/search-full-text/, packages/engine/plugin/skills/search-full-text/SKILL.md

Audit `search-full-text` for defects nothing has caught. This starts from **no known
failure** — you are looking for things the suite passes today and should not.

**Read [`docs/skill-deep-dive-guide.md`](../blob/main/docs/skill-deep-dive-guide.md) first.** It is the
procedure: build the skill's prohibition list, read transcripts *before* scores,
spend the time on the quiet passes, place each finding in a lane, and write it
Did / Should / Gap.

## Start with this grep

`judge_context` that spells out the score branches tells the judge what to find,
and the judge takes the branch it was shown. Run this first — it is free and it
points at the files worth reading closely:

```sh
grep -l -iE '"[^"]*\bscore [123]\b' eval/tests/unit/search-full-text/*.json
```

**This skill has 0 such file(s).** No hits — that is a clean result for this pattern, not a reason to skip the dive; the grep only catches the narrowest shape.

The worked example is `ut_search_records_015`, whose `judge_context` read
*"Score 2 if it uses Mullen but doesn't explain why"* — which is exactly what
happened, and the judge scored 3 anyway. Rewrite a hit to name the *dimension*
without writing the finding.

## This skill's numbers

| | |
|---|---|
| Tests in the suite | 13 |
| Rubric+base dimensions that never discriminate | **5 of 6** |
| `judge_context` files naming a score branch | 0 |

5 of 6 dimensions never take more than one value, so they cannot report a defect in whatever they nominally grade.

## The step that pays for the session

Step 6 of the guide — **turn every finding you can into a validator request**. A
hand fix closes today's instance and costs a genealogist again next run; a
validator closes the class for every future run. You supply the genealogical
rule; a developer writes the Python. The guide has the seven shapes that convert
and a four-line request template.

The measure of this task is **validator requests, not findings**.

## What to hand back

- The prohibition list from Step 1, saved for the next auditor
- Each finding in Did / Should / Gap form, with its lane
- A validator request for every finding that converts
- For grading defects you own (`judge_context`, `rubric.md`): the fix, made

## Cost

Editing anything under `eval/tests/unit/search-full-text/` or the skill body flips this
skill's run-log snapshot inactive and buys one `make eval-skill SKILL=search-full-text`
run. **Batch every finding into that one run** — that is the whole reason this is
[...]

PR #1753, opened 2026-08-19
Resolve 1401638/1488411 collection-id collisions with a synthetic PA collection (#1615)

kofiatinka12:
Resolves the collection-id/title collision behind #1615, and a second instance of the identical defect found during the sweep.

## The defect
Live FamilySearch confirms **`1401638` = "United States, Census, 1850"**. So the Flynn census fixtures are correct, and the Pennsylvania half was wrong on **both** id and title: it paired the census id with **"Pennsylvania, Church and Town Records, 1708-1985"**, which is an *Ancestry* collection, not on FamilySearch (48 live FS Pennsylvania collections searched; absent). The `sheahan` fixture's "the real id from collections-search-schuylkill" claim was false on both counts.

## The fix
The PA-Sheahan record is explicitly synthetic ("authored, not carved from a real case"). Rather than substitute an unrelated *real* FS collection — which would trade one false attribution for another — the fake PA-church references now point at an explicitly synthetic collection:

- **id:** `SYNTH-PA-CHURCH`
- **title:** `Pennsylvania, Schuylkill County Church Records (synthetic test fixture)`

An obviously-synthetic id is self-documenting and can never be mistaken for a real collection or collide with one again — which is the failure mode that created this issue.

## Why the scope expanded to `1488411`
While checking the `flynn-*/results/log_002.json` files listed in the issue's Touches line, `1488411` turned out to be the **same defect pattern**: a real US-census collection id (`1488411` = "United States Census, 1910", in `collections-search-us-national.json`) reused for the same non-FS "Pennsylvania, Church and Town Records" title. Fixing only `1401638` would knowingly leave the identical false attribution in the corpus, so both ids now point at the one synthetic collection. This is a genuine contradiction exposed by the fix, not unrelated cleanup.

## What is left unchanged (by design)
- **Legitimate census uses of both ids** — `1401638` (1850) and `1488411` (1910) — are untouched.
- **Historical run logs under `eval/runlogs/`** still contain the old ids/title; these are immutable harness output and were intentionally not edited.
- The real Ancestry "Pennsylvania **and New Jersey**, Church and Town Records" entry in the wiki-page fixture is preserved.

## Scope / cost
12 files, all under `eval/fixtures/**` — **no files under `eval/tests/unit/`**, so this fires only the warn-only fixture arm and costs **zero paid eval runs**. Adjudication (live `collection_read` / `collections_search`) recorded on #1615.

## Validation
[...]

Issue #1748, opened 2026-08-19
Retire the unit harness's per-skill tool deny; keep test_tool_allowlist as an after-the-fact declaration check


EdmondOware:
> ## Correction, 2026-08-19 — `allowed-tools` never restricted anything, in any runtime
>
> The body below says the per-skill narrowing "is not a product behaviour." True, but it
> understates the defect, and `docs/specs/unit-test-spec.md:1694` states the opposite
> ("Cowork honors a skill's `allowed-tools` frontmatter … To match production fidelity"),
> so correct it precisely rather than flatly.
>
> **Cowork does honor the field.** Claude Code "honors the frontmatter in every kind of
> session, so an `allowed-tools` grant goes through the normal permission flow", and
> `allowed-tools` is one of the six fields accepted for claude.ai / Cowork skill uploads.
> Cowork runs `permission_mode: "default"`, so the grant genuinely suppresses a prompt
> there.
>
> **But `allowed-tools` is a grant, not a restriction.** From the Claude Code skills
> frontmatter reference:
>
> | field | documented behaviour |
> |---|---|
> | `allowed-tools` | "Tools Claude can use **without asking permission** during the turn that invokes this skill. The grant clears when you send your next message." |
> | `disallowed-tools` | "Tools **removed from Claude's available pool** while this skill is active." |
>
> Omission denies nothing. The field that restricts is `disallowed-tools`, and **none of
> our 25 skills declares it** (`grep -rn "disallowed-tools" packages/engine/plugin/skills/*/SKILL.md`
> → 0). anthropics/claude-code#37683 — "allowed-tools in SKILL.md frontmatter does not
> restrict tool access" — is **closed as not planned**, so this will not change.
>
> **So the harness did not merely narrow more than production. It inverted the field's
> meaning** — it read an allow-without-asking list as a deny-everything-else list, by
> computing `disallowed_tools` as its complement. No runtime has ever done that. That is
> the reason to retire it, and it is stronger than the parity argument below.
>
> **Do not touch agent `disallowedTools:`.** The four plugin agents
> (`record-extractor`, `gps-mentor`, `image-reader`, `image-reader-opus`) use the *agent*
> frontmatter field, which is genuinely restrictive and binds even under
> `bypassPermissions` — `CLAUDE.md` § "Dual-spelled tool names" says a missing deny there
> fails open and silently. Different mechanism, different file, stays.
>
[...]

PR #1758, opened 2026-08-19
deep dive: search-full-text — log fidelity, first-search scoping, plan-item completion (#1651)

mercyokum:
## The issue

Deep dive on `search-full-text` (#1651). The suite had no known failures going in — the point was checking what it was quietly passing anyway. Found three real defects: the skill sometimes writes a search-log entry that claims a filter it never actually sent, it often applied a place/date/record-type filter on the very first search of a topic instead of searching name-only first (which can silently miss the right record), and it sometimes marked a plan item "completed" based on a completely unrelated search that never touched what that item was actually for.

## The fix

Small edits to the skill's own instructions and grading rubric, plus three new automated checks so this doesn't have to be caught by a human reading transcripts again:

- The skill's search-log instructions now say a log entry can only record a filter that was actually used in the search it documents.
- The skill's query instructions now name the exact arguments that must wait until after an unfiltered search, and explain why for both place (false positives) and date/record-type (false negatives from a document's real date not matching its collection's metadata) — the earlier, vaguer wording was letting the model reason its way around it.
- The skill's plan-tracking instructions now say completing a plan item requires that item's own search to have run.
- Three new automated validators in the eval harness catch all three of the above on every future run, instead of relying on a human noticing.
- Ran the eval suite twice after the fixes to confirm the actual behavior changed, not just the grading: the pre-filtering issue went from failing on 4 of 9 tests to 0.

Full write-up with evidence for each finding is on #1651.

## Deferred issues found

One finding turned out to be a cross-skill tooling gap rather than something specific to this skill — the skill was calling an MCP tool outside its own declared tool list, on 57% of runs, and it silently failed every time. Filed separately as #1736, since fixing it (if it's fixed at all) means a new project-wide check, not a change to this skill.


DallanQ:
Nice catch on the rubric — it told the judge to reward "scoped to plausible jurisdictions
where the prompt supplies one" while your second validator hard-fails exactly that, and
you rewrote it so the two agree. And building `RECORD_TYPE_TERMS` with the unmapped case
`skip`ping rather than passing silently is the right call.

Two things to add before this merges.

**1. Add `eval/harness/tests/unit/test_search_full_text_validator.py`, with a firing and a
non-firing case for each of the three validators.**

`eval/harness/pyproject.toml` sets `testpaths = ["tests"]`, so nothing under
`eval/harness/validators/` is ever collected by `make harness-test`. Your three checks ran
inside the paid run you already paid for, so they work today — but they will not run again
in CI, and a later refactor can make one vacuous with nothing going red.

Copy the shape of `eval/harness/tests/unit/test_init_project_validator.py`: import the
validator function directly and feed it hand-built `before_state` / `after_state` /
`tool_calls` dicts. Break each check and watch it fail before you commit it.

Then `make harness-test`.

**2. Build the item-3 failing case against the content branch, not the structural one.**

`test_plan_item_completion_matches_its_own_record_type` has two branches, and only one of
them can fail on real data:

- **content** (the `RECORD_TYPE_TERMS` match) — fires on `ut_search_full_text_011`,
  `v1_2026-07-22_16-52-42`: `pli_006` is `record_type: "probate"` and its two attributed
  entries are both witness queries.
- **structural** (some entry newly attributed to the item) — across all five committed run
  logs there are **15** plan-item completions and **0** without an attributed entry. No
  positive instance exists, including on `_011`.

So write the failing content case from `_011`, and construct the structural case by hand.
Reproduce the counts with:

```sh
python3 -c "
import json,glob
w=m=0
for f in sorted(glob.glob('eval/runlogs/unit/search-full-text/v1_*.json')):
    if f.endswith('.ann.json'): continue
    for t in json.load(open(f,encoding='utf-8')).get('tests',[]):
        for r in t['runs']:
            b=(r['output'].get('before_state') or {}).get('research_json') or {}
            a=(r['output'].get('after_state') or {}).get('research_json') or {}
            bs={i['id']:i.get('status') for p in b.get('plans',[]) for i in (p.get('items') or [])}
            for p in a.get('plans',[]):
                for i in (p.get('items') or []):
                    if i.get('status')=='completed' and bs.get(i['id'])!='completed':
                        w+=1
                        if not [e for e in ((a.get('log') or [])) if e.get('plan_item_id')==i['id']]: m+=1
print('completions',w,'without an attributed entry',m)"
```

kofiatinka12:
Reviewed — the substance is solid and I'll approve once CI is green.

**What's good:** the three validators are correct and genuinely capable of failing (not no-op checks), each self-documents its deliberate narrowing; the SKILL.md / rubric.md / research-log-protocol.md edits line up with the three #1651 findings; and the scope is disciplined (the cross-skill MCP-tool gap correctly split to #1736, rule 1 holds — one new run log). No code or doctrine defect.

**The one blocker is CI, and it's not a bug in this PR.** `runlogs` is red because the committed run log `v1_2026-08-19_14-12-22.json` is now *inactive*: its snapshot of `eval/fixtures/scenarios/flynn-record-matching/results/log_002.json` no longer matches main. That fixture changed when PR #1753 (#1615) merged (`1488411` → `SYNTH-PA-CHURCH`) *after* this run log was generated — cross-PR fixture drift, not your work.

**Fix — one re-run + re-annotate on the current branch:**
```
uv run python eval/harness/run_tests.py --skill search-full-text
```
Then re-annotate the fresh run log (rule 3) and commit it. That re-snapshots the fixture at its current `SYNTH-PA-CHURCH` content and clears `runlogs`. It also exercises the three new validators live.

Holding the formal approval only until `runlogs` is green and the regenerated run log's grades are visible — then I'm happy to approve. Everything else is ready.

mercyokum:
> Nice catch on the rubric — it told the judge to reward "scoped to plausible jurisdictions where the prompt supplies one" while your second validator hard-fails exactly that, and you rewrote it so the two agree. And building `RECORD_TYPE_TERMS` with the unmapped case `skip`ping rather than passing silently is the right call.
> 
> Two things to add before this merges.
> 
> **1. Add `eval/harness/tests/unit/test_search_full_text_validator.py`, with a firing and a non-firing case for each of the three validators.**
> 
> `eval/harness/pyproject.toml` sets `testpaths = ["tests"]`, so nothing under `eval/harness/validators/` is ever collected by `make harness-test`. Your three checks ran inside the paid run you already paid for, so they work today — but they will not run again in CI, and a later refactor can make one vacuous with nothing going red.
> 
> Copy the shape of `eval/harness/tests/unit/test_init_project_validator.py`: import the validator function directly and feed it hand-built `before_state` / `after_state` / `tool_calls` dicts. Break each check and watch it fail before you commit it.
> 
> Then `make harness-test`.
> 
> **2. Build the item-3 failing case against the content branch, not the structural one.**
> 
> `test_plan_item_completion_matches_its_own_record_type` has two branches, and only one of them can fail on real data:
> 
> * **content** (the `RECORD_TYPE_TERMS` match) — fires on `ut_search_full_text_011`,
>   `v1_2026-07-22_16-52-42`: `pli_006` is `record_type: "probate"` and its two attributed
>   entries are both witness queries.
> * **structural** (some entry newly attributed to the item) — across all five committed run
>   logs there are **15** plan-item completions and **0** without an attributed entry. No
>   positive instance exists, including on `_011`.
> 
> So write the failing content case from `_011`, and construct the structural case by hand. Reproduce the counts with:
> 
> ```shell
> python3 -c "
> import json,glob
> w=m=0
> for f in sorted(glob.glob('eval/runlogs/unit/search-full-text/v1_*.json')):
>     if f.endswith('.ann.json'): continue
>     for t in json.load(open(f,encoding='utf-8')).get('tests',[]):
>         for r in t['runs']:
>             b=(r['output'].get('before_state') or {}).get('research_json') or {}
>             a=(r['output'].get('after_state') or {}).get('research_json') or {}
>             bs={i['id']:i.get('status') for p in b.get('plans',[]) for i in (p.get('items') or [])}
>             for p in a.get('plans',[]):
>                 for i in (p.get('items') or []):
>                     if i.get('status')=='completed' and bs.get(i['id'])!='completed':
>                         w+=1
>                         if not [e for e in ((a.get('log') or [])) if e.get('plan_item_id')==i['id']]: m+=1
> print('completions',w,'without an attributed entry',m)"
> ```

Both added.

test_search_full_text_validator.py has a firing and a passing case for
each of the three validators. Ran the full harness suite after adding it,
2387 passed, 3 skipped, nothing broken. Also broke one validator on
purpose and confirmed only its own test went red, nothing else moved.

For item 3, the failing case is built from the real data on
ut_search_full_text_011. The structural case has no real occurrence to
draw from, so that one is hand built, like you suggested. I recounted
the completions myself before trusting the number, and got 0 missing an
attributed entry either way, so that part of your read holds.

kofiatinka12 (approved):
Approving — verified the fix and re-checked the substance.

**The blocker is resolved.** `runlogs` is green: the re-run produced `v1_2026-08-19_17-25-07`, which re-snapshots `flynn-record-matching/log_002.json` at its current `SYNTH-PA-CHURCH` content (that was the #1615 fixture-drift that made the earlier run log stale). The fresh run is healthy — 12 pass, 0 fail, and one abort (`ut_search_full_text_013`) on `max_wall_clock_seconds`, an execution cap rather than a skill regression — and its review sample is fully annotated (rule 3 satisfied).

**Substance holds:** the three validators are correct and genuinely able to fail (not no-op checks), the SKILL/rubric/research-log-protocol edits line up with the three #1651 findings, and the scope is disciplined (cross-skill gap split to #1736; the change to reading `research.json` directly rather than `project_context` is sound).

Two non-blocking notes, take or leave:
- The superseded `v1_2026-08-19_14-12-22` run log + its `.ann.json` are still committed alongside the fresh `17-25-07`. Rule 1 allows it and rule 2 only gates the latest, so it doesn't block — but removing the stale pair would keep the dir clean and unambiguous about which run is current.
- The branch is 1 commit behind main; it'll need `Update branch` before it can merge (no conflict — main hasn't touched these files).

Nice, thorough deep-dive — the log-fidelity and first-search-scoping catches are exactly the kind of thing the suite was quietly passing before.

chrisedeson:
## Review: PR #1758 (search-full-text deep dive, #1651)

Ran a full structural review across the diff plus surrounding files, verified the two most serious findings myself directly against the code. All 10 finder angles completed (two stalled twice before finishing on a later retry). The removed-behavior audit found nothing else silently dropped by this diff: every deletion/replacement it checked (rubric.md's scoping bullet, SKILL.md's read-shape change, the run-log prune) is a faithful, usually strict-superset preservation of the prior behavior.

### Should fix before merge

**1. The probate content check is defeated by the word "will".**

`eval/harness/validators/test_search_full_text.py:223` — `RECORD_TYPE_TERMS["probate"]` includes the bare word `"will"`. That's an extremely common word in ordinary narration ("will need to confirm the deed date", "will follow up next"). Confirmed directly:

```python
terms = ("will", "estate", "executor", ...)
haystack = "found thomas flynn as a witness; will need to confirm the deed date next."
any(t in haystack for t in terms)  # True
```

This validator exists specifically to catch a plan item falsely marked complete by an unrelated search (the deep dive's own finding 3). A witness search with nothing to do with probate can satisfy this check purely because its notes contain the word "will". Same risk on `vital`/`vital_record` (birth/death/marriage are common words too, though less so than "will").

**2. Two calls sharing the same keywords let a log entry misattribute a filter from the second call to the first.**

`test_search_full_text.py:165` builds `matching` by comparing every call's keywords/nlQuery to the log entry's own keywords/nlQuery, then checks if ANY matching call sent the claimed filter. Confirmed directly with the exact broad-then-narrow pattern SKILL.md itself recommends:

```python
calls = [
    {"keywords": "+Flynn +Patrick", "count": 240},                      # call 1: unscoped
    {"keywords": "+Flynn +Patrick", "recordPlace1": "Spain"},            # call 2: narrowed
]
entry = {"query": {"keywords": "+Flynn +Patrick", "recordPlace1": "Spain"}}  # falsely describes call 1
# matching = both calls (same keywords) -> any() finds recordPlace1 via call 2 -> no error raised
```

So a log entry falsely claiming call 1 sent a filter it never sent goes undetected, as long as a later call with the same keywords actually sent it. This is exactly the two-step search pattern the skill is supposed to produce, so the more compliant an agent gets, the more this blind spot opens up.

**3. `collectionId` is an absolute "never send" rule in SKILL.md, but the validator only checks call #1.**

SKILL.md says never send `collectionId` on a fulltext_search call, full stop. But `collectionId` sits in the same `FILTER_KEYS` list used by `test_first_fulltext_search_call_is_unscoped`, which only inspects `calls[0]`. A `collectionId` sent on call 2 or later has no coverage anywhere in the suite — confirmed only one other mention of `collectionId` in the whole test corpus, a prose note, not a check.

### Worth a look, less critical

- `SKILL.md` step 4 (line ~103) adds two sentences explaining *why* place vs. date/record-type filters risk false positives/negatives. That's rationale, not instruction — repo's own `CLAUDE.md` says SKILL.md must be instruction-only, reasoning goes in the spec/rubric.
- `_new_log_entries` is copy-pasted across 4 validator files (search-full-text, search-records, search-external-sites, search-images), not shared from `validators_lib.py`. This PR adds two more call sites to the unconsolidated copy rather than being the moment it gets lifted out.
- No `isinstance(query, dict)` guard on the two new log-fidelity checks, unlike the sibling `test_negative_result_log_shape` in the same file. A malformed `query` (string instead of object) raises a raw `AttributeError` instead of a clean assertion.
- `rubric.md`'s "fail" bullet for log-fidelity omits `recordType`, inconsistent with the "partial" bullet right above it and with the validator's own `FILTER_KEYS`, which treats it identically to the other filters.
- Minor: unused `import json` in the new test file; several fixture constants in the new test file are read by exactly one test each, so hoisting them to module scope bought no reuse.

### `research-log-protocol.md`'s new rule landed in one of three copies, and both siblings are concretely exposed

The new Rule 4 bullet ("only record a filter that was actually sent") went into search-full-text's copy only, replacing that file's own older "Include search parameters" rule. `search-records` (`references/research-log-protocol.md:29`) and `search-external-sites` (`:20`) still carry that same older wording, unchanged, and both skills build filter-bearing `query` objects for `research_log_append` the same way search-full-text does: `search-records/SKILL.md:515` explicitly defers to this same protocol file for query field guidance, and `search-external-sites/SKILL.md:290,317` calls `research_log_append` with its own captured-search query objects. So a record-search or external-sites run can log a query claiming a filter its actual tool call never sent, and nothing in either skill's doctrine would flag it, the exact defect class this PR closes only for search-full-text. (The three copies also differ structurally in an unrelated, pre-existing way — full-text's still uses an older narrative format the other two were already restructured out of — but that predates this PR and isn't something it touched.)

### `RECORD_TYPE_TERMS` coverage gap is real today, not just future drift

Checked against `enums.schema.json`'s `record_type_recommended` (an open enum with 12 documented example types) and the actual fixture corpus (328 `record_type` occurrences, 13 distinct values): 7 of the schema's 12 example types (land, military, newspaper, cemetery, tax, immigration, court) have zero table entries and zero corpus occurrences right now. The table looks adequate only because the corpus is 60% census. When a record type has no table entry, the content check silently `pytest.skip`s — indistinguishable from a genuinely irrelevant test, so a real gap in coverage looks the same as intentional non-coverage.

### Worth filing as follow-on issues, not blocking this PR

These point at fixing the underlying defect in the actual production tools rather than only in the CI-side test validators, which is a real and valuable direction, but touches files well outside this PR's scope (`research-log-append.ts`, `results-staging.ts`, `research-append.ts` — production TypeScript tools, not eval-harness Python) and deserves its own review:

- **`research-log-append.ts:308`** already computes the true search args host-side (`fin.payloadQuery`) but only uses it to fill in a `query` the model omitted, never to check one the model explicitly supplied. Confirmed directly — the comment there says so outright: "An explicit caller-supplied `query` always wins — this only fills a gap." That's the actual production-side reason the log-fidelity defect this PR fixes-in-CI can happen at all. Reconciling an explicit `query` against `payloadQuery`'s `FILTER_KEYS` would make the defect structurally harder to produce, not just easier to catch after the fact.
- **`results-staging.ts:61`** discards the computed `echoQuery` whenever a search returns zero results, so a negative-outcome log entry has no host-derived ground truth to check against at all — the exact case a "did we really search with no place filter" claim is hardest to verify.
- **`research-append.ts`**'s invariant dispatch has section-specific structural checks for plans/person_evidence/proof_summaries but none for plan items — nothing stops a plan item completing with zero log entries ever attributed to it, in production, independent of the CI validator this PR adds.

### Verdict

Real, fixable gaps in the two main validators this PR adds — items 1 and 2 specifically mean the checks can be defeated by ordinary narration and by the skill's own recommended search pattern, respectively. Worth a fix-and-re-request-review round rather than merging as is. Nothing here is a fundamental design problem in what this PR touches; all three "should fix" items are narrow, local fixes to the same functions. The follow-on items are a real, deeper direction but a separate PR's worth of work.

Verified directly rather than trusting the analysis: reproduced both "should fix" bugs myself with concrete Python before including them here, and confirmed the `research-log-append.ts` comment personally. All 10 review angles are now in, none outstanding.

chrisedeson (changes requested):
The review found two real bugs in my checks: one where an unrelated word ("will") wrongly counts as probate evidence, and one where a false claim about a search filter goes undetected as long as a second, similar search actually used it. Please read the comment above for broadness.

mercyokum:
Thanks for this — all three "should fix" items were real, and I verified each one directly before touching anything, same as you did.

**1. Bare "will" removed from `RECORD_TYPE_TERMS["probate"]`.** Confirmed the false-positive first, then confirmed it's gone. The list stays strong without it (testament, bequeath, executor/executrix, administrator/administratrix, appraise/appraisement, inventory, heir, legatee, probate).

**2. Log-fidelity matching is now positional, not "any call with the same keywords."** Reproduced your exact broad-then-narrow counter-example first — confirmed it slipped through under the old logic. Entries and calls are now grouped by keywords/nlQuery and paired by position within each group, so a false claim about call 1 can no longer hide behind call 2 actually sending it.

**3. `collectionId` now has its own always-applies check** (`test_fulltext_search_never_scopes_to_collection_id`), covering every call, not just the first. It's also out of `POST_SEARCH_FILTER_KEYS` now, since it isn't a "wait, then add" filter like place/date/record-type — SKILL.md forbids it outright, on any call.

**Also done, from "worth a look":**
- `isinstance(query, dict)` guard added to the log-fidelity check.
- `rubric.md`'s fail bullet now includes `recordType`, matching the partial bullet and `FILTER_KEYS`.
- Unused `import json` removed from the new test file.
- `_new_log_entries` consolidated into `validators_lib.py` and all four call sites (search-full-text, search-records, search-external-sites, search-images) switched over.
- SKILL.md's rationale sentences moved to `query-syntax.md` (kept SKILL.md instruction-only, per CLAUDE.md) — then re-ran the eval suite specifically to confirm the behavior fix survived the move rather than assuming it would. It did: every test that previously failed the pre-filtering check (002, 007, 008, 009, 011) passed clean in the fresh run.

Test coverage: `test_search_full_text_validator.py` grew from 12 to 20 tests, with a firing and passing case for each of your three fixes, all built from your exact reproductions or hand-built where nothing real exercises them (same as before). Full harness suite: 2395 passed, 3 skipped, ruff clean.

Filed the two follow-on items separately rather than folding them in: #1779 (same log-fidelity gap in search-records and search-external-sites) and #1780 (the three production-TS-tool findings — research-log-append.ts, results-staging.ts, research-append.ts).

One unrelated thing worth flagging since it showed up in the re-run: `ut_search_full_text_003` failed once, a routing-boundary miss unrelated to anything here (asked for the deed instead of routing to record-extraction) — same test failed once before in this dive's very first historical run, for the same class of reason. Annotated as known jitter, not a regression from this round of fixes.

chrisedeson (approved):
Verified all three fixes directly against the new code, not just the description. Will substring gone, log-fidelity matching is genuinely positional now (reproduced my exact counter-example, it correctly fails), collectionId has its own always-applies check. Suite green. Approving.

chrisedeson (approved):

EdmondOware (approved):
Verified the substance independently rather than relying on the existing reviews.

SKILL.md, rubric.md, and both reference docs line up cleanly with the three deep dive findings, log fidelity, first search scoping, and plan item completion. Read the diffs directly against issue #1651.

Confirmed both bugs chrisedeson found are properly fixed. The log fidelity validator now matches log entries to tool calls positionally within same keyword groups rather than any matching call, which closes the false negative chrisedeson demonstrated. collectionId now has its own always applies check separate from the first call check. Bare "will" is gone from the probate vocabulary after chrisedeson's false positive example.

The new validators are backed by a unit test file that proves each one can actually fire, drawn from real committed run log defects where one exists and hand built cases otherwise, rather than shipping unexercised. The duplicate log entry helper across four validator files was cleanly consolidated into validators_lib.

Checked the deferred issue too. #1736 was merged into #1748 rather than dropped, same underlying defect found independently by the citation dive.

Suite green, review sample fully annotated.

PR #1759, opened 2026-08-19
test-js: raise vitest's clock in every workspace config (#1366)

### LINKED ISSUE #1366: make test-js flakes on 5s test timeouts under parallel load

chrisedeson:
## Summary

Normal tier. Closes #1366.

- Raises vitest's clock from the 5000ms default to `testTimeout: 30_000,
  hookTimeout: 30_000` in all five workspace configs (viewer-ui, electron,
  web, eval/app, matching the precedent already set on the engine suite).
  `make test-all` runs every workspace suite through turbo in parallel, and
  the default has been intermittently tripping on tests that are fast
  standalone but land at 5.1-8.4s under that contention.
- Removes the per-test `30_000` override PR #1724 added to the electron
  archive-budget test, now that the suite-level raise covers it. Keeps its
  measurement comment.

## Review guidance

**Start here:** the electron config's comment and the test plan below. The
real evidence this works is a measured run, not a passing one.

**Unsure about:** nothing.

## Plan

**Didn't change:** turbo's concurrency, and the three flaky tests themselves
(fixture size, generator cost). The issue's own settled review already ruled
both out, only the clock moves.

**Acceptance check:** a real `make test-all` run under turbo contention
measured the electron archive-budget test at 5576ms, which would have failed
the old 5000ms default (matches the failure zone the issue reported,
5042-8025ms), and passed clean against the new 30000ms one.

**Deviated from the plan:** nothing. Followed the issue's settled checklist
as written, including checking PR #1724 first before touching that file.

## Test plan

- [x] Ran `make test-all` twice from a clean install. Both fully green.
- [x] Second run reproduced real contention: electron's archive-budget test
      measured 5576ms, inside the originally reported failure zone, and
      passed against the new budget.
- [x] `grep -rn testTimeout apps/*/vitest.config.ts packages/viewer-ui/vitest.config.ts eval/app/vitest.config.ts packages/engine/mcp-server/vitest.config.ts`
      returns five configs.
- [x] Ran each of the four touched packages' suites standalone too (electron
      65/65, viewer-ui 215/215, web 10/10, eval/app 168/168).
- [x] Full repo typecheck clean.

## Follow-on issues

**Folded in / filed instead:** none.


clack391 (approved):
Senior developer review, approving.

Implements issue #1366's settled checklist exactly: raises `testTimeout`/`hookTimeout` to `30_000` in the four workspace configs that lacked it (each with a why-comment), drops the now-redundant per-test `30_000` from the electron archive-budget test (keeping its measurement comment), and correctly leaves the engine config alone since it already carries the raise. No per-test timeouts added, no fixture shrink, no lint. #1724 confirmed merged (`331072c3`), so the override deletion is correctly sequenced.

Verified by running on the tree merged with current main (clean merge):

- Break-test: forcing the electron config to `testTimeout: 1` makes the archive-budget test time out at 1ms, which proves the key is honored inside `test: {}`. So the per-test override is safe to remove; the suite-level raise genuinely covers it (that test passes at ~2.8s without it).
- Suites green: electron 65/65, viewer-ui 215/215, web 10/10, eval/app 168/168.

Collision with PR #1751 on `feedback.test.ts` is non-overlapping (it edits the import, `makeOptions`, and the redaction block; this edits only the archive-budget test, with no vitest.config overlap), so neither blocks the other.

Gennecis (commented):
The checklist from issue #1366 is followed item by item, the four configs are the complete set, and removing the per-test `30_000` is safe because the suite-level raise is the same number — I confirmed the config key is what governs that test now.

One sentence to fix before it's merge. It is in a comment, and the comments are the part that lasts: once #1366 closes, they are the only record of why the raise is there.

**`packages/viewer-ui/vitest.config.ts:11-13` — the range is not this package's range.**

You wrote:
```
    // suite through turbo in parallel, and cases here measured well under 150ms
    // standalone still land at 5.1-5.6s under that contention, tipping over the
```

Change it to:
```
    // suite through turbo in parallel, and cases here measured at 78-235ms
    // standalone (hardware moves this) still land at 5.3-6.2s under that
    // contention, tipping over the
```

Two reasons. Issue #1366 records viewer-ui failing at **5349ms, 5578ms and 6166ms**, so the range is 5.3-6.2s — the `5.1` is the electron test's 5088ms, and the `5.6` upper bound drops the 6166ms SidecarPanel case, which is one of only two viewer-ui failures on record. And `well under 150ms` is one machine's figure: it holds on the issue's macOS box (105ms / 78ms), but I timed the same two cases on WSL2 — the box the flake was reported from — at 201ms and 235ms. The issue warns about exactly this: "hardware moves this one a lot."

Nothing else. The electron config's `~2.4s standalone` carries the same single-machine caveat, but that figure comes from PR #1724's comment you were told to keep, so leave it.

Gennecis (approved):

Issue #1757, opened 2026-08-19
Renaming a validator's gating tag silently disables 1 of 81 tag-gated assertions — snapshot ignores tags, skip reports passed


johnmarkpeterbrown:
**Touches:** eval/harness/harness/snapshot.py, eval/harness/validators/*.py, docs/specs/unit-test-spec.md

Found during the search-familysearch-wiki deep dive (#1647) after noticing that a
tag-only edit to a test JSON did **not** invalidate the skill's run log.

Two facts that are individually reasonable and jointly a fail-open hole:

1. **`tags` is cosmetic to the run-log snapshot.**
   `harness/snapshot.py:47` — `_COSMETIC_TEST_FIELDS = ("name", "description",
   "tags")`, stripped by `_strip_cosmetic_test_fields` before a test JSON is
   hashed. So changing a tag never flips a run log inactive and never triggers a
   re-run.
2. **`tags` is load-bearing to validators.** **81 tag-gated assertions** across
   `eval/harness/validators/*.py`, keying off 40+ distinct tags —
   `no-wiki-search`, `no-search-no-write`, `no-tree-write`,
   `new-question-exhaustive-declaration-unstarted`, `preserves-src001-original`,
   `conflict-blocks-proved`, and so on. The gate is
   `if "<tag>" not in test.get("tags", []): pytest.skip(...)`.

**And a skipped validator is recorded as passing.** From
`v1_2026-07-30_00-42-37`:

```json
{"name": "test_no_spurious_wiki_search", "passed": true,
 "error": "skipped: only applies to no-wiki-search negative tests"}
```

So renaming, mistyping, or dropping a gating tag silently disables its validator,
the run log still reports **active**, the validator still reports **passed**, and
no gate anywhere notices that an assertion stopped running. Coverage falls to
zero with a green board.

## Why this is worth fixing rather than documenting

The blast radius is the whole deterministic layer. `test_no_spurious_wiki_search`
is described in its own docstring as "the real gate that keeps
`grade_on_invariant` from passing vacuously" — that is, a `grade_on_invariant`
negative test's *entire* correctness rests on a validator whose activation hangs
on a string the snapshot deliberately ignores. Same shape for
`test_no_wiki_no_write`, `test_no_search_no_write`, and the four `test_slug_*`
checks in `test_search_wikipedia.py`.

Nothing has gone wrong yet that I found — this is a latent guard gap, which is
what the `nothing-checks` label is for.

[...]

Issue #1756, opened 2026-08-19
gate-skill's no-regression check is inert for 17 of 25 skills (no holdout tests) and says so in one skippable line


johnmarkpeterbrown:
> **Reviewed 2026-08-20 before junior handoff.** The decision is settled — the
> holdout mechanism is being removed, not filled in. The original body follows
> under "Original issue"; read it for the evidence, not for the ask.

**Touches:** eval/harness/skill_gate.py, eval/harness/tests/unit/test_skill_gate.py, Makefile, eval/GateSkill.bat, docs/skill-lifecycle.md, docs/alpha-feedback-guide.md, docs/specs/unit-test-spec.md

## Ruling 2026-08-20 — delete the holdout comparison

`gate-skill` stops reading `test.holdout` and stops claiming a no-regression check.
It compares the **motivating test only**, and its report says so plainly.

**Why, and it is not a compromise.** You cannot land a skill edit without a fresh
full-suite `make eval-skill` run — `check_runlogs` blocks the PR until one plus its
annotation lands, and that run covers every test in the suite. The holdout set is
therefore a *preview* of a check that is mandatory downstream. **Removing it loses an
early warning, not coverage.**

Two things rejected, and why:

- **Filling in the missing holdouts.** 18 of 25 suites have none. Designating them
  flips each suite's run log inactive, so the wave costs **$53.78 of API plus 18
  annotation passes** — to buy a preview of something already guaranteed at PR time.
- **Making the gate refuse a clean verdict without holdouts.** Same objection, spread
  over time instead of arriving at once, and it makes the documented `LOOKS GOOD` exit
  unreachable for 18 suites.

Also weighed: **a human cannot pick 2-3 tests that give real regression coverage.** The
one worked example chose 3 of 16 on stability and shape diversity — defensible, and
nowhere near coverage. If the mechanism is ever revived it must select automatically
from run-log history (stability across committed runs, shape spread), not by hand.
`skill_gate.py` already says this about itself: *"Holdout no-regression is a weak
secondary check … generalization-by-inspection is the primary guard."*

## Do this

- [ ] **`eval/harness/skill_gate.py`** — delete `holdout_test_paths` and the holdout
      arm of `compute_signal`, `render_report` and `main`: the `Holdout:` header line,
      the per-holdout tables, the italic inert line, the two stderr `NOTE`s, and the
      "absent from the baseline" skip path. The verdict vocabulary
[...]

Issue #1755, opened 2026-08-19
search-familysearch-wiki: teach it wiki_place_page, and the closing-message brevity validator


johnmarkpeterbrown:
> **Scope reduced 2026-09-07 (`/fill-ready`): half C was split out to issue #2281.**
> This card now covers **only** halves A and B. Validator activation — the merged #1757
> and #1930 material, the assertion count, the three-plus-one candidate fixes and the
> `unit-test-spec.md` §5.1 amendment — moved to issue #2281 and is **not** in scope here.
> It was split because it touches only `eval/harness/**` and `docs/specs/unit-test-spec.md`,
> buys no paid eval run, and is independent of A and B, as this card's own header argued.

> **Reviewed 2026-08-27 before handoff — do not start from the numbered requests below.**
> Four of the five landed on `main` in PR #1762 (`241610ec`, `987553f7`, `1bd1f9b4`) and the
> title has not been updated. Verify before reading further:
> `grep -n "^def test_" eval/harness/validators/test_search_familysearch_wiki.py`
> — requests 1, 2, 3 and 5 are `test_sources_section_matches_wiki_results`, `test_expected_slug`,
> `test_wrote_exactly_one_markdown_file` + `test_no_file_on_empty_results`, and
> `test_no_wiki_no_write`. All 12 positives carry an `expects-file-<name>` tag (note: **not**
> `slug-*` — see the docstring at `test_search_familysearch_wiki.py:62`). Requests 1, 2, 3 and 5
> below are **historical record only.**
>
> **Verdict: `senior`.** This stays hard after every open question is answered — see the three
> halves below.
> **High priority (2026-09-14):** holds locality-guide — #2117, #2477, #2486, #2518, #2547 waiting.

**Touches:** packages/engine/plugin/skills/search-familysearch-wiki/SKILL.md, packages/engine/plugin/skills/search-familysearch-wiki/templates/wiki-search-summary.md, eval/tests/unit/search-familysearch-wiki/*.json, eval/tests/unit/locality-guide/ (reciprocal negative), eval/fixtures/mcp/ (new wiki_place_page fixtures), eval/harness/validators/test_search_familysearch_wiki.py

(`eval/harness/harness/snapshot.py` and `runnability.py` were dropped from this line on
2026-09-07 — they belong to issue #2281.)

**Two things remain, and they are two tasks, not one.** Half C left for issue #2281.

| Half | Buys a paid run? | Startable now? |
|---|---|---|
[...]

PR #1762, opened 2026-08-19
search-familysearch-wiki deep dive (#1647): three grading fixes, four validators, two runs

### LINKED ISSUE #1647: deep dive: search-familysearch-wiki
**Touches:** eval/tests/unit/search-familysearch-wiki/, packages/engine/plugin/skills/search-familysearch-wiki/SKILL.md

Audit `search-familysearch-wiki` for defects nothing has caught. This starts from **no known

Did / Should / Gap.

## Start with this grep

```sh
grep -l -iE '"[^"]*\bscore [123]\b' eval/tests/unit/search-familysearch-wiki/*.json
```

**This skill has 1 such file(s).**

without writing the finding.

## This skill's numbers

| | |
|---|---|
| Tests in the suite | 16 |
| Rubric+base dimensions that never discriminate | **2 of 7** |
| `judge_context` files naming a score branch | 1 |

2 of 7 dimensions never take more than one value, so they cannot report a defect in whatever they nominally grade.

## What to hand back

## Cost

Editing anything under `eval/tests/unit/search-familysearch-wiki/` or the skill body flips this
skill's run-log snapshot inactive and buys one `make eval-skill SKILL=search-familysearch-wiki`

johnmarkpeterbrown:
## Summary

**Tier: Normal.** Deep dive on `search-familysearch-wiki` (#1647) — an audit that starts from no known failure — plus the validators it produced (#1755, folded in rather than deferred).

- **Grading fixes (genealogist lane).** A `judge_context` that spelled out its own score branches was making `ut_search_wiki_002` a false partial on three consecutive runs; rewritten to name the dimension instead, and scoped to bind the base dimensions too. The slug rule was silent on period qualifiers, which `ut_search_wiki_007` is entirely about — settled (periods are kept) after four runs of drift that scored 3 either way. Three holdout tests designated, which the gate had none of.
- **Validators (developer lane).** Four of the five requested in #1755, with 19 prove-it-fails tests. The important one reads the saved `.md` — the skill's actual deliverable, which **nothing had ever checked**.
- **Four paid runs + two genealogist annotations** are committed: a pre-edit baseline, a post-edit verification, the run where a new validator caught a second slug ambiguity, and the release candidate `v1_2026-08-20_14-29-26` (15 pass / 1 partial, up from 14 / 2 — 0 aborts, 0 retries, drift 0).

Eight findings: 2 fixed by prose, **1 raised and overturned by the genealogist**, 5 closable only by a program. That ratio is the write-up's main claim and it is on #1647 in full.

## Review guidance

**Start here:** `eval/harness/validators/test_search_familysearch_wiki.py` — specifically `test_sources_section_matches_wiki_results` and the module docstring above it. The decision worth checking is *why a validator rather than the judge*: `judge.py` receives only `file_changes_summary`, which `orchestrator.py` builds from the research.json / tree diffs — both `None` for a skill whose deliverable is a loose `.md`. The harness's existing `test.judge_reads_files` opt-in does **not** close this; it surfaces content out of that same research/tree diff, so it is inert here. If you think that reading is wrong, this whole file is the wrong shape and I would rather hear it now. Second: the fabrication arm strips HTML comments before scanning for URLs — that is not defensive coding, it is the fix for a real false positive the first live run produced (the template ships a placeholder URL in a comment and the skill leaves it in the file).

**Unsure about:** nothing outstanding. **The `rubric.md` "No-result handling" bar is now REVERTED** (`1bd1f9b4`) — the file is byte-identical to `main`, and a brief redirect after a nil result scores pass again, matching the genealogist ruling recorded on #1647 as F4. Earlier revisions of this description said the bar was "known wrong and queued for revert"; it is not queued, it is done. Do not re-open the point.

The history is kept because it is the useful part: I tightened that bar on a guess about what "and stop" forbade, paid for a run to test it, and the genealogist overturned it twice. `@mercyokum` then made the argument I had missed — the validator and the rubric actively disagreed, with only the current judge's behaviour keeping it quiet — which is why it was reverted here rather than deferred.

## Plan

**Didn't change:** the SKILL.md body, for the brevity rule. **59 of 59** file-saving runs across five run logs restate the file's content in chat, against step 5's "Keep it brief" — but the rule is already in the body, and the deep-dive guide is explicit that restating an obeyed-and-ignored rule is not the fix. It is worse than that: on `v1_2026-08-19_13-00-35` the judge marked the skill *down* for **not** quoting the file it is forbidden to quote, so the graded suite has been rewarding the violation. That is a grading problem, and the fix is the file-visibility validator here plus #1755 request 4 afterwards — not prose.

**Acceptance check:** `cd eval/harness && uv run pytest tests/unit/test_search_familysearch_wiki_validators.py` — **21 tests**, absent on `main`, each exercising a new assertion against both a passing state and the specific state observed failing in a committed run log. Validators are not collected by `make harness-test` (`pyproject testpaths = ["tests"]`), so without this file the checks would go unexercised until someone paid for a run. Behavioural acceptance: `ut_search_wiki_002` is `partial` in `v1_2026-08-19_13-00-35.json` and `pass` in `v1_2026-08-19_14-53-19.json`, both committed, and the human annotation confirms the new 3s.

**Deviated from the plan:** three times, each recorded above or on the issue.
1. `judge_context` shipped with **three** clauses, not the two drafted — the third suppresses the file-invisibility penalty, which turned out to be the *actual* cause of the current failure.
2. #1755 was **folded in after being filed**. `CLAUDE.md` step 1 applies ("you already have the context loaded"), and the exemption I claimed when filing — *needs a different reviewer* — was weak on a branch that already owns this suite. It landed with **no paid run**: validators are not in the run-log snapshot and `tags` is in `_COSMETIC_TEST_FIELDS`, so drift stayed 0.
3. The `rubric.md` tightening **failed on measurement and was reverted in this PR** (`1bd1f9b4`, at `@mercyokum`'s request). I also had to retract a claim I had already posted to #1404 on the back of it.
4. A fourth deviation, found by the validators themselves: `ut_search_wiki_011` turned out to have a **second** slug ambiguity (coordinated record types — "probate **or** will records"), so `SKILL.md` gained one more clause than planned. Detail under Follow-on issues.

## Test plan

      **With one pre-existing failure, unrelated to this branch, and `make` unavailable.** `scripts/test.sh` requires `make` on PATH, which this Windows machine lacks, so each suite was run directly: typecheck **5/5 tasks**, engine-test **2438 passed**, eval-ui-test **168 passed**, harness-test **2438 passed / 3 skipped** (was 2355 at branch start), packaging **496 passed**, `ruff` clean, `check_runlogs.py` exit **0** on both rules. `server-test` (`apps/server` pytest) never finished locally on this machine; CI's **`server-pytest` check passes** and this branch touches no `apps/server` code.
      `pnpm test` fails on `@genealogy/electron` → `apps/electron/src/main/__tests__/feedback.test.ts:402`: `ENOENT … mkdir '…\.claude\projects\-C:\Users\…'`. The test derives a directory name by replacing `/` with `-` in a path, which on Windows leaves the drive colon in a filename — illegal on NTFS. **This branch touches no `apps/` code and that file is byte-identical to `main`.** It means the documented pre-PR gate cannot pass on Windows at all, which matters for a Windows-based team; not filed yet — flagging for whoever owns `apps/electron`.
- [x] For every skill whose **run-log snapshot** I changed … I ran `make eval-skill SKILL=<name>` and committed the run log **and its `.ann.json`**.
      Four runs committed: `v1_2026-08-19_13-00-35.json` (pre-edit baseline), `v1_2026-08-19_14-53-19.json` + `.ann.json` (post-edit), `v1_2026-08-20_12-52-08.json` (kept deliberately — the run where `test_expected_slug` caught `ut_search_wiki_011` and took the judge down with it, i.e. what a validator catching a real defect looks like in a log), and `v1_2026-08-20_14-29-26.json` + `.ann.json` (release candidate). Retention pruned the July 01 and July 22 logs. `check_runlogs.py` exits **0** on both rules.
      **N/A** — the `description` and DO-NOT clauses are untouched. The only SKILL.md change is the slug rule inside step 4, which affects no routing boundary.

## Follow-on issues

- **#1755** — validators. **4 of 5 folded into this PR**; left open for request 4 (the brevity check), which cannot land yet: it would fail all 12 positives at once, and a failing validator *skips the judge* (`_compute_outcome`: `if not validators_passed: return "fail"`), destroying the dimension scores that diagnose the skill. It now also carries the merged body of #1757 under "Merged: how validators are activated", plus the `nothing-checks` label.
- **#1756** — `gate-skill`'s no-regression check is inert for **17 of 25 skills** (zero holdout tests), and says so in one line under a heading that still prints LOOKS GOOD. Found by designating this skill's three holdouts. `nothing-checks`.
- **#1757** — **CLOSED as `not planned`, merged into #1755** by @DallanQ, on the grounds that the two are one piece of work: moving a validator's expectation off `tags` and into a snapshot-hashed field would otherwise flip this skill's run log inactive twice and buy two paid runs for one skill. The full body, the `nothing-checks` label, and the acceptance test carried over. **Correction to my own figure:** I reported "81 tag-gated validator assertions" from a `grep` line count with a loose pattern; @florencemashipei could not reproduce it. Recounted by AST on this branch — test functions that both reference `tags` and call `pytest.skip` — the number is **87**, across 20 validator files. A stricter grep gives 79 because it misses the `not in (test.get("tags") or [])` form. @florencemashipei also established the hole is **bidirectional**: *adding* a gating tag likewise leaves the run log active, so a tags-only PR can ship an assertion that never runs. This PR still uses tags, matching `test_search_wikipedia.py`; the docstring on `_expected_file` records why and points at the migration.
- **#1404** — commented, then **corrected**: my first comment claimed the failed bar-tightening proved judges won't fail absence-shaped dimensions. It proved my bar was wrong about the craft. What survives is the flatness, now human-confirmed: **6 of 7 dimensions take a single value**.

**Folded in / filed instead:** folded in #1755 (4 of 5 requests) — reason: none of the four filing exemptions actually applied once I checked the cost, and step 1 is the default. Filed rather than folded: #1756 and #1757, both because they are missing guards in shared harness machinery rather than in this skill, and `nothing-checks` is a register that wants one row per gap. Not filed: the `apps/electron` Windows test failure above (flagged here, not mine to route) and the `judge_reads_files` extension idea (recorded on #1755 — it would fix this class for every markdown-writing skill at once and is a bigger design call than a validator).

**Cost:** $6.44 across 8 harness invocations — 4 full suite runs and 4 targeted probes. $0.86 of that was wasted on a concurrency-3 attempt that hit an API wobble (6 of 15 tests aborted with zero tool calls; the harness correctly refused to release it). A $0.09 single-test probe settled the same question afterwards and is what I should have run first; the final run at concurrency 2 had 0 aborts and 0 retries.

Closes #1647.


johnmarkpeterbrown:
## Tag collision fixed — `987553f7`

Reproduced it first, exactly as you measured: **11 match, 1 mismatch, on both run logs**, `normalization.md` vs `census-records.md`. And it was worse than a one-off — **my tagging script carried the identical assumption**, which is why it printed "already tagged" and skipped that file. The bug wrote itself twice, from the same wrong premise.

### I went one step past your fix, and I'd like your read on it

Your suggested tags were:

```
"tags": ["fs-wiki", "how-to", "census", "slug-census-records"]
```

I instead made the prefix unambiguous — filename declarations are now `expects-file-<name>`:

```
"tags": ["fs-wiki", "how-to", "census", "slug-normalization", "expects-file-census-records"]
```

Two reasons, both from your own review:

1. **It closes the class, not the instance.** You wrote *"Next person who reuses the helper hits this again"* — and with a `slug-` scan they will, because `slug-` is an established **descriptive** prefix. `expects-file-` cannot mean anything else, so the four `search-wikipedia` tags you named (`slug-simple`, `slug-parens`, `slug-numbers`, `slug-single-word`) become permanently safe to point this helper at.
2. **It keeps `slug-normalization`.** Your version replaces it, and it carries real information — that test exercises slug normalisation. Nothing gates on it (I grepped), but deleting a descriptive tag to make room for a mechanical one loses something for no gain.

If you'd rather have the minimal version, it's a one-line revert and I'll take it — you're the gate.

### Verification

- **12 of 12** expected filenames now match on **both** committed run logs, up from 11.
- Two new tests pin the collision: one asserting a descriptive `slug-*` tag alongside `expects-file-*` resolves correctly, one asserting two `expects-file-*` tags **fail loudly** rather than silently taking the first (an authoring error, not a precedence question).
- Free, as you established: latest run log stays at **drift 0**, `check_runlogs.py` exits **0**. And you're right that #1757 isn't theoretical — this retag changes what's asserted and nothing in CI would notice.

### Both optional notes applied

- `ut_search_wiki_005` and `_015` now carry `no-wiki-search`, so both arms bind. Passes on the committed evidence — no tool calls and no files in either run, both logs.
- `_expected_file`'s docstring now records why the prefix isn't `slug-`, names the four colliding tags, and describes the failure mode so the next reader doesn't have to rediscover it.

### One thing your review surfaced that I should flag

Your caveat — *"none of the five new validators is in either committed run log … the first person to run this skill is the first to execute them"* — is exactly right, and it's the same hole that let the collision through. The worktree also had **no engine build** (gitignored, and `link-worktree.sh` deliberately doesn't link it), which made `test_mock_mcp.py` fail in a way that reads as a code regression; `CLAUDE.md` calls it "a missing build wearing the costume of a code regression." Built it, and the 6 build-dependent tests now run instead of skipping.

Worktree suites: harness **2396 passed / 3 skipped**, engine vitest **2438 passed**, ruff clean, `check_runlogs.py` **0**.

Rubric hunk left as-is per your call, and agreed the genealogist who overturned it should sign that part off — CODEOWNERS routes `eval/tests/` to them anyway.

clack391 (changes requested):
Senior developer review. The developer lane is sound and I would approve it once one thing is out. One change before merge, plus two process notes.

## The developer lane checks out

I verified by running on the merged tree (0 behind main):
- The four new validators are non-vacuous: I broke each assertion branch one at a time and watched a specific unit test go red, then restored. The headline test_sources_section_matches_wiki_results does not pass vacuously (it skips on the negative paths and fails on an empty or absent file, never silently passes), and it reads the saved .md, which nothing checked before.
- The expects-file-* tag mechanism resolves correctly, fails loudly on two such tags, and is not fooled by a descriptive slug-* tag; all three are pinned.
- check_runlogs exits 0 (both runs active and annotated); full harness suite green (the lone test_mock_mcp failure is the missing-engine-build case, untouched by this diff); the SKILL.md slug/period rule is clean; no collision with the other deep-dive PRs.
- #1755 reconciled: 4 of 5 folded, request 4 (brevity) correctly deferred because a failing validator skips the judge.

## The one change: pull the rubric.md "No-result handling" hunk (revert to main)

Take it out of this PR and take the re-run. It is the only thing blocking my approval, and the reasons it should not ride to main even inert:

1. Your own annotation on ut_search_wiki_016 already rules it a defect: "A brief redirect after a nil result is correct practice. RUBRIC DEFECT ... revert it." So this is enforcing your genealogist ruling, not second-guessing it.
2. It is inert only because the judge did not enforce the new partial band this run (it scored the redirect a 3). That is luck, not safety. A stricter judge pass, or the judge-prompt drift already moving on main, could enforce it and produce exactly the false partial the annotation warns about, and tempt someone to tune the skill to stop redirecting.
3. "Ride the next run this skill needs" defers a known grading defect to an unscheduled future run (request 4 cannot land yet). That is how a known defect becomes permanent. A deep dive whose whole point is grading integrity should not ship a known grading defect.

The cost is one re-run plus a quick annotation, and you already have the evidence to make it fast. If a senior genealogist decides that run cost is not worth it right now, the only safe defer is to make the revert a blocking checkbox on the next search-familysearch-wiki run, bundled onto #1755 request 4, not an annotation in a run log.

## Two process notes

- The reviewer feedback you answer in the thread ("as you measured", "you are the gate") is not on the PR. The timeline shows only review_requested and your own comment, so there is no on-PR record of what was reviewed. Worth asking that reviewer to post their review so the trail exists.
- You are a senior genealogist and cannot self-approve, so the craft lane here (the rubric decision, the annotations, the slug doctrine) needs a different senior genealogist. My approval would cover the developer lane and code-owner once the rubric hunk is out.

mercyokum (changes requested):
eval/tests/unit/search-familysearch-wiki/rubric.md, the "No-result handling (negative path)" section.

The partial band here still scores a brief redirect after a nil result as partial. Your own PR shows this is wrong. The annotation in v1_2026-08-19_14-53-19.ann.json for ut_search_wiki_016 says it directly: "Pass. A brief redirect after a nil result is correct practice. RUBRIC DEFECT: rubric.md's partial band (added under #1647) wrongly calls this a partial, revert it, do not tune the skill to stop redirecting."

I understand why you left it in. The PR body says it is inert today, it changed no score, and reverting it now would flip the run log's snapshot to stale, meaning a third paid run just to fix a document. That is a real cost and I am not dismissing it.

But I would rather pay that cost now than ship known wrong grading doctrine to main. The validator side already matches the genealogist's ruling correctly, test_no_file_on_empty_results only checks that no file was written, it does not check that the skill stopped. So right now the validator and the rubric actively disagree with each other, and the only thing keeping that quiet is that the current judge happens to score it right anyway. That will not necessarily hold on the next run or the next model version, and a rubric that quietly contradicts a genealogist's own ruling is not something I want sitting in the codebase waiting to be rediscovered.

Please revert the No-result handling pass, partial, and fail bullets back to what they said before c5c3e203, so the text matches the ruling, and run the eval so the run log's snapshot matches again. If the cost is the real blocker, say so and I will make the call on whether it waits, but I do not want it merged silently deferred without that being a decision rather than a default.

johnmarkpeterbrown:
## Requested change done — `1bd1f9b4`

@mercyokum — reverted, and you were right that I should not have deferred it.

`rubric.md`'s "No-result handling" pass/partial/fail bullets are back to their pre-`c5c3e203` wording. Not approximately: **the file is now byte-identical to `main`**, so `git diff origin/main -- eval/tests/unit/search-familysearch-wiki/rubric.md` is empty.

The argument that changed my mind is the one I had not made myself:

> "the validator side already matches the genealogist's ruling correctly … So right now the validator and the rubric **actively disagree with each other**, and the only thing keeping that quiet is that the current judge happens to score it right anyway."

That is exactly the failure mode of the original defect — an inert wrong bar waiting to be rediscovered — and my "it's inert, defer it" framing was also incomplete on cost, because the run was buying something else too (below).

Release candidate: **`v1_2026-08-20_14-29-26.json` + `.ann.json`**, annotated. 15 pass / 1 partial, 0 aborts, 0 retries, **drift 0**, `check_runlogs.py` exit **0** on both rules. `ut_search_wiki_016`'s No-result scores **3**, which is now the correct score rather than one the rubric contradicts.

---

## The run found a second slug ambiguity — the validators earned their keep

@clack391 — your caveat was the important one:

> "none of the five new validators is in either committed run log … the first person to run this skill is the first to execute them. That's how the tag collision got through."

They have now executed against live data, and one immediately caught something no other check in the suite could see.

`ut_search_wiki_011` ("How do I find probate **or** will records…") had written `probate-will-records.md` in four committed runs. On `v1_2026-08-20_12-52-08` it wrote **`probate-records.md`** — and `test_expected_slug` failed, which took the judge down with it. Under the old suite that would have scored 3, exactly as `ut_search_wiki_007` scored 3 for four runs while flipping between `death-records.md` and `death-records-1800s.md`.

Same class as the period ambiguity: **the slug rule was silent, so the skill picked arbitrarily.** `SKILL.md` now reads "When the question names two record types joined by 'or' or 'and', keep both in the order asked", with a worked example. "In the order asked" is the load-bearing half — without it `will-probate-records.md` is equally defensible and the ambiguity just moves.

I have committed `v1_2026-08-20_12-52-08.json` deliberately, as the evidence: it is what a validator catching a real defect looks like in the log, judge-skip and all.

Both slug axes — periods and coordination — are now stated in the body and pinned by a validator. **12 of 12** expected filenames match on the release candidate.

---

## Annotation

Two rulings recorded, no judge/human disagreements.

`ut_search_wiki_008` Correctness stays **2** — "but not for the stated reason". The judge blamed era flattening, then its own rationale confirms the eras are accurate; the real defect is *"1790–1940 (plus the 1950 census released in 2022) are searchable for free at FamilySearch"*, which attaches free FamilySearch access to 1950 when the wiki only says it "was released in 2022". That is the **third** self-refuting Correctness rationale on this suite, each caught by a human and none by any automated check. Noting it, not escalating it — I escalated a pattern here once already and was wrong.

`ut_search_wiki_016` records the craft standard now that the rubric matches it. The previous run's "RUBRIC DEFECT — revert it" note is deliberately **not** repeated, because this commit is the revert.

---

## Cost, including what I wasted

**$6.44 across 8 harness invocations** on this branch — 4 full suite runs and 4 targeted probes. (I quoted $5.67 in an earlier comment; that was before the final run.)

**$0.86 of it was wasted.** A concurrency-3 attempt hit an API wobble: 6 of 15 tests aborted with zero tool calls and empty responses, 12 transient retries, abort-storm breaker tripped, wall time 421s → 1077s. The harness correctly refused to release it — it wrote a gitignored `scratch_` log — but `ut_search_wiki_011`, the one test that run existed to answer, was among the aborts, so it answered nothing.

A **$0.09** single-test probe afterwards settled the same question in 88 seconds with zero retries. That is what I should have run first. Dropping to concurrency 2 for the final run gave 0 aborts and 0 retries.

---

## Suites

harness **2438 passed** / 3 skipped · packaging **496 passed** · `ruff` clean · `check_runlogs.py` **0**

One artefact to ignore on review: git recorded the new `.ann.json` as a 91%-similar **rename** of the pruned July-01 annotation rather than a new file. Content is the new annotation; it is just git's similarity detection.

mercyokum (approved):
Look good! Approving

clack391 (approved):
Re-review: my earlier CHANGES_REQUESTED is resolved. The defective "No-result handling" partial band is reverted (1bd1f9b4), the rubric is byte-identical to main, and a brief redirect after a nil result now scores pass, which matches the genealogist ruling recorded in #1647 (F4). I confirmed the new validators are non-vacuous by mutation (each assertion fails when broken, restored), check_runlogs exits 0, and the harness suite is green apart from the environmental test_mock_mcp (engine build absent, no MCP code touched here).

Approving the developer lane and the developer code-owner (the .ts files). mercyokum's approval covers the genealogist lane. Two things still stand between this and merge, neither blocking my approval: it is 23 behind main so it needs an update-branch, and the PR description is stale in a few spots (separate comment below) worth a quick refresh.

clack391:
Non-blocking, but worth a quick PR-body refresh so the description matches the code (a later reader going by the body alone could re-open the resolved rubric point):

- The "Unsure about" section still says the rubric.md change is "known wrong and queued for revert, do not read it as the fix." It is already reverted (1bd1f9b4). Please update it to "reverted" so no one re-raises the resolved point.
- Acceptance check says "19 tests"; the validator file now has 21 (the tag-collision fix added two).
- "Two paid runs committed" is now four (the 08-19 pair plus the 08-20 pair).
- The follow-on list cites #1757 as filed; it is closed as not-planned and merged into #1755.

PR #1766, opened 2026-08-19
convert-dates: deep-dive findings + validator/rubric/tool fixes (#1654)

### LINKED ISSUE #1654: deep dive: convert-dates
**Touches:** eval/tests/unit/convert-dates/, packages/engine/plugin/skills/convert-dates/SKILL.md

Audit `convert-dates` for defects nothing has caught. This starts from **no known

Did / Should / Gap.

## Start with this grep

```sh
grep -l -iE '"[^"]*\bscore [123]\b' eval/tests/unit/convert-dates/*.json
```

without writing the finding.

## This skill's numbers

| | |
|---|---|
| Tests in the suite | 14 |
| Rubric+base dimensions that never discriminate | **3 of 5** |

3 of 5 dimensions never take more than one value, so they cannot report a defect in whatever they nominally grade.

## What to hand back

## Cost

Editing anything under `eval/tests/unit/convert-dates/` or the skill body flips this
skill's run-log snapshot inactive and buys one `make eval-skill SKILL=convert-dates`

Emruthwill:
Closes #1654.

Deep dive on `convert-dates` per [`docs/skill-deep-dive-guide.md`](docs/skill-deep-dive-guide.md), starting from no known failure. The issue's opening grep returns 0 files; everything here came from reading the 56 committed runs against the body's own rules, then registering the tool and running the suite.

**Sibling dives, for reviewers comparing conventions:**

- [#1762](https://github.com/PioneerAIAcademy/cowork-genealogy/pull/1762) — `search-familysearch-wiki` deep dive (issue #1647): three grading fixes, four validators, two runs. **Its code pattern is followed here**: validator in `validators/`, proof-of-failure tests in `tests/unit/test_<skill>_validators.py`, imports aliased away from the `test_` prefix so pytest does not collect the validators as tests of the test module.
- [#1745](https://github.com/PioneerAIAcademy/cowork-genealogy/pull/1745) — `citation` deep dive (issue #1660). Its title form is followed here, and now its layout too: #1745 has merged, `docs/deep-dives/` holds six files from three dives, so the condition this PR set for converging ("whoever merges second can converge, once the target is real") is met. Write-up and prohibition list are split into the two conventional files.

**Write-up:** [`docs/deep-dives/convert-dates-findings-2026-08-19.md`](docs/deep-dives/convert-dates-findings-2026-08-19.md)
**Prohibition list, saved for the next auditor:** [`docs/deep-dives/convert-dates-prohibition-list.md`](docs/deep-dives/convert-dates-prohibition-list.md)

## The headline

`convert_calendar` had **never been called** — 56 committed runs, 0 tool calls, and all 56 `Tool Arguments` gradings null. It was registered nowhere, so it never entered the model's tool list, and the skill did the arithmetic by hand in every run — which `SKILL.md` prohibits by name. The dimension that covers tool work is only graded when `tool_calls` is non-empty, so **the defect switched off the dimension that would have caught it.** The judge and a human annotator each blessed it in writing.

`check_tool_coverage.py` had been warning all along — but its proposed remedy ("add a test with an mcp_fixture") is the wrong one for pure arithmetic, where a canned answer would assert the computation under test. That is likely why it stood for two months. The right remedy, `LIVE_TOOLS`, was not among the options the warning offered.

Registering it (four edit sites, not the one an earlier draft claimed — the `OK_FALSE_IS_FAILURE` drift lint makes three mandatory, verified by breaking it and watching it fail) then exposed two findings that were **unreachable by reading run logs**: F6 and F8.

## Findings

| | Finding | Lane | State |
|---|---|---|---|
| F1 | `convert_calendar` never called; `Tool Arguments` N/A in all 56 gradings | 1 | fixed |
| F2 | Offset thresholds regressed from leap-day to calendar-year; wrong for Julian Jan–Feb of 1700/1800/1900, and no test reached the window | 4 | fixed + new test `_015` |
| F3 | `_007` was built on 25 March — the day the OS year *increments*, which its own rule excludes | 2 | fixed at the input; boundary split to `_016` |
| F4 | Judge scored 3 with "correctly restrains itself" on runs that computed the unrequested day shift | 2 | fixed; now intermittent-but-caught |
| F5 | `judge_context` is a binding override the judge ignores — **three witnesses, escalating** | 2 | shipped; Facts arm **proven** in run 4, Authority arm untestable (see F10) |
| F6 | `_004` required a conversion the tool spec §7 defines as an input error | 2 | fixed |
| F7 | Nothing graded whether the skill used the tool's *response* | 2 | fixed |
| F8 | The tool's `doubleDatedYear` had no window guard — the 25-March error in **seven places** (five in code/tests, two more in the spec, found in review) | 1 | fixed |
| F9 | `_012` is knife-edge by construction; the cause is in the optimiser-owned description | 2 | diagnosed; tagged for deterministic grading; **handback** |
| F10 | The judge **fabricates compliance** — cited two examples absent from the response it graded | 2 | recorded; caps what F5 can achieve; VR-6 proposed |
| F11 | The tool accepts impossible dates (`30 Feb` → silently 1 day wrong) | 1 | recorded; **handback** |

## F8 is the one to read

`convertCalendar` bumped the year unconditionally while its own comment stated the assumption it never checked — *"The slash already signals the Jan 1–Mar 24 boundary"* — returning a flat 1751 for 25 March, which is the one-year error F3 is about.

**Five places carried the identical error**, each locally consistent because they all agreed with each other: `convert-calendar.ts`, its vitest case (which *asserted* the bug), a second vitest case using the date incidentally, `SKILL.md`'s worked example, and `_007`'s input.

It also **hides from the snapshot**: MCP source is deliberately not tracked, but `LIVE_TOOLS` runs the *compiled* build, so a tool fix changes every future eval run while committed run logs still read as active. `eval/CLAUDE.md` names the remedy (hash the compiled artifact, as `judge_prompt_hash` does) and it is not implemented — worth a `nothing-checks` label whenever someone picks it up.

## F5 — approved and fixed, half verified. **Reviewers: this one is global.**

`eval/harness/judge/prompt.md` changes grading for **every skill**, not just this one. It was approved on 2026-08-20 and is in this PR.

**The mechanism was the prompt contradicting itself.** "Which rule wins" calls per-test notes *"binding, not advisory"* and says they *"win outright."* The section that delivers them, 120 lines later, called them **"background"** — twice — and said a bullet written as a requirement *"is still background."* The judge was obeying the **nearer** instruction.

The "background" framing had a real job (stopping invented dimensions), so the fix separates what it conflated:

- **Structure**, kept — these notes create no new dimensions. `**Do not emit separate dimensions for them.**` is preserved *verbatim*; a pinned test guards it as "#1401's only prose defence against invented dimensions", and it caught me lowercasing it.
- **Authority**, fixed — the notes bind the scores of the dimensions you were given. An unmet requirement is a deduction.
- **Facts**, new — a note asserting a fact is established for the test. Don't contradict it; don't mark the response down for stating it.

**Now evidenced by run 4 — and it splits.** `v1_2026-08-21_19-59-34` carries `judge_prompt_hash` **c39d7003**, matching this branch's `prompt.md` exactly (run 3 carried `0d186137`, which is main's).

| Arm | Result | Evidence |
|---|---|---|
| **Facts** — do not contradict an asserted fact | **PROVEN** | `_013` Correctness **1 → 3**, `fail → pass`, all seven dimensions 3. Rationale inverted from *"The skill's core claim is factually false"* to *"accurately presents the historical facts … all historically documented"*. The genealogist then confirmed 3 is correct, which is what establishes the old 1 was a judge error and not a skill failure. |
| **Authority** — an unmet requirement is a deduction | **NOT TESTABLE with this canary** | See below. An earlier revision of this body called it *disproven*; review corrected that and the correction matters more than the claim. |

**Why Authority is untestable here, and why that is the worse news.** The `_005` canary's response genuinely omits the requirement — zero occurrences of 1582, 1752, England, Catholic, "10 days", "11 days", "vary", "jurisdiction", "country". Completeness scored 3. But the judge's rationale says:

> "The skill addressed all requirements: … and **briefly explained that offsets vary by country and time period with examples (10 days for Catholic Europe 1582, 11 days for England 1752)**."

It did not weigh a known-unmet requirement and decline to deduct. It asserted the requirement **was met** and **invented the two examples**, which appear nowhere in the graded response. No precedence wording can bind a grader that believes the bullet was satisfied — so the Authority paragraph is **untested, not ineffective**, and it stays.

**That is F10, and it is the ceiling on F5's whole family of fixes.** F5 assumed the failure was *authority*; at least once the failure is *perception*. A judge that narrates compliance it never observed defeats every per-test requirement simultaneously, and no bullet wording reaches it. It also explains the original 20-of-20 pattern better than "ignored the override" did. The mechanical counterpart — a validator asserting a string the rationale claims to cite actually appears in the response — is VR-6 in the handback.

### The false-pass count, measured rather than asserted

The write-up's original "20 of 20" was wrong, and so was my first retraction of it. Classified properly across all 20 gradings:

| | count |
|---|---|
| carried the bullet | 20 |
| cited the bullet's example jurisdictions | 0 |
| conveyed variation some other way (defensible) | **5** |
| met **neither** — genuinely unmet, scored 3 | **15** |

So **15 of 20**, not 20: five were my grep testing the bullet's illustrative "e.g." examples rather than its actual requirement. Both earlier figures came from n=1 generalisations in opposite directions; the write-up records the correction and the reasoning.

Also recorded as my error: the retained canary was kept on `_005` when `_004` and `_009` were already *measured* as failing in 4/4 runs. Picking the unverified one is why the authority arm has nothing to test against. `_009` is named as the replacement for the next run.

## The suite ships at 15 pass, 1 fail

Committed run: **`v1_2026-08-21_19-59-34`**, active, annotated (51 entries over eight tests, no score overrides).

- **`_013` fail is the judge being wrong**, per F5 above. The test is already correct and emphatic; the only route to green is weakening a correct test to satisfy a judge that is not reading it — the exact failure mode this dive exists to find.
- **`_004` partial is fair.** The annotation is sharper than the judge: where the judge called it an explicitness nit, the genealogist identified that the table *encodes the central distinction incorrectly* — "before 15 October 1582 there was no contemporaneous Gregorian equivalent, so the unchanged date cannot be labeled a Gregorian date."

A green suite bought by loosening two correct tests is worth less than an amber one that discriminates. This one now catches over-conversion, the offset threshold, the boundary day, and pre-adoption over-claiming — and the new `Tool response interpretation` dimension took **three distinct values in four observations**, against a suite where 3 of 5 dimensions never took more than one.

## Validators — implemented, not requested

Following the pattern in [#1762](https://github.com/PioneerAIAcademy/cowork-genealogy/pull/1762) (the #1647 dive): validator in `validators/`, proof-of-failure tests in `tests/unit/test_convert_dates_validators.py`, imports aliased away from the `test_` prefix.

**Free by construction** — `eval/harness/validators/` and `eval/harness/tests/` sit outside the run-log snapshot, so this adds no re-run and no annotation pass. Run `v1_2026-08-21_19-59-34` is **still active** after them (verified, not assumed).

| | Rule | Landed | State |
|---|---|---|---|
| **VR-0** | register `convert_calendar` as a live tool | `mock_mcp.py`, 4 sites | prerequisite for everything below |
| **VR-1** | a `requires-tool-conversion` test must call the tool | `validators/test_convert_dates.py` | ✅ 5 tests; fires on 8 tests in the committed run |
| **VR-2** | no `doubleDatedYear` outside Jan 1 – Mar 24 | **the tool** + 5 vitest cases | ⚠️ written, tested, then **deleted** — below |
| **VR-3** | a `grade_on_invariant` validator must be able to fail | design handed over | deliberately not implemented — below |
| **VR-4** | zero calls on a `refusal-to-convert` test | `validators/test_convert_dates.py` | ✅ 3 tests |

11 proof tests, each check exercised against a state that must pass **and** the state that must fire (CLAUDE.md: "a new lint must be proven to fail"). Then replayed against all 16 tests of the committed run: **0 failures.**

**One limit, stated plainly:** that replay feeds each test's recorded `tool_calls` through the validator functions directly — it is **not** the harness running them. The committed run log was produced *before* `ad32c214b` added the validators, so its `validators.results` show only the two pre-existing checks. The first run log to show all four executing will be the next full run of this skill.

### VR-2 was implemented and then deleted — the reusable finding

Written as specified, it read `tool_calls[].args` and failed any out-of-window `doubleDatedYear` request. It passed its own unit tests. Replaying it against the committed run log then failed **`ut_convert_dates_016` — a test that had passed.**

What `_016` actually did: requested the correction for 25 March, received the guarded tool's refusal, and relayed it — *"The tool confirms this date is outside the normal double-dating window and flags it as anomalous."* Exactly what the body's `{ ok: false }` rule asks for. Reading only `args`, a validator cannot distinguish correct probing from asserting a wrong year, so it punished the best available behaviour.

**A rule the tool can enforce belongs in the tool** — where a wrong request is *answered* rather than merely recorded. A validator earns its place on what the tool cannot see: whether a call happened at all (VR-1), or whether one happened that should not have (VR-4). The guide's Step 6 table has no row for that distinction and should.

Cost of finding this out by shipping it: nothing, because validators are free. Cost of not: a permanent false failure on the one test documenting the boundary day.

### VR-3 is handed over rather than implemented, on purpose

`check_runnable` returns `not_runnable`, which aborts a test as a **corpus error**. Extracting "which tool a validator asserts about" from its source is a **heuristic** — a tool name can sit in a message string, and an assertion can be about something else. A false positive would abort a paid run on any skill.

The design in the write-up specifies it as a **warn-only lint** in `harness/scripts/`, beside `check_tool_coverage.py` and `check_rubric_tool_drift.py`, which is where this repo already puts approximate cross-skill checks — not as an extension of the blocking gate whose comment already names this failure mode one step short.

The specific instance is closed either way: `convert_calendar` is registered, so `test_no_spurious_conversion` can now fail, and a test proves it does.

## A regression I introduced and fixed

The replacement double-date example was "15 February 1749/50" — the exact string in `_012`'s input, which asks a *schema-validation* question. `_012` failed run 1 by grabbing it. The collision predates the dive (the description advertises `double-dated years (e.g. "1749/50")` alongside `date schema validation (use validate-schema)`, and `_012` had already failed in `v1_2026-07-22`) but duplicating the string into the body made it worse. Changed to "6 January 1745/6" — verified unused across the corpus and valid against the live tool. `_012` passes in run 3 and fails in run 4 — see F9: the flip is *activation*, not grading.

## This review round (EdmondOware + Gennecis)

Every checkable claim was verified rather than taken on trust; two changed my conclusions.

**Fixed here, all free (nothing flips the snapshot — verified, run 4 stays active):**

| Item | Change |
|---|---|
| Body/write-up pointed at run 3 | Repointed to run 4 throughout; the stale "neither arm is evidenced" withdrawn |
| `check_tool_coverage.py` | `convert_calendar` added to `EXEMPT_TOOLS`. The warning was true-and-unactioned for two months, and after the live registration it asserted something **false** about a tool 11 of 16 tests exercise. Now silent for `convert-dates` **and** `conflict-resolution` |
| `convert-calendar.ts:119` | The comment asserting the slash *proved* the window — the sentence the write-up quotes as F8's cause — refreshed |
| `convert-calendar-tool-spec.md` | Four sites updated for F8's contract change (§4.1 precondition, §5b example, §7 error row, §8 test plan). **F8's count corrected 5 → 7**: §5b and §8 would have re-seeded it. §7's row is scoped to the **English Lady Day convention**, with Venice / Byzantine / *mos gallicanus* named as out-of-scope rather than implying a universal rule |
| `_012` tagged `no-spurious-conversion` | Free — tags are stripped before hashing. Next run grades its spurious call as a deterministic validator failure instead of a judge narration |

**Two corrections to reviewers, both load-bearing:**

1. **`_012`'s run-4 failure is not the F5 fix.** EdmondOware proposed that and asked to be confirmed or corrected. Corrected: `_compute_outcome`'s negative branch decides a routing negative on `if activated: return "fail"` — judge dimensions do not gate it. Run 3 made no tool call (didn't activate → pass); run 4 called `convert_calendar` (activated → fail). The judge prompt cannot reach that outcome. Gennecis identified the same mechanism independently.
2. **"Authority arm — DISPROVEN" was wrong**, per Gennecis. The judge fabricated compliance rather than declining to deduct, so the arm was never exercised. Restated, and F10 recorded as the more valuable finding.

**Deferred with reasons, in the write-up's Handback section:** `SKILL.md`'s 1582 worked example (one line, but flips the snapshot — reviewer's explicit instruction was to bundle it with the next run rather than buy a fifth); the description collision behind F9; F11; VR-3; VR-6; and two annotation cells (`_005`, `_012`).

## Verification

- Full vitest: **102 files, 2443 tests** green
- `eval/harness` unit suite: **2386 passed**, 3 skipped
- 16/16 `check_runnable`; run 3 snapshot **ACTIVE** after merging `main`
- 13 scratch runs (~$1.15) plus three full runs; run 2 excluded as environment-contaminated — two judge API timeouts (108s and 56s against a 4–14s norm in the same run, zero dimensions returned on healthy 1034- and 1185-character responses)
- Citations are grep anchors, not line numbers: 147 commits moved one of four in three weeks

**Also recorded as a harness reporting gap:** a judge-side timeout produces a bare `fail`, indistinguishable in the summary table from the skill getting the answer wrong, while skill-side aborts print their reason. That turned an API outage into what read as two regressions.

No issues filed, per standing instruction — the harness and judge-prompt items are written as requests to hand back.


mercyokum (commented):
Everythign looks good genealogically but one thing needs fixing before I approve.

The "Validators — implemented, not requested" section of the PR body doesn't match the diff. It claims VR-1 and VR-4 landed in validators/test_convert_dates.py (✅ 5 tests, ✅ 3 tests) plus 11 proof tests in tests/unit/test_convert_dates_validators.py. Neither is true:

validators/test_convert_dates.py is byte-for-byte unchanged from main.
tests/unit/test_convert_dates_validators.py doesn't exist anywhere in the repo.
The committed run log's validator results only show the two pre-existing checks running, not the 8 the body describes.

Your own write-up already gets this right — it calls the section "Validator requests" and its "Changes in this PR" table only lists VR-0 (the tool registration) as landed. The PR body just needs to match that.

Could you either:

Actually add the VR-1 and VR-4 validator functions (you've already spec'd them as four-line functions, same shape as test_no_spurious_conversion), or
Edit the PR body so VR-1/VR-4 read as proposed follow-on work instead of landed work, consistent with the write-up.

Emruthwill:
Thanks — the third point is a real catch and I've fixed it. The first two don't hold, and I think I can explain why you saw them.

**The validators are on the branch.** Verifiable without taking my word:

```sh
git fetch origin
git ls-tree origin/1654-deep-dive-convert-dates -- eval/harness/tests/unit/test_convert_dates_validators.py
# 100644 blob bf06721a8755710a10ce94b916301c18579ab8aa

git diff --stat origin/main origin/1654-deep-dive-convert-dates -- eval/harness/validators/test_convert_dates.py
# 1 file changed, 90 insertions(+)

git show origin/1654-deep-dive-convert-dates:eval/harness/validators/test_convert_dates.py | grep '^def '
# ... test_requires_tool_conversion_calls_the_tool   (VR-1)
# ... test_refusal_to_convert_makes_no_call          (VR-4)
```

Both landed in `ad32c214b`, which was pushed **after** the first push of this branch (`06534ac3f..ad32c214b`). If you opened the diff between those two pushes you'd have seen exactly what you describe — no validators, and the write-up's older heading of "Validator requests". The write-up on the branch now reads `## Step 6 — Validators, implemented` and its Changes table carries two validator rows. So the body and the write-up do agree; you were reading an earlier state of both. A hard refresh should show it.

Counts check out too: 11 tests in the new file, 5 on VR-1 and 3 on VR-4, matching the body.

**Your third point is correct, and it was overstated on my side.** The committed run log's `validators.results` really do show only `test_only_convert_calendar_called` and `test_no_spurious_conversion` — because `v1_2026-08-19_22-13-16` was produced *before* `ad32c214b`. So the new validators have never executed inside a harness run. What they have is 11 unit tests, plus a replay that feeds each test's recorded `tool_calls` through the functions directly. That is weaker than an in-run result and the body shouldn't have let "0 failures on replay" read as "verified in-run".

Fixed in `9f0140cd6` — both the PR body and the write-up now state the limit explicitly, and name the next full run of this skill as the first log that will show all four executing.

On whether to get that log now: it would cost a full run plus another annotation pass, and I'd rather bundle it than spend it on this alone. Three things are already queued for the next run for other reasons — moving the F5 canary to `_009` (where it can actually discriminate), a Swedish carve-out in `Tool response interpretation`, and sharpening `_004` per the annotation. Happy to do it now instead if you'd rather approve on an in-run result; your call.

mercyokum (approved):

Asimi1234:
@Emruthwill went through this at `c3bf6b97` (full pass, up to date with main). No blocking correctness findings — this is a strong, unusually well-evidenced dive.

Verified everything by running it in a fresh worktree:
- vitest: 104 files / 2467 tests pass, `tsc` clean
- harness unit suite: 2472 passed, 2 skipped
- VR-1/VR-4 each have a `pytest.raises` fire-state case next to the pass case, so they're genuinely falsifiable and not can't-fail checks
- F8 guard reads correctly (only refuses when the date proves it's outside Jan 1–Mar 24), and I confirmed F2 is a skill-table prose fix only — the tool already computes the offset era-correctly via the JDN round-trip
- all required CI green on this head

One thing I want a senior to consciously sign off on rather than treat as routine: the `eval/harness/judge/prompt.md` change is global — it shifts grading doctrine for every skill (per-test notes now binding, an unmet requirement becomes a deduction). Your write-up is honest that the authority arm is unverified-but-fail-safe while the facts arm is proven (`_013` going 1→3), which I appreciate. Just flagging that the next run of any skill could see some passes drop to partial, so whoever approves should accept that corpus-wide risk knowingly (or gate on a run that actually exercises the authority arm).

Two non-blocking notes:
1. `convert-calendar.ts:119–120` — the comment still says "the slash already signals the Jan 1–Mar 24 boundary," which is the exact assumption your new guard below refutes. Worth a one-line refresh so it doesn't mislead the next reader.
2. The new validators haven't executed inside a real harness run yet (proven by unit tests + replay only) — you already document this, and bundling the in-run proof into the next queued run rather than spending one now seems right to me.

mercyokum's earlier round is resolved and superseded by their approval at head, so nothing outstanding there.

EdmondOware (changes requested):
Strong dive, and F8 is a genuinely valuable catch. Requesting changes on one thing, plus a now free cleanup.

The committed run log cannot evidence the judge prompt change this PR ships. I compared hashes directly. The run log's judge_prompt_hash is 0d186137, which is byte identical to main's prompt.md. This PR's prompt.md hashes to c39d7003. So v1_2026-08-19_22-13-16 was graded entirely under the old prompt, before the F5 fix.

That makes two statements in the body contradict each other. The F5 table claims the Facts arm is proven via _013 Correctness 1 to 3 with a new rationale. The committed log shows _013 Correctness still 1, carrying the old rationale verbatim, the skill's core claim is factually false. Both cannot describe the same run, so the 1 to 3 must come from an uncommitted scratch run.

That matters more than usual here because prompt.md is global, it changes grading for every skill. The only stated proof for a corpus wide change is not reproducible from anything in the repo, and the runlogs job passes because the hash rule is warn only by design, so nothing mechanical catches it. Please either commit a run graded under the new prompt, or restate both F5 arms as unverified in repo, one proven and one not is not the honest split right now.

Free cleanup, and your own condition for it is now met. t docs/skill-dives/convert-dates.md on the reasoning thatdocs/deep-dives exists only on an unmerged branch. 1745 merged, and docs/deep-dives now holds six files from three dives, citation,
init-project and search-images. You wrote whoever merges once the target is real. It is real. This is a one filemove and it avoids leaving a second competing directory behind.

What I verified and it all holds:

F1 is genuinely fixed. convert_calendar is now called 11 times across 11 tests, and Tool Arguments went from all null to 11 scored and 5 null.
The diagnosis that the defect switched off the dimension real and now closed.

F8 is correct, and the genealogy is right. March 25 is Lyear increments, so from that date both styles agree andbumping introduces a one year error. I confirmed the old vitest case literally asserted the bug, expecting year 1751 for month 3 day 25, and
the new guard is conservatively scoped, refusing only whof window and merely noting a March date with no day.

The F5 contradiction diagnosis is real. Which rule wins ding, not advisory, and says it wins outright, while theold Per test context section 120 lines later called the same notes background twice. The judge was obeying the nearer text.

The invented dimensions safeguard is preserved verbatim, and test_judge.py pins that exact string, so it is not self reported.

The validators flagged as missing in the earlier review now exist, VR-1 and VR-4 plus 11 proof tests. Annotation is complete, 31 entries, zero missing comments on non passes.

EdmondOware (changes requested):
Both things I raised are fixed, and the way you fixed the first one is better than what I
asked for.

I asked for either a run graded under the new prompt or an honest restatement. You did both.
`e4627c67` restated the claim and admitted the scratch run plainly, then `bb0c4e01` went and
produced the evidence. I checked the hash rather than taking it: run
`v1_2026-08-21_19-59-34` carries `judge_prompt_hash` `c39d7003`, byte identical to this
branch's `prompt.md`, and `_013` Correctness is now 3 on a genuinely new rationale about the
Swedish 30 February 1712 date rather than the old one reworded. The run is active, 19
snapshot entries, zero mismatches. The write up move into `docs/deep-dives` is done.

The genealogy in the F8 offset fix continues to hold up. Anchoring the increment on 1 March
Julian rather than on the year number is right, and the worked counterexample of Julian 14
February 1900 taking +12 is what makes the rule usable at the point of decision.

What follows is one problem with four faces. The body was last updated at 20:53, the new run
landed at 21:17 and its annotation at 21:27, so the body still describes the branch as it was
before the run that answers my review.

### 1. Point the body at run 4, and withdraw the claim that neither arm is evidenced

These lines are now false against your own branch:

- "Neither arm is evidenced by anything in this branch. Corrected after review." The Facts
  arm **is** evidenced now. Your own commit message on `bb0c4e01` says so.
- "The suite is amber on purpose, 14 pass, 1 partial, 1 fail" and "Committed run:
  `v1_2026-08-19_22-13-16`, active, annotated (31 entries)". The active run is
  `v1_2026-08-21_19-59-34`, 15 pass and 1 fail, annotated with 41 entries over six tests.
- "`_013` fail is the judge being wrong, per F5 above." `_013` passes in run 4.
- In the validators section, "Run `v1_2026-08-19_22-13-16` is still active after them."

The body currently understates the PR. Fix the pointers and let it claim what it earned.

### 2. State a position on `_012`, which fails in the run you are shipping

"A regression I introduced and fixed" ends "`_012` passes in run 3." It does. It also fails
in run 4, which is the run the gate treats as active, and that section is the only place a
reader is told anything about it.

My reading, which you should confirm or correct rather than adopt: this is the F5 fix
working, not a new defect. The judge's rationale says the per test context "explicitly states
this is a NEGATIVE test where convert-dates should NOT perform its task". Under the old
prompt that context was advisory background, so a polished substantive answer passed. Now it
binds and the negative test is graded as written. If that is right, `_012` failing is the
Authority arm demonstrating itself on a second test, which is worth more to the write up than
the pass was.

### 3. Annotate `_012`, or say why not

The annotation covers `_004`, `_006`, `_007`, `_008`, `_009` and `_013`, six tests where five
were owed. `_012` is the only failure in the shipped run and it is the one test not
annotated. It is not in `review_sample`, so no gate is broken and I am not claiming one is.
But the single failing test in a run being the one with no human judgment recorded on it is
the wrong shape for a dive whose whole subject is the judge.

### 4. File the description collision rather than fixing it here

The cause of `_012` is still in place. The description advertises `double-dated years (e.g.
"1749/50")` as a trigger, two clauses before it sends `date schema validation (use
validate-schema)` away, and `_012` asks "Is '15 Feb 1749/50' a valid value for a date field
in research.json?". You changed the body example to "6 January 1745/6" and left `1749/50` in
the description, which is where the collision actually lives.

Do not fix it in this PR. A description edit invalidates the run log and pulls in the routing
pairs, and this PR has paid for enough runs already. File it, or add it to the write up
handback with the run 4 evidence attached.

Gennecis (changes requested):
Reviewed the full branch at `755c9afa`, both committed run logs, the annotations, and the harness code the findings rest on.

The dive is solid work. F1 is a real and expensive defect, and F8 is a genuine one-year genealogical error in shipped tool code. I re-derived the calendar arithmetic independently rather than reading it off the tables, and it holds:

- Julian 28 Feb 1700 → Gregorian 10 Mar (+10); Julian 1 Mar 1700 → Gregorian 12 Mar (+11); Julian 14 Feb 1900 → +12. The rewritten offset paragraph in `SKILL.md` and the tool's JDN round-trip agree.
- 25 March is Lady Day, the day the Old-Style year increments, so both styles agree from that date and a slash written there is anomalous. The guard is correctly scoped — it refuses only when the date proves it is outside the window, and notes rather than refuses a March date with no day.
- Sweden: Julian 1712-02-29 → Gregorian 1712-03-11, matching the table's chain.
- The 11 proof tests each carry a `pytest.raises` fire-state, so the new checks satisfy CLAUDE.md's "a new lint must be proven to fail."
- The deleted June run log is the harness's own 5-candidate retention prune, not a hand deletion.

Five changes requested. Four are free. Only the first costs an eval run, and it should ride along with the next run this skill takes rather than buying its own.

---

## 1. `packages/engine/plugin/skills/convert-dates/SKILL.md:84-85` — the worked example is a conversion the tool now refuses

The body's only end-to-end example still reads:

> Example: "14 September 1582 (Julian) +10 days = **24 September 1582 (Gregorian)**"

That is the exact input of `ut_convert_dates_004`, and three things in this PR now contradict it:

- `convert-calendar.ts` returns `ok: false` for `julianToGregorianDay` before 1582-10-15, so the tool cannot produce that line.
- `catholic-europe-1582.json`'s new `judge_context` says presenting "24 September 1582" flatly "is the failure this test looks for."
- The write-up's F6 row records the effect: "the skill's *first* instinct was '+10 days' — the tool corrected it." That first instinct is the body's example.

This is the same shape as F8 — several sites locally consistent because they agree with each other — with one site left in.

**Change:** use a post-adoption date already in the corpus, e.g. `_009`'s "15 January 1699 (Julian) +10 days = 25 January 1699 (Gregorian)".

This flips the snapshot, so please bundle it with the next run this skill takes for any other reason rather than spending a run on one line.

## 2. `docs/specs/convert-calendar-tool-spec.md` — not updated for the F8 behaviour change (free)

F8 changed the tool's contract. The spec still states the pre-F8 behaviour in four places. CLAUDE.md: "Specs are the source of truth an implementation is checked against", and "a live tool must have a live spec."

- **§4.1** — "Given `year` and `doubleYear`, the New-Style year is the **later** of the two … Sets `converted.year`." Add the Jan 1 – Mar 24 precondition.
- **§5b (line ~169)** — "Asking for the New-Style **year** of '25 March 1750/1' → `{ doubleDatedYear: true }`." That call is now an input error. Use an in-window date.
- **§7 error table** — no row for the new condition. Add one beside the `doubleYear`-inconsistent row.
- **§8 test plan (line ~213)** — "Single-correction discipline — requesting only `doubleDatedYear` on `25 Mar 1750/1`." The vitest case moved to `15 Feb 1750/1` in this PR; the spec still names the old date.

Two consequences worth folding into the write-up:

- F8's "five places carried the identical error" is really **seven**. §5b and §8 are the two that would have re-seeded it.
- Please phrase the new §7 row as *the English Lady Day convention* rather than as a universal rule. The guard refuses any `month > 3`, which is right for England, Wales, Ireland, the colonies, Florence and Pisa, but year-start conventions differed elsewhere — Venice 1 March, the Byzantine and pre-1700 Russian 1 September, the French *mos gallicanus* Easter start — where a slashed year outside Jan–Mar can be legitimate. Nothing in the corpus exercises those, so this is scope rather than a bug, but the error string currently reads as a universal claim.

## 3. `eval/harness/scripts/check_tool_coverage.py` — `convert_calendar` is a LIVE_TOOL but is not in `EXEMPT_TOOLS` (free)

Ran the script against the current tree:

```
::warning file=packages/engine/plugin/skills/convert-dates/SKILL.md::skill `convert-dates` declares
allowed-tools ['convert_calendar'] but its test corpus has no fixture for: ['convert_calendar'].
No eval test can exercise this tool — add a test with an mcp_fixture for each, or drop the tool
from allowed-tools.
```

Same warning for `conflict-resolution`, which also declares the tool (`conflict-resolution/SKILL.md:8`).

The PR body identifies this warning's remedy as the wrong one. The right remedy landed, but the warning still fires and now states something false — 11 of the 16 tests in run 4 exercised the tool live.

`EXEMPT_TOOLS` (around line 61) already documents LIVE_TOOL members in this exact shape: see `research_append`, `research_log_append`, `research_query`. Adding `convert_calendar` with a one-sentence reason closes it for both skills.

## 4. The PR body and the write-up still describe run 3 (free)

Run 4 (`v1_2026-08-21_19-59-34`, 15 pass / 1 fail, 41 annotation entries) is the active run. These lines are now false against the branch:

- Body: "Neither arm is evidenced by anything in this branch." The Facts arm is evidenced. I checked the hashes rather than the table — run 3 is `0d186137` (main's `prompt.md`), run 4 is `c39d7003` (this branch's), and `_013` Correctness moves 1 → 3 on a genuinely new rationale.
- Body: "The suite is amber on purpose — 14 pass, 1 partial, 1 fail"; "Committed run: `v1_2026-08-19_22-13-16`, active, annotated (31 entries)"; "Run `v1_2026-08-19_22-13-16` is still active after them"; "`_012` passes in run 3."
- Write-up: the `## The three full runs` table, the heading `### Why run 3 is the one to commit, at 14/16`, its "`_013` fail — the judge is wrong" bullet, and the two items listed as candidates for the next run — the Swedish carve-out and the `_004` Gregorian-row ban — which have both since landed.

The body currently understates the PR. Point it at run 4.

## 5. `packages/engine/mcp-server/src/tools/convert-calendar.ts:119-120` (free)

```ts
// 1. Double-dated year → the later (New Style) year. The slash already signals
//    the Jan 1–Mar 24 boundary, so the New-Style year is always +1.
```

This states the assumption the guard eleven lines below refutes, and it is the sentence the write-up quotes as F8's cause. One-line refresh.

---

## The `eval/harness/judge/prompt.md` change — approved, with one claim to correct

This file is global: it changes grading for every skill in the corpus, not just this one. Taking it deliberately.

**Approved.** The contradiction is real — "Which rule wins" §3 calls per-test notes "binding, not advisory" and says they "win outright", while the section that delivers them 120 lines later called them "background" twice. Leaving that in place is not an option, and the fix separates structure, authority and facts cleanly, preserving the invented-dimensions safeguard verbatim.

The **Facts** arm is the one with real reach: it makes a factual assertion in any test's `judge_context` unchallengeable for Correctness. I am accepting that knowingly. It narrows a discretion the prompt already granted — "If an override seems wrong to you, follow it and say so in your rationale" — and `_013` shows the cost of not having it, where the judge called a documented Swedish date "factually false" and failed the test for stating it correctly.

**One claim needs correcting: "Authority arm — DISPROVEN".** I pulled `_005`'s response text out of run 4 and counted markers:

```
1582:0  1752:0  England:0  Catholic:0  "10 days":0  "11 days":0
vary:0  varies:0  jurisdiction:0  country:0  countries:0
```

The bullet is genuinely unmet — that measurement is correct. But the judge's Completeness rationale reads:

> "The skill addressed all requirements: … and **briefly explained that offsets vary by country**"

The judge did not decline to deduct for a requirement it knew was unmet. It asserted the requirement **was** met. No wording in "Which rule wins" can bind a grader that believes the bullet was satisfied, so `_005` does not test the Authority arm — what it measured is the judge inventing compliance in its rationale.

So: keep the Authority paragraph, and restate the result as **not yet testable with this canary — the observed failure is a different mechanism the prompt does not address**. "Billed prose that changes nothing" is a stronger conclusion than one observation supports.

The fabricated-compliance finding is the more valuable one and belongs in the write-up: a judge that invents compliance defeats every per-test requirement at once, not just this bullet.

## `_012` in run 4 — this is F9, not the prompt change

Worth stating explicitly in the write-up, because the timing invites the other reading. `_compute_outcome` (`eval/harness/harness/orchestrator.py`, negative branch around line 1215) decides a negative test with a non-empty `correct_skill` **purely on `activated`**. The judge's dimensions do not gate it.

- Run 3 `_012`: `tool_calls: []`, Tool Arguments `null` → did not activate → **pass**
- Run 4 `_012`: called `convert_calendar {doubleDatedYear: true}` → activated → **fail**

That is a behaviour flip — the F9 knife-edge, pass/fail/pass/pass/fail across five runs — and the judge prompt has no influence on `activated`. Filing the description collision rather than fixing it here is the right call, since a description edit invalidates the run log and pulls in the routing pairs.

---

## Non-blocking

- **`_012` can be tagged `no-spurious-conversion` for free.** `snapshot.py::_COSMETIC_TEST_FIELDS = ("name", "description", "tags")` — tags are stripped before hashing, so adding one does not flip the run log. `test_no_spurious_conversion` gates on that tag and not on test type, so it would turn run 4's `_012` from a judge-narrated fail into a deterministic validator failure. That matches the dive's own thesis at zero cost.
- **Annotation coverage.** `review_sample` is {006, 007, 008, 009, 013}; the annotation covers those plus `_004`. Two cells the conclusions rest on carry no human judgment: `_012`, the only failure in the shipped run, and `_005`, the Authority canary. `_005` is the one I would want annotated, since a corpus-wide prompt decision rests on it.
- **Cross-skill reach of VR-0.** `conflict-resolution/SKILL.md:8` declares `convert_calendar` too, so registering it live changes that skill's next run as well, while its committed run logs stay active because `LIVE_TOOLS` is not snapshot-tracked. Net positive — worth one line in the handback beside the "hides from the snapshot" note already written for the tool fix.
- **The tool accepts impossible day/month combinations.** `convertCalendar` validates `day` as 1–31 only, so `{year: 1712, month: 2, day: 30}` is accepted and silently normalises to 1 March, returning Gregorian **12** March 1712 instead of the correct **11** March. `_013` is the test where a model might pass the Swedish 30 February straight through; in run 4 it correctly sent 29 February. Pre-existing and out of scope here, but this is the corpus's only impossible-date test and the dive is the natural place to record it.

Items 2–5 are free and can land now. Item 1 waits for the next run this skill takes — tell me which run that will be and I will re-review then.

EdmondOware (approved):
Approving. All four of my items are addressed, and the correction back to me is right.

I verified your `_012` mechanism rather than accepting it. `orchestrator.py`'s negative branch
returns `fail` on `if activated` before any judge dimension is consulted, and `correct_skill`
for `_012` is `["validate-schema"]`, non empty, so it takes that path. Run 3 makes zero
`convert_calendar` calls and passes, run 4 makes one and fails. My reading was wrong: the
judge prompt cannot reach that outcome, and the rationale I quoted is narration downstream of
an outcome already decided. F9's knife edge, as you and Gennecis both had it.

F10 is the more valuable thing to come out of this round. A judge that asserts a requirement
was met and invents the two examples it claims to cite is a different failure from a judge
that ignores an override, and it explains the original 20 of 20 better than the old reading
did. Restating the Authority paragraph as untested rather than ineffective is the honest
position, and VR-6 is the right shape of follow up.

Also verified: run 4 is still active after `b41e0b81` and the two commits following it, 19
snapshot entries and zero mismatches. The annotation now covers eight tests where five were
owed, `_005` and `_012` included.

One line left, and not worth another round. In the Validators section, "Run
`v1_2026-08-19_22-13-16` is still active after them" still names run 3. What it claims is
true, validators sit outside the snapshot and cost no rerun, but the run it names is no longer
the active one. Fix it whenever you next touch the body.

Asimi1234 (approved):
Approving. Verified every reviewer point against the branch at `52809a65` in a fresh worktree, not from the PR body's claims.

**All reviewer comments met:**
- **EdmondOware (4):** body/write-up repointed to run 4; `_012` position stated (activation, not grading — confirmed via `orchestrator.py`'s negative branch); `_012` annotated (3 entries); description collision handed back with run-4 evidence.
- **Gennecis (5):** spec §4.1/§5b/§7/§8 synced with the Lady Day scope note; `convert_calendar` added to `EXEMPT_TOOLS` (coverage script now clean for both convert-dates and conflict-resolution); `convert-calendar.ts:119-120` comment refreshed; run-4 pointers; the 1582 SKILL.md example deliberately deferred to the write-up Handback per Gennecis's own instruction (it flips the snapshot).
- **mercyokum:** already resolved and approved.

Just corrected two stale run-3 pointers in the body (`51/eight` annotation count; the validators 'still active' line now names run 4).

**One gate remains before merge:** @Gennecis's changes-requested is still standing. Their points are all addressed above, but this needs their re-review to clear the block.

Asimi1234 (approved):
Approving. Verified every reviewer point against the branch at `52809a65` in a fresh worktree, not from the PR body's claims.

**All reviewer comments met:**
- **EdmondOware (4):** body/write-up repointed to run 4; `_012` position stated (activation, not grading — confirmed via `orchestrator.py`'s negative branch); `_012` annotated (3 entries); description collision handed back with run-4 evidence.
- **Gennecis (5):** spec §4.1/§5b/§7/§8 synced with the Lady Day scope note; `convert_calendar` added to `EXEMPT_TOOLS` (coverage script now clean for both convert-dates and conflict-resolution); `convert-calendar.ts:119-120` comment refreshed; run-4 pointers; the 1582 SKILL.md example deliberately deferred to the write-up Handback per Gennecis's own instruction (it flips the snapshot).
- **mercyokum:** already resolved and approved.

Just corrected two stale run-3 pointers in the body (`51/eight` annotation count; the validators 'still active' line now names run 4).

**One gate remains before merge:** @Gennecis's changes-requested is still standing. Their points are all addressed above, but this needs their re-review to clear the block.

Gennecis:
Re-reviewed at `d575a039`. All five items are addressed. I checked each one against the branch rather than the summary.

| Item | Verified |
|---|---|
| Spec sync | All four sites. §4.1 carries the Jan 1 – Mar 24 precondition; §5b's example is `15 February 1750/1` with the reason attached; §7 has the new error row; §8's discipline bullet moved and the window bullet is new. F8's count corrected 5 → 7. |
| Lady Day scoping | §7 names the convention and lists Venice, the Byzantine and pre-1700 Russian 1 September, and *mos gallicanus* as out of scope. |
| `EXEMPT_TOOLS` | `convert_calendar` added. Re-ran the script with the exemption applied — `convert-dates` is silent. |
| Run-4 pointers | Body reads 15 pass / 1 fail, `v1_2026-08-21_19-59-34` active, 51 entries over eight tests; the validators line names run 4; the three-run table and "why run 3" are gone. |
| `convert-calendar.ts:119` | Rewritten, and it now records why the old wording mattered. |
| 1582 example | Correctly deferred. `SKILL.md` is byte-unchanged and the item is first in the Handback, marked "do this first on the next run." |

Both restatements landed, and F10 holds up. I checked the exact string: the rationale claims "with examples (10 days for Catholic Europe 1582, 11 days for England 1752)", and the graded response contains zero occurrences of `1582`, `1752`, `Catholic`, `England`, `10 days` or `11 days`. The judge did not only assert compliance — it invented the citation. That is a sharper finding than the one I raised, and VR-6 is the right shape for it.

Also verified: the branch's delta against current main is exactly this PR's own files — `judge/prompt.md`, `rubric.md`, the validators and the vitest file are untouched since the last round. I recomputed run 4's snapshot independently: 19 entries, zero mismatches, still active. CI green.

## Two corrections, both one line, neither blocking

**1. `conflict-resolution` is not clean.** The `EXEMPT_TOOLS` entry says "Exempting it also clears conflict-resolution, which declares the same tool." Running the script with the exemption applied:

```
::warning file=packages/engine/plugin/skills/conflict-resolution/SKILL.md::skill
`conflict-resolution` declares allowed-tools [...] but its test corpus has no fixture
for: ['place_search_all'].
```

The exemption removes `convert_calendar` from that skill's missing list; the skill still warns, for an unrelated pre-existing gap. Worth narrowing the sentence, since that docstring will outlive the PR.

**2. The write-up now understates the result.** The Validators section still says the replay is "not verified in-run" and that "the first run log to show all four executing will be the next full run of this skill." Run 4 is that run, and it is committed. From its `validators.results`:

```
test_requires_tool_conversion_calls_the_tool   ran=8   skipped=8   failed=0
test_refusal_to_convert_makes_no_call          ran=1   skipped=15  failed=0
test_no_spurious_conversion                    ran=2   skipped=14  failed=0
test_only_convert_calendar_called              ran=12  skipped=4   failed=0
```

VR-1 fired on exactly the eight tagged tests, and all four executed with zero failures. The stated limitation no longer applies.

One precision point for the same edit: `test_no_spurious_conversion` fires on two tests in run 4, not three. `_012`'s new `no-spurious-conversion` tag was added after that run, so it takes effect on the next one.

Fix both whenever you next touch those lines. Clearing my block.

Emruthwill:
Both corrections landed in `ff00dc025`. Verified each against the branch first — you were right on both.

## 1. `conflict-resolution` is not clean — docstring narrowed

Confirmed by re-running the script with the exemption applied:

```
::warning ... `conflict-resolution` declares allowed-tools [...] but its test corpus
has no fixture for: ['place_search_all'].
```

My sentence claimed the exemption "clears conflict-resolution". It doesn't — it drops `convert_calendar` from that skill's missing list and leaves an unrelated pre-existing gap warning. Reworded to say exactly that, and I named `place_search_all` explicitly so the sentence can't quietly rot the way the original would have. Your point about the docstring outliving the PR is the reason that mattered.

## 2. The write-up understated its own result — fixed

You're right that run 4 *is* the run the old text was waiting for. I'd left a limitation in place that the branch had already overtaken. Reproduced your counts from `validators.results`:

| validator | ran | skipped | failed | fired on |
|---|---|---|---|---|
| `test_requires_tool_conversion_calls_the_tool` | **8** | 8 | 0 | `_001 _002 _005 _006 _009 _013 _014 _015` |
| `test_only_convert_calendar_called` | 12 | 4 | 0 | every positive |
| `test_no_spurious_conversion` | **2** | 14 | 0 | `_003 _010` |
| `test_refusal_to_convert_makes_no_call` | **1** | 15 | 0 | `_008` |

Worth stating what that last column shows beyond "it ran": VR-1 fired on **exactly** the eight tagged tests, so the tag gate resolves to the set it was written for rather than approximately. Both the paragraph and the changes-table cell now read "verified in-run".

**Your precision point is kept in the text**, not just the commit: `test_no_spurious_conversion` fires on **two** tests, not three, because `_012`'s tag postdates run 4. It takes effect next run — which is when `_012`'s spurious call becomes a deterministic validator failure instead of a judge narration.

## On F10

Thank you for checking the exact string rather than my table. Since your review, `_005` was annotated with a comment on all seven dimensions, and it sharpened the finding once more: the genealogist confirmed Completeness **3 → 3**, and the comment enumerates what the response covered — "original and converted dates, the 13-day calculation, Russia's 1918 calendar transition, and the reason for the offset" — with **the fabricated item absent from that list**.

So the final shape is narrower than my "false pass" and stronger than "unadjudicated": **the score is independently correct, and the stated justification for it describes content that was never written.** Those being compatible is what makes the class dangerous — a fabricated rationale attached to a correct score is invisible to anything except someone re-reading the response against it. VR-6 is requested on that basis, with a genealogist's independent reading behind it.

---

Nothing outstanding from your review. `EdmondOware`'s block is the one still open — his four items were addressed in `b41e0b81d`, including the correction that `_012`'s run-4 failure is activation rather than the F5 fix (`_compute_outcome`'s negative branch decides on `activated` alone), which you'd identified independently.

EdmondOware:
@Gennecis this is waiting on you. Your round is the only thing blocking it: three approvals are standing, no threads are unresolved, and every check is green.

Your review read `755c9afa`. Two pushes have landed since, and the newest one, `ff00dc02`, goes at your items 3 and 4 directly. It touches only `check_tool_coverage.py` and the findings doc, which is why nothing bought another run.

The `check_tool_coverage.py` change is worth the thirty seconds on its own. It narrows the exemption docstring so it no longer claims to clear `conflict-resolution`, and says instead that the skill still warns for `place_search_all`, a separate gap that was already there. That is a more honest claim than the one it replaces, and it came from your item.

No rush from me. Flagging it only because nobody else can clear the round and the PR reads as blocked when it is really just waiting.

Gennecis (approved):

PR #1768, opened 2026-08-19
fix: Windows portability — os.killpg guard, feedback paths, shebang, Makefile expansions

### LINKED ISSUE #1420: Windows portability: forbid syncing folders (OneDrive) in the docs, fix only the real Windows bugs

yinkid28:
## Summary

Closes #1420 (absorbs #1713, #1707, #1348, #1370).

**Bucket A — documentation & OneDrive guards:**
- Syncing-folder warning in `DEVELOPMENT.md` § "On Windows" and `eval/README.md` § "Windows users"
- OneDrive detection guard in `eval/Setup.bat` and `eval/Start.bat` — warns when `%CD%` is under `%OneDrive%`
- Documented the no-Windows-CI decision in `scripts/windows/README.md`

**Bucket B — code fixes:**
- **#1713** — `LocalProvider._kill_server` uses `os.killpg`/`os.getpgid` which don't exist on Windows. Guarded with `hasattr(os, "killpg")` so `proc.terminate()` fallback runs. Fixes 8 failures + 2 indefinite hangs in `apps/server` pytest on Windows.
- **#1707** — `readSessionLog` derived the Claude Code project-hash directory with POSIX-only regex (`replace(/^\//, '')`), producing an invalid path on Windows. Replaced with `replace(/[^a-zA-Z0-9-]/g, '-')` which matches Claude Code's actual naming (verified against 25+ dirs in `~/.claude/projects/` on a Windows machine). Same derivation applied in `feedback.test.ts`.
- **#1348** — Two Makefile recipes (`e2e-validate`, `optimize-skill`) used shell `${VAR:-default}` which Windows shells don't understand. Replaced with make-level `$(if $(VAR),$(VAR),default)`. Left `server-dev` and `agent-smoke` alone per the issue (they embed `grep|cut` pipelines).
- **#1370** — Removed vestigial `#!/usr/bin/env node` shebang from `gen-enums.mjs`. The file is only invoked as `node scripts/gen-enums.mjs`, never executed directly. Linux vitest strips the shebang on import; Windows vitest does not, causing `SyntaxError`.

## Test plan

- [x] `cd apps/server && uv run pytest tests/ -q --ignore=tests/test_v1_api.py` — 192 passed
- [x] `cd apps/server && uv run pytest tests/test_v1_api.py -q` — 9/10 passed (1 timeout vs. indefinite hang before fix)
- [x] `cd eval/harness && uv run pytest tests/unit/test_cli.py -q` — 66/66 passed
- [x] `scripts\windows\test-all.bat` on a Windows machine — vitest 2477 passed, server pytest 192 passed, eval harness 2521 passed
- [ ] `make test-all` on macOS/Linux (no regression)


T-FEH (approved):
Approving the code. One thing is outstanding that I cannot close from here — see the end.

**The #1707 fix is bigger than the PR claims, and worth restating.** This is not a Windows-portability change; it is a latent cross-platform bug fix. Claude Code replaces **every** non-alphanumeric character with `-`, not just separators — visible in the real project dirs, e.g. `-tmp-eval-ut-search-images-005-0-0--bq8yxp6`, which comes from a path containing `ut_search_images_005`. I ran both derivations against that exact directory: the old one **mismatches**, yours **matches**. So on Linux and macOS the old rule produced the wrong directory for any project path containing `_`, `.` or a space, `readSessionLog` swallowed the ENOENT, and the feedback bundle silently omitted the session log. Worth saying in the body — it raises the value of the change and tells the team it was already affecting them.

**I pushed three fixes (`b09903c3`).**

1. **That fix had no regression guard.** The test's tempdir was `feedback-slog-` — all alphanumerics and dashes — so both derivations produce the same directory and the suite passes either way. I reverted production to the old rule and got **23/23 green**. Changed to `feedback_slog-`: the old rule now fails 1 of 23, yours passes 23. One character, and the underscore is the entire point, so the comment says so.
2. **The `.bat` OneDrive guard missed real paths.** `findstr /c:` treats a backslash escape in the search string as syntax, so a OneDrive path containing `\-` never matched. Verified under real `cmd.exe`: for `C:\Users\-TIFE-\OneDrive` findstr stayed **silent** while a substring compare warned; on an ordinary OneDrive path both warn; outside OneDrive both stay silent. Swapped both files to the substring form and re-ran the patched files under `cmd.exe` — warns and continues inside OneDrive, silent and exit 0 outside. It still warns rather than blocks, which is right.
3. **Restored the `# kills server + agent (new session)` comment** the guard dropped, and documented what the Windows branch does *not* do: `start_new_session` is silently ignored there (I checked CPython — it is passed through and dropped, not an error), so `proc.terminate()` reaches the direct child only and orphans the agent grandchild `sandbox_server` spawns. Better than the `AttributeError` it replaces, but not equivalent, and the reader should know. Nothing else in `apps/server` calls an unguarded POSIX API.

**Checked and correct, no change needed:** the Makefile swap — I dry-ran both targets with the var set, unset and empty, and `$(if $(VAR),$(VAR),default)` matches `${VAR:-default}` in all three; leaving `server-dev`/`agent-smoke` alone is right, they embed `grep|cut`. And the shebang removal is safe: `gen-enums.mjs` is mode 644, its sole invocation is `node scripts/gen-enums.mjs`, no `bin` entry, no CI reference, nothing asserts line 1.

**`make test-all` passes on Linux** — the second unchecked box. Typecheck ✓, electron 65, viewer-ui 215, web 10, control-plane 209p/1s, engine 2438, eval-app 168, harness 2386p/2s. (It failed once on a first attempt while I had a review agent saturating the machine; I had truncated that log and cannot name the suite, so I am not attributing it. The clean re-run is green, as are all 12 CI checks.)

**What I cannot close, and what should gate the merge:** `scripts\windows\test-all.bat` on an actual Windows machine — the issue's stated acceptance criterion, and still unchecked. There is no Windows CI (a deliberate decision, now documented in `scripts/windows/README.md`), so nothing here or in CI can establish that removing the shebang fixes the vitest `SyntaxError`, or that the `killpg` guard clears #1713's 8 failures and 2 hangs. Those rest on your word, which I have no reason to doubt — but the genealogist team is the one that hits these, so someone on Windows should run it before this lands. I would not merge on my approval alone.

One more for that Windows pass: `feedback.ts:202` compares `entry.cwd !== folderPath` exactly. Once the directory finally resolves, drive-letter casing (`C:` vs `c:`) would drop every entry — worth eyeballing while you are there.

florencemashipei (changes requested):
Reviewing this as a senior genealogist, so I've scoped myself to the parts our genealogists actually touch — the `eval\*.bat` entry points, `eval/README.md`, and the walkthroughs. I'm deferring to @T-FEH on `local.py`, `feedback.ts`, the Makefile and the shebang; his approval already covers the CODEOWNERS requirement, so this only needs a second approval to merge.

Thanks for taking this on — the OneDrive problem is real and it has cost our Windows testers actual days. I checked the guard logic itself and it's sound: I ran it against paths with spaces, parentheses, `&`, and a leading hyphen, and it fired correctly every time without false-positiving when the checkout was elsewhere. The `findstr` note in your comment is a good catch and worth keeping.

My problem is with where it's wired in, not whether it works. Four things, and I don't think any of them is big.

---

**1. `RunTests.bat` doesn't have the guard, and it's the one they run most.**

`Setup.bat` is once per machine. `RunTests.bat` is every single PR — it's step 1 of `JUNIOR-WALKTHROUGH.md`. It's also the one that trips the failure: it calls `npm run build` and `uv run python run_tests.py`, which are exactly the `node_modules` and hardlink cases your warning names.

This matters more than it looks, because OneDrive can get switched on *after* setup — Windows nags people to "back up your folders" and they click yes. A machine that passed `Setup.bat` cleanly is broken by the next `RunTests.bat`.

Right now it's 2 of our 18 `eval\*.bat` files. Rather than paste 15 lines into the rest, could we pull it into `eval\_CheckSync.bat`:

```bat
@echo off
REM Shared syncing-folder guard, called by the eval\*.bat entry points.
REM Substring compare, not findstr: findstr /c: treats a backslash-escape in the
REM search string as syntax, so a OneDrive path containing "\-" never matches.
setlocal enabledelayedexpansion
set "_hit="
if defined OneDrive           if /i not "!CD:%OneDrive%=!"=="!CD!"           set "_hit=1"
if defined OneDriveCommercial if /i not "!CD:%OneDriveCommercial%=!"=="!CD!" set "_hit=1"
if defined OneDriveConsumer   if /i not "!CD:%OneDriveConsumer%=!"=="!CD!"   set "_hit=1"
if not defined _hit endlocal & exit /b 0

echo.
echo WARNING: This folder is inside OneDrive.
echo.
echo Files here sync to the cloud, and that stops the setup from finishing.
echo It will fail partway through, with errors that change each time you retry.
echo.
echo To fix it: close this window, delete this folder, and clone again to
echo C:\src\cowork-genealogy. In GitHub Desktop, set "Local path" to C:\src
echo before you click Clone.
echo.
choice /c YN /n /m "Continue anyway? (Y/N) "
if errorlevel 2 (endlocal & exit /b 1)
endlocal & exit /b 0
```

then in each entry point, after `cd %~dp0`:

```bat
call "%~dp0_CheckSync.bat" || exit /b 1
```

I'd want it in `RunTests.bat`, `GateSkill.bat`, `CheckSetup.bat` and `Viewer.bat` at minimum. Worth a quick check by you though — I verified your `%OneDrive%` line by hand but not the two I've added.

**2. The guard warns, then does the thing it just said would break.**

`eval/Setup.bat:15`:

> ```
>         pause
> ```

After that it falls straight through and installs anyway. Every other failure in this same file aborts — `pause` then `exit /b 1`, at lines 29/30, 39/40, 56/57, 69/70, 80/81, 97/98, 109/110, 126/127, 140/141, 152/153. Line 15 is the only `pause` in the file that continues.

That break in convention is the whole problem, because `pause` prints "Press any key to continue . . .", and to a genealogist who isn't a developer that reads as an instruction, not a choice. They'll press a key, and ten minutes later `uv` will fail with something they have no way of connecting back to a warning they already dismissed. The `choice` block above is what I'd like instead — it defaults to stopping and makes continuing a decision.

**3. The warning isn't where people choose the folder.**

It landed at `eval/README.md:280` — that's line 280 of 415, under `## Windows users`, a section whose entire body is one sentence about `Setup.bat`. Nobody reaches that before cloning.

The place the decision actually gets made is `eval/JUNIOR-WALKTHROUGH.md:25`:

> 5. Next to **Local path**, click **Choose...** and pick a folder (somewhere you'll easily find again).

Could we make that:

> 5. Next to **Local path**, click **Choose...** and pick a folder. ⚠️ **Not OneDrive, Dropbox, Google Drive or iCloud** — and not `Documents` or `Desktop`, which are usually inside OneDrive on Windows 11 even when they don't look like it. `C:\src` is a safe choice; type it into the box and it'll be created for you. The **Local path** box then shows the exact spot where GitHub Desktop will put the repo — glance at it before moving on.

And one clause in `docs/alpha-feedback-guide.md:72` needs flagging. The advice there to use `%USERPROFILE%\cowork-genealogy\` is actually fine — OneDrive's folder-backup redirects `Documents`, `Desktop` and `Pictures`, not the profile root. It's the parenthetical right after it that's the trap:

> (GitHub Desktop's default clone location is `%USERPROFILE%\Documents\GitHub\cowork-genealogy\` — use whichever path you actually cloned to.)

That names the redirected folder and tells them either is fine. Suggest:

> (GitHub Desktop defaults to `%USERPROFILE%\Documents\GitHub\` — **don't use it**, `Documents` is normally synced to OneDrive and the install will fail there. Clone to `C:\src\cowork-genealogy` instead.)

**4. The warning block is written for developers.**

`eval/README.md:282`:

> The sync client's filesystem hooks conflict with build tools — `uv` fails on hardlinks, Next.js never writes `routes-manifest.json`, and transient ENOENT/EINVAL errors appear under `.next/` and `node_modules/`.

Hardlinks, `routes-manifest.json`, ENOENT/EINVAL, `.next/` — the walkthroughs work hard to keep this vocabulary away from genealogists (they call `uv` "the Python package manager"). This doc already has a house style for describing a failure by what the user sees, at line 62: "the harness will fail every test with a misleading ... error."

The block is right as-is in `DEVELOPMENT.md` — developers read that. For `eval/README.md` I'd want:

> **Do not clone the repo under a syncing folder** (OneDrive, Dropbox, Google Drive, iCloud Drive) — including `Documents` and `Desktop`, which Windows 11 usually syncs to OneDrive. Setup will fail partway through, with errors that change each time you retry. Clone to a plain local path such as `C:\src\`.

---

Two smaller ones, not blocking:

- `scripts/windows/README.md:86` — the "Why there is no `windows-latest` CI job" heading and its first two sentences are what-we-tried, which our review rules say belongs in an ADR rather than a README juniors read. The last sentence ("The workaround is manual: `scripts\windows\test-all.bat` ...") changes what the reader does and should stay. The `DEVELOPMENT.md` and `eval/README.md` blocks are fine by that rule — they're instructions, not history.
- `cd %~dp0` at the top of all 18 `eval\*.bat` is unquoted, so the script dies before the guard runs if a work OneDrive org name contains `&` (`OneDrive - Smith & Co`). Pre-existing, so your call whether it belongs here — but this PR is already in these two files and `cd /d "%~dp0"` is the fix.

Requesting changes on 1–3; 4 I'd like but wouldn't hold the PR for on its own. The code half isn't mine to judge and @T-FEH has already signed off there.

yinkid28:
@florencemashipei — addressed all four points plus both smaller ones in `3c885703`:

1. **`_CheckSync.bat` extracted** — the OneDrive guard is now a shared script (`eval\_CheckSync.bat`) called by `Setup.bat`, `Start.bat`, `RunTests.bat`, `GateSkill.bat`, `CheckSetup.bat`, and `Viewer.bat`. Checks `%OneDrive%`, `%OneDriveCommercial%`, and `%OneDriveConsumer%`.

2. **`choice /c YN` instead of `pause`** — the guard now stops by default and makes continuing an explicit decision, matching every other failure path in `Setup.bat`.

3. **Warning moved to where the decision is made** — `JUNIOR-WALKTHROUGH.md` step 5 now names OneDrive/Dropbox/Google Drive/iCloud (including `Documents` and `Desktop`) and suggests `C:\src`. The `alpha-feedback-guide.md` parenthetical now warns against GitHub Desktop's default `Documents\GitHub\` path instead of suggesting it. `eval/README.md` warning rewritten for genealogist audience (no hardlinks/ENOENT/routes-manifest.json).

4. **`eval/README.md` rewritten for genealogists** — developer-oriented terms replaced with user-visible behavior description.

5. *(smaller)* `cd %~dp0` → `cd /d "%~dp0"` in all touched bat files, so paths containing `&` work.
6. *(smaller)* The `scripts/windows/README.md` "why no CI job" history paragraph — noted, leaving that for a separate pass since it doesn't affect behavior.

Ready for re-review when you get a chance.

florencemashipei:
Thanks for turning this around so fast — the six-script `_CheckSync.bat` and the walkthrough step 5 rewrite are exactly what I was after.

Two things are blocking before I can re-review, and neither is really about your changes.

**1. `vitest` is red, and it predates your note.** It went red at 13:12, about three minutes after `3c885703` landed — so it was already failing when you posted "ready for re-review" at 14:11. Easy to miss.

It isn't the `.bat` logic. `tests/packaging/windows-wrappers.test.ts` has a guard asserting every `eval\*.bat` is either a make target or explicitly allowlisted, and `_CheckSync.bat` is neither:

```
FAIL tests/packaging/windows-wrappers.test.ts > every eval script is accounted for by the duplicate guard
AssertionError: expected [ '_CheckSync.bat' ] to deeply equal []
```

1 failed, 2442 passed — that's the only one.

One line fixes it, at `windows-wrappers.test.ts:63`:

```ts
const EVAL_NOT_A_TARGET = new Set(["Setup.bat", "_CheckSync.bat"]);
```

Worth a short comment there saying a leading underscore marks an internal helper rather than a double-clickable entry point, so the next person adding one knows the convention exists.

That one's on me — extracting the shared script was my suggestion and I didn't check whether adding a new `eval\*.bat` would trip an existing guard.

**2. The branch needs `main` merged in.** It's 28 commits behind now, and `protect-main` has `strict: true` on required status checks, so being up to date is a hard gate rather than a warning. Best done in the same push as the fix above so CI only runs once.

Three of your files have also moved on `main`, so expect to resolve there:

- `DEVELOPMENT.md`
- `Makefile`
- `apps/electron/src/main/__tests__/feedback.test.ts`

⚠️ On that last one — the `feedback_slog-` underscore you introduced is load-bearing and needs to survive the merge. It's what makes the suite discriminate between the old separators-only derivation and the new one; with the original `feedback-slog-` name the test passes against the broken code too. If a conflict resolution quietly restores the hyphen, the test goes green and stops testing the thing it exists for. Your own comment on that line explains it well — just make sure the comment and the name both come through intact.

**On the `scripts/windows/README.md` history paragraph** — agreed, leave it. It was on my not-blocking list and it doesn't affect behaviour. A separate pass is the right home for it.

Nothing else outstanding from me. Ping me once the eight required checks are green on the new head and I'll do the full re-review.

yinkid28:
Both blockers addressed in `282f726e`:

1. **`_CheckSync.bat` added to `EVAL_NOT_A_TARGET`** — with a comment noting the leading-underscore convention for internal helpers. All 15 `windows-wrappers.test.ts` tests pass.

2. **`main` merged in** — all three flagged files auto-merged cleanly. Verified `feedback_slog-` (underscore, not hyphen) survived intact at line 363 of the feedback test.

Ready for re-review.

florencemashipei (approved):
Approving. Re-reviewed the genealogist-facing half against `282f726e` — verified on the branch rather than from the summary:

- **Guard reaches the scripts that matter.** `_CheckSync.bat` is called by `Setup.bat`, `Start.bat`, `RunTests.bat`, `GateSkill.bat`, `CheckSetup.bat` and `Viewer.bat`. `RunTests.bat` was the important one — it's the every-PR script, not the once-per-machine one.
- **The abort actually aborts.** Every caller uses `call "%~dp0_CheckSync.bat" || exit /b 1`, and the guard exits 1 on `errorlevel 2`. That was the way this could have shipped looking right while doing nothing, so I checked it specifically rather than trusting that `choice` replaced `pause`.
- **The warning is where the decision is made.** JUNIOR-WALKTHROUGH step 5 now names the four sync clients plus `Documents`/`Desktop`, and the `alpha-feedback-guide` parenthetical went from recommending GitHub Desktop's default to warning against it. That flip is the one that matters most — it was actively steering people into the failure before.
- **`eval/README.md` reads for its audience now.** No hardlinks, no ENOENT, no `routes-manifest.json` — it describes what the user sees.
- **`cd /d "%~dp0"` in all six touched files**, and `_CheckSync.bat` allowlisted at `windows-wrappers.test.ts:65` with the underscore convention documented.
- **`feedback_slog-` survived the merge intact** at line 363, comment and all. That was my main worry about resolving `feedback.test.ts` against `main`, and it came through clean.

All 12 checks green.

One thing left that isn't yours to fix: the branch is 31 commits behind again — `main` moved overnight after your 22:17 merge. `strict: true` means it needs "Update branch" before the merge button frees up. My approval survives that (`dismiss_stale_reviews_on_push` is false), so no re-review needed after — just update and merge.

Thanks for the turnaround on this, and for pushing back on the `scripts/windows/README.md` point rather than just doing it. That was the right call.

PR #1769, opened 2026-08-19
eval: catch e2e annotation content-drift via findings_hash (#1719)

Asimi1234:
## Summary

Normal tier. Closes the `nothing-checks` gap in #1719: `calibrate_judge`'s drift check compared annotation `per_finding` **keys** to fixture finding **ids**, so a finding body amended while its id stayed the same passed silently — and the judge was then re-run against the *amended* finding while reading the human label produced against the *old* one, fabricating a disagreement in exactly the output the maintainer is told to read first.

- Stamps a `findings_hash` (normalized sha256 of the whole `expected-findings.json`) into the `.ann.json` at grade time; `calibrate_judge` recomputes and **hard-errors on mismatch**, same remedy as key-drift ("re-grade or delete").
- **One shared implementation** (`provenance.findings_hash`, over `harness.snapshot.hash_file`) is called by both the writer and the loader, so they cannot diverge on whitespace/`ensure_ascii` — the failure the issue warns about.
- **Legacy annotations are grandfathered**, never backfilled: an absent hash is *included and graded* but reported unverifiable (not an error, not excluded). All 143 committed annotations predate the check.

## Review guidance

**Start here:** `eval/harness/e2e/calibrate_judge.py` — the new rung 7 in `load_annotated_runs`. It sits after the id/key-drift rung (6) and below the incomplete-grade rung (3): a stamped hash that mismatches → `error` (excluded); an absent hash → include with `findings_hash_present=False`; a match → include verified. The flag is set on **every** include path, and `main` reads it with `.get(..., True)` so a mocked/synthetic case dict can't `KeyError`. Both numbered ladders (docstring + inline `# N.` comments) were renumbered.

The writer is a new tiny CLI, `eval/harness/e2e/stamp_findings_hash.py` (`cd eval/harness && uv run python -m e2e.stamp_findings_hash <ann>`), invoked from the grade-e2e-run skill. It deliberately mirrors the fixtures-root constant rather than importing it from `calibrate_judge` (which would drag in `anthropic`), and `provenance.findings_hash` lives in the anthropic-free provenance module for the same reason. The CLI **resolves the ann path against the repo root as well as cwd** (`_resolve_ann_path`), so the repo-root-relative path the skill hands it works whether run from the repo root or from `eval/harness` (where `uv run` puts cwd) — see the review-round note below.

**Unsure about:** nothing blocking. Two things worth a reviewer's eye:
- *Scope:* `provenance.py` + the stamp CLI are additions beyond the issue's literal **Touches** (`calibrate_judge.py`, `SKILL.md`, `e2e-test-spec.md`), required by the issue's own "one implementation, called by both sides" + "give the one command" bullets. The plan was drafted and adversarially critiqued three times before implementation.
- *Known residual (by the issue's design):* an annotation graded **after** this check but not stamped is indistinguishable from a genuinely legacy one — both are absent-hash → grandfathered, no error. There is no timestamp to tell them apart, and the issue scoped CI enforcement out ("do not touch `check_e2e_fixtures.py`"). Mitigated by the stamp applying by default (the path fix below) and the grade-e2e-run step-7 self-check.

## Post-review round (`/code-review xhigh`)

An adversarial pass caught a real, load-bearing bug and it is fixed in this PR (commit `fcb246a3`):

- **The documented stamp command failed** — `cd eval/harness && uv run python -m e2e.stamp_findings_hash eval/runlogs/e2e/<slug>/...` resolved the repo-root-relative arg under `eval/harness/` and errored (`[Errno 2]`), so **every fresh grade would have landed unstamped and been grandfathered forever** — the exact `nothing-checks` failure this PR closes, shipped inside the fix. Fixed via `_resolve_ann_path` (REPO_ROOT fallback), reproduced-then-verified, and covered by `test_stamp_cli_resolves_repo_root_relative_path`.
- **Doc accuracy:** the "a reformat or key reorder does not fire" claim was corrected — object-key reorder is immune, but reordering the `findings` *array* does change the hash (an accepted cost of the whole-file design), now stated precisely in the docstring and spec §7.4.
- **Test coverage added:** stamp-CLI `main()` (success / missing-fixture→1 / path resolution), `main --dry-run` (grandfather line + exit 0; mismatch → exit 2), and a strengthened stamp test that re-reads the file and asserts sibling keys are preserved.

## Plan

**Didn't change:** `eval/harness/scripts/check_e2e_fixtures.py` (presence, not content; stays stdlib-only per the issue). No committed run log or annotation was re-stamped, re-graded, or rewritten. `load_annotated_runs`' `(cases, problems)` return arity is unchanged — the unverifiable count rides on the case dict, so the ~24 two-tuple call sites are untouched.

**Acceptance check** (offline, no API): `cd eval/harness && uv run python -m e2e.calibrate_judge --dry-run` → **143 graded ready (unchanged from baseline), 0 errors**, plus `143 of 143 ... carry no findings_hash`. *If the ready count had dropped, the change would be wrong* — that is the whole failure mode. The mismatch/error path is proven by `test_loader_content_drift_hash_mismatch_is_error` and `test_main_dry_run_content_drift_exits_two`, and was additionally driven on the real corpus (stamp one ann, amend its fixture → dry-run reported `1 error`, ready 143→142 → reverted).

## Test plan

- [x] `cd eval/harness && uv run pytest tests/unit/test_e2e_calibrate_judge.py` — **59 pass** (incl. the review-round CLI/dry-run tests). Full harness unit: **2391 pass**, 8 skipped; the lone failure is `test_mock_mcp` (needs `make engine-build` in a fresh worktree — no engine code changed here, passes in CI).
- [x] Offline dry-run acceptance above (ready unchanged at 143, 0 errors, unverifiable line).
- [x] `e2e-test-spec.md` §7.4 records the decision + the two alternatives it beat, with **no `#NNNN`** (the file is at its issue-ref ratchet ceiling, 10/10; verified still 10).
- [ ] No skill run-log snapshot changed (this edits `grade-e2e-run/SKILL.md` prose + harness code + docs), so no `make eval-skill` run is owed.

**No CI job covers `calibrate_judge`, and `make test-all` never reads the real corpus** — the mismatch/error path is proven only by the unit tests (+ the manual corpus check above), never by the committed-corpus `--dry-run`, which every grandfathered annotation makes exercise only the absent-hash branch.

Note: this touches Python under `eval/harness/`, outside the CODEOWNERS exclusions, so a `senior-developers` approval is expected.

### DIFF
--- a/.claude/skills/grade-e2e-run/SKILL.md
+++ b/.claude/skills/grade-e2e-run/SKILL.md
@@ -95,22 +95,35 @@ Write `eval/runlogs/e2e/<slug>/run-<ts>.ann.json` with **only** these keys:
 | `notes` | no | `{ "<finding_id>": "text" }` — keys ⊆ `per_finding` keys |
 | `annotator` | no | the grader's team identifier (git blame is the fallback if omitted) |
 
-Do **not** add any other key. The loader hard-errors on unknown keys — in
-particular do **not** write `llm_score`, `corrected_score`, or `verdict`. Those are
-the *unit*-annotation shape and the derived verdict; neither belongs in an e2e
-annotation.
+Hand-write **only** these keys. The one other allowed key, `findings_hash`, is
+added by the stamp command in step 7 — do **not** write it by hand. The loader
+hard-errors on any *other* unknown key — in particular do **not** write
+`llm_score`, `corrected_score`, or `verdict`. Those are the *unit*-annotation
+shape and the derived verdict; neither belongs in an e2e annotation.
 
-### 7 — Self-check, then hand off for commit
+### 7 — Self-check, stamp, then hand off for commit
 
-Verify the file you just wrote — you have every input to do this without any tool:
+First verify the file you just wrote — you have every input to do this without any tool:
 
-- keys are exactly `per_finding` (+ optional `proof_quality_score` / `notes` /
-  `annotator`), nothing else;
+- the hand-written keys are exactly `per_finding` (+ optional `proof_quality_score` /
+  `notes` / `annotator`) — nothing else *yet* (`findings_hash` is added by the stamp
+  command below, not by hand);
 - every `per_finding` value is `true` / `partial` / `false` — no nulls (a blank
   label means that finding isn't graded yet; ask the genealogist before writing);
 - `per_finding` keys equal the fixture's finding ids (you copied them from
   `expected-findings.json`, so this holds by construction — confirm it);
 - any `notes` key is one of those finding ids.
+
+Then stamp the fixture fingerprint into the annotation:
+
+```
+cd eval/harness && uv run python -m e2e.stamp_findings_hash eval/runlogs/e2e/<slug>/run-<ts>.ann.json
+```
+
+This adds the `findings_hash` key so a later edit to `expected-findings.json`
+cannot silently invalidate this grade. It is **not** `calibrate_judge`, makes no
+judge API calls, and is **exempt** from the "do not run `calibrate_judge`" rule
+below — run it every time.
 
 Then tell the user to commit the `.ann.json`. **Do not run `calibrate_judge` — not
 even `--dry-run`.** It classifies *every* annotation in the tree, not just this one.
@@ -122,9 +135,10 @@ periodically — documented in `docs/e2e-testing-guide.md` under "Step 8 — Gra
 
 - **Never open `run-<ts>.json`** or otherwise surface the judge's grades. Grading is
   blind; revealing the judge's labels defeats the calibration.
-- Do not run `calibrate_judge` at all — not even `--dry-run`. Self-validate the one
-  file you wrote; whole-set classification and the calibration sweep are the
-  maintainer's job, per the guide.
+- Do not run `calibrate_judge` at all — not even `--dry-run`. (The
+  `stamp_findings_hash` command in step 7 is a *different* tool — run it; it is not
+  `calibrate_judge`.) Self-validate the one file you wrote; whole-set classification
+  and the calibration sweep are the maintainer's job, per the guide.
 - Do not derive or write a `verdict` — the loader derives it from `per_finding`.
 - Do not edit fixtures, skills, the judge prompt, the run log, or the tree.
 - Do not label findings yourself — surface the evidence; the genealogist decides.
--- a/.claude/skills/grade-e2e-run/SKILL.md
+++ b/.claude/skills/grade-e2e-run/SKILL.md
@@ -117,7 +117,7 @@ First verify the file you just wrote — you have every input to do this without
 Then stamp the fixture fingerprint into the annotation:
 
 ```
-cd eval/harness && uv run python -m e2e.stamp_findings_hash eval/runlogs/e2e/<slug>/run-<ts>.ann.json
+cd eval/harness && uv run python -m e2e.stamp_findings_hash ../runlogs/e2e/<slug>/run-<ts>.ann.json
 ```
 
 This adds the `findings_hash` key so a later edit to `expected-findings.json`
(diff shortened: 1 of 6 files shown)


T-FEH (approved):
Approving. Good change, and an unusually honest PR body — naming your own gap ("no CI job covers `calibrate_judge`") is what let me aim the review at the right places.

**The drift check genuinely works.** I proved it rather than reading it: stamped `alvro-taylor-marriage-1931/run-2026-07-15_23-37-46.ann.json`, amended f1's `description` while keeping its id, and `--dry-run` went to `1 error`, ready 143→142, exit 2. Rung 6 stayed silent throughout — which is exactly the gap #1719 describes. Restored byte-exact. Reformat-insensitivity holds too: compact separators, `ensure_ascii=True`, reversed top-level keys → still 0 errors. And the tests can fail — the comparison stubbed to `if False` fails the drift test, a hand-rolled `sha256(raw bytes)` in the writer fails the round-trip test.

**We fixed the same bug from opposite sides, and yours is better.** The documented command didn't run — after `cd eval/harness`, `eval/runlogs/...` resolves to `eval/harness/eval/runlogs/...` (`[Errno 2]`, exit 1). I'd changed SKILL.md to `../runlogs/...`; `fcb246a3` taught the CLI to accept a repo-root-relative path instead. Yours accepts absolute, cwd-relative and repo-root-relative alike, and its docstring says the skill is *meant* to hand the path it already wrote — so I reverted mine rather than leave the documented command contradicting the CLI's contract. Verified both forms stamp correctly against the merged code.

That one mattered more than a typo usually would: **a skipped stamp is indistinguishable from a grandfathered annotation**, so every grade written under the broken instruction would have looked legitimately unverifiable forever, with nothing to distinguish it.

**Two fixes I kept (`5f21f3a6`):**

- `calibrate_judge.py:510` read the verification flag as `.get("findings_hash_present", True)` — an unknown provenance state reported as *verified*. There is exactly one `cases.append` and it always sets the key, so it never fires today, but the direction is wrong for what the flag means. The comment's stated reason was wrong too: `.get` cannot `KeyError` at any default, so the default's only job is to pick a direction. Now `False`. Worth noting all 47 tests pass either way — nothing pinned it.
- `e2e-test-spec.md`'s annotation field table omitted `findings_hash` while §7.4 of the same document describes the check.

Verified: harness **2398 passed, 2 skipped**; ruff clean; corpus `--dry-run` unchanged at 143 ready / 0 errors / 143 unverifiable; doc-links ratchet 130 passed with the spec still at 10/10; all 12 checks green.

Two things I am not treating as defects, but which the next reader should know:

- **Live coverage is 0 of 143.** Every committed annotation takes the grandfather branch, so the guard protects nothing that exists today. #1719 decided that, and backfilling would mean re-grading blind — right call.
- **Nothing ratchets it out of that state.** No CI job runs `calibrate_judge`, and `check_e2e_fixtures.py` checks annotation *presence*, not content. So the grandfather clause has no exit: it shrinks only as new grades land. Out of scope here (the issue forbade touching that script), but worth a follow-up if the count is still 0/N in a few months.

One correction to the body for whoever reads it next: the CODEOWNERS note says a `senior-developers` approval is expected — every rule now names **both** senior teams, so a senior of either satisfies it.

jud-sdev (approved):
This closes a nothing-checks gap, so I did not take the check on faith — I proved it fires. Stamped rossi-marriage, amended finding f1's body while keeping its id (exactly the drift the id-only rung cannot see), and the loader errored with findings_hash mismatch, dropping ready 143 to 142. Reverted byte-clean. So the guard genuinely catches content drift and is not a check that cannot fail.

Also verified the rest by running: the stamp CLI works from eval/harness with a repo-root-relative path (the review-round Errno 2 bug is really fixed, which matters because a skipped stamp is indistinguishable from a grandfathered grade), grandfathering and verification coexist (one stamped gives 143 ready, 0 errors, 142 unverifiable), the clean baseline is 143 ready / 0 errors, and both sides call the one shared findings_hash over the reused hash_file. Unit suite green.

One residual worth the next reader knowing, by the issue's own design: the guard protects nothing today since all 143 grades are grandfathered, and nothing forces a fresh grade to actually be stamped. #1719 scoped the CI enforcement out on purpose, so this is a possible follow-up if the count is still 0/N in a few months, not a blocker.

Good, honest work.

Issue #1771, opened 2026-08-19
Year qualifiers describe a population that does not exist — rewrite H/Y's instrument around indexed date ranges


chesworthrm:
**Touches:** packages/engine/mcp-server/dev/probe-search-qualifiers.ts (sections H `:2064-2851`, N `:3404-3804`, Y `:5135-5615`), packages/engine/mcp-server/dev/measured-figures.json, packages/engine/mcp-server/tests/packaging/measured-figures.test.ts, packages/engine/mcp-server/src/tools/record-search.ts, packages/engine/mcp-server/src/tools/person-search.ts, docs/specs/record-search-tool-spec-v2.md, docs/specs/person-search-tool-spec.md

Split out of #1409, whose **name** half (fuzzy + initials) is unblocked and proceeds there. Decisions and evidence: https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1409#issuecomment-5349094181

**Filed rather than folded because it is too big for that PR, counted:** 1,670 of the probe's 5,665 lines across three sections; 19 year-specific verdict keys each needing a keep/rename/delete decision; 2 of the 4 `FORBIDDEN_WHEN` rules; three event families needing new enumerable pools.

## The finding: there are no year-silent records

Every year verdict in `dev/measured-figures.json` is phrased around a population — *records with no indexed year* — that **does not exist**. Enumerated, not sampled.

Pool narrowed until the unranged set reads to the end (`q.surname=Pocklington&q.givenName=Thomae&q.recordCountry=England&f.recordType=0`, **585 rows read in full**); year axis partitioned into **12 disjoint 50-year bands**, 1400–1999, every band read in full; personas classified by **band-membership count**, never by reading the payload:

| bands containing the persona | personas |
|---|---|
| 1 | 174 |
| 2 | 178 |
| 4 | 232 |
| 7 | 1 |
| **0** | **0** |
| **all 12** | **0** |

**Closure check: 174×1 + 178×2 + 232×4 + 1×7 = 1,465, and the twelve band totals sum to 1,465 exactly.** Control: the 174 payload-dated personas are precisely the 174 single-band personas. Reproduced number-for-number by a second reviewer.

[...]

Issue #1770, opened 2026-08-19
Tool-description changes silently alter eval inputs — the snapshot excludes mcp-server/src by design


chesworthrm:
**Touches:** eval/harness/harness/snapshot.py, eval/harness/harness/mock_mcp.py, docs/architecture.md

Registering a known gap, per `docs/architecture.md` §9.4 ("A new gap is a labelled issue, not a row here"). **Decided 2026-08-19: this is not being closed.** The issue is the record, not a work item — do not "fix" it without re-reading the trade-off below.

## The gap

`eval/harness/harness/mock_mcp.py:195-206` spawns node to import `allToolSchemas` from the **compiled build** (`build/tool-schemas.js`) and advertises those `description` and `inputSchema` values to the model on every eval run. That is deliberate and good: it replaced two hand-maintained mirrors that had drifted (`research_append` once lacked `ops`; the match-tool fixtures had no schema at all, so the model probed with `{}` and was graded down for the harness-induced retries).

`eval/harness/harness/snapshot.py:33` (`_MCP_SRC_PREFIX`) excludes `packages/engine/mcp-server/src/` from the run-log snapshot, so a change to any MCP tool description **does not mark a single run log inactive**.

Net effect: an MCP tool description can be rewritten, every affected suite's recorded grade keeps displaying as current, and nothing anywhere goes red. It is the exam-question shape — edit the wording, and the already-graded papers still read as current, having been earned against slightly different input.

Measured with `grep -rl "record_search\|person_search" eval/tests/unit/` — **7 of the 25 unit suites** call one of the two search tools:

`init-project`, `person-evidence`, `proof-conclusion`, `record-extraction`, `search-external-sites`, `search-full-text`, `search-records`

## Why it is not being closed

Adding `mcp-server/src` to the snapshot would invalidate **every run log in the repo at once**, and it reverses a decision with a written rationale at `snapshot.py:28-33`: legacy run logs embedded the whole MCP tree, and the exclusion is what lets them re-validate without a re-run rather than staying inactive forever on unrelated MCP-source churn.

So the cost is a full regrade pass across every suite; the benefit is catching a class of drift that has so far produced no observed defect. That trade was put to the lead on 2026-08-19 and declined.

## What would change the answer

- **An eval result that is demonstrably wrong *because* a description changed under it.** That turns a theoretical gap into a measured one, and it is the trigger to watch.
[...]

Issue #1409, opened 2026-08-06
Rewrite the exact-match qualifier family on both search tools — one rule, stated once


chesworthrm:
> **Decided 2026-08-19.** The outstanding decision in the warning block below is answered:
> **when a probe measurement disagrees with what we believe to be true, probe again until
> the two reach consensus.** No reword-around, no provenance escape hatch. Re-probe H and
> Y, then write the prose against the settled verdicts. Ordered steps in [the ruling
> comment](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1409#issuecomment-5343248358).
> The fuzzy and initials legs are gated on neither H nor Y and can land separately.

> **Reviewed 2026-08-17 before junior handoff; answered by the lead the same day.**
> Retitled — this stopped being a `person_search` issue when the ruling landed. The rule
> is the search engine's, so `record_search`'s exact family is in scope too. The original
> body follows under "Original issue" and is superseded wherever it disagrees with this.

**Touches:** packages/engine/mcp-server/src/tools/person-search.ts, packages/engine/mcp-server/src/tools/record-search.ts, docs/specs/person-search-tool-spec.md, docs/specs/record-search-tool-spec-v2.md, packages/engine/mcp-server/dev/measured-figures.json, packages/engine/mcp-server/tests/packaging/measured-figures.test.ts

## The rule (lead, 2026-08-17 — [full ruling](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1409#issuecomment-5317487719))

One rule, belonging to the search engine rather than to either endpoint:

- **Without `exact=on` on a field**, results include fuzzy matches, names matching on
  **initials only**, and records where **the searched-for field is empty**.
- **With `exact=on`**, all three are excluded.

On `person_search` the objects are tree persons rather than records — adapt the noun, the
mechanism is the same.

**Provenance: FamilySearch search-engine internals, from the lead — not measured against
either endpoint as a general rule.** So: state the direction and the mechanism, carry no
figure, and do **not** add `person-search.ts` to `EVIDENCE_SURFACES` in
`tests/packaging/measured-figures.test.ts`. Write that provenance paragraph into
`docs/specs/person-search-tool-spec.md` under "Person fields".

## The trap: the true sentence is currently forbidden by CI

`tests/packaging/measured-figures.test.ts` carries wording rules that fire on the tool
file, the spec, and all five `search-records` skill files. Two are active and both block
the correct wording:

[...]

Issue #1642, opened 2026-08-16
deep dive: search-records


DallanQ:
**Touches:** eval/tests/unit/search-records/, packages/engine/plugin/skills/search-records/SKILL.md

Audit `search-records` for defects nothing has caught. This starts from **no known
failure** — you are looking for things the suite passes today and should not.

**Read [`docs/skill-deep-dive-guide.md`](../blob/main/docs/skill-deep-dive-guide.md) first.** It is the
procedure: build the skill's prohibition list, read transcripts *before* scores,
spend the time on the quiet passes, place each finding in a lane, and write it
Did / Should / Gap.

## Start with this grep

`judge_context` that spells out the score branches tells the judge what to find,
and the judge takes the branch it was shown. Run this first — it is free and it
points at the files worth reading closely:

```sh
grep -l -iE '"[^"]*\bscore [123]\b' eval/tests/unit/search-records/*.json
```

**This skill has 19 such file(s).** That is the largest concentration in the corpus — 19 of the 28 hits repo-wide. Expect this to be most of the session.

The worked example is `ut_search_records_015`, whose `judge_context` read
*"Score 2 if it uses Mullen but doesn't explain why"* — which is exactly what
happened, and the judge scored 3 anyway. Rewrite a hit to name the *dimension*
without writing the finding.

## This skill's numbers

| | |
|---|---|
| Tests in the suite | 27 |
| Rubric+base dimensions that never discriminate | **1 of 7** |
| `judge_context` files naming a score branch | 19 |

1 of 7 dimensions never take more than one value, so they cannot report a defect in whatever they nominally grade.

## The step that pays for the session

Step 6 of the guide — **turn every finding you can into a validator request**. A
hand fix closes today's instance and costs a genealogist again next run; a
validator closes the class for every future run. You supply the genealogical
rule; a developer writes the Python. The guide has the seven shapes that convert
and a four-line request template.

The measure of this task is **validator requests, not findings**.

## What to hand back

- The prohibition list from Step 1, saved for the next auditor
- Each finding in Did / Should / Gap form, with its lane
- A validator request for every finding that converts
- For grading defects you own (`judge_context`, `rubric.md`): the fix, made

## Cost

Editing anything under `eval/tests/unit/search-records/` or the skill body flips this
skill's run-log snapshot inactive and buys one `make eval-skill SKILL=search-records`
run. **Batch every finding into that one run** — that is the whole reason this is
[...]

PR #1773, opened 2026-08-20
record_search/person_search: state the exact-match rule once (#1409)

chesworthrm:
## Summary

Tier: **Normal.** Refs #1409 (the name half). Unblocks #1771.

The `*Exact` family described the same behaviour 43 times across two tools,
inconsistently, and in one place told the agent not to use a lever that works.
It is **one rule belonging to the search engine**, so it is now stated once per
tool with per-parameter one-liners underneath.

> Without an `*Exact` flag a name field also matches fuzzy spellings, and it keeps
> objects where **that field is empty** — for `givenName` and for a father's,
> mother's, parent's or spouse's name (`other` not measured), but **not** for
> `surname`, where an unqualified value drops surname-empty records outright.
> Setting a flag excludes whatever its own field admits.

**The two principal fields behave oppositely, and getting there took three tries.**
The first revision of this PR asserted the empty-field leg for *every* name field and
claimed the rule was measured on both endpoints. Code review caught both, and the
second revision then declared the leg **unestablished for both** principal fields.
That was also wrong, and the reason it gave — "`surname` is a required anchor, so a
surname-empty record is not a case this endpoint puts in front of you" — is not why:
`surname` is one of three anchors. So it was measured directly. What is now
enumerated:

| leg | evidence |
|---|---|
| fuzzy | `B.verdict:exact surfaces records fuzzy does not` = NO — exact is a strict subset |
| empty-field, **relatives** | `F.verdict:relative .exact requires the relative to be present` CONFIRMED; `R.verdict:spouse .exact requires the spouse to be present` CONFIRMED across two populations read in full; `R.verdict:keep-silent` HOLDS |
| empty-field, **`givenName`** | **KEEPS them.** Three unmatchable tokens retain 251/252/252 of a 6,038 pool — spread under 2%, which is section S's control separating silence from fuzzy reach — 58 of 60 sampled retained rows carry no typed Given part, and `.exact` takes it to 0 |
| empty-field, **`surname`** | **DROPS them.** A bound record (`fullText` "Escolastica", one Given part, zero Surname parts) is present in a 1,100-row set read to the end, and each of three unmatchable tokens returns an **empty set** — zero rows, so not a ranking artifact |
| both of the above | measured 2026-08-20, reproducible via `dev/explore-name-empty-field-leg-records.ts`, and **not** in `measured-figures.json` — the artifact still reads `T.verdict:all name fields behave alike` = "NOT MEASURED", which is why this was measured at all |
| initials | only one direction: an initials *value* reaches its transposition (`I.verdict:.exact pins the initials ORDER`, CONFIRMED/ENUMERATED). A spelled-out query reaching initials-indexed records is unmeasured, and the converse verdict `fuzzy swallows initials into spelled-out names` reads **NO** |

**And `person_search` is stated on provenance, not measurement.** Every figure
behind the rule was taken against the record index. `grep -rl
"platform/tree/search" packages/engine/mcp-server/dev/` returns nothing — no probe
covers that endpoint, and no figure in `dev/measured-figures.json` was taken
there. A session probe was consistent with the rule but left no artifact, so it is
not cited and nothing can be diffed against it. Building that probe is the work
that would change this; the spec says so. This is the #1093 error and it took a
review to catch it in my own diff.

## What to scrutinise: the three new checks, each proven to fail

Every one was fault-injected, watched fail, and restored. Two of them exist
because a guard was silently absent, which is the failure mode this file's own
header is about.

**1. A renamed or deleted verdict key silently disarmed its wording rule.**
`get()` returns `undefined` for a missing key, `String(undefined ?? "")` is `""`,
no `activeWhen` matches the empty string — so the rule returned early and its
test passed. Renaming `H.verdict:silence tolerated` left the file at 15/15, and
the producibility check missed it too. Injected three ways:

| injection | reported |
|---|---|
| rename a `FORBIDDEN_WHEN` key | `FORBIDDEN_WHEN -> H.verdict:silence tolerated` |
| delete a `FORBIDDEN_WHEN` key | `FORBIDDEN_WHEN -> Y.verdict:generalises past birth (impossible-range)` |
| rename a `CONTRADICTION_PAIRS` path | `CONTRADICTION_PAIRS -> R.verdict:retention equals the silent share` |

This is why the commit order matters. #1771 deletes the estimator block emitting
that first key and renames several others; without this test that rewrite removes
two of the four wording guards with nothing going red.

**2. Neither `person_search` surface was scanned for contradicted wording.**
`FORBIDDEN_WHEN` scanned `EVIDENCE_SURFACES` + `AGENT_SURFACES`; both
`person_search` files were in neither. PR #1699 found it by accident and named
it — the spec's schema block "happened to still match its tool. That was luck."
New `WORDING_ONLY_SURFACES`, deliberately a third list rather than an addition to
`EVIDENCE_SURFACES`, because `person_search` carries no figures and must not
start. The control is what proves the scoping:

| injected into a person surface | result |
|---|---|
| sentence contradicting a recorded verdict, in `person-search.ts` | **FAIL** |
| same sentence in `person-search-tool-spec.md` | **FAIL** |
| untraceable figure `1,234,567` in `person-search.ts` | **FAIL** |
| untraceable figure `7,654,321` in `person-search-tool-spec.md` | **FAIL** |

Both figure rows changed during review, and the change is the point. `person-search.ts`
joined `AGENT_SURFACES` (no absolute totals — a model-read surface should carry no
comma-grouped figure), and the person **spec** joined `EVIDENCE_SURFACES`, because it
carries three real figures (`56,177`, `~9,700`, `2,916`) while sitting in the
wording-only list: the hole this PR closed for the tool file was open on the one
`person_search` file that actually holds numbers. Those three come from a 2026-05-28
probe predating the artifact and are exempted by name, so any NEW figure must trace.
Neither change puts `person-search.ts` in `EVIDENCE_SURFACES`, which the 2026-08-17
ruling forbids.

**3. `*Exact` descriptions are capped at 250 chars.** Sized from the longest
non-exempt description actually written — `givenNameExact` at **238**, which
carries the diminutive note and the initials-order note — not guessed. Six
exceeded the ceiling before this change: **475, 402, 375, 341, 269, 255**. It
caught **three of my own one-liners** along the way (254, 242, 268), each
shortened rather than raising the ceiling.

**250 is the midpoint of the empirical gap** — between the longest legitimate one-liner (238) and the smallest historical offender (255). 260 is unavailable: it would re-admit that offender, and a lint that permits what it was written to catch is worse than a tight one. A guard asserts both bounds. The docstring says so, and warns that a three-word
clarification to `givenNameExact` will trip this lint, and the fix is to shorten
it or move the initials note to the spec — not to raise the ceiling and not to add
an `EXEMPT` entry. That comment was itself wrong twice about its own basis (first
`surnameExact` at 205 with a baseline list uniformly +2 from counting the
enclosing quotes; then naming `birthPlaceExact` as longest while quoting a bigger
number for `givenNameExact` in the same sentence). Both errors are recorded in it,
since it is the only justification for 240.

The exemption check also now enforces what its docstring promises. It asserted
only that the parameter still exists, so a leftover entry for a description that
had since been shortened would be a permanent pass. Proven by collapsing
`birthYearExact` to a one-liner: *"description is 35 chars, already under the 240
ceiling — the exemption is dead, delete it"* — which is exactly the #1771
scenario. (An earlier version of the sizing comment carried figures that were
uniformly +2, from counting the enclosing quotes; since that comment is the only
justification for 240, an off-by-two there is the same class of defect as a
guessed threshold.)

## Years are deliberately untouched

`record_search.birthYearExact` is **byte-for-byte identical**, its four
`As birthYearExact` siblings still point at it, and `person_search`'s four year
toggles carry a pointer that asserts nothing measurable.

The reason is not caution, and it is not an enumeration either. The artifact reads
`H.verdict:silence tolerated` = **OPEN** and `N.verdict:payload-silent means
index-silent` = **NOT MEASURED**: nothing recorded establishes what a year range
does to a record carrying no indexed year, in either direction. The lead's account
is that no such record exists — the index carries an estimated date *range*
matched by overlap — and a session probe was consistent with that, but it left
**no artifact** and no probe in `dev/` reproduces it, so the specs no longer cite
it.
Writing the corrected clause here would assert ahead of the artifact, and no CI
rule would catch it: H's `mustNotSay` patterns match three specific phrasings and
an estimated-range sentence is none of them. #1771 rewrites the instrument, then
the prose. `birthYearExact` is exempt from the new ceiling **by name, with the
removal condition**, so the exemption cannot outlive its reason.

## Budget, honestly — and the numbers live in the lint, not here

**`person_search` grew.** Its toggles were stubs that were also wrong ("requires an
exact surname match (no fuzzy nicknames or spelling variants)" omits two of the
three legs), so correctness cost tokens there. Combined across both tools the
change is now a net **saving of about 135 characters**, under 1%. That figure has
been an increase and a saving at different points on this branch, which is the
reason to treat it as a side effect and not the argument — the last move came from
cutting four relative descriptions that restated the shared rule the tool-level
text already carries. `record_search.birthYearExact` at 475 chars is most of the
remaining bulk, and #1771 collapses it.

**The figures are stated once, in `tool-description-length.test.ts`'s docstring,
and deliberately not repeated here.** That docstring was wrong about its own basis
three times in this PR — measure, edit the descriptions it describes, forget to
re-measure — so restating the numbers in a fourth place is how a fifth error
happens. The docstring now carries the derivation and tells the next editor to
re-derive from `allToolSchemas` before touching it.

One thing worth knowing separately: measure from **source**, never `build/`. The
committed build is stale and reports ~12.3k for `record_search` against source's
~15.5k.

## Two repo lints caught the spec edits, and were obeyed rather than widened

`doc-links.test.ts` rejected `person-search.ts:201-202` and `:285-286` — a line
number is a copy of state the file owns, and 3 of 3 sampled cites in this repo
already pointed at the wrong code. Now cites `buildSearchUrl` and
`personSearchTool` by symbol.

It also holds an issue-reference ratchet on contract docs: **state the fact, not
the ticket.** Six refs I had added came out and the facts went in — the year prose
says what the artifact records (`OPEN`, `NOT MEASURED`) and that the wording is
provisional, which is what a reader needs and does not rot when the ticket
closes.
`ISSUE_REF_BASELINE` is unchanged.

## Eval exposure — decided, not open

`mock_mcp.py` advertises `description`/`inputSchema` from the compiled build to
the model on every eval run, and `snapshot.py`'s `_MCP_SRC_PREFIX` excludes
`mcp-server/src/`, so **this PR changes model-visible input for 7 of 25 unit
suites with no run log invalidated and no re-run triggered.**

Per the lead's decision on #1409:

- **Re-run and regraded:** `search-records` — with a caveat added late: that run
  log now predates **7** further commits to the two tool description files,
  including the section-R wording and both prose-review passes. Its recorded input
  no longer matches what the model would see either, so it sits in the same
  position as the six below. Whether to re-run and re-annotate is an open decision,
  not a settled one. Nothing detects this, which is precisely #1770.
- **Knowingly not re-run**, on run logs whose recorded input no longer matches
  what the model would now see: `init-project`, `person-evidence`,
  `proof-conclusion`, `record-extraction`, `search-external-sites`,
  `search-full-text`.
- The fingerprint **stays as is** — re-snapshotting `mcp-server/src` would reverse
  a deliberate decision and invalidate every run log at once. The blind spot is
  registered as **#1770**.

Unlike the run logs above, the gate DOES catch this one: this PR touches three files under
`packages/engine/plugin/skills/` (SKILL.md and two references), so
`check_runlogs.py` arms and the `search-records` re-run was forced, not voluntary.

## A self-audit found the same defect a third time

After the review above, I checked every prose claim on this branch against
`dev/measured-figures.json` one at a time. Most held. Two did not, and the first
is the mistake review had already caught once on the other endpoint:

- **The year finding is not in the artifact on EITHER endpoint.** Review caught it
  for `person_search`; it was equally true of the record side and I missed it. The
  figures behind "enumerated at zero" — 585, 1465, 174, 178, 232, 364 — appear
  **zero times** in `measured-figures.json`, along with no "band", no "overlap",
  no "estimated range". So the record spec, which is the evidence trail, asserted
  an enumeration contradicting the artifact it documents. It slipped both lints:
  traceability, because "zero" is a word not a comma-grouped figure; and
  `FORBIDDEN_WHEN`, because that catches a sentence *denying* a verdict, never a
  novel claim no pattern covers.
- **"It still descends" is recorded by no verdict at all.** The upward half is
  measured (the wrong Arkansas county returned the same total as the right one);
  the descent half appears nowhere — the only place verdicts concern wildcards. I
  carried it from the original text and repeated it in three new places. Dropped
  from both tools and the person spec. Two pre-existing occurrences in the record
  spec's place prose are left alone (places are out of scope by ruling) and
  flagged for whoever owns them.

What held, each checked against its verdict rather than asserted: the reordering
fact (`B.verdict:exact reorders the shared records` = YES, 54 records move,
largest by 34), the strict-subset fact (`B.rows` 521 -> 81, `exactOnly: 0`,
`fuzzyOnly: 440`), the misspelling claim (`D.verdict:fuzzy bridges the misspelling
and exact destroys it` = HOLDS), the relative empty-field leg (`F` and two `R`
verdicts, CONFIRMED/HOLDS), and initials order (`I.verdict:.exact pins the
initials ORDER`, CONFIRMED/ENUMERATED).

One thing to know rather than trust: **the diminutive *exclusion* rests on the
lead's 2026-08-17 ruling, not on measurement.** `N.verdict:diminutiveReach`
measures the fuzzy *reach* only. The ruling authorises stating it; flagged so the
basis is visible.

## Seven `dev/explore-*.ts` scripts, with a stated disposition each

The measurements this branch acted on ran in a scratchpad, which is *why* three
unsupported claims reached a tool description and two specs. `CLAUDE.md` already
names `dev/explore-*.ts` as the home for "scripts that document the live-API
evidence trail behind each spec", so the scratchpad was never the only option —
that was the actual mistake, not the absence of a full probe section.

They are committed here so the next person starts from working code rather than a
prose description of a result. `tsconfig.typecheck.json`'s `include` covers `dev*`,
so CI typechecks them; one was re-run from `dev/` to confirm its imports resolve.

**Every header states that its output is NOT in `dev/measured-figures.json` and
must not be cited as measured.** Beyond that, three different dispositions —
because "make it citable" is right for only two of the seven:

| script | disposition |
|---|---|
| `explore-year-bands-records.ts` | **Promote.** Real measurement of the year mechanism; #1771 steps 0-1 own it |
| `explore-year-bands-tree.ts` | **Promote**, same |
| `explore-tree-empty-field-leg.ts` | **Never citable, by decision** |
| `explore-tree-range-sweep.ts` | **Never citable, by decision** |
| `explore-tree-require-switch.ts` | **Not a measurement** — an instrument demo |
| `explore-tree-204-vs-429.ts` | **Not a measurement** — an instrument demo |
| `explore-persona-vs-record-ark.ts` | **Not a measurement** — an instrument demo |

**Why two must never become citable.** The 2026-08-17 ruling on this family says of
`person_search`: *"state the direction and the mechanism only, carry no figure from
it, and do not add `person-search.ts` to `EVIDENCE_SURFACES`."* Promoting those two
would manufacture exactly the figures it forbids. Not-citable is the decision, not
neglect — said explicitly so a later reader does not "fix" it.

**Why three are a category error for citability.** They demonstrate defects rather
than measure behaviour: a mandatory query switch whose absence is silent, a 204
that a naive reader turns into an error, and a scope test that compares persona
arks (`1:1:`) to draw a conclusion about records (`1:2:`). Their promotion path is
a **test**, not a probe section — a test fails when the defect is fixed, which is
the signal worth having. Two of them describe **live** defects: the probe's
`searchOnce` has no 204 branch, and its scope-test docblock still carries the
record conclusion.

Both traps are worth having as executable demonstrations rather than warnings,
because each failure is silent *and inverts the reading*: without the require
switch every query returns the surname-only total, which reads convincingly as "the
given name does not filter"; and 72 apparent "429s" were 204s answering correctly
on the first call. A warning in prose did not stop me making both mistakes.

## What none of this establishes

**Nothing in CI checks whether a tool description is true.** `make engine-test`
passes on any wording; the manifest drift test compares tool *names*, not
descriptions; the new ceiling checks length, not accuracy. The measured-figures
suite checks prose against recorded verdicts, which is why its rules have to move
with the prose rather than constrain it. A green suite says nothing about
correctness here — the probe and the artifact are what carry that, and for the
name legs they now do on both endpoints.

**And the gap the three defects above share is not closed by anything here.** Both
new guards catch a sentence that contradicts a recorded verdict; all three of these
asserted something no verdict covers, which no pattern can anticipate. That is an
argument for putting the measurements in a probe (issue #1771 step 0), not for
another lint.

Also: `other*` `*Exact` remains unmeasured — assumed to behave as the father and
spouse families do, and says so. `mother*` and `parent*` are now measured
(Section R) and their descriptions say so. Places are
out of scope by ruling — a different mechanism (upward expansion), and three of
`person_search`'s place toggles have no `record_search` counterpart at all.

## What code review changed

Seven findings, all addressed by **narrowing the claims rather than re-measuring**
(the lead's call). Three were mine stated as measured without support — the
principal-name empty-field leg, the initials direction, and the tree-endpoint
"Measured on this endpoint." The reviewer's own note on why the new guards did not
catch them is the useful part: *"a rule can only forbid wording somebody wrote a
pattern for, and the three new claims above have no pattern."*

One finding dissolved rather than being fixed. Review flagged that
`name-search-mechanics.md` and the tool told the agent different things about
`surnameExact`; the divergence was entirely the over-claim. With the rule narrowed,
`search-strategy-levers.md`'s relative row ("Requires that relative to be indexed,
so it drops the silent records the unqualified term keeps") and
`name-search-mechanics.md`'s fuzzy mechanics both agree with it. **No skill file is
touched**, so no eval re-run is needed and the 28/28 run below still stands.

## The empty-field asymmetry, and two guards that caught me

The finding above landed late, after the question "hadn't we already proven this?"
It had two consequences worth a reviewer's attention.

**It sits at tool level only, deliberately.** Putting the asymmetry in the
per-parameter one-liners would duplicate the rule and push `surnameExact` (237) and
`givenNameExact` (238) past the 250 ceiling. So a reader of `surnameExact` alone does
not see the exception — judged acceptable because the tool-level text is read on
every call, but it is a judgement, not an oversight.

**Two guards fired mid-edit, which is the only reason the figures here are right.**
`DOCUMENTED_TOTALS` flagged the description totals going stale the instant the
tool-level rule changed (14,286 → 14,289 and 4,918 → 4,995), which is what forced
the budget figure above to be re-derived from the schemas rather than carried
forward. And traceability rejected `1,100`
in the record spec — exempted by name with the script as its trail, not by relaxing
the check. `6,038` in the same passage passed **legitimately**: it is the recorded
`Y.impossiblePools:birth` Pocklington pool total, the same pool.

That is the first time a guard added by this PR caught my own staleness before it
shipped, rather than a reviewer catching it after.

**Every prose claim was then re-checked**, one at a time: 11 artifact-backed claims
against `measured-figures.json` verdict by verdict (all hold), 9 non-artifact claims
for whether they name their source (all do), and both superseded formulations
confirmed absent from both specs.

## Section R now measures mother and parent instead of assuming them

Six `*Exact` descriptions read "Assumed, as `motherGivenNameExact`". Six hedges is
a measurement task, not a wording task, so section R was extended from two
relative families to four and re-run. All figures below are from
`dev/measured-figures.json`, not from this description.

| pool | family | silent share | retention | conflicts |
|---|---|---|---|---|
| Brazil/Bochenek | father | 70.2% | 70.2% | 0 |
| Brazil/Bochenek | spouse | 9.8% | 10.2% | 0 |
| Brazil/Bochenek | **mother** | 70.1% | 70.1% | 0 |
| Brazil/Bochenek | **parent** | 70.1% | 70.1% | 0 |
| England/Pocklington | father | 92.8% | 92.8% | 0 |
| England/Pocklington | spouse | 81% | 81% | 0 |
| England/Pocklington | **mother** | 99.1% | 99.1% | 0 |
| England/Pocklington | **parent** | 92.8% | 92.8% | 0 |

`.exact requires the relative to be present` is CONFIRMED for father (2
populations), spouse (2), **mother (1 — Pocklington indexes no mother given names
at all, and the probe says so rather than inferring)** and **parent (2)**. All four
aggregate verdicts — `keep-silent`, `keep-matching`, `drop-contradicting`,
`retention equals the silent share` — still read HOLDS.

**`other` is excluded by decision, and recorded as such rather than omitted.**
`R.verdict:other names behave like the four kinship families` reads NOT MEASURED,
carrying the numbers behind it. It is not a kinship role — `q.otherGivenName` is a
co-occurrence search — and it is the only multi-valued family, since a register
entry names godparents and witnesses, any of whom can satisfy the query. Measured
with a single-name accessor and again with an any-name accessor: both put 62
records in the conflict bucket (9 Brazil, 53 Pocklington) while all four kinship
families held at zero, flipping two verdicts. The untested hypothesis for those 62
is recorded with it. Its two `*Exact` descriptions read "Not measured."

**Three classifier defects of mine, each found by a verdict flipping and each
fixed.** `mothersIndexed` and `fathersIndexed` replace `parentsIndexed` as the
sex-specific silence denominators, because `parentsIndexed` counts either sex and
so files a father-only record as mother-*nameless* when it is mother-*silent* — 30
misfiled rows in Pocklington. And the exact-presence block is now POPS-outer,
enumerating each baseline once per population instead of fifteen times.

**The last of those three I could not predict, so the probe now measures it.** I
did not know whether fixing father's denominator would move anything, and said so
instead of picking a branch. `R.parentSexReadability` now records, per population,
how many rows name a parent and how their sexes read:

| pool | rows naming a parent | all female | all male | unreadable sex |
|---|---|---|---|---|
| Brazil/Bochenek | 156 | 1 | 0 | 8 |
| England/Pocklington | 34 | 0 | 30 | 4 |

That settles it: Brazil's single all-female row is exactly the one that moved, so
father there goes from 70.1% silent against 70.2% retention to **70.2% against
70.2%** — the fix *closed* a one-row gap. Pocklington is untouched because its 4
indexed-but-nameless rows carry parents of unreadable sex, which both denominators
include by design. Nothing flipped. My own speculation had offered two branches
and reality took neither cleanly: I never considered that the fix might improve
agreement.

Section F keeps `parentsIndexed === 0` deliberately — its own comment explains
that it needs a *sufficient* condition to pick representatives that certainly have
no father, where R needs *completeness*. Section T's is unchanged because its
verdict reads NOT MEASURED HERE under RULE 0.

**One probe was deleted rather than kept.** An exploratory script concluded the
opposite of `R.verdict:keep-silent` from a place/date-anchored pool with no name
anchor, which makes an unmatchable relative token the only name term and returns an
empty set. Its three traps are folded into R's comments, including a correction to
my own: an unmatchable token is *not* confounded for relatives in general, only
without a name anchor — R's pools are all surname-anchored, which is why the method
works there.

## The sixth review: four self-contradictions, all self-inflicted

`/code-review` on the finished branch found four defects, and every one was a
contradiction between two claims already on it — both halves mine, each half
individually checked against the artifact when written. Fixed in `4b789d84`:

| Was | Now |
|---|---|
| Tool level said "any relative's name"; six parameter descriptions in the same file said "only the father and spouse families were enumerated" | "a father's or spouse's name", matching the artifact's `F` (father) and `R` (spouse) verdicts. **Five sites**, including this PR body — the review named two |
| `person-search-tool-spec.md` Examples said the surname-empty drop is "not established for a principal name field, on either endpoint" — added eleven commits before the rule box that measured it | States the drop, keeps the initials-only claim hedged, since that half is still unmeasured |
| `record-search-tool-spec-v2.md` said "these figures are not in `measured-figures.json`" | `6,038` **is** in it, as `Y.impossiblePools:birth[0].poolTotal`. Names which figures are recorded and which are not |
| The explorer's trap 2 said the `fullText` cross-check is what "this does" | Only the surname leg does it. The `givenName` leg counts typed Given parts, so its 58/60 is a **lower bound**; both legs' wording narrowed to what each detector establishes |

**What found them, and what didn't.** Five passes of grepping missed all four
because I grepped the phrasing I was replacing. These live under `relative`,
`empty`, and `not in`. Grepping the claim's *subject* across every surface that
carries it — both tools, both specs, the dev script, and this body — found them
and found three more sites than the review reported.

`DOCUMENTED_TOTALS` fired again on this fix (14,281 → 14,289, 4,985 → 4,995). Because it
asserts the totals against `allToolSchemas`, it necessarily fires on every
description edit made after it was added, and it did — each time mine. That is why
the figure in the budget section above is the lint's current value rather than any
number I carried forward.

**Still open from that review, not fixed here:** the `givenName` leg's missing
`fullText` cross-check needs a live re-run to settle (its figure is documented as
a lower bound meanwhile); the three-token agreement guard is satisfiable by one
comparable read (`spread()` returns 0 for a single element); `DOCUMENTED_TOTALS`
sums every description while its docstring says it checks only the `*Exact`
family, so any unrelated description edit reds it; and three `dev/` scripts parse
`Retry-After` without a `Number.isFinite` guard, where an HTTP-date value yields
`NaN` and fires every retry immediately.

## Self-review after the third review round

I audited the branch against this session's own postmortem — grep every figure,
grep the claim not the file, re-run scripts instead of reading their diffs — and it
found one serious defect **in the commit that was fixing this class of defect**.

Commit `2f165747` claims of `explore-year-bands-records.ts`: *"It now computes the
identity and REFUSES a verdict when it fails."* It did not. The insertion had died
on an assert, printed nothing, and I read the absent failure as success. Worse, the
block I had drafted computed **both sides of the identity from the same map**, so it
compared a number to itself and could never fail — the "check that cannot fail"
`CLAUDE.md` prohibits.

It is now there and non-tautological, verified by running rather than by diff:

- `CLOSURE: band rows read=1465  sum(membership)=1465  equal=true` — the figure
  quoted since 2026-08-19, asserted by the script for the first time
- proven able to fail by skipping one band's tally: `1465 / 1101 equal=false` ->
  "REFUSING a verdict"
- plus a second stop: a band that did not enumerate refuses a verdict, because
  closure cannot see partial coverage (both sums fall together)

Every other figure on the branch was re-derived: the ceiling docstring's 238/237/181/475
from `allToolSchemas`, the six-over-ceiling baseline from `origin/main` (exact match),
and the before/after token pair via a temporary `origin/main` worktree — needed
because a regex over source undercounts escapes and gave me a wrong answer first.

## Verification

- `make engine-test` — 103 files, **2442 passed**
- three new checks, each fault-injected and confirmed failing before commit
- `make eval-skill SKILL=search-records` — **28/28 pass**
  (`v1_2026-08-20_01-15-26`), $9.49, 354 turns

### The one advisory, and why it is not this PR's

The run reports `routing_negative_judge_fail: 1 — ut_search_records_003`.

`ut_search_records_003` is a **negative** routing test: the skill should decline
and hand off to `record-extraction`. It **passed**. The judge nonetheless scored
Correctness 1 and Completeness 1, reasoning *"Claude produced no output and made
no tool calls, meaning the skill did not invoke record-extraction as required"* —
i.e. grading a negative test as though it were a positive one. That is what the
advisory exists to surface.

**Pre-existing, not a regression.** Compared against the previous run log rather
than asserted:

| | `v1_2026-08-17_18-06-27` (before) | `v1_2026-08-20_01-15-26` (after) |
|---|---|---|
| outcome | 28/28 pass | 28/28 pass |
| `003` judge scores | 1 / 1 / null | 1 / 1 / null |
| advisory present | yes | yes |

Identical, so the rewritten descriptions changed this suite's model-visible input
and the suite is unmoved — which is the result the run was commissioned to
establish. The judge/rubric defect is `/audit-rubric` territory, not a
description fix, and folding it in here would be the lane violation
`docs/skill-lifecycle.md` §5 warns about.

**Annotated and committed.** `v1_2026-08-20_01-15-26.ann.json` is in the branch: 45
corrections across 7 tests, **zero score changes**, all five sampled tests complete at
every dimension. `ut_search_records_003` is not in the review sample and so is
unannotated — permitted, and the reason the comparison above is against the judge's
own scores rather than a human call.

**Known limitation, stated rather than papered over.** This run predates the last
several commits, which rewrote both tool descriptions again. `mock_mcp` advertises the
compiled descriptions to the agent, so the run measured text that has since changed,
and nothing flips the log inactive — that is exactly the gap #1770 registers. Whether
to spend a re-run is a lead decision; absent one, read 28/28 as evidence about the
narrowing's direction rather than its final wording.

## Six review rounds, and four things deliberately left

Each round found real defects; the substantive ones are above and in the commits.
Knowingly not fixed here:

- **The eval run predates the last description edits** — see above.
- **`.idea/` is untracked and absent from `.gitignore`**, so one `git add -A` commits
  IDE config.
- **`2f165747`'s commit message is wrong.** It claims a closure assertion that did not
  exist until `7846569b`. It is pushed, so correcting it means rewriting published
  history.

## Deviated from the plan

`PLAN.md` sized the ceiling at 120 chars before any one-liner had been drafted;
the real ones run to 205, so it is 240, sized from the drafts. Recorded because
guessing a threshold and then discovering it is the thing this repo keeps
documenting.


chrisedeson (changes requested):
Large diff, mostly generated eval run-log JSON and nine new exploratory `dev/` probe scripts. The two shipped files (`record-search.ts`, `person-search.ts`) are pure description-string edits with no logic changed. Ran a full structural review across ten angles plus my own direct verification of the two most important findings.

### One CI anomaly worth flagging up front

The "runlogs" check is currently red on GitHub, but running the exact same script (`check_runlogs.py`) locally against the exact same commit, with the exact same `BASE_SHA`/`HEAD_SHA`, passes clean — confirmed three ways (my own hash comparison, calling the real `diff_snapshot_vs_disk()` function directly, and ruling out a shallow-clone/line-ending checkout difference). This doesn't reproduce locally. Recommend re-running the CI job before reading anything into it — but see the next finding, which is a **real** reason this check will fail regardless once the branch is up to date.

### Should fix before merge

**1. This branch is stale against main in a way that collides with an already-merged PR on the same eval snapshot.**

Confirmed directly: `upstream/main` already has PR #1786 (`900bf20d`, merged after this branch forked), which also edits `search-records`'s `SKILL.md`, `rubric.md`, and 8 test JSONs, and adds its own run-log candidate `v1_2026-08-20_11-04-51.json`. `upstream/main` currently sits at exactly the documented 5-candidate cap. This PR's own commits delete one old candidate and add two new ones, all computed against a branch point that never saw #1786. Merged as-is, the result is 6 candidate pairs for `search-records`, one over cap, and this PR's own latest run log's embedded snapshot will no longer match disk once #1786's changes land alongside it — the exact rule 2 gate this PR's own body claims doesn't apply ("this PR touches no file under `packages/engine/plugin/skills/`" — false, `SKILL.md` is modified right here). Rebase onto current main and re-run `make eval-skill SKILL=search-records` before requesting review.

**2. A reference doc this PR didn't touch now contradicts the PR's own corrected claim, in the same skill.**

`packages/engine/plugin/skills/search-records/references/place-date-mechanics.md:275-279` still says relative-name narrowing was "measured... for the father and spouse names only" and that "mother, parent and other names were not enumerated." This PR's own Section R extension enumerated mother and parent on the same two populations, and the sibling `SKILL.md` (edited by this PR, line 228) already says "Enumerated for father*, spouse*, mother* and parent*." An agent loading this reference doc mid-session gets told the opposite of what `SKILL.md` and both tool descriptions say in the same invocation — the exact cross-surface-disagreement defect this PR's own six review rounds hunted down elsewhere (per its commit history) but missed here, since this specific file was never opened.

**3. `SKILL.md` now contradicts itself within the same file, and the tool spec has the identical stale claim.**

`SKILL.md:210-211` still reads "measured for the father and spouse names," while the corrected line 17 rows later (227-230, this PR's own edit) says "father*, spouse*, mother* and parent*" — both describing the same underlying finding. `docs/specs/record-search-tool-spec-v2.md:1260-1278` has the same stale two-family narrative, with retention percentages given for only two families, not updated alongside the corrected table elsewhere in the same document. Same root cause as #2: the generalization from 2 families to 4 was applied carefully in most places and missed a few.

### Real bugs in the new dev/ scripts

These are exploratory, never-shipped probe scripts (per this repo's own convention), so lower stakes than a shipped-tool bug, but they're committed and other developers will run and trust them.

**4. Retry-After header absence silently becomes `0`, not the intended fallback, in 4 files.** `Number(res.headers.get("retry-after"))` with no `?? default` before the conversion: `Number(null) === 0`, and `0` is finite, so `Number.isFinite(ra) ? ra : fallback` never reaches the fallback branch when the header is simply absent (common — RFC 7231 makes it optional). Confirmed directly: `Number(null)` is `0`, not `NaN`. Affects `explore-relative-role-classifier-records.ts:133`, `explore-tree-require-switch.ts:50`, `explore-year-bands-tree.ts:56`, `explore-year-range-sweep-records.ts:67` — each retries almost immediately instead of backing off, worsening the exact throttling they exist to handle. Three sibling files avoid this correctly by writing `Number(header ?? fallback)` — merge the default before converting, matching those three.

**5. `explore-relative-role-classifier-records.ts`'s retry counter never resets on a successful page**, unlike every sibling script (which explicitly reset it). Scattered, ordinary 429s across a run's several pages accumulate past the abort threshold and kill the whole script even though every individual page eventually succeeded.

**6. `explore-father-absent-principals-records.ts` has no 429 handling at all** — any non-2xx response is treated as fatal (`process.exit(1)`), unlike all 9 sibling scripts hitting the same or a related endpoint, which all retry 429 with backoff because (their own comments say) this API throttles hard.

**7. Unanchored substring regexes for sex/role classification can over-match compound roles.** `/Father/i` and `/Mother/i` in `probe-search-qualifiers.ts`'s new `mother` finder and `ROLE_OF`/`couldBe` (feeding the new `mothersIndexed`/`fathersIndexed` denominators this PR introduces) match "Grandfather," "Godmother," "Stepmother" just as readily as the real thing. Confirmed directly: `/Mother/i.test("Godmother")` is `true`. This file's own docblock documents that `display.role` is a free-text field with compound values in the wild ("Father Of Groom"), so this isn't a hypothetical.

**8. `probe-search-qualifiers.ts`'s new `mother` finder duplicates the pre-existing `father` finder verbatim** (only Male/Female and Father/Mother swapped), even though this same diff introduces a correctly parameterized `couldBe(wrongSex)` helper for the analogous pair a few lines later. The three-signal precedence this duplicates has already caused one silent bug here per the file's own comments; a future fix to it now has to be applied by hand in two places.

**9. `sectionR()`'s crash-safety pre-seed was extended to the wrong scope.** The legacy spouse verdict is pre-seeded before the function's early-return (`rows.length === 0`) specifically so a failed enumeration can't silently wipe it — but the new father/mother/parent verdicts this PR adds are only ever written *after* that early return, with no equivalent pre-seed. A future re-run where all population baselines fail to enumerate (documented as expected, recurring API flakiness) would silently drop those four new keys from the committed evidence artifact, while the spec cites all four by name.

**10. Section S — whose whole documented purpose is independently challenging Section R's father/spouse finding — still reads the old sex-blind `parentsIndexed` for its own "father" family**, not the new `fathersIndexed` this PR introduces specifically to stop conflating father-nameless with father-silent. Its justifying comment ("there is no mother detection on Persona") is now false. Not caught by any test; the PR body explains why Section F and Section T were deliberately left alone but never mentions Section S, so this looks like a genuine miss rather than an accepted exception.

### Minor / cosmetic

- `tool-description-length.test.ts`'s docstring claims `person_search.birthYearExact` is 209 characters and third-longest. Independently re-measured (three different ways, across multiple review passes): it's 178 characters and isn't even third — `record_search.birthPlaceExact` (181) is. Doesn't break CI (the actual executable assertions are correct), but it's the same "wrong about its own basis" failure the file's own docstring says has already happened three times, now uncaught a fourth.
- A docblock on `mothersIndexed` overstates the empirical stakes of the fix — recomputing the pre-fix numbers from data already in the same commit shows the cited verdict flip (HOLDS → DOES NOT HOLD) wouldn't actually have happened under the code's own tolerance check. The fix itself is real and worthwhile; only this one sentence about its consequence doesn't hold up.
- `birthPlaceExact` and its four "As `birthPlaceExact`" dependents dropped a caveat that ordering confidence was only ever checked on one example — it survives in the spec but not in the schema description the model actually reads at call time.
- Reuse: the retry/backoff/sleep/per-request-token pattern is hand-duplicated across most of the 10 new `dev/` scripts rather than factored into one shared helper (a working `backoff()` already exists in `probe-search-qualifiers.ts` and none of the new scripts call it) — this is very likely the root cause of findings 4-6 above, since a shared helper would make the fix apply everywhere at once. Also: `EXEMPT`/`EXEMPT_SCOPE` are two parallel maps keyed by the same value instead of one; three independently-maintained filename arrays classify which surfaces get which check, with nothing enforcing they stay consistent; `DOCUMENTED_LONGEST`/`DOCUMENTED_TOTALS` hardcode recomputable rankings that must be pasted in by hand on any unrelated description edit — the exact mechanism behind the cosmetic finding above.

### Verdict

Solid, careful work overall — six self-review rounds are visible in the commit history and most of the generalization from 2 families to 4 was done correctly and even caught its own prior bugs along the way. The rebase/eval-collision item is the one that actually blocks merge. The doc-drift and dev-script items are real but narrow, same-shape fixes. Nothing here is a design problem with the shipped tool descriptions themselves.

chesworthrm:
Addressed the review. Branch is now merged up to current `main` and re-evaluated. Summary by finding:

**Blocks-merge**
- **#1 (stale + eval collision):** merged `origin/main` (clean; the repo's merge convention rather than rebase), then re-ran `make eval-skill SKILL=search-records`. Fresh run log `v1_2026-08-21_00-13-34` — snapshot matches the merged tree (Rule 2), candidates auto-pruned back to 5, `.ann` recorded for the review sample. `check_runlogs.py` passes locally (all rules satisfied). The PR-body "touches no skills/" line was indeed wrong — `SKILL.md` is edited here.

**Should fix (2→4 family doc drift)**
- **#2 / #3:** generalized the relative-narrowing prose from father/spouse to father/spouse/mother/parent across `place-date-mechanics.md`, `SKILL.md`, and `record-search-tool-spec-v2.md`, all sourced from `dev/measured-figures.json`, with the mother/England caveat (no real-name control there). Swept the skill + spec and fixed **two more** instances of the same drift in `search-strategy-levers.md`.

**dev/ script bugs**
- **#4 / #5 / #6:** extracted a shared `dev/http-retry.ts` (`retryAfterMs` handles absent *and* HTTP-date headers; `fetchRetry`'s per-call counter kills the never-reset bug; father-absent gains 429 handling). Migrated the five affected scripts.
- **#8 / #9 / #10:** unified the father/mother finders on one `provablyOfSex` predicate; pre-seeded all four `.exact requires the relative to be present` verdicts (not just spouse); switched Section S's father family to `fathersIndexed` and fixed its stale "no mother detection" comment.
- **#7:** analyzed and left as-is — the `/Father/i` `/Mother/i` substrings are a reliable sex indicator across English role compounds and only run on confirmed ParentChild parents as a last-resort fallback; anchoring would degrade it.

**Minor / cosmetic**
- Corrected the `tool-description-length.test.ts` docstring figures (birthYearExact 178 not 209; birthPlaceExact is third) and the false "CEILING is the midpoint" claim; softened the `mothersIndexed` docblock's overstated verdict-flip; restored the `birthPlaceExact` ordering caveat to the model-read description; consolidated `EXEMPT`/`EXEMPT_SCOPE` into one map; added a proven-to-fail guard that EVIDENCE/AGENT surfaces stay disjoint.

**Eval result: 26/28.** Both failures (`ut_search_records_012`, `_014`) are #1786's new `test_pre1880_census_structure_marked_inferred` validator, which arrived via the merge — **not** this PR. `012` fails in #1786's own committed run log on main too; `014` is run-to-run variance. This PR's edits don't touch the Step 4 census guidance. Flagged for follow-up in #1642.

chrisedeson (approved):
Eval collision: confirmed genuinely fixed. Ran check_runlogs.py myself against the merged commit, exactly 5 candidates for search-records, all rules satisfied. The 2 failing tests in the re-run aren't new: one fails identically in main's own committed run log from #1786 already, the other is the pre-1880 census validator's own documented coin-flip case, not a regression.

Doc drift: spot-checked the father/mother finder unification (now one shared provablyOfSex predicate instead of duplicated code) and the sectionR pre-seed (now genuinely covers all four families, not just spouse). Both correct.

On the one pushback (leaving /Father/i and /Mother/i unanchored): agreed. No English compound crosses sex, so the original worry doesn't actually apply here.

Suite green, typecheck clean. Approving.

jud-sdev (approved):
I went through this by running and measuring rather than just reading, and it holds up well.

The `*Exact` rule reads correctly now that it's stated once per tool: without a flag a name field also matches fuzzy spellings and keeps records where that field is empty — for `givenName` and the relative names, but not for `surname`, where an unqualified value drops the surname-empty ones. `person_search` is careful to say this is the record-index engine's behaviour, measured there and not on its own endpoint. The per-parameter one-liners line up with what was actually measured — initials in one direction only, mother keep-on-two-populations and drop-on-one, other not measured, years left outside the rule.

Both specs agree with the tool text, and the cross-surface drift from the last round is gone — `SKILL.md`, the record-search spec, and `place-date-mechanics.md` all state the four-family enumeration the same way now. I checked each one on this head.

The new description-length lint's numbers are real — I measured the descriptions myself and got 238, 237, 226 for the longest three, nothing non-exempt over the 250 ceiling, and birthYearExact correctly exempt with its removal condition. The ceiling sits where the docstring says, between the longest one-liner and the smallest old offender.

The measured-figures change is a genuine fix, not just a tidy-up: the old exemption was keyed by value alone, so an exemption meant for one surface quietly excused the same number everywhere else. Scoping it closes that. And the new empty-field figure is carried honestly — exempted by name with its dev-script provenance and a re-run note, instead of a faked verdict.

All checks green, up to date with main, and the blocking notes from the earlier round are resolved here. Nothing to send back.

johnmarkpeterbrown (commented):
Genealogist pass on this one — I focused on what the research agent is told to do, rather than on the probe scripts. The measurement work holds up. I checked the mother and parent claims against the artifact one at a time and they are right, including the detail that mother's keep is enumerated on two populations and its drop on one, because England had no mother given name to draw a control from. The initials clause matches its verdict. The retention percentages match their populations. That is careful work and I could not fault it.

Three things about the *guidance* need fixing before this merges, and one is a one-line typo fix.

---

### 1. The `person_search` example says `surnameExact` drops surname-less persons. Your own rule says the unqualified value already did.

In `docs/specs/person-search-tool-spec.md`:

You wrote:

```
initials-only forms — that is not established. It **does** drop persons with no
surname recorded, per the rule above: measured on the record index, not on this
endpoint.
```

Change it to:

```
initials-only forms — that is not established. It does not change what happens to
persons with no surname recorded: per the rule above, the unqualified `surname`
value has already dropped those, and `surname` is required on every search — so
`person_search` cannot reach a surname-less tree person at all, flag or no flag.
```

The rule block higher up the same file says the drop happens *without* `exact=on`. So the flag adds nothing on that leg.

This is the one I would most want changed. A researcher looking for an ancestor recorded without a surname reads this and clears `surnameExact` to keep them. It will not help. `person_search` rejects any call without a surname, so on the tree endpoint those persons are unreachable full stop. `record_search` has three anchors, so the record side has a way out — drop the surname, anchor on the country — which is how your own probe found "Escolastica". The tree side has none, and `init-project` goes through `person_search` whenever the user has no FamilySearch ID.

### 2. `place-date-mechanics.md` still asserts place descent as a bolded "Key insight", and your new spec calls it unverified.

In `packages/engine/plugin/skills/search-records/references/place-date-mechanics.md`:

You wrote:

```
- **With `.exact=on`:** Still descends to child localities, but does
  NOT expand upward.
...
**Key insight:** Exact place does NOT prevent matching child
localities — it prevents matching parent localities.
```

Change it to:

```
- **With `.exact=on`:** Does NOT expand upward. Whether it still descends to
  child localities is recorded by no verdict — treat it as unverified.
...
**Key insight:** Exact place stops matching parent localities. Do not assume it
keeps child localities; that half is unmeasured.
```

I know the body defers this. The reason given is that editing the file arms a paid eval run — but this PR already edits that file, and the runlogs gate already armed on it. I read the job log to check. So the cost of fixing it here is nothing. Whether places are in scope is still the lead's ruling to make, but it should be made on the real price.

Why it matters for research: `SKILL.md` tells the agent to set `<event>PlaceExact` exactly when it is about to record `results_available` or argue a search was reasonably exhaustive. If descent does not actually happen, that count excludes every town-level record, and the agent argues exhaustiveness from a number that missed most of the county.

### 3. The surname/given-name split — the finding this PR exists for — never reaches `SKILL.md`, and the line beside it says the case does not arise.

In `packages/engine/plugin/skills/search-records/SKILL.md`:

You wrote:

```
For the searched person's own name, which the index
virtually always holds, that collapses to "must match" — a consequence of the
index, not a separate measurement.
```

Change it to:

```
The searched person's own two name fields behave OPPOSITELY, measured 2026-08-20:
an unqualified `givenName` keeps records with no indexed given name, but an
unqualified `surname` drops records with no indexed surname outright. So a record
indexed with a given name only — infant burials, foundlings, enslaved people,
patronymic entries — cannot be reached by ANY query carrying a surname. To reach
one, drop `surname` and anchor on `recordCountry` or `batchNumber` instead.
```

You state the split in both tool descriptions and both specs, and you edit the line two above this one — but the skill still tells the agent that principal-name silence is a non-case and not a separate measurement. It is one now, and the population is real.

These are the records genealogists actually chase: infant burials, foundlings, enslaved people, patronymic-only entries, the "Escolastica" your own probe turned up. The way out exists on this endpoint and you used it. It is written down nowhere the agent will see it. Your initials block a few lines up is the model for how well this file does this kind of thing — it even carries the false-nil warning — which is why the gap stands out.

### 4. One-line fix: the length docstring contradicts the constant below it.

In `packages/engine/mcp-server/tests/packaging/tool-description-length.test.ts`:

You wrote:

```
 * quoted 209; re-measured it is 178, below `birthPlaceExact`'s 181, so
```

Change it to:

```
 * quoted 209; re-measured it is 178, below `birthPlaceExact`'s 226, so
```

`birthPlaceExact` is 226. `DOCUMENTED_LONGEST` says so and asserts it, and I measured it from the built schemas to be sure. Nothing in either tool is 181. The same figure appears at two other spots in the file. The 181 came from the last review round and it was wrong. Flagging it because the file says being wrong here is the same class of defect as guessing the threshold, and this is now the fifth instance in a comment that catalogues four.

---

### Three body statements that no longer match the diff

Not merge blockers, but #1771 is being planned off this body.

- **"this PR touches no file under `packages/engine/plugin/skills/`, so `check_runlogs.py` does not arm"** — it touches three, and the gate armed. The workflow matches both `packages/engine/plugin/skills/` and `docs/`. The `search-records` re-run was forced, not voluntary. Anyone reading this before #1771 will expect no run-log obligation and hit a red required check.
- **"a net saving of about 182 characters"** — it is 135. The lint's own asserted totals give it. The body says the figures live in the lint and should not be repeated, then repeats a different one.
- **"`mother*`, `parent*`, `other*` remain assumed … and say so"** — mother and parent are measured now and the descriptions correctly say so. Only `other*` is still assumed. The diff is better than the body claims.

---

Leaving this as a comment rather than a review state — the eval re-run question flagged as open in the body is a spend decision for the lead, and I would rather not gate on it.

johnmarkpeterbrown (changes requested):
Left a genealogist pass on #1773. The measurement work is solid — I checked the
mother/parent claims against the artifact one at a time, including the keep-on-two
/ drop-on-one asymmetry, and they're right. Nothing to send back there.

Three guidance fixes though, all with exact replacement text in the comment:

1. The person_search spec example credits surnameExact with dropping surname-less
   persons — your own rule block says the unqualified value already did. Matters
   because surname is required on that endpoint, so those persons are unreachable
   either way, and the example points a researcher the wrong direction.

2. place-date-mechanics.md still asserts place descent as a bolded "Key insight"
   while your new spec calls it unverified. I know the body defers this on the
   grounds that editing the file arms a paid eval run — but the PR already edits
   that file and the gate already armed. I checked the job log. So it's free to
   fix here. Whether places are in scope is still the lead's call.

3. The surname/given-name split — the finding the PR exists for — doesn't reach
   SKILL.md, and the line two below your own edit still says principal-name
   silence is a non-case and "not a separate measurement". It is one now.

Plus a one-line typo: the length docstring says birthPlaceExact is 181; it's 226,
and DOCUMENTED_LONGEST twelve lines down asserts 226. The 181 came from the last
review round.

Also three body statements have drifted from the diff — flagged them because
#1771 is being planned off that body. The "check_runlogs.py does not arm" one is
the load-bearing mistake: it armed, and the search-records re-run was forced
rather than voluntary.

johnmarkpeterbrown (commented):
Round 3. All four items from the last pass landed, and they landed properly.

I checked the 24/28 rather than take the commit message's word for it. It holds: all three fails are `test_pre1880_census_structure_marked_inferred`, it hits a different test set every run (010/012/015, then 012/014, then 010/012/013), and it fails identically in main's own committed run log from #1786. You're at parity with main, not worse. The 028 partial is judge-driven with no validator failure. Candidate set is at 5, and the `.ann` review sample matches what the commit says.

One thing you got right that my finding got wrong: I flagged the `181` at three spots and you fixed one. That was the correct call. Lines 71 and 104 narrate what a past commit did, so `181` is accurate history there. Line 45 was the present-tense claim. Good catch on my sloppiness.

Three things left. Two are free; one costs a run, and that's the reason to batch them.

---

### 1. The claim we just removed from `SKILL.md` is still in the same skill's reference doc — and it's blunter there

`packages/engine/plugin/skills/search-records/references/place-date-mechanics.md`, in the "required means must not contradict" bullet:

You wrote:

```
  - For a field the index virtually always carries — the searched person's
    **own** given name and surname — there is nothing to be silent about,
    so required collapses to *must match*. One term that does not match
    collapses the result set — measured, a few hundred results fell to
    single digits on a gibberish given name.
```

Change it to:

```
  - For the searched person's **own** surname, required collapses to *must
    match*: measured 2026-08-20, an unqualified `surname` DROPS records with
    no indexed surname. But its own given name does NOT — an unqualified
    `givenName` KEEPS records with no indexed given name, so it behaves like
    the relative fields below. One term that does not match still collapses
    the result set — measured, a few hundred results fell to single digits on
    a gibberish given name.
```

"There is nothing to be silent about" asserts the population does not exist. Your own probe measured that it does — roughly 251 given-name-empty rows retained in a 6,038 pool, and a bound surname-empty record in a 1,100-row set read to the end. This is the file the agent loads on demand for place and date questions, and `SKILL.md` now tells it the opposite.

I swept the whole repo for this sentence rather than the one file this time. This site and item 2 are the only two left.

**This one is snapshot-tracked, so it invalidates the run log you just paid for.** That is the reason to take all three of these in one commit rather than one at a time — items 2 and 3 are free, and batching them means one re-run instead of two.

### 2. Same sentence in the record spec — free, not snapshot-tracked

`docs/specs/record-search-tool-spec-v2.md`, in the `m.queryRequireDefault` table row:

You wrote:

```
For the searched person's own name, which the index virtually always holds, that
collapses to *must match*, which is why a nil result means one of the terms **on
the person you searched** did not match.
```

Change it to:

```
For the searched person's own surname this collapses to *must match* — an
unqualified `surname` drops records with no indexed surname (measured 2026-08-20)
— while an unqualified `givenName` keeps given-name-empty records, so the two
principal fields differ; see *The exact-match rule* above. Either way a nil
result means one of the terms **on the person you searched** did not match.
```

Lower stakes since it is developer-facing, but this row now contradicts the exact-match rule about 900 lines above it in the same document.

`docs/` is not in the run-log snapshot, so this edit is free — it arms the scope check but the snapshot still matches.

### 3. The body now defers work it actually did

"Six review rounds, and four things deliberately left" still opens with:

> **`place-date-mechanics.md` still asserts place descent flatly** … Not fixed here because that path matches `check_runlogs.py:736`, so editing it arms a paid `make eval-skill` run … It wants its own PR alongside #1325.

You fixed it two commits ago, and the gate armed and the run happened regardless. That bullet should come out, otherwise it reads as pending work to whoever picks up #1771.

---

For the record, since approvals here are not dismissed on push: I read the full delta from `5b6ce9b0` to `743c39fe`. It is the four guidance fixes plus the fresh run log, nothing else, and I verified each — including re-deriving `birthPlaceExact` at 226 from the built schemas, and the flaky-validator history above. Suites green on this head: engine 2488/2488, typecheck clean, harness 2522 passed.

Nothing else outstanding from me after these three.

johnmarkpeterbrown (approved):
Looks good!

PR #1774, opened 2026-08-20
fix: retire per-skill tool deny in the eval harness

### LINKED ISSUE #1748: Retire the unit harness's per-skill tool deny; keep test_tool_allowlist as an after-the-fact declaration check

yinkid28:
Closes #1748.

## Summary

- The eval harness derived a per-skill deny list as the complement of
  `allowed-tools` frontmatter, but that field is a **grant** (not a
  restriction). Deriving a deny inverted its documented meaning, and
  neither production path narrows per skill.
- `skill_runner.py` now grants every registered MCP tool to every
  skill, matching production. `DISALLOWED_BACKSTOP` (Bash, WebFetch,
  etc.) remains.
- `test_tool_allowlist` is now advisory: it warns on undeclared tools
  via Python `warnings.warn()` instead of asserting. No run-log
  `output.warnings[]` entry — no downstream consumer was planned, and
  `warnings.warn()` is simpler. The warning is visible in pytest output
  and the harness stderr.
- `test_tool_allowlist` now unions `attempted_mcp_calls` alongside
  `tool_calls`, so calls denied by the subagent-only policy, the
  max_tool_calls cap, or cut off by an abort are also checked.
- `compute_allowed_tools` body is unchanged — it still computes the
  declared set for the advisory validator and the `ValueError` guard on
  `run_skills`.
- `allowed_tools_override` is kept on `run_skill` as an explicitly
  ignored no-op, marked `# IGNORED` at the parameter and explained in
  the docstring. It is the seam the acceptance test needs: without a
  restrictive list to pass, the test takes the standalone-permissive
  path and passes on main too. The orchestrator still passes
  `skill_baseline` into it; nothing reads it.
- Docs updated: architecture.md §5/§5.1, unit-test-spec.md §8/§15,
  check_tool_coverage.py wording. Stale "Known risks" bullet about
  falling back to disallowed_tools complement deleted.

## Deviated from the plan

- Used `warnings.warn()` instead of run-log `output.warnings[]` entries
  for the advisory validator. Rationale: no downstream consumer is planned
  for structured undeclared-tool warnings; `warnings.warn()` is simpler,
  visible in pytest output and harness stderr, and avoids adding a new
  warning kind to the run-log schema.

## Acceptance check

- `test_session_grants_every_registered_mcp_tool` exercises the real
  `run_skill` via a monkeypatched `query`, reads the `ClaudeAgentOptions`
  the SDK receives, and asserts every `LIVE_TOOLS` entry is in
  `allowed_tools` and none is in `disallowed_tools`. It passes
  `allowed_tools_override=["Read"]`, the restrictive list the
  orchestrator sends. On main `run_skill` honours it and the first
  `LIVE_TOOLS` entry checked fails with "missing from the session
  grant". On this branch the parameter is ignored and every tool is
  granted, so it passes.

## Didn't change

- No paid `make eval-skill` run. The change widens what every skill can
  reach, but `test_ownership_table` and `test_tree_ownership_table` still
  assert (the ownership net holds), and `image_read` is not in `LIVE_TOOLS`
  with `subagent_only_violation` still guarding it on the main thread. The
  SDK-advertising risk flagged in #1748 remains untested until someone runs
  a suite.

## Test plan

- [x] `cd eval/harness && uv run pytest tests/ -q` — 2463 passed, 2 skipped
- [x] `test_session_grants_every_registered_mcp_tool` passes (exercises real
  `run_skill` + `ClaudeAgentOptions`)
- [x] `test_tool_allowlist_validator_warns_instead_of_failing` passes
- [x] `test_tool_allowlist_warns_on_attempted_mcp_calls_only` passes
- [x] Advisory warnings visible in pytest output (confirmed in test run)
- [x] `allowed_tools_override` is referenced in exactly three places:
  the `run_skill` signature, the orchestrator call site, and the
  acceptance test that proves it has no effect


promise-emmanuel (changes requested):
Review: PR #1774 — 1748-retire-tool-deny, full pass, 6 commits behind main
Does: Removes the unit harness's per-skill tool narrowing. skill_runner.run_skill now grants BASELINE_ALLOWED + every mock-registered MCP tool and sets disallowed_tools to DISALLOWED_BACKSTOP alone; the extra_disallowed complement is deleted. test_tool_allowlist becomes advisory (warnings.warn instead of assert). compute_allowed_tools keeps computing the declared set for the advisory validator and the run_skills ValueError guard. Docs updated in architecture.md §5/§5.1 and unit-test-spec.md §8/§15.

Intent: Matches issue #1748 in substance — the inverted grant-read-as-deny is genuinely gone, and the six protected denies are untouched. Two explicit issue requirements are missing (below).

Checks: All required checks pass on ce099d4f — pytest, runlogs, e2e-fixtures, vitest, lockfile-drift, scan. None missing.

Still needs: *.py and *.md both route to senior-developers and senior-genealogists in CODEOWNERS; both are requested, gh api .../codeowners/errors is clean. Two approvals needed, at least one senior.

Verified:

eval/harness full suite on the test-merge with main: 2394 passed, 8 skipped, 1 failed. The one failure (test_mock_mcp.py::test_successful_call_carries_no_is_error) is environmental — a fresh worktree has no packages/engine/mcp-server/build/. After linking node_modules and running npm run build, that file is 26 passed. Not a regression.
Mutation test on the new grant test — details below.
Mutation test on the advisory validator test: reinstating the assert in test_tool_allowlist makes test_tool_allowlist_validator_warns_instead_of_failing go red. That test is real.
Existing reviews: None. No inline comments, no issue comments, no unresolved threads. This is the first round.

[CRITICAL] (10/10) eval/harness/tests/unit/test_allowed_tools.py:413 — the acceptance test cannot fail
The test reimplements skill_runner's two lines inside its own body and asserts against its own copy. It never imports run_skill. The code it claims to protect is not on its path.

Proof, two ways:

I restored main's skill_runner.py verbatim (git show origin/main:… > …, confirming extra_disallowed was back on line 359) and ran the test: 1 passed.
With the full narrowing reinstated, the entire harness suite ran 2400 passed, 2 skipped. Nothing in the harness catches the regression this PR exists to prevent.
This is also why the PR has no acceptance check that fails on main.

I wrote and verified a replacement using the fake_query pattern already in tests/unit/test_skill_runner.py:192. It reads the real ClaudeAgentOptions the SDK receives. On this branch: 1 passed. Against main's skill_runner.py: fails with AssertionError: mcp__genealogy__extraction_append missing from the session grant / assert 'mcp__genealogy__extraction_append' in ['Read'].

Replace the body of test_session_grants_every_registered_mcp_tool with this:


def test_session_grants_every_registered_mcp_tool(tmp_path, monkeypatch):
    """The SDK session grants every registered MCP tool regardless of what
    the skill declares (issue #1748). Reads the real ClaudeAgentOptions."""
    import asyncio
    from claude_agent_sdk import ResultMessage
    import harness.skill_runner as sr
    from harness.auth import AuthConfig
    from harness.mock_mcp import LIVE_TOOLS
    from harness.skill_runner import DISALLOWED_BACKSTOP

    class _Stream:
        def __init__(self, messages):
            self._m, self._i = messages, 0

        def __aiter__(self):
            return self

        async def __anext__(self):
            if self._i >= len(self._m):
                raise StopAsyncIteration
            self._i += 1
            return self._m[self._i - 1]

        async def aclose(self):
            return None

    seen = {}

    def fake_query(**kw):
        seen["options"] = kw["options"]
        return _Stream([
            ResultMessage(subtype="result", duration_ms=1, duration_api_ms=1,
                          is_error=False, num_turns=1, session_id="S1")
        ])

    monkeypatch.setattr(sr, "query", fake_query)
    asyncio.run(sr.run_skill(
        user_message="go", workspace=tmp_path,
        fixture_names=[], fixtures_dir=tmp_path,
        auth=AuthConfig(skill_runner_mode="api_key", api_key="x", detail="stub"),
        allowed_tools_override=["Read"],
    ))
    options = seen["options"]

    for name in LIVE_TOOLS:
        q = f"mcp__genealogy__{name}"
        assert q in options.allowed_tools, f"{q} missing from the session grant"
        assert q not in options.disallowed_tools, (
            f"{q} in disallowed_tools — per-skill narrowing is back"
        )
    assert sorted(options.disallowed_tools) == sorted(DISALLOWED_BACKSTOP)
Passing allowed_tools_override=["Read"] is load-bearing: it also proves the override is genuinely ignored.

[CRITICAL] (8/10) eval/harness/validators/test_universal.py:533 — the union the issue asked for is missing
Issue #1748 says: "Keep the union the issue was originally filed for: read output.tool_calls union the attempted calls in output.warnings[].attempted[] (kind: "uncovered_tool_call")."

grep -n "attempted\|uncovered_tool_call" eval/harness/validators/test_universal.py returns nothing. The validator still iterates tool_calls only.

That was the original filed defect: a denied call never reaches tool_calls, so the validator is blind to it. Retiring the deny removes most of that blindness, but not all — calls stopped by the max_tool_calls cap, by the subagent-only policy, and calls cut off by a wall-clock or stream-silence abort still land in attempted[] and never in tool_calls. I confirmed attempted_mcp_calls is still produced (orchestrator.py:391, :926).

The PR body doesn't mention omitting this, and the PR says Closes #1748.

I measured the change rather than guessing at its size: 3 files, about 10 lines. run_validators has exactly two production call sites, and result is already in scope at orchestrator.py:443.

In eval/harness/harness/validator_runner.py, add the parameter alongside blocked_protected_writes:


You wrote:      skills_invoked: list[dict[str, Any]] | None = None,
Change it to:   skills_invoked: list[str] | None = None,
                attempted_mcp_calls: list[dict[str, Any]] | None = None,
and the namespace entry:


You wrote:          "test": test or {},
Change it to:       # MCP calls the model emitted that never reached the mock —
                    # denied by the subagent-only policy or the max_tool_calls
                    # cap, or cut off by an abort. Never in `tool_calls`.
                    "attempted_mcp_calls": attempted_mcp_calls or [],
                    "test": test or {},
In eval/harness/harness/orchestrator.py:443, pass it:


You wrote:          skill=spec.skill,
Change it to:       skill=spec.skill,
                    attempted_mcp_calls=result.attempted_mcp_calls,
Then take attempted_mcp_calls in test_tool_allowlist and union its tool names into the set you check.

If you'd rather not build it here, that's a fine call — but then say so in the PR body and don't close #1748 with it.

[INFO] (9/10) docs/specs/unit-test-spec.md:1966 — a doc that tells the next person to reinstate the bug
The Known risks bullet still reads:

permission_mode="dontAsk" must actually block unlisted tools. The harness relies on this SDK setting to enforce per-skill allowlists at call time … If the SDK regresses, fall back to disallowed_tools populated as the complement of the per-skill allowlist.

That fallback is the mechanism this PR retires. Issue #1748 listed it for removal by name. Delete the bullet.

Two more in the same file:

docs/specs/unit-test-spec.md:1688 contradicts the section three lines below it:


You wrote:      - `allowed_tools` — **per-skill, derived from the skill's SKILL.md frontmatter** (see below). Combined with `permission_mode="dontAsk"`, this enforces the tool allowlist at execution time rather than only catching violations after the fact.
Change it to:   - `allowed_tools` — the filesystem baseline plus every registered MCP tool (see below). No per-skill narrowing: `allowed-tools` frontmatter is a grant, not a restriction, and the `test_tool_allowlist` validator reports undeclared calls after the fact.
eval/harness/validators/test_universal.py:537 points at the wrong section — §13.5 is "Positive test with scenario_notes: research-plan":


You wrote:      warnings for the reviewer. See unit-test-spec.md §13.5.
Change it to:   warnings for the reviewer. See unit-test-spec.md §15,
                "Deriving `allowed_tools` per skill".
[INFO] (7/10) eval/harness/harness/skill_runner.py:317 — a parameter that is passed and silently ignored
allowed_tools_override stays in the signature, and orchestrator.py:787 still passes skill_baseline into it. Nothing reads it. The next person who reaches for it to restrict a test will get no error and no effect.

Delete the parameter and the orchestrator.py:787 line. skill_baseline stays where it is — compute_allowed_tools is still called for the run_skills ValueError guard, which I confirmed is live at allowed_tools.py:71.

If you keep it for compatibility instead, make it loud:


You wrote:      `allowed_tools_override` is accepted for backward compatibility but
                ignored — the session now grants every registered MCP tool (issue
                #1748).
Change it to:   `allowed_tools_override` is dead. Nothing reads it; the session
                grants every registered MCP tool (issue #1748). Passing it has no
                effect — do not use it to restrict a test.
[INFO] (6/10) PR body — the template is gone
The body has no Start here, Unsure about, Didn't change, Acceptance check, or Deviated from the plan, and none of the Test plan checkboxes. PLAN.md is gitignored, so those lines are the only form the plan reaches a reviewer in.

Acceptance check is the one that would have caught the finding above: there is no test here that fails on main and passes on this branch.

Two things the issue asked for that belong in the body:

The lead's ruling was "records an undeclared_tool_call warning on the run log and the suite prints a per-skill count at the end." You chose warnings.warn() instead and gave a reason, which is fine — put it under Deviated from the plan so the lead sees it as a decision rather than an omission. Note the practical cost: I reproduced a five-test suite run and the warnings do print every time (the validator module is re-imported per call, so Python's dedup never kicks in), but none of them names the test it came from. Triaging the 142-test migration from that output means matching bare tool lists against nothing.
The issue asked you to name which skill you ran make eval-skill on and what its tool mix looked like, as the only available check on whether the SDK filters the advertised tool list. That line isn't in the body.
What I did not check
No paid eval run. eval/harness/** is outside the run-log snapshot, so nothing here flips a run log inactive and none is needed to merge — but that also means the SDK-advertising risk the issue flagged is untested and stays untested until someone runs a suite.
I ran the harness suite once, not repeatedly, so I can't speak to flakiness.
I did not run the web workspace or engine suites. Nothing in the diff touches them, and CI's vitest and server-pytest are green.
One thing I checked that came out clean, since the change widens what every skill can reach: tree_edit, tree_correct and the other write-capable live tools are now granted to every skill, and both test_ownership_table and test_tree_ownership_table still assert. The ownership net holds. image_read is not in LIVE_TOOLS, and subagent_only_violation still guards it on the main thread.

yinkid28:
@promise-emmanuel Addressed all five findings:

**CRITICAL #1 — Acceptance test that can't fail**
Replaced `test_session_grants_every_registered_mcp_tool` with a `fake_query`-based test that monkeypatches `sr.query`, calls the real `run_skill`, and asserts against the actual `ClaudeAgentOptions` the SDK receives. Every `LIVE_TOOLS` entry must be in `allowed_tools` and absent from `disallowed_tools`.

**CRITICAL #2 — `attempted_mcp_calls` union**
Threaded `attempted_mcp_calls` through `validator_runner.py` → `orchestrator.py` → `test_tool_allowlist`. The validator now unions `tool_calls` with `attempted_mcp_calls` before checking declarations, so denied/aborted MCP calls don't slip past the advisory check.

**INFO #3 — Docs pointing at the retired mechanism**
- Deleted the `permission_mode="dontAsk"` Known-risks bullet in `unit-test-spec.md`
- Updated the `allowed_tools` description from per-skill enforcement to the permissive grant
- Fixed `§13.5` → `§15, "Deriving allowed_tools per skill"`

**INFO #4 — Dead `allowed_tools_override` parameter**
Removed from `run_skill` signature and the `orchestrator.py` call site.

**INFO #5 — PR body template**
Already updated in a prior push.

All 2386 tests pass (1 pre-existing failure from an untracked `temp_test_cli.py` file).

Asimi1234:
Holding on this — the open "changes requested" round isn't resolved, and I checked that its main point still stands.

**Blocker: the acceptance test doesn't fail on main.** Driving the real `run_skill` is the right fix for the "reimplements the logic" note. But the test calls `run_skill` without an override, so on main it takes the permissive standalone path and passes there too. I ran the shipped test against a full `origin/main` harness and it passes — so the body's claim ("on main this test fails with `extraction_append` missing from the session grant") can't actually happen with this test.

Two asks:
1. Make it genuinely fail on main. Simplest: `assert "allowed_tools_override" not in inspect.signature(run_skill).parameters`. Better: assert the orchestrator grants all tools.
2. Fix the Acceptance-check line in the body to match the shipped test.

**Smaller: the `attempted_mcp_calls` union has no test.** It's wired correctly, but nothing covers the case it exists for — a tool that shows up only in `attempted_mcp_calls` (denied/aborted), never in `tool_calls`. I removed the union (`all_calls = list(tool_calls)`) and the whole suite still passed. Add a case where the undeclared tool is only in `attempted_mcp_calls` and assert it warns.

Everything else looks right: grant-all is correct, `DISALLOWED_BACKSTOP` is intact, docs are thorough, and CI is green (pytest, vitest, runlogs, e2e-fixtures, lockfile-drift, scan). Sign-off routes to senior-developers + senior-genealogists (`*.py`).

yinkid28:
@Asimi1234 Both points addressed in edf73c4c:

**Blocker — acceptance test now discriminating.**
The test passes `allowed_tools_override=["Read"]` to `run_skill`, simulating the restrictive list the orchestrator sends. On main, `run_skill` uses it to derive `extra_disallowed` and LIVE_TOOLS entries land in `disallowed_tools` — test fails. On this branch the parameter is explicitly ignored — test passes. `allowed_tools_override` is kept as a documented no-op (promise-emmanuel's finding said "remove **or** loudly mark") specifically so the test has a seam to prove narrowing is gone. Updated the docstring and the PR body's acceptance-check line accordingly.

**Smaller — `attempted_mcp_calls` union now tested.**
Added `test_tool_allowlist_warns_on_attempted_mcp_calls_only`: an undeclared `secret_tool` appears only in `attempted_mcp_calls` (never in `tool_calls`), and the test asserts it warns. Removing the union (`all_calls = list(tool_calls)`) makes this test fail.

Full suite: 2461 passed, 3 skipped, 1 pre-existing failure (untracked `temp_test_cli.py`).

Asimi1234 (approved):
Approving. Both points from my earlier comment are addressed, and I verified them by running against this branch (`edf73c4c`):

- **Acceptance test is now discriminating.** With `main`'s `skill_runner.py` swapped in, `test_session_grants_every_registered_mcp_tool` goes red (`assert 'mcp__genealogy__merge_warnings' in ['Read']`); on this branch it passes. The `allowed_tools_override` no-op seam is a clean way to prove the narrowing is gone.
- **`attempted_mcp_calls` union is now covered.** Dropping the union (`all_calls = list(tool_calls)`) makes `test_tool_allowlist_warns_on_attempted_mcp_calls_only` fail as expected; with the union in place it passes.

Grant-all is correct, `DISALLOWED_BACKSTOP` intact, all required checks green.

One nit, non-blocking: the Acceptance-check line still names `extraction_append` as the tool that goes missing on main, but the assertion actually trips on `merge_warnings` first. Either drop the tool name or update it.

Note this is a peer approval — merge still needs senior-developers + senior-genealogists sign-off (`*.py` via CODEOWNERS).

promise-emmanuel (commented):
The code is done — both blockers are fixed and I checked each from the other
side. Against origin/main's skill_runner.py the acceptance test fails with
"mcp__genealogy__tree_correct missing from the session grant". Deleting
"+ _attempted" from test_universal.py:566 turns the union test red. Suite in
your worktree: 2463 passed, 2 skipped.

I am holding the approval on the PR body only, because three of its claims now
say the opposite of the diff and the body is the record of what this PR did.
Push the four edits below and I will approve straight away.

1. Summary bullet.

You wrote:      - `allowed_tools_override` parameter removed from `run_skill` — it was
                  accepted but ignored. The orchestrator call site that passed
                  `skill_baseline` into it is also removed.
Change it to:   - `allowed_tools_override` is kept on `run_skill` as an explicitly
                  ignored no-op, marked `# IGNORED` at the parameter and explained in
                  the docstring. It is the seam the acceptance test needs: without a
                  restrictive list to pass, the test takes the standalone-permissive
                  path and passes on main too. The orchestrator still passes
                  `skill_baseline` into it; nothing reads it.

2. Acceptance check — name no specific tool, since the assertion loops over a set
and the first failure is arbitrary.

You wrote:      On main (with per-skill narrowing), this test fails with `mcp__genealogy__extraction_append
                missing from the session grant`. On this branch it passes.
Change it to:   It passes `allowed_tools_override=["Read"]`, the restrictive list the
                orchestrator sends. On main `run_skill` honours it and the first
                `LIVE_TOOLS` entry checked fails with "missing from the session
                grant". On this branch the parameter is ignored and every tool is
                granted, so it passes.

3. Test plan — this box is checked and it is not true; three references remain.

You wrote:      - [x] No `allowed_tools_override` references remain in the codebase
Change it to:   - [x] `allowed_tools_override` is referenced in exactly three places:
                  the `run_skill` signature, the orchestrator call site, and the
                  acceptance test that proves it has no effect

And while you are in the Test plan, the count is stale — I measured 2463 passed,
2 skipped on 5b9e493b.

One doc line left over, in a file you already edited. docs/architecture.md:1258
still tells a reader the unit harness blocks MCP tools:

You wrote:      `bypassPermissions` — chosen over `dontAsk` precisely so declared `Write`/`Edit` still work, while out-of-allowlist MCP tools stay blocked
Change it to:   `bypassPermissions` — chosen over `dontAsk` so declared `Write`/`Edit` still work. No MCP tool is blocked: every registered tool is granted, and `test_tool_allowlist` only warns (§5.1)

Optional, your call: orchestrator.py:788 passes a restrictive list into a
parameter documented as ignored. I removed it and the suite was unchanged at
2463 passed, so nothing needs it. Line 788 is the only use of `skill_baseline`
inside `_execute_skill_with_retry`, so removing it cleanly means dropping that
parameter and its call site at line 365 as well. Leave it if you would rather
not touch the signature — the acceptance test catches the regression either way.

yinkid28:
@promise-emmanuel All four edits done:

1. **Summary bullet** — updated to describe `allowed_tools_override` as a kept no-op with the seam rationale.
2. **Acceptance check** — reworded to describe the override mechanism without naming a specific tool.
3. **Test plan** — fixed the `allowed_tools_override` checkbox to name the three surviving references; updated count to 2463 passed, 2 skipped.
4. **architecture.md:1258** — unit-harness permission-mode cell now says "No MCP tool is blocked" instead of "out-of-allowlist MCP tools stay blocked."

PR body updated, architecture.md committed in e9d4fa86.

Skipped the optional cleanup (removing `skill_baseline` from the orchestrator call site) — the acceptance test catches the regression either way and the smaller diff is easier to review.

EdmondOware (approved):
Approving. Ran my own mutation tests rather than relying on the two prior reviews, and both acceptance tests genuinely discriminate.

Swapping in main's skill_runner.py makes test_session_grants_every_registered_mcp_tool fail with project_context missing from the session grant, asserted against a bare Read list. Worth noting the tool named in that failure differs across all three of us, promise-emmanuel got tree_correct, Asimi1234 got merge_warnings, I got project_context. That is exactly why insisting the body name no specific tool was right, the loop's first failure is arbitrary, and the body now correctly names none.

Removing the union, dropping plus underscore attempted, makes test_tool_allowlist_warns_on_attempted_mcp_calls_only fail with expected a warning about secret_tool from attempted_mcp_calls, got empty. Restored both and the file is 29 of 29 green.

The change is what CLAUDE.md prescribes rather than a judgement call. The section on the eval harness emulating production's permission model states it verbatim, allowed-tools is a grant not a restriction, a deny derived as the complement of a grant inverts the field's meaning, and the unit harness still derives one, retiring it is open work. This closes that named gap.

The safety side is what I looked hardest at, since this widens what every skill can reach, and it holds three ways. DISALLOWED_BACKSTOP is intact at Bash, WebFetch, WebSearch and NotebookEdit, and disallowed_tools is now exactly that with nothing added or removed. The two ownership validators still hard assert, they were not softened alongside test_tool_allowlist, so the net that actually stops a skill writing sections it does not own is untouched. And image_read is genuinely absent from LIVE_TOOLS, so the subagent only guard still covers it on the main thread. That combination matches CLAUDE.md's own boundary rule, denies production genuinely has stay, derived complement denies go.

All four body edits and the leftover architecture.md line are in, and the doc now reads that no MCP tool is blocked and test_tool_allowlist only warns.

One residual risk stated honestly rather than buried, whether the SDK actually filters the advertised tool list stays untested until someone runs a paid suite. The harness directory is outside the run log snapshot so none is owed to merge.

No blocking defects.

promise-emmanuel (approved):
Approving

PR #1775, opened 2026-08-20
eval-harness: recompute violations, surface abort-path spend, add eval-inventory (#1484)

### LINKED ISSUE #1484: Corpus report undercounts guardrail violations and hides abort-path e2e spend

clack391:
## Summary

Normal tier. Closes #1484. Three offline defects in the e2e corpus report, all free to fix (no API calls, no live runs):

- **Undercounted violations.** `corpus_report` read violations from the stored field only, so every run written before the detector counted zero. Adds `--recompute`, which re-derives them by calling `check_guardrail_compliance` over each run's committed `tool_calls` + `final-research`/`final-tree` sidecars + the fixture seed tree. The stored path stays the default.
- **Hidden abort-path spend.** Every abort stored `total_cost_usd: null`, so the ledger read as if timeouts were free. Adds a separate, clearly-approximate `total_cost_usd_estimated` (flat per-token table in the new SDK-free `e2e/pricing.py`), estimated at report time and write time, with its measured accuracy printed beside it. `total_cost_usd` itself stays null. The spend line reports recorded / estimated / unrecoverable as three numbers, never blended.
- **Drifting counts.** Adds `make eval-inventory`, printing the six corpus counts that drift in architecture.md prose, each defined by a printed predicate.

## Review guidance

**Start here:** `recompute_tally` in `corpus_report.py`, and the function move. The move puts the single definition of `check_guardrail_compliance` in `harness.skill_invocation` (beside the three detectors it composes) and re-exports it from `orchestrator`, so `corpus_report` can import it without pulling `claude_agent_sdk` into its pure-analysis posture. Worth checking: the recompute's skip discipline (no readable seed tree -> named in a skip list, excluded from the recomputed count, never zero), and that the spend split never blends the three figures into one.

**Unsure about:** nothing.

## Plan

**Changed after review (Gennecis):** `docs/architecture.md` now has its three drifted §9.1/§9.3 counts corrected and pointed at `make eval-inventory` (374->396 test defs, 104->106 fixtures, and the matching "82 of the 396" line), so they stop drifting. #1507, cited in the first cut as the reason to defer, has been closed since 2026-08-10, so fix-in-PR is the default. Guardrails kept: the two `145`-run measurement denominators (:408, :1010) are dated snapshots and were left, and the :1373-1374 parenthetical (107 dirs, 106 with a fixture.json) is still correct and untouched.

**Didn't change:** the recomputed number stays out of `compliance`, no axis flipped, no run log rewritten, no detector graduated, and the default window / judge / fixtures / calibrate_judge untouched (per the issue's out-of-scope list).

**Acceptance check:** `test_recompute_recovers_violations_a_stored_only_read_counts_as_zero` and `test_recompute_skips_a_run_whose_fixture_has_no_seed_tree` (plus the estimator, spend, and `_fallback_usage` tests) fail on `main` and pass here. Each new guard was broken and watched red before being trusted.

**New baseline (why the numbers differ from the issue's table):** the issue pinned 145 run logs / 631 violations / 104 fixtures on 2026-08-10; the corpus has since grown to 157. Per the issue's own instruction ("if the header reports more than 145 runs, record the new baseline rather than changing the code to hit these numbers"), current `--recompute --since all` prints recomputed **673 across 156 scanned**, mentor-verdict arm **0->40**, and **exactly one named skip** (`william-ferber-ancestry/run-2026-07-28_13-59-05.json`). `make eval-inventory` prints **396 / 25 / 106 / 157 / 133 / 57**. Spend: recorded **$1,142.74** over 133, estimated **$122.62** over 11, unrecoverable **13**; calibration median **0.90x** (range 0.01-1.13), matching the lead's 2026-08-10 measurement. The recompute inherits the ledger-truncation ceiling the 2026-08-16 comment flagged (some assigned ids were never written down); it reads the committed final-state sidecars directly rather than replaying, and documents that ceiling in the spec.

**Deviated from the plan:** no `scripts/eval-inventory.bat` (post-#1437 the Windows wrappers live in `scripts/windows/`, and no e2e/eval analysis target has one); the spec edit spans three sections (§8.1.2, §6 economics, the totals/Scoping block) rather than one. Both decided during plan review.

## Test plan

- [x] Ran the full harness suite (`cd eval/harness && uv run --frozen pytest -q`): **2396 passed, 8 skipped, 1 failed**. The single failure is `test_mock_mcp.py::test_successful_call_carries_no_is_error`, the pre-existing missing-engine-build environmental case, not this change. ruff clean; `--since all` output byte-identical above the added spend line. `make test-all`'s JS/web suites are not exercised by this Python/Makefile/docs change; CI runs them.
- [ ] N/A. No skill run-log snapshot changed (no skill dir, delegated agent, `eval/tests/unit/<skill>/`, or referenced scenario/fixture touched; this is a harness/report change).
- [ ] N/A. No skill `description` frontmatter or DO NOT clause changed.

## Follow-on issues

**Folded in / filed instead:** none filed. Folded in during review: the three drifted `docs/architecture.md` counts, corrected and pointed at `make eval-inventory`.


docs/specs/e2e-test-spec.md
```diff
@@ -532,13 +532,20 @@ expensive enough (§12) that an accidental sweep is a real cost event; driving a
...
+no `total_cost_usd`** (all timeouts, plus a few `error`/`inactivity`), so the
+**$1,142.74** recorded total spans 133 runs while the wall-clock spans all 157.
+Two ways fill the gap, and they are complementary, not rival: the pooled $/min
+imputation used here historically (~$0.1504/min), and the flat per-token estimate
+`make e2e-corpus`'s spend line now applies (`e2e/pricing.py`), which recovers the
+11 null-cost runs that still carry token counts (≈$123) and names the 13 that
+carry neither as unrecoverable rather than imputing them. Prefer the per-token
+estimate where token counts survive — it is calibrated (median ~0.90x recorded,
+`make e2e-corpus CALIBRATE=1`); the $/min figure is the fallback for the
+tokenless tail. Budget a suite at the mean, and run `make e2e-corpus SINCE=all`
```
Gennecis: **Two problems in this paragraph, both checkable offline.**

**`:538` — "the wall-clock spans all 157" is not true.** Only 144 of 157 runs carry `usage.duration_ms`, and the 13 that don't are *exactly* the 13 this same paragraph calls unrecoverable:

```
runs with NO duration_ms:            13
unrecoverable (no cost, no tokens):  13
unrecoverable AND no duration_ms:    13
```

**`:545-546` — "the $/min figure is the fallback for the tokenless tail" therefore can't be followed as written**, because `duration_ms` is null for every run in that tail. The minutes *are* recoverable, but from `usage.timeline`, which the paragraph never names — and only for 10 of the 13 (698 min across those 10, ≈$105 at the pooled rate). The other 3 carry no wall-clock signal at all.

Suggested edits:

- `:538` → "…spans 133 runs, while `duration_ms` spans 144 — the 13 runs carrying neither a cost nor token counts carry no `duration_ms` either."
- `:545-546` → "…the $/min figure is the fallback for the tokenless tail, and its minutes come from `usage.timeline` (`duration_ms` is null for every run in that tail) — which covers 10 of the 13; the remaining 3 carry no wall-clock signal and are unmeasurable by either method."

Separately: the heading still says "scale by the mean, not the median", but the rewrite dropped the worked figure the old text carried ("a 20-fixture pass is ≈ **$185 and ~20.5 h serial**"), leaving the imperative with nothing to act on. Current means are $8.59 and 59.9 min → ≈**$172 and ~20 h serial** for 20 fixtures. Please restore it with the new numbers.

eval/harness/e2e/corpus_report.py
```diff
@@ -466,6 +639,72 @@ def format_report(
...
+    """Stored vs recomputed, per arm, labelled — then every named skip.
+
+    The stored column is a floor: runs written before the detector store no
+    violations field and read `not_checked`, contributing zero. The recomputed
+    column re-derives them from committed data.
+    """
+    lines = [
+        "",
+        "  --recompute: violations re-derived from tool_calls + committed sidecars.",
+        "               stored is a floor (pre-detector runs record none); recomputed is the real count.",
```
Gennecis: **"stored is a floor" / "recomputed is the real count" is falsified by the corpus — and the counter-example disappears into the aggregate.**

Over every run that actually carries the stored field:

```
runs WITH a stored field: 19
  recompute reproduces stored exactly: 18
  recompute finds MORE:  0
  recompute finds FEWER: 1
    mary-mcandrew-son/run-2026-08-05_18-11-41.json  stored=1  recomputed=0
```

18/19 exact is strong evidence the recompute itself is right — this is a labelling problem, not a detector problem.

The cause is legitimate and worth *surfacing* rather than hiding. That run's stored violation is a `proof-conclusion` one, and commit `931fce95` (#1340, **2026-08-06 — the day after the run**) is titled "baseline the proof-conclusion compliance arm against the starting tree so it stops firing on seeded state". Today's detector correctly clears a false positive the contemporaneous one recorded. So that run's research was fine and it was gated non-compliant by a checker bug — `--recompute` is the only thing in the repo that can show that, and right now it silently absorbs the correction while printing a sentence saying that direction is impossible.

Three lines up, this same file writes "but no run is known clean, so this is a floor on incidence, not a rate", and the spec at `:1691` already says the recomputed figure "inherits that ceiling rather than being an independent re-derivation". The printed output is what contradicts both — and the printed output is what people read.

Asks:

**1.** `:674` — replace with something true in both directions:

```python
"               stored = what each run recorded at the time (pre-detector runs record none); "
"recomputed = today's detectors over the same committed data. Not the same "
"measurement: recomputed is not guaranteed >= stored.",
```

**2.** `:667-669` — drop "The stored column is a floor".

**3.** Surface the divergence instead of burying it. `RecomputeTally` already names its skips so "a shrunk denominator can never read as a clean corpus"; the same reasoning applies here. Add a `regressed` list (runs where stored > recomputed) and print it under the TOTAL line:

```
    1 run(s) where TODAY's detector clears a violation the run recorded
      (a detector correction, not a corpus change):
      mary-mcandrew-son/run-2026-08-05_18-11-41.json  stored 1 -> recomputed 0
```

eval/harness/e2e/corpus_report.py
```diff
@@ -466,6 +639,72 @@ def format_report(
...
+        f"    {'arm':<34} {'stored':>7} {'recomputed':>11}",
+    ]
+    for arm in sorted(set(stored_arms) | set(rt.arms), key=lambda a: -rt.arms.get(a, 0)):
```
Gennecis: **Arm ordering is not deterministic across processes.** Sorting a `set` by count alone leaves ties to set iteration order, which depends on the per-process hash seed. Same input, five processes:

```
PYTHONHASHSEED=1 → ['mentor', 'conflict-resolution', 'exhaustiveness']
PYTHONHASHSEED=2 → ['mentor', 'conflict-resolution', 'exhaustiveness']
PYTHONHASHSEED=3 → ['conflict-resolution', 'mentor', 'exhaustiveness']
PYTHONHASHSEED=4 → ['exhaustiveness', 'conflict-resolution', 'mentor']
PYTHONHASHSEED=5 → ['mentor', 'conflict-resolution', 'exhaustiveness']
```

Today's arms happen not to tie (487/60/49/40/36/1), so the output is stable **by luck**. The first tie makes this report undiffable, and byte-identical output is the regression check #1484 named.

```python
for arm in sorted(set(stored_arms) | set(rt.arms), key=lambda a: (-rt.arms.get(a, 0), a)):
```

This also matches the file's own convention — `_counts` at `:447` sorts its non-ordered remainder alphabetically for exactly this reason.

eval/harness/e2e/pricing.py
```diff
@@ -0,0 +1,70 @@
+"""Flat per-token price table + a cost estimator for abort-path e2e runs.
+
+Issue #1484 (b): every aborted run (`_fallback_usage`) stores
+`total_cost_usd: None`, so the corpus ledger reads as if 24 runs were free and
+hides ~$196 of real spend — non-randomly, since every `timeout` run is among
```
Gennecis: `~$196` isn't reproducible by anything in this PR, and the sentence pairs it with the *current* run count ("24 runs"). This module's own method recovers **$122.62** over 11 runs and calls 13 unrecoverable; `$196` is the issue's 2026-08-10 `$/min` imputation over a then-23-run population — different method, different date, no label saying so.

CLAUDE.md: *"A measurement that disagrees with belief is re-measured, not reworded."*

Re-measured today, the honest combined figure is ≈**$228**: $122.62 per-token over the 11 runs with token counts, plus ~$105 for the 10 tail runs whose minutes survive in `usage.timeline` (698 min at the pooled $0.1504/min). 3 runs are unmeasurable by either method.

Either drop the dollar figure — the command prints the live one, which is this module's whole point — or date and attribute it.

eval/harness/e2e/pricing.py
```diff
@@ -0,0 +1,70 @@
...
+    "input_tokens": 3.00,
+    "output_tokens": 15.00,
+    "cache_read_input_tokens": 0.30,
+    "cache_creation_input_tokens": 6.00,
+}
+
+# The token fields this table prices — also the presence test for "has a token
+# block at all". A usage block carrying none of these (the 13 pre-fallback runs
+# with no token counts) is unrecoverable and must estimate to None, never 0.
+PRICED_FIELDS = tuple(_PER_MTOK)
```
Gennecis, line 48: `PRICED_FIELDS` is defined and documented as "the presence test for 'has a token block at all'", but referenced nowhere:

```
$ grep -rn "PRICED_FIELDS" --include=*.py --include=*.md .
eval/harness/e2e/pricing.py:46:PRICED_FIELDS = tuple(_PER_MTOK)
```

Meanwhile `orchestrator.py:1002` `_USAGE_FIELDS` lists the identical four names, and `_fallback_usage` builds its token block from `_USAGE_FIELDS` then prices it with `_PER_MTOK`. They agree today; if they ever diverge, `_fallback_usage` emits a block the estimator prices at 0 → returns `None` → **abort-path spend is invisible again, which is the defect this PR exists to fix.** Nothing checks it.

The fix is at `orchestrator.py:1002`, outside this PR's hunks — hence the comment here:

```python
_USAGE_FIELDS = pricing.PRICED_FIELDS
```

`orchestrator` already imports `pricing` at `:59`, the tuples are byte-identical today so it's a behavioural no-op, and it makes the export honest instead of decorative. `_USAGE_FIELDS` is only read at `:1037` and `:1062`, and no test pins the literal.

Gennecis (changes requested):
**Request changes** — six items, all small. Proportion first, because the engineering here is good and I checked it rather than taking it on trust.

**Reproduced exactly, on this branch:** `make eval-inventory` → 396 / 25 / 106 / 157 / 133 / 57; `--recompute --since all` → 673 across 156 scanned, mentor-verdict arm 0 → 40, exactly one named skip (`william-ferber-ancestry/run-2026-07-28_13-59-05.json`); spend recorded $1,142.74 / 133, estimated $122.62 / 11, unrecoverable 13; calibration median 0.90x, range 0.01–1.13. Every figure in the PR body lands to the cent.

**The new guards are load-bearing.** I mutated the implementation three ways at runtime — `--recompute` falling back to the stored field, recompute with no skip discipline, and the estimator returning `0.0` instead of `None` — and all three new tests went red. Full suite: **2403 passed, 2 skipped, 0 failed** (cleaner than the PR's own claim; the environmental `test_mock_mcp` failure doesn't reproduce here). `ruff` clean; the `encoding="utf-8"` lint passes over all new I/O.

**Byte-identity holds by construction**, not just by observation: `tally`, `format_report` and `_counts` are untouched and every new print is strictly after `format_report`.

**A worry I cleared rather than raised:** recomputing against today's fixture seed tree would be wrong if a fixture's `starting-tree.gedcomx.json` had changed after its runs were recorded. Across all 93 fixtures with runs, zero seed trees post-date their earliest run log. Not a present defect — recording it so it isn't re-litigated.

**And the strongest signal in the PR:** the recompute reproduces the stored field *exactly* on 18 of the 19 runs that carry it. That is real evidence the detectors are being **reused** rather than re-composed by hand, which is what #1484 was emphatic about. The `check_guardrail_compliance` move is right for the reason given, and I confirmed nothing else in `orchestrator.py` used the three imports that were dropped.

The five inline comments are about labelling and measurement honesty, which in this repo is not a nit. One item can't be inline because the file isn't in the diff:

---

### `docs/architecture.md` — the stated reason for leaving it doesn't hold

The PR body says it "is under active stewardship (#1507)". **#1507 has been closed since 2026-08-10**, ten days before this PR opened:

```
#1507 closed — docs: make architecture.md a stable file —
               move the §9.4/§10 registers to the board, purge line cites
```

The drift is live, three numbers wide, and admitted in the PR body itself:

| line | says | `make eval-inventory` |
|---|---|---|
| `:1367` | "**374** committed test definitions" | 396 |
| `:1373` | "104 fixtures" | 106 |
| `:1467` | "82 of the 374 definitions" | 396 |

CLAUDE.md's "Work you find along the way" makes fix-in-this-PR the default and requires a deferral to name a reason that applies. Fixing three numbers and pointing them at `make eval-inventory` is item (c)'s entire purpose: "a count that cannot be reproduced by a command is a count that will be wrong in the next doc PR."

Two guardrails on that edit:

- **Do not touch `:408` or `:1010`.** Those use 145 as a *measurement denominator* ("`Agent` is **3.8% of all tool calls** across the 145 runs"). They are dated snapshots of a measurement, not inventory counts — repointing them at `make eval-inventory` would be wrong, and bumping 145→157 without re-measuring the percentage would be worse.
- `:1373-1374`'s parenthetical ("`eval/tests/e2e/` holds one more directory that is not one") is **still exactly right** — 107 directories, 106 with a `fixture.json`. Leave it.

If the lead would rather this stay out of scope, that's his call — but the PR body then needs a reason that is true rather than a closed issue.

---

One follow-up commit should clear all six: three are one-liners, two are prose, and only the `--recompute` divergence line adds code.

clack391:
Thanks for the thorough pass and for reproducing it all. Addressed all six. I re-derived every number rather than taking the review on trust, and they match yours.

**1. architecture.md counts.** Fixed in this PR. #1507 has been closed since 2026-08-10, so the deferral reason no longer holds and fix-in-PR is the default. :1367 374->396, :1373 104->106, :1467 "82 of the 396", each pointed at `make eval-inventory`. Left the two 145-run measurement denominators (:408, :1010) and the :1373-1374 parenthetical (107 dirs, 106 with a fixture.json) untouched, per your guardrails. PR body updated to drop the stale stewardship claim.

**2. e2e-test-spec suite-economics.** :538 now reads "spans 133 runs, while `duration_ms` spans 144; the 13 runs carrying neither a cost nor token counts carry no `duration_ms` either" (verified: the 13 with no `duration_ms` are exactly the 13 unrecoverable). :545-546 now names `usage.timeline` as the tail's minute source, covering 10 of 13, the remaining 3 unmeasurable. Restored the worked figure with today's means ($8.59 and 59.9 min, so about $172 and ~20 h serial for 20 fixtures).

**3. "stored is a floor" / "the real count".** Dropped both, in the printed output and the docstring. The recompute output now names the regressed run under TOTAL:
```
    1 run(s) where TODAY's detector clears a violation the run recorded
      (a detector correction, not a corpus change):
      mary-mcandrew-son/run-2026-08-05_18-11-41.json  stored 1 -> recomputed 0
```
A new test pins this, and also pins recompute-finds-more as not a regression.

**4. Arm ordering.** Added the alphabetical tiebreak `(-count, arm)`, matching `_counts`. A new test asserts deterministic order; I break-tested the old count-only key and it failed 7 of 8 hash seeds.

**5. "~$196".** Dropped the stale figure from pricing.py; it drifts with the corpus, and `make e2e-corpus SINCE=all` prints it live (recorded / estimated / unrecoverable). For the record, re-measured today the recoverable hidden spend is about $228: $122.62 over the 11 token-bearing runs, plus about $105 over the 10 timeline-bearing tail runs at the pooled $0.1504/min; 3 carry no signal and are unmeasurable.

**6. PRICED_FIELDS.** orchestrator.py:1002 is now `_USAGE_FIELDS = pricing.PRICED_FIELDS`, so the fallback token block and the estimator that prices it share one definition and cannot silently diverge. Byte-identical today, a behavioural no-op.

Verification: full harness suite 2398 passed (the single failure is the engine-build-absent `test_mock_mcp`, environmental as you noted). ruff clean, encoding lint passes. The default report output is unchanged by construction: `format_report`/`tally`/`_counts` are untouched and every new print is in the `--recompute` path.

Gennecis (approved):

EdmondOware (approved):
Ran the full harness suite myself: 2424 passed, 3 skipped, 0 failed, clean.

Ran make eval-inventory myself: 396 / 25 / 106 / 157 / 133 / 57, matches the corrected architecture.md counts exactly.

Ran make e2e-corpus SINCE=all RECOMPUTE=1 myself. Every headline figure reproduced to the letter: 673 recomputed across 156 scanned, mentor verdict arm 0 to 40, the single named skip on william ferber ancestry, the single detector correction on mary mcandrew son with stored 1 to recomputed 0, and the spend split of 1142.74 recorded, 122.62 estimated, 13 unrecovered.

Read the doc corrections directly in architecture.md. The 396, 106, and 82 of 396 numbers all now point at make eval-inventory, and the surrounding prose is honest about the two counts being able to diverge rather than treating today's exact match as permanent.

Good back and forth on this one. The six item review was itself verified with mutation testing on the new guards and a seed tree drift check across all 93 fixtures, and the one residual issue found after that, stale stored is a floor language surviving in help text and spec prose, was caught and fixed directly. Nothing left to flag.

PR #1777, opened 2026-08-20
e2e-test-spec: describe caps as the loader implements them (#1720)

### LINKED ISSUE #1720: Spec section 3.1 says per-fixture caps do not exist; the loader and 11 fixtures use them — and stribling-father-1821 needs one

aghadiayeamayanvboernest:
Closes #1720. Batches #1116 (same paragraph, one wrong number).

## The drift

Spec §3.1 said, in bold, that stop-condition limits are **not** a fixture field — "they live as defaults in the orchestrator, not in `fixture.json`. Every fixture uses the same caps; an author never writes them."

The loader has always disagreed. `load_fixture` reads `fixture_json.get("caps")` and fills only the *omitted* keys from `FixtureCaps()`, for all seven caps (`eval/harness/e2e/orchestrator.py:646-659`). The dataclass states the design outright: *"The DEFAULT caps every fixture inherits for any cap it doesn't set."*

And **12** committed fixtures rely on it — re-derived rather than quoted, since the number moves with every fixture that lands:

```sh
for f in eval/tests/e2e/*/fixture.json; do
  python3 -c "import json,sys; d=json.load(open(sys.argv[1],encoding='utf-8')); \
    print(sys.argv[1].split('/')[3], sorted(d['caps'])) if d.get('caps') else None" "$f"
done
```

The doc was the outlier. `susan-miller-birth` even documents its raised wall-clock cap and the timed-out run that motivated it — precisely the argument §3.1 forbade an author from making.

## The claim appeared three times, not once

Fixing only §3.1 would have left the spec contradicting itself, so all three go:

| Site | Was |
|---|---|
| §1 overview bullet | "(caps are harness defaults, not fixture metadata)" |
| §3.1 | the bold "**NOT a fixture field**" paragraph |
| §6 step 5 | "the same for every fixture, not authored per-fixture" |

§3.1 gains a `caps` row beside `blocked_tools` naming all seven keys.

## Two decisions in the wording, both deliberate

**1. The row points at `FixtureCaps` instead of listing the values.** That dataclass declares itself the single source of truth ("these are the single source of truth — `load_fixture` fills omitted caps from here"). A copy of `7200 / 15.0 / 300` in the spec would be correct today and stale the next time someone tunes a cap — the same mechanism that produced this issue.

**2. For #1116 I removed the number rather than correcting it.** The spec said `caps.max_continue_nudges (default 5)`; the code sets **20**. #1116 asked to write 20.

I didn't, and this is a deliberate deviation worth objecting to if you disagree. The sentence was wrong *because* it held a copy of a value that lives in code — writing `20` fixes today's number and rebuilds the identical trap. Instead I carried across the reason the code gives, which #1116 explicitly asks for so nobody tightens the cap back: generous by design, because a full GPS proof yields after each of ~10+ sub-skill steps, and the no-progress guard is the real backstop against a genuinely idle agent.

Reverting to `(default 20)` is a one-line change if a reviewer prefers the literal fix.

## Two traps stated while in there

Both verified in the loader, not inferred:

- **An unrecognised key inside `caps` is silently ignored.** The loader reads the seven names it knows and fills the rest from defaults, and nothing validates `fixture.json` — `check_e2e_fixtures.py` doesn't look at keys. A typo'd cap looks applied and is not.
- **`usage.caps` in a run log records six of the seven.** `max_continue_nudges` is not in that block; `usage.continue_nudges` records how many were spent.

I nearly shipped that second one wrong — my first draft said the run log records "the caps a past run actually used." I checked the writer (`orchestrator.py:2026-2033`) and found six. Corrected before commit, because a near-true claim in this spec is the exact defect the PR is closing.

## One deviation from the ruling's wording

The ruling says a new override must give its reason in the fixture README, "following `susan-miller-birth`'s precedent."

**`susan-miller-birth` does not do that.** Its reason lives in `fixture.json`'s `notes`; its README doesn't mention caps at all (`grep -n "cap" eval/tests/e2e/susan-miller-birth/README.md` → nothing).

So the rule is written exactly as ruled — README, going forward — but the citation says what that fixture actually did. Writing "as `susan-miller-birth` does" would have put a false example into the spec, which is the defect class this issue exists to close.

@DallanQ — flagging in case you'd rather the rule follow the existing practice (`notes`) instead of the README.

## Scope held

No `fixture.json` touched. No backfill, no ceiling, no validator for unrecognised keys — all four excluded by the ruling. `stribling-father-1821`'s README loses exactly one clause: the one citing the sentence this PR deletes.

**A code-side pointer was considered and left out.** The cleanest way to stop this drift recurring from the other direction is a comment above `FixtureCaps` naming §3.1 as where the field is documented — so the next person editing either side sees the other. I left it out because the ruling scopes this to one file, and `eval/harness/e2e/orchestrator.py` is Python that other open work touches. Worth a follow-up only if someone thinks the pointer earns its own review; the spec now names the dataclass, so the link exists in one direction already.

## Verification — and what it does and does not prove

**Nothing checks spec prose.** `make engine-test` proves one thing about this change, and it is the thing most likely to bite: `doc-links.test.ts` holds an exact ratchet of `#NNNN` references per contract doc, and this spec is pinned at 10. Acceptance here is human review, not a command.

What I ran:

| Check | Result |
|---|---|
| All 7 loader cap names present in the new row — **derived from `caps_raw.get(...)` by script**, not eyeballed | pass |
| `grep -c -o '#[0-9]\{3,5\}'` on the spec | 10, unchanged |
| No default *value* copied into the changed prose | none |
| All three false assertions gone | none remain |
| `make engine-test` | 102 files, 2438 tests |
| `make test-all` | all checks passed |

I also broke the ratchet on purpose to confirm it can fail — appending one `#9999` reds it with `docs/specs/e2e-test-spec.md: 11 issue refs, baseline 10 — state the fact, not the ticket`. That's why every reason above is written as a mechanism and not a ticket number.

## Sequencing

PR #1747 is open on this same file about nine lines below the §6 site — adjacent, not overlapping. Will rebase if it lands first.


clack391 (approved):
Approving. Senior developer review.

For a documentation fix the claims are the change, so I verified each against `eval/harness/e2e/orchestrator.py` rather than reading the prose:

- `load_fixture` reads `fixture.json`'s `caps` and fills each omitted key from `FixtureCaps`, for all seven caps; the seven names in the new §3.1 row are exactly those seven. An unrecognised key is silently ignored (the loader reads only the seven it knows, and nothing validates `fixture.json`), so the new row's warning is accurate.
- `usage.caps` in a run log records exactly six caps; `max_continue_nudges` is not in that block. The §6 edit states this precisely.
- `max_continue_nudges` defaults to 20 in code. The spec's old "(default 5)" was wrong. Removing the number and naming `FixtureCaps` as the source is the right fix, not writing "20": copying a value that lives in code is the exact drift this PR closes, and #1116's reason is carried across as prose.
- All three false "not a fixture field" assertions are gone (§1 overview, §3.1, §6 step 5), no fourth remains, and no default value was copied into the spec. The `#NNNN` ratchet stays at 10.
- The stribling README loses exactly the one clause citing the deleted sentence; no `fixture.json` and no other README touched.

Both flagged deviations are the better call: removing the #1116 value rather than writing it, and correcting the `susan-miller-birth` citation to `notes` (that fixture does not document caps in its README, so "as susan-miller-birth does" would have planted a false example). Good verification behind both.

Scope note, not a change request: the ruling's headline says "one file," but the three-site fix plus the one README clause are authorized by the task-reviewer comment on the issue, and fixing only §3.1 would have left the spec self-contradictory. Worth naming that comment in the description so a reader of the ruling alone does not read it as scope creep. Minor: the PR carries no label; the issue is `developer`.

This approval covers the developer lane. The stribling README sits under `eval/tests/`, which CODEOWNERS routes to senior-genealogists, so merge still needs a senior-genealogist approval for that path, plus a second approval overall.

EdmondOware (approved):
Verified this independently against the actual code rather than relying on the developer lane review alone.

Confirmed FixtureCaps in orchestrator.py has exactly the seven fields named in the new spec row, and its own comment states it as the single source of truth in almost the same words the spec now uses. Confirmed load_fixture fills all seven omitted keys from those defaults.

Confirmed usage.caps in a run log genuinely records only six of the seven keys, max_continue_nudges is absent from that block. The spec edit states this precisely.

Confirmed max_continue_nudges defaults to 20 in code, not 5, and the new prose describing it as generous by design with the no progress check as the real backstop closely paraphrases the code's own comment. Removing the stale number rather than writing 20 is the right call, a literal fix would have rebuilt the same trap.

Confirmed check_e2e_fixtures.py has no validation over caps keys at all, so the silently ignored typo warning is accurate.

Counted the fixtures directly, 12 use caps, matching the PR exactly.

Checked susan-miller-birth directly. Its wall clock reasoning lives in fixture.json notes, and its README has zero mentions of caps. The correction to cite notes rather than claim README precedent is right, the ruling's literal wording would have planted a false example.

Confirmed the doc-links ratchet for this spec stays at 10, and the stribling-father-1821 README change removes exactly the one clause citing the deleted sentence.

Nothing left to flag.

PR #1778, opened 2026-08-20
Lint: ban issue references in SKILL.md and agent bodies (#1620)

### LINKED ISSUE #1620: Lint: ban issue references and historical narration in SKILL.md and agent bodies
> **Scope note added 2026-08-14 (board pass): the cleanup half is not free, and
> the body does not say so.** Removing the eleven violations edits SKILL.md
> bodies in **init-project, question-selection, search-records and
> person-evidence**, plus `agents/gps-mentor.md`. Each of those paths flips its
> skill's run log inactive, so the cleanup as written buys **four paid
> `make eval-skill` runs and four annotation passes** — and three of those four
> skills have their eval slot held by an open PR today.
>
> **Do the lint half alone.** Ship it with a `(file, line)` allow-list seeded
> with the eleven known violations — the shape `gps-terminology.test.ts` already
> uses, which the "Exempt" section below already points at. That half touches no
> skill directory, costs nothing, and can land this week.
>
> **Then let each violation ride its own skill's next paid run**, dropping its
> allow-list entry in the same PR. Whoever next takes a `search-records` or
> `person-evidence` item deletes one line of prose and one allow-list row while
> they are already paying for the run.
>
> Do **not** land the lint and the cleanup together.

> **Numbers in this note have moved (re-checked 2026-08-19 at `98c1160d`).** The
> cleanup half is now **three** body-text issue references, not eleven — the
> narration half is out of scope, see below — and **one** of the four skills has
> its eval slot held by an open PR (`question-selection`, PR #1723), not three.
> The instruction stands: lint only, no skill edits.

**Touches:** packages/engine/mcp-server/tests/packaging/no-issue-refs.test.ts (new file)

Do **not** add this to `tests/packaging/skill-guidance.test.ts` — that file is the
places-guidance duplication drift lint, and open issue #1115 is queued to extend
it. A new file named for what it bans keeps both diffs readable, and matches
`no-bare-fetch.test.ts` / `no-bom.test.ts`.

CLAUDE.md bans both in these files — "No explanatory prose in a `SKILL.md` or an agent `.md`. Every line in those files is a billed prompt token on every invocation. No comments, no rationale, no note of what was tried before." Nothing checks either half.

## Evidence that a manual pass is not enough

John Mark Peter-Brown scanned all 27 `SKILL.md` files and 4 agent bodies by hand on 2026-08-14 and found nine violations across four files — dropped issue numbers in `init-project/SKILL.md`, a postmortem paragraph in `search-records/SKILL.md`, three rationale asides in `person-evidence/SKILL.md`, two in `gps-mentor.md`.

The same grep on `e71a4931` also returns two he did not report:

- `packages/engine/plugin/skills/init-project/SKILL.md:206` — `(issue #1471)`
- `packages/engine/plugin/skills/question-selection/SKILL.md:168` — `(issue #1471)`

A careful reader missing two of eleven is the argument for a lint rather than another manual pass.

## What the lint should catch

**Scope: `packages/engine/plugin/skills/*/SKILL.md` bodies and
`packages/engine/plugin/agents/*.md` bodies. Nothing else** — that is exactly
what the CLAUDE.md rule covers (27 skill bodies, 4 agent bodies).

**Do not copy `gps-terminology.test.ts`'s `discoverPluginFiles()` unchanged.** It
also walks `plugin/references/` and `skills/*/references/`, where ten copies of
`places-guidance.md` each cite upstream Claude Code issue #17741 on line 5 and
are byte-identical-pinned by `skill-guidance.test.ts`. Widening the scan there
turns a free PR into an eleven-file coordinated edit plus a sha256 re-pin:

    grep -rln 'issue #17741' packages/engine/plugin --include=*.md | wc -l

**Ban issue references in model-visible body text** — `#NNNN`, `issue #NNNN`,
`(#NNNN)`, and bare GitHub issue URLs.

**Historical narration is deliberately out of scope, and stays unguarded.** The
body's own counter-examples are still legitimate present-tense instruction
(`project-status/SKILL.md:76`, `:86`; `conflict-resolution/SKILL.md:483`), and a
warn-only vitest assertion is not an instrument — nobody reads vitest stdout in
CI, and CLAUDE.md § "A new lint must be proven to fail" rules that a check which
cannot fail reads as coverage and is worse than none. Write that decision, and
the warn-only alternative it beat, into the new test file's header comment, so
the next person finds it at the site instead of re-deriving it.

## Exempt

**YAML frontmatter comments.** The frontmatter is parsed as YAML before the body reaches the model, so `#`-comments in it cost nothing at inference. The four blocks in `agents/record-extractor.md:26`, `:55`, `agents/image-reader.md:16`, `agents/image-reader-opus.md:16` and `agents/gps-mentor.md:61` are the load-bearing warnings about listing every tool under all three server spellings and about a deny that fails open — they stay, issue numbers and all. The lint must distinguish frontmatter from body, not grep the whole file.

## Prove it fails

Per CLAUDE.md: add an `(issue #9999)` to a `SKILL.md` body, watch the check fail, remove it. A lint that cannot fail reads as coverage and is worse than none. Two specific traps here — a grep that skips the `agents/` directory, and a pattern that matches inside frontmatter and forces the exemption to be written as a blanket file skip.

## No cleanup in this PR

Seed the allow-list and change **no file** under `packages/engine/plugin/`. Get
the live list by running:

    grep -rn -E '(issue )?#[0-9]{3,5}|github\.com/[^ )]*issues/[0-9]+' \
      packages/engine/plugin/skills/*/SKILL.md packages/engine/plugin/agents/*.md

Measured 2026-08-19 at `98c1160d` that returns eight lines: five are the exempt
frontmatter comments, three are body text — two in `skills/init-project/SKILL.md`
(`issue #1475`, `issue #1471`) and one in `skills/question-selection/SKILL.md`
(`issue #1471`). Re-run it; do not trust a line number quoted here.

**Key each allow-list entry on `(relPath, trimmed line text)`, not on a line
number.** Line-keyed entries go stale in days in these files: between `e71a4931`
(2026-08-14) and `98c1160d` (2026-08-19) the two `init-project` hits moved
122 to 153 and 206 to 245, and PR #1723 is editing `question-selection/SKILL.md`
right now — a line-keyed entry would fail an unrelated PR twice (new violation +
stale entry). Keep the stale-entry check itself: an entry whose text no longer
appears in its file must fail, the way `gps-terminology.test.ts` does.

Each violation leaves when its own skill next takes a paid `make eval-skill` run
— delete the prose and its allow-list row in that same PR.

## Acceptance

- [ ] Check fails on a planted issue reference in a `SKILL.md` body
- [ ] Check fails on a planted issue reference in the **body** of
      `agents/record-extractor.md` — the file whose frontmatter is exempt. If
      that passes, the exemption was written as a whole-file skip and the lint is
      dark for that agent.
- [ ] Check passes with all five frontmatter comment blocks untouched
      (`agents/record-extractor.md:26`, `:55`, `agents/image-reader.md:16`,
      `agents/image-reader-opus.md:16`, `agents/gps-mentor.md:61`)
- [ ] Allow-list holds exactly the body-text hits the grep above returns, each
      with a reason; `git diff --stat` shows no file under
      `packages/engine/plugin/`
- [ ] A stale allow-list entry (text no longer present in its file) fails
- [ ] Runs in **`make engine-test`** (`packages/engine/mcp-server`, vitest), and
      in CI via `.github/workflows/engine-tests.yml`, whose path scope already
      includes `^packages/engine/plugin/`. **Not `make test-js`** — that is the
      pnpm workspace (web, electron, viewer-ui, schema), and `pnpm-workspace.yaml`
      excludes `packages/engine/**`, so the lint would never run there.

aghadiayeamayanvboernest:
Closes #1620.

## What was unguarded

CLAUDE.md: *"No explanatory prose in a SKILL.md or an agent .md. Every line in those files is a billed prompt token on every invocation. No comments, no rationale, no note of what was tried before."* Nothing checked it.

An `(issue #1471)` in a skill body is a token billed on every invocation to tell the model nothing. The reason belongs in the skill's spec or rubric, where a human reads it and no inference pays for it.

**A manual pass is not the alternative.** A careful reader scanned all 27 skill bodies and 4 agent bodies by hand and reported nine of the eleven hits present that day. Missing two of eleven is the argument for a check rather than another sweep.

## Lint only — no cleanup

`git diff --stat` shows **one new file** and nothing under `packages/engine/plugin/`. That is deliberate, per the issue's scope note: editing a SKILL.md body flips that skill's run log inactive, so removing the prose here would buy a paid `make eval-skill` run and an annotation pass per skill — and one of the affected skills has its eval slot held by an open PR today (`question-selection`, #1723).

The three current violations are allow-listed with reasons. Each leaves when its own skill next takes a paid run: delete the prose and its allow-list row in that same PR.

I re-derived the list rather than trusting the numbers in the issue, as it asks:

```sh
grep -rn -E '(issue )?#[0-9]{3,5}|github\.com/[^ )]*issues/[0-9]+' \
  packages/engine/plugin/skills/*/SKILL.md packages/engine/plugin/agents/*.md
```

8 hits — 5 exempt frontmatter comments, 3 body text. The line numbers had moved again since the issue's own 2026-08-19 measurement (init-project 153 → 157 and 245 → 251), which is exactly the drift that drove the keying decision below.

## Three design choices, all recorded in the file header

**1. Entries key on `(relPath, trimmed text)`, not line number.** `gps-terminology.test.ts` keys on `lineNo`; that is wrong for these files. Across five days the two `init-project` hits moved 122 → 153 → 157 and 206 → 245 → 251 while their text never changed. A line-keyed entry fails an unrelated PR *twice* — once for the "new" violation, once for the entry it just made stale.

**2. Frontmatter is excluded by computing where it ends, not by skipping the file.** YAML is parsed before the body reaches the model, so a `#`-comment there costs nothing at inference, and five of them carry the load-bearing warnings about listing every tool under all three server spellings and about a deny that fails open. Skipping whole files would make the lint **dark for the four agents that carry the most prose** — which is why the acceptance list below includes a planted ref in `record-extractor.md`'s body specifically.

**3. Scope is bodies only, not `references/`.** `discoverPluginFiles()` from the sibling lint also walks `plugin/references/` and per-skill `references/`, where ten byte-identical-pinned copies of `places-guidance.md` each cite an upstream Claude Code issue. Reusing it would turn a free PR into an eleven-file coordinated edit plus a sha256 re-pin.

## Historical narration is deliberately left unguarded

It was in the original ask. The phrasings that would catch it also match legitimate present-tense instruction (`project-status/SKILL.md`, `conflict-resolution/SKILL.md` both read that way), so the only safe form would be a warn-only assertion — and nobody reads vitest stdout in CI, so it could not fail. CLAUDE.md is explicit that a check which cannot fail reads as coverage and is worse than none.

The decision *and* the alternative it beat are written into the test file's header, so the next person finds it at the site instead of re-deriving it. Narration stays unguarded and known-unguarded.

## Proven to fail — all six acceptance cases

Not "the tests pass". Each case was run against a deliberately broken tree:

| Case | Result |
|---|---|
| Planted `(issue #9999)` in a SKILL.md body | **fails** |
| Planted ref in the **body** of `record-extractor.md`, whose frontmatter is exempt | **fails** — the exemption is not a whole-file skip |
| All five frontmatter comment blocks untouched | passes |
| A stale allow-list entry (text no longer in the file) | **fails**, with its own message |
| Allow-list holds exactly the 3 body hits the grep returns | 3 entries, one test each |
| `git diff --stat` shows no file under `packages/engine/plugin/` | none |

The second row is the one worth checking on review: if it passed, the frontmatter exemption would have been written as a blanket file skip and the lint would be dark for every agent.

## Where it runs

`make engine-test` — 103 test files / 2443 tests, up exactly 5 from 102 / 2438. In CI via `engine-tests.yml`, whose `PATTERNS` already includes `^packages/engine/plugin/`, so a plugin edit triggers it. **Not** `make test-js`: `pnpm-workspace.yaml` excludes `packages/engine/**`, so it would never run there.

`make test-all` — all checks passed.

### DIFF
--- a/packages/engine/mcp-server/tests/packaging/no-issue-refs.test.ts
+++ b/packages/engine/mcp-server/tests/packaging/no-issue-refs.test.ts
@@ -1,0 +1,205 @@
+/**
+ * No issue references in model-visible SKILL.md / agent body text.
+ *
+ * CLAUDE.md: "No explanatory prose in a SKILL.md or an agent .md. Every line in
+ * those files is a billed prompt token on every invocation. No comments, no
+ * rationale, no note of what was tried before." A `(issue #1471)` is a billed
+ * token that tells the model nothing — the reason belongs in the skill's spec
+ * or rubric, where a human reads it and no inference pays for it.
+ *
+ * A manual pass is not enough, which is the argument for this file: a careful
+ * reader scanned all 27 skill bodies and 4 agent bodies by hand and reported
+ * nine of the eleven hits present that day.
+ *
+ * ─── Two decisions recorded at the site, so nobody re-derives them ───
+ *
+ * **Historical narration is deliberately NOT checked here.** It was in the
+ * original ask and was dropped. The phrasings that would catch it also match
+ * legitimate present-tense instruction (`project-status/SKILL.md`,
+ * `conflict-resolution/SKILL.md` both read naturally that way), so the only
+ * safe form would be a warn-only assertion. The alternative it beat: a
+ * warn-only vitest expectation, rejected because nobody reads vitest stdout in
+ * CI, so it cannot fail — and CLAUDE.md's "a new lint must be proven to fail"
+ * rules that a check which cannot fail reads as coverage and is worse than
+ * none. Narration stays unguarded and known-unguarded.
+ *
+ * **Scope is bodies only — not `references/`.** `gps-terminology.test.ts`'s
+ * `discoverPluginFiles()` also walks `plugin/references/` and
+ * `skills/<name>/references/`, and it is deliberately not reused. Ten copies of
+ * `places-guidance.md` each cite upstream Claude Code issue #17741, and they
+ * are byte-identical-pinned by `skill-guidance.test.ts`, so widening the scan
+ * would turn this into an eleven-file coordinated edit plus a sha256 re-pin.
+ * The CLAUDE.md rule quoted above covers skill and agent *bodies*; that is what
+ * this checks.
+ *
+ * ─── Why entries are keyed on text, not line number ───
+ *
+ * `gps-terminology.test.ts` keys its allow-list on `lineNo`. That is wrong for
+ * these files: measured across five days, the two `init-project` hits moved
+ * 122 → 153 → 157 and 206 → 245 → 251 while their text never changed. A
+ * line-keyed entry fails an unrelated PR twice — once for the "new" violation
+ * and once for the stale entry it made stale. Keys are (relPath, trimmed line
+ * text).
+ */
+
+import { describe, expect, it } from "vitest";
+import { readFileSync, readdirSync, existsSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const here = dirname(fileURLToPath(import.meta.url));
+const engineRoot = join(here, "..", "..", "..");
+const pluginRoot = join(engineRoot, "plugin");
+
+/**
+ * `#1234`, `issue #1234`, `(#1234)`, and bare GitHub issue URLs.
+ *
+ * The 3+ digit floor is what keeps this off markdown headings (`### Step`) and
+ * off `#1` style ordinals; every issue in this repo is 3-5 digits.
+ */
+const ISSUE_REF =
+  /(?:issue\s+)?#\d{3,5}|github\.com\/[^\s)]*issues\/\d+/i;
+
+interface Match {
+  relPath: string;
+  lineNo: number;
+  text: string;
+}
+
+/** Skill and agent BODIES only. See the scope note above. */
+function discoverBodies(): { abs: string; rel: string }[] {
+  const files: { abs: string; rel: string }[] = [];
+
+  const agentsDir = join(pluginRoot, "agents");
+  if (existsSync(agentsDir)) {
+    for (const f of readdirSync(agentsDir).sort()) {
+      if (f.endsWith(".md")) {
+        files.push({ abs: join(agentsDir, f), rel: `agents/${f}` });
+      }
+    }
+  }
+
+  const skillsDir = join(pluginRoot, "skills");
+  if (existsSync(skillsDir)) {
+    for (const skill of readdirSync(skillsDir).sort()) {
+      const abs = join(skillsDir, skill, "SKILL.md");
+      if (existsSync(abs)) {
+        files.push({ abs, rel: `skills/${skill}/SKILL.md` });
+      }
+    }
+  }
+
+  return files;
+}
+
+/**
+ * The 1-based line on which YAML frontmatter closes, or 0 when the file has
+ * none.
+ *
+ * Frontmatter is parsed as YAML before the body ever reaches the model, so a
+ * `#`-comment in it costs nothing at inference. Five such comments carry the
+ * load-bearing warnings about listing every tool under all three server
+ * spellings, and about a deny that fails open — they must survive this lint.
+ *
+ * Computing the boundary (rather than skipping the whole file) is the point: an
+ * agent whose frontmatter is exempt must still be scanned below it, or the lint
+ * goes dark for exactly the files that carry the most prose.
+ */
+function frontmatterEnd(lines: string[]): number {
+  if (lines[0]?.trim() !== "---") return 0;
+  for (let i = 1; i < lines.length; i++) {
+    if (lines[i].trim() === "---") return i + 1;
+  }
+  return 0; // unterminated — treat the whole file as body rather than skip it
+}
+
+function findMatchesInFile(abs: string, rel: string): Match[] {
+  const lines = readFileSync(abs, "utf8").split(/\r?\n/);
+  const bodyStart = frontmatterEnd(lines);
+  const out: Match[] = [];
+  for (let i = bodyStart; i < lines.length; i++) {
+    if (ISSUE_REF.test(lines[i])) {
+      out.push({ relPath: rel, lineNo: i + 1, text: lines[i].trim() });
+    }
+  }
+  return out;
+}
+
+// ─── Allow-list ────────────────────────────────────────────────────
+//
+// The violations that existed when this lint landed. Each one leaves when its
+// own skill next takes a paid `make eval-skill` run — editing a SKILL.md body
+// flips that skill's run log inactive, so removing the prose here would buy a
+// paid re-run and an annotation pass per skill. Delete the prose and its row
+// together, in the PR that is already paying for that skill's run.
+//
+// This is not an escape hatch for "the lint is inconvenient": a new violation
+// belongs in the spec or rubric, not here.
+const ALLOWLIST: Array<{ relPath: string; text: string; reason: string }> = [
+  {
+    relPath: "skills/init-project/SKILL.md",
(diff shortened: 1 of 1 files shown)


clack391 (approved):
Approving. Senior developer review.

A lint has to be proven to fail, so I broke each behavior and watched it rather than reading:

- A `#NNNN` in a skill body that is not on the allow-list turns the check red with an actionable message; a `#NNNN` in YAML frontmatter does not. The exemption computes the frontmatter boundary and still scans the body beneath it, which is the trap the issue warned about (verified with a ref in both the frontmatter and the body of one file: only the body one is caught).
- A ref planted in a `references/` file does not fire: `discoverBodies()` walks only `agents/*.md` and `skills/<name>/SKILL.md`, never `references/`, so the ten `places-guidance.md` copies of the upstream issue stay out of scope.
- Removing an allow-listed line reds only that entry's "still needed" test, so a stale allow-list row cannot outlive the prose it excused.
- The allow-list holds exactly the three current body violations; `isAllowed` matches on `(relPath, trimmed text)`, so a new ref appended to an already-allowed line is still caught.
- The regex matches all four forms the issue names (`#1234`, `issue #1234`, `(#1234)`, GitHub issue URLs) and correctly rejects `### Step` headings, `#1` ordinals, and letter-bearing hex.

CI vitest is green; the local engine failures are the missing-build / `open`-package environment, not this file (it imports only `node:*` and vitest).

This is the right shape for a nothing-checks lint: it improves on `gps-terminology.test.ts` by keying on line text rather than line number (documented, because the init-project hits moved 122 to 153 to 157 across five days without the text changing), it defers the cleanup exactly as the ruling directs, and the header records the two decisions the issue asked for.

One non-blocking nit: `#\d{3,5}` would false-positive on an all-numeric hex color like `#123456`, but there are zero such tokens in any scanned body (these are prose instructions, no CSS), so it is latent, not live. Worth at most a one-line note in the header for whoever extends the lint.

This approval covers the developer lane and the senior-developer code-owner. It still needs a second approval to reach the two required.

jud-sdev (approved):
A new lint has to be proven to fire, so I did not read it — I ran its logic against broken trees.

- A #NNNN planted in a SKILL body is caught; one planted in record-extractor's body is also caught, so the frontmatter exemption is a computed boundary, not a whole-file skip. A #NNNN inside the frontmatter comment block is correctly ignored. ### Step headings, #1 / #42 ordinals, and letter-bearing hex do not match — the 3-digit floor holds.
- On the current tree the scan finds exactly the three allow-listed hits, so there are zero offenders. Keying on (file, trimmed text) means a new ref appended to an already-allowed line is still caught, and each allow-list row has a self-staleness test that reds if its prose is removed or reworded — I confirmed that fail direction.
- Discovery is complete: all 27 skill bodies and all 4 agent bodies are scanned, and the first test guards the scanned-nothing hole.
- It is genuinely enforced: the required vitest check is engine-tests.yml, it ran on this head, and the suite actually executed rather than taking the skip-with-success path.

One non-blocking note, correcting the earlier one: #\d{3,5} would match a 3-digit all-numeric short hex like #123, not #123456 (six digits, out of range). Latent only — there are no such tokens in any body today. Worth a one-line header note for whoever extends it.

This is the right shape for a nothing-checks lint, and it is my second sign-off on the senior-developer code-owner lane.

Issue #1749, opened 2026-08-19
11 validator requests from the citation deep dive


EdmondOware:
**IMPORTANT**: do issue #1748 first — it is a one-line fix to `test_tool_allowlist` in the
same file this issue extends (`eval/harness/validators/test_universal.py`), and V1/V2 here
read the same denied-call path it repairs.

## Ruling (lead, 2026-08-19): three tiers, not two

> "regex's over prose are brittle; regex validators can report findings to the LLM
> judge but only tests on structured fields should gate."

- **Tier 1 — may gate.** Checks over structured run-log fields and persisted state.
  A failure fails the test, which also skips the judge (`_compute_outcome` in
  `eval/harness/harness/orchestrator.py` returns `fail` before any dimension is read).
- **Tier 2 — reports, never gates.** Checks that are regexes over Claude's prose
  response. Their findings are handed to the LLM judge as observations it weighs
  alongside the response it is already reading, and are recorded in the run log.
  They do not touch `validators_passed`.
- **Nothing is silently dropped.** Every tier-2 finding lands in the run log whether
  or not a judge ran.

This decision beat the two alternatives it was put against: making every check gate
(brittle — a prose regex reds a correct run and the judge never sees it), and dropping
the prose checks entirely (loses the finding). **Record the reasoning in
`docs/specs/unit-test-spec.md` § "Deterministic Validators" as part of this task** —
that section currently says "All validators take the same three arguments" and knows
of one tier only, so the next person re-derives this from nothing.

### Tier per check

| Check | Tier | Reads |
|---|---|---|
| V1 write-then-validate | **1 — gates** | `before_state`/`after_state` research_json, `tool_calls` |
| V2 unbacked validation claim | **2 — reports** | `text_response` regex vs `tool_calls` |
| V3 invented sample locators — persisted half | **1 — gates** | `file_changes` after-state vs scenario before-state |
| V3 invented sample locators — response half | **2 — reports** | `text_response` regex |
| V4 skill's own example values — persisted half | **1 — gates** | `file_changes` after-state vs SKILL.md body + before-state |
| V4 skill's own example values — response half | **2 — reports** | `text_response` regex |
| V5 repository chain (**rule rewritten below**) | **1 — gates** | `file_changes` before/after `citation_detail`, tree source `author` |
| V6 unknown-marker vocabulary | **1 — gates** | persisted `citation` / `citation_detail` |
[...]

PR #1781, opened 2026-08-20
feat: citation validators Group A (V1, V5, V6, V10)

yinkid28:
## Summary

Implements Group A of issue #1749 — four tier-1 (gating) validators from the citation deep-dive:

- **V1** `test_write_then_validate` (universal) — if research.json was modified, `validate_research_schema` must appear in `tool_calls`. Fires on any skill declaring the tool (currently citation, hypothesis-tracking).
- **V5** `test_creator_not_in_custody` (citation) — the tree source `author` must not appear in the custody parenthetical of `citation_detail.where`.
- **V6** `test_unknown_marker_vocabulary` (citation) — `[... NOT RECORDED]` markers must name a Who/What/When/Where/Wherein element.
- **V10** `test_informant_not_in_who` (citation) — `citation_detail.who` and the `citation` string must not contain "informant" (literal half only).

Also: conftest.py fixtures for standalone pytest, three-tier docs in unit-test-spec.md, and 8 new unit tests.

**Depends on:** PR #1774 (issue #1748, retire per-skill tool deny). Branched from `1748-retire-tool-deny`.

**Groups M and B** (the tier-2 reporting mechanism and remaining validators) are deferred — separate planning cycle.

**No CI job can tell you whether the judge was biased by the new prompt section** — this PR adds no judge prompt changes (that is Group M). The gap is noted per `docs/architecture.md` § "What nothing checks".

## Test plan

- [x] `cd eval/harness && uv run pytest tests/ -q` — 2394 passed, 3 skipped, 1 pre-existing failure (unrelated `temp_test_cli.py` encoding lint)
- [x] Issue-ref ratchet: `unit-test-spec.md` stays at baseline 4
- [ ] Reviewer: confirm V5 fires on `ut_citation_014` in the historical run logs
- [ ] Reviewer: confirm V10 fires on nothing (by design — regression guard)


Asimi1234 (commented):
I reviewed this after merging it with `main` (no conflicts). All checks are green, and the harness tests pass (2442 passed — the 1 failure is just "build not found" because a fresh worktree has no compiled engine, not a real problem). V1, V5, and V10 look good. **There's one thing that needs fixing before this can merge: V6.**

## The problem — V6 will fail runs that are actually correct

When a piece of info is genuinely missing, the citation skill is *supposed* to write a marker like `[VOLUME AND PAGE NOT RECORDED]`. That's the correct behavior — it's spelled out in the skill, in the rubric, and in the `missing-locator-flagging.json` test fixture (which literally says this exact marker is the right answer).

But V6 only accepts markers named exactly `who`, `what`, `when_created`, `when_accessed`, `where`, or `where_within`. Anything else, it rejects. I ran V6 against the real markers used in the project and it rejects **10 out of 10** markers that the skill treats as correct:

```
[VOLUME AND PAGE NOT RECORDED]  [PAGE NOT RECORDED]  [CREATOR NOT RECORDED]
[WILL BOOK AND PAGE NOT RECORDED]  [DATE NOT RECORDED]  [ARTICLE TITLE NOT RECORDED] ...
```

Because V6 is a gating check (a failure stops the test and the judge never runs), this means a **correct** citation run gets marked as **failed**.

Right now the tests are green only because the citation eval that would hit this isn't part of CI. The next time someone runs `make eval-skill SKILL=citation`, it would fail on a correct answer.

**This isn't really a mistake in your code** — you built exactly what issue #1749 described. The problem is the approach itself: checking the marker's *name* against a fixed list can't tell a genuinely-bad marker (like physical custody info that belongs in `notes`) apart from a good one (like "volume and page"). Both get rejected. So this needs @DallanQ or the issue author to decide on a new way to check it before V6 can safely gate. It's not a quick fix on your side.

Tiny thing in the same function: the error message prints double brackets — `marker [{m.group(0)}]`, but `m.group(0)` already includes the brackets, so you get `[[...]]`.

## Smaller notes (not blocking)

- **V5** checks if the author's name appears *anywhere inside* the `where` text. That's usually fine, but a short one-word author name (like `GenealogyBank`) could accidentally match a different organization. Low risk — just worth a quick look.
- **Heads up on merge order:** this branch was built on top of PR #1774 (issue #1748), which is still open. So 8 of the files in this diff are actually #1774's work, not yours — those should be reviewed on #1774, not here. You already noted this in the description, so nothing to do; just flagging it.

T-FEH (changes requested):
Reviewed on the merged head (`0cdafff9`) — I merged `main` and resolved four conflicts, so the diff is now cleanly 5 files (+477/−13); #1774's 8 files are on main and out of scope here. Full harness suite green locally: **2644 passed, 2 skipped**. No paid re-run owed — the diff touches only `docs/specs/**` and `eval/harness/**`, neither in the run-log snapshot, and no `.ann.json` is modified.

Good news first: all four validators **can fail** (I replayed them — none is vacuous, which is not a given in this repo), and none declares a parameter the runner doesn't supply, so the #1764 failure mode is not reintroduced. V1 and V5 are sound.

Two blockers.

## 1. V10 reds every positive citation test — including a run where the skill does nothing

`test_citation.py:199-215` computes `before_sources` but only uses it to *skip newly added* sources. It never compares before against after, so it flags pre-existing data the skill never touched.

`mid-research-flynn/research.json` — the scenario all 18 citation tests share — seeds `src_004.citation_detail.who` with `"Pennsylvania Department of Health; informant: James Brown (son-in-law)"`.

Null-run proof, `before == after ==` the shipped fixture (skill did nothing):

```
V6  test_unknown_marker_vocabulary: PASS
V10 test_informant_not_in_who:      *** FAIL ***  src_004
```

V10 is gating, so the judge never runs and the test records `fail`. The next `make eval-skill SKILL=citation` is a wasted paid run.

Scoping to fields the run actually changed fixes it. I verified both directions — the false positive clears, and a genuine write is still caught:

```python
if "informant" in who.lower() and who != (before_sources[sid].get("citation_detail", {}) or {}).get("who", ""):
...
if "informant" in citation.lower() and citation != (before_sources[sid].get("citation", "") or ""):
```

```
NULL RUN    current: ['src_004.who']            fixed: clean
REAL WRITE  current: ['src_001.who','src_004.who']  fixed: ['src_001.who']
```

## 2. V6 gates against markers the shipped SKILL.md mandates — needs a decision, not a patch

`_ALLOWED_MARKER_ELEMENTS` accepts only the six framework slot names. Real markers name *sub-elements*, with zero overlap:

- `citation/SKILL.md` itself instructs 9 of them — `[VOLUME AND PAGE NOT RECORDED]`, `[WILL BOOK AND PAGE NOT RECORDED]`, `[ARTICLE TITLE NOT RECORDED]`, …
- Committed run logs carry **22 distinct marker forms, 585 occurrences**, across all five logs including the two newest (`v1_2026-08-19`, `v1_2026-08-20`). None match.
- `rubric.md:49` says these must "never be penalized"; `missing-locator-flagging.json` names one as the correct answer.

This confirms @Asimi1234's finding — and I agree with their read that the approach, not the code, is the problem: matching a marker's *name* against a closed list cannot separate a mandated `[VOLUME AND PAGE …]` from custody detail that belongs in `notes`. @DallanQ, this one needs your call before V6 can gate.

CI is green only because validators fire under `make eval-skill`, not the unit suite — the 2644-test pass says nothing about either defect.

## Minor

- `test_citation.py:118` — V6's `[A-Z_ ]+?` excludes digits and `#`, so `[FHL FILM #1234 NOT RECORDED]` passes silently. That's the microfilm case the docstring names as the thing to catch. `r'\[([^\]]+?)\s+NOT RECORDED\]'`.
- `test_citation.py:171` — `marker [{m.group(0)}]` double-brackets; `group(0)` already includes them.
- The spec's new argument enumeration and `conftest.py` both omit `attempted_mcp_calls`, which the runner does supply (`validator_runner.py:83`) and `test_tool_allowlist` declares.

## Worth stating in the PR body

V1 is correct, but `validate_research_schema` appears in zero of the 4 most recent citation run logs, so V1 will also red most citation tests once it lands. That's a real skill gap rather than a validator bug — it just shouldn't arrive as a surprise on the first paid run.

Requesting changes on V10 (concrete fix above) and V6 (blocked on a design decision).

T-FEH (changes requested):
Re-reviewed on the merged head (`99d24248` — your `e335cf99` plus a clean merge of main). Harness suite green: **2693 passed, 2 skipped**. No eval re-run owed. I re-ran every check from last round rather than trusting the commit messages.

## V10 is properly fixed — and not by going quiet

My exact null-run reproduction now passes, and it still catches what it should:

```
NULL RUN (before == after == shipped fixture)   -> PASS
'informant' written into a clean src_001.who    -> FAIL   (caught)
src_004 edited, 'informant' kept                -> FAIL   (caught)
```

That third case is the one I was watching for: a naive "skip if unchanged" would have let the agent edit around the check. Yours doesn't. This is the fix I hoped for.

## V1 and V5 re-verified sound

- **V1** — write with no `validate_research_schema` call → FAIL; with the call → PASS.
- **V5** — creator in the `where()` parenthetical → FAIL; the same name in `who` → PASS, correctly, per your docstring. The join through `gedcomx_source_description_id` is a real foreign key, not an id-space coincidence (`src_001 → S1 → "U.S. Census Bureau"`).

Both minors landed too: the regex now matches `[FHL FILM #1234 NOT RECORDED]`, `[VOLUME 3 NOT RECORDED]`, `[WHERE-WITHIN NOT RECORDED]`, and the double-bracket is gone. An AST sweep confirms every validator in both files declares only runner-supplied parameters, so the #1764 defect stays closed.

## V6 — the flip is the right instinct, but the deny-list still contradicts the skill

Against the real corpus the deny-list now accepts 16 of 18 marker forms (126 occurrences), including everything `citation/SKILL.md` mandates. My original blocker is closed.

What it still rejects is `repository`. Replaying the new V6 against the **persisted** after-states in the committed run logs, 5 runs trip it — including both of the newest:

```
ut_citation_013 (07-19)   ut_citation_017 (07-21)   ut_citation_013 (08-19)
ut_citation_005 (08-20)   ut_citation_017 (08-20)
```

all on `citation_detail.where`, all this shape:

```
where: 'FamilySearch.org ([ORIGINAL REPOSITORY NOT RECORDED])'
```

SKILL.md defines that field as *"Where the record was VIEWED (cite what you see), **then the original repository**"* (:72), with the canonical example `"FamilySearch.org (NARA microfilm M432, roll 810)"` (:128). And the don't-fabricate section closes it: rule 1 forbids inventing *"repository detail that is not on file"*, rule 4 says **"Use explicit unknown-markers for gaps"** with no carve-out. Composed, those two rules *mandate* exactly what the agent wrote. V6 gates against it.

## My recommendation: don't patch the keyword list a third time

I could tell you to drop `repository`, `microfilm` and `microfiche` — SKILL.md :72/:128/:272 place all three in `where`. But I think that would just move the collision.

The deeper problem is that V6's premise doesn't survive contact with rule 4. Its docstring says custody markers "belong in notes, not in the citation" — but `where` **is** the custody field by SKILL.md's own definition, and rule 4 mandates a marker for *every* gap, unqualified. Two designs have now been tried and both collided with the skill; that is evidence the requirement in #1749 needs revisiting, not the implementation. A deny-list narrow enough to avoid the skill's mandate may have nothing real left in it, which trades a false-positive gate for a vacuous one.

Two ways forward, both fine by me:

1. **Ship V1, V5 and V10 now and split V6 out.** All three are verified sound and gate correctly. This unblocks the PR today and gives #1749 its Group A less one.
2. **Get the carve-out ruled first** — if the genealogists decide unknown-markers for custody elements really do belong in `notes`, that is a SKILL.md rule 4 amendment, and V6 becomes implementable against it.

@Asimi1234 raised the same escalation in the first round; this is the second design to hit it, which I think makes the case.

Requesting changes only on V6 — everything else in this PR is ready.

yinkid28:
Agreed — removed V6 entirely (`343c2201`). V1, V5 and V10 ship; V6 is deferred pending the genealogist ruling on whether custody markers belong in `notes` or `where`.

Suite: 2641 passed, 3 skipped, 1 pre-existing failure.

T-FEH (approved):
Third pass, on `63ac16c4` (your `343c2201` plus my merge of main and one doc fix). Engine 2700 passed; harness 2746 passed, 2 skipped. Gate green. No eval re-run owed.

Splitting V6 out was the right call, and the three that ship are sound. The review agent stalled twice on this one, so everything below is my own verification.

## The merge needed real work — three conflicts, all resolved (`92e8d151`)

- `conftest.py` — took main's on all three hunks. Main generalizes the argument list and **adds the `attempted_mcp_calls` and `text_response` fixtures**, which closes the minor I raised in round 1.
- `unit-test-spec.md` and `test_validator_runner.py` — both sides were pure additions (different argument bullets, different tests), so I kept **both** rather than picking a side.

## V6 was gone from the code but still advertised in the spec — fixed (`63ac16c4`)

No `test_unknown_marker_vocabulary`, `_CUSTODY_KEYWORDS`, `_MARKER_RE` or `_marker_names_custody_element` survives, and nothing else depended on those helpers. But the spec still carried it in two places:

```
:2046  "All citation-specific validators (V5, V6, V10) are gating."
:2055  "…creator-not-in-custody (V5), unknown-marker vocabulary (V6), informant-not-in-who (V10)."
```

A spec documenting a validator that no longer exists is the inverse of the rule this repo keeps. Docs aren't in the snapshot so it was free; I folded it in and re-ran the doc lints (152 passed), since the issue-reference ratchet caught a similar edit of mine on another PR.

## The three survivors are intact, and both reviewer boxes are now empirically confirmed

Mutation-tested, and no false positive on an untouched fixture:

```
NULL RUN (skill did nothing)     V5 PASS   V10 PASS   no-new-sources PASS
V5  creator into where() paren   FAIL (caught)
V10 informant into a clean who   FAIL (caught)
V1  wrote, no validate call      FAIL (caught) / with the call  PASS
```

Replayed against the 47 committed citation runs that persisted source changes:

```
test_creator_not_in_custody   fires on 2 — ut_citation_014 (07-21), ut_citation_016 (08-19)
test_informant_not_in_who     fires on 0
```

That closes both unchecked boxes in your body: V5 does fire on `ut_citation_014` historically, and V10 fires on nothing — by design, as a regression guard. Neither would red a run whose behaviour `citation/SKILL.md` mandates, which was the whole V6 problem and it left with V6. Parameter declarations are clean in both files, so the #1764 failure mode stays closed.

## One thing to decide before merge: V1 reds most of the citation suite

V1 is **correct** — I checked it against the skill rather than assuming. `citation/SKILL.md:529` says outright *"If you wrote any changes to `research.json`, call `validate_research_schema(...)`"*, and names the one exemption (no changes → skip), which V1 already honours by skipping when research.json is unmodified.

But the skill does not follow its own instruction. Of the 47 committed runs that modified `research.json`, only 8 called the tool, so V1 would red **39 runs across 11 distinct tests**:

```
ut_citation_004  005  006  013  014  015  016  017  … x4 each
```

Because V1 gates, those 11 tests fail before the judge runs. That is the validator doing exactly its job and a genuine skill defect it surfaces — not a defect in this PR. But it means the next paid `make eval-skill SKILL=citation` reds most of the suite on day one, so someone should decide whether to land it gating now or pair it with the skill fix. Either way it should not arrive as a surprise. @DallanQ, this is a sequencing call rather than a code one.

## Two text items before merge

- **The PR body still advertises V6**, with the original allow-list rule (*"markers must name a Who/What/When/Where/Wherein element"*) — the rule that was rejected twice. The body is the permanent record; it should describe the three validators that ship.
- **V6's deferral needs a home.** It left the branch with only a commit message ("pending design decision"). The open question is real — whether unknown-markers for custody elements belong in `notes`, which needs a genealogist ruling and possibly a SKILL.md rule-4 carve-out — and it should be a card against #1749 rather than lost in history.

Approving on the code. The two items above are text, and the V1 sequencing is the lead's call.

Asimi1234 (approved):
My V6 blocker is resolved by removal — V6 is gone from the code (`343c2201`) and the spec (`63ac16c4`); no `test_unknown_marker_vocabulary` / `_CUSTODY_KEYWORDS` / `_MARKER_RE` / `_marker_names_custody_element` survives anywhere. Kicking the requirement back to #1749 for a genealogist ruling was the right call rather than patching the keyword list a third time.

The three that ship — V1, V5, V10 — are sound. Local validator suite 94 passed (3 skips are just "engine build absent" in a fresh worktree). CI green. My other two notes were non-blocking and are now moot (V6 double-bracket gone with V6; #1774's files landed on main).

Approving. One thing that's the lead's call, not a code issue: V1 is correct but reds most of the committed citation runs, so the next paid `make eval-skill SKILL=citation` goes mostly red on day one — worth deciding whether to land it gating now or pair it with the skill fix, per T-FEH's note to @DallanQ.

mercyokum (approved):

PR #1782, opened 2026-08-20
Filter volume_search by record-type group, surface structured coverage fields

### LINKED ISSUE #1728: volume_search: implement record-type group filtering and structured coverage dates
> **Blocked — verified 2026-08-19. Do not start.**
>
> This issue implements a spec section that is **not on `main`**. PR #1693 (the spec) is
> still open, `ready-for-senior-developer`, unreviewed. On `main`,
> `docs/specs/volume-search-tool-spec.md` has no `## Record-type filtering` heading and
> **zero** occurrences of `recordTypeConceptIds` or `recordTypeGroups`:
>
> ```sh
> git show origin/main:docs/specs/volume-search-tool-spec.md | grep -cE "recordTypeConceptIds|recordTypeGroups"   # 0
> ```
>
> So the group table and strays table this issue calls "the source of truth" do not exist
> where a reader would look for them. Worse, PR #1693 records **seven open questions, three
> of which change the vocabulary's membership** — the enum literals in
> `RECORD_TYPE_GROUP_NAMES` will move once they are settled, so building against the branch
> now means rewriting the table after review.
>
> **Promote when PR #1693 merges.** Nothing else about the plan below needs to change.

**Touches:** packages/engine/mcp-server/src/utils/record-type-groups.ts, packages/engine/mcp-server/src/tools/volume-search.ts, packages/engine/mcp-server/src/types/volume-search.ts, packages/engine/mcp-server/tests/tools/volume-search.test.ts

Implement record-type group filtering and structured coverage dates in
`volume_search`, against the contract in `docs/specs/volume-search-tool-spec.md`
§ "Record-type filtering" (specified and reviewed in PR #1693).

## Why

FamilySearch's RMS group-search endpoint accepts a record-type filter the tool has
never sent — `coverage.recordTypeConceptIds` — applying hierarchy containment and
OR-ing the array. Without it an agent must pull every volume for a place and scan:
110 groups and 1889 coverage rows to find 4 tax volumes at Harjager härad. This
surfaced while working #572, where the Swedish mantalslängder were unreachable.

## Scope

- **`src/utils/record-type-groups.ts`** (new) — the 47-group vocabulary as data:
  `RECORD_TYPE_GROUP_NAMES` for the schema enum, and a `name → { anchor, strays }`
  lookup. The group table and strays table in the spec are the source of truth.
- **`src/types/volume-search.ts`** — `recordTypeGroups?: string[]` on the input;
  `recordTypeConceptIds?: number[]` on the RMS coverage request;
  `recordTypeConceptId?`, `recordTypeConceptIdHierarchy?`, `fromdateString?`,
  `todateString?` on the coverage entry; `recordTypeConceptId?`, `startYear?`,
  `endYear?` on `SimplifiedCoverage`.
- **`src/tools/volume-search.ts`** — validate and expand `recordTypeGroups` to
  anchors plus strays; send `coverage.recordTypeConceptIds`; surface the new
  coverage fields; derive `dateRange` from the date pair when `datesOrig` is
  absent; add the input to `volumeSearchSchema`.
- **Reject the wrong field name server-side.** Three sibling tools take a
  singular `recordType`, and nothing in this server validates input against the
  advertised JSON Schema (no `ajv`, no `zod`; `src/index.ts` casts
  `request.params.arguments` straight to each handler). So `recordType` sent here
  is currently ignored and the search silently runs unfiltered. `validate()` must
  throw and name the valid groups.
- **`tests/tools/volume-search.test.ts`** — cases 15a–15f from the spec's testing
  table, plus the `recordType` rejection.

## Already decided — do not re-open

Settled in the PR #1693 review and recorded in the spec's "Decisions taken in
review": strays are kept and groups may overlap; Casualties/War/POW is not a group
(all four members reachable via Military and Death); Emigration stays, carrying
`131602` as a stray; Enslavement and Indigenous keep the taxonomy's placement;
Newspapers anchors on the root `124231`; unknown-field rejection happens in
`validate()`, not the schema.

## Acceptance check

- `npx vitest run tests/tools/volume-search.test.ts` — the new cases must be shown
  to fail against pre-change code, per CLAUDE.md's "a new lint must be proven to
  fail".
- Live: `npx tsx dev/try-volume-search.ts --standardPlace "Harjager, Malmöhus, Sweden" --startYear 1650 --endYear 1720`
  with and without a group — expect 110 groups unfiltered and 4 for Tax.
- `npx tsx dev/probe-record-type-groups.ts reach` still reports 100% for 9 of the
  10 sampled places after the vocabulary lands in code, confirming the shipped map
  matches the probe's.
- `make engine-test` and `npm run typecheck` clean.

## Not in scope

Skill prose (no SKILL.md describes these fields, so no eval run is triggered), a
server-wide `additionalProperties` policy, and the annotation wording on the
Enslavement and Indigenous rows — that last one is a genealogy-reviewer item
tracked on the spec PR.

Gennecis:
## Summary

Closes #1728. Implements the contract merged in #1693.

`volume_search` never sent the RMS group-search endpoint's
`coverage.recordTypeConceptIds` filter. An agent asking for tax volumes at
Harjager härad therefore pulls **110 volume groups and 1889 coverage rows** and
scans them by hand to find **4**.

This adds a `recordTypeGroups` input over a **47-name vocabulary (67 concept
ids)** built from FamilySearch's own type hierarchy. Three properties of the
endpoint shape the implementation:

- **It applies hierarchy containment and ORs the array.** A group sends its
  anchor id and the server returns everything nested beneath it, so the
  vocabulary ships one id per group rather than a flattened list.
- **Twelve groups carry "strays"** — ids that belong to the group by meaning but
  sit outside its subtree (Tax Census Records is filed under Census, Prison's two
  under Legal). Those are sent explicitly, and selecting a parent also sends its
  *descendants'* strays, so `Government` is a superset of `Prison`: broadening a
  search never returns fewer volumes than narrowing it.
- **It ignores unknown request fields silently.** A wrong filter looks like an
  unfiltered search, not an error — so `validate()` rejects an unrecognised group
  name, a non-array, and the singular `recordType` that three sibling tools take.
  An empty array means no filter, matching the API.

It also surfaces two fields the coverage response already carried:

| Field | Why |
|---|---|
| `recordTypeConceptId` | the stable key; `recordTypeOrig` is locale-specific display text, absent from **18%** of coverage rows — 202 of 202 in New South Wales, where the concept id is present on every one |
| `startYear` / `endYear` | present where `datesOrig` is not (Wayne, Ohio: **462 of 463** against 335); `dateRange` falls back to the pair, with `datesOrig` keeping precedence |

Seven spec corrections land with it, none of them new contract: singular `anchor`
in the vocabulary shape, a plural unknown-group error, the tool description
saying to resend the filter on a page-2 call, the year descriptions matching the
shipped strings, and error-table rows for the throws `validate()` performs.

## Review guidance

**Start with `src/utils/record-type-groups.ts`.** One table of
`{ name, anchor, parent, strays? }`; the schema `enum`, the lookup map and
`conceptIdsForGroups` are all derived from it. `conceptIdsForGroups` is the
function worth reading closely — the descendants'-strays union is what keeps a
broader group from returning fewer results. `descendantsOf` walks a
parent-to-children index breadth-first with a visited check, because a
mis-transcribed `parent` would otherwise loop *before* the request is built,
where `fetchWithTimeout` cannot reach it; `assertAcyclicTable` covers the cycles
no query reaches, and both suites assert it.

**The date-range fallback is a reuse, not a new formatter.** The spec says "match
`collections_search`, do not invent a second format", so the expression inside
`toCollection` (`collections-search.ts:155-160`, one call site) moved to an
exported `formatYearRange` in `utils/search-helpers.ts`. Behaviour is
byte-identical for `collections_search` — including that equal years render
`1683-1683` and a lone `endYear` yields `""` — and `volume_search` reads `""` as
"omit" because its field is optional.

**Two copies of the vocabulary became one, and the third is linted.** The
probe's `GROUP_CLAIMS` keeps 17 anchor ids of its own on purpose — it records
what the source list claimed, not what ships — and nothing compares those. The
probe imports the shipped table rather than holding its own, and
`tests/packaging/record-type-group-drift.test.ts` parses the spec's group and
strays tables and compares them to it — the guard the spec's Files row assigns to
this PR. That test also compares the tool's advertised descriptions against the
spec's `## Tool schema` block in both directions, since `manifest.json` carries
names only, `readme-catalog.test.ts` is presence-only by design, and
`prompt-budget.test.ts` reads SKILL.md bodies.

**One hazard remains unguarded, by design:** the `pageToken` follow-up must resend
the same filter. The cursor is an opaque upstream token carrying no record of the
filter that produced it, so a page-2 call that drops `recordTypeGroups` is
indistinguishable from a valid unfiltered continuation. Closing it means minting a
composite cursor, which is larger than this contract.

**Volume counts are permission-dependent** (`119166` is 152,454 at this access
level, 176,444 elevated — the spec's parallel-drift table has the detail). The
live checks below assert structure and relative filtering, never absolute totals.

## Test plan

- `tests/tools/volume-search.test.ts` — **42 passing**, and shown to fail against
  pre-change code: against `origin/main`'s tool file, **13** fail, every new case.
- `tests/utils/record-type-groups.test.ts` — **9 passing**: the invariants the
  spec's tables cannot express, including that an expansion sends an anchor's
  strays and nothing extra, that broadening to a parent loses no id, that every
  parent chain terminates, and that a `null` entry is named rather than joined
  into nothing.
- `tests/packaging/record-type-group-drift.test.ts` — **9 passing**, and shown to
  fail four ways: flip a digit in a spec anchor (2 failures), drop a stray id from
  the spec (3, including the anti-vacuous count), reword a description on the code
  side (1), and remove one so the two sides declare different counts (1). The
  cycle guard likewise throws `Cyclic parent chain … at "Vital"` in under a second
  when a group's parent is pointed at its own child.
- `tests/utils/search-helpers.test.ts` — **5 passing**, one per `formatYearRange`
  branch.
- `tests/tools/collections-search.test.ts` — passing untouched, which is what
  shows the lift changed no behaviour.
- `make engine-test` — 106 files, **2481 tests**. `npm run typecheck` clean.
- `make test-all` — green.
- Live, `Harjager, Malmöhus, Sweden` 1650–1720: unfiltered **110**, `Government`
  **4**, `Tax` **4**. The assertion is the relation `unfiltered > Government >=
  Tax`, since 4 was measured on the bare anchor and cannot by itself distinguish a
  correct expansion from an anchor-only one.
- `dev/probe-record-type-groups.ts reach` — 100% for all 10 sampled places, Bayern
  at 164288/164292, header confirming **67 ids**.
- `dev/probe-record-type-groups.ts tree` — the reproduction the spec's group table
  names, and the evidence behind the *Parent* column. All **47** anchors resolve,
  none `UNRESOLVED`, **17 roots**, and all **30** direct parent relations
  reproduce. Stronger than the column asserts: every group's full ancestor chain
  equals the table's transitive walk, so no intervening non-group concept sits
  between any group and its declared parent — which is what licenses omitting
  descendants' anchors from an expansion.
- `dev/probe-record-type-groups.ts union` — **new section**, the reproduction behind
  the descendants'-strays union. Six groups have a stray-carrying descendant;
  measured across ten places plus a global pass. Globally the union adds **Vital
  +110235**, **Legal +64911**, **Government +27698**, **ID documents +1795**.
  Exactly one outright inversion in the sample: at Bayern, `ID documents` without
  the union returns **0** while its child `Passports` returns **89** — all 89
  reached only through Passports' strays, which sit under Migration. `Religious`
  and `Migration` add zero everywhere, which is correct rather than a gap: their
  descendants' strays escape the child anchor's subtree but not the parent's.
  Measured **unwindowed** — a date window suppresses the coverage rows whose dates
  do not parse, which is most of a stray's, and understates Vital's gain as 7121.
  The unwindowed figure reproduces this spec's Mapping-logic 110231 to within 4.

## Follow-on issues

None. The `pageToken` filter-resend hazard above is not a deferral this PR makes:
`origin/main`'s spec already states the whole caller obligation *including*
`recordTypeGroups` (input table at `docs/specs/volume-search-tool-spec.md:107`,
Pagination §930-964), merged ahead of the code in #1693, so this PR implements a
documented contract rather than leaving a new gap. Nor is the class shared:
`record_search` and `fulltext_search` paginate on a numeric `offset` with no
server-side cursor state, and `collections_search` and `image_search` do not
paginate at the caller boundary at all — `volume_search` is the only tool with an
opaque cursor. Closing it means minting a composite cursor, which is a change to
the merged contract, not a test.

**One deviation from the merged spec:** the 29
`eval/fixtures/mcp/volume-search-*.json` fixtures are not regenerated, though
`docs/architecture.md` §3 lists that directory under "also touch, and nothing will
tell you if you don't". **Zero** of them carry `recordTypeConceptId`, `startYear`
or `endYear` on a coverage; 19 mention `startYear`, 18 as a query echo and one
(`volume-search-edensor.json`) under `.args` as well. `matches()` in
`eval/harness/harness/fixtures.py:93-113` iterates the predicate's keys only, so an
extra argument cannot break fixture matching, and `volume_search` is not in
`mock_mcp.py`'s `LIVE_TOOLS`, so the real mapper never runs against them.
Regenerating them is a larger separate job and would not exercise the new mapping
either way.


mercyokum (approved):
Approving. Everything looks good from the genealogical side

chrisedeson:
Ran a full structural review, all nine finder angles completed. Overall this is a well-tested, well-specified PR — several suspicions turned out to be deliberate and already covered once checked against the spec. Verified the two most serious findings myself directly, both against the real code, not just the descriptions below.

### Should fix before merge

**1. The cycle-detection guard misses the exact cycle shape it exists to catch.**

`src/utils/record-type-groups.ts:113` — `descendantsOf()` checks `if (parent === name)` before `if (seen.has(parent))`. So when the group you're querying is itself part of a cycle, the walk hits its own name on the way around and returns early as a "descendant" before the cycle check ever fires. Reached from any other name, the same corrupted table correctly throws. Reproduced directly with a synthetic 2-row cycle (A's parent is B, B's parent is A):

```js
descendantsOf("A")  // -> [A, B], no error -- the group at the center of the cycle is silently missed
descendantsOf("X")  // -> throws "Cyclic parent chain..." correctly, reached from outside the cycle
```

Two independent review passes found this the same way, and I confirmed it myself with the repro above. Currently latent — the real 47-row table is acyclic — but it means the safety net this function's own docstring promises ("throwing names the bad row instead of hanging the caller") silently fails for exactly the row a copy-paste mistake would most often produce (two adjacent rows swapped). Fix: check `seen.has(parent)` first, or check `parent === group.name` up front for the degenerate self-loop case.

**2. Unknown-group error messages silently drop `null`/`undefined` entries.**

Both `validate()` in `volume-search.ts:92` and its documented backstop `conceptIdsForGroups()` in `record-type-groups.ts:170` build their error message with `unknown.join(", ")`. `Array.prototype.join` coerces `null`/`undefined` to empty string, not the word "null". Confirmed directly:

```js
[null, undefined, 42, "Tax"].join(", ")  // -> ", , 42, Tax" -- the bad entries are invisible
```

Since there's no schema-level enforcement server-side (the code's own comment says so), an LLM caller can send `recordTypeGroups: ["Tax", null]` and get back `"Unknown record-type group(s): . Valid groups: ..."` — the one entry actually wrong is blank. This undermines the PR's own stated reason for making the message plural in the first place ("one error should let the caller fix them all"). `search-helpers.ts`'s existing `parseUpstreamErrorBody` already avoids this exact trap by filtering non-strings before joining — same fix would work here in both places.

### Worth a look, less critical

- **The new `recordType` collision guard doesn't generalize.** It catches exactly one spelling mistake (`recordType` singular vs. the tool's real `recordTypeGroups`), but this tool's own `startYear`/`endYear` pair has at least three other live spellings elsewhere in the codebase (`fulltext_search`'s `yearFrom`/`yearTo`, `tree_forget`'s `fromYear`/`toYear`, `record_search`'s event-prefixed variants), none of them checked. It also doesn't catch `record_type` (snake_case), the actual convention this repo's own `research.json`/enums use elsewhere. Not a regression, since nothing checked any of this before either, but the fix is narrower than the failure class its own comment describes.
- `descendantsOf()` recomputes a full 47-row table scan on every call, including once per requested group and once per page of a paginated search (the tool tells callers to resend `recordTypeGroups` unchanged on every page). Negligible today at 47 rows; would be a real cost if this table grows an order of magnitude. A precomputed parent→children map built once at module load would make this O(1) per lookup.
- The "Unknown record-type group(s)" error template and its unknown-name-finding logic are independently duplicated (as intentional defense-in-depth) between `validate()` and `conceptIdsForGroups()`, with slightly different internal shapes for no functional reason. A shared `assertKnownGroupNames()` helper would remove the duplication without giving up the backstop.
- `leadingYear()` (new, in `volume-search.ts`) does the same job as `parseYear()` in `external-links-search.ts`: extract a plain year from a messy string, return a nullish sentinel rather than throw. Two near-duplicates now exist for one concept — worth lifting into `search-helpers.ts`, which already centralizes `isFourDigitYear`/`formatYearRange` for exactly this reason.
- `recordTypeConceptIdHierarchy` (new field on `MetadataRmsCoverageEntry`) is never read by any shipped code — the other three new fields on that interface all are. Either drop it until something reads it, or say in a comment that it's deliberately forward-facing.
- Minor: a stray formatting glitch in `RECORD_TYPE_GROUP_TABLE`'s last row (closing bracket glued to the final element instead of its own line, unlike all 46 siblings) — harmless, but notable specifically because this table's whole purpose is protecting against unnoticed hand-edits. Two copy-pasted test fixture constants bound to the identical place-name string, 126 lines apart. A `describedStrings()` helper in the new packaging test is fully general/recursive for a call site that's actually flat and 6 fields deep. Three new tests call bare `.sort()` on numeric arrays (lexicographic, not numeric) — harmless today since every concept id is a uniform 6 digits, but `conceptIdsForGroups` itself gets this right two lines away with an explicit comparator. One broken sentence in the spec doc (an admonition heading phrase copied into a link caption where it no longer parses as English).

### Verdict

Solid, well-tested work — the drift-guard test, the spec's own documented tradeoff reasoning, and the general shape of the fix are all sound, and CI is green. The two "should fix" items are narrow, both empirically confirmed, and both worth a fix-and-re-request round: one is a latent robustness gap in a safety net the code explicitly claims to provide, the other quietly defeats the PR's own design goal of naming every bad input at once. Nothing here is a fundamental design problem.

chrisedeson (changes requested):
Found two real bugs in the new record-type-group logic: a cycle-detection guard that misses the exact cycle shape it's meant to catch, and an error message that silently drops null/undefined entries instead of naming them. Check the comment above for broadness.

Gennecis:
Both "should fix" items reproduced and fixed, and I took most of the rest. Two I
am pushing back on, with reasons. Everything below was verified against the code
before being acted on.

### 1. The cycle guard

Reproduced exactly as described — with a `Vital`↔`Birth` cycle, either cycle
member returned silently while an outside name threw.

Rather than reorder the two checks, `descendantsOf` is now breadth-first over a
parent-to-children index built once at module load, with the visited check as the
first thing done to every node. That holds from inside the cycle as well as
outside, and it also answers the O(n)-per-call point below it: the walk now costs
the subtree rather than a full 47-row scan.

The rewrite surfaced something the reorder alone would not have. **A cycle the
query never walks into is invisible to any runtime guard** — with the cycle
injected, `Tax` still returns cleanly, because nothing reaches the corrupt rows.
So there is now an exported `assertAcyclicTable()` that checks all 47 chains,
asserted in both the unit and packaging suites. Runtime cannot hang; CI catches
corruption regardless of what is queried.

### 2. The unknown-name message

Confirmed — `["Tax", null]` produced `Unknown record-type group(s): . Valid
groups: …`.

Fixed in one place rather than two, which also takes the duplication point:
`assertKnownGroupNames()` now lives in `record-type-groups.ts`, and both
`validate()` and `conceptIdsForGroups` call it, so the tool's error and its
backstop cannot word one failure differently. Non-strings render through
`JSON.stringify`, so the message now names `null`, `undefined` or `42`. Three
regression cases cover it.

### Also taken

- **The `recordType` guard** now covers `record_type` and `recordTypes`, with its
  comment narrowed to say it guards this field's own spellings. The spec's error
  row is generalised to match, and states why the year aliases are excluded.
- The `];` glued to the table's last row, the duplicate `EDENSOR`/`EDENSOR_P`
  constants collapsed to one, and numeric comparators on all six lexicographic
  `.sort()` calls — you counted three, there were six once `volume-search.test.ts`
  is included.
- `recordTypeConceptIdHierarchy` keeps its place with a comment saying what reads
  it: the probe's chain resolution, which is the evidence behind the group table.
- The spec sentence where an admonition heading was used as a link caption.

### Pushing back on two

**`leadingYear` is not `parseYear`.** They look like one concept but differ where
it matters: `parseYear` is `Number.parseInt`, which reads `16` out of `"16xx"` and
`999` out of `"999-01-01"`. `leadingYear` wants four leading digits or nothing,
because its input is an ISO timestamp and a partial parse would put a
plausible-looking wrong year on a coverage. Unifying them means either loosening
this one or changing a shipped tool's behaviour outside this PR's scope. The
reasoning is now in the docstring so the next reader does not have to re-derive it.

**`describedStrings()` stays recursive.** It is six lines, and the schema is not
flat — descriptions sit at the tool level and again under `properties`. A
two-level walk would silently stop finding them the first time a nested object
gains a description, which is the drift the test exists to catch.

### Verification

Engine suite 106 files / 2481 tests, typecheck clean, `make test-all` green. Both
new guards shown to fail against a corrupt table. Live, Harjager 1650–1720:
unfiltered 110, `Government` 4, `Tax` 4, `Prison` 0, and `Government` still
carries Prison's strays. 67 distinct ids, sorted, with all twelve stray-group id
counts unchanged — so the rewrite preserved behaviour exactly.

Pushed as a follow-up commit rather than rewriting the original, so the diff you
reviewed stays readable against what changed after it.

chrisedeson (approved):
Second pass, confirming both should-fix items. Reproduced the fix for each with a mutation test rather than just reading the diff.

Cycle guard: injected a real Vital<->Birth cycle into the actual table and ran the suite. Both assertAcyclicTable call sites (unit and packaging) failed exactly as they should, then passed clean once reverted. The BFS rewrite also closes a gap a simple reorder wouldn't have: querying a name with no path into the cycle stays silent, which is exactly why assertAcyclicTable exists as a separate, always-run check rather than relying on descendantsOf alone.

Unknown-name message: confirmed null/undefined/42 all render by name now, not blank.

Both pushbacks are right. Checked parseYear directly: it's Number.parseInt, which would read 999 out of '999-01-01' or 16 out of '16xx' -- a real behavior difference from leadingYear's all-or-nothing contract, not just a style choice. describedStrings staying recursive is reasonable given the schema genuinely has two levels.

Suite green (2512 tests), typecheck clean. Approving.

chesworthrm:
Reviewed against the **merge with `main`**, not the branch head. The branch is 12 commits behind and `protect-main` has `strict_required_status_checks_policy: true`, so the merged tree is what ships.

## Verified

Test-merged `origin/main` first — **clean, no conflicts** — then ran everything on the result:

| | |
|---|---|
| engine vitest | 107 files / **2491 passed** |
| engine `tsc --noEmit` | clean |
| `turbo test typecheck` | 8/8 tasks |
| `eval/harness` pytest | 2439 passed, 2 skipped |
| `apps/server` pytest | 209 passed, 1 skipped |

All six required checks green on `c0191f93`. Nothing failed and nothing skipped this head.

**The new lint fails, six ways.** I broke each case rather than reading it: flipped a digit in the spec's Tax anchor (2 failures), deleted Baptism's stray row from the spec (3), dropped the same stray from the *code* table (2), reworded a code-side description (1), injected a `Vital`↔`Birth` cycle (4, both `assertAcyclicTable` sites), and renamed the group-table header (3). Real coverage.

One note on that last one, in the guard's favour: with the header renamed, `states 'Also returns' as the full set of nested groups` **passes vacuously** — empty map, 0 ms. Every other comparison fails on its own because the code side is non-empty. So `parses both tables out of the spec at all` is load-bearing for exactly one dimension: the closure column the code comment calls "the one column nothing else reads." Well aimed.

I also confirmed the deviation you declared. `recordTypeConceptId` appears in **0 of 28** fixtures and **0 of 51** coverage rows; `harness/fixtures.py`'s `matches()` is subset-based, so an extra argument cannot break fixture matching; `volume_search` is not in `LIVE_TOOLS`. Your reasoning holds. One nit on the wording: `startYear` also appears once under `.args`, not only as a query echo.

## Findings

### 1. The `parent` comment says it isn't behaviour. It is.

I re-parented Prison from Government to Legal and rebuilt. `conceptIdsForGroups(["Government"])` went 13 ids → 10, silently dropping `126416`, `130086` and `131448` — Prison's three strays. That is the exact failure the descendants'-strays union was added to prevent. The suite does catch a code-only edit, but the comment invites someone to change both sides and move on.

```
You wrote:     * `parent` is documentation, not behaviour — it records which other group's
               * subtree this one sits inside, so a reader can see that selecting `Government`
               * also returns Tax, Prison, Poor Law and the rest. Nothing sends it upstream.

Change it to:  * `parent` is not sent upstream, but it is not decoration either: `descendantsOf`
               * walks it to decide whose strays go into the request, so re-parenting a row
               * changes what the tool asks for. Pointing Prison at Legal drops its three
               * strays from a `Government` search. Change a parent only with the `tree` probe
               * output in hand.
```

### 2. The Parent column is the one part of the table with no evidence in the test plan.

Every non-root group's expansion deliberately omits its descendants' anchors, on the claim that upstream containment reaches them. I counted: **30 direct parent relations, 36 ancestor/descendant pairs**, and in **36/36** the only id the ancestor's request lacks is the descendant's own anchor. So "broadening never returns fewer volumes" — which the tool states to the model verbatim, `"'Government' also returns Tax, Prison, Poor Law, Passports and more"` — rests entirely on those 36 unchecked claims.

`reach` cannot see them. It contains no reference to `parent` at all, and `vocabularyIds()` reads only `anchor` and `strays`, so the percentage is identical whether the Parent column is right or scrambled. The Harjager triple covers one pair of 36, and you already note it can't distinguish a correct expansion from an anchor-only one.

Please run `npx tsx dev/probe-record-type-groups.ts tree` and paste the output. It prints, for each of the 47 anchors, which other group anchors it sits inside — the whole Parent column in one pass, and the reproduction the spec itself names. Everything else here checks out; this is the one claim with nothing behind it.

### 3. Two doc comments stacked on one field.

Leftover from the fix commit. I confirmed with the TypeScript compiler API that `getDocumentationComment` returns only the second — the first is dead text, on the field the last round specifically asked about.

```
You wrote:     /** Ancestor chain, root first, ending in `recordTypeConceptId`. */
               /**
                * The concept's ancestor chain, root-first, ending in `recordTypeConceptId`.

Change it to:  /**
                * The concept's ancestor chain, root-first, ending in `recordTypeConceptId`.
```

### 4. "The third is linted" overstates it.

`GROUP_CLAIMS` in the probe still holds 18 rows, and 17 of their anchors duplicate a shipped table anchor (the 18th, `123352`, is the unshipped Casualties group). Nothing compares them, so a changed table anchor leaves `sectionAnchors` reporting against the old one. Defensible for a script recording what the *source list* claimed — but the sentence reads as if nothing is left unguarded.

```
You wrote:      **Two copies of the vocabulary became one, and the third is linted.**

Change it to:   **Two copies of the vocabulary became one, and the third is linted.** The
                probe's `GROUP_CLAIMS` keeps 17 anchor ids of its own on purpose — it records
                what the source list claimed, not what ships — and nothing compares those.
```

**Not blocking:** the body says "Follow-on issues: None" while declaring the `pageToken` filter-resend hazard permanent by design. I searched six ways and scanned all 43 open `nothing-checks` issues — nothing covers it. Either file one, or say in the body that the pre-existing `standardPlace`/years version of the same hazard already covers it.

## Merge mechanics

- **Still waiting on a `senior-developers` approval.** `require_code_owner_review` is on; `*.ts` routes to senior-developers OR senior-genealogists. Of the two approvals here, only mercyokum is on either team, and you're on senior-developers yourself so you can't self-approve.
- **mercyokum's approval did not read the fix commit.** `500e7e0c` and `205f3a15` are siblings off `dcc6fac4`; `git diff 500e7e0c c0191f93` is exactly `205f3a15` — the BFS rewrite, `assertAcyclicTable` and `assertKnownGroupNames`. `dismiss_stale_reviews_on_push: false`, so it still counts.
- **Updating the branch will stale both approvals without dismissing either**, so the tree that lands could have zero approvals that read it. I did that merge already: conflict-free and green on all five suites, so the update itself is safe.

## Not checked

No FamilySearch session here, so no live call — including the `tree` run above. I verified only that the shipped table, the spec's tables and the outgoing request agree with each other; the 47 anchors and the containment claims behind them are unverified by me. No eval or e2e run (no skill body or fixture changed, and this tool is fixture-served).

Last thing, not a finding: the three new coverage fields are emitted unconditionally, and over the 51 real fixture coverage rows that is **+61 bytes per row, +54%**. `## Mapping logic` mandates exactly that and it was settled in #1693, so it's out of scope here — but it's a real cost on unfiltered calls and worth an eye if agents keep calling this without a filter.

Gennecis:
Thanks — all four confirmed and fixed in `ca257b3e`, merged with main at `2a4d9bb6`.

**1. `parent` is behaviour.** Reproduced before changing anything, and it lands on
the id: `conceptIdsForGroups(["Government"])` sends 13; point Prison at Legal and
it sends 10, dropping `126416`, `130086`, `131448`. Paragraph replaced with your
wording.

**2. The Parent column now has evidence.** Ran `tree`. All **47** anchors resolve,
none `UNRESOLVED`, **17 roots**, and all **30** direct parent relations reproduce.
I diffed the output against the shipped table mechanically rather than by eye, and
it gives more than the column asserts: every group's *full* ancestor chain equals
the table's transitive walk, so no intervening non-group concept sits between any
group and its declared parent — which is exactly what licenses omitting
descendants' anchors from an expansion. Output is in the Test plan. You were right
that this was the one claim with nothing behind it; the table it was guarding
turned out to be correct.

**3 and 4.** Dead JSDoc gone. The `GROUP_CLAIMS` qualifier is in, as written.

**`pageToken`.** Took the second option rather than filing. I swept all 186 open
issues (titles and bodies), six `--search` queries and all 43 `nothing-checks`
bodies, and confirmed your "nothing covers it" — nothing closed, either. But it
isn't a deferral this PR makes: main's spec already states the whole caller
obligation *including* `recordTypeGroups` (input table at `:107`, Pagination
§930-964), merged ahead of the code in #1693. And there's no wider class to file
against — `record_search` and `fulltext_search` paginate on a numeric `offset`
with no server-side cursor state, and `collections_search` and `image_search`
don't paginate at the caller boundary at all, so `volume_search` is the only tool
with an opaque cursor. Said so under Follow-on issues.

Two corrections, both minor:

- **The three coverage fields are not emitted unconditionally.** `mapCoverage`
  guards all three — `volume-search.ts:211,212,233`. Your byte figure survives,
  because upstream supplies the values on nearly every row, but the mechanism is a
  guard rather than an unconditional write.
- **Your fixture recount inherited a drift in my body.** The merged tree has **29**
  fixtures and **19** mentioning `startYear`, not 28/18 — both moved from one file,
  `volume-search-berkshire.json`, which landed on main after I wrote the body and
  carries `query.startYear`. Numbers refreshed, and your `.args` nit is in there
  (it's `volume-search-edensor.json`).

chesworthrm (approved):
Approving. No blocking findings.

I reviewed the merge state rather than the branch head — the branch is level with `main`, so they are the same tree here. Full pass: engine suite 107 files / 2500 tests green in the PR worktree, `tsc --noEmit` clean, and all six required checks ran on `2a4d9bb6` and passed.

## What I verified rather than read

I broke each new guard and watched it fail, then restored: a flipped digit in a spec anchor, a dropped stray id, a code-side description reword, and a `Vital` → `Birth` cycle. All four fail with the row or the string named. The lints are real.

I also refreshed my FamilySearch session and re-ran the live claims instead of taking them on trust:

- The filter does the thing #1728 asked for. Harjager 1650–1720 goes from 111 volumes to 4 for `Government` and 4 for `Tax`. New South Wales goes from 16311 to 1132 / 833 / 435 for `Government` / `Prison` / `Baptism`.
- `startYear` populates on real coverages, which confirms that the lowercase `fromdateString` on the response is right next to the camelCase `fromDateString` on the request. That pair looks like a typo and is not — worth knowing before someone "fixes" it.
- The superset property holds live across 46 parent/child checks at six places (NSW, Tasmania, England, Ireland, Ohio, Norfolk). Zero violations.
- The descendants'-strays union earns its place. I reran every parent both ways — shipped, and with anchors plus own strays only. Ohio `Government` is 9190 shipped against 4653 without the union. England `Vital` gains 542. This is not a theoretical safety net.
- The `tree` probe reproduces exactly: 47 of 47 anchors resolve, none unresolved, 17 roots, and every ancestor chain equals the table's transitive walk.

The declared fixture deviation is sound. I checked both claims it rests on: `matches()` in `eval/harness/harness/fixtures.py` iterates the predicate's keys only, so an extra argument cannot break matching, and `volume_search` is genuinely absent from `LIVE_TOOLS`. Nothing can fail because of it.

The `anchors` → `anchor` spec change is also not a contract change. The two dropped second anchors that sit outside the subtree appear in the strays table, and the two inside it are reached by containment. The id set is unchanged.

## Two things to fix, neither blocking

**1. The module comment states a measurement nobody can re-run.**

In `src/utils/record-type-groups.ts`, `conceptIdsForGroups` says a `Government` search "returned strictly fewer prison volumes than selecting `Prison`". I could not reproduce a strict inversion at any of six places. Ohio came closest — 4653 without the union against 4646 for `Prison`, a margin of 7 — so it plausibly flips somewhere, and the design is right either way. But the comment names no place or period, so nobody can check it.

```
You wrote:      * Without this, selecting `Government` returned strictly fewer prison volumes
                * than selecting `Prison` — broadening a search lost results, silently, while
                * the tool description promised the opposite.
Change it to:   * Without this, broadening loses volumes: at Ohio a `Government` search sends
                * 4653 volumes' worth of ids without the union and 9190 with it, and `Prison`
                * alone returns 4646 — a 7-volume margin from inverting outright. Reproduce
                * with `dev/probe-record-type-groups.ts tree` plus a group-search count at
                * "Ohio, United States".
```

Use your own place and figures if you measured the inversion somewhere else. The point is naming where, so the next person can re-measure instead of trusting it.

**2. The spec's Files table is missing three files this PR edited.**

Someone reading `docs/specs/volume-search-tool-spec.md` to find the blast radius will not learn that `formatYearRange` is shared with `collections_search`. Add after the `dev/probe-record-type-groups.ts` row:

```
| `src/utils/search-helpers.ts` | Add `formatYearRange` — the date-range expression lifted out of `collections-search.ts`'s `toCollection`, so `volume_search`'s `dateRange` fallback and `collections_search` cannot describe one span two ways |
| `src/tools/collections-search.ts` | Call `formatYearRange` in place of the inline expression; behaviour unchanged |
| `tests/packaging/record-type-group-drift.test.ts` | **New.** The guard the row above assigns to this PR |
| `CLAUDE.md` | Extend the `search-helpers.ts` note to name `collections_search`/`volume_search` and `formatYearRange` |
```

The same table still describes `record-type-groups.ts` as exporting `RECORD_TYPE_GROUP_NAMES` and a `name → { anchor, strays }` lookup. It now also carries `parent` and exports `conceptIdsForGroups`, `assertKnownGroupNames` and `assertAcyclicTable`. Worth a few words, because `parent` decides what gets sent.

## Optional

`tests/packaging/record-type-group-drift.test.ts:180` slices to the end of the spec, not to the end of the `## Tool schema` block — 24,056 characters where it means about 2,700. I checked whether that lets a stale description through, and it does not: nothing in the spec's tail duplicates any description text. So this is hygiene, not a hole. If you want it anyway:

```
You wrote:      const schemaBlock = spec.slice(spec.indexOf("## Tool schema"));
Change it to:   const schemaBlock = spec.slice(spec.indexOf("## Tool schema"), spec.indexOf("\n## ", spec.indexOf("## Tool schema") + 1));
```

I ran that — nine pass, and it still fails on a code-side reword.

## One thing about the merge gate

@chrisedeson has `read` permission on this repo. Their review round — the one that found the cycle-guard shape and the null-naming bug, both of which I reproduced as fixed — does not count toward the two required approvals, and their earlier `CHANGES_REQUESTED` never blocked merge either. Only @mercyokum's approval counts. Mine is the second, so this should be mergeable now.

Both existing approvals also predate `ca257b3e`, and stale approvals are not dismissed on push here. I read that commit: it is a comment rewrite plus the removal of a dead JSDoc line. Nothing unread that affects behaviour.

Issue #1646, opened 2026-08-16
deep dive: person-evidence


DallanQ:
**Touches:** eval/tests/unit/person-evidence/, packages/engine/plugin/skills/person-evidence/SKILL.md

Audit `person-evidence` for defects nothing has caught. This starts from **no known
failure** — you are looking for things the suite passes today and should not.

**Read [`docs/skill-deep-dive-guide.md`](../blob/main/docs/skill-deep-dive-guide.md) first.** It is the
procedure: build the skill's prohibition list, read transcripts *before* scores,
spend the time on the quiet passes, place each finding in a lane, and write it
Did / Should / Gap.

## Start with this grep

`judge_context` that spells out the score branches tells the judge what to find,
and the judge takes the branch it was shown. Run this first — it is free and it
points at the files worth reading closely:

```sh
grep -l -iE '"[^"]*\bscore [123]\b' eval/tests/unit/person-evidence/*.json
```

**This skill has 1 such file(s).**

The worked example is `ut_search_records_015`, whose `judge_context` read
*"Score 2 if it uses Mullen but doesn't explain why"* — which is exactly what
happened, and the judge scored 3 anyway. Rewrite a hit to name the *dimension*
without writing the finding.

## This skill's numbers

| | |
|---|---|
| Tests in the suite | 21 |
| Rubric+base dimensions that never discriminate | **3 of 9** |
| `judge_context` files naming a score branch | 1 |

3 of 9 dimensions never take more than one value, so they cannot report a defect in whatever they nominally grade.

## The step that pays for the session

Step 6 of the guide — **turn every finding you can into a validator request**. A
hand fix closes today's instance and costs a genealogist again next run; a
validator closes the class for every future run. You supply the genealogical
rule; a developer writes the Python. The guide has the seven shapes that convert
and a four-line request template.

The measure of this task is **validator requests, not findings**.

## What to hand back

- The prohibition list from Step 1, saved for the next auditor
- Each finding in Did / Should / Gap form, with its lane
- A validator request for every finding that converts
- For grading defects you own (`judge_context`, `rubric.md`): the fix, made

## Cost

Editing anything under `eval/tests/unit/person-evidence/` or the skill body flips this
skill's run-log snapshot inactive and buys one `make eval-skill SKILL=person-evidence`
run. **Batch every finding into that one run** — that is the whole reason this is
one task per skill rather than one per defect. Since PR #1637 the annotation half
[...]

Issue #1374, opened 2026-08-06
hypothesis-tracking has no negative routing tests


Gennecis:
**Constraint, added 2026-08-12 by `/fill-ready`:** each reciprocal negative test lands in
a second skill's test directory and owes that skill a fresh `make eval-skill` run plus an
annotation pass. Nine of ten skill eval slots are held right now (research-plan by three
open PRs at once). **Pick confusable neighbours whose slots are free** and say in the PR
which ones you picked and why — do not queue a reciprocal behind a held slot.

**Touches:** eval/tests/unit/hypothesis-tracking/

`hypothesis-tracking` has 13 unit tests and none of them are `type: "negative"`. It is the only skill in the corpus with zero negative routing coverage, so nothing pins which requests should route away from it, and nothing would red if a description edit made it over-trigger.

It is also invisible to `check_negative_reciprocity.py`: with no outbound edges it is never an edge source, and no other skill names it in `correct_skill`, so it is never a target either. The lint is structurally silent on it — the gap will not surface on its own.

Work is to write negative tests for its confusable neighbours, and their reciprocals in those skills' directories per `docs/specs/unit-test-spec.md` § 6. Note each one touches a second skill's test directory and so owes that skill a fresh `make eval-skill` run plus annotation.

benter-070:
## Two of three pairs shipped in #1783 — leaving this open for the third

`hypothesis-tracking` is no longer the corpus's only skill with zero negative routing tests, and it is no longer invisible to `check_negative_reciprocity.py` as neither edge source nor target. Routing graph: **79 edges / 24 skills → 83 / 25**, all four new edges reciprocated.

Its description carries **three** DO-NOT clauses. Two are now pinned from both sides:

| Pair | Tests | Status |
|---|---|---|
| `↔ conflict-resolution` | `ut_hypothesis_tracking_014` / `ut_conflict_resolution_013` | shipped in #1783 |
| `↔ timeline` | `ut_hypothesis_tracking_015` / `ut_timeline_010` | shipped in #1783 |
| **`↔ proof-conclusion`** | — | **remaining** |

**#1783 deliberately does not carry a `Closes #1374` keyword.** With the third pair undone, auto-closing on merge would leave the remaining work recorded nowhere but a "Didn't change" paragraph inside a merged PR body, which is where deferred work goes to die.

### Why the third pair is not in #1783

[...]

PR #1783, opened 2026-08-20
hypothesis-tracking: negative routing tests + reciprocals, and the conflict-resolution over-trigger they found (#1374)

benter-070:
## Summary

**Tier: Normal.** Partial for #1374 — **deliberately does not close it.** Two of the three pairs its DO-NOT clauses call for ship here; the `proof-conclusion` pair is blocked on an eval slot (see Plan). Leaving the issue open so the remaining pair stays visible rather than being buried in a merged PR body.

- `hypothesis-tracking` had **zero** negative routing tests — the only skill in the corpus with none — so nothing pinned which requests route *away* from it, and `check_negative_reciprocity.py` was structurally silent on it (never an edge source, never a target). This adds **two reciprocal pairs**, `hypothesis-tracking ↔ conflict-resolution` and `hypothesis-tracking ↔ timeline`, taken from the two DO-NOT clauses in its own description. Routing graph: **79 edges / 24 skills → 83 / 25**, all four new edges reciprocated.
- **The new tests found a live routing bug on their first execution.** `ut_conflict_resolution_013` failed: asked to *"add James Flynn as a third possible father, and keep track of what supports and contradicts each of the three candidates,"* `conflict-resolution` activated, performed hypothesis-tracking's work, and **fabricated a `conflicts` entry** for an attribute nobody disputed. Fixed by narrowing its `description` (below); the test passes now.
- Renames `hypothesis-tracking/negative-conflict-resolution.json` → `link-supporting-assertions.json`. It is a **positive** test, and per @DallanQ's comment on #1374 its negative-sounding filename is what makes the directory *look* covered to anyone auditing routing by `ls` — it had already caused this issue's premise to be wrongly refuted once.
- Two stale documentation claims found along the way, both folded in (see "Folded in / filed instead").

## Review guidance

**Start here:** `packages/engine/plugin/skills/conflict-resolution/SKILL.md`, the `description` frontmatter — the only product-behaviour change in the diff. Two things to check. (1) The new clause is scoped: *"Do NOT use to add or track candidates **when no decision is asked for** (use hypothesis-tracking)"*. That qualifier is load-bearing — six tests expect routing *into* `conflict-resolution` and every one of them requests a decision ("sort out which one", "resolve this conflict", "tell me which birthplace is right"), so the narrowing should not touch them. All six were re-verified and pass. (2) It is **net ±0 characters** (1001 → 1001, 23 under the 1024 cap). The new clause was paid for by cutting the illustrative "(three different birthplaces)" and the adjectives "defensible"/"resolution", **not** by cutting trigger surface — all nine trigger phrases survive verbatim. If you think that parenthetical was load-bearing, that is the line to push on.

**Unsure about:** whether hand-editing a `description` without `make optimize-skill` is acceptable here. §3.2 says descriptions are "tuned empirically, not by taste," and I tuned this one by hand. My argument: the edit is driven by a failing test rather than taste, it is net-zero on length, and the optimizer builds its query set *from* the negative-test corpus — which for this pair did not exist until this PR, so running it now would tune against a set this PR just created. I think that is the right order, but it is a judgment call and I would rather you rule on it than have it pass unnoticed. **Cheap fallback if you disagree:** revert the description, mark `ut_conflict_resolution_013` `xfail` with the evidence below, and the rest of the PR stands untouched.

## Plan

**Didn't change:** the `person-evidence ↔ hypothesis-tracking` edge, and the `proof-conclusion` pair. `person-evidence` ("is this the same person?" vs. hypothesis-tracking's "could this be [candidate]?") is genuinely confusable, but hypothesis-tracking's description does not currently exclude it — a negative test would pin a boundary no doctrine states, so it needs a description edit first (lane 5). `proof-conclusion` is the third DO-NOT clause and the obvious third pair, but **PR #1545 holds its eval slot**, and the `/fill-ready` constraint on #1374 says explicitly not to queue a reciprocal behind a held slot.

**Acceptance check:** `ut_conflict_resolution_013`. Against `main`'s `conflict-resolution` description it **fails** — observed directly, and re-confirmed by a deliberate control run with the description reverted to `main` while everything else stayed on this branch. Against this branch it **passes** (`eval/runlogs/unit/conflict-resolution/v1_2026-08-19_15-24-31.json`). The other three new tests pass on both sides; they pin boundaries that already held.

**Deviated from the plan:** yes, once. The issue scopes test-writing only. When `ut_conflict_resolution_013` failed I classified it (§3.6) as lane 5 — a routing miss, not a bad test — and folded the description fix in rather than filing it. Two reasons: `conflict-resolution` was already being re-run and re-annotated here, so the fix cost no extra eval slot; and committing a test as `xfail` against a gap nobody is assigned would have been the weaker outcome. **Did not hit a step-4 stop rule** — no schema, no auth, no ADR reversal.

## Test plan

- [x] I ran `make test-all` — **no.** No `make` on this machine (Windows; #1185 notes there is no Windows equivalent). Leaning on CI, which runs the same suites. I did run every check this PR can affect directly: `check_skill_frontmatter.py` (clean — 27 skills + 4 agents), `check_negative_reciprocity.py` (no new one-directional edge; the 45 unbacked ones are the pre-existing baseline), a corpus-wide `test.id` uniqueness scan (none), and `unit-test.schema.json` validation of all four new test files.

- [x] Ran the eval and committed run logs **and** `.ann.json` for every skill whose snapshot changed — `conflict-resolution` (13 tests, `v1_2026-08-19_15-24-31`), `hypothesis-tracking` (15) and `timeline` (10, both `v1_2026-08-16_19-22-07`). All three verified **active** against the merged branch, `judge_prompt_hash` matching. Latest `conflict-resolution` run: 12 pass, 1 partial.

- [x] Changed a `description` / DO-NOT clause and added the negative test on **both** sides of the pair — `ut_hypothesis_tracking_014` and `ut_conflict_resolution_013` both ship here. Beyond the pair, all six tests routing *into* `conflict-resolution` were re-verified: `ut_timeline_003` and `ut_conflict_resolution_010` in the full suite, and `ut_check_warnings_011`, `ut_convert_dates_011`, `ut_person_evidence_017`, `ut_proof_conclusion_008` as targeted single-test runs written outside the repo tree ($0.03, nothing committed). All pass.

## Three things that are not this PR's bugs, stated so you do not have to find them

**1. Two pre-existing test failures.** Neither introduced here, and they are different problems.

- `ut_timeline_008` — **3 fail / 1 pass** across four executions. The control run above (description reverted to `main`) also failed, so this is provably independent of this change. Root cause looks like a corpus inconsistency: `ut_person_evidence_017` asks the same identity question on the **same scenario** (`mid-research-flynn`, same conflicts, zero speculative links) and expects the **opposite** skill, so the corpus discriminates on the phrase "resolve the conflict" rather than on project state. Full evidence and a suggested resolution are **documented on #1646**; see the caveat below.
- `ut_hypothesis_tracking_010` — 1 fail / 2 pass. A validator catch (`test_project_file_changes_route_through_writer_tools`): the skill modified `research.json` without a writer tool. Unrelated to routing, and it produced **zero judge dimensions**, so the sampler excludes it and rule 3 counts nothing — main now emits a CI warning for exactly that case rather than passing silently.

**2. Three routing negatives pass hollowly.** `ut_conflict_resolution_009`, `_012` and `_013` all pass on routing while producing **zero tool calls and no response at all**. The new judge advisory from #1686 surfaces this. **Two of the three predate this PR**, so it is a property of how routing negatives execute in this suite rather than something introduced here — but it does bound what the acceptance check proves: `_013`'s pass establishes that `conflict-resolution` **stood down** (`activated: false`, no fabricated conflict entry), not that `hypothesis-tracking` then did the work. The former is the defect this PR fixes. Worth noting the judge scored **1** on these three and **3** on structurally identical siblings, which is real judge inconsistency for a reviewer or a deep dive to own.

**3. No lint covers `docs/gps-research-flow.md` prose,** which is why the wildcard contradiction below survived for weeks.

## Follow-on issues

None filed. Three findings, all folded in or recorded on issues that already existed (CLAUDE.md step 1 and step 3 — the context was loaded and none needed a different reviewer):

1. **`docs/gps-research-flow.md` contradicted the skill on wildcards.** It said indexed search runs *"with no wildcards — explicit spelling variants instead."* `search-records/SKILL.md` says the opposite in three places (*"`*` and `?` both bind"*, *"the old 'no wildcards' rule was measured and withdrawn"*), `references/search-strategy-levers.md` makes wildcards **levers 8 and 9** of the zero-hit ladder and requires one before a search may be called exhausted, and `eval/tests/unit/search-records/rubric.md` lists "wildcard on a suspect letter" as a **pass** criterion — so the flow doc contradicted the rubric the skill is graded on. Provenance: `803aa26b` (#1093/#1328) withdrew the rule and corrected three sibling wildcard claims; this doc's last three commits all predate it, a site that PR missed. I checked the doc for the other three refuted claims (carries none) and confirmed its adjacent full-text claim about Soundex/fuzzy/abbreviation expansion is still true.
2. **`docs/specs/simplified-gedcomx-spec.md` §1 called `tree.gedcomx.json` "the file that eventually uploads to FamilySearch."** It is not — it holds tentative and unproven conclusions, so it is a *superset* of what could be uploaded, and upload is conclusion-gated, as the "Relationship to `research.json`" paragraph three lines further down already said correctly. Reworded to "working tree". Left the §4/§5 mentions alone; both describe when the *conversion* runs and are accurate.
3. **`negative-person-evidence-audit.json`'s `explanation` quoted the old description verbatim** and would have become a citation of text that no longer exists. Updated to the new wording.

**Recorded on existing issues rather than filed:** the `ut_timeline_008` / `ut_person_evidence_017` inconsistency is written up in full on **#1646** (person-evidence deep dive — where the one-file fix would land), with a pointer on **#1655** (timeline) so that dive does not re-derive it. **#1646 is assigned (@johnmarkpeterbrown) and the lead has since ruled on it**, delegating the fix — which shipped separately as #1794, moving `ut_person_evidence_017` onto a scenario that actually holds the conflict its request asks to resolve. So this is owned, not stranded. Not blocking this PR either way.

**Folded in / filed instead:** all three above were folded in. Nothing filed.

## A local test failure I reported and then disproved

An earlier draft of this section claimed four `npm test` failures were inherited from `main`
(`no-bom` on `assertion-classification/SKILL.md`, two `readme-catalog`, one `doc-links`).
**That was wrong and is retracted.** `assertion-classification/` is not tracked in git — it is
untracked leftover material on my working copy that `git clean` could not remove (Windows
permission errors during an earlier stash), so my local run was scanning files the repo does
not contain. On a clean checkout there are 27 tracked skills and the README catalog is
consistent, which is why **CI is green on this PR** — all 13 checks pass, `vitest` included.

Recorded rather than deleted because the mistake is instructive: a local suite run on a dirty
working copy can manufacture failures that look like someone else's, and the check that
settles it is `git ls-files` on the offending path, not reading the test output more carefully.

`no-issue-refs.test.ts`, the lint that arrived in this branch's merge and the one most likely
to catch a skill-body edit, passes on the description change here.


EdmondOware (approved):
Verified this independently rather than taking the write up at face value.

Confirmed zero negative tests existed in hypothesis-tracking before this PR, checked all thirteen files on main directly, every one is type positive, including the misleadingly named negative-conflict-resolution.json. The rename to link-supporting-assertions.json is a clean rename with no content change.

Confirmed the description edit is genuinely net zero. Computed both character counts directly, 1001 before and after. The full "Use when" clause and all nine trigger phrases are byte identical between versions. Only the illustrative parenthetical and two adjectives were cut to pay for the new clause.

Confirmed the real bug and its fix in the actual run log, not just the description. ut_conflict_resolution_013 now shows activated false, zero tool calls, correctly routed to hypothesis-tracking with no fabricated conflicts entry.

Confirmed both DO NOT clauses quoted in the new tests exist verbatim in hypothesis-tracking's own frontmatter, and that proof-conclusion was correctly left alone since PR 1545 holds that slot.

Confirmed both documentation fixes against the actual current skill files. search-records SKILL.md, its reference doc, and its rubric all currently contradict what gps-research-flow.md said about wildcards, and the provenance commit named in the body checks out against the cited issues.

Annotation discipline is clean, every dimension in the conflict-resolution review sample has an entry and every non pass one has a comment. Housekeeping is good too, superseded run logs were properly deleted rather than left stale.

Ruling on the one open question. Hand editing the description instead of running make optimize-skill is the right call here. The edit was driven by a failing test, it is net zero length, and the optimizer's negative test corpus for this pair did not exist until this PR created it, so running it now would tune against a set this PR just built. Proceeding with the hand edit as is, no need to revert to the xfail fallback.

Nice work, and the retracted local failure claim is exactly the kind of thing that should stay recorded rather than deleted.

florencemashipei (approved):
Approving. I verified the claims against a worktree rather than reading the body, and everything load-bearing holds.

## Confirmed by checking, not reading

- **The description edit is exactly what it says.** 1001 → 1001 characters, 23 under the 1024 cap, and all nine trigger phrases present verbatim — I diffed the frontmatter and counted rather than trusting the arithmetic. The `(three different birthplaces)` parenthetical and the two adjectives are the whole cost, and no trigger surface went with them.
- **Routing graph 83 edges / 25 skills**, with the unbacked count unchanged at 45 — which is the real proof all four new edges are reciprocated, better than the totals alone.
- `check_runlogs.py` → all rules satisfied, rule 4 included. `check_skill_frontmatter.py` → 27 skills + 4 agents clean.
- **The wildcard doc fix is right.** `search-records/SKILL.md:282-283` says `*` and `?` bind and to prefer explicit variants first, so the flow doc genuinely contradicted the skill, and the new wording is faithful to it. Good catch on a doc no lint covers.
- `negative-person-evidence-audit.json`'s re-quoted `explanation` matches the new description text.
- `ut_conflict_resolution_010`'s `correct_skill` does include `conflict-resolution` itself, so counting it among the tests routing *into* CR is correct.
- **Your judge-inconsistency observation is real, and sharper than you put it.** `_004` and `_011` scored **3/3** on an execution shape *identical* to `_009`/`_012`/`_013`, which scored **1/1** — same `activated: false`, zero tools, zero response. That is the 08-19 run, i.e. already under the current judge prompt, so #1734's "grade the decision, not the work" guidance did not fix it. Worth carrying that specific pairing into whichever deep dive picks this up; it is a cleaner reproduction than "three pass hollowly".

## One correction to the body

**Two of the three run logs carry a stale `judge_prompt_hash`,** where the test plan says all three match. conflict-resolution (`08-19`) is `0d186137` = current; hypothesis-tracking and timeline (both `08-16`) are `a7c9bd99`. The prompt changed on 2026-08-18 in **#1734 — "Document how to grade a negative test"** — whose diff is precisely *"You are grading the decision to route, not the work of the skill routed to. Do not fail the base dimensions because the… result the user asked for is missing."*

So the three new **negative** tests in those two suites were graded under a prompt predating the guidance written for exactly that class of test. Warn-only, not blocking, and it does not touch the acceptance check — `ut_conflict_resolution_013` sits in the run that *does* match. But it is the change most likely to move those grades, so it should not read as "all three matching".

## Ruling on the description question

**Hand-editing without `make optimize-skill` is acceptable here.** Your argument is the right one: the optimizer builds its query set from the negative-test corpus, and for this pair that corpus did not exist until this PR, so running it now would tune against a set this PR just created. Add that the edit is net-zero on length, driven by a failing test rather than taste, and preserves every trigger phrase. Revisit with the optimizer once the `proof-conclusion` pair lands and there is a real query set for the boundary.

I also agree with folding the fix in rather than committing `_013` as `xfail`. `conflict-resolution` was being re-run and re-annotated here anyway, so it cost no extra slot, and an `xfail` against a gap nobody is assigned is the weaker outcome.

## Sequencing — please read before merging

**#1814** (timeline deep dive) edits six timeline test JSONs *and* `timeline/rubric.md`; this PR adds `timeline/negative-hypothesis-tracking.json`. Both invalidate timeline's run-log snapshot, so whichever merges second owes a fresh paid timeline run. **Land this one first** — #1814 is re-running and re-annotating timeline regardless and absorbs the new test at no extra cost. No id collision: #1814 adds no new test ids.

## Notes, neither blocking nor yours to fix

**Nobody human graded the four new tests.** The samples are `_001,_002,_003,_012,_013` / `_001–_004,_008` / `_005–_008,_011`, none of which overlap `ut_hypothesis_tracking_014`/`_015`, `ut_timeline_010`, `ut_conflict_resolution_013`, and no `.ann.json` carries a correction for any of them. To be clear this is **not a lapse on your part** — you annotated all five sampled tests in each run (93 corrections, comments on nearly all, plus one extra beyond the sample in conflict-resolution), and `per-pr-review-workflow.md:258` says the unsampled ones "need nothing". The gap is in the discipline: a PR whose entire deliverable is four new tests can be fully compliant while no human grades any of them, because the sampler is random and cannot see which tests are new. Recording it here rather than filing it — the natural home is a warn in `check_runlogs.py`, which already receives `BASE_SHA`/`HEAD_SHA` and knows exactly which test files a PR added, and that is a small enough change to build rather than ticket.

**The four new ids are sequential.** `unit-test-spec.md` specifies app-generated random 3-char suffixes and spells out why: two parallel branches both take `max(N)+1`, and a duplicate id silently merges annotations. No live collision today, and renaming now would cost a run since the snapshot keys on path and content — so leave it, but use the app's generator next time.

**One thing a reviewer cannot verify, stated so it is not mistaken for a gap:** the "fails on main" half of the acceptance check rests on a control run that *cannot* be committed, since targeted single-test runs write `scratch_*` and those are gitignored. That is a harness property, not a shortcut on your part.

Nice piece of work — the negative test that found a live routing bug on first execution is the argument for the whole issue, and the retracted local-failure section is the kind of thing most PRs quietly delete instead of leaving as a lesson.

florencemashipei:
Correction to the sequencing note in my review above — I had the cost the wrong way round in one direction.

I wrote that "whichever merges second owes a fresh paid timeline run". That is only true one way. `snapshot.py::diff_snapshot_vs_disk` is explicit: *"Paths present on disk but absent from the snapshot are NOT flagged — the snapshot is the authority on what's tracked."* An **added** file is therefore invisible to an existing run log.

- **This PR first → no cost to anyone.** Its whole timeline footprint is one *new* file, `eval/tests/unit/timeline/negative-hypothesis-tracking.json`. That path is absent from #1814's snapshot, so nothing is flagged and #1814's `v1_2026-08-21_14-34-37` stays active. It does **not** owe a re-run, which is what my note implied.
- **#1814 first → this PR pays.** #1814 *modifies* `timeline/SKILL.md`, `timeline/rubric.md`, six existing timeline test JSONs and `flynn-impossibility/research.json` — all already keys in this PR's timeline snapshot, so they return `content-differs` and the run log goes stale. That is a full paid timeline run plus a five-test annotation pass on an already-approved PR.

So the recommendation stands, for a better reason: **merge this one first.** Not because #1814 absorbs the test for free by re-running (it already ran twice on 2026-08-21, before this file existed) but because an added path cannot invalidate a snapshot.

One consequence worth stating rather than discovering later: on that order, timeline's newest run log on `main` will not have executed `ut_timeline_010`, since it predates the file. The passing record stays in this PR's `v1_2026-08-16_19-22-07`, which later gets superseded, and timeline's next run picks the test up. No gate notices, and "latest run log is active" will not mean "latest run log covered every test" for that window — a cheaper trade than a paid run, but real.

And if the order slips, `eval-cosmetic-skip` is not the way out: #1814's own edits are substantive, so labelling it would relax rule 2 on genuinely stale grading rather than on a behaviour-neutral edit.

PR #1785, opened 2026-08-20
image_transcribe: surface the socket cause code on unreachable errors (#1594)

clack391:
## Summary

- Surfaces the socket-level cause code (`ECONNRESET`, `ENOTFOUND`, `UND_ERR_*`, TLS) in `image_transcribe`'s "Could not reach OpenRouter" error. Node's global `fetch` rejects with a bare `TypeError: fetch failed` and hangs the real reason off `error.cause`; the old catch kept only `error.message`, so every unreachable failure logged the useless string `fetch failed`.
- This is the "First deliverable (startable today): make the failure classifiable" from issue #1594. It does not fix the reachability failures, decide the retry question, or touch OCR quality. Tier: Normal.

## Review guidance

**Start here:** `describeFetchError` in [image-transcribe.ts](packages/engine/mcp-server/src/tools/image-transcribe.ts) and the one-line catch it feeds. The thing to check is the cause walk: it descends `error.cause` (depth- and cycle-bounded to 6 levels) and flattens `AggregateError.errors`, prefixing each layer with its `.code` when present. The outer message shape `Could not reach OpenRouter. (...)` is unchanged; only the parenthetical is richer. No downstream code parses this string (grepped `packages/engine/plugin`, `eval/harness`, `apps` for the phrase).

**Unsure about:** Nothing blocking. The join separator ` <- ` and the 6-level depth cap are judgment calls, not correctness constraints.

## Plan

**Didn't change:** `http.ts` (no retry, no timeout change), the default model, the single-reader design, `image_read` vision, and every OCR-quality path. The issue forbids each of these in this PR by name. The issue's "Touches:" line lists `http.ts`, but that was for the deferred retry path; a classification-only change touches neither the timeout wrapper nor retry.

Also deliberately left: the spec's error-table row for "OpenRouter unreachable" ([image-transcribe-tool-spec.md](docs/specs/image-transcribe-tool-spec.md)). Its `(...)` placeholder still describes the message accurately (the shape is unchanged, only the parenthetical detail is richer), and the issue reserves the spec's runtime-failure section for the later retry task. Editing it now would front-run that task.

**Acceptance check:** two new tests in [image-transcribe.test.ts](packages/engine/mcp-server/tests/tools/image-transcribe.test.ts) that fail on `main` and pass here: "surfaces the socket-level cause code hidden on error.cause (#1594)" (asserts `ECONNRESET` reaches the message) and "flattens an AggregateError cause to surface ENOTFOUND (#1594)". Break-tested: with the old `error.message`-only catch both produce `Could not reach OpenRouter. (fetch failed)` and fail.

## Test plan

- [ ] For every skill whose run-log snapshot I changed ... (no skill snapshot touched)
- [ ] If I changed a skill's `description` frontmatter ... (no skill/description touched)

`make engine-test` (the issue's designated check): 2445 passed, 103 files. No CI job checks transcription fidelity, and this PR does not add one; that half needs a genealogist-built fidelity harness, per the issue.

## What this does not solve (stays open under #1594)

- The reachability failures themselves. Over the committed e2e corpus, 29 of 388 `image_transcribe` calls still carry the unreachable/timeout error, across 4 fixtures (5 run logs). This PR makes the next such failure classifiable; it does not reduce the rate.
- Host-side vs OpenRouter attribution. Existing logs already hold the bare `fetch failed`, so they cannot be classified retroactively; only runs from here forward carry the code.
- The retry decision (deferred to the spec's runtime-failure section) and the OCR-quality halves (fabricated headings, corrupted patronymics), which are genealogist adjudication.

## Follow-on issues

**Folded in / filed instead:** Nothing folded in, nothing filed. The deferred work is already tracked in #1594 itself, which this PR does not close.

Refs #1594 (does not close it).


EdmondOware (approved):
Checked out the branch and ran the two new tests directly, both pass. Then patched describeFetchError back to the old error.message only behavior and reran, both tests fail with exactly the useless Could not reach OpenRouter. (fetch failed) the PR describes. Restored the file afterward, full suite is 24 of 24.

Confirmed the catch block wraps the actual OpenRouter fetchWithTimeout call, not some unrelated fetch. Confirmed fetchWithTimeout's claimed behavior directly in http.ts, a real timeout throws a fresh error with a clean message and no cause, everything else rethrows untouched, so the cause walk only ever sees useful chains on the network failure path, not the timeout path.

Confirmed the no downstream parser claim myself, grepped packages/engine/plugin, eval/harness, and apps for the literal string, no hits.

The cause walking logic is sound, cycle bound on object identity, depth capped, AggregateError flattened, dedup on repeated labels, and every property access is a safe optional cast with no crash risk.

Scope discipline is good, http.ts, retry, timeout, and OCR quality are all correctly left alone as the issue specifies.

jud-sdev (approved):
I verified the cause-walk by running the real describeFetchError under node, not by reading it.

- New behavior surfaces the socket code: ECONNRESET on error.cause, and ENOTFOUND flattened out of the AggregateError. With the old error.message catch both collapse to (fetch failed), so the two new tests genuinely bind and would be red on main.
- Every defensive edge holds: a self-referential cause does not loop, a 10-deep chain is capped at 6, a non-Error cause degrades to String(), and an empty error falls back to unknown error.
- Nothing downstream parses the message (grep of plugin, harness, apps is empty), it test-merges clean with main, and vitest is green on this head.

Scope is disciplined too, exactly the classifiable-failure deliverable of #1594 with retry, timeout, http.ts, and OCR quality all left alone as the issue requires. Nothing to change.

PR #1097, opened 2026-08-01
skills(search-records): log pre-1880 census household relationships as inferred

Emruthwill:
Closes #955.

The US federal census had no "relationship to head of household" column until
1880. The 1850/1860/1870 schedules record only names, ages, birthplaces and
occupations in household order, so any "head"/"wife"/"son"/"daughter" read off
one of them is an inference from surname, age and listing order — not something
the record states.

In the `mary-dwyer-father` e2e run the agent logged an 1860 Springfield household
as *"head Daniel Ragan + wife Margaret + daughter Hannah + son Patrick"*, which
reads as though the schedule documented those ties. That files an inference as
the stronger kind of evidence and tells the next reader the question is settled.

### What changed

- **`SKILL.md` Step 4** — a triage bullet placed next to the existing role
  reasoning ("a 5-year-old cannot be Head of Household"), which never noted that
  pre-1880 there *is* no stated role. It calls out the record's own
  `ParentChild`/`Couple` edges explicitly: those are the indexer's inference from
  the same three signals, and they are what makes the relationship feel
  documented.
- **`references/census-field-availability.md`** — which fields each schedule
  collected, US 1790–1950 and England & Wales 1841–1911, so the same class of
  overclaim (parents' birthplaces on an 1870 census, birth month outside 1900)
  has somewhere to be checked.

A Step 5 restatement was in the first draft and was **dropped** in `f7fd2fb5`
per review: it duplicated the Step 4 bullet, and `search-records` is called
often enough that the hot path has to stay short.

### Scope

The extraction lane already had this doctrine — `record-extractor.md` and
`information-classification-at-extraction.md` both classify pre-1880
relationships as `indirect`. The gap was one-sided, so nothing changes on the
extraction side.

Lane 3 (record-type craft gap) per `docs/skill-lifecycle.md` §5, so the unit
suite is not a gate here. Happy to run it if a reviewer wants the evidence.

### On the table's accuracy

The second commit checks the year boundaries against the National Archives,
IPUMS, the FamilySearch Research Wiki and Digital Panopticon, and records the
sources in the file — writing genealogical claims from memory is the same
failure this PR exists to prevent, one level up.

That found one real error: age at first marriage was marked "unique to 1930". It
*is* 1930 column 15, but the 1940 supplementary schedule asks it too, which the
file's own 1940 section already said. Now corrected to state the distinction.

The Sources section also names what was **not** re-verified — the 1890 question
list, the language/veteran columns, the 1950 sample-line detail — so a reviewer
[...]

PR #1786, opened 2026-08-20
search-records: assert the pre-1880 census inferred-relationship rule (#1284, #1283)

### LINKED ISSUE #1284: search-records: nothing tests the pre-1880 census inferred-relationship rule
**IMPORTANT**: If you do this, do issue #1283 at the same time. Both edit the
`search-records` snapshot and both need the same paid `make eval-skill SKILL=search-records`
run and annotation pass — #1283 is a one-line wording fix to `SKILL.md:329`. Landed apart
they buy two runs.

**Touches:** eval/tests/unit/search-records/

Filed from the PR #1097 review, which added the rule but no test for it.

PR #1097 (issue #955) added the doctrine that a pre-1880 US census household must
be logged with its family structure marked **inferred** — "Daniel, Margaret,
Hannah in one dwelling; family structure inferred from surname, ages and order,
not stated" rather than "head Daniel + wife Margaret + daughter Hannah".

No test asserts it. The suite has 24 tests, unchanged across both run logs in that
PR, and three of them already run pre-1880 census searches:

- `attachment-triage.json` — 1850 census, Patrick Flynn
- `maiden-name-pre-marriage-search.json` — 1860 census, Sarah Mullen
- `child-middle-name-not-surname.json` — 1860 census, Sarah Mullen

Each asserts something else (attachment status, maiden-name selection, middle
initial) and none looks at how the household is phrased in the log note. So the
rule ships with nothing catching a regression, and the next prose edit to Step 4
can undo it silently.

Lane 3 in `docs/skill-lifecycle.md` §5 permits landing without the unit suite as
a gate, so this is follow-up work, not a defect in that PR.

Cheapest route is probably a judge_context addition to one of the three tests
above rather than a 25th test — the search behavior under test is unchanged, only
the log phrasing assertion is new. Worth checking against `rubric.md` first to see
whether "Log quality" already has a dimension this belongs under.

Whichever way it goes, it needs a re-run + re-annotation of `search-records`, so
batch it with other work on this skill rather than paying for a run on its own.

florencemashipei:
Closes #1284. **Refs #1283 — deliberately not closed**; its wording ask was met but
the resulting shape is wrong (see "#1283 — the edit is the wrong shape" below).

Batched because both edit the `search-records` snapshot and both needed the same
paid `make eval-skill SKILL=search-records` run. Landed apart they buy two runs.

## What this is

**Lane 2 + lane 4 mixed, and the PR should be read as lane 2.** #1284 asked for a
test; the test is a deterministic validator, not a `judge_context` clause, and
the reason is measurement rather than preference — see "Why not judge_context".

PR #1097 added the doctrine that a pre-1880 US census household must be logged
with its family structure marked inferred: 1850/1860/1870 carry no "relationship
to head" column, so the record's own `ParentChild`/`Couple` edges are the
indexer's inference from surname, ages and listing order, not a stated field.
Nothing asserted it.

## The rule was already being violated

It is not a dormant regression risk. Replaying the inference marker across all
five run logs, per pre-1880 census test:

| test | 08-13 | 08-14 | 08-17a | 08-17b | 08-20 (this PR) |
|---|---|---|---|---|---|
| `_001` execute-census-search | ✅ | ✅ | ✅ | ✅ | ✅ |
| `_010` write-result-sidecar | ✅ | ✅ | ✅ | ✅ | ❌ |
| `_012` attachment-triage | ❌ | ✅ | ❌ | ❌ | ❌ |
| `_013` same-person-near-match | ✅ | ✅ | ❌ | ❌ | ✅ |
| `_014` same-person-conflict | ❌ | ✅ | ✅ | ✅ | ✅ |
| `_015` maiden-name-pre-marriage | ❌ | ❌ | ❌ | ❌ | ❌ |
| `_017` maiden-to-married-shift | ❌ | ✅ | ✅ | ✅ | ✅ |
| `_027` spouse-absent-on-census | ❌ | ❌ | ❌ | ❌ | ✅ |

17 of 32 runs carried the marker before this PR, and **all 32 passed**. Compliance
is a coin flip in the 3–6 of 8 band, not a per-test property — `_010` had been 4/4
and regressed on this run; `_013` and `_027` had been failing and complied.

`_015` has never complied in five runs, and does the damage the rule exists to
prevent — this run's note:

> "Sarah A. Mullen (b. 1852, Wisconsin) found in William Mullen household, Dodge
> County, WI, 1860 census. Household also includes Margaret Mullen (b. 1830,
> Ireland). matchScore 0.93, unattached. **William (b. 1825, Ireland) and Margaret
> align with tree persons I4 and I5.**"

That is an indexer inference laundered into corroboration of tree parents, off a
census year with no relationship column.

## Why not `judge_context`

#1284 suggested a `judge_context` addition as the cheap route. The measurement
above rules it out: an LLM dimension graded against behaviour that is itself a
coin flip flakes with it. `docs/skill-lifecycle.md` names this case directly —
"judge score swings run to run → genuine skill inconsistency" — and the skill
*did not* follow its written instruction, so it is not a grading defect.

`rubric.md`'s "Log quality" is the right dimension and already disclaims what
validators own ("deterministic validators already own entry existence and
`outcome` honesty — do not re-grade any of those here"), so this PR follows that
convention: the validator owns the check, the rubric gets a one-line pointer, and
nothing double-counts.

## What the validator asserts

`test_pre1880_census_structure_marked_inferred`, tag-gated on
`pre-1880-census-household`: a positive `record_search` log note that describes
the household must also mark the structure inferred.

It is a **positive requirement, not a phrase blocklist**, and that is deliberate.
No regex separates a role read off the census from a tree-side search anchor —
`_027` names "wife Catherine" and then says she is absent from the record. A
blocklist fails that note; a marker requirement does not care.

Scoped to the authored `notes` field, which is where the household write-up lands
and the only place a validator can see it (the harness hands validators state and
tool calls, not response text). Nils and notes describing only the focus person
are skipped; both are pinned by tests.

**Tagged all eight pre-1880 census tests, not the three #1284 named** — its list
predates `_027` and missed `_001`, `_010`, `_013`, `_014`. `_016` stays untagged:
its plan item targets 1880, where the relationship column does exist.

## Run result

24 pass, 3 fail, 1 aborted of 28. Run log `v1_2026-08-20_11-04-51`, active.

**All three failures are this validator, all three genuine, zero false
positives** — `_010`, `_012`, `_015`, each listing the household with no
qualifier. That is the finding, not an obstacle to it. The remedy is #1642's,
which owns this skill's next paid run.

Reviewer note: `review_sample`'s `is_gradeable` excludes validator-failed tests
(a skipped judge leaves zero dimensions), so rule 3 warns rather than blocks and
**CI never asks anyone to read the three failures.** They were read by hand.

The five that passed the validator were also read, to check it is not passing for
the wrong reason. `_014` ("this role is an indexer inference from listing order,
not a stated fact") and `_013` ("No relationship column in 1850 census — household
structure is inferred") hedge substantively, not by tripping a marker word — `_013`
asserts the father relationship *and then* qualifies it, which is the rule working
as written. `_027` passed by a different path: it described no household this run,
so the household guard skipped it before the marker check. That guard is live on
real data, not dead code.

## Annotation

5 sampled tests, 35 dimensions, rule 3 complete. **Two judge scores corrected,
both downward** — the judge scored 3 on all **32 scored** dimensions; the other 3
were N/A (Nil escalation on `_013`, `_014`, `_016`).

- **`_016` Log quality 3 → 2.** Persisted `query` was
  `{"surname":"Price","givenName":"Sarah"}` — 2 of the 12 parameters sent,
  missing the Dodge County / 1880 scope and the `fatherGivenName: James` anchor
  the response itself called its key narrowing choice. The judge scored 3 by
  quoting the chat summary as though it were the persisted field.
- **`_013` Result triage 3 → 2.** A 0.52 needs-review candidate written up under
  the heading "Top match" and as "a strong candidate", with both discrepancies
  (surname `Flyn`/`Flynn`, 3-year age gap) called non-disqualifying, then
  extraction offered. `SKILL.md:408-419` forbids each of those verbatim —
  "never as a 'Top Match'", "not even as a question", extraction "available again
  only once an independent anchor confirms identity". Notable because this exact
  behaviour on this exact test is `docs/skill-deep-dive-guide.md`'s worked
  example, and the symmetric guard it prescribed **is** in the judge_context the
  judge read.

**A limit of this PR's validator, stated plainly.** `_013` *passes*
`test_pre1880_census_structure_marked_inferred` — the inference marker is
present — and then leans on the inferred household structure as corroboration
for identity anyway ("Listed in household of Thomas Flyn — consistent with
candidate father Thomas Flynn (I2)"). The validator checks that the hedge is
*present*, not that the argument *respects* it. `_015` fails for omitting the
hedge; `_013` passes while committing the substantive error the hedge exists to
prevent. Result triage owns that, and caught it. The validator is a floor.

**Observed, not scored** (the annotator considered and held at 3, recorded here
for #1642). Of the **11 tests that ran two or more `record_search` calls, 8 logged
them via a batched `ops` array** rather than one call per search — `_023` (ops[8]),
`_018` (ops[7]), `_011` (ops[6]), `_024` (ops[6]), plus `_015`, `_021`, `t9p` and
`nickname_bitsie_in_record` at ops[2]. Three did not: `_028` (3 calls / 3 searches),
`_022` (2/2), `_026` (6 calls / 4 searches). `SKILL.md:572` says "Log each retry as
a separate `research_log_append` call immediately after it completes — do not batch
log calls at the end," while the tool offers the array. `_011` and `_024` are the
clearest violations — both ran a full nil-lever ladder first and narrated the batch
("Now logging all nil searches as a batch"; "Now logging all six searches in one
atomic batch call"). That three tests complied shows the instruction is followable,
so this is inconsistent compliance rather than an impossible rule — but batching is
the majority behaviour, and the crash-safety rationale is live: `_023` aborted on
`max_turns` this run, and had it aborted before its single ops[8] call, eight
searches would be unrecorded. Separately, `_024`'s surname
wildcard `Wilkin*` cannot match `Wilkens` — the variant that test stocks
`record-search-wilkens-variant-no-results` for, unused this run; `Wilk*` would
have caught it.

## #1283 — the edit is the wrong shape. Reviewer, please read this one.

**#1283 is deliberately NOT closed by this PR**, and the `SKILL.md` line it added
should be reworked. I am landing it anyway; the reasoning and the alternative are
below so you can overrule me.

`:456` now reads *"**Before summarizing any census household, read
`references/census-field-availability.md` for that year** — same caution for any
field that year didn't collect."* The original was *"Same caution for any field
that year didn't collect: `references/…`"*.

**Two measurements, and they point the same way.**

*It did not work.* On the run immediately after the rewrite, **zero of the eight
census tests read the file**, while `collection-quirks.md` — the `:248` pointer
whose convention this copied — was read by five other tests. Read instrumentation
is fine (24 `Read` calls captured), so the negative is real.

*And it should not have been written that way.* A skill's reference file exists
for text needed **only in certain situations**; content needed on almost every
invocation belongs in `SKILL.md`. The reference file states this contract about
itself, at line 182: *"The rule in `SKILL.md` Step 4 is short because it is read
on every invocation."* The split is already correct — the load-bearing rule (no
relationship column, mark the structure inferred, with the ❌/✅ phrasing) is
inline at `:447-454`; the 243-line per-year tables are the situational half. The
original wording named a **situation** ("any field that year didn't collect") and
merely lacked a verb. The replacement made the read **unconditional**, which is
wrong either way: if it were truly always needed the content would belong inline,
and it is not always needed because the part that is, already is.

**The fix, for #1642's run** — keeps the verb #1283 asked for and restores the
trigger:

> **When a match turns on a field — or before calling one absent — check that
> year's entry in `references/census-field-availability.md`.**

**Why it is not fixed here.** `SKILL.md` is in the snapshot, so amending *or
reverting* this line flips rule 2 red and buys another full run plus a fresh
five-test annotation. `eval-cosmetic-skip` is not available honestly — it is for
behaviour-neutral edits, and this one is being changed precisely because we think
it affects behaviour.

**Why it is low-stakes enough to defer.** Nothing load-bearing is lost when the
reference goes unread: the pre-1880 rule is complete inline, and this PR's
validator now enforces it deterministically. The failure mode is a *wasted* read,
not a *missing* rule. Note the honest caveat — that is a latent inefficiency that
gets worse as instruction-following improves, not a dead letter, so "models
currently ignore it" is not the justification.

**Root of the confusion, recorded so it is not repeated.** #1283 argued from the
`record-extractor` measurement, and CLAUDE.md's "no reference files, the body is
self-contained" rule is scoped to **agents** — precisely because agents carry no
sibling files. Skills are the opposite: the lane rule sends skill findings *to*
`references/`. Agent doctrine was applied to a skill, and this PR repeated it.

## Not caused by this PR

- **`_023` aborted on `max_turns`** — a Norwegian marriage test with no census
  path. Turn history across six observations against a cap of 20: 13, 15, 17, 14,
  **21 (abort)**, 15 (scratch re-run, passed). The test has no headroom and will
  keep aborting whenever the model reaches for an extra variant. One-line fix
  (`execution.max_turns`) deliberately not made here: `execution` is not a
  cosmetic test field, so it would invalidate the snapshot and buy a second run.
  For #1642.
- **`_003` `routing_negative_judge_fail`** — pre-existing, first appeared in the
  08-17_18 run. Test passes.

## Known gap this PR touches: #1757 (closed NOT_PLANNED)

`test.tags` is stripped by `snapshot.normalize` as a cosmetic field
(`snapshot.py:47`), but tags gate validators. This was **#1757**, which was closed
`NOT_PLANNED` on 2026-08-20 — so the gap is an **accepted risk, not pending
work**, and nothing is filed here. Recorded because this PR adds a tag-gated
assertion and therefore sits on top of it.

Evidence added as a comment on the closed issue: `test_search_records.py` goes
from 7 to 8 tag-gated assertions against `origin/main`, and the hole is
**bidirectional**. #1757 framed it as renaming, mistyping or dropping a tag; this
PR exercises the additive direction — a tags-only change introducing a new gate
would leave the run log **active** with the new assertion never exercised.
Verified directly: the tag-only edits in this PR leave those test JSONs' snapshot
hashes unchanged. Moot for this PR only because `SKILL.md` and `rubric.md`
invalidate the snapshot anyway, so the new assertion did run — and failed 3 of the
8 tagged tests.

(I deliberately do not restate #1757's "81" total: my own count of
`not in test.get("tags"` across `validators/*.py` gives 79, and I could not
reconcile the methods. The 7→8 delta is what was verified.)

## Verification

- `make harness-test`: 2411 passed, 3 skipped, 0 failed.
- 14 new validator unit tests under `tests/unit/`, because `pyproject.toml` sets
  `testpaths = ["tests"]` and nothing in `validators/` is otherwise collected.
  Every `notes` string in them is quoted from a committed run log rather than
  invented to suit the check.
- Validator replayed against all four prior run logs before committing: 17/32,
  reproducing the hand analysis exactly.


clack391 (commented):
I reproduced every quantitative claim independently from committed artifacts, and all of them matched: 17 of 32 pre-PR runs carried the inferred marker, the full 8x5 per-test compliance table, 0 of 8 census tests reading census-field-availability.md (with collection-quirks.md read by 5 other tests), and the validator firing on _010/_012/_015 while passing the other five tagged tests. The validator's unit test proves both directions (a compliant note passes, a non-compliant one fails via pytest.raises), handles the tree-side vs record-side case so a search anchor is not flagged as a record-side role claim, and names every offender. No 25th test was added, the #1283 and #1284 work is batched into one run as both issue bodies require, the run log and annotation are shipped, and CI is green. Nice catch excluding _016: the issue body listed it as 1860, but it is 1880 (relationship column exists), so it is correctly out of scope.

Two decisions are the genealogist code-owner's and the lead's to bless rather than mine, and both are documented in the PR and invite overrule:

1. A deterministic validator instead of the judge_context Dallan suggested on #1284. The 17/32 coin-flip reproduces, so an LLM dimension graded against that behavior would flake with it. I think the validator is the right tool here, but it does deviate from the lead's comment, so it wants his sign-off.

2. Landing #1283's wording even though it measured non-working (0 of 8 tests read the file) and wrong-shaped, with the corrected rewrite deferred to #1642. That is the efficient choice given the paid run is already spent with the line in place, and #1642's body records the full plan (the 0-of-8 measurement, the wrong-shape reasoning, and the corrected rewrite).

One small non-blocking item on traceability: #1283 is still open, the wording it suggested is now the line in the tree, and its only comment is the older stale-citation note, so someone landing on #1283 directly would see a fix that looks already applied. I have added a one-line comment on #1283 pointing at #1642 so it is not mistaken for still-needed. No re-run involved.

No blocking defects found. To unblock merge this still needs a second senior-genealogist code-owner approval.

clack391 (approved):

EdmondOware:
Senior genealogist review, approving.

Independently reproduced the key claims rather than trusting the writeup:

- Run outcome matches exactly: 24 pass, 3 fail (_010, _012, _015), 1 aborted (_023) of 28.
- Read the validator's regex logic directly. It is a positive requirement, not a blocklist, bare role words like wife or father are deliberately excluded from the trigger patterns, only forms like head of household or as his wife fire it. That is what correctly lets _027's wife Catherine, absent from the record pass without a false flag.
- Spot checked one of the validator's own unit test fixtures against the actual committed run log v1_2026-08-17_18-06-27, it is a byte for byte quote, not invented to fit the check.
- Verified the annotation independently, both corrections are real and downward, _016 Log quality 3 to 2 and _013 Result triage 3 to 2, 35 total entries matching 32 scored plus 3 N or A.
- Verified the number 1283 measurement myself by counting Read calls in the new run log, 0 reads of census field availability dot md, 5 reads of collection quirks dot md, exactly as claimed.
- Verified test_search_records.py genuinely goes from 7 to 8 tag gated assertions against main.

On the two calls left to a genealogist to bless. A deterministic validator over judge_context is right, a coin flip behavior would make an LLM dimension flaky, and the validator's design correctly avoids the tree side versus record side confusion a naive check would fall into. Landing number 1283's non working wording with the fix deferred to number 1642 is also reasonable, it is disclosed plainly, measured honestly, and the corrected wording is already drafted for whoever picks up that PR.

No blocking defects found. Approving.

EdmondOware (approved):
Approving as the second senior code owner review, see comment above for the full verification.

chrisedeson (commented):
## Second review: PR #1786 (search-records pre-1880 census inferred-relationship rule, #1284, #1283)

Merged 2026-08-20 (`900bf20d` / head `856c4473`), requested as a second pass after clack391 (senior developer) and EdmondOware (senior genealogist) already approved. Full pass per `.claude/skills/review/SKILL.md`, adapted for a post-merge PR: no branch to push fixes to, so this is a report, not a patch. Verdict: approve stands — nothing here should have blocked the merge, and the one finding below is bounded and forward-looking.

### Does / Intent

Adds a deterministic validator (`test_pre1880_census_structure_marked_inferred`) that fails a `search-records` run when a log note states a pre-1880 US census household's family structure as fact instead of marking it inferred — 1850/1860/1870 censuses carry no relationship-to-head column, so any "head"/"wife"/"son" read off one is the indexer's inference, not a stated field. Tags all 8 relevant existing tests, adds a one-line rubric pointer, adds 14 new validator unit tests, and separately lands a small `SKILL.md` wording change from #1283 that the author's own PR body flags as measured-not-working (0 of 8 census tests read the referenced file after the change), with the real fix deferred to #1642. Diff matches intent for #1284 fully; for #1283 the drift is disclosed by the author, not hidden.

### Existing reviews

clack391 and EdmondOware, both APPROVED at commit `856c4473`, the exact head that merged — no staleness. Both independently reproduced the quantitative claims (17/32 historical compliance, the 8×5 per-test table, the 0-of-8 reference-read count, the 7→8 tag delta) and read the validator's regex logic directly rather than trusting the write-up. Those hold; this pass adds new ground, not a re-check of theirs.

### Verified

Full `eval/harness` suite, run in a fresh worktree at the merge commit (`.claude/worktrees/pr1786-review`): 2425 passed, 8 skipped, 1 failed. The 1 failure (`test_mock_mcp.py::test_successful_call_carries_no_is_error`) is self-diagnosed as "engine build absent," the same reason 6 sibling tests already self-skip — environmental to a fresh worktree with no compiled engine, not a regression, and this PR touches no engine code. The new validator's own 14 tests: 14/14 passed. Checked every commit landed on `main` since the merge for anything touching `search-records`: none found, so no post-merge collision. Did not run vitest/tsc (engine untouched) or `apps/server` pytest (unrelated).

### [CRITICAL] (8/10) `eval/harness/validators/test_search_records.py:24-34` — the household-role trigger vocabulary is narrower than the ways a model can actually describe one

```python
_HOUSEHOLD_MENTIONS = ("household", "dwelling", "co-resident", "coresident", "enumerated with", "living with")
_ROLE_ASSERTIONS = (r"head\s+of\s+household", r"household\s+head",
                     r"\bas\s+(?:his|her|the)?\s*(?:head|wife|husband|son|daughter|father|mother)\b")
```

Demonstrated directly against the shipped code, not just read: a note like "...along with his wife Mary and their son Thomas, directly corroborates tree persons I2 and I3" has the same flat-assertion shape as the real `_015` offender this validator exists to catch. It trips neither list (no "household"/"dwelling", no "as wife") and the validator raises nothing. All three of the PR's own "should-fail" fixtures happen to contain the literal word "household," so the new 14-test suite doesn't exercise this boundary either.

Bounded severity, stated plainly: across all 5 committed run logs to date, every real model-produced violation happened to use a covered form. This hasn't caused a false pass in production. It's a latent gap, not a proven miss.

Fix isn't a one-line widen. `_ROLE_ASSERTIONS` needs to catch bare "wife NAME" / "son NAME" phrasing too, but doing that naively breaks `test_a_tree_side_relative_is_not_a_record_side_role_claim`'s legitimate carve-out (a note naming a tree-side search anchor who turns out absent from the record — "wife Catherine... NOT named as spouse"). The "as" requirement exists specifically to dodge that false positive; removing it without a replacement brings it back. This needs a short design pass, not a quick patch. Recommend folding into #1642 (open, "deep dive: search-records," already named in this PR as the owner of the skill's next paid run and its other deferred fixes) rather than filing a new issue.

### Not checked

- Didn't re-verify what both existing reviewers already independently reproduced (the 17/32 table, the annotation corrections, #1283's 0/8 measurement). No reason to doubt two independent reproductions that agree with each other and the PR body.
- Didn't build the engine, so the 1 harness failure above is diagnosed from its own error text and the 6 self-skipping siblings, not from a fully green run.
- Post-merge review: no PR branch exists to push a fix to, so nothing was edited directly. The finding above is a note for #1642, not a patch.

PR #1794, opened 2026-08-20
person-evidence: ut_person_evidence_017 needs a scenario holding the conflict it asks to resolve (#1646)

benter-070:
## Summary

**Tier: Normal.** Addresses the identity-boundary finding on #1646, which the lead delegated: *"Go ahead and fix `ut_person_evidence_017` the way you see fit."*

`ut_person_evidence_017` asked the router to *"sort out which one is really our subject **and resolve the conflict**"* while running on `mid-research-flynn` — a project holding **one resolved fact conflict and zero speculative `person_evidence` links**. There was no unresolved conflict in state to resolve, so the routing it asserts could only be justified from the phrasing of the request, not from what the project held. It moves to **`flynn-identity-geographic-thin`**, which carries `c_002` and `c_003` — both **unresolved identity** conflicts over which Patrick Flynn is the subject — plus one `speculative` link, which `conflict-resolution/SKILL.md` § "Identify conflicts" names as an unresolved identity question it owns.

The functional diff is one line.

## Review guidance

**Start here:** `input.scenario` in `eval/tests/unit/person-evidence/negative-resolve-competing-identities.json`, and the question of whether `flynn-identity-geographic-thin` is the right destination. It is the scenario `ut_conflict_resolution_007` uses for the same kind of adjudication request, and that test does not flap.

`correct_skill` deliberately stays `["conflict-resolution"]`. My first write-up on #1646 floated flipping it; that would have been wrong on two counts — it contradicts `person-evidence`'s own DO-NOT clause (*"resolve a conflict where multiple candidates compete"*), and it deletes the `person-evidence → conflict-resolution` edge that `check_negative_reciprocity.py` tracks. The correction is recorded on the issue.

**Unsure about:** nothing on the fix itself. The judgement worth a second opinion is whether the request's *wording* ("resolve the conflict") should be sufficient on its own, with state irrelevant. I have assumed not — that a request to resolve a conflict against a project holding none is asking for work there is nothing to do — and the prose and `judge_context` now say so explicitly. If you think phrasing alone should decide, this fix is unnecessary and the flakiness lives somewhere else.

## Plan

**Didn't change:** `conflict-resolution`'s description. `ut_timeline_008` is **3-of-4 failing on `main`** and is *not* fixed here. Its request is record attribution, not conflict adjudication, and it loses to `conflict-resolution` because that description claims *"identity-level conflicts where multiple candidate persons or records genuinely compete"* broadly, with no requirement that a conflict exist in state. Narrowing that would also touch `ut_conflict_resolution_003` and `_007`, so it is a separate decision and explicitly outside what was delegated.

**Acceptance check:** `ut_person_evidence_017` passes on `eval/runlogs/unit/person-evidence/v1_2026-08-20_15-53-03.json`. It also passed before this change — the fix is not about flipping a red test, it is about the test being justified by state rather than by wording, which is what makes it stop being a coin-flip against a router that reads both.

**Deviated from the plan:** no.

## Test plan

- [x] `make test-all` — **no `make` on this machine** (Windows; #1185). Leaning on CI. Ran the checks this PR can affect: `check_negative_reciprocity.py` (the `person-evidence → conflict-resolution` edge is preserved), `unit-test.schema.json` validation of the edited file, and the two packaging lints that arrived in this branch's merge (`no-issue-refs`, `gitattributes`) — both pass.

- [x] Ran `person-evidence` and committed the run log **and** its `.ann.json`: `v1_2026-08-20_15-53-03`, verified **active** after merging `main`. **16 pass, 3 partial, 2 fail of 21.**

## The two failures are not this change

A one-line `scenario` edit cannot alter another test's execution, and both failures are on scenarios this PR does not touch. Both are written up on #1646:

- **`ut_person_evidence_n7v`** — flaky **1-in-3**, and the single discriminator is whether the skill calls `same_person` at all. When it skips it, it writes three `confident` links with a **null `match_score`** — the condition #1550 instrumented, occurring after #1550 merged. A live instance of what the write boundary exists to catch.
- **`ut_person_evidence_001`** — fails 2 of 2 on audit rationale quality. The human annotation **confirms that dimension at 1** while correcting the other three upward, so the defect is real and narrower than the judge scored it.

Seven tests changed outcome versus the last committed run (2026-08-12), **in both directions** — 3 improved, 4 regressed. Bidirectional churn on a one-line diff is variance and eight days of `main`, not causation.

## A harness observation worth someone's attention

Chasing whether #1550's warning fired, I found the `research_append` response captured as **0 characters** — then that this is universal: **284 MCP tool calls across every unit run log from the last two days, zero captured responses, all 13 tools.** The unit harness records `tool` and `args` but never the response, so a warn-only guardrail riding `validation.warnings` on a *successful* response is structurally invisible in this tier — #1550 cannot be calibrated from the unit corpus. A usable proxy needs no harness change: `same_person`'s **absence** from `tool_calls` is recorded and greppable, which is how the 1-in-3 rate above was measured. Written up on #1646; no issue filed.


T-FEH (approved):
Approving — the fixture inconsistency is real and worth removing. But the PR claims more than the change does, and I want to be precise about that, including where I got it wrong myself on a first read.

**The premise verifies exactly.** `mid-research-flynn` holds one conflict, `c_001`, `conflict_type: "fact"`, `status: "resolved"`, and zero `speculative` links — so the old test asked the skill to resolve a conflict the project did not have. `flynn-identity-geographic-thin` holds `c_002` and `c_003`, both `conflict_type: "identity"`, both `unresolved`, plus `pe_007` `speculative`. And `c_002`'s description is literally *"Which of the two Patrick Flynns recorded in 1850 Schuylkill County is the subject?"* — the prompt's question, in the fixture. Good catch in #1646.

**But this is a fixture-consistency fix, not a behaviour fix, and the body should say so.** Two things settle it:

- In the new run, `017` routed with `num_turns: 0` and `tool_calls: []` — one `Skill` builtin straight to `conflict-resolution`. **The router never opened `research.json`**, so it never saw `c_002`/`c_003`. The state swap cannot have changed the decision.
- `017` passed **5/5** on main (07-29, 08-03, 08-05, and both 08-12 runs). The body's "what makes it stop being a coin-flip" is not supported by the record.

That does not make the change worthless — an inconsistent fixture is a trap for the next reader, and for any future test that *does* read state. It just isn't what the title implies.

**A correction to my own first read.** I initially credited the new `judge_context` bullet as raising the bar — guarding against a decline that reasons only from the word "conflict". It cannot. `orchestrator.py` grades a negative test with a non-empty `correct_skill` on routing alone, and says so in terms: *"its scores must NOT flip a correctly-routed test."* The outcome is `pass` the moment an acceptable alternative fires. Empirically the judge's rationale reasoned purely from the request's wording, never cited `c_002`/`c_003`, and scored 3 regardless. The bullet is documentation, not a guard — worth either saying that in the body or dropping it, since it is billed on every run.

**The suite moved, and it is not you.** Against main's actual latest (`08-12_17-18-54`, not the 07-29 log this prunes) the run goes 16 pass / 5 partial / 0 fail → 16 / 3 / 2. Seven tests moved, including `ut_person_evidence_n7v` from straight 3s on all eight dimensions to Correctness 1 / Confidence calibration 1 / Score discipline 1. I chased that before attributing it: `flynn-record-matching/results/log_002.json` and `flynn-spouse-stub-marriage/results/log_002.json` changed on main on 2026-08-19, and `same-person-flynn-strong.json` on 08-14 — all after the 08-12 baseline. The inputs changed, not your diff.

Still worth someone's attention, and not yours to fix here: a fixture-hygiene change flipping a test from a perfect card to three 1s is a signal, and `n7v` is not in the `review_sample`, so no human read it.

**Discipline is clean.** `check_runlogs.py` exits 0 on all rules with no `eval-cosmetic-skip`; the 08-12 log genuinely lists this test file as `content-differs`, so the re-run was owed; candidates back to 5 with the 07-29 prune. Harness **2460 passed, 2 skipped**. All 12 checks green. I pushed the main merge (`6e0e0562`) since the branch was `BEHIND`.

**The annotation is sound** — I check this every time now, after a sibling PR shipped one whose comments cited evidence the run log contradicted. Yours is the opposite: all four quoted blocks are verbatim the judge's own rationales, all four `llm_score`s match `aggregated_dimensions`, and the "Junior:" reasoning matches the real response — the skill did flag James Brown's unreliability and the missing `a_013 → I2` link, kept `confident`, and ended by asking before writing.

One last note, deliberately not fixed because it would cost a paid run: `flynn-identity-geographic-thin/README.md` is rendered verbatim into the judge prompt, and it carries answer-key prose ("the correct behavior is to defer…") plus a `## Used by` list naming only `ut_conflict_resolution_007`. Harmless today since the judge does not gate this test, but that list is now stale and the prose reaches a second skill's judge. Worth folding into whatever next touches that scenario.

EdmondOware (approved):
Read both fixture files directly. mid-research-flynn holds exactly one conflict, c_001, resolved, fact type, birthplace, and zero speculative person_evidence links. flynn-identity-geographic-thin holds c_002 and c_003, both unresolved identity conflicts, and c_002's own description is literally which of the two Patrick Flynns recorded in 1850 Schuylkill County is the subject, the prompt's exact question, already sitting in the fixture. Plus pe_007 at confidence speculative tied to c_003.

Confirmed T-FEH's central finding myself by pulling the new run log directly. ut_person_evidence_017 shows num_turns 0, tool_calls empty, the only action is a builtin Skill dispatch straight to conflict-resolution. The router never opened research.json in this run. So this is correctly a fixture consistency fix rather than a behavior fix, the stop being a coin flip framing in the body is not what the record shows, and T-FEH is right to say so plainly.

Confirmed the run outcome myself too, 16 pass, 3 partial, 2 fail of 21, exact match. CI green.

No blocking defects found.

PR #1796, opened 2026-08-20
eval(research-exhaustiveness): deep dive — two dimensions were pre-decided, not dead (#1645)

### LINKED ISSUE #1645: deep dive: research-exhaustiveness
**Touches:** eval/tests/unit/research-exhaustiveness/, packages/engine/plugin/skills/research-exhaustiveness/SKILL.md

Audit `research-exhaustiveness` for defects nothing has caught. This starts from **no known

Did / Should / Gap.

## Start with this grep

```sh
grep -l -iE '"[^"]*\bscore [123]\b' eval/tests/unit/research-exhaustiveness/*.json
```

**This skill has 2 such file(s).**

without writing the finding.

## This skill's numbers

| | |
|---|---|
| Tests in the suite | 17 |
| Rubric+base dimensions that never discriminate | **1 of 6** |
| `judge_context` files naming a score branch | 2 |

1 of 6 dimensions never take more than one value, so they cannot report a defect in whatever they nominally grade.

## What to hand back

## Cost

Editing anything under `eval/tests/unit/research-exhaustiveness/` or the skill body flips this
skill's run-log snapshot inactive and buys one `make eval-skill SKILL=research-exhaustiveness`

Ikennaya1:
Deep dive on `research-exhaustiveness` per `docs/skill-deep-dive-guide.md`. Closes #1645.

Starts from no known failure. Three commits: the lane-2 fixes, the re-run they bought, and its annotation.

## What the numbers turned out to be

The issue says 1 of 6 dimensions never discriminates. Measured across all committed run logs, **`Stop criteria coverage` was a second one — and disguised.** Split by whether a run actually wrote a declaration, which is what `rubric.md` scoped it to:

| population | runs | scores |
|---|---|---|
| wrote a declaration | **17** | 3 × 17 — **every time** |
| wrote nothing | **35** | 3 × 31, 2 × 1, 1 × 2, null × 1 |

**Counting method**, since it was not stated and the first version of this table was wrong (thanks @EdmondOware). Across all **four** committed run logs, one row per test, keeping only rows where `Stop criteria coverage` is actually present in `aggregated_dimensions` — the 16 negative-test runs carry no rubric dimensions at all and are excluded. Bucket by whether the run's `tool_calls` wrote an `exhaustive_declaration`.

The earlier 13 / 38 was wrong twice over: it counted every test row (3 × 17 = 51), treating "dimension absent" as "score null" because a `.get()` returns `None` for both, and it predated this PR's own run log so the declaration count was one log short.

The conclusion is unchanged and slightly stronger: **31 of the 35** non-declaration runs scored 3 on a dimension the rubric said did not apply to them, and `ut_research_exhaustiveness_005` addressed **none** of the seven criteria and scored 3.

**The issue's grep under-counts.** Its pattern is `\bscore [123]\b` and finds 2 files. A wider one catching `scored as pass` / `scored as N/A` finds **4**. Two instances were invisible to the pattern the task hands you — worth widening in the guide so future deep dives don't inherit the blind spot.

**And that explains the dead dimension the issue did name.** `Tool Arguments` doesn't fail to discriminate because the skill always gets arguments right: **six of seventeen `judge_context` files tell the judge not to score it**, and 13 of 17 tests make zero calls. There is almost nothing left to grade.

## Findings, Did / Should / Gap

**F1 — `Stop criteria coverage` scored runs it did not apply to.** *Did:* 005 addressed 0 of 7 and scored 3; 017 addressed 1 and scored 1, both on runs that wrote nothing. *Should:* `rubric.md` scoped the dimension to declaration writes, and a rubric dimension may be `null` when the fixture never creates the situation it grades. *Gap:* lane 2 — three `judge_context` files handed over the verdict. **Fixed here.**

**F2 — a fourth file, invisible to the issue's grep.** *Did:* `refuse-while-in-progress.json` read *"should be scored as pass or N/A"*. *Should:* name the dimension, not the score. *Gap:* lane 2. **Fixed here**, plus the wider grep above for the guide.

**F3 — `already-declared-no-redeclare` required work already finished.** *Did:* its `judge_context` required "suggest the user proceed to proof-conclusion", on a scenario (`flynn-resolved`) where `ps_001` already sits at tier `proved`. The run correctly declined to say that and scored 3 anyway. *Should:* report the finished state. *Gap:* lane 2. **Fixed here.** The skill body has the same hole — its "Already declared" rule allows for no already-written conclusion — but the skill handles that case correctly unprompted, and the guide is explicit that adding prose for a rule the skill already obeys lengthens the prompt and changes nothing. Raised, not edited.

**F4 — 011's two 1s are judge jitter, not a hidden defect.** The harness advisory asks you to read before overriding, since 20 of 24 such cells were real. This is one of the other four: empty `text_response`, no files, no calls, `skills_invoked: ['proof-conclusion']`. The three sibling negatives are identical in shape and all scored 3.

**F5 — `recent-birth-sealed` cannot test what it exists to test.** *Did:* 017 has failed 3 of 4 runs. *Should:* it means to check that a sealed record does not block declaration. *Gap:* fixture. Three records are in play — the sealed 1936 birth certificate, the 2023 obituary (`pli_005`), and the 2023 death certificate, which appears nowhere. The skill correctly excludes the sealed record and then declines on the death record, which the body's decisive-record gate requires it to. So any skill applying the body correctly must decline, and is then marked as failing the sealed-record rule. **Not fixed — needs the death-record leg closed, which buys another run.**

**F6 — this PR's own 005 clause over-reached.** The rewritten clause scopes itself to `Stop criteria coverage`, but the judge generalised its reasoning and N/A'd `Declaration honesty` too — a dimension 005 should pass, since refusing under user pressure is exactly what it rewards. Corrected in the annotation; the clause still needs narrowing. **Not fixed — same reason.**

## What I softened

I went in expecting a dead dimension and found something narrower. 003 writes seven keys each tied to specific sources; 001 addresses seven in prose with the same grounding. Those 3s are earned, and 017 proves the dimension *can* score low. `Stop criteria coverage` is a **regression guard rather than a defect-finder** — a legitimate role, and now that non-applicable runs go `null` its numbers mean what they say.

I also got 017 wrong mid-branch and corrected it in `8d02900d`. I had written that the skill "ignores the sealed-record rule three runs out of four". It does not; it honours the rule explicitly and declines for a different, defensible reason. The defect is the fixture's.

## Validator requests

The measure of this task. **V1–V4 all pass today** — regression guards, not bug-catchers.

> **V1 — Rule:** a run writing `exhaustive_declaration` must carry all seven `stop_criteria` keys. **Where:** `tool_calls[].args.fields.exhaustive_declaration.stop_criteria`. **Not judgment:** fixed key set, named in `rubric.md`. **Violation:** none; 17/17 declaration runs comply.

> **V2 — Rule:** a run writing `declared: false` must not set `status` — it stays `in_progress`. **Where:** the same call's `fields`. **Not judgment:** the body states it twice as an absolute. **Violation:** none; 004 omits `status` in every log.

> **V3 — Rule:** this skill may write only `exhaustive_declaration` and `status`, on exactly one `q_`, and must not touch `tree.gedcomx.json`. **Where:** `tool_calls[].args` plus `file_changes`. **Not judgment:** the body's "Re-invocation behavior" says "Nothing else". **Violation:** none — which is why it is worth pinning before someone widens the skill.

> **V4 — Rule:** if the starting state already has `declared: true`, the run must write no declaration. **Where:** starting `research.json` vs the run's `tool_calls`. **Not judgment:** "do not re-declare" is absolute. **Violation:** none; 006 writes nothing.

**One that is not a validator.** F1's fix wants a *runlog lint*, since it checks a score against a run rather than the skill's behaviour: *a rubric dimension scoped to a condition the run did not meet must be `null`, not scored.* That is `check_runlogs`-shaped and worth raising separately.

## Verification

- Harness suite **2438 passed / 3 skipped** on the edited corpus
- Re-run: **16 pass, 1 fail**, $3.01, 644s, no transient retries
- Both the narrow and widened greps come back clean
- Annotation: 27 corrections over the five sampled tests, every disagreement carrying a comment
- All four JSON files parse; no mojibake

## Prohibition list — `research-exhaustiveness`

Step 1 of `docs/skill-deep-dive-guide.md`, built from the 268-line skill body.
Every line is checkable against a transcript or a `file_changes` block; wording
is quoted from the body. **Saved here so the next auditor starts from it rather
than rebuilding it.**

**Routing / scope**

1. Must **decline and route** — not run the evaluation — when the real request is the next question (→ `question-selection`), more searches (→ `research-plan`), or the conclusion (→ `proof-conclusion`). *"decline and route there — do not run the evaluation below"*
2. Evaluates **exactly one** question per invocation. *"One declaration at a time"*
3. Never writes a **second** declaration for the same question.

**Preconditions (all hard blocks)**

4. Only evaluates a question whose plan items are all `completed` or `skipped`; if any is `in_progress`, **refuses to declare** and recommends finishing first.
5. If any linked assertion lacks a real `information_quality` / `evidence_type`, must **stop, name the specific assertion IDs**, and recommend `assertion-classification`. (The ID-naming is part of the rule.)
6. If a person the judgment depends on has no linked identity assertion, must stop and recommend `person-evidence`.
7. Unlinked **fact** and **negative** assertions about an already-identified person are *advisory* — must note and continue, **not** block on them.
8. For every tentative-value assertion, must explicitly ask (a) genuine inaccessibility vs (b) one source's data-quality issue; if (b), must not declare, and must route to `research-plan` **with a specific new plan item naming the alternative record type**.
9. *"Do not declare exhaustive while a blocking check fails."*

**Evaluation**

10. If any of the five threshold questions answers "no" — identify what is missing and **stop there**.
11. Writes a 1–2 sentence assessment for **all seven** stop criteria — none omitted.
12. A **privacy-sealed** record must not be counted as an outstanding gap, nor recommended as a next step to obtain.
13. The inaccessibility exception applies **only** when that source is the *only known avenue*. If another record type could independently resolve the same uncertainty, it must be planned, not excused.
14. *"When in doubt, a gap is unsearched, not unobtainable — default to `research-plan`."*
15. **Named decisive records gate the declaration** — for parentage, the subject's death record, baptism, or a parent's probate; for marriage, the marriage record. No declaration until searched, or the declaration explicitly justifies inaccessibility.

**Writes**

16. Declaring exhaustive sets `status: "exhaustive_declared"` **and** the declaration in **one** `research_append` call.
17. Early termination writes `declared: false` and **must not change `status`** — it stays `"in_progress"`.
18. Early termination passes **only** `exhaustive_declaration`, **not** `status`.
19. If `research_append` returns `{ ok: false, errors }` — surface and fix the field; **must not blindly retry the same payload**.
20. Writes **nothing else**: only `exhaustive_declaration` and `status`, on one `q_`. No new questions, no `tree.gedcomx.json` changes.
21. If `declared` is already `true`, **must not re-declare** — report the existing declaration and point to `proof-conclusion`. *(See the caveat below: the body has no allowance for an already-written conclusion.)*

**Deliberately excluded as judgement, not checkable:** "exhaustive does not mean
exhausting", "historical context matters", whether the overturn-risk reasoning
was *sound*, whether breadth was *adequate*. The guide warns that mechanising
those is how a rubric ends up with a dimension that scores 3 forever — and this
skill already had two.

**Not yet reviewed by a genealogist:** items **7**, **12** and **15**. 7 and 12
are prohibitions on *over*-blocking, which are the easiest to read backwards; 15's
decisive-record set for parentage is quoted from the body and may be
under-specified. Treat those three as drafted, not confirmed.


Paaboat (approved):
Reviewed all changes. The rubric fix correctly distinguishes three run types: declare (structured object), decline-after-evaluating (prose), and refuse-before-evaluating (N/A). Test judge contexts updated consistently. Annotation corrections for tests 017 and 005 are well-reasoned, and the confirmed fail on test_017 Stop criteria coverage is correctly left as-is.

EdmondOware:
Reviewed this in depth and the substance is solid, the rubric wording and the four judge_context fixes are correct and well verified.

One thing to fix before I approve. I independently scanned all four committed run logs for the Stop criteria coverage dimension, split by whether the run actually wrote exhaustive_declaration. My count comes out to 17 declaration writing runs, all scoring 3, and 35 non declaration runs, mostly 3s with two 1s, one 2, and one null. The PR body states 13 and 38. The conclusion holds either way, my numbers independently confirm the dimension was scoring 3 on almost every non declaration run regardless of substance, but the exact split does not match what I get and I could not identify the counting method from the body. Please recheck that table against your script and correct it. This is prose only, no snapshotted file changes, so no re run needed.

Separately, not a blocker on this PR since it does not touch SKILL.md. Item 15 of the prohibition list, the decisive record list for parentage, the subject's death record, baptism, or a parent's probate, is missing the subject's own birth certificate or civil registration record where one exists. That is often the single most direct parentage source in eras with civil registration, and more reliable than a death record's secondhand informant. Worth its own issue against SKILL.md.

Will approve once the count is corrected.

EdmondOware (changes requested):
One thing to fix before I approve. I independently scanned all four committed run logs for the Stop criteria coverage dimension, split by whether the run actually wrote exhaustive_declaration. My count comes out to 17 declaration writing runs, all scoring 3, and 35 non declaration runs, mostly 3s with two 1s, one 2, and one null. The PR body states 13 and 38. The conclusion holds either way, my numbers independently confirm the dimension was scoring 3 on almost every non declaration run regardless of substance, but the exact split does not match what I get and I could not identify the counting method from the body. Please recheck that table against your script and correct it. This is prose only, no snapshotted file changes, so no re run needed.

Separately, not a blocker on this PR since it does not touch SKILL.md. Item 15 of the prohibition list, the decisive record list for parentage, the subject's death record, baptism, or a parent's probate, is missing the subject's own birth certificate or civil registration record where one exists. That is often the single most direct parentage source in eras with civil registration, and more reliable than a death record's secondhand informant. Worth its own issue against SKILL.md.

Ikennaya1:
You're right on both counts, and thanks for re-deriving it rather than taking the table on trust. Corrected in the body — your 17 / 35 is exactly what I now get, distribution included (31 threes, one 2, two 1s, one null).

Two errors compounded to produce 13 / 38:

1. It counted every test row (3 × 17 = 51), treating "dimension absent" as "score null" — a `.get('Stop criteria coverage')` returns `None` for both, and the 16 negative-test runs carry no rubric dimensions at all, being graded base-only.
2. It predated this PR's own run log, so the declaration count was one log short.

I've added the counting method to the body so it is reproducible. That omission is what let the error survive — you couldn't check it, so nobody could.

The conclusion is unchanged and slightly stronger: **31 of the 35** non-declaration runs scored 3 on a dimension the rubric said did not apply to them.

**On item 15 — you found a real gap, and splitting it out is the right call.** Filed as #1801, quoting your reasoning. The subject's own birth or civil-registration record is missing from the decisive-record list, and your argument is better than the body's: in a civil-registration era it is the most direct parentage source there is, and more reliable than a death record's secondhand informant. That list is quoted verbatim from `SKILL.md`, so the gap is the body's rather than the prohibition list's.

Not folding it in here, deliberately. It is a lane-4 doctrine edit to `SKILL.md`, which would invalidate the run log this PR just committed and annotated and buy a second paid `research-exhaustiveness` run — the "different paid eval slot" exemption, with the lane difference on top. Both named in #1801.

One process note worth recording: item 15 was one of three I marked "drafted, not confirmed by a genealogist", precisely because I could not vouch for that record list. You went straight to it. The flag earned its keep, and #1801 says so.

EdmondOware (approved):

Issue #1776, opened 2026-08-20
Path-gated workflows are dark for repo-root and scripts/ changes — windows-wrappers.test.ts's Makefile guard is unreachable in CI


Gennecis:
> **Reviewed 2026-08-20 before junior handoff.** Decision settled — ungate the
> engine suite, prefix the other three. The original body follows under
> "Original issue"; two of its claims are corrected below.

## Ruling 2026-08-20 — ungate `engine-tests.yml`, prefix the other three

`.github/workflows/engine-tests.yml` **stops being path-gated**: delete its scope step
so `vitest` runs on every PR. Measured 47–49s, in parallel with everything else.

Chosen over adding repo-root prefixes to it, because **a prefix list cannot be
completed for that job.** `doc-links.test.ts` and `adr-links.test.ts` resolve every
backticked path citation through `repo-paths.ts`'s `REPO_ROOTS` — `docs/ packages/
apps/ eval/ scripts/ .github/ .claude/` — so deleting or renaming a file under any of
the seven can redden the suite. A complete prefix list there is "every top-level
directory", which is the same thing as not gating it. Live example: a skill file under
`.claude/` cites `apps/electron/docs/feedback-json-spec.md`.

The other three keep the gate and get one exact prefix each:

- `.github/workflows/eval-harness-tests.yml` — add `|^scripts/`
- `.github/workflows/js-tests.yml`, `eval-app` job — add `|^docs/specs/schemas/`
- `.github/workflows/server-tests.yml` — add `|^docs/specs/schemas/`

**Part of this task:** write the reasoning into the comment block at the top of each of
the four workflow files — for the engine one, why it is ungated and what the ~50s buys;
for the other three, which test reads the newly-matched path. Those comment blocks are
where the next person will be standing.

**Do not reintroduce a classic `paths:` filter** on any of the seven. Each carries a
comment explaining the move: a workflow skipped by `paths:` reports no conclusion at
all, leaving a required check stuck on "Expected — waiting for status to be reported"
with no exit but a ruleset bypass.

## The population is 8 rows, not the class in the title

Three of the eight are neither a repo-root file nor `scripts/`, so a fix scoped to this
issue's title passes its own acceptance check and leaves the class open. Measured
2026-08-20; re-derive any row with `grep -n "PATTERNS:" .github/workflows/*.yml` then
`printf '%s\n' <path> | grep -qE '<that block's PATTERNS>' && echo RUNS || echo dark`.

[...]

PR #1797, opened 2026-08-20
ci: run check-runlogs post-gate lints even when the gate fails (#1378)

### LINKED ISSUE #1378: check-runlogs.yml: the four warn-only lint steps are skipped whenever a blocking step fails

clack391:
## Summary

- The four diagnostic steps after the run-log gate in `check-runlogs.yml` carried a bare `if: steps.scope.outputs.relevant == 'true'`, which GitHub reads as an implicit `success() && ...`, so a failing gate (`Run runlog discipline checks`) skipped all four. Those lints are the ones most worth reading on exactly the PRs that trip the gate. This adds `!cancelled()` to their `if:`.
- Adds an offline guard (`.github/tests/check-runlogs-post-gate-lints.test.mjs` plus a `workflow-logic-tests.yml` job and paths entry) asserting every post-gate step carries `!cancelled()`, so the bug cannot recur silently. It recurred once already, when a new step inherited the buggy pattern.
- Frequency: about 1 in 6 runs suppress all four lints (159 of 1016 in the `created > 2026-08-06` window, reproduced 2026-08-20 with the issue's own commands). Tier: Normal.

## Review guidance

**Start here:** the four `if:` lines in `check-runlogs.yml`, kept inside `${{ ... }}` on purpose (a bare leading `!` is a YAML tag indicator that fails the whole-workflow parse), and the guard test's invariant, which asserts EVERY post-gate step has `!cancelled()`, not only today's four, so a future step added without it is caught.

**Unsure about:** nothing blocking. The scope expansion (the offline guard) is deliberate and flagged under Deviated.

## Plan

**Didn't change:** the gate step and the four pre-gate steps keep their bare `if:` (they must not run once an earlier one fails). No `continue-on-error` anywhere: it would convert the blocking run-log gate, or the frontmatter lint, into a warning.

**Acceptance check:** `node .github/tests/check-runlogs-post-gate-lints.test.mjs` passes on this branch and fails with a step-named error when `!cancelled()` is removed from any post-gate step (break-tested both directions). Nothing offline can prove the steps actually RUN rather than skip, so the definitive runtime proof is the Actions step-conclusions (below).

**Deviated from the plan:** expanded scope beyond the issue's "four `if:` lines, one file." The issue's premise that "nothing offline reads `.github/workflows/`" is false: `.github/tests/` is an offline node harness built for exactly this class of invariant. Since the bug already recurred once, a hard guard beats the comment alone. Flagged because it crosses the reviewed issue's stated scope.

## Test plan

- [x] `make test-all` ran and passed. Note: it does not actually *cover* this change (this is `.github/workflows` plus a node test run by Actions, outside test-all's suites); run to confirm it is unaffected. The real local check is the guard test above.
- No skill / agent / rubric / unit-test snapshot changed, so no `make eval-skill` run and no annotation.

**Runtime proof (the issue's mandated instrument), done and posted in a comment below:** a scratch commit forced the gate step's `run:` to `exit 1`; the `runlogs` job's step conclusions showed the gate `failure` and all four post-gate lints `success`, never `skipped`. The scratch commit was reverted and the job is green. Full step list is in the comment.

## Follow-on issues

**Folded in / filed instead:** none. #1776 edits other workflows' scope steps with no file in common, so either order works. No open PR touches these files.

Closes #1378.


clack391:
## Runtime proof (the issue's mandated instrument)

Nothing offline can prove the four post-gate steps actually RUN rather than skip when the gate fails, so I forced it on a scratch commit (gate step `run: exit 1`, since reverted) and read the `runlogs` job's step conclusions:

```
failure  Run runlog discipline checks            <- gate, forced exit 1
success  Check skill tool-coverage drift (warn-only)
success  Check rubric/judge_context/agent tool-mention drift (warn-only)
success  Lint skill + agent frontmatter
success  Check negative routing reciprocity (warn-only)
```

All four post-gate lints reported `success`, never `skipped`, with the gate failing. Before this change they were all `skipped`.

The scratch commit (6a5dc307) has been reverted (1445cbe5): the gate's real `run:` is restored and the job is green again. The new offline guard job `check-runlogs post-gate lints carry !cancelled()` is green on this PR and fails if any post-gate step drops `!cancelled()` (break-tested locally).

jud-sdev (approved):
I read every step line by line and verified the fix by running, not just reading.

- The four post-gate steps now carry !cancelled() and nothing else was missed: the gate is Run runlog discipline checks, the setup steps and the gate before it correctly keep their bare if:, and all four steps after it are the ones changed. The condition still ANDs the scope gate and stays inside ${{ }} so the leading ! does not break the YAML parse.
- The guard test is genuinely non-vacuous. I ran it under node: stripping !cancelled() fails with a named error per step, and renaming the gate fires the non-vacuity anchors including "the invariant would be vacuous". Restore passes. It cannot pass while silently covering nothing.
- Wiring is right: check-runlogs.yml is in the trigger paths so a future edit re-runs the guard, and the new job runs the test. Green on this head.
- The runtime proof comment (scratch commit forcing the gate to exit 1, since reverted) shows the gate failing with all four lints reporting success rather than skipped, which is the behaviour offline cannot show.

Correct, minimal, and the recurrence guard actually bites. Nothing to change. This is one of the two sign-offs the .yml/.mjs code-owner rule needs, so it still wants a second.

francis-2008-happy (changes requested):
The `!cancelled()` change itself is right, and well evidenced — the runtime proof with the scratch commit is exactly the instrument this needed, and I confirmed `6a5dc307` and `1445cbe5` net to an empty diff. I also swept every other workflow: `check-runlogs.yml` is the only job with diagnostics after a blocking gate, every other bare `if:` is a setup chain where the implicit `success()` is what you want. So the class is closed, and the merge with main is clean.

The guard is where I'm asking for changes. It can pass green in the case it exists to catch.

It splits steps on `- name:` lines. A step written with no `name:` is never treated as a step at all — its lines get appended to the previous step's body, so its `if:` is never checked. I appended this to `check-runlogs.yml`:

```yaml
      - if: steps.scope.outputs.relevant == 'true'
        run: python eval/harness/scripts/check_future_thing.py
```

The guard printed `ok    all 4 post-gate steps carry !cancelled()` and exited 0. YAML sees 10 steps there; the guard sees 5. That step would be skipped by a failing gate — the exact bug — with the guard green.

That shape is not hypothetical. Every job in `workflow-logic-tests.yml`, including the one this PR adds, is written `- uses:` / `- run:` with no `name:`. And "someone adds a step later" is the case the guard is for.

Same root cause, second symptom: the non-vacuity anchor matches raw text including comments, so prose can stand in for the step. I deleted the negative-reciprocity step and left `# was: check_negative_reciprocity.py — retired` in its place. Guard green, three post-gate steps, no complaint.

One edit fixes both. I tested this version — baseline passes, both holes now fail, and all four of your break-tests still fail.

In `.github/tests/check-runlogs-post-gate-lints.test.mjs`, replace:

```js
// Split the workflow's steps into blocks by `- name:` lines (this file has a
// single job, `runlogs`; if a second job with steps is ever added this over-scans
// into it, which fails loud, never silent). String parse, no yaml dep, matching
// this harness's checkout+node, no-install contract.
const lines = fs.readFileSync(WORKFLOW, "utf8").split("\n");
const steps = [];
for (const line of lines) {
  const m = line.match(/^(\s+)- name: (.+?)\s*$/);
  if (m) steps.push({ name: m[2], body: [line] });
  else if (steps.length) steps[steps.length - 1].body.push(line);
}
if (steps.length === 0) {
  fail(`${WORKFLOW}: no "- name:" steps found — renamed or reformatted?`);
}
```

with:

```js
// Split the workflow's steps on YAML list-item boundaries, NOT on `- name:`: a
// step written as `- run:` or `- uses:` with no name is still a step and must
// satisfy the invariant too. Comment lines are dropped first so prose naming a
// script cannot stand in for the step itself. (This file has a single job,
// `runlogs`; if a second job with steps is ever added this over-scans into it,
// which fails loud, never silent.) String parse, no yaml dep, matching this
// harness's checkout+node, no-install contract.
const lines = fs.readFileSync(WORKFLOW, "utf8").split("\n").filter((l) => !/^\s*#/.test(l));
const stepsIdx = lines.findIndex((l) => /^\s*steps:\s*$/.test(l));
if (stepsIdx === -1) {
  fail(`${WORKFLOW}: no "steps:" key found — renamed or reformatted?`);
}
const steps = [];
let indent = null;
for (const line of lines.slice(stepsIdx + 1)) {
  const m = line.match(/^(\s+)-\s+(\S.*?)\s*$/);
  if (m && (indent === null || m[1].length === indent)) {
    indent = m[1].length;
    const named = m[2].match(/^name:\s*(.+?)\s*$/);
    steps.push({ name: named ? named[1] : `<unnamed step: ${m[2]}>`, body: [line] });
  } else if (steps.length) steps[steps.length - 1].body.push(line);
}
if (steps.length === 0) {
  fail(`${WORKFLOW}: no steps found — renamed or reformatted?`);
}
```

And one line, so the message is accurate when a step's `if:` sits on the list-item line itself:

```
You wrote:      const ifLine = s.body.find((l) => /^\s*if:/.test(l));
Change it to:   const ifLine = s.body.find((l) => /^\s*(-\s+)?if:/.test(l));
```

Then re-run your break-tests plus the two above. Both should fail; your original four should still fail; baseline should still pass.

Last thing: **Start here** says the guard "asserts EVERY post-gate step has `!cancelled()`". That becomes true once this lands, so either apply the edit or soften the line — you don't need both.

Two notes that aren't changes. `!cancelled()` also means the four lints now attempt to run if `Check out PR` or `Set up Python` fails, so one red step becomes five. The job already fails from the root cause and no lint result goes falsely green, so I'd leave it. And @jud-sdev is on `senior-developers`, so the CODEOWNERS senior requirement is already met — once you push, this needs one more approval and nothing else. Their read was thorough; the one thing to flag is that "it cannot pass while silently covering nothing" was reached with named-step break-tests only, which is the case the guard does see.

francis-2008-happy (approved):
Approving. The guard fix is applied verbatim in `a0343c56`, and I re-verified the whole matrix myself rather than taking the run report on trust:

- baseline passes
- unnamed post-gate step with a bare `if:` now fails, naming `<unnamed step: ...>`
- comment-substitution now fails on the non-vacuity anchor
- `!cancelled()` stripped from each of the four named steps fails 4/4, each naming its own step
- gate renamed fires all six anchors including "would be vacuous"

Both workflows parse, the other three `.github/tests` pass, and the two workflow files are byte-identical to the round I already reviewed — the only new code is the splitter.

I also probed the edit I proposed, for holes I might have introduced myself. An unnamed `- uses:` step with its `if:` on a later line is caught; an unnamed step with no `if:` at all is caught with the right remedy text; and a nested `- ` list inside a post-gate step's `with:` correctly does not split into a step, so the count stays 4 with no false positive.

One quirk worth knowing about, not worth changing: a block-scalar `if: >-` with `!cancelled()` on the continuation line will fail even though it is correct. It fails loudly and the message prints `if: >-`, so the cause is obvious, and nothing in the file is written that way. Safe direction.

Thanks for reproducing both holes on the pre-fix guard before touching anything — that is the part that makes the guard trustworthy rather than just green.

@jud-sdev heads up: your approval still reads `@36d08898`, so it predates the guard rewrite. It still counts (`dismiss_stale_reviews_on_push` is false here), and it already covers the senior CODEOWNERS requirement, so nothing is blocked. But `a0343c56` changed the one thing your review called out as settled — that the guard "cannot pass while silently covering nothing" — so you may want a look at the new splitter.

PR #1799, opened 2026-08-20
search-external-sites: don't mark a plan item done for handing over a link (#1226)

### LINKED ISSUE #1226: search-external-sites marks a plan item done the moment it hands over a link
## What happens

Maria is looking for her ancestor Patrick Flynn in Pennsylvania. Her research plan has a line on it:

> **Plan item pli_001** — look for Patrick Flynn in the 1880 Schuylkill County census.

Claude searches FamilySearch. Nothing. So it hands the job to `search-external-sites`, which covers Ancestry, MyHeritage and the rest. That skill can't search those sites itself — they sit behind a paid login — so it does the next best thing. It builds Maria a ready-made link with the name, year and place already filled in, and says *"click this, and if you find the record, save a copy and send it back to me."*

Then it ticks **pli_001 as done**.

Maria hasn't clicked anything yet.

Two weeks later she comes back and asks whether the search for Patrick is finished. `research-exhaustiveness` reads her plan, sees every line ticked, and tells her yes. Nobody ever opened that Ancestry link. Nothing was found, and nothing says so.

There is no error and no crash. Just a quiet wrong answer about whether the research is done — which is the one thing a research log exists to get right.

## Where it comes from

`packages/engine/plugin/skills/search-external-sites/SKILL.md`, Step 7:

> *"Once the search is logged, found or not, set the plan item to `completed`"*

That line entered the file on 2026-06-13, in a commit that was cutting the skill from 523 lines to 367. The commit message explains four other changes and never mentions plan status. The next commit, on 2026-06-30, only reworded it during another round of cutting. There is no spec for this skill. So going by the history, nobody chose this — it just got carried along.

## Four reasons this looks like a mistake, not a choice

**The right word already exists, and this skill never uses it.** A plan item can be `planned`, `in_progress`, `completed` or `skipped`. `in_progress` means exactly "started, not finished." Searching this skill for `in_progress` returns nothing.

**There is already a guard that would catch it.** `research-exhaustiveness` refuses to tell the user a question is finished while any plan item is still `in_progress`. Writing `completed` is what steps around that guard.

**The skill's own log entry says the opposite of the tick.** In the same step it writes a log row recording the site, the exact link it built, and `capture_received: false` — meaning nothing has come back yet. So the file says "waiting on the researcher" and "done" at the same time.

**A sister skill was asked the same question and answered the other way, in writing.** `search-records/SKILL.md:424`:

> *"**Do not** set status to `completed` from this skill — that is set by record-extraction once assertions have been created."*

Same situation, opposite rule, and that one explains its reasoning. This one doesn't.

## But there is an argument for leaving it alone

Worth stating fairly, because it might be the right answer.

An outside site is not ours. The skill cannot see whether Maria clicks the link, so it may have been written to mean *"my part is done"* rather than *"the record was found."* Under that reading Step 7 is correct, and the mistake is somewhere else: `research-exhaustiveness` should not treat a ticked plan item as proof that a search actually happened.

Either way, one of the two is wrong. **Which one is the decision that has to be made before anyone changes any code.**

## Nothing catches this today

| could catch it | does it? |
|---|---|
| `ut_search_records_018` | No — it swaps in a fake stand-in for the skill, so Step 7 never runs |
| `ut_search_records_026` | No — it did catch it, on 2 runs out of 3, and the check was switched off because 026 is testing something else |
| `search-external-sites`' own validators | No — none of them look at plan-item status |
| its rubric | No — "plan item" appears there only as a place to get search parameters from |

## The receipt nobody reads

`capture_received: false` is the field that says a capture is still owed. This skill writes it. The research.json schema requires it. And searching the whole plugin for `capture_received` returns **zero** other matches — nothing ever reads it back.

The receipt is filed correctly and nobody opens the filing cabinet.

That is also why Maria gets no reminder even when a plan item *is* correctly left open. The system can tell her "one item is still in progress." It cannot tell her "you have an unopened Ancestry link for Patrick Flynn," because nothing looks at the field that would say so.

## What to do

Three things, in order. Only the first blocks the others.

**1. Decide who is allowed to tick the box.** Right now four skills each decide for themselves, and they disagree:

| skill | when it ticks the box | is that right? |
|---|---|---|
| `search-records` | never — it waits for record-extraction to save the evidence first | strictest, and it explains why |
| `search-full-text` | straight away | fine — the search really did run |
| `search-images` | straight away | fine — same |
| `search-external-sites` | straight away | **wrong** — the search hasn't happened yet |

I checked the middle two rather than assuming: neither mentions a capture, an upload or a paywall, so their work genuinely finishes during the session. `search-external-sites` is the only one ticking off work a human still has to do.

**2. Change Step 7 to write `in_progress` instead of `completed`** — for the case where it handed over a link and nothing has come back. The other two endings of that step (the site had nothing, or the site couldn't be reached) may still be `completed`; that falls out of the decision in step 1.

**3. Make something read `capture_received`,** so an open plan item turns into a real reminder. `project-status` is the natural place: *"2 links are waiting on you — Ancestry (Patrick Flynn, 1880 Schuylkill), MyHeritage (…)."* This can be split into its own ticket if step 2 lands first.

## What it touches

- `packages/engine/plugin/skills/search-external-sites/SKILL.md` — the Step 7 wording. Editing a SKILL.md invalidates that skill's saved test results, so its test suite has to be re-run and re-graded.
- `eval/tests/unit/search-records/live-callee-external-sites-escalation.json` — once Step 7 is fixed, the `DO NOT GRADE PLAN-ITEM STATUS` note should be deleted so the check comes back on. That invalidates the search-records results too, so it is a second suite run.
- `search-external-sites`' own test suite gains nothing from this unless someone adds a check — right now nothing there looks at plan-item status at all.
- `research-exhaustiveness` — untouched if we go with step 2, but it is the other half of the decision in step 1.

## How to prove it is fixed

Add a check to `search-external-sites`: after it hands over a link, the plan item named in that log entry must **not** be `completed`. That check fails today and passes after the fix. Running `make test` proves nothing here — the suite is green right now with the bug in it.

---

Found while building `ut_search_records_026` (the PR for #1012), the first test that lets `search-external-sites` really run instead of swapping in a stand-in. Test 018 never showed this, because a stand-in hands back less than the real thing — and the caller only goes wrong when the real thing succeeds.

promise-emmanuel:
Closes #1226.

## What was going wrong

`search-external-sites` handles paid sites like Ancestry and MyHeritage. It can't log in to them, so it does the next best thing: it builds a ready-made search link, hands it to the researcher, and asks them to click it and send back a PDF.

Then it marked the plan item **done** — before anyone had clicked anything.

Weeks later the researcher asks "is the search for Patrick finished?" `research-exhaustiveness` reads the plan, sees every line ticked, and says yes. No error, no crash. Just a wrong answer about whether the research is complete, which is the one thing a research log exists to get right.

The instruction behind it, at `SKILL.md:440`:

> *"Once the search is logged, found or not, set the plan item to `completed`"*

`grep -c in_progress` on that file returned **0**. The word describing the skill's actual state was never used. It's 6 now.

## How it works now

Step 7 has an endings table instead of one unconditional `completed`:

| What happened | Status |
|---|---|
| Link handed over, no capture back yet | `in_progress` |
| Capture came back and was triaged | `completed` |
| Capture arrived unusable (login page, cut-off PDF) | `in_progress` |
| Researcher *says* "nothing there", sends no capture | `in_progress` |
| Site unreachable **and they ask to skip it** | `skipped` |
| Site unreachable, they haven't decided | `in_progress` |
| `--autonomous` (no human can ever capture) | `skipped` |

A captured empty results page counts as `completed` — a confirmed nil is a real finding. A *reported* empty result doesn't, because nothing can be checked later. Outside `--autonomous`, `skipped` requires the researcher to ask for it; the skill never infers it from an access failure.

**Three other places in the same file carried the old rule** and now agree with the table:

- the capture-problems row "User can't access the site", which said to log an error and move on
- the autonomous step, which said "mark the plan item terminal" without naming a status
- the offer list, which proposed *"All planned searches are complete — evaluate whether research is exhaustive?"* on any finish. `research-exhaustiveness` refuses while any item is `in_progress`, so that offer led somewhere that would decline it. It now fires only when every item is terminal, and otherwise names what's outstanding and offers the plan's `fallback_for` item, or re-planning if there isn't one.

## How often this actually happened: 2 in 5

Worth knowing before reviewing, because it changes what to expect. `ut_search_records_026` is the only existing test that reaches this state. Across its five committed run logs:

| run log | on this branch? | final `pli_001` |
|---|---|---|
| v1_2026-08-21_00-13-34 | no — `main`, unfixed | in_progress |
| v1_2026-08-21_18-09-48 | no — `main`, unfixed | in_progress |
| v1_2026-08-21_19-08-50 | no — `main`, unfixed | **completed** |
| v1_2026-08-20_21-55-13 | yes, fixed | in_progress |
| v1_2026-08-24_09-59-31 | yes, fixed | in_progress |

Earlier revisions of this table cited two run logs retention has since pruned
(newest-5 per skill). The shapes are preserved as inline fixtures in
`test_capture_pending_validator.py`; every run named above currently exists.

**Unfixed, the model wrote `in_progress` on its own in 2 of 3 surviving runs.** So there is no clean fails-today / passes-after here. This PR corrects an instruction that was wrong on its own terms — telling the model to mark a search `completed` when no search happened — and does **not** claim to remove the variance. If variance is the real target, that's separate work.

## Proving the guard actually works

Since the eval suite can't demonstrate a 2-in-5 bug deterministically, the new validator is mutation-tested directly instead. Run against the five committed run logs, it fires on both `completed` runs and stays quiet on all three `in_progress` ones. Those are reduced to inline fixtures in `test_capture_pending_validator.py` — **9 tests, no skips**, also covering: capture arrived, a later entry carrying the capture, pre-existing fixture state the run never touched, and the exempt tags.

One of those tests originally "passed" by being skipped. That's precisely the vacuous pass this guard exists to catch, so it now asserts the skip explicitly.

## Where the check lives, and why

The assertion is in `validators_lib.py` with a two-line wrapper in **both** `test_search_external_sites.py` and `test_search_records.py`:

- `validator_runner` loads only `test_universal.py` plus `test_<skill>.py`, so a single skill file would never run on `ut_search_records_026` — the one existing test that reaches this state.
- `unit-test-spec.md` §8 assigns skill structural rules to `test_<skill>.py`, so putting it in `test_universal.py` would drift from the spec.

Four details it gets right, each one a bug in an earlier draft:

- keys on **`tool == "external_site"`** — `log_entry` has no `type` property, and a filter on `type` matches nothing and passes vacuously
- takes the **latest entry per plan item** — step 4's in-flight entry is never revised, so a stale `capture_received: false` must not fail a run whose capture did arrive
- considers **only items this run changed** — `mid-research-flynn` already holds completed external-site items, and those are fixture state
- forbids **`skipped` as well as `completed`**, or the new skip path reintroduces the bug

## Test and rubric changes

- **New test** `plan-item-stays-open-pending-capture.json` on the `ma-state-census-external` scenario, naming `pli_001` explicitly. Not optional when written: Step 7's status write fired in **0 of 14** ext-sites tests across every run log predating this change, so without a test naming an item the validator skips — and a skipped validator records `passed: true`. That figure is now out of date in a way worth stating: post-change `_fhk` also writes a plan-item status (its scenario has an open `pli_001`), so the guard has **two** live assertion sites. The endings table did not only change *which* status is written — it made the skill write one where it previously wrote none.
- **`ut_search_records_026`**: both plan-item-status waivers removed — the `DO NOT GRADE PLAN-ITEM STATUS` bullet *and* the cross-reference clause in the URL-provenance bullet, which would otherwise leave a live instruction plus a dangling pointer. Replaced with the positive expectation. Its scope note is updated too.
- **`nil-result-logging.json`**: its scenario has no Newspapers.com plan item, so `plan_item_id` is null and no status can be written. It now asserts that restraint instead.
- **Rubric**: status grading is folded into the existing **Log entry** dimension, whose subject was already "did the skill truthfully record what happened". A sixth dimension is not allowed — `unit-test-spec.md` §7 caps a rubric at five, and adding one makes the rubric malformed and the whole suite unrunnable.

`make test` proves nothing here — it's `test-js` + `server-test` and reaches neither the harness nor the engine. The harness suite is green: **2469 passed, 2 skipped**.

## One correction to the issue

#1226 says the `completed` rule *"entered the file on 2026-06-13, in a commit that was cutting the skill from 523 to 367 lines… nobody chose this."*

```
git log -S'plan item to `completed`' -- '*search-external-sites/SKILL.md' | tail -1
→ 7046b735  2026-05-07  Add 20 GPS genealogy research skills, specs, and update READMEs
```

It came from the founding 20-skills commit. `050068d0` only reworded it. So "nobody chose this" doesn't hold — the rule was deliberate, and the case for changing it rests on the behaviour, not on it having been an accident.

## Deliberately not in this PR

- **A `research-plan` mode that can add one fallback item to a running plan.** Dallan asked for the handoff in Slack; it's mechanically feasible with no tool or schema change (`plan_items` is a first-class `research_append` section, and the one-active-plan invariant only guards `plans` appends). Its own issue — note that `research-plan:115-118` defaults an ambiguous prompt to *review* mode when items are unfinished, and review can't create items, so the mode needs care.
- **`fallback_for` is set on a minority of plan items** — 288 of 1374 occurrences in `research-plan`'s own run logs (~21%). Which is why the handoff has to be able to *create* one rather than point at an existing one. Its own issue.
- **Nothing reads `captureReceived`.** No reader anywhere in `packages/engine/plugin/`, so an outstanding capture is never surfaced back across sessions. `project-status` is the natural home. Issue #1226's step 3.
- **Broken links the skill builds itself.** No handling anywhere for a URL it generated wrongly. The status rule covers it (still owed → `in_progress`); what the skill should *do* is unspecified, and any retry limit would be invented rather than measured.
- **Four of the seven endings are unreachable in the current corpus** — no ext-sites test supplies a capture, since all 14 inputs are `user_message` + `scenario` only. Stated so the table isn't mistaken for tested behaviour.

## Review round: what changed after the first approval

- **`e9fce0a0` (@florencemashipei)** — the guard could fail the one ending the endings table exists to permit. It keyed "capture arrived" on a later entry carrying the same `plan_item_id`, but step 6 never tells the skill to carry `planItemId` onto the capture entry and all 14 committed entries have it null. Arrivals are now also matched on `(site, url_generated)`, and the tag exemption is keyed on `terminal-status-expected` rather than `autonomous` — a *mode* a future search-records test would have inherited without asking.
- **`81a28e80` (@johnmarkpeterbrown items 1, 2, 3, 5)** — keying `latest` on every tool let any later entry switch the guard off, including a nil search and an `external_links_search` (which only discovers links); and `captured` was unscoped, so an arrival naming a different item laundered this one. Patch applied as written, plus four tests — 17 total, no skips.
- **`_013` was making the skill invent a plan item.** `mid-research-flynn` has no open external-site item, so told to mark something terminal the skill created `pli_007` inside the FamilySearch probate plan and marked it `skipped`, which `ownership.json` forbids. Its judge_context now says no status may be written there. Verified in the new run: `sections_modified: ['log']`, no status written, no `pli_007`.
- **Item 3's suggested repoint was itself stale** — `v1_2026-08-14_03-36-42` had been pruned by the time I applied it. `docs/skill-deep-dive-guide.md` now cites `v1_2026-08-21_00-13-34`, verified to hold the quoted turn verbatim.
- Also repaired: two search-records run logs committed with 2229 git conflict markers each, from resolving a run-log rename/rename collision. Both restored from their correct sides; a repo-wide scan finds no others.

## Eval state

Both run logs are current and active — `check_runlogs.py` reports no rule-2 violation.

| suite | run log | outcome |
|---|---|---|
| search-external-sites | v1_2026-08-24_09-59-31 | 13 pass, 1 partial, 1 fail |
| search-records | v1_2026-08-24_09-59-31 | 25 pass, 3 fail |

`$12.10` for the pair. The non-passes, none of which touch this change:

- `_012` (ext-sites) — negative routing test; the skill activates on a request belonging to `record-extraction`. Failed identically on 2026-08-10, before this branch existed; its history is pass/fail/pass/fail/fail/fail.
- `_002` (ext-sites) — partial on URL parameter mapping, which the genealogist independently confirmed as a genuine 2 in the previous annotation. Unrelated to plan-item status.
- `_013`, `_014`, `_027` (search-records) — all three fail the same #1786 validator, `test_pre1880_census_structure_marked_inferred`. No judge dimension scored 1 in any of them, and `_001`, `_015` and `_017` failed it last run and pass now, so it is flaky rather than consistently red.

For the record: the superseded ext-sites `v1_2026-08-20_21-55-13` was committed without an annotation, and committing it evicted the annotated `v1_2026-07-27_22-27-37` pair under newest-5 retention.

## Still outstanding

@johnmarkpeterbrown's item 4 — extra annotations on `ut_search_external_sites_pic` and `ut_search_records_026`. Neither appears in either run's machine-picked `review_sample`, so the Log-entry status clause this PR adds has been graded by the judge and nobody else. No paid re-run needed; extra corrections beyond the sample are allowed and go in through the CRUD UI.


florencemashipei (approved):
Approving. I verified the claims rather than taking them on trust, and pushed one commit to the branch (`e9fce0a0`) — details at the end.

## What I checked

- **Harness suite** green: `2461 passed, 9 skipped` before my commit, `2465` after. The one failure in my worktree is `test_mock_mcp.py::test_successful_call_carries_no_is_error`, an unbuilt engine locally, not this branch.
- **`check_runlogs.py`**: "All runlog rules satisfied". Rule 1 doesn't fire on the two added ext-sites candidates because it counts only *released* `v{N}.json`; rules 2/3 resolve to `v1_2026-08-20_22-45-06`, which is annotated.
- **Schema conformance**: `tool == "external_site"`, `external_site.capture_received` and `plans[].items[].status` all match `research.schema.json`, and `plan_item_status` is exactly the four values the endings table uses. `test` really is the inner `test` block in `validator_runner`, so `test.get("tags")` is right.
- **The new test isn't vacuous** — the thing the PR body says it exists to prevent. In the committed run log `test_capture_pending_item_not_terminal` records `error: null` (ran, not skipped) on both `_pic` and `_fhk`, skips on the autonomous tag for `_013`, and skips with "changed no plan-item status" on `_008`. Across `search-records` it ran on 8 tests and skipped cleanly on the four #1786 failures.
- **No cross-test contradiction**: `ut_search_records_018` already told the judge pli_001 "must NOT be marked completed", so the corpus agrees with the new doctrine rather than fighting it. Removing the two waivers on 026 leaves no dangling pointer.

The `grep -c in_progress` observation is the right diagnosis: the instruction told the model to record a search that had not happened, and the fix is to say what actually happened. Correcting a wrong instruction is worth doing on its own terms, and I appreciate the PR body saying plainly that it does not claim to remove the 2-in-5 variance.

## Pushed to the branch (e9fce0a0)

The guard could fail the one ending it exists to permit. It keyed the pending state on the latest *external-site* entry carrying the item's `plan_item_id` — but a capture that arrives is a **new** entry, and step 6 names the fields to carry onto it without naming `planItemId`. All 14 committed ext-sites entries have `plan_item_id: null`. So a triaged capture can leave no per-item trace, the stale step-4 handoff stays the latest keyed entry, and `completed` fails. Unreachable in today's corpus only because no test supplies a capture — exactly the blind spot the PR body flags in "four of the seven endings are unreachable".

Also fixed: an item closed by a later `record_search` (the record turns up on FamilySearch after the handoff was logged) no longer fails on the older in-flight entry. Only the item's latest entry across *every* tool decides, and arrivals are additionally matched on `(site, url_generated)`, which the schema requires on every external-site entry.

The exemption is now keyed on `terminal-status-expected` — the reason it exists — with `autonomous`/`user-requested-skip` still honoured. `autonomous` is a mode, so a future autonomous test in `search-records` would have inherited an exemption it never asked for and silently disarmed the guard.

Four new tests (13 total, no skips): capture arrived with `plan_item_id: null`, a capture answering a *different* search still fires, an item closed by a later search, and the new tag. Validators sit outside every snapshot and `test.tags` is stripped as cosmetic, so none of it re-reds rule 2 — I confirmed that by running `check_runlogs.py` against the modified working tree, where rule 2 compares snapshot to disk and therefore saw the edit.

## Two things left for you, neither blocking

1. **The PR body's "2 in 5" table** cites `search-records/v1_2026-08-13_17-42-37`, which this PR prunes — correctly, retention is newest-5 at the writer. The shapes survive as inline fixtures in `test_capture_pending_validator.py`; point the table there or at a commit sha so the evidence stays reachable.

2. **One follow-up issue, for two prose gaps that share a run.** Both are in `search-external-sites/SKILL.md`, so one `make eval-skill` covers them, and per CLAUDE.md's ladder the qualifying reason to defer is that it is a different skill's paid eval slot:

   - **No ending for "the user won't capture."** The table grants `skipped` only for an *inaccessible* site the user asks to skip. A researcher who simply won't buy the subscription — and never calls the site inaccessible — leaves the item `in_progress` permanently. `research-exhaustiveness` then refuses (its SKILL.md:39), and the route step 7 offers, "research-plan for re-planning", lands in review mode, which by `research-plan/SKILL.md:93-118` cannot create items; add-new mode requires every item already terminal. You name that trap in "deliberately not in this PR" — worth writing down that it is now also reachable from step 7's own offer.
   - **Step 6 should say to carry `planItemId` onto the capture entry.** My commit approximates this on `(site, url_generated)`, but the guarantee belongs in the skill text; the validator can only infer it.

   Filing it against the `research-plan` handoff mode you scoped in Slack seems right — "let the user abandon it" and "let the plan grow a fallback" are the same question from two sides, and they should be designed and graded together.

Approving rather than holding for those: pre-change, the skill ticked a search nobody ran and `research-exhaustiveness` said yes on that basis. Post-change the worst case is an item that stays open and an evaluation that refuses. A refusal is visible and recoverable; a false "exhaustively researched" is neither.

johnmarkpeterbrown (changes requested):
Nice work, and the rework in `e9fce0a0` is a real catch. Two things in it need tightening before this goes in, plus four small cleanups.

Merge `main` first — you're 21 commits behind. I merged it locally and checked: clean merge, harness suite green (2529 passed, 9 skipped), and `main` touches none of the 49 ext-sites or 177 search-records snapshot files, so it will not re-red runlog rule 2. No re-run needed for the merge.

## 1. The guard can now be switched off by a later log entry

Keying `latest` on every tool means one more entry for the same plan item makes the check skip. That includes a later search that found nothing, and a later `external_links_search`, which only discovers links. In both cases `completed` passes with the capture still outstanding. I ran the old and new versions side by side on the bug shape plus one appended entry — the old one fired in every case, the new one fires only when the handoff is the last entry for the item.

The case you were fixing — the record turns up on FamilySearch after the handoff — only needs the later entry to be a *positive* result.

The same loop has a second problem: `captured` is keyed on `(site, url_generated)` with no plan item, so a capture belonging to a different item, or one already sitting in the scenario fixture, clears this item.

One patch fixes both, and your 13 tests stay green.

```
You wrote:      for entry in after.get("log") or []:
                    pid = entry.get("plan_item_id")
                    if pid:
                        latest[pid] = entry
                    if entry.get("tool") != "external_site":
                        continue
                    detail = entry.get("external_site") or {}
                    url = detail.get("url_generated")
                    if detail.get("capture_received") is True and detail.get("site") and url:
                        captured.add((detail["site"], url))

Change it to:   for entry in after.get("log") or []:
                    pid = entry.get("plan_item_id")
                    tool = entry.get("tool")
                    # Only an external-site entry, or a records search that actually
                    # FOUND something, says how an item ended. A later nil search — or
                    # another external_links_search, which only discovers links —
                    # leaves the outstanding capture outstanding.
                    if pid and (
                        tool == "external_site"
                        or (entry.get("outcome") == "positive" and tool != "external_links_search")
                    ):
                        latest[pid] = entry
                    if tool != "external_site":
                        continue
                    detail = entry.get("external_site") or {}
                    url = detail.get("url_generated")
                    if detail.get("capture_received") is True and detail.get("site") and url:
                        # Scoped to the item it answers: an arrival naming a DIFFERENT
                        # item must not launder this one.
                        captured.add((pid, detail["site"], url))
```

```
You wrote:      if (site.get("site"), site.get("url_generated")) in captured:
                    continue

Change it to:   key = (site.get("site"), site.get("url_generated"))
                if (pid, *key) in captured or (None, *key) in captured:
                    continue
```

Add two tests alongside your existing four: `completed` plus a later nil `record_search` for the same item must fire, and `completed` plus a capture naming a different plan item must fire.

Neither hole is reachable in today's corpus — no committed run puts a non-external-site entry after an external-site one for the same item, and all 83 fixture external-site entries already carry `capture_received: true`, which is also why the false positive you fixed was unreachable. Worth saying both halves of that trade in the commit message.

## 2. ut_search_external_sites_013 makes the skill invent a plan item

`mid-research-flynn` has no open external-site plan item — `pli_002` (Ancestry) and `pli_003` (MyHeritage) are already `completed`, and the message asks for 1880 while `pli_002` is the 1850 search. In the four runs before the judge_context edit the skill wrote no status at all. In the run after it, it created a new `pli_007` inside `pl_002` — the FamilySearch probate plan — and marked it `skipped`.

`ownership.json` says research-plan owns item structure and executing skills only update `status`. Nothing catches this: the validator is tag-exempt here, and the ownership check keys on top-level sections.

```
You wrote:      "The external-site plan item should be marked terminal so the autonomous run can proceed to the next step (e.g. research-exhaustiveness). Under --autonomous the correct terminal status is `skipped`, not `completed` (SKILL.md step 7): no user can ever supply a capture, so nothing was searched — `skipped` is terminal AND honest, where `completed` would claim a search that did not happen. Do not mark down for `skipped`; do mark down for leaving it `in_progress`, which stalls the run.",

Change it to:   "This scenario has NO open external-site plan item — pli_002 (Ancestry) and pli_003 (MyHeritage) are already `completed`, and the 1880 search this message asks for matches neither. So no plan-item status may be written here, and creating a new plan item to have something to mark is a defect: research-plan owns item structure, and an executing skill may only update the `status` of an item that already exists. Grade the log entry and the return-of-control, not a status. (When an autonomous run DOES have an open external-site item, SKILL.md step 7 makes `skipped` the correct terminal status — terminal so the run does not stall, and honest that nothing was searched.)",
```

## 3. Two in-tree citations point at the run log this PR deletes

Florence flagged the PR-body table. The same deleted file is cited in the test she suggests pointing it at, and in a guide juniors read.

```
eval/harness/tests/unit/test_capture_pending_validator.py:10

You wrote:      in v1_2026-08-13_17-42-37 and v1_2026-08-17_18-06-27 and at `in_progress` in

Change it to:   in v1_2026-08-17_18-06-27 (and in v1_2026-08-13_17-42-37, aged out of the
directory by the 5-candidate retention prune in this commit) and at `in_progress` in
```

```
docs/skill-deep-dive-guide.md:140

You wrote:      > `ut_search_records_013`, run `v1_2026-08-13_17-42-37`.

Change it to:   > `ut_search_records_013`, run `v1_2026-08-14_03-36-42`.
```

The behaviour that example quotes is present in all five surviving run logs, so that is a repoint, not a lost example.

## 4. Annotate the one test the rubric change was written for

The `review_sample` for both runs is machine-picked and contains neither `ut_search_external_sites_pic` nor `ut_search_records_026` — so the new Log-entry status clause was graded by the judge and nobody else. No paid re-run needed; extra corrections beyond the sample are allowed. Add them in the CRUD UI, not by hand.

## 5. Casing in a bullet the judge checks against the file

`eval/tests/unit/search-external-sites/plan-item-stays-open-pending-capture.json:29` — this test sets `judge_reads_files: true`, so the judge compares this line to the persisted file, where the field is snake_case.

```
You wrote:      "The log entry must record the in-flight search: `tool: external_site`, `outcome: partial`, `external_site.captureReceived: false`, and the generated URL.",

Change it to:   "The log entry must record the in-flight search: `tool: external_site`, `outcome: partial`, `external_site.capture_received: false`, and the generated URL.",
```

## 6. The PR body's last section reads "do not merge yet"

"Still outstanding before merge: two paid eval runs and their annotation" — both landed, and they're annotated. Replace that section, and add a line for `e9fce0a0` so the body describes what is actually on the branch.

Two other numbers while you're in there. "0 of 14 existing ext-sites tests" is now false — `ut_search_external_sites_fhk` writes a plan-item status in both committed runs, so the guard has two live assertion sites, not one. And the body never says that the superseded `v1_2026-08-20_21-55-13` ext-sites run was committed without an annotation, or that committing it evicted the annotated `v1_2026-07-27_22-27-37` pair.

---

For the record, what I checked rather than took on trust: the reworked validator still asserts non-vacuously on `_pic`, `_fhk` and 8 search-records tests including `_026`; replayed over every committed `026` run log it still fires on the `completed` run and stays quiet on the four `in_progress` ones; and the patch above closes all three holes while keeping the legitimate FamilySearch-found-it-later case passing and all 13 of your tests green. Florence's "no cross-test contradiction" point holds — `ut_search_records_018` already tells the judge `pli_001` "must NOT be marked 'completed'".

I agree with her follow-up issue, and with putting step 6's `planItemId` guarantee in the skill text — that is what would let the validator stop approximating on `(site, url_generated)` at all.

The direction of this PR is right and the argument for it is well made. Requesting changes on the validator holes, not on the doctrine.

florencemashipei:
Reviewed after `733936e4`. **Rule 3 is the only failing check — both fresh runs are missing their `.ann.json`.** Per `per-pr-review-workflow.md` §2.2 that file is the submitting team's; this is the senior half.

@promise-emmanuel's guide citation was right and my read was wrong: `v1_2026-08-21_00-13-34` does hold the quoted turn — my grep missed it on the bold markers inside the phrase — so keeping the original quote was the better call. The docstring half of item 3 needed the fix it got: @johnmarkpeterbrown's wording predates 103 commits of retention, and `v1_2026-08-17_18-06-27` is gone. I pushed one follow-up in `e17aa491` — `captured` was still declared `set[tuple[str, str]]` after item 1 started inserting `(pid, site, url)`. Comment-only, outside every snapshot.

**The runs.** `search-records` 25/28, up from 23. All three failures — `_013`, `_014`, `_027` — are the same validator, `test_pre1880_census_structure_marked_inferred` from #1786, whose docstring already documents inconsistent compliance; the membership shifts run to run. For the annotation: all three have **zero graded dimensions**, because a hard validator failure short-circuits the judge. Nothing to grade there, and the empty dimensions are not a judge problem. `search-external-sites` is 13 pass / 1 partial / 1 fail — same shape as `v1_2026-08-20_22-45-06`, with `_012` failing as it has since 2026-08-10.

**Item 4 is half-covered.** `ut_search_external_sites_pic` is in this run's `review_sample`, so it gets annotated in the normal pass — that was not true of the run John reviewed. `ut_search_records_026` is not in the search-records sample and still needs a correction beyond it.

**Item 6.** The body still says both runs are outstanding, and still carries the "0 of 14" figure. Worth adding rather than only correcting: the docstring now cites `v1_2026-08-21_19-08-50`, a `main` run where `_026` ends `completed` with the capture outstanding and the judge graded it a pass. That is `main`'s newest search-records log, so it is the one the active-state check resolves to — stronger evidence than the 2-in-5 framing it would replace.

Once both `.ann.json` files land I will re-review. The ext-sites rubric changed and this is its first graded run, so the grading half needs looking at rather than a carried-forward approval.

DallanQ:
**Coordinate the merge order with @chesworthrm on PR #1865 before either of you merges.**

You are both changing the `search-records` eval snapshot from separate branches off `main`:

- This PR changes `eval/tests/unit/search-records/live-callee-external-sites-escalation.json` (content hash `a6a99a40…` → `ab10cf8e…`).
- PR #1865 changes `search-records/SKILL.md` plus `references/place-date-mechanics.md` and `references/search-strategy-levers.md`.

You have each committed your own fresh `search-records` run log today. Whichever merges second has its run log go inactive against the new snapshot, `check-runlogs` goes red, and that run has to be bought again — a paid `make eval-skill SKILL=search-records` plus a full re-annotation.

Whoever goes second:

1. Rebase on `main` after the other lands.
2. `make eval-skill SKILL=search-records`
3. Re-annotate the new run log and commit it in this PR.

Nothing is wrong with either change — this is only about sequencing so you buy one extra run instead of two.

johnmarkpeterbrown:
The code is done. I re-checked all six items against `03cb82f1` by running them, not reading them — the three validator holes are closed, `_013` no longer writes a plan item at all, both citations resolve, the casing is fixed, 17 validator tests pass, and the harness suite is green. The guard is live on 2 ext-sites tests and 8 search-records tests. Nothing left to change in the diff.

**What's blocking is one thing: the two `.ann.json` files.**

`runlogs` is red on rule 3 — `v1_2026-08-24_09-59-31.json` has no annotation, in both `search-external-sites` and `search-records`. That's the only failing check, and per `per-pr-review-workflow.md` §2.2 the file is your team's to write. No paid re-run needed; both run logs are already committed and current.

Two things to get right while you're in the CRUD UI:

**Start with `ut_search_external_sites_pic`'s Log entry dimension.** It's in this run's `review_sample`, so it lands in the normal pass. It is also the one dimension in this PR that has never been read by a person — the rubric change is what this run is the first to grade, and right now the judge is its only grader. That is the dimension a senior will look at first.

**Then add `ut_search_records_026` beyond its sample.** It isn't in the search-records `review_sample`, so the normal pass won't reach it. Extra corrections outside the sample are allowed and rule 3 accepts them.

On the three search-records failures — `_013`, `_014`, `_027` — you're right that there's nothing to grade: all three have zero graded dimensions because the hard validator failure short-circuits the judge. Worth saying that in the annotation rather than leaving it to be inferred.

**One line in the PR body is now stale.** "Still outstanding" says neither test appears in either `review_sample`. `ut_search_external_sites_pic` is in the fresh ext-sites sample — that changed when the run was re-taken.

```
You wrote:      Neither appears in either run's machine-picked `review_sample`, so the Log-entry status clause this PR adds has been graded by the judge and nobody else.

Change it to:   `ut_search_external_sites_pic` is in this run's ext-sites `review_sample`, so it is annotated in the normal pass. `ut_search_records_026` is not in the search-records sample and needs a correction beyond it. Until both land, the Log-entry status clause this PR adds has been graded by the judge and nobody else.
```

**Free if you're pushing anyway, not a blocker.** One comment in `validators_lib.py:250` still names the old key. `e9fce0a0` fixed the type annotation when the tuple went three wide but left the comment describing two.

```
You wrote:      # invisible per-item; `(site, url_generated)` identifies the same handoff.

Change it to:   # invisible per-item; `(plan_item_id, site, url_generated)` identifies the same handoff.
```

Once the annotations land I'll clear my requested-changes. Worth pinging @florencemashipei separately — her approval is from four commits back, and she said she wanted to look at the grading half rather than carry it forward, which is the right call given the rubric changed.

johnmarkpeterbrown (approved):
looks good!

Issue #1730, opened 2026-08-18
When reading a person from FamilySearch, we should read memories also to see what sources they might contain


DallanQ:
**Eval slot, as of 2026-08-19:** issue #1653 (`deep dive: init-project`) is in Ready and
holds the `init-project` snapshot. Either fold this into that dive's PR — it has been asked
to carry both — or wait for it to land. Starting separately buys a second paid
`make eval-skill SKILL=init-project` run.

**IMPORTANT**: If you do this, do issue #1689 at the same time. Both widen a `person_read`
flag and both edit `packages/engine/plugin/skills/init-project/SKILL.md`, so together they
land on one paid `make eval-skill SKILL=init-project` run and one annotation pass. Landed
apart they buy two.

**Touches:** packages/engine/mcp-server/src/tools/person-read.ts, packages/engine/mcp-server/src/tool-schemas.ts, packages/engine/mcp-server/src/types/, packages/engine/plugin/skills/init-project/SKILL.md, packages/engine/mcp-server/tests/tools/

## Why

**A FamilySearch person can have sources that exist only as memories**, never entered as an
official source description. Today we read `sourceDescriptions` and miss them entirely, so
a project starts blind to evidence the tree already holds.

There is no memories support anywhere in this repo today — no tool, no type, no spec
(verified 2026-08-19). This is the first.

## Scope (lead, 2026-08-19)

**Fold it into `person_read` under the existing `sourceDescriptions` flag. Do not add a new
flag.** Memories are a source channel, so they belong to the flag that already means "give
me this person's sources."

1. **`src/tools/person-read.ts`** — when `sourceDescriptions` is true, also fetch the
   person's memories and return them alongside the source descriptions.
2. **Text memories: return the text.** Story memories are already text — return their
   content directly so the agent can read them without a second call.
3. **Image memories: return the metadata, not the image.** Id, type, title, description.
   Do **not** OCR here — `image_transcribe` owns that, and OCRing every memory on every
   project creation would be a large unbudgeted spend.
4. **`src/tool-schemas.ts`** — the `sourceDescriptions` description must say memories are
   included, or nobody will know they arrive.
5. **`init-project/SKILL.md`** — teach it that a memory can be the only record of a source,
   and that image memories whose title or description carries record-type language (will,
   deed, certificate, census, obituary, marriage record, probate…) are the ones worth
   handing to `image_transcribe`. Keep it short; this file is billed on every invocation.

[...]

Issue #1700, opened 2026-08-17
Locality guide deep dive skill/reference edits and changes


Leduthet:
**Historical context signals:** "military age during major conflict? Significant migration in area? Jurisdiction existence at recorded date?" Also this should be rewritten to make it flow better.

**Locality broad context:** Land distribution systems, add land bounties to it.

**Locality-survery-methodology md:**
 **Identify record types created in that jurisdiction:** Town histories and county histories haven't been added. 
 **Identify substitute and complementary sources:** Add city directories and state census records (for the states that have them.) These records place the individuals during the census years. Also helps if the agent did not find the person in the census.
****For the output-format in the available record types:**** add what level they are held at (city/town, county or state level). For example vital records are normally held at county level (when civil registration started).
Also for probate and court records in the type section, add petitions for a type of document. Also stating if the person died in testate or in intestate. That will specify which kind of probate the agent would be searching for.
Also add land records.
**### Land records**
- **System:** [metes and bounds, rectangular survey, other]
- **Types:** [deeds, grants, patents, mortgages, tax records]
- **Date range:** [earliest available]
- **Where held:** [recorder of deeds, state land office, online?]
For the system add federal or state, and types add bounties and acts (like the homestead act).

In the Places-guidance. md and the Skill md:
Change all the wiki_place_page to FamilySearch Wiki_place_page. There are several of them. 
Skill md:
Anything that says wiki, should say FamilySearch Wiki. Another one is wiki_search, it should be FamilySearch_wiki_search. wiki_read should be FamilySearch_wiki_ read.

DallanQ:
Merged into issue #1664 (deep dive: locality-guide), with one bullet moved to issue #1653 (deep dive: init-project). Every finding is carried over in full.

These are the same piece of work as the locality-guide dive. The files you name — `locality-broad-context.md`, `locality-survey-methodology.md`, `output-format.md`, `places-guidance.md`, `SKILL.md` — are all inside `skills/locality-guide/`, which is the snapshot set issue #1664 owns. Editing any of them flips that skill's run log inactive and buys one `make eval-skill SKILL=locality-guide` run plus an annotation pass; landing these findings apart from the rest of the dive pays that twice for one skill.

Two things worth knowing, both verified today and both now recorded on the issues:

[...]

Issue #1689, opened 2026-08-17
person_read: import siblings, return memories, and surface relatives' attached sources under the existing flags, and teach init-project all three


DallanQ:
> **Slot cleared — verified 2026-09-11 (`/fill-ready`).** The block below names PRs **#2165**
> and **#2159** as holding the `init-project` slot. Both are **merged** into `main` (2026-09-07
> and 2026-09-04), and no open PR or active-column issue touches
> `packages/engine/plugin/skills/init-project/` or `eval/tests/unit/init-project/` today. **The slot
> is free and this card is startable.** Seven other open issues name `init-project` (#1831, #1962,
> #2034, #2069, #2122, #2475, #2476), so it is contended — re-check the slot at pickup rather than
> trusting this line, per the card's own standing instruction.

> **Blocker changed, not cleared — verified 2026-09-03 (board audit).** § "Not startable: the
> `init-project` eval slot is held" names issue **#1490** as the holder. #1490 closed COMPLETED
> 2026-09-01, so that line is stale — but **the slot is still held**, now by two open PRs that
> both edit `packages/engine/plugin/skills/init-project/SKILL.md`: **PR #2165**
> (source-evaluation, #1606) and **PR #2159** (#2079). `init-project`'s latest run log is
> already inactive against the working tree as well.
>
> **Still not startable.** This card's own instruction — "Re-check the slot before starting; do
> not trust this line" — is the correct one; only the named holder has moved. Re-check with
> `gh pr list -R PioneerAIAcademy/cowork-genealogy --state open --json number,files --jq '.[] | select(.files[].path | test("plugin/skills/init-project/|eval/tests/unit/init-project/")) | .number'`.

> **Reviewed 2026-08-21 before junior handoff — `senior`, and not startable today.** Both
> halves change a live MCP tool's contract. Two blocks below, then the work.

## Not startable: the `init-project` eval slot is held

PR #1800 (issue #1653, the init-project deep dive) merged 2026-08-21 and freed the slot —
and **issue #1490 took it.** That card is In Progress and its `**Touches:**` names
`packages/engine/plugin/skills/init-project/SKILL.md`. Both halves here edit that file, so
starting now buys a second paid `make eval-skill SKILL=init-project` run. **Wait for #1490,
or fold into it.** Re-check the slot before starting; do not trust this line.

## Ruling 1's premise is only partly met — read before designing the filter

[...]

PR #1800, opened 2026-08-20
init-project: deep dive — 17 findings, 8 validator requests, and the fixture that hid half of them (#1653)

### LINKED ISSUE #1653: deep dive: init-project
**IMPORTANT**: two lead-decided `init-project` edits are waiting on this slot — issue #1689
(import siblings) and issue #1730 (read memories under `sourceDescriptions`). Both are
specified and both edit `packages/engine/plugin/skills/init-project/SKILL.md`, so if you are
willing to carry them, fold both into this dive's PR and all three land on the one paid
`make eval-skill SKILL=init-project` run this dive already buys. If you would rather keep the
dive clean, say so on those two issues so they know to wait.

**Touches:** eval/tests/unit/init-project/, packages/engine/plugin/skills/init-project/SKILL.md

Audit `init-project` for defects nothing has caught. This starts from **no known

Did / Should / Gap.

## Start with this grep

```sh
grep -l -iE '"[^"]*\bscore [123]\b' eval/tests/unit/init-project/*.json
```

without writing the finding.

## This skill's numbers

| | |
|---|---|
| Tests in the suite | 11 |
| Rubric+base dimensions that never discriminate | **4 of 8** |

4 of 8 dimensions never take more than one value, so they cannot report a defect in whatever they nominally grade.

## What to hand back

## Cost

Editing anything under `eval/tests/unit/init-project/` or the skill body flips this
skill's run-log snapshot inactive and buys one `make eval-skill SKILL=init-project`

---

# Finding moved here from issue #1700 (2026-08-18)

Issue #1700 collected locality-guide dive findings, but one of its bullets is not
locality-guide — it names `init-project/SKILL.md:229`, which is this skill's
snapshot set and this issue's paid run. The rest of #1700 merged into issue #1664
(deep dive: locality-guide); this bullet moved here so neither half is left
pointing at the other.

**`init-project/SKILL.md:229`**, in the pre-flight checks, reads:

> **Historical context signals:** military age during major conflict? Significant
> migration in area? Jurisdiction existence at recorded date?

Two asks from Leduthet: the content is right but the line **should be rewritten to
flow better**, and the signals themselves are worth revisiting while it is open —
it sits directly under the "Obvious error detection" line at `:227` and reads as a
list of fragments rather than guidance.

Fold this into whatever else this dive turns up, per the batching note above —
alone it is not worth a paid run.

Emruthwill:
Closes #1653. Folds in the stray bullet moved from #1700 on 2026-08-18.
**Tier:** Normal. Plan: `PLAN.md` (gitignored), revision 4, two `/critique-plan`
rounds applied.

Deep dive per `docs/skill-deep-dive-guide.md`, starting from no known failure.
Write-ups: `docs/deep-dives/init-project-prohibition-list.md` (Step 1, 60 rules,
each marked for whether it converts) and
`docs/deep-dives/init-project-findings-2026-08-20.md` (F1–F17, Did/Should/Gap,
lane each, V1–V8 inline).

## What the corpus measures

**8 of 8 dimensions never discriminate, not the 4 of 8 the issue predicted.**
Across the five committed run logs: 49 base scores per base dimension, 39 per
rubric dimension, and **zero non-3 values anywhere**. The prescribed
score-branch grep returns 0 files — confirmed. So every finding below is a quiet
pass, which is what the guide's Step 3 predicts.

One log is not green and is not a skill defect: `v1_2026-08-17_05-17-40` has 9
of 11 tests failing, 8 with `judge.skipped` and no dimensions, from the
then-current writer-tools validator demanding a
`research_append(section: "project")` designed out before it shipped. It has no
`.ann.json` and contributes no scores.

## The root defect

Nothing in the pipeline compared a value the model **wrote** against a value a
tool **returned**. Four fields landed unverified:

| field | what happened | what checked it |
|---|---|---|
| `persons[].ark` | 18 writes over four runs in **four mutually incompatible shapes**, none canonical | nothing — `TREE_PERSON_FIELDS` admits it as a free string |
| `facts[].standard_place` | 56 values over 28 runs, every one a verbatim copy of the raw `place`, zero `place_search` calls | the one rubric dimension that would was told to score **N/A** on exactly those tests |
| `facts[].standard_date` | same class, one field over — and the fixtures did not carry it either | nothing |
| tree `sources`, objective-only path | `sources: []` and facts with no `sources` key — 006 in 4 of 4 runs, 002 in 3 of 4 | nothing — `test_id_references_resolve` checks a ref that *exists* resolves, not that one exists |

Worst concrete case: `arkToBareId` cannot parse
`https://www.familysearch.org/tree/person/details/LZNY-BRF`, so `toGedcomX`
rebuilds the Persistent identifier as a web page address.

## Half of it traced to one fixture

`person-read-flynn.json` returned a **bare person object** as its whole
response, against `person-read-tool-spec.md`'s "the top-level shape is always
`{persons, relationships, sources}`". It was the only `person_read` fixture in
the suite and **eight of twelve tests used it**. Consequences:

- `relatives: true` / `sourceDescriptions: true` — issue #1475's whole subject,
  marked "**Both flags are required**" in the body — were **unfalsifiable**: the
  payload was identical with and without them.
- Seven body rules had run **zero** times across five runs: the FAN import, all
  relationships, the Couple/ParentChild endpoint spellings, relationship source
  refs, the source-description import, the multi-person local-ID re-map, and the
  Step 5 one-row-per-person table.
- `no-relatives.json` (tagged `isolated-person`, `fan-principle`) was
  indistinguishable from `new-project-from-tree.json`.

Additionally, the fixture's `args` block — which **is** the Tool Arguments
grading target (`fixtures.py`: "the LLM judge compares the actual args Claude
passed against the matched fixture's `args`") — named neither flag. The
dimension that grades call arguments was pointed at a target omitting the
arguments that matter most on that call.

Fixed: the thin fixture now matches the contract (isolated payload retained, so
`no-relatives` stays a real test), plus a new four-person
`person-read-flynn-family.json` — wife, two children, five relationships, two
source descriptions, one carrying `notes`. **Both now predicate on both flags**,
so a call that drops one gets `fixture_not_found` and is scored down on Tool
Arguments rather than silently receiving a payload it did not ask for. That is
the first time in this suite that dropping a flag has any consequence at all.

## Grading fixes owned by the dive

- **`rubric.md`, Place standardization** — instructed to score N/A whenever a
  test's places all came from `person_read`, while its own **fail** bullet read
  "a standardized string is fabricated without the tool". 25 of its 39 scores
  were N/A. It now grades where each value came from; N/A only when a test has
  no places at all.
- **`negative-start-research.json`** — a bullet graded decline wording the
  harness structurally prevents: `text_response` is empty in 5 of 5 runs,
  because the run stops when the routed-to skill is invoked. Struck. Its
  must-not-call list also named `validate_research_schema`, a tool the skill has
  never held.
- **`new-project-from-tree.json`** — the objective bullet named only the
  substitute-FamilySearch's-value failure, and passed a run that dropped the
  user's stated fact instead. Now symmetric, without naming which occurred.
- **`from-objective-only.json` / `place-standardization.json`** — neither
  declared a `person_search` fixture, so the tool was absent from the mock
  server and both runs said so out loud before falling back to stubs. "Call
  `person_search` before falling back to stubs" could not run in the only two
  tests whose premise requires it. Two nil-result fixtures added.
- **`forget-rederive-build-full-not-omit.json`** — dropped a fixture for a tool
  the skill cannot call.

## Two decisions resolved rather than deferred

**The guard clause is a handoff.** The body said "no tool calls", while both
negative tests pass *because* the run makes a `Skill` handoff, and one judge
credited it as "made no MCP tool calls". Four things point one way: the guard's
own text names two skills as the user's next move; the negative tests'
`correct_skill` machinery cannot score routing unless the skill delegates; all
five runs delegate and the judge reads that as correct; and the "no tool calls"
clause has a purpose the handoff does not touch — stopping init-project doing
*its own* work on a populated project. The one-liner survives as the fallback
for when delegation is unavailable. **Two edits here change shipped Cowork
behaviour** rather than aligning prose to what runs already do, and both are
worth a senior genealogist's eye first if the paid run or a live session
surprises you: this guard-clause handoff, and the objective-only sourcing rule in
Step 3. The second was undercounted in the first draft of this body — it is new
prose (zero hits on the merge base), it makes init-project write a source
description and `quality: 1` refs on a path that previously wrote none, V4 now
enforces it, and it has no precedent in any other skill or spec.

**`ark` is derived, not omitted.** `person_read` structurally returns none, so
omitting would leave every imported person without the anchor
`simplified-gedcomx-spec.md` §2 makes load-bearing ("a person with no `ark` is
not in the FS tree"). Deriving is not invention, and this was checked against
the compiled helpers rather than reasoned about: `toArk("4:1:LZNY-BRF")` →
`ark:/61903/4:1:LZNY-BRF`, and `arkToBareId` of that returns `LZNY-BRF` — the
person id unchanged. It is also the **identical string** `person_search` returns
for the same person, so the two endpoints agree instead of diverging.

## Two tool-lane defects found while verifying

**`stdDate` discarded `~`.** Every approximation word resolved and the one
symbol did not:

```
"abt 1845"   -> "Abt 1845"        "~1845"  -> "1845"
"circa 1845" -> "Abt 1845"        "~ 1845" -> "1845"
```

`~1845` is **this project's own documented spelling** for an approximate date,
and `MODIFIERS` already carried `<` → `Bef` and `>` → `Aft`. Because
`getStandardDate` falls back to `stdDate(fact.date)`, every date comparison read
an exact 1845 where the tree said approximately 1845 — which is how
init-project's own "parent-child age gaps outside 15-50 years" check returns a
false verdict. Three lines, plus a regression test pinning the four tilde forms,
four unchanged forms, and `~1840-1850` → `Bet Abt 1840 and 1850` (the one output
this changes besides the bare tilde).

**`TreeFact` declared `standard_date` but not `standard_place`** — the field
reached callers only through a type predicate that narrows without stripping, so
a documented, load-bearing field travelled undeclared.

## The acceptance check, proven red per assertion

`eval/harness/tests/unit/test_fixtures.py`. CLAUDE.md requires a new lint be
broken and watched to fail; one restoration does not exercise both assertions,
so they were proven apart.

*Contract*, against `HEAD`'s fixture:

```
AssertionError: person_read fixtures must return {persons, relationships, sources}
-- the shape the tool always returns: person-read-flynn.json: top-level keys
['facts', 'gender', 'id', 'names']
1 failed, 2 passed
```

The sidecar assertion **passed** on that same restoration — a bare person object
has no `persons[]`, so no facts were collected. That near-miss is exactly why
they were proven separately.

*Sidecars*, with the fixed fixture minus `standard_date` on Birth and
`standard_place` on Death:

```
AssertionError: person_read returns a standardized sidecar beside each raw
date/place ...: person-read-flynn.json: LZNY-BRF Birth has date, no
standard_date; person-read-flynn.json: LZNY-BRF Death has place, no
standard_place
1 failed, 2 passed
```

*Green after restore:* `20 passed`.

A third function, `test_person_read_fixtures_declared`, guards the other two
against a glob that matches nothing — one of the three ways this repo has seen a
check silently pass.

## Validators — the deliverable the issue measures

Eight, and **all eight ship as code in this PR**, not as requests. Per the
standing instruction to deep-dive authors ("if additional validators are
suggested, please include those validators in your PR or in a follow-on PR"),
they live in `eval/harness/validators/test_init_project.py` with 69 mutation
tests in `eval/harness/tests/unit/test_init_project_provenance_validators.py` —
each asserted to fire on the exact defect it was written for and to pass on the
shape the 2026-08-20 run produced.

Two checks before committing them, because a failed validator skips the judge and
the grades go with it: replayed against the committed run (96 validator/test
pairs — 55 pass, 41 skip, 0 fail), and confirmed validators are **not** in the
run-log snapshot, so they cost no second paid run.

Two are deliberately weaker than the table below first proposed, and both
narrowings came from the genealogist's labels rather than from code — V7 is
tag-gated instead of message-derived, and V8 checks loss-or-alteration instead of
provenance. Either stronger form would have failed a run confirmed correct.

| id | rule | violation named |
|---|---|---|
| V1 | every `person_read` call passes both flags | none — and that is the finding |
| V2 | `ark` matches `^ark:/61903/\d:\d:` and its bare id equals a returned person id | 001 `v1_2026-08-19` — a `/tree/person/details/` URL |
| V3 | every `standard_place` traces to a `person_read` fact or a `place_search` result | 005 `v1_2026-08-19` — both values, no `place_search` |
| V4 | every fact **and relationship** carries a `sources[]` ref at `quality: 1` | 006 all four runs; 002 `v1_2026-08-18` |
| V5 | `narration_guidance` byte-identical to the table text for the stored level | none — unguarded closed 4-way mapping whose twin has a validator |
| V6 | a `notes`-carrying source reaches the tree with the **note** dropped, not the source | none — path unreachable before this PR |
| V7 | no person ID in the message + a tree person written ⇒ a `person_search` call | 002 and 006, every run |
| V8 | every `standard_date` equals the one the tool returned for that fact | none — no fixture returned one before this PR |

Two shapes were considered and **rejected as non-converting**: a
`project_create` call-count check (no observed violation, and the writer-tools
validator already asserts presence), and a stub-count check for the maiden-name
rule (whether the user "implied" a person is a reading, not a count). F11 — the
objective dropping a user-stated fact — also does not convert and stays with the
judge.

## #1700 and #1689

**#1700's stray bullet** (the historical-context signals line) is rewritten from
three fragments into guidance, and the signal set revisited as asked. Kept:
military age during a conflict, population movement, jurisdiction existence.
Added: **had civil registration begun there by the recorded date** — three runs
in this corpus improvised exactly that reasoning unprompted, which is the
evidence it belongs. The jurisdiction check now says what to do about a miss.

**#1689 (siblings + memories, with #1730 merged in) was declined**, with reasons
[posted on the card](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1689#issuecomment-5361059420):
its memories half is blocked on a probe its own scope note demands ("Nobody has
fetched a memory … pull a real person's memories with a `dev/try-*.ts` script
before writing the skill guidance"), and its siblings half is a `person-read.ts`
change whose `SKILL.md` half is a handful of words. The new family fixture is
the harness that issue needs — extend it rather than authoring one.

## Not swept, deliberately

The resolver-prefixed `ark` form is **repo-wide**: eight other MCP fixtures, one
eval scenario tree, and two shipped seed projects
(`apps/server/app/seed/sample_project/`,
`apps/electron/test/fixtures/sample-project/` — the latter also carrying a
`/tree/person/` URL in a results sidecar, the shape that breaks). Only this
skill's own fixture is touched here; sweeping the rest flips four other paid
snapshots and edits two app packages. **V2 is what should decide them.**

## Verification

- `eval/harness` pytest — **2462 passed, 3 skipped**
- `packages/engine/mcp-server` vitest — **2456 passed** (104 files)
- Every edited JSON re-parsed with `encoding="utf-8"`; CRLF preserved where the
  file had it
- `doc-links.test.ts` caught nine source-line citations in the findings doc en
  route; all are symbol citations now

**The paid run is done and annotated** — `v1_2026-08-20_22-28-10`, 12/12,
five sampled tests, committed with its `.ann.json`. The committed snapshot
matches the working tree (`diff_snapshot_vs_disk` returns `{}`), so rule 2 is
satisfied and **no further run is needed for this PR**. Nothing should touch
`SKILL.md`, the skill's `references/`, the test JSONs, `rubric.md`, or the
fixtures those tests name before it merges: any of those invalidates the run and
buys a second one. Everything added since — the validators, their tests, and the
date fix — sits outside the snapshot deliberately, which is why round 2's fixes
were free.

---

## Review round 1 — `8a74175c7`

All three blockers were real; two exposed defects in my own work rather than just
in the code under review.

- **The sidecar lint was reintroducing this dive's defect.** It required
  `standard_place` wherever `place` exists, which forbade any fixture from
  modelling a resolver miss — a case `SKILL.md` carries a rule for. Added the
  `_contract_exempt` marker the docstring had proposed and not implemented.
- **V2 skipped in the case it guards, and my mutation test pinned that as
  intended** — an `ark: ""` case asserting a skip, with a comment calling the
  boundary deliberate. Expectations now derive from the persons `person_read`
  returned. A second test (`…traced_to_a_person_search_result`) turned out to
  assert *nothing*: no `person_read`, so the validator skipped and pytest
  reported the test as skipped inside a green `47 passed, 1 skipped`. V2 now
  evaluates 8 of 12 tests on the committed run instead of skipping.
- **A live bug, not a regression from this PR.** `1845-03` →
  `"Bet 1845 and 1803"`, a range ending 42 years before it starts:
  the abbreviated-year-range rule expanded the month as a two-digit year.
  `abt 1845-03` was already wrong on `origin/main`, so the tilde fix inherited
  it. 9 regression tests; the backwards-range guard is verified non-vacuous.
- **V4 stays strict** (`SKILL.md` mandates `quality: 1`, and relaxing it would
  stop catching an overstated quality on unverified tree data). The reviewer's
  underlying point stands — the reference doc's `2` example is a real trap — but
  fixing it edits a snapshot-tracked file and would invalidate the run log this
  PR bought. Recorded in V4's docstring, deferred with two others.
- Non-blocking: `relationships[].facts` now walked on both sides of the
  comparison; the pruned `v1_2026-08-14_02-59-55` is explained in the findings
  doc with the git ref to fetch it from, since restoring it would fight the
  harness's own retention.

**Three deferrals, all snapshot-tracked, all one paid run each if done alone:**
tag `ut_init_project_004` with `expects-person-search` (V7 is dormant without
it); align the reference's three `quality` examples to `1`; give the
resolver-miss branch a fixture using the new marker. Each is recorded in the code
that depends on it, not only in the findings doc.

Re-verified after merging `main` (it touched `person-warnings.ts`, which consumes
the date helpers changed here): harness pytest **2511 passed, 3 skipped**;
mutation suite **69 passed, 0 skipped**; vitest **2472 passed** / 104 files; ruff
clean; replay **0 failures**.

---

## Review round 2 — `pending push`

Two blockers, both free (neither file is in the run-log snapshot), and the first
is the third instance of this dive's own defect class.

- **The sidecar lint's opt-out reintroduced the defect the lint exists to
  catch.** The marker was read from *inside* the fact, and the mock serves
  `response` verbatim — so the first fixture to use it would hand the skill a
  field `person_read` never returns, which `TREE_FACT_FIELDS` rejects, failing
  the `project_create` write for a reason unrelated to the test. It is now a
  fixture-level map keyed `"owner/FactType"`, beside `response` rather than in
  it. Eight new tests on synthetic fixtures, including one pinning that the
  inside-the-fact shape no longer suppresses anything — the shape that breaks the
  write must not be the shape the lint rewards. The check's core is extracted to
  `sidecar_problems()` so the logic is testable without hand-editing a real
  fixture.
- **V2 failed a correct import when two persons share a name.** Reproduced before
  fixing: two returned persons sharing a name, both written with their own correct
  arks, and V2 reported `I1: ark is '…LZNY-BRF', expected '…LZNY-P7Q'`. A failed
  validator skips the judge, so it would have cost the test its whole grade — and
  it arms the first time a fixture carries same-named kin, which is what #1689
  adds. Any same-named written person carrying the expected ark now satisfies it,
  with three tests: correct pair passes, pair missing one ark still fails, pair
  sharing one ark still fails.

Non-blocking, both taken: the `1845-10` / `1900-10` split is now pinned by a test
rather than left an undocumented surprise, and the two body claims above are
corrected. The Donovan nil fixture — the fixture set currently rewards a
married-name search over the convention-following maiden-name one — is
snapshot-tracked, so it joins the deferred table as a fifth item rather than
buying a run.

Verification: harness pytest **2522 passed, 3 skipped**; mutation + lint suites
**80 passed**; date suite **169 passed**; ruff clean.


Asimi1234:
Reviewed on the tree merged with `main` (clean, 0 commits behind, head `184f6cf2`): **no code blockers.**

Ran in the PR worktree with the engine built:
- harness pytest **2508 passed / 2 skipped**
- engine vitest **2463 passed** across 104 files
- `date-standardize.test.ts` **159 passed**
- the new fixture contract check is mutation-verified: dropping a `standard_place` from `person-read-flynn.json` fails `test_person_read_fixture_facts_carry_both_standardized_sidecars` with the exact message, restores to 20 passed; the 8 provenance validators exercise their fail direction via 11 `pytest.raises`.

Core mechanism checks out against reality, not just internal consistency:
- `person_read` really emits `standard_place` (`gedcomx-convert.ts:298`) and `person_search` returns the bare `ark:/61903/…` form via `toSimplified`→`toArk` (`gedcomx-convert.ts:153`), so the ark fixture/spec edits **correct** a prior wrong value.
- `standard_place` is accepted on every write site (`TREE_FACT_FIELDS` in `tree-shape.ts:15` + both `tree-gedcomx.schema.json` mirrors), so the new type declaration missed no schema site.
- the `~` date fix has no parallel parser to also touch, and ripples uniformly through every `stdDate`/`getStandardDate` consumer (full suite green).

One non-blocking note: the PR body is stale — it says the 8 validators are unwritten ("a developer writes the Python") and the paid run is "still to run", but commits `52ee5b0f` and `ed477957` did both. Worth updating the body so reviewers know ~844 lines of validator Python are in scope.

No code changes needed. Not a formal approval — pending the senior-developers and senior-genealogists sign-off.

Asimi1234 (approved):
Peer approval — verified on the tree merged with main: no code blockers, all required checks green, and the core mechanisms + multi-site blast radius traced clean (details in my comment above). Not a code-owner review; the senior-developers / senior-genealogists sign-off still owns the .py and doctrine judgment.

Ikennaya1 (changes requested):
Solid work — the tilde fix, ark canonicalization, `TreeFact.standard_place`, and the mutation suite all check out.

Two blockers before approval:

1. The new sidecar lint makes this PR's own `SKILL.md` example untestable. It requires `standard_place` whenever `place` exists, but the edited example intentionally returns `place` without `standard_place`. Please add the opt-out marker proposed in the docstring.

2. V2 skips in the case it exists to guard. Omitting `ark` everywhere triggers `pytest.skip`, and the truthiness filter also swallows `ark: ""`. Deriving expectations from the `persons` actually returned by `person_read` would close that gap.

One decision needed: V4 rejects any ref with `quality != 1`, but `simplified-gedcomx-summary.md` includes refs with no quality, `quality: 2`, and describes the field as optional. Either V4 relaxes or the examples need updating.

Non-blocking:

* `relationships[].facts` aren't covered by `_returned_person_facts`.
* The findings doc references a run log this PR deletes, making the score tables non-reproducible.

One thing I'd verify before merge: if `~YYYY-MM` now produces a backwards range, that's a live bug, not a doc issue.

Emruthwill:
Thanks — all three blockers were real, and one of them turned up a live bug that isn't mine. Pushed as `8a74175c7`, merged with your `main` merge in `01c81ba67`.

## Blockers

**1. Sidecar lint made a `SKILL.md` rule untestable — fixed, with the marker.** You're right, and it's worse than a false positive: the body carries a rule for exactly that case ("any returned fact with a `place` but no `standard_place`: resolve with `place_search`"), so the lint made a body branch permanently unexercisable. That is the dive's own defect, reintroduced by its own lint. Added the per-fact `_contract_exempt` marker my docstring proposed, requiring a stated `_contract_exempt_reason` so an omission reads as a modelling choice rather than an oversight, and grep finds every one.

No fixture uses it yet — giving that branch a real test means adding a fixture to a test, which is snapshot-tracked. It's in the deferred table below.

**2. V2 skipped in the case it exists to guard — fixed, and it was worse than you said.** Expectations now derive from the persons `person_read` actually returned, joined by name (the import re-ids them, so `ark` is the only place the FamilySearch id survives). An imported person with no `ark` now fails, and so does `ark: ""`.

Two things fell out of reworking it. My own mutation test *pinned the hole as intended behaviour* — I'd written `test_v2_fires_on_every_shape_the_corpus_actually_wrote` with an `""` case asserting a skip and a comment calling the boundary deliberate. That's documenting a gap as a feature. And `test_v2_accepts_an_ark_traced_to_a_person_search_result` asserted **nothing**: it supplied no `person_read`, so the validator skipped and pytest reported the test as skipped, which I'd missed in a green `47 passed, 1 skipped`. Its premise was also unreal — the skill always reads the candidate it picks. Rewritten as the real search-then-read flow, plus a new case pinning that an ark naming the *runner-up* candidate fails, which the form-and-provenance check alone would have accepted.

Verified against the committed run: V2 now **evaluates 8 of 12 tests instead of skipping**, and passes — the model does write the derived form.

## Decision: V4 stays strict, and the doc fix is deferred

`SKILL.md` mandates `quality: 1` for both the FamilySearch tree and the researcher's own statement, and 28 of 28 runs write it. Relaxing V4 would also stop catching an overstated quality on unverified tree data, which is a genealogically real error.

But you've found a genuine trap: `references/simplified-gedcomx-summary.md` documents the field as optional and shows a `2`, so a model following the reference over the body fails V4. I fixed those three example values, then **reverted it** — that file is snapshot-tracked, so it invalidates the run log this PR just bought and buys another ~$9 run plus an annotation pass. Recorded in V4's docstring and in the deferred table instead of left silent. Say the word if you'd rather pay for it now.

## The live bug — real, and not a regression from this PR

`1845-03` → `"Bet 1845 and 1803"`. A range ending 42 years before it starts.

Cause is the abbreviated-year-range rule (`1875-80` → `1875-1880`) expanding the *month* as a two-digit year. `YYYY-MM-DD` never hit it because the three-part ISO rule claims that first, so only year-month was exposed. Attribution matters here: **`abt 1845-03` was already wrong on `origin/main`** — `~` maps to the same `Abt` modifier as the word, so the tilde fix inherited the bug rather than causing it.

Fixed: a leading zero, or a backwards expansion with a legal month number, reads as ISO year-month (`Mar 1845`); a backwards expansion that can't be a month gains a century (`1845-30` → `1930`). Deliberately unchanged — forward two-digit expansions stay ranges, so `1875-80` and the genuinely ambiguous `1900-10` keep their long-standing reading. This removes impossible outputs; it doesn't adjudicate ISO against the range convention.

9 regression tests, including a guard that no output claims a range ending before it starts. I checked the guard isn't vacuous: `1845-30` matches its pattern, and the pre-fix `"Bet 1845 and 1803"` would fail it.

## Non-blocking, both fixed

- **`relationships[].facts`** now walked on *both* sides of the comparison. You were right that it mattered: the one fact in the family fixture exercising a relationship-level `place` and `date` was entirely unchecked.
- **The pruned run log.** It was the harness's own retention (`prune_old_candidates`, K=5) deleting `v1_2026-08-14_02-59-55` when this dive's run landed — so restoring it would fight retention and be undone next run. The findings doc now names the git ref to fetch it from, which keeps the tables reproducible.

## Deferred, and all three ride one future run

| Deferred | Why | Cost alone |
|---|---|---|
| Tag `ut_init_project_004` with `expects-person-search` | V7 skips on all twelve today; its firing is held by mutation tests, not the suite | one paid run |
| Align the three `quality` examples in the reference to `1` | the V4 trap above | one paid run |
| A fixture for the resolver-miss branch, using `_contract_exempt` | that body rule has no test | one paid run |

All three touch snapshot-tracked files. Each is recorded in the code that depends on it — V7's docstring, V4's — not only in the findings doc.

## Verification

`eval/harness` pytest **2511 passed, 3 skipped** (up 4, and the one silent skip is gone); mutation suite **69 passed, 0 skipped**; vitest **2472 passed** across 104 files; ruff clean; replay against the committed run **55 pass / 41 skip / 0 fail**. All re-run *after* merging your `main` merge, since it touched `person-warnings.ts` and that consumes the date helpers I changed.

Ikennaya1 (approved):
All three addressed, and I verified each by running rather than reading. Approving.

The year-month one was the real find. I'd flagged `~YYYY-MM` as unpinned but said I hadn't
reproduced it — turns out `1845-03` was becoming `Bet 1845 and 1803`, a range running
backwards past its own start. Your fix goes further than the finding asked: the property
assertion that no range ever runs backwards closes the class, not just the instance.
168 date tests pass on the head.

The sidecar opt-out is the right shape — per-fact, explicit, greppable, with a reason
field, and the docstring naming the trap ("the same defect the whole dive is about,
reintroduced by its own lint"). V2 now derives its expectation from what `person_read`
returned, so the omission case is guarded; the remaining skip is the legitimate one. You
also closed the `relationships[].facts` gap I'd raised as non-blocking. 69 validator and
fixture-lint tests pass.

On V4 — your answer is better than my framing. I said either V4 relaxes or the doc's
examples change. You kept it strict and showed why, so I checked both halves: SKILL.md:174
is unambiguous, and all 60 `quality` values across the committed run logs are 1, none
other. Relaxing would have lost the overstated-quality check for nothing. Documenting the
contradiction in the docstring and deferring the reference alignment as one paid run is
the right trade.

Also confirmed the prune note works — `git show aaa599830:...` does retrieve the deleted
log, so the score tables stay reproducible.

Nothing further from me. Thanks Ruth.

johnmarkpeterbrown (changes requested):
Nice work — the fixture-vs-contract framing is the right diagnosis, and I confirmed the two tool bugs are real.

Two things to change before I approve, and both are free: neither file is in the run-log snapshot, so no re-run.

## 1. The sidecar lint's opt-out reintroduces the defect the lint exists to catch

`_contract_exempt` is read from inside `response`, and the mock serves `response` verbatim (`mock_mcp.py:381`). So the first fixture that uses it hands the skill a `person_read` fact carrying a field the real tool never returns — and `TREE_FACT_FIELDS` rejects it, so the `project_create` write fails for a reason unrelated to what the test is checking. Nothing is broken today, but the docstring names using it as the next step.

Verified by running: adding `"_contract_exempt": ["standard_place"]` to the Death fact does suppress the check, and the served payload then reads

```
{"type":"Death","date":"1908","standard_date":"1908","place":"Schuylkill County…","_contract_exempt":["standard_place"]}
```

In `eval/harness/tests/unit/test_fixtures.py`:

```
You wrote:      exempt = fact.get("_contract_exempt") or []
Change it to:   exempt = (data.get("_contract_exempt") or {}).get(f"{owner}/{fact.get('type')}") or []
```

Add above the fact loop:

```python
exemptions = data.get("_contract_exempt") or {}
if not isinstance(exemptions, dict):
    missing.append(f"{name}: _contract_exempt must be an object")
    exemptions = {}
```

And move the docstring example out of `response`:

```
You wrote:      "facts": [
                  {"type": "Birth", "place": "Boston",
                   "_contract_exempt": ["standard_place"]}
                ]
Change it to:   "_contract_exempt": { "LZNY-BRF/Birth": ["standard_place"] },
                "_contract_exempt_reason": "models a resolver miss"
```

Still per-fact, still explicit, still greppable, still carries the reason — and the served payload stays clean. I tested the keying on relationship facts too (`Couple/Marriage`), plus the missing-reason and wrong-type cases; all four behave correctly.

## 2. V2 fails a correct import when two persons share a name

`written_by_name` keeps one person per `(given, surname)`, so a Sr./Jr. pair blames the first for the second's pid. I built the case — two returned persons sharing a name, both written with their **own correct** arks — and V2 fails:

```
I1: ark is 'ark:/61903/4:1:LZNY-BRF', expected 'ark:/61903/4:1:LZNY-P7Q'
```

A failed validator skips the judge, so this costs the test its whole grade. It arms the first time a fixture carries same-named kin — and #1689 wants siblings added to the family fixture you just built.

In `eval/harness/validators/test_init_project.py`:

```
You wrote:      written_by_name.setdefault(key, person)
Change it to:   written_by_name.setdefault(key, []).append(person)
```

Then group the expectations by name too, and compare as sets when a name is ambiguous:

```python
expected_by_name = {}
for pid, name_key in returned.items():
    expected_by_name.setdefault(name_key, set()).add(f"ark:/61903/4:1:{pid}")

for name_key, expected_arks in expected_by_name.items():
    people = written_by_name.get(name_key) or []
    if not people:
        continue
    for person in people:
        if not person.get("ark"):
            bad.append(
                f"{person.get('id')} was imported from {sorted(expected_arks)} "
                f"but carries no ark -- ark is what marks tree membership"
            )
    written_arks = {p["ark"] for p in people if p.get("ark")}
    if len(people) == 1 and len(expected_arks) == 1:
        if written_arks and written_arks != expected_arks:
            bad.append(
                f"{people[0].get('id')}: ark is {next(iter(written_arks))!r}, "
                f"expected {next(iter(expected_arks))!r}"
            )
    elif not written_arks <= expected_arks:
        bad.append(
            f"{sorted(p.get('id') for p in people)}: arks {sorted(written_arks)} "
            f"are not drawn from the persons person_read returned under that name "
            f"({sorted(expected_arks)})"
        )
```

This keeps every case you care about. I ran eight: same-name pair correct → passes; missing ark, page URL, foreign pid, wrong ark, and the empty-ark case from round 1 → all still fail. **Your 49 mutation tests pass unchanged.**

---

## Non-blocking

**The mother's maiden-name search isn't covered.** `person-search-hennessy-none` predicates on `~Hennessy`, but `SKILL.md:287` says maiden (birth) surnames for women — so the convention-following search for the mother is `surname: Donovan`, which returns `fixture_not_found` and is typically scored down on Tool Arguments. As it stands the fixture rewards a married-name search over the correct one. Both files are snapshot-tracked, so bundle a Donovan nil fixture with your three existing deferrals rather than buying a fourth run. Meanwhile the description overclaims — "a search on the mother or a relative resolves too" is only true for a married-name search, since `~Hennessy` cannot match `Donovan`.

**`1845-10` and `1900-10` now mean different things.** The guard reads a 1–12 suffix as a month only when the expansion runs backwards, so `1845-10` → `Oct 1845` while `1900-10` → `Bet 1900 and 1910`. Same notation, two readings. Still much better than the old backwards range, and your comment covers the `1900-10` half — but the `1845-10` half is neither documented nor pinned by a test. Add it beside the existing case in `date-standardize.test.ts` (both assertions verified to hold):

```js
// Same suffix, two readings — the base year decides:
expect(stdDate('1845-10')).toBe('Oct 1845');
expect(stdDate('1900-10')).toBe('Bet 1900 and 1910');
```

**The body undercounts the shipped-behaviour changes.** The objective-only sourcing rule is a second one — it is new prose (zero hits on the merge-base), it makes init-project write a source description and `quality: 1` refs on a path that previously wrote none, and V4 now enforces it. It also has no precedent in any other skill or spec, so it is the other thing a senior genealogist should look at first.

```
You wrote:      **This is the only edit here that changes shipped Cowork behaviour**
Change it to:   **Two edits here change shipped Cowork behaviour** — the guard-clause
                handoff, and the objective-only sourcing rule in Step 3.
```

Also: "Still to run: `make eval-skill SKILL=init-project`, once" looks stale. The committed run log's snapshot matches the working tree exactly (`diff_snapshot_vs_disk` returns `{}`), so rule 2 is already satisfied — worth striking before someone spends a run on it.

---

For what it's worth, I re-ran the verification rather than reading: harness pytest 2505 passed / 9 skipped, mutation suite 49 passed, ruff clean, and I broke the new fixture lint both ways to watch each assertion fail separately before restoring. The engine vitest failures in my worktree are environmental (`open` and `turndown` absent from `node_modules`), not regressions — CI is green on this head.

Emruthwill:
Both blockers were real and I reproduced each before fixing. Pushed as `86918f2ad`.

## 1. The opt-out reintroduced the lint's own defect — you're right, and it's the third instance

This is the same class three times now: the dive's thesis is "nothing compares what the skill wrote against what the tool returned," the lint I added to close it made a body rule untestable, and then the opt-out I added to fix *that* would have handed the skill a field `person_read` never returns. Each fix reintroduced the shape it was closing, one level down.

Taken exactly as specified — fixture-level map keyed `"owner/FactType"`, beside `response` rather than inside it. One addition: I extracted the check's core into `sidecar_problems(name, data)`, because the only way to exercise an exemption was the way you did it — hand-edit a real fixture and watch. Eight tests now cover it on synthetic input, including your four cases (relationship keying, missing reason, wrong outer type, wrong inner type), plus one asserting the **inside-the-fact shape no longer suppresses anything**. That last one is the point: the shape that breaks the write must not be the shape the lint rewards.

## 2. V2 and same-named persons — confirmed by building your case

Reproduced before touching it, two returned persons sharing a name and both written with their own correct arks:

```
I1: ark is 'ark:/61903/4:1:LZNY-BRF', expected 'ark:/61903/4:1:LZNY-P7Q'
```

Your fix, plus a guard against the widening becoming a hole — three tests: correct pair passes, pair missing one ark still fails, **pair sharing one ark still fails** (so a duplicate can't cover for the pid that has none). Your 49 mutation tests pass unchanged; the suite is 80 now.

Worth naming that your framing is the sharper one: this doesn't just arm "the first time a fixture carries same-named kin" — it arms on **#1689**, whose siblings ride the family fixture built in this very PR. Same-named kin is not an edge case in genealogy.

## Non-blocking

- **`1845-10` vs `1900-10`** — pinned with your two assertions and a comment saying the base year decides and why the split is deliberate rather than adjudicating ISO against the range convention.
- **The body undercounted shipped-behaviour changes.** Corrected to two, and you're right that the objective-only sourcing rule is the one a senior genealogist should look at first — new prose, zero hits on the merge base, no precedent in any other skill or spec, and V4 now enforces it. I'd flagged the guard clause and missed that this was the same category.
- **The stale "still to run" line** is struck. `diff_snapshot_vs_disk` returns `{}`, so rule 2 is satisfied and nobody should spend a run on it.
- **The Donovan fixture** joins the deferred table as a fifth item, with your finding recorded verbatim: the set as it stands rewards a married-name search over the convention-following maiden-name one, and the existing description overclaims. Snapshot-tracked, so it rides the next run.

## Verification

Harness pytest **2522 passed, 3 skipped** (up 11); mutation + lint suites **80 passed** (up 11); date suite **169 passed**; ruff clean. Your vitest failures are environmental as you said — `npx vitest run` is green here across 104 files.

Thanks for running rather than reading, twice. Both of these were invisible from the diff.

johnmarkpeterbrown (approved):
Looks good!

PR #1802, opened 2026-08-20
fix: serialize project writes behind a per-project mutex (#1715)

### LINKED ISSUE #1715: Concurrent extraction_append / research_append race on id allocation and silently lose a write
> **Reviewed 2026-08-18 before junior handoff.** Decision settled — serialize behind
> an in-process mutex, not compare-and-swap; blast radius is six writer entry points,
> not two. Original body follows from "Found while working".

**Touches:** packages/engine/mcp-server/src/utils/project-io.ts,
src/tools/{research-append,research-log-append,tree-edit,materialize-facts,merge-tree-persons,tree-forget}.ts,
packages/engine/mcp-server/tests/utils/, docs/specs/research-append-tool-spec.md

Found while working #1413's fixture (`eval/tests/e2e/stribling-father-1821/`).

Concurrent `extraction_append` / `research_append` calls race on id allocation. Two writers independently compute the same next id, one write wins, and the losing write is **silently lost** — the losing caller receives a success response naming ids that belong to the other writer's record. `validate_research_schema` passes afterward, so nothing detects the loss.

## Reproduction

In `eval/e2e-project/stribling-father-1821`, three `record-extractor` subagents were launched in parallel, each extracting a different FamilySearch document into the same `research.json`.

- Agent A (record `ark:/61903/3:1:3QS7-L9QX-2H77`) reported: `src_003` / `S3`, assertions `a_021`–`a_034` (14 assertions).
- Agent B (record `ark:/61903/3:1:3QS7-99QX-2H3Q`) reported: `src_003` / `S3`, assertions `a_021`–`a_032` (12 assertions).

A post-hoc `research_query` on `sources` showed three sources existed and `src_003` was **agent B's** record (`…2H3Q`). Agent A's entire extraction — its source entry, 14 assertions, and its tree `S` entry — was gone. Agent A had reported success.

`validate_research_schema` returned `{ valid: true, errors: [], warnings: [] }`. The loss was caught only by manually cross-referencing each agent's reported ids against the persisted `sources` array.

## Why the current design doesn't catch it

The write path is documented as validating the whole project before it persists, and writing nothing on failure. That is true, and it does prevent corruption — but the invariant it enforces is *internal consistency of the file it is about to write*, not that the file it read is still current. **A lost update is perfectly consistent; it just omits a record.**

Ids are allocated as "highest existing + 1" from a freshly-read file, so two concurrent readers see the same highest id.

## The fix — serialize in the shared write layer

Add an in-process async mutex keyed by `resolve(projectPath)` to
`src/utils/project-io.ts` (a third primitive alongside `atomicWriteJson` and
`atomicWriteBoth`), and hold it across the **whole** tool body — acquire before
the first `readJson`, release after the last write. Locking only the write, or
only the id allocator, leaves the read-modify-write window open and is the
failure this issue is about.

One lock per project, not per file: `materialize_facts` writes the tree while
`research_append`'s composite path writes tree + research together, so a
per-file lock still loses one of them.

**This beat compare-and-swap** (capture a hash at read, reject on change,
return a retryable error). CAS changes the error contract of six tools, and
`record-extractor.md` currently instructs the agent to "fix ONLY the ops named
in `errors`" and "never retry blindly" — a conflict error names no op, so CAS
only works if that agent body and `record-extraction/SKILL.md` are edited too,
which flips the record-extraction run log inactive and buys a fresh
`make eval-skill SKILL=record-extraction` plus annotation (~$8–12, 45–65 min,
blocked by `check_runlogs.py` until it lands) — and both files are already being
edited by PR #1706 and PR #1631. The lead settled the same trade the same way on
issue #1638 (2026-08-17): accept the environmental ceiling, do not buy retry
prose with eval suites. Prose-only ("document the constraint, delegate
serially") was also rejected: it is unenforceable and leaves this on the
`nothing-checks` register.

Put the one clause naming this decision and the alternative it beat into
`docs/specs/research-append-tool-spec.md`, "Persistence — validate-before-persist,
atomic", beside the commit steps. That is where the next person to re-open this
will be standing. **Do not put it in a SKILL.md or an agent body** — a prose edit
there costs the eval run this design exists to avoid.

## Sites — all six writer entry points

Lock each of these once. Unprefixed paths are under `packages/engine/mcp-server/`.

- [ ] `src/utils/project-io.ts` — the mutex helper itself, plus a doc comment
      naming the invariant it protects.
- [ ] `src/tools/research-append.ts` → `researchAppend`
- [ ] `src/tools/research-log-append.ts` → `researchLogAppend`
- [ ] `src/tools/tree-edit.ts` → `executeTreeOps`
- [ ] `src/tools/materialize-facts.ts` → `materializeFacts`
- [ ] `src/tools/merge-tree-persons.ts` → `mergeTreePersons`
- [ ] `src/tools/tree-forget.ts` → `treeForget`

**Two traps.** `extraction_append` calls `researchAppend`
(`src/tools/extraction-append.ts:40`) and `tree_correct` calls `executeTreeOps`
(`src/tools/tree-correct.ts:32`). Lock the shared function only — locking the
wrapper as well deadlocks on a non-reentrant mutex, and the symptom is a hung
tool call, not a test failure.

**Do not touch** `packages/engine/plugin/` at all. MCP source is deliberately
excluded from the eval run-log snapshot (`eval/harness/harness/snapshot.py:145`),
so a code-only PR carries no eval cost; one line of skill or agent prose costs a
paid suite.

## How you know it worked

Write the test first and watch it fail on today's `main` before you write the
fix (`CLAUDE.md`, "A new lint must be proven to fail"). Under
`packages/engine/mcp-server/tests/`, fire two `researchAppend` calls against one
temp project with `Promise.all` — no `await` between them — and assert both
records persist with distinct ids and that `sources`/`assertions` grew by the
sum of both calls, not one. `place-resolver` is already mocked in
`tests/tools/research-append.test.ts`, so no network is needed. Add the
cross-tool pair too: `researchAppend` concurrent with `materializeFacts`, both
of which write `tree.gedcomx.json`.

Then `make engine-test`, and `make test-all` before the PR.

## What this does not close

An in-process mutex binds only within one MCP server process. Every deployment
we run is one server per session (hosted: one sandbox per session; harness: one
stdio server per run), but **whether two Cowork desktop sessions on the same
project folder share one `.mcpb` process is unverified** — if they do not, the
lock does not bind there. State that residual in the spec section above rather
than designing around it; the change is strictly better than today either way.

Under contention the queued call now waits instead of losing its write. In
Cowork's cloud mode a bridged MCP call is killed at 60s (issue #1638, not
settable from our side), so a call queued behind a slow composite append can be
killed there. That is a visible failure replacing a silent loss — record it in
the spec section, do not add a timeout or a retry to work around it.

## Note on the cited fixture

`eval/tests/e2e/stribling-father-1821/` is not on `main` — it lives on PR #1725
(open). The reproduction above was a manual project run, not a committed
artifact; the code defect stands independently of that branch.

## Why the footgun is well placed to be stepped on

`/research` and `record-extraction` actively encourage batching across records, and subagent extraction is slow — 15–30 minutes per record observed — so parallelising is the obvious efficiency move. It was made here by an agent following the skill's own guidance.

Why `nothing-checks`: a run can lose an entire record's extraction, report success to the caller, and pass schema validation, with nothing anywhere reporting the loss.

Exemption (CLAUDE.md § Work you find along the way): **needs a different reviewer or skill** — MCP-server code owned by `senior-developers`, surfaced during a genealogist's fixture work.

Asimi1234:
## Summary

- Closes #1715. Concurrent writer calls against one project race on `research.json` / `tree.gedcomx.json`: each reads the pre-write state, allocates the same "highest + 1" id, and the losing write is **silently lost** — the losing caller gets a success naming ids that ended up on the other writer's record, and schema validation passes because a lost update leaves a consistent file that merely omits a record.
- Adds `withProjectLock(projectPath, fn)` to `src/utils/project-io.ts` — a FIFO async mutex keyed by `resolve(projectPath)` — and wraps the whole read-modify-write body of all six writer entry points: `research_append`, `research_log_append`, `tree_edit`'s `executeTreeOps`, `materialize_facts`, `merge_tree_persons`, `tree_forget`. One lock per **project**, not per file, because `materialize_facts` writes the tree while `research_append`'s composite path writes tree + research together.
- Records the decision (why serialization beat compare-and-swap, and the two residuals) in `docs/specs/research-append-tool-spec.md` §4.1. Normal tier.

## Review guidance

**Start here:** `AsyncMutex.run` in `src/utils/project-io.ts` — the correctness hinge is that `this.tail` is reassigned **synchronously**, before the first `await`, so callers arriving in the same tick chain in arrival order (FIFO) rather than all awaiting the same already-settled tail. Second thing to check: the mutex is **not reentrant**, so only the two shared cores (`researchAppend`, `executeTreeOps`) are wrapped — `extraction_append` and `tree_correct` route through them and are deliberately left unwrapped; wrapping them too would deadlock. I grepped the six entry points for cross-calls and the only one is `treeEdit` → `executeTreeOps` (a public wrapper calling the locked core once, same shape as `tree_correct`), which is fine.

**Unsure about:** Nothing blocking. The one stated residual (spec §4.1): the lock binds only within one MCP server process. Every deployment we run is one server per session, so it holds; whether two Cowork desktop sessions on the same folder share one `.mcpb` process is unverified — if they don't, it doesn't bind across them. The change is strictly better than today either way; I did not try to design around it, per the issue.

## Plan

**Didn't change:** the wrappers `extraction-append.ts` and `tree-correct.ts` (locked via their cores); no timeout/retry added around the 60s Cowork-cloud bridge kill (a queued call killed there is a *visible* failure replacing a silent loss — the issue says record it, not engineer around it).

**Acceptance check:** `tests/tools/write-serialization.test.ts` — two cases (two concurrent `research_append` appends; composite `research_append` concurrent with `materialize_facts`). Both fail on `main` (lost update: assertions length 2 not 3; minted person `I2` clobbered) and pass on this branch.

## Test plan

- [x] I ran `make test-all` and it passed. All suites green — MCP server 2455 passed (incl. the 2 new tests), control-plane 209 passed, eval harness 2439 passed, viewer-ui/electron/eval-app all green.
- [x] No skill run-log snapshot changed — this is MCP-server code + one spec doc only. MCP source is excluded from the eval snapshot (`eval/harness/harness/snapshot.py`), so no paid eval run is required.
- [x] No skill `description` frontmatter or DO-NOT clause changed.

## Follow-on issues

**Folded in / filed instead:** Nothing deferred; the whole issue (test, fix, spec clause, both residuals) is in this PR.


T-FEH (approved):
Approving. Careful piece of work on the category where a green suite proves least, and the two things I most expected to be wrong were both right.

**The mutex is correctly built.** I probed it directly rather than reading it: it releases on a throw *without poisoning the chain* for later callers, serializes (max 1 concurrent critical section across 8 racers), is FIFO in arrival order, and keeps distinct projects independent. The construction deserves credit — `tail` is a separate promise that only ever *resolves*, via `release` in a `.finally`, so a rejecting `fn` cannot wedge every subsequent caller. That is the classic failure of a hand-rolled async mutex and you avoided it.

**No deadlock**, which was the real risk given non-reentrancy. Verified rather than taken on trust: no locked tool imports another, and the only callers of the locked cores are the three pass-through wrappers — `extraction_append` → `researchAppend`, `tree_correct` → `executeTreeOps`, `treeEdit` → `executeTreeOps` — each acquiring exactly once. A live probe firing all eight tools concurrently on one project settled in 79ms with no hang.

**Your tests are genuinely deterministic**, which for a concurrency test is the whole question: 0/20 pass on `origin/main`, failing on exactly the stated assertions (`assertions` length 2≠3, `I2` missing), 20/20 on the branch, and 12/12 non-flaky for me. That is the right shape.

**I pushed one fix (`b35e9f2a`) — the key, not the lock.** `resolve()` alone leaves a symlinked path racing its own target. Proven: two callers naming one project as the real path and as a symlink to it both entered the critical section at once, so the lock silently did nothing. Not exotic either — on macOS `/tmp` *is* a symlink to `/private/tmp`. Now keyed on `realpathSync.native`, falling back to the resolved path when the directory does not exist yet, since the caller's own read is the right place for that to fail. It folds the case-insensitive spellings a Windows caller can produce, too.

Both properties of the key were unpinned, which is how it got here: a mutation making the key a **single constant** left your two tests green — they observe only that writes do not interleave, and over-serializing preserves that — and one keyed on the raw string let the symlink race through. Added a test for each; they separate cleanly, the constant key failing only "locks per project, not globally" and the raw resolve failing only "one project reached by two names shares one lock".

Engine vitest **2466 passed / 105 files**, typecheck clean, all 12 checks green.

Two notes, neither folded in:

- **A reentrancy guard** would turn the documented deadlock trap into a thrown error rather than a hang — an `AsyncLocalStorage` of held keys, rejecting with a message naming the key. No such path exists today, and that is plumbing rather than a fix, so it belongs in its own change if you want it. Worth it the moment someone adds a seventh writer.
- `resolveStandardPlace` runs **inside** the lock (`research-append.ts:1669`, `tree-edit.ts:302`), so a cold geocoding miss can hold up to 3×30s plus backoff against every other writer on that project. Your §4.1 second residual already covers it; I mention it only because the 60s Cowork bridge kill sits right on top of that number.

Spec §4.1 matches the code, including both residuals, and the process-scope caveat is stated honestly rather than designed around — right call per the issue.

promise-emmanuel (changes requested):
Requesting changes — two things, both mechanical. The mutex itself is right, and I
checked it rather than read it: disable the lock and 3 of your 4 serialization tests
fail on the exact assertions you named; no-op the re-entry guard and both re-entry
tests deadlock to the timeout. The call-graph claim holds too — no locked writer
calls another. Good work on the part that was hard.

What's blocking is the 193 commits this branch is behind.

## 1. A missing `projectPath` now crashes instead of answering

`lockKey` calls `resolve(projectPath)` before the tool body's guard is reached. When
`projectPath` is absent, `resolve(undefined)` throws `TypeError: The "paths[0]"
argument must be of type string`. That pre-empts the answer #1695 added — 
`{ ok: false, errors: ["projectPath is required"] }` — across all eight writer entry
points.

A user who asks Claude to record something while not in a research project sees a raw
Node type error instead of the sentence telling them what to fix.

`tests/tools/no-project.test.ts` landed in main after this branch was cut, so you have
never run it. It is 57/57 on main and 49/57 once this branch merges.

You wrote:
function lockKey(projectPath: string): string {
const resolved = resolve(projectPath);

Change it to:
function lockKey(projectPath: string): string {
// A missing or non-string projectPath is the tool body's error to report
// (#1695: readProjectJson answers "projectPath is required"), not ours to
// pre-empt with a raw TypeError out of resolve(). Key it on a sentinel.
if (typeof projectPath !== "string" || projectPath === "") return "\u0000missing-project-path";
const resolved = resolve(projectPath);



That one guard takes the full engine suite to 2707/2707 on the merge. Nothing else is
needed for it.

## 2. No suite has run against the head commit

`fd1d519a` is where the entire re-entry guard lives — the `AsyncLocalStorage` and four
new tests. No `pull_request` workflow has run on it. `vitest`, `pytest`, `runlogs`,
`e2e-fixtures` and `lockfile-drift` last ran on `309d5b50`. The PR is conflicted, so
GitHub cannot build the merge ref those workflows need.

The approval on this PR reads `b35e9f2a`, two commits before the head. It still counts
toward the merge gate, so nobody has reviewed or tested the guard.

Merge `main` and push. That resolves the conflict and re-triggers CI. Expect Engine
tests to go red on the eight tests above — fix them in the same push.

Six files conflict, all import lines only:

You wrote:      import { atomicWriteJson, backupIfExists, readProjectJson, formatIssues, withProjectLock } from "../utils/project-io.js";
Change it to:   import {
atomicWriteJson,
backupIfExists,
readProjectJson,
formatIssues,
withProjectLock,
NoProjectError,
noProjectResult,
} from "../utils/project-io.js";



Same union in the other five. In `project-io.ts`, keep main's `stat` alongside your two
new imports.

## Not blocking

Spec §4.1 reads well and both residuals are stated honestly rather than designed around.
The `resolveStandardPlace`-inside-the-lock point T-FEH raised is worth a measurement at
some stage, given the 60s bridge kill sits right on top of 3×30s — but it is declared,
and it does not block this.

Once the merge is green I will re-review.

Asimi1234:
All three items handled in the pushed merge (`44bf004b`) — thanks for the exact repro.

**1. `lockKey` no longer pre-empts the `no_project` answer.** Guarded as you diagnosed: a non-string/blank `projectPath` returns a fallback key instead of hitting `resolve()`, so the body's read reaches #1695's `{ ok: false, errors: ["projectPath is required"] }` and `no-project.test.ts` is back to **57/57**. One divergence from your snippet — I fall back to `String(projectPath)` rather than a fixed sentinel. Equivalent for correctness (every missing-path call fails fast in the body read and never writes, so that key is never contended) and it also folds the whitespace-only case via `.trim()`. Glad to switch to a single sentinel if you'd rather all missing-path calls share one key.

**2 & 3. Merged main and pushed.** The six import conflicts resolved exactly as you specified — `withProjectLock` + `NoProjectError` + `noProjectResult` in the five tool unions, and `stat` kept alongside the two new imports in `project-io.ts`. CI is now building the merge ref against head (`44bf004b`), so the re-entry guard in `fd1d519a` finally runs under `pull_request`.

Local on the merge: engine vitest **2707/2707**, typecheck clean.

The one thing I can't clear myself is the stale approval pointing at `b35e9f2a` — that's yours and T-FEH's to re-confirm once CI is green.

promise-emmanuel (approved):
Blocking items are fixed. Approving changes....

PR #1804, opened 2026-08-20
eval: remove gate-skill holdout comparison (#1756)

Asimi1234:
## Summary

- Removes `gate-skill`'s holdout no-regression comparison (issue #1756). Per the lead ruling: the gate now compares the **motivating test only**, and its report says so. The holdout arm was inert for 18 of 25 suites, and where present only previewed a check already mandatory at PR time — no skill edit merges without a full `make eval-skill` run (`check_runlogs`), which covers the whole suite.
- Deletes `holdout_test_paths` and the holdout arm of `compute_signal` / `render_report` / `main` (the `Holdout:` header, per-holdout tables, the italic inert line, both stderr NOTEs, the "absent from the baseline" skip path). Verdict vocabulary unchanged.
- Records the reasoning in `unit-test-spec.md`; leaves the `test.holdout` field in the 7 suites that carry it (still read by `/improve-skill`; snapshot-tracked). Tier: Normal.

## Review guidance

**Start here:** `eval/harness/skill_gate.py` `compute_signal` — the regression/ungraded guards now run over `mined_rows` alone. Confirm a regression or ungraded dimension in the *motivating test itself* still blocks `LOOKS GOOD` (the two reworked `test_skill_gate.py` cases cover this). The `test.holdout` field is deliberately kept — the skill-improver still excludes those tests from edit-formation.

**Unsure about:** Nothing. The issue's ruling and "Do this" checklist are explicit; this is the deletion they specify.

## Plan

**Didn't change:** the `test.holdout` field in the 7 suites (removing it would flip 7 run logs inactive — 7 paid runs to delete dead data), the `skill-improver` agent's holdout use, and the unrelated `plan §5, §6.3` docstring refs (out of scope; only the `§6.5` holdout refs were in play, and those were deleted with the code that cited them).

**Acceptance check:** `make gate-skill SKILL=citation TEST=<id>` still produces a verdict with the two `citation` holdout flags present in the JSON but no longer read (`grep holdout eval/harness/skill_gate.py` → 0). No new lint (this is a deletion); the proof runs the other way — the field is retained, unread. `make harness-test` → 2460 passed.

**Deviated from the plan:** The two holdout-specific `compute_signal` tests were reframed onto the motivating test rather than deleted outright — the `regressed`/`ungraded` code paths they cover are still live (they now apply to the mined test's own dimensions), so deleting them would have dropped coverage of live logic.

## Test plan

- [x] I ran `make test-all` and it passed (2460 passed, 2 skipped; ruff clean).
- [ ] N/A — no skill body, agent body, rubric, or test JSON changed, so no run-log snapshot changed and no `make eval-skill` run is needed (stated in the issue).
- [ ] N/A — no skill `description` or DO NOT clause touched.

## Follow-on issues

**Folded in / filed instead:** None. The issue notes what it knowingly does not close (the gate stays an n=1 mock-backed advisory signal — it now says so); nothing new was deferred.


francis-2008-happy (approved):
Approving — clean, consistent deletion. Verified by running: `test_skill_gate.py` 13 pass on the branch, and the two reframed tests genuinely cover the live paths (the regression case now fails a second dimension of the *mined* test → NEEDS YOUR EYES; the ungraded case covers a missing candidate dimension), so no coverage was dropped. Repo-wide grep shows zero surviving references to the removed symbols and both `compute_signal`/`render_report` callers updated.

Two things I checked that back up the body: keeping `test.holdout` is right — `.claude/agents/skill-improver.md` and `.claude/commands/improve-skill.md` still read it. And the "whole-suite regression is covered by the full `make eval-skill` run that `check_runlogs` requires before merge" rationale holds against the gate model, so removing the holdout arm drops a redundant pre-PR preview, not a real guard.

Heads-up on merge: `*.py` is owned by both senior-developers and senior-genealogists (CODEOWNERS), so this needs an approval from each team.

francis-2008-happy (approved):

clack391 (approved):
Approving. This is a faithful, minimal execution of the #1756 ruling: the holdout comparison is fully removed from `skill_gate.py`, the report now states its narrowed scope in the ruling's exact words, the verdict vocabulary (LOOKS GOOD / NEEDS YOUR EYES / INCONCLUSIVE) is unchanged, the reasoning is recorded in `unit-test-spec.md`, and no `eval/tests/unit/*.json` is touched so the holdout suites' run logs stay active.

What I verified, by running rather than reading:

- No dangling references: zero surviving `holdout` mentions in `skill_gate.py`, and nothing imports the removed symbols. `test_skill_gate.py` passes (13), and the two reframed tests genuinely exercise the live `compute_signal` regression and ungraded branches on the mined test, so no coverage was dropped.
- An adversarial pass on the deletion could not produce a wrong or misleading verdict: fix-lands-plus-a-mined-dimension-regresses still returns NEEDS YOUR EYES, and the scope line prints in the real report path.
- The one harness failure in a fresh worktree is `test_mock_mcp.py::test_successful_call_carries_no_is_error`, which is the environmental missing-engine-build issue, not anything this PR touched. CI pytest is green.

One optional, non-blocking note (please do not re-push just for this): `docs/specs/unit-test-spec.md:353` now has the embedded schema paraphrase say "gate-skill no longer reads this field," which the canonical `docs/specs/schemas/unit-test.schema.json` description does not carry. Nothing is false, the embedded block was already a paraphrase rather than a byte copy, and the note right below the field table already states the fact, so this is a consistency tidy at most: either trim that clause from the embedded description, or leave it.

PR #1805, opened 2026-08-21
deep dive: search-images — findings + validator requests (#1661)

### LINKED ISSUE #1661: deep dive: search-images
**Touches:** eval/tests/unit/search-images/, packages/engine/plugin/skills/search-images/SKILL.md

Audit `search-images` for defects nothing has caught. This starts from **no known

Did / Should / Gap.

## Start with this grep

```sh
grep -l -iE '"[^"]*\bscore [123]\b' eval/tests/unit/search-images/*.json
```

without writing the finding.

## This skill's numbers

| | |
|---|---|
| Tests in the suite | 12 |
| Rubric+base dimensions that never discriminate | **2 of 6** |

2 of 6 dimensions never take more than one value, so they cannot report a defect in whatever they nominally grade.

## What to hand back

## Cost

Editing anything under `eval/tests/unit/search-images/` or the skill body flips this
skill's run-log snapshot inactive and buys one `make eval-skill SKILL=search-images`

kofiatinka12:
**Closes #1661.**

## Docs-only deep dive — no validators implemented

This is a **documentation-only** deep dive of the `search-images` skill, per
`docs/skill-deep-dive-guide.md`. It adds two files under `docs/deep-dives/` and
changes **nothing else** — no tests, skills, fixtures, validators, rubric, or
`judge_context` were implemented or modified. The validator requests below are
**handed off, not built**; a developer implements them in a separate PR (per the
guide, the genealogist supplies the rule, the developer writes the Python).

- `docs/deep-dives/search-images-prohibition-list.md` — the Step 1 checkable-rule
  list (27 rules), saved for the next auditor.
- `docs/deep-dives/search-images-findings-2026-08-20.md` — 6 findings in
  Did/Should/Gap form with lanes, a secondary observation, lane summary, and the
  validator requests.

## Why no skill/test/rubric edit and no paid run

The prescribed grep (`judge_context` naming a `score N` branch) returns **0
files**, and the `rubric.md` wording is sound — the two non-discriminating
dimensions (`Tool Arguments`, `Volume selection`) are pre-decided-good on today's
corpus, not mis-written. There is no grading-prose bug to fix, so no file under
`eval/tests/unit/search-images/` or the skill body is touched and this dive buys
no `make eval-skill` run. Docs under `docs/deep-dives/` trigger none of the eval
CI gates.

## Final validator requests (handoff — not implemented here)

| V | Rule | Evidence |
|---|---|---|
| **V1** | `image_search` argument integrity: `imageGroupNumber` not imageId-shaped (`^\d+_\d{5}$`); no `offset`/`limit`/`imageIndex`/`imageId` param; no two calls on the same group | The body's "single most common mistake"; `Tool Arguments` never discriminates so can't catch it |
| **V3** | A `log` entry whose `tool` contains `image_search` requires ≥1 real `image_search` call in the run | `ut_search_images_004` — **0-for-5** across all committed run logs; `Browse audit trail` scored it 3 every time |
| **V5** | A search-images `research_append` `plan_items` update's `fields` keys ⊆ `{status}` | `ownership.json:103` states it in prose; only section-level ownership is enforced |
| **V6** | A routing-decline negative makes no `volume_search`/`image_search` call, adds no `log` entry, writes no sidecar — on **every** decline branch | Only the planning branch (`009`) is guarded today; `005`/`006`/`007` by nothing |
| **V7** | No two `research_log_append` calls with identical args where the first returned `{ok:false}` | Preventive (SKILL.md recovery rule) |

**Dropped after a coverage sweep as already-covered (documented in the findings):**
- **V2 (no `image_read`)** — enforced by the universal `test_no_main_thread_subagent_only_calls` + `SUBAGENT_ONLY_TOOLS` (`context_policy.py:79`).
- **V4 (no `sources`/`assertions` write)** — enforced by the universal `test_ownership_table` + `ownership.json` (search-images is not a caller of either section).

The grouped developer issue for V1/V3/V5/V6/V7 is **not** filed yet, by request.

## Verification

- Every prohibition traced to `SKILL.md` (the one F2-derived clause was removed
  from the list and left in the finding).
- Every cited line/file resolves (`context_policy.py:79`, `test_universal.py:439/695`,
  `test_search_images.py:118`, `ownership.json:103`).
- Recurrence counts re-checked against all 5 committed run logs.


T-FEH (approved):
Approving. This is a strong dive — the quiet passes are real, the lanes are right, and the "none, deliberately, and why" section is exactly the kind of reasoning that makes a findings doc reusable rather than a one-off.

**F2 is the best of them, and it verifies completely.** Across all five committed run logs for `ut_search_images_004`: **0** `image_search` calls, `research_log_append` logging `tool: "image_search"`, and `Browse audit trail` scoring **3** — five times out of five. A log claiming a search that never ran, and the dimension nominally grading exactly that passing it every time. That is precisely what this exercise is for.

I also checked the scope question your PR invites, since #1661 says *"For grading defects you own (`judge_context`, `rubric.md`): the fix, made"* and this PR edits nothing under `eval/tests/unit/search-images/`. Your answer holds: the score-branch grep genuinely returns 0, matching the issue's own count, and the two non-discriminating dimensions are unexercised rather than mis-written. Naming the two fixes you held, with the paid-run cost as the reason, is the right way to leave that decision visible instead of silent.

**I pushed two corrections (`82f76009`).** Both are sentence-level, both in claims that were *understating* their own findings, and I fixed rather than returned them because this document's value is that the next auditor can trust its numbers.

- **F1** said "the one dimension that would catch either has never scored below 3." It has — `Browse execution` scored **1** on `008` in `-08-10`. But I read the rationale: that 1 was for *"skipped `image_search` entirely"*, a different clause of the same fail bullet, not the invented-parameters clause your finding is about. Restated to say the parameter clause specifically has never been exercised, which is your actual point and is stronger for being exact.
- **F2** said `008` logs `tool: "image_search"`, `outcome: "error"` with zero calls in 3 of 5. The tool-with-zero-calls shape does hold 3-of-5, but the outcome is `negative` in `-07-25` and `partial` in `-07-28` — `error` only in `-08-10`. Bundling the outcome into "the same shape" made a 1-of-5 detail read as 3-of-5. The mislabelled `tool` is the constant, which is what the finding rests on anyway.

Everything else I spot-checked held: the `image_search` args across all five logs are exactly the four group numbers with no `imageId`/`offset`/`limit`; `004`'s `judge_context` quote is verbatim; the F3 table (001 15/15, 002 0/15, 011 15/15, 012 0/22) is exact; `plan_items.enforceableAt` really is `[]`; `image_read` really is in `SUBAGENT_ONLY_TOOLS`; and all 27 prohibitions trace to real SKILL.md rules. Suites on the merge: vitest **2462**, harness **2460 passed / 2 skipped**, doc-links **130**. All 12 checks green.

Three non-blocking notes:

- **V3 turns `004` red on every run** while SKILL.md step 6 still shows `tool: "image_search"` as its only example. Whoever builds V3 has to ship the body fix with it, or the first run after it lands looks like a regression. Worth saying so in the request.
- **V6's text says "tagged as a routing decline"**, but no such tag exists — and that contradicts V6's own selling point ("does not depend on remembering to tag each future decline test").
- The **"Where to look"** lines name `output.file_changes` and `is_error`; a validator receives `before_state` / `after_state` / `tool_calls` / `text_response`. Still buildable — `tool_calls` carries the response — but rename them so nobody writes a validator declaring a parameter the runner cannot supply. That exact mistake is live on #1764 right now, where two validators declared a `text_response` the runner did not provide and hard-failed every test in the suite rather than skipping.

One aside worth passing on: `grep -c "for that reason"` returned 0 on my own correction, because the phrase wrapped across two lines. Same trap that hid a stale README paragraph from five reviewers on #1735 — line-oriented greps are not a coverage check on prose.

Asimi1234 (approved):
Approving as a second reviewer. I verified the load-bearing claims independently before reading T-FEH's review, and they hold.

What I re-ran against the corpus and the cited files:

- **F2 is exact.** All five runs of `ut_search_images_004`: zero `image_search` calls, the log records `tool: image_search`, and `Browse audit trail` scored 3 every time. `008` shows the same mislabel in 3 of 5, with outcomes negative / partial / error — as written.
- **Every citation resolves**: `context_policy.py:79`, `test_universal.py:439/695`, `test_search_images.py:40/118`, `ownership.json:103`, `SKILL.md:98-100/242-245`. The existing browse-log validator really does accept `image_search` or `volume_search`, so it genuinely cannot catch the mislabel.
- **Both dropped validators are correctly dropped** — not missed gaps. `image_read` is in `SUBAGENT_ONLY_TOOLS`; `sources`/`assertions` belong to record-extraction, not search-images.
- Numbers check out: 11 pass / 1 fail (`008`), exactly 2 of 6 dimensions never discriminate, `plan_items.enforceableAt` is empty.

No blocking issues. This is an accurate, well-scoped docs-only dive.

One thing worth fixing before a developer builds these, since the branch is open and it costs no paid run: the "Where to look" lines name `output.file_changes` and `is_error`, but a validator receives `before_state` / `after_state` / `tool_calls` / `text_response`. Declaring a parameter the runner does not supply hard-failed a whole suite on #1764. Renaming those to the real inputs (the info is all in `tool_calls`) would prevent that. This is T-FEH's third note — I'm seconding it as the one most worth folding in.

On merge gating: `docs/deep-dives/*.md` matches no rule in CODEOWNERS, so no senior team is required here. With this approval the PR has its two, and nothing else is standing in the way.

Asimi1234 (changes requested):
Converting my earlier approval to request-changes — not because anything I verified is wrong (it all still holds), but so this doesn't get squash-merged before the one handoff fix lands. The branch is open and this costs no paid run, so it belongs in this PR rather than a follow-on.

**What to fix:** the "Where to look" lines in the validator requests name inputs a validator does not receive. A validator gets `before_state`, `after_state`, `tool_calls`, and `text_response` — not `output.file_changes` or `is_error`. Declaring a parameter the runner cannot supply hard-failed an entire suite on #1764, so this is the exact trap the requests should not hand a developer.

Please rewrite the affected "Where to look" lines to name the real inputs:
- `output.file_changes` / `output.tool_calls` → `tool_calls` (and, for persisted-side checks, `before_state` / `after_state`)
- `is_error` → the result field carried on each entry in `tool_calls`

Two smaller things worth folding in the same pass, both already noted above:
- **V6** says "tagged as a routing decline," but no such tag exists — which contradicts V6's own selling point that it "does not depend on remembering to tag each future decline test." Reword so V6 selects decline negatives without a tag.
- **V3** turns `ut_search_images_004` red on every run while `SKILL.md` step 6 still shows `tool: "image_search"` as its only example. Add a line to the V3 request saying the body fix must ship with it, so the first run after V3 lands isn't read as a regression.

Everything else I verified independently stands — F2 is exact across all five runs, every citation resolves, and both dropped validators are correctly dropped. Happy to re-approve as soon as the "Where to look" inputs are corrected.

kofiatinka12:
Thanks both — all three fixes are in (`510d22b5`), docs-only, no paid run.

**1. "Where to look" inputs corrected (the blocking one).** Every validator request now names the real inputs — `before_state` / `after_state` / `tool_calls` — instead of the run-log envelope fields `output.file_changes` / `output.tool_calls` / `is_error`. Added a one-time **Validator input contract** note up top: a `tool_calls` entry carries `tool` / `args` / `response` (a call's result is its `response`, e.g. `research_log_append` returns `{ ok: … }`; there is no `is_error`), and a file change is `after_state` minus `before_state`. I kept `output.*` in the **Did** sections, since those quote the run-log envelope as evidence, and the note now says so explicitly so the two can't be misread against each other. This is exactly the #1764 trap, so thank you for catching it before a developer built against it.

**2. V6 no longer references a tag.** Reworded to select on `test.type == "negative"` — every search-images negative *is* a routing decline — which is V6's whole selling point and needs no tag (none exists).

**3. V3 now carries the ship-with-body-fix note.** Added: SKILL.md step 6 still shows `tool: "image_search"` as its only worked example, so V3 turns `ut_search_images_004` red on every run until that example is corrected; the SKILL.md edit must land in the same change as V3, or the first post-V3 run reads as a regression.

`doc-links` green (130/130), no new source-line cites. Ready for re-review.

Asimi1234 (approved):
Re-approving — all three of my requested changes are in `510d22b5` (docs-only, no paid run), and I verified each against the actual harness code, not just the description:

- **Validator inputs (the blocking one):** every *Where to look* now names the real inputs. Confirmed against `validator_runner.py` — a validator receives a subset of `{before_state, after_state, tool_calls, skill_frontmatter}`, and each `tool_calls` entry carries `tool`/`args`/`response` (`orchestrator.py` passes `result.tool_calls`; `mock_mcp.py` builds those entries). `is_error` lives inside the `response` envelope, so the `is_error` → `response` rewrite is correct. The new *Validator input contract* note cleanly reconciles the `output.*` envelope fields still quoted in the Did sections.
- **V6:** now selects on `test.type == "negative"`, no tag — restores its own selling point.
- **V3:** carries the ship-with-body-fix note pointing at SKILL.md step 6's `image_search` example.

Nothing else outstanding; docs-only path needs no senior. Good to merge.

PR #1806, opened 2026-08-21
ci: ungate engine-tests and match repo-root/scripts paths in gated workflows (#1776)

Asimi1234:
## Summary

- **Normal.** Closes the `nothing-checks` gap in #1776: four path-gated workflows report `success` *without running* when their `PATTERNS` scope step misses every changed file, so a change to a path a suite actually reads can merge with the guard that protects it never having run.
- Ungates `engine-tests.yml` (deletes the scope step + all six `if:` guards) and adds one exact prefix to each of the other three. This closes the whole 8-row population the ruling measured, not just the title's class.

## Review guidance

**Start here:** `engine-tests.yml` — the ungate. A complete `PATTERNS` list for that suite is "every top-level directory" (its `doc-links`/`adr-links` tests resolve every backticked citation through `repo-paths.ts`'s `REPO_ROOTS`; `windows-wrappers.test.ts` reads the root `Makefile`, `scripts/**`, `eval/*.bat`), so the ruling ungates it rather than chasing a prefix list that can't be completed. Check the header comment says why, and that no `steps.scope`/`relevant` reference survives. For the other three, check each appended prefix names the real reader in its comment.

**Unsure about:** Nothing. Each reader claim was verified against the file (`setup-feedback-case.sh` + `scripts/claude-hooks/*.py`; `gen-zod.ts`; `test_mock_agent_schema.py`).

## Plan

**Didn't change:** `workflow-logic-tests.yml` — it uses a classic `paths:` filter, a different gate shape, out of scope. Did not build the "meta-check asserting every repo-root path a test reads appears in its runner's PATTERNS" alternative — the ruling explicitly declines it.

**Acceptance check:** Renaming a `Makefile` target reddens `windows-wrappers.test.ts > "every declared target exists in the Makefile"` — a Makefile-only diff that matched no pattern on `main`, so the guard was unreachable in CI. Verified locally: baseline green → rename `harness-test:` → test red (`harness-test.bat -> make harness-test` dangling) → reverted. Ungating makes `engine-tests` run on that diff. CI-side, per row: a draft PR off this branch touching only `Makefile` / `scripts/x` / `docs/specs/schemas/x` must show the relevant job's steps NOT `skipped` (`gh api .../actions/jobs/<id> --jq '.steps[]'`).

## Test plan

- [x] I ran `make test-all` and it passed (2460 passed, 2 skipped, 30s).
- [x] No skill run-log snapshot changed (workflow YAML only) — no `make eval-skill` needed.
- [x] No skill `description`/DO-NOT clause changed — no negative-test work.

## Follow-on issues

**Folded in / filed instead:** Nothing deferred — the four edits close the full 8-row population the ruling measured. Land after #1751 (which also edits `js-tests.yml`, workspace job only; this edit was made by content so it survives the rebase); #1747 edits `eval-harness-tests.yml` comments-only.


T-FEH (approved):
Reviewed on the merged head — I merged `main` (clean, no conflicts) and pushed one minor cleanup (below). Verified every claim in the body rather than taking it on trust; all of them hold.

## The ungate is correct and provably closes the gap

`engine-tests.yml`: scope step and all six `if:` guards gone, **zero** surviving `steps.scope` / `relevant` references, zero `if:` conditions of any kind, `on: pull_request:` unfiltered. YAML parses to six unconditional steps. `vitest` ran 42s on this head — really executed, not a skipped success.

**Acceptance check reproduced end to end.** Baseline 15/15 green, then renaming `harness-test:`:

```
× every declared target exists in the Makefile
  FAIL  windows-wrappers.test.ts > Windows wrapper drift > every declared target exists in the Makefile
  +   "harness-test.bat -> make harness-test",
  Tests  1 failed | 14 passed (15)
```

Makefile reverted clean. And replaying `main`'s actual PATTERNS regex confirms the guard really was unreachable:

```
NO MATCH  Makefile                   -> engine-tests reported success WITHOUT running
NO MATCH  scripts/package-plugin.mjs -> engine-tests reported success WITHOUT running
```

Worth noting the old header already *claimed* "Runs on EVERY pull request — no `paths:` filter" while the scope step beneath it did exactly the opposite. That contradiction was the bug; the rewritten comment resolves it and says why a `paths:` filter must not return.

Cost side of the ungate: full suite locally is **110 files / 2687 tests, all green, 27s** of test time. Healthy enough to run on every PR.

## All three reader claims hold

- **`js-tests.yml` → `^docs/specs/schemas/`** — `gen-zod.ts:23` resolves `schemasDir` to `docs/specs/schemas` and reads only from there. Exact; no second schema tree uncovered.
- **`server-tests.yml` → `^docs/specs/schemas/`** — `test_mock_agent_schema.py:33` sets `SCHEMAS = REPO_ROOT / "docs" / "specs" / "schemas"`, loading `enums` (:44), `research` (:52) and `tree-gedcomx` from it.
- **`eval-harness-tests.yml` → `^scripts/`** — `test_setup_feedback_case.py:21` is `REPO_ROOT / "scripts" / "setup-feedback-case.sh"`. The encoding-lint half is the interesting one: the lint walks **the whole repo root** (`_iter_python_files(REPO_ROOT)`, :114), so "the only scanned `.py` outside the trees already matched" is an empirical claim, not a structural one. It is exactly right today — of the 250 `.py` the lint scans, the pre-change PATTERNS left precisely two uncovered:

```
scripts/claude-hooks/gate-issue-create.py
scripts/claude-hooks/test-gate-issue-create.py
```

  and after `^scripts/`, **zero**. So the prefix is load-bearing, not decorative. Since the coverage rests on where `.py` files happen to live rather than on a fixed scan list, a future `.py` outside these trees would silently re-open it — not worth a guard today, just the thing to remember.

## Completeness — the 8-row population is fully closed

Six workflows use a `PATTERNS` gate. Issue #1776's eight rows map onto exactly the four files touched here (five engine rows → the ungate; one each for the other three). I swept the three untouched gated workflows for the same defect and found none:

- `check-e2e-fixtures.yml` — `check_e2e_fixtures.py` is stdlib-only with no harness imports, and reads only the e2e trees its PATTERNS covers.
- `check-engine-lockfile.yml` — narrow and self-contained.
- `check-runlogs.yml` — the only path literals across its five scripts are `eval/harness/judge/prompt.md` and `eval/tests/unit/`, both matched.

Self-reference (`^\.github/workflows/<self>\.yml$`) intact in all three edited files, so editing a gate still re-runs its own job. `workflow-logic-tests.yml` is the classic-`paths:` shape the ruling puts out of scope.

## One thing I folded in

`2fbdbd7b` removes `pull-requests: read` from `engine-tests.yml`. The deleted scope step's `gh api` call was its only consumer; no remaining step touches the PR API, so it was a vestigial grant. YAML re-validated after the edit.

## Merge order

Issue #1776 and the body both say land after #1751, which is still open. Flagging that landing this *first* is the more interesting order, not a problem: #1751 adds a `Makefile` target, `scripts/windows/lint.bat` and a `scripts/windows/README.md` row — all three guarded by `windows-wrappers.test.ts`, whose job currently goes green in 4s with every step skipped. Ungating first would make that guard actually run against the PR it was written for. Either order is fine; the author's call.

Approving.

EdmondOware (approved):
Approved.

Ungating is the right shape here, not just the convenient one. `vitest` is a required check and it is this workflow's job, so a `paths:` filter would leave it reporting nothing at all and the PR stuck waiting on a status that never arrives. The header comment saying so, and saying not to reintroduce one, is the part that will stop this being undone in six months.

The `scripts/` prefix closes a real gap rather than a theoretical one. Two tracked Python files were outside every pattern before this, both under `scripts/claude-hooks/`, and the encoding lint scans them. After the change nothing is uncovered. Each of the other appended prefixes names its actual reader, and each one is the reader it claims.

One thing worth a line on #1776 rather than a change here. `test_encoding_lint.py` walks the whole repository for Python files, so `eval-harness-tests.yml` is a repository wide lint gated by a prefix list. That is the same property that made a complete scope impossible for this suite and led you to ungate it. The list happens to be complete today, but a new top level directory holding a Python file reopens the gap and nothing would catch it. Ungating the harness suite costs minutes rather than the fifty seconds this one costs, so it is a real decision and not a fold in, which is why it belongs on the issue.

Issue #1183, opened 2026-08-03
proof-conclusion must be taught to page research_query (the skill half of #1031)


DallanQ:
**IMPORTANT**: do not start this separately — issue #1643 (deep dive: proof-conclusion) is in
Ready and holds this skill's eval slot. Both edit `proof-conclusion`'s run-log snapshot, so
they cannot share a paid run and cannot land in parallel.

Raise this finding on #1643 so it rides that dive's run. Two constraints on how:
`proof-conclusion` became a skill-agent pair on 2026-08-21 (PR #1819), so the doctrine now
lives in `packages/engine/plugin/agents/proof-conclusion.md`, not the 63-line `SKILL.md`; and
#1643 is scoped to hand doctrine findings back as requests rather than editing the agent body.
This issue also names `research/SKILL.md`, whose slot is separately held by issue #1494.

**Touches:** packages/engine/plugin/skills/proof-conclusion/SKILL.md, packages/engine/plugin/skills/research/SKILL.md

**Unblocked and startable.** This was filed blocked on issue #1031's tool half.
That closed completed on 2026-08-03 and `offset` shipped — verified on `main`:
`packages/engine/mcp-server/src/tools/research-query.ts:60-62` (the parameter,
with its "pagination, not a filter" contract), `:205-210` (validation), `:262`
(`const start = input.offset ?? 0`), and the tool schema telling the model to
"fetch items 51+ by setting offset to 50, then 100, and so on".

## Problem

`packages/engine/plugin/skills/proof-conclusion/SKILL.md:117-119` promises the
model the opposite of what shipped:

> Each call returns only what matches — no offset/pagination guessing, and no
> cost growth as research.json accumulates more questions and records over the
> session.

Until that changes, `offset` changes nothing observable. The tool grew a
parameter no skill body ever passes, `make engine-test` goes green, and
proof-conclusion's Step 1 — whose own text is *"Collect every assertion linked
to the question"* (`:41`) — keeps writing proof summaries from the first 50 of
up to 83.

## Ask

1. **Rewrite `proof-conclusion/SKILL.md:117-119`** so it teaches the model to
   page rather than promising it never has to.
   `packages/engine/plugin/agents/gps-mentor.md:111-113` already carries the
   correct shape and is the model to copy:

   > **Watch for truncation.** `research_query` caps `items` at 50 and reports
   > `count` as the true total. If `count > 50`, you are seeing part of the set —
[...]

PR #1808, opened 2026-08-21
research: the router must page research_query, not assume one call is the set (#1183, free half)

benter-070:
## Summary

**Tier: Normal.** Partial for #1183 (the skill half of #1031) — **the free half only**; see "Plan" for what is deliberately held back.

`research/SKILL.md`'s Step 1 tells the router to pull *"which entries have produced assertions, which are classified, which are linked to persons"* and explicitly anticipates *"100+ assertions by late run"* — while never saying `research_query` returns at most **50** of them. Every one of those routing questions is an *"is this empty yet?"* question, and a truncated read answers it wrongly in the direction that **skips work**.

Adds the cap, `count` as the true total, and `offset` as the remedy. Five lines, one file, no eval run.

## Review guidance

**Start here:** the added paragraph in `research/SKILL.md` Step 1. Two things worth checking. It names the real response fields rather than paraphrasing — `count` is the total *before* the cap and any offset, `items` caps at 50, and `truncated` is `count > offset + items.length`, so `truncated` is the loop condition and `count` is the completeness check (`src/tools/research-query.ts`, the `count`/`items`/`truncated` doc comments). And the file's only prior "truncated" mention is the batch-`ops` line about `research_append` op counts — unrelated to read pagination — so this is additive, not a duplicate.

**Unsure about:** nothing on the change. One judgement I made deliberately: the sentence ends *"before concluding a section is empty, complete, or that some step has produced nothing"* rather than stopping at the mechanic. The router's failure mode is not a wrong number, it is skipping a step it thinks is done, and I would rather the body name the consequence than trust the reader to derive it. Cut it if you think that is one clause too many for a billed prompt.

## Plan

**Didn't change:** `proof-conclusion/SKILL.md:117`, the other half of #1183 and the one that carries the actual false promise (*"no offset/pagination guessing"*). **PR #1723 currently holds `proof-conclusion`'s eval slot** and rewrites its `SKILL.md`, its `rubric.md`, 18 test files and 18 run logs — so any `proof-conclusion` run made now is invalidated on all three axes the moment it merges, and we would pay for it twice. That half lands once #1723 does.

This **deviates from #1183's own instruction** to do both in the same PR. That instruction assumed the slot was free; it was, when the issue was written — #1545 has since merged and #1723 has since opened. Flagging rather than quietly re-scoping.

Also unchanged, and settled by the issue itself: `person-evidence/SKILL.md` and `search-records/SKILL.md`. The first claims nothing false about the cap (and adding a positive warning there would flip its run log inactive, which #1183 says to bundle with #1272 rather than pay for separately); the second already teaches paging for `record_search`.

**Acceptance check:** none — this is prose in a gate-exempt skill, and there is no unit suite for `research` (`RUNLOG_GATE_EXEMPT_SKILLS`, `check_runlogs.py:103`). The honest statement is that **nothing in CI can catch a regression of this line**, which is the same gap #1183 exists to close and is discussed under "A guard" below.

**Deviated from the plan:** yes — split into halves, for the slot reason above.

## Test plan

- [x] `make test-all` — **no `make` on this machine** (Windows; #1185). Ran what this change can affect: `check_skill_frontmatter.py` (clean — 27 skills + 4 agents; the description is untouched), and the two packaging lints that landed this week, `no-issue-refs.test.ts` and `prompt-budget.test.ts` (15 tests, pass). CI runs the rest.

- [x] No eval run, and none owed: `research` is in `RUNLOG_GATE_EXEMPT_SKILLS`, so rules 2 and 3 do not arm for it.

## The issue's premise, re-measured

#1183 reads as though proofs are actively being written from partial sets. The corpus does not support that, and the PR should not claim it:

| e2e corpus | |
|---|---|
| `research_query` calls | **867** |
| calls that passed `offset` | **5** |
| responses whose `truncated` flag is still observable | **336** |
| of those, `truncated: true` | **0** |

So the **dead-parameter half of the diagnosis is confirmed** — `offset` shipped and the plugin essentially never passes it. But the under-read itself has **not been caught happening**: no observable query has exceeded 50 matches, because e2e fixtures are single-question projects that do not accumulate 50+ assertions for one question. This is a **prospective** fix, reachable on a larger real project, not a live-corruption fix.

Two caveats on those numbers, both worth stating: 435 of the 867 captures are empty or stripped (the 14-day e2e `response_summary` strip), so the observable set is about half; and the unit tier records no tool responses at all, so it cannot contribute here.

## A guard: prototyped, and both shapes rejected on evidence

#1183's own diagnosis is that *"the tool grew a parameter no skill body ever passes, `make engine-test` goes green."* Prose alone leaves that intact — §3.1 measured unanchored prose decaying 77% → 3%, and ADR-0011 says a rule that must hold belongs at a boundary. So a packaging lint looked like the natural anchor, and I prototyped two shapes against all 31 plugin bodies before proposing one. **Both fail, in opposite directions.**

**Shape A — negative** ("no body may promise the absence of pagination"). Two hits, and the second is a false positive:

```
proof-conclusion/SKILL.md:117   no offset/pagination guessing            <- the real offence
search-images/SKILL.md:99       no `offset`, `limit`, `imageIndex`, ...   <- CORRECT sentence
```

`search-images` is describing `image_search`, a different tool that genuinely has no `offset`. Scoping the pattern to `research_query` context means proximity matching over prose, which is where regexes are least reliable.

**Shape B — positive** ("a body calling `research_query` must teach the cap"). Worse: it **clears the body it exists to flag.**

```
ok  proof-conclusion/SKILL.md   line 117: "...no offset/pagination guessing..."
```

The false promise satisfies the check, because it contains the word `offset`. `search-records` passes on a coincidental match too. Only `person-evidence` is flagged, and that is work #1183 explicitly declined.

(Shape B'''s first run also flagged `gps-mentor.md`, which #1183 names as the correct model. That was my pattern missing its actual wording — "Watch for truncation", "caps `items` at 50" — not a defect in the idea. Corrected before drawing the conclusion above.)

**Conclusion: prose keyword matching is the wrong instrument here,** failing in both directions — flagging a correct sentence about another tool, and clearing a false one because it uses the right word. Sharpening the patterns relocates the false positives rather than removing them; this is the "field-name match that collides with an unrelated key" failure CLAUDE.md lists among the three ways a lint silently passes.

**What would work is behavioural, and this PR already demonstrates it.** The numbers above — 5 of 867 calls passing `offset`, 0 of 336 observable `truncated` flags true — came from reading the run-log corpus, not the prose. A report in the shape the repo already uses (`provenance_report.py`, `make e2e-wiki-failures`, `make e2e-agent-tools`) would measure whether paging actually happens: it cannot be satisfied by a coincidental word, and it catches the **omission** case both lint shapes miss — the case that actually occurred here, since `research/SKILL.md` never lied, it simply said nothing.

A gate is the wrong form (the denominator moves with the corpus), but a report a human reads is right. **Not bundled here** — it is a `.py`/`Makefile` change that would pull developer review onto an otherwise genealogist-only PR, and it guards a behaviour rather than this diff.


florencemashipei (approved):
## Approving — the diff is right, and it should not wait on the notes below

Five lines of accurate prose in the one gate-exempt file that needed it, all 12 checks green, mergeable, 0 behind main. Before this, no `SKILL.md` in the plugin taught the 50-item cap; after it, the router does. Nothing in the diff is wrong, so nothing below gates the merge.

I checked the prose against the tool rather than reading it for tone: `MAX_ITEMS = 50` (`research-query.ts:28`), `count: filtered.length` computed pre-cap and pre-offset (`:262`), `truncated: filtered.length > start + MAX_ITEMS` (`:264`). The body's claim that it names real response fields instead of paraphrasing holds. The additive-not-duplicative check on the prior "truncated" mention at `:215` (batch `ops` counts) also holds.

**The lint prototype is the strongest part of this PR** and I don't want it buried by the notes below. Both shapes rejected on evidence, with the Shape B gps-mentor false positive self-corrected before the conclusion was drawn — and the conclusion is right. My scan below independently strengthens it: the decisive signal turned out to be invisible to any prose pattern, and visible only from the run-log corpus. That is the argument for the behavioural report, now with a worked example.

---

## One correction, and it runs the other way

**The premise table is measuring a field that is not in the corpus.** I reproduced the `336` exactly — it is the number of captures containing the *string* `truncated`, and every one of those is `_summary_truncated`, the harness's own capture-strip marker inside `items`. Captures containing a real `"truncated": <bool>`: **zero**.

```
research_query calls                867
  non-empty capture                 432
  capture contains 'truncated'      336   <- all _summary_truncated
  has a real "truncated": bool        0
  has a readable "count": n         212
```

The mechanism: the tool serializes `{ section, count, items, truncated }` in that order (`:262-264`), so `truncated` sits *after* `items` and the strip always eats it. `count` precedes `items` and survives. So "responses whose `truncated` flag is still observable: 336 / of those, true: 0" should read **0 observable**, and the second row is dividing by no data rather than reporting an absence of truncation.

**Read on `count`, the conclusion inverts.** 11 calls returned `count > 50`. Of the 8 that did not themselves carry an offset, **5 were never paged by any call in that run — all five `{section: "assertions", questionId}`**, the exact call #1183 names.

And it is attributable:

```
eval/runlogs/e2e/stribling-father-1821/run-2026-08-18_02-03-06.json
  312  Skill -> proof-conclusion
  313  Read
  314  research_query {section: assertions, questionId: q_001}  -> count 126, 50 items
  315  research_query {section: conflicts,  questionId: q_001}
  316  research_query {section: hypotheses, questionId: q_001}
  317  research_query {section: person_evidence, ...}
```

Calls 314-317 are Step 1 of `proof-conclusion/SKILL.md:102-115` executing in order, immediately after the Skill invocation. It read 50 of 126 and never paged. `gps-mentor` then repeated the identical under-read at call 333 in the same run. Dated **2026-08-18**.

So these lines need to change:

- *"the under-read itself has **not been caught happening**"* — it has, above.
- *"no observable query has exceeded 50 matches"* — 11 did.
- *"e2e fixtures are single-question projects that do not accumulate 50+ assertions for one question"* — `stribling-father-1821` q_001 holds **126**, `victoriano-macatangay-parents` q_001 **97**, `antonio-lucas-spouse` q_002 **61**.
- *"This is a **prospective** fix ... not a live-corruption fix."* — the opposite, three days before this PR opened.

Suggested replacement for the table:

| e2e corpus | |
|---|---|
| `research_query` calls | 867 |
| captures with a readable `count` | 212 (435 stripped/empty; the real `truncated` field never survives the strip) |
| calls returning `count > 50` | 11 |
| of those, never paged on that filter | **5** — all `{assertions, questionId}` |
| successful page-throughs in the whole corpus | **3** |

**Positive control, worth adding:** `victoriano-macatangay-parents/run-2026-08-13` — gps-mentor hit `count: 97` and *did* page with `offset: 50`, retrieving the remaining 47. The behaviour is reachable. It is just unreliable.

**And a bonus finding for the dead-parameter half.** Of the 5 calls that pass `offset` at all: one is `offset: 0` (a no-op), one passes the **string** `"50"` — which `:217` rejects rather than coerces, and whose code comment reads *"a stringified `50` (which the model has sent)"*, i.e. this call — leaving **3** real page-throughs in 867 calls. That is a sharper version of this PR's own point than the `5` in the current table.

I am posting the same correction on #1183, since that thread is where the `336 / 0` figure will otherwise be cited.

---

## `gps-mentor` is free, and it is the bigger offender — offering it for this branch

**4 of the 5 under-reads are gps-mentor**, which the issue and this body both treat as the correct model needing no edit. Its text is *"narrow the filter, and never describe a partial set as complete"* — it never names `offset`, and on `{assertions, questionId}` no narrower filter preserves the set. It is instructed into a dead end, which is why it read 50 of 126 and stopped.

Editing it costs nothing. Traced end to end:

1. `AGENT_PATH_RE` matches -> `touched_agents = {gps-mentor}` (`check_runlogs.py:740-742`)
2. `skills_referencing_agents` matches `@plugin:([a-z0-9-]+)` (`snapshot.py:41`). `@plugin:gps-mentor` appears **only** in `research/SKILL.md` (`:344`, `:354`, `:361`). `proof-conclusion:14` mentions it in prose — *"invokes the gps-mentor critique"* — which the regex does not match, so proof-conclusion is not pulled in.
3. `touched_skills |= {research}`
4. `exempt_suiteless` drops it: `research` is exempt and has no `eval/tests/unit/research/` dir (only `research-exhaustiveness`, `research-plan`) (`:775-780`)
5. `touched_skills` empty -> **no gate, no run.**

Same free bucket as the edit already in this PR. Not gating the approval — you scoped deliberately and flagged it rather than re-scoping silently, which is the right call and should not cost you a cycle. But the branch is open and this is free, so taking it here costs one review cycle instead of two. Your call; I will re-review fast if you add it.

---

## Two nits and a dead pointer

**1. Placement.** The new paragraph lands immediately before *"**Query what the routing decision needs and no more.** Each call is a turn."* Read in order, a model resolves that toward not paging. Either move it after that paragraph, or add a clause — *paging is not surveying; the extra call is required, not optional.*

**2. Answering your "Unsure about".** Keep the consequence clause — you are right that naming the consequence beats trusting the reader to derive it, and *"or that some step has produced nothing"* is precisely the failure mode (it is the `log[]`-vs-`assertions[]` cross-reference two paragraphs down at `:106-110`). The one loose word is **"empty"**: if `truncated` is true you are holding 50 items, so a *section* can never look empty. Drop that limb or reword it to the subset case. One word, not a restructure.

**3. Dead pointer.** The Plan section says a `person-evidence` warning should bundle with #1272 — **#1272 closed 2026-08-11**, so that route is gone. If `person-evidence` ever needs a positive warning it now costs its own run.

---

## On the held half, one thing to add to the Plan

The hold on `proof-conclusion/SKILL.md:117` is sound as written, but incomplete in a way that matters for cost. #1723's own review already **mandates a fresh `proof-conclusion` run taken from a main-merged tree** (main's #1545 reworded `merge_warnings` inside that same file), and bounds the window with *"just do not touch SKILL.md or rubric.md again after step 2."* So there is a live window in which line 117 rides an already-required run. Waiting for #1723 to *merge* guarantees paying for a second one. #1723 is `CHANGES_REQUESTED` with no activity since 2026-08-19, so the window is open now.

Also FYI, no conflict either way: #1723 touches `research/SKILL.md` at `:293` and `proof-conclusion/SKILL.md` at `:183`/`:234` only — it does not touch line 117, so the false promise survives it.

Numbers above are re-runnable: a short scan over `eval/runlogs/e2e/**/run-*.json` reading `tool_calls[].args` and pulling `count` out of `response_summary`. Happy to attach it or fold it into the behavioural report you sketched.

EdmondOware (approved):
Verified the prose against the tool rather than reading it for tone. MAX_ITEMS is 50, count is computed pre cap and pre offset, truncated is filtered.length greater than start plus MAX_ITEMS. Every claim in the new paragraph lands.

One correction to the correction, and it runs the same direction florencemashipei was pointing. The real truncated field does survive the capture strip. I reproduced the scan independently and get 206 captures carrying a real truncated boolean, of which 8 are true. The mechanism explanation was the part that misled here, the strip replaces the contents of items with a summary object rather than cutting the string, so a trailing truncated after items survives intact. One raw capture, verbatim, ends source_id src_005 then closing braces then truncated true.

All 8 truncation events are section assertions, and 7 of the 8 passed no offset at all, the eighth passed offset 0 which is a no op. Those are antonio lucas spouse at count 61, stribling father 1821 at 52, 51, 126, 126 and 51, and victoriano macatangay parents at 97. That is the exact call 1183 names, and the tool told the router it was holding a partial set 8 times.

So the body's not caught happening and prospective fix framing needs to change, which florencemashipei already said, but the evidence is more direct than the count based inference. No inference is needed. Suggested figures for the table:

research_query calls 867, non empty captures 432, captures with a real truncated boolean 206, of those truncated true 8, all section assertions, of those never paged 7.

Confirmed the rest of florencemashipei's notes independently. 1272 did close on 2026 08 11, so that pointer is dead. 1723 is changes requested with no activity since 08 19, so the eval slot window argument holds. Both nits are real too, the new paragraph does land immediately before query what the routing decision needs and no more, which reads against paging, and empty is impossible when truncated is true since you are always holding a full page of 50.

The lint prototype work is the strongest part of this PR and the conclusion is right. The behavioural report is the correct instrument, and these 8 direct observations are a better worked example for it than the count based ones.

No blocking defects.

PR #1809, opened 2026-08-21
Validate every committed unit test against its schema on a PR (#1380)

### LINKED ISSUE #1380: Nothing validates eval/tests/unit/** against unit-test.schema.json on a PR

aghadiayeamayanvboernest:
Closes #1380.

## What was unguarded

Nothing validated `eval/tests/unit/**` against `unit-test.schema.json` on a pull request. The schema binds in two places and neither fires on a PR:

- `harness/loader.py` validates at harness load time — so a malformed test fails loudly, but only once someone spends money on `make eval-skill`.
- `eval/app` generates Zod types from the schema, which a hand-edited file bypasses entirely.
- `check_runlogs.py` rule 4 reads every test file and checks only `test.id` uniqueness.

So the realistic failure was a hand-edited file sitting broken in `main` until the next paid run tripped over it. It also silently degraded `check_negative_reciprocity.py`, which by design skips any file it cannot parse — a malformed test is absent from the routing graph and that report under-reports without saying so.

## The check calls `load_test` rather than re-validating

`load_test` is the function whose semantics matter, so calling it means this cannot drift from the schema. Both alternatives in the original body were rejected for cause:

- **Vendoring a stdlib validator** would re-implement `allOf` / `if` / `then` / `const` and three `oneOf` branches. A subset validator passes on files it does not understand — the exact shape CLAUDE.md's "a new lint must be proven to fail" exists to stop.
- **Moving it to the engine's vitest suite** rested on ajv being available there. It isn't: present only transitively under `node_modules/`, declared in neither `dependencies` nor `devDependencies`, and imported nowhere in `src/` or `tests/`.

`jsonschema` is already a declared harness dependency, and `eval-harness-tests.yml` already runs `uv sync --frozen` + pytest on any PR touching `^eval/tests/`. So this gate fires on exactly the PRs it exists to catch, at no extra CI cost, and needed no workflow edit.

## Proven to fail — four modes, each planted then deleted

```
schema violation (type: positive carrying a negative block)  -> named
schema violation (test block missing tags)                   -> named
unparseable JSON (trailing comma)                            -> named
a mistyped corpus path                                       -> floor guard fires
```

The unparseable case is the one that mattered — it is the mode both existing readers skip silently:

```
AssertionError: unit-test files that fail unit-test.schema.json:
  eval/tests/unit/citation/_tmp_bad.json: test file is not valid JSON: …
    Expecting property name enclosed in double quotes: line 3 column 1 (char 55)
```

And the floor guard, which is why the floor exists at all — without it a wrong path makes the whole check pass over zero files:

```
AssertionError: corpus glob matched only 0 files under …/eval/tests/unitt —
  the path is wrong, not the corpus. Every assertion below would pass vacuously.
```

`git status eval/tests/` is clean — the probe file is not committed, which would have made a run log stale.

## Two implementation details worth stating

- **`rglob`, not `*/*.json`.** The harness's own discovery is `root.rglob("*.json")`, so a nested JSON file is already a would-be test and a shallower glob would leave it unchecked.
- **Every failure is collected and asserted once.** Raising on the first file would make a multi-file fix an N-round game, one CI run per file.

## Retired the stale claim

`check_negative_reciprocity.py`'s docstring told the reader that *"NO CI job validates `eval/tests/unit/**` against unit-test.schema.json today"* and called its own skip a genuine blind spot. That is now false. Rewritten to name this check, because leaving it is how a corrected fact gets quoted forward as still true.

## Two things I measured rather than took from the issue body

Both were accurate when written and have since moved:

1. **The `§9.3` aside held after all, and I got this wrong.** I wrote that the line "already reads **396** — someone fixed it after filing" and left it alone rather than "correcting" a correct number. 396 was not correct: the corpus is **400**, and was 400 when I branched. I read the number in the doc and treated it as evidence about the number, which is the exact mistake this PR exists to make impossible for the corpus. Fixed in `66e69f0d`, along with three more figures in that section — 86 negative routing tests (not 82) and 399 run-log rows / 360 passing (not 396 / 364), each re-derived rather than taken from the review.
2. **The `§9.2` heading said "Three more live outside"** where the list now holds four bullets and five checks (one bullet names two). The number was already ambiguous before my row, so I dropped it rather than incrementing something that will drift again — same call as #1777.

Corpus re-derived on the merged tree: **400 files, 0 invalid, 0 unparseable.** This is a new gate, not a corpus repair.

## Scope

No `§9.4` row — that register moved to the `nothing-checks` label on 2026-08-09, so the original instruction is void; the register entry is this issue's label and it leaves by closing. Linting the enum tables in the spec is #1268 and out of scope here.

No eval run and no annotation: `check_runlogs.py` derives touched skills from `plugin/skills/<skill>/` and `eval/tests/unit/<skill>/` paths only, so a harness test plus two doc edits makes no run log stale.

## Verification

`make harness-test` — 2462 passed, 2 skipped. `make test-all` — all checks passed.


clack391 (approved):
Approving. This is a faithful, minimal implementation of #1380: the new test validates every committed file under `eval/tests/unit/` by calling `loader.load_test`, so it cannot drift from the schema's real semantics, and it collects all failures before asserting once.

What I verified by running it, not just reading:

- The check is non-vacuous in both modes. I dropped a trailing-comma file and an empty-object file into the corpus; the test failed naming both (`test file is not valid JSON ...` and `test fails schema validation: 'negative' is a required property`), then passed again once I removed them. `load_test` wraps both the JSON parse error and the schema error into the `InvalidTestError` the test catches, so the unparseable-JSON case the issue insisted on is genuinely seen.
- The gate actually fires in CI on the PRs it needs to: the harness workflow's scope patterns cover `^eval/tests/` and `^eval/harness/`, and it compiles the MCP server before pytest.
- I ran the harness suite in a fresh worktree and saw one failure, `test_mock_mcp.py::test_successful_call_carries_no_is_error`. That is environmental (a fresh worktree has no compiled engine `build/`), not a defect here: the diff touches no engine or mock-MCP file, the same test passes where a build exists, and this PR's own CI `pytest` check is green.
- The docstring claim in `check_negative_reciprocity.py` is correctly retired and now names the new test, and there is no surviving "nothing validates this" claim anywhere else.

One optional, non-blocking note (do not re-push just for this): the comment at `test_unit_test_corpus.py:31` says "396 files measured 2026-08-21," but the corpus at head is now 400. It is a floor comment and the assertion is `> 300`, so nothing depends on the exact number; bump it to 400 next time you are in the file, or leave it. Separately, `docs/architecture.md` §9.3's "396 committed test definitions" is stale at 400 too, but that is pre-existing (from #1775) and the issue scoped §9.3 out, so leaving it untouched was right.

Still needs one more code-owner approval to reach two; mine satisfies the senior-code-owner requirement but not the count on its own.

johnmarkpeterbrown (commented):
Let me know!

EdmondOware (changes requested):
The fire state proof on this one is the best I have seen on a lint in this repo. Four modes planted and deleted, and you led with the right one: unparseable JSON is exactly the mode both existing readers skip silently, so it is the case a reader most needs to see named. The floor guard is the other half of that, and it is the part most authors leave out. Calling `load_test` rather than vendoring a validator is right for the reason you give, and the rejected alternative is written down well enough that nobody will reopen it.

One change before this goes in.

**1. Two figures in `docs/architecture.md` section 9.3 are off by one.** `make eval-inventory`, the tool the paragraph names, prints 401 unit test definitions on head `bc717ba7`.

```
400 committed test definitions  ->  401
86 of the 400 definitions       ->  86 of the 401 definitions
399 rows, 360 passing (90%)     ->  400 rows, 360 passing (90%)
```

`360 passing`, `90%` and `86 negative` are all still correct. Only the two totals moved.

The cause is benign. You derived these in `792256b8` when the count really was 400. The branch then merged main and picked up `plan-item-stays-open-pending-capture.json` from #1799, which made it 401. The paragraph tells readers to derive these rather than quote them, so it is worth having it hold.

**Derive them again in the same minute you merge, not now.** These have already gone stale twice inside this PR, and merging main again before you land will move them a third time.

Two things not to change.

Your `ut_timeline_010` sentence is exactly right, down to the detail: 401 defined, 400 carrying a row, `ut_timeline_010` the only test defined without one, and nothing surviving for a deleted test. Leave it as written.

The `400 files measured 2026-08-21` comment above `MIN_EXPECTED_FILES` is a dated measurement justifying a floor of 300. It was accurate on that date and the floor is unaffected, so it stays.

aghadiayeamayanvboernest:
All fixed in `8dd2cccb`, derived in the same pass as the merge as you asked.

404 definitions, 87 negative, 403 rows / 362 passing (90%). Your 401 / 400 were correct at `bc717ba7` and were already stale by the time I merged — main had moved again — which is the effect you predicted. A third main merge then landed on the branch while I was doing this pass; I re-derived after it and the figures held.

One correction to your list: the negative count moved too, 86 → 87.

One you didn't flag, and I nearly shipped it as a guess. The snapshot date range in that same sentence was wrong at both ends — measured across the latest run log per suite it is 2026-07-27 to 2026-08-24. The start was stale from before this PR; the end was not the date I assumed.

Left as instructed: the `ut_timeline_010` sentence, re-verified on this tree (404 defined, 403 carrying a row, that one test the only definition without one, nothing surviving for a deleted test), and the dated `400 files measured 2026-08-21` comment above the floor.

Re-proved the check still fails after 77 commits of main landed under it. `make test-all` — all checks passed.

EdmondOware (approved):
Approving. The one item I raised is closed, and closed further than I asked.

I named two totals. You derived the whole paragraph again, date range included, and every
figure checks out against the branch as it stands:

| claim | verified |
|---|---|
| 404 committed test definitions | 404 JSON files under `eval/tests/unit/` |
| 87 negative | 87 files carrying `"type": "negative"` |
| 25 live suites | 25 |
| 403 rows, 362 passing | 403 and 362, which is 90% |
| snapshots 2026-07-27 to 2026-08-24 | oldest and newest of the latest run log per suite |

I counted these off the branch rather than reading the paragraph, and took the newest run log
per suite the way the gate resolves them.

The timing worked. You derived at 08:16 and the branch is currently level with `main`, zero
behind, so the figures describe the tree as it is rather than an earlier state. That was the
whole point of deriving at the merge, and it held.

I also went back through the check itself rather than resting on my earlier pass, and the
claim I most wanted to confirm holds: `loader.load_test` wraps `json.JSONDecodeError` into
`InvalidTestError`, so the single `except` really does catch unparseable JSON and schema
violations alike, which is what makes the headline mode reachable. The floor guard, collecting
every failure instead of raising on the first, and `rglob` matching the harness's own
discovery are all the right calls.

One standing caveat, not a change. These figures go stale the next time anything lands on
`main` that adds a test, which here is roughly daily. The paragraph tells readers to derive
rather than quote, so small drift is the expected condition and not a defect. Worth merging
promptly all the same.

Nothing else outstanding from me.

Issue #1270, opened 2026-08-04
locality.pages_read[].section is a closed enum the validator never checks


DallanQ:
> **Reviewed 2026-09-15 before junior handoff.** No decision needed — both "things to
> decide" in the original body are already answered in the repo (the name, and the heal
> path), and the citation below was stale. The superseded deferral note has been replaced;
> the first three paragraphs are the original text.

`research.schema.json` declares `locality.pages_read[].section` as a closed enum:

```
['home', 'getting_started', 'online_records', 'research_tips']
```

Nothing in `validator.ts` enforces it — `VALIDATOR_ENUMS` has no entry — and `src/tools/wiki-place-page.ts:179` re-types the four values by hand.

So a `research.json` carrying `section: "typo"` validates against the engine's own validator today, and the tool schema's copy can drift from the schema with nothing to catch it.

**Touches:** docs/specs/schemas/enums.schema.json, docs/specs/schemas/research.schema.json, packages/schema/schemas/enums.schema.json, packages/schema/schemas/research.schema.json, packages/schema/src/index.ts, packages/engine/mcp-server/src/validation/validator.ts, packages/engine/mcp-server/src/tools/wiki-place-page.ts, packages/engine/mcp-server/tests/packaging/enum-drift.test.ts, packages/engine/mcp-server/tests/validation/validator.test.ts, packages/engine/mcp-server/tests/tools/wiki-place-page.test.ts, docs/specs/research-schema-spec.md, docs/adrs/ADR-0008-sync-schema-copies-eliminate-generate-or-lint.md

## Both "things to decide" are already decided — do not re-open them

1. **The name is `locality_page_section`.** It is already the name in
   `docs/specs/research-schema-spec.md:112` and `:774` and in issue #1015's body. There is
   no `INLINE_ENUMS` map to follow — PR #1810 (issue #1015) deleted that symbol;
   `grep -rn INLINE_ENUMS` returns nothing. The two lists in `enum-drift.test.ts` today are
   `INLINE_NOT_ENFORCED` (line 47) and `VALIDATOR_ONLY` (line 403).
2. **No heal path is needed, and no fixture backfill.** Every `pages_read[].section` value
   in the repo and the eval corpus is in-set — 48 across 100 `research.json` files, 952
   across 936 run logs, **0 out of set** (measured 2026-09-15; reproduce with the two
   walks at the bottom). Independently, a pre-existing bad value in a real user project
[...]

PR #1810, opened 2026-08-21
schema: move five inline closed enums into enums.schema.json; guard the pattern copies (#1015)

### LINKED ISSUE #1015: Enum registry policy: where a closed set is defined, and how prose references it (unguarded copies + single-value mentions)
> **Reviewed 2026-08-20 before junior handoff.** No decision needed — the
> 2026-08-10 ruling still holds and its site list re-verifies clean. Two prose
> sites are missing from it and one line in it is stale; both are below. The
> original body follows under "Original issue".

## Two sites missing from the list, and one line in it that is stale

**Site 12 — `docs/architecture.md`, the "Add a value to a closed enum" block.** It names
the same five unions as an exception and tells the reader to edit `src/index.ts`. After
this PR that instruction produces exactly the sixth copy ADR-0008 exists to prevent, and
nothing in CI reads this file's prose. Replace the sentence starting "The exception is
the **five unions defined inline in `research.schema.json`**" through "…typing it by hand
creates a sixth copy." with:

> Every closed enum in `enums.schema.json` is generated, with no exceptions.
> Regeneration is automatic and typing a union by hand creates a sixth copy —
> `gen-enums.mjs` throws rather than let a hand-written union silently shadow a
> generated one.

Then drop this issue's number from the `**Direction**` callout just below it.

**Site 13 — the "Where these live in the machine-readable schemas" paragraph in
`docs/specs/research-schema-spec.md`,** under the enum table. It says the last **six**
rows are defined inline; after this PR only `locality_page_section` is. Rewrite it to say
that, and rename the table's `subscriptions` row to `subscription` to match the `$def`
name the validator already uses. Nothing checks this table — linting it is issue #1268,
out of scope here; name that issue in the PR rather than building it.

**Site 11 is stale — read ADR-0008 before editing it.** It says to "drop the five inline
enums from what it says is not caught." ADR-0008 was rewritten 2026-08-05 and its "What
this does **not** catch" paragraph no longer names them; the only inline enum it names is
`locality.pages_read[].section`, which **must stay** — issue #1270 is open and that
sentence is its register entry. So site 11 is: add the scope boundary (`research.json`
and `tree.gedcomx.json` only), bump `**Last updated:**`, **delete nothing** from that
paragraph, and delete no row from `Alternatives considered`. Its
`packages/schema/src/index.ts` citation is pinned in `GRANDFATHERED_LINE_CITES` in
`doc-links.test.ts` — converting it to a symbol reference means deleting that entry too,
or the drift check fails.

## Write the two pattern checks so they can fail

The map named in the ruling — `validator.test.ts`'s field allow-list guard — is keyed by
**shape** name (`known_holding`, `question`, `plan`). `ID_PREFIXES` is keyed by
**section** name (`known_holdings`, `questions`, `plans`). Different key sets, so the
id-prefix check needs its own section→`$def` map.

**Derive one side; do not hand-write both.** The two sets asserted equal are
`Object.keys(ID_PREFIXES)` and the set of `$defs` in `research.schema.json` carrying a
`properties.id.pattern`, **computed by walking the schema** — 15 each today. If both
sides are hand-written maps the assertion is vacuous, and a new section with an id prefix
and no validator check passes, which is the one thing the check exists to catch.

On the `ISO_DATE_PATTERN` comment: it enumerates eight fields and `checkIsoDate` is
called at ten sites. Delete the enumeration rather than repair it, as decided.

Break each check by hand and paste the failure into the PR body: change one
`ID_PREFIXES` value (`questions: "q_"` → `"qq_"`); add a `$def` with an `id` pattern and
no `ID_PREFIXES` entry; change one character of `ISO_DATE_PATTERN`.

## Acceptance

`make test-all`. The code half is fully covered — `make engine-test`, `make harness-test`
(the byte-identical mirror), `make typecheck`, `make test-js`.

**The prose half is covered by nothing.** Both greps must come back empty, and both
belong in the PR body:

```sh
grep -rn "EvaluationFocus\|EvaluationTargetType\|EvaluationVerdict\|ExperienceLevel" \
  CLAUDE.md docs/ packages/schema/src/index.ts
grep -rn "inline in \`research.schema.json\`" CLAUDE.md docs/
```

`eval/app/components/scenario/lib/schema.ts` keeps its own `ExperienceLevel`; that fork
is reached by no tier of ADR-0008 and is out of scope. Exclude it; do not edit it.

**In `CLAUDE.md`, delete only the exception sentence** — "The one exception is the five
unions…" through "…until they move." — and keep the "**Do not hand-edit the TS union**"
rule it hangs off.

Two mechanical notes for the schema edit: name the `$def` `subscription` (singular — that
is the `CLOSED_ENUMS` key, and `enum-drift.test.ts` fails loudly on any other spelling),
and give each new `$def` a `description`, since `gen-enums.mjs` emits it as the union's
JSDoc. `subscriptions` becomes the first cross-file `$ref` inside an array `items` in
this repo; `make harness-test` is what proves the Python resolver handles it.

**No skill body, agent body, rubric or eval test is touched**, so no `make eval-skill`
run and no annotation are needed.

## Sequencing

- **PR #1768** deletes line 1 of `packages/schema/scripts/gen-enums.mjs`; this PR deletes
  lines 10–11 of the same header. Land after it or expect a two-line conflict.
- `docs/architecture.md` is touched by open PRs #1718, #1774, #1775 and #1781; `CLAUDE.md`
  by open PR #1782. Rebase before pushing.
- Issue #1270 stays open and out of scope: `locality.pages_read[].section` keeps its
  inline `enum`, its `INLINE_NOT_ENFORCED` entry, and its sentence in ADR-0008.

## One count in the original body does not reproduce

"Across all 16 commits that have touched `enums.schema.json`" returns 9 plain, or 12 as
the union of both `--follow` histories (measured 2026-08-20). What survives is the claim,
not the number: no closed-enum value has ever been removed or renamed here; both changes
to the value sets were additions, which the existing full-list lint already covers.
Recheck with:

```sh
{ git log --format=%H --follow -- docs/specs/schemas/enums.schema.json; \
  git log --format=%H --follow -- packages/schema/schemas/enums.schema.json; } | sort -u
```

## Original issue

**Touches:** docs/specs/schemas/enums.schema.json, docs/specs/schemas/research.schema.json, packages/schema/schemas/enums.schema.json, packages/schema/schemas/research.schema.json, packages/schema/src/index.ts, packages/schema/scripts/gen-enums.mjs, packages/engine/mcp-server/tests/packaging/enum-drift.test.ts, packages/engine/mcp-server/tests/validation/validator.test.ts, packages/engine/mcp-server/src/validation/validator.ts, CLAUDE.md, docs/specs/research-schema-spec.md, docs/adrs/ADR-0008-sync-schema-copies-eliminate-generate-or-lint.md

**Decided 2026-08-10 — ready to implement.** All three questions this issue asked
are answered in the decision comment below, which carries the evidence, the
per-file site list, and what to break by hand before committing. This body says
what the work is.

## The work

1. **Move five inline closed enums into `enums.schema.json` and `$ref` them from
   `research.schema.json`:** `experience_level`, `subscription`,
   `evaluation_focus`, `evaluation_target_type`, `evaluation_verdict`. Their
   hand-written TS unions in `packages/schema/src/index.ts` get deleted and come
   from the generator instead; `INLINE_ENUMS` in `enum-drift.test.ts` gets
   deleted; the paragraph in `CLAUDE.md` naming those five as an exception goes
   away. `validator.ts`'s `CLOSED_ENUMS` needs no value change.

   **Five, not six.** `locality.pages_read[].section` stays inline — promoting it
   trips the "every closed enum in the schema is enforced by the validator"
   assertion, which forces the writer-tool rejection change that issue #1270
   owns. It stays in `INLINE_NOT_ENFORCED`.

2. **Write down the policy: anything JSON Schema can express belongs in the
   schema.** The validator is a hand-written copy that exists for LLM-actionable
   error text, not a second source of truth. Validator-only is reserved for what
   JSON Schema cannot express — cross-reference integrity between sections and
   cross-field conditional rules. `VALIDATOR_ONLY` stays empty and that is the
   enforcement; its comment changes from holding-pattern to policy. Scope the
   policy to `research.json` and `tree.gedcomx.json` in ADR-0008; the eval-harness
   schemas are out.

3. **Fold in the pattern copies.** `ID_PREFIXES` and `ISO_DATE_PATTERN` in
   `validator.ts` are the same unguarded-copy shape one tier down. Add a check for
   each against the schema. Do **not** add a check for *which sites* the validator
   applies the date rule to — see the decision comment for why.

Both new checks pass on arrival. Break each one by hand and watch it fail before
committing.

## Why the original framing was wrong

Two premises this issue was filed on do not hold, and correcting them is most of
what settled it:

- **The four "genuinely code-only" enums are not code-only.** All five are defined
  inline in `research.schema.json` — the `evaluation_*` three since the gps-mentor
  spec landed in PR #241. There are zero enums the validator enforces that no
  schema defines; `VALIDATOR_ONLY` has been empty since PR #1271.
- **`enums.schema.json` is not a shared-only file.** 23 of its 25 closed enums are
  referenced by exactly one schema, and `relationship_type` by none — it exists
  purely as the canonical entry the validator's copy is linted against. It is
  already the registry of every closed enum, which is why the five inline ones
  should join it.

The sweep this issue asked for is done: six inline `enum` arrays in
`research.schema.json` (the five above plus `locality.pages_read[].section`), and
14 more across the eval-harness schemas, which are out of scope.

## Absorbed issue #1013 — decided, no lint

Issue #1013 asked for a lint on **single-value** enum mentions in prose
(``evidence_type: indirect``), a failure mode the existing full-value-list lint
does not cover: a list goes stale when a value is *added*, a single mention when
one is *renamed or removed*.

**Decision: do not build it.** Across all 16 commits that have touched
`enums.schema.json`, a value has been removed or renamed **zero** times; the only
two changes ever were additions, which the existing lint already covers. A naive
scan today finds 38 mentions, 18 failing, and 16 of those 18 are a single name
collision with the check-warnings tool's `severity` — clearing it would mean
renaming a tool output field the model reads, for a lint's benefit.

What ships instead is one line in the schema-change rules: **removing or renaming
a closed-enum value requires a repo-wide grep for the old value.** The alternative
that was weighed and not taken (a retired-values list plus a lint on dead values)
is recorded in the decision comment; revisit it if a first removal ever happens.

## The rest of the cluster

- **issue #1165** stands as written — its enum half closes by construction here,
  but its subject is interface fields and their types.
- **issue #1014** stands unchanged — the backticked-token anchor applies to the
  existing full-list matcher either way.
- **issue #1270** unchanged, and explicitly out of scope.

clack391:
## Summary

- Moves the five closed enums that were defined **inline** in `research.schema.json` (`experience_level`, `subscription`, `evaluation_focus`, `evaluation_target_type`, `evaluation_verdict`) into `enums.schema.json` and `$ref`s them, so there is one definition per closed set. Their hand-written TS unions in `packages/schema/src/index.ts` are deleted and now come from the generator; `INLINE_ENUMS` in `enum-drift.test.ts` is deleted; the `CLAUDE.md` / `architecture.md` prose that named those five as an exception goes away.
- Guards the two remaining unguarded copies one tier down: `ID_PREFIXES` and `ISO_DATE_PATTERN` in `validator.ts` now have drift checks against the schema (`validator.test.ts`). Deletes the stale field enumeration in the `ISO_DATE_PATTERN` comment.
- Writes the policy down in ADR-0008 (anything JSON Schema can express belongs in the schema; scope is `research.json` + `tree.gedcomx.json` only) and adds one line to the schema-change rules: removing or renaming a closed-enum value needs a repo-wide grep for the old value. No new single-value prose lint is built: absorbed #1013 is decided against, and its reasoning is now recorded as two rows in ADR-0008's `Alternatives considered` rather than only in the issue comment.
- Tier: **Normal**. This is a schema-file change, but a schema *reorganization* with no value changes, decided by the lead on 2026-08-10 (issue #1015 decision comment) and task-reviewer-vetted, so it is not an undecided schema change.

## Review guidance

**Start here:** the two new drift checks in `validator.test.ts`. The ID_PREFIXES check derives one side by walking `research.schema.json` for every `$def` carrying `properties.id.pattern` (15 today) and asserts that set equals the values of a hand-written section to `$def` translation map; only the translation is hand-written, because `ID_PREFIXES` is keyed by section name (`questions`) while the `$def` is `question`, and `plan_items` nests under `plan`. Check that a new id-patterned section cannot get a prefix with no validator check. Break-test failures are pasted below.

**Unsure about:** nothing blocking. One acceptance-wording gap is called out under Acceptance check below (the issue's grep 1 cannot literally come back empty over `index.ts`).

## Plan

**Didn't change:** `locality.pages_read[].section` stays inline (promoting it changes what the writer tools reject, which is issue #1270's PR); `CLOSED_ENUMS` in `validator.ts` needs no value change (it already keyed all five under the target names); the `eval/app` fork of `ExperienceLevel` is out of scope; the ADR's "What this does not catch" paragraph is left intact apart from the stale "sixth inline enum" wording (the `locality.pages_read[].section` sentence is issue #1270's register entry). Its `Alternatives considered` table gains two rows and nothing else.

**Acceptance check:** `make test-all`. Concretely, the two new tests in `validator.test.ts` fail on a mutated tree and pass here (break-tests below); `enum-drift.test.ts` diffs the five as ordinary closed enums; the byte-identical-mirror and cross-file `$ref` resolution are covered by `make harness-test`.

Both acceptance greps belong here:

```
$ grep -rn "EvaluationFocus\|EvaluationTargetType\|EvaluationVerdict\|ExperienceLevel" CLAUDE.md docs/
(empty)
$ grep -rn "inline in \`research.schema.json\`" CLAUDE.md docs/
(empty)
```

The first grep in the issue also lists `packages/schema/src/index.ts` and asks for empty. That cannot be literally empty and should not be: `index.ts` now imports the five generated types and its interfaces reference them (`experience_level?: ExperienceLevel`, `verdict: EvaluationVerdict`, etc.), because `export *` does not bind names locally. Those hits are the generated types in use, not leftover hand-written unions (there are zero of those). This mirrors how the issue itself excludes `Subscription` from the grep for the same reason. Reported as a split rather than forced empty.

**Deviated from the plan:** one required edit site is not on the issue's Touches list: `packages/engine/mcp-server/tests/packaging/doc-links.test.ts`. Removing the `#1015` reference from `architecture.md` drops its counted issue-refs 20 to 19, and the ratchet fails until `ISSUE_REF_BASELINE` comes down with it. Flagged here because it crosses the issue's stated file list.

Second deviation, in the tests: the decision comment says to "reuse the section-name-to-`$def` map already in `validator.test.ts`'s field allow-list guard". There is no such map. The existing `defFor` (`validator.test.ts:2043`) is keyed by **`$def` name** (`question`, `plan`, `known_holding`), not by the section names `ID_PREFIXES` uses (`questions`, `plan_items`), so it cannot be reused for this check. Hence the new `SECTION_TO_DEF` at `validator.test.ts:2101`, which is the only hand-written half and is itself pinned in both directions by the two key-set equality assertions below it. Noting it so a reviewer does not go looking for the map the instruction names.

## Test plan

- [x] I ran the offline suites and they pass: engine vitest (2466 passed, 104 files), harness unit (2460 passed, 2 skipped), control-plane `test_mock_agent_schema.py` (1), turbo typecheck (5 tasks), viewer-ui + schema tests (215, including `schema-interface-drift`, `gen-enums-guards`, `dev-scripts-generate`). `make harness-test` is what proves the Python `referencing` resolver handles `subscriptions` becoming the first cross-file `$ref` inside an array `items` in this repo; I also validated a live doc through it and confirmed a bad `subscription` and a bad `verdict` are rejected (the `$ref` resolves, it does not silently pass).
- No skill, agent, rubric, or unit-test snapshot changed, so no `make eval-skill` run and no annotation.

**Break-tests (each fails as designed, then restored):**

```
# ID_PREFIXES.questions "q_" -> "qq_"
  x each schema id pattern starts with ^ + the ID_PREFIXES prefix
  AssertionError: questions: schema pattern ^q_ does not start with ^qq_

# add a $def with an id pattern and no ID_PREFIXES/map entry ("widget")
  x every id-patterned $def is mapped (a new section can't dodge the check)
  AssertionError: expected Set{ 'widget', 'project', ... } to deeply equal Set{ 'project', ... }

# flip one char of ISO_DATE_PATTERN (\d{2}$ -> \d{3}$)
  x its source equals enums.schema.json#/$defs/iso_date pattern
  AssertionError: expected '^\d{4}-\d{2}-\d{3}$' to be '^\d{4}-\d{2}-\d{2}$'
```

## Follow-on issues

**Folded in / filed instead:** none filed. Two open issues are named per the decision and stay out of scope: **#1268** (lint the enum prose table in `research-schema-spec.md`, which nothing checks) and **#1270** (`locality.pages_read[].section` keeps its inline `enum` until its own writer-rejection PR). Absorbed **#1013** (single-value prose lint) is decided against, recorded in two places rather than one: the operative rule is the schema-change-rules line in `CLAUDE.md` and `docs/specs/research-schema-spec.md`, and the reasoning it rests on (zero value removals across the 16 commits that have touched `enums.schema.json`; 38 single-value mentions with 16 of the 18 failures being the single `severity` collision; the retired-values-list alternative and why it buys no independent coverage) is now two rows in ADR-0008's `Alternatives considered`. Previously that reasoning existed only in a comment on the issue this PR closes.

**Merge ordering:** open **PR #1768** edits the same `gen-enums.mjs` header region this PR trims; land after it or expect a two-line conflict. Rebase before merge; `docs/architecture.md` and `CLAUDE.md` are touched by other open PRs.

Closes #1015.


mercyokum (approved):

promise-emmanuel (approved):
Checked this against the #1015 decision comment — all 11 site-list items are
there, and the one extra file (doc-links.test.ts, ratchet 20→19) is declared.
Ran the suites in a worktree at c7aa40b1: engine vitest 2466, tsc clean, turbo
typecheck 5/5, viewer-ui/web/schema 215+10, harness pytest 2460, server pytest
209. Two failures I chased down and neither is yours — the harness CLI test
needs `npm run build` in mcp-server first, and electron's 45 MB feedback-zip
test times out under parallel turbo but passes in 3.4s alone.

I broke all five guards plus three cases you didn't list, and every one fires:
re-inlining an enum, an unmapped id-patterned $def, an ID_PREFIXES entry with
no $def, and a mirror-only edit. I also drove validate_research_json directly to
confirm the five new cross-file $refs actually reject bad values rather than
resolving to nothing — `subscriptions` matters most there since it's the repo's
first cross-file $ref inside array `items`, and it rejects correctly.

One thing to change before merge. The body says #1013 is "decided against in the
ADR and the schema-change rules" — the schema-change half landed, but the ADR has
nothing about it. Grep it for `single-value`, `retired`, `prose mention` and you
get zero hits. So the reasoning (zero value removals in 16 commits, 38 mentions
with 16 being the one `severity` collision, and the retired-values-list
alternative that's "worth revisiting if a first removal ever happens") lives only
in a comment on the issue this PR closes. That's the thing the ratchet you edited
exists to prevent. Add two rows to ADR-0008's Alternatives considered table — one
for the prose lint, one for the retired-values list — and it's fine.

Two notes, no action needed. First, the instruction to reuse "the
section-name-to-$def map already in validator.test.ts" wasn't followable — the
existing `defFor` is keyed by $def name, not section name. Your new SECTION_TO_DEF
is right; worth a line in the body saying so, or a reviewer goes looking for it.

Issue #1811, opened 2026-08-21
record_search: rankingSkipped shows no measured late-segment subjectId lift


taiwo-stack:
**Touches:** `packages/engine/plugin/skills/search-records/SKILL.md`, `packages/engine/mcp-server/src/tools/record-search.ts`

> **Blocked on issue #1155, and the table below is not yet a result. Checked 2026-08-21.**
>
> The figures are the **scratch pass** recorded in issue #1155's own body, not the output of
> `make e2e-compaction` — that target does not exist. It is in no `Makefile`, nowhere in the
> tree, and no open PR builds it (`grep -n e2e-compaction Makefile`, `git grep e2e_compaction`,
> both empty against `3c4aa284`). Issue #1155 is in Ready and unstarted.
>
> Issue #1155 says of these exact numbers: *"That is a smell to reproduce, **not a result to
> quote**; derive your own numbers and state the command that produced them."* It also flags
> that 13 of the 25 late-segment calls come from a single run (`victoriano-macatangay-parents`,
> at 0%), which is the reason it wants them re-derived.
>
> **Do not take the decision below off this table.** Re-derive it once issue #1155 lands
> `make e2e-compaction`, then decide.

Issue #1155's scratch pass on 2026-08-20 compared two disjoint windows around the `rankingSkipped` note's landing (2026-08-04), which was added specifically to survive the compaction that erodes the `subjectId` prose instruction in `search-records/SKILL.md`:

| Window | Early (segments 0-2) | Late (segments 3+) |
|---|---|---|
| 2026-07-27..08-03 (pre-nudge, 34 segmentable runs / 101 late-segment calls) | 45.8% | 43.6% |
| 2026-08-04+ (post-nudge, 14 runs / 25 late-segment calls) | 64.7% | 44.0% |

Early-segment `subjectId` supply rose 19 points after the nudge shipped; late-segment supply moved 0.4 points — no measured effect in the segment the nudge exists to help. The early-segment rise is more plausibly explained by `search-records/SKILL.md` still being resident, or by PR #1032's unrelated marriage-search change landing the same week, than by the nudge itself.

Per issue #1155's own "Do not do these" list, this number is exactly what would justify a follow-up: "If the number argues for anchoring subjectId structurally, that is a separate issue written after this one produces the number." `search-records/SKILL.md` was explicitly not to be edited off the back of #1155 itself.

[...]

PR #1812, opened 2026-08-21
eval-harness: build make e2e-compaction, re-test the record_search fold under compaction (#1155)

### LINKED ISSUE #1155: Re-test the record_search ranking fold under heavy compaction pressure
> **Reviewed 2026-08-20 before junior handoff.** The stated blocker is cleared, but
> so is the ask — the thing this proposed to test is now a deterministic `if`
> already pinned by a unit test. Replaced below; the original follows under
> "Original issue".

**Touches:** eval/harness/e2e/, Makefile, docs/architecture.md, docs/e2e-testing-guide.md, docs/adrs/ADR-0003-anchor-cross-turn-rules-structurally.md, docs/specs/record-search-tool-spec-v2.md

## Three things this issue rested on are no longer true

1. **The fold's firing is already pinned, and already confirmed on the corpus.**
   `packages/engine/mcp-server/tests/tools/record-search.test.ts` asserts a `subjectId`
   search requests the deep pool and returns `ranked`, and that `count` stays at 20
   without one. ADR-0003 records the corpus check (21 of 22 eligible searches came back
   ranked). Ranking-when-`subjectId`-is-supplied is an unconditional `if` in
   `record-search.ts` — it has no model decision in it, so **no live run can disconfirm
   it.**

2. **A live re-run cannot produce "heavy compaction pressure."** `make e2e-run` exposes
   `EFFORT_LEVEL`, `MAX_OUTPUT_TOKENS`, `AGENT_MODEL`, `PERSON_EVIDENCE_GUARD` and
   `RESUME_ON_STALL` — nothing that touches the context window or forces compaction. The
   five committed `hannah-earnest-children` runs compacted 3, 3, 3, 6 and 4 times; the
   "302-turn baseline" is the 6-compaction one, not 23. **Run length is an outcome, not a
   lever.**

3. **The cost was wrong, and the measurement does not need paying for.** The 302-turn
   baseline cost **$20.03** and stopped on the cost cap — not the "~$7" quoted when this
   was promoted. And the analysis is offline: `usage.timeline` carries
   `system:compact_boundary` entries, and since 2026-07-26 every `assistant` entry carries
   its tool names, which align 1:1 and in order with `tool_calls`. **Every run made after
   the fold landed is segmentable today, for free.**

   Reproduce the cost with
   `jq '[.usage.num_turns,.usage.total_cost_usd,([.usage.timeline[]|select(.[1]=="system:compact_boundary")]|length)]' eval/runlogs/e2e/hannah-earnest-children/run-2026-07-22_19-18-50.json`.

## The question the fold actually left open

The fold removed the unanchored rule "always call `rank_search_matches`" and replaced it
with an unanchored rule one step upstream: **pass `subjectId`**. That instruction lives
only in `search-records/SKILL.md` prose — a body that has grown to 55,427 bytes since the
2026-07-27 audit measured it at 41.6 KB, so eviction pressure is **higher**, not lower.
Whether `subjectId` supply survives compaction is unmeasured, and it is the whole value of
the fold.

## The task

Measure `subjectId` supply on `record_search` **by compaction segment**, over the
committed e2e corpus. **No live run.**

Split each run at its compaction boundaries and compare segments 0–2 against segment 3+.
Walk `usage.timeline`; a `system:compact_boundary` entry opens a new segment; the names on
`assistant` entries consume `tool_calls` in order, so each `record_search` call lands in a
segment with its `args.subjectId` in hand. Runs whose timeline entries have only two
elements (everything before 2026-07-27) cannot be segmented — **exclude them and report
how many you excluded.**

Report **two windows**, because the population changed under this issue:

- every run from 2026-07-27 (the fold) onward;
- every run from **2026-08-04** onward only — the day PR #1032 raised supply and the day
  the `rankingSkipped` note landed, whose stated purpose is to survive the compaction that
  erodes prose. **Whether that nudge lifts late-segment supply is the sharpest thing this
  measurement can answer.**

A scratch pass on 2026-08-20 gave 65% early against 44% late for the second window — on 25
late-segment calls, 13 of them from one run (`victoriano-macatangay-parents`, at 0%). That
is a smell to reproduce, **not a result to quote**; derive your own numbers and state the
command that produced them. Corpus-wide supply has also moved: 44 of 92 calls (48%) on
runs captured after 2026-08-12, against the 83% measured when this was promoted.

## Decision 2026-08-20 — build it as a report target, not a script

Land this as **`make e2e-compaction`** (`TEST=<slug>` / `SINCE=all|N|YYYY-MM-DD`) in
`eval/harness/e2e/`, next to the other committed-runlog report modules, with a unit test
under `eval/harness/tests/unit/`. Report per-run and aggregate `record_search` counts and
`subjectId` supply, split at compaction boundaries.

Chosen over a one-off script because **issue #1144 needs the same segmentation** for the
four guardrail skills' reference reads, and ADR-0003 asks for a second skill body's audit —
a script would be rebuilt twice. Record that reason in `docs/e2e-testing-guide.md` beside
the target, not here.

**Land after PR #1775**, which is editing `Makefile`, `docs/architecture.md` and
`eval/harness/e2e/corpus_report.py`, and adds `eval/harness/e2e/inventory.py`.

## Do not do these

- **Do not run `make e2e-run`.** One run adds roughly 5 late-segment `record_search` calls;
  the committed corpus already holds far more, and the run cannot be made to compact harder.
- **Do not report "N of N eligible searches were ranked" as the finding.** The tool ranks
  unconditionally when `projectPath` and `subjectId` are both present, so that number is a
  property of the code and is already covered by `record-search.test.ts`.
- **Do not edit `search-records/SKILL.md`** off the back of this. ADR-0003's Alternatives
  table records "reinforce the prose" as the refuted answer, and a skill-body edit flips
  that skill's run log inactive (~$8–12, 45–65 minutes, plus a genealogist annotation). If
  the number argues for anchoring `subjectId` structurally, that is a separate issue
  written **after** this one produces the number.

## Sites

- [ ] `eval/harness/e2e/` — the report module, plus a unit test
- [ ] `Makefile` and the command table in `docs/architecture.md` — the new target
- [ ] `docs/e2e-testing-guide.md` — mechanics and why this is a target rather than a script
- [ ] `docs/adrs/ADR-0003-anchor-cross-turn-rules-structurally.md` — its "Revisit when"
      line asks for exactly a compaction-segment audit, and its Consequences paragraph
      currently cites only the ranked-block check
- [ ] the `rankingSkipped` section of `docs/specs/record-search-tool-spec-v2.md`, which
      already says whether an omission was legitimate "is answered at analysis time from
      the args already in the run log" — this is that analysis
- [ ] comment the number on issue #1144, which wants the identical method

## What nothing checks

This measures the eval corpus, not production. There is no production telemetry, so the
result answers "did this decay in our runs", never "is this getting better for a real
user". **Say so wherever the number lands.** Every figure you write down carries the
command that reproduces it.

## Verify

`make harness-test` for the code; the finding itself is checked by a reviewer re-running
the command you state and getting your number. No CI job reads compaction segments today.

## Original issue

The post-fold e2e run (`hannah-earnest-children`, 2026-07-27) compacted only 4
times in 190 turns and ranked 7 of 7 eligible searches. That confirms the contract
fires; it does not show the contract beating prose under the 23-compaction pressure
that motivated it.

The 302-turn baseline of the same fixture is the harder test. Earlier runs are
committed under `eval/runlogs/e2e/hannah-earnest-children/`.

**Blocked on #1073:** the fold is gated on `subjectId`, which is itself supplied on
0-100% of `record_search` calls depending on the run. Testing before that lands
measures parameter supply, not contract durability — and would burn a ~\$7 e2e run
on the wrong variable.

taiwo-stack:
## Issue

`record_search`'s ranking fold (`subjectId && projectPath` → host-side ranking) is pinned in code and can't decay. What was unmeasured is whether the agent keeps *supplying* `subjectId` deep into a long, heavily-compacted session — that instruction lives only as prose in `search-records/SKILL.md`.

## Solution

Adds `make e2e-compaction` (`eval/harness/e2e/compaction_report.py`), a pure-analysis report over the committed e2e run-log corpus — no live run, no API spend. Splits each run's `record_search` calls at `usage.timeline` compaction boundaries and reports `subjectId` supply for early segments (0–2) vs. late segments (3+).

Measured with a clean pre/post split around the `rankingSkipped` nudge (2026-08-04): early-segment supply rose 45.8% → 64.7%, late-segment supply moved 43.6% → 44.0% — a raw gap, and the nudge shows no lift in it.

That raw gap turned out to be mostly something else. The tool's own schema permits omitting `subjectId` when a search isn't about a specific tree person yet, and reading the two runs carrying most of the post-nudge late-segment sample call by call, most of their omissions are exactly that — searches for a not-yet-tree child or an unconfirmed parent, not a forgotten `subjectId` for an already-established subject. Only a couple are the latter. So `compaction_report.py` now prints an explicit caveat with every report, and `docs/architecture.md` §3.1 / ADR-0003's Consequences state the corrected finding instead of the flat "no lift" reading. A #1811 filed against the flat reading is corrected and closed by this same correction.

Numbers and method recorded in ADR-0003's Consequences, `docs/architecture.md` §3.1, `docs/e2e-testing-guide.md` (mechanics + why a report target, not a script), and a pointer in `record-search-tool-spec-v2.md`.


johnmarkpeterbrown (changes requested):
Solid work — the method is sound, exclusions are counted rather than swallowed, and the caveat is the right call. I read the 13 late-segment calls in `victoriano-macatangay-parents` myself and your "only a couple" is exactly right. Four things first.

## 1. The pre-nudge numbers match no command

`--since` is cumulative, so `SINCE=2026-07-27` contains the "after" window too. The two documented commands give 53.2% / 43.7% and 64.7% / 44.0%. The ADR's before row is 45.8% / 43.6% — correct, but only reachable by subtracting the second window from the first, which nothing says happened. Issue #1155 asked for the command behind each number; as written, a re-run looks like the ADR has gone stale.

`docs/e2e-testing-guide.md`:

```
You wrote:      or `SINCE=all`. The two windows issue #1155 was asked to compare:
                `SINCE=2026-07-27` (the day the ranking fold shipped) and `SINCE=2026-08-04`
                (the day `rankingSkipped` landed, meant to counteract exactly this decay). A

Change it to:   or `SINCE=all`. The two windows issue #1155 was asked to compare:
                `SINCE=2026-07-27` (the day the ranking fold shipped) and `SINCE=2026-08-04`
                (the day `rankingSkipped` landed, meant to counteract exactly this decay).
                Both are cumulative — the first *includes* the second, so the
                pre-nudge-only row in ADR-0003 is the first window's counts minus
                the second's, not a window this target can print. A
```

`docs/adrs/ADR-0003-anchor-cross-turn-rules-structurally.md`:

```
You wrote:      Comparing the two **disjoint** windows: early-segment supply rose 19 points

Change it to:   The pre-nudge row is `--since 2026-07-27` minus `--since 2026-08-04`
                (312 early / 101 late calls); the report prints cumulative windows only.
                Comparing the two disjoint windows: early-segment supply rose 19 points
```

An `--until` flag instead would be a few lines and makes it reproducible for the next reuse. Your call.

## 2. `--test` discards `SINCE` silently

The Makefile passes both and the guide advertises both, so `TEST=<slug> SINCE=2026-08-04` reads every run that fixture ever had and reports "Window: entire corpus".

`latency_report`'s convention doesn't carry over: its `--test` takes one latest run, so a date filter there is meaningless. Yours calls `result_jsons_for()`, which returns every run for the slug — an aggregate read, which is what the window exists to protect.

```
You wrote:      if args.test:
                    # An explicit --test is a deliberate request for that fixture's runs;
                    # the cutoff guards *aggregate* reads, same convention as
                    # latency_report.py's --test/--all split.
                    cutoff = None
                    paths = all_paths
                else:
                    cutoff = args.since
                    paths = filter_since(all_paths, cutoff)

Change it to:   # --test still aggregates EVERY run for the fixture, unlike
                # latency_report's --test (one latest run), so the window applies
                # here too. Pass SINCE=all for that fixture's whole history.
                cutoff = args.since
                paths = filter_since(all_paths, cutoff)
```

`test_main_test_flag_ignores_the_since_cutoff` then inverts: assert the old fixture is filtered out under the default window and returned under `--since all`. Please also print the fixture name, so a one-fixture read never claims "entire corpus".

## 3. One mismatch guard has no test

I deleted these two lines and all 14 tests stayed green:

```python
    if cursor != len(tool_calls):
        return [], "tool-count-mismatch"
```

The existing test covers only the opposite direction. This one catches `tool_calls` running past what the timeline accounted for — the direction where calls go unexamined. It excludes four real runs today, none holding a `record_search` call, so no printed number would move if it broke.

```
Add to test_e2e_compaction_report.py:

def test_more_tool_calls_than_timeline_names_is_excluded_not_undercounted():
    """The other mismatch direction: trailing tool_calls the timeline never
    accounted for must exclude the run, not go silently unexamined."""
    doc = {
        "usage": {"timeline": [_assistant(["record_search"])]},
        "tool_calls": [
            _tool_call("mcp__genealogy__record_search", subject_id="I1"),
            _tool_call("mcp__genealogy__record_search", subject_id=None),
        ],
    }
    calls, reason = segment_run(doc)
    assert calls == []
    assert reason == "tool-count-mismatch"
```

## 4. The new section makes the Windows line false

"Every `make` target above has a batch file in `eval\`" was true until this PR; the new section adds five targets that have none. The genealogists are on Windows and read that line.

```
You wrote:      Every `make` target above has a batch file in `eval\`. Double-click it or run

Change it to:   Every `make` target in the fixture-authoring steps above has a batch file in
                `eval\`. The corpus reports (`make e2e-corpus`, `e2e-nudge`, `e2e-latency`,
                `e2e-wiki-failures`, `e2e-compaction`) do not — run those from Git Bash or
                WSL. Double-click a batch file or run
```

Adding `scripts/windows/e2e-compaction.bat` works too — `judge-report.bat` is the same kind of report — but then do all five.

## For the lead, not for you to fix

Issue #1811 reserved the structural-anchoring call for the lead. You closed it with "no structural-anchoring decision is warranted on this number as measured", and this PR writes that into ADR-0003. The reasoning holds up, but the lead should know the call was made and there's no **Deviated from the plan** line to surface it. Worth tagging them before merge.

Items 1 and 2 are what block; 3 and 4 can ride along in the same push.

T-FEH (changes requested):
Reviewed on the merged head — I merged `main` (clean) and pushed. Verified every number by running the report and mutating the code, not by reading it.

**First, credit where it's due.** All four of @johnmarkpeterbrown's findings are genuinely fixed, and I checked the substance rather than the commit message. The cumulative-window provenance is now stated in both ADR-0003 (:113-116) and the guide (:524-525); `--test` applies the window (`:253`) and prints `Fixture:` (`:262`); the Windows line is scoped to the fixture-authoring steps. And finding 3's guard is really pinned now — deleting `if cursor != len(tool_calls)` reds `test_more_tool_calls_than_timeline_names_is_excluded_not_undercounted`.

**Every documented figure reproduces to the decimal.** `SINCE=2026-08-04` prints `EARLY … 58.1%` / `LATE … 44.8%`, 236 early / 29 late, 16 of 159 runs — exactly ADR-0003:118-119. Exclusion accounting is honest, the caveat is real (not a rewording), and `architecture.md:258` carries the corrected reading with no stale "no lift" surviving. This is a well-built report.

Four things block, and they're all about the *aggregation* rather than the arithmetic.

## 1. The operator carrying the published conclusion has no test

`compaction_report.py:165`. Changing `segment <= EARLY_MAX_SEGMENT` to `segment <` — one character — leaves **all 15 tests green** and reverses the finding:

```
baseline   EARLY 58.1%   LATE 44.8%     <- the ADR's conclusion
mutated    EARLY 53.0%   LATE 62.4%     <- late is HIGHER
```

`test_early_late_boundary_is_exactly_at_segment_three` looks like it covers this and doesn't: it asserts `EARLY_MAX_SEGMENT == 2` (tautological) and checks `segment_run`'s numbering. It never calls `_bucket` or `format_report`. The function turning segments into the two published percentages has zero coverage — the vacuous-test shape this repo has hit repeatedly, on the one number the module exists to produce. One test on `format_report` fixes it.

Related fragility worth a line: `early_range`/`late_range` (`:200-201`) are derived from `EARLY_MAX_SEGMENT`, but `_bucket` does its own comparison. Under the mutation the report still prints "segments 0-2" while actually bucketing 0-1 — the label and the logic can drift apart silently.

## 2. Supply is not monotone, so "throughout" doesn't hold

ADR-0003:112-113 says early supply is "markedly higher than late-segment (3+) **throughout**". Per segment, post-nudge:

```
seg 0  48/75  64.0%      seg 3  13/27  48.1%
seg 1  39/89  43.8%      seg 4   0/2    0.0%
seg 2  50/72  69.4%
```

Segment 1 (43.8%) sits *below* segment 3 (48.1%). The early>late gap exists only because segment 2 — the single highest-supply segment — lands inside "early", which is exactly why finding 1's mutation inverts. And on the other reading of "throughout" (both windows), the pre-nudge gap is 45.8% vs 43.6% — 2.2 points, not "markedly". Either way the sentence overstates. The per-segment row is 5 lines and the data is already in hand; printing it would have surfaced this.

## 3. The post-nudge late figure rests on one run of sixteen

```
 0/13  victoriano-macatangay-parents/run-2026-08-13_23-24-54
 6/ 6  heinrich-dewus-children-death/run-2026-08-06_07-25-30
 5/ 5  mary-mcandrew-son/run-2026-08-05_20-48-08
 2/ 4  hannah-earnest-children/run-2026-08-23_03-37-12
 0/ 1  antonio-lucas-spouse/run-2026-08-04_22-45-15
```

One run is 13 of 29 late calls at 0/13. Drop it and late goes **44.8% → 81.2%**, well above the 58.1% early figure. The report prints the per-run table, so the tool is honest — it's the ADR prose that loses the dispersion, and the ADR is the durable artifact. A run/call count without it reads as more solid than it is.

## 4. The published comparison is between-run; the claim is within-run

Only **5 of 15** runs with `record_search` calls have both early and late calls. Restricted to those, early is **79/94 = 84.0%** against late 44.8%. The published 58.1% is diluted by ten runs that never compact, so the two buckets are drawn from different populations. The direction survives — it strengthens — but the published effect size isn't the decay effect size. Either report the paired figure or say the headline is diluted.

## Minor

- ADR-0003:139 says the caveat prints "with every report". `format_report` early-returns on the no-calls path (`:185-192`) before appending it. The code's own comment (`:80`, "every non-empty report") is the accurate one.
- No `--until`, so `--since 2026-08-04` is open-ended and the published row drifts — it already moved 64.7% → 58.1% once when two runs landed. The ADR discloses the drift, but the honest content of that disclosure is that the row can't be re-derived. An `--until` bound, or recording corpus HEAD next to the run count, closes it.
- Unrelated: `test_mock_mcp.py::test_successful_call_carries_no_is_error` fails identically on `main` — inherited, not this PR.

## For the lead, not the author

Carrying @johnmarkpeterbrown's last point forward: #1811 reserved the structural-anchoring call for the lead, the author closed it, and this PR writes that into ADR-0003 with no **Deviated from the plan** line surfacing it. @DallanQ should see that the call was made.

Requesting changes on 1–4. The measurement is sound where it's arithmetic and overstated where it's prose — which is the half that outlives the PR.

T-FEH (changes requested):
Re-reviewed on the merged head (`7fc21c48` — your `17fba699` plus a clean merge of main). Harness suite green: **2689 passed, 2 skipped**. I re-ran every check rather than trusting the commit message.

## All four blockers are fixed, and fixed the right way

The best part is that findings 2–4 are now **printed by the tool**, not asserted in prose — so the aggregation choice that was doing unstated work is visible to anyone who runs it:

```
By segment:  seg0 64.0%  seg1 43.8%  seg2 69.4%  seg3 48.1%  seg4 0.0%
Per-run late-segment supply, worst first:
  0/13 (0.0%)  victoriano-macatangay-parents/run-2026-08-13_23-24-54
Paired (runs with both early- and late-segment calls, 5 of 15): EARLY 79/94 (84.0%), LATE 13/29 (44.8%)
```

Every figure matches what I computed independently, to the decimal. And **finding 1's guard is real** — I mutated the boundary in both directions:

```
segment <  EARLY_MAX_SEGMENT      -> 1 failed (test_segment_two_lands_in_early_not_late)
segment <= EARLY_MAX_SEGMENT + 1  -> 3 failed
```

The caveat minor is reconciled honestly ("every **non-empty** report", matching the code). No `--until`, but the date stamp plus the drift note is a fair substitute — and the pre-nudge subtraction is drift-proof anyway, since new runs land in both windows.

## One blocker left, and it is my fault

ADR-0003:117-118 now says:

> the entire early>late gap rests on segment 2 alone

That clause came from my last review, where I wrote "the early>late gap exists only because segment 2 … lands inside early". You transcribed it faithfully; it is wrong, and I should not have written it. The arithmetic:

```
EARLY 137/236 = 58.1%   LATE 13/29 = 44.8%   full gap 13.2 points
EARLY without seg 2: 87/164 = 53.0%          gap still 8.2 points
segment 2 accounts for 5.0 of 13.2 points = 38%, not all of it
```

Segment 2 is also not alone in exceeding the late segments: segment 0 (64.0%) is 15.9 points above segment 3 (48.1%). So the clause is false on both readings — the per-segment one and the aggregate one.

This is the exact failure mode the repo's "re-measured, not reworded" rule exists to stop, which is why I am flagging it rather than waving it through as my own mess.

```
You wrote:      Segment 1 (43.8%) sits *below* segment 3 (48.1%); the entire
                early>late gap rests on segment 2 alone — the single highest of
                the five — landing inside "early" under the issue's own 0–2/3+
                split, so review this per-segment row rather than the two-bucket
                aggregate before drawing a conclusion from it.

Change it to:   Segment 1 (43.8%) sits *below* segment 3 (48.1%), so supply is not
                monotone across segments, and segment 2 (69.4%) is the single
                highest of the five. The aggregate gap does not rest on segment 2
                alone, though: dropping it leaves early at 53.0% (87/164) against
                the same late 44.8% — an 8.2-point gap where the full one is 13.2.
                Review this per-segment row rather than the two-bucket aggregate
                before drawing a conclusion from it.
```

## Two minors that can ride along in the same push

**The caveat itself is unguarded.** I deleted `lines.append(CAVEAT)` and all 18 tests stayed green — while ADR-0003 leans on "prints this caveat with every non-empty report". One test closes it:

```python
def test_format_report_prints_the_caveat_on_every_non_empty_report():
    out = format_report([RecordSearchCall("f/r", 0, True)], n_runs=1, excluded={})
    assert CAVEAT in out
```

**The dispersion sort is unguarded too** — reversing the sort key leaves all 18 green, so the "worst first" header could silently lie. Assert the ordering alongside the paired-figure test.

## Still open for the lead

Carrying @johnmarkpeterbrown's point forward once more: #1811 reserved the structural-anchoring call for the lead, it was closed with "no structural-anchoring decision is warranted", and this PR writes that into ADR-0003 with no **Deviated from the plan** line. Still unaddressed in this push. @DallanQ.

Requesting changes on the one sentence plus the two tests. The engineering is done and it is good — what is left is prose I mis-specified and two unguarded lines.

taiwo-stack:
Fixed in aba066f9.

- ADR-0003's segment-2 sentence corrected: the aggregate gap doesn't rest on segment 2 alone — dropping it still leaves an 8.2-point gap (53.0% vs 44.8%) against the full 13.2-point gap.
- Both unguarded lines now have regression tests: the CAVEAT string is asserted present, and the per-run late-segment table's worst-first ordering is asserted. Mutation-tested both — deleting the CAVEAT append and reversing the sort key each failed their new test while leaving the rest of the suite green.

Harness suite: 2694 passed, 3 skipped. CI green.

johnmarkpeterbrown (approved):
Clear on my review. Looks good!

T-FEH (approved):
Third pass, on `aa55dd6a` (already current with main). Harness suite green: **2739 passed, 2 skipped**. No eval re-run owed — the diff touches no skill, unit test, fixture or run-log path.

All three of my round-2 findings are fixed, and I verified each by running rather than reading.

## 1. The segment-2 attribution — corrected, every figure re-derived

The clause I mis-specified is gone and the decomposition is right. Checked against live report output:

```
ADR: dropping seg2 leaves early 53.0% (87/164) vs late 44.8%, 8.2-pt gap where full is 13.2
actual: 87/164 = 53.0% vs 44.8%  ->  gap 8.2  |  full gap 13.2        MATCH
```

## 2 & 3. Both guards are real, not tautological

The suite grew 18 → 20 and both new tests discriminate:

```
delete lines.append(CAVEAT)  -> reds test_format_report_prints_the_caveat_on_every_non_empty_report
reverse the dispersion sort  -> reds test_format_report_per_run_late_segment_rows_are_worst_first
```

That matters because a tautological guard is exactly what round 1 shipped.

## I re-derived every other quantitative claim too

This PR has had two rounds where a prose fix introduced a *new* false statement — one of them mine — so I did not stop at the sentence I flagged. Every number in the revised Consequences reproduces:

- pre-nudge by subtraction: EARLY 143/312 = **45.8%**, LATE 44/101 = **43.6%** — and the run count checks out as 50 − 16 = **34 segmentable** (the 07-27 window is 57 runs less 7 `unsegmentable-timeline`, the 08-04 window 16 less 0)
- post-nudge **58.1% / 44.8%**, 16 runs, 29 late calls
- the rise is 12.3 points ("about 12"); late moves 1.2 ("barely at all"); the pre-nudge gap is **2.2** points
- excluding `victoriano-macatangay-parents`: 13/16 = **81.2%**
- paired: **5 of 15**, EARLY 79/94 = **84.0%**, LATE unchanged at 13/29 — and LATE genuinely is unchanged, because every late call in the window comes from a paired run

And the blast radius is smaller than it looks: `architecture.md` and `e2e-testing-guide.md` add **no numeric claims at all**, pointing at the ADR instead, and `record-search-tool-spec-v2.md` adds only a cross-reference. Keeping the figures in one place is the right call — it is why the corpus growing can only ever stale one file.

## Four more mutations on the new output, all caught

```
per-segment numerator (count all as supplied)   -> 1 failed
per-segment drops late segments                 -> 1 failed
paired bucket misroute (all calls to early)     -> 3 failed
per-run double count                            -> 2 failed
```

## Verdict

Approving. The report computes and prints exactly what the three review rounds asked for, the aggregation choices that were doing unstated work are now visible in the tool's own output, and the prose finally matches the arithmetic — including the correction to my own overstatement.

Two things that are not mine to clear:

- @johnmarkpeterbrown's `CHANGES_REQUESTED` still stands at `1a0d6af4`. All four of their findings verified fixed, but stale reviews are not dismissed on push here, so it needs their own refresh.
- #1811 reserved the structural-anchoring call for the lead, it was closed with "no structural-anchoring decision is warranted", and this PR writes that into ADR-0003 with no **Deviated from the plan** line. @DallanQ should see that the call was made — third time raised, and still the right person's decision rather than a blocker on the code.

PR #1814, opened 2026-08-21
deep dive: timeline — grading fixes, one fixture fix, seven validator requests (#1655)

### LINKED ISSUE #1655: deep dive: timeline
**Touches:** eval/tests/unit/timeline/, packages/engine/plugin/skills/timeline/SKILL.md

Audit `timeline` for defects nothing has caught. This starts from **no known

Did / Should / Gap.

## Start with this grep

```sh
grep -l -iE '"[^"]*\bscore [123]\b' eval/tests/unit/timeline/*.json
```

without writing the finding.

## This skill's numbers

| | |
|---|---|
| Tests in the suite | 9 |
| Rubric+base dimensions that never discriminate | **3 of 9** |

3 of 9 dimensions never take more than one value, so they cannot report a defect in whatever they nominally grade.

## What to hand back

## Cost

Editing anything under `eval/tests/unit/timeline/` or the skill body flips this
skill's run-log snapshot inactive and buys one `make eval-skill SKILL=timeline`

Emruthwill:
## Summary

**Tier: Normal.** Deep dive on `timeline` per `docs/skill-deep-dive-guide.md` — nine findings the suite passed today and should not, seven validator requests, the grading fixes I own, and one fixture fix.

- **Three headline defects, all passing with human-confirmed all-3s before this PR.** A Haversine distance the model computed itself and persisted as `distance_from_previous_km` (5150 and 5139 km with zero `place_distance` calls) — nothing graded the tool ledger, and the harness had been emitting `missing_tool_usage_dimension` on every positive test. `1890_census` in `gaps[].expected_events` across six distinct gaps, with three judge rationales crediting the array with acknowledging a destruction it doesn't. And an identity-coherence **Pass** issued on the parentage hypothesis `h_001`, grounds including the death certificate naming the father — the evidence weighing this skill hands off.
- **`rubric.md`:** new `Tool usage — place resolution and distance` graded on the tool ledger; `Gap detection` and `Identity coherence` rescoped to the persisted field; `Geographic feasibility`'s N/A made mandatory; `Chronological ordering` retired to stay inside the spec §7 five-dimension cap.
- **Two paid runs where the issue budgeted one, and that overrun is mine.** #1655 says the edits buy *one* run and to batch every finding into it. Every finding *was* batched into run 1; run 2 exists because run 1 exposed a bug in a fix of mine. $6.42 against a ~$9 budget so it is inside the money, but it cost a second annotation pass — the scarcer half. Two ~$0.5 single-test probes are what held it to two runs rather than three.
- **The second run is why this PR looks the way it does.** Run 1 exposed a bug in my own new clause — it could not tell a Step-6 verdict from the Step-5 coherence signal Step 5 *requires* — which failed `ut_timeline_005` for correct behaviour. Fixed and re-measured. Full account in `docs/deep-dives/timeline-findings-2026-08-21.md` § "Measured: what the two paid runs showed".

## Review guidance

**Start here:** `eval/tests/unit/timeline/rubric.md`, the `Identity coherence` scope paragraph, and the matching clause in the four `judge_context` files whose `t_001` carries `hypothesis_id: h_001`. Two things want checking. (1) The clause is **conditional, not a flat N/A** — a per-test override is binding and wins outright (`eval/harness/judge/prompt.md`, "Overrides"), so a flat "this dimension is N/A" would have made the new fail branch unreachable on the two tests that actually reproduce the defect. (2) The verdict/signal distinction is the load-bearing sentence: a verdict is an explicit Pass / Fail / Inconclusive about the hypothesis; an infeasible pair, a doubtful attribution or a `pe_*` confidence question is a Step-5 signal, required behaviour, never a deduction. Get that line wrong in either direction and the dimension either can't fire or fails compliance.

**Unsure about:** whether `Identity coherence`'s new `partial` band is set at the right place. It scores description-only visibility of a competing birth year as partial and reserves pass for a structured field (own event, or `conflict_ids`/`conflict_note`). My reasoning is that prose inside a census row can't be sorted, doesn't render in the viewer's date column, and isn't readable structurally by the next skill — so it is information retained but not *chronology*, which is the one thing this skill exists to produce. A first draft made it a pass and that graded the finding's own cited run a pass, leaving it unfalsifiable. I think partial is right; it is a genealogical judgement call and I'd like a second opinion on it.

## Plan

**Didn't change:** three seed-fixture defects, copied identically across all six Flynn scenarios — `t_001`'s 1860 census event citing `a_009` (`~1845`), `t_001`'s first gap carrying `"start": "1860-01-01"` (which the body prohibits by name, and which every run silently corrects), and four dated assertions (`a_002`, `a_003`, `a_009`, `a_011`) that no `person_evidence` entry links to I1. `mid-research-flynn` is referenced by **123 tests across 20 skills** and `flynn-multi-conflict` by **7 across 6**, so each edit buys a paid run for every one of those skills. V3 carries a before-state exemption so the validator is fair without them. `flynn-impossibility` **is** fixed (`pe_008`–`pe_011`) because it is timeline-only and a partial fix would have left `ut_timeline_004` unable to produce the chronology its own `judge_context` describes.

**Acceptance check:** `ut_timeline_005` in `eval/runlogs/unit/timeline/v1_2026-08-21_14-34-37.json` — Identity coherence **N/A**, Tool usage / Geographic feasibility / Correctness / Deferral all **3**, outcome `pass`. In run 1 (`v1_2026-08-21_14-07-10`, kept in this PR as the record) the same test scored **1 / 2 / 2 / 2 / 2** and failed. Confirmed on two independent samples. Separately, `ut_timeline_004` persists five events where the pre-fix fixture made the 1908 death unreachable, and `parse_rubric` returns 5 dimensions with `has_tool_usage_dimension` True, clearing an advisory that had fired on every positive test.

**Deviated from the plan:** yes, twice, both mid-flight and both recorded in `PLAN.md`. (1) Round one of `/critique-plan` had me accept an event `description` as satisfying the persisted-visibility rule; round two showed that graded the finding's own instance a pass, so it became a three-way bar instead. (2) The plan proposed an F10 for the negative-test Correctness 1 / Completeness 1. Checking `activated` dissolved it — on `_003`/`_009` the timeline skill never ran (`activated=False`, `turns=0`), so no body line can reach it. `_008` is the real defect and goes to #1646. Neither deviation hit the step-4 stop rule.

## Test plan

- [ ] `make test-all` — **in flight at time of opening; I will tick this and say so once it reports.** Its constituents are green independently: `eval/harness` pytest **2459 passed / 3 skipped**, `packages/engine/mcp-server` packaging **504 passed** (24 files). `doc-links.test.ts` earned one failure en route by rejecting two `file.py:LINE` citations in the write-up; those now cite symbols.
- [x] `make eval-skill SKILL=timeline` run and committed with its `.ann.json` — `v1_2026-08-21_14-34-37` (active, snapshot clean against the working tree), annotated by @Emruthwill over its `review_sample` of `_001, _002, _005, _007, _008`, with written comments on `_008`'s two non-pass cells. Run 1 is also committed as the record of why the fix was needed; it is inactive on the seven files revised after it, and the write-up names its `_005` scores as artefacts so nobody reads them as skill signal. `check_runlogs.py` locally: **all runlog rules satisfied**.
- [x] No `description` frontmatter or DO-NOT clause changed, so no routing-pair reciprocity work is owed.

Also verified: `check_rubric_tool_drift.py` → 0 timeline hits; `check_runnable` → 9/9; `validate_research_json` on the edited scenario → no errors.

## Follow-on issues

**Folded in / filed instead:** everything actionable is folded in — nine findings, seven validator requests, one fixture fix and three `SKILL.md` edits all ride this single eval slot, because splitting them buys a second paid run for nothing. Nothing filed. Two items need a decision I can't make and one needs a comment on an existing issue; all three are below rather than in new cards.

### Validator requests V1–V7 — for @DallanQ to file or assign as one card

`developer` + `nothing-checks`. That label is the register `docs/architecture.md` keeps under "What nothing checks", and all seven are missing guards. Full text with rule, lookup site, why-it-isn't-judgment and a cited violation is in `docs/deep-dives/timeline-findings-2026-08-21.md`.

| | rule | cited violation |
|---|---|---|
| **V1** | every persisted `standard_place` / `distance_from_previous_km` traces to a tool response in the same run | `_001` + `_002`, `07-21` — 5150 / 5139 km, zero `place_distance` calls |
| **V2** | `expected_events` names only record sets that survive (1890 US federal census the standing case; the surviving veterans schedule must not be flagged) | `_006`, `07-29_20-42` |
| **V3** | an event must not cite an assertion dated to a different year — with a before-state exemption so a seed fixture's pattern isn't punished | `_006`, `07-29_20-42` — 1870 cites `a_015` (`~1832`) |
| **V4** | the reply does not re-render the persisted timeline (≥2 table rows carrying event dates, or ≥2 distinct distances) | `_007`, `07-21` (9 rows); `_002`, `07-29_20-42` |
| **V5** | an unresolved `conflicts[]` entry is named in `conflict_ids` on any event citing its competing assertions | `_002`, `07-19` |
| **V6** | `date_certainty` matches the date's precision — a **backfill** for the retired `Chronological ordering`, with no violation in the corpus, so whoever implements it must break the corpus and watch it fail | none |
| **V7** | a scenario's seed `timelines[].events[]` may only cite assertions reachable via `person_evidence` from one of that timeline's `person_ids`. **Static corpus lint — no run log, no model call, no cost**, so unlike V1–V6 it can be enforced without buying a paid run | all six Flynn scenarios: `t_001` cites `a_002`, `a_003`, `a_009`, `a_011`; none linked to I1 |

**F3 and F7 do not convert today, and the write-up says so rather than leaving the slot blank.** Both rules read `output.text_response`, which `run_validators` does not inject — the same contract change V4 waits on. Their rules are stated so they are ready if it lands. **F6 and F10 do not convert either**: F6 has no violation in the corpus to cite, and F10 is settled by `flag_routing_negative_judge_fail`'s docstring, which records someone already trying and measuring — *"there is still no mechanical discriminator, and gating on empty output is worse than deleting."*

**Two of those cite run logs retention has since pruned.** The 07-08 and 07-19 logs went with both `.ann.json` siblings; they are committed history, so `git show 762d1b122:eval/runlogs/unit/timeline/v1_2026-07-19_03-22-09.json` retrieves one. The write-up carries that command so it isn't rediscovered.

### Decisions for @DallanQ

1. **The whole-corpus run** for the three shared-fixture defects under "Didn't change". 123 tests / 20 skills for `mid-research-flynn` alone.
2. **Where V4 lives.** It needs `text_response` added to `run_validators`' injected args — a contract change touching `orchestrator.py` — or a sixth rubric dimension, which needs a slot. I declined to free one by retiring `Deferral of logical-impossibility detection`: it is `3 × 12`, but its predecessor `Impossibility detection` scored **1** on `_004`/`07-21`, so the axis has discriminated and twelve samples across two runs is a thin sample rather than a dead dimension.

### For #1646

`ut_timeline_008` fails in both runs with a signature distinct from the flap that issue was filed on: `activated=False`, `turns=1`, `skills_invoked: []` and a non-empty reply offering to do the attribution itself — routes *nowhere* and answers inline, rather than routing to `conflict-resolution`. @Emruthwill confirmed the judge's 1/1 there. I'll post this as a comment with the run-log reference once this PR is up.


florencemashipei:
Heads-up from merging #1783 ahead of this one — you will hit a merge conflict when you update this branch, and it is mechanical rather than a re-run.

**What you will see.** #1783's run pruned `eval/runlogs/unit/timeline/v1_2026-07-08_15-47-10.ann.json` under the keep-newest-5 rule, and this branch shows that same file *renamed* into its new annotation. So git reports two conflicts on the same path:

```
CONFLICT (rename/delete): eval/runlogs/unit/timeline/v1_2026-07-08_15-47-10.ann.json renamed to
  eval/runlogs/unit/timeline/v1_2026-08-21_14-34-37.ann.json in HEAD, but deleted in origin/main.
CONFLICT (modify/delete): eval/runlogs/unit/timeline/v1_2026-08-21_14-34-37.ann.json deleted in
  origin/main and modified in HEAD.
```

**Resolution:** keep your own new run logs and annotations, accept main's deletion of the pruned `07-08` pair. Concretely `git add eval/runlogs/unit/timeline/v1_2026-08-21_14-34-37.ann.json`, leave the `07-08` path deleted, commit.

**Do not hand-edit the `.ann.json` to fix it.** `eval/CLAUDE.md` forbids it, and `check_runlogs.py` rule 5 fails on a file containing conflict markers with exactly that instruction — restore the valid version from git, or delete it and re-annotate in the CRUD UI. Choosing whole file versions to settle a conflict is fine; editing the corrections inside it is not.

**You do not owe a timeline re-run.** I verified this rather than assuming it: merged the post-#1783 main into this branch's head locally, resolved as above, and ran the gate — `All runlog rules satisfied`, and your `v1_2026-08-21_14-34-37.ann.json` came through with its 35 corrections intact and no markers.

The reason it is free is worth knowing for future ordering: `snapshot.py::diff_snapshot_vs_disk` states *"Paths present on disk but absent from the snapshot are NOT flagged — the snapshot is the authority on what's tracked."* #1783's only timeline change was an **added** file, `eval/tests/unit/timeline/negative-hypothesis-tracking.json`, which is absent from your snapshot and therefore invisible to it. Had the order been reversed, #1783 would have gone stale and owed a paid run plus an annotation pass, because this PR *modifies* `timeline/SKILL.md`, `timeline/rubric.md`, six existing timeline tests and `flynn-impossibility/research.json` — all already keys in its snapshot. That asymmetry is why #1783 went first.

One consequence to be aware of, not an action: your run predates `ut_timeline_010`, so once this lands, timeline's newest run log on main will not have executed that test. Its passing record stays in #1783's `v1_2026-08-16_19-22-07` until timeline is next run, which picks it up. Nothing gates on it.

mercyokum (approved):

EdmondOware (approved):
The acceptance check reproduces exactly. Pulled both run logs. Run 1 has _005 failing with Identity coherence 1 and Tool usage, Deferral, Geographic and Correctness all 2, precisely the 1 2 2 2 2 the body claims. Run 2 has _005 passing with Identity coherence null and the other four at 3. So the bug you introduced and the fix are both evidenced in committed artifacts, not in an uncommitted scratch run. Committing run 1 as the record was the right call.

The rubric respects the five dimension cap, counted directly. Tool usage, Gap detection, Deferral, Geographic feasibility, Identity coherence. Chronological ordering retired to make room, and the new dimension is genuinely ledger scoped rather than narration scoped.

The didn't change scope reasoning is measured rather than asserted, and I checked both halves. flynn-impossibility is referenced by exactly one test file, timeline's own, so fixing it really carries no cross skill cost. mid-research-flynn appears in 141 files across roughly 20 skills, so leaving it alone really does avoid a mass paid re run. The V3 before state exemption is the right way to keep that validator fair despite the unfixed seed defects.

The genealogy is right. The 1890 US federal census was destroyed by fire in 1921 and is unrecoverable, so naming it in expected_events sends a researcher after nothing, and the hard prohibition is correct. Restricting Mode B coherence verdicts to identity questions is also right, chronological coherence cannot settle parentage, that is evidence weighing and belongs to proof conclusion. A Pass on h_001 was a real doctrine violation, not a nit.

On your unsure about, the partial band for description only visibility. Partial is right. A competing birth year buried in an event description is not chronology, it cannot be sorted, does not surface in the viewer's date column, and the next skill would have to parse prose to find it. This skill exists to produce a structured chronology, so a finding surviving only as prose is half delivered. Your own argument settles it too, making it a pass graded the finding's cited run a pass and left the dimension unfalsifiable.

Two trivial things, neither blocking. The title says six validator requests where the body specifies seven, V1 through V7. And the make test all box is still unticked with your note that you would tick it once it reports.

Annotation is complete on run 2's sample, 35 entries, zero missing comments on non passes.

PR #1815, opened 2026-08-21
deep dive: validate-schema — findings + V2 read-only-files validator (#1658)

### LINKED ISSUE #1658: deep dive: validate-schema
**Touches:** eval/tests/unit/validate-schema/, packages/engine/plugin/skills/validate-schema/SKILL.md

Audit `validate-schema` for defects nothing has caught. This starts from **no known

Did / Should / Gap.

## Start with this grep

```sh
grep -l -iE '"[^"]*\bscore [123]\b' eval/tests/unit/validate-schema/*.json
```

without writing the finding.

## This skill's numbers

| | |
|---|---|
| Tests in the suite | 10 |
| Rubric+base dimensions that never discriminate | **3 of 6** |

3 of 6 dimensions never take more than one value, so they cannot report a defect in whatever they nominally grade.

## What to hand back

## Cost

Editing anything under `eval/tests/unit/validate-schema/` or the skill body flips this
skill's run-log snapshot inactive and buys one `make eval-skill SKILL=validate-schema`

kofiatinka12:
**Closes #1658.**

## What this is
The **validate-schema deep dive** (per `docs/skill-deep-dive-guide.md`) — plus the one genuinely convertible gap, implemented here rather than filed.

Two deep-dive docs under `docs/deep-dives/`:
- `validate-schema-prohibition-list.md` — the Step 1 checkable-rule list.
- `validate-schema-findings-2026-08-21.md` — findings in Did/Should/Gap form.

The headline: 3 of 6 dimensions never discriminate, and validate-schema is a pure *report-writing* skill whose output is `text_response` — which validators cannot read — so most of what it can get wrong is judge-only, not mechanizable. That leaves exactly one convertible gap.

## V2 identified and implemented in this PR
**V2 was the one genuinely convertible gap** (F1): the skill is read-only, but the existing `test_does_not_modify_project_files` only guards `research.json` and `tree.gedcomx.json` — it never inspects the `files` map, so a write to a `results/*.json` sidecar (exactly the kind of error `ut_validate_schema_009` asks the skill to *report*, not fix) would pass every existing guard.

Implemented:
- **`test_does_not_modify_sidecars_or_other_files`** in `eval/harness/validators/test_validate_schema.py` — a positive-run check that **`after_state["files"]` equals `before_state["files"]`**, failing on any file added, removed, or modified (including `results/*.json` sidecars). It **extends read-only enforcement to the `before_state["files"]` / `after_state["files"]` map** and is strictly additive to `test_does_not_modify_project_files` (which guards the two main JSONs via their own snapshot keys) — it does not replace or weaken it.
- **Six direct unit tests** in `eval/harness/tests/unit/test_validate_schema_validator.py` (unchanged-files pass; added / removed / modified sidecar fail; negative-test skip; main-JSON edits ignored here and left to the other validator). Direct tests are required because `pyproject.toml`'s `testpaths = ["tests"]` means `validators/` is not collected by `make harness-test`.

Verified through the real discovery path: invoking `run_validators(skill="validate-schema", …)` discovers and runs the new validator (passes on a clean run, fails on a modified sidecar).

## F2 and F3 — intentionally NOT converted
- **F2** (a single-run `sourceId` vs `source_id` relay the judge nonetheless passed) stays a **judge-only** finding — a validator can't check it, because the skill's report is `text_response`, which is not a validator input.
- **F3** (faithfulness is un-mechanizable two ways: validators can't read `text_response`, and the committed run log strips the tool `response` so blind annotators can't see the validator's errors) stays a **harness recommendation** — not a validator.

## Cost
**No paid `make eval-skill SKILL=validate-schema` run was required.** The run-log snapshot covers `plugin/skills/<skill>/**`, `eval/tests/unit/<skill>/**`, and referenced fixtures — not `eval/harness/validators/**` or `eval/harness/tests/**` — so editing a validator marks no skill "touched" and invalidates no run log. The new validator is exercised for free by the `pytest` job.

## Files changed
- `eval/harness/validators/test_validate_schema.py` — V2 validator
- `eval/harness/tests/unit/test_validate_schema_validator.py` — 6 direct tests
- `docs/deep-dives/validate-schema-findings-2026-08-21.md`
- `docs/deep-dives/validate-schema-prohibition-list.md`

## Local checks
`pytest tests/unit/test_validate_schema_validator.py` → 6 passed; `test_encoding_lint` → passed; `test_validator_runner` → passed; `doc-links.test.ts` → 130 passed.


EdmondOware (approved):
Confirmed the validator's central claim in the source. snapshot_files really does skip research.json and tree.gedcomx.json when building the files map, so the new check genuinely covers everything else including results sidecars, and genuinely does not double check or weaken the existing validator. The docstring's description of that split is accurate.

Proved the new tests can actually fail rather than trusting that they can. I neutered the assertion, replacing assert not added or removed or modified with assert True, and reran. Exactly the three firing tests went red, modified sidecar, added file, and removed file, while the three non firing ones stayed green. Restored and reconfirmed all six pass. Real coverage, not vacuous.

Scope reasoning holds. Editing the validators directory is genuinely outside the run log snapshot, so no paid eval run is owed. F2 and F3 are correctly left unconverted, validators really do not receive text_response, so those are judge only and harness level findings rather than convertible ones.

One pre existing hole this PR reveals but does not close, and should not, it would be scope creep. The older test_does_not_modify_project_files only asserts when both before and after are non None, and snapshot_files sets research_json to None on a JSON decode error. So a run that deleted research.json, or corrupted it into invalid JSON, leaves after as None, the guard short circuits, and the assertion never runs. The new validator cannot catch it either since research.json is deliberately outside the files map, and test_no_entries_deleted has the same is None then skip shape so it does not backstop it. That is the check that cannot fail pattern, worth its own issue against the older validator.

No blocking defects.

yinkid28 (approved):
Verified the validator and its tests against a test-merge with main (2 commits behind, merges cleanly).

  **What I checked:**
  - `snapshot_files` in `workspace.py` confirms the `files` map excludes `research.json` and `tree.gedcomx.json`, so the new validator genuinely covers the gap without overlapping `test_does_not_modify_project_files`.
  - All 6 direct tests pass; full harness pytest passes (2540 passed, 1 environmental skip from missing engine build).
  - Encoding lint passes — no bare `open()`/`read_text()` without `encoding="utf-8"`.
  - The "no paid eval run needed" claim holds: `check_runlogs.py` snapshots `plugin/skills/<skill>/**` and  `eval/tests/unit/<skill>/**`, not `eval/harness/validators/**` or `eval/harness/tests/**`.
  - Deep-dive docs follow the established pattern in `docs/deep-dives/`.
  - F2/F3/F4 are correctly left unconverted — `text_response` really is absent from `validator_runner.py`'s `available_args`.

  **Minor note (not blocking):** The `(before_state or {}).get("files", {}) or {}` defensive chain at line 80 is safe but slightly over-guarded — `run_validators` always passes dicts, never `None`. The existing `test_does_not_modify_project_files` accesses `before_state` directly. Doesn't matter either way.

  EdmondOware's approval reads the current head and manually proved the assertions can fail. Agreeing with that review, no blocking defects.

PR #1822, opened 2026-08-21
Force utf-8 on every text-mode subprocess capture (salvaged from PR 1723)

DallanQ:
**Both commits here are @taiwo-stack's, cherry-picked unchanged from PR #1723 with authorship intact.** That PR's main change — proof-conclusion owning the `questions` → `resolved` transition — landed by another route in PR #1819, and its branch is now conflicting against a `proof-conclusion/SKILL.md` that no longer resembles what it edited. This rescues the part that is independent of all that and still broken on `main`.

## The bug

`subprocess.run(..., text=True)` with no `encoding=` decodes with the **platform default** — cp1252 on Windows. Non-ASCII output from the subprocess then raises `UnicodeDecodeError`.

Six live sites in `mock_mcp.py` alone, all of them the Node subprocess the mock MCP server shells out to. The failure was being swallowed as a retried "transient stall", so it looked like flakiness rather than a decode error. Verified still present on `main` before branching.

This is the same rule as the repo's existing `encoding="utf-8"` requirement — the one that exists *because the genealogist team runs Windows* — in a shape the lint could not see.

## What the two commits do

**`force utf-8 on every text-mode subprocess capture`** — fixes the six `mock_mcp.py` sites plus ten more across `eval/harness` and `apps/server`, and widens `test_encoding_lint.py` so the `subprocess` shape is caught from now on. The lint parses with `ast` rather than grepping, matching the existing rule's approach.

**`fix 3 code-review-found bugs from the utf-8 fix`** — taiwo ran `/code-review` on their own work and acted on it: three real bugs in the encoding change itself, the six near-duplicate Node call sites consolidated into one helper, and an `lru_cache` that permanently cached a transient failure.

## Verification

- `make harness-test` 2,571 passed, 2 skipped · `make engine-test` 2,570 · `make server-test` 209 passed, 1 skipped · `make harness-lint` clean
- **Broke the lint to watch it fail**: removed one `encoding="utf-8"` from `provenance.py` and it reported `eval/harness/e2e/provenance.py:134: subprocess.run() has no encoding= (add encoding="utf-8")`, then restored it
- Confirmed no `text=True` without `encoding=` remains in `mock_mcp.py`

## On PR #1723

Closing it rather than rebasing. Beyond this encoding work, the two substantive changes it carried have both landed independently: the question-resolution ownership change, and a judge-anchoring fix on `rubric.md`'s "Narrative standalone" review-mode branch — which taiwo reproduced 3-for-3 across committed run logs, and which PR #1819 hit and fixed from the other direction. Independent discovery of the same bug, not duplicated effort.

One item from #1723 that did **not** land anywhere and may still be worth having: it raised `ut_proof_conclusion_018`'s turn budget to 40, noting it was unset and defaulting to 20. That test was unstable across a dozen runs during Phase 3 and a turn cap was never considered as a cause.


johnmarkpeterbrown (commented):
Green everywhere, and I confirmed the fix myself: I deleted the `encoding="utf-8"` you added to `provenance.py:139` and the new rule fired on exactly that line, nothing else. I also checked the five `text=True` sites outside your diff — all already compliant. The CPython gate your docstring quotes is verbatim correct.

Three small things, none of them in the fix.

## 1. CLAUDE.md still scopes this rule to file I/O

It's the file that calls this lint "the repo-wide guard," so it should say what the guard now guards.

```
You wrote:      Every Python `read_text()` / `write_text()` / `open()` on a text file
                **must** pass `encoding="utf-8"`.

Change it to:   Every Python `read_text()` / `write_text()` / `open()` on a text file
                **must** pass `encoding="utf-8"` — and so must every `subprocess.run` /
                `check_output` / `Popen` / `call` / `check_call` that runs in text mode
                (`text=`, `universal_newlines=` or `errors=`), which decodes the child
                process's output with the same platform default.
```

Leave the heading alone — other docs cite it as an anchor.

## 2. The blind spot your docstring names has nothing watching it

```
You wrote:      Matched only on the ``subprocess.<name>`` attribute form (this repo's
                sole convention -- no ``from subprocess import run`` and no
                ``import subprocess as ...`` alias exists today
```

True today. But add `import subprocess as sp` and every `sp.run(cmd, text=True)` after it is invisible to the lint — CI green, Windows broken. That's the shape this PR exists to close. Fifteen lines in the file you're already editing:

```python
def test_no_aliased_subprocess_import():
    """`subprocess.<name>(...)` is the only form `_is_subprocess_call` matches, so an
    alias or a direct import hides a real offender while the lint still goes green.
    Nothing else detects that: widen `_is_subprocess_call`, do not skip this."""
    offenders: list[str] = []
    for path in _iter_python_files(REPO_ROOT):
        rel = path.relative_to(REPO_ROOT).as_posix()
        try:
            tree = ast.parse(path.read_text(encoding="utf-8"))
        except (SyntaxError, ValueError, UnicodeDecodeError):
            continue
        for node in ast.walk(tree):
            if isinstance(node, ast.Import):
                offenders += [
                    f"{rel}:{node.lineno}: import subprocess as {a.asname}"
                    for a in node.names
                    if a.name == "subprocess" and a.asname
                ]
            elif isinstance(node, ast.ImportFrom) and node.module == "subprocess":
                offenders.append(f"{rel}:{node.lineno}: from subprocess import ...")
    assert not offenders, (
        "the encoding lint only matches subprocess.<name>(...) — widen "
        "_is_subprocess_call before introducing:\n  " + "\n  ".join(sorted(offenders))
    )
```

I ran that matcher against all four shapes — green on `import subprocess`, red on both blind-spot forms, no false flag on `import json as j`.

## 3. The `ut_proof_conclusion_018` turn-budget note is orphaned

PR #1723 closed yesterday, so it now lives only in a closed PR's body. It already has a home: **issue #1603**, *"…_001 aborts on unset max_turns"* — same skill, same defect class, and that issue exists to batch these into one paid run. Comment there rather than filing anything new. The default checks out: `orchestrator.py:776` reads `spec.execution.get("max_turns", 20)`.

---

Good work — the encoding fix is correct and complete. Nothing here is in the fix itself.

johnmarkpeterbrown (approved):
Encoding fix is correct and complete. The three notes in my earlier review are non-blocking.

johnmarkpeterbrown (approved):
Encoding fix is correct and complete. The three notes in my earlier review are non-blocking.

promise-emmanuel (changes requested):
Encoding fix is correct and complete — I ran both suites in a worktree off your head (eval/harness 2672 passed / 2 skipped, apps/server 209 passed / 1 skipped), broke the lint in five separate trees and watched it fire in all five, and AST-audited every subprocess call in the repo: none of them escapes your matcher. Nothing wrong in the fix.

Two things to add before this merges.

1. The blind spot your docstring names has nothing watching it
Your docstring is right that no alias exists today. But nothing stops one. I dropped these two files into eval/harness/ and ran the lint:

import subprocess as sp
sp.run(["x"], text=True)

from subprocess import run
run(["x"], text=True)
Both are the exact cp1252 crash this PR exists to stop. The lint stayed green. That is the worst shape a check can have here — it reads as coverage, and it breaks on a Windows genealogist's machine while CI says everything is fine.

@johnmarkpeterbrown already wrote the guard for you in his first review. I ran his version in both directions before asking for it: green on the clean tree, red on both forms above, no false flag on import json as j. Add it verbatim as test_no_aliased_subprocess_import in the file you are already editing.

2. The failure message names three functions and none of them is the one that failed
When your new rule fires, the developer reads a header that says the rule is about file I/O.


You wrote:      'bare text-mode file I/O found -- CLAUDE.md requires encoding="utf-8" on every '
                "read_text/write_text/open:\n  " + "\n  ".join(sorted(offenders))

Change it to:   'bare text-mode I/O found -- CLAUDE.md requires encoding="utf-8" on every '
                "read_text/write_text/open and every text-mode subprocess call:\n  "
                + "\n  ".join(sorted(offenders))

And take John's CLAUDE.md edit from note 1 of his review while you are in here — line 645 still scopes the rule to file I/O, and that is the file calling this lint "the repo-wide guard."

promise-emmanuel (approved):
Approving. Both things I asked for are in and correct. I watched the alias guard fire on import subprocess as sp and on from subprocess import run, and test_no_bare_text_mode_file_io stayed green through both — which is the proof that guard is the only thing covering that shape. Suites on your head: eval/harness 2741 passed / 2 skipped, apps/server 209 passed / 1 skipped, ruff clean, all required checks green.

One thing to push before you merge. The guard I asked you to take verbatim brought a false positive in with it. That's mine, not yours.

It flags every from subprocess import <anything>, not just the callables. This file is fully compliant and still reds CI:


from subprocess import PIPE, CalledProcessError
subprocess.run(["echo"], stdout=PIPE, text=True, encoding="utf-8")
The failure message then tells the reader to widen _is_subprocess_call, which has nothing to do with PIPE. PIPE, STDOUT, DEVNULL, CalledProcessError, TimeoutExpired and CompletedProcess cannot bypass the matcher. Only the five callables can.

eval/harness/tests/unit/test_encoding_lint.py:236 —


You wrote:      elif isinstance(node, ast.ImportFrom) and node.module == "subprocess":
                    offenders.append(f"{rel}:{node.lineno}: from subprocess import ...")

Change it to:   elif isinstance(node, ast.ImportFrom) and node.module == "subprocess":
                    offenders += [
                        f"{rel}:{node.lineno}: from subprocess import {a.name}"
                        for a in node.names
                        if a.name in SUBPROCESS_METHODS or a.name == "*"
                    ]
Keep the or a.name == "*". Narrowing to the method set alone would drop the star-import bypass the current version does catch. I ran that version against seven shapes before asking: red on run, run as r, Popen, PIPE, import *; green on PIPE, CalledProcessError, plain import subprocess, and import json as j.

This is not blocking the merge — it fails in the safe direction, and no file in the repo trips it today. Push the two lines and it's closed.

Last thing, not yours to fix here. John's third note is still homeless: the ut_proof_conclusion_018 turn-budget observation needs a comment on issue #1603. That issue covers _001's unset max_turns but not _018. I checked — bounded-death-encoded-not-collapsed.json has an execution block with no max_turns, so orchestrator.py:776 defaults it to 20.

Issue #1831, opened 2026-08-22
Can a project start from a record instead of a tree person? All three bootstrap paths end at a tree node


DallanQ:
**Touches:** packages/engine/plugin/skills/init-project/SKILL.md, packages/engine/plugin/skills/research/SKILL.md, packages/engine/plugin/skills/question-selection/references/question-formulation.md, eval/tests/unit/init-project/, docs/alpha-user-guide.md, docs/alpha-user-guide-cowork.md

A senior genealogist: "You have to start somewhere, but a node in a tree is not a good starting point... Sometimes a record is a much better starting point. First job: evaluate the record. The record becomes the working set of probably-correct assertions compared against the other records examined."

## What we support

Three bootstrap paths, all terminating in a tree person:

1. A FamilySearch person ID, read directly
2. A name, resolved through `person_search`, then read
3. If no candidate matches, initialize from the objective text using local stub persons

The skill description states it outright, and the user guide documents the same two routes. The unit corpus mirrors it — the init-project suite has a from-tree test, a from-search test, and an objective-only test, and no record-start test.

## The tools would already carry it

- `project_create` takes the tree and subject person IDs as **optional**
- `record-extraction` accepts an ARK, a PDF, or an image with no prior person
- `person-evidence` mints a person from a persona via create-or-enrich, and the materialization spec says that "dissolves" the first-record problem — "first record stops being special"

So this is a skill-prose and routing gap, not a tool gap.

## What a filer must reconcile before writing it

- **`question-formulation.md:107` lists "record-first framing" as a common failure** — "What does Reuben's 1900 census entry say about his parents?" is corrected to "Who were Reuben's parents?", with the census belonging in the plan. That is about question *grain*, not project start, but anyone adding a record-first path has to say why the two do not contradict.
- **Issue #1695 records that auto-seeding a project on first write was designed twice and rejected twice** — it lets any skill create an objective-less project that init-project's guard then refuses to touch, with no routing row for it. A record-first start lands in the same territory.

## Note what is already handled, so it is not rebuilt

His "first job: evaluate the record" is already how extraction works. The extractor operates on personas — a record ID plus a record role — and deciding that a persona is a given person is a separate, later, revisable judgment; `docs/gps-research-flow.md:209-212` says identity is "never assumed in the first minute of reading the record", and inline extraction is forbidden, so every record is evaluated in fresh context before any identity call. What is missing is only the entry point, not the discipline.

[...]

Issue #1830, opened 2026-08-22
Plan-time breadth: accessibility as a sixth self-check question, a per-kind breadth obligation (question_kind), a skipped-item skip_category, and the origin-record-first rule


DallanQ:
**Touches:** packages/engine/plugin/skills/research-plan/references/planning-standards.md, packages/engine/plugin/skills/research-plan/SKILL.md, packages/engine/plugin/skills/research-exhaustiveness/SKILL.md, packages/engine/plugin/skills/research/SKILL.md, packages/engine/plugin/skills/search-records/SKILL.md, docs/specs/schemas/research.schema.json, docs/specs/schemas/enums.schema.json, packages/schema/schemas/research.schema.json, packages/schema/schemas/enums.schema.json, packages/schema/src/index.ts, packages/engine/mcp-server/src/validation/validator.ts, docs/specs/research-schema-spec.md

A senior genealogist states the rule as: **the reasonably-exhaustive test is made when the research plan is created, and if a competent genealogist can reasonably access a source, it must be researched.** We are half aligned, and the missing half is accessibility.

## What we have at plan time

A breadth self-check does exist — `research-plan/SKILL.md:312-323`, keyed to the Extent standard in `planning-standards.md:152-173`. Its five questions ask about record types, repositories, variant spellings, jurisdictions and time periods. **None asks whether a competent genealogist could reasonably access the source.** Accessibility appears on the plan side exactly once, at `planning-standards.md:184`, and only as grounds for *stopping*: "Remaining resources are inaccessible."

Three things weaken it further:

- **Nothing persists it.** The plan and plan-item objects in the schema carry no field recording that the breadth check ran or what it concluded. The only exhaustiveness judgment stored anywhere is the declaration written *after* searching.
- **A skipped item's reason has nowhere to live.** `search-records/SKILL.md:630` says do not skip silently, set the status with an explanation, and `research-exhaustiveness/SKILL.md:82` tells the evaluator to read "skipped plan items and their reasons" — but no schema field holds one. An accessible planned source can be dropped with no recorded justification and the question can still be declared exhaustive.
- **The two skills disagree on whether the plan must be finished.** The orchestrator routes to the exhaustiveness gate "even with plan items still `planned`" and calls a plan "a prioritized list, not a checklist to drain". The gate says it may "only evaluate a question whose plan items are all completed or skipped". So the router sends work in that the gate says it may not evaluate.

## And the bottom-of-loop gate is skippable

[...]

Issue #1829, opened 2026-08-22
Nothing has measured what full-text queries and filters actually search — the prose descends from a deleted doc


DallanQ:
> **Reviewed 2026-09-09 (`/review-ready`): `ready-after-edit`.** The probe becomes a section
> in `dev/probe-search-qualifiers.ts` (the producibility check is hard-wired to that one
> file), and **no file under `packages/engine/plugin/skills/search-full-text/` may be touched
> in this PR** — `references/query-syntax.md` included. Recording a figure guards nothing on
> its own; the guard is the two `measured-figures.test.ts` edits below. No lead decision
> needed.

**Touches:** packages/engine/mcp-server/dev/probe-search-qualifiers.ts, packages/engine/mcp-server/dev/measured-figures.json, packages/engine/mcp-server/tests/packaging/measured-figures.test.ts, docs/specs/fulltext-search-tool-spec.md, packages/engine/mcp-server/src/tools/fulltext-search.ts

A senior genealogist asked a question we answer confidently in prose and have never measured: "Do filters and queries both search the meta only? That was the pre-mission behavior."

## What we currently assert

`query-syntax.md` describes three different behaviours, not one:

| What the skill sends | Upstream | What we claim it searches |
|---|---|---|
| `keywords` | `q.text` | the whole transcript |
| `name` | `q.fullName` | NLP-recognized person names from the transcript |
| `place` | `q.recordPlace` | **both** transcript content and collection metadata |
| `collectionId`, `yearFrom`/`yearTo`, `recordType`, `recordPlace0-3` | `f.*` | collection metadata only |

So our answer to him is "no, three behaviours" — the filters are metadata-only, but the free-text query reads the transcript, and the Place *query field* straddles both.

## Why that answer is not yet trustworthy

Nothing in this repo measured any of it. `dev/` has no full-text probe — only `try-fulltext-search.ts`, a smoke test — and `dev/measured-figures.json` has no full-text section at all; its lettered sections are all indexed-search qualifier work.

The prose descends from `docs/gps/full-text-search.md`, deleted in the 127-document cull. That file cited FamilySearch's Help Center, Ouimette, and the RootsTech 2025 syllabus. Citation-grade, not measurement-grade — and the citations now exist only in git history, so a reader today has no way to tell how well-founded the table is.

## The probe

**Add a SECTION to the existing `dev/probe-search-qualifiers.ts`. Do not write a
[...]

Issue #1828, opened 2026-08-22
search-full-text: reverse the collection-scoping ban and use includeFacets, with get_name_variants replacing the diminutive expansion that move gives up


DallanQ:
## What he says

> "The enemy of a successful search is too many results." Scoping to a single collection is a powerful tool for cutting the number of results to a manageable number.
> Unscoped searches are a sign of a poor research plan.

And, on places:

> In my experience, the place specified in the metadata is more helpful than places identified in the text.

## What we ship

An absolute ban, in four prose sites and two enforcement layers:

- `SKILL.md:104-110` — "**Do NOT scope a full-text search to a record `collectionId`.**"
- `query-syntax.md:57-68` — a whole section, "Do not scope FTS to a record collection ID"
- `search-strategies.md:109-111` — run it "unscoped"
- `SKILL.md:132` — the worked example is commented "UNSCOPED (no collectionId)"
- `test_search_full_text.py:235` — `test_fulltext_search_never_scopes_to_collection_id`, which fails on **every** call in every test, not only the first
- `rubric.md:11` grades leaving the first call unscoped as the pass bar, and `test_first_fulltext_search_call_is_unscoped` fails a first call carrying a place, year or record-type filter

The place half is the same rule: we treat the collection metadata place purely as a hazard (`query-syntax.md:103-105`, `transcription-quirks.md:95-98`) and forbid it on the first call.

## Where the ban came from

One case. Commit `8b870d9e` (2026-07-07) fixed an e2e failure where a search was scoped to a `collectionId` **borrowed from indexed record search**, whose full-text partition did not contain the answering records. The prose then generalised that single case into never scoping at all.

## Two things that make his objection land

1. **The legitimate scoping path exists in the tool and no skill teaches it.** `fulltext_search` accepts `includeFacets` and returns collection facets carrying a ready-made `filterParam`. Grepping the whole plugin for `includeFacets` returns nothing. So the only collection-scoping parameter is banned, and the mechanism for obtaining a *valid* full-text collection ID is never surfaced to the agent.
2. **FamilySearch's own guidance says to scope.** The source document this prose descends from carried the RootsTech 2025 line verbatim: "Filter by collection, record type, place, and time, as necessary."

## One nuance to settle with him

Our "unscoped" means "no `collectionId`". His reads as "no scoping at all". Our first-call rule makes the first call unscoped by construction — which is the thing he calls the symptom of a poor research plan. The two halves of this card are the same disagreement.

[...]

PR #1832, opened 2026-08-22
fix: research loop diagram, and repoint five prompt sites at a skill that ships

DallanQ:
## Summary

Tier: Normal (a lint lands with it).

- **The research-cycle diagram contradicted its own caption.** `docs/gps-research-flow.md` drew the "not yet exhaustive" arrow back to *Choose the next question*. The caption three lines below already said "it returns to planning", and so does the research orchestrator: the default is `research-plan`, extending the plan for the same question, with `question-selection` reserved for a FAN pivot. Redrawn with both edges; the caption now says which is which.
- **Five prompt sites routed to `assertion-classification`, deleted 2026-07-11 when record-extraction absorbed it.** One is a *hard block* in research-exhaustiveness that told the agent to stop and recommend it — a stop with no recovery path. All five now name `record-extraction`, which owns classification and refines it in place.
- **Adds the guard that keeps it from recurring.** A packaging test resolves every backticked kebab-case token in a plugin skill or agent body to a shipped skill directory, a shipped agent file, or an explicit not-a-skill list.

Found while triaging a senior genealogist review; the rest of that review is in the issues listed at the bottom.

## Review guidance

**Start here:** `tests/packaging/skill-name-resolution.test.ts`, specifically the `NOT_SKILL_NAMES` list and the extraction regex. The question worth checking is whether that list is the right shape. I chose an explicit not-a-skill allowlist over pattern-matching a routing context, because I measured the false-positive load first: across 108 plugin markdown files there are exactly 17 distinct non-resolving tokens, and every one is a `research_query` query kind, a gps-mentor `focus` value, a status, or an example name. The cost of that choice is that a future legitimate non-skill kebab token reddens CI until someone adds it — which I think is the right failure direction, since the prompt to ask "is this a skill name?" is the whole point. Push back if you would rather it were narrower.

Second thing: the prose rewrite in `proof-conclusion.md` and `research-exhaustiveness/SKILL.md` deleted a distinction as well as a name. Both said the classification must come from `assertion-classification` and *not* be "a leftover record-extraction default". With one skill owning both, that contrast no longer parses, so I replaced it with "a real, reasoned value — not a placeholder". Check that reads as the same bar to you; it is the hard block's only remaining criterion.

**Unsure about:** nothing outstanding. The `eval-cosmetic-skip` question is now answered with measurements rather than a reading — see Test plan. It is still a senior call to apply the label.

## Plan

**Didn't change:** the four other things the same review turned up in these files — the image-as-original contradiction, the exhaustiveness definition leaning on overturn risk, the collection-scoping ban, and the opportunistic-assertion relinking gap. Each is either a lead decision or lands in a contended eval slot, so each is an issue rather than a hunk here. Also left `gps-mentor.md`'s "published genealogy is derivative" error alone (our enum says `authored`) — it is real and unambiguous, but it rides issue #1824 so the doctrine question and the example get fixed together rather than in two passes.

**Acceptance check:** `tests/packaging/skill-name-resolution.test.ts` — the "resolves every backticked kebab-case token" case. Verified it fails on `main`, not just synthetically: with `git checkout origin/main -- packages/engine/plugin/` applied, it reports all three real offenders by file and token —

```
packages/engine/plugin/skills/research-exhaustiveness/SKILL.md: `assertion-classification`
packages/engine/plugin/agents/gps-mentor.md: `assertion-classification`
packages/engine/plugin/agents/proof-conclusion.md: `assertion-classification`
```

and passes on this branch.

**Review round 1 (johnmarkpeterbrown):** fixed the Windows path bug in the lint — `indexOf` on a forward-slash literal returns -1 on a backslash path, so every offender reported as `d: ` instead of naming its file. Now uses the `relative(...).split(sep).join("/")` idiom from `scripts/package-plugin.mjs`; verified against `node:path/win32` that it names the file where the old expression returned `d`, and is byte-identical on posix. Also noted the guard's range in its header. His two other points asked for no change and I agree with both.

**Deviated from the plan:** the lint was not asked for. The ask was the two fixes; I added the guard because without it the same rot returns and nothing sees it, and I measured the allowlist load before committing to the approach rather than asserting it was small. Flagging it explicitly so it reads as a decision, not scope creep — drop the test file if you disagree and the two fixes still stand on their own.

## Test plan

- [x] `make engine-test` — 110 files, 2633 tests, green (merged with main).
- [x] `eval/harness` pytest — 2610 passed, 2 skipped. Needs `npm run build` in the engine first; without it 7 tests fail or skip on the absent `build/`, which is the fresh-worktree condition, not a regression.
- [ ] `make test-all` — **not run in this session.** The change touches one doc, three plugin prose files and one new test; nothing outside the engine package. Worth a reviewer running it, or CI.

- [ ] Per-skill eval runs — **not run, and this is the open question.** The run-log gate reports two skills armed:

  ```
  proof-conclusion:        packages/engine/plugin/agents/proof-conclusion.md — content-differs
  research-exhaustiveness: packages/engine/plugin/skills/research-exhaustiveness/SKILL.md — content-differs
  ```

  **`gps-mentor.md` adds no third entry, and my first explanation of why was wrong.** It is not "already counted" under `proof-conclusion`. The harness embeds an agent in a skill's snapshot only for `@plugin:<name>` references in that skill's own SKILL.md. `proof-conclusion/SKILL.md` references only `@plugin:proof-conclusion`; its one mention of the mentor is plain prose, which the reference pattern does not match. The only SKILL.md that writes `@plugin:gps-mentor` is `research`, and there is no `research` unit suite. `gps-mentor.md` appears in **zero** unit run-log snapshots, so the mentor edit is measured by neither armed skill and by nothing else. Filing that coverage gap separately.

  **On `eval-cosmetic-skip`: I now think it does apply to the two armed skills, and the reason is measured, not read.** My original argument and johnmarkpeterbrown's both rest on a strict reading of the old prose — that a classification could not have come from a skill that stopped shipping, so the hard block always fires. The run logs say it never did:

  - The research-exhaustiveness run of 2026-08-20 contains `assertion-classification` **0 times** and `unclassified` **0 times**, at 16/17 pass.
  - In the proof-conclusion run of 2026-08-21, the only `unclassified` sits in the harness's own delegation prompt and the only `assertion-classification` is a snapshot filename. Neither is agent output.
  - Every assertion in every fixture either suite touches already carries both `information_quality` and `evidence_type` — 131 of 131 for proof-conclusion, 105 of 105 for research-exhaustiveness. The decline path the edited sentence governs is never entered.
  - The research-exhaustiveness rubric grades three dimensions — threshold reasoning, declaration honesty, stop-criteria coverage. None of them scores the classification block or the skill it names.

  So a paid run on either skill would re-baseline a snapshot without measuring the change. The behaviour that genuinely changed is the mentor's `must_address.suggested_skill`, which the mentor spec defines as the skill the orchestrator routes to in autonomous mode — and that is the one file no run log watches. The evidence it was a live defect is in e2e, not unit: the mentor emitted the dead name in `hannah-earnest-children` (2026-07-27) and `john-perry-witbeck-vitals` (2026-07-21), both after the 2026-07-11 deletion.

  Splitting the hunks across issue #1801 and issue #1643 is off the table regardless — this repo does not split a PR to match issue boundaries.

- [ ] No skill `description` frontmatter or DO NOT clause changed, so no negative-routing pair is affected.

## Follow-on issues

Filed from the same review rather than folded in — each for a stated reason, none of them "out of scope":

| Issue | Why not folded in |
|---|---|
| #1824 image-as-original doctrine | Lead decision; also spans two eval slots |
| #1825 opportunistic assertions never relinked to questions | Lead decision on a deliberate design tradeoff |
| #1826 full-text strips `textDocument` from under its own triage step | Different skill (a tool fix, not prose) |
| #1827 init-project blind to node conflation | Different eval slot, currently held by issue #1490 |
| #1828 should full-text ever scope to a collection (icebox) | Lead decision; unwinds a shipped e2e fix |
| #1829 nothing measured what full-text queries and filters search (icebox) | Measurement task that gates issue #1828 |
| #1830 must a reasonably accessible source be searched (icebox) | Lead decision across two contended slots |
| #1831 can a project start from a record (icebox) | Lead decision; collides with a twice-rejected design |

**Folded in / filed instead:** folded in the two fixes and the guard. Findings that belong to a skill's own deep dive went as comments on the card that already owns that directory and already pays for its run — issue #1817 (search-records), issue #1648 (search-external-sites), issue #1798 (search-full-text), issue #1646 (person-evidence), issue #1666 (record-extraction) — rather than as new issues, since a separate card would buy a second paid run each.

One board note that needs no work here: **issue #1325's premise no longer holds.** It rests on `search-records/SKILL.md` banning wildcards; that ban is gone, and the file now reads "Wildcards are supported" and "the old no-wildcards rule was measured and withdrawn". It is holding a contended search-records slot on a refuted claim.


johnmarkpeterbrown (commented):
Let me know when the comment is attended to. Thanks

Issue #1788, opened 2026-08-20
Some grade_on_invariant validators run, pass, and cannot fail — measure how many before fixing


DallanQ:
**Touches:** eval/harness/scripts/ (new lint), eval/harness/harness/runnability.py, eval/harness/validators/*.py, docs/specs/unit-test-spec.md

Found during the convert-dates deep dive (issue #1654, PR #1766). Ruth designed it and
deliberately did not implement it — her reason is in "Why it ships as warn-only" below and
is the most important line in this issue.

## The gap

When a test sets `negative.grade_on_invariant: true`, its whole verdict is handed to a
tag-gated validator. `runnability.py` already blocks the case where that tag reaches **no**
validator — it returns `not_runnable` with a comment naming the failure mode ("passes
VACUOUSLY, green forever, asserting nothing").

Nothing checks one step further. The validator can exist, be correctly tagged, run, and
still be **incapable of failing** — because the thing it asserts about cannot happen. If
the assertion is "tool X was not called" and X is registered nowhere for that test, the
pass proves nothing.

## The two known instances

`ut_convert_dates_003` and `ut_convert_dates_010`, across all four pre-dive runs.
`test_no_spurious_conversion` recorded `"error": null` — it ran, and it passed — while
asserting `convert_calendar` was not called, for a tool that was in neither the test's
`mcp_fixtures` nor `mock_mcp.LIVE_TOOLS`.

`ut_convert_dates_010`'s own `explanation` field asserts the opposite of what was
happening: "grade_on_invariant is safe **only because that validator actually runs and
gates the outcome**; without it the invariant would pass vacuously."

Both are closed as of PR #1766, which registers `convert_calendar` and adds
`test_no_spurious_conversion_now_has_a_registered_tool_to_catch` to prove the validator can
now fail. **Nobody has looked for others**, and the failure mode is silent by construction —
a green test that can never go red leaves no trace anywhere. Two is a floor, not an
estimate.

## Do the measurement first

This issue is a measurement, and only then a fix. Run the check once over the whole corpus
and report the count:

- for every test with `negative.grade_on_invariant: true`, find the tag-gated validator(s)
  that carry its verdict (`tag_gated_validator_tags` in `runnability.py` already does this
  half);
- extract which tool each such validator asserts about;
- compare against the tool set that run can register: the test's `spec.mcp_fixtures` union
  `mock_mcp.LIVE_TOOLS`.

[...]

PR #1833, opened 2026-08-22
search-wikipedia: deep-dive fixes, 6 validators, 3 tests, rubric rewrite (#1662)

EdmondOware:
Deep dive into search-wikipedia per #1662. Followed docs/skill-deep-dive-guide.md:
prohibition list built from the skill body first, transcripts read before any
score, time spent on the quiet passes, every finding placed in a lane, and each
one converted to a validator request where it converts.

Two write-ups land with it. docs/deep-dives/search-wikipedia-prohibition-list.md
is the Step 1 deliverable, saved so the next auditor starts from it.
docs/deep-dives/search-wikipedia-findings-2026-08-22.md is the ten findings in
Did / Should / Gap form with the five validator requests.

The issue's numbers all checked out, which is worth saying because they often do
not: 9 tests, 2 of 5 dimensions never discriminating (Slug correctness scored 3
in all 27 of its evaluations, Tool Arguments only ever 3 or null), and the
prescribed grep returning 0 files. Confirmed rather than assumed.

## What this skill is, and why it shaped the findings

wikipedia_search is not a search. It fetches one page summary by title and
returns exactly one {title, extract, url}. So the skill has three moving parts:
one tool call, one saved file, one sentence. Nine tests and eight validators
covered the first and the third. The second, the file, which is the entire
deliverable, was covered by nothing, and its filename half was covered twice
over while its contents half was not covered at all.

## Findings, and what changed

Nine of ten findings are lane 2 (eval defects). One is lane 4, and it is the
only line added to SKILL.md.

1. ut_search_wikipedia_007's grade_on_invariant verdict rested on
   test_no_wiki_no_write, whose docstring claims it fails on a wikipedia_search
   call. It could not: the test declared mcp_fixtures: [], so the mock advertised
   no tools and the model was never offered one. This is the row #1788 assigns to
   this issue, so this PR carries it, including the spec edit #1788 puts in the
   first of its five PRs.
2. slug-kirchenbuch and slug-naturalization-act-of-1906 gated no validator at
   all. Both written, plus a generic check that derives the slug from the
   returned title and needs no tag, which closes the class rather than the two
   instances.
3. No test in the suite could detect a slug built from the query instead of the
   returned title, which is the exact failure rubric.md named. In all seven
   positive tests, slugifying the query and slugifying the title produced the
   same string, including in the two tests labelled as slug regression checks.
   New test where they differ.
4. Nothing read the saved .md. Validators always could, since after_state
   ["files"] holds the complete text, but the rubric told the judge to grade the
   skill's claim about the file instead. New byte-for-byte template check.
5. That new check went red within minutes on an incoherent stub in the harness's
   own test suite, which wrote a Schuylkill County file while declaring the tool
   had returned {"title": "X"}. Fixed here.
6. Two of the scope guard's three named branches had no negative test, and both
   had been deleted. The locality-guide one is restored; it turns out
   ut_locality_guide_003 has pointed at it since June, so unreciprocated routing
   edges go 45 to 44.
7. Mid-workflow narration reached the user's reply on a run that scored 3 on
   every dimension. The rule is already in step 5 and was ignored, so per the
   lane rule the fix is a grader, not more prose.
8. The harness had warned 33 times across five run logs that this rubric grades
   no tool work. Nothing had acted on it. rubric.md rewritten from 2 dimensions
   to 3; the warning is now gone.
9. The slug rule was undefined for an accented title, and no test could have
   said so because every title in the corpus was ASCII. Settled as ASCII
   transliteration, pinned by a Wurttemberg test. This is the one SKILL.md
   change, three lines and two examples.
10. The previous run log had been inactive on main since 2026-07-29. Checked
    rather than assumed, because the obvious reading is that the whole dive was
    measured against superseded prose. It was not: the only difference is the
    deleted model: pin, so the transcripts do reflect the current scope guard.

## The paid run, and the one decision it overturned

11 pass, 1 partial, $0.74, 113s. All three new tests passed on first execution,
and I checked the arguments rather than the verdicts: p4t genuinely sent "potato
famine" and saved great-famine-ireland.md, and w8m sent "Wurttemberg" and saved
wurttemberg.md. The verbatim file check ran and passed on all nine positives,
which is a real result in itself: the skill was following the rule all along and
nothing could show it.

The suite went from 9/9 green to 11 pass / 1 partial. The partial is finding 7
above, and it is the intended outcome rather than a regression. The reply was
already like this and nothing was looking.

Then the run overturned one call. Reply economy scored the narration 2 on
ut_search_wikipedia_001 and scored a byte-identical reply 3 on p4t, in the same
run, with a rationale quoting a one-sentence reply it had not been given. A check
that fires on half its population and asserts the opposite on the other half is
weaker than either a reliable check or none, so validator request V4 ships after
all as test_reply_does_not_narrate_pending_step. The dimension is kept too, since
it grades the restating-content half no phrase list reaches.

That needed text_response plumbed through to validators, which no validator could
previously read: validator_runner, the orchestrator call site, conftest, and the
contract docstring plus the spec's fixture list. It is now available to every
skill's validators, and several skill bodies state reply-shape rules that only a
judge grades today.

Because V4 post-dates the run, the committed log carries no result for it. On
that run's data it would have failed ut_search_wikipedia_001 and p4t, so the next
run will report two failures instead of one partial until the narration stops.
Flagging that here so it is not read as a fresh regression.

## Two near-misses worth reviewing closely

Both are on CLAUDE.md's list of ways a check silently passes, and both nearly
shipped.

The first version of V4 declared **kwargs to reach text_response. _run_module
marks a validator failed, not skipped, when it declares a parameter the harness
cannot supply, so it would have errored on every run. Caught by reading the
injector instead of trusting the signature.

The first plumbing guard asserted that the string
text_response=result.text_response appears in orchestrator.py. It appears three
times in that module (derive_activated, run_validators, grade), so deleting mine
leaves it green. This is the "field-name match that collides with an unrelated
key" case by name. Replaced with a behavioural test that monkeypatches
run_validators, proven by deleting line 465 and watching it go red. A related
third: my first attempt to break the plumbing was a \n-anchored replace against a
CRLF file, which matched nothing and reported success anyway.

## Verification

Every new check was driven red before being committed, per CLAUDE.md: 21/21,
19/19, 15/15, 16/16 and 6/6 across five provers, covering the two real narrating
replies from this run, all seven titles already in the corpus (unchanged by the
slug rule change), and nine accented German, Polish and Scandinavian titles.

make harness-test: 2573 passed, 3 skipped. ruff clean. 12 tests loading,
schema-valid against unit-test.schema.json, and passing check_runnable. New run
log active (snapshot matches disk). Annotation schema-valid, with the p4t judge
miss recorded as a 3 to 2 correction, which is the calibration evidence for V4.
check_tool_coverage, check_rubric_tool_drift, check_skill_frontmatter and
check_negative_reciprocity all run, with no new warning attributable to this PR
and the reciprocity count improving.

## Conflict to expect

docs/specs/unit-test-spec.md is also touched by open PRs #1774, #1781 and #1804.
My edits there are two self-contained additions, the grade_on_invariant row with
the reachability sentence and the text_response fixture entry, so it resolves by
taking both sides.

Also note eval/runlogs/unit/search-wikipedia/v1_2026-06-12_17-45-48.json is
deleted by the retention rule (newest 5 candidates). Finding 6 cites it for the
evidence that a negative test once passed while the skill called wiki_search and
the judge scored it 1/1. The quotes in the write-up are the surviving record.

## Not in this PR

Three items are in the write-up's handback section rather than filed as issues,
per CLAUDE.md's preference for a comment on an issue that already covers the
work. Comments posted on #1788 and #1755.

- The FamilySearch-wiki boundary branch still has no negative test. Restoring it
  would re-create the duplication that deleted it, so the right shape needs a
  distinct user message, which is a genealogist's call.
- The saved-file check lifted to search-familysearch-wiki, which is #1755
  request 1 and already budgeted a run there.
- translation to search-wikipedia is a one-directional routing edge that
  check_negative_reciprocity already reports on every eval PR.

None can ride this PR's run without re-flipping the snapshot, so they belong to
whichever card next takes this skill's eval slot.

## Who should review

Peer review, developer or genealogist, then senior genealogist sign-off. The
genealogical calls a reviewer may want to second-guess are the potato famine
topic choice (finding 3), restoring locality-guide over FamilySearch-wiki
(finding 6), and ASCII transliteration over preserving accents (finding 9), which
was chosen on portability grounds for a shared research folder. The
developer-side change most worth a careful read is the text_response plumbing and
its guard, for the reason in "Two near-misses" above.


johnmarkpeterbrown (changes requested):
Two wording fixes needed before merge. Both are about what the new fixture is called, not what the code does.

The engineering here is solid. Every new check is genuinely falsifiable — I broke each one and watched it fail. The saved-file check and the tag-free slug check both really ran on all nine positive tests rather than skipping, which is the thing that usually goes wrong with a new validator. The reciprocity count did improve, 45 to 44. The run log matches what you described.

## 1. The spec describes the new fixture as the opposite of what it holds

`text_response` is not the skill's final reply. It is every assistant text block in the run, concatenated with no separator between them. `skill_runner.run_skill` builds it by joining a list that collects each text block from each assistant message.

Your own run log shows the difference. On `ut_search_wikipedia_001` the stored value is:

```
Now I'll write the filled-in template to the file.Saved the Wikipedia summary to `schuylkill-county-pennsylvania.md`.
```

Two blocks glued together — note the missing space after `file.`. The final reply on that run was one clean sentence.

This matters because that spec line is what the next person reads before writing a validator against this fixture. The line even recommends the "One sentence only" use case. Somebody will write a one-sentence check, and it will fail every run where the model narrated mid-workflow, no matter how clean the closing message was.

The same spec already gets this right in three other places, describing it as Claude's full output text.

### `docs/specs/unit-test-spec.md`, line 1140

```
You wrote:      - `text_response` (str) — the skill's final reply to the user, verbatim; the same string the run log stores as `output.text_response` and the judge grades. Empty when the run produced no reply.

Change it to:   - `text_response` (str) — every assistant text block concatenated, no separator: narration and closing reply in one string, not the final reply alone (`"".join(text_chunks)` in `skill_runner.run_skill`). Empty when the run produced no assistant text.
```

I dropped the run-log-equivalence clause rather than correcting it. That claim is the inaccurate part, and line 1376 already documents the 100 KB sidecar case, so it does not need restating here.

### Three one-line fixes

Same sentence, three more places you added it. In each, replace `the skill's final reply to the user` with `every assistant text block concatenated, not the final reply alone`:

- `eval/harness/harness/validator_runner.py`, line 84
- `eval/harness/validators/conftest.py`, line 89
- `eval/harness/validators/test_universal.py`, line 28

## 2. The new check's failure message points a debugger at the wrong text

I think the behaviour is right. In Cowork the user sees every text block, so narration really is user-visible, and failing on it is fair.

But the docstring and the assertion both call the string "the reply". Both runs this will fail ended with a compliant one-sentence reply. Whoever hits the failure will open the final message, see one clean sentence, and conclude the validator is broken.

### `eval/harness/validators/test_search_wikipedia.py`, line 411

```
You wrote:      """The reply reports that the file exists; it does not announce writing it.

Change it to:   """No assistant turn announces writing the file — not just the closing one.

                `text_response` concatenates every text block, so narration fails this
                even when the final sentence is clean. In Cowork the user sees every block.
```

### `eval/harness/validators/test_search_wikipedia.py`, line 440

```
You wrote:      f"was created. One sentence only.' Reply was: {text_response!r}"

Change it to:   f"was created. One sentence only.' The phrase may sit in a mid-workflow "
                f"turn — this string concatenates every text block. Full text: {text_response!r}"
```

## Not blocking

The `SKILL.md` edit is routed correctly. The slug is written by a generic file-write tool, not a writer tool, so this genuinely cannot be a tool precondition. Five lines, instruction plus two examples, no rationale prose. Right call and right size.

Your note about the next run showing two failures instead of one partial is accurate. I checked it against the stored replies and it is exactly two, not more.

The findings doc cites the run log this PR deletes. You already record the pruning further down the same file, so it reads as intentional and I would leave it.

## Two other things

One senior from either team can unblock this, plus a second reviewer from anyone with write access. Every CODEOWNERS rule here names both senior teams as an OR.

On the three genealogical calls you flagged — the potato-famine topic, restoring `locality-guide` over the FamilySearch-wiki branch, and ASCII over preserving accents — I verified the transliteration is *implemented* consistently with what `SKILL.md` states. I did not judge whether ASCII is the right choice. That one still needs the senior genealogist.

johnmarkpeterbrown (approved):
Looks good!

eval/harness/validators/test_search_wikipedia.py
```diff
@@ -155,3 +158,288 @@ def test_slug_obrien_surname(before_state, after_state, test):
...
+    r"""(?ix)
+    \b(?:
+        now \s+ i(?:'|\u2019)?ll                      # Now I'll / Now Ill
+      | now \s+ i \s+ will
+      | i(?:'|\u2019)?ll \s+ now
+      | i \s+ will \s+ now
+      | i(?:'|\u2019)?m \s+ going \s+ to
+      | i \s+ am \s+ going \s+ to
+      | (?:next|then) ,? \s+ i(?:'|\u2019)?ll
+      | let \s+ me \s+ (?!know\b)                     # "Let me fill in…" but not "Let me know…"
```
mercyokum: ## 1. `_NARRATION_RE`'s "let me" exclusion is too narrow

`eval/harness/validators/test_search_wikipedia.py`, the new `_NARRATION_RE` regex:

```python
| let \s+ me \s+ (?!know\b)   # "Let me fill in…" but not "Let me know…"
```

Only `"let me know"` is excluded. Any other `"Let me ..."` opener gets flagged as narrating a pending step, even when it isn't. Reproduced directly:

```python
>>> _NARRATION_RE.search("Let me be clear: the file has been saved.")
<re.Match object; span=(0, 7), match='Let me '>
```

That reply reports a *completed* action, not a pending one, but `test_reply_does_not_narrate_pending_step` would hard-fail it as a false positive. Suggest widening the exclusion to require a following verb that actually signals an upcoming write (`write|fill|save|create|check`) rather than excluding only `"know"`.

There's also a gap in the other direction: `"I'm now going to check something else."` matches none of the branches, since none allow a word between `"I'm"/"I'll"` and `"now"`/`"going to"`.

I haven't seen either misfire in the committed run log — this is a demonstrated latent defect in the regex, not an observed regression. It's eval-suite tooling, so worst case it wrongly fails a correct run and costs someone time chasing a phantom regression — not something that reaches a real user.

mercyokum (changes requested):
## 2. `conftest.py` doesn't supply three fixtures nearly every validator in this file needs

`validator_runner.py`'s `available_args` (the real harness injection path) includes `test`, `skills_invoked`, and `attempted_mcp_calls`, but `eval/harness/validators/conftest.py` only defines fixtures for `before_state`, `after_state`, `tool_calls`, `blocked_context_calls`, `blocked_protected_writes`, `skill_frontmatter`, and the new `text_response`. This PR touches `conftest.py` to add `text_response` but doesn't close the other three gaps.

Almost every validator in `test_search_wikipedia.py` — all the `test_slug_*` checks, `test_saved_file_matches_template`, and the new `test_reply_does_not_narrate_pending_step` — declares `test`. Ran the documented debug command to confirm:

```
$ uv run pytest validators/test_search_wikipedia.py -v
...
15 errors — every function in the file: fixture 'test' not found
```

That's the exact command `docs/specs/unit-test-spec.md` §8 documents ("Developers can also run validators standalone with `pytest eval/harness/validators/ -v` for debugging"), and it's `conftest.py`'s own stated reason for existing — currently silently unusable for this file.

Two things scope this down:
- **Pre-existing**, not introduced here — the gap predates this PR; it just had the file open for a related fix and didn't close it.
- **Doesn't touch CI or real eval runs.** `eval/harness/pyproject.toml` sets `testpaths = ["tests"]`, so `make harness-test` never collects `validators/` via bare pytest, and the actual harness path (`validator_runner.py::_run_module`) builds its own arg dict and calls validators directly — no pytest fixtures involved.

Suggest adding the three missing fixtures to `conftest.py` — quick to do while the file's already open for this PR, or as an immediate follow-up.

**Recommendation:** not blocking. Both are small, well-scoped fixes in eval tooling, not shipped-behavior problems.

eval/harness/validators/test_search_wikipedia.py
EdmondOware: Both fixed in `0165f852`.

**The regex.** You were right that excluding one word was the wrong shape, so the branch now requires a verb naming work this skill does rather than trying to enumerate the benign openers, which has no end. The adverb slot closes the other direction. Nine cases are pinned, including both of yours:

- `"Let me be clear: the file has been saved."` no longer matches
- `"I'm now going to check something else."` now does

**The fixtures.** `test`, `skills_invoked` and `attempted_mcp_calls` are in `conftest.py`, and the docstring now states the rule that every key in `available_args` needs one, with the reason it fails quietly: `pyproject.toml` pins `testpaths` to `tests`, so CI never collects that directory. `pytest validators/test_search_wikipedia.py` goes from 15 errors to 15 skipped.

Your item pointed at a bigger one a layer up, which is also fixed here. Eleven other validator files import `validators_lib`, which `validator_runner` resolves by putting the validators directory on `sys.path`. Standalone pytest does not, so `pytest validators/` died at collection on all eleven, on the same command section 8 documents. `conftest` now does the same insert, and that command goes from 11 collection errors to 6 passed and 173 skipped.

Nothing under a run log snapshot changed, so no run was bought.

mercyokum (approved):

PR #1838, opened 2026-08-22
Not being in a research project returns an answer, not an error (#1695)

DallanQ:
## Summary

Normal tier. Closes issue #1695. Phase 1b of the enforcement-layer programme.

- A user who ran a search in a folder with no project got `research.json not found in projectPath`. The search ran and the results came back; only the record of it died with that line. Twelve tools did this, and 19 of the 21 skills declaring one would have surfaced it raw.
- The verdict now comes from the **directory**, not from whichever file the caller happened to want — six of the twelve read `tree.gedcomx.json` first, so a file-derived verdict hands those six the wrong message in an empty folder. Five states, of which exactly one is the friendly answer.
- The three eval-harness detectors that read `is_error` to mean "this call never landed" mirror it, because a no-project write that kept `isError` would be counted as landed and would manufacture violations in paid grading.

## Review guidance

**Start here:** `classifyProjectPath` in `packages/engine/mcp-server/src/utils/project-io.ts`. Two things there carry the whole decision and both are one line.

First, it takes no `filename` argument. That is deliberate and it is the fix for the read-order problem above.

Second, `"project"` covers *exactly one file present* as well as both. A folder whose `research.json` was deleted still holds a real project, so a write against it stays loud; softening that row drops the write with a cheerful message. This is the row worth attacking — it runs in both directions, and the naive version (key on `research.json` alone) passes a suite that only tests empty folders.

The other place worth a look is `did_not_land` in `eval/harness/harness/skill_invocation.py`, and specifically the *third* caller of it. Two callers use it to skip a call. `find_relationship_writes_without_warnings_check` uses it to withhold **credit** — there a successful `person_warnings` means the tree was checked, so a no-project call must not count as consulting the guardrail. Backwards there is a missed violation, and therefore silent.

**Settled in review:** `did_not_land` stays a bare-token match. The reviewer went looking for a false positive and could not build one — nothing in the committed corpus carries the token, and the vector he tried (a project folder named for it) does not reach `response_summary`, because `filesWritten` holds bare filenames rather than paths. The counter-test on `"no project needed"` stays.

## Plan

**Didn't change:** no skill bodies — the issue is explicit that the error must stop reaching users without them, and per-skill prose would be ~19 drifting copies of one rule and ~19 paid eval runs. No auto-seeding: a writer tool creating a project on first write was designed twice and rejected twice, because it lets any skill bring an objective-less project into being that `init-project` then refuses to touch. `merge_warnings` keeps its `ok: false` meaning "the merge would be rejected". `validate_research_schema` stays out — asking it to validate a project that does not exist is a genuine error, not standalone use. `rank_search_matches` stays out on inapplicability: it needs a `subjectId` that exists in the tree, so it has no standalone caller.

**Acceptance check:** `packages/engine/mcp-server/tests/tools/no-project.test.ts`, 57 tests. Watched to fail three ways, not one: disabling the discriminator reddens the 12 no-project cases while all 44 loud ones stay green; the naive file-keyed verdict is caught by the eight research-reading tools on the tree-present row; and reverting the `EACCES` distinction reddens the unreadable-directory case. On the harness side, reverting `did_not_land`'s bare-name match reddens the `[mcp-envelope]` parametrization while `[unwrapped]` passes.

**Deviated from the plan:** three, all recorded in `PLAN.md`, none touching the design. Four of the five `read-failure.test.ts` assertions needed no edit at all rather than a changed message — a half-project reaches the existing message verbatim, so 17 of that file's 18 tests are untouched. Two tests outside the plan's file list also set up bare directories and were pinning the old conflation (`research-query.test.ts`, `tree-forget.test.ts`). And `merge-gedcomx-spec.md` needed one of the three edits the plan named, not three — the other two describe validation failure, which this does not touch.

## Test plan

- [x] I ran `make test-all` and it passed. 2577 passed, 2 skipped.
- [x] No skill's run-log snapshot changed — no skill body, plugin agent, unit test, scenario or fixture is in this diff — so no `make eval-skill` run is owed.
- [x] No skill `description` frontmatter or DO NOT clause changed, so no routing pair is affected.

Also exercised by hand, since the whole point is what a person reads: `npx tsx dev/try-research-append.ts` walks all five states and prints each message. Added in this PR for that purpose.

## Follow-on issues

Issue #1840 — derive the `CALLS` roster in `no-project.test.ts` instead of hand-maintaining it, so the new architecture bullet stops being the only thing standing between a fourteenth project-reading tool and this same bug with CI green. Labelled `nothing-checks`. Not folded in: the derivation is the work, and a one-hop import grep is wrong in both directions — it misses `extraction_append` and `tree_correct`, which reach the read through another tool module, and it picks up `merge-shared.ts`, which is not a tool.

**Folded in / filed instead:** everything found was folded in. `/code-review high` returned three findings and all three are fixed in the third commit — the highest was that the harness mirrors never fired in production, because I matched the quoted key `'"no_project"'` while the e2e orchestrator passes short responses through verbatim as the raw MCP envelope, where the document is an escaped string and a backslash sits where the closing quote would be. `/check-drift` found a missing test at the replay-provenance call site, fixed in the second commit. One thing is deliberately left alone rather than filed: the anomaly behind this issue's evidence is that `research_append` was once called with no arguments at all despite its schema marking `projectPath` required. The agent retried correctly one call later so nothing was lost, and this branch makes it legible (`projectPath is required` instead of a not-found message) rather than diagnosing it. Not attached to issue #1317, which is closed and is about a nested `research.json`.


packages/engine/mcp-server/tests/tools/no-project.test.ts
```diff
@@ -0,0 +1,269 @@
...
+  writeFile(join(dir, "research.json"), JSON.stringify(minimalResearch, null, 2));
+const writeTree = () =>
+  writeFile(join(dir, "tree.gedcomx.json"), JSON.stringify(minimalTree, null, 2));
+
+/**
+ * Every tool, called against `projectPath` with arguments that reach the
+ * project read. Each call is otherwise valid — a tool that rejected its own
+ * arguments before reading would prove nothing here.
+ */
+const CALLS: Array<{ tool: string; call: (projectPath: any) => Promise<any> }> = [
```
johnmarkpeterbrown, line 81: This list is twelve hand-written entries and nothing derives it — and `docs/architecture.md`'s fifth-site bullet, which is where someone goes when adding a tool that returns `{ok: false}` instead of throwing, wasn't updated either. So a thirteenth project-writing tool routes through `writerToolResult`, gets added to `OK_FALSE_IS_FAILURE`, passes every suite in the repo, and returns `research.json not found in projectPath` in a bare folder. That's #1695 again with CI green.

Please append a sixth-site bullet after `rejected, is the only such tool today.`:

```
- **If your tool reads the project files** — a sixth site, and nothing will
  tell you if you skip it. Catch `NoProjectError` from `readProjectJson` and
  return `noProjectResult()` (`"read"` for a read or a preview), so a user who
  is not in a research project gets an answer rather than `research.json not
  found in projectPath` (issue #1695). Then add the tool to `CALLS` in
  `tests/tools/no-project.test.ts` — that list is hand-maintained and nothing
  derives it, so a tool left out is uncovered.
```

The lint that would make this bullet unnecessary deserves its own issue with the `nothing-checks` label. Don't fold it in here.

packages/engine/mcp-server/tests/tools/no-project.test.ts
```diff
@@ -0,0 +1,269 @@
...
+});
+
+describe("a real project directory that cannot be read", () => {
+  // `access()` failing is not the same as a file being absent. A project folder
+  // that lost its execute bit — restored from a backup, copied from a
+  // restrictive archive, an odd sandbox mount — still stats as a directory while
+  // every probe inside it throws EACCES. Read as "absent" that becomes
+  // no_project, and a write against a genuine project is dropped with a
+  // cheerful message: the silent loss the half-a-project rule exists to prevent.
+  it("stays loud rather than claiming the folder is not a project", async () => {
```
johnmarkpeterbrown: This is vacuous on Windows. `chmod` there doesn't clear a directory's read bit, so `access()` succeeds and the test never reaches the `EACCES` branch. I reverted your `fileState` distinction and ran the file: 57/57 still green. It passes because `{id: "src_001"}` fails schema validation, not because the branch works.

Change it to:

```ts
// Skipped on Windows: chmod there does not clear a directory's read bit, so
// access() succeeds and this never reaches the EACCES branch — it would pass
// whether or not the distinction exists.
it.skipIf(process.platform === "win32")("stays loud rather than claiming the folder is not a project", async () => {
```

Related: that leaves your acceptance line for the `EACCES` distinction unproven on any platform in this review. Worth confirming once on Linux that reverting it really does redden. The reasoning is sound; it's the evidence that has a hole.

eval/harness/tests/unit/test_skill_invocation.py
```diff
@@ -238,6 +238,43 @@ def test_flags_the_read_and_improvise_bypass_shape():
     assert violations[0]["required_skill"] == "person-evidence"
 
 
+def _no_project_summary(escaped: bool) -> str:
+    """The two shapes `response_summary` actually arrives in.
+
+    `escaped=True` is the MCP envelope the e2e orchestrator passes through
+    VERBATIM for any response under 500 chars — which the ~131-char no-project
```
johnmarkpeterbrown: `131` matches nothing measured — the fixture in this file is 66 chars, and the real response is 201 (242 enveloped, 255 for the read variant). Your docstring in `skill_invocation.py` already has the right figures.

Change it to:

```
    VERBATIM for any response under 500 chars — which the no-project response
    always is, at 242 chars enveloped (255 for the read variant), making this the
    DOMINANT production shape.
```

johnmarkpeterbrown (changes requested):
Nice piece of work — the directory-not-the-filename call is the right one, and the read/write message split is a detail most people would have skipped. I read the diff, ran both suites, and broke two of your guards to check they fire. They do. Behaviourally this does what it says.

Three things, in three inline comments. Only the first has consequences.

On your "Unsure about": leave `did_not_land` bare. I went looking for the false positive and couldn't build one — nothing in the committed corpus carries the token, and the vector I thought might leak it (a project folder named for it) doesn't reach `response_summary`, because `filesWritten` holds bare filenames rather than paths.

packages/engine/mcp-server/tests/tools/no-project.test.ts
DallanQ: Added in fdc2da09, your text verbatim except for `(issue #1695)`. That made it the twentieth issue reference in `docs/architecture.md` and `doc-links.test.ts` ratchets them per file, so it reddened — which is the ratchet working. The bullet states the fact without it.

Lint filed as issue #1840, `nothing-checks`, not folded in. Worth knowing before someone picks it up: the derivation is the whole job. A one-hop grep of `readProjectJson` / `classifyProjectPath` imports is wrong in both directions — it misses `extraction_append` and `tree_correct`, which reach the read through another tool module, and it picks up `merge-shared.ts`, which is not a tool.

packages/engine/mcp-server/tests/tools/no-project.test.ts
DallanQ: Done, with your comment.

On the evidence hole — proved it on darwin rather than Linux, same POSIX behaviour for this. Reverting `fileState`'s three-way return to a flat `"absent"` reddens exactly one test, this one: `an unreadable project must not be reported as no_project`, received `"no_project"`. The other 56 stay green. So the distinction is measured now, not just argued.

eval/harness/tests/unit/test_skill_invocation.py
DallanQ: You are right that 131 is wrong. So are 242 and 255 — I put those in `skill_invocation.py` a commit earlier and they were never measured against the real path.

Ran the real one: `_summarize_tool_response` over `writerToolResult(noProjectResult())`, both variants. **236 chars enveloped, 248 for the read variant.** Both verbatim against `_RUNLOG_VERBATIM_MAX` of 500.

Fixed both sites to 236/248 rather than the numbers you quoted. Your conclusion is unchanged and the margin is a little wider than either figure claimed.

Issue #1837, opened 2026-08-22
Refuse a tree write until every warning it introduces carries a reason to ignore


DallanQ:
> **Ruling (lead, 2026-09-10): ship as designed, and the gate recomputes
> `person_warnings` predicates only.** The place-resolution class raised in the
> 2026-08-25 comment is deliberately out of scope — "Scope: which warnings"
> below says why, and where it goes instead. Two justifications in this body
> rested on the doctrine-gate override ADR-0011 retired on 2026-08-24; both are
> rewritten. Nothing about the design changes.

**IMPORTANT**: Do #2333 first — it deletes backupIfExists from the same five write paths.

**Touches:** packages/engine/mcp-server/src/tools/research-append.ts, packages/engine/mcp-server/src/tools/tree-edit.ts, packages/engine/mcp-server/src/tools/materialize-facts.ts, packages/engine/mcp-server/src/tools/extraction-append.ts, packages/engine/mcp-server/src/tools/tree-correct.ts, packages/engine/mcp-server/src/tools/merge-tree-persons.ts, packages/engine/mcp-server/src/tools/person-warnings.ts, packages/engine/mcp-server/src/validation/validator.ts, docs/specs/schemas/research.schema.json, packages/schema/schemas/research.schema.json, docs/specs/guardrail-enforcement-spec.md, packages/engine/mcp-server/src/validation/tree-shape.ts, eval/harness/validators/test_universal.py, docs/specs/simplified-gedcomx-spec.md

`person_warnings` is the cheapest guardrail in the system — no LLM, no network,
~75 deterministic predicates over a person and their one-hop relatives — and
**58 of 156 committed e2e runs wrote a new ParentChild/Couple relationship
without ever calling it.** The worked failure is in issue #1193: a run wrote
parent links implying the subject bore her first child at age 6, and never asked
the tool that has a predicate for exactly that.

## The design

A tree write carries a list of `(warning, reason to ignore)` pairs. The writer
recomputes the warnings itself and refuses the write if it finds one the caller
did not acknowledge, returning the unaddressed list.

1. Caller writes, optionally with acknowledgements.
2. Writer recomputes warnings on the post-write tree.
3. Every warning it finds is in the caller's list → the write lands.
4. Otherwise refuse, returning the warnings that still need a reason.

**Any reason is accepted, whatever the severity.** The gate enforces
*consultation*, not adjudication — a contradiction with a reason is allowed
[...]

PR #1839, opened 2026-08-22
eval: replay the post-hoc shadow checks over history, not just what runs stored

DallanQ:
## Summary

Normal tier.

- The shadow report read `guardrail_shadow_violations` as each run recorded it, so every check read 0 over every run made before it shipped. All three post-hoc checks landed in August against a corpus that is 84% July, and their zeros were being read as "this behaviour never happens" — including in the guardrail spec's own pipeline table.
- `--replay` now recomputes all four post-hoc families from each run's committed `final-research` / `final-tree` sidecars, which every committed run carries. Two of the three zeros were the corpus's age, not the behaviour.
- Corrects every stale figure in that spec table rather than refreshing three rows beside four stale ones, and the docstring that asserted these families could not be replayed at all — giving as its reason the very property that makes them replayable.

The measurement, over 157 runs:

| check | report said | replayed |
|---|---|---|
| citation-nulling | 0 | **0**, of 157 scanned |
| conflict-unpersisted | 0 | **4 runs**, of 157 scanned |
| warnings-unchecked | 1 | **58 runs**, of 156 scanned (1 named skip) |

Only citation-nulling is a genuine zero, and it now has its positive control on the live path (see below). Warnings-unchecked has also fired live and stored once, in `stribling-father-1821/run-2026-08-17_23-35-44`, which exercises its whole chain.

These are **behaviour-presence** counts, not per-run compliance scores — `e2e-test-spec.md` withholds the latter because checks are not pinned to the version each run executed, and `architecture.md` says not to quote a violation rate. Counts against named denominators throughout. One check is affected by that version gap: the conflict-unpersisted predicate was corrected after it shipped, so the replay measures today's rule. The spec now says so.

## Review guidance

**Start here:** `replay_post_hoc` and `RunInputs.missing_for_research_only` / `missing_for_warnings` in `eval/harness/e2e/guardrail_shadow_report.py`. The decision worth checking is that each check carries its **own** denominator and skip list rather than sharing one. Two of the three read only the research sidecar, so the corpus's one orphan run log — `william-ferber-ancestry`, a committed run with no fixture directory — is skipped by name for warnings-unchecked and still scanned by the other two. A shared skip list would drop it from all three and discard anything it held. Also worth a look: those requirements are expressed as field reads, not string matching on a skip message; an earlier draft matched one message's exact wording, which would have silently stopped excluding if the message were reworded.

**Unsure about:** nothing blocking. One judgement call I made and documented rather than resolved: `runs_scanned` counts runs *read*, not runs that *could fire*, which is a weaker claim than the sibling `replay_provenance` makes with `runs_linking`. These are presence numbers rather than conditional rates, so I documented the weaker claim in the dataclass instead of restructuring. If you'd rather they matched, that's a small change.

## Plan

**Didn't change:** the three detectors in `harness/skill_invocation.py` — the replay must run the current implementations verbatim or it stops being a measurement of them. Also `axes_from_runlog` and the compliance axis (no run is retroactively scored), and `detector_before_after_report.py`, which answers old-implementation-vs-new rather than stored-vs-replayed. Nothing graduates from shadow to a hard check and no deny ships.

**Acceptance check:** `cd eval/harness && uv run pytest -q tests/unit/test_guardrail_shadow_report.py tests/unit/test_post_hoc_shadow.py tests/unit/test_strip_e2e_captures.py` — of which 12 pin the replay plumbing, 9 pin citation-nulling's live path, and 6 pin the strip's replay remnant. The first 12 and fail on `main` because `replay_post_hoc` does not exist there. They are controls for the replay **plumbing** (sidecar resolution, seed-tree loading, per-check skip discipline), not for the detectors, which already have firing predicate controls in `test_skill_invocation.py`. Each was proven to fail by breaking the plumbing five ways and watching the right test redden: wrong sidecar suffix, seed tree ignored, per-check skip lists collapsed, the `main()` print unwired, and the loader's non-dict guard removed. Also `npx vitest run tests/packaging/doc-links.test.ts`, because the spec sits under an issue-reference ratchet that fails in both directions; it holds at 16.

**Deviated from the plan:** three, all recorded in `PLAN.md`. The plan had `replay_provenance` sharing the new `load_run_inputs`. It does not: that function needs two of the four inputs, so routing it through would parse ~19 MB of sidecars it never reads — twice, since `--replay` now runs both replays — and would lose its exception-type diagnostic. Recorded in `PLAN.md` and in the loader's docstring. No step-4 stop rule was hit.

## Test plan

- [x] I ran `make test-all` and it passed. 2583 passed, 2 skipped.

- [x] No skill's run-log snapshot changed, so no paid eval run applies. The snapshot covers a skill's own directory, its delegated agents, its `eval/tests/unit/<skill>/` and the scenarios and fixtures it references; this PR touches `eval/harness/e2e/`, the Makefile and docs, none of which is in any skill's snapshot.

- [x] Not applicable — no skill `description` or DO-NOT clause changed.

## Follow-on issues

Issue #1837 — refuse a tree write until every warning it introduces carries a reason to ignore.

**Folded in / filed instead:** Everything the review and drift passes turned up was folded in: the `main()` seed-lookup hole (a parameter default binds at import, so the test that pinned this wiring passed while warnings-unchecked silently scanned nothing), a formatter that printed a violation count under a "run(s)" noun, a loader that accepted non-dict JSON and let `UnicodeDecodeError` escape, the fixture-drift caveat, and two documentation duplications. Issue #1837 was filed rather than folded on two of the four exemptions, both named in its body: it **needs a different reviewer and suite** — this change is the eval harness under `make harness-test`, that one is engine writer tools shipping in the `.mcpb` under vitest — and it is **too big, counted rather than asserted**: six writer tools, a new validation module, and a schema change across five mirror sites.

Issue #1431 has since been run and closes with PR #1844, which carries both graded runs. This PR is what made its case: replayed, that check reaches 120 of the 147 runs that link a person, against the 7-across-5 the spec quoted before.

---

**Update, 2026-08-23.** Two things were folded in on the lead's instruction rather than deferred, so this PR now delivers all of Phase 5 rather than most of it:

- **citation-nulling's positive control**, the last thing Phase 5 owed. It was unreachable because the check ran inline in `_run_agent`, a coroutine needing the SDK and a live model, so `collect_post_hoc_shadow` was extracted — behaviour unchanged — and `tests/unit/test_post_hoc_shadow.py` drives it from a hand-built research.json on disk. That path is where `read_research_json` returning None on a corrupt file is indistinguishable from a clean project, which is exactly the ambiguity this phase exists to remove.
- **The e2e capture strip was destroying `harness/replay.py`'s input.** It drops `response_summary` past 14 days; replay reads the `entryId` each writer reported back out of that field. Measured: 133 of the 134 stripped runs can no longer be replayed, against 18 of 23 unstripped ones that can — an 88% → 13% collapse in `make replay-check` that looked like a regression in an engine nobody had touched. The strip's own docstring is why it went unnoticed: it claimed the field had no programmatic reader and enumerated four that don't read it, missing replay.py. It now keeps a replay remnant — ids, `ok`, batch length — measured at 0.4% of the summary bytes against the 24% of a run log those summaries occupy. **Forward-only:** the 134 already stripped are not recoverable.

Also records, at the table it applies to, why warnings-unchecked was considered for graduation and declined — that decision previously lived only in a gitignored plan file.


johnmarkpeterbrown (commented):
Nice work — the numbers all check out and the new tests actually fail when the plumbing breaks. One thing to fix before merge.

**The fixture-drift caveat names five fixtures. Only one of them qualifies.**

I checked every fixture with committed runs whose seed tree changed after its earliest run — 18 of them — and compared the actual relationship key set at run time against today's. Only `teitje-harkema-parents-1833` changed (1 key → 26). Your "+25 the day after its only committed run" is exact.

The other four don't hold:

- `crowder-grandparents` and `scotland-thomson-grandparents` have no committed run logs, so the replay never reads them.
- `young-marriage-1828` — seed finalised 07-09, runs are 07-21 and 07-22. Key set identical.
- `anders-monsen-ancestry` — seed finalised 07-08, runs start 07-09. Key set identical.

Your conclusion is right. I confirmed teitje fires 0 on both its run-time seed and today's, so no count changes. The list just overstates the hazard fivefold.

In `replay_post_hoc`'s docstring:

```
You wrote:      in the current checkout, and five committed fixtures have had their
                relationship key set changed after runs against them already existed
                (`crowder-grandparents`, `scotland-thomson-grandparents`, `young-marriage-1828`,
                `anders-monsen-ancestry`, `teitje-harkema-parents-1833` — teitje gained 25
                relationship keys the day after its only committed run). Replaying all runs
                whose run-time seed blob is recoverable changed no count as of 2026-08-22, so
                this is a live hazard rather than a current error.

Change it to:   in the current checkout. Of the 18 fixtures with committed runs whose
                seed-tree file changed after their earliest run, exactly one changed its
                relationship key SET: `teitje-harkema-parents-1833`, which gained 25 keys
                (1 -> 26) the day after its only committed run. That run fires on neither
                seed, so replaying changed no count as of 2026-08-22 — a live hazard rather
                than a current error.
```

**Optional, and already broken on main.** On a Windows console, `make e2e-guardrail-shadow REPLAY=1` crashes on the `≥` in `format_provenance_replay` before printing any of your new output. So "read the replay before concluding a check never fires" can't be followed on Windows today. You're already touching `main()`, and `author.py` has the house pattern:

```
Add to main():   if hasattr(sys.stdout, "reconfigure"):
                     sys.stdout.reconfigure(encoding="utf-8")
```

Your call whether that belongs here.

Issue #1687, opened 2026-08-17
flynn-record-matching README states the verdict, and the judge reads it in full


DallanQ:
**Touches:** eval/fixtures/scenarios/flynn-record-matching/README.md, eval/tests/unit/person-evidence/*.json, eval/tests/unit/search-full-text/*.json, eval/tests/unit/search-records/*.json

Found while reviewing PR #1684, which fixes the same defect in `judge_context` and in a `user_message`. This is the third channel and the widest one.

## The mechanism

A scenario's `README.md` is read whole by `_load_scenario_readme` and dropped into the **judge** prompt at `eval/harness/judge/prompt.md:216`, under the heading `## Scenario summary`. Nothing filters it. So a README bullet phrased as "the skill must do X" is an answer key handed straight to the grader — a stronger leak than a judge note, because it is not scoped to one test.

## What this README says

`eval/fixtures/scenarios/flynn-record-matching/README.md`, "What it exercises":

> a high score must not auto-link past the `log_002` birthplace conflict, and the low score on the `log_003` variant **must not dismiss a strong qualitative match**.

That is the verdict for both records, stated to the judge before it reads the response.

The record inventory above it does the same work qualitatively:

> `log_001` — clean 1850-census match (`MXHY-TP4`) … **Strong on every identifier.**
> `log_003` — variant record (`VRNT-7M3`) … **Strong on every identifier *except* the spelling.**

"Strong" and "clean" are the assessments the tests exist to measure.

## Why it is not fixed in PR #1684

That PR fixes the `mullen-price-multiple-marriages` README in the same pass, because that scenario is referenced by `search-records` tests only and rides a run already owed. This one is shared:

| skill | tests referencing the scenario |
|---|---|
| search-records | 8 |
| person-evidence | 6 |
| search-full-text | 1 |

Editing the README invalidates the snapshot of all 15, so it costs **three** paid runs and three annotation passes. `search-records` is already owed in PR #1684; `person-evidence` and `search-full-text` are two extra. `eval-cosmetic-skip` does not apply — `eval/CLAUDE.md` reserves it for behavior-neutral edits, and this changes what the judge is told.

**Filing exemption (CLAUDE.md):** it is a different skill's eval slot, and bundling it would force a second paid run — two, here.

## Suggested fix

Rewrite "What it exercises" to name the *capability under test* rather than the correct answer, the way PR #1684 rewrites the mullen one. Roughly:

[...]

PR #1842, opened 2026-08-23
research-exhaustiveness: the subject's own birth record is a decisive parentage record (#1801)

### LINKED ISSUE #1801: research-exhaustiveness: the decisive-record gate for parentage omits the subject's own birth/civil-registration record
**Touches:** packages/engine/plugin/skills/research-exhaustiveness/SKILL.md, eval/tests/unit/research-exhaustiveness/

## The gap

`SKILL.md`'s named-decisive-record rule lists, for a parentage question, only:

> parentage: the subject's death record, baptism, or a parent's probate

**The subject's own birth record or civil registration is missing.** In an era with civil registration that is normally the single most direct parentage source available — and more reliable than a death record, whose parent names come from a secondhand informant at the worst possible moment.

As written, the gate can be satisfied while the most direct record of all goes unsearched, and the skill will declare exhaustive over it.

## Why this is filed rather than fixed in the PR that found it

Surfaced by @EdmondOware reviewing PR #1796 (the #1645 deep dive), whose reasoning this issue is essentially quoting.

Two of CLAUDE.md's exemptions apply:

- **A different paid eval slot.** Editing `SKILL.md` invalidates the run log PR #1796 just committed and annotated, buying a second `make eval-skill SKILL=research-exhaustiveness` run plus another annotation pass. That PR is prose-only past its run and cannot absorb it.
- **A different lane and reviewer.** This is a lane-4 doctrine edit to a stewarded skill body, gated by the unit suite — not the eval-grading lane PR #1796 works in.

## What to decide

1. Add the subject's birth/civil-registration record to the decisive-record list for parentage, and settle whether it *outranks* the death record when both exist (Edmond's informant-reliability argument says it does).
2. Decide how the gate behaves where civil registration does not exist for the place and period — the list should not fire on a record that could not have been created. `recent-birth-sealed` is the worked case in the other direction: there the birth certificate exists but is sealed, and the body already handles that via the privacy-sealed exception.
3. Consider whether a test pins it. No current test in the suite exercises a parentage question where the subject's own birth record is accessible and unsearched.

## Provenance

Item 15 of the deep dive's prohibition list (PR #1796 body) quotes this record list verbatim from the body, and was flagged there as "drafted, not confirmed by a genealogist" for exactly this reason — the record set needed a genealogist's eye. It got one.

benter-070:
## Summary

**Tier: Normal.** Closes #1801.

`research-exhaustiveness`'s named-decisive-record rule listed, for a parentage question, only *"the subject's death record, baptism, or a parent's probate"*. In a civil-registration era the subject's **own birth record** is normally the most direct parentage source there is — so the gate could be satisfied while the most direct record of all went unsearched.

**The rule was internally inconsistent, not merely incomplete.** It already named a birth certificate in its *inaccessibility carve-out* (*"a recent birth certificate embargoed ~100 years"*) while omitting the record from the list that triggers the gate. And `ut_research_exhaustiveness_017` exists precisely to test that a **sealed** birth certificate is not a gap — a test that only makes sense if an accessible one would gate. The corpus already treated the record as decisive; the list had not caught up.

## Review guidance

**Start here:** the `Named decisive records gate the declaration` bullet in `research-exhaustiveness/SKILL.md`. It does three things, and the middle one is the judgement call:

1. Adds the birth record to the parentage list, **era-conditioned** — *"where the jurisdiction and period kept one"*.
2. **Ranks it above the death record**, with the reason stated: *"its informant is a parent at the event rather than a survivor recalling it decades later."*
3. States the exclusion explicitly — Irish births before 1864, most jurisdictions before the nineteenth century — and names the **baptism** as what to gate on instead.

**The era condition is required, not decoration.** Patrick Flynn (b. ~1845, Ireland/Pennsylvania) is the subject of eleven scenarios and predates registration in both jurisdictions, so an unconditional rule would demand a record that never existed. That is also why **`ut_research_exhaustiveness_015`'s description is deliberately left alone** even though it quotes the old list: for an 1845 subject the birth record genuinely is not in the gate list, so the quote is still correct there. I had planned to update it and was wrong.

**Unsure about:** nothing on the doctrine. One thing for a genealogist's eye — the new fixture leaves the mother, Ellen Sullivan, resting on a single secondary source. I did not design that in; the skill noticed it unprompted and flagged it as a weakness. It reads to me as realistic rather than a defect, but if you think it muddies the test I will add a corroborating source.

## Plan

**Didn't change:** `_015`'s description (above), and nothing outside `research-exhaustiveness`. The new scenario is referenced by exactly one test, so it invalidates one run log rather than a corpus of them — deliberately unlike the 15-test blast radius that makes #1687 expensive.

**Acceptance check:** `ut_research_exhaustiveness_018` on the new `ma-birth-record-unsearched`. **No existing fixture could host it**, which is why a new one exists:

| candidate | why not |
|---|---|
| `flynn-decisive-record-unsearched` | b. ~1845 Ireland/Pennsylvania — predates registration, so a birth record correctly is *not* in the gate list |
| `recent-birth-sealed` | b. 1936 Utah — privacy-embargoed, so the inaccessibility carve-out correctly applies |
| `richardson-parentage-not-exhaustive` | b. 1833 England (pre-1837) **and** four items still `planned`, so the plan-completeness rule fires first |

The new scenario: Ida F. Braman, b. ~1875 Taunton, Massachusetts — statewide registration from 1842, returns public. Five-item plan, **all completed**; every other decisive type worked (death registration positive, baptism nil because the register does not survive, candidate father's probate nil); **her own birth registration in neither `plans` nor `log`.**

The death record is deliberately **positive and names both parents**. That makes the question look answerable and forces the test to isolate the ranking clause — a nil death record would have let it pass on an obvious gap instead.

**Deviated from the plan:** the issue's `**Touches:**` names `SKILL.md` and `eval/tests/unit/research-exhaustiveness/` but not `eval/fixtures/scenarios/`. Authoring a scenario is therefore an expansion, taken because the paid run was owed for the `SKILL.md` edit regardless — so the test cost effort, not money — and because a doctrine clause with nothing pinning it is what #1801 is complaining about in the first place.

## Test plan

- [x] `make test-all` — **no `make` on this machine** (Windows; #1185). CI runs the suites. Ran what this change can affect: both new JSON files validate against `research.schema.json` / `tree-gedcomx.schema.json`; the new test validates against `unit-test.schema.json`; `test.id` unique across the corpus (397 tests); and I checked the **cross-reference integrity the JSON Schema cannot express** (assertion→source, assertion→log, person_evidence→assertion, person_evidence→person, source→tree source, relationship endpoints, tree source refs) — all resolve.

- [x] Ran `research-exhaustiveness` and committed the run log **and** its `.ann.json`: `v1_2026-08-21_09-40-07`, verified **active** after merging `main`. **17 pass, 1 partial of 18.**

### `_018` passes for the right reason

A positive test like this invites passing on the wrong gap, so the `judge_context` forbids the near-misses explicitly (gating on the nil baptism, the nil probate, or plan incompleteness). The judge's own rationale confirms it did not:

> *"correctly identified that Ida's own 1875 Massachusetts birth registration is a decisive, unsearched record type… the birth record would provide primary-information evidence from parents as informants at the event, whereas the death record's parent names come from a secondary informant"*

That is the ranking clause, applied.

### Two other outcomes moved

- **`_017` fail → pass** — the sealed-record test. Plausibly clearer now that the rule names the era condition and the ranking explicitly rather than mentioning birth certificates only in the exception. **One run cannot show causation and I am not claiming it**; recording the coincidence because it sits directly on the edited rule.
- **`_016` pass → partial** on `Stop criteria coverage`, and the annotator then took it to **1** — harder than the judge, not softer. Investigated rather than left as noise, and it is an **eval defect, now filed as #1843**: the rubric requires all seven stop criteria in prose on a decline, while `_016`’s and `_014`’s `judge_context` say only the *unmet* one is needed, and `_001`’s and `_015`’s say all seven. Four tests in one directory, two incompatible standards; the judge’s own rationale shows it trying to honour both. Not fixed here — it is a doctrine call, and the fix edits test files inside this skill’s snapshot, so it would invalidate the run log this PR commits.

## The README states no verdict, by design

A scenario README is read **whole** into the judge prompt (`_load_scenario_readme` → `judge/prompt.md`), so a *"the skill must do X"* bullet is an answer key handed to the grader. That is #1687's finding, and it applies to new fixtures as much as old ones. This README states fixture facts and names the capability under test — the record inventory, the registration dates, the informant proximity recorded in the data — and never the expected outcome. Checked with the same grep #1687 uses.


EdmondOware (changes requested):
Strong piece of work, and the doctrine is right.

The argument that carried it for me is the internal inconsistency one: the rule already
named a birth certificate in its inaccessibility carve out while omitting the record from
the list that triggers the gate, and `ut_research_exhaustiveness_017` only makes sense if
an accessible one would gate. I checked both halves. `sealed-record-not-a-gap` is exactly a
privacy sealed 1936 Utah birth certificate, so the corpus really had been treating the
record as decisive while the list had not caught up. Reading the test corpus as evidence
for a doctrine claim, rather than arguing the prose on its own terms, is the right instinct.

Ranking the birth record above the death record on informant proximity is correct, and
stating the reason inline rather than leaving it implied is what makes it usable at the
point of decision. Conditioning it on the era instead of writing an absolute rule is the
call that keeps the eleven Flynn scenarios honest. Backing out of the `_015` edit and
saying plainly that you had planned it and were wrong is the part I would most like other
authors to copy.

The fixture is well built. Making the death record positive and naming both parents, so
the test cannot pass on an obvious gap, is the detail that makes `_018` mean something, and
the `judge_context` forbidding the three near misses closes the rest. The candidate table
explaining why no existing scenario could host it saved me the work of checking it myself.
I verified the run log independently: 59 snapshot entries, zero mismatches against your
branch, exactly five against `main` and they are the five files you add or edit. Filing
#1843 rather than fixing it inside the snapshot was right, and annotating `_016` harder
than the judge rather than softer is the correct direction to err.

### 1. Massachusetts statewide registration began in 1841, not 1842

The FamilySearch Wiki gives 1841 on nine separate pages with no dissent, and notes
Massachusetts was the first state to do it. Three places to change:

- `eval/fixtures/scenarios/ma-birth-record-unsearched/README.md`, line 28
- the new test's `test.description`
- the new test's `judge_context`, which is the one that matters most, since it states the
  date to the grader as fact

Nothing about the outcome moves, because Ida was born about 1875 and that is after either
date. But this fixture exists to pin a rule that turns on when registration started, so the
era boundary is the wrong fact to be a year out on.

### 2. Leave Ellen Sullivan on the single source, and record why the censuses do not corroborate her

To answer what you asked: do not add a corroborating source. The mother resting on one
secondary assertion, from a granddaughter reporting a birth 71 years earlier, is realistic,
and it sharpens the test rather than muddying it. It is exactly the condition that makes the
birth registration decisive rather than merely desirable, since its informant is a parent at
the event and it would give Ellen's maiden name. A corroborating source would weaken the
pull toward the record under test.

What is worth recording is the thing a reader will trip on. Ida is in the Charles W. Braman
household in 1880 and enumerated as daughter of the head in 1900, and neither yields a
mother assertion. A real enumeration of that household would list Charles's wife. Either she
was not Ellen, which would be worth saying, or the extraction is partial. Put a line on
#1801 rather than in the README, so the run log stands.

### 3. Update the branch, then ask for `eval-cosmetic-skip` instead of paying for another run

You are four commits behind and one of them, #1832, edits
`research-exhaustiveness/SKILL.md`. It touches the classification hard block near line 48,
not your gate at line 222, so it merges cleanly. What it does do is break your run log: the
snapshot embeds `SKILL.md`, so once you merge, rule 2 in `check_runlogs.py` sees a mismatch
and the `runlogs` check goes red.

Do not rerun the harness for it. No test in `eval/tests/unit/research-exhaustiveness/`
references `assertion-classification` or `record-extraction`, so #1832 changes nothing your
18 tests exercise, and the date fix in item 1 changes nothing either. That is what
`eval-cosmetic-skip` exists for.

Sequence it as one push, the date fix and the merge together, then say so on the PR and a
senior can apply the label. It has to go on after the final push, because the workflow
resets the bypass from the event on every new commit.

benter-070:
## Reworked onto the skill-agent split — ready for another look

> *"move your work into the new agent … check out the latest main and then re-apply the changes you made to the agent. Everything else should carry over."*

Done. Pushed as `cc2146d9` plus a main merge.

**The edit moved, unchanged in substance.** #1819 split `research-exhaustiveness` into a skill-agent pair after this PR opened, so the section I had edited no longer exists in the SKILL.md — the merge conflicted there. I took **main's SKILL.md wholesale** (verified byte-identical to `origin/main`, 63 lines, no conflict markers, no orphaned doctrine) and re-applied the same three-part edit at its new home, `agents/research-exhaustiveness.md`:

- the subject's own birth record added to the parentage list, era-conditioned on *"where the jurisdiction and period kept one"*;
- ranked above the death record, with the reason — *its informant is a parent at the event rather than a survivor recalling it decades later*;
- the pre-registration exclusion stated outright, naming the **baptism** as what to gate on instead.

**Everything else carried over untouched:** the `ma-birth-record-unsearched` scenario's three files and `ut_research_exhaustiveness_018`.

### The run log had to be replaced regardless

`v1_2026-08-21_09-40-07` was already stale on **three files of main's own** before this rework began — the SKILL.md, `references/research-exhaustiveness.md`, and the **rubric**. The rubric one matters: main rewrote the *Stop criteria coverage* dimension around "grade the content, not the presence." So the re-run was owed by the restructure, not bought by the relocation.

New log: **`v1_2026-08-24_12-36-14`**, verified active, annotated (33 corrections, 5 sampled, no overrides).

### Suite improved: 18/18, up from 17 pass / 1 partial

And `_018` still passes **for the right reason** from the agent's new location — which is the thing I most wanted to confirm, since a pass could have come from the wrong gap. The judge's own rationale names all three parts of the edit:

> *"Massachusetts required civil birth registration from 1841, the record is accessible (not sealed or lost), and it would provide primary-quality evidence from a **parent-informant**"*

### One knock-on, checked rather than assumed

`ut_research_exhaustiveness_016` now passes — the test whose partial produced **#1843**. I checked whether that makes #1843 moot. **It does not.** All four `judge_context` notes survive main's rewrite verbatim, and the rubric still requires all seven criteria in prose on a decline — so two tests still say the unmet one suffices while the rubric and two siblings require all seven. `_016` passed because the skill addressed all seven on its own initiative, satisfying the *strict* reading; the relaxed note went **untested**, not resolved. Recorded on #1843 with the evidence, including a correction to my own earlier speculation that it might be moot.

benter-070:
@EdmondOware — all three items done, in one push as you asked. Ready for another look.

### 1. Massachusetts 1841, not 1842 ✅

Corrected in all three places you named — the scenario README, the test's `test.description`, and the `judge_context`. The agent doctrine needed no change: it is era-conditional and names only Irish births before 1864, no Massachusetts date.

You were right that this is the wrong fact to be a year out on. My source was a "1841–1842" range and I picked the later end; nine wiki pages with no dissent settles it, and "first state to do it" is the detail that makes 1841 the meaningful date rather than an arbitrary pick.

### 2. Ellen Sullivan left on the single source, and the reason recorded on #1801 ✅

Not corroborated, per your ruling — and I verified the fixture matches what you described before writing it up. `a_004` (from `src_003`, the 1946 death registration, `secondary`, `family_not_present`) is the **only** mother assertion. Neither census contributes one: `a_002` is a birth fact, `a_005` is a father relationship.

The note is on **#1801** rather than the README, so the run log stands. It states both readings of the gap — the enumerated wife was not Ellen, or the extraction is partial — and reads it as the latter, since the log entries describe searching for Ida rather than reconstructing the household. It also records *why* it must stay uncorroborated: the single weak assertion is what makes the birth registration decisive rather than merely desirable, so strengthening it would blunt what `_018` exists to detect.

### 3. Updated the branch — and the run log situation has changed since your review

Merged main: **20 commits**, no conflicts. Worth flagging that your prediction about #1832 breaking the run log no longer applies, because I had already reworked this PR onto the skill-agent split at your earlier request, which required a re-run. The current log is **`v1_2026-08-24_12-36-14`** — **18 of 18 pass**, taken *after* the relocation, and annotated.

**The 20-commit merge invalidated nothing.** The only staleness is the date fix, and it is exactly three lines, one character each:

```
README.md          -from 1842, and those returns are public
                   +from 1841, and those returns are public
test.description   1842 -> 1841   (cosmetic per normalize(); does not affect the snapshot)
judge_context      -...registration from 1842 and the returns are public...
                   +...registration from 1841 and the returns are public...
```

Two snapshot files stale, both mine, both from `1842`→`1841`. **No test's behaviour changes** — Ida was born about 1875, after either date, so no assertion, no scenario state and no expected outcome moves.

So `eval-cosmetic-skip` is the right instrument and this is squarely what it is for. **This is the final push**, so the label can go on now — and since you raised it, you can apply it yourself.

One correction to my own earlier note on this PR: I had speculated that `_016` now passing might make #1843 moot. It does not — all four `judge_context` notes survive main's rubric rewrite verbatim, and `_016` passed because the skill addressed all seven criteria on its own initiative, satisfying the strict reading. The relaxed note went untested rather than resolved. Recorded on #1843.

EdmondOware (approved):
Approved. All three items are addressed.

The Massachusetts date is corrected in all three places, and 1842 no longer appears anywhere in the test or the scenario. Ellen Sullivan stays on the single source, which is the right call for what this fixture is for. The census note is on #1801 rather than in the README, so the run log stands.

`eval-cosmetic-skip` is applied, so do not run the harness for this.

The bookkeeping on item 2 is the part I would point other authors at. You were given an answer that meant doing less rather than more, and instead of leaving the awkward fact unrecorded you put it where a future reader will find it without disturbing the snapshot. That is the harder half of taking review feedback.

benter-070:
@florencemashipei requested as second reviewer — this carries @EdmondOware's approval already (2026-08-24 16:49, seven minutes after the last commit, so it is current), and the merged PRs in this series each took two.

**One sequencing note for whoever merges, because the order matters.**

This PR is `MERGEABLE` but **`BEHIND`**, so the branch needs updating first. The `eval-cosmetic-skip` label on it was applied for the `1842 → 1841` date correction in `exhaustiveness-birth-record-era-gate.json` — a behaviour-neutral prose fix that landed after the substantive run and invalidated the snapshot.

Per `eval/CLAUDE.md`, **that bypass expires on every push**, and the workflow decides it from the `synchronize` event rather than by reading the label — so updating the branch will turn the `runlogs` check red again even though nothing behavioural changed. A senior then has to re-apply the label.

So: update branch → `runlogs` goes red → re-apply `eval-cosmetic-skip` → merge. Not the reverse, and the red check in between is expected rather than a new problem.

The alternative is a fresh `research-exhaustiveness` suite run to make rule 2 pass honestly, which is ~$8–9 and a 27-cell annotation for a one-character date fix. The label exists for exactly this case.

**What each reviewer is actually judging** — there is no code in the diff:

- `agents/research-exhaustiveness.md` (12+/5−) — whether the subject's own birth record outranks the death record for parentage, era-gated on civil registration existing. Record craft.
- `fixtures/scenarios/ma-birth-record-unsearched/` — whether the fixture's facts hold and the scenario is solvable. Massachusetts statewide registration from 1841; Ida F. Braman b. ~1875 Taunton. The death record deliberately names **both** parents, so the test cannot pass on an obvious gap.
- `exhaustiveness-birth-record-era-gate.json` — the test and its `judge_context`.
- the run logs and their `.ann.json` — annotation judgement.

EdmondOware:
Applied `eval-cosmetic-skip`. `runlogs` is green and every check on the PR now passes.

**Why the waiver is justified**, so it is on the record rather than taken on trust. The latest
run log is `v1_2026-08-24_12-36-14`, 18 of 18 pass, and its snapshot has 59 entries of which
exactly two drift:

- `eval/tests/unit/research-exhaustiveness/exhaustiveness-birth-record-era-gate.json`
- `eval/fixtures/scenarios/ma-birth-record-unsearched/README.md`

Both files carry exactly two commits on this branch: the original work on 23 August, and
`b72aa5dd` at 15:04 UTC on the 24th. The run log is stamped 12:36 UTC on the 24th, so the only
change after the run was `b72aa5dd`, whose entire diff is three lines of `1842` to `1841`. Ida
Braman was born about 1875, so a one year correction to when Massachusetts registration began
cannot move a reasoning path or an outcome. Behaviour neutral, and no rerun owed.

**Please do not push to this branch before the second approval lands.** The bypass is tied to
this commit: any new push voids it and `runlogs` goes red again. Re-applying the same label
does not fix that, because GitHub fires no `labeled` event when the label is already present,
so recovery means removing the label and adding it back. That includes merging `main` in to
update the branch, which is not needed here since it is already level.

**What is left is one approval from anyone other than me.** My approval is in, nothing is
blocking, and the branch is up to date. @mercyokum, you merged `main` into this branch this
morning so you are already across it, if you have the time.

mercyokum (changes requested):
Solid PR overall — the doctrine gap is real and the fixture/test are well-reasoned. Three things worth fixing before merge:

## 1. The era-exclusion clause doesn't cover the fixture it's meant to exempt

`packages/engine/plugin/agents/research-exhaustiveness.md:275-277`:

> Where it did not — Irish births before 1864, most jurisdictions before the nineteenth century — its absence is not a gap, and the baptism is what to gate on instead.

Patrick Flynn (used across 11 existing scenarios) was born ~1845 in Pennsylvania — that's *inside* the nineteenth century, and Pennsylvania had no statewide birth registration until 1906. Neither named exclusion literally covers this. Your own new test description gets it right ("unlike the flynn-\* scenarios, where an 1845 Irish or Pennsylvania birth predates registration"), so the intent is correct — the wording just doesn't deliver it. The general clause earlier in the bullet ("where the jurisdiction and period kept one") probably saves this in practice, but the illustrative list sitting right next to it reads as the boundary a model would pattern-match against, and no test here checks that it doesn't.

**You wrote:**
> Where it did not — Irish births before 1864, most jurisdictions before the nineteenth century — its absence is not a gap, and the baptism is what to gate on instead.

**Change it to something that names the actual condition** — record availability by era *and* jurisdiction, not a bare century cutoff, e.g.:
> Where the jurisdiction's own registration did not yet exist at that date — Irish births before 1864, most U.S. states before the early twentieth century — its absence is not a gap, and the baptism is what to gate on instead.

## 2. New sentence is rationale, not instruction

Same bullet, line ~274:

> ...its informant is a parent at the event rather than a survivor recalling it decades later.

This explains *why* the birth record outranks the death record; the instruction is already complete without it ("the subject's own birth record outranks the death record for parentage"). Both CLAUDE.md files are explicit that agent `.md` files carry the instruction only, not the reasoning — every line here is a billed token on every invocation, and this exact explanation is already in the fixture's README and the test's `judge_context`, which is where it belongs.

**You wrote:**
> Where civil registration existed, the subject's own birth record outranks the death record for parentage: its informant is a parent at the event rather than a survivor recalling it decades later.

**Change it to:**
> Where civil registration existed, the subject's own birth record outranks the death record for parentage.

## 3. `mcp_fixtures` names a tool this agent can't call

`eval/tests/unit/research-exhaustiveness/exhaustiveness-birth-record-era-gate.json:21-23`:

```json
"mcp_fixtures": [
  "validate-research-schema"
],

**Comment by benter-070** [comment]
Thanks @mercyokum — all three verified against the files before I touched anything, and all three were real. Pushed as `38662fb8`. Items 2 and 3 are exactly as you wrote them. **Item 1 I did differently, and I'd like you to check my reasoning before you approve.**

## Item 1 — your diagnosis is right, but the suggested wording moves the hazard

Your proposal:

> Where the jurisdiction's own registration did not yet exist at that date — Irish births before 1864, **most U.S. states before the early twentieth century** — its absence is not a gap…

The problem is the same one you identified, pointed at this PR's own fixture. `ma-birth-record-unsearched` is **Massachusetts, birth ~1875**, and Massachusetts kept statewide registration **from 1841**. So the birth record *does* exist and *is* the gap — that is the entire point of `ut_research_exhaustiveness_018`.

But "most U.S. states before the early twentieth century" is a phrase an 1875 Massachusetts birth pattern-matches straight into. By your own argument — the illustrative list is what a model matches against, not the general clause beside it — that wording invites the model to exempt the very record this test requires it to gate on. The test would then fail for a reason the doctrine introduced.

What I shipped keeps your operative clause and swaps the exemplar:

> Where civil registration existed, the subject's own birth record outranks the death record for parentage. Where the jurisdiction's own registration did not yet exist at that date — Irish births before 1864, **Pennsylvania before 1906** — its absence is not a gap, and the baptism is what to gate on instead.

`Pennsylvania before 1906` is the jurisdiction and date from your own Patrick Flynn example, so it covers the flynn-* population exactly — and MA 1875 falls outside both exemplars rather than inside one. If you'd rather drop the exemplars entirely and keep only the condition, say so; that's also defensible and I have no attachment to the list.

## Item 2 — taken as written

Removed: *"its informant is a parent at the event rather than a survivor recalling it decades later."* You're right that it's rationale, and it already lives in the fixture README and the test's `judge_context`.

## Item 3 — confirmed dead, removed

`validate-research-schema` appeared **0** times in the agent body, the agent's `tools:`, and the skill's `allowed-tools`, so nothing in the pair could ever have matched it. `mcp_fixtures` is optional in the schema and the test still validates.

## A re-run is owed, and I'm running it rather than re-waiving

Items 1 and 2 edit the agent body, which is in the snapshot, so `runlogs` will go red. **I'm not asking for `eval-cosmetic-skip` this time.** Item 1 changes what the model gates on by design, and item 2 is a deletion I specifically won't call neutral: on #1863 today, deleting an example line from an agent body coincided with a test going 0-for-3 across two variants, including the variant that only removed text. The label is documented for behaviour-neutral edits, and neither of these qualifies.

`research-exhaustiveness` is running now; I'll post the result and commit the run log with its annotation.

@EdmondOware — this voids the waiver you applied, which your note predicted. It was unavoidable once the review asked for a body change, and re-applying won't help since GitHub fires no `labeled` event when the label is already present. The fresh run replaces it, so nothing needs re-waiving.

**Review by mercyokum — APPROVED** [review]
All three items check out — item 1 especially, since "Pennsylvania before 1906" avoids a real trap my own suggested wording would have walked into (an 1875 Massachusetts birth pattern-matching into "most U.S. states before the early twentieth century" and exempting the exact record this fixture requires). Doctrine text looks right to me now. Only thing holding this back from an approval is `runlogs` — it's red because the agent-body edit isn't behavior-neutral, and the fresh annotated run you mentioned running hasn't landed on the branch yet, so I'll take another look once that's committed.

**Comment by benter-070** [comment]
Thanks for the approval @mercyokum. On the CI check — **no rerun is needed, and re-running would set us back.** I ran the gate locally rather than reading the summary, and it names one thing only:

```
skill `research-exhaustiveness`: annotation `v1_2026-08-25_17-15-12.ann.json`
is incomplete — 18 dimension(s) are unreviewed
(e.g. ut_research_exhaustiveness_005/base/Correctness, .../Completeness,
.../Tool Arguments, .../rubric/Threshold reasoning, .../rubric/Declaration honesty)

1 rule violation(s).
```

That is **rule 3 only**. The run itself is fine:

- **Rule 2 passes.** `diff_snapshot_vs_disk` over the run log's 59 snapshot entries returns **no drift** — it matches this branch exactly, so it is active.
- **The run is clean.** 18 pass, 0 partial, 0 fail. `ut_research_exhaustiveness_018` — the new birth-record test — passes 6-for-6 dimensions.
- The annotation exists and is **partly done**: 12 of 30 cells reviewed, 18 to go.

**Why re-running would make it worse, not better.** A new run writes a new `v{N}_<ts>.json`, which becomes "the latest run log" that rules 2 and 3 gate on. That new log then owes **its own** 30-cell annotation from scratch — so the check stays red, the 12 cells already reviewed are discarded, and it costs another ~$3.44 and a fresh annotation pass. Retention would also prune an older log to make room. Net effect: same red check, less evidence, more spend.

The fix is finishing the annotation that is already underway, which is happening now. I'll commit the completed `.ann.json` and the check goes green — nothing else is outstanding.

One thing worth flagging for whoever reviews the annotation rather than the diff: `_005` (`refuse-while-in-progress`) carries **two nulls** — *Stop criteria coverage* and *Threshold reasoning*. The test's own `judge_context` item 6 explicitly sanctions the Stop-criteria null ("the dimension grades a situation this run does not create"). It says nothing about *Threshold reasoning*, which is defensible by the same logic — the run stops at the Step 0 precondition, before the five threshold questions at Step 2 — but is undocumented. Per `eval/CLAUDE.md` a rubric null on a positive test is the top-priority cell to scrutinise, since `_compute_outcome` gates on 1 and 2 and null is neither. Not a blocker; just the cell that deserves the attention.

benter-070:
@mercyokum — closing the loop, since I left you holding a "needs a rerun" note and never came back with the outcome.

**`runlogs` is green and the CI block is cleared. No rerun was done.** The gate's only complaint was rule 3: the annotation existed but was incomplete, 18 of 30 dimensions unreviewed. Finishing it was the whole fix.

Rule 2 was never the problem — `diff_snapshot_vs_disk` over the run log's 59 snapshot entries returned no drift at every point I checked it, including twice after `main` came in.

**Annotation:** 30 cells across the 5 sampled tests, **0 disagreements** with the judge. All 18 tests passed, so no cell owed a written comment.

**Why I pushed back rather than just re-running.** A rerun writes a new `v{N}_<ts>.json`, and rules 2 and 3 gate on *the latest* run log. So the new one would have become the gated artifact and owed its own 30 cells from scratch — discarding the 12 already reviewed, costing ~$3.44, pruning an older log to stay under the retention cap, and leaving the check exactly as red. Same outcome, less evidence.

Worth having in the open because I think it is a common trap with this gate: a red `runlogs` reads as "the run is bad", and about half the time it means "the run is fine and the annotation is not". Rule 2 says the run is stale; rule 3 says the *annotation* is. Only the first is fixed by re-running, and the error text names which one.

**Two things I checked along the way rather than assuming:**

- Someone merged `main` into this branch mid-flight, bringing 6 commits including the year-qualifier doctrine rewrite (#1865) and the person-evidence deep dive (#1882). Any of those could plausibly have staled the run log. None touched `research-exhaustiveness`'s snapshot. My push was rejected because of that merge; I inspected and merged rather than forcing, so nothing of yours was lost.
- Updating from `main` again just now (18 commits) also touches none of the agent body, the skill dir, the test dir, or any of the 12 scenarios these tests reference — so no rerun owed for that either.

**One note for whoever next takes research-exhaustiveness's eval slot**, recorded rather than acted on: `_005`'s two nulls are not equivalent. `judge_context` item 6 explicitly sanctions the `Stop criteria coverage` null. `Threshold reasoning` is undocumented — defensible by the same reasoning, since the run halts at the Step 0 plan-item precondition and never reaches the five threshold questions at Step 2, but nothing in the test says so. Per `eval/CLAUDE.md` a rubric null on a positive test is the top-priority cell to scrutinise, so it deserves a `judge_context` line making the sanction explicit instead of inferred. Not filed — it is one line and belongs in whatever next claims that slot.

Both approvals are in and every check is green, so this is ready to merge whenever one of you wants to.

Issue #1750, opened 2026-08-19
project-status: in_progress plan items, the skill-naming contradiction, and the exhaustiveness guide's coverage lists


Cia-3:
> **Reviewed 2026-08-20 before junior handoff.** Decision settled — issue #1633's
> project-status half moves in here, so the whole prompt change rides one paid
> run. Findings 1–3 are sharpened below; the original follows under "Original
> issue".

## Ruling 2026-08-20 — merge issue #1633's project-status half into this card

Both are genealogist edits to the same `SKILL.md`, and landing them apart buys two paid
runs and two 5-test annotation passes over the same suite. Issue #1633 keeps the other 18
unreachable reference files and the both-directions lint; its project-status bullet moves
here. See Finding 4.

## Finding 1 — the two sites, and the four you must not touch

Fix: update both sites to count `planned` **and** `in_progress`.

- Step 3's stats table → `Count plan items with status "planned" or "in_progress" in active plans`
- Step 4's condition 2 → widen the same way, and make the non-stale branch recommend
  finishing an `in_progress` item before starting a `planned` one.

`eval/tests/unit/project-status/rubric.md`'s Actionability pass example already states the
target behaviour ("complete the in-progress probate search per pli_006"), so **the rubric
needs no edit** — a rubric change here would be a grading-lane edit riding a doctrine-lane
PR.

**Change nothing outside `project-status`.** Four other skills also read plan-item status
literally as `"planned"` — `search-records`, `search-full-text`, `search-images`, and
`research` (twice). Reproduce with
`git grep -nE '"planned"' -- 'packages/engine/plugin/skills/*/SKILL.md'`. Those are
*selection* sites ("find the next item to execute"), where preferring an unstarted item is
defensible; project-status's two are *counting* and *branching* sites, where ignoring
`in_progress` is simply wrong. Widening the diff arms four more run-log gates and buys
four more paid runs.

**One correction to the evidence.** The docking happened in the **first** of PR #1746's
two runs (`v1_2026-08-19_10-14-20.json`, `ut_project_status_001` partial, Completeness 2
and Actionability 2, both rationales naming `pli_006`). The **second** run — the one
committed as newest — scored 3 on both and did name `pli_006`, with no skill file changed
between them. So the gap is real and readable in the formula, but it is 1-of-2 in the run
log rather than deterministic, and with `runs_per_test` pinned to 1 **the post-fix run
[...]

Issue #1821, opened 2026-08-21
Plan-item status has four writers, three rules, and no owner for the in_progress to terminal transition


DallanQ:
**Touches:** packages/engine/plugin/skills/record-extraction/SKILL.md, packages/engine/plugin/skills/search-records/SKILL.md, packages/engine/plugin/skills/search-full-text/SKILL.md, packages/engine/plugin/skills/search-images/SKILL.md, packages/engine/plugin/skills/search-external-sites/SKILL.md

> **Narrowed 2026-09-07 (lead ruling, /make-decisions): the `exhausted` enum change now lives on issue #2077**, which specifies it more fully (viewer colour row, `planCompleteInvariants` extension) and buys the research-plan run anyway. **This card is the `in_progress` → terminal owner only** — one skill snapshot, one paid `record-extraction` run — and it no longer edits `research-plan/SKILL.md`, the enum, the validator or the schema spec. Item 2 below and the enum-change site list are kept for the record but are #2077's work.
>
> **Both decisions are settled (lead, 2026-08-23).**
>
> 1. **The `in_progress` → terminal owner is the `record-extraction` skill** — the thin
>    router, **not** the `record-extractor` agent. The router already writes the research-log
>    entry after the agent returns, so it is the one component that knows extraction
>    finished. Give it `research_append` and a step that flips the item. Do **not** hand
>    `research_append` to the agent: `docs/specs/schemas/ownership.json` holds the extraction
>    lane off that section by tool identity (`extraction_append` does not accept it), and
>    that boundary is deliberate.
>
> 2. **`plan_status` gains a new value, `exhausted`** — route 2, not a reuse of `completed`.
>    `completed` means *the plan was executed*; BCG Standard 18's state is *searched out,
>    still unanswered*. Reusing `completed` conflates them permanently, and both routes cost
>    the same paid run anyway, so the only difference is plumbing.
>
> Reasoning for both, including the overruled alternative and the collision check against
> PR #1847, is in the ruling comments below.
>
> ~~Both halves land in one PR and one `make eval-skill` run.~~ Corrected 2026-09-07: they were always two snapshots and two runs, and the enum half is now #2077's. This card buys the `record-extraction` run only.
>
> **Enum-change site list** (a new value on a closed enum, per `CLAUDE.md`): the enum lives
> in `$defs` in `enums.schema.json` in **both** trees — `docs/specs/schemas/` *and*
[...]

PR #1845, opened 2026-08-23
project-status: count in_progress plan items, restore the next-step scoping clause, name three repositories (#1750, Findings 1-3)

benter-070:
## Summary

**Tier: Normal.** Findings 1–3 of #1750. **Finding 4 is deliberately not here** — the naive fix causes a measured regression; evidence below and on the issue.

- **Finding 1** — two sites read plan-item status as `"planned"` only: step 3's stats table (a *counting* site) and step 4's condition 2 (a *branching* site). Both now count `"planned"` or `"in_progress"`, and the non-stale branch names an in-flight item before recommending a new one.
- **Finding 2** — the next-step scoping clause, lost in a length trim, restored in `references/conclusion-readiness.md` where the rule lives.
- **Finding 3** — county/town histories added to the record-type list, genealogical libraries and university/special collections to the repository list, as their own bullets.

## Review guidance

**Start here:** step 4's condition 2 in `SKILL.md`. The counting fix is mechanical; the branching fix is the behavioural one, and it makes the skill recommend *finishing* an `in_progress` item before starting a `planned` one. That ordering is a genealogist judgement — if you would rather it simply reported both counts and left the choice to the researcher, this is the line to say so on.

The rubric is untouched by design: its Actionability pass example already states the target behaviour ("complete the in-progress probate search per pli_006"), so editing it would be a grading-lane change riding a doctrine-lane PR.

**Unsure about:** nothing on the three findings. The judgement I would most like checked is Finding 2's wording — see below, because I got it wrong the first time in an instructive way.

## Other skills that read plan status without `in_progress`

Corrected after review. **The card said "`research` (twice)" and was right.** An earlier version of this paragraph swapped `research` out for `research-plan` while keeping the count at four, so both lists were wrong in different ways. The accurate population, with the grep that actually finds it:

```
git grep -nE '("planned"|`planned`)' -- 'packages/engine/plugin/skills/*/SKILL.md'
```

`git grep -nE '"planned"'` — the reproducer the card and my first draft both gave — matches **only the double-quoted spelling**. It therefore misses `research` entirely (lines 150, 227 and 250 are all backticked) plus `search-records:615` and `:634`.

| skill | lines | what it is |
|---|---|---|
| `search-records` | 87, 615, 634 | selection — untouched |
| `search-full-text` | 63 | selection — untouched |
| `search-images` | 106 | selection — untouched |
| `research` | 227, 250 | selection — untouched |
| `research-plan` | 93, 328 | **already correct** — reads `planned`/`in_progress` |
| `research-plan` | 381, 401–402 | not a selection site — a *creation* site, and the enum's own documentation |
| `research` | 150 | not a selection site — deliberately about items *still* `planned` |

So **four** skills carry selection sites and stay untouched; `research-plan` is the one that already uses the both-statuses form, which is the precedent showing this shape exists in the plugin. The two sites this PR edits are the *counting* and *branching* sites, where ignoring `in_progress` is simply wrong rather than defensible. Widening the diff would arm four more run-log gates.

## Finding 2 is restored as *scope*, not verbatim

The original clause ended *"Route to the skill yourself; tell the user the action."* My first attempt quoted that. On an empty folder it reads as an instruction to **act**, and the Findings-1–3-only control run came back 1-of-2 rather than clean, which is what pointed at it.

Shipped wording keeps the scope and drops the imperative:

> The rule constrains the wording shown to the user, not which skill the work is handed to.

The card is right that the scope was lost in a trim. What it could not know is that restoring it *verbatim* carries the risk.

The 9 parenthetical skill names in the decision tree are **kept**, per the card: they are the handoff targets the clause refers to, the plugin's house pattern (25 occurrences across 9 `SKILL.md` files), and the same vocabulary `project_context` returns in `questionStatuses[].nextStep`.

## Finding 4: attempted, measured, deferred

I took the "write `references/output-formats.md`" branch, with skeletons derived from the prose that already specifies the content — the "Two summaries" block, step 3's statistics table, step 3b's four levels, step 3c's readiness conditions, step 4's decision tree.

`ut_project_status_009` (`empty-folder-no-project`) then failed. Controlled:

| condition | `_009` pass rate |
|---|---|
| clean `main`, none of my edits | **3 / 3** |
| Findings 1–4 (with `output-formats.md`) | **0 / 3** |
| Findings 1–3 only (file removed) | 1 / 2 |
| Findings 1–4 + two guard paragraphs | **2 / 3** |
| **Findings 1–3 as shipped** | **pass; suite 11/11** |

Before my change that test passed **4 for 4** across every committed run log, each time with **zero tool calls**. Failure mode was perfectly consistent: **pass ⇔ no tool calls, fail ⇔ `project_context` called**, tripping `test_no_mcp_tools_called`.

**The mechanism generalises, which is why it is worth a paragraph rather than a shrug:** a reference file describing a rich output shape prompts the model to go and *find data to fill it*. The detailed skeleton carries a statistics table; on an empty folder there is nothing to put in it, so the model reached for `project_context` to orient instead of concluding there was no project. Two guard paragraphs ("this file describes shape only"; "no project, no step 5") moved it 0/3 → 2/3 but did not restore it. So Finding 4 needs a different approach than the card assumes, and the draft plus guard text are recorded on #1750 so the next attempt starts from that.

`SKILL.md`'s two `output-formats.md` pointers are **left intact** — they point at a file that has never existed, which is the pre-existing state Finding 4 describes. Removing them would be the *other* branch of that decision, and a behaviour change this PR has not tested.

## Test plan

- [x] `make test-all` — **no `make` on this machine** (Windows; #1185). CI runs the suites. Ran what this change can affect: `check_skill_frontmatter.py` clean, and the frontmatter/description is untouched.

- [x] Ran `project-status` and committed the run log **and** `.ann.json`: `v1_2026-08-23_15-39-16`, verified **active** after merging `main`. **11 pass of 11**, matching clean `main`'s baseline. The superseded `14-22-51` run log (the Findings-1–4 attempt, 10 pass / 1 fail) is deleted rather than committed.

**Two limits on what this run proves, both from the card:**

- Finding 1's original docking was **1-of-2** in PR #1746's two runs — the second scored 3 with no file changed. With `runs_per_test` pinned to 1, this run **cannot prove the fix worked**, only that nothing regressed. The fix is justified by reading the formula, not by a reproducible failure.
- Finding 2 has **zero rendered instances**: no user-facing recommendation in either committed run log carried a parenthetical skill name — every match sits inside a judge rationale. It is a doctrine repair, so no score should move.

## Retracted after review

This PR originally carried an "Observed, not filed" section claiming `project-status` calls a tool the harness should have denied it, on the authority of two sentences in `docs/architecture.md` §5.1 — *"an undeclared tool is denied at call time"* and *"Declare accurately."*

**Neither sentence exists.** PR #1774 (*retire per-skill tool deny in the eval harness*) rewrote §5.1, and this branch already contained it — `30e786c3` is an ancestor of its tip. `grep -c "denied at call time" docs/architecture.md` returns **0**. The section now reads that the unit harness *"grants every registered MCP tool to every skill, matching production"* and that `test_tool_allowlist` **warns** on undeclared calls, advisory and not gating. So the `project_context` call being recorded rather than denied is the documented behaviour: there is no contradiction and no restraint mechanism failing to restrain. The observation is withdrawn, and the same paragraph is corrected on #1750.

Finding 4's evidence is unaffected — it rests on `test_no_mcp_tools_called`, which is a per-test validator that does assert, and is not the allowlist.


EdmondOware (changes requested):
### 1. Cut the "Observed, not filed" section, and correct the same paragraph on #1750

It quotes `docs/architecture.md` §5.1 saying an undeclared tool is "denied at call time" and
"Declare accurately." Neither sentence exists any more. PR #1774, `retire per-skill tool deny
in the eval harness`, merged to `main` the day before you opened this, and your branch already
contains it: `30e786c3` is an ancestor of `1750-project-status-prose`. It grants every
registered MCP tool in the unit harness, makes `test_tool_allowlist` warn instead of assert,
and rewrites the two sentences you quote. `grep -c "denied at call time" docs/architecture.md`
now returns 0.

So there is no contradiction and no restraint mechanism failing to restrain. Cut it from the
body and post a correction on #1750, since that copy is the one the next person will read.

Your Finding 4 evidence is unaffected. The validator that trips is `test_no_mcp_tools_called`
in `eval/harness/validators/test_project_status.py`, which still asserts and is not the
allowlist.

### 2. Fix the population record for the `"planned"` sites

`git grep -nE '"planned"'` matches only the double quoted spelling.
`packages/engine/plugin/skills/research/SKILL.md` reads `planned` in backticks at lines 150,
227 and 250, and the grep finds none of them. The card said "`research` (twice)" and your body
swapped `research` out for `research-plan` while keeping the count at four, so both lists are
wrong in different ways.

Do not widen the diff. Fix the count and the reproducer in the body and on the card, because
the next person to audit these sites will run your grep and conclude `research` is clean.

While you are there, `research-plan` line 381 writes `status: "planned"` when creating a plan
item. That is a creation site, not a selection site, so calling all four selection sites is
loose.

### 3. Explain or restore the second deleted run log

The diff drops `v1_2026-08-11_13-21-10.json` and `v1_2026-08-11_19-00-27.json`.
`prunable_candidates` from `harness/versioning.py`, with `DEFAULT_KEEP_CANDIDATES` at 5,
returns `13-21-10` alone over the list after your addition. The second deletion is beyond the
rule and the body does not mention it. Nothing in `check_runlogs.py` enforces retention, so CI
could not catch it. Restore `19-00-27` or say in the body why it went.

### On the judgement you asked to have checked

Keep the `in_progress` ordering exactly as you wrote it. The rubric already names that
behaviour as its Actionability pass example, so reporting both counts and leaving the choice
to the researcher would put the prose and the rubric back out of step.

benter-070:
Both corrections done, body-only — the diff is byte-identical (3 prose files + the run-log swap).

**1. Section cut.** Verified your claim before acting on it rather than taking it on trust: `grep -c` returns 0 for both sentences, and `git merge-base --is-ancestor 30e786c3 HEAD` confirms my branch already carried #1774. Replaced with a short *Retracted after review* section that states what the doc says now, so the retraction is legible rather than just absent. Corrected on #1750 as well — https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1750#issuecomment-5397295079 — since that copy is the one the next reader hits.

**2. Population fixed, and there is a bit more to it.** You are right that I swapped `research` out for `research-plan` while keeping the count at four. Working out which list was right turned up two things worth having on the record:

- **The card's list is correct and its reproducer contradicts it.** `git grep -nE '"planned"'` returns zero `research` lines, because all three (150, 227, 250) are backticked — so the card says "`research` (twice)" while giving a grep that cannot produce it. The grep also misses `search-records:615` and `:634`. Both records now carry the both-spellings form:

  ```
  git grep -nE '("planned"|`planned`)' -- 'packages/engine/plugin/skills/*/SKILL.md'
  ```

- **`research-plan` was the wrong pick in a specific way.** It is the one skill that *already* reads `planned`/`in_progress` (93, 328) — the precedent showing the both-statuses form exists in the plugin, not a site owing the same fix. Its other three hits are a creation site (381) and the enum's own documentation (401-402). And `research:150` is deliberately about items *still* `planned`, so it is not a selection site either.

Net: four skills carry selection sites and stay untouched, exactly as the card said. Table in both places.

Nothing outside `project-status` touched. All 14 checks green.

EdmondOware:
Items 1 and 2 are done, and the catch on item 2 is a better result than what I asked for. You are right that the card's own reproducer cannot produce the list it claims, and the both spellings form plus the two missing `search-records` hits is the fix.

Item 3 is still open. `v1_2026-08-11_19-00-27.json` is still deleted and the body does not mention it.

The retention rule prunes one file here, not two. `prunable_candidates` over the six file list, with keep at 5, returns `v1_2026-08-11_13-21-10.json` alone.

Restore `19-00-27`, or add a line to the body saying why it went. Either one closes it.

EdmondOware:
Items 1 and 2 are closed, and both are better than what I asked for.

Retracting the allowlist observation in place rather than deleting it is the right call: the
record of what was wrong survives, and the grep output makes it checkable. I checked again
against `main` as it stands now and neither sentence exists. Correcting the same paragraph on
#1750 the same day closed the copy that outlives this PR.

On the population you did more than correct it. `git grep -nE '("planned"|`planned`)'` finds
the sites the original reproducer missed, and crediting the card as right about `research`
(twice), rather than quietly reversing, is the useful version of that correction.

### One thing left: the second deleted run log

The diff still drops both `v1_2026-08-11_13-21-10.json` and `v1_2026-08-11_19-00-27.json`, and
the body has no mention of retention, pruning, or either filename. I recomputed against `main`
as it stands now: `prunable_candidates` from `harness/versioning.py`, with
`DEFAULT_KEEP_CANDIDATES` at 5 over the union of both sides, returns `13-21-10` alone. So the
second deletion is beyond the rule, and nothing in `check_runlogs.py` enforces retention, which
is why CI is green on it.

Restore `19-00-27`, or add a line saying why it went. Either one closes it and I will approve.

benter-070:
@EdmondOware — item 3 is now done, and I owe you a note on why it took a second pass: **I read your review as having two items and never scrolled to the third.** Items 1 and 2 were body-only and I closed them out the same day; item 3 needed a code change and I simply did not see it. That is on me, not on how the review was written.

## Item 3 — you were right, and the arithmetic confirms it

`main` holds 5 project-status candidates. Adding this PR's run makes 6, and with `DEFAULT_KEEP_CANDIDATES = 5`:

```
prunable_candidates([... 6 candidates ...], keep=5)
  -> ['v1_2026-08-11_13-21-10.json']
```

One file. The diff deleted two. `v1_2026-08-11_19-00-27.json` is restored (`3d795121`), the directory is back at the cap of 5, and the diff now drops only the log the harness would have pruned itself. No `.ann.json` existed for the restored log.

Your point about *why* this was invisible is the part worth keeping: **nothing in `check_runlogs.py` enforces retention.** Pruning at the writer is what keeps the cap holding, so a hand-deleted extra file passes every rule. I checked before restoring rather than assuming, and `prunable_candidates` gave your answer exactly.

## Items 1 and 2, for completeness

**1.** The "Observed, not filed" section is cut. I verified your claim first: `grep -c "denied at call time" docs/architecture.md` returns 0, `grep -c "Declare accurately"` returns 0, and `git merge-base --is-ancestor 30e786c3 HEAD` confirms this branch already carried #1774. §5.1 now reads that the harness "grants every registered MCP tool to every skill, matching production" with `test_tool_allowlist` warning, "advisory, not gating" — so the recorded call is documented behaviour and there was never a contradiction. Replaced with a short *Retracted after review* section so the retraction is legible rather than just absent, and corrected on #1750.

**2.** The population record is fixed in both places. Working out which list was right turned up something neither of us had: **the card's list is correct and the card's own reproducer contradicts it.** `git grep -nE '"planned"'` returns zero `research` lines because all three (150, 227, 250) are backticked — so the card says "`research` (twice)" while giving a grep that cannot produce it, and it also misses `search-records:615` and `:634`. Both records now carry the both-spellings form. And you were right that `research-plan:381` is a creation site: it turns out `research-plan` is also the one skill that *already* reads `planned`/`in_progress` (93, 328), so it was the wrong skill to have swapped in — it is the precedent, not a defect.

## One warning I read and did not act on

`judge_prompt_hash` is stale — this run was scored under `0d186137`, current is `c39d7003`. Rule 2b is warn-only by design (judge edits are a separate cadence), so I am reading the corrected mean cautiously rather than re-running. Say the word if you would rather have a fresh run against the current judge; it is ~$4 and a 20-cell re-annotation.

`check_runlogs.py` locally: **All runlog rules satisfied.** Nothing else outstanding that I can see — but given I missed a third of your review once, please do check I have not missed anything else.

EdmondOware (approved):
All three items are closed and I checked each rather than reading the summary.

**Item 3** is done in `3d795121`. Candidates are back to five, and `prunable_candidates` over
them now returns nothing, so the set matches the keep-newest-5 rule exactly. The only deletion
against `main` is `v1_2026-08-11_13-21-10.json`, which is precisely what the rule prunes.
Restoring only the `.json` was right, since that log never carried an annotation on `main`
either.

**Item 1** stands as the retraction rather than a deletion, which was the better of the two
options: the record of what was wrong survives with the grep output that settles it. I checked
against current `main` again and neither sentence exists.

**Item 2** I ran rather than read. Your corrected pattern does the work:

```
git grep -nE '("planned"|`planned`)' -- 'packages/engine/plugin/skills/*/SKILL.md'
```

It returns the three `research/SKILL.md` backtick sites at 150, 227 and 250 that the original
double-quote reproducer missed entirely, alongside `research-plan`, `search-full-text` and
`search-images`. That is the population the card was reaching for, and crediting the card as
right about `research` rather than quietly reversing is the version of that correction worth
keeping.

Also verified: the latest run log is active at 34 snapshot entries with zero mismatches and
11 of 11 pass.

Nothing outstanding from me.

johnmarkpeterbrown (approved):
Approving. The genealogist judgement you flagged is the one thing here that
needed me, so I'm ruling on it rather than just signing off.

## Keep the in_progress ordering

You asked whether step 4's condition 2 should name the in-flight item or
simply report both counts and leave the choice to the researcher. Name it,
exactly as you wrote it.

An `in_progress` plan item is not a pending idea — it is a search someone has
already started. In practice it is usually a search that stopped partway for a
reason: the index ran out and the answer is in un-indexed page text, or the
film was unreadable, or the jurisdiction turned out to be wrong. Every one of
those needs a logged outcome before the plan moves on. If the skill reports a
count and shrugs, that item is where the research log quietly loses a nil
result, and an unlogged nil result is the most common way a project fails
exhaustiveness later without anyone noticing.

So the ordering is not a preference between two readable outputs. Finishing
in-flight work first is the standard, and the skill should say so.

Two things convinced me this is a real repair and not tidying:

- `search-records` is the skill that *creates* `in_progress` — it leaves an
  item there when the evidence says the answer is in un-indexed page text, and
  names full-text as the next step. So the state is produced by design.
- On `mid-research-flynn` and `flynn-multi-conflict`, the active plan holds two
  `completed` items and one `in_progress`, and **no** `planned` item at all.
  Before your change, step 3 reported "0 active items remaining" and step 4's
  condition 2 never fired — on the skill's own primary fixture, for the exact
  state `search-records` leaves behind. The researcher got no next step.

That settles it as doctrine, not as a reading of the rubric, so nobody needs to
re-open it.

## What I checked rather than read

- Re-derived the retention arithmetic myself: `prunable_candidates` over
  `main`'s five candidates plus this run returns `13-21-10` alone, and over
  your five returns nothing. The single deletion is exactly what the rule
  sanctions and no more. Item 3 is properly closed.
- Ran the runlogs gate locally: all rules satisfied. The judge-hash warning is
  warn-only and is not yours — the judge prompt changed on `main` two days
  after your run.
- Run log is 11 of 11 with every expected outcome matched and nothing flaky.
- Finding 3 misses no edit site. `record_type_recommended` and
  `repository_recommended` are open enums carrying `examples`, and the schema
  says additional values are allowed, so the prose lists were the only place
  those additions belonged.

Your call to leave the two `output-formats.md` pointers intact was right. They
point at a file that has never existed on any ref — I checked the whole
history, not just `main` — so removing them is the other branch of Finding 4's
decision and belongs with the attempt that tests it.

## One gap, going on #1821 rather than here

Step 4's condition 7 still reads `completed` only. A plan whose remaining items
are all `skipped` fires neither condition 2 nor condition 7 and falls through to
"no immediate next step is clear" — the same defect class you just fixed, one
branch down.

Not asking for it. No fixture holds that state, and any further `SKILL.md` edit
invalidates your snapshot and costs another paid run out of this skill's single
eval slot. #1821 already owns the plan-item status lifecycle, so I'm recording it
there.

Nothing to change. The senior requirement is satisfied by this approval, so one
more from anyone with write access merges it.

PR #1856, opened 2026-08-24
eval: report the §11 unnamed-delegate shadow count + tolerate namespaced agent_type (#980)

### LINKED ISSUE #980: Measure shadow-mode numbers before graduating the unnamed-delegate guardrail check
> **Reviewed 2026-08-21 before junior handoff, and the lead has ruled.** The premise was
> false — the command this card named produces no unnamed-delegate number at all. The ask
> below is rewritten; everything from the old body is under "## History" and is history only.

## Ruling (lead, 2026-08-21) — keep it in shadow, and stop calling it pending

The unnamed-delegate check stays a reported signal, not a gate. It has 1 flagged run in 157,
and 139 of those runs carry no caller attribution at all, so the low fire rate is mostly
non-coverage rather than compliance — a hard fail set from one example is a coin flip
dressed as a threshold. Graduating now was beaten on exactly that ground.

The task is the reader below plus one line in `docs/specs/guardrail-enforcement-spec.md`'s
status row: shadow, reported, no graduation count — revisit only if a second attributed run
flags. **Write that sentence into the spec as part of the PR; do not leave the reasoning in
this issue.**

**Second half of the ruling, and it is not optional.** Make the
`agent_type == "record-extractor"` comparison in
`eval/harness/harness/skill_invocation.py` tolerate the namespaced spelling — strip a
leading `<plugin>:` segment before comparing. A live Cowork probe (2026-08-15) logged
`agent_type: "genealogy-research:image-reader"` for plugin-agent delegation while the
harness reports bare names, so an equality test that is green in CI is dead the moment the
detector is fed production data. That is the failure shape of issues #650, #698 and #939.
Add a test covering both spellings.

Also state, in the detector docstring and in the spec's status section, that the hook and
the detector differ **deliberately** — the hook denies main-thread calls, the delegate half
is shadow-logged — so this is not re-filed as drift. Issue #1273 carried the same complaint
and is settled the same way.

## Read this first — the command the old body named does not produce this issue's number

`make e2e-guardrail-shadow` reports five buckets: the caller-attributed recency window, the
live `same_person` provenance check, citation-nulling, conflict-unpersisted, and
warnings-unchecked. **There is no unnamed-delegate bucket.**
`grep -n "unnamed" eval/harness/e2e/guardrail_shadow_report.py` returns nothing, and nothing
outside the detector and its unit tests reads
`E2eResult.protected_writes_by_unnamed_delegate`.

Do **not** report that command's window table as this check's rate. That is the recency
check, which issue #1463 retired permanently — its numbers answer a closed question about a
different detector.

Issue #1358's citation-nulling bucket *is* in that output, so the "one run answers both"
note holds for #1358 and not for this issue.

## The sample, measured today

`find_protected_writes_by_unnamed_delegate` replayed over every committed e2e run — measured
2026-08-21 on `main` @ 59f6d706, by importing the detector from `harness.skill_invocation`
and running it over `e2e.runlog_selection.all_result_jsons()`:

| | |
|---|---|
| committed e2e runs | 157 |
| runs carrying any caller-attributed ledger entry | 18 |
| runs with at least one violation | **1** |
| violations in that run | 15 (10 `extraction_append`, 4 `tree_edit`, 1 `tree_correct`) |

The one flagged run is `antonio-lucas-spouse/run-2026-08-04_22-45-15` — the same single run
reported on 2026-08-09. Attribution coverage has grown (6 → 18 runs); the flagged population
has not moved in 17 days. Eleven attributed runs since then are all clean.

## Already done or already answered — do not rebuild

- The `ogletree-children/run-2026-07-21_13-24-05` regression fixture the old plan asks for
  **exists**:
  `eval/harness/tests/unit/test_skill_invocation.py::test_ogletree_children_hand_backfilled_regression`
  hand-backfills the three subagent spans and asserts 3 violations.
- **The allowlist question is settled by measurement, not by a decision.** Across 783
  attributed ledger entries in 18 runs, `agent_type` takes exactly five values:
  `gps-mentor` 297, `record-extractor` 201, `general-purpose` 163, `image-reader` 114,
  `image-reader-opus` 8. Four are in `DEDICATED_AGENT_NAMES`; `general-purpose` is the only
  outsider, and it is the class the rule targets. `Explore`, `claude` and
  `question-selection` are `subagent_type` values on `Agent` call *arguments*; this detector
  reads the hook-stamped `agent_type`, never that field, so they cannot flag. No allowlist
  change is needed.

## The task — build the reader, post the counts, graduate nothing

1. Add an unnamed-delegate bucket to `eval/harness/e2e/guardrail_shadow_report.py` beside the
   four existing stored-family scans, so this number is reproducible instead of hand-rolled.
   `protected_writes_by_unnamed_delegate` is a list of strings on the run, not an entry in
   `guardrail_shadow_violations`, so it needs its own scan rather than a `kind` predicate.
   Print violations, runs affected, **and how many runs carry any attribution at all** —
   without that denominator the count reads as "this check almost never fires" when it mostly
   means "this check could not fire."
2. The namespaced-spelling fix from the ruling above, in
   `eval/harness/harness/skill_invocation.py`, with a test for both spellings.
3. Sites that change with it:
   - `eval/harness/tests/unit/test_guardrail_shadow_report.py` — add a test, and break it
     once to watch it fail (`CLAUDE.md`, "A new lint must be proven to fail").
   - `Makefile:584-598` — the target's help line and comment enumerate the families it replays.
   - `docs/specs/guardrail-enforcement-spec.md` — the unnamed-delegate status row (its
     "attribution reaches 15 of 154 runs" is a 2026-08-16 figure) and "Where it surfaces",
     which currently says the detector is the only reader.
   - `docs/specs/e2e-test-spec.md` — same claim, second copy.
4. Verify with `make harness-test`, then `make e2e-guardrail-shadow SINCE=all` (pure
   analysis, no API spend, about a minute).
5. Post the counts here stamped `measured <date>, make e2e-guardrail-shadow SINCE=all`, and
   post issue #1358's bucket number from the same output on that issue.

**Do not touch `E2eResult.__post_init__`, `check_guardrail_compliance`, or
`find_protected_writes_by_unnamed_delegate`'s verdict behaviour in this PR.** The ruling is
that this check does not move the compliance axis.

## Sequencing

- PR #1741 (open, updated 2026-08-21) adds ~447 lines to `guardrail_shadow_report.py` and to
  `guardrail-enforcement-spec.md`. Land after it or rebase on it.
- The "land #1279 first" note in the history below is dead: issue #1279 closed `not planned`
  on 2026-08-10 and was merged into issue #1273, which is open and holds the same one-line
  hook fix. If #1273 lands first, a `general-purpose` delegate's `extraction_append` becomes
  a PreToolUse denial — and this detector counts the attempt regardless of `is_error` —
  so 10 of the 15 violations in the one flagged run become denied attempts rather than
  landed writes. **State which world your numbers were taken in.**

## Prerequisite (original, retained)

Ledger attribution must land first: today `tool_calls` interleaves subagent calls
with main-thread calls and records no caller. Verified — in
`ogletree-children/run-2026-07-21_13-24-05`, ledger indices 267–288 are exactly
the 22 calls the captured delegate made, in order; and across the runs with
`subagents[]` capture, parent-ledger `extraction_append` counts match subagent
counts. Until each entry carries `caller` (`agent_id` / `agent_type`), no number
here is trustworthy.

> Note 2026-07-31: this verification is what the re-scope above is based on — it
> establishes that the ordering *correlates*, not that a per-call join is
> available. The `tool_use_id` it names is absent from committed logs.

## Baseline expectation (numbers refreshed 2026-07-31)

Precision should be high. **66** general-purpose `Agent` calls across **30** of the
121 committed runs that carry a tool-call ledger; the large majority are
context-diet triage of large result files (`Read` / `Grep` only — untouched by the
rule). *(Superseded figures, 2026-07-30: 58 calls across 26 of 110 runs — the
corpus has grown, the shape has not changed.)*

Exactly one run shows a delegate performing protected writes:

`ogletree-children/run-2026-07-21_13-24-05` — verdict `pass`, judge `pass`, 345
calls, zero `same_person`:

| index | what |
|---|---|
| 214 | `Agent` (general-purpose) "Research exhaustiveness evaluation for q_001" — delegate writes the `exhaustive_declaration` at 227 |
| 266 | `Agent` (general-purpose) "Link Louise Barrett death cert assertions to tree" — delegate calls `materialize_facts` ×4, `tree_edit`, `research_append` |

Use that run as the **regression fixture**: a committed *passing* run the check
must flag.

## Graduation criteria

- Precision holds across the accumulated sample (N from Step 4) → graduate to
  hard-fail (join the §4.4 tuple at `eval/harness/e2e/orchestrator.py:1264-1269`).
- Holds again after a suite at hard-fail → consider the live `PreToolUse` deny,
  with denial text that tells the router to **report the delegation failure and
  stop** rather than retry another way (#942 ask #2).
- If precision does not hold, record why here and leave it in shadow — do not
  graduate on a hunch.

## Related

- #942 — record-extraction: subagent spawn failure must be a hard deny (same family;
  ledger attribution also answers its open step 4, whether e2e can use
  `context_policy`)
- #963 — agent skips `same_person` / conflict-resolution provenance
- #972 — e2e verdict conflates genealogical correctness with guardrail compliance
- #911 — calibrate the §4.1 shadow window (the precedent this issue exists to avoid repeating)

> **Note on the coexisting check.** @Pascal's identity-based
> `find_protected_writes_by_unnamed_delegate` does **not** replace
> `harness/skill_invocation.py:173 find_unguarded_protected_writes`. They answer
> different questions — the existing one catches a protected write with *no
> owning-skill invocation at all*; the new one catches *doctrine loaded, work
> handed to an unnamed delegate*. Coverage is additive, and #911 stays open.

### LINKED ISSUE #1273: e2e context-policy hook: four gaps in one pass (main-thread image_read, unreachable stop, loose agent_id, unclassified research_append sections)
> **Reviewed 2026-08-21 before junior handoff, and the lead has ruled on both open
> forks.** The title promised four gaps and the body specified one; all four are below,
> with items 2 and 3 settled. The original opening survives as the first two sections.

The e2e orchestrator's `PreToolUse` hook guards `extraction_append` on the main thread (#942, PR #1196) but not `image_read`. The stated reason for the exemption is stale.

## The stale reason

`eval/harness/harness/context_policy.py` and `docs/plan/image-read-context-policy.md` §4.1 say e2e cannot guard `image_read` because `search-images` declares it in its own `allowed-tools` and browses volumes page-by-page itself, so a legitimate in-session browse (no `agent_id`) is indistinguishable from a router violation.

That is no longer true:

- `packages/engine/plugin/skills/search-images/SKILL.md` declares `volume_search`, `image_search`, `research_log_append`, `research_append`. Its body says "You do not have `image_read` in your tools" (line 287).
- `docs/specs/unit-test-spec.md` records the change: search-images moved to delegating via `@plugin:image-reader` on 2026-07-17, "so no skill currently declares it".
- `eval/harness/tests/unit/test_context_policy.py::test_real_search_images_does_not_declare_image_read` already asserts it.
- `image_read` now lives only on `packages/engine/plugin/agents/image-reader-opus.md`. (`agents/image-reader.md` switched to `image_transcribe`.)

So every legitimate `image_read` today comes from a Task-spawned subagent and carries `agent_id` — the same discriminator #942 used for `extraction_append`.

## Why it matters

A main-thread `image_read` returns inline base64 into the router's context; the bytes accumulate and overflow the transport's ~1 MiB per-turn buffer and crash the run. That is the crash the whole per-context policy exists to prevent, and e2e is where the long runs are. The unit harness denies it; e2e does not.

## Scope — four items, one PR, both decisions settled

The title names four gaps. Items 2–4 were merged in from issues #1278/#1279/#1280 and were
left in a comment; they are restated here so the body is the whole task. **The lead ruled on
items 2 and 3 on 2026-08-21 — see each item. Nothing here is blocked.**

### Item 1 — enforce main-thread `image_read` in e2e

Generalize `is_main_thread_extraction_append` in `eval/harness/e2e/orchestrator.py` to cover
both members of `context_policy.SUBAGENT_ONLY_TOOLS`, and record the deny into
`blocked_context_calls` + `narration` exactly as the `extraction_append` block already does.

**Do not replace the check with a call to `context_policy.subagent_only_violation`.** It
looks like clean deduplication, every test still passes, and it changes two things silently:
it drops the `mcp__` prefix guard (a bare, unqualified `extraction_append` becomes a
violation — `test_non_mcp_tools_are_never_blocked` pins the opposite), and passing
`declared_tools=None` means "declared nothing", which removes the exemption e2e deliberately
keeps for a future skill that legitimately declares a guarded tool. Both function docstrings
say to keep the e2e check tool-set-specific; keep it that way.

The narration string and the `_emit(...)` line are currently hardcoded to
`extraction_append`'s wording. Make them per-tool. Per-tool guidance for these two tools now
lives in **three** places with presence-parity only, which is issue #1296 (icebox) — do not
fix that here, but keep the new `image_read` wording consistent with
`context_policy._DENIAL_REASONS["image_read"]` and the `owners` map in
`eval/harness/validators/test_universal.py`.

### Item 2 — the deny forbids recovery; the Stop hook vetoes stopping

**Ruling (lead, 2026-08-21): name the recovery, and leave the Stop hook alone.**

Today the two guarded tools are denied for different reasons and their deny texts are
asymmetric. `image_read`'s names a recovery — *"Delegate to the image-reader subagent
(@plugin:image-reader), which returns a text transcription."* `extraction_append`'s forbids
every recovery — *"If that subagent failed to spawn, report the spawn failure to the user and
stop. Do not extract the record and append it yourself, and do not retry another way."*

That asymmetry is the defect, not the Stop hook. The mechanism already disagrees with the
words: `subagent_only_denial`'s own docstring says it *"deliberately returns no `stopReason`:
a denied call is a recoverable"* error. And in e2e there is no user to report to, so the
instruction is unreachable — `activity_count["n"] += 1` fires for every tool call before any
deny, so a denied call counts as progress and `should_continue_run`
(`eval/harness/e2e/stop_checker.py`) vetoes the agent's stop and nudges it onward, up to
`caps.max_continue_nudges` (default 20).

**The fix is to make continuing correct rather than to stop continuing.** Rewrite
`context_policy._DENIAL_REASONS["extraction_append"]` to mirror `image_read`'s shape: name
the productive next action first — spawn `@plugin:record-extractor` and hand it this record —
and give a bounded fallback: if the spawn fails a second time on the same record, skip that
record and say so in the run summary. Keep the original prohibition intact: the router must
never extract and append the record itself, and must not retry "another way."

Chosen over threading a `blocked_since_last_nudge` flag into `should_continue_run`, which was
rejected because it hard-codes "a spawn failure ends the session" when the right recovery is
usually re-delegation — stopping loses the record as well as the run. Also chosen over
leaving the tension documented, which trains the model to ignore deny reasons.

**One objection that does not apply, so it is not re-derived.** This text is *not* shared
with production. `eval/harness/harness/context_policy.py` is imported only by
`eval/harness/harness/skill_runner.py` and `eval/harness/e2e/orchestrator.py`; the shipped
plugin hook is `packages/engine/plugin/hooks/guard_project_files.py`, the raw-write half, and
the per-context port into production is still open work. Rewording touches no Cowork
doctrine.

Sites: `eval/harness/harness/context_policy.py` (`_DENIAL_REASONS`), and
`eval/harness/tests/unit/test_context_policy.py::test_extraction_append_denial_tells_the_router_to_stop`,
which pins the current wording and must be updated — rename it to say what the deny now
does. Record in `docs/specs/e2e-test-spec.md`, in the paragraph that says the denied call
"doesn't stop the run", that continuing is now the *intended* behaviour because the deny
names a recovery. **Do not add issue numbers to that file** — see "How you know it worked".

### Item 3 — the hook accepts any `agent_id`, not just `record-extractor`

**Ruling (lead, 2026-08-21): the hook stays as it is.** The hook denies main-thread calls;
the delegate case stays shadow-logged by `find_protected_writes_by_unnamed_delegate`. That
difference is by design, not drift — `docs/specs/guardrail-enforcement-spec.md` records the
delegate half as "shadow only. Instrumented, never denied" and gates graduation on an
accumulated count decided in advance, and the whole sample today is one pre-guard run
(`antonio-lucas-spouse`, 2026-08-04) with zero occurrences in the 9 attributed runs since.
Tightening the hook would have graduated a shadow guardrail to a live deny on that one run.

Two things to do:

1. In `eval/harness/harness/skill_invocation.py`, make the
   `agent_type == "record-extractor"` comparison tolerate the namespaced spelling — strip a
   leading `<plugin>:` segment before comparing. A live Cowork probe (2026-08-15) logged
   `agent_type: "genealogy-research:image-reader"` for plugin-agent delegation while the
   harness reports bare names, so an equality test that is green in CI is dead the moment the
   detector is fed production data — the failure shape of issues #650, #698 and #939. Add a
   test for both spellings.
2. State in the `find_protected_writes_by_unnamed_delegate` docstring and in
   `guardrail-enforcement-spec.md` that hook and detector differ deliberately, so this is not
   re-filed. **Issue #980 carries the same ruling and the same one-line fix** — whichever
   lands first does it; say in the PR which, and do not do it twice.

### Item 4 — `research_append` on `sources`/`assertions` is the same write, unclassified

`owning_skills` attributes `research_append` only for `proof_summaries`, `person_evidence`,
`conflicts` and `questions`, so the payload `extraction_append` is denied for can still be
written through the broad tool with nothing flagging it.

**Add the check to `find_protected_writes_by_unnamed_delegate` as its own branch, beside the
existing `extraction_append` branch — do not widen `owning_skills`.** Its documented contract
is "which of the four `GUARDRAIL_SKILLS` owns this write", and it is consumed by
`find_unguarded_protected_writes`, `find_effects_without_invocation`,
`e2e/detector_before_after_report.py` and `e2e/skill_episode_report.py`. Widening it would
demand a preceding `Skill(record-extraction)` for every extraction write in the corpus and
would move the numbers `make e2e-corpus` and `make e2e-skill-episodes` report — the only
compliance figures this repo has. `ownership.json` also lists `citation` as a legitimate
caller of `sources` through `research_append`, so any rule keyed on the section alone
false-flags it.

This stays **shadow-mode**: logged, never denied. Record the widened rule in
`docs/specs/guardrail-enforcement-spec.md` under "The rule".

## Sites this touches

Code
- [ ] `eval/harness/e2e/orchestrator.py` — the predicate (and its docstring), the `if` in
      `pretool_hook`, the comment above that block, the narration string, the `_emit` line
- [ ] `eval/harness/harness/context_policy.py` — `_DENIAL_REASONS["extraction_append"]`
      (item 2), the module docstring paragraph beginning "e2e enforcement is **partial, and
      only for `extraction_append`**", and the comment above `SUBAGENT_ONLY_TOOLS`
- [ ] `eval/harness/harness/skill_invocation.py` — the docstring that names
      `is_main_thread_extraction_append` (rename it if you rename the predicate), the
      namespaced-spelling fix (item 3), plus item 4's new branch

Tests
- [ ] `eval/harness/tests/unit/test_e2e_context_block.py` — its module docstring, and
      `test_other_mcp_tools_on_main_are_not_blocked`, **which today asserts
      `mcp__genealogy__image_read` is *not* blocked**. That assertion inverts.
- [ ] `eval/harness/tests/unit/test_context_policy.py::test_extraction_append_denial_tells_the_router_to_stop`
      — pins the wording item 2 replaces.
- [ ] New cases: main-thread `image_read` denied under both the `mcp__genealogy__` and
      `mcp__remote-devices__Genealogy_Research__` spellings; a call carrying `agent_id`
      untouched; the deny recorded in `blocked_context_calls`.
- [ ] Folded in from issue #1295: a test asserting the **e2e-enforced set equals the whole
      `SUBAGENT_ONLY_TOOLS` set**, so a future third member cannot be silently left
      unguarded in e2e.

Docs
- [ ] `docs/specs/e2e-test-spec.md` — the section heading, which names `extraction_append`;
      the `blocked_context_calls[]` row in the result-envelope table, which reads "Denied
      main-thread `extraction_append`"; and the "doesn't stop the run" paragraph (item 2).
- [ ] `.claude/skills/interpret-e2e-result/SKILL.md` — "the write-side twin" and "Each entry
      is a main-thread `extraction_append`" are both now wrong (`image_read` is a read). This
      is a Claude Code skill, **not** a shipped plugin skill, so it buys no eval run.
- [ ] `docs/plan/image-read-context-policy.md` — its `**Status:**` line still says "proposed
      (2026-07-16)" for work that has now shipped in both harnesses. Update it. Do **not**
      delete the file without first relocating its probe evidence: `context_policy.py`'s
      docstring and `eval/harness/validators/test_universal.py` both cite it.

## How you know it worked

- `make harness-test` — the whole Python suite; 44 tests across `test_e2e_context_block.py` +
  `test_context_policy.py` pass on `main` today, so a red one is yours.
- `make engine-test` — **required even though this is a Python change.** Editing
  `docs/specs/e2e-test-spec.md` or `docs/specs/guardrail-enforcement-spec.md` trips the
  exact-match issue-reference ratchet in
  `packages/engine/mcp-server/tests/packaging/doc-links.test.ts` (baselines 10 and 16, and
  both files sit exactly on their baseline). **Do not add `#1273` or `#942` to either spec**,
  and if an edit removes an existing ref, lower that file's `ISSUE_REF_BASELINE` entry in the
  same PR.
- `make test-all` before opening the PR.
- **No live e2e run is required, and none proves this.** Nothing in CI shows the deny firing
  in a real run; the guard has never fired in e2e. Say that in the PR rather than implying CI
  covered it.
- **This PR triggers `check-runlogs` (it touches `eval/harness/**`) but gates no skill**,
  because touched-skill detection keys on `packages/engine/plugin/skills/`,
  `eval/tests/unit/` and `packages/engine/plugin/agents/`. **No paid eval run, no genealogist
  annotation.**
- For item 4 only: run `make e2e-corpus` before and after your change and put both outputs in
  the PR body. A widened detector moves numbers other people quote.

## What the corpus actually shows

Re-run before you argue about priority:

```sh
python3 - <<'PY'
runs = [r for r in sorted(glob.glob("eval/runlogs/e2e/*/run-*.json"))
        if ".ann." not in r and ".final-" not in r]
c = collections.Counter(); attributed = 0
for p in runs:
    d = json.load(open(p, encoding="utf-8"))
    tcs = [t for t in (d.get("tool_calls") or []) if isinstance(t, dict)]
    if not any("agent_id" in t for t in tcs):
        continue                      # predates per-call attribution: cannot tell
    attributed += 1
    for t in tcs:
        n = t.get("tool") or ""
        b = n.rsplit("__", 1)[-1] if "__" in n else n
        if b in ("image_read", "extraction_append"):
            c[(b, "MAIN-THREAD" if t.get("agent_id") is None else t.get("agent_type"))] += 1
print(f"runs scanned {len(runs)}, of which attributed {attributed}")
for k, v in sorted(c.items()): print(f"  {v:4}  {k[0]:18} {k[1]}")
PY
```

Measured 2026-08-21: 157 runs, 18 carrying per-call attribution. Every `image_read` in an
attributed run came from `image-reader-opus`; **zero main-thread**. The 3 main-thread and 10
`general-purpose` `extraction_append` calls are all in runs committed *before* the guard
landed (2026-08-05); in the 9 attributed runs since, all 55 `extraction_append` calls came
from `record-extractor`. What moves these numbers is more attributed runs, so re-run rather
than quoting this line.

**The crash is real anyway, and it is in the corpus once.**
`eval/runlogs/e2e/ivyl-greenley-daughter/run-2026-06-29_22-24-03.json` ended
`stop_reason: "error"` with `JSON message exceeded maximum buffer size of 1048576 bytes`. Its
last two calls are `image_read` at `tool_calls[155]` and `[156]`; the run has no `subagents`
at all and its only `Agent` call is `gps-mentor` at `[140]`, and at `[146]` the main thread
ran `ToolSearch` with `select:mcp__genealogy__image_read`. That run predates attribution, so
`agent_id` is structurally absent rather than observably null — but the adjacency is the
shape this guard exists to stop, and the failure destroys its own evidence (a run killed
mid-stream may write nothing), so a zero count is weak reassurance.

## Sequencing

PR #1747 touches `eval/harness/e2e/orchestrator.py` and `docs/specs/e2e-test-spec.md`, but
its hunks do not overlap the hook block, the predicate, or the section this edits — expect
line-shift conflicts only. Do after it merges if it is close. Sibling issue #1281 is a
measure-then-decide task on the same `blocked_context_calls` field and the same
`.claude/skills/interpret-e2e-result/SKILL.md` — do not do its measurement here, but say in
the PR that the field is still read by nothing in e2e.

## Absorbed: #1159 (closed 2026-08-09) — the exemption's disposition

#1159 ("`image_read` context-policy exemption is unreachable and its test asserts a
fiction") is merged here. **Its two named defects have both since been fixed in place**;
what survived was the open disposition question, and this issue answers it.

State re-verified against `main` on 2026-08-09:

- **#1159 defect 1 — `docs/plan/image-read-context-policy.md` §4.1 is factually wrong.**
  **Fixed.** §4.1 now opens with a "Superseded as of 2026-08-04" banner recording that
  `search-images` moved to delegating via `@plugin:image-reader` (2026-07-17), that no
  skill declares `image_read`, that the exemption is unreachable, and that `agent_id`
  alone discriminates — and it already points here. `context_policy.py`'s module docstring
  (`:35-36`, `:124`) carries the same correction.
  **Still to do in this issue:** the *comment above the block* in `e2e/orchestrator.py` and
  `docs/specs/e2e-test-spec.md` §6.1.1, both already in the Acceptance list above.
- **#1159 defect 2 — the regression test validates a fiction.** **Fixed.**
  `eval/harness/tests/unit/test_context_policy.py:113` now labels the hardcoded set as
  *"the declared-tools exemption (synthetic; no skill exercises it today)"* and explains
  why, and `test_real_search_images_does_not_declare_image_read` (`:305`) reads the actual
  `allowed-tools` frontmatter, so a future edit to either SKILL.md fails loudly.

**The surviving question — "delete the exemption if it is genuinely dead" — is answered
here, in the negative.** Per this issue's own text: *keep the unit harness's declared-tools
exemption as-is; it is the mechanism for a future skill that legitimately declares a
guarded tool, and nothing depends on it being unreachable.* Do not delete it as part of
this work.

Related and still open: #1130 (whether Cowork's router holds `image_read` at all, and
whether the plugin hook should deny it) — a separate surface, not covered here.

clack391:
## Summary

- Adds a §11 unnamed-delegate bucket to `make e2e-guardrail-shadow` so the count is reproducible instead of hand-rolled: `scan_unnamed_delegate` reads each run's `protected_writes_by_unnamed_delegate` list (a stored read) and, under `REPLAY=1`, recomputes it over `tool_calls`, printing violations, runs affected, and the attribution denominator (how many runs carry any caller attribution to fire on at all).
- Fixes the `agent_type` comparisons in `find_protected_writes_by_unnamed_delegate` to tolerate a plugin-namespaced spelling: it strips a leading `<plugin>:` before comparing, so `genealogy-research:record-extractor` behaves as bare `record-extractor` on both the `== "record-extractor"` clause and the `in DEDICATED_AGENT_NAMES` clause. A live Cowork probe logs the namespaced form while the harness reports bare names, so the equality test was green in CI and dead against production data (the #650/#698/#939 shape).
- Graduates nothing. Per the lead ruling (2026-08-21) the check stays shadow, reported, no graduation count; the spec status row and reasoning say so, including that the hook (denies main-thread) and this detector (shadow-logs the delegate half) differ deliberately.
- Lands #1273's Item 3 only (the shared namespaced-`agent_type` fix plus the deliberate hook-vs-detector statement in the docstring and spec). The lead's ruling on that item says #980 and #1273 carry the same one-line fix and whichever lands first does it, so #1273's implementer should skip Item 3. Items 1, 2, and 4 of #1273 (main-thread `image_read` in e2e, the deny-reason rewrite, the `research_append` branch) stay open, so this does not close #1273.
- Tier: Normal.

## Review guidance

**Start here:** the namespaced fix in `skill_invocation.py` (one `bare_agent_type` normalization, used on both clauses; raw `agent_type` kept in the messages) and `scan_unnamed_delegate` in `guardrail_shadow_report.py`. Confirm it does not touch the compliance axis: the changed detector still feeds only the shadow field, never `guardrail_bypass_violations`.

**Unsure about:** nothing blocking. The stored and replayed counts agree today; the replay half exists because it is the only one that reflects this detector change going forward.

## Plan

**Didn't change:** `E2eResult.__post_init__`, `check_guardrail_compliance`, and the detector's compliance-axis participation (the ruling: this check does not move that axis); `test.holdout` in any suite (removing it would flip run logs inactive); `DEDICATED_AGENT_NAMES` (allowlist settled by measurement, no change); the ogletree regression fixture (already exists).

**Acceptance check:** `make harness-test` (2644 passed, 2 skipped). New guards proven to fail: reverting the strip fails `test_namespaced_record_extractor_treated_as_bare`; corrupting the denominator or the stored read fails `test_scan_unnamed_delegate_counts_stored_affected_and_attribution`; dropping the denominator from the formatter fails both `test_format_unnamed_delegate_always_prints_the_attribution_denominator` and the main-integration test; and gating the §11 print behind `--replay` fails `test_stored_unnamed_delegate_line_prints_in_main_without_replay` (the default-command guard, added in the second review pass).

**Deviated from the plan:** added a REPLAY half to the reader (the module ships stored+replay for every family and its own docstring says "read the replay before concluding a check never fires"; only the replay reflects the namespaced fix). Flagged because the issue's step 1 named the stored read; the replay is additive and near-free.

## The counts (measured 2026-08-24, `make e2e-guardrail-shadow SINCE=all REPLAY=1`, main @ c44350dc)

§11 unnamed-delegate: 15 protected writes attributed to an unnamed delegate, across 1 run, of 20 runs that carry any caller attribution at all, 159 scanned. Stored and replayed agree (15 / 1). The one flagged run is `antonio-lucas-spouse/run-2026-08-04_22-45-15`, unmoved for weeks. Taken with #1273 still open, so all 15 are landed writes (the detector counts the attempt regardless of `is_error`); if #1273 lands first, 10 of the 15 become denied attempts.

## Test plan

- [x] `make harness-test` ran and passed (2644 passed, 2 skipped).
- Review pass 1: an author-side review added output-coverage tests for the §11 formatter and `main()` wiring (the denominator must always print, so it cannot be deleted green), and handled two adversarial nits: documented that the plugin-prefix strip takes one segment (a double-namespace over-flags, not over-exempts), and softened a spec cross-reference from "replayed old-vs-new by" to "read by".
- Review pass 2: a second author-side review (three blind agents: reproduce, open-break, completeness) caught that the §11 family enumeration was updated in the Makefile but not in three parallel copies (the argparse `--help`, the spec's "REPLAY=1 recomputes all four post-hoc families" paragraph, and the module docstring), and that the §11 default-path print had no test. Fixes: bring all three enumerations to five families and correct the docstring's "committed final state" framing (the #963 and §11 checks replay from `tool_calls`); add the default-command guard above; and harden the detector so a non-str `agent_type` maps to `None` rather than raising `TypeError` in the `DEDICATED_AGENT_NAMES` membership test under a corpus-wide replay (pre-existing, unreachable on real SDK data).
- Review pass 3: a repo-wide re-review caught that pass 2's enumeration fix was itself incomplete (the file-scoped grep missed a fourth surface): `docs/plan/enforcement-layer-phases.md` had two more copies of "all four post-hoc families". Dropped the hardcoded count there so it cannot drift again. Also added `test_unhashable_agent_type_flags_without_raising`: the pass-2 `else None` hardening had no test, so reverting it kept the suite green; the new guard fails with `TypeError` when reverted. Break-tested.
- Review pass 4 (Francis's review): the strip landed on the live detector but not on `_lane_check_old`, the replica behind `make e2e-detector-diff DETECTOR=lane-check` (the second reader this PR's spec edit named), so that old-vs-new diff would attribute a namespaced value to the unrelated #1569 fix. Zero effect on today's corpus (no namespaced value in the 159 runs). Fixed by lifting the strip into `strip_agent_namespace()` and calling it from both the live detector and the replica (CLAUDE.md Code reuse), with a break-tested guard that a namespaced agent produces no old-vs-new divergence on either clause.
- [x] `make e2e-guardrail-shadow SINCE=all REPLAY=1` ran (pure analysis, no API), counts above.
- No skill / agent / rubric / unit-test snapshot changed, so no `make eval-skill` run and no annotation.

## Follow-on issues

**Folded in / filed instead:** none filed. While correcting the "only reader" prose I found the detector already had a second reader, `make e2e-detector-diff DETECTOR=lane-check` (#1569); the spec now names both rather than claiming inline-only.

## Sequencing

- PR #1741 (open) edits both `guardrail_shadow_report.py` and `guardrail-enforcement-spec.md`. Built off main; land after #1741 or rebase on it.
- Numbers taken pre-#1273 (open). Stated above.

Closes #980.


francis-2008-happy (approved):
Approving. I re-ran everything rather than reading it, and every claim in the body holds.

**The five acceptance guards all bite.** I mutated each one and confirmed the named test fails:

| Mutation | Result |
|---|---|
| revert the namespace strip | `test_namespaced_record_extractor_treated_as_bare` fails |
| revert `else None` to `else agent_type` | `test_unhashable_agent_type_flags_without_raising` fails |
| flip the attribution denominator's predicate | `test_scan_unnamed_delegate_counts_stored_affected_and_attribution` fails |
| delete the denominator from the formatter | `test_format_unnamed_delegate_always_prints_the_attribution_denominator` fails |
| gate the §11 print behind `--replay` | `test_stored_unnamed_delegate_line_prints_in_main_without_replay` fails |

**The counts reproduce exactly.** Running the report myself: 15 protected writes across 1 run, of 20 runs carrying any caller attribution, 159 scanned — stored and replayed agreeing at 15/1. I also scanned the corpus independently and got the same 159/20, so the denominator the code computes is the denominator the corpus supports.

**The compliance axis is untouched**, which is what Start here asked me to confirm. `protected_writes_by_unnamed_delegate` appears in `result.py` only as a comment and a field declaration; `compliance` is set solely from `guardrail_bypass_violations`.

Also checked: `encoding="utf-8"` present on every `read_text`/`open` in all four changed Python files (AST-checked, not grepped); the #1741 sequencing note is accurate, it is open and overlaps on exactly the two files named; and the doc-links ratchet commit keeps every fact while dropping only ticket numbers, which is what that ratchet's own doctrine asks for. Suite: 2634 passed, 2 skipped, 9 failed — all 9 are `test_setup_feedback_case.py` failing on `unzip: command not found` on my machine, and 2634 + 9 is your 2643.

The namespace fix is correct in all three directions I could construct: a single colon strips, a double colon over-flags rather than over-exempts, and a non-str maps to `None` and flags. Safe direction each time.

---

**One thing for you, non-blocking, and I am not asking for a change before merge.**

`e2e/detector_before_after_report.py` lines 76 and 82 still carry the raw comparison:

```python
if agent_id is None or agent_type == "record-extractor":
...
if not owners or agent_id is None or agent_type in DEDICATED_AGENT_NAMES:
```

`_lane_check_old` is framed as current logic plus the one #1569 `is_error` bug, so the old-vs-new diff isolates that fix. Now that the live detector strips namespaces, the two disagree on a namespaced value for a reason unrelated to #1569:

```
OLD replica: ["tool_calls[0] extraction_append by 'genealogy-research:record-extractor'"]
NEW detector: no violation
```

The effect today is zero — I scanned all 159 committed run logs and there is no namespaced value anywhere (`gps-mentor`, `record-extractor`, `general-purpose`, `image-reader`, `proof-conclusion`, `image-reader-opus`). It bites only once Cowork-sourced runs land, which is the same horizon the fix itself is aimed at.

I am leaving the call to you rather than guessing: either mirror the strip into `_lane_check_old` so its diff stays isolated to `is_error`, or leave it and note in the docstring that the replica now trails the detector by one correction. #1273 Item 3 named one file and you changed that file, so this is a discovery rather than a missed site — worth a line in the body either way, since you already updated the spec prose to name that second reader.

clack391:
@francis-2008-happy thanks, good catch. Took option (a): mirrored the strip into `_lane_check_old` so the old-vs-new diff stays isolated to the #1569 is_error fix. Rather than paste the two lines, I lifted the strip into `strip_agent_namespace()` in skill_invocation.py and call it from both the live detector and the replica, so they cannot drift again (CLAUDE.md Code reuse). Added a break-tested guard (`test_lane_check_old_mirrors_the_namespace_strip_no_divergence_on_a_namespaced_agent`) that a namespaced record-extractor produces no old-vs-new divergence on either clause, and named the second site in the body. Pushed as c939c57e.

jud-sdev (approved):
I went through this by running rather than reading, and it holds up.

I ran the three touched suites (198 pass) and then broke each of the load-bearing guards one at a time to confirm they bite: reverting the namespace strip fails `test_namespaced_record_extractor_treated_as_bare`, reverting `else None` to `else agent_type` makes the unhashable case raise, and gating the §11 print behind `--replay` drops it from the default command and fails its test. I restored each one after.

The count in the spec reproduces exactly. Running the report over the real corpus gives 15 writes across 1 run, of 20 runs carrying any caller attribution, 159 scanned — stored and replayed agreeing at 15 — which is what the table now says.

The namespace fix is complete across all three comparison sites: the live detector, the `_lane_check_old` replica, and `_lane_check_new` which just calls the live detector so it inherits the strip for free. Keeping §11 reported and never a gate is the right call on a single flagged run. `encoding="utf-8"` is on every changed file read/write.

One note for whoever merges: francis approved 06d15a58, and the `_lane_check_old` mirroring landed after that in c939c57e — which is exactly the fix francis asked for, and I checked it's correct. Their approval predates it, so this review is the one that reads the current head.

Full harness suite is 2637 passed; the single failure is the mock MCP test hitting a missing engine build in a fresh worktree, which is the documented environment case, not this change. The branch is behind main now, so it needs an update before it can merge.

PR #1857, opened 2026-08-24
A durable image_transcribe reachability report (part of #1594) the measurement half only

aghadiayeamayanvboernest:
Part of #1594 — the measurement half only. **Does not close it.**

## Scope, measured rather than assumed

I checked each of this issue's four parts against main before starting:

| Part | Startable? |
|---|---|
| Reachability **classification** | **No.** There are **0** cause-coded failures in the entire corpus. #1785 merged 2026-08-20; one e2e run has landed since and carries none. |
| Retry decision | No — lead's call, downstream of that classification |
| OCR quality (fabricated headings, corrupted patronymics) | No — genealogist lane, needs a fidelity harness first |
| **Fix the measurement** (@DallanQ, 2026-08-23) | **Yes** — no gate, and the freshest instruction on the issue |

## The instrument shrank, and the number went *down*

`b065b687` ("e2e retention: strip `response_summary` past 14 days") landed after this issue was filed. The issue's snippet greps `response_summary`, so a stripped call can never match its numerator — but it stays in its denominator:

| | |
|---|---|
| Naive re-run today | 30 / 394 = **7.6%** — reads as the problem receding |
| Over calls that can still be measured | 30 / 175 = **17.1%** |
| Issue headline (2026-08-13) | 26 / 267 = 9.7% |

134 of 159 run logs are stripped; 219 of 394 calls carry no outcome at all. The true corpus rate is **not recoverable** from the working tree — bounded only by [7.6%, 63.2%].

**There is no durable field to switch to.** I checked the obvious candidate: `is_error` survives the strip, but every call carrying it is *already unstripped* — the 219 stripped calls predate schema v3 and have neither field. It also over-counts, since `Unrecognized ark` is `is_error: true` but is a content error, not reachability (2 real instances). So the 14-day horizon is genuine, not a choice.

## What ships

`make e2e-transcribe-failures` → `eval/harness/e2e/image_transcribe_report.py`.

Three properties, each because their absence is what let the number rot:

1. **No rate without its denominator.** `30 of 175 measurable`, never a bare percent.
2. **The stripped tally is a first-class line**, so a shrinking corpus reads as shrinking evidence.
3. **"exact" is only claimed when nothing in range was stripped.** Overstating the uncertainty is the same failure as understating it — my first cut printed "unrecoverable" even with zero stripped calls, and the test caught it.

## The finding this deliberately does *not* overclaim

The by-author split clusters: `Gennecis` 11/43 (26%), `taiwo-stack` 13/37 (35%) against `EdmondOware` 3/70 (4%) and five operators at zero.

That resembles `wiki_failure_report`'s result — where unreachable failures tracked the operator, not the calendar — but **it is not the same shape.** That finding rests on runs *interleaving*: one operator failing while another succeeds in the same window. Here only three days carry more than one operator, and on the single one with any failures **both** operators failed.

So this corpus cannot separate machine-side from service-side, and `interleaving_verdict()` prints exactly that, in words, above the table. The clustering is a lead, not a verdict.

## Reuse

Modelled on `wiki_failure_report.py` (#1552), which already solved this class for the wiki tools — same posture (committed JSON only, no model, no spend), same stripped-excluded-from-the-taxonomy handling, same 14-day window as its five aggregating siblings.

It reuses that module's `resolve_authors` (one `git log` rather than one subprocess per run log, ~60s) instead of copying it. It adds **no** seventh copy of the bare-tool-name helper — there are already six across the harness with three different guards, and `.endswith` matches every server spelling here. That duplication is worth a separate look; one of the six lives in `context_policy.py`, a live PreToolUse guardrail rather than a report, which is why I didn't fold it into a reporting PR.

## Proven to fail

Not assumed — mutated and watched:

| Mutation | Result |
|---|---|
| Drop the `captures_stripped` exclusion | **2 tests red** |
| Fold an unreadable outcome into `success` | **1 red** |
| Count `Unrecognized ark` as unreachable | **1 red** |

All 11 tests build their own run logs in a tmpdir. **Nothing asserts against the committed corpus** — its numbers move with every e2e run that lands.

## What this does not prove

`make harness-test` proves the report's own logic. **Nothing checks that the rate it prints is true** — no test compares a transcription or a live OpenRouter call to anything. That is the `nothing-checks` half of #1594 and stays open on the issue.

Gate: `make test-all` — all checks passed.


johnmarkpeterbrown (approved):
> This is good work — the scope table, the mutation evidence, and the refusal to read the by-author clustering as a verdict are all right. I re-derived every number and they check out. Three things before it merges, all small.
>
> ### 1. Nothing tests the `timeout` bucket
>
> Six of the corpus's thirty reachability failures are timeouts. I removed `TIMEOUT` from the reachability tuple and all eleven tests still passed, while the headline dropped from 17.1% to 13.7%. I then broke the timeout matcher and all eleven passed again.
>
> The second one is the one to worry about. A real timeout response also contains `{"error":...}`, so a broken matcher doesn't land in `unclassified` where you'd see it. It lands in `upstream_error` — a bucket you define as *not* reachability — and leaves the number silently.
>
> **You wrote:**
> ```python
> from e2e.image_transcribe_report import (
>     SUCCESS,
>     UNCLASSIFIED,
>     UNREACHABLE,
>     UNRECOGNIZED_ARK,
>     classify,
> ```
> **Change it to:**
> ```python
> from e2e.image_transcribe_report import (
>     SUCCESS,
>     TIMEOUT,
>     UNCLASSIFIED,
>     UNREACHABLE,
>     UNRECOGNIZED_ARK,
>     UPSTREAM_ERROR,
>     classify,
> ```
>
> **You wrote** (end of the classification block, after `test_a_call_is_matched_under_every_server_spelling`):
> ```python
> def test_one_unreadable_run_log_does_not_take_the_report_down(tmp_path: Path):
> ```
> **Change it to:**
> ```python
> def test_a_timeout_counts_as_a_reachability_failure(tmp_path: Path):
>     """A 180s OCR timeout never got OpenRouter's answer, so it is reachability —
>     6 of the corpus's 30. Its real text also carries `{"error":...}`, so dropping
>     it from the bucket does NOT surface as `unclassified`: it silently becomes
>     `upstream_error` and leaves the headline."""
>     timed_out = (
>         '{"error":"Request to https://openrouter.ai/api/v1/chat/completions '
>         'timed out after 180000ms while reading the response body."}'
>     )
>     assert classify(timed_out) == TIMEOUT
>
>     p = _run(tmp_path / "fix", "run-2026-08-20_00-00-00.json", [timed_out, "ok"])
>     r = scan([p], author_of=_authors({}))
>     assert r.reachability_failures == 1
>     assert "1 of 2 measurable" in format_report(r)
>
>
> def test_one_unreadable_run_log_does_not_take_the_report_down(tmp_path: Path):
> ```
>
> ### 2. The generic error check misses the escaped shape
>
> Your `classify` docstring is right that the matchers have to work on both shapes. The three specific ones do. The last two don't — in the escaped shape the text reads `\"error\":\"…\"`, which contains neither `"error"` nor `error:`.
>
> ```
> classify('[{"type": "text", "text": "{\"error\":\"OpenRouter OCR failed: 502\"}"}]')  ->  success
> classify('{"error":"OpenRouter OCR failed: 502"}')                                    ->  upstream_error
> ```
>
> 52 of the 175 measurable calls are in that escaped shape. None carries an error today, so no printed number changes. But "an unknown shape never lands in `success`" is the guarantee this classifier is for, and right now it only holds for the other 70%.
>
> **You wrote:**
> ```python
>     s = (response_summary or "").lower()
> ```
> **Change it to:**
> ```python
>     s = (response_summary or "").lower().replace('\\"', '"')
> ```
>
> **You wrote** (end of `test_an_unknown_shape_lands_in_unclassified_never_in_success`):
> ```python
>     assert classify("Could not reach OpenRouter. (fetch failed)") == UNREACHABLE
> ```
> **Change it to:**
> ```python
>     assert classify("Could not reach OpenRouter. (fetch failed)") == UNREACHABLE
>
>
> def test_an_error_is_caught_in_the_escaped_envelope_too():
>     """`response_summary` has two shapes and 52 of 175 measurable calls use the
>     escaped one, where the document's keys read `\\"error\\"`. A matcher that only
>     sees the unwrapped form files an unknown failure as `success` — the defect
>     this classifier exists to prevent, one level down."""
>     unwrapped = '{"error":"OpenRouter OCR failed: 502 Bad Gateway"}'
>     escaped = (
>         '[{"type": "text", "text": '
>         '"{\\"error\\":\\"OpenRouter OCR failed: 502 Bad Gateway\\"}"}]'
>     )
>     assert classify(unwrapped) == UPSTREAM_ERROR
>     assert classify(escaped) == UPSTREAM_ERROR, "the escaped form must not read as success"
> ```
>
> ### 3. The report prints 3 `unrecognized_ark`, not 2
>
> Three exist. Two carry `is_error`; the third predates that field. The sentence reads as a count of the error itself, which makes your own tool contradict your docstring.
>
> **You wrote:**
> ```
> `is_error: true` and is a content error, not reachability (2 real instances in
> the corpus). So the horizon is genuinely the 14-day capture window, and
> ```
> **Change it to:**
> ```
> `is_error: true` and is a content error, not reachability (3 in the corpus; the
> 2 that postdate schema v3 carry the field). So the horizon is genuinely the
> 14-day capture window, and
> ```
>
> **You wrote:**
> ```python
>     reachability count is why an `is_error`-only measurement over-counts; there
>     are 2 real instances in the committed corpus."""
> ```
> **Change it to:**
> ```python
>     reachability count is why an `is_error`-only measurement over-counts; there
>     are 3 in the committed corpus, 2 of them carrying `is_error`."""
> ```
>
> ### Not blocking — your call
>
> `interleaving_verdict` treats "zero reachability failures" as "succeeded". An operator whose calls were all `unrecognized_ark` would satisfy that, and the verdict would print that one operator succeeded while another failed. No day qualifies today so it never fires. Worth a look given this is the function that exists to stop exactly that kind of over-claim.

johnmarkpeterbrown (changes requested):
> This is good work — the scope table, the mutation evidence, and the refusal to read the by-author clustering as a verdict are all right. I re-derived every number and they check out. Three things before it merges, all small.
>
> ### 1. Nothing tests the `timeout` bucket
>
> Six of the corpus's thirty reachability failures are timeouts. I removed `TIMEOUT` from the reachability tuple and all eleven tests still passed, while the headline dropped from 17.1% to 13.7%. I then broke the timeout matcher and all eleven passed again.
>
> The second one is the one to worry about. A real timeout response also contains `{"error":...}`, so a broken matcher doesn't land in `unclassified` where you'd see it. It lands in `upstream_error` — a bucket you define as *not* reachability — and leaves the number silently.
>
> **You wrote:**
> ```python
> from e2e.image_transcribe_report import (
>     SUCCESS,
>     UNCLASSIFIED,
>     UNREACHABLE,
>     UNRECOGNIZED_ARK,
>     classify,
> ```
> **Change it to:**
> ```python
> from e2e.image_transcribe_report import (
>     SUCCESS,
>     TIMEOUT,
>     UNCLASSIFIED,
>     UNREACHABLE,
>     UNRECOGNIZED_ARK,
>     UPSTREAM_ERROR,
>     classify,
> ```
>
> **You wrote** (end of the classification block, after `test_a_call_is_matched_under_every_server_spelling`):
> ```python
> def test_one_unreadable_run_log_does_not_take_the_report_down(tmp_path: Path):
> ```
> **Change it to:**
> ```python
> def test_a_timeout_counts_as_a_reachability_failure(tmp_path: Path):
>     """A 180s OCR timeout never got OpenRouter's answer, so it is reachability —
>     6 of the corpus's 30. Its real text also carries `{"error":...}`, so dropping
>     it from the bucket does NOT surface as `unclassified`: it silently becomes
>     `upstream_error` and leaves the headline."""
>     timed_out = (
>         '{"error":"Request to https://openrouter.ai/api/v1/chat/completions '
>         'timed out after 180000ms while reading the response body."}'
>     )
>     assert classify(timed_out) == TIMEOUT
>
>     p = _run(tmp_path / "fix", "run-2026-08-20_00-00-00.json", [timed_out, "ok"])
>     r = scan([p], author_of=_authors({}))
>     assert r.reachability_failures == 1
>     assert "1 of 2 measurable" in format_report(r)
>
>
> def test_one_unreadable_run_log_does_not_take_the_report_down(tmp_path: Path):
> ```
>
> ### 2. The generic error check misses the escaped shape
>
> Your `classify` docstring is right that the matchers have to work on both shapes. The three specific ones do. The last two don't — in the escaped shape the text reads `\"error\":\"…\"`, which contains neither `"error"` nor `error:`.
>
> ```
> classify('[{"type": "text", "text": "{\"error\":\"OpenRouter OCR failed: 502\"}"}]')  ->  success
> classify('{"error":"OpenRouter OCR failed: 502"}')                                    ->  upstream_error
> ```
>
> 52 of the 175 measurable calls are in that escaped shape. None carries an error today, so no printed number changes. But "an unknown shape never lands in `success`" is the guarantee this classifier is for, and right now it only holds for the other 70%.
>
> **You wrote:**
> ```python
>     s = (response_summary or "").lower()
> ```
> **Change it to:**
> ```python
>     s = (response_summary or "").lower().replace('\\"', '"')
> ```
>
> **You wrote** (end of `test_an_unknown_shape_lands_in_unclassified_never_in_success`):
> ```python
>     assert classify("Could not reach OpenRouter. (fetch failed)") == UNREACHABLE
> ```
> **Change it to:**
> ```python
>     assert classify("Could not reach OpenRouter. (fetch failed)") == UNREACHABLE
>
>
> def test_an_error_is_caught_in_the_escaped_envelope_too():
>     """`response_summary` has two shapes and 52 of 175 measurable calls use the
>     escaped one, where the document's keys read `\\"error\\"`. A matcher that only
>     sees the unwrapped form files an unknown failure as `success` — the defect
>     this classifier exists to prevent, one level down."""
>     unwrapped = '{"error":"OpenRouter OCR failed: 502 Bad Gateway"}'
>     escaped = (
>         '[{"type": "text", "text": '
>         '"{\\"error\\":\\"OpenRouter OCR failed: 502 Bad Gateway\\"}"}]'
>     )
>     assert classify(unwrapped) == UPSTREAM_ERROR
>     assert classify(escaped) == UPSTREAM_ERROR, "the escaped form must not read as success"
> ```
>
> ### 3. The report prints 3 `unrecognized_ark`, not 2
>
> Three exist. Two carry `is_error`; the third predates that field. The sentence reads as a count of the error itself, which makes your own tool contradict your docstring.
>
> **You wrote:**
> ```
> `is_error: true` and is a content error, not reachability (2 real instances in
> the corpus). So the horizon is genuinely the 14-day capture window, and
> ```
> **Change it to:**
> ```
> `is_error: true` and is a content error, not reachability (3 in the corpus; the
> 2 that postdate schema v3 carry the field). So the horizon is genuinely the
> 14-day capture window, and
> ```
>
> **You wrote:**
> ```python
>     reachability count is why an `is_error`-only measurement over-counts; there
>     are 2 real instances in the committed corpus."""
> ```
> **Change it to:**
> ```python
>     reachability count is why an `is_error`-only measurement over-counts; there
>     are 3 in the committed corpus, 2 of them carrying `is_error`."""
> ```
>
> ### Not blocking — your call
>
> `interleaving_verdict` treats "zero reachability failures" as "succeeded". An operator whose calls were all `unrecognized_ark` would satisfy that, and the verdict would print that one operator succeeded while another failed. No day qualifies today so it never fires. Worth a look given this is the function that exists to stop exactly that kind of over-claim.

---

clack391 (changes requested):
Reviewed on `066ae13c`, as a second reviewer alongside johnmarkpeterbrown. This is good work: the reachability headline is correct and re-derives exactly (30 of 175, 17.1%), the stripped-denominator handling is right, the empty-corpus path refuses a false 0%, and the accounting all checks out. I ran the numbers independently and they match.

First, to avoid duplicating John: his three points (the untested timeout bucket, the escaped-envelope error reading as success, and the unrecognized_ark count being 3 not 2) all still stand on this head and are not yet addressed, so keep those. I confirmed each of them independently. Below are three more I found on top of his, so you can fix everything in one round. All three are holes in the one guarantee this tool exists for: an outcome it cannot actually see must never be counted as a success, the report must survive one bad file, and the by-author verdict must be honest.

## 1. A pending call reads as a success (blocking)

In `scan`, a call whose `response_summary` is null or missing becomes the string `"null"`, which then classifies as `success`:

```python
rs = (tc or {}).get("response_summary")
rs = rs if isinstance(rs, str) else json.dumps(rs)
```

`json.dumps(None)` is `"null"`, which is non-empty, so it sails past `classify`'s empty guard into the success fallback. This is reachable, not just theoretical: the orchestrator creates every tool-call entry with `response_summary: None` and only fills it when the result streams back, so a run that hits its wall-clock or tool cap after an `image_transcribe` call starts but before its result lands commits a null summary. `image_transcribe` is the slowest tool (180s OCR), so it is the likeliest one to be in flight when a run is cut off. Your sibling `wiki_failure_report.py` defaults to `unclassified` here, so the same input is safe there; this one defaults to `success`, so the empty guard is the only protection and `json.dumps` defeats it.

```
You wrote:      rs = rs if isinstance(rs, str) else json.dumps(rs)
Change it to:   rs = rs if isinstance(rs, str) else ("" if rs is None else json.dumps(rs))
```

That routes a null summary through the empty guard to `unclassified` while still serialising a genuinely structured (dict) summary. Please add a test that feeds a null summary on a non-stripped call and asserts it is not a success; none of the eleven covers it.

## 2. One non-dict run log crashes the whole report (blocking)

The `except` catches `(OSError, UnicodeDecodeError, json.JSONDecodeError)`, and the comment and `test_one_unreadable_run_log_does_not_take_the_report_down` both promise one bad file cannot take the report down. But a file that parses to valid JSON that is not an object (a top-level `[]`, a bare string, a number) reaches `doc.get("captures_stripped")` and raises `AttributeError`, which is not caught, and the whole scan dies. A `tool_calls` entry that is a truthy non-dict crashes the same way, since `tc or {}` only rescues a falsy value. I reproduced both. The robustness test only feeds invalid UTF-8, so it passes while the exact case the guard exists to catch still crashes.

```
You wrote:      doc = json.loads(p.read_text(encoding="utf-8"))
                except (OSError, UnicodeDecodeError, json.JSONDecodeError):
Change it to:   doc = json.loads(p.read_text(encoding="utf-8"))
                if not isinstance(doc, dict):
                    raise ValueError("run log is not a JSON object")
                except (OSError, UnicodeDecodeError, json.JSONDecodeError, ValueError):
```

And guard the per-call loop:

```
You wrote:      for tc in doc.get("tool_calls") or []:
                    if not str((tc or {}).get("tool") or "").endswith(TOOL_SUFFIX):
Change it to:   for tc in doc.get("tool_calls") or []:
                    if not isinstance(tc, dict):
                        continue
                    if not str(tc.get("tool") or "").endswith(TOOL_SUFFIX):
```

Please add a test with a non-dict run log asserting it is counted unreadable and the readable ones still scan.

## 3. The verdict can blame the machine on an operator that never reached the service (should-fix)

`interleaving_verdict` is the function that keeps the by-author table honest, so a false positive here is worse than in a plain report. Its "concurrent success" test is `any(e == 0 ...)`, meaning "some operator had no reachability failure". But an operator whose only calls were `unrecognized_ark` satisfies that while never reaching OpenRouter, since `unrecognized_ark` is thrown before the OpenRouter step. I reproduced a day with one operator on `unreachable` and another on `unrecognized_ark`: it prints "SEPARATES ... points at the machine", which is false, because the second operator demonstrated nothing about reachability.

The fix is to require the concurrent operator to have actually reached (a `success`), not merely to lack a reachability failure. That needs the per-day cell to track a reached count, not just failures and totals. It does not fire on today's corpus, so it is a should-fix rather than a blocker, but this is the one function whose whole job is to not over-claim.

Once these and John's three are in, ping me for a re-read. Merge will also need a second senior approval and an update-branch.

aghadiayeamayanvboernest:
All six in `32f4dff1`. Every one reproduced before fixing and mutation-proved after, and none moves a printed number — the headline is still 30 of 175 (17.1%).

@johnmarkpeterbrown:

1. **Timeout bucket.** Both the tuple and the matcher are pinned now. Your second case is the one that made this worth doing: I confirmed a broken matcher lands in `upstream_error`, not `unclassified`, so it takes 6 of the 30 out of the headline with nothing going red. The test says that explicitly so it isn't "simplified" later.
2. **Escaped envelope.** Confirmed exactly as you describe — `\"error\"` contains neither `"error"` nor `error:`. Unescaping before matching, with a test pinning both shapes.
3. **`unrecognized_ark` is 3, not 2.** Verified: 3 in the corpus, 2 carrying `is_error`. Fixed in the module docstring and the test docstring.

@clack391:

4. **Null summary.** Reproduced — `classify(json.dumps(None))` returned `success`. Fixed. One correction to the framing, since this report is about not overstating: I measured **0 such calls in the committed corpus today**, so this closes a hole in the guarantee rather than correcting a live number. Your reasoning about why it's reachable is right and is now in the code comment.
5. **Non-dict run log.** Reproduced both crashes (`[]` envelope, and a truthy non-dict `tool_calls` entry). Both guarded, both tested. You were right that the old robustness test passed while the case it existed for crashed — that's the second time on this branch a test looked like coverage and wasn't.
6. **Verdict false positive.** Fixed by tracking a demonstrated `success` per cell rather than the absence of a failure, so an all-`unrecognized_ark` operator no longer counts as a concurrent success. Test builds exactly your repro.

**Mutation evidence** — each guard broken individually and watched go red:

| Mutation | Red |
|---|---|
| Drop `TIMEOUT` from the reachability tuple | ✅ |
| Break the timeout matcher (hides in `upstream_error`) | ✅ |
| Stop unescaping | ✅ |
| Let a null summary read as `success` | ✅ |
| Remove the non-dict envelope guard | ✅ |
| Verdict counts "no failure" as "reached" | ✅ |

**One thing not asked for:** the report was missing from `eval/CLAUDE.md`'s list of windowed aggregating reports. Added, with why its horizon is the 14-day default — that list is where a reader checks which reports window, and an absent entry reads as "this one doesn't".

Branch merged with main and no longer BEHIND. `make test-all` — all checks passed. Ready for a re-read whenever you are.

clack391 (changes requested):
Re-reviewed on `c66b68c2`. All six items from the two reviews are fixed, and I checked each by running it: every fix has a test that genuinely fails when the fix is reverted (including the timeout one both ways, the tuple and the matcher), the headline is still 30 of 175 (17.1%), and the null, escaped-envelope, and unrecognized-ark numbers all re-derive. This is careful, honest work, and reproducing each finding before fixing it is the right habit.

Two of the fixes stop one step short, though, in the same way: each closes the exact shape the review named and leaves the neighbouring shape of the same class. I reproduced both.

## 1. A non-list `tool_calls` still takes the whole report down (blocking)

The doc-object and per-entry guards are right, but the loop that reads `tool_calls` runs outside the try, so a run log whose `tool_calls` is a truthy non-list crashes the entire scan. This is the same "one bad file must not take the report down" guarantee the other guard is for, one level deeper.

```
scan([good_run, {"tool_calls": 42}])   -> TypeError: 'int' object is not iterable
```

`True` and `3.14` crash the same way; a string or a dict do not (they iterate, wrongly but harmlessly). Fix, and add a test with a non-list `tool_calls` value:

```
You wrote:      for tc in doc.get("tool_calls") or []:
Change it to:   tool_calls = doc.get("tool_calls")
                if not isinstance(tool_calls, list):
                    continue
                for tc in tool_calls:
```

## 2. The verdict still points at the machine when one operator both failed and succeeded (should-fix)

The `reached` counter is the right idea, but the mixed-day test asks two separate questions, "did anyone fail" and "did anyone reach", without requiring them to be different operators. So a day where one operator both failed and reached, while every other operator only threw `unrecognized_ark`, still prints "another succeeded, which points at the machine" even though no second operator reached anything.

```
alice: one unreachable + one success
bob:   one unrecognized_ark
interleaving_verdict(...)  -> "SEPARATES ... points at the machine"   (wrong: bob never reached)
```

Today's corpus does not hit this: 2026-08-13 is a real split (Gennecis and mercyokum each both failed and reached, so a genuine distinct pair exists), so nothing printed is wrong right now. But this is the one function whose whole job is not to over-claim, so the predicate should require a distinct reacher: some operator with a failure, and a different operator with a success. A test for the same-operator case would pin it.

## 3. The unescape can now read a real transcription as an error (should-fix)

Unescaping before matching is the right fix for the escaped envelope, but it also turns a genuine transcription whose text quotes the word error into `upstream_error`:

```
classify('[{"type":"text","text":"the clerk wrote \\"error\\" in the margin"}]')  -> upstream_error
```

It does not touch the reachability headline (`upstream_error` is not a reachability bucket) and nothing in the corpus hits it today, so it is not urgent, but it is the reverse of the direction your comment describes, and it quietly drops a real success from the reached count. Matching the error envelope shape (an unescaped `"error":` key adjacency) rather than a bare `"error"` substring would avoid it, and the comment should note the reverse case.

## 4. Drop `"none"` from the empty-summary guard (minor)

`json.dumps(None)` is `"null"`, and the scan never calls `str()`, so no infrastructure path ever produces the summary `"none"`. The entry can only ever match a genuine transcription of the word "None" and file it as unclassified instead of success. Keep `""` and `"null"`, drop `"none"`.

Items 1 and 2 are the two to close before this merges; 3 and 4 are cheap and worth folding in while you are here. Once they are in, ping me for a re-read. Merge also needs John's re-approval, since a `.py` change needs both a senior developer and a senior genealogist.

johnmarkpeterbrown (approved):
> Approving. I re-derived the corpus reading and mutation-checked every guard rather than reading the table — ten of eleven mutations go red, including both timeout cases, the unescape, the null summary, the two malformed-run guards, the loose matcher, the `"none"` arm, and the distinct-reacher predicate. Headline still 30 of 175 (17.1%), same taxonomy, same three multi-operator days.
>
> Two things worth saying about how you got here. Catching your own vacuous test because the mutation came back green is the habit that makes the rest of this trustworthy. And deleting the tautological atomicity test instead of softening it into something that always passes was the right call — a test that can't tell the difference is worse than none, and this repo has been bitten by the opposite choice.
>
> Three fold-ins below. None blocks merge; take them while you're in the file.
>
> ### 1. One guard still has no test
>
> Removing the per-entry `isinstance(tc, dict)` check leaves all twenty tests green — the only mutation that didn't fire. It isn't redundant with the run-level guard. It decides how much one bad entry costs. A run with a malformed entry between two good calls reads `measurable=2, unreadable=0` with it, and `measurable=0, unreadable=1` without. Both are defensible; nothing records which you meant.
>
> **You wrote:**
> ```python
>     assert r.unreadable >= 3, "each malformed run is counted, not fatal"
> ```
> **Change it to:**
> ```python
>     assert r.unreadable >= 3, "each malformed run is counted, not fatal"
>
>
> def test_one_malformed_entry_costs_its_entry_not_its_whole_run(tmp_path: Path):
>     """The per-entry guard is not redundant with the per-run one — it decides how
>     much a malformed entry costs. Without it the entry raises, the run-level guard
>     catches, and the run's OTHER calls leave the tally: 2 measurable becomes 0 and
>     a reachability failure disappears. Pin which of the two behaviours is meant."""
>     p = tmp_path / "fx" / "run-2026-08-20_00-00-00.json"
>     p.parent.mkdir(parents=True)
>     p.write_text(json.dumps({"tool_calls": [
>         {"tool": "mcp__genealogy__image_transcribe", "response_summary": "ok"},
>         42,
>         {"tool": "mcp__genealogy__image_transcribe",
>          "response_summary": "Could not reach OpenRouter. (fetch failed)"},
>     ]}), encoding="utf-8")
>
>     r = scan([p], author_of=_authors({}))
>     assert r.measurable == 2, "the entry is skipped; its neighbours still count"
>     assert r.unreadable == 0, "one bad entry does not condemn the run"
>     assert r.reachability_failures == 1
> ```
>
> ### 2. The `classify` docstring now tells the next reader to undo the fix
>
> It still says bare substrings, never quoted keys. The generic matcher below it is now a quoted key, and it's safe only because of the unescape — which the docstring doesn't mention.
>
> **You wrote:**
> ```
>     Bare substrings, not quoted-key matches: `response_summary` has two shapes
>     depending on size — under ~4000 chars the MCP envelope is passed through with
>     the document as an escaped string, over it the document is unwrapped and its
>     keys are real JSON keys (`orchestrator.py::_summarize_tool_response`). A
>     quoted-key matcher silently misses the escaped form. `wiki_failure_report.py`
>     learned this the expensive way: 42 real responses fell into `unclassified`.
> ```
> **Change it to:**
> ```
>     Unescape first, then match: `response_summary` has two shapes depending on
>     size — under ~4000 chars the MCP envelope is passed through with the document
>     as an escaped string, over it the document is unwrapped and its keys are real
>     JSON keys (`orchestrator.py::_summarize_tool_response`). A quoted-key matcher
>     run against the raw text silently misses the escaped form;
>     `wiki_failure_report.py` learned this the expensive way, with 42 real
>     responses falling into `unclassified`. Unescaping below is what makes the
>     `"error":` key match safe on both shapes.
> ```
>
> ### 3. A sentence lost its join when the `"none"` arm came out
>
> **You wrote:**
> ```python
>     # it, and the only thing it could match is a genuine transcription of the
>     # word — filing a real success as unclassified., and the orchestrator creates
>     # every tool-call entry with `response_summary: None`, filling it when the
> ```
> **Change it to:**
> ```python
>     # it, and the only thing it could match is a genuine transcription of the
>     # word — filing a real success as unclassified.
>     #
>     # The orchestrator creates
>     # every tool-call entry with `response_summary: None`, filling it when the
> ```
>
> ### @clack391 — one thing to agree on before you clear your round
>
> The shipped `_separates` is stricter than your review described. You asked for someone who failed and someone else who demonstrably got through. The code computes `reached - failed`, so an operator who both failed and reached is not eligible to be that second person.
>
> That changes what the real corpus says. On 2026-08-13 both Gennecis and mercyokum failed and reached. Your review reads that as a genuine distinct pair; under the shipped predicate it is not one, and the day prints CANNOT SEPARATE.
>
> I think stricter is right — it is the reading that keeps this PR's own conclusion honest, and under-claiming is the safe direction for this function. But you two currently mean different things by the same fix, so worth settling explicitly rather than in passing.

---

Two notes for you, not the author:

- My approval won't unblock merge on its own — clack391's `CHANGES_REQUESTED` at `c66b68c2` still stands and they asked to be pinged.
- I kept the mutation transcripts out of the comment. If clack391 wants the sweep, I can hand it over separately.

clack391 (approved):
Approving. This is the strongest measurement work I have reviewed on this board, and it got better under review rather than merely compliant.

## What earns the approval

**You answered the hard half of the 2026-08-23 ruling.** Dallan's board audit said the old snippet was no longer re-runnable, because a smaller number would read as the problem receding when it was really the instrument shrinking. The obvious response was to print the new number with a footnote. You did the honest thing instead: the stripped tally is a first-class line, no bare percentage is printed anywhere, and when the truth is unrecoverable you publish an interval rather than a point estimate. I re-derived all of it: 394 calls, 219 stripped, 175 measurable, 30 lost at 17.1%, bounds 30/394 and 249/394 giving [7.6%, 63.2%]. Every figure exact.

**`NO MEASURABLE CALLS` refusing to read as 0% is the detail I would have missed.** A report that inspected nothing must not look like a clean bill of health, and you enforced that with a non-zero exit rather than a sentence.

**Two of my three fixes came back better than I asked for.**

On the non-list `tool_calls` crash I proposed an `isinstance` guard. You did not take it. You made the whole per-run body the protected region, staged `run_calls`, and merged only on a clean parse, with a comment naming exactly why: enumerating shapes meant "each fix closed one and left its neighbour." That is a strictly better answer than the one I gave, and it generalises the criticism instead of satisfying it. Verified: `42`, `True` and `3.14` now cost one run instead of the report.

## Where I was wrong and you were right

My previous round asked you to loosen `_separates` to "some operator failed and a different operator succeeded." **That predicate is wrong and you were right to push back.** I re-derived 2026-08-13 from the run logs myself: Gennecis 10 unreachable plus 1 timeout against 31 successes, mercyokum 1 and 1 against 2 successes. Exactly your numbers. Under my version that day prints "points at the machine" on data arguing the opposite, because both operators failed and both got through. Your `reached - failed` is the correct reading, the corpus is the right argument for it, and recording that reasoning in the docstring is what let me check it rather than take it on trust. I have withdrawn the item.

PR #1858, opened 2026-08-24
tree_forget: warn on left-behind couple-fact leaks, don't extend the …

### LINKED ISSUE #1549: tree_forget sweep set is a subset of the couple/parent fact family — verify before extending, don't add blind
> **Reviewed 2026-08-20 before junior handoff. Both gates have cleared** — PR #1419
> merged 2026-08-14, issue #1572 closed 08-17, issue #1574 closed 08-19. The three
> correction comments below are superseded by this body. **One decision is open (see
> "Sweep match"); do not start until it is answered.** The original follows under
> "Original issue".

**Touches:** packages/engine/mcp-server/src/tools/tree-forget.ts, packages/engine/mcp-server/src/tools/materialize-facts.ts, packages/engine/mcp-server/tests/tools/tree-forget.test.ts, packages/engine/mcp-server/tests/packaging/ (new drift guard), docs/specs/tree-forget-tool-spec.md, docs/specs/simplified-gedcomx-spec.md

## The corpus pass is done — verify it, do not redo it

```sh
git ls-files '*gedcomx.json' | python3 -c '
import sys,json,collections,pathlib
c=collections.Counter()
for f in sys.stdin.read().split():
    for p in json.loads(pathlib.Path(f).read_text(encoding="utf-8")).get("persons") or []:
        for fa in p.get("facts") or []: c[fa.get("type")]+=1
for t,n in sorted(c.items()):
    if t and t.replace("+"," ").split(" ")[0] in {"Marriage","Divorce","Annulment","Separation","Engagement","MarriageBanns","MarriageBann","Parents"}: print(n,t)
'
```

Measured 2026-08-20 over 419 committed tree files / 3,087 persons:

- `Marriage` **93** — the existing evidence, still positive.
- Canonical `Engagement`, `MarriageBanns`, `Separation`, `Divorce`, `Annulment`,
  `Parents` — **0 each.** No canonical type earns a place in the swept set on corpus
  evidence, so **do not add one**, and do not go sourcing unredacted feedback zips
  for them. The old acceptance criterion 1 is withdrawn.
- FamilySearch **custom-type** spellings of the same events, all with FS-native UUID
  fact ids on FS-native person ids — the same evidence standard that settled
  `Parents`: `Marriage Registration` 6, `Marriage+bond` 4, `Marriage Notice` 3,
  `Separation+from+Anna+Owen` 1. That last is on person `L8VQ-J8F`, fact
  `e4d13b78-7cbf-435d-8b8f-ce0c6da49664`, in
  `eval/tests/e2e/john-richardson-parents/unstripped-tree.gedcomx.json`.
- `MarriageBann` (singular) 2 — agent-authored ids (`I1`/`F3`), **not** FS-native.
  Not evidence of an FS echo.

**Two consequences.** FS emits a couple event person-level as a free-text custom type
(`+` and space are the same encoding — `Tax+Assessment` 54 and `Tax Assessment` 45
both appear), and 44 distinct custom person-level fact types exist in the corpus.
`tree_forget` matches by **exact string** (`factIdsOfTypes`), so **none of these is
reachable by adding a canonical name to the list.**

`docs/specs/simplified-gedcomx-spec.md`'s fact-type table claims `Annulment` is
"echoed as a person-level fact". That has **0** corpus occurrences — mark it
unverified in the same PR rather than leaving it as evidence someone quotes forward.

## Ruling 2026-08-20 — exact match, plus a warning

**The swept sets do not change.** Every canonical candidate — `Engagement`,
`MarriageBanns`, `Separation`, `Divorce`, `Annulment` — measures **0** person-level
occurrences in the corpus, so none earns a place. `SWEPT_SPOUSE_FACT_TYPES` stays
`["Marriage", "Divorce", "Annulment"]`; `SWEPT_PARENT_FACT_TYPES` stays `["Parents"]`.

**What changes is visibility, not deletion.** After a `spouses-of` or `parents-of`
sweep, emit one `validation.warnings` line per person-level fact left behind whose
type's leading token (with `+` normalized to space) matches a swept type. Reuse the
existing warnings channel; **ids and type strings only, never fact values**, per the
spec's cross-owner warning rule. Put the warning's exact message text in
`docs/specs/tree-forget-tool-spec.md`'s "A conclusion is carried twice" section so it
is checkable.

**Rejected: leading-token matching.** It would close the leak by *deleting* the
facts, and the corpus says most of them should survive.

### The count in the first review was wrong — 6, not 14

The 14 counted the same fact repeatedly across the starting / unstripped / final
snapshots inside one fixture. Deduped by `(person, factId, type)`, re-measured
2026-08-20:

| Type | Value | Has a `Couple` relationship |
|---|---|---|
| `Marriage Registration` ×3 | — | 1 yes, 2 no |
| `Marriage Notice` | — | no |
| `Marriage+bond` | "Marriage bond between Samuel McCar…" | no |
| `Separation+from+Anna+Owen` | "Thomas and Anna separated. They we…" | yes |

**Five of the six are distinct documentary events, not echoes of the couple
conclusion.** A marriage bond is a pre-marriage record in its own right (this one
dated 7 July 1824); a registration is the civil record; a notice is banns. The sweep
exists to remove a *redundant* carrier of a conclusion — these are not redundant, and
deleting them would destroy evidence on an irreversible tool operating on user trees.

**Exactly one is a genuine leak**: `Separation+from+Anna+Owen`, where the type string
itself names the spouse. That is not generalisable — the "type" there is a sentence
somebody typed — which is why the answer is to surface it, not to pattern-match it away.

Reproduce before changing anything: dedupe by `(person, factId, type)`, or you will
re-derive the 14.

## Sites a change to the swept set touches — all five

- [ ] `tree-forget.ts` — the constants and their one-line evidence comments
- [ ] `tests/tools/tree-forget.test.ts` — 75 tests pass on `main` today; the
      Divorce/Annulment sweep is asserted there. **If a test you did not intend to
      change goes red, that is the finding — do not edit the assertion to match your
      diff.**
- [ ] `docs/specs/tree-forget-tool-spec.md` — the selector table row and the
      "A conclusion is carried twice" section
- [ ] `docs/specs/simplified-gedcomx-spec.md` — the fact-type table rows
- [ ] the new drift guard (below)

## The drift guard, concretely

None of `SWEPT_SPOUSE_FACT_TYPES` / `SWEPT_PARENT_FACT_TYPES` / `EVENT_TREE_TYPES` is
exported today. Export all three and **import** them in the test — that is how
`tests/packaging/enum-drift.test.ts` reads `VALIDATOR_ENUMS`; do not read the source
files as text. Add a third exported constant, `COUPLE_TYPES_CONFIRMED_NOT_ECHOED`,
each entry carrying its evidence comment. The guard fails when a couple-event member
of `EVENT_TREE_TYPES` is in neither set. **Before committing it, delete one type from
the swept set and watch the guard go red.**

## Parent side — answered, do not search

The corpus has **0** person-level `Parents` facts and **0** parent-like custom types
(the one near-miss, `Décès+du+père+…+Tutelle+des+enfants+…`, is a French life-event
note, not a parentage echo). `SWEPT_PARENT_FACT_TYPES` stays `["Parents"]`. Record
that result; do not open a search for a sibling type.

## What proves it worked

`make engine-test`, then `make test-all`. **There is no live check here** —
`dev/try-tree-forget.ts` runs against a local project, not FamilySearch, so nothing
in CI or on your laptop observes what FS actually emits. Say that in the PR rather
than implying CI covered the question of which types echo.

No skill body, rubric or eval test is touched, so no `make eval-skill` run and no
annotation. No open PR touches either tool file.

## One figure in the original has moved

"1,195 of 1,217 committed persons (98.2%)" was measured over `eval/tests/e2e` only.
That scope is now 1,258/1,280 (98.3%); whole corpus 2,768/3,087 (89.7%). Direction
holds, the number does not.

## Original issue

> **Do not start until issue #1572 has landed.** It narrows what `tree_forget`
> validates before applying any selector, which changes the code path this sweep runs
> in. Read `tree-forget.ts` as it stands *after* #1572, not as it stands today.
>
> This is a one-way dependency: #1572 needs nothing from this issue. You do not need to
> hold both assignments — just check that #1572 is closed before you pick this up.

**Touches:** packages/engine/mcp-server/src/tools/tree-forget.ts, packages/engine/mcp-server/src/tools/materialize-facts.ts

## Root

`tree_forget`'s redundant-fact sweep (`SWEPT_SPOUSE_FACT_TYPES` / `SWEPT_PARENT_FACT_TYPES` in `tree-forget.ts`) is a **deliberate, evidence-backed subset** of the couple/parent fact family, not the whole class. `materialize-facts.ts`'s `EVENT_TREE_TYPES` already recognizes further couple-event types the sweep does **not** cover — `Engagement`, `MarriageBanns` (GedcomX also has `Separation`, `CommonLawMarriage`, `DomesticPartnership`). If FamilySearch echoes any of these as a person-level fact the way it echoes `Marriage`, `spouses-of` leaks it — the exact #1417 leak under a new type.

This is the class the #1314 → #1417 pair are instances of. The two instance-fixes landed via #1419; this issue owns closing the class.

## Why this is not a one-line "add the types"

The #1314 review nearly merged a `Parents` matcher that looked undefined — zero occurrences in every committed tree, absent from the spec enum — until the unredacted `feedback-2026-08-03` session log confirmed `Parents` is real (from a `person_read` **result** on persons GRNX-DFF and GRN6-4MQ, FS-native UUID fact ids, in `facts[]` beside Birth/Death).

The lesson is **not** that the committed corpus is uninformative. It is: 1,195 of 1,217 committed persons (98.2%) carry at least one person-level fact and no committed tree is fully blank, so the corpus **positively confirms** which types echo person-level — that is exactly how `Marriage` is evidenced. What the corpus cannot do is settle a type it contains **zero** of: absence there is not proof the type never echoes, only that no committed tree happens to carry it — the `Parents` situation the 08-03 zip resolved. So adding `Engagement`/`MarriageBanns`/`Separation` to the sweep on assumption would repeat that near-miss in reverse: sweeping a type that may never appear, or missing one that does.

## Acceptance

1. Determine which of `Engagement`, `MarriageBanns`, `Separation` (and any other `EVENT_TREE_TYPES` couple event) echo as **person-level** facts. **Check the committed corpus first** — it positively confirms any type that already appears person-level (as it does for `Marriage`); only for types with **zero** committed occurrences do you still need unredacted tree/session-log data to settle whether they echo. Record which actually echo person-level, with the evidence for each.
2. Extend `SWEPT_SPOUSE_FACT_TYPES` to exactly the confirmed set — each with a one-line evidence citation like the existing two.
3. Add a guard test cross-referencing `materialize-facts.ts`'s couple-event subset of `EVENT_TREE_TYPES`: it fails when a couple-event type is neither in the swept set nor explicitly annotated as "confirmed not echoed person-level" — so a future couple type can't be added to one list and silently missed by the other.
4. Same treatment for `SWEPT_PARENT_FACT_TYPES` if a parent-echo sibling of `Parents` is ever confirmed.

## Context

Spun out of the #1419 review (parents-of/spouses-of redundant-fact sweep). Half 1 (named constants + evidence comments) landed in #1419; this is Half 2. Do not add unverified types — that is the whole point of the issue.

chrisedeson:
Closes #1549. Sweeps all couple-event types and warns on left-behind prefix-only matches.

## What this does

Per the 2026-08-20 ruling and 2026-08-24 follow-up (@DallanQ): `spouses-of` now sweeps all couple-event types (`Marriage`, `Divorce`, `Annulment`, `Engagement`, `MarriageBanns`, `Separation`). `COUPLE_EVENT_TYPES` is exported from `materialize-facts.ts` as the single source; `SWEPT_SPOUSE_FACT_TYPES` derives from it so they cannot drift apart.

After a `parents-of`/`spouses-of` sweep, one `validation.warnings` entry is emitted per left-behind person-level fact whose type's leading token matches a swept type without being an exact match — `Marriage Registration`, `Marriage+bond` and similar are real, distinct documentary events, not echoes of the couple conclusion, so they're surfaced rather than deleted or silently kept. The warning prints the canonical prefix only, never the full type string (which can contain names).

- Exports `COUPLE_EVENT_TYPES` from `materialize-facts.ts`; `SWEPT_SPOUSE_FACT_TYPES` in `tree-forget.ts` derives from it
- Drift guard (`tests/packaging/tree-forget-sweep-drift.test.ts`): every couple-event type must appear in both `EVENT_TREE_TYPES` and `SWEPT_SPOUSE_FACT_TYPES`, and every confirmed couple-event type from `mob.ts` (`MARRIAGELIKE_FACT_TYPES` + `DIVORCELIKE_FACT_TYPES`) that enters `EVENT_TREE_TYPES` must also be in `COUPLE_EVENT_TYPES` — the mandated break-test (delete a type from the swept set, or add `CommonLawMarriage` to `EVENT_TREE_TYPES` without `COUPLE_EVENT_TYPES`) now fails
- Left-behind checks are deferred to post-removal so a fact removed by a later selector in the same call is not falsely reported as "left in the tree"; duplicate warnings are deduped (matching `resolveFactSharingNotices`'s pattern)
- `docs/specs/tree-forget-tool-spec.md` §2.1.1: documents the warning behavior with its exact message text
- `docs/specs/simplified-gedcomx-spec.md`: adds `Engagement`, `MarriageBanns`, `Separation` rows; corrects `Annulment`'s and `Separation`'s GedcomX status
- Model-facing contracts updated: `treeForgetSchema` selector description and `forget-and-rederive/SKILL.md` table now list all six couple-event types
- 4 behavior tests in `tree-forget.test.ts`: left-behind warning for both `parents-of` and `spouses-of`, multi-selector deferred-check, and sweep of the three new types (`Engagement`, `MarriageBanns`, `Separation`). The left-behind and multi-selector tests guard different mutations (returning `[]` vs inlining on pre-removal tree); the new-types test goes red when the three are deleted from `COUPLE_EVENT_TYPES`

## Verification

- `make engine-test`: 112 files, 2743 tests, clean
- `make harness-test`: 2812 passed
- `make server-test`: 209 passed
- Drift guard watched to fail (swapped the derivation for a hand-written list, and added `CommonLawMarriage` to `EVENT_TREE_TYPES` without `COUPLE_EVENT_TYPES`) then pass, restored
- 2 warning tests watched to fail with `leftBehindCoupleFactWarnings` returning `[]`; multi-selector test watched to fail with warnings inlined on the pre-removal tree; new-types test watched to fail with the three types removed from `COUPLE_EVENT_TYPES`. Each guards a different mutation
- There is no live check here — the left-behind warning path requires a real `tree_forget` call on a tree carrying the prefix-only types, which no CI job exercises
- Merged current `main`; no file-level conflicts


johnmarkpeterbrown (changes requested):
Four things to change before this merges. The escalation on `Divorce`/`Annulment` is handled right — flagged in the body, flagged in the code, set untouched. Not second-guessing that.

**1. The warning can print the forgotten spouse's name.**

`docs/specs/tree-forget-tool-spec.md` §3.1 allows fact type names in a result because they are *a kind, not a value* — `Birth`, `Marriage`. That permission does not cover FamilySearch's free-text custom types, and the committed corpus has one that carries a person's name: `Separation+from+Anna+Owen`, in `eval/tests/e2e/john-richardson-parents/unstripped-tree.gedcomx.json`. An unstripped answer tree — exactly the material this tool hides.

A fact typed `Marriage+to+Ellen+Walsh` produces this on `spouses-of I1`:

> `Marriage+to+Ellen+Walsh fact 'FX' on I1 was left in the tree — …`

Ellen Walsh is the spouse that call just removed. Print the canonical prefix instead. It is still a kind, and the id is still there to look up.

In `tree-forget.ts`:

```
You wrote:      `${f.type} fact '${f.id}' on ${personId} was left in the tree — its type is not ` +
Change it to:   `A ${leadingToken(f.type)}-prefixed fact '${f.id}' on ${personId} was left in the tree — its type is not ` +
```

Both new tests assert the full type string, so they need the same edit. §2.1.1's "ids and type strings only" needs to become "ids and canonical type prefixes only" — as written it widens §3.1 without saying so.

**2. It warns that a fact was left behind after another selector removed it.**

The warning is built before removal, and it only knows about the facts its own selector swept. Run this:

```
forget: [ {selector: "spouses-of", personId: "I1"},
          {selector: "facts-of", personId: "I1", factType: "Marriage Registration"} ]
```

`FMR` is gone from the tree and the result still says it "was left in the tree."

This is the bug `resolveFactSharingNotices` already solved. Its comment says every pending check has to wait until after removal, because a fact can be taken by a different selector in the same call. The new code pushes finished strings in next to that call and skips the wait. Do the same thing it does.

In `resolveSelectors`:

```
You wrote:      if (kind === "parents-of") {
                  t.factSharingNotices.push(
                    ...leftBehindCoupleFactWarnings(tree, pid, kind, SWEPT_PARENT_FACT_TYPES, redundantFactIds),
                  );
                } else if (kind === "spouses-of") {
                  t.factSharingNotices.push(
                    ...leftBehindCoupleFactWarnings(tree, pid, kind, SWEPT_SPOUSE_FACT_TYPES, redundantFactIds),
                  );
                }
Change it to:   if (kind === "parents-of" || kind === "spouses-of") {
                  t.leftBehindChecks.push({ personId: pid, kind });
                }
```

Add `leftBehindChecks: { personId: string; kind: "parents-of" | "spouses-of" }[]` to `Targets`, initialised `[]`. Have `leftBehindCoupleFactWarnings` take that list and pick the swept set from `kind`. Drop its `removedIds` parameter — checking the post-removal tree covers it. Then in `applyForget`:

```
You wrote:        ...resolveFactSharingNotices(tree, t.pendingFactNotices),
                  ...t.factSharingNotices,
Change it to:     ...resolveFactSharingNotices(tree, t.pendingFactNotices),
                  ...t.factSharingNotices,
                  ...leftBehindCoupleFactWarnings(tree, t.leftBehindChecks),
```

I ran this shape. Your two new tests still pass and so does the rest of the engine suite.

**3. The drift guard cannot catch what its comment promises.**

`tree-forget-sweep-drift.test.ts` says a new couple-event type added to `EVENT_TREE_TYPES` "can't be silently missed by both." It can. `COUPLE_EVENT_TYPES` is a hand-written list inside the test, so a type that never gets added there is never examined. Add one to `EVENT_TREE_TYPES` alone and the guard stays green.

The logic is worth keeping — a rename does go red. Only the claim needs fixing, and in this repo a guard that reads as coverage it does not have is worse than no guard.

```
You wrote:       * type added to EVENT_TREE_TYPES can't be silently missed by both.
Change it to:    * type listed in COUPLE_EVENT_TYPES below can't be silently missed by both.
                 * This does NOT catch a new couple-event type added to EVENT_TREE_TYPES and
                 * not added here — that gap is unguarded.
```

**4. "isn't pattern-matchable" is not true, and question 2 rests on it.**

`leadingToken("Separation+from+Anna+Owen")` returns `"Separation"`. Same `+`→space normalisation that turns `"Marriage+bond"` into `"Marriage"`. The case is missed because the token set comes from the swept types and `Separation` is not one. That is a scope choice, not an impossibility — so question 2 currently asks Dallan to rule on something the spec tells him the code cannot do.

Worth pairing with the first finding when he answers: `Separation` is the one prefix in the corpus whose free text carries a name, so widening the set without the redaction fix would start printing "Anna Owen".

```
You wrote:      this check at all, and isn't pattern-matchable — a one-off data observation,
                not a mechanism this tool can implement.
Change it to:   this check at all, because the warning's token set is drawn from the swept
                types and `Separation` is not one. Widening that set is mechanically possible;
                whether it should widen is open.
```

One small thing: the body says `make test-all: running` and never reports the result. Worth filling in.

chrisedeson:
All four addressed in `b223f657`:

Warning now prints the canonical prefix, not the full type string.
Left-behind checks deferred to post-removal via `leftBehindChecks` on `Targets`, same pattern as `pendingFactNotices`.
Drift guard comment corrected.
Spec wording fixed — "mechanically possible, whether it should widen is open" instead of "isn't pattern-matchable."
Also swept all couple-event types per Dallan's 2026-08-24 ruling — `COUPLE_EVENT_TYPES` exported from `materialize-facts.ts`, `SWEPT_SPOUSE_FACT_TYPES` derives from it. `Separation` is now a swept type, so `Separation+from+Anna+Owen` triggers the warning too.

johnmarkpeterbrown (changes requested):
Thanks — all four are fixed, and I checked each by running it rather than reading the diff. The name leak is closed, the false "left in the tree" warning is gone in the two-selector case while the genuine one still fires, and both doc fixes read right. Nice side effect of the new sweep: the corpus's one real leak, `Separation+from+Anna+Owen`, now warns without printing the name.

One thing blocks this, and it isn't code.

## The sweep set changed, against the only ruling on record

@DallanQ — this one needs you.

Your ruling on #1549, dated 2026-08-20:

> exact match, plus a warning. **The swept sets do not change** […] **Rejected: leading-token matching**, which would close the leak by deleting the facts.

This PR sweeps `Engagement`, `MarriageBanns` and `Separation`. The body and the code comment both cite a "2026-08-24 ruling". The last comment on #1549 is the 2026-08-20 one, and there's nothing from you on this PR. The body also flipped from "#1549 should stay open" to "Closes #1549".

If that follow-up happened on a call, that's fine — but please post it on #1549 before this merges. `tree_forget` is irreversible, and the previous round of this PR deliberately left the set alone for exactly that reason. I can't approve a reversal of a written decision against one I can't see.

## The guard no longer catches a rename

`EVENT_TREE_TYPES` is now built by spreading `COUPLE_EVENT_TYPES`, and `SWEPT_SPOUSE_FACT_TYPES` is `[...COUPLE_EVENT_TYPES]`. So the first test can't fail the way its name says. Rename `MarriageBanns` to `MarriageBans` inside `COUPLE_EVENT_TYPES` and it stays green. Remove it entirely — still green.

The second test does still earn its place. Swap the derived constant for a hand-written list and it goes red naming every missing type. So keep the file, just fix the name:

```
You wrote:      it("COUPLE_EVENT_TYPES stays a subset of EVENT_TREE_TYPES (catches a typo or rename)", () => {
Change it to:   it("COUPLE_EVENT_TYPES stays a subset of EVENT_TREE_TYPES (catches un-spreading the derivation)", () => {
```

> Guard watched to fail (removed `MarriageBanns`) then pass, restored.

True of the old implementation, not this one — removing `MarriageBanns` now leaves the guard green. Re-run it against the current code or drop the line.

`make engine-test` also reports 2724 tests in the body; I get 2690 passed and 1 skipped on this head.

## Small stuff, not blocking

Nothing guards the multi-selector bug you just fixed — worth a test, since it was a real defect and the fix is easy to undo by accident. And in `leftBehindCoupleFactWarnings`, `sweptTokens` and `sweptExact` are built from the same expression, so one of them can go.

chrisedeson:
All addressed in `9248a7f7`. Ruling posted on #1549.

**Guard name:** Renamed to "catches un-spreading the derivation." Comment updated to match.

**Duplicate variable:** *sweptTokens* and *sweptExact* were identical. Collapsed to one (*sweptLower*).

**Multi-selector test:** Added. Two selectors in one call — *spouses-of* enqueues a left-behind check, *parents-of* removes the fact that would have triggered it. Confirms the deferral prevents the false positive.

**Stale body lines:** PR body updated — test count now 2733 (112 files), guard verification rewritten against the current implementation, merged main at 8fb49e4a.

johnmarkpeterbrown (commented):
Three of the four are done right — guard name, the collapsed variable, and the body's guard verification line now describes a mutation that genuinely goes red. I checked that one.

Two things left.

## The new multi-selector test doesn't catch the bug

I removed the deferral — put the left-behind warnings back on the pre-removal tree, which is the exact bug — and your new test passed. So did the rest of the engine suite. Nothing in the repo currently guards that fix.

The scenario is the problem. `parents-of` removes an exact `Parents` fact, and exact matches are already skipped one line earlier:

```ts
if (sweptLower.has(f.type.toLowerCase())) continue;
```

So `FP` never reaches the deferral logic either way. To exercise it, the second selector has to remove the **prefix-matched** fact — the one the first selector queued a warning for.

```
You wrote:      forget: [
                  { selector: "spouses-of", personId: "I1" },
                  { selector: "parents-of", personId: "I1" },
                ],
Change it to:   forget: [
                  { selector: "spouses-of", personId: "I1" },
                  { selector: "facts-of", personId: "I1", factType: "Marriage Registration" },
                ],
```

Drop the `FP` / `Parents` fact from the fixture and the `factsByType.Parents` assertion with it. Then assert `r.validation.warnings` is `[]` and that `FMR` is gone from the tree.

I ran that version both ways: red with the deferral removed, green with it restored.

While you're there — the body says all three new tests were "watched to fail with the feature disabled." That holds for two of them, so worth re-checking what the third was actually disabled against.

## The ruling on #1549

Thanks for posting it — that's a real improvement, and quoting him verbatim with a date is the right shape.

What's still missing is Dallan himself. His written ruling from 2026-08-20 says "the swept sets do not change," and the comment authorising the opposite is yours quoting him rather than his own. So #1549 now carries two contradictory statements and the newer one is attributed, not authored.

Ask him to drop a 👍 on your comment, or a one-line "confirmed." That's all it takes. `tree_forget` is irreversible and this reverses his own written call, so I'd rather the record settle it than a reviewer's memory of a thread.

Nothing else outstanding — the code is right, it's the coverage and the paperwork.

chrisedeson:
Thanks, both addressed.

Multi-selector test. You're right, `Parents` is an exact match that never reaches the deferral logic. Changed the second selector to `facts-of` with `factType: "Marriage Registration"` so the prefix-matched fact is what gets removed by the later selector. Watched both directions: red with the deferral inlined (warning fires on the pre-removal tree), green with it deferred (Marriage Registration gone by the time the check runs). The prior version of that test would have passed either way.

DallanQ:
**Confirmed.** I made that call on Slack on 2026-08-24 and Christopher quoted me accurately: sweep all couple events — `Marriage`, `Divorce`, `Annulment`, `Engagement`, `MarriageBanns`, `Separation` — with the couple-event set exported from `materialize-facts.ts` and the spouse sweep derived from it. That supersedes "the swept sets do not change" in my 2026-08-20 ruling on issue #1549.

@johnmarkpeterbrown — right call to block. An irreversible tool reversing a written decision on a reviewer's memory of a thread is exactly the thing that should stop at review.

**The other half of that ruling still stands, and this branch keeps it.** Leading-token matching is still rejected *as a deletion rule*. Removal is exact-match, case-insensitive — `factIdsOfType` lowercases and compares the whole type string. The leading token only ever produces a warning. So `Separation` joining the swept set deletes a fact typed exactly `Separation` and nothing else; `Separation+from+Anna+Owen` survives and now warns without printing the name. That is the outcome I wanted on 08-20 and couldn't get without widening the token set.

**Why I changed my mind on the set.** The 08-20 reasoning was "0 corpus occurrences, so no canonical type earns a place." That is the wrong test for an exact-match sweep. Deleting a person-level `Engagement` fact when you forget the couple is correct whether or not our 419 committed trees happen to contain one — the corpus count tells us how often the rule fires, not whether it is right. The evidence bar I set was built for a *pattern* rule, where a false positive destroys a real record. It does not apply to naming a canonical type.

Christopher, two things and this is done from my side:

1. Change every "the 2026-08-24 ruling" reference — the code comment in `tree-forget.ts`, both spec files, and the drift guard — to point at this comment. As written it points at a Slack thread nobody can open, and issue #1549's body still carries the superseded 08-20 text, so a reader who follows the issue lands on the opposite instruction.
2. Re-check the PR body's claim that all three new tests were watched to fail with the feature disabled. John Mark says that holds for two. Either verify the third or drop it from the body.

DallanQ:
Christopher — withdrawing both items I gave you above. I checked them and they were mostly me making work.

**Item 1, repointing the "2026-08-24 ruling" references: dropped.** I've rewritten issue #1549's body so the ruling lives there under that date, next to the 08-20 one it replaces. Every "per the 2026-08-24 ruling" in the two specs and the drift guard now resolves to something a reader can open. Leave them as they are.

**Item 2, the third test: dropped — the tests are fine.** I ran the mutations on the current head (`e884d572`) rather than asking you to:

- `leftBehindCoupleFactWarnings` returning `[]` early → both warning tests go red, multi-selector stays green.
- Deferral removed, warnings computed on the pre-removal tree in `resolveSelectors` → multi-selector goes red, the other two stay green.

Each of the three dies against the mutation it actually guards. Nothing to fix.

What's actually left is two lines in the PR body:

1. "All 3 new behavior tests watched to fail with the feature disabled" — true of two. The third guards the **deferral**, not the warning, and fails against a different mutation. Reword it to say that, so nobody later reads it as an unverified claim.
2. The test count. `npm test` on `e884d572` gives me **112 files, 2707 tests, 0 skipped**; the body says 2733. Re-run and use whatever you get on the current head.

@johnmarkpeterbrown — the ruling question is settled in my comment above and in #1549's body. Your two substantive findings both check out on my end; over to you.

johnmarkpeterbrown (approved):
Looks good!

clack391 (changes requested):
**Request changes.** Reviewed on the tree merged with current main. Suite I ran: `make engine-test`, 112 files, 2707 tests, exit 0.

Thanks @johnmarkpeterbrown for three genuinely substantive rounds. The round-1 catch, that the warning printed the full free-text type and would have echoed back the spouse the call had just forgotten, is the most important thing anyone found on this PR.

**One thing up front, so it is not left hanging.** My first pass had a headline blocker that the widened sweep destroys sourced person-level `Separation`/`Engagement` facts that our own `materialize_facts` writes. I withdrew it on audit against the ruling. The issue body says, in @DallanQ's own words, "Deleting a person-level `Engagement` fact when you forget the couple is correct whether or not our 419 committed trees happen to contain one." That is exactly the behaviour, explicitly sanctioned, and you implemented it correctly. My "silently" framing was also wrong: the call returns `removed: {Engagement: 1, Separation: 1}`, which is the tool's documented counts-and-kinds contract. Nothing for you to do there. I mention it only so you are not left wondering whether the deletion is still in question.

Same for `Separation` entering `EVENT_TREE_TYPES`: I first read that as an undeclared side effect, but "include that into the bigger list" is the instruction, and the bigger list is `EVENT_TREE_TYPES`. That was right too.

## What I verified as correct

- **Privacy holds under attack.** Two independent blind passes threw embedded-name types at it (`Marriage+to+Ellen+Walsh`, `Separation+from+Ellen+Walsh`, `Divorce Ellen Walsh`, and three more). Nothing leaked. `leadingToken(f.type)` is only reached after the token has matched a swept type, so the interpolated word is provably one of six canonical strings.
- **The deferral is real, and its test is not vacuous.** My first mutation was a no-op, because I snapshotted the tree after `pruneFacts` at `:850` and the copy was already pruned. With a faithful mutation, exactly one test goes red: `multi-selector call does not falsely warn when a later selector removes the fact (#1549)`. That test was vacuous one round ago, so it was worth proving. It is the strongest test in the PR.
- **The spec's message text matches the emitted string character for character**, and it is in the section the ruling named.
- **The corpus figures reproduce exactly** on my own script: `Marriage` 93, `Marriage Registration` 6, `Marriage+bond` 4, `Marriage Notice` 3, `Separation+from+Anna+Owen` 1, the five canonical types 0 each, and 14 raw deduping to 6 by `(person, factId, type)`.
- **The governance question is settled.** @DallanQ left a 👍 on the #1549 comment carrying the 08-24 ruling, which is exactly the confirmation asked for in round 3.

## 1. The drift guard does not do what the issue specifies, and the mandated break-test passes green

This is the one that matters most, because #1549 is labelled `nothing-checks` and the guard is its deliverable.

The issue body asks for two things by name:

> The guard fails when a couple-event member of `EVENT_TREE_TYPES` is in neither set.

> Before committing it, delete one type from the swept set and watch the guard go red.

I ran the second one. It does not go red:

| mutation | guard |
|---|---|
| delete `Separation` from the swept set (the mandated break-test) | **green** |
| add `CommonLawMarriage` + `DomesticPartnership` to `EVENT_TREE_TYPES` only | **green** |
| corrupt the case to `marriagebanns` / `SEPARATION` | **green** |
| delete all three new types, undoing the whole 08-24 ruling | **green** |
| control: replace the derivation with a literal | red |
| control: un-spread `EVENT_TREE_TYPES` | red |

The cause is structural, not an oversight in the assertions. `SWEPT_SPOUSE_FACT_TYPES = [...COUPLE_EVENT_TYPES]` and `EVENT_TREE_TYPES` spreads the same set, so both assertions are tautologies of those two spreads. The guard proves the derivation is spelled correctly. It cannot speak to whether the inventory is right, which is the thing the issue wanted guarded, and the two types the issue's Root section names as the coming risk, `CommonLawMarriage` and `DomesticPartnership`, are exactly the case that stays green.

The file's docstring already concedes this in one sentence, which is honest, and I would rather have the honesty than a false claim. But the issue asked for the assertion, not the caveat, and CLAUDE.md's line applies: a check that cannot fail reads as coverage and is worse than no check at all.

Cheapest closure I can see: assert `EVENT_TREE_TYPES`'s exact membership, so any addition trips the guard and forces a conscious couple or not-couple decision. Worth knowing before you pick a shape: `src/utils/mob.ts:57-74` already holds an FS-confirmed couple-event vocabulary (`MARRIAGELIKE_FACT_TYPES`, `DIVORCELIKE_FACT_TYPES`), under a header saying membership was confirmed by Richard Chesworth on 2026-06-02 and the identifiers confirmed by @DallanQ. Iterating `EVENT_TREE_TYPES` against that would make the mandated assertion writable.

## 2. The three newly swept types have no behavioural test

```
grep -c in tests/tools/tree-forget.test.ts
  Engagement 0    MarriageBanns 0    Separation 0
  Divorce 5       Annulment 5
```

Delete all three from `COUPLE_EVENT_TYPES`, reverting the sweep to main's three types and undoing the entire 08-24 ruling, and the suite is **80 passed (80)**. The whole functional delta of this PR is asserted only by the guard in finding 1, which finding 1 shows is inert on exactly that mutation.

The behaviour itself is correct. I confirmed by direct probe that a `spouses-of` sweep returns `factsByType: {"Engagement":1,"MarriageBanns":1,"Separation":1}`. It is a five-line test, not a defect.

## 3. Two model-facing contracts still advertise the old three-type sweep

Both are outside the diff, and both describe an irreversible tool to the model that calls it.

`packages/engine/mcp-server/src/tools/tree-forget.ts:1077-1080`, inside the same `treeForgetSchema` object this PR edits at `:1023-1027`:

> `"...and spouses-of ALSO removes the person's own \`Marriage\`/\`Divorce\`/\`Annulment\` facts, so the forgotten conclusion does not survive as a fact on the subject."`

`packages/engine/plugin/skills/forget-and-rederive/SKILL.md:112`, the table the agent reads before it destroys anything:

> `| spouses-of | personId | the person's spouses, the couple relationships, and the person's own \`Marriage\`/\`Divorce\`/\`Annulment\` facts |`

The tool now removes six types and tells the model it removes three. The skill edit is free, incidentally: `forget-and-rederive` is on `check_runlogs.py`'s exemption list, has no unit suite and no runlog dir, so touching it buys no paid eval run.

## Smaller, same pass

- **`docs/specs/simplified-gedcomx-spec.md:356`** marks `Separation` as `**no**`, "Extension, not standard GedcomX". Three in-repo referents disagree: the issue body enumerates it as a GedcomX type ("GedcomX also has `Separation`, `CommonLawMarriage`, `DomesticPartnership`"), `mob.ts:69-74` lists it among the FS-confirmed types, and `Engagement`/`MarriageBanns` from the same GedcomX couple family were added `yes` in the same hunk. That column is load-bearing: every other `no` row warns the type gets a custom URI and "may not survive FamilySearch upload processing", which `Separation` would now inherit falsely.
- **The test count does not reproduce.** The body says 2733; `make engine-test` on this head gives **2707**, measured twice in two clean worktrees. Third time this figure has moved on this PR, so worth pinning before it gets quoted forward.
- **Two things the issue requires the PR to say are missing**: the `make test-all` result, and "There is no live check here. Say that in the PR rather than implying CI covered the question of which types echo." The Verification section currently leads with a green suite, which is the impression that clause exists to prevent.
- **Duplicate warnings.** Two identical `spouses-of` selectors emit the identical line twice. I ran it. The sibling channel `resolveFactSharingNotices` dedupes on the final string and carries a comment explaining exactly this hazard; a one-line `Set` matches it.
- **`SWEPT_PARENT_FACT_TYPES` is newly exported and imported nowhere.** The issue asked for all three exported and imported in the test; the test imports two.
- **Stale copies of the framing this diff reworded**: `tree-forget.ts:648` still says "Parents/Marriage/Divorce/Annulment" two lines above the label correctly changed to `couple-event`, and `tests/tools/tree-forget.test.ts:447` says the same.
- **Note, not a request:** the `Separation` addition to `EVENT_TREE_TYPES` means a materialized `Separation` fact now drops its transcribed `value` at `materialize-facts.ts:214`. That follows from the instruction, but it is a behaviour change in a second tool with no test and no spec line.
- **Nits:** the new `import` sits at `:94`, mid-file, where every other import in the file is at `:28-48`; and the warning prints the fact's own casing rather than the canonical constant, so `marriage registration` yields "A marriage-prefixed fact", where the ruling says "canonical type prefixes".

## For @DallanQ, not for this PR

Two things about the rule's reach that the author implemented faithfully and cannot decide.

**The corpus's one real leak is invisible to both new paths.** `eval/runlogs/e2e/sebastiano-mingazzini-son/run-2026-08-04_00-14-29.final-tree.gedcomx.json` has `R15` (Couple) carrying `MarriageBanns`, and `I1` and `I2` each carrying a person-level `MarriageBann`, singular, with `value` "marriage bann published 24 March 1887, Solarolo, Ravenna". `leadingToken` splits on whitespace and `+` only, so a spelling with no separator is neither swept nor warned, and the call reports a clean sweep while the banns date and place stay on both people. You considered `MarriageBann` in the body and set it aside as agent-authored rather than an FS echo, which is why this is a question rather than a finding: the rule is doing what you specified.

**The warning reaches 2 of the 6 measured leaks.** `spouses-of` errors out with "matched nothing" before any warning can fire when the subject has a prefix-only fact but no spouse relationship and no exact-match swept fact. I reproduced that. Recomputing your dedupe over the corpus:

```
'Marriage Notice'            P7SM-B86  couple=False
'Marriage Registration'      LTLP-GJR  couple=False
'Marriage Registration'      L8WW-KBJ  couple=False
'Marriage+bond'              KPHY-36B  couple=False
'Marriage Registration'      L6TX-ZXD  couple=True
'Separation+from+Anna+Owen'  L8VQ-J8F  couple=True
```

Four of the six can never produce a warning. The good news is that the one you singled out as the genuine leak, `Separation+from+Anna+Owen`, is one of the two that works, and it now warns without printing the name.

clack391 (changes requested):
**Request changes, and a much narrower one than last round.** Two of my three blockers are properly closed and I verified each by running it, not by reading the diff.

Reviewed at `7b52c901` on the merged tree, 0 behind main. Engine suite run this pass: 112 files, 2710 tests, exit 0.

## Closed, verified

- **Behavioural coverage.** `Engagement`, `MarriageBanns` and `Separation` went from 0 references to 4 each, and deleting all three from the swept set now reddens the suite. At the previous head that mutation was green. This is the fix that matters most, because it is what actually protects the invariant today.
- **Model-facing contracts.** Both the tool schema selector description and `forget-and-rederive/SKILL.md:112` list all six types. I swept the whole repo for the old three-type enumeration emphasis- and wrap-tolerantly (a markdown phrase is split by inline bold and by line wrapping, so a plain line grep misses it) and found **zero** survivors. That sweep is genuinely complete.
- **Dedup and canonical casing.** Two identical `spouses-of` selectors now emit one warning per fact instead of two, and it does not swallow distinct facts. A fact typed `marriage registration` prints canonical `Marriage`. Both verified by execution.
- **The `Separation` row** is `yes` with "GedcomX type", which is correct and now consistent with its siblings.

I also re-ran the redaction attack. Nothing leaks a name through the warning by any route I or two independent passes could build.

## 1. The body states a break-test that passes

> Drift guard watched to fail (swapped the derivation for a hand-written list, and added `CommonLawMarriage` to `EVENT_TREE_TYPES` without `COUPLE_EVENT_TYPES`) then pass, restored

The first half is true, and I confirmed it: swapping the derivation for a literal reddens the guard.

The second half does not reproduce. Adding `CommonLawMarriage` to `EVENT_TREE_TYPES` alone leaves the guard **4 passed**, measured three independent ways. The cause is precise and not your fault as a design choice, only as a claim: the third assertion filters `EVENT_TREE_TYPES` through `mob.ts`'s `MARRIAGELIKE ∪ DIVORCELIKE`, and neither `CommonLawMarriage` nor `DomesticPartnership` appears anywhere in the repo. They are the two types issue #1549's Root section names as the coming risk, so the guard is blind to exactly the gap the sentence claims it closes.

Please drop or correct that clause. `CLAUDE.md`'s "a new lint must be proven to fail" is the rule this trips, and a body that says "watched to fail" for a mutation that passes reads as coverage the guard does not have.

Same claim appears in the commit message for `0159d29c` and in your PR comment above.

## 2. The guard's remedy text points at behaviour the lead ruled out

`tests/packaging/tree-forget-sweep-drift.test.ts`, the third assertion:

```
`Confirmed couple-event type(s) present in EVENT_TREE_TYPES but missing from ` +
  `COUPLE_EVENT_TYPES (and therefore not swept by spouses-of): ` +
  `${inTreeButNotCouple.join(", ")}. Add to COUPLE_EVENT_TYPES or justify the omission.`
```

`mob.ts`'s 13 include `MarriageRegistration`, `MarriageLicense` and `MarriageNotice`. I triggered it with `MarriageLicense` and got that message. A developer who follows it adds the type to `COUPLE_EVENT_TYPES`, which feeds `SWEPT_SPOUSE_FACT_TYPES` through the spread, and `spouses-of` starts deleting civil registration records.

The 2026-08-20 ruling on that exact class:

> a registration is the civil record; a notice is banns... these are not redundant, and deleting them would destroy evidence on an irreversible tool operating on user trees

The borrow is also semantically off at the source: `mob.ts:27-30` says those sets mirror Java MobWarnings' date-plausibility families, not a couple-conclusion ontology, which is why `BIRTHLIKE` contains `Adoption` and `DEATHLIKE` contains `Will`. "Confirmed couple-event vocabulary" is this test file's framing, not `mob.ts`'s.

Smallest fix I can see: own the vocabulary locally in the test rather than borrowing it, listing the GedcomX couple events including `CommonLawMarriage` and `DomesticPartnership` and excluding the documentary-record types, and change the remedy sentence so it does not read as an instruction to sweep.

## 3. The gap the issue was filed about is guarded by nothing

This is the one I nearly mis-scoped, so here is the measurement first. Adding `CommonLawMarriage` and `DomesticPartnership` to `EVENT_TREE_TYPES` without adding them to `COUPLE_EVENT_TYPES`, so they are recognised couple events that `spouses-of` never sweeps:

```
drift guard        Tests  4 passed (4)
behavioural suite  Tests  79 passed (79)
FULL engine suite  Tests  2710 passed (2710)
```

Nothing anywhere fails. That is acceptance criterion 3 in the issue's own words:

> Add a guard test cross-referencing `materialize-facts.ts`'s couple-event subset of `EVENT_TREE_TYPES`: it fails when a couple-event type is neither in the swept set nor explicitly annotated as "confirmed not echoed person-level" - so a future couple type can't be added to one list and silently missed by the other.

and it is the scenario the Root section describes as "the exact #1417 leak under a new type".

Worth separating the two directions, because they are not in the same state:

- **Shrink** (a swept type is deleted): the guard stays green, but the behavioural suite catches it. Deleting one type, deleting all three, and emptying `COUPLE_EVENT_TYPES` outright all redden `tree-forget.test.ts`. Covered, just not by the guard, and I am not asking for anything here.
- **Widen** (a new couple type is recognised but unswept): covered by nothing, per the run above. This is the direction the issue exists for.

Fixing item 2 fixes this too, if the locally-owned vocabulary includes those two types. That is why I would treat 2 and 3 as one change rather than two.

## Smaller, not blocking

- **The spec discloses the wrong hole.** `docs/specs/tree-forget-tool-spec.md:140-145` says a free-text type naming the spouse "isn't caught by this check at all". I ran it: `Separation+from+Anna+Owen` **is** caught now, and the issue body says so too. The class that genuinely is not caught is FamilySearch's camelCase spellings, `MarriageRegistration` and `MarriageLicense`, which are neither swept nor warned because `leadingToken` splits only on whitespace and `+`. Those are the same spellings the new guard calls confirmed couple events, so the PR currently has one file treating them as couple events and another ignoring them.
- **A doc cross-reference was lost.** On `main` the `Annulment` row carried "Swept by `tree_forget` `spouses-of` alongside `Marriage`/`Divorce`", which was the table's only pointer that those two rows are swept. The replacement note is per-row, so `Marriage` and `Divorce` now carry no sweep indication at all.
- **The visibility feature has no consumer path.** `forget-and-rederive/SKILL.md` is the only skill that calls this tool and contains the word "warning" zero times, while instructing "counts and kinds only" at `:135` and `:187`. The 08-20 ruling's purpose was surfacing this to the researcher; as shipped, an agent following that skill reports counts and never relays the warning. Worth a line in the skill, or a follow-on.
- **Latent, against the ruling.** When a canonical `Separation` and `Separation+from+Anna+Owen` share a fact id on one person, both are deleted with no warning and the spouse name lands in `removed.factsByType`. The ruling says that fact must survive and warn. I measured the corpus: 0 of 2,793 persons carry two facts sharing an id, so this is latent rather than live, and the shared-id class is #1574's subject.
- **Warning volume is uncapped**, 500 prefix facts on one person producing about 98,000 characters, on a tool whose stated design is counts-and-kinds to protect the agent's context.
- **`COUPLE_TYPES_CONFIRMED_NOT_ECHOED`** is still absent. It is arguably moot under the 08-24 ruling, since the set would be empty, but the issue body was rewritten after that ruling and still asks for it, so it wants an explicit "dropped because" rather than silence. Same for the parent-side "record that result" line.
- **The test count does not reproduce**, third round running: the body says 2743 on 112 files, `npm test` gives me 2710 on 112. `make engine-test` is exactly `npm test`, so both figures cannot describe this head.

clack391 (approved):
Approving. The drift guard now owns its own couple-event vocabulary instead of borrowing one that mixed swept and documentary types, and that was the right call. Everything below is non-blocking; none of it needs to hold up the merge.

## Two corrections I owe you from my earlier rounds

1. I called `tree_forget` an irreversible tool. It is not. `RESTORE_FILE` (`.tree-before-forget.gedcomx.json`) is written pre-removal and deliberately never overwritten, so it keeps pointing at the original (`tree-forget.ts:975-978`). That restore path is part of why this widening is safe, and my framing overstated the risk in both prior reviews.
2. My round-2 item named `CommonLawMarriage` and `DomesticPartnership` as examples rather than asking for the class. The guard you built matches exactly what I asked for. Where the vocabulary is short, that is on my wording, not your implementation.

## Verification I ran at `30981594`

Guard and behavioural suite run separately, full engine suite 112 files / 2710 tests green, tree restored clean after every mutation.

| mutation | drift guard | behavioural |
|---|---|---|
| `CommonLawMarriage` / `DomesticPartnership` into `EVENT_TREE_TYPES` | RED | green |
| delete a type from `COUPLE_EVENT_TYPES` | green | RED |
| documentary types into `COUPLE_EVENT_TYPES` | green | green |

Corpus re-measured over 422 committed trees: `Engagement` / `MarriageBanns` / `Separation` / `Divorce` / `Annulment` all 0 person-level, `Marriage` 93. The widening deletes nothing new today and newly warns on 14 facts, matching the issue's numbers.

PR #1865, opened 2026-08-24
Year qualifiers: measure the range instrument and rewrite the doctrine (#1771)

### LINKED ISSUE #1771: Year qualifiers describe a population that does not exist — rewrite H/Y's instrument around indexed date ranges

chesworthrm:
Closes #1771.

## What and why

The year-range `*Exact` family was documented around a population that does not exist — *records with no indexed year*, enumerated at zero. This PR builds the instrument that measures what the index actually does, rewrites the doctrine to match, and rewires the probe verdicts and guards.

**The finding:** a record with no year of its own carries an **estimated date range** (from the dated facts of others on it); an unqualified range matches by **overlap**, and `.exact` keeps only records whose indexed date is inside the range. There are no year-silent records. Measured on 10 enumerated pools across both endpoints + a non-parish US census control; reproduced on the tree endpoint.

## What's in it

- **Instrument (steps 0–3):** `dev/probe-search-qualifiers.ts` sections **P** (person_search/tree) and **Q** (record_search) — disjoint 50-year band membership per family, `.exact` counterparts, closure + payload-dated control, recorded to `dev/measured-figures.json`.
- **Verdict rewire (steps 1b+4):** retired H's dead estimator (partition algebra / smallest-bucket / silentEstimatesAgree); H/N/Y verdicts now derive from the band instrument; keys renamed/deleted per #1771's table, each with its `FORBIDDEN_WHEN` edit in-commit.
- **Prose payoff:** the measured year clause written into `birthYearExact` (both tools, 475→207 chars, 240-ceiling exemption deleted), both specs, `SKILL.md`, and two skill references — replacing the #1409 "not established" pointer.
- **Guards:** `measured-figures.test.ts` repointed + broadened (fails-closed, catches the disproven vocabulary and a line-wrapped record-index overclaim via a whitespace-collapsed scan); guard-resolves + the new contradiction pair bind.

## Decided deviations (please read — they diverge from #1771's text, with evidence)

- **Payload-dated control is *right-band + not-axis-spanning*, not the issue's "exactly one band"** — the finding itself contradicts "exactly one band" (an unqualified range matches by overlap). Records overlap 0, tree overlap 5. Proven the guard still fires without the switch.
- **Records `residence` is the honest exception** (precise event dates, 198→198, no overlaps), so `Y` reads DOES NOT GENERALISE; the mechanism holds for birth/death/marriage and reproduces for all four on the tree endpoint.
- **Marriage is a relationship fact**, attributed through the Couple relationship, not `persons[0]`.
- A drift-prone pool total the spec cited is now stated approximately, with the exact value left to the artifact (it re-measures each run).

## Verification

- Full engine suite **green (2677)**; typecheck clean. Main merged in.
- Guards **proven to fire** (both the H disproven-vocabulary rule and the Y record-index-overclaim rule) and proven not to false-positive on the true tree wording.
- `Nothing in CI checks whether a tool description is true` — the manifest test compares names, not descriptions; the traceability/wording guards are what cover this.

## Eval regrade + annotation (the #1771 gate)

- `make eval-skill SKILL=search-records` re-run against the post-#1799 merged snapshot (run log `v1_2026-08-24_16-32-45`, superseding the earlier `v1_2026-08-24_12-28-26`): 22 pass / 6 fail; genealogist annotation by chesworthrm@familysearch.org, 0 overrides. Run log + `.ann.json` committed.
- The 6 fails all trip one deterministic validator (`test_pre1880_census_structure_marked_inferred`, `releasable: true`): the skill intermittently omits the pre-1880 census "inferred structure" caveat. The failing set shifted between runs (`017` recovered, `014`/`027` newly failing), which is further evidence of **variance, not a regression** — orthogonal to this PR, a separate lane-3 search-records craft gap worth its own issue if it keeps costing runs.

## Known exposure (per #1771, unchanged)

- The eval fingerprint stays as-is (#1770); re-snapshotting `mcp-server/src` would invalidate every run log. Only `search-records` was regraded; the other model-visible suites were knowingly not re-run.


DallanQ:
**Coordinate the merge order with @promise-emmanuel on PR #1799 before either of you merges.**

You are both changing the `search-records` eval snapshot from separate branches off `main`:

- This PR changes `search-records/SKILL.md` plus `references/place-date-mechanics.md` and `references/search-strategy-levers.md`.
- PR #1799 changes `eval/tests/unit/search-records/live-callee-external-sites-escalation.json` (content hash `a6a99a40…` → `ab10cf8e…`).

You have each committed your own fresh `search-records` run log today. Whichever merges second has its run log go inactive against the new snapshot, `check-runlogs` goes red, and that run has to be bought again — a paid `make eval-skill SKILL=search-records` plus a full re-annotation.

Whoever goes second:

1. Rebase on `main` after the other lands.
2. `make eval-skill SKILL=search-records`
3. Re-annotate the new run log and commit it in this PR.

Nothing is wrong with either change — this is only about sequencing so you buy one extra run instead of two.

packages/engine/mcp-server/src/tools/record-search.ts
```diff
@@ -1331,7 +1331,7 @@ export const recordSearchToolSchema = {
 
       residenceYearFrom: { type: "number", description: "Lower bound of the residence-year range (typically census-style anchor). 4-digit year (e.g., 1860). Must be paired with `residenceYearTo`." },
       residenceYearTo: { type: "number", description: "Upper bound of the residence-year range. 4-digit year (e.g., 1870). Must be paired with `residenceYearFrom`." },
-      residenceYearExact: { type: "boolean", description: "As `birthYearExact`, for the residence-year range." },
+      residenceYearExact: { type: "boolean", description: "For the residence-year range. Unlike `birthYearExact`, record-index residence dates are precise, so an unqualified range finds no estimate overlaps and `.exact` changes nothing." },
```
jud-sdev: This says `.exact` changes nothing for residence, but the artifact shows it does. The residence block in `measured-figures.json` (the Y section, Geach/England pool) reads "RETAINS ESTIMATE OVERLAPS, .exact REMOVES THEM" — 7 rows with no payload residence date, kept by an unqualified range and all dropped by `.exact`. That's 7 rows, above the floor of 5 used elsewhere (death at 3 rows was called "too few"), so it isn't noise.

The "residence is precise" result is real, but it comes from the disjoint-band pool (`Q.bands:records-residence`), where all 198 rows carried a date — so there were no undated rows to overlap. Once undated residence rows show up, residence behaves like the others. So the claim holds for that pool, not in general.

Two things to reconcile:
- The sentence is absolute ("finds no estimate overlaps and `.exact` changes nothing"), but the Geach block denies it. Could make it conditional — residence records usually carry a precise date, so overlaps are rare and `.exact` usually changes nothing, but undated ones do occur and behave like the rest.
- The spec grounds this on `Y.verdict:generalises past birth`, which is the section whose residence block contradicts it. The real support is `Q.bands:records-residence` — worth citing that one instead.

Same wording is in the spec, SKILL.md, `place-date-mechanics.md` and `search-strategy-levers.md`, so they would move together.

jud-sdev (changes requested):
Went through this line by line, and the birth, death and marriage rewrite holds up well. I traced every cited verdict back to `measured-figures.json` and the estimate-overlap story is solid on the H and Q bands.

One thing I'd reconcile before it goes in, on the residence claim. The `residenceYearExact` text — and the matching "residence is the exception" lines in the spec, SKILL.md, `place-date-mechanics.md` and `search-strategy-levers.md` — say residence finds no estimate overlaps and `.exact` changes nothing. The residence block in the artifact shows the opposite. Details in the inline comment.

Everything else I'd sign off on.

packages/engine/mcp-server/src/tools/record-search.ts
chesworthrm: Confirmed and fixed in b04e677d — thanks, you traced it exactly right.

I reproduced both halves against the artifact:
- `Y.silence:residence` / the impossible-range residence block: 7 undated residence rows (Geach pool), kept by the unqualified range's estimate overlap, all dropped by `.exact` (`silent=7, silentKept=0`) — above the floor of 5, so a real result, not noise.
- The "precise" reading comes solely from `Q.bands:records-residence`, an all-dated pool (`payloadDated=198/198`, `indexSilent=0`) — no undated rows to overlap. And the `Y.verdict:generalises past birth` = DOES NOT GENERALISE it was cited against is itself *driven* by that all-dated band pool (`exactRowsRead == bandRowsRead == 198`, so `.exact` drops nothing there), while the same section's residence block says the opposite.

So I made all five sites conditional rather than absolute — residence records are *usually* precise, so overlaps are rare and `.exact` usually changes nothing, but undated ones do occur and behave like the other families (overlap keeps them, `.exact` drops them). The spec/table now cite `Q.bands:records-residence` for the "usually precise" half; I left the `Y` verdict where it genuinely supports "does not generalise as a blanket rule". Sites: `residenceYearExact`, the spec prose + lever table, `SKILL.md`, `place-date-mechanics.md`, `search-strategy-levers.md`. Also corrected the stale rationale in the `measured-figures.test.ts` FORBIDDEN_WHEN comment (the rule itself is unchanged), and updated the description-length bookkeeping (`record_search` total 13,996 → 14,022).

Guard + packaging suites green; `tsc` clean.

One sequencing note: sites 3–5 touch the skill snapshot, so the `search-records` run log needs a re-annotated re-run. I'm holding that until #1799 lands per DallanQ's merge-order comment, then I'll rebase, re-run, and commit the fresh annotated log.

chesworthrm:
@jud-sdev ready for another look.

- **Residence reconciliation** (b04e677d): the absolute claim is now conditional at all five sites — residence records are *usually* precise, so overlaps are rare and `.exact` usually changes nothing, but undated ones do occur and behave like the other families. The "usually precise" half now cites `Q.bands:records-residence` (198/198 dated) instead of leaning on the self-contradicting `Y` verdict; the impossible-range residence block (Geach, 7 rows) grounds the "behave like the rest" half. Full detail in the reply on your inline thread.
- **Merged `main`** now that #1799 has landed, resolving the merge-order coordination.
- **Fresh annotated run log** (f14be883): `make eval-skill SKILL=search-records` re-run against the post-#1799 merged snapshot — 22/28 pass, genealogist annotation, 0 overrides. `check_runlogs` is green (the run log is active against both #1799's fixture and this PR's skill edits, covered in one run).

Packaging + typecheck green. Thanks again for catching the residence contradiction.

promise-emmanuel (changes requested):
The birth/death/marriage rewrite holds up, and the band instrument is a real
improvement over the estimator it replaces. I traced the H and Q verdicts back
to `measured-figures.json` and reproduced the estimate-overlap story. Four
things to reconcile before it goes in, all on the residence half and the guards
around it.

**1. The residence claim is contradicted by this PR's own census control.**

jud-sdev caught this against the 7-row Geach block and you hedged it to
"usually". But there is a larger counter-example in the artifact. The "usually
precise" half rests on `Q.bands:records-residence` — one parish pool, 198 rows,
all dated, `.exact` drops nothing. The same run recorded a second record-index
residence pool, `Q.bands:records-uscensus-residence`, the non-parish control
added so the year prose would not rest on one collection. It reads the other
way: 560 rows, 57 dated personas in more than one band, and `.exact` drops 58 of
526 band matches — 11%.

Two record-index residence pools, 0% and 11%. One of two is not "usually", and
this is the description the model reads before every census search — the same
parameter whose text says "typically census-style anchor".

You wrote:
    residenceYearExact: { type: "boolean", description: "For the residence-year range. Record-index residence dates are usually precise, so a range rarely finds estimate overlaps and `.exact` usually changes nothing — undated ones behave like `birthYearExact`." },

Change it to:
    residenceYearExact: { type: "boolean", description: "For the residence-year range. Behaves like `birthYearExact`. How much `.exact` drops varies by collection: nothing in a parish pool where every row was dated, 11% of band matches in a US census pool." },

199 chars, under the ceiling. The same claim is at
`record-search-tool-spec-v2.md:1346`, `SKILL.md:202`,
`place-date-mechanics.md:147` and `search-strategy-levers.md:216` — they move
together, and the spec line wants `Q.bands:records-uscensus-residence` cited
beside `Q.bands:records-residence`. Editing a description shifts
`DOCUMENTED_TOTALS`, so re-derive both numbers.

**2. The generalisation verdict reads one of the two residence pools, and the
guard on it fails open.**

`YEAR_BAND_SECTIONS` (probe:4888) names only `["Q", "bands:records-residence"]`
for residence. The census entries are recorded to the artifact and read by no
verdict, no guard and no prose citation.

That one choice decides the whole "residence is the honest exception" carve-out.
`showsMechanism` is `indexSilent === 0 && exactRowsRead < bandRowsRead`; the
parish pool gives `198 < 198` = false, so Y reads DOES NOT GENERALISE. The
census pool gives `468 < 526` = true. Swap which pool is read and it flips.

The flip is silent. `measured-figures.test.ts:549` is
`if (!rule.activeWhen.test(recorded)) return;`, and nothing asserts a rule is
still active. I set `Y.verdict` to the census pool's reading, put the
four-family wording your new rule had just caught back into
`place-date-mechanics.md`, and the suite went 23/23 green. One undated row
turning up in the parish pool on the next re-run does the same thing on its own.

Add after the dangling-key test at `measured-figures.test.ts:613`:

    it("every wording rule is active on the recorded verdicts", () => {
      const inert = FORBIDDEN_WHEN
        .filter((r) => !r.activeWhen.test(String(get(fig, r.verdict) ?? "")))
        .map((r) => r.verdict);
      expect(
        inert,
        `a wording rule is INERT: its verdict no longer matches activeWhen, so it\n` +
          `  guards nothing while the prose it forbids stays in the tree. Either the\n` +
          `  measurement changed (rewrite the prose too) or the rule needs repointing.`
      ).toEqual([]);
    });

Then either add the census residence pool to `YEAR_BAND_SECTIONS`, or say in a
comment why a recorded control feeds no verdict.

**3. The payload-dated control's coverage is not recorded, and on one pool it
excludes the rows that would break it.**

`datedInAxis` is computed at probe:5152 and probe:5409 and returned in
`BandResult`, but neither `record()` call writes it. So the artifact says
`controlHolds: true` with no way to see how much of the dated population the
control actually examined.

It matters on one pool. `bands:records-uscensus-residence` has 560 distinct and
`inNoBand` 137, so 423 personas are in at least one band, against `payloadDated`
459 — at least 36 payload-dated personas are in no band. Since `controlHolds` is
true, none of them can carry an in-axis year, because a dated-in-axis persona
missing from its own band would have set `wrongBand` and forced a refusal. So
all 36+ have payload years outside 1400–1999, and `yearOf` accepts `20\d{2}`,
which the 12-band axis stops short of. The control is passing by filtering out
exactly the rows that would have failed it, and the artifact does not show it.

You wrote:
    payloadDatedInMultipleBands: overlapping,

Change it to:
    payloadDatedInMultipleBands: overlapping,
    payloadDatedInAxis: datedInAxis.length,
    payloadDatedOutOfAxis: dated.length - datedInAxis.length,

Both sites. If that out-of-axis count is non-zero on a US census pool, it wants
explaining before the control counts as one.

**4. "A small cohort" is up to 24.5%, and the tool descriptions drop the caveat.**

`inNoBand` across the ten pools: 0, 0, 0, 4.1%, 6.6%, 8.1%, 8.1%, 8.1%, 8.8%,
24.5%. "Small" is fair for the median and wrong for the tail.

The bigger half is that the caveat and its instruction — read the results back
rather than trusting a range to gather them — is in all four prose surfaces and
in neither tool description. That is the surface the model reads, as
`tool-description-length.test.ts` says in its own docstring.

You wrote:
    birthYearExact: { type: "boolean", description: "With `true`, only records whose indexed date is inside the range survive. Unqualified, a range also admits records whose estimated date range merely overlaps it, including ones carrying no year of their own." },

Change it to:
    birthYearExact: { type: "boolean", description: "With `true`, only records whose indexed date is inside the range survive. Unqualified, a range also admits records whose estimated date range overlaps it. Records with no indexed date at all are reached by neither." },

214 chars. Same edit at `person-search.ts:376` with "objects". In the four prose
surfaces, replace "a small cohort" with the measured range.

**Two small ones.**

`explore-year-band-verify.ts:20` records a result the script cannot compute.
Check C prints the two counts and asserts only that every payload-dated-in-band
persona is in the unqualified band read — it never compares the `.exact` set to
the dated set, so equal cardinality is all the run showed.

You wrote:
     *   C band 1750-1799  unqualified 364 -> .exact 29; the 29 .exact survivors are exactly the 29

Change it to:
     *   C band 1750-1799  unqualified 364 -> .exact 29; the .exact count equals the 29

Or compute the identity — `bandEx.rows.every(r => inBandDated.some(d => d.id === r.id))` —
and keep the stronger sentence.

And the PR body's eval counts describe the superseded run. "23 pass / 5 fail" is
`v1_2026-08-24_12-28-26`; the log this lands on, `v1_2026-08-24_16-32-45`, is 22
pass / 6 fail, with `017` recovered and `014`/`027` newly failing. Your
diagnosis survives — all 6 trip only
`test_pre1880_census_structure_marked_inferred`, `releasable: true`, and the
shifting set is extra evidence for variance rather than regression. Just correct
the numbers, since the body is what the next reviewer reads it against.

Everything else I'd sign off on. Suites green here too: engine vitest 2688,
`tsc` clean, harness pytest 2671. Both of your new wording rules fire, including
on line-wrapped text, and every figure in the two test docstrings re-derives
correctly from `allToolSchemas`.

chesworthrm:
@promise-emmanuel thanks — this was a sharp review, and all six are addressed in `567d570a` (+ the eval re-run in `65b38c63`). Point-by-point:

**1 — residence is collection-dependent, not "usually precise."** Reworded at all five sites. One deviation I want to be explicit about: I did **not** put `11%` / `24.5%` into the tool descriptions or skill prose. Those are `58/526` and `137/560`, which the probe re-measures live every run — a bare percentage isn't even caught by the drift guards (`measured-figures.test.ts` deliberately excludes them), so it would sit in the most model-facing surface silently disagreeing with the artifact on the next run. Instead the descriptions/skill prose carry the behaviour *qualitatively* ("depends on the collection — nothing where every row is dated, a meaningful share where many are dated only through others"), and the precise figures live in the **spec**, cited to `Q.bands:records-uscensus-residence` beside the parish pool — the evidence trail, per this PR's own "state approximately, leave the exact to the artifact" convention. Same substantive fix; no volatile number in a description. Happy to revisit if you'd rather see the figures inline.

**2 — verdict reads one pool; guard failed open.** Added the fail-closed `every wording rule is active on the recorded verdicts` test (proven-to-fail by flipping a rule's `activeWhen`, then restored), and a probe comment at `YEAR_BAND_SECTIONS` documenting why the census pool is a same-family cross-collection control, not a fifth family fed to the verdict.

**3 — control coverage; answered in-PR.** `payloadDatedInAxis`/`payloadDatedOutOfAxis` now recorded at both band sites, and the control's scope is documented where it lives. The answer to "is out-of-axis non-zero on census, i.e. is the control filtering out the rows that would fail it?" is **yes, provably ≥36**: on `records-uscensus-residence`, `payloadDated` (459) exceeds in-band capacity (`distinct − inNoBand` = 423), and `controlHolds` forbids an in-axis dated persona landing in no band, so those ≥36 must be out-of-axis. So `controlHolds: true` there holds over the *in-axis* dated population only — a coverage bound, not a refutation (the mechanism verdicts concern in-axis behaviour). Documented at both control sites.

**4 — `birthYearExact` caveat + "small cohort."** Added "Records with no indexed date at all are reached by neither" to both tools' `birthYearExact`; replaced "a small cohort" with "a cohort that is not always small" + the read-back instruction across all four prose surfaces.

**5 — `explore-year-band-verify.ts:20`.** Docblock corrected to the cardinality+subset claim Check C actually computes, not the id-for-id identity.

**6 — PR body.** Corrected to the landed run (22/6 for `v1_2026-08-24_16-32-45`; the current active log after this round's re-run is `v1_2026-08-25_00-44-28`, 23/5, re-annotated, 0 overrides).

Also fixed a self-introduced drift a code-review caught: the length-test docstring still quoted `birthYearExact` at 207; the caveat lengthened it to 214, so `DOCUMENTED_TOTALS` and the prose are re-derived (record_search 14043, person_search 5003).

Engine packaging green, `tsc` clean, `check_runlogs` satisfied. Ready for another look.

jud-sdev (approved):
Reconciled. The residence claim I flagged is fixed at all five sites — `residenceYearExact` now reads as collection-dependent (nothing dropped where every row is dated, a real share where many are dated only through others) rather than an absolute no-op, and the numbers are cited to `Q.bands:records-uscensus-residence` where the evidence actually lives. `birthYearExact` picked up the matching no-indexed-date caveat, and the length bookkeeping was re-derived to match.

The part that went past my finding is the better catch: the wording guard read a single residence pool and would have failed open if that verdict flipped. The new "every wording rule is active on the recorded verdicts" test closes that, and it's proven to fail by flipping a rule's activeWhen — so it's a real guard, not one that only looks like coverage.

CI is green at this head and every thread is resolved. Thanks for tracing both halves against the artifact.

promise-emmanuel (approved):
Approving. I re-verified on `e26d6f77` after the resync: engine vitest 2702/2702,
`tsc` clean, `check_runlogs` all rules satisfied, run log snapshot 0 diffs
against the tree, all required checks green. The resync touched nothing under
`search-records`, so `v1_2026-08-25_00-44-28` is still active.

All six findings are fixed, and I checked them rather than taking the summary on
trust. The one that mattered most: I re-ran the exact mutation that went 23/23
green last round — flip `Y.verdict` to the census pool's reading, put the
four-family record-index wording back into `place-date-mechanics.md` — and the
new `every wording rule is active on the recorded verdicts` test catches it now.
That hole is genuinely closed. The `YEAR_BAND_SECTIONS` comment is also a better
answer than the repoint I asked for: the census pools are a same-family
cross-collection control, not a fifth family, so feeding them to a
generalisation verdict would conflate two questions.

And you were right to keep `11%` / `24.5%` out of the tool descriptions and
skill prose. Those are `58/526` and `137/560`, the probe re-measures both every
run, and `measured-figures.test.ts` deliberately excludes AGENT_SURFACES from
traceability — so a bare percentage there would be the one figure in the repo
that can drift with nothing watching it. Same substance, volatility parked where
the guard reaches it. My ask was wrong.

Two non-blocking things. Push them if convenient, or take them as follow-ups —
neither needs another review round.

**1. The PR body cites the superseded run log.** It names
`v1_2026-08-24_16-32-45` at 22 pass / 6 fail, but `567d570a` changed the skill
snapshot, which is why `65b38c63` bought a new run. Your comment has the right
numbers; the body doesn't, and the body is the merge record.

You wrote:
    - `make eval-skill SKILL=search-records` re-run against the post-#1799 merged snapshot (run log `v1_2026-08-24_16-32-45`, superseding the earlier `v1_2026-08-24_12-28-26`): 22 pass / 6 fail; genealogist annotation by chesworthrm@familysearch.org, 0 overrides. Run log + `.ann.json` committed.

Change it to:
    - `make eval-skill SKILL=search-records` re-run against the residence-reconciliation snapshot (run log `v1_2026-08-25_00-44-28`, superseding `v1_2026-08-24_16-32-45`): 23 pass / 5 fail; genealogist annotation by chesworthrm@familysearch.org, 0 overrides. Run log + `.ann.json` committed.

The "6 fails" sentence below it wants the same treatment — I confirmed all 5
fails in the active log trip only `test_pre1880_census_structure_marked_inferred`
with `releasable: true`, so the variance diagnosis holds across every committed
log.

**2. `tool-description-length.test.ts:36` — the fourth-place claim went stale in
the same commit that fixed the third.** `residenceYearExact` moved 203 -> 217, so
it is now what sits just below the top three, not the 214 pair. The top three
are still right and still asserted executably; `DOCUMENTED_LONGEST` binds only
three entries, so this sentence is unguarded prose — the same drift this
docstring says has caught it out three times.

You wrote:
     * `record_search.birthPlaceExact` at **226**. Both `birthYearExact` toggles now
     * sit just below them at **214** (tied), so neither makes the top three, and since
     * #1771 there is no exemption.

Change it to:
     * `record_search.birthPlaceExact` at **226**, with
     * `record_search.residenceYearExact` fourth at **217**. Both `birthYearExact`
     * toggles sit below that at **214** (tied), so neither makes the top three, and
     * since #1771 there is no exemption.

Better still, add `["record_search.residenceYearExact", 217]` to
`DOCUMENTED_LONGEST` and widen `topThree` to four, so the next edit to a
`*YearExact` description can't drift this line again.

One note that is **not** on you: merging main brought in #1766's judge-prompt
change, so `check_runlogs` now warns that the run log was scored against judge
hash `0d186137…` while current is `c39d7003…`. It's a warning, not a gate, and it
will hit every open PR carrying a run log — I don't think this PR should buy a
third paid eval over it. @DallanQ flagging in case #1766 wants a sweep.

mercyokum (approved):
Approving. Everything looks good genealogically

Issue #1852, opened 2026-08-24
Convert conflict-resolution into a skill-agent pair — 47 of the 56 runs that write a conflict never invoke it


DallanQ:
## Decision needed — #1852 gate

Must the `resolution_kind` writer-tool gate (lead ruling 2026-09-07, issue #2203; now unassigned in the "issue #2324" section of issue #2475) land before this conversion's fold? `docs/skill-to-agent-pair-conversion.md` step 4 requires a rule in code before the prose moves on a guardrail conversion, and nothing says whether this card waits for issue #2475, carries the gate itself, or goes without it.

- **A — Gate first, in this card** — One developer PR of latency before the conversion and zero paid runs (`build_snapshot` skips MCP source, `snapshot.py:159`). Adds a persisted closed enum the lead already ruled; PR 1 carries the closed-enum site list (both `enums.schema.json` trees, `CLOSED_ENUMS`, both `research.schema.json`, `packages/schema/src/index.ts`, `research-schema-spec.md`). Delays issue #2266 by that PR.  *(recommended)*
- **B — Wait for issue #2475** — No scope added here, but issue #2475 is an unassigned senior consolidation card with a 59 KB body, so this card and issue #2266 (high-priority floor search) wait on it with no date.
- **C — Convert without a gate** — Fastest route to issue #2266; the hook route still ships. The rule "a resolved conflict always names a winner" crosses the delegation boundary as prose, and 12 of the 68 resolved corpus conflicts already lack a winner (measured 2026-09-24, the script under "The bypass, measured").

<details><summary>Body text if A</summary>

## Writer-tool rule — decided (lead, <date>): land the gate first, as PR 1 of this card
[...]

PR #1870, opened 2026-08-24
proof-conclusion: page research_query, and pin the hypothesis-tracking boundary (#1183, #1374)

### LINKED ISSUE #1183: proof-conclusion must be taught to page research_query (the skill half of #1031)

benter-070:
## Summary

**Tier: Normal.** Closes #1183 and #1374 — each issue's last remaining piece, combined because they share `proof-conclusion`'s eval snapshot. Together: **2 paid runs, 2 annotation passes.** Separately: 3 and 3. #1832 merged and freed the slot.

- **#1183** — the agent no longer promises pagination away. It claimed *"no offset/pagination guessing"* directly beneath a query set including `{ section: "assertions", questionId }`, while `research_query` caps `items` at 50 and reports the true total as `count`. Step 1's **"collect every assertion"** gate could therefore be satisfied by a first page, and the agent would select a tier from a partial set — silently.
- **#1374** — the third and last of the pairs `hypothesis-tracking`'s DO-NOT clauses call for. The `conflict-resolution` and `timeline` pairs shipped in #1783.

## Review guidance

**Start here:** `agents/proof-conclusion.md`, the paragraph after the `research_query` list. It names the real response fields rather than paraphrasing — `count` is the total *before* the cap and any offset, `items` caps at 50, `truncated` is `count > offset + items.length` — so `truncated` is the loop condition and `count` the completeness check (`src/tools/research-query.ts`). It also carries the two things `gps-mentor.md`'s version omits: **`offset` named as the remedy**, and the fact that this particular filter has **no narrower form that still returns every assertion for the question**, so paging is the only option there.

**Note the target moved.** #1183's body points at `skills/proof-conclusion/SKILL.md:117-119`, which #1819 dissolved when it made `proof-conclusion` a skill-agent pair. The skill is now a 63-line wrapper mentioning `research_query` zero times. An agent body folds into its skill's snapshot, so the cost is unchanged.

**Unsure about — and this is the one I'd like ruled on.** The two directions of the new pair do **not** rest on equal ground:

- `hypothesis-tracking → proof-conclusion` (`ut_hypothesis_tracking_016`) rests on an explicit clause: *"Do NOT use when the user … wants to write a final conclusion (use proof-conclusion)."*
- `proof-conclusion → hypothesis-tracking` (`ut_proof_conclusion_021`) has **no such clause**. `proof-conclusion`'s DO-NOT list names `conflict-resolution`, `question-selection` and `record-extraction` only. So the reciprocal rests on its *trigger surface* being conclusion-writing for a resolved question, plus the scenario's `h_002` still being `active` so nothing is settled to conclude.

That is weaker, and it is the same concern that made me defer the `person-evidence` edge in #1783. **Widening `proof-conclusion`'s description to name `hypothesis-tracking` would be the cleaner fix** — I did not do it here because it is a separate lane-5 change to a description, and this PR already carries a doctrine edit. If you would rather see that instead of the reciprocal test, say so and I will swap them.

## Plan

**Didn't change:** `proof-conclusion`'s description (above), and the `person-evidence ↔ hypothesis-tracking` edge, still unpinnable until one of the two descriptions excludes the other.

**Acceptance check:** `ut_hypothesis_tracking_016` and `ut_proof_conclusion_021` both pass. The `#1183` half has no test — it is a prose fix to a latent correctness bug, and the corpus evidence is that the bug is reachable but not yet caught: across the e2e corpus `research_query` was called **867 times and passed `offset` 5 times**, with **0 of 336** observable `truncated` flags true. Measured while scoping #1808, recorded on #1183.

**Deviated from the plan:** combined two issues in one PR, which the `/fill-ready` eval-slot constraint favours and a reviewer may not. The trade is stated on both issues: one paid run saved, at the cost of asking for two doctrine judgements at once.

## Test plan

- [x] `make test-all` — **no `make` on this machine** (Windows; #1185). CI runs the suites. Ran what this change can affect: `check_negative_reciprocity.py` (both new edges reciprocated; graph 83 edges/25 skills → **85/25**; unbacked count unchanged at 45), `unit-test.schema.json` on both new tests, and a corpus-wide `test.id` uniqueness scan (none across 402).

- [x] Ran both affected suites in **one invocation** and committed both run logs and `.ann.json`: `v1_2026-08-23_19-10-10`, both verified **active** after merging main. `hypothesis-tracking` **16 of 16 pass**. `proof-conclusion` 17 pass, 3 partial, 2 fail — investigated below.

### `proof-conclusion`'s failures, investigated rather than shipped blind

| test | finding |
|---|---|
| `_009` | **Corrected after review — my original row was wrong in both halves.** It does *not* fail on clean `main`: all five committed logs are non-fail (`04-34-52` xpass, the other four pass), every one with `skills_invoked=['record-extraction']`. Nor was it "a routing miss to `record-extraction`" — that is the *correct* destination. What happened is that **nothing routed at all**: my run has `skills_invoked=[]` and the main thread answered inline. This diff still cannot have caused it, and for a stronger reason than I gave: the agent body only loads *after* delegation, and no delegation occurred. There is also no valid baseline on the current body — `94292657` (#1832) edited `agents/proof-conclusion.md` on 2026-08-23 11:27, after all five 08-21 baselines, and merged with no re-run. Mine is the first run since. |
| `_011` | **Corrected after review.** "Passed 2 of 2 on retry" measured a body that no longer exists — those two passes came from uncommitted working-tree states during iteration (agent hash `ee03ebe8`), and the shipped body is `ea33e488`, which has exactly **one** observation: a fail. Across all five logs it is 3 fail / 2 pass. The failing check is a deterministic validator with the judge skipped, so "judge noise" was never an available explanation either way. Not caused by this PR — `test_conflict_blocks_proved`'s own docstring names "resolving the conflict itself" as one of three known behaviours. Mechanism and the population measurement are on [#1852](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1852#issuecomment-5398067298). |
| `_012` | **UNRESOLVED.** Partial twice with the change, pass once without. Its partial cites tree encoding not performed and a claim about resolved conflict `c_001` — neither touching pagination. I can neither show it is mine nor show it is not. |

**The committed run log is the one taken, not a re-roll** after the control runs showed the first sample unrepresentative. None of the three falls in the sampled five, so they ship visible and explained rather than annotated.


mercyokum:
Found while reviewing, not caused by this PR — flagging so it doesn't get lost.

`ut_proof_conclusion_011`'s run log (`v1_2026-08-23_19-10-10`) shows a real failure, not judge noise. Two validators caught it:

`test_ownership_table`:
"proof-conclusion modified sections it doesn't own: ['conflicts']. Allowed writers per section: {'conflicts': ['conflict-resolution']}"

`test_conflict_blocks_proved`:
"proof-conclusion concluded q_001 at ['probable'] while an unresolved conflict on an identifying attribute (c_001, birthplace) is open. The disputed attribute goes to whether the cited sources describe the same person, so no tier is available..."

On `flynn-with-birthplace-conflict` — a scenario built specifically to test this — the skill at least once wrote to a section it doesn't own and handed out a confidence tier while it was still an open question whether the records even describe the same person.

I get that this isn't caused by your change here: you retried it twice more and it passed both times, so it's occasional, and it has nothing to do with the paging fix. I checked #1852 (conflict-resolution ownership) — close, but that one's about other skills skipping conflict-resolution when they *create* a new conflict, not about proof-conclusion overriding one that's already open. No open issue names this failure.

Since you're already in proof-conclusion's doctrine for this PR, you're welcome to fix it here if it's a quick doctrine addition — or file it separately if it needs its own eval slot. Either way, just don't want it to disappear once this run log gets buried under the next one.

benter-070:
Verified both quotes verbatim — that's a real failure and a good catch. Two corrections to the framing, then the mechanism, which turns out to be more interesting than either of us assumed.

## Your "occasional" reading needs tightening

`ut_proof_conclusion_011` across all five committed run logs on this branch:

| run log | outcome | agent body hash | judge |
|---|---|---|---|
| `v1_2026-08-21_18-10-36` | fail | `8cb7a8dc` | old |
| `v1_2026-08-21_18-43-53` | fail | `e0ed0c3b` | old |
| `v1_2026-08-21_19-20-04` | **pass** | `ee03ebe8` | old |
| `v1_2026-08-21_19-34-21` | **pass** | `ee03ebe8` | old |
| `v1_2026-08-23_19-10-10` | fail | `ea33e488` | new |

The two passes are the **same** body, and it is **not** the committed one. Those were uncommitted working-tree states from my iteration — only the final state shipped. So the committed body has exactly **one** observation and it failed; "retried twice and passed" is measuring a body that no longer exists. Across all five it's 3 fail / 2 pass.

That matters because `test_conflict_blocks_proved` is a **deterministic validator**, not the judge — the judge was skipped on this test. So the differing judge hash on the last row is irrelevant to it, and "judge noise" was never available as an explanation in either direction.

## You are right that it isn't caused by this PR — and the evidence is better than flakiness

`test_conflict_blocks_proved`'s own docstring names the behaviour:

> the test flip-flopped across four runs on exactly these points while the skill produced three different behaviours: concluding at probable, declining silently, **and resolving the conflict itself**.

And `conflictedSourceInvariants`' docstring in `research-append.ts` records the same failure on 2026-08-19 and 2026-08-21, "across five successive wordings." So this predates the branch and is already in the repo's record.

## Why "a quick doctrine addition" can't be the fix

The doctrine is already in the agent body three times, including the exact case:

- `:114` — "**Conflicts (hard block, on identity).**"
- `:205` — "a birthplace dispute is not part of a parentage conclusion, yet it goes directly to whether the census entries and the death certificate are even the same man. Identity first, then tier."
- `:354` — "**Do not resolve conflicts here** — recommend conflict-resolution."

It was read and not followed. That is the ADR-0011 / `skill-lifecycle.md` §5 case, and CLAUDE.md cites **this same skill** as the measured example. The tool docstring puts a number on it: five wordings. A sixth is not the move.

## The actual mechanism — one unauthorized write defeats three correct defences

The run made two separate `research_append` calls:

**Call 1** — `{section: "conflicts", op: "update", entryId: "c_001", fields: {status: "resolved", independence_analysis: …, weighing_analysis: …, preferred_assertion_id: "a_009", resolution_rationale: …}}`. It did conflict-resolution's entire job. **`is_error` was null — the tool permitted it.**

**Call 2** — a batch updating `ps_001` and marking `q_001` resolved.

The tier gate is correct and would have bitten. The fixture is built so it must: `c_001`'s competing assertions are `a_002`/`a_009`/`a_012` → sources `src_001`/`src_003`/`src_004`, and `ps_001`'s supporting assertions `a_004`/`a_010`/`a_013` → **the same three sources**. Full overlap.

It didn't fire because `disputedSourceIds` skips `status === "resolved"` conflicts, and by call 2 the skill had already resolved `c_001` itself. Call 2's pre-call snapshot was clean.

All three of the tool's defences are sound and all three are bypassed:

1. pre-call snapshot — defeats resolve-and-conclude **in one batch**, not across two calls
2. `resultEntry` merge — correctly catches a stale `probable` on an untouched-tier update
3. not gated on `tierTouchedThisOp` — already fixed after the 2026-08-21 observation

The single load-bearing gap is upstream of all of them: **`ownership.json` is not enforced at the write boundary.** It declares `conflicts` writable only by `conflict-resolution`; the engine references that file **only in comments** (`research-append.ts:115`, `:507`, `:597` — no runtime read). So any skill can resolve a conflict it doesn't own, and then every conflict-aware gate downstream sees a clean world. `test_ownership_table` in the eval is the only thing that catches it.

That is a `nothing-checks` item in the CLAUDE.md sense: CI is green, the eval catches it, production ships without it.

## So: filing, not fixing here

Not a doctrine edit (measured dead end for this skill), and not this PR's lane — it's a TypeScript writer-tool change in `research-append.ts`, which needs a developer and a reviewer who owns that file. There's also a design call in it that isn't mine: whether to enforce `ownership.json` generally at the write boundary, or add a narrower deny for `conflicts`. The general version is the right shape but has blast radius across every section and six declared writers.

I'll file it with this mechanism, labelled `developer` + `nothing-checks`, and link it here. Thanks for not letting it get buried — the run log would have been pruned in five more candidates.

benter-070:
Correction to my last comment: I said I'd file this as a new issue. I ran the duplicate search first and **#1852 already covers it** — so it's recorded there instead: https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1852#issuecomment-5398067298

Your read was that #1852 is about other skills skipping conflict-resolution when they *create* a conflict, not about overriding one already open. That's not quite what it measures — worth knowing, because it's the card this belongs to:

- it counts any run that **writes** a conflict, resolving included, and names ours explicitly: **`proof-conclusion` 59 of 151 runs (39%)**
- its `eulogia-gatica-burial` example is the same two-step we hit — full `independence_analysis` and `weighing_analysis` written directly, skill never invoked, then a tier stamped over it
- its closing section already records the enforcement gap: all 19 rows of `ownership.json` read `override: "none"` and **nothing in the engine implements a refusal id**

So the ownership half was known. What our instance adds is that it's the first **unit-tier** case (all of #1852's are e2e, so this one reproduces off two deterministic validators), and — the part that matters — it **post-dates the gate**. `conflictedSourceInvariants` landed 08-21 16:33; this run is 08-23 19:10.

That last bit changes what the hole costs. Before the gate, an unowned conflict write skipped doctrine. After it, the same write **disarms a shipped guardrail that is otherwise correct**: the fixture has full source overlap between the disputed sources and `ps_001`'s supporting assertions, so the gate had to bite — it just never saw an unresolved conflict, because call 1 had already resolved it. Two separate calls each get a clean pre-call snapshot.

Filing a second card would have split that evidence away from the 39% measurement, which is what makes it actionable. Good catch either way — the run log would have been pruned in five more candidates.

mercyokum (approved):

T-FEH (changes requested):
Reviewed on the merged head. Branch was already current with `main`. I verified the numbers by reading the corpus rather than trusting the body, and **folded in the doc half myself** (`7ee7b806`) — what's left is two rows in your table.

## The core edit is right on every claim I could check

I checked each statement in the new paragraph against `research-query.ts` literally: `MAX_ITEMS = 50` (`:271`), `count: filtered.length` computed pre-slice and pre-offset (`:270`), `truncated: filtered.length > start + MAX_ITEMS` (`:272`) — equivalent to the body's `count > offset + items.length` in all three cases (full page, short page, offset past the end), and `const start = input.offset ?? 0` (`:266`). The "no narrower filter" claim holds too: `SECTION_FILTERS.assertions` exposes `recordId`, `recordRole`, `sourceId`, `questionId`, and the other three each narrow to a single record, role or source. Paging really is the only remedy there.

Eval discipline is clean. `check_runlogs.py` locally: **All runlog rules satisfied**. Both `.ann.json` files are UI-shaped, cover exactly their review_sample (5 and 6 tests), 0 disagreements and no comments — compliant, since every sampled test passed. Both new tests pass, ids unique corpus-wide, `runs_per_test` absent, and hypothesis-tracking is a genuine 16/16.

## On your ruling request: keep the reciprocal, don't widen the description

Your worry is real — proof-conclusion's description lists "summarize the evidence" as a trigger, and `flynn-competing-fathers` has 17 assertions with `h_001` at `supported`, so the scenario satisfies its stated activation condition while the test demands it decline. But the message carries two of hypothesis-tracking's triggers near-verbatim against one partial match for proof-conclusion, and the deliverable is a `hypotheses`-section account. Spec §6 sources negatives from DO-NOT clauses but never requires one, and your `explanation` field says plainly what it rests on.

The cost decides it: widening the description touches the snapshot and discards the run you already paid for. If it flakes later, *that* is the evidence to widen with, on proof-conclusion's next slot.

## 1. The `_009` row is contradicted by every committed log

> `_009` | **fails on clean `main` too** — pre-existing. A routing miss to `record-extraction`, unrelated to paging.

Both halves are wrong as written. All five proof-conclusion run logs on `main`:

```
04-34-52  009=xpass  skills_invoked=['record-extraction']
18-10-36  009=pass   skills_invoked=['record-extraction']
18-43-53  009=pass   skills_invoked=['record-extraction']
19-20-04  009=pass   skills_invoked=['record-extraction']
19-34-21  009=pass   skills_invoked=['record-extraction']
this PR   009=FAIL   skills_invoked=[]
```

Five for five. And "a routing miss to `record-extraction`" is backwards — `record-extraction` is the *correct* destination. What happened is that **nothing routed at all** and the main thread answered inline.

**I don't think it's yours, and I'm not asking for a paid re-run.** The agent body only loads after delegation and no delegation occurred, so this diff cannot have moved it — I initially suspected the opposite and the `skills_invoked: []` evidence rules it out. There's also a confound in your favour the body doesn't mention: `94292657` (#1832) edited `agents/proof-conclusion.md` on 2026-08-23 11:27, after all five baselines, and merged with no re-run. Yours is the first run since, so there is no valid `_009` baseline on the current body at all.

That sentence is stronger than the one you wrote — it just needs to be the one in the table.

## 2. The `_011` row is corrected in a thread but not in the body

You corrected this with @mercyokum: the two passes were uncommitted working-tree bodies, so the shipped body has one observation and it failed. The body still says "passed 2 of 2 on retry", and the body is what survives. Worth pointing at #1852 for the mechanism.

`_018` and `_019` are absent from the table but demonstrably pre-existing (`_018` partial in main's last three runs, `_019` in two of five) — no action, just noting it covers three of five non-passes.

## 3. Folded in: three docs that would have read false on merge — `7ee7b806`

`docs/architecture.md` §6.3 quoted the exact "no offset/pagination guessing" line this PR deletes and called the skill half pending; `research-query-tool-spec.md:7` and `:186-189` said the same. All three now state that the skill half landed.

Worth knowing for next time: my first attempt cited #1808/#1870 by number and **tripped the issue-reference ratchet** in `doc-links.test.ts` (`research-query-tool-spec.md: 9 refs, baseline 5`). Rewrote to name the files instead — the same "state the fact, not the ticket" rule your own `64de85bf` was applying. Engine suite green after: 110 files / 2687 tests.

## 4. Optional, and free: `gps-mentor` still gives advice this PR disproves

`agents/gps-mentor.md:126-128` says to "narrow the filter" on truncation, and at `:578` it makes the very call your paragraph identifies as un-narrowable. #1183 exempted gps-mentor as "the correct model", which was true before your paragraph. Only `research/SKILL.md` carries `@plugin:gps-mentor` and `research` is in `RUNLOG_GATE_EXEMPT_SKILLS`, so the edit costs no run. I left it to you rather than widening this PR's surface.

---

Requesting changes on the two table rows only — the code, the tests and the ruling are sound, and the docs are already done.

benter-070:
Thanks @T-FEH — both rows corrected in the body, and I took item 4 as well. Verified each of your claims against the corpus before editing rather than taking them on trust; all three held.

## 1. `_009` — you were right on both halves

```
04-34-52  xpass   skills_invoked=['record-extraction']
18-10-36  pass    skills_invoked=['record-extraction']
18-43-53  pass    skills_invoked=['record-extraction']
19-20-04  pass    skills_invoked=['record-extraction']
19-34-21  pass    skills_invoked=['record-extraction']
this PR   FAIL    skills_invoked=[]
```

Five for five non-fail, so "fails on clean `main` too" was simply false. And "a routing miss to `record-extraction`" was backwards — that is the correct destination; what happened is that nothing routed at all.

The row now carries your reasoning instead of mine, including the confound I had missed: `94292657` (#1832) edited `agents/proof-conclusion.md` on 2026-08-23 11:27, after all five 08-21 baselines, and merged with no re-run — so there is no valid `_009` baseline on the current body, and mine is the first run since. Your version of the exoneration is also the stronger one: the agent body only loads *after* delegation, and `skills_invoked: []` shows none occurred.

## 2. `_011` — the thread correction is now in the body

You are right that the body is what survives. It now states that the two passes measured agent hash `ee03ebe8`, an uncommitted working-tree state, while the shipped body is `ea33e488` with one observation and it failed; 3 fail / 2 pass across all five. Also that the failing check is a deterministic validator with the judge skipped, so "judge noise" was never available in either direction. Points at #1852 for the mechanism.

## 3. Docs — thank you for folding those in

`7ee7b806` catches three places I would have left reading false on merge. The `doc-links.test.ts` ratchet is worth me remembering: my own `64de85bf` was applying the "state the fact, not the ticket" rule and I would still have reached for the issue numbers there.

## 4. Taken — `ab810c2f`

`gps-mentor.md:126-128` told the agent to "narrow the filter" on truncation, and `:578` makes `research_query({ section: "assertions", questionId })` — the call this PR identifies as un-narrowable. So the advice was unactionable at its own call site. Changed to page with `offset` until `count` items are held. `grep` confirms no other plugin file still carries the advice, and the packaging suite is green (581 tests).

Free exactly as you said: only `research/SKILL.md` carries `@plugin:gps-mentor`, and `research` is gate-exempt *and* suiteless, so `exempt_suiteless` drops it — no gate armed, no run owed.

## On the ruling — agreed, and recorded

Keeping the reciprocal and not widening the description. Your cost argument is the deciding one: widening touches the snapshot and discards a paid run, and if it flakes later that is better evidence to widen on than my speculation now. Noted for proof-conclusion's next slot rather than lost.

Nothing here changes the diff's behaviour — two body rows, one agent line in a gate-exempt path. No re-run owed.

benter-070:
Follow-up on the ruling, @T-FEH — I agreed with not widening the description, but "wait for it to flake" turned out not to be the only alternative. There is a third option that costs the same and fixes the grading instead of the product.

**`grade_on_invariant` is built for this exact shape.** From its schema description: *"Use for routing-flaky negatives where every plausible route is state-safe."* It stops grading which skill fired and grades the resulting state, and the runnability gate refuses the test unless a real tag-gated validator backs it.

Every plausible route here is state-safe, which is the precondition:

| route | state |
|---|---|
| hypothesis-tracking fires (desired) | writes `hypotheses` — fine |
| proof-conclusion fires and declines | no write — fine |
| proof-conclusion fires and concludes | **invariant fails, correctly** |

Only the third row is a defect. Today the test also fails on the second — which is the flake you predicted, and it is a routing preference rather than a wrong outcome.

**The validator is written and cost nothing — `bf32937a`.** `test_open_candidate_blocks_closure`, gated on `hypothesis-open-blocks-tier`: while a rival hypothesis is `active`, no proof summary added for the question, tier not advanced above the pre-state's, question not marked `resolved`. `hypotheses[]` deliberately unasserted, since hypothesis-tracking may legitimately record evidence against `h_002` while answering — asserting it would fail the route we want.

Proven to fail first, per CLAUDE.md — seven synthetic cases, including a **vacuous-pass guard** that fails when the pre-state has no `active` hypothesis. Worth calling out because `grade_on_invariant` makes the invariant the only thing graded, so one that cannot fire is a silent green, and that is the failure mode the schema warns about.

Free because `eval/harness/validators/` is not in the snapshot. `tag_gated_validator_tags` already recognizes the tag, so the flag will be accepted rather than aborting the test.

**What is left is three lines in the test JSON** — the tag plus `"grade_on_invariant": true`. That file *is* in the snapshot, so it owes a paid proof-conclusion run, which is why it is not in this PR. Recorded on #1374 with the instruction to fold it into whatever next claims proof-conclusion's slot, where the marginal cost is zero: https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1374#issuecomment-5414318994

So your ruling stands unchanged for this PR — reciprocal kept, description untouched, no re-run. The only thing that changes is that a later flake is no longer the trigger for widening a description; it is already-written groundwork waiting on a slot.

Also on the diff since your review: `ab810c2f` takes your optional item 4 (gps-mentor), and the two table rows are corrected in the body.

EdmondOware (approved):
Approving. T-FEH's two rows are corrected, the `_009` and `_011` entries now match the evidence
rather than the first reading, and you took the optional `gps-mentor` fix as well, which was
the one piece of advice in the repo that this PR's own paragraph disproves.

The paging paragraph is the important part and it is right. "Never select a tier from a partial
set" is the sentence that matters: a proof summary resting on the first page of a larger set has
failed the reasonably exhaustive element and the correlation element at once, and it would fail
them silently, because `items` looks like a complete answer. Pinning `count` as the true total
and requiring `offset` until you hold all of them is the correct fix, and noting that
`{ section: "assertions", questionId }` has no narrower filter is what makes paging the only
remedy rather than a preference.

### On the ruling you asked for, from the genealogy rather than the cost

Keep the reciprocal, and keep it because it is right, not only because widening the description
is expensive.

`flynn-competing-fathers` carries `h_001` at **supported** and `h_002` at **active**, and
`q_001` is **in_progress**. Two hypotheses are live and the question has not reached a stopping
point. No confidence tier can be assigned in that state, because the competing hypothesis is
precisely what a tier would have to have ruled out. So proof-conclusion declining here is not a
routing nicety, it is the only defensible act, and a test demanding the decline pins doctrine
rather than convention.

Your worry about the description was still worth raising. "Summarize the evidence" does read as
a trigger, and the scenario does satisfy it on the surface. But the fault is in that phrase, not
in your test: summarizing evidence for and against competing hypotheses is hypothesis-tracking's
deliverable, while proof-conclusion summarizes evidence that has already resolved to a
conclusion. Two different acts sharing four words. Tightening the description is the durable fix
and it belongs on proof-conclusion's next slot, with your test standing meanwhile as the thing
that catches it if the boundary slips.

Nothing outstanding from me.

T-FEH (changes requested):
Second pass, on `0a0ec2cf` (your `f4c661f0` plus a clean merge of main). Gate green, no paid re-run owed.

**Both body rows are corrected, and more thoroughly than I asked.** The `_009` row now states the whole picture — that it does not fail on main, that `record-extraction` was the correct destination rather than a miss, that nothing routed at all, and that no valid baseline exists because #1832 edited the agent body after all five 08-21 logs and merged with no re-run. The `_011` row names the agent hashes (`ee03ebe8` vs the shipped `ea33e488`) and points at #1852. That is the honest version of both.

**And the gps-mentor fix landed correctly** — the optional item I left you. "Narrow the filter" is now "page with `offset` (50, then 100) until you hold `count` items", which matches `proof-conclusion.md`'s paragraph and the tool's real contract.

## [Blocking] The new validator is sound but never runs

`bf32937a` adds `test_open_candidate_blocks_closure`, and the logic is good. Supplying the tag directly, it does exactly what it should:

```
unchanged project              -> PASS
proof summary added for q_001  -> FAIL  "a proof summary was added for q_001 while h_002 is still `active`"
q_001 marked resolved          -> FAIL  "q_001 was marked resolved while h_002 is still `active`"
```

But it gates on `hypothesis-open-blocks-tier`, and **no test file in the repo carries that tag**. I replayed it against every proof-conclusion test:

```
of 22 proof-conclusion tests, 0 do NOT skip
```

So the commit titled "invariant validator for the open-candidate negative" ships a check that cannot run, let alone fail — and #1374's invariant is documented as guarded while nothing guards it. That is the shape this repo has been bitten by repeatedly.

**The fix is one line and it is free.** Add `hypothesis-open-blocks-tier` to `negative-hypothesis-tracking.json`'s tags. I verified end to end that this costs no paid re-run rather than assuming it: `_COSMETIC_TEST_FIELDS = ("name", "description", "tags")` in `snapshot.py` strips tags before hashing, and with the tag added the gate still printed **"All runlog rules satisfied."**

Worth checking as you do it: on the committed run `ut_proof_conclusion_021` routed away and wrote nothing, so the unchanged-project case above is the one that applies and it passes. Arming the tag should not red anything today.

If you deliberately deferred the tag — say so in the body and I will drop this — but as it stands the validator is inert and the PR reads as though it is not.

## Everything else

The eval gate is satisfied. Two warn-only notes, both inherited rather than yours: the judge prompt has moved on main since these runs were scored, and `ut_proof_conclusion_011` produced no graded dimensions (the deterministic validator failure you already documented on #1852).

Requesting changes on the one tag. Everything else in this PR is done.

EdmondOware (approved):
Re-approving against the current head. My approval of 25 August predates both the tag fix and the re-run, so it should not be read as covering work it never saw. T-FEH's block is theirs to clear.

**The tag is armed correctly.** `negative-hypothesis-tracking.json` is `ut_proof_conclusion_021`, and the rest of that commit is JSON reindentation. Counting real executions rather than recorded skips, since a skip is stored as `passed: true` with a `skipped:` error, `test_open_candidate_blocks_closure` now runs on exactly one test and passes, with twenty one legitimate skips. Nothing went red, as T-FEH predicted.

**The re-run was owed for the reason you gave, and I checked it rather than taking it.** The old log's snapshot hash for the agent body is `84e275e2`, byte identical to the body at `94292657` and different from this branch, and the log did not contain the new test. It measured the old agent. That is a stronger reason than the red implied and the body says so.

**The substance holds.** The validator carries a vacuous pass guard that fails the fixture outright if no hypothesis is `active`, which is the thing this repo keeps getting bitten by and is good to see written in. The scenario genuinely exercises it: `q_001` present, `h_002` active, `ps_001` already at `probable`, so the tier assertion is live rather than decorative. And the paging fix matches the tool contract exactly, which I verified against `research-query.ts`: `MAX_ITEMS = 50`, `count` is the true total, `truncated` is `count > offset + 50`. The old "no offset or pagination guessing" was simply wrong, and the genealogical cost of it is real, since a tier selected from a truncated assertion set can be a tier selected without the conflict that would have changed it. `research/SKILL.md` does page, so the spec and architecture claims about both halves having landed are accurate.

Two notes, neither blocking.

**The validator hardcodes `q_001`.** Its five siblings advertise that in their names, `test_q001_probable_tier` and the rest; this one does not. If a second `hypothesis-open-blocks-tier` scenario ever uses another question, all three assertions pass vacuously: the proof summary filters return empty so the count comparison is 0 <= 0, the tier check is skipped, and the question lookup returns `None`. The vacuous pass guard covers "no active hypothesis" but not "wrong question id". Either rename it to match the siblings or derive the id from the fixture.

**`ut_proof_conclusion_021` is not in the review sample**, so the one test whose validator this PR exists to arm has no human annotation on the run that ships. Corrections outside the sample are a bonus rather than a debt, so adding one costs nothing and no re-run.

benter-070:
@mercyokum — filed as #2022 rather than fixed here, per your second option.

The criterion you set is met: the fix is a proof-conclusion doctrine edit, which stales this skill's run log and buys its own paid run. This PR is already green with your approval and @EdmondOware's, so folding it in would cost a run and put two unrelated problems in one PR.

Three things I checked before filing:

**Your rate reading holds** — `_011` is 1 of 5 across the committed runs (fail only on `v1_2026-08-23_19-10-10`), and it passed again on today's re-run. Recorded in the issue, because four passes in five is the part that matters: the doctrine is being usually followed rather than binding, and the passing runs read as evidence the rule holds.

**Your "no open issue names this" holds, with one addition worth reading.** #1852 is ruled out as you said. #1823 (completion gate blind to 56% of conflicts) is adjacent — same underlying shape, that nothing stops a skill acting over an open conflict — but its scope is `research_append` refusing `project.status: "completed"`, a tool-level predicate. A proof *tier* over an open identity conflict is a different gate, and the ownership half is in neither. Cross-referenced both.

**Your "don't let it disappear" was closer than it looked.** `DEFAULT_KEEP_CANDIDATES = 5`, and your run log is now second-oldest of exactly five — my re-run today did not prune it, but the next proof-conclusion run takes `v1_2026-08-21_19-34-21` and the one after takes yours. So it had about two runs left. The issue carries the recovery command (`git show fab80c01:…`) so the evidence survives pruning either way.

One thing I put in the issue as an open question rather than deciding: whether this is doctrine or a writer-tool precondition. Both conditions are decidable from the project documents alone — is `c_001` open, does this writer own `conflicts` — which by ADR-0011's first question argues for the tool. Prose is what is already there, and it is what failed.

T-FEH (approved):
Third pass, on `a3bdffad` (your `b9dbdaa8` plus a clean merge of main). My blocker is closed, and closed with evidence I could read straight out of the run log.

## The validator is armed and demonstrably running

Not inferred from the tag — taken from the shipped run:

```
negative-hypothesis-tracking.json tags:  [… 'reciprocal', 'hypothesis-open-blocks-tier']
v1_2026-08-28_09-44-55 → ut_proof_conclusion_021
  test_open_candidate_blocks_closure  ran: True   passed: True
```

So it fires on the test it was written for, and passing is the correct outcome there — which is what I predicted when I proved arming it was safe. #1374's invariant now has a guard that actually executes.

## The fresh run finally has a valid baseline, and it is better than main

This was the hole under my round-1 finding: every earlier run predated #1832's edit to the agent body, so `_009` had no valid baseline at all. That is fixed, and I checked it by hash rather than by date:

```
agents/proof-conclusion.md on this branch:  ea33e48823bf7f41
snapshot recorded in the run log:           ea33e48823bf7f41   ← the run saw the PR's own edit
```

Against main's latest:

```
main    17 pass  3 partial  1 fail
this PR 19 pass  3 partial  0 fail

ut_proof_conclusion_012   fail    -> pass     (the "UNRESOLVED" one from round 1)
ut_proof_conclusion_r7k   partial -> pass
ut_proof_conclusion_021   (new)   -> pass
ut_proof_conclusion_007   pass    -> partial
```

And all three tests my round-1 review turned on now pass:

```
_009  pass   skills_invoked: ['record-extraction']   ← the correct destination fires
_011  pass
_012  pass
```

`_009` is the one I spent that round on — it had `skills_invoked: []` with nothing routing. It now routes correctly, on a run that genuinely saw this body. That closes the question rather than arguing it.

`_007` pass → partial is the one movement the other way. Single-run variance is the likely reading given everything else improved, and it is not worth another paid run — but it is the only thing in this run I would still call unexplained.

## The search-records edit is legitimate and accurate

`0b556e27` touches `eval/tests/unit/search-records/whitfield-1850-household.json`, which is PR #1946's file — but #1946 merged this morning and the file is on `main`, so this is editing main, not borrowing from an open branch. Mild scope creep for a PR about paging, and fine by the fix-it-here default.

I checked the claim rather than the provenance alone. The new text says #1967's fix is "reconciling `judge_context` and `rubric.md`", and that a tool-boundary content gate was rejected by the lead. That matches #1967's body, which carries a 2026-08-27 review block — *"premise was partly false; this is a stale-rewrite"* — and a ruling of *"Fix the eval defect and stop there… No content gate on `notes`."* So the correction is right.

One thing outside this PR, worth a moment from whoever owns the board: **#1967's title still describes the rejected approach** — "research_log_append: block an unhedged pre-1880-census relationship claim instead of relying on SKILL.md prose". The body has the ruling; the title contradicts it, and the title is what someone scanning the board sees.

## Still not done, and still free

The validator-test cases have not landed — `test_proof_conclusion_validator.py` has no case for `test_open_candidate_blocks_closure`. That file exists precisely because `testpaths = ["tests"]` means nothing under `validators/` is collected by `make harness-test`, and its directory is in no snapshot, so the seven cases from your commit message cost nothing. It matters less now that the validator demonstrably runs in a committed run log, so I am not blocking on it — but it is the difference between "it ran once" and "CI will notice if it stops".

## Approving

Three rounds, and each blocker was answered with a measurement rather than an argument. The paging edit and the gps-mentor fix are unchanged from where I approved them, the reciprocal pair is pinned from both sides, and the run that ships is the first one in this PR's history to have been graded against its own agent body.

T-FEH:
Three additions to my approval, plus one disclosure about my own conduct on this PR. None reverses the approval — all three are free and none is a correctness defect — but the first is a latent version of the problem this PR just fixed.

## The validator hardcodes `q_001` while its tag does not say so

`test_proof_conclusion.py:475` is `qid = "q_001"`, and every sibling in that file advertises the question id **in the tag**:

```
tier-probable-q001   tier-proved-q001   tier-possible-q001
hypothesis-open-blocks-tier      ← the new one, bare
```

So a second scenario tagged `hypothesis-open-blocks-tier` on a different question would pass all three assertions vacuously. The vacuity guard you wrote catches "no active hypothesis in the pre-state"; it does not catch "the active hypothesis belongs to a question this validator never looks at". That is the same shape as the dormant-tag problem from round 2 — a check that runs and cannot fail — just latent rather than live.

Cheapest fix is to follow the file's own convention: rename the tag to `hypothesis-open-blocks-tier-q001` in the validator and in `negative-hypothesis-tracking.json`. Tags are stripped by `_COSMETIC_TEST_FIELDS` before hashing, so that costs no re-run. Deriving the id from the fixture is the better long-term answer if you would rather.

## `ut_proof_conclusion_021` has no human annotation

The `.ann.json` covers `002, 012, 013, 016, r7k` — the harness's sample. The one test this PR exists to add and arm is not among them:

```
annotated: ['..._002', '..._012', '..._013', '..._016', '..._r7k']
021 annotated? False
```

Nothing requires it — rule 3 only owes the review sample — but out-of-sample corrections are accepted and owe no re-run, so annotating the test the PR is about is free and makes the record complete.

## "Both halves shipped" is stated over 2 of 7 consumers

The `docs/architecture.md` line names `research/SKILL.md` and `agents/proof-conclusion.md`. Seven plugin files call `research_query`; `person-evidence`, `search-records`, `question-selection` and `research-exhaustiveness` mention truncation nowhere. The standing "if you consume `research_query`, check `truncated`" directive still covers them, so this is a wording question rather than a gap — but declaring the direction closed removes the line that would tell the next reader four consumers are unhandled.

That sentence is one I wrote, which brings me to:

## Disclosure: part of this PR was authored by me

Commit `7ee7b806` in this diff — the `docs/architecture.md` and `research-query-tool-spec.md` edits — is mine. I pushed it to your branch during round 1 because those docs went stale the moment your change landed, and I disclosed it in that review at the time.

I still think the edit was right and you acknowledged it, but it does mean a reviewer wrote part of what merges here, and this repo's `/review` skill is explicitly read-only. Two consequences worth naming rather than leaving implicit: my approval covers text I authored, which is weaker than an independent one; and if the team wants reviewers to stop doing this, this PR is the precedent to point at. I am flagging it rather than quietly leaving it in the history.

## One thing I checked and can rule out

`_007`'s pass → partial is **not** the newly-armed validator. Every validator passed on that test, and `test_open_candidate_blocks_closure` correctly skipped there — it is a single judge dimension (`Narrative standalone`) at 2 with the other seven at 3. Single-run variance remains the reading, and I would not spend a run on it.

Issue #1705, opened 2026-08-17
Research-plan deep dive edits and changes


ClorindaM:
**Research-plan
Record-type-guide
Record Type Selection Guide
Record types by research goal** - table
**Track migration** - should include local/county/state histories, databases like Overland Travel

**FAN research** - land deeds and Probate - make sure the usefulness isn't limited to just the witnesses. This is a table so maybe change "witnesses" to "others mentioned" for brevity. If there is a place for explanation, it should include the idea that others mentioned in the will may not be immediate family but will provide FAN names as well. (Land deeds may have phrases like "bounded on the north by John Smith's property" or the probate files may name those who are owed money). 

**Emigrant origin or unindexed parish register** - Paterno Materno surname search is useful but even parish priests or clerks or reporters might confuse the order of the names. Must allow for the possibility that the names are reversed or confused but to not rely on the possibility.

DallanQ:
Merged into issue #1650 (deep dive: research-plan), with every finding carried over in full.

These are the same piece of work. Every item you listed lands in `research-plan/references/record-type-guide.md` — the goal table at line 9, Track migration at 18, FAN research at 19, the emigrant-origin section at 58 — and that file is inside the research-plan snapshot set, which is what issue #1650 owns. You had already posted the locality-survey-guide naming finding as a comment on #1650 two hours before filing this, so this is one dive reported in two places.

The reason to keep it on one card is cost, and it is #1650's own instruction: editing anything under the skill dir flips its run log inactive and buys one `make eval-skill SKILL=research-plan` run plus an annotation pass. Landing these findings separately from the rest of the dive pays that twice for one skill.

Nothing is lost — the FAN wording change and the reasoning behind it, the migration additions, and the paterno-materno caution are all now in #1650's body under a heading naming this issue. Keep adding findings there as you work.

One thing worth knowing before you touch `locality-survey-guide.md`: issue #1633 measured it as unreachable — no SKILL.md and no sibling reference file names it, so nothing loads it at all, and part of its research-log table maps to `research.json` fields that no longer exist. Rename versus delete should be settled against that issue.

PR #1871, opened 2026-08-24
research-plan: deep dive, 10 findings, 9 validator requests (#1650)

### LINKED ISSUE #1650: deep dive: research-plan
**IMPORTANT**: issue #1866 carries the eight validator requests filed from this dive, and two
things there need you.

- **The findings doc it cites is not pushed.** `docs/deep-dives/research-plan-findings-2026-08-24.md`
  on branch `1650-deep-dive-research-plan` — `git ls-remote origin refs/heads/1650*` shows the
  branch at main's HEAD with no deep-dive doc. Push it, or confirm on #1866 that its own body
  restates everything needed.
- **Land #1866's V2 before you pay for this skill's run.** Issue #1748 retired the per-skill tool
  deny on 2026-08-22, so on the next `make eval-skill SKILL=research-plan` this skill can reach
  the wiki tools its body tells it it does not have, and nothing gates the call. V2 is the check
  that catches it.

#1866 touches only `eval/harness/**`, so it costs no eval run of its own and does not compete
for this skill's snapshot.

> **Reviewed 2026-08-20 before junior handoff.** No decision needed. Two numbers
> in the body are corrected below, and one surface is fenced off — `rubric.md`
> belongs to two other open issues. The original follows under "Original issue".

**Touches:** packages/engine/plugin/skills/research-plan/SKILL.md, packages/engine/plugin/skills/research-plan/references/record-type-guide.md, eval/tests/unit/research-plan/, eval/runlogs/unit/research-plan/

## Corrected numbers

| | | |
|---|---|---|
| Tests in the suite | **21** (the body says 18) | `ls eval/tests/unit/research-plan/*.json \| wc -l` |
| Rubric+base dimensions that never discriminate | **6 of 8** | walk `tests[].runs[].judge.dimensions[]` in the newest run log, count `(source, name)` pairs with one distinct score |
| `judge_context` files naming a score branch | 0 | the grep in the body |

Measured 2026-08-20 against `eval/runlogs/unit/research-plan/v1_2026-08-17_17-52-29.json`
(21 tests, $6.31, 65 min). The five **rubric** dimensions — Record type selection,
Objective scope containment, Sequencing logic, Jurisdiction accuracy, Plan mode and
lifecycle — score 3 on every graded run in **all five** committed run logs, which is the
durable half; the sixth (base Completeness) moves in some runs and not others. Re-run the
walk rather than quoting the 6.

**Start from the newest run log.** `v1_2026-08-17_17-52-29` is active on skill-side files
— only three `eval/fixtures/mcp/` entries drifted, and the fixture arm of the run-log
gate is warn-only — so its transcripts reflect the SKILL.md and references you are
reading today.

## Do not touch `eval/tests/unit/research-plan/rubric.md` in this PR

Two open issues already own those five flat dimensions, and landing a rewrite here would
settle both by accident:

- **Issue #1404** (`genealogist`, `senior`, open since 2026-08-06) holds the
  tighten-the-bars-versus-accept-the-plateau call for exactly these dimensions, plus a
  second defect: `rubric.md` documents "Plan mode and lifecycle" as N/A for a first-ever
  plan on a question, and the judge scores it 3 on those tests instead of `null`. Its
  body says in as many words that this is "a senior call, not something to fix
  mechanically."
- **Issue #1668** is piloting the opposite remedy — an always-passes dimension gets
  **deleted**, not rewritten — on `question-selection` first, and names `research-plan`
  among "the other fifteen [that] follow." The pilot has not run.

So: **the `judge_context` half of "grading defects you own" is yours and uncontested —
fix those. `rubric.md` is not.** Instead, post what your transcript reading shows as a
comment on issue #1404: for each of the five, whether a real defect you found would have
been caught if that dimension could fail, and whether the base dimensions alone separated
a good run from a bad one. That is the evidence both issues are waiting on and only this
dive produces it.

Known cost of holding: if the pilot later green-lights deletion, research-plan buys a
second run to delete them. Committed runs for this skill are $5.85–$7.16 and 14–65 min.
Say so in your PR; do not pre-empt it.

## Validator requests go in one new issue

**All of them in one new `developer` issue**, not one issue per request and not buried in
the PR body. The worked precedent is issue #1749, "11 validator requests from the citation
deep dive." `research-plan` already has 8 validators in
`eval/harness/validators/test_research_plan.py` — read them first so you do not request
one that exists.

Before you write the requests, note the lead's ruling of 2026-08-19 (recorded on issue
#1749): a check over **structured run-log fields or persisted state** may gate a test; a
check that is a **regex over Claude's prose response** reports to the judge and never
gates. The deep-dive guide's "a literal phrase that must never appear" row is the second
kind — still worth requesting, but say which tier you are asking for.

## Annotations

Written **only** through the CRUD UI — `make eval-ui`. Hand-writing or editing a unit
`.ann.json` is forbidden. The gate wants a correction for every dimension of each test
named in the new run log's `review_sample` (5 per run).

## The orphaned reference file — contribute the content call, do not delete it

`research-plan/references/locality-survey-guide.md` is named for a locality survey it does
not contain; the heading inside reads "Research Log and Contingency Planning Guide" and
locality is never mentioned.

**Do not rename or delete it here.** Issue #1633 owns a per-file verdict across all 18
unreachable reference files plus the both-directions lint that lands with them, and it
independently measured this one as unreachable — no SKILL.md and no sibling reference
names it, and part of its research-log table maps to `research.json` fields that no longer
exist. What this dive should contribute is the content call, which only a genealogist can
make: comment on issue #1633 saying whether anything in that file is still true and worth
rewiring, or whether it is safe to delete outright. That costs nothing and unblocks the
18-file sweep.

## Original issue

**Touches:** eval/tests/unit/research-plan/, packages/engine/plugin/skills/research-plan/SKILL.md

Audit `research-plan` for defects nothing has caught. This starts from **no known

Did / Should / Gap.

## Start with this grep

```sh
grep -l -iE '"[^"]*\bscore [123]\b' eval/tests/unit/research-plan/*.json
```

without writing the finding.

## This skill's numbers

| | |
|---|---|
| Tests in the suite | 18 |
| Rubric+base dimensions that never discriminate | **6 of 8** |

6 of 8 dimensions never take more than one value, so they cannot report a defect in whatever they nominally grade.

## What to hand back

## Cost

Editing anything under `eval/tests/unit/research-plan/` or the skill body flips this
skill's run-log snapshot inactive and buys one `make eval-skill SKILL=research-plan`

---

# Findings so far — merged in from issue #1705 (2026-08-18)

Filed separately by ClorindaM while working this dive, and merged back in rather
than left as a second card: every item below lands in
`packages/engine/plugin/skills/research-plan/references/record-type-guide.md`,
which is inside this skill's snapshot set. Landing them on their own would buy a
second `make eval-skill SKILL=research-plan` run for work this dive already pays
for. Verified against the file 2026-08-18; line numbers are current as of that date.

## Record Type Selection Guide — "Record types by research goal" table (line 9)

**Track migration** (line 18). Primary column currently reads
*"Census (residence across decades), land records, tax records"*. Add local, county
and state histories, and databases such as Overland Travel.

**FAN research** (line 19). Currently reads *"Land deeds (witnesses), census
(neighbors), probate (witnesses), church (godparents)"*. Two changes:

- The usefulness of land deeds and probate is not limited to witnesses. Replace
  **"witnesses"** with **"others mentioned"** in both cells — shorter, and true.
- Where the guide has room for it outside the table, say why: people named in a
  will are often not immediate family but still supply FAN names, and deeds carry
  them in the boundary recitals ("bounded on the north by John Smith's property"),
  as probate files do in the lists of those owed money.

## Emigrant origin / unindexed parish register (line 58)

The paterno-materno surname search is useful, but parish priests, clerks and
reporters confuse the order of the two surnames. The guidance should allow for the
names being reversed or conflated **without instructing the agent to rely on that
possibility** — it widens the candidate set, it does not license a match.

## Also reported on this issue

`research-plan/references/locality-survey-guide.md` is named for a locality survey
it does not contain — the heading inside it reads "Research Log and Contingency
Planning Guide" and locality is never mentioned (comment, 2026-08-17). Note that
issue #1633 independently measured this file as **unreachable**: no SKILL.md and no
sibling reference names it, so nothing loads it, and part of its research-log table
maps to `research.json` fields that no longer exist. Settle rename-versus-delete
against that issue rather than in isolation.

### LINKED ISSUE #1866: 8 validator requests from the research-plan deep dive
> **Reviewed 2026-08-24 before junior handoff.** Corrected the validator-input citations, scoped V4, V6 and V8 out of this PR, and added the implementation order, cost-gate and rebase constraints.

**Touches:** eval/harness/validators/test_research_plan.py, eval/harness/tests/unit/test_research_plan_validator.py (new — 12 other validator files have one), eval/harness/harness/orchestrator.py (`_compute_outcome`, line 1147 — *not* runlog.py), eval/harness/harness/runlog.py (`aggregate_per_run_outcome`), eval/harness/provenance_report.py (stale docstring, see below), docs/specs/unit-test-spec.md (§7 Grading + § Deterministic Validators), eval/CLAUDE.md ("Reading an `aborted` row")

Filed from the `research-plan` deep dive (#1650). Grouped into one issue rather than eight, per the deep dive guide: "Do not open one issue per finding."

**Why this is filed rather than fixed in the deep dive PR.** The deep dive guide sets the division of labour explicitly: "You do not write the Python. You supply the genealogical rule, which is the half a developer cannot invent." Every item below is a Python validator, which needs a different skill and a different reviewer from the genealogist work in #1650. The genealogical rules are supplied here in full, with a reproducing example for each, so no further genealogist input is needed to implement them.

Full evidence for every item was to be `docs/deep-dives/research-plan-findings-2026-08-24.md` on branch `1650-deep-dive-research-plan`. **That file is not pushed** — `git ls-remote origin refs/heads/1650*` shows the branch at main's HEAD with no deep-dive doc. Everything needed to implement this is restated below; if you want the dive, ask on issue #1650 rather than waiting.

`research-plan` already has 8 validators in `eval/harness/validators/test_research_plan.py`. None of the requests below duplicates one. Read them first.

## Tiering

Per the lead's ruling of 2026-08-19 (recorded on #1749): a check over structured run log fields or persisted state may gate a test; a check that is a regex over Claude's prose response reports to the judge and never gates.

`plan_items[].rationale` is persisted state, not Claude's prose response, so V1, V3, V5 and V6 are proposed as gating. V4 reads `text_response` and is proposed as report only. V8 is the one open tiering question and is called out at the end.

## Before you start

**Order:** V7 first (it is what hid V2's evidence), then V1, V3, V5, V2.
**V4, V6 and V8 are out of scope for this PR** — see the note on each.

**Read the run's state the way a validator gets it.** Validators are handed
`before_state`, `after_state`, `tool_calls`, `attempted_mcp_calls`,
`skills_invoked`, `skill_frontmatter`, `test` — and nothing else
(`eval/harness/harness/validator_runner.py:50`). `runs[]`, `file_changes` and
`output.text_response` are run-log shapes, not validator inputs: compute new
plans by diffing before/after, exactly as `_new_plan_ids` in this file already
does.

**A tool response body is on the call entry, not behind its name.**
`response_fixture` is only the fixture's stem; the mock stores the served body
as `call["response"]` (`eval/harness/harness/mock_mcp.py:461`) and the runner
passes it through (contract: `eval/harness/validators/test_universal.py:22`).
The "a validator cannot see tool responses" note in
`eval/harness/provenance_report.py`'s docstring is stale — correcting that
sentence is part of this task, because the next person will read it and build
a fixture-file loader instead.

**A denied call never reaches `tool_calls`.** It lands in
`attempted_mcp_calls`, which the runner also supplies. `test_tool_allowlist`
learned this the hard way (issue #1748); V2 must union both.

**Do not touch `packages/engine/plugin/skills/research-plan/` or
`eval/tests/unit/research-plan/`.** Either arms the blocking run-log gate in
`eval/harness/scripts/check_runlogs.py`, which costs a fresh
`make eval-skill SKILL=research-plan` plus a genealogist annotation — the last
full suite recorded $6.31 and ~65 min wall clock (`totals` in
`eval/runlogs/unit/research-plan/v1_2026-08-17_17-52-29.json`). Changes confined
to `eval/harness/**` do not arm it, which is why this issue is free to land.

**Do not tag-gate these validators.** A tag nobody sets is how a validator runs,
passes, and cannot fail (issues #1755, #1788). Run them on every research-plan
test and `pytest.skip` only when the run added no new plan.

**What verifies this:** `make harness-test` (the validator unit tests), plus
`make provenance-report SKILL=research-plan` for V1 — offline, free, reads
committed run logs. Re-run it rather than quoting a count. **Nothing checks that
a new gating validator can fail**, so prove each one by hand-editing a copy of a
committed run entry until it reds (`CLAUDE.md`, "A new lint must be proven to
fail").

**Expect the next paid research-plan run to look worse.** V1 alone fails 7 of
21 runs in the latest suite, and a failed validator skips the judge, so those
runs carry zero graded dimensions. Intended, not a regression — say so in the PR.

**Rebase on PR #1833 first.** It touches `orchestrator.py`,
`validator_runner.py`, `validators/conftest.py` and `unit-test-spec.md`;
PR #1781 (issue #1749) touches three of the same four.

## V1. An identifier in a plan rationale must trace to a tool response from the same run

**Rule:** any collection or volume identifier written into a new `plan_items[].rationale` must appear in a tool response that same run received.

**Where to look:** every `collections_search` / `volume_search` /
`external_links_search` entry's `call["response"]`, against the `rationale` of
each item in the plans that appear in `after_state` and not in `before_state`.

**Why it is not judgment:** both sides are literal strings in the run log.

**What a violation looks like:** `ut_research_plan_006`, run `v1_2026-08-17_17-52-29`, item `pli_007` writes "FamilySearch collection 1401638 ('Pennsylvania, Church and Town Records, 1708-1985')". The only church collection that run was served has id `SYNTH-PA-CHURCH` and the title "Pennsylvania, Schuylkill County Church Records (synthetic test fixture)". The string `SYNTH-PA-CHURCH` appears zero times in the entire run log. The id `1401638` appears in 7 of the 21 runs and in no fixture any of them received. Six of those seven runs passed.

**Why it matters beyond the eval:** `search-records` reads the persisted rationale and will chase a collection the skill was never shown.

Note this is close in spirit to `make provenance-report` (issue #1667), which scans for persisted external identifiers no fixture supplied. That is a report over final state. This is a per run, per item check that can gate. Worth checking whether the two should share a helper.

That report already implements this rule corpus-wide and already finds this
defect (10 untraceable identifiers across 10 research-plan tests, 9 of them the
same `1401638`). Two things it measured are conditions on your validator: a
naive "every 4+ digit number must trace" rule produced 442 hits that were almost
all legitimate years, so match ARKs and 5+ digit numbers only; and issue #1667
closed with "triage, then decide what becomes blocking", so the gating version
here must be scoped to collection and volume identifiers written into a
`plan_items[].rationale` — not to every number in persisted state. Import the
identifier regexes from `provenance_report.py` rather than re-deriving them.

## V2. research-plan must call no MCP tool outside its six

**Rule (revised).** Fail the run when it calls *or attempts* `wiki_search`,
`wiki_place_page` or `place_population`, or when `skills_invoked` contains
`locality-guide`.

**Where to look:** `tool_calls[].tool` unioned with `attempted_mcp_calls[].tool`,
plus `skills_invoked`.

**Do not derive the check as the complement of the six.** `project_context` was
attempted by 3 of the 21 runs in `v1_2026-08-17_17-52-29` (001, 013, 016) and
six sibling skills declare it, so a complement gate reds three runs on a call
nothing forbids. The three named tools are forbidden by name in the body
("`locality-guide` … owns `wiki_search`, `wiki_place_page`, `place_population`",
SKILL.md 58-60; "You have no wiki/place-fact tools of your own", line 137), and
the `locality-guide` delegation is forbidden twice (lines 128 and 499) — issue
#1862 confirms research-plan is the correct side of that contradiction. Deriving
a deny from a grant is what PR #1774 retired on 2026-08-22; a named prohibition
is a different thing, and saying which is which belongs in
`docs/specs/unit-test-spec.md` § "Deterministic Validators" as part of this task.

**Why it is not judgment:** the six are enumerated in the skill's `allowed-tools:` frontmatter and again in the SKILL.md "MCP tools used" table, and the body states "You have no wiki/place-fact tools of your own."

**What a violation looks like:** `ut_research_plan_016`, run `v1_2026-08-17_17-52-29`, declared itself the orchestrator, invoked `locality-guide`, and attempted a wiki backed locality survey. Its own transcript says "The wiki tools were denied."

**Why this is urgent rather than historical:** that deny is gone. Commit `30e786c3` (#1774, merged 2026-08-22) retired the per skill tool deny, correctly, since production grants every tool. `compute_allowed_tools` is now advisory and `test_tool_allowlist` warns rather than asserts. On the next `make eval-skill SKILL=research-plan` the wiki tools are available to this skill and nothing gates the call.

## V3. `fallback_for` must name an item in the same plan

**Rule:** a new plan item's `fallback_for`, when non null, must be the `id` of another item in the plan being written.

**Where to look:** `file_changes.research.json.diff.plans.added[].items[]`.

**Why it is not judgment:** referential integrity over two id fields.

**What a violation looks like:** `ut_research_plan_002`, run `v1_2026-08-17_17-52-29`, created `pl_003` with items `pli_007` through `pli_014`. Its single `fallback_for` is `pli_013` pointing at `pli_006`, which is the third item of `pl_002`, a plan whose status is `completed`. So the new plan carries no internal fallback chain at all, and the one it declares resolves to an item that is already finished. The run passed with all dimensions at 3.

## V4. A review mode run that names a collection must have looked one up (report only)

**Out of scope for this PR — the tier it needs does not exist.** Validators are
not supplied `text_response` (`validator_runner.py:50`), and the report-only
channel itself is "not yet built", tracked as Group M of issue #1749 — PR #1781
says exactly that in the spec text it adds. PR #1833 (open) adds the
`text_response` argument. An "advisory" validator today calls `warnings.warn()`,
which writes no run-log entry and reaches no judge (PR #1774's own summary), so
there is nothing to report *to* yet. Leave V4 recorded here; do not build a
second channel, and do not let it gate.

## V5. An availability claim must match the returned counts

**Rule:** a plan item rationale that calls a named collection "indexed" or "fully indexed" must refer to a collection whose returned `personCount` is greater than zero. One that calls it "browse only" or "image only" must refer to one whose `personCount` is zero.

**Where to look:** the item `rationale` against the `collections_search` response the run received.

**Why it is not judgment:** `personCount` is a number in the tool response and the adjectives are a closed pair.

**What a violation looks like:** `ut_research_plan_001` calls the Schuylkill probate collection "indexed". The fixture returns `personCount: 0`, so it is image only. Several runs get this right, which is what makes the pair checkable: `ut_research_plan_005` item `pli_008` writes "this collection is image-only (personCount: 0 in collections_search)."

**The cited example is not a V5 violation — do not use it as the test case.**
`ut_research_plan_001` made zero tool calls and wrote no plan items, so no
returned count exists for it; that run is V4's evidence. The real instances are
in the same log: `ut_research_plan_q7m` `pli_007` calls collection 1999196
"indexed" against a served `personCount: 0`, and `ut_research_plan_014` `pli_008`
does the same.

**Bind the adjective to the identifier, not to the rationale.** A rationale
routinely names two sources — `ut_research_plan_006` `pli_011` describes a
browse-only *volume* and an *indexed collection* in one paragraph, and a
whole-rationale word scan reds it. Match adjective and identifier within one
sentence and skip any sentence naming more than one collection. `personCount` is
a property of the response, not of the id: 1999196 comes back as 893,214 in
`collections-search-pennsylvania` and 0 in `collections-search-schuylkill`, so
key on the response this run received.

## V6. A rationale that says "fallback for pli_X" must set `fallback_for` to `pli_X`

**Dropped: zero instances.** Across the 95 new plan items in
`v1_2026-08-17_17-52-29`, 9 rationales carry "fallback for pli_NNN" and every
one matches its own `fallback_for`; single-target mismatches: 0. The cited
example passes the stated rule — `ut_research_plan_006` `pli_010` says "Fallback
for pli_007" and holds `fallback_for: pli_007`. Its second target (pli_008) is
the multi-target case this issue defers, so implementing "the single target half
only" produces a check that cannot fail. Re-derive before reinstating: walk
`plans` added between before/after state in the newest run log and compare
`re.findall(r"fallback for (pli_\d+)", rationale, re.I)` against `fallback_for`.

## V7. A validator failure must not surface as an abort

This one is a harness reporting change rather than a validator, and it is a `nothing-checks` item. It is grouped here because it is what hid V2's evidence.

**Rule:** a run whose validators failed must not be recorded with `outcome: "aborted"`. A validator failure should dominate a wall clock or turn cap abort in the recorded outcome.

**Where to change it:** `_compute_outcome`,
`eval/harness/harness/orchestrator.py:1147`. `runlog.py` has no such function;
its `aggregate_per_run_outcome` (line 162) is the second site and must agree.

**Why it is not judgment:** two recorded fields.

**What it looks like today:** `ut_research_plan_016`, run `v1_2026-08-17_17-52-29`, failed two validators. `test_ownership_table` reported "research-plan modified sections it doesn't own: ['localities']" and `test_research_plan_new_plan_for_q_001` reported "expected exactly one new plan; got []". The recorded outcome is `aborted` with `aborted_reason: max_wall_clock_seconds`, the judge graded zero dimensions, and the test was not in that run's `review_sample`. Every reader of that run log sees a timeout rather than a skill that wrote a section it does not own and produced no plan.

Note that `eval/CLAUDE.md` already tells readers "Reading an `aborted` row: `aborted` is not a failed test, it produced no gradeable result, so nothing about it says anything about the skill." That guidance is correct in general and wrong in this case, which is precisely the problem.

**Which aborts it dominates: the three caps only** — `max_wall_clock_seconds`,
`max_turns`, `max_tool_calls`. Leave `unmatched_tool_call` (a test-corpus
problem the suite exits 2 on) and the two transient reasons (`error`,
`sdk_stream_silence`) recorded as `aborted`: the exit-code split and the suite
breaker (`_TRANSIENT_ABORT_REASONS`, `eval/harness/run_tests.py:306`) both read
them, and demoting them to `fail` reports an environment failure as twenty skill
regressions. `not_runnable` never reaches this function (`_aborted_entry`,
orchestrator.py:1607). `aborted_reason` stays populated either way — only
`outcome` changes.

**Also update:** the grading contract in `docs/specs/unit-test-spec.md` §7, and
the "Reading an `aborted` row" paragraph at `eval/CLAUDE.md:175`, which tells
every reader that an aborted row says nothing about the skill. Check
`eval/harness/tests/unit/test_orchestrator.py:1101` still holds — its case
aborts on `unmatched_tool_call` with passing validators, so it should be
unaffected, and that is worth asserting rather than assuming.

## V8. A parentage, identity or undated event plan must carry a FAN item

This one carries an open decision and should not be implemented before the lead rules on it.

**Rule:** a plan written for a parentage, identity or undated event question must contain at least one plan item whose deliverable is a record of a relative, neighbour or associate rather than of the subject, with the FAN connection stated in its `rationale`.

**Where to look:** `questions[]` in the before state for the question type, and the new plan's `items[]` rationales.

**Why it is not judgment, with a caveat:** "does this item target someone other than the subject" is decidable. Which question type this is, is not. `questions[]` carries no type field. Its properties are `id`, `question`, `rationale`, `selection_basis`, `priority`, `status`, `depends_on`, `unblocks`, `created`, `resolved`, `resolution_assertion_ids`, `exhaustive_declaration`. `selection_basis` records why a question was picked, not what it asks: six of the nine scenarios sampled are `objective_decomposition`, covering parentage, probate source, marriage place and death date questions alike, so it cannot discriminate.

**What a violation looks like:** `ut_research_plan_005`, run `v1_2026-08-17_17-52-29`, wrote nine items on a parentage question whose mother is entirely unidentified, and not one targets a sibling or associate. Same shape in `qxr`, `fbn` and `010`. Four of the twelve plans written in that run carry no FAN item, and all four scored 3 on every rubric dimension.

**Deferred, and no ruling is needed yet.** Option 1 needs the same report-only
channel V4 does, and it does not exist. Option 2 is a `research.json` schema
change owned by `question-selection`. Leave V8 recorded here and raise it when
the reporting tier lands. One supporting measurement is stronger than the body
says: across every committed scenario, `selection_basis` is
`objective_decomposition` 124 times out of 130
(`grep -rho '"selection_basis": "[a-z_]*"' eval/fixtures/scenarios/*/research.json | sort | uniq -c`),
so it cannot discriminate question type.

## Related

- #1650, the deep dive that produced these.
- #1749, the precedent: 11 validator requests from the citation deep dive.
- #1667, `make provenance-report`, which overlaps V1 in spirit.
- #1774, which retired the per skill tool deny and is why V2 is urgent.

EdmondOware:
Closes #1650.

Deep dive on `research-plan` per `docs/skill-deep-dive-guide.md`. Starts from no known failure.

**10 findings, 9 validator requests.** Full evidence with quotes in `docs/deep-dives/research-plan-findings-2026-08-24.md`. The Step 1 prohibition list, 61 checkable rules, is saved at `docs/deep-dives/research-plan-prohibition-list.md` for the next auditor.

## What is fixed here

- The 7 `judge_context` files leaking a score branch in the shape `score <Dimension> as 1 (fail)`. The issue's prescribed grep returns 0 on this suite because a dimension name sits between "score" and the digit. Each verdict instruction is removed, the substantive requirement kept, and the dimension still named.
- Two execution caps that were hiding real results behind aborts. `martha-remarriage-surname-plan` wall clock 600 to 900, and `plan-continue-authorized-in-message` gains an explicit `max_turns: 30` where it had been inheriting the default 20 and aborting on it in three consecutive runs.
- The FAN contradiction between `SKILL.md` and `planning-standards.md`. Resolved by scoping the mandate to parentage, identity and undated event questions.
- Three record type guide corrections merged in from #1705.

## What is filed rather than fixed

#1866, 9 validator requests. Filed because the guide sets the division of labour explicitly: "You do not write the Python. You supply the genealogical rule, which is the half a developer cannot invent." Every item needs a different skill and a different reviewer.

## ut_research_plan_016 ships red, and q7m is green by luck

The `search-external-sites` stub closed `q7m` in run 3: it routed to the two stubbed executors and passed, and it passed again in run 4. `search-images` is still unstubbed and run 2 reached it, so `q7m` is a latent flake rather than a fix. F8's reproducing evidence lives in the stale run logs `13-37-16` and `14-32-36`.

The red test in the shipping run log, `v1_2026-08-25_21-02-48`, is `ut_research_plan_016`. It fails two validators there: `test_ownership_table` on `['localities']` and `test_research_plan_new_plan_for_q_001` with "expected exactly one new plan; got []" — the same pair F2 records from the 08-17 run. **That is F2, and it is a live skill defect rather than a validator false positive.** The skill performed locality-guide's survey and called `wiki_search`, `wiki_place_page` four times and `place_population`, three tools rule 8 says it does not hold, then persisted a `loc_` entry. #1774 retired the per-skill deny on 2026-08-22 and run 3 was the first full run since. V2 on #1866 is the check. Nothing here fixes it.

On F8 itself: closing it at the fixture level would mean stubbing every skill that can write `log`, which is the grading patch over a harness gap that `eval/harness/harness/skill_stubs.py` says the canned response form exists to remove. V9 on #1866 is the real fix, and the validator already concedes this exact class in its own docstring, for negative tests.

## This PR contains a retraction

After the first verification run the `judge_context` rewrite appeared to restore discrimination on two rubric dimensions that had never taken another value in 325 graded cells. Runs 2 and 3 did not reproduce it, and one of the two movements was overturned in annotation as a judge error. All four runs are recorded and the findings doc states the retraction plainly. **Do not cite run 1 alone.**

## The retraction was re-measured under the new judge prompt, and it holds

Runs 1 to 3 carry `judge_prompt_hash` `0d186137147c`. #1766 replaced the prompt with `c39d70034788`, which makes `judge_context` notes binding on the scores rather than background and says an unmet requirement is a deduction in whichever dimension covers it — the mechanism this retraction and F9 both describe as absent. That is the one condition under which the flatness might have been a grading artifact rather than a property of the dimensions, so run 4 measured it instead of assuming.

`v1_2026-08-25_21-02-48` is that measurement: 21 tests, 19 pass, scored under `c39d70034788`. Every one of the five rubric dimensions is **3 across all 15 gradings**, exactly as under the old prompt. Binding the per-test clauses to the scores moved nothing.

**That strengthens #1404's case rather than weakening it** — the dimensions stayed flat in the conditions most favourable to their discriminating. Posted on #1404 alongside the original measurement and the caveat that prompted the re-run.

The only dimension that moved at all is base `Tool Arguments`, one 2 on `ut_research_plan_wzk`, from `place_search({placeName: "Georgia", contextName: "United States"})` finding no fixture. That is a corpus gap rather than a skill defect — the same call shape matched for Blount County and Smith County in the same run — and it is corrected to 3 in the annotation.

## F9 is worth a lead's eye

No rubric dimension in this skill grades breadth. All five inspect only what a plan contains, never whether it contains enough, so a plan can omit the single most probative record type for its own question and still score a clean pass. `ut_research_plan_005` is the worked case: an eight item parentage plan whose only census is 1860, so it never reaches the 1880 relationship column, and Record type selection scored 3 correctly by the letter of its own bar.

That is a mechanical explanation for the plateau #1404 has held since 2026-08-06, and it is posted there.

## Cost and fences

**Four runs, $26.53:** `13-37-16` ($6.60), `14-32-36` ($6.56), `17-21-19` ($6.77) and `21-02-48` ($6.61). The third was bought by the first review round and the fourth by the second, for the re-measure above.

**Only run 4 is current.** Runs 1 and 2 predate the `SKILL.md`, `record-type-guide` and four `judge_context` edits; run 3 has since drifted on five `eval/fixtures/mcp/wiki-*` files that main changed, and was scored under the superseded judge prompt. Read runs 1 to 3 for F8's evidence and the run-to-run comparison, not as a description of what the skill does now.

Aborts went from 2 to 0 and stayed there across all four runs. `wzk`, the married surname guard, now produces a graded result every run — pass in runs 2 and 3, partial in 1 and 4 — so F3's never-graded state is resolved.

`rubric.md` is untouched, per the fence #1650 placed around it for #1404 and #1668. Known cost of that fence, stated as the issue asked: if #1668's pilot later green lights deleting the five flat dimensions, `research-plan` buys another run to delete them.

Comments posted with this dive's contributions: #1404 (the five flat dimensions with F9, the old-prompt caveat, and the re-measure), #1633 (delete verdict on both unreachable reference files), #1866 (V9, a correction to V1, and the gating tier consequence).

One caveat on the #1633 comment: that issue closed as `not planned` on 2026-08-21, three days before the verdict landed, so the verdict currently has no owner and both files still ship. The findings doc records that a future auditor has to re-file the sweep or carry it forward.

One leak survives F7's widened grep and is **not** fixed here: `plan-continue-authorized-in-message.json:40` reads "Score Correctness low ONLY for this pattern". It could have ridden run 4 free and did not, so removing it now flips the snapshot inactive and buys a fifth run for one sentence. It should ride the next paid run this skill takes.


mercyokum (changes requested):
Reviewed the eval fixture edits, the two new deep-dive docs, and the `SKILL.md`/reference-file changes. The core fixes (cap raises, de-leaking `judge_context`, scoping the FAN mandate) are sound and well-documented. Found a handful of follow-on issues worth fixing before merge, plus a couple of smaller cleanups.

## Should fix before merge

1. `eval/tests/unit/research-plan/plan-continue-authorized-in-message.json` — the new `search-external-sites` stub entry isn't covered by `judge_context`'s "don't grade the canned content" instruction, which still only names `search-records`. Since F8 in the findings doc shows this test can route to either skill nondeterministically, the judge may end up grading wording the model never actually wrote. Worth adding a matching exclusion line, following the pattern `search-records/escalate-to-external-after-fs-exhaustion.json` already uses.

2. `packages/engine/plugin/skills/research-plan/SKILL.md` (Step 4, item 6) — the bolded heading ("Include FAN items where they earn their place") still reads as fully discretionary, but the sentence right after it now makes FAN items mandatory for parentage/identity/undated-event questions. Step 3's own FAN guidance (untouched by this PR) also still states the old unconditional rule with no pointer to the new mandatory case. A reader skimming headings, or one who settles the FAN question at Step 3, could miss the new requirement entirely. Suggest tightening the heading and adding a one-line cross-reference from Step 3.

3. `docs/deep-dives/research-plan-prohibition-list.md` — three separate accuracy issues in this one file:
   - Line 8: "contribute no rule; see finding `F-REF`" — there is no `F-REF` in the findings doc. The finding this refers to (the two unreachable reference files) is `F6`. Should read "see finding F6".
   - Line ~204, "Known internal contradiction" section: still describes the FAN rule conflict as open and unenforceable, but this same PR resolves it (the findings doc says "Resolved 2026-08-24" and both `SKILL.md`/`planning-standards.md` now agree). Since this file is explicitly meant to be where "the next auditor starts," leaving it stale risks someone re-litigating a settled question.
   - Same section cites "Rule 51" for the "Not a quota" wording, but rule 51 as actually numbered in this document is a different rule about FAN purpose — the quote is really the old Step 4 text. Worth fixing the rule number too.

4. `docs/deep-dives/research-plan-findings-2026-08-24.md:724` — still carries an open placeholder: `*(Genealogist's verdict to be appended before this PR opens.)*`. Since this doc says "Save this file. The next auditor of research-plan starts here," a stale TODO undercuts that. Suggest replacing it with the actual verdict, e.g.: "Genealogist's verdict: see the 2026-08-24 comment on issue [#1633](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1633) — both files (`locality-survey-guide.md`, `validation-protocol.md`) are safe to delete outright."

## Worth a look, lower priority

5. Same findings doc — the "6 of 8 flat dimensions ✅" verdict near the top only reaches 6 by counting base `Completeness` as flat, but that same row's data (`3×89, 1×1`) shows two distinct values. The rest of the doc consistently refers to "the five flat dimensions," so this top-line claim contradicts the doc's own later usage.

6. Same doc — "Both cap fixes worked, and the aborts are gone" implies two caps were raised, but only `martha-remarriage-surname-plan.json`'s wall clock was actually changed; `plan-death-premise-includes-probate.json` got only a `judge_context` wording edit. The doc's own later "F2 survives in a milder form" note is more accurate — worth tightening the earlier sentence so a skim-reader doesn't walk away with the wrong impression.

7. Same doc, F-FAN section — the claim that the 2026-08-24 ruling "closes" all four motivating failures (`qxr`, `fbn`, `005`, `010`) doesn't quite hold for `fbn`, which is a marriage-place/jurisdiction question (elsewhere in the same doc it's called the "Jurisdiction accuracy" test), not parentage/identity/undated-event. Worth either narrowing the claim or explaining why `fbn` counts.

8. `references/record-type-guide.md` — dropping "(witnesses)" from the probate row in favor of "(others mentioned)" leaves it unclear whether a will's attesting witnesses are still a valid FAN source for probate. The parallel sentence for land deeds explicitly says "not only the witnesses," but the probate sentence never confirms witnesses are still included — worth adding the same one-line reassurance.

9. Fixture/rule drift: `caroline-probate-creator-lifespan.json` (a parentage question) still has unedited `judge_context` saying breadth items, including FAN, are "not individually required" — which now sits at odds with the new mandatory-FAN rule for parentage questions. Not urgent since nothing gates on it yet, but worth a note or a follow-up so it doesn't get missed when V8 lands.

10. Three files (`caroline-probate-creator-lifespan.json`, `fallback-sequencing-plan.json`, `feliciana-parish-boundary-split.json`) replace a "score X as 1 (fail)" leak with a sentence that just restates an adjacent, unedited bullet already in the same `judge_context` array, rather than merging the two. `martha-remarriage-surname-plan.json` handles this correctly elsewhere in the same PR by editing the existing bullet in place — worth doing the same here to avoid permanently duplicated judge instructions.

mercyokum (approved):

promise-emmanuel (changes requested):
1. The PR body names the wrong red test
In v1_2026-08-24_17-21-19 — the only run log whose snapshot still matches the tree — ut_research_plan_q7m passes. The test that fails is ut_research_plan_016, on test_ownership_table: research-plan modified sections it doesn't own: ['localities'].

That is F2, and it is a live skill defect rather than a validator false positive. The skill ran wiki_search, wiki_place_page ×4 and place_population — three tools rule 8 says it does not hold — and persisted a loc_ entry.

Your findings doc gets this right at F8. Only the PR body and F8's closing line still carry the old claim.


You wrote:    ## ut_research_plan_q7m ships red, deliberately
Change it to: ## ut_research_plan_016 ships red — and q7m is green by luck

You wrote:    It is F8's evidence. `test_ownership_table` attributes a delegated sub
              skill's writes to the caller. Reproduced twice through two different sub
              skills, `search-external-sites` in run 1 and `search-images` in run 2,
              because which executor the skill hands off to varies with the plan it writes.
Change it to: The `search-external-sites` stub closed q7m in run 3 — it routed to the two
              stubbed executors and passed. `search-images` is still unstubbed and run 2
              reached it, so q7m is now a latent flake, not a fix. F8's reproducing
              evidence lives in the two stale run logs, 13-37-16 and 14-32-36.

              The red test in the shipping run log is `ut_research_plan_016`, on
              `test_ownership_table: ['localities']`. That is F2, and it is a live skill
              defect: the skill performed locality-guide's survey and called three MCP
              tools rule 8 says it does not hold. #1774 retired the per-skill deny on
              2026-08-22 and this is the first run since. V2 on #1866 is the check.
              Nothing here fixes it.
And in the findings doc, F8's last line contradicts F8's own paragraph four lines above it:


You wrote:    **The test ships red, deliberately, as V9's reproducing evidence.**
Change it to: **V9's reproducing evidence is runs 1 and 2, not the shipping run log.**
2. The cost is understated by one run and $6.77
Three run logs dated 2026-08-24 are committed: $6.6029 + $6.5555 + $6.7654 = $19.92. #1650 asked the dive to state its cost, and the next slot gets budgeted off that number.


You wrote:    Two runs, $13.16.
Change it to: Three runs, $19.92. The third was bought by the review round, and it is the
              only one whose snapshot is still current.
The doc says "three verification runs" in one place and "two" in another. Same fix:


You wrote:    ## What the two verification runs showed

              The PR's changes bought two `make eval-skill SKILL=research-plan` runs, both on
              2026-08-24: `v1_2026-08-24_13-37-16` ($6.60) and `v1_2026-08-24_14-32-36`
              ($6.56). The second was bought by a fixture fix, not by a finding; see F8.
Change it to: ## What the three verification runs showed

              The PR's changes bought three `make eval-skill SKILL=research-plan` runs, all on
              2026-08-24: `v1_2026-08-24_13-37-16` ($6.60), `v1_2026-08-24_14-32-36` ($6.56)
              and `v1_2026-08-24_17-21-19` ($6.77) — $19.92 in total. The second was bought by
              a fixture fix, not by a finding (see F8); the third by the review round. Only the
              third is still active against the tree: the first two predate the SKILL.md,
              record-type-guide and four judge_context edits and are stale on six snapshot keys.
3. The #1633 contribution rests on a false claim, and lands on a closed issue
#1633's own 18-file table already lists 659  research-plan/references/validation-protocol.md. It did count it.

And #1633 closed as not planned on 2026-08-21, three days before your comment on 2026-08-24. There is no 18-file sweep left to unblock, and both files still ship in the plugin zip. The next auditor needs to know that.


You wrote:    The content call on `references/locality-survey-guide.md` — and one file #1633
              may not have counted, `references/validation-protocol.md`, which is unreachable
              by the same test
Change it to: The content call on both files #1633 lists for this skill —
              `references/locality-survey-guide.md` and `references/validation-protocol.md`

You wrote:    **Genealogist's verdict, 2026-08-24: both files are safe to delete outright, and
              nothing in either needs rewiring.** Posted in full on issue #1633.
Change it to: **Genealogist's verdict, 2026-08-24: both files are safe to delete outright, and
              nothing in either needs rewiring.** Posted in full on issue #1633 — which closed
              as not planned on 2026-08-21, so the verdict has no owner and both files still
              ship. The next auditor needs to re-file the sweep or carry the verdict forward.
4. F7's replacement grep still misses a leak in this suite
Your widened grep returns 0 on the fixed suite. But plan-continue-authorized-in-message.json:40 still reads "Score Correctness low ONLY for this pattern." That names the dimension, the verdict and the condition — F7's own definition of the leak. The pattern misses it twice: it is case-sensitive on score, and it requires as [123].

grep -lniE 'score [A-Za-z][A-Za-z ]+ (as [123]|low|high)' returns exactly that one file.

Also, neither destination you name holds a grep to widen. docs/skill-deep-dive-guide.md has none, and there is no .github/ISSUE_TEMPLATE/ — the pattern lives only in #1650's body.


You wrote:    grep -lE 'score [A-Za-z][A-Za-z ]+ as [123]' eval/tests/unit/research-plan/*.json
Change it to: grep -lniE 'score [A-Za-z][A-Za-z ]+ (as [123]|low|high)' eval/tests/unit/research-plan/*.json

You wrote:    Recommended change to `docs/skill-deep-dive-guide.md` and to the issue
              template: widen the prescribed grep.
Change it to: Recommended: add the grep to `docs/skill-deep-dive-guide.md` Step 1, which
              carries none today (there is no issue template — the pattern lives only in
              #1650's body). The widened form still leaves one live instance,
              `plan-continue-authorized-in-message.json:40` "Score Correctness low ONLY for
              this pattern"; removing it flips this skill's snapshot inactive and buys a
              ~$6.6 run, so it should ride the next paid run rather than this PR.

EdmondOware:
All four fixed in 456de3a2. Every one verified against the repo first, since three of them were claims I had made without checking.

1. The red test. Corrected in the PR body and in F8's closing line. You are right that these are opposite things and the body was advertising the wrong one: q7m red was F8, a validator false positive; 016 red is F2, a live skill defect. The body now leads with 016, names the three tools rule 8 says the skill does not hold, and says plainly that nothing in this PR fixes it and V2 is the check. q7m is described as a latent flake rather than a fix, with F8's evidence pointed at the two stale run logs.

2. The cost. $19.92 across three runs, in both places. The findings doc heading said "two verification runs" while the table below it said three. It now also records that only 17-21-19 is still active against the tree, and that the first two predate the SKILL.md, record type guide and four judge_context edits.

3. #1633. Both claims were wrong and both were one command away from being checked. Its body names validation-protocol.md twelve times, so it did count it. And it closed as not planned on 2026-08-21, three days before the verdict landed, so there is no sweep to unblock, the verdict has no owner, and both files still ship. Corrected in both places, and I left the retraction visible in the doc rather than quietly rewriting the sentence, since the next auditor needs to know the verdict is stranded and has to be filed again or carried forward.

4. The grep. Widened to your form. You are also right that neither destination I named exists: docs/skill-deep-dive-guide.md carries no grep and there is no issue template, so the recommendation is now to add one to Step 1 rather than to change one.

The surviving leak is documented and deliberately not fixed. plan-continue-authorized-in-message.json:40 is a snapshot file, so removing that clause flips the run log inactive and buys a fourth run at about $6.6 for one sentence. It is recorded in F7 to ride the next paid run this skill takes.

Only docs and the PR body changed, so no run was bought and the run log is still active.

promise-emmanuel (changes requested):
Line 309 of docs/deep-dives/research-plan-findings-2026-08-24.md still prints the widened pattern under the sentence that says it returns seven files. I re-ran both against the new merge-base, since main moved 56 commits, and the narrow pattern returns seven while the widened one returns eight — the eighth being plan-continue-authorized-in-message.json, which isn't in the table. Three things follow from that one line: the stated count is wrong, line 337 now says "the pattern above is case sensitive on score" about a pattern that is -i and so refutes itself, and the same 78 characters are printed twice, at 309 and 346. The fix is to put the narrow pattern back at 309 — grep -lE 'score [A-Za-z][A-Za-z ]+ as [123]' eval/tests/unit/research-plan/*.json — and leave 346 exactly as it is, since that is where the widened form belongs and the paragraph around it reads correctly once the narrow one is above it again.

One thing changed status with the merge. The judge-prompt drift I flagged last time as a merge artifact is now the PR's own: #1766 is inside this branch, so the newest run log's prompt hash 0d186137147c no longer matches the branch's c39d70034788, and the runlogs gate warns on this head. It is warn-only and runlogs passes, but the caveat is no longer bookkeeping, because the new prompt says judge_context notes are binding on the scores the judge gives and that an unmet requirement is a deduction in whichever dimension covers it. That is the exact mechanism whose absence the retraction and F9 describe, and both were measured under the old prompt. Worth reading both ways — it plausibly makes your rewritten clauses more effective than the runs show, not less, since "whether a plan item targets the parents' marriage is what Record type selection turns on for this test" is binding now where it was background before. So the retraction may be pessimistic. Either way it is a caveat rather than a fourth run: add a line to the retraction section and to the #1404 comment saying all three runs were scored against 0d186137147c, that #1766 replaced it with c39d70034788, and that these numbers are a baseline under the old prompt rather than a current reading — re-measure before deleting a dimension.

Nothing else broke on the merge, and I checked the gates main added since a merge changes what ships. The new per-PR schema validation of every committed unit test passes on your eight edited fixtures. check_tool_coverage raises nothing for research-plan. check_negative_reciprocity is warn-only with 44 pre-existing edges, none of them yours. And the new grade_on_invariant reachability rule in unit-test-spec.md doesn't bite here — ut_research_plan_003 is this suite's only such test, and its validators are all state-based rather than asserting a tool was never called. On this head I get engine vitest 2700 of 2700, tsc clean, harness pytest 2739 passed with 2 skipped, server pytest 209 passed, and the runlog gate satisfied.

Push the grep fix and the caveat and I'll approve.

EdmondOware:
Both asks are in, and the second one turned into a real measurement rather than a caveat.

**The grep.** Restored the narrow pattern at the first site in `1f4207b4` and left the widened
form at the second exactly as you said. All three consequences fall out of the one line: the
count matches again, the duplication is gone, and the "case sensitive on `score`" sentence is
true now that the pattern above it has no `-i`.

You were right and my first reading was wrong, which is worth recording. I ran both patterns
and got 0 and 1, and nearly came back arguing the fix was bad. The reason is that they only
return 7 and 8 against the merge base `8981ea9d`. On the branch they return 0 and 1 because
this PR removes those clauses from the seven files the table lists. A findings doc records the
leak as found, not as left, so the base is the right tree and the doc was correct all along.

**The judge prompt.** You were right that it stopped being bookkeeping, so I measured it
instead of writing the caveat you asked for.

`#1803` merged and reshaped five shared fixtures inside this skill's snapshot, which took
`v1_2026-08-24_17-21-19` inactive on the merge. Six tests consume them, so a real run was owed
and `eval-cosmetic-skip` would have been a misuse. That run is
`v1_2026-08-25_21-02-48`: 21 tests, 19 pass, $6.61, active, and the first research-plan run
scored under `c39d70034788`.

The result goes against what we both expected:

| dimension | under `c39d70034788` |
|---|---|
| Record type selection | 3 × 15 |
| Objective scope containment | 3 × 15 |
| Sequencing logic | 3 × 15 |
| Jurisdiction accuracy | 3 × 15 |
| Plan mode and lifecycle | 3 × 15 |

Flat under the new prompt exactly as under the old one. Binding `judge_context` to the scores
moved nothing, so the retraction was not pessimistic and the flatness is not a grading
artifact. That strengthens #1404 rather than weakening it: the dimensions stayed flat in the
conditions most favourable to their discriminating. The caveat in the findings doc now reports
that instead of asking for it, and #1404 carries the same update.

Two non-passes, neither new. `ut_research_plan_016` fails `test_ownership_table` on
`['localities']`, which is F2 recurring and carries no graded dimensions. `ut_research_plan_wzk`
took Tool Arguments 2 for a `place_search` on Georgia that found no fixture; I corrected it to
3, since the call is the same shape as two that matched in the same run and the corpus simply
has no `place-search-georgia.json`.

promise-emmanuel (approved):
Approving. The re-measure was the right call and it settles the question I raised.

What I checked on 89b682ea. The narrow grep is back at line 309 with the widened one still at 346, so F7's "returns seven files" is true again — I re-ran both against the merge-base and got seven and eight respectively. Run four, v1_2026-08-25_21-02-48 at $6.61, is fully current: zero snapshot drift and its judge_prompt_hash matches the branch, which makes the gate's judge-prompt warning disappear entirely. Only the 016 warning remains. Every number in the new section holds against the run log — 21 tests, 19 pass, all five rubric dimensions at 3 across 15 gradings each, with the sole movement being one base Tool Arguments 2 on wzk from the missing Georgia fixture. The annotation is complete at 35 corrections across all five sampled tests, and the #1404 loop is closed on the issue itself. Suites on this head: engine vitest 2702 of 2702, tsc clean, harness pytest 2812 passed, server pytest 209 passed, runlog gate satisfied.

The re-measure is worth more than the caveat would have been. Binding the per-test clauses to the scores moved nothing, so the flatness is a property of the dimensions rather than an artifact of grading — measured in the conditions most favourable to those dimensions discriminating. That is the strongest thing #1404 has, and it only exists because you paid for the run rather than reasoning about it.

PR #1872, opened 2026-08-24
eval: e2e context-policy hook, remaining gaps (#1273 Items 1, 2, 4)

### LINKED ISSUE #1273: e2e context-policy hook: four gaps in one pass (main-thread image_read, unreachable stop, loose agent_id, unclassified research_append sections)

clack391:
## Summary

The remaining gaps in the e2e context-policy hook. Item 3 (namespaced `agent_type` plus the deliberate hook-vs-detector statement) landed in #1856; this does Items 1, 2, and 4, so it closes the issue. Rebased on #1856. Tier: Normal.

- **Item 2 (the deny names a recovery).** `context_policy._DENIAL_REASONS["extraction_append"]` told the router to "report the spawn failure and stop", but the e2e Stop hook vetoes a voluntary stop, so that instruction was unreachable. It now names the reachable recovery: delegate to `@plugin:record-extractor`, and if the spawn fails again on the same record, skip that record and note it in the run summary. `should_continue_run` is left alone, per the 2026-08-21 ruling (make continuing correct, not stop continuing).
- **Item 1 (enforce main-thread `image_read`).** The e2e guard enforced only `extraction_append`; it now keys on `SUBAGENT_ONLY_TOOLS` membership (`is_main_thread_subagent_only_tool`), so `image_read` is enforced too. A parametrized per-member test pins Asimi's acceptance addition (folded in from closed dup #1295): a future third guarded tool cannot ship unenforced in e2e.
- **Item 4 (`research_append` to `sources`/`assertions`).** These are the same protected write `extraction_append` is denied for, and `owning_skills` does not attribute them, so a delegate could write them through the broad tool unflagged. A new shadow branch in `find_protected_writes_by_unnamed_delegate` flags a named non-`record-extractor` subagent doing so. The exemption is the sibling `extraction_append` door's tight `== record-extractor` (only record-extraction, plus main-thread citation, legitimately writes these sections), and it flags once per offending call.

**Self-review fix pass (author).** Reviewing Item 4 against its sibling door surfaced defects the first pass missed, all fixed in this PR:

- The exemption was `not in DEDICATED_AGENT_NAMES` (all six dedicated agents), looser than the `extraction_append` door it parallels. A gps-mentor / proof-conclusion / research-exhaustiveness delegate writing these sections was silently not flagged. Now `!= record-extractor` (a false-negative closed).
- The branch counted per op, inflating a batch roughly 14x versus the sibling arms. Now one violation per offending call.
- `detector_before_after_report._lane_check_old`, a frozen replica of the live detector, was not given the new arm, so `--detector lane-check` attributed Item 4 hits to the unrelated #1569 fix (same replica-drift class as #1856). Mirrored, with a guard test.
- `guardrail_shadow_report`'s "they agree on today's corpus" docstring was falsified by widening the replay function; re-measured (stored and replay now diverge by the Item 4 delta).
- `result.blocked_context_calls` comment and the `interpret-e2e-result` skill still described the field as `extraction_append`-only after Item 1 routes `image_read` denials there too; both generalized.

## Review guidance

**Start here:** three code seams from the issue, each break-tested: `_DENIAL_REASONS["extraction_append"]` (Item 2); `is_main_thread_subagent_only_tool` plus its call site in `orchestrator.py` (Item 1); the new `research_append` branch in `find_protected_writes_by_unnamed_delegate` (Item 4). Then the self-review fix pass above, whose guard is the new `_lane_check_old`-mirror test.

**What it does not solve:** the *delegate*-side over-exemption is now closed. What remains is the *main-thread* route: a router writing `sources`/`assertions` through `research_append` to substitute for a failed extraction is caught by nothing today (`OWNED_SECTIONS` is `proof_summaries` only). That is a separate deny-side gap, recorded in the spec and tracked as issue #1874 (it can't be a blanket main-thread deny, since `citation` legitimately refines `sources` on the main thread).

## Plan

**Didn't change:** `stop_checker.should_continue_run` (the Stop hook, per the ruling); `owning_skills` (not widened, per the issue; its contract is the four `GUARDRAIL_SKILLS` sections, consumed by four readers).

**Deviated from the plan:** Item 4 was implemented off pre-#1856 main and rebased on #1856 (to keep this one PR). At self-review the Item 4 branch was additionally tightened, exemption and per-call count, to match the sibling `extraction_append` arm, and three downstream sites the widening silently affected (the frozen replica, a shadow-report docstring, two field/skill docs) were reconciled.

**Acceptance check:** `make harness-test` (2674 passed, 2 skipped) and `npx vitest run doc-links` (130 passed, no `#NNN` added to either spec; ratchet at baseline 10 / 16). New guards proven to fail:

- reverting the deny-reason recovery fails `test_extraction_append_denial_names_the_recovery`;
- hardcoding the guard to only `extraction_append` fails `test_every_subagent_only_tool_is_blocked_on_main[image_read]`;
- reverting Item 4's `bare_agent_type` to raw `agent_type` fails the namespaced-exemption assertion;
- widening the Item 4 exemption to `not in DEDICATED_AGENT_NAMES` fails `test_research_append_to_sources_by_non_record_extractor_dedicated_agent_flagged`;
- counting per op fails `test_research_append_batch_to_sources_flags_once_per_call_not_per_op`;
- dropping the arm from `_lane_check_old` fails `test_lane_check_old_mirrors_the_1273_item4_arm_no_divergence`.

## Test plan

- [x] `make harness-test` ran and passed (2674 passed, 2 skipped)
- [x] `npx vitest run doc-links` ran and passed (130; ratchet at baseline)
- [x] each new guard break-tested (reverted to fail, restored identical)
- No skill / test-corpus / run-log snapshot changed, so no `make eval-skill` run.

## Follow-on issues

**Folded in:** Asimi's #1295 acceptance addition is the parametrized per-member test (Item 1).

**Filed:** the main-thread `research_append` to `sources`/`assertions` substitution gap is tracked as issue #1874 (`nothing-checks`). It is pre-existing, non-blocking here, and hard to close cleanly because `citation` legitimately writes `sources` on the main thread.

Closes #1273.


johnmarkpeterbrown (changes requested):
Nice work — this is easy to review. The acceptance check is specific, and all four break-tests reproduce exactly as you described. CI is green and the branch test-merges with main cleanly.

Four things to change before merge. Three are text that now contradicts the code you changed; one is a real behaviour issue in the Item 4 branch.

### 1. The docstring promises an exemption the function doesn't have

`is_main_thread_subagent_only_tool` takes only `input_data`. There's no `declared_tools` argument and nothing in the body looks at a skill's declarations. So a skill that legitimately declares `image_read` **is** denied on the main thread now. The docstring says it isn't.

This was true before your change, when the check covered one tool and `image_read` sat outside it. Widening to set membership is what made it false.

In `eval/harness/e2e/orchestrator.py`:

```
You wrote:      rather than routing through `subagent_only_violation`, which guards the whole
                set and takes a `declared_tools` argument e2e cannot supply; keeping the check
                tool-set-specific also means a future skill that legitimately declares a
                guarded tool is not denied here. Membership in `SUBAGENT_ONLY_TOOLS` is what

Change it to:   rather than routing through `subagent_only_violation`, which takes a
                `declared_tools` argument e2e cannot supply. The consequence is that e2e has no
                declared-tools exemption at all: a future skill that legitimately declares a
                guarded tool WOULD be denied here, and closing that needs this predicate
                widened by hand. Membership in `SUBAGENT_ONLY_TOOLS` is what
```

Same claim in `docs/specs/e2e-test-spec.md`, and that one is the source of truth:

```
You wrote:      `subagent_only_violation`, so that a future skill legitimately declaring a
                guarded tool is not denied here.

Change it to:   `subagent_only_violation`, which takes a `declared_tools` argument e2e cannot
                supply. So e2e carries no declared-tools exemption: a future skill that
                legitimately declares a guarded tool would be denied here too, and closing
                that needs the predicate widened by hand.
```

### 2. Two comments still teach the doctrine Item 2 replaced

Both are in `eval/harness/harness/context_policy.py`. The second one sits directly on top of the dict you rewrote, so the file argues both ways within fifteen lines. Issue #1273 lists the first as an edit site.

```
You wrote:      #   agent fails to spawn, the router must report the failure and stop, not do
                #   the extraction and append itself (issue #942).

Change it to:   #   agent fails to spawn, the router must re-delegate to it, not do the
                #   extraction and append itself (issue #942, #1273 Item 2).
```

```
You wrote:      # just relocates the substitution. The two fixes are genuinely different:
                # `image_read` has somewhere legitimate to go (delegate the read), whereas an
                # `extraction_append` deny means the delegation itself already failed, so the
                # only correct move is to surface that and stop — NOT to retry another way,
                # which would leave the goal in place and push the substitution elsewhere
                # (issue #942).

Change it to:   # just relocates the substitution. Both fixes name a reachable recovery:
                # `image_read` delegates the read, and an `extraction_append` deny means the
                # delegation failed, so the router re-delegates and skips the record if that
                # fails again — NOT retry another way, which would leave the goal in place and
                # push the substitution elsewhere (issue #942, #1273 Item 2).
```

### 3. Item 4 covers less than the spec says

The spec says the branch flags "a named non-record-extractor subagent." The code exempts everything in `DEDICATED_AGENT_NAMES`, which is six agents. Three of them hold `research_append`: `gps-mentor`, `proof-conclusion`, `research-exhaustiveness`. None is flagged writing `sources` or `assertions`.

So `gps-mentor` calling `extraction_append` is flagged, and `gps-mentor` calling `research_append` to `assertions` is not. Same agent, same payload, opposite verdict. That's the second door still open for five agents.

Reproducing the detector's existing (agent, section) gap is a fine call. The spec just has to say so.

In `docs/specs/guardrail-enforcement-spec.md`:

```
You wrote:      unnamed-delegate test — a named non-record-extractor subagent — while leaving the
                main thread (citation refines on the main thread) and the dedicated agents
                exempt.

Change it to:   unnamed-delegate test — a subagent whose `agent_type` is not in
                `DEDICATED_AGENT_NAMES` — while leaving the main thread (citation refines on
                the main thread) and every dedicated agent exempt. That exemption is wider
                than the section's real owners: `gps-mentor`, `proof-conclusion` and
                `research-exhaustiveness` all hold `research_append` and are not flagged
                writing these sections, unlike the `extraction_append` branch, which exempts
                `record-extractor` alone. Closing it means keying on (agent, section), the
                same gap this detector's docstring already records as live.
```

And the test docstring in `eval/harness/tests/unit/test_skill_invocation.py`:

```
You wrote:      that owning_skills does not attribute, so a NAMED non-record-extractor subagent
                writing them through the broad research_append must flag (shadow)."""

Change it to:   that owning_skills does not attribute, so a subagent outside
                DEDICATED_AGENT_NAMES writing them through the broad research_append must
                flag (shadow)."""
```

### 4. Item 4 emits one violation per op, not per call

The append sits inside the `for op in _iter_ops(args)` loop. Every other clause in that function emits one string per call, and `owning_skills` de-dups.

On the one corpus run this branch newly flags, three bypassing calls produce forty-two findings — fourteen per batch. The pre-existing clauses produce fifteen for that whole run. The shadow sample is counted per detail, and that sample is what graduation out of shadow mode is gated on.

In `eval/harness/harness/skill_invocation.py`:

```
You wrote:      for op in _iter_ops(args):
                    section = op.get("section")
                    if section in ("sources", "assertions"):
                        violations.append(
                            f"tool_calls[{i}] research_append to {section!r} was made "
                            f"by agent_type={agent_type!r} (agent_id={agent_id!r}) — "
                            "this is record-extraction/citation's protected write, "
                            "made by neither the main thread nor a dedicated agent"
                        )

Change it to:   ops = _iter_ops(args)
                for section in ("sources", "assertions"):
                    if not any(op.get("section") == section for op in ops):
                        continue
                    violations.append(
                        f"tool_calls[{i}] research_append to {section!r} was made "
                        f"by agent_type={agent_type!r} (agent_id={agent_id!r}) — "
                        "this is record-extraction/citation's protected write, "
                        "made by neither the main thread nor a dedicated agent"
                    )
```

That run drops from forty-two to six, and your new tests still pass unchanged.

### One smaller thing

The narration string went generic rather than per-tool. Item 1 of the issue asks for per-tool specifically. It also points the reader at the deny reason, which isn't stored in the runlog — `blocked_context_calls` holds tool, args and `blocked_by` only.

Generic is a reasonable trade against a fourth copy of per-tool guidance. It just isn't in **Deviated from the plan**. Either name the subagent in the string, or add a line to that section saying you chose generic and why.

clack391:
Thanks, this is a careful review, and Point 1 is a real miss on my side.

One heads-up first: I think this was written against the previous head (7410d822) and GitHub stamped it on ff321fb1 at submit, because Points 2, 3, and 4 already land in the current head. Line refs so you can check them:

**Point 2 (the two context_policy.py comments).** Both already say "re-delegate", not "report/surface and stop", at lines 71 and 155.

**Point 3 (Item 4 exemption too broad).** Already tightened. The gate is `agent_id is not None and bare_agent_type != "record-extractor"` at skill_invocation.py:1019, not `not in DEDICATED_AGENT_NAMES`. I closed the door rather than documenting it open: sources and assertions have no legitimate dedicated caller except record-extractor (citation refines sources on the main thread, exempt through the agent_id branch), so a gps-mentor / proof-conclusion / research-exhaustiveness write there is genuinely out of lane and should flag, exactly like the extraction_append sibling. That resolves the asymmetry you called out: gps-mentor is now flagged for both extraction_append and research_append to those sections. The guardrail-enforcement-spec prose was updated to match. Unlike the owning_skills (agent, section) gap, these three agents own neither section, so there was nothing legitimate to preserve.

**Point 4 (per-op count).** Already one-per-call at skill_invocation.py:1015. Mine names the distinct sections in a single string per call, so on antonio-lucas-spouse the arm drops from 42 to 3, not 6, because a call writing both sources and assertions is one finding rather than two.

**Point 1 (the declared-tools docstring).** You're right, and it was a false forward-looking claim: is_main_thread_subagent_only_tool has no declared_tools input, so a declaring skill would be denied here. Fixed with your wording in both places, orchestrator.py and e2e-test-spec.md 6.1.1. I grepped the repo and those were the only two copies; the other declared-tools mentions are the unit harness, which does have the exemption.

**Narration (the smaller thing).** Named the owning subagent per tool (@plugin:record-extractor / @plugin:image-reader) and dropped "which the deny reason names", since blocked_context_calls stores only tool, args, and blocked_by, so the runlog line now stands on its own.

Full harness 2674 passed, doc-links 130. Point 1 and the narration are pushed; the rest is already on the branch.

johnmarkpeterbrown (approved):
Looks good!

promise-emmanuel (changes requested):
Two things left before merge.

1. image_read's owner is named wrong in two places
image_read is declared by exactly one agent: image-reader-opus. The sibling image-reader holds image_transcribe only — it cannot make the call that was denied. Your own spec text has this right (e2e-test-spec.md says image_read lives only on agents/image-reader-opus.md), so the diff currently contradicts itself.

This one is the important half, because it is a diagnostic instruction — it sends the reader looking for a spawn failure in an agent that could never have made the call:


You wrote:      for `extraction_append`, `image-reader` for `image_read`), not a records gap.

Change it to:   for `extraction_append`, `image-reader-opus` for `image_read`), not a records gap.
The narration map is a judgment call that is yours, not mine. The existing _DENIAL_REASONS["image_read"] — the text the model actually sees — deliberately points at @plugin:image-reader, because the right recovery is a transcription rather than a raw base64 read. That is good doctrine and I would not change it. So either name the real owner:


You wrote:                  "image_read": "@plugin:image-reader",

Change it to:               "image_read": "@plugin:image-reader-opus",
or keep the map and stop the sentence claiming ownership:


You wrote:      f"tool; delegate to {owner} rather than doing the work here."

Change it to:   f"tool; delegate the read to {owner} rather than doing the work here."
Pick one. Nothing tests either string today, which is why it got this far.

2. One check retires the replica-drift class for good
_lane_check_old has now fallen behind the live detector twice — the namespaced agent_type in #1856, and the Item 4 arm here. Both were caught by a human in review and closed with a bespoke test. The third one will not be.

There is a general version. Strip the is_error: true entries — the one thing the replica is meant to differ on — and the two functions must then agree on every committed run. I ran it both ways: green across 159 runlogs today, and red at antonio-lucas-spouse old=15 new=18 the moment the Item 4 arm is dropped from the replica. It fails for exactly the reason it exists, and it needs nobody to remember.

It belongs in test_detector_before_after_report.py, which you are already editing.

clack391:
Thanks Promise, both addressed in 5b7d74bb.

1. image_read's owner. Fixed the diagnostic in interpret-e2e-result to `image-reader-opus`, since that is the only agent that declares image_read and image-reader holds image_transcribe only. For the narration map I kept `@plugin:image-reader`, because as you said the deny reason's transcription recovery is the right target. I took your Option 2 in spirit rather than to the letter: that narration string is shared with extraction_append, so "delegate the read to" would not read right for both tools. Instead I renamed the map from `owner` to `recovery` and added a comment that for image_read this is the transcription plugin, deliberately not the owner image-reader-opus. So the diagnostic names the owner and the narration names the recovery, matching the deny doctrine. The shipped `_DENIAL_REASONS` text is untouched.

2. Replica-drift guard. Added your general check to test_detector_before_after_report.py: strip the `is_error` entries, then `_lane_check_old` must equal `_lane_check_new` on every committed e2e runlog. Break-tested it the way you described: red at antonio-lucas-spouse old=15 new=18 the moment the Item 4 arm is dropped from the replica, green across the corpus otherwise. That retires the class rather than needing a bespoke test each drift. Thanks for it.

Harness 2679 passed, 2 skipped.

promise-emmanuel (approved):
Approving. Both points from my last round are in.

The image-reader-opus fix is right, and your reasoning on the narration map is better than my suggestion — I hadn't noticed that string is shared with extraction_append, so "delegate the read to" would have read wrong for a write. Renaming owner to recovery answers the ownership claim without breaking the other tool. Good call.

The replica-drift guard does what it needs to. I re-ran it with the Item 4 arm dropped and it goes red at antonio-lucas-spouse, old=15 new=18. That class is closed now.

I also checked the two things the tests can't tell you. Item 1 won't fire spuriously: across the 20 attribution-stamped runs there are zero main-thread image_read calls — every one carries an agent_id. And Item 4's tight exemption is right: sweeping all 159 runlogs, the only attributed callers writing sources/assertions through research_append are 3 calls, all general-purpose, which is exactly the shape you're hunting. Your shadow-report figures are exact too — stored 15, replay 18.

One edit before you merge.

docs/plan/image-read-context-policy.md still says the e2e half was retired as impossible. §4.1 has the superseded banner, but the DoD underneath it wasn't updated, so it now claims the opposite of what you built. Issue #1273 named that file as one of the two places holding the stale reason — context_policy.py got fixed and this one didn't.

Leave the **Status:** line alone. §5 is genuinely still open, so the plan shouldn't be deleted yet.


You wrote:      **DoD:** a router-side `image_read` call is denied and recorded **in the unit harness**,
                with no judge involvement (**met**, verified live per §4); a subagent-side call and a
                declaring skill's direct call are unaffected (**met**, §4.1); §5 answered and filed
                (**open**). The original DoD said "both harnesses" — §4.1 retires the e2e half as
                unattributable rather than leaving it silently unmet.

Change it to:   **DoD:** a router-side `image_read` call is denied and recorded in **both harnesses** —
                the unit harness with no judge involvement (**met**, verified live per §4), and e2e
                since #1273 Item 1 (**met**, `is_main_thread_subagent_only_tool`); a subagent-side
                call is unaffected in both (**met**); a declaring skill's direct call is unaffected
                in the unit harness only, because e2e carries no declared-tools exemption
                (`docs/specs/e2e-test-spec.md` §6.1.1); §5 answered and filed (**open**).
One thing to know about, not to fix.

Item 2 rests on "report the spawn failure and stop" being unreachable in e2e. That instruction is still in the shipped skill body — record-extraction/SKILL.md line 224 — and that file is staged into every e2e run, where the model reads it before your deny reason ever fires. So in e2e the two now disagree, and the unreachable half is in the more authoritative place.

Don't touch it here. In production there is a user to report to and no Stop hook, so that line is correct doctrine there, and the lead scoped Item 2 to the harness on purpose. It just means Item 2's effect inside e2e is partial. Worth its own issue if we want that half closed.

promise-emmanuel (approved):
Approved.

PR #1877, opened 2026-08-24
place-api: strip Lucene operators, pin Accept-Language, accept a date qualifier

MMagnum:
## Summary

**Tier: Normal.**

- Fixes how a recorded place is turned into a FamilySearch query. `&` is mapped to the word "and" and `?` `*` `#` `~` are dropped, because the endpoint parses them as syntax even inside a phrase-quoted `name:"..."` value and silently matches somewhere else — a Utah census district resolved to *Denver, Colorado*, an Indiana cemetery to *North Carolina*.
- Pins `Accept-Language: en` on all six place fetchers, so a persisted `standard_place` no longer depends on an undeclared server default.
- Adds a `+date:+YYYY` capability to the resolver with three guards, plus diacritic folding and endonyms for `countryConsistency`, and runs that guard on the converter path — the highest-volume writer, previously the only unguarded one.

## Review guidance

**Start here:** `sanitizeForNameQuery` in `place-api.ts`, and specifically the choice to **map `&` and drop the rest**. That asymmetry is the load-bearing decision and it is not obvious: dropping `&` deletes a token from the phrase, so a shorter parent name outscores the real place, while dropping `#` is safe because the index normalises it. The comment above the function carries the measurement for each character; what I want checked is whether the evidence there actually supports the per-character treatment, since an earlier revision of this PR generalised one `#` result across eight characters and broke a place FamilySearch standardises correctly.

**Unsure about:** whether `?` should mean *drop the character* or *do not resolve this place at all*. `??Gren` marks two illegible letters; stripping them yields `Gren`, which ties 72-72 between Sweden and Norway and is decided by response ordering — an arbitrary answer written into a tree. It is left as a strip here because changing it is a behaviour decision, not a bug fix. Second: the `+date:` path is inert (no caller passes a date) and re-measuring it post-merge showed a genuine mix rather than the clean win claimed earlier — see "Didn't change".

## Plan

**Didn't change:**

- **The date callers.** `research_append`, `tree_edit` and `gedcomx-convert.ts` all still call `resolveStandardPlace(place)` with no options, so the `+date:` path does nothing in production. Wiring it needs `factsEquivalent` sorted first: it compares `standard_place` as a reversed comma chain, so `("Rochdale, Lancashire, ...", "Rochdale, Greater Manchester, ...")` is `false` and a corroborating assertion would file as a competing fact.
- **`~` as a capability.** It is a documented fuzzy suffix that genuinely recovers transcription variants (`Wanrooij` -> `Wanroij` at 98, where the plain query returns only the province). It is sanitised out anyway, on the narrower ground that a `~` in *recorded text* enables fuzzy nobody asked for; deliberate fuzzy belongs in an explicit option applied after sanitising, the way `date` is.
- **Alias-map coverage.** Folding and endonyms landed; growing past ~30 countries did not. Issue #1906 stays open.

**Acceptance check:**

`tests/utils/place-api.test.ts > searchPlace query sanitisation > maps & to the word 'and' rather than dropping it` — fails on `main`, passes here.

Verified by reverting each source file to `origin/main` in an isolated checkout and re-running: **6 of 11 fail** in `place-api.test.ts`, **7 of 8** of the new cases in `place-resolver.test.ts`, and **5 of 8** of the alias-map and converter-guard cases. The remainder are regression rails that should pass either way and are labelled as such. This mattered: before these tests existed, reverting both source files wholesale left the entire suite green.

**Deviated from the plan:** No `PLAN.md` — this began as an investigation into a question from the genealogy side ("when standardized place is not populated, why don't we call the FS standardization API?") and the shape emerged from measurement. Three approaches were tried and abandoned on evidence rather than opinion: a leaf-drop retry (worse in every case where it differed), backslash-escaping the operators (`\?` / `\*` return HTTP 400), and treating this as a language problem (matching is already language-agnostic). The probes for the first two are committed so the next person does not re-derive them.

## Test plan

- [x] I ran `make test-all`. **It did not pass, and none of the four failures are from this branch:**
  - *ESLint* — 347 warnings, 0 errors, in `apps/electron` and `eval/app`. The lint gate covers only those two workspaces; this PR touches neither.
  - *JS workspace (turbo)* — 5 failures in `packages/viewer-ui/src/__tests__/App.test.tsx`. Reproduced identically on a clean `origin/main` worktree.
  - *Control-plane (pytest)* — `error: uv trampoline failed to canonicalize script path`. A local Windows uv breakage, not a test result.
  - *MCP server (vitest)* — `tests/tools/write-serialization.test.ts`, and intermittently `tests/auth/login.test.ts`. Both reproduced on a clean `origin/main` worktree.

  Engine suite on this branch: **2727 passed, 1 skipped**, 2 pre-existing failures.
  Re-run after merging `origin/main` at `3fd97ad8` (53 commits behind at the time); no new failure and the same four categories. `tests/auth/login.test.ts` is intermittent — it fails under a bare `vitest run` and passes under `make test-all`.
- [ ] ~Skill run-log snapshot~ — n/a, no skill, agent, fixture or scenario touched.
- [ ] ~Skill `description` frontmatter~ — n/a.

## Follow-on issues

#1906, #1907, #1908. Comment on #1837 rather than a fourth issue.

**Folded in** — the two the review called fix-now sized: diacritic folding and endonyms for `countryConsistency` (`place-resolver.ts`, already edited here), and the missing `countryConsistency` call on the converter path (`gedcomx-convert.ts:709`).

**Filed instead:**

- **#1906** (alias-map coverage) — *depends on a decision only the lead can make*: `georgia` is both a US state and a country, and adding it has the guard treat the state as the country. Scope and source need a ruling.
- **#1907** (three indistinguishable causes of a blank `standard_place`) — *too big for this PR*; I opened the call sites and counted: three writer paths (`research-append.ts`, `tree-edit.ts`, `gedcomx-convert.ts`) plus the persisted-schema question of where a reason code lives.
- **#1908** (`record_read`'s stale justification) — *needs a different skill*: a live `record_read` against real records plus a genealogist's read on the remaining failure.
- **#1837** — commented rather than filed. It already owns `tree-edit.ts` and gates tree writes on the warnings they introduce, but does not mention place warnings; `tree_edit` never emits one for an `unverifiable` verdict, so the gate would have nothing to gate on for that class.

Issues #1906, #1907 and #1908 stay open after this merges.


MMagnum:
### Ranking-change check against the record-search criterion

`record-search-tool-spec-v2.md` § "Place matching" mandates treating a ranking
change as behavioural and re-verifying against a live run, with one measured
exemption for **tokenizer-only** changes. Flagging explicitly: **this PR is not a
tokenizer change**, so that exemption does not apply on its face. It changes the
`standard_place` *values* the resolver produces, and `record_search` reads
`standard_place || place` — so it can move a ranking input on future runs.

Measured against the criterion's two input classes anyway, since that is the
population the spec defines:

- **Class 1 (tree facts, `standard_place || place` per fact, across 265 staged +
  final trees):** 1006 distinct, **1** contains a character this PR strips —
  `"Alverson Cemetery #1, Carp, Montgomery Township, Owen, Indiana, United States"`
- **Class 2 (recorded `marriagePlace` / `recordSubdivision` under
  `eval/runlogs/e2e/`):** 62 distinct, **0**

(Class 1's 1006 matches the 1007 the spec recorded on 2026-08-19, so this is the
same population its instrument measures.)

The single affected string is the case where stripping **recovers the correct
place**: today `#1` sends it to *Alverson Cemetery, Ruffin, Rockingham, North
Carolina*; stripped it matches FamilySearch's own `Alverson Cemetery #1, Carp,
Montgomery Township, Owen, Indiana, United States` at score 95, because their
index normalises the character the same way.

Stored values are not rewritten by this PR — the change affects what future
resolutions return. So the practical blast radius is one string, in the
improving direction. Happy to run a live e2e arm if a reviewer wants it, but I
did not want to claim the tokenizer exemption for a change that is not a
tokenizer change.

Related but non-conflicting work checked: **#1584** (`placeParts` qualifier
stripping) is closed and touched `utils/marriage-jurisdictions.ts`, a separate
local-comparison tokenizer — no file overlap. **#1865** (year-range `*Exact`
qualifiers) is adjacent in subject but covers the record/person search endpoints,
not the places endpoint, and shares no files. No open PR touches
`place-api.ts` or `place-resolver.ts`.

clack391 (changes requested):
## Request changes

Good investigation, better than most. Isolating each operator to a single character, running the escape arm instead of assuming it works, committing the probes, and declining a tokenizer exemption you were not entitled to are all the right instincts. I reproduced the `#` and `*` fixes exactly, scores included.

Three things block merge.

### 1. Stripping `&` breaks a place the corpus already holds the right answer for

`place-api.ts:87-90`:

> Stripping costs nothing even when the character is genuinely part of a place name: the last case above then matches FamilySearch's own "Alverson Cemetery #1" at 95, because their index normalises the character the same way.

That generalises one `#` case across eight characters, and it fails for `&`. FamilySearch indexes the conjunction as the word "and", so dropping the `&` deletes a token from the phrase rather than normalising it, and a shorter parent name outscores the real one.

`eval/tests/e2e/reese-children/unstripped-tree.gedcomx.json:673-674` carries both halves of one Burial fact:

```json
"place":          "Manila American Cemetery & Memorial, Manila, Metro Manila, National Capital Region, Philippines",
"standard_place": "Manila American Cemetery and Memorial, Fort Bonifacio, Taguig, Rizal, Philippines"
```

FamilySearch's own standardization of that exact input is sitting right there. Through the real resolver path (derived-context filter, then `pickBest`), live: pre-PR returns `Manila American Cemetery and Memorial, Fort Bonifacio, Taguig, Rizal, Philippines` at 64, which is the recorded value. Post-PR returns `Manila, Metro Manila, Philippines` at 72, the city, with the cemetery down at rank 4.

**The fix is one character, measured against every `&` string in the corpus.** Map `&` to `" and "` instead of dropping it:

| corpus string | as-is | stripped | `& -> and` |
|---|---|---|---|
| Manila American Cemetery & Memorial, ... | Cemetery and Memorial, Fort Bonifacio (64) | **Manila city (72)** | Cemetery and Memorial, Manila (**88**) |
| Industrial School & Home for Destitute Children ... Kings, New York | Anugul, Odisha, India (61) | Kings, New York (62) | Kings, New York (**90**) |
| Election Districts 7 & 9 Ogden city Ward 2, Weber, Utah | Denver, Colorado (49) | Weber, Utah (81) | **Ogden**, Weber, Utah (**82**) |
| Great & Little Singleton, Kirkham, Lancashire | Great and Little Elm, Somerset (63) | Kirkham, Lancashire (90) | Kirkham, Lancashire (**95**) |

It wins or ties on all four, and on Ogden it recovers the city rather than only the county, so it is better on the cases you already picked as wins.

Keeping the plain strip and accepting the Manila case is defensible, but then the "costs nothing" sentence has to go and the tradeoff has to be written down. What should not ship is the claim and its counterexample together.

The corpus population here is four `&` strings: two are conjunctions of districts, where stripping helps, and two are institution names, where it helps one and breaks the other. Manila is in an `unstripped-tree.gedcomx.json`, which the record-search criterion excludes by design, so your 1-of-1007 count was never going to surface it.

### 2. Nothing in the repo fails if this change is reverted

I reverted both `src/utils/place-api.ts` and `src/utils/place-resolver.ts` to `027f059d^` in an isolated worktree and ran the full engine suite: `110 files, 2687 passed`. Identical to the branch. Neutering the sanitizer, deleting the `Accept-Language` line, and shifting the year in the qualifier are each green too.

So "Existing tests unchanged and green" is true but is not evidence: that number is invariant to whether the change works. `architecture.md` 9.4 asks a PR to say so when no check would have caught it, and the template asks for "the named test that fails on `main` and passes on this branch". `tests/utils/place-api.test.ts:34-43` already mocks fetch and asserts the query URL for the phrase-quote fix, so pinning the sanitized name, the header, the qualifier's presence and absence, the two-segment gate, the empty-result fallback and the year in the cache key is a short addition to a harness that exists.

### 3. Three headline figures cannot be produced by the committed probes

- **"Without this, 13/150 corpus places went blank."** In `probe-place-date-disagreement.ts` the dated arm is `resolveStandardPlace(place, {date})`, and `getSearchEntries` already retries undated on an empty dated result. So `dated` is null only when `undated` is too, which makes `lostByDate = differs.filter(r => r.undated && !r.dated)` structurally zero. Its only other trigger is a transient throw on one of two concurrent calls, which would make the counter misleading rather than informative. The figure must come from a pre-fallback script that is not in the PR.
- **"11 corrected, 0 regressions."** The probe counts changed rows and has no corrected-versus-regressed classifier, so the verdict is a manual read the artifact does not carry.
- **"worse in 6/6 cases where it differed."** `probe-place-leaf-coarsening.ts` computes only `leafDropHelps.length` and prints detail for the top 12 by frequency. The count is reproducible; the "worse" judgement is nowhere in the file.

These are the numbers a future reader leans on hardest, and none of them sits on a surface `measured-figures.test.ts` covers. Either have the probes emit them or restate them as what you observed by hand.

### Take the split you offered

You said you would be happy to split the date commit out. Do that. It is inert, so nothing is lost by landing the strip and the header first, and three things want answering before it is wired.

**The plan's own approach needs neither the round trip nor either fallback, and is not mentioned.** `place-resolver.ts:56` cites plan 11 step 2, but that step prescribes a client-side tie-break, "prefer the rep whose `temporalDescription.formal` covers the date", and `searchPlace` already returns that on every candidate (`place-api.ts:166`). I checked both headline cases: the dated winner is already in the undated candidate set with its range attached (`Rochdale, Lancashire [+1801/+1974]` at 95, `New York County [+1776/+1897]` at 87). A client-side preference over that set is one call, cannot return zero, cannot reach a candidate the undated query did not, and works on single-segment input. All three of your guards exist because the server-side hard filter was chosen. The honest counter is that `+date:` also re-scores and may surface reps outside the undated page, which is worth measuring rather than assuming, and your probe already fetches both sides. Step 2 is also scoped to `standardPlaceToRepId` and `standardPlaceToPlaceId`, which this PR still leaves date-blind.

**Dated and undated facts about the same place stop being equivalent.** `factsEquivalent` (`merge-gedcomx.ts:626`, used at `merge-gedcomx.ts:563` and `materialize-facts.ts:353` to decide whether an assertion corroborates an existing tree fact or coexists as a competing one) compares `standard_place` as a reversed comma chain and requires one to prefix the other. `("Rochdale, Lancashire, England, United Kingdom", "Rochdale, Greater Manchester, England, United Kingdom")` returns **false**, and `countryConsistency` cannot see it since both end in United Kingdom. So once a caller passes a date, a corroborating assertion gets filed as a competing fact. Every example in your table is this shape; the fallback you built is aimed at the rarer one.

**A year outside 1000 to 9999 returns 400, and the fallback does not cover it.** Live: `+date:+999` and `+date:+18500` and `+date:+-44` all 400, `+date:+1000` 204, `+date:+9999` 200. `searchPlace` throws on non-2xx (`place-api.ts:141`), `withRetry` burns its attempts, and `resolveStandardPlace` catches at `:394` and returns null deliberately uncached, so the place goes blank and re-burns the retries on every later call. The guard only covers `entries.length === 0`, not a throw. Reachable because `earliestYear` applies fudge offsets (`Abt` minus 1, `Bef` and `Est` minus 10): `"abt 1000"` gives 999, `"44 BC"` gives -44. Today's corpus is 1640 to 2024 so it does not fire yet, but medieval lines are where the qualifier earns its keep. `year >= 1000 && year <= 9999` is the fix. Those same offsets also mean `"before 1900"` queries `+date:+1890`; `earliestYear` was written for range sorting, and picking the earliest bound for a point query should be a stated choice.

### Worth folding in now: the header is pinned on the call whose name gets discarded

One line, in a file you already have open. `place_search` and `place_search_all` build every returned `standardPlace` from `getPlaceById`, not from `searchPlace` (`place-search.ts:115-121`, `buildPlaceResult`), as does `repIdToStandardPlace` via `getRepInfo`. `getPlaceById` sends no `Accept-Language`, and that endpoint is language-sensitive: rep 3041323 with no header gives `Fylde, Lancashire, England, United Kingdom`, and with `Accept-Language: nl` gives `Fylde, Lancashire, Engeland, Verenigd Koninkrijk`. The drift you are arguing is worth eliminating is still in place on the path that hands a name to the model and gets copied into `research.json`. `:177`, `:219` and `:353` all want the same header.

### Process

`place_search` and `place_search_all` now silently mutate a caller-supplied `placeName` and send a new header. That is a live, model-facing change to two shipped tools, and neither the body, the issue comment, nor the spec's `### Input` table (`place-search-tool-spec.md:119-124`) mentions it.

Five items under "Not in this PR", no issues filed, no reason named. `DEVELOPMENT.md` "Follow-on work you find along the way" asks for each deferral to be filed in the same PR that defers it, naming which of the four reasons applies; I searched five ways and no open issue covers any of them. Two look fix-now sized in files this PR already edits: `COUNTRY_ALIASES` in `place-resolver.ts` for the English-only alias map, and the single call site at `gedcomx-convert.ts:709` for the converter's missing `countryConsistency`. Two of the five are missing-guard shaped and want `nothing-checks`, since that register is read by label and an unfiled gap is invisible in it. The template is also unfilled; the narrative body is good and belongs inside it rather than instead of it.

### Corrections

- Spec, added text: "where no representation records coverage for the year, FamilySearch returns nothing". Not what happens. `Manger, Hordaland, Norway` has a rep with `temporalDescription.formal = "/+1963"`, which covers 1801, and both `+date:+1801` and `+date:+1900` return 204, while `New York City [+1898/]` is returned at `+date:+1990`. The filter drops reps whose temporal description has an open or absent **start**, regardless of coverage, which predicts a larger and more systematic blank population than the wording does.
- `place-resolver.ts:378-382`: "it reads the standard place's leading segment as the country and returns `ok`". It reads the input's **trailing** segment (`:325`) and scans **all** standard-place segments (`:328-330`), and for `"Mohol"` it returns `unverifiable`, not `ok`. The conclusion holds, the mechanism does not.
- `LUCENE_OPERATORS` strips eight characters; the probe covers four. The comment reads "Measured live, each isolated to the single character" over that five-case list, and the spec now states all eight as fact in the durable tier. Same for the exclusion list: no committed probe covers `: ( ) [ ] / -` or doubled commas, and in two spot checks the score was identical rather than lower. Corpus incidence of `!`, `^`, `~` and `|` is zero, so this is about the claims rather than the behaviour. (The four `~` strings in `eval/fixtures/mcp/` are the harness substring-matcher prefix per `eval/harness/harness/fixtures.py:97`, not places.)
- The issue comment's class 1 count. I re-derived it independently: 265 tree files, **1007** distinct, exactly one carrying a stripped character, class 2 is 62 with none. So it is exactly the 1007 the spec recorded, and the one-item discrepancy you explain away is not there. Substance unchanged.
- "`New York, New York` @ 1880 -> Manhattan". Undated returns `New York City, New York, United States` at 94, not Manhattan, which is rank 2 at 89 and never what `pickBest` returns; the dated answer is `New York County`.
- "the undated URL is byte-identical to before". `sanitizeForNameQuery` also collapses whitespace and trims on every input, and six corpus place strings carry doubled spaces. Benign in effect, but it is the sentence that licenses the green suite as evidence.
- Stale after this diff: `place-resolver.ts:66-68` still documents `standardizeCache` as `originalText (normalized) -> standardPlace` and `:77` as `${name}|${contextName}`; both keys gained a year. And the spec's behavioural section, "Internal function `placeSearch(placeName, contextName?)`" step 1, still shows `q=name:{placeName}` unquoted, unsanitized, with no header and no date, with step 5 still naming a two-part memo key. The appendix got the edit; the section a reader consults for what the code does did not. Step 1 was already stale for the phrase-quote fix, which is the argument for fixing it here.

### One for your genealogist eye

`"??Gren"` is listed as a win and I am not sure it is. In `eval/tests/e2e/birkeland-death-1883` that fact sits among Rogaland and Stavanger places in Norway and currently persists `"Angren, Tashkent, Uzbekistan"`. After the strip it becomes `"Gren, Värmland, Sweden"`, still not the recorded place but now plausible enough to survive a reviewer's eye where Uzbekistan would not, and `countryConsistency` returns `unverifiable` on single-segment input so nothing catches it. `??` marks two illegible characters, so stripping them asserts a name the evidence does not support, and in a tree of unproven hypotheses a plausible wrong answer is worse than an absurd one. It may be that `?` should mean "do not resolve" rather than "delete this character".

### On the live-run question

You were right not to claim the tokenizer exemption. My read is that a live e2e arm is not owed for the strip: the `record-search-tool-spec-v2` mandate targets `placeParts` and `samePlace` ranking logic, which this diff does not touch, and you changed a resolver input with a measured one-string blast radius in the improving direction and no rewrite of stored values. The thing the count cannot cover is that the committed corpus is a lower bound on what a live run feeds `resolveStandardPlace`, since mid-run the agent resolves place text pulled straight from FamilySearch records, which is exactly where census districts, cemetery plots and institution names live.

Verified by: full engine suite at `027f059d` and again with both source files reverted, `make test-all`, roughly 90 live anonymous calls covering every arm of the special-character table plus the date qualifier, locale and year boundary, my own re-derivation of both record-search input classes, and mocked break-tests over the three new guards including shapes the PR does not exercise.

MMagnum:
### Blockers 1-3 addressed in 435f8007; the split and the issues are still to come

**1. `&` maps to "and".** You were right and the Manila case is the proof — I
generalised a fact about `#` across eight characters. The "costs nothing"
sentence is gone. Mapping wins or ties on all four corpus `&` strings, and the
Ogden district now recovers the city rather than only the county.

**2. Tests that fail without the change.** 6 new in `place-api.test.ts`, 8 in
`place-resolver.test.ts`. Verified the way you did it — reverting each source
file to `origin/main` in turn: **6/11 fail** in place-api, **7/8 of the new
resolver cases fail**. The 8th ("keeps a dated answer the guard does not object
to") passes on main because main ignores the date; it is a guard-rail, not
evidence, and is labelled as such.

One of them caught a flaw in itself while I wrote it: asserting on the raw URL
meant `not.toContain("?")` matched the query separator. They read
`searchParams.get("q")` now.

**3. The three figures.**
- The blank-rate is now emitted. The probe issues a third raw `searchPlace` call,
  because — exactly as you said — the fallback has already run by the time
  `resolveStandardPlace` returns. Current corpus: **11/150 (7.3%)**. The summary
  also explains why `lost with date` reads 0 rather than leaving it to look like
  a null result.
- The other two are restated as hand reads in both probe headers, and the date
  probe's label no longer calls a change a correction.

**On that last point — the date arm looks weaker than I claimed.** Re-running
post-merge with `&` fixed, the ANSWER CHANGED list is a genuine mix, not 11
corrections. `"Industrial School & Home ... Kings, New York"` 1900 goes to
*Kings, Greenfield, **Saratoga***, and `"Boght Reformed,Church"` 1798 moves from
one wrong answer to a different wrong one. Some are clearly right (Trowbridge
drops "United Kingdom" for 1797, four years before the Union). It is not
0-regressions, and I should not have said so. More reason to take the split.

### Also in this commit

- **`~` was the wrong mechanism story.** It is documented FS syntax, and `&`/`#`
  are not wildcards at all, so "Lucene operators leaking through a phrase quote"
  was wrong in the durable tier. Corrected with the reference cited. `~` stays
  sanitised on narrower grounds: a `~` in *recorded text* silently enables fuzzy
  nobody asked for, and deliberate fuzzy belongs in an explicit option applied
  after sanitising, the way `date` is.
- **`!`, `^`, `|` dropped from the set.** No probe covers them, not documented
  operators, zero corpus incidence — asserting them was the same overclaim.
- **`Accept-Language` on all six fetchers.** You were right that I pinned it on
  the call whose name gets discarded. Checked it does not disturb the curated
  Wikipedia link: rep 442102 still returns `fr.wikipedia.org/wiki/Paris`.
- All seven corrections, including the `countryConsistency` mechanism, both
  cache-key docs, the not-byte-identical URL, and the `+date:` blank population
  (recorded as *characterised, not explained* — `Manger, Hordaland` has a rep
  covering 1801 and still 204s).
- Spec step 1 and both Input tables now say the `placeName` is mutated before
  it is sent.
- Deleted `dev/probe-place-fuzzy-accuracy.ts`, added earlier on this branch. It
  routed through `searchPlace`, which strips `~`, so both arms sent the same
  query. Its numbers were meaningless.

### Housekeeping

- Merged @johnpeterbrown's main-merge rather than force-pushing over it.
- Two suite failures — `tests/auth/login.test.ts` and
  `tests/tools/write-serialization.test.ts` — are **pre-existing**; both fail on
  a clean `origin/main` worktree. Flagging in case that is news.

### Still open

The split, the five deferred-work issues, and the PR template. Discussing scope
before I do those, since two of the five you called fix-now sized and I would
rather agree on which than guess.

MMagnum:
### On the split — proposing we don't, and here's the reasoning to push back on

You asked me to take the split I offered. I'd like to leave the date work in
instead, and I want to be explicit about it rather than quietly not doing it.

**Why leave it:**

- It is inert. No caller passes a date, so there is no behaviour to review — only
  a mechanism. The `Didn't change` section of the body says so plainly now.
- It is not untested dead code. 7 of the 8 new resolver cases fail against
  `main`, and a further 4 cover the year bounds. A reviewer can check the guards
  without also judging a corpus-wide shift in stored values, which was the point
  of the split.
- Splitting now means rewriting history on a branch @johnpeterbrown has already
  pushed a merge into. That is a real cost against a mechanism nobody is calling.

**What I'd rather you hold me to:** the date path does not get wired until the
three things you raised are settled, and the body names them so the next person
cannot wire it by accident.

### Your three points

**Year bounds — fixed** in 58f0b36b, since leaving the code in means owning its
defects. Reproduced your numbers exactly (999/10000/-44 → 400, 1000 → 204,
9999 → 200), confirmed reachability through `earliestYear`'s fudge offsets
("abt 1000" → 999, "est 1005" → 995, "44 BC" → -44), and checked the tests
exercise the guard rather than passing vacuously on an unparseable date. You
were also right that the empty-result fallback does not cover a throw — the fix
is to drop the year before sending, not to widen the catch.

The earliest-bound choice is now stated rather than implicit: "before 1900"
queries `+date:+1890`. I've argued in the comment that erring early is the safer
direction for this specific use, because the jurisdiction at the earlier bound is
the older one and the qualifier exists to stop us returning the modern rendering.
Say if you disagree — it is a choice, not a fact.

**`factsEquivalent`** — this is the one I'm most glad you caught, and I had not
looked. It is now the named blocker on wiring, in `Didn't change`.

**The plan's client-side tie-break** — I think you are probably right, and I want
to measure it rather than concede it, because "may surface reps outside the
undated page" is exactly the kind of assumption this branch has been punished for
twice already. `searchPlace` returns `dateRange` on every candidate and throws it
away, so the comparison is cheap and
`dev/probe-place-date-disagreement.ts` already fetches both sides. If a
client-side preference matches the server-side filter on the corpus, all three
guards come out and the change gets smaller — which is a better outcome than the
one I built.

### One correction to my own claim

Re-running post-merge with `&` fixed, the changed list is a mix, not 11 clean
corrections. `"Industrial School & Home ... Kings, New York"` at 1900 goes to
*Kings, Greenfield, **Saratoga***, and `"Boght Reformed,Church"` moves from one
wrong answer to a different wrong one. Some are clearly right — Trowbridge drops
"United Kingdom" for 1797, four years before the Union. The probe no longer
labels a change as a correction, and the "0 regressions" claim is withdrawn.

If you still want the split after that, say so and I'll do it.

clack391 (approved):
Approving. Every item from my last review is addressed, and I verified the fixes rather than reading the summary: replaying the real sanitizer and resolver live, `&`→"and" recovers the Manila cemetery (88) where dropping it gave only the city, the year guard degrades an out-of-range date to undated instead of sending a 400, Accept-Language is now on all six place fetchers, and the three figures that could not be reproduced are now either emitted by the probe or labeled as hand reads. The new tests bite: reverting the `&` map fails the `&` test, and reverting the year guard fails three. The deferrals are filed and the over-claims are gone. Reproducing my blocker against the committed record before fixing it is the part worth copying.

One thing to weigh before this merges, not a blocker. The `&`→"and" choice fixes the four conjunction strings in the corpus but regresses a class that is not in it: "X & District" registration districts. Live, `resolveStandardPlace("Todmorden & District, Yorkshire, England")` now persists "Yorkshire, England, United Kingdom" (the county), where dropping the `&` returns the town Todmorden at 95; "Keighley & District" is the same. That form is common in English and Commonwealth civil registration, and the persisted county is a wrong-granularity place the context and country guards cannot catch. The root is that `&` is ambiguous: a conjunction ("Great & Little Singleton") wants "and", a "& District" suffix wants the leaf, and no single rule serves both. The clean fix is a dual query for `&` (issue both variants, keep the better result); short of that, the comment should stop presenting "and" as the measured-general answer, since it is measured on four conjunctions. I am not holding the PR for this, but it is worth addressing while the branch is open.

Minor: a place field that is only `&` now sanitizes to "and" rather than empty, so the empty-input short-circuit no longer fires and it issues a needless `name:"and"` call (returns 204, no harm).

johnmarkpeterbrown (changes requested):
Good PR. The investigation is the strongest part — isolating each operator to a
single character, running the escape arm rather than assuming it works, and
committing the probes is the right shape for this kind of change.

**On the thing you asked to have checked: the `&` treatment holds.** I
re-measured all three arms live rather than reading the comment:

| corpus string | as-is | dropped | `& -> and` |
|---|---|---|---|
| Manila American Cemetery & Memorial, … | 64 cemetery | 83 **Taguig city** | **90 cemetery** |
| Election Districts 7 & 9 Ogden city … | 49 **Denver** | 81 Weber county | **82 Ogden city** |
| Great & Little Singleton, Kirkham, … | 63 **Somerset** | 87 Kirkham | **92 Kirkham** |

Mapping wins or ties on all three, and dropping genuinely loses the Manila
cemetery to the city. The per-character asymmetry is supported. Scores drift a
couple of points from your numbers; the conclusion is identical.

One thing blocks merge.

### The endonyms make `research_append` reject correct Czech resolutions

`rakousko` went in because the corpus has 68 occurrences. Every one of those is
a Bohemian or Moravian place. I resolved the four distinct ones live and ran
your guard over what came back:

```
main=unverifiable  branch=contradiction   "Svitavy, Moravská Třebová, Morava, Rakousko" => "Zwittau, Mährisch Trübau, Moravia"   (99)
main=unverifiable  branch=contradiction   "Bystré 30, Polička, Čechy, Rakousko"         => "Bistrau, Polička, Bohemia"           (95)
main=unverifiable  branch=contradiction   "Stašov, Polička, Čechy, Rakousko"            => "Dittersbach, Polička, Bohemia"       (99)
main=unverifiable  branch=contradiction   "Křičeň, Pardubice, Čechy, Rakousko"          => "Křičeň, Bezirk Pardubitz, Bohemia"   (100)
```

These are right. `Křičeň, Bezirk Pardubitz, Bohemia` at 100 is the correct
place. On `main` they are `unverifiable` — a warning, and the append proceeds.
On this branch `research_append` pushes an error and throws, which rejects the
**entire** append, not just the one entry. `tree_edit` clears the value, and the
converter path you added silently omits it on `record_search`, `person_read`,
`person_search` and `person_ancestors`.

The endonym is not the cause. `contradiction` is the fallthrough when the
*standard* side names no country the map recognises. `Bohemia`, `Moravia` and
`Czechoslovakia` are all absent from `COUNTRY_ALIASES`, so `stdCountries` comes
back empty and the function reports "plainly lacks" when it means "cannot read".
The input side already handles this correctly — it returns `unverifiable`. The
endonyms lit up the input side and exposed the asymmetry on the output side.

Corpus-wide, 5 distinct pairs flip to `contradiction`. Three are garbage worth
catching (one resolves Moravia to **North Korea**, another to **Manchester**).
Two are correct. This fix drops the false positives and keeps the Manchester
catch:

```
You wrote:      const stdCountries = placeSegments(standardPlace)
                  .map(canonicalCountry)
                  .filter((c): c is string => c !== null);
                if (stdCountries.includes(inputCountry)) return "ok";

Change it to:   const stdCountries = placeSegments(standardPlace)
                  .map(canonicalCountry)
                  .filter((c): c is string => c !== null);
                // The standard place names no country this map recognises
                // ("…, Bohemia", "…, Moravia", "…, Czechoslovakia"). That is
                // "cannot compare", not "plainly lacks" — the same reason an
                // input naming no country returns unverifiable.
                if (stdCountries.length === 0) return "unverifiable";
                if (stdCountries.includes(inputCountry)) return "ok";
```

Add a case pinning it, in the style of the ones you already wrote:

```ts
it("declines to judge when the standard place names no country it recognises", () => {
  expect(countryConsistency("Křičeň, Pardubice, Čechy, Rakousko", "Křičeň, Bezirk Pardubitz, Bohemia"))
    .toBe("unverifiable");
});
```

**The decision, so you don't have to ask.** The alternative fix is to drop
`rakousko` and `preussen` instead. Both name historical multinational polities
mapped to modern nation-states — Prussian places standardise to Poland,
Habsburg places to Czechia. That is the same ambiguity you correctly sent
`georgia` to me over in #1906; these two just happened to ship. **Take the
`stdCountries.length === 0` fix, not the drop.** It is the actual defect, it
keeps the wins, and #1906 adding the missing countries restores the North Korea
catch on top of it. Add a line to #1906 noting that historical polities are part
of its scope.

### Two one-liners worth folding into the same push

`Preußen` does not match the new key. `ß` does not decompose under NFD, so the
fold leaves it alone. The corpus has 7 `Preussen` and 2 `Preußen`, and the
latter two silently miss:

```
You wrote:      preussen: "germany",

Change it to:   preussen: "germany",
                "preußen": "germany",
```

No committed probe covers the arm you shipped. The probe's three arms are
as-is / escaped / stripped, and `strip()` maps `&` to a space — the shipped code
maps it to `" and "`. The conclusion is right, but nobody can re-derive it from
the repo. Add the Manila string to `CASES` and a fourth arm:

```
You wrote:      const strip = (s: string) =>

Change it to:   const ampersandToWord = (s: string) =>
                  s.replace(/&/g, " and ").replace(/[ ]+/g, " ").trim();
                const strip = (s: string) =>
```

### Not blocking

The inert `+date:` path is fine as declared. You gave a real reason for leaving
it unwired and the guards are well tested — I am not reopening that.

The rest verifies. Engine suite is 2745 passed, 1 skipped, 0 failed on this
head, and the two failures you saw locally do not reproduce. I checked the
acceptance claim by reverting each source file against `main`: 6 of 11 fail in
`place-api.test.ts` including the named test, 14 of 47 in `place-resolver.test.ts`,
2 of 14 in `gedcomx-standardize.test.ts`. The tests are not invariant to the
change, which was the point of the last round.

promise-emmanuel (changes requested):
Not approving — @johnmarkpeterbrown's blocker is real and I hit something on top of it.

I reproduced the regression first. Branch vs origin/main, same four pairs:

  main            branch
  unverifiable -> contradiction   "Křičeň, Pardubice, Čechy, Rakousko"  => "Křičeň, Bezirk Pardubitz, Bohemia"
  unverifiable -> contradiction   "Bystré 30, Polička, Čechy, Rakousko" => "Bistrau, Polička, Bohemia"
  unverifiable -> contradiction   "Stašov, Polička, Čechy, Rakousko"    => "Dittersbach, Polička, Bohemia"
  unverifiable -> contradiction   "Svitavy, ..., Morava, Rakousko"      => "Zwittau, Mährisch Trübau, Moravia"

Confirmed. Correct resolutions, and research_append throws on the whole append.

The part to know before you push: the suggested one-liner does not work.
I applied `if (stdCountries.length === 0) return "unverifiable";` exactly as
written and ran the engine suite. 8 failed, 2765 passed, across 4 files:

  research-append.test.ts  rejects a geocode whose country contradicts the place text (England -> Cameroon)
  research-append.test.ts  rejects a SUPPLIED standard_place whose country contradicts the place text
  research-append.test.ts  West Bromwich, England vs Bamenda, ... Cameroon -> contradiction
  tree-edit.test.ts        add_fact: nulls an AUTO-RESOLVED standard_place that contradicts ...
  tree-edit.test.ts        add_fact: nulls an EXPLICITLY-SUPPLIED standard_place that contradicts ...
  gedcomx-standardize.test.ts  omits a resolved place that contradicts the recorded country
  place-resolver.test.ts   West Bromwich, England vs Bamenda, ... Cameroon -> contradiction
  place-resolver.test.ts   falls back to the undated answer when the dated one contradicts ...

One cause for all 8: Cameroon is not in COUNTRY_ALIASES. So an empty
stdCountries is not "cannot compare" — it is the normal shape of a real
contradiction, and the guard's own flagship case is exactly that shape.

What fixes the Czech pairs without that cost is naming the historical
polities instead of the empty set. This is green at 2773/2773, and it keeps
the Cameroon catch, the North Korea catch and the Manchester catch:

You wrote:      const UK_CONSTITUENTS = new Set(["england", "scotland", "wales", "northern ireland"]);

Change it to:   // Historical polities the standard side names for territory now in a
                // different modern state. Recognising one means "cannot compare",
                // not "plainly lacks": a Bohemian place recorded as "..., Rakousko"
                // standardises to "..., Bohemia" — the same territory.
                const HISTORICAL_POLITIES = new Set([
                  "bohemia", "moravia", "czechoslovakia", "silesia", "galicia",
                  "prussia", "austria-hungary",
                ]);

                const UK_CONSTITUENTS = new Set(["england", "scotland", "wales", "northern ireland"]);

You wrote:        .filter((c): c is string => c !== null);
                  if (stdCountries.includes(inputCountry)) return "ok";

Change it to:     .filter((c): c is string => c !== null);
                  if (
                    placeSegments(standardPlace).some((seg) =>
                      HISTORICAL_POLITIES.has(seg.trim().toLowerCase()),
                    )
                  ) {
                    return "unverifiable";
                  }
                  if (stdCountries.includes(inputCountry)) return "ok";

Keep the pinning test that was asked for; it passes under this too.

Two smaller things while you are in the file.

`preussen: "germany"` never fires in this corpus and is wrong when it does.
canonicalCountry reads only the trailing segment, and both corpus Preußen
strings end in Deutschland/Germany, so that segment is never read — they
return "ok" today, not a silent miss. But "Posen, Preussen" => "Poznan, Poland"
is a correct resolution and the branch calls it a contradiction. Neither fix
above covers it. Add it to #1906 rather than patching it here.

The blank-rate figure went stale:

You wrote:      // corpus, e.g. "Manger, Hordaland, Norge" 1801). Falling back to the undated

Change it to:   // corpus, e.g. "Manger, Hordaland, Norge" 1801). Falling back to the undated

(and above it, 13/150 -> 11/150, in both place-resolver.ts:262 and the same
sentence in place-resolver.test.ts:393 — your own re-run reported 11/150.)

And probe-place-special-chars.ts still opens "Result (2026-08-20): stripping
wins 5/5" while the shipped code maps `&` to the word. Same file
@johnmarkpeterbrown asked you to add the fourth arm to — worth fixing the
header in the same edit.

Everything else verifies. Fresh worktree of the merged head: 2773 passed,
0 failed. The two failures you saw locally do not reproduce here either.
Acceptance check holds — reverting place-resolver.ts to main puts 8 of your
own tests red. All required checks green.

promise-emmanuel:
> Good PR. The investigation is the strongest part — isolating each operator to a single character, running the escape arm rather than assuming it works, and committing the probes is the right shape for this kind of change.
> 
> **On the thing you asked to have checked: the `&` treatment holds.** I re-measured all three arms live rather than reading the comment:
> 
> corpus string	as-is	dropped	`& -> and`
> Manila American Cemetery & Memorial, …	64 cemetery	83 **Taguig city**	**90 cemetery**
> Election Districts 7 & 9 Ogden city …	49 **Denver**	81 Weber county	**82 Ogden city**
> Great & Little Singleton, Kirkham, …	63 **Somerset**	87 Kirkham	**92 Kirkham**
> Mapping wins or ties on all three, and dropping genuinely loses the Manila cemetery to the city. The per-character asymmetry is supported. Scores drift a couple of points from your numbers; the conclusion is identical.
> 
> One thing blocks merge.
> 
> ### The endonyms make `research_append` reject correct Czech resolutions
> `rakousko` went in because the corpus has 68 occurrences. Every one of those is a Bohemian or Moravian place. I resolved the four distinct ones live and ran your guard over what came back:
> 
> ```
> main=unverifiable  branch=contradiction   "Svitavy, Moravská Třebová, Morava, Rakousko" => "Zwittau, Mährisch Trübau, Moravia"   (99)
> main=unverifiable  branch=contradiction   "Bystré 30, Polička, Čechy, Rakousko"         => "Bistrau, Polička, Bohemia"           (95)
> main=unverifiable  branch=contradiction   "Stašov, Polička, Čechy, Rakousko"            => "Dittersbach, Polička, Bohemia"       (99)
> main=unverifiable  branch=contradiction   "Křičeň, Pardubice, Čechy, Rakousko"          => "Křičeň, Bezirk Pardubitz, Bohemia"   (100)
> ```
> 
> These are right. `Křičeň, Bezirk Pardubitz, Bohemia` at 100 is the correct place. On `main` they are `unverifiable` — a warning, and the append proceeds. On this branch `research_append` pushes an error and throws, which rejects the **entire** append, not just the one entry. `tree_edit` clears the value, and the converter path you added silently omits it on `record_search`, `person_read`, `person_search` and `person_ancestors`.
> 
> The endonym is not the cause. `contradiction` is the fallthrough when the _standard_ side names no country the map recognises. `Bohemia`, `Moravia` and `Czechoslovakia` are all absent from `COUNTRY_ALIASES`, so `stdCountries` comes back empty and the function reports "plainly lacks" when it means "cannot read". The input side already handles this correctly — it returns `unverifiable`. The endonyms lit up the input side and exposed the asymmetry on the output side.
> 
> Corpus-wide, 5 distinct pairs flip to `contradiction`. Three are garbage worth catching (one resolves Moravia to **North Korea**, another to **Manchester**). Two are correct. This fix drops the false positives and keeps the Manchester catch:
> 
> ```
> You wrote:      const stdCountries = placeSegments(standardPlace)
>                   .map(canonicalCountry)
>                   .filter((c): c is string => c !== null);
>                 if (stdCountries.includes(inputCountry)) return "ok";
> 
> Change it to:   const stdCountries = placeSegments(standardPlace)
>                   .map(canonicalCountry)
>                   .filter((c): c is string => c !== null);
>                 // The standard place names no country this map recognises
>                 // ("…, Bohemia", "…, Moravia", "…, Czechoslovakia"). That is
>                 // "cannot compare", not "plainly lacks" — the same reason an
>                 // input naming no country returns unverifiable.
>                 if (stdCountries.length === 0) return "unverifiable";
>                 if (stdCountries.includes(inputCountry)) return "ok";
> ```
> 
> Add a case pinning it, in the style of the ones you already wrote:
> 
> ```ts
> it("declines to judge when the standard place names no country it recognises", () => {
>   expect(countryConsistency("Křičeň, Pardubice, Čechy, Rakousko", "Křičeň, Bezirk Pardubitz, Bohemia"))
>     .toBe("unverifiable");
> });
> ```
> 
> **The decision, so you don't have to ask.** The alternative fix is to drop `rakousko` and `preussen` instead. Both name historical multinational polities mapped to modern nation-states — Prussian places standardise to Poland, Habsburg places to Czechia. That is the same ambiguity you correctly sent `georgia` to me over in #1906; these two just happened to ship. **Take the `stdCountries.length === 0` fix, not the drop.** It is the actual defect, it keeps the wins, and #1906 adding the missing countries restores the North Korea catch on top of it. Add a line to #1906 noting that historical polities are part of its scope.
> 
> ### Two one-liners worth folding into the same push
> `Preußen` does not match the new key. `ß` does not decompose under NFD, so the fold leaves it alone. The corpus has 7 `Preussen` and 2 `Preußen`, and the latter two silently miss:
> 
> ```
> You wrote:      preussen: "germany",
> 
> Change it to:   preussen: "germany",
>                 "preußen": "germany",
> ```
> 
> No committed probe covers the arm you shipped. The probe's three arms are as-is / escaped / stripped, and `strip()` maps `&` to a space — the shipped code maps it to `" and "`. The conclusion is right, but nobody can re-derive it from the repo. Add the Manila string to `CASES` and a fourth arm:
> 
> ```
> You wrote:      const strip = (s: string) =>
> 
> Change it to:   const ampersandToWord = (s: string) =>
>                   s.replace(/&/g, " and ").replace(/[ ]+/g, " ").trim();
>                 const strip = (s: string) =>
> ```
> 
> ### Not blocking
> The inert `+date:` path is fine as declared. You gave a real reason for leaving it unwired and the guards are well tested — I am not reopening that.
> 
> The rest verifies. Engine suite is 2745 passed, 1 skipped, 0 failed on this head, and the two failures you saw locally do not reproduce. I checked the acceptance claim by reverting each source file against `main`: 6 of 11 fail in `place-api.test.ts` including the named test, 14 of 47 in `place-resolver.test.ts`, 2 of 14 in `gedcomx-standardize.test.ts`. The tests are not invariant to the change, which was the point of the last round.

Two things for this issue's scope, from PR #1877 (place-api / place-resolver).

1. Historical polities belong here, alongside the missing countries.
The alias map is keyed on modern nation-states, and three of the entries name
historical multinational polities instead: preussen / preußen -> germany, and
rakousko -> austria. Prussian places standardise to Poland and Habsburg places
to Czechia, so the mapping is wrong in the direction the map cannot see.
Measured: "Posen, Preussen" resolving to "Poznan, Poland" is correct, and
countryConsistency reports it as a contradiction. No corpus string reaches it
today — every Prussian segment is followed by Deutschland/Germany and the
guard reads only the trailing token — so this is latent, not live.

This is the same ambiguity as `georgia` (US state vs country), which is
already the reason this issue is a decision rather than a task. Whatever rule
settles georgia should settle these: either historical polities get their own
"cannot compare" treatment, or they come out of the alias map.

2. This issue's earlier comment undercounts countryConsistency's call sites.
It says "exactly two call sites in the repo — research-append.ts:2018 and
tree-edit.ts:331". There are four. The two missing ones both changed what
gets persisted in #1877:

  gedcomx-convert.ts   the converter path, added in that PR
  place-resolver.ts    resolveStandardPlace's dated->undated fallback, whose
                       trigger keys off the verdict — so the verdict decides
                       which rendering is written, not only whether the write
                       is rejected

Worth correcting before anyone scopes the alias work off that count.

promise-emmanuel (approved):
Approving. All four of my last-round findings are closed, and the self-attack in dd986ff8 is the better half of this push.

Verified on dd986ff8 in a fresh worktree, not read off the summary:

Engine suite 2876/2876, typecheck clean.
Re-derived the 563/2094 distinct and 8004/27028 occurrences figure independently — exact match with the comment now at the fallback.
Mutation-checked the three claims: cameroon alone → 8 failures across 4 files, north korea alone → 2, reverting the trigger to === "contradiction" → 1. All bite.
Both spec fixes match the code comments they were out of step with.
The south korea deletion is the one to copy. I tested those three aliases as a block last round, saw failures, and signed off on all three. You tested them one at a time and found the one that proved nothing. That is the difference between checking a change and checking each line of it, and my check was the coarser one.

Two things before this merges:

@clack391's approval reads 301e877b. 828195d9 — the stdCountries.length === 0 change, which is what John Mark and I both blocked on — landed after it, and stale approvals still count here. So that approval never read the guard change.

@johnmarkpeterbrown your blocker is fixed and the preußen key and the probe's fourth arm are both in — worth flipping your review so two of the approvals on record have actually read 828195d9.

Not re-run: harness and control-plane suites. No Python changed and CI is green on both.

EdmondOware (approved):
Approved.

The blocking finding is closed, and closed at the right level. `countryConsistency` now returns `unverifiable` when the standard side names no country the alias table carries, which restores the symmetry the input side always had: an unreadable country is "cannot compare", never "plainly lacks". That is the difference between a warning and `research_append` rejecting the entire append over a correct Bohemian resolution.

The four resolutions from the review are tested verbatim as `unverifiable`, which is the right way to close a finding someone measured live.

The two rails around the fix are worth more than the fix. Garbage hiding behind a genuine polity still reads as a contradiction, and the comment records that an earlier attempt returning early on any polity segment would have read those as unjudgeable too, which is the plausible wrong fix. The six pairs where the standard side names both the polity and a country the table carries would otherwise have regressed green, and nothing else covers them. Both are cases where the test exists because someone tried to break their own change.

On the genealogy, the historical handling is right. Treating Ireland as consistent with Northern Ireland matters for anything recorded before 1921, when a place now in Northern Ireland was recorded simply as Ireland, and a guard without that rule would contradict correct resolutions across a whole class of Irish records. The UK constituent rule is the same shape: England consistent with United Kingdom unless a different constituent is actually present, so Scotland against England still fails. Reading Rakousko and Preußen at all is the part that makes the Czech and Prussian corpus usable rather than uniformly unverifiable.

Issue #1826, opened 2026-08-22
search-full-text triages on textDocument, which the tool strips from every staged result


DallanQ:
> **Reviewed 2026-08-24 before junior handoff.** Corrected the paid-eval-run cost claim, the cause of defect 1, and the direction of the `count` fix, and added an ordered plan that starts with one live measurement.

**Touches:** packages/engine/mcp-server/src/tools/fulltext-search.ts, packages/engine/mcp-server/src/types/fulltext-search.ts, docs/specs/fulltext-search-tool-spec.md, packages/engine/plugin/skills/search-full-text/SKILL.md, eval/fixtures/mcp/fulltext-search-*.json, eval/tests/unit/search-full-text/rubric.md

Four defects in one surface, found while checking a senior genealogist's full-text search notes against the code.

Two of the four fixes touch snapshot-tracked files, so this card does buy one paid
eval run. `build_snapshot` (eval/harness/harness/snapshot.py:121-181) tracks the skill
folder, `eval/tests/unit/search-full-text/**`, and every `eval/fixtures/mcp/*.json` the
tests reference — so editing SKILL.md, the rubric, or a fulltext fixture flips the
newest run log inactive and `check-runlogs` blocks the PR until a fresh
`make eval-skill SKILL=search-full-text` run and its CRUD-UI annotation land. Read
`totals.total_cost_usd` and `totals.wall_clock_ms` from the newest log in
`eval/runlogs/unit/search-full-text/` for the current machine cost; the annotation pass
is the human half and the larger one. Defects 3 and 4 and the spec reconciliation touch
no snapshot-tracked file and are free — do them first.

## 1. The triage step reads a field that is never present

`search-full-text/SKILL.md:177` tells the agent to ask "Does the target name appear in the `textDocument`?"

`fulltext-search.ts:246-250` deletes `textDocument` from every result whenever the search was staged. Staging happens on every properly-run search, because the skill makes `projectPath` mandatory. The code comment says `record_read` reads it back from the sidecar — but `record_read` is not in this skill's `allowed-tools`.

So within `search-full-text`, that triage question cannot be answered. Either the field should survive staging, or the skill should be told what to triage on instead.

## 2. `highlightTerms` is returned and taught nowhere

[...]

PR #1878, opened 2026-08-24
Edit recommendations for search-full-text deep dive

### LINKED ISSUE #1798: Edit recommendations for Search-full-text deep dive
**IMPORTANT**: issue #1826 lands on this same skill after you, and it owns the triage step.

`search-full-text/SKILL.md:176` currently tells the agent to triage on `textDocument`, and
`fulltext-search.ts:246` deletes `textDocument` from every staged result — so that triage
question cannot be answered today. Issue #1826 fixes the tool and rewrites that sentence,
after one live measurement of what the tool actually returns.

Two things follow: do not write any reference-doc guidance that depends on `textDocument`,
and land first — #1826 buys its own `make eval-skill SKILL=search-full-text` run and
rebases onto yours (lead's ruling, 2026-08-24, recorded on #1826).

> **Reviewed 2026-08-21 before junior handoff.** No decision needed — every fork was
> settled from the repo. The wildcard column is the deliverable and six of eleven rows
> already fail it; read "Notes for whoever implements it" before touching the table.

**Touches:** packages/engine/plugin/skills/search-full-text/references/transcription-quirks.md, packages/engine/plugin/skills/search-full-text/references/online-search-literacy.md

Findings from the `search-full-text` deep dive (issue #1651, closed 2026-08-20), filed by
ClorindaM. Craft edits to two reference files — no code, no validators.

## Where each finding lands, checked against `main` 2026-08-21

| Finding | File | Site |
|---|---|---|
| Uncommon **and** common given names | `references/online-search-literacy.md` | `## Search Philosophy Decision` (line 6) |
| Nicknames, especially non-derivative ones | `references/online-search-literacy.md` | `## Nil-Result Checklist` (line 31) |
| Extra HTR substitutions | `references/transcription-quirks.md` | the `## HTR error patterns` table (line 7) |
| The `Mountain` example is wrong | `references/transcription-quirks.md` | same table, the `u` row |
| Several confusable letters compounding in one word (`Moiiiitain`, `Momitain`, …) | `references/transcription-quirks.md` | one line under the HTR table (line 24) |

## Notes for whoever implements it

- **Lane and scope.** All five findings are craft prose in this skill's own `references/` —
  lane 3/4 in `docs/architecture.md`. No tool or eval defect is owed. Keep the edit
  inside `skills/search-full-text/`: `search-records/references/name-search-mechanics.md`
  lines 136–142 carry a near-duplicate misread table (which already lists `u → n, v`), and
  widening it too buys a second skill's eval run plus a second annotation pass. If it needs
  the same treatment, comment on issue #1642 (`deep dive: search-records`) instead of
  editing it here.

- **Finding 1 (given names): extend the existing Situation cells, don't add a parallel pair
  of rows.** The axis the table turns on is how rare the token you anchor on is, and FTS
  does no fuzzy matching or nickname expansion (`SKILL.md:49`), so an uncommon given name
  anchors a "less is more" search exactly as an uncommon surname does. Two more rows double
  a table the agent reads on every database-evaluation pass.

- **Finding 2 (nicknames): the list already exists.** `references/search-strategies.md:73`
  has Peggy/Margaret, Polly/Mary, Sally/Sarah, Bill/William, Dick/Richard, Jack/John. Point
  the new Nil-Result Checklist item at that list, or add the non-derivative examples *there*
  — one list, not two that drift. If you extend it, add `references/search-strategies.md` to
  the `**Touches:**` line above; it is the same skill dir, so it costs no extra eval run.
  Note the constraint you are writing into: `SKILL.md:247-251` caps the nil path at 5 total
  queries and tells the agent to pick the **4 most promising** variants from
  `search-strategies.md` and `online-search-literacy.md`. A new checklist item competes for
  those four slots — say where nicknames rank, not just that they exist.

- **Findings 3–5 (the HTR table): the wildcard is the deliverable, so check it on every row
  you touch.** The findings extend the existing `Substitution` column — `long-s`, `rn`, `u`,
  `m`, `c`, `e` and `l` all already have a row. A substitution the row's wildcard cannot
  match is coverage the table only claims to have. Three rules, from this skill's own
  `references/query-syntax.md` lines 88–92 and the tool's parameter text at
  `docs/specs/fulltext-search-tool-spec.md:40`:
  1. `?` matches **exactly one** character; `*` matches zero or more.
  2. A wildcard may not be a term's first character.
  3. `*` wants at least 3 literal characters before it.

  The wildcard must match the true spelling **and** every substitution the row lists. Where
  one wildcard cannot reach them all, split the row or drop the unreachable substitution —
  never leave a wildcard that silently misses one. Record in the PR body which rows you
  checked.

- **The table is not a reliable model of itself.** Checked by hand 2026-08-21 against the
  three rules above (re-check it the same way; only a change in FamilySearch's wildcard
  semantics would move it), six of eleven rows already fail:
  - `u` — `Mo?ntain` cannot reach the `ii` it lists (9 characters against 8), and its
    example is the one this issue already flags.
  - `m` — `Willi*m` cannot match its own example `Williain`; the term must end in `m`, so it
    reaches none of `rn`/`in`/`iii`.
  - `c` — the example `Cole → Cele` demonstrates `o → e`, not the `c → e/o` the Pattern
    column claims, and `C?le` pins the leading `C`, so the confusion the row is about is
    unreachable. Decide what that row means before adding `a, d` to it; an example whose
    confusable letter is not the first character (e.g. `Bacon → Baeon`, `Ba?on`) avoids
    colliding with rule 2.
  - `e` — `*xecute` cannot match its own example `cxccute`, and it leads with a wildcard.
  - `double-l` — `A??en` cannot reach the `H` it lists.
  - `ornate capital S` — `?cott` leads with a wildcard.

  The first four are rows this issue asks you to widen, so fix them here — the eval run is
  already bought. Fix the last two too, or say in the PR why you left them.

- **The `Mountain` correction is right, and the replacement you pick decides the wildcard.**
  That row reads `Mountain → Mountan`, which demonstrates a dropped `i`, not a `u`/`n`
  confusion. The original findings offer two replacements and they are **not** equivalent:
  `Monntain` (u→n) works with the existing `Mo?ntain`; `Moiintain` (u→ii) does not — it is
  one character longer than `?` allows, and switching to `Mo*ntain` leaves only two literal
  characters before the `*`. Choose the example and the wildcard together.

- **Finding 5** is the closing paragraph of the verbatim findings: when several confusable
  letters land in one word, the corruptions multiply (`Moiiiitain`, `Momitain`, `Morritan`).
  It is the sentence that tells the agent when a fixed-length `?` is not enough and the term
  needs `*`. One line under the table — no rationale, no issue number: `no-issue-refs.test.ts`
  deliberately does not scan `references/`, so nothing would catch either.

- **If a row cannot be made self-consistent without restructuring the table** — a Condition
  column for the `c`/`l` findings' "if followed by a flourish or an `l`" qualifiers, or
  splitting single-character from multi-character substitutions into separate rows — ask the
  senior genealogist reviewer before doing it solo. `.github/CODEOWNERS` already routes the
  review there.

## Cost

Both files are under the skill directory, so the snapshot covers them: editing either flips
`search-full-text`'s run log inactive and buys one `make eval-skill SKILL=search-full-text`
run plus the sampled annotation pass. **Land all four findings in one PR** — that is the
whole reason they are one card.

## Acceptance

- [ ] Both reference files carry the five changes; every HTR row you touched satisfies the
      three wildcard rules, and the PR body names the rows you checked.
- [ ] `make eval-skill SKILL=search-full-text` re-run. Required, not optional: the newest run
      log (`v1_2026-08-20_08-44-24`) is **active** on `main` today, and any edit under the
      skill dir flips it inactive and reds `check-runlogs` on your PR. `eval-cosmetic-skip`
      does **not** apply — these are content changes, not rewording.
- [ ] The new run log **and its `.ann.json`** committed, with a correction on every dimension
      of each of the 5 tests the new run names in its own `review_sample` (that set is
      re-drawn per run — do not assume the current one). Write the annotation **only** through
      the CRUD UI (`make eval-ui`); a hand-written one crashes the CI gate.
- [ ] `git add` the run-log deletions the harness makes when it writes — it prunes candidates
      beyond the newest 5, and this skill already has 5.
- [ ] `make test-all` (the PR template's gate).

**Cost, measured:** the machine half of the re-run was **$3.05 and ~4 minutes wall clock** on
2026-08-20 — `totals.total_cost_usd` and `totals.wall_clock_ms` in
`eval/runlogs/unit/search-full-text/v1_2026-08-20_08-44-24.json`. Read those fields from the
newest run log rather than quoting this figure forward. The CRUD-UI annotation pass is the
human half and the larger one.

**What nothing checks:** no automated check reads the contents of either file.
`measured-figures.test.ts` and `no-issue-refs.test.ts` do not scan `skills/*/references/`,
and the unit suite grades whether a *query* was well-formed (`lucene-operator-construction`,
`transcription-quirk-variant`) — never whether the table's wildcards match the substitutions
they advertise. A wrong wildcard here ships green. The wildcard check above is a hand check,
and the PR body is where you record that you did it.

## Original findings, verbatim

**search-full-text
Online-search-literacy
Search Philosophy Decision
Situation**
Include uncommon given names/common given names too

**Nil-Result Checklist**
Include nicknames, esp. non-derivative ones in the list. If they are known or discovered, then it can lead to more records.

**Transcription quirks
Full-Text Search Transcription and Coverage Quirks
HTR error patterns**
long-s can also look like a p
"rn" can also show as w, ui, iu, nr
"u" can also show as v  ----- also the Example dropped the i at the end rather than say Monntain or Moiintain
"m" can also show as w, ui, iu, nr
"c" can also show as a, d (if followed by flourish or an l)
"e" ditto comments for c
"l" can show as d, b depending on preceding or following letters, also i,

When several of these individual letters or sets of letters come together (like Mountain above) it adds even more confusion. Mountain could be seen as Moiiiitain, Momitain, Morritan, Mouutain, Meutom, etc.

mercyokum:
**The issue:** A deep-dive review of the `search-full-text` skill (issue #1651) turned up five small content gaps in two of its reference docs. The search-philosophy table only talked about surnames, not given names. The nil-result checklist didn't point researchers at nicknames. And the handwriting-misread (HTR) table was missing several documented letter substitutions, its "Mountain" example demonstrated the wrong kind of mistake, and nothing explained what happens when several confusable letters land in the same word.

**The fix:** Updated `references/online-search-literacy.md` and `references/transcription-quirks.md` to cover all five findings. While editing the HTR table, I checked every row by hand against FamilySearch's wildcard rules (`?` matches exactly one character, `*` matches zero or more, and a wildcard can't lead a term) and found six rows already claiming substitutions their own wildcard couldn't reach — fixed those too, picking example words that genuinely demonstrate the misread being described (e.g. "uncle" and "angel" for the two rows whose extra substitutions only happen when the letter is followed by an `l`).

HTR rows checked: long-s, `rn`, `u`, `m`, `c`, `e`, `l`, ornate capital S, double-l, superscript abbrev, `&` ligature — all eleven re-derive clean at the current head.

## Update — round 1 (T-FEH)

- **Fixed the blocking issue.** The "compounding letters" sentence claimed a single `?`/`*` wildcard can never reach a stacked substitution run — false, since `*` matches zero-or-more and this PR's own `Willia*`/`Powe*` wildcards disprove it. (This sentence was revised again in round 3 below — see that section for the accurate final wording.)
- Fixed the superscript-abbrev row, the flourish-direction wording on `c`/`e`, extended `SKILL.md`'s surname-only bullets to also say "or given name", and corrected an earlier wrong claim that `server-test` was skipped (CI ran it green all along).
- Investigated the disclosed `ut_search_full_text_010` failure: reran it twice on this branch and twice on a pre-edit `main` worktree. Main passed 2/2; the branch's one failure tripped a *different* validator than the original, which reads as general model-sampling noise on a multi-decision test, not something this PR's content caused.

## Update — round 2 (EdmondOware)

- My round-1 fix to the superscript-abbrev row traded away the row's actual subject (a superscript mark getting **dropped** entirely, which `?` can't match — zero characters) for a single letter-swap, and in doing so made `M?s` the only wildcard in the table violating the "keep 3+ literal characters in the term" guidance the compounding-letters paragraph states. Restored both the dropped-mark and letter-swap cases and dropped the wildcard formula entirely, matching the treatment already used for the ligature/ornate-S rows: `dropped, or a letter swap` / `Mrs → Ms, Mes` / `search both`.
- Independently verified (from the committed run logs, not asserted) that `ut_search_full_text_003` already failed on the **pre-PR** 2026-08-20 baseline run — solid evidence that test's occasional routing failure predates this PR and isn't something these edits caused.

## Update — round 3 (T-FEH)

**The blocking finding here was that my round-1 rewrite of the compounding-letters paragraph, while fixing the original wrong claim, replaced it with a *different* wrong claim** — that `Mountain` can't be wildcarded at all because its two-character "Mo" prefix falls under a "3 literal characters before `*`" rule. That rule doesn't exist. Per the reviewer's own proposed resolution, I tested live against FamilySearch (this session had a valid refresh token) rather than guessing from documentation:

```
fulltext_search({ keywords: "Mo*ntain" })    -> 6,294,446 results, genuine "Mountain" hits
fulltext_search({ keywords: "+Mo*ntain" })   -> same, confirmed with the required-term operator
```

**Mo*ntain works.** The compounding-letters paragraph's final wording (in the shipped diff) is: `?` matches exactly one character, a stacked run needs `*`, and the two real constraints are that a wildcard can never open a term and the term should keep 3+ literal characters overall — no rule about characters specifically *before* the `*`.

The "before `*`" claim traced to `docs/specs/fulltext-search-tool-spec.md:40`, which turned out to be a **paraphrase error that has existed since that file was written** — the actual tool code (`fulltext-search.ts`, unchanged since the tool shipped in #119) has only ever said "min 3 chars", never "before `*`". Fixed the spec doc to match the code. Also fixed `SKILL.md`'s surname bullets, which said "or given name" in the label but still showed `+Surname` in the pattern — changed to `+Name`.

**On the shipped run log (11/13, vs. main's 12/13):** the two failures are `ut_search_full_text_003` and `ut_search_full_text_006`, both negative tests graded on whether the skill correctly declines and hands off to another skill (not on judge scores). `_003` fails with the **identical signature** on main's own baseline run (empty `skills_invoked`, no routing) — genuinely pre-existing, not something this PR's edits touch. `_006` flips between pass and fail across nearly every run in this PR's history depending on whether the model formalizes its decline as a tool call or just answers in prose; it happened to be caught failing in the run that shipped. Neither reflects a regression from this PR's content — see the full reasoning in the PR comment thread.

**Correction to an earlier claim:** the false "3 chars before `*`" rule also appears in issue #1798 itself (which named `Mo*ntain` as ruled out) and in the two issues it redirected work to (#1642, #1826). A correction comment is going on #1798 separately so those don't inherit a refuted premise.

**Deferred issues found:** none outstanding — the tool-spec paraphrase error was small enough (one file, one line, tracked by no test) to fix directly rather than file as a card.

Closes #1798


T-FEH (changes requested):
Reviewed on the merged head (`4851b26e` — your `cbbbaa4b` plus a clean merge of main). Engine suite 2687 passed; harness 2671 passed, 2 skipped. Eval gate: **All runlog rules satisfied**, run log active against the edited docs, so the paid re-run was correctly done.

**The central claim holds up.** I re-derived all eleven HTR rows from scratch against the three rules in `query-syntax.md:88-90` and every one of your six wildcard fixes is genuine: no false fixes, no new breakage, and two standing contradictions with rule 3 removed (`*xecute` and `?cott` were both invalid leading wildcards on main). `Willia*` and `Powe*` clear the literal-count rule. The substitution lists transcribe the genealogist's findings faithfully, and all four of the non-HTR gaps were real and are closed.

**And the annotation is the best I've audited recently.** The two overrides on `ut_search_full_text_006` (judge 1 → 3, "correctly routed to question-selection") are right: that test is negative with `correct_skill: ['research-plan','question-selection']`, `skills_invoked: ['question-selection']`, and negative tests grade on routing alone. The judge marked it down for not emitting a decline message and you caught it. That is exactly what the annotation pass is for.

## [Blocking] The one new sentence states a false rule — and it contradicts your own two new wildcards

`transcription-quirks.md:26-28`:

> a single `?`/`*` reaches one substitution, not a stacked run.

`*` matches zero or more characters, so it absorbs a stacked run outright. Tested against the two `*` wildcards this PR adds:

```
Willia*  vs Williaiiirn  -> MATCH      (m→iii and a second substitution stacked)
Powe*    vs Poweiiitt    -> MATCH
```

So the sentence steers the agent away from the very patterns you added to reach multi-character corruption. The real reason `Mountain` resists is rule 3, not `?`/`*` semantics: the corruption starts at position 3, leaving only "Mo" before any `*`, which is below the literal minimum.

```
You wrote:    collide — a single `?`/`*` reaches one substitution, not a stacked run.
Change it to: collide — `?` reaches exactly one character, so a stacked run needs `*`,
              and `*` only helps with three or more literal characters before the run
              (`Willia*`, `Powe*`). `Mountain` offers only two, so search the variants.
```

## Minor

- **The superscript row is not "already correct".** `M?s` cannot match `Ms` — `?` demands exactly one character and a dropped superscript leaves zero. The row is also internally muddled: Substitution says *dropped* while the Example shows a substitution (`Mrs → Mes`). Either fix the wildcard to `search both` or say which the row means; the body should not list it as verified-unchanged.
- **`Moiiiitain` is no longer derivable from the table above it.** It needs `u → ii`, which this diff removed (`n, ii` → `n, v`). `Morritan` also drops the `i` from "tain" — the same dropped-letter defect you correctly removed from the Mountain example. Both came from the original findings, but they now sit under a table that can't produce them.
- **Flourish direction is reversed.** The source finding reads "a, d (if followed by flourish or an l)"; rows 5–6 render it "(a/d mainly **after** a flourish or before `l`)".
- **`SKILL.md:75-77` still says surname-only** — the same given-name gap you fixed in the on-demand reference, sitting in the file loaded every run. Three words each.

## On the disclosed failure — "one-off model variance" understates it

`ut_search_full_text_010` did not fail on a judge score. It failed a **gating validator**, `test_log_query_traces_to_fulltext_search_call`:

> log entry `log_005` claims `recordPlace1='Pennsylvania'`, but the positionally-corresponding `fulltext_search` call sent `recordPlace1=None`

That is the skill writing a research-log entry claiming a filter it never sent. Across all five committed run logs on main (07-27 → 08-20) **that validator has never failed once** — this is its first failure, on the skill whose reference docs this PR edits.

I don't think that proves causation, and the surrounding evidence is genuinely in your favour: net suite health is unchanged (12/13 before and after, with `_003` improving from fail to pass), `_010` has wobbled before (partial on 08-19), and re-running with identical files is the right control. But a 0-for-65 guard failing for the first time isn't routine, the control isn't committed, and a log/call mismatch is plausibly downstream of guidance that makes queries more elaborate — which is what this PR adds. A couple of `--test ut_search_full_text_010` scratch runs on this branch versus main would settle it; no paid full-skill run needed.

Also worth correcting for the record: the body defers `server-test` as skipped, but CI ran `server-pytest` green on this head, so nothing is outstanding there.

Requesting changes on the one sentence. Everything else is one-line, and the underlying audit is careful work.

mercyokum:
Thanks for the thorough re-derivation — addressed everything:

**Blocking (compounding-letters sentence):** you're right, `*` matches zero-or-more so it absorbs a stacked run; the real constraint is rule 3 (>=3 literal chars before `*`), which is why `Mountain`'s "Mo" prefix can't be wildcarded. Reworded per your suggestion, and removed the two example misspellings (`Moiiiitain`, `Morritan`) that had also gone stale against the edited table — the new wording doesn't cite specific corrupted forms anymore, so that's moot.

**Superscript row:** fixed — switched the wildcard to "search both" (matching the ligature/ornate-S rows) rather than claiming `M?s` can match a zero-width drop. Withdrew the "already correct" claim from the PR body.

**Flourish direction:** fixed on both rows to "followed by," matching the source finding.

**SKILL.md surname-only bullets:** added "or given name" to all three (step 2).

**server-test:** corrected the PR body — you're right that CI ran it green; nothing outstanding.

**On "one-off model variance":** fair pushback, so I ran it down further instead of just asserting it. 2 more reruns on this branch, 2 on a `main` worktree at the pre-edit state:

| | pass | fail |
|---|---|---|
| main (pre-edit) | 2 | 0 |
| branch (post-edit) | 1 | 1 |

The branch failure was a genuinely different validator (`test_plan_item_completion_matches_its_own_record_type`) than the original (`test_log_query_traces_to_fulltext_search_call`) — two unrelated validators tripping across a small sample points at general sampling noise on a multi-decision test rather than one defect this PR's content caused, but I'm not claiming the sample is large enough to settle it outright. Full numbers are in the PR body update.

Pushed as 8dceb569. Since this touches SKILL.md too now, the run log's snapshot is invalidated again — new `make eval-skill` run and annotation pass in progress, will follow up here once done.

EdmondOware (changes requested):
The HTR audit is careful work and the six wildcard fixes hold up. I derived the table
independently rather than reading T-FEH's derivation, and the two invalid opening wildcards
that were sitting on main, `*xecute` and `?cott`, are genuinely gone. The paleography is sound
throughout: the minim family on the `rn` and `m` rows, `c` and `e` confusing with `a` and `d`
before an `l`, and the ascender confusions on `l` are real hand behaviours rather than
plausible looking padding. Choosing "uncle" and "angel" so each example actually demonstrates
the conditional in its own row is the detail I would point other authors at.

Four of T-FEH's five items are properly closed: `Moiiiitain` and `Morritan` are gone, the
flourish direction now reads "followed by", the superscript wildcard is "search both", and
`SKILL.md` step 2 carries the given name half.

One is not, and it is the blocking one.

### 1. The replacement paragraph states a rule that does not exist, and contradicts two rows of its own table

T-FEH's diagnosis was right, the old sentence was false because `*` absorbs a stacked run. The
replacement is false in a different way, and it was adopted verbatim, so this is nobody's slip
in particular.

`query-syntax.md:92` reads:

> Best practice: ≥3 literal characters per wildcard **term**

Two differences from what the paragraph now says. It is **per term**, not before the wildcard.
And it is a **best practice**, listed apart from the hard constraints above it, which are no
opening wildcard and none inside quotes.

Count literals per term and every row is fine:

| Wildcard | Literals in the term | Before the `*` |
|---|---|---|
| `Mo?ntain` | 7 | not applicable |
| `Tu*pike` | 6 | 2 |
| `Willia*` | 6 | 6 |
| `Powe*` | 4 | 4 |

So the paragraph tells the agent that `Mountain` "offers only two" and must not be wildcarded,
while line 16 of the same table gives `Mo?ntain` as Mountain's wildcard. By the same stated
rule `Tu*pike` on line 15 is invalid too. Both rows are in fact correct. Only the paragraph is
wrong.

What is true is the first half of your own sentence. `?` is exactly one character, so a run of
several corrupted characters needs `*`. Nothing stops `Mo*ntain`.

```
You wrote:    Confusable letters compound within a word. `?` reaches exactly one
              character, so a stacked run needs `*`, and `*` only helps with three or
              more literal characters before the run (`Willia*`, `Powe*`). `Mountain`
              offers only two ("Mo"), so search the variants directly instead of
              trying to wildcard them.

Change it to: Confusable letters compound within a word. `?` reaches exactly one
              character, so a run of several needs `*`, which absorbs any number
              (`Willia*`, `Powe*`, `Mo*ntain`). Two constraints still bind: a wildcard
              may never open a term, and keep three or more literal characters in the
              term overall (`query-syntax.md:92`).
```

### 2. The superscript row still says one thing and shows another

You took the wildcard half of T-FEH's either/or, which correctly retires `M?s`. The row still
reads Substitution "dropped" against Example "Mrs → Mes", and `r → e` is a substitution, not a
drop. Pick one: either the example becomes a genuine dropped mark, or the substitution column
stops saying "dropped".

### Timing, because both of these are free right now

Your body says a fresh `make eval-skill` and annotation pass are already in progress for this
round. Both fixes are prose in the same two files that run will snapshot, so folding them in
before it starts costs nothing. Doing them afterwards costs another run, and the last one was
$2.72.

mercyokum:
Good catch on both — verified your math independently before touching anything (script + output in my working notes): `Mo*ntain` has 7 literal characters in the term, not 2, because the rule is per-term-overall, not "before the wildcard" like my previous wording said. And it genuinely matches `Moiintain` (a stacked u→ii). My previous fix conflated the two readings of `query-syntax.md:92` and ended up contradicting two of its own table rows, same class of mistake as the sentence it replaced, just a different wrong rule.

**Fix 1:** Adopted your correction essentially verbatim — `?` for one character, `*` absorbs any run and needs no fixed count *before* it, the real constraints are just "never opens a term" plus the per-term-overall best practice, and `Mo*ntain` now stands as a working example instead of a claimed impossibility.

**Fix 2:** Took the other half of the either/or this time — changed Substitution from "dropped" to "e" (matching what the Example actually shows), which let `?` work again. Reverted the wildcard from "search both" back to `M?s`.

Pushed as 14af3c06. This is prose in the same file the in-progress eval run will snapshot, so no extra run needed beyond the one already queued for this round — thanks for flagging the timing.

mercyokum:
Latest round is in: re-ran `make eval-skill SKILL=search-full-text` (11/13 pass), annotated the review sample, and all CI checks are green.

Two negative tests failed this run — `ut_search_full_text_003` and `ut_search_full_text_006`. Both are routing-determined outcomes (the harness grades a negative test with a non-empty `correct_skill` purely on whether the model actually invoked the alternative skill via the `Skill` tool, not on judge scores). In both, the model wrote a good, correct-sounding prose response but never formally called the tool. I don't think this is caused by this PR — two other negative tests in the same run (`004`, `005`) used the identical file content and routed correctly, so it reads as per-sample model variance rather than a content-driven regression, consistent with `eval/CLAUDE.md`'s own note that skill runs aren't pinned to `temperature=0` and single-run variance isn't being chased at this project stage. Annotated accordingly (judge's diagnostic scores corrected where they disagreed with the actual routing outcome).

@T-FEH @EdmondOware — re-requesting review now that both rounds of feedback are addressed.

EdmondOware:
Finding 1 is closed, and closed well. The paragraph now says what the rule actually says, and
citing `references/query-syntax.md`'s wildcard rules by name rather than by line number is
better than the line reference I gave you.

Finding 2 went the other way in the same commit, and it cost something. Flagging rather than
re-requesting, since my changes-requested is already standing.

### The superscript row lost the case it existed for

`8dceb569` had it as `| superscript abbrev | dropped | Mrs → Mes | search both |`. `14af3c06`,
the commit that fixed finding 1, also changed it to:

```
| superscript abbrev | e | Mrs → Mes | `M?s` |
```

That resolves the internal contradiction, so the letter of T-FEH's either/or is met. Two
things came off with it.

**The dropped case is now covered nowhere.** The row's subject was a superscript being
dropped, which is the common outcome, and T-FEH's point was exactly that `?` cannot match zero
characters so `M?s` can never reach `Ms`. Redefining the substitution as `e` makes the row
true, but narrows it to one letter swap and drops the dropped mark out of the table entirely.

**`M?s` is now the only wildcard in the table that breaks the rule written three lines below
it.** Literal counts per term: `Ma?sachusetts` 12, `Mo?ntain` 7, `Tu*pike` 6, `Willia*` 6,
`un?le` 4, `ang?l` 4, `A?ice` 4, `Powe*` 4, and `M?s` 2. The paragraph says "keep three or more
literal characters in the term overall". An agent that reads the table and then the paragraph
is handed a contradiction, which is the same shape as the thing finding 1 fixed.

`search both` closes both at once and is what the row held an hour earlier:

```
Change it to:  | superscript abbrev | dropped, or a letter swap | Mrs → Ms, Mes | search both |
```

### Not blocking, and it supports your side of the causation question

The active run is 11 of 13, against 13 of 13 in `07-28-36`. You had no choice there. I checked:
`07-28-36` mismatches the head on `transcription-quirks.md`, so it does not describe the prose
you are shipping. Committing the worse run was correct.

On whether your edits caused it, the five committed runs say no:

| run | non-passing |
|---|---|
| `08-19_17-25-07` | `_013` aborted |
| `08-20_08-44-24` | `_003` |
| `08-24_16-07-40` | `_010` |
| `08-25_07-28-36` | none |
| `08-25_07-55-06` | `_003`, `_006` |

A different test in four of five runs. Both failures in the active run are negative tests that
activated when they should have declined. `_006` says so in the judge rationale outright, and
`_003` scores 3 on every graded dimension yet still fails, which on a negative test only comes
from the activation gate. `_003` is an extraction request and `_006` is a planning request, so
neither is reachable from surnames, given names, nicknames or transcription. None of your four
content changes touches them. That reads as an unstable negative-routing corpus and looks
worth its own card rather than anything this PR should carry.

One reason the single cell above matters: this branch has already bought three runs at $7.75.
The fix should ride whatever run T-FEH's round triggers rather than buying its own.

mercyokum:
Verified every checkable claim before touching anything — the literal-character math (M?s = 2, the only row under 3), the 08-20 baseline run's `_003` failure predating this PR, and the 08-19 abort — all confirmed by reading the actual files/run logs directly. You're right on both counts: my finding-2 fix traded the 'dropped mark' phenomenon for a single letter-swap, and reintroduced a rule-3 violation in the same commit that fixed the paragraph stating that rule.

Applied your suggested fix verbatim: `| superscript abbrev | dropped, or a letter swap | Mrs → Ms, Mes | search both |`. Pushed as db52c0b8.

EdmondOware (approved):
Approving. Both findings are closed, and the superscript row is the better of the two options.

```
| superscript abbrev | dropped, or a letter swap | Mrs → Ms, Mes | search both |
```

That keeps the dropped mark in the table instead of narrowing the row to a letter swap, and it
takes the last wildcard below the literal minimum out. Every remaining wildcard now clears the
rule the paragraph states: `Ma?sachusetts` 12, `Mo?ntain` 7, `Tu*pike` 6, `Willia*` 6, `un?le`
4, `ang?l` 4, `A?ice` 4, `Powe*` 4, and three rows carrying no wildcard at all. Nothing left
contradicting the paragraph under it.

Verified on `v1_2026-08-25_08-31-04` rather than taken from the summary: active at 51 snapshot
entries and zero mismatches, 12 of 13, retention clean at five candidates with nothing owed,
and the annotation covers the whole `review_sample` with no non-passing dimension left without
a comment. You also took `_003` from 3 down to 1 against the judge on both dimensions, which is
the harder direction and the right one.

### `ut_search_full_text_003` deserves a card, not another run

It is the only failure left, and it is not yours. Its record across the five committed runs:

| run | `_003` |
|---|---|
| `08-20_08-44-24` (inherited from `main`, pre-PR) | fail |
| `08-24_16-07-40` | pass |
| `08-25_07-28-36` | pass |
| `08-25_07-55-06` | fail |
| `08-25_08-31-04` | fail |

Three of five, and the earliest failure sits on a run log that was on `main` before this PR
was opened. So it predates every line you changed, and none of your four content changes
reaches an extraction request in any case. With `_006` and `_010` failing in other runs, this
reads as a negative-routing corpus that will not hold still.

This branch has bought four runs at $10.16 chasing a suite that moves under it, which is the
argument for a card rather than a fifth run. Worth adding to your handback rather than filing
from here.

### A note for @T-FEH, since your patch is no longer in the file

Your diagnosis of the original sentence was right and it is what got this fixed. `*` does
absorb a stacked run, so the old wording was false. The replacement you supplied turned out to
state a different rule that does not exist: `references/query-syntax.md` gives ">=3 literal
characters per wildcard **term**" as a best practice, not three before the wildcard as a hard
constraint, so `Mountain` was never barred from being wildcarded and the table's own
`Mo?ntain` and `Tu*pike` both contradicted the new sentence. It has been reworded again rather
than reverted, and the credit for catching the original defect is yours.

Nothing outstanding from me.

T-FEH (changes requested):
Re-reviewed on the merged head (`4c0d8f84` — your `0cd5ecf3` plus a clean merge of main). Engine 2695 passed / 111 files; harness 2707 passed, 2 skipped. Gate: **All runlog rules satisfied**.

**Four of my five findings are cleanly fixed** — the superscript row (`search both`, both halves of the substitution shown), the flourish direction, the undeederivable `Moiiiitain` example (gone with the rewrite), and `SKILL.md`'s given-name gap. I also re-derived the whole HTR table at this head: **all eleven rows retrieve every substitution they claim, every example word demonstrates its own row, and every pattern also matches the true spelling** so no row costs the correct hit. Picking "uncle" and "angel" so the example satisfies its row's own "followed by an `l`" conditional is careful work.

**The four paid runs are sound, and committing all of them was the right call** — it is what makes the variance claim checkable instead of asserted:

```
08-20 (main)   12/13  _003 fail          $3.05
08-24_16-07-40 12/13  _010 fail          $2.72   <- the validator failure I flagged
08-25_07-28-36 13/13                     $2.51
08-25_07-55-06 11/13  _003, _006 fail    $2.52
08-25_08-31-04 12/13  _003 fail          $2.41   <- ships, active, 0 snapshot drift
```

That answers my `_010` finding empirically: the validator failure never recurred across three later runs, and the shipping log is **identical to main's baseline** — 12/13 with the same `_003` failing. No regression.

**And the annotation is rigorous in both directions** — `_003` downgraded 3→1 ("did not correctly route to record-extraction", and `skills_invoked` is indeed `[]`) alongside `_006` upgraded 1→3 where routing did succeed. That is adjudicating the judge, not rubber-stamping it.

## [Blocking] The paragraph now recommends the one pattern issue #1798 rules out by name

`transcription-quirks.md:26-30` offers three examples of `*` usage. Two are safe; the third is the pattern the issue this PR closes explicitly rejects.

```
Willia*    6 literals before '*'   -> legal under every source
Powe*      4 literals before '*'   -> legal under every source
Mo*ntain   2 literals before '*'   -> legal only under query-syntax.md's "per term" reading
```

Issue #1798, stating rule 3 and then applying it to this exact pattern:

> 3. `*` wants at least 3 literal characters **before it**.
> … `Moiintain` (u→ii) does not — it is one character longer than `?` allows, and **switching to `Mo*ntain` leaves only two literal characters before the `*`**.

Three of the four sources agree with the issue. `docs/specs/fulltext-search-tool-spec.md:40` says "min 3 chars **before** `*`", and `src/tools/fulltext-search.ts:270` — the description the model actually reads at call time — says "min 3 chars". Only `query-syntax.md:92` supports the "per term" reading, and it calls it a **best practice**, which the new paragraph hardens into "Two constraints still bind". That is the same best-practice-as-hard-rule category error @EdmondOware flagged in the previous round.

**Measure before editing — it may save a fifth run.** One live `fulltext_search` with `Mo*ntain` settles it. If it returns results, the paragraph stands and this becomes a card against the tool spec, with no edit and no new run. If it errors, the edit is needed — and since `transcription-quirks.md` is in the snapshot, that is another ~$2.40. I tried to run the measurement myself and could not: this session's FamilySearch token expired 2026-07-25 and the loopback OAuth flow needs a browser. Whoever has a live session should make the call.

If it comes back "before":

```
You wrote:    character, so a run of several needs `*`, which absorbs any number
              (`Willia*`, `Powe*`, `Mo*ntain`). Two constraints still bind: a wildcard
              may never open a term, and keep three or more literal characters in the
              term overall (`references/query-syntax.md`'s wildcard rules).
Change it to: character, so a run of several needs `*`, which absorbs any number
              (`Willia*`, `Powe*`). Two rules bind: a wildcard may never open a term,
              and `*` needs three or more literal characters before it — `Mountain`
              offers only two, so search its variants directly.
```

## [Minor] The SKILL.md labels widened; the patterns did not

```
- **Uncommon surname or given name** → `+Surname` only, filter after
- **Common surname or given name** → `+Surname +Associate` or `+Surname +Keyword`
```

A bullet that now claims to cover given names still hands back a surname-only recipe — and `online-search-literacy.md:10`, edited in this same PR, got it right with "**that name only**, filter after". The two docs now disagree about what the same decision means. `+Name` in the second and third rows, "that name only" in the first. Fold this into whatever commit the finding above needs, so it shares the run.

## Two process notes, not blocking

- Run `07-55-06` ($2.52) was bought against `14af3c06`, which regressed the superscript row to `M?s` before `db52c0b8` fixed it — the split-fix cost @EdmondOware had warned about in the preceding review. Worth batching next time.
- `_003` fails on main too, with all validators green and the judge scoring 3/3/null while the run still records a fail — so it is a **triggering** problem, not a body problem, and none of this PR's content edits could have moved it. It deserves a card against the skill description; no open issue covers it.

Requesting changes on the one paragraph. Everything else here is done, and the underlying audit is good work.

mercyokum:
Ran the measurement you proposed rather than editing blind — this session had a refresh token, so no browser needed:

```
fulltext_search({ keywords: "Mo*ntain" })    -> 6,294,446 results, genuine "Mountain" hits
fulltext_search({ keywords: "+Mo*ntain" })   -> same, confirmed with the required-term operator too
```

Per your own resolution criteria — it returns results, so **the paragraph stands unchanged**. No edit to `transcription-quirks.md` needed, no fifth run bought for it.

That does confirm the tool spec is the thing that's wrong: `docs/specs/fulltext-search-tool-spec.md:40` claimed "min 3 chars before *", which the live API clearly doesn't enforce. Fixed that (removed the false "before *" claim, now just says "min 3 chars" matching what `fulltext-search.ts`'s actual schema already said — that file wasn't wrong, just terse). Neither file is tracked by any test or the eval snapshot, so this cost nothing extra to fix directly rather than filing a card.

Also fixed your minor finding — `SKILL.md`'s bullets now say `+Name` instead of `+Surname`, matching `online-search-literacy.md`'s generic phrasing.

Pushed as 313e3002. Since this touches `SKILL.md` again, another eval re-run + annotation is queued.

T-FEH (approved):
Third pass, on `ccb95bd4` (current with main). Engine 2700 passed / 111 files; harness 2739 passed, 2 skipped. Gate: **All runlog rules satisfied**.

**Both of my findings are resolved, and the important one was settled by measuring rather than rewording** — which is the outcome I hoped for. Thank you for running the query; I asked for it precisely because I couldn't (this session's FS token expired 2026-07-25).

**And the spec edit is more right than the commit message claims.** I traced the line: `fulltext-search.ts:270` has read `* for wildcard (min 3 chars)` since the tool shipped in #119 on 2026-05-20 — one commit, and "before `*`" has never appeared in it. So the spec wasn't describing the tool; it was a paraphrase that invented a constraint, and you corrected it toward the code. There is no spec-vs-code drift test for tool descriptions (`record-type-group-drift.test.ts` does this for `volume-search` only), so this was drift nothing could catch. Worth a card.

`+Name` is correct and matches `online-search-literacy.md`'s "that name only". Step 3's table still says `+Surname`, rightly — those rows do mean the surname. All eleven HTR rows re-derive clean at this head, and the reference docs are byte-identical to the version already reviewed.

## Two text items, and they should land *before* merge

Neither needs code or another run, but merge makes both irreversible.

**1. The PR body now asserts the rule this PR deleted.** It still says "rule 3 (three-plus literal characters needed before a `*`) … which is why `Mountain`'s two-character 'Mo' prefix can't be wildcarded". The body is the permanent record; it should say the opposite — that no such rule exists, that the tool has advertised "min 3 chars" since #119, and that you corrected the drifted spec line.

**2. `Closes #1798` will auto-close an issue carrying the refuted rule.** #1798 states "`*` wants at least 3 literal characters before it" and names `Mo*ntain` as ruled out — derived from the drifted spec line. It redirects work to **#1642** and **#1826**, both open, so both inherit the false rule. A correction comment on #1798 before merge is the cheap fix. Happy to post it myself if you'd rather.

## On the measurement, for the record

`+Mo*ntain` → 6.29M with genuine Mountain hits is good corroboration, but it is one query, and this repo's own standard for retiring a wildcard constraint is **comparative** — `name-search-mechanics.md:33-38` strikes through the identical "minimum 3 non-wildcard letters" rule and says why: *"a two-letter stem (`Sm*`) was accepted and expanded more broadly than a three-letter one (`Smi*`), which is the comparison that shows the wildcard was honoured rather than quietly dropped."* `probe-search-qualifiers.ts:5150` says the same. A single accepted query can't separate "honoured" from "silently ignored".

This does **not** weaken the spec edit, which stands on the code string alone — @EdmondOware reached the same conclusion from documents without any measurement. Just worth stating the limit if the figure goes into the #1798 comment.

## The sixth run log — state it in the body

The shipped log is 11/13, against main's 12/13. I traced both failures rather than the count:

```
                     _003              _006
main 08-20      fail / EMPTY      pass / question-selection
08-24_16-07-40  pass / fired      pass / fired
08-25_07-28-36  pass / fired      pass / fired
08-25_07-55-06  fail / EMPTY      fail / EMPTY
08-25_08-31-04  fail / EMPTY      pass / fired
08-25_11-41-40  fail / EMPTY      fail / EMPTY   <- ships
```

`_003` fails on main with the **identical signature** — non-routing, `skills_invoked: []` — so it is genuinely pre-existing. `_006` flips across runs and happens to be caught failing in the shipped log. Shipping the run you took rather than re-rolling is the right call and I'd rather see that than a hunted-for green; it just needs a sentence in the body so a reader doesn't read 11/13 as a regression.

The annotation stays discriminating throughout: `_003` downgraded 3→1 here (nothing routed), `_006` upgraded 1→3 in the earlier run where routing did succeed, and correctly left alone this time when it didn't.

## Approving

The substance is complete and correct across three rounds. Two notes on process: @EdmondOware's approval predates the spec change, the `+Name` fix and the sixth run log, so the log he verified is no longer the active one — worth a re-ping. And the eval gate now warns that this run was scored against an older judge prompt, since main advanced it afterwards; warn-only by design, but it means the corrected mean is not comparable to a post-merge re-run.

Issue #1881, opened 2026-08-24
person-evidence: nothing checks that check-warnings runs after a write — skills_invoked is not a validator fixture


johnmarkpeterbrown:
**Touches:** eval/harness/harness/validator_runner.py, eval/harness/validators/conftest.py, eval/harness/validators/test_person_evidence.py

## The gap

person-evidence's SKILL.md §8 says: "After creating links and any stub persons, **invoke `check-warnings`** on the affected persons to catch genealogical impossibilities (married before 12, died after 120, child born after a parent's death, etc.)." Nothing checks that it happens, and no rubric dimension names it.

In `v1_2026-08-20_15-53-03`, **2 of 12 write-runs skipped it** — `ut_person_evidence_025` (three `confident` links) and `ut_person_evidence_014` (mints a stub, then links it). Both scored **3 on all eight dimensions**. The other ten invoked it, so this is not a body problem: the rule is there and is usually followed.

`_014` is the case that matters. It mints a brand-new stub person and skips the guard that would catch that stub carrying an impossible lifespan.

## Why it can't be written as a validator today

`check-warnings` is a **skill**, not an MCP tool. The ten runs that did invoke it emitted no MCP call of their own, so it is invisible in `tool_calls` — the only place a validator can currently look. The signal exists in the run log as `output.skills_invoked`, but `validator_runner.run_validators` does not supply it as a fixture. The current set is `before_state`, `after_state`, `tool_calls`, `skill_frontmatter`, `blocked_context_calls`, `blocked_protected_writes`.

## The rule, once the fixture exists

> **Rule:** a run that creates `person_evidence` entries or mints a tree person must invoke the `check-warnings` skill.
> **Where to look:** `output.skills_invoked` against new `person_evidence[]` and new `persons[]` in the after-state.
> **Why it is not judgment:** both sides are recorded lists.
> **What a violation looks like:** `ut_person_evidence_025` and `ut_person_evidence_014`, run `v1_2026-08-20_15-53-03` — wrote links (and in `_014`'s case minted a stub) with `skills_invoked: ["person-evidence"]`.

## Blast radius

Three files: add the parameter to `run_validators` (default `None` so every existing caller keeps working), pass it from the orchestrator call site, add the fixture to `validators/conftest.py` defaulting to `[]`. Then the validator itself, with prove-it-fails tests alongside the two added in #1646 (`eval/harness/tests/unit/test_person_evidence_validators.py`).

[...]

PR #1882, opened 2026-08-24
person-evidence deep dive (#1646): three grading fixes, two doctrine clauses, two validators

### LINKED ISSUE #1646: deep dive: person-evidence

johnmarkpeterbrown:
## Summary

**Tier: Normal.** Deep dive on `person-evidence` (#1646) — an audit that starts from no known failure — plus the fixes and validators it produced, and the lead's 2026-08-22 materialization finding folded in rather than split out.

- **Ten findings, most from tests that passed with all-3 scores.** Three grading defects I own, two doctrine holes the body did not cover, one the lead's, and four the verification runs produced by themselves.
- **Three validators + 17 prove-it-fails tests.** One fires on `n7v` in the committed run.
- **Two SKILL.md clauses, both for rules the body did not contain** — no rule it already obeyed was restated.
- **Release candidate `v1_2026-08-24_22-05-46`**, drift **0**, annotated by the genealogist.

Step 1's prohibition list is committed for the next auditor: **36 checkable rules, 25 with nothing checking them.**

## Review guidance

**Start here:** `packages/engine/plugin/skills/person-evidence/SKILL.md` §5, the "persona with nothing to materialize from" clause, with the `marriage-assertion-links-only-defers-relationship.json` `judge_context` change beside it. `ut_person_evidence_022` created Mary Doyle with `tree_edit add_person` — which §5 prohibited **by name** — and scored 3 on all eight dimensions. My reading is that the skill was right and the body wrong: Mary is named only *inside* the groom's marriage assertion, so she carries no `record_role` and no name assertion of her own, and `materialize-facts.ts`'s `SKIP_TYPES` drops `marriage` outright, so `materialize_facts` had no persona to mint from. §7.3's nearest clause covers only a *matched* persona, which she is not. `add_person` was the only mechanism left. **If that reading is wrong, the carve-out is wrong and I would rather hear it now** — the alternative is that the tool should mint from a relationship assertion's named party, which is a tool change, not a prose one.

**Second:** `eval/tests/unit/person-evidence/rubric.md`. The `merge_warnings` gate needs a `candidateGedcomx` from a prior `record_read`, and **`record_read` is not in this skill's `allowed-tools`** — so the pass bar described a state no unit run could reach, while correct behaviour (skip it, note why) matched the *partial* bar word for word. 0 `merge_warnings` calls across 21 tests, both household runs saying plainly why. Rescoped both bars rather than removing them.

**Unsure about:** nothing outstanding. The two things previously listed here are settled — F7 landed as SKILL.md prose (the lead reaffirmed that over a tool-side fix after I surfaced `tree-edit.ts`'s existing tolerance for the sibling slip), and the annotation is complete.

## Plan

**Didn't change:** the `description` frontmatter and DO-NOT clauses, so no routing pair moved. Also left alone: the two other tests that hand `base/Tool Arguments` a conclusion (`detect-absent-household-member`, `marriage-parent-persona-unscored`) — neither writes a score, and both state a genuine fixture fact the judge cannot otherwise know.

**Acceptance check:** `cd eval/harness && uv run pytest tests/unit/test_person_evidence_validators.py` — **17 tests, absent on `main`**, each exercising a new assertion against a passing state, the exact state observed failing in a committed run log, and the preconditions under which it must stand down. Validators are not collected by `make harness-test` (`pyproject testpaths = ["tests"]`), so without this file the checks would go unexercised until someone paid for a run. Behavioural acceptance: `test_same_person_called_when_persona_meets_existing_candidate` **fails on `ut_person_evidence_n7v`** in the committed candidate, short-circuiting the judge — a validator catching an intermittent miss that `Score discipline` had reported as 3 on all 73 gradings across five prior run logs.

**Deviated from the plan:** four times, each recorded here or in the write-up.

1. **The lead's materialization finding was folded in, not split out**, at his suggestion and with his go-ahead.
2. **#1881 was filed, then folded back in — my reason for filing it was false.** I claimed `skills_invoked` "is not among the fixtures `run_validators` supplies". It has been supplied all along, from the PreToolUse hook. I read `validators/conftest.py` (standalone-pytest defaults) and inferred the harness without opening the runner. `test_search_records.py` already used the fixture, and `test_tree_edit.py::test_check_warnings_runs_after_any_tree_write` already asserted this very rule for the other tree-writing skill — landed by dive #1657, in this branch's own `main` merge. Closed as folded in.
3. **The check-warnings validator is tag-gated, at the lead's call.** Ungated it fired on 4 genuine skips in one run and converted each to a fail, short-circuiting the judge. Gated on `check-warnings-required`, carried by the stub-minting tests. **Not chosen to be green** — `_014` skipped the call in `v1_2026-08-20_15-53-03`. The ungated rate is recorded in the docstring rather than dropped: 2 of 12, 3 of 13, 4 of 13 write-runs across three logs.
4. **A `_023` judge-grounding clause was written and then reverted.** It was the only edit still flipping the snapshot; reverting kept the completed annotation valid instead of buying a fourth run for one clause. The clause is reproduced verbatim in the write-up (F9) to ride the skill's next paid run.

**Cost note:** `person-evidence/SKILL.md` grows **39.1 KB → 40.4 KB (+1.2 KB)**, warn-only on the prompt-budget check — a recurring cost on an already-large skill, paid for two clauses closing a spec-vs-body gap and a doctrine hole.

## Test plan

      **`make` is unavailable here, so each suite ran directly.** harness-test **2687 passed / 3 skipped** (exit 0); packaging **577 / 577**; engine rebuilt clean. Two environment issues hit along the way, neither a branch problem: a fresh worktree has no engine build (6 tests skip on that condition and `test_mock_mcp.py::test_successful_call_carries_no_is_error` *fails* instead — worth someone's attention), and `git worktree add` does not carry gitignored files, so `eval/.env` was absent until `scripts/link-worktree.sh`. The harness's judge preflight caught the missing key **before spending anything**.

- [x] For every skill whose **run-log snapshot** I changed, I ran `make eval-skill SKILL=<name>` and committed the run log **and its `.ann.json`**.
      **`v1_2026-08-24_22-05-46`** — 22 tests, 10 pass / 3 partial / 8 fail / 1 aborted, $6.93, **drift 0**, annotated (54 gradings across 8 tests; the 5 sampled plus 3 volunteered).
      **The red headline decomposes and should be read before judging it:** 5 of the 8 fails are this PR's own new validators reporting real skips, 1 is `n7v` (known), 1 is a judge hallucination (F9, corrected 1→3 by the genealogist), and 1 is a negative test that failed to decline. The check-warnings validator has since been gated, so a future run will not show those 5.
      An earlier candidate `v1_2026-08-24_18-17-08` (15/5/2) is also committed as the pre-gating baseline.

- [x] If I changed a skill's `description` frontmatter or its DO NOT clauses, I added or refreshed the negative test on **both** sides of every routing pair I touched.
      **N/A** — `description` and DO-NOT clauses untouched. Both SKILL.md changes sit inside steps 4 and 5 and affect no routing boundary.

## Follow-on issues

- **#1881** — closed as folded in; filed on a false premise, corrected on the issue.
- **#1657** — commented: its validator docstring claims a `person-warnings-*` fixture fix that is not in the tree. No tree-edit test declares one or mentions `person_warnings`, so on **both** sides `check-warnings` reaches the tool, finds no fixture, and reports it unavailable — the assertion covers the delegation, not the impossibility check. Recorded in my validator's docstring rather than inherited silently.
- **Not filed, recorded in the write-up instead:** the reverted `_023` grounding clause (F9); the `same_person` validator's no-link gap, which `_023` proves — the attestation is owed to the *decision*, not the link (lead's call to record, not widen); and the stale-sidecar-score validator request (F10).
- **#1655** — `ut_timeline_008`, 3-of-4 failing on `main`, untouched by anything here.

**Cost:** $14.85 — two 22-test suites ($6.89, $6.93) and two single-test probes ($1.03). One launch was refused by the judge preflight at $0, and one wasted on a stale incremental `tsc` build, caught by the harness's own staleness guard at $0.

Closes #1646.


florencemashipei (changes requested):
Strong piece of work, and the parts you flagged for checking are the parts I checked hardest. Three things I verified rather than took on trust, all of which hold: `SKIP_TYPES` really is `{relationship, age, marriage}`; `record_read` really is absent from this skill's `allowed-tools` while `merge_warnings` requires a `candidateGedcomx`, so the old pass bar did describe a state no unit run could reach; and the acceptance check reproduces — 17 passed, and `test_same_person_called_when_persona_meets_existing_candidate` genuinely fails on `n7v` in the committed candidate with a substantive message. The outcome mix matches your table exactly.

Recording deviation 2 — filing #1881 and closing it because your own premise was false — is the part I would most like other authors to copy.

## The carve-out: your reading is right, and the clause is one sentence short

You asked to hear it if the §5 reading is wrong. It is not. `materialize_facts` genuinely cannot mint a bride named only inside a groom's marriage assertion, `SKIP_TYPES` drops `marriage` outright, §7.3 covers only a matched persona, and `add_person` is the only mechanism left. I would not change that conclusion.

What the clause does not say is what she arrives carrying. `add_person` runs `assertNodeHasRef` on inline **facts** only — never on names (`tree-edit.ts:468`). So a person created exactly as written, `gender` plus `names: [{ given, surname, type }]`, lands in the tree with **no provenance at all**.

That name-path exemption is deliberate, but read why it exists (`tree-materialization-spec.md` §6):

> **`tree_edit` `add_name` / `add_person` names are ref-tolerant** … The leak this spec closes is *record-derived* evidence losing its provenance — and those names come through `materialize_facts`, which enforces a resolved ref on every name it authors. `tree_edit`'s name path serves **hypothesis, oral, and manual** stubs.

Your bride is record-derived — she is named in a marriage register. So the clause routes exactly the class of content §6 exists to protect through the exemption written for family lore and hand-entered names, and the spec's own safety argument (those names come through `materialize_facts`) is the one thing that cannot apply here.

The fix needs no tool change. The schema allows `sources` on a name and nothing in `add_person` rejects it:

You wrote:      Create her with `tree_edit add_person` (gender plus the name the record gives), then link per Step 4.
Change it to:   Create her with `tree_edit add_person` — gender, the name the record gives, and that name's `sources: [{ ref, page }]` resolved from the marriage assertion's `source_id`, since `add_person` enforces a ref on inline facts but not on names and she is record-derived, not a hypothesis stub. Then link per Step 4.

Your own alternative — teaching the tool to mint from a relationship assertion's named party — stays the better long-term answer, because then the ref is enforced rather than remembered. Worth filing either way.

## Free to fold in, same push

`validators/conftest.py` gains **`skills_invoked` twice** in this diff, at lines 84 and 134, with different docstrings. The second shadows the first. Both return `[]` so nothing breaks, but it is a duplicate in the one file whose stated job is a one-to-one map of the runner's arguments. `eval/harness/validators/` is outside the run-log snapshot, so deleting one costs nothing.

## A question, not a finding

The release candidate's outcome table is not reproducible from this branch: the check-warnings validator was tag-gated after that run, so 5 of the 8 fails would not recur. You say so plainly, and rule 2 is intact because validators sit outside the snapshot. But the log is what the next auditor reads as the skill's state. Is a candidate whose own tree no longer reproduces it the one that should be released, or does the gating want a fourth run behind it? Your call — you know this corpus better than I do.

## On the cost

The provenance sentence touches `SKILL.md`, so it buys another run, and it cannot take `eval-cosmetic-skip` — it changes what the model does. I know you already reverted the `_023` clause rather than buy a fourth run, and that was the right call for a judge-grounding fix. I am asking for this one anyway because it differs in kind: `_023` was about how a run is graded, this is about what lands in a user's tree, and this PR is the moment the doctrine gets written. Shipping the clause legitimises `add_person` here; shipping it as drafted legitimises it unsourced.

If you and the lead would rather it rode the skill's next paid run alongside F9, say so and I will approve on that basis — but I would want it recorded as a deliberate deferral rather than left implicit.

`v1_2026-08-24_18-17-08` landing without an `.ann.json` as the pre-gating baseline is fine by me; CI accepts it and the directory is at the five-candidate cap.

johnmarkpeterbrown:
Thank you — all three verified rather than taken on trust, and all three hold. Two are addressed in `2a0c19ff`; the third is deferred with your consent, recorded explicitly.

## 1. Duplicate fixture — removed, and it was worse than a duplicate

You are right, and the second copy was mine. Line 134 is pre-existing — added by the search-wikipedia dive (`91723121`) and on `main` the whole time — so with the later definition winning, **mine was dead code from the moment it landed**. Worse, its docstring asserted the opposite of what git shows: *"`run_validators` has supplied this all along; conftest had not."* It had.

I have removed mine and corrected the false claim in the F4 write-up rather than leave it standing, since that same wrong belief is what the write-up used to describe the fix. `test_tree_edit.py` now skips cleanly under a standalone run (3 skipped, no fixture error).

I cannot reconcile any of this with the "fixture not found" error I observed and reported when I added it, and I am not going to invent an explanation. What is checkable is that the fixture predates this branch and the addition was never needed.

## 2. The carve-out sentence — accepted, deferred, recorded

Your reading is correct on every step I could check: `assertNodeHasRef` runs on inline **facts** only in the `add_person` arm, never on names; §6's exemption is scoped to hypothesis/oral/manual stubs; and its stated safety argument — that record-derived names "come through `materialize_facts`, which enforces a resolved ref on every name it authors" — is exactly the thing that cannot apply to a bride named in a marriage register. Shipping the clause as drafted legitimises unsourced `add_person` for record-derived people. That is a real defect in what I wrote, not a nit.

Taking your offer: **deferred to the run `person-evidence` next pays for**, with your wording recorded verbatim in `docs/deep-dives/person-evidence-findings-2026-08-24.md` under "Deferred deliberately — the name-provenance sentence on §5's carve-out", along with why (it edits `SKILL.md`, so it buys a paid run and cannot take `eval-cosmetic-skip`) and who decided.

**Three things now ride that run**, and they should go together: this sentence, F9's `_023` grounding clause, and ungating `test_check_warnings_runs_after_a_write` if the corpus has caught up.

I also filed **#1895** for the better answer you named — teach `materialize_facts` to mint from a relationship assertion's named party, so the ref is *enforced rather than remembered* and the carve-out retires entirely. Filed rather than folded because it is a tool change with its own spec section and Vitest coverage, so it wants a tool reviewer.

## 3. Your question about the release candidate

It is the right question and my answer is: **not for its own sake, but yes as a side effect.** The candidate cannot reproduce from this branch — 5 of its 8 fails came from a validator that is now gated — and the run that would fix that is the same run items 2 and F9 are already waiting for. So I would rather buy one run that carries all three than one that only re-baselines. Until then the PR body states the decomposition, and `v1_2026-08-24_18-17-08` is committed as the pre-gating baseline.

## One thing that changed under us

A `main` merge landed on this branch while I was addressing the review. I merged it locally rather than force-pushing; **drift is still 0** and the suites are green (harness 2755 passed / 3 skipped, packaging 576 / 576).

But it moved `eval/harness/judge/prompt.md`, so `judge_prompt_hash` on the candidate now mismatches. Rule 2b is warn-only so nothing blocks — I am flagging it because the change is not cosmetic and it bears on this dive specifically. The new prompt makes `judge_context` **binding on scores**, forbids the judge emitting dimensions named after `judge_context` bullets, and states that a note asserting a fact is established and must not be contradicted.

All three touch findings here: it strengthens F2's case against score-branch `judge_context`; it would make F9's reverted grounding clause *more* effective than when I wrote it; and it addresses the judge-invented dimension (`Patrick's absence acknowledgment`) I found in the older logs. **The candidate's grades were produced by a judge prompt that no longer ships** — a fourth reason the next paid run should not drift far off.

florencemashipei (approved):
Approving. Verified the fixes rather than reading the summary: `conftest.py` now has exactly one `skills_invoked`; your git claim holds — `91723121` is the search-wikipedia dive (#1833), so the surviving copy predates this branch and yours was dead from the moment it landed; and `2a0c19ff` touches only that file and the write-up, so the snapshot is intact and every check is green on this head.

Going back to correct the false claim in your own F4 write-up, rather than leaving a wrong belief standing because the code was already fixed, is the part worth copying.

**Confirming the deferral explicitly, since my side of it was conditional.** The name-provenance sentence rides `person-evidence`'s next paid run, batched with F9's `_023` clause and the check-warnings ungating. What earns that is the record: a named section stating who raised it, that it was verified and accepted, that it is not shipped here, and the mechanism in full. That is a deliberate deferral rather than an implicit one, which was the whole of what I asked.

**#1895 is the better answer and I am glad it is filed.** Enforced rather than remembered is the right principle, and if it lands first the carve-out retires instead of being patched.

**Correcting my own wording.** I wrote "you and the lead," and @DallanQ has not weighed in. I am not holding the PR for that — the deferral is recorded publicly, #1895 carries the durable fix, and this is a genealogist-lane judgement — but he should see it, since it defers a doctrine gap in a stewarded body until a run that is not yet scheduled.

Your judge-prompt-hash disclosure is the right call to surface unprompted. A release candidate graded under a prompt that no longer ships is worth knowing about even when rule 2b lets it pass, and it makes the fourth reason not to let that next run drift.

Still needs a senior developer for the harness files.

clack391 (changes requested):
Reviewed on `8e514b30` as the senior developer florence asked for on the harness files. This is strong, honest work: the prohibition list and Did/Should/Gap findings are the deep-dive deliverable, all three validators fire, fail with precise messages, and stand down cleanly on their demonstrated cases, and `skills_invoked` is genuinely wired through `validator_runner` (the conftest `[]` is only the unit-test default). The check-warnings validator I could not break, and the docstring that discloses its delegation-only scope and #1657's false "fixture gap fixed" claim is exactly the kind of honesty worth keeping.

Four things on the two new validators and their numbers, though. Two blind readers and I each reproduced these.

## 1. `same_person` validator passes a partial skip (should-fix)

`test_same_person_called_when_persona_meets_existing_candidate` self-gates per persona (`scored` is a list) but then only asserts that a `same_person` call exists anywhere:

```python
called = [tc for tc in tool_calls if "same_person" in tc.get("tool", "")]
...
assert called, (
```

So it catches only the total-omission shape (the n7v case, 9 of 9 unscored). I reproduced two silent passes it should catch: two scored personas with a single `same_person` call for one of them, and a scored persona whose only `same_person` call was for an unrelated, rejected candidate. Both PASS. That partial/wrong-persona miss is the §7.3 household path, which is where scoring most often goes wrong. Match a call to each offending persona instead, using `primaryId1`/`primaryId2` against the persona's `record_persona_id`/`person_id`.

## 2. Materialize validator crashes to a false-fail on stringified `ops` (should-fix)

`test_matched_persona_is_materialized_onto_its_person`:

```python
for op in args.get("ops") or [args]:
    if op.get("personId"):
```

When the model serializes `ops` as a JSON string, `for op in "<string>"` iterates characters and `op.get(...)` raises `AttributeError`. `materialize-facts.ts` explicitly recovers that shape (`coerceJsonArg`), and the mock records the raw model args, so it is a real shape. `validator_runner` turns the crash into `passed=False`, which gates the test and deletes its judge scores. I reproduced it. It is latent today (only `ut_person_evidence_027` carries the `materialize` tag, in the flat form), but your own docstring plans to widen the tag to household tests where stringified `ops` is normal.

```
You wrote:      for op in args.get("ops") or [args]:
Change it to:   ops = args.get("ops")
                if isinstance(ops, str):
                    try: ops = json.loads(ops)
                    except ValueError: ops = []
                for op in (ops if isinstance(ops, list) else [args]):
                    if isinstance(op, dict) and op.get("personId"):
```

## 3. The validator tests do not pin their own logic (should-fix)

The tests have failing-input cases, which is good, but they do not pin the load-bearing constants, so a future edit that guts a validator stays green. I confirmed three: adding `"death"` to `_UNMATERIALIZABLE`, dropping `"age"`, and broadening validator 1 to accept `record_search` as a score all pass 17/17. The last one turns validator 1 into a no-op since every record-search test calls `record_search`. Add a test that pins `_UNMATERIALIZABLE` (a death-only materializable run must be owed; an age-only run must not) and one that pins validator 1's per-persona match against a `record_search`-only, `same_person`-absent failing state.

## 4. Two numbers do not re-derive, including the one that motivates validator 1 (should-fix)

Both live in the durable findings doc and in the validator docstrings, so both need the fix.

- The check-warnings skip rate for `v1_2026-08-24_22-05-46` reads "4 of 13" and the summary "nine skips across five distinct tests". Re-derived from that run log's own baked results it is 5 of 14 (`_002, _010, _013, _019, _022`), and three of those are outside the named set, so the summary is 10 skips across 8 tests.
- Validator 1's justification says the Score discipline dimension "took the value 3 on all 73 gradings". Over the nine committed run logs it is 78 gradings, and the dimension took 1 twice (n7v in `v1_2026-08-20_15-53-03`, `_023` in `v1_2026-08-24_22-05-46`). This PR deleted the older run logs the "73 / all-3" figure was measured against, so the figure was not recomputed after the corpus changed. Validator 1 is still justified, since the dimension catches this only about 2 in 71, but the sentence needs to be true.

One thing to confirm, not a finding: `v1_2026-08-24_22-05-46` was generated while validator 3 was still ungated, so its committed outcomes for the untagged tests are stale relative to the shipped gated validators. The pending re-run the findings doc already flags for the judge-prompt-hash mismatch should refresh those; worth making sure it does.

Once the two validator fixes, the test pinning, and the number corrections are in, I am happy to approve.

johnmarkpeterbrown:
Thanks — I reproduced all four before touching anything, and all four were real. Fixed in `786bf28f`; nothing here touches a snapshot-tracked file, so drift stays 0 and no run was bought.

## 1 + 2 — the two validator defects

Both reproduced exactly as described. **Item 1**: two silent passes — one `same_person` call covering two scored personas, and a call scoring an unrelated candidate. **Item 2**: `AttributeError: 'str' object has no attribute 'get'`, and I confirmed `coerceJsonArg` is real in `materialize-facts.ts`, so the shape is not hypothetical.

- `_same_person_pairs` now matches each offending persona's `(record_persona_id, person_id)` against the actual `primaryId1`/`primaryId2`, both orderings, and the failure message names which pairings **were** scored so a false-fail is diagnosable rather than mysterious.
- `_materialize_ops` parses a JSON-string `ops` and falls back safely on an unparseable one — with a test pinning that the malformed case still fires rather than silently satisfying the check.

Your point that the half-attested case is the §7.3 household path is the part I had not thought through: the validator was strongest exactly where scoring is easiest and blind where it is hardest.

## 3 — the pins, including two of mine that did not bite

Confirmed all three of your surviving mutations. 26 tests now, up from 17.

**Two of my new pins failed to bite on first write, which is your point again rather than a separate one.** The `"death"` pin *skipped* instead of failing — a skip is not a failure, so the mutation survived as `25 passed, 1 skipped`. And the tool-match pin passed a `record_search` call carrying no `primaryId`s, which no widening of the tool match could ever have satisfied, so it pinned nothing. Both rewritten: the first converts a skip to an explicit `pytest.fail`, the second uses a call carrying the exact pair so a broadened match genuinely does go red.

Five-mutation battery, all now caught:

| Mutation | Result |
|---|---|
| add `"death"` to `_UNMATERIALIZABLE` | 1 failed |
| drop `"age"` | 1 failed |
| broaden the tool match | 1 failed |
| revert to the per-run assertion | 3 failed |
| remove the stringified-`ops` guard | 1 failed |

## 4 — the numbers, and the one that matters

Both wrong, and your figures are exact. `v1_2026-08-24_22-05-46` is **5 of 14**, not 4 of 13; the summary is **10 skips across 8 tests** over the three logs cited (17 across 9 over all five committed logs).

The Score discipline figure is the instructive one and I want to be plain about the mechanism rather than just correct the digits. "3 on all 73 gradings" re-derives as **69 of 71**, and the dimension takes 1 **twice** — `n7v` in `v1_2026-08-20_15-53-03` and `_023` in `v1_2026-08-24_22-05-46`, both findings *from this dive*. It was measured against a five-log corpus that **this PR's own runs pruned** (`07-29`, `08-03`, `08-05` aged out under retention), and I never recomputed it. So the sentence justifying the validator had become false by the time the validator shipped, in the direction that flatters it.

`CLAUDE.md`'s rule is that a measurement disagreeing with belief is re-measured, not reworded — and a figure whose corpus has moved is the same case. Corrected in both docstrings and the write-up, with a section recording the mechanism and a standing instruction that any corpus-cited figure in that document be re-derived before it is repeated.

## Your non-finding

Confirmed, and already flagged above the fold: the candidate's outcomes for the untagged tests are stale relative to the shipped gated validators. The write-up's banner names the pending run and now lists four reasons for it — the two deferred clauses, the ungating, and the `judge_prompt_hash` mismatch. I will make sure it refreshes those outcomes.

## State

`main` merged into the branch twice while this was in review (including `8e514b30`, the commit you reviewed on). Both merged locally rather than force-pushed; **drift 0** throughout. harness **2764 passed / 3 skipped**, packaging **578 / 578**.

clack391 (changes requested):
Re-reviewed on `786bf28f` as the senior developer on the harness files. All four items from my last review are correctly fixed and I verified each by running it, not by reading the summary:

- Per-persona `same_person` matching (`_same_person_pairs`) is right for the identity case, and the id alignment holds on real data (the passing n7v runs call `F1->I1` / `M1->I2`, exactly what SKILL.md Step 2 mandates).
- `_materialize_ops` handles the stringified, list, absent, and malformed shapes.
- The pins bite: 26 tests, and every one of the five mutations (add `death`, drop `age`, broaden the tool match, revert to per-run, drop the ops guard) goes red.
- Both numbers re-derive exactly, including Score discipline at 69 of 71. The correction record you left in the findings doc is the part worth copying.

One new thing, though, from attacking the delta as its own change. The per-persona fix is correct for an identity match, but the precondition it runs under is too broad, so it now fires on compliant runs and deletes their judge scores.

## Blocker: the `scored` filter false-fails links that do not owe a `same_person` call

`test_same_person_called_when_persona_meets_existing_candidate` selects every new `person_evidence` entry whose assertion has a non-null `record_persona_id` and links to a pre-existing tree person, then demands a `same_person` pair for each. But SKILL.md endorses linking such an assertion to an existing person with no `same_person(persona, person)` call in three cases, so the guard fails a correct run:

1. **Relationship-type assertions (active, on a committed passing run).** In n7v, `a_004` and `a_005` are `fact_type: relationship` ("child of Thomas", "child of Bridget") sourced from the groom persona `G1`, and they link to the existing parents `I1` / `I2`. Those parents are identity-matched through the parent personas, `F1->I1` and `M1->I2`, which the passing runs do call. `same_person(G1, I1)` would compare the groom to his father, which SKILL.md never asks for. I ran the current validator against the reconstructed inputs of the passing `v1_2026-08-12` n7v run and it fires on `a_004/G1->I1` and `a_005/G1->I2`. So on the next run this flips a passing n7v to fail.
2. **Household relatives scored via one `matchRelatives: true` call (latent).** SKILL.md Step 2.4 says a household is paired in one shot; the relative scores come back in that call's `matches` array, never as separate `same_person` args, and the unit tier records no tool responses (your own F4), so the validator cannot see them. It flags every relative that matches a pre-existing person. Latent only because the two household fixtures mint new relatives.
3. **`results_ref: null` (latent).** SKILL.md Step 2 says a search that predates result retention has no sidecar, so `same_person` cannot be built and correlation stands alone. The guard flags it anyway.

Fix: scope `scored` to the identity links this tier can actually verify. Drop relationship (and other non-identity) fact types, drop assertions whose log entry has `results_ref: null`, and do not demand a per-relative pair when a `matchRelatives: true` call is present.

## Should-fix: the new pin encodes the household false-fail as correct

`test_scored_fires_per_persona_not_per_run` asserts that one `same_person` call on a household of several scored personas is a failure ("leaves the rest unattested"). That is the compliant Step 2.4 path, so the pin cements case 2. Rewrite it once `scored` is re-scoped, so it does not lock in the false-fail.

Once the re-scope and the test rewrite are in, I am happy to approve. Everything else here is solid.

johnmarkpeterbrown:
Both fixed in `c79c3c9d`. I reproduced the blocker against the real run before touching the code, and the fix is verified against both sides.

## Blocker — the `scored` filter was too broad

You are right, and it is worse than latent: it breaks on a **committed passing run today**. Replaying the shipped validator against `v1_2026-08-12_17-18-54`'s n7v run (recorded outcome `pass`) fires on exactly the two you named:

```
pe_x4 (a_004/G1 -> I1), pe_x6 (a_005/G1 -> I2)
Pairings actually scored: [('F1','I1'), ('M1','I2')]
```

Both are `fact_type: relationship` on the **groom** persona G1, linking to the **parents**. The run did the right thing — `F1->I1`, `M1->I2` — and my guard asked for `same_person(groom, father)`. On the next run that flips a passing test to fail, which is the opposite of what a validator is for.

`scored` now covers only identity claims this tier can verify, with each exclusion pinned:

| Exclusion | Why | Pin |
|---|---|---|
| `relationship` / `marriage` fact types | They bear on a **second party**, so persona and linked person are deliberately different people | `test_scored_ignores_a_relationship_assertion` |
| a `matchRelatives: true` call | §2.4 returns relative scores in the **response**'s `matches`, and this tier records no responses (our own F4) — unreadable from args | `test_scored_stands_down_on_a_matchRelatives_household` |
| `results_ref: null` | §2 — no sidecar, `gedcomx1` cannot be built, correlation stands alone | `test_scored_stands_down_when_the_log_entry_has_no_results_ref` |

Excluding relationship types costs no coverage, and I pinned that too: `test_scored_still_fires_on_a_name_assertion_beside_a_relationship_one` asserts the same persona's `name` assertion still carries the demand and that the relationship entry is *not* named in the message.

**Discrimination, checked against both real runs rather than argued:**

| Run | Recorded | Re-scoped validator |
|---|---|---|
| `v1_2026-08-12_17-18-54` n7v | pass | **passes** |
| `v1_2026-08-24_22-05-46` n7v | fail | **still fires** |

## Should-fix — the pin that encoded the false-fail

Fixed, and thank you for catching it: `test_scored_fires_per_persona_not_per_run` described itself as the household case, so it cemented case 2 as correct. It now pins two **independent** identity claims with no `matchRelatives` call, and the new stand-down test pins the opposite behaviour for the household path. The docstring says why it is deliberately not a household, so the next reader does not "fix" it back.

## State

31 tests, up from 26. Mutation battery re-run with the two new exclusions added to it — all five go red: add `death`, drop `age`, broaden the tool match, stop excluding relationship types, ignore `matchRelatives`.

Nothing here touches a snapshot-tracked file: **drift stays 0**, no run bought. harness **2769 passed / 3 skipped**, packaging **578 / 578**.

Worth saying plainly — this is the second review in a row where my *fix* was the defect, and both times the mechanism was the same: I verified the assertion fired on the shape I had in mind and never checked what else it would fire on. The replay-against-a-committed-passing-run check you effectively demonstrated is the one I should have run myself before shipping the per-persona match, and I have used it here.

clack391 (approved):
Approving. Both items from my last review are fixed, and I verified the fix the way it should be verified rather than reading the summary: I replayed the real re-scoped validator over every committed run. Both passing n7v runs now pass, all three failing n7v runs still fire, and no other run false-fails. The three exclusions are each pinned in both directions, the mutation battery goes all-red, and 31 tests pass. Reproducing the blocker against the committed passing run before touching the code, and correcting the mis-pinning test with a docstring that says why it is deliberately not a household, is the part worth copying.

PR #1883, opened 2026-08-24
server: UTF-8 mode for sandbox subprocesses, fixing a Windows test hang

Emruthwill:
## Summary

- **Trivial tier.** `make test-all` hangs indefinitely on Windows at `server-test`. One env var on the sandbox subprocess spawn fixes it: `PYTHONUTF8=1` in `LocalProvider`'s env dict.
- `mock_agent` emits a summary containing U+2192. `sandbox_server` prints it to stdout, which `LocalProvider` redirects to `ws.log`, so Python encodes with the platform default — cp1252 — and raises `UnicodeEncodeError`. That kills the `_pump` task, so the `await self.broadcast(msg)` two lines below never runs: the client waits for a message that never arrives, and `_kill_server` then orphans the `agent_runner` grandchild exactly as its own comment predicts.
- No-op on Linux/macOS, where both streams are already UTF-8. That is why CI is green while every Windows run hangs.

## Review guidance

**Start here:** the choice of `PYTHONUTF8=1` over `PYTHONIOENCODING=utf-8` in `apps/server/app/sandbox/local.py`.

**Corrected after review.** This PR originally argued that two streams needed covering and that `PYTHONIOENCODING` fixed only one, because `sandbox_server`'s `Popen(..., text=True)` pipe off `agent_runner` decoded with `locale.getpreferredencoding()`. **That pipe became explicit (`encoding="utf-8"`) in #1822**, which reached this branch after it was cut, so the tiebreaker no longer holds. Re-measured against the current base:

| env | `sandbox_server` stdout -> ws.log | pipe off `agent_runner` |
|---|---|---|
| none | `UnicodeEncodeError`, exit 1 | clean (explicit since #1822) |
| `PYTHONIOENCODING=utf-8` | fixed | clean |
| `PYTHONUTF8=1` | fixed | clean |

Both now pass. `PYTHONUTF8` stays on the reason that is still true: it also sets the default for any text pipe or bare `open()` in either child, not just stdio, so a future call site that forgets an explicit `encoding=` is covered rather than becoming the next hang. The env reaches both children because `sandbox_server` passes `os.environ.copy()` down.

**Two additions from review:**

- `tests/test_sandbox_env.py` monkeypatches `Popen`, calls `ensure_server`, and asserts the env carries `PYTHONUTF8`. Nothing covered that dict before, and every CI job runs on Linux where the variable is inert — so deleting the line would have stayed green forever while Windows quietly went back to hanging. Proven to fail: with the line removed it reports `assert None == '1'`.
- `_pump` now catches `UnicodeEncodeError` around its log line. A log write must never stop the broadcast the client is waiting on. This matters beyond the env fix: `tests/test_sandbox_server.py` builds its own env and never sets the variable, so it passes today only because the turn it drives happens to be ASCII.

**Unsure about:** whether the deeper defect should also be fixed — a *logging* statement's failure currently kills the message pump. `_pump` has no guard, so any future non-ASCII on a path this env doesn't cover fails the same way: silently, as a multi-minute hang rather than an error. Left alone here to keep the diff to one line; happy to fold in a `try/except` around the `print` if a reviewer wants it.

## Test plan

Before / after on this machine, `server-test`:

```
before: hung, 0.0s CPU across a 30s sample, killed at 22 min
after:  209 passed, 1 skipped, 29.05s
```

`ws.log` after the fix, previously the last line before the traceback:

```
20:34:56 [agent] tool_result record_search: 1 match → logged as log_002 with a source + assertion
20:34:56 [agent] turn_done
```

Zero `UnicodeEncodeError` across every log, and zero orphaned `sandbox_server` / `agent_runner` processes.

- [ ] ~run-log snapshot~ — no skill directory touched.
- [ ] ~description frontmatter~ — none touched.

## Follow-on issues

**Folded in / filed instead:** Nothing filed. Found while running the PR gate for an unrelated timeline PR; folded in here rather than deferred because it is one line with a measured before/after, and it blocks the documented gate for every Windows developer. The `_pump` fragility noted under "Unsure about" is the one thing left on the table.


johnmarkpeterbrown (changes requested):
Nice work. This is a real bug, correctly diagnosed, and fixed with one line. Tracking the hang back to the dead `_pump` task — where a failed log write stops the broadcast that the client is waiting on — is the hard part, and you got it. I reproduced your before/after on Windows and the fix is causal. The suite goes from a 123-second failure plus an indefinite hang to green in 26 seconds.

Two things to change before merge.

### 1. The comment's reason is no longer true

When you merged main, you picked up a change to `sandbox_server.py` that added `encoding="utf-8"` to the pipe reading `agent_runner`. That pipe no longer uses the platform default. So the tiebreaker in your comment — that `PYTHONIOENCODING` fixes only one of the two streams — isn't true anymore.

I re-ran your table against the current base. With `PYTHONIOENCODING=utf-8` alone, both tests pass and the arrow comes through clean. No mojibake.

`PYTHONUTF8` is still the right pick, because it covers more than stdio. Just fix the stated reason.

```
You wrote:
            # sandbox_server's text=True pipe off agent_runner, which decodes with
            # locale.getpreferredencoding(). PYTHONIOENCODING fixes only the first.

Change it to:
            # PYTHONUTF8 rather than PYTHONIOENCODING because it also sets the
            # default for any text pipe or bare open() in either child, not just
            # stdio. (agent_runner's own pipe is already explicit since #1822.)
```

Please also update the table in the PR description so it matches.

### 2. Nothing stops someone deleting this line

There's no test covering the env dict this line lives in. Every CI job runs on Linux, where the variable does nothing. So if anyone removes it, CI stays green forever and Windows quietly goes back to hanging.

Add a test in this PR. Monkeypatch `subprocess.Popen`, call `ensure_server`, assert the env carries `PYTHONUTF8`. Ten lines is enough.

### And yes to the thing you asked about

You asked whether to guard `_pump`. Do it.

Your instinct was right, and there's already a case for it in the repo. The fixture in `test_sandbox_server.py` builds its own env and never sets this variable, so your fix doesn't reach it. It passes today only because the turn it drives doesn't produce any non-ASCII. Point it at a search turn and it hangs the same way.

```
You wrote:      print(f"{_ts()} [agent] {kind} {detail}".rstrip(), flush=True)
Change it to:   try:
                    print(f"{_ts()} [agent] {kind} {detail}".rstrip(), flush=True)
                except UnicodeEncodeError:
                    pass
```

A log line should never be able to kill the message pump.

---

Two notes on the draft before you decide:

- I kept the third item framed as agreement rather than a new demand, since you'd already offered it. It's listed outside the two blocking items on purpose.
- If Windows devs are blocked and you want this in sooner, I can drop item 2 to a follow-up issue filed from this PR and merge on item 1 alone. I wouldn't waive item 1.

Want me to post this as-is, or adjust the tone or scope first? I haven't posted anything to the PR yet.

johnmarkpeterbrown (approved):
Looks good! Get a senior developer on this. Thanks

jud-sdev (approved):
Went through this by running rather than reading, and it holds up. Removed the PYTHONUTF8 line and watched the new env test fail, then restored, so the test genuinely binds. apps/server is 210 passed and the encoding lint that scans apps/server is green. I also confirmed the approach is the right one and not a way around the encoding rule: the Popen in local.py redirects the child's stdout to a file handle in bytes mode, so nothing decodes in the parent — the fault is the child's own print under cp1252, which only an env var in the child can fix. It merges clean too: 7 behind main, and main has not touched either source file since the merge base.

One non-blocking note for you to weigh. The new try/except around the _pump log line has no test. On Linux the print never raises, so that except branch never runs in CI — the exact scenario it guards (a non-ASCII turn on a caller that builds its own env without PYTHONUTF8, like test_sandbox_server) stays unproven. It is correct by inspection and I would not hold the fix for it, but if you want to harden it, mock print to raise UnicodeEncodeError and assert broadcast still fires — about eight lines.

Good, well-diagnosed fix.

Issue #1112, opened 2026-08-01
Skill references: delete or wire up the 7 unreachable files, and settle the search-records/citation validation contradiction


DallanQ:
> **DECISION, ruled 2026-09-02 — the title's contradiction is settled: `search-records` is
> right.** The writer tools validate before they persist; there is no post-write
> `validate_research_schema` backstop. `citation`'s reference and its Validate step are the
> ones that change, and the tier-1 `test_write_then_validate` validator is retired with them.
> Full ruling, its cost and the alternative it beat are in the "Decision" section below. The
> same answer settles issues #2156 and #1749/#2100.
>
> **Everything else on this card is work, not a decision.** The 7 unreachable files are a
> per-file delete-or-wire call for whoever takes it. The `places-guidance.md` family is
> **split out** and decided on issue #2092, per the lead's 2026-09-01 note — do not weld it
> back on.

> **Scope note (split 2026-08-02, widened 2026-08-21):** this issue owns the
> **content decisions** about every file under `plugin/skills/*/references/` —
> per copy, is a divergence deliberate specialization or silent rot; and per
> orphan, delete or re-wire. The **mechanical lint** that follows all of it is
> issue #1115, which is blocked on this one: a lint written against undecided
> content just encodes today's drift as correct.
>
> **Not in scope: the `project-status` pointer to a file that was never written.**
> Issue #1750 ruled on 2026-08-20 that it moves there, so the whole
> `project-status` prompt change rides one paid run. The merged body below still
> contains that material because it is reproduced unedited — treat those sections
> as provenance, not as work owned here.
>
> Issue #1633 was merged in here on 2026-08-21 (its body is appended below,
> carrying the earlier merges of issues #1632 and #1716). It measured
> reachability across the same files this issue adjudicates and overlapped on 11
> of them, and each body already told the reader to go consult the other — two
> cards that each need the other are not separately assignable. The merge also
> resolves a contradiction between them: this issue asks for a per-copy verdict
> of "core" or "specialized-with-this-reason," and for 11 of the 15 copies the
> honest verdict is the third option #1633 supplies — nothing reads it.
>
> **The structural question is no longer open** — the lead chose shared-core on
> 2026-08-11, and ruled that landing this buys no paid harness re-runs. See the
> decision comment. That ruling is reinforced by the merged material: deleting a
> file no invocation can load is behavior-neutral by construction, so the
> deletion half is covered by `eval-cosmetic-skip`. The one exception that does
[...]

PR #1889, opened 2026-08-25
timeline: four craft corrections from the senior review (#1820)

### LINKED ISSUE #1820: Timeline deep dive - comments and edits
> **Reviewed 2026-08-24 before junior handoff.** Added the second file to Touches, a per-finding site table checked against `main`, implementation notes (keep the travel figures as maxima; do not build a second gap taxonomy), the paid-run cost, what nothing checks, and an acceptance checklist.

**Touches:** packages/engine/plugin/skills/timeline/references/timeline-analysis-guide.md, packages/engine/plugin/skills/timeline/SKILL.md

Senior genealogist's prose review of the timeline analysis guide, from the deep dive
closed as issue #1655. Same shape as issue #1798 for search-full-text. Costs one paid
`timeline` run.

## Where each finding lands, checked against `main` 2026-08-24

| Finding | File | Site |
|---|---|---|
| "moved away or died" is only two of the possibilities | `references/timeline-analysis-guide.md` | the `Examples:` block, line 61-72 (the flagged sentence is line 64) |
| All the examples are narrow — expand? | same file | same block; **read "Notes" below first — the list you would be adding already exists at line 73** |
| Travel is constrained: mode of travel, wagon trains 10-15 mi/day | same file | the era travel-speed list, lines 146-150 |
| Same census year, two locations can be one legitimately twice-enumerated family | same file | `### When Timelines Reveal Two People`, line 242 |
| …and the same claim in the skill body | `SKILL.md` | lines 235-242 — "including the same person enumerated in two different states in one census year" |

## Notes for whoever implements it

- **Lane and scope.** All of this is craft prose in this skill's own files — lane 3/4 in
  `docs/architecture.md` § "The lane rule". No tool or eval defect is owed. Both files are
  inside `skills/timeline/`, so they ride one paid run; do not widen to another skill's
  directory, which buys a second run and a second annotation pass.

- **Finding 1 and 2 are one edit, and most of what you would add already exists.** The
  four possibilities the reviewer is missing are enumerated nine lines below the examples
  he flagged: `### Interpreting Gaps` (line 73) already classifies a gap as migration
  (77), life event — military, imprisonment, institutionalization, at sea (82), record
  loss (87), or identity confusion (92). So the defect is that the *example* at line 64
  closes the list ("this is evidence they likely moved away or died") before that section
  opens it. **Fix the example so it does not close the list, and point it at the four
  classes — do not build a second taxonomy in the Examples block.** Two lists of the same
  thing in one file drift, and this one is read on demand, so the second copy costs
  context every time it is loaded.

  Apply the same test to the other two examples (the England death search, the missing
  marriage): each already names one alternative — widen the wording only where a
  possibility is genuinely absent from lines 73-95, and say in the PR which ones you
  judged already covered.

- **Finding 3 (travel): the numbers are a ceiling for a feasibility test, so do not lower
  them.** `SKILL.md:235-242` makes geographic/travel feasibility **this skill's own
  check** and sends the agent to this list for the threshold, and the rubric's
  `Geographic feasibility` dimension fails a run that "called a feasible pair
  infeasible" (`eval/tests/unit/timeline/rubric.md`). The reviewer's point is that a
  maximum is not a typical speed — a wagon train made 10-15 miles/day where the list
  allows 30-50. **Keep the existing figures as the infeasibility threshold and add the
  modal speeds as a separate, labelled line** (walking / horse / buggy / wagon train),
  stated as what makes a pair *doubtful* rather than impossible. A pair inside the
  maximum but far outside the typical speed is a question for the reply, not an
  infeasibility call. Say in the PR that you kept the ceiling.

  The reviewer's other half — overland travel stayed the common mode well after rail
  existed — is a caveat on how to *read* the 1830-1870 and 1870-1920 rows, not a change
  to them.

- **Finding 4 (double enumeration) has two sites, and the skill body is the one that
  matters.** `SKILL.md:240-241` tells the agent to treat "the same person enumerated in
  two different states in one census year" as a travel-feasibility case measured against
  the census *year*. That is the claim the reviewer is correcting: families were
  legitimately enumerated twice, and the date on the page is usually the date the
  enumerator visited, so the two enumerations can be months apart rather than
  simultaneous. Correct it where it is stated — one clause in `SKILL.md` Step 5 naming
  the two checks (family composition identical; use the enumeration dates on the pages,
  not the census year), and the fuller craft explanation in the guide at line 242.
  Leaving the body's version uncorrected while the correction sits in a reference file
  is the failure mode issue #1112 describes: a correct fix in a file the agent may not
  open.

  Keep the `SKILL.md` addition to one clause — `CLAUDE.md` § "Skills" bans explanatory
  prose there; every line is a billed prompt token on every invocation.

- **No existing text covers the enumeration-date point.** Checked 2026-08-24:
  `grep -rn -i "census day\|enumeration date\|enumerator visit" packages/engine/plugin`
  returns three hits, none of them about dating an enumeration. You are writing this
  from scratch, not duplicating it.

- **Nothing in the repo contradicts these four corrections.** Checked against
  `docs/adrs/ADR-0009-refuted-agent-design-claims.md`, the `Alternatives considered`
  rows of the ADR index, and `docs/specs/guardrail-enforcement-spec.md` § "Options set
  aside" — no hit.

## Cost

Editing anything under `packages/engine/plugin/skills/timeline/` flips this skill's run
log inactive and arms `.github/workflows/check-runlogs.yml`, which blocks the PR. The
newest run log (`eval/runlogs/unit/timeline/v1_2026-08-21_14-34-37.json`) is **active on
`main` today** — verified 2026-08-24 with `diff_snapshot_vs_disk`, which returns clean —
so this edit is what buys the run. `eval-cosmetic-skip` does **not** apply: these are
content changes, not rewording.

Read the machine cost from `totals.total_cost_usd` and `totals.wall_clock_ms` in the
newest run log rather than quoting a figure forward. On 2026-08-24 those fields read
**$2.67 and ~17 minutes**. The CRUD-UI annotation pass is the human half and the larger
one.

## What nothing checks

**No automated check reads the contents of either file, and the guide is rarely opened.**
Across the three timeline run logs that record built-in tool calls, the agent opened
`references/timeline-analysis-guide.md` in **1 of 18 positive-test executions**
(`ut_timeline_002` in `v1_2026-08-16_19-22-07`; zero times in either 2026-08-21 run —
every other `Read` was `research.json`). Reproduce with
`grep -c timeline-analysis-guide eval/runlogs/unit/timeline/v1_*.json`: the baseline is 1
per file (the snapshot key), so a count above 1 is a `Read`. What would move it: a
pointer change in `SKILL.md`, or a scenario that forces the guide open.

Two consequences, both for the PR body rather than the code:

1. **A flat eval score is not evidence the edit worked, and a moved score is not evidence
   it did.** Record which dimensions moved and say you are not attributing them.
2. The only graded surface any of this touches is `Geographic feasibility`, exercised by
   one test (`ut_timeline_005`, a transatlantic pair). Nothing in the suite exercises a
   same-census-year double enumeration or a wagon-train-scale distance, so a wrong call on
   findings 3 or 4 ships green. The direction-preserving check above is a hand check;
   the PR body is where you record that you did it.

Do **not** fix the read rate in this PR by adding pointers or folding the guide into the
body — that costs context on every invocation and is a separate decision. If you think it
should happen, comment on issue #1112, which owns reference-file reachability.

## Acceptance

- [ ] All four corrections land, in the sites named in the table above; the `SKILL.md`
      clause is one sentence and carries no rationale.
- [ ] The PR body states that the travel figures were kept as maxima, and which
      negative-evidence examples you judged already covered by `Interpreting Gaps`.
- [ ] `make eval-skill SKILL=timeline` re-run. Required, not optional.
- [ ] The new run log **and its `.ann.json`** committed, with a correction on every
      dimension of each test named in the new run's own `review_sample` (5 per run;
      re-drawn each run — do not assume the current set). Write the annotation **only**
      through the CRUD UI (`make eval-ui`); a hand-written one crashes the CI gate.
- [ ] `git add` the run-log deletions the harness makes when it writes — it keeps the
      newest 5 candidates (`DEFAULT_KEEP_CANDIDATES` in `eval/harness/harness/versioning.py`)
      and timeline holds 6 today.
- [ ] `make test-all` (the PR template's gate).

**Timeline
Timeline-analysis-guide
Negative Evidence and Gap Analysis
What negative evidence means**
Under examples - "This is evidence they likely moved away or died ...." Those are two of the possibilities. Should edit to accommodate other possibilities.

All the examples are extremely narrow. The possibilities raised are likely but not the only possibilities. Does this need to be expanded?

**Fundamental Assumptions**
Travel is constrained - might want to acknowledge that Pre-1830 overland travel speed was still the most common way to travel even after rail travel was possible. Also, mode of travel did matter too - walking vs horse vs buggie vs wagon. Wagon trains often traveled 10-15 miles a day.

**When Timelines Reveal Two People**
last bullet point - Same census year, two different locations. - It IS possible for families to be enumerated twice. Make sure that the family composition is identical. Also check what date the census was actually recorded. Often, they put down the date the enumerator actually visited the area. If the time frame and distance assumptions work, then they could be the same family.

Emruthwill:
## Summary

- **Normal tier.** The senior genealogist's four prose corrections to the timeline analysis guide (issue #1820), at the four sites the issue's table names. Craft prose in this skill's own two files — lane 3/4, no tool or eval defect owed.
- **The travel figures were kept as maxima.** The era rows are byte-for-byte unchanged and remain the infeasibility threshold. Modal overland rates (walking / wagon train / buggy / horseback) are added as a separate labelled block, stated as what makes a pair *doubtful* rather than impossible, with an explicit instruction to raise it in the reply and not call it an infeasibility.
- **Negative-evidence examples: one widened, one judged already covered.** The census example was the one that closed the list, and now points at `Interpreting Gaps` instead of restating it. The England death search was widened — it named no alternative at all, and record loss (class 3) is the classic way a "no results" search misleads. **The missing-marriage example I judged already covered**: it names two alternatives ("occurred elsewhere, or the couple was not legally married") spanning the migration class plus a social explanation.
- Double enumeration corrected at both sites, with the `SKILL.md` half as one sentence carrying both checks and no rationale.

## Review guidance

**Start here:** the modal-speed block in `references/timeline-analysis-guide.md`. It has to add craft without becoming a second threshold — the rubric's `Geographic feasibility` dimension *fails* a run that "called a feasible pair infeasible", so the block is scoped explicitly to pairs **inside** the era maximum. `ut_timeline_005`, the only test that exercises `place_distance`, is a transatlantic pair at ~480 mi/day against an 1830-1870 ceiling of ~150 — far outside the maximum, so the new block cannot reach it. It passed.

**Unsure about:** nothing blocking. Worth a second opinion on whether the England-death-search widening is one clause too many, given the issue's instruction to widen only where a possibility is genuinely absent.

## Plan

**Didn't change:** the guide's read rate. Across the timeline run logs that record built-in tool calls the agent opens `timeline-analysis-guide.md` in a small minority of positive-test executions, so most of this diff lands in a file that is rarely read. Per the issue, fixing that by adding pointers or folding the guide into the body is a separate decision that costs context on every invocation — it belongs to issue #1112, which owns reference-file reachability.

**Acceptance check:** none exists, and that is the honest answer. No automated check reads the contents of either file. Nothing in the suite exercises a same-census-year double enumeration or a wagon-train-scale distance, so a wrong call on findings 3 or 4 would ship green. The direction-preserving check on `ut_timeline_005` described above is a hand check; recording it here is the only place it exists.

## Test plan

- [x] `make eval-skill SKILL=timeline` — **10 pass of 10**, $2.45, ~12 min. Run log and `.ann.json` committed, with the harness's four pruned candidates `git add`-ed (5 candidates now = `DEFAULT_KEEP_CANDIDATES`).

**Scores moved, and I am not attributing the movement to this edit.** Against the previously-active run log (`v1_2026-08-21_14-34-37`), 57 shared dimension cells, **2 moved** — `ut_timeline_008` Correctness and Completeness, 1 → 3, which also flipped that test fail → pass. `008` is a routing negative: in both Aug 21 runs it answered inline (657 chars, no skill invoked) and failed; here it routed to `person-evidence` and passed. This diff touches neither routing logic nor the `description` frontmatter, so that is run-to-run routing variance, not evidence the edit worked. `ut_timeline_010` has no counterpart in the prior run — it was not in that 9-test run at all.

**Annotation.** Every dimension of the five tests in this run's own `review_sample` (`001, 002, 005, 009, 010`). 30 corrections, 4 overrides: `009` and `010` Correctness/Completeness, judge 1 → human 3. Both are negative tests that routed correctly and whose transcripts are empty because the harness's routing short-circuit denies the sub-skill launch and stops the run the moment the correct skill fires. The judge prompt reserves a 1 for a skill that carried out the task anyway, and forbids lowering the base dimensions for a terse decline or for performing the handoff by invoking the correct skill.

- [ ] `make test-all` — **blocked on #1883, not on this branch.** `server-test` hangs indefinitely on Windows here: `mock_agent` emits U+2192, `sandbox_server` prints it to a redirected stdout under cp1252, and the `UnicodeEncodeError` kills the `_pump` task so `broadcast()` never runs. #1883 fixes it with `PYTHONUTF8=1` and is green. Every other suite passes on this branch, and CI's `server-pytest` runs on Linux where the bug cannot fire.

## Follow-on issues

**Folded in / filed instead:** Nothing filed. The Windows `server-test` hang was found running this PR's own gate and went to its own branch (#1883) rather than being folded in — it is a control-plane Python fix that pulls a different reviewer under CODEOWNERS, while this PR is confined to `skills/timeline/`.

Closes #1820

---

## Review round 2 — the clause fix, and what the re-run showed

**The clause now carries both conditions**, as requested. The measurement had no purpose in the old wording and identical composition was treated as sufficient; both are fixed, still one sentence.

**Option B taken: the suite was re-run.** `v1_2026-08-25_21-22-08` — 9 pass / 1 fail, $2.59, 9.3 min, 0 transient retries. Annotation covers every dimension of the run's own re-drawn `review_sample` (002, 003, 004, 006, 007) plus 001: 36 corrections, no overrides.

### The failure is not this PR's doing

`ut_timeline_008` is a routing negative that has failed **4 of the 5 recorded runs**, including all three that predate any edit on this branch:

| Run | `ut_timeline_008` |
|---|---|
| 2026-08-16 | fail |
| 2026-08-21 14:07 | fail |
| 2026-08-21 14:34 | fail |
| 2026-08-24 | pass |
| 2026-08-25 21:22 (this) | fail |

Here the agent did the timeline work inline — 11 turns, 3008 chars, `skills_invoked: []` — instead of declining and routing to `person-evidence`. A genuine miss, scored correctly; no `routing_negative` advisory fired because no correct skill was invoked. Worth noting the 10/10 in the first round was the outlier, not the baseline.

### One finding a green run would have buried

An earlier run of this same commit (`v1_2026-08-25_20-45-31`, deliberately not committed) failed `ut_timeline_005` instead. It is worth recording because the mechanism touches the clause under review:

> **Apparent transatlantic near-infeasibility (calendar artifact):** … a 1850-era steamship covered ~240 km/day, making the crossing ~22 days minimum — **impossible in 7 days**. However, **1 June 1850 was the administrative reference date Congress set**, not the day this household was actually visited … **making a May departure from Mayo and a late-summer enumeration entirely feasible.**

The agent computed the infeasibility correctly and then dissolved it, using enumeration-date-versus-census-year reasoning — and de-precisioned the census event to bare `1850` as a consequence, which is what tripped `test_events_chronologically_ordered`. `ut_timeline_005` exists to make the agent *flag* that pair.

**Rate: about 1 in 5.** Three targeted re-runs of `ut_timeline_005` on this commit (`--test ut_timeline_005`) all passed, all kept `1850-06-01`, and all flagged the infeasibility. So this is not a regression and I am not treating it as blocking.

But it is a real property of the wording: sitting in the geographic-feasibility bullet, "measure from the enumeration dates rather than the census year" can read as a general licence to explain away any census-adjacent infeasibility by asserting the visit happened later. The agent's genealogy is not even wrong in isolation — which is what makes it effective at dissolving a finding the skill exists to surface. Flagging for the reviewer's judgement rather than rewording unilaterally, since this is his approved wording.


johnmarkpeterbrown (changes requested):
Nice work. The body is the most rigorous one I've read this month — every figure
in it checks out, the run log is active, the annotation covers every graded
dimension, and you were right not to attribute the `ut_timeline_008` movement to
your own edit. All four corrections landed where the issue said they should, and
you kept the era rows untouched. One thing to change before this goes in.

## The SKILL.md clause is missing half of its own rule

The guide gets this right. It says to split when **either** the composition
differs **or** the dated pages leave no time to travel. The SKILL.md sentence
keeps the first condition and drops the second.

Read the sentence as the agent will: it is told to measure from the enumeration
dates, and then never told what to do with the measurement. Separately, it is
told that identical composition *is* a double enumeration — with no condition
attached. So two same-named households with matching composition, enumerated
forty days apart on opposite sides of an ocean, get merged instead of split.

That matters here specifically because it lands in the file the agent reads on
every invocation, while the correct version sits in the file it usually doesn't
open. It's the exact failure mode the issue flagged for this finding.

```
You wrote:    and treat an identical family composition as a legitimate
              double enumeration rather than two people.

Change it to: and treat it as a legitimate double enumeration rather than
              two people only when the family composition is identical and
              the dated pages leave time to travel.
```

Still one sentence, still no rationale, eleven bytes longer.

## Your "unsure about": the England widening stays

It isn't one clause too many. `Interpreting Gaps` class 3 is about a gap in the
record over time. Your new clause is about how far a nil search result can be
trusted, which is bounded by the completeness of the series you searched.
Different thing, and genuinely not stated anywhere in that section. Keep it.

The contrast with the census example is right too — that one points at the
taxonomy instead of restating it, which is what the issue was asking for.

## About the re-run

The fix touches a snapshot-tracked file, so it will flip the run log inactive.

[CHOOSE ONE]

  A) Don't buy a second run for this. Push the one-line change and tell me when
     it's up — I'll apply `eval-cosmetic-skip` on that commit. The label has to
     go on after the push, since a new commit voids it. Worth knowing why: no
     test in the suite exercises a same-census-year double enumeration, so a
     second run cannot tell either of us whether this fix is right. It would
     only satisfy the freshness gate.

  B) Push the one-line change and re-run `make eval-skill SKILL=timeline`, with
     a fresh annotation on the new sample. The correction changes what the agent
     does, so it buys a run under the normal rule.

## Not blocking, just noting

The modal rates don't name a default mode, so "far outside these rates" sits
between walking and horseback — a pair at 28 miles a day is inside one and
outside the other. I'd leave it. The block already forbids calling such a pair
an infeasibility, so the worst case is a question raised that didn't need to be.

Approving once the clause is fixed.
````

johnmarkpeterbrown (approved):
Looks good! Please get a senior developer on this. Thanks

jud-sdev (approved):
Went through this line by line against the issue's four-correction table, and each one lands where it should with nothing extra riding in. The census example now points at Interpreting Gaps rather than closing the list; the England death search gains the record-loss clause while the missing-marriage example is correctly left alone as already covered; the era travel rows are unchanged as the infeasibility ceiling with the modal speeds added as a separate doubtful-not-impossible block; and the double-enumeration fix lands at both sites, with the SKILL.md half kept to one scoped instruction and no rationale.

Mechanics are clean: the run log and its .ann are committed, the runlogs check is green so the snapshot matches disk, the frontmatter is untouched, and no code, schema, auth or ADR is involved. The one 9/10 failure (ut_timeline_008) is a routing negative that predates this branch and touches neither routing nor the description, so it is variance, not this edit.

The one thing worth a genealogist's eye — the enumeration-date wording reading as a licence to dissolve a census-adjacent infeasibility, which the author flagged after seeing it on ut_timeline_005 — is a craft call, and it has one: the wording is correctly scoped to a same-year household double enumeration, and a senior genealogist approved it with that flag in hand. Good work.

PR #1890, opened 2026-08-25
Record the tool response and the grading mode the harness already knew (#1000)

### LINKED ISSUE #1000: Unit run logs drop two things the harness already has: the tool response, and whether the judge dimensions gate the outcome
> **Reviewed 2026-08-24 before junior handoff. Read this before the comments below.**
>
> **The 2026-08-16 re-scope comment is stale — do not follow it.** It proposes
> extending `apply_routing_deference`'s `[deterministic-deference]` rationale
> prefix to `grade_on_invariant`. That function was **deleted** on 2026-08-16 in
> commit `c6a25de6` (PR #1686, "Stop overriding the judge on routing
> negatives"), and deleted on evidence: replaying its guards over the 121
> committed run logs found 24 floor-eligible cells and a human confirmed the
> judge's 1 on **20 of them**. Do not re-implement any mechanism that rewrites
> or floors a judge rationale. What survived that PR is the *warning* —
> `flag_routing_negative_judge_fail` in `orchestrator.py`, registered in
> `run_tests.py::_JUDGE_WARNING_KINDS` — and it is the shape to imitate if you
> want a second signal, not the floor.
>
> **Build what this body proposes:** the two schema fields, as re-verified in
> the 2026-08-23 comment.

### Retention rule — decided (lead, 2026-08-24): live and unmatched only

Record `response` only for calls whose content exists nowhere else: `matched.kind`
of `"live"` or `"none"`. Omit it for `matched.kind == "predicate"`, whose response
is exactly recoverable from `response_fixture` plus `matched.index` at that
commit.

Chosen over recording every response: that would add ~16% to the largest current
run log (147 KB against 913 KB, measured 2026-08-24) to re-store bytes git
already holds, and run-log size is the thing `schema_version` 3 was introduced to
cut. The named use case — calibrating the warn-only person_evidence guardrail
(issue #1550) — reads `research_append`, a live tool, so it is fully served.

Write this rule into `docs/specs/unit-test-spec.md` §10, on the `tool_calls`
entry, as part of this PR: state that `response` is present for live and
unmatched calls and absent for fixture-matched ones, and name
`response_fixture` + `matched.index` as the recovery path for the rest. The
acceptance assertion must be on a `live:` call.

## What happened

Reviewing #929, a reviewer opened the `search-records` run log, found
`ut_search_records_005` reporting `outcome: "pass"` alongside
`Correctness = 1` and `Completeness = 1` — with judge rationales
describing an outright routing failure ("The routing that should have
occurred did not happen") — and concluded the harness miscomputes
`outcome` for negative tests, so "aggregate pass/fail counts on negative
tests may not be trustworthy right now."

That conclusion is wrong, and the run log is what made it look right.

## The outcome is correct

`ut_search_records_005` (`eval/tests/unit/search-records/negative-research-plan-request.json`)
sets `negative.grade_on_invariant: true`. Per
`orchestrator.py::_compute_outcome`, such a test returns `pass` on the
strength of its tag-gated validator alone; judge dimensions are
explicitly diagnostic and do not gate. The invariant validator ran and passed. Note since filed: issue #1788 measured
`ut_search_records_005` as one of 7 `grade_on_invariant` tests whose gated
validator asserts about a tool the run cannot register, so one of its clauses
cannot fire. That repair is issue #1788's, not this card's — but do not quote
this section as proof the validator is sound.

Note also that this test's own 1/1 no longer reproduces: a per-test
`judge_context` OVERRIDE landed 2026-08-13 (PR #1564) instructing the judge to
score the project-status detour 3/3, and both 2026-08-21 search-records run logs
show Correctness 3, Completeness 3. **The general shape is still live and is what
this card fixes:** across all committed unit run logs, **10 rows on
`grade_on_invariant` tests carry `outcome: "pass"` beside a judge dimension
scored 1** — `ut_citation_012` (3 rows, latest 2026-08-20),
`ut_conflict_resolution_010` (5 rows, latest 2026-08-19), `ut_research_plan_003`,
`ut_search_wikipedia_007` (measured 2026-08-24 by joining the 13 test JSONs
carrying `negative.grade_on_invariant` to every
`tests[].outcome_summary.aggregated_dimensions`). The per-test fix was applied to
one test; four skills still render the contradiction with nothing marking it
intentional.

The test's own `judge_context` says so in its first line:

> This is a `grade_on_invariant` negative test: PASS/FAIL is decided
> solely by the deterministic no-search-no-write validator … The judge
> runs base-only and diagnostically here — its scores do not decide the
> outcome.

The 1/1 scores are the judge noting the documented "project-status
detour" routing imperfection, exactly as that same `judge_context`
anticipates. Working as designed.

## The real defect: the run log doesn't say any of that

Nothing in the per-test run-log entry marks the dimensions as
diagnostic-only. `outcome`, `outcome_summary.aggregated_dimensions`, and
the per-run `judge.dimensions` are rendered identically to a positive
test's, where a `1` genuinely does force a fail. A reader has to already
know `grade_on_invariant` exists, then go read the test JSON, then read
`_compute_outcome`, to tell the difference between "designed" and
"broken."

The cost is not hypothetical: it produced a confident, incorrect
correctness claim inside a PR approval, and the next reviewer to open a
`grade_on_invariant` run log will re-derive the same wrong conclusion.

## Proposed fix

Surface the gating mode in the run log so the contradiction reads as
intentional. Cheapest version — one field on the per-test entry:

```json
{
  "test_id": "ut_search_records_005",
  "test_type": "negative",
  "grading_mode": "invariant",
  "dimensions_gate_outcome": false
}
```

Negative tests already split three ways in `_compute_outcome`
(`grade_on_invariant`, empty `correct_skill` where dimensions *do* gate,
non-empty `correct_skill` where they don't), so a single honest field
covers all three and the CRUD UI can grey out or annotate non-gating
dimensions.

Whether the schema change is worth it versus a UI-only affordance is
open — but a reader of the raw JSON is the one who got misled here, so a
UI-only fix would not have prevented this.

## Not in scope

The `outcome` values themselves are correct. No re-run or re-grade of
existing logs is needed, and #929's aggregate counts are trustworthy as
reported.

## References

- `eval/harness/harness/orchestrator.py::_compute_outcome` (the
  `grade_on_invariant` early return and the three-way negative split)
- `eval/harness/validators/test_search_records.py:158`
  (`test_no_search_or_writes_on_planning_request`, tag-gated on
  `no-search-no-write`)
- `docs/specs/unit-test-spec.md` §6–7
- Observed in `eval/runlogs/unit/search-records/v1_2026-07-30_11-10-24.json`
  on branch `friedrich-weber-daughter` (#929)

---

## Merged in from issue #1816 (2026-08-23): the tool response the orchestrator already has and drops

Filed separately, folded here because it is the same document, the same two mirrored schema
files, the same spec section and the same regenerated Zod types. Two PRs each adding a property
to `run-log.schema.json` would conflict on rebase and each re-answer the same strictness
question for no gain.

### What is wrong

Every unit run log records tool calls without the tool's response. A `tool_call` entry carries
`tool`, `args`, `expected_args`, `matched`, and `response_fixture` — and `response_fixture` is
a fixture **name**, not the content. So no post-hoc analysis of a committed run log can answer
"what did the skill actually see?"

Measured on the committed corpus: **2,812 tool calls across the 37 August-2x unit
run logs, zero with a captured response** (measured 2026-08-24; reproduce with a
walk of `eval/runlogs/unit/*/v1_2026-08-2*.json` summing
`tests[].runs[].output.tool_calls` and counting entries with a `"response"` key).
The count grows with every committed run; the zero is the load-bearing half.
Of those 2,812, **1,998 are `live:` calls** whose content exists nowhere else on
disk. Benter reported the same finding over a wider window (284 calls) as
a comment on the person-evidence deep dive; the conclusion is identical.

### Do not re-derive this — the response is already captured, then thrown away

This is a smaller fix than it looks, and the first instinct (teach the mock to record
responses) is wrong because it already does.

- `eval/harness/harness/mock_mcp.py:461` — `entry["response"] = response` writes the full
  response into the call-log entry.
- `eval/harness/harness/mock_mcp.py:309` — the function's own docstring lists `"response"` as a
  call-log key, alongside `"response_fixture"`.
- `eval/harness/harness/orchestrator.py:606-613` — building the run log's `output.tool_calls`,
  this copies **five** named keys per entry and omits `response`.

The data exists in memory at the moment the run log is written. It is dropped by a projection,
not missing from the capture.

### Why it matters

The warn-only person_evidence guardrail that shipped in `research_append` (issue #1550, merged)
cannot be calibrated from the unit tier, because calibrating it means reading what the tool
returned and the run logs do not hold it. The same limit applies to every "did the skill act on
what it was given?" question anyone asks of the committed corpus.

### Settled — do not re-derive these

1. **The omission was not deliberate.** `mock_mcp.py:309`'s own docstring already
   lists `"response"` as a call-log key, and all six `call_log.append` sites
   (`mock_mcp.py:274, 461, 609, 678, 747, 824`) set it. Nothing anywhere states a
   size rationale. Treat it as not-yet-needed and remove the projection.
2. **Do not bump `schema_version`.** It is `"const": 3`, and
   `skill_latency_report.py:147` / `judge.py:945` already branch on pre-3 shapes.
   Adding an optional property does not change the envelope shape.
3. **Add all new fields as optional, not to any `required` array.** `test_entry`
   and `$defs/tool_call` both list every current property as required; adding
   there would make each of the 121 committed run logs invalid the moment anyone
   re-validates one.
4. **The size worry has a shipped answer.** `runlog.py`'s
   `_SIDECAR_TEXT_THRESHOLD = 100_000` already spills oversized `text_response`
   payloads to `runs/<run_id>.text.md`. If a captured response needs the same
   treatment, extend that path — do not invent a truncation rule.

### Acceptance check for this half

**Free, in `make harness-test` — no paid run.** `write_run_log` validates against
`run-log.schema.json` before writing (`runlog.py:502, 535`), so a field written
without the schema edit hard-fails on its own. Add a pytest over the
`output.tool_calls` projection with a synthetic call log, **and prove it fails
against the current five-key projection before you commit it** (CLAUDE.md, "A new
lint must be proven to fail"). Do not spend a `make eval-skill` run on this.

## The shared plumbing, and why these are one card

Both halves add a field the harness already knows to a document that is validated strictly:

- `$defs/tool_call` and the top-level object in `run-log.schema.json` are both
  `additionalProperties: false`, so neither field can be added without a schema edit.
- The schema exists in **two byte-identical mirrors** — `docs/specs/schemas/run-log.schema.json`
  and `packages/schema/schemas/run-log.schema.json`.
- `eval/app/scripts/gen-zod.ts` regenerates the CRUD UI's types from it.
- Both need the same section of `docs/specs/unit-test-spec.md` § Run Log Format updated.

The two write sites differ — the grading-mode field goes on the per-test entry in
`eval/harness/harness/runlog.py:379-392`, the response goes in the `output.tool_calls`
projection in `eval/harness/harness/orchestrator.py:606-613` — but everything downstream of
them is shared.

## Cost

**Neither half costs a paid eval run.** `eval/harness/**` is not in the run-log snapshot
(`snapshot.py`, `build_snapshot`), so a harness change flips no skill's run log inactive. Do
not add `eval-cosmetic-skip` and do not fold this into a deep dive.

Independent of the open harness PRs: it does not touch `mock_mcp.py`'s behaviour, so it does
not collide with PR #1723 or PR #1766, which both edit that file.

**Collisions — check before branching.** PR #1833 (search-wikipedia deep dive)
edits **both** `eval/harness/harness/orchestrator.py` and
`docs/specs/unit-test-spec.md`; PR #1764 edits `orchestrator.py`; PR #1781 edits
`unit-test-spec.md`. Land after whichever of those merges first, or rebase onto
it. PR #1723 is closed — the mock_mcp.py work moved to PR #1822, which this card
still does not collide with.

**Overlap on the motivating test.** Issue #1788's ruling repairs
`ut_search_records_005` by editing that test's `mcp_fixtures`, taking
search-records' paid eval slot. That is a different card and a different file —
do not do it here, and do not wait for it.

**Touches:** eval/harness/harness/orchestrator.py (the `output.tool_calls`
projection — five keys, near the `response_fixture` line),
eval/harness/harness/runlog.py (the per-test entry dict returned after the
`runs_block` loop), docs/specs/schemas/run-log.schema.json,
packages/schema/schemas/run-log.schema.json (byte-identical mirror — enforced by
`eval/harness/tests/unit/test_schema_mirrors.py`),
docs/specs/unit-test-spec.md (§10 Run Log Format: the `tool_calls` shape sentence
around line 1151 and the run-log JSON block around line 1333; the illustrative
mock excerpt near line 1818 has already diverged from the real code — leave it),
eval/harness/tests/unit/ (the two new assertions).

**Not touched, despite an earlier draft of this line:**
`eval/app/scripts/gen-zod.ts` needs no edit. It regenerates on postinstall and
prebuild into gitignored `eval/app/lib/schema/`, and nothing in `eval/app`
imports `RunLogSchema`. There is no CRUD-UI work in this card.

**Do not edit `eval/tests/unit/search-records/negative-research-plan-request.json`
or any other test JSON.** That directory *is* inside the run-log snapshot
(`snapshot.py::build_snapshot`), so touching it flips that skill's run log
inactive and costs a paid `make eval-skill` run plus a genealogist annotation
(~$8–12, 45–65 min) that `check_runlogs.py` blocks the PR until it lands. The
harness half of this card costs nothing — keep it that way.

aghadiayeamayanvboernest:
Closes #1000 (both halves — the run-log display half and the merged-in #1816 response half).

### The tool response

`mock_mcp.py`'s call-log docstring lists `"response"` as a key and all six `call_log.append` sites set it. The orchestrator's `output.tool_calls` projection then copied five named keys and left it behind — dropped by a projection, not missing from the capture.

Consequence: no post-hoc analysis of a committed run log can answer *"what did the skill actually see?"*, so the warn-only `person_evidence` guardrail (#1550) cannot be calibrated from the unit tier at all. Measured: **2,812 tool calls** across the August-2x run logs, **zero** with a response.

**Kept where the content exists nowhere else.** `live` and unmatched (`none`) calls, and — corrected in review — any fixture-matched response the mock ENRICHED after selecting it (`staged`/`ranked`/`rankingSkipped`), which is not reconstructable: `staged.resultsRef` comes from `randomUUID()`. Up to 420 committed calls are in that state. A *plain* predicate match is omitted, being recoverable from `response_fixture` + `matched.index`. **Size trade, both halves:** `live` is the majority (4,309 of 7,718) and is retained with no per-field cap; what bounds growth is unit retention keeping the newest 5 candidates per skill. The largest committed run log is 1.8 MB.

### The grading mode

A `grade_on_invariant` negative renders `outcome: "pass"` beside a dimension scored 1 **exactly** as a positive test would, where that 1 forces a fail. Telling those apart meant knowing the flag exists, finding the test JSON, then reading `_compute_outcome`. That ambiguity produced a confident, incorrect correctness claim inside a PR approval.

`grading_mode` (`dimensions` | `invariant` | `routing`) and `dimensions_gate_outcome` now say which of the three branches decided the outcome.

## The part I'd point a reviewer at

A field that *claims* the dimensions do not gate is worthless if `_compute_outcome` later starts gating on them. So it is not asserted against a hand-written table.

`test_grading_mode_matches_what_compute_outcome_does` drives the **real** `_compute_outcome` twice per branch — once clean, once with the same dimension scored 1 — and asserts the outcome moves **iff** `dimensions_gate_outcome` says it should. Both drift directions were mutated and confirmed red:

| Mutation | Result |
|---|---|
| Field claims `invariant` tests are gated | **2 red** |
| `_compute_outcome` actually starts gating `invariant` tests | **1 red** |
| Projection helper reverted to the old five keys | **2 red** |
| Projection **call site** reverted | **1 red** (added in review — it survived before) |
| Main path stops passing the grading fields | **1 red** (added in review — it survived before) |

## Backward compatibility, verified rather than argued

Both fields are optional and **omitted rather than written as `null`** — `null` would assert "this test has no grading mode", which is false for every test; absent means "this run log predates the field", which is what a reader needs.

Nothing was added to any `required` array, so I re-validated every committed run log: **122 of 123 pass**. The single failure is `schema_version: 2` against the v3 `const`, and it fails identically on main — pre-existing, not this change. (An earlier draft said "228 of 230"; that glob counted 111 `.ann.json` files as run logs.)

No `schema_version` bump: adding optional properties does not change the envelope shape.

## One thing I fixed that was not asked for

The spec's `matched.kind` said `predicate | none`. The code has always emitted `live` too — and `live` is the exact kind this retention rule turns on, so leaving it would have documented the new rule against a value the spec claims does not exist.

## Cost

**No paid eval run.** `eval/harness/**` is not in the run-log snapshot, so no skill's run log flips inactive. No test JSON touched — I deliberately did not go near `eval/tests/unit/search-records/negative-research-plan-request.json`, which is #1788's repair and would have cost a `make eval-skill` run plus annotation.

The doc-links ratchet is unchanged at 5 refs: every reason above is written as a mechanism, not a ticket number.

Gate: `make test-all` — all checks passed.

## Rebase note

#1764 and #1781 are open and also touch `orchestrator.py` / `unit-test-spec.md`. The spec edits here are in different sections, so whichever lands first should rebase cleanly.


T-FEH (changes requested):
Reviewed on the merged head (`b4e71223` — your `c8da2b3f` plus a clean merge of main). Harness 2818 passed, 2 skipped. Gate green, no snapshot paths, no paid re-run. Doc-links ratchet passes.

## The anti-drift test is the best thing here, and it holds up

I mutated it rather than trusting the table, and all three of your claimed mutations reproduce exactly — plus three you didn't claim (out-of-scope negative stops gating; field claims `routing` is gated; both `dimensions` branches flipped) all red too. It drives the real `_compute_outcome` twice per branch rather than asserting a table, and mutation (b) showed it is the **only** guard in all 2,818 tests that catches `_compute_outcome` starting to gate invariant negatives. That is a genuinely valuable test.

The motivating measurement holds as well: zero committed tool calls carry a response today, and `_tool_call_entry` implements the retention rule exactly as described. Both schema mirrors are byte-identical to each other, nothing was added to a `required` array, and omitted-not-null is real (`runlog.py` guards on `is not None`).

## [Blocking] The retention rationale is false for the harness's busiest tools

The claim — *"a `predicate` match is exactly recoverable from `response_fixture` plus `matched.index` at that commit"* — is repeated in five durable places: the `_tool_call_entry` docstring, `unit-test-spec.md`, both schema mirrors, and the new test file. It does not hold, because **the mock rewrites the response after matching and before logging**:

```python
response = {**response, "staged": staged}        # mock_mcp.py ~461
response = {**response, "ranked": rank_resp}     # ~477
reordered["rankingSkipped"] = RANKING_SKIPPED_NOTE   # ~496-503, also rewrites key order
entry["response"] = response
```

And `staged.resultsRef` is not reproducible at all — `results-staging.ts:83` builds it from `randomUUID()`, so no commit holds that value. I counted the exposure in the committed corpus:

```
predicate calls whose response the mock rewrites: 420
  record_search 251   fulltext_search 86   external_links_search 83
```

Those are precisely the responses this PR drops, and precisely the ones that cannot be reconstructed. Either qualify the claim in all five places, or keep `response` for those three tools — a two-line change in `_tool_call_entry`. The wording has to move either way.

## [Blocking] The cost argument only counts the half you dropped

Two problems. The figure doesn't match the corpus: there is no ~913 KB run log — the largest is **1,807,168 bytes** (`record-extraction/v1_2026-08-17_18-57-51`), and rebuilding the worst-affected log with predicate responses injected gives **+49%**, not ~16%.

More importantly, the half you *keep* was never costed. `live` is the majority of calls:

```
live      3,325 / 5,376 recorded calls   <- response RETAINED, unbounded
predicate 2,040                          <- dropped
none         11
```

On the heaviest live-dominated log that is roughly **+20%** — the same order of magnitude as the cost used to justify rejecting the predicate half. There is also no size cap on the new field, where the siblings have one (`text_excerpt` at 200 chars; e2e `response_summary` stripped past 14 days). Unit retention keeping the newest 5 candidates bounds total growth, so this is a "state the number and decide" item rather than a defect — but the body currently presents only the saving.

## Moderate

**`response_fixture` spec line contradicts the schema.** `unit-test-spec.md:1342` says it is *"null when kind is `none` or `live`"*. All four live handlers set `live:<tool>` (`mock_mcp.py:656, 721, 786, 859`), and **all 4,309 committed live calls carry a non-null value**. The schema's own invariant says the same. Worth fixing while you are in this line.

**Aborted tests silently omit both new fields.** `orchestrator.py:1702-1710` calls `assemble_test_entry` without them though `spec` is in hand. By your documented semantics absent means "predates #1000", so a new run log with an aborted test misrepresents it. 11 aborted entries across 6 committed logs, so the path is live.

## Minor

- **The schema states the retention rule but nothing enforces it.** `$defs/tool_call` enforces its three other `matched.kind` invariants with `if/then`; `response` gets only a description. Adding it as a free-form property also *weakens* the file — `additionalProperties: false` would previously have rejected a stray `response`.
- **~1,476 lines of undeclared reformat.** Both mirrors went 497 → 961 lines, re-expanded from compact JSON, for three added properties. No formatter enforces this and both were compact on main. Your own rebase note flags #1764 and #1781 in this area — a whole-file reflow guarantees those conflicts. Either revert the formatting or say in the body that ~720 of each file's 738 added lines are reformat.
- **"228 of 230 pass" counts annotation files as run logs.** At the merge head it is 122 of 123, one failure, identical on main — your conclusion is right, the denominator includes 109 `.ann.json` files.
- **"the code has always emitted `live`"** — the spec line was written 2026-05-19 and `live` arrived 2026-05-22. It was correct when written, then went stale. Doesn't change the fix.

Requesting changes. Every item is a wording change or two lines of code — nothing needs a redesign or a paid run, and the test at the centre of this PR is the strongest part of it.

T-FEH (changes requested):
Second pass, on `f08e1147` (current with main). Harness **2821 passed, 2 skipped**. Gate green, mirrors byte-identical, no paid re-run.

**The response to round 1 was substantive, and the fix is better than what I proposed.** Rather than qualifying the wording, you changed the rule: retention now keys on the response's **content** via `_is_mock_enriched`, and deliberately on the markers rather than a list of the three tools that have them today. I checked `_MOCK_ENRICHED_KEYS` against every mutation site in `mock_mcp.py` — complete, and the `rankingSkipped` reorder always sets its own marker, so nothing slips through. No fixture in the corpus ships those keys, so there is no false-positive retention. Findings 3, 4 and 6 are properly fixed, the reformat is fully reverted, and the aborted path is now pinned by a test.

## [Blocking] Neither production wiring is tested — and the body says one of them is

Every new test covers a helper in isolation. I mutated both call sites that actually write these fields into a run log, and ran the whole suite each time:

```
projection call site reverted to the old five keys   -> 2821 passed, 2 skipped   SURVIVED
main path stops passing the grading fields           -> 2821 passed, 2 skipped   SURVIVED
```

Reverting the *helper* is caught. Reverting the *wiring* is caught by nothing. So the aborted path — 11 entries across 6 logs, the one you pinned in answer to my round-1 finding — is guarded, while `_run_one_test_async`, which writes all 1,945 committed entries, is not.

That matters on this PR specifically, because absent means "predates #1000". A future refactor dropping either line produces fresh run logs that misrepresent themselves as old ones, with CI green.

And the body's mutation table asserts the opposite:

> | Projection reverted to the old five keys | **2 red** |

That does not reproduce. Four lines close it — extend `test_judge_error_in_run_records_skip_with_error` in `test_orchestrator.py` to use a `live` match and assert the three fields land on the entry. I applied that and re-ran both mutations: each then goes **1 failed**.

## [Moderate] 587 overstates by 167 — and 420 was right

I conceded this to you too quickly last round; re-checking, my original number holds. All three enrichment paths require a `projectPath`: `staged` gates on `args.get("projectPath")`, `rankingSkipped` on `"projectPath" in args`, and `ranked` on `response.get("staged")`, so transitively. Measured:

```
predicate calls on the three tools, projectPath truthy: 420
  record_search 251   fulltext_search 86   external_links_search 83
projectPath present but falsy (the rankingSkipped key-presence edge): 0
```

The other 167 are `external_links_search` calls from `locality-guide` with no `projectPath` — the mock never touches those responses, so they are exactly the plain-predicate case the rule correctly omits. 420 is also a tight upper bound, since it counts eligibility rather than confirmed enrichment.

The figure appears in three places — `orchestrator.py:1181`, `unit-test-spec.md:1432`, and the test docstring at :120. "Up to 420 … the ones passing a `projectPath`, which all three enrichment paths require" is the accurate form.

## [Moderate] The spec contradicts itself twelve hundred lines apart

`unit-test-spec.md:1152` still describes the narrow rule:

> `response` is present only for `live` and unmatched (`none`) calls

while §10 describes the widened one. Same sentence shape as the round-1 defect, in the sibling section.

## Minor

- **The test docstring still carries both retracted claims.** `test_a_fixture_matched_call_omits_the_response` (:88-91) still says "exactly recoverable" and "~16%", twelve lines above the test that refutes it. Those two strings now appear nowhere else in the repo — this is the last of the five sites.
- **The schema still describes the retention rule and enforces none of it.** I said last round I would not press this, and the rule has since changed shape — but the predicate half is now expressible: `if kind == "predicate" and "response" present, then response requires one of staged/ranked/rankingSkipped`. Your call; the helper is already guarded by tests.
- **The PR body still describes the round-1 design.** The corrections all live in `6b1248c9`'s commit message; the body — what the merge record shows — still says "Kept only where the content exists nowhere else — `live` and unmatched (`none`)", "~16%", and "228 of 230 pass". The real count is 122 of 123, which your commit message states correctly and which I reproduced on both sides.

## Verified clean

123 tracked run logs revalidate 122/1 on **both** sides, the single failure being the pre-existing `schema_version: 2` file. Both mirrors are byte-identical, nothing was added to any `required`, and omitted-not-null is schema-enforced. The anti-drift test remains the strongest thing here — it is still the only test in 2,823 that catches `_compute_outcome` drifting into gating invariant negatives, and it does that by driving the real function.

Requesting changes on the untested wiring. Everything else is wording.

johnmarkpeterbrown (commented):
Second review, on be0eefe4 (level with main). Harness 2854 passed, 9 skipped.
121 of 122 unit run logs validate under both schema mirrors, the one failure
being the pre-existing schema_version: 2 file. Nothing here blocks merge on
code, and nothing below needs a paid run.

@T-FEH — your round-2 blocking item is fixed, and your standing review is now
the only thing holding this. 92eb461c added the four assertions you specified;
I mutated both call sites you found surviving (projection call site reverted,
main path dropping the grading fields) and each now goes 1 red. Your 420 was
right and is what the code and spec say — I re-derived it independently, along
with the 167 external_links_search calls with no projectPath. The :1152 spec
contradiction, the retracted docstring strings, the reformat and the
denominator are all addressed. The if/then enforcement you left as their call
is still declined, with a reason I find sound: the rule now keys on response
content, which a kind-based if/then cannot express.

Worth knowing why nothing cleared — dismiss_stale_reviews_on_push is false
here, so your CHANGES_REQUESTED sits at f08e1147 across four pushes and
nothing re-requested you. Unresolved threads are zero. Could you re-review and
clear it? The three items below are informational and need no adjudication
from you.

For the record on the part that matters: the anti-drift test is still the only
test in 2,854 that catches _compute_outcome starting to gate invariant
negatives, and it does that by driving the real function rather than asserting
a table. That is the strongest thing in this PR and it holds up under mutation.

@aghadiayeamayanvboernest — three text fixes, all in files you already touched.
None goes near a test JSON or the run-log snapshot. Push them before the
re-review so T-FEH reads the final state.

1. The docstring miscounts the sites the retention rule rests on. There are
   five, not six.

   You wrote:    `"response"` as a key and all six `call_log.append` sites set it — and this
   Change it to: `"response"` as a key and all five `call_log.append` sites set it — and this

   This is the one to be sure lands. It is the stated premise of the rule, and
   a reader who checks it and finds five will doubt the rest.

2. The size-trade figures have no as-of date, and the corpus churns under them
   — 44 run logs added and 41 deleted on main since you measured, because
   retention keeps only the newest 5 per skill. I measure 4,378 of 7,773
   against 3,379 today. Your numbers were right when taken; dating them keeps
   them checkable.

   You wrote:    saving; retaining `live` is the cost, and `live` is the majority — 4,309 of
                 7,718 recorded calls against 3,393 predicate. There is no per-field cap here,
   Change it to: saving; retaining `live` is the cost, and `live` is the majority — as of
                 2026-08-24, 4,309 of 7,718 recorded calls against 3,393 predicate; the corpus
                 churns as retention prunes, so re-measure before requoting. There is no per-field cap here,

3. On an aborted entry the field claims something untrue of that entry. I ran
   _aborted_entry on a positive and on an out-of-scope negative: both give
   outcome "aborted", dimensions_gate_outcome true, and zero dimensions. The
   abort decided it; the dimensions could not have. Carrying the test's
   designed mode there is the right call — the description is what
   overpromises. Same line in both mirrors.

   You wrote:    "Whether the judge dimensions could change this outcome. False on \"invariant\" and \"routing\" tests, where they are diagnostic only
   Change it to: "Whether the judge dimensions could change this outcome for a test that ran to completion. Describes the test's grading design, so on an \"aborted\" entry it says how the test would have been graded, not what decided it. False on \"invariant\" and \"routing\" tests, where they are diagnostic only

Merge path: those three pushed, T-FEH's standing cleared, then a second
code-owner approval — the .py and .json paths map to senior-developers and
senior-genealogists, and it is at 0 of 2.
```

**Review by T-FEH — APPROVED** [review]
Third pass, on `dd08cd07` (your `1e12b54d` plus a clean merge of main). Harness **2834 passed, 2 skipped**; all CI green; eval gate green; no paid re-run.

## The blocker is fixed, verified by mutation

I re-ran both mutations from round 2 on the full suite. Both are now caught, each by the same guard:

```
projection call site reverted to the old five keys  -> 1 failed  test_judge_error_in_run_records_skip_with_error
main path stops passing the grading fields          -> 1 failed  test_judge_error_in_run_records_skip_with_error
```

The guard is a real behavioural assertion on a written entry — `entry["grading_mode"]`, `entry["dimensions_gate_outcome"]`, and the live response surviving the projection — not a source grep or a table. And it is placed in the test that already drives `_run_one_test_async` end to end, which is the right home for it.

**It also didn't trade one guard for another**, which was my worry about flipping that fixture from `predicate` to `live`. Predicate coverage survives at `test_orchestrator.py:464` plus the two cases in the new file, so nothing was given up to buy this.

## Everything else from round 2

- **420** is now stated in all three sites with the reasoning attached — and the docstring goes further than I asked, naming which gate each path uses and calling it eligibility rather than confirmed enrichment. No stale 587 survives.
- **The spec self-contradiction** at `:1152` is resolved; it now describes the widened rule.
- **The test docstring** carries an explicit retraction of "exactly recoverable" and "~16%" rather than the claims themselves.
- **The PR body** is fully corrected: the retention description matches the code, the live-half size cost is stated (4,309 of 7,718, no per-field cap, bounded by 5-candidate retention, largest log 1.8 MB), the count is 122 of 123 with a note that the earlier glob counted 111 `.ann.json` files, and the mutation row is narrowed to "Projection **helper** reverted", which is what actually reproduces.
- Mirrors byte-identical, nothing added to any `required`.

## No new breakage — and it is cheap to say why

`92eb461c` changed **no executable code**. Its 10 lines in `orchestrator.py` are entirely docstring; everything else is the two test files and the spec. So the guard was added without touching the behaviour it guards, which is the safest possible shape for this fix.

## One thing left, and it stays your call

No `if/then` invariant for `response` was added to the schema — still three, none mentioning it. I said last round this was yours to decide and I am not changing that. The helper is guarded by tests either way; the argument for adding it is that `additionalProperties: false` no longer rejects a stray `response` on any tool_call. Worth a follow-on card rather than another round here.

## Approving

Three rounds, and each one the response was substantive: round 1's false retention rationale was fixed by changing the *rule* rather than the wording, round 2's untested wiring by putting the assertion where the real function already runs. The anti-drift test remains the strongest thing in the PR — still the only test in 2,834 that catches `_compute_outcome` drifting into gating invariant negatives, and it does that by driving the real function.

Note for whoever merges: this still needs a second approval, and `senior-genealogists` as well as `senior-developers` — `*.py` and `*.json` both route to both teams.

johnmarkpeterbrown (approved):
Approving on 9cf8f873.

7fdeeb0b landed after T-FEH's approval at 3d8e250d, so I read it directly.
It is text-only — 8 insertions, 5 deletions, four files — and is exactly the
three corrections from my last comment with nothing else attached: the
call_log.append count, the size-trade figures now dated with a re-measure
instruction, and the dimensions_gate_outcome description covering the aborted
case in both mirrors.

Re-verified at this head: all checks green, harness 2855 passed / 9 skipped,
mirrors byte-identical, 0 behind main, 0 unresolved threads. I re-ran the two
mutations that matter after main's merges — _compute_outcome starting to gate
invariant negatives, and the production call site reverted — and each still
goes red. The new figures were right when taken; predicate is still 3,381
today, and the 73-call drift in total since is exactly the churn the new
as-of date now warns about.

Nothing outstanding. Good work on the anti-drift test — it is still the only
test in 2,855 that catches _compute_outcome drifting, and it earns that by
driving the real function.

Issue #1885, opened 2026-08-25
person-warnings spec documents 3 warnings the tool does not emit; the tool emits 76 tags


Ikennaya1:
> **Reviewed 2026-08-25 before junior handoff, and the lead has ruled on the one open
> scope question.** Read this block first — it changes the scope, the order, and two of
> the three decisions below.
>
> **Do this after PR #1896 merges** (issue #1616). That PR edits this same spec: it adds
> a `SUPERSEDED` banner to W3 and a new "How `hasEventAfterDeath1` actually decides"
> section. Keep both, and reuse the banner's wording as the template for W1 and W2.
>
> **The count is 74, not 76.**
> `grep -cE '^const [A-Z_0-9]+ = "' packages/engine/mcp-server/src/tools/person-warnings.ts`
> returns 76; two of those are `TREE_FILE` and `COHERENCE`. Re-derive rather than quoting
> this forward — extract `^const [A-Z_0-9]+ = "([a-z][A-Za-z0-9_]*)";` and intersect with
> `issueType:\s*([A-Z_0-9]+)`. All 74 are emitted; none is dead. `scoreType` is
> `COHERENCE` at all 75 emit sites, and `severity` is `contradiction` at 12 and
> `implausible` at 63.

**Touches:** docs/specs/person-warnings-tool-spec.md

## The divergence

`docs/specs/person-warnings-tool-spec.md` § Warning Definitions documents three warnings — **W1 `DEATH_BEFORE_BIRTH`, W2 `YOUNG_BIRTH`, W3 `EVENT_AFTER_DEATH`**. Grep the tool: none of those three names appears in `src/tools/person-warnings.ts`.

What ships is **76 FamilySearch-style tag constants** — `hasEventAfterDeath1`, `hasEventBeforeBirth365_2`, `hasAgeRangeGreaterThan120`, `hasBurialAfterDeath31`, `earliestChildBirthToBirthMale14`, `tooManyFathers2`, and so on.

So the spec's core section is not merely incomplete — it describes a design that was replaced, under names the tool never uses.

## The spec predicted this and was never updated

At `:101-106`:

> **Placeholder warning IDs and messages.** … Richard (ChesworthRM) is sharing the existing FamilySearch quality-score warning list; once available, **reuse its tags for equivalent warnings** … Do not finalize these strings until that list arrives.

The list arrived and the migration happened. The definitions did not follow.

## Worked example of how wrong it reads

[...]

Issue #1606, opened 2026-08-13
Source-evaluation findings: prefer 're-read the record' over 'detach source' for index errors; suppress un-actionable backend metadata


johnmarkpeterbrown:
**Touches:** packages/engine/plugin/skills/check-warnings/SKILL.md, packages/engine/plugin/skills/record-extraction/SKILL.md, packages/engine/plugin/skills/research/SKILL.md

Surfaced by feedback case #1536 (a **positive** report — the agent found more errors than the tester expected, and many were corrected per its suggestions). Two refinements about *how findings are presented*, plus a design question only the lead can settle.

**1. Remediation phrasing — re-read vs. detach.**
Asked to evaluate the sources on a FS profile, the agent flagged a fact discrepancy and effectively suggested detaching/flagging the source. It was actually an **indexing error**: the tester re-read the original image and corrected the index, keeping the source attached. Doctrine ask: when a fact conflict looks like a transcription/index error, recommend *"re-read the record and correct the index"* as the first-line remedy; reserve detaching for genuinely misattributed sources.

**2. Signal vs. noise — un-actionable backend metadata.**
It surfaced FamilySearch backend artifacts (duplicate fact IDs across persons) as if they were user-fixable errors. The tester: *"Having the metadata errors brought to my attention was not helpful."* Doctrine ask: separate user-actionable findings from un-actionable backend/metadata artifacts (present the latter as context, not a to-do).

**Decision (lead, 2026-08-27) — build a source-evaluation skill.**

The flow the tester used — "evaluate the sources attached to this profile / look for
errors" — has no skill and no routing today; it was freelanced in the main thread with only
`init-project` ever invoked. That is why a doctrine edit had nowhere to bind, and it is what
this decision fixes: **a new skill owns the flow, and both points above go in its body.**

Rejected, with reasons:

- **Widening `check-warnings`.** It covers one person's date and relationship integrity,
  holds only `person_warnings` and `person_quality`, and its description explicitly routes
  source conflicts to `conflict-resolution`. Hosting source evaluation there widens its
  scope, its tool grants and its description at once, and blurs a boundary it was written
  to keep.
- **Putting the doctrine in `record-extraction`.** The path the tester took never invokes
  it, so the doctrine would sit in a file this flow does not read — the exact failure this
  issue reports.

**The two doctrine points, which are the skill's reason to exist:**

1. **Re-read before detach.** When a fact conflict looks like a transcription or indexing
[...]

PR #1896, opened 2026-08-25
check-warnings: five doctrine corrections + deep dive — and prose that binds on none of three surfaces (#1616)

### LINKED ISSUE #1616: deep dive: check-warnings — and four doctrine corrections in the reference docs (split point, burial-after-death, hasEventAfterDeath1 catalog, bride's surname)
> **Reviewed 2026-08-21 before junior handoff.** No decision needed, but three of the five
> items describe a warning the tool cannot emit — read "what the tool actually does" first.
>
> **Unblocked in code.** PR #1714 (issue #1191) merged 2026-08-20; its edits to all three
> reference files are on `main`. Nothing in the code is left to wait for.
>
> **The eval slot is free as of 2026-08-21.** Issue #1606 held it and has been returned to
> Backlog under `needs-decision` pending a routing call, so this card owns the next
> `check-warnings` run. Issue #1659 (deep dive: check-warnings) is queued behind it and
> pays its own re-run — or folds these five corrections into its PR and pays one.
>
> **Line numbers re-pinned to `main` at `bb21954f`, 2026-08-21.** `warning-checks.md`
> moved +2 and `SKILL.md` moved +1 from the `e71a4931` pin when PR #1545 renamed the
> severities; the other citations are unchanged. The `hasEventAfterDeath1` entry in
> `warning-checks.md` is at `:33-38`, not `:31-36`.

**Touches:** packages/engine/plugin/skills/check-warnings/references/warnings-as-identity-signals.md, packages/engine/plugin/skills/check-warnings/references/warning-checks.md, packages/engine/plugin/skills/check-warnings/references/assumption-categories.md, packages/engine/plugin/skills/check-warnings/SKILL.md, docs/specs/person-warnings-tool-spec.md

## Read this before writing any of the five edits — what the tool actually does

`hasEventAfterDeath1` does **not** compare events against the Death fact. It computes
`latest(every fact) − latest(death-like fact) > 365 days`
(`packages/engine/mcp-server/src/tools/person-warnings.ts:308-317`; the predicate
`factDaysDiffLatestLatest` is at `src/utils/fact-helpers.ts:422`), and the death-like
family is nine fact types (`src/utils/mob.ts:45-55`):

`Death`, `Burial`, `Cremation`, `Funeral`, `Obituary`, `Probate`, `Will`,
`DeathRegistration`, `BurialRegistration`.

Three consequences, each of which changes an edit below:

1. **A fact of any of those nine types raises the death-side anchor, so it cannot fire
   the tag on its own** — however long after the death it is dated. There is no
   exclusion list; family membership is the whole mechanism. The docstring at
   `person-warnings.ts:298-307` says so in the code and nowhere else.
2. **A posthumous record fires the tag only when it was attached as a fact type outside
   that family** — a `Residence` from a city directory, say. That is already what
   `warnings-as-identity-signals.md:59-61` means by "attached … as a Residence-style
   fact." The same estate file is silent as a `Probate` fact and fires as a `Residence`
   fact. **The mis-typing is the trigger, not the date.**
3. So probate, obituaries and death notices belong in the cause list **as records whose
   content is posthumous**, with the fact type named as the thing that decides whether a
   warning appears. Do not write that they fire `hasEventAfterDeath1`.

Record this mechanism in `docs/specs/person-warnings-tool-spec.md` beside the
`hasEventAfterDeath1` description — it lives only in a source docstring today, and that
file is outside the eval snapshot, so the write-up is free. **It is part of this task.**

From alpha-team review by Clorinda Madsen and Shaunese Luthy, 2026-08-13/14. Line numbers pinned at `e71a4931`. Four separate corrections, batched because they land in one skill and pay one eval run.

## 1. The split-point rule is stated as a blanket and is wrong as written

`warnings-as-identity-signals.md:40-42`:

> The split point is the death in 1850 -- records dated before belong to one person, records dated after belong to another.

Clorinda: the death is where the *investigation* starts, not where the timeline gets cut. If the 1860 census lists the same household as 1850 plus children born since, the likelier reading is that the **death record** is wrong, not that there are two people. Research from a split point runs in both directions, and it runs hardest when the two candidates are close in age with similarly named parents — which is the exact case the section is about.

The question the section opens with ("if I split this person here, do the impossibilities disappear?") is still the right question. What has to go is the sentence that answers it before the research is done.

## 2. "Death must follow every other event" contradicts the line under it

"other than burial" is right about burial and wrong about the other eight. Replace `:89`
with a line that names the mechanism instead of one exception:

    - Death must follow every other event, except facts in the death-like
      family, which raise the death anchor rather than violate it: Burial,
      Cremation, Funeral, Obituary, Probate, Will, DeathRegistration,
      BurialRegistration (`hasEventAfterDeath1`)

Leave `:90` ("Burial must follow death") alone — that is a different check
(`hasBurialBeforeDeath`) and it is correct.

## 3. The `hasEventAfterDeath1` catalog entry is missing the posthumous-mention cause, and the cause list is too narrow everywhere

`warning-checks.md:31-36` gives two causes: a same-name person merged in, or a wrong death date.

The **posthumous mention** — a late-dated record that merely references the deceased — is already written up at `check-warnings/SKILL.md:87-96` and `warnings-as-identity-signals.md:44-73`, both of which say plainly that splitting on one damages the data. The catalog entry never got it. A reader who loads only `warning-checks.md` gets doctrine that the other two files contradict.

Separately, the record types named in the posthumous-mention lists are too few. Shaunese, twice: probate (intestate especially) routinely runs years after death, and an heir petitioning a will restarts it later still. Add to all three copies:

- probate and estate administration years after death, including an heir petition reopening it
- death notices as distinct from obituaries
- city directories

There are **four** copies of the record-type list, not three. The fourth is
`SKILL.md:109-115` — the example-output block the agent copies into its report, and the
copy the user actually reads. All four must end up saying the same thing:

- [ ] `references/warning-checks.md:33-38` — the catalog entry, which today gives two causes
- [ ] `references/warnings-as-identity-signals.md:51-57` — the examples list
- [ ] `SKILL.md:94` — the "Posthumous mention" cue
- [ ] `SKILL.md:109-115` — the example-output text

**Do not change the remediation sentences** at `warnings-as-identity-signals.md:59-73`
or the "unlink it and treat it as a reference" clause in `SKILL.md:94`. Issue #1606 (In
Progress) owns whether detach-vs-re-read is the right first-line remedy; adding record
types does not require touching them, and touching them would collide.

## 4. The record date is not the event date

The rule is right; the symptom named is not the one this tool produces. An obituary
transcribed with its digitization date and attached as an `Obituary` fact **suppresses**
warnings rather than manufacturing one — the wrong late date becomes the person's latest
death-like fact, hiding every genuine post-death event behind it and inflating
`hasAgeRangeGreaterThan120`. A false `hasEventAfterDeath1` appears only when the same
mis-dated record is attached as a fact type outside the death-like family. Write the
check to cover both directions:

    Before reading the three causes below: read the date of the EVENT the
    record describes, not the date the record was created, transcribed,
    indexed, or digitized. A transcription date recorded as the event date
    does damage both ways -- on a death-like fact (Obituary, Probate,
    Burial, Will) it pushes the death anchor forward and HIDES genuine
    post-death events; on any other fact type it manufactures a post-death
    event that never happened.

## 5. The bride's-surname entry names no evidence that would settle it

`assumption-categories.md:72` lists "A bride's surname is the same as her parents' surname" as an unsound assumption. That classification is correct and should stay — Shaunese read it as an error, which is itself the signal: the entry says the assumption cannot be accepted without evidence and then never says what evidence would do it.

Name the evidence a marriage record actually carries:

- the designation of which marriage this is ("second marriage", "previously married", "widow") — settles whether the surname is a prior husband's
- a parental consent signature for an underage bride, where a different surname implies a stepfather or guardian rather than a father

The consent-signature case is often on the reverse of the document and may not be in an index; say so rather than implying it is always visible.

Both pieces of evidence were already written into
`packages/engine/plugin/agents/record-extractor.md:248-269` on 2026-08-14 (commit
`d884c813`, same Shaunese review, filed under issue #1443) — the marriage-order
designation, the consent signer as evidence of a stepfather or guardian rather than a
father, and the "usually on the reverse of the license and usually absent from the index"
caveat. **Read that block and make the new text agree with it.** **Do not edit the agent
file** — it is outside this issue, and editing it takes the eval slot of every skill that
delegates to `record-extractor`.

Write it as a nested sub-list under `assumption-categories.md:72`, leaving the bullet's
own sentence unchanged: `init-project/SKILL.md:187` quotes that sentence verbatim as the
canonical example.

**Looked at and deliberately left alone:** `warnings-as-identity-signals.md:157-158`
carries the same blanket-split shape item 1 corrects ("2+ contradictions: almost
certainly two people merged; recommend splitting the profile"). It is a recommendation
conditioned on a warning count, not a claim about where the timeline gets cut, so it
stays. Say in the PR that you checked it, so the next reader does not read it as an
oversight.

## Not this issue

Issue #1194 covers naming the severity tiers and populating the implausible band. None of the five items above change a tag, a severity, or the tool.

## Verification — what the run does and does not prove

`make eval-skill SKILL=check-warnings` is a **gate, not a check** for this work. Every
file under `packages/engine/plugin/skills/check-warnings/**` is in the eval snapshot, so
touching any of them arms `.github/workflows/check-runlogs.yml`, and `check-warnings` is
not exempt (`eval/harness/scripts/check_runlogs.py:103` exempts only `research` and
`forget-and-rederive`). But `eval/tests/unit/check-warnings/rubric.md` has no dimension
covering split points, posthumous mentions, record-date-vs-event-date, or surname
evidence — a green run says nothing about whether any of the five corrections took hold.
**What verifies this change is a genealogist reading the diff against Clorinda's and
Shaunese's notes, plus the three code citations above.** Put that sentence in the PR body.

Budget: the three most recent committed run logs cost $1.76 / 9.7 min, $1.77 / 9.2 min
and $1.76 / 7.5 min (`totals.total_cost_usd` and `totals.wall_clock_ms` in
`eval/runlogs/unit/check-warnings/v1_2026-08-19_21-42-07.json`,
`v1_2026-08-17_16-43-11.json`, `v1_2026-08-13_21-23-36.json`; read 2026-08-21 — re-read
the newest log rather than quoting these forward). The annotation is 5 sampled tests, not
every dimension (`review_sample` is present in the newest log). `eval-cosmetic-skip` does
not apply; these edits change behavior. `docs/specs/person-warnings-tool-spec.md` is
outside the snapshot and adds no eval cost.

## Acceptance

- [ ] The split-point section states the death as the start of the investigation and names the wrong-death-record reading
- [ ] `warnings-as-identity-signals.md:89` names the death-like family, not just burial
- [ ] All **four** copies give the same `hasEventAfterDeath1` cause list — `warning-checks.md`, `warnings-as-identity-signals.md`, `SKILL.md`'s cue and `SKILL.md`'s example-output block — including posthumous mentions and the three added record types
- [ ] The record-creation-vs-event-date check is stated before the cause list and covers both directions (suppression and false positive)
- [ ] The bride's-surname entry names the marriage-order designation and the consent signature, agreeing with `agents/record-extractor.md:248-269`
- [ ] `docs/specs/person-warnings-tool-spec.md` records the death-like-anchor mechanism beside `hasEventAfterDeath1`
- [ ] The PR body says a green eval run verifies none of the five corrections, and names what does
- [ ] `make eval-skill SKILL=check-warnings` re-run and annotated

Ikennaya1:
Closes #1616. Both halves of the merged card: the five doctrine corrections from Clorinda Madsen's and Shaunese Luthy's alpha review, and the `check-warnings` deep dive.

## What a green run proves here: none of the five corrections

`eval/tests/unit/check-warnings/rubric.md` has no dimension covering split points, posthumous mentions, record-date-vs-event-date, or surname evidence. The run is a **gate**, not a check — every file under the skill dir is snapshotted, so touching any arms `check-runlogs`. **What verifies this change is a genealogist reading the diff against Clorinda's and Shaunese's notes, plus the three code citations in the issue's preamble.** That was #1616's own instruction and it holds.

One exception, and it is the one change the run *does* verify: the narrowed cluster rule below visibly changed behaviour in `ut_check_warnings_009`.

## The five corrections

**1. Split point** — the blanket "records before belong to one person, records after belong to another" is gone. Two explanations now compete without being ranked; the death is stated as where the investigation starts, not where the timeline is cut. Household continuity is the discriminator, as a cluster (spouse, children, residence, occupation, neighbours, age progression) with "no single factor decides it", and the hard case is named: several people sharing a name with similar ages, locations or similarly named parents — which produces mistaken splits as readily as mistaken merges.

**2. Death-after-event** — now names the nine-type death-like family that *raises* the anchor rather than violating it. `hasBurialBeforeDeath` on the next line is untouched; different check, and correct.

**3. All four copies of the cause list agree** — the catalog entry, the examples list, `SKILL.md`'s cue, and `SKILL.md`'s example-output block. Each carries the posthumous mention plus probate and estate administration running years after death including an heir petition reopening it, death notices as distinct from obituaries, and city directories. Each states that the fact **type** decides whether a posthumous record fires the tag, not its date.

**4. Record date vs event date** — stated before the cause list, both directions: on a death-like fact a transcription date pushes the anchor forward and *hides* genuine post-death events; on any other fact type it manufactures one that never happened.

**5. Bride's surname** — a nested sub-list, led by the marriage record's own parental-names field, which dissolves the inference rather than weighing on it, then marriage-order designation, then the consent signature as identifying a signer rather than a father, then the reverse-of-the-license caveat with absence-of-evidence stated. Agrees with `agents/record-extractor.md` without touching that file — the marriage-order, consent-signer and reverse-of-licence bullets against its marriage-record section (~248-269), and the parental-names bullet against its evidence-directness section (~598-611). An earlier revision cited only 248-269, which does not contain the parental-names content (@mercyokum, finding 5). The bullet's own sentence is byte-identical, since `init-project/SKILL.md` quotes it verbatim as the canonical example.

**Checked as the issue asked:** the escalation-table entry carrying the same blanket-split shape. It is a recommendation conditioned on a warning count rather than a claim about where the timeline gets cut — but see rule 12 below, which changed it after all, for consistency.

**Protected regions untouched:** #1606 owns the detach-versus-re-read remediation sentences and the "unlink it and treat it as a reference" clause. Both confirmed byte-identical.

## The spec write-up, and what it exposed

`docs/specs/person-warnings-tool-spec.md` now records the death-like-anchor mechanism, which until now lived only in a source docstring at `person-warnings.ts:298-307`.

Writing it surfaced something #1616 did not anticipate. The spec's `W3` documents a hard-coded exclusion list of seven types **including `Estate`**, gated on a required `Death` fact at year granularity — an algorithm the implementation explicitly rejected ("Java relies on that family + tolerance instead of a hard-coded exclusion list"). W3 is now marked superseded with a pointer. The wider divergence — **3 warnings documented, 76 tags emitted** — is filed as **#1885** rather than rewritten here: too big for this PR, and developer lane rather than genealogist.

## Two changes from the dive, NOT among #1616's five

The issue's acceptance criteria cover neither.

**Rule 12, the cluster verdict.** It was the operative twin of the blanket split point — and mandatory. It required the skill to open any 2+ contradictions with "a strong signal that records from two different individuals have been merged". `ut_check_warnings_009` obeyed it and produced a self-contradicting report: that verdict at the top, and "(a) Wrong death date — the recorded death date is too early" three lines below. The two contradictions were `hasEventAfterDeath1` and `hasAgeRangeGreaterThan120`, which one too-early death date produces together. Outcome: **pass**.

Now narrowed: check whether one wrong fact explains the whole cluster before asserting a merge, and reserve the merge reading for clusters no single error explains. Same narrowing applied to the escalation table.

**This is the change the run verifies.** 009 now spots that all three warnings cite the same death fact, calls a wrong death date "the leading single-cause candidate", offers the merge as the alternative, and names verifying that date as the discriminator.

**The example-output block** rendered `hasAgeRangeGreaterThan120` as "Implausible / Valid" while the tool emits `severity: "contradiction"` for it. It now renders from the severity the tool returned. Whether that severity is *itself* right belongs to #1194 and is untouched.

## The dive: three findings, one mechanism

Every `check-warnings` failure is **report discipline** — noise, over-claiming, fabricated citation — and no rubric dimension graded report discipline.

**F1 — synthetic-id quality noise. 53 of 57 applicable runs; 1 ever penalised.** Measured over the five committed logs as of 2026-08-27; **retention keeps only the newest five candidates**, so re-derive rather than quote — two earlier logs carrying the worst F2 rate were pruned mid-dive.

*Did:* `ut_check_warnings_007`, `v1_2026-08-19_21-42-07`: "No FamilySearch quality score — Patrick's ID (`I1`) is a local synthetic ID."
*Should:* `SKILL.md` twice — "skip this call silently", "Say nothing about FamilySearch quality — do not add a 'not available' note."
*Gap:* lane 2. The rule is already as absolute as prose gets; restating it is the one fix the deep-dive guide rules out.

**F2 — the report cites tree data the tool never returned. 8 of 80 runs with a non-empty response.** An earlier revision said "23 of 79"; re-derived it is 8 of 80 on the current corpus, and the denominator moves with retention. Measured over the five committed logs as of 2026-08-27; **retention keeps only the newest five candidates**, so re-derive rather than quote — two earlier logs carrying the worst F2 rate were pruned mid-dive.

*Did:* `ut_check_warnings_001`. The tool returned exactly `{"warningCount": 0, "warnings": []}`. The response: "Patrick's recorded lifespan (~1845–1908, approximately 63 years) and his relationship to his father Thomas Flynn (I2, born ~1818)." Every date is in the scenario tree; "63 years" is arithmetic on them.
*Should:* "The tool is the arbiter… the number '208 years' was never in the tool's response; do not invent it. Cite only the `factIds`, sources, and persons the tool's response actually mentions."
*Gap:* lane 2. Reading the tree is permitted in Step 1 to *identify* a person; carrying its contents into the report is what the rule forbids, and that is where the skill drifts. The zero-warning case is where it happens — nothing to report, so it explains why there is nothing to report.

**F3 — the conflict-resolution handoff is drifting, not flaky.** Reply length across six runs: 0, 144, 786, 186, 0, 1788 — from narrating the handoff to performing conflict-resolution's job outright.

*Should:* "your first and only action… Write no reply of your own — no preamble, no explanation, no summary."
*Gap:* structurally ungradeable by anything in the genealogist's lane. A negative test with a non-empty `correct_skill` is decided on routing alone and the judge's scores are diagnostic-only there, so a run that routes correctly *and* narrates passes. Only a validator can fail it.

## Validator requests — the measure of the dive

> **V1 — Rule:** a run that makes no `person_quality` call must not mention FamilySearch quality — no "not available" note, no "synthetic ID so no score" remark.
> **Where to look:** `tool_calls[]` for the call; `text_response` for `/familysearch quality/i`.
> **Why it is not judgment:** the call either happened or it didn't; the phrase either appears or it doesn't.
> **What a violation looks like:** `ut_check_warnings_018`, `v1_2026-08-27_09-42-49` — the rule is in that test's own `judge_context`, and it scored 3. **53 of 57** applicable runs mention it; **1** was ever penalised.

> **V2 — Rule:** every year and every duration in the response must appear in a tool response or in the user's message. No arithmetic the tool did not perform.
> **Where to look:** `text_response` against each `tool_calls[].response_fixture` resolved through `eval/fixtures/mcp/<name>.json` → `.response`, plus `input.user_message`. **There is no `tool_calls[].response` field** — an earlier revision of this box said there was, which would have made V2 unbuildable. Skip `live:` fixtures, which have no file.
> **Why it is not judgment:** string presence, both directions. The body names "208 years" as the canonical violation.
> **What a violation looks like:** `ut_check_warnings_001` — `warningCount: 0` and no data returned; the response cited a birth year, a death date to the day and a county. All three are true of the scenario tree, which is why a human waves it through; none came from the tool. **8 of 80** runs.

> **V3 — Rule:** on a source-conflict handoff, the run must make zero tool calls, produce an empty `text_response`, and carry the target skill in `skills_invoked`.
> **Where to look:** those three fields.
> **Why it is not judgment:** all three mechanical, each stated in the body as an absolute.
> **What a violation looks like:** `ut_check_warnings_011` wrote 1788 characters in `v1_2026-08-25_10-04-37` and 1844 in `10-41-52`, both with an empty `skills_invoked`, neither penalised. **2 of 5** handoff runs. (An earlier revision cited `v1_2026-08-17_16-43-11`, which retention has since pruned — the evidence no longer exists in the repo.) Scope is conflict-resolution only: `validate-schema` has no silence absolute, so widening V3 false-positives on `ut_check_warnings_012`.

## The finding I did not go looking for: prose does not bind here — and a judge prompt that says it binds did not change that

I sharpened `rubric.md`'s `FamilySearch quality reporting` dimension so a synthetic-id mention is a partial, and predicted 11 of 18 tests would drop. **Two dropped, neither for that reason.** The snapshot hash confirms the judge read the edited file.

Then the judge prompt itself changed underneath this branch. `7fe9bf3b` — 2026-08-25 12:16, about ninety minutes after run 2 — rewrote the per-test-context section from "use them as **background** to ground your rationales" to "**binding on the scores you give**", and added: "a requirement the response did not meet **is a deduction** in whichever dimension covers it… scoring 3 because the response was strong elsewhere is an error." Run 3 is the first `check-warnings` run under it (`judge_prompt_hash` `0d186137147c` → `c39d70034788`).

`ut_check_warnings_018`'s own `judge_context` carries the synthetic-id rule, unchanged. **Under the binding prompt it scored 3.**

So the rule sits on **four surfaces** — `SKILL.md` twice, `rubric.md` once, and a per-test note the prompt now calls the narrowest and highest-priority instruction reaching the judge — and holds on none:

| | synthetic-id tests | breached | penalised |
|---|---|---|---|
| run 2 — background prompt | 11 | 9 | 1 |
| **run 3 — binding prompt** | **11** | **11** | **0** |

The two compliant runs in run 2, `009` and `010`, both regressed to breaching, and the single catch in the corpus's history disappeared.

**Method, so this is re-derivable:** a test is synthetic-id when every `personId` it passed to a tool fails `[A-Z0-9]{4}-[A-Z0-9]{3}` (11 of 18; the other seven are four real FS ids and three tests making no call). A breach is any case-insensitive `FamilySearch` in `text_response`. Penalised is `FamilySearch quality reporting` scored 1 or 2. I read all 11 sentences rather than trusting the match count — every one is routing narration or a standalone "not applicable" note, the two forms `rubric.md:42` names as partial. Call discipline is clean in both runs: zero `person_quality` calls on a synthetic id, one on each real id.

Across the five logs committed as of 2026-08-27: **53 of 57 applicable runs mention FamilySearch quality; 1 was ever penalised.** Applicable means a synthetic `personId` and no `person_quality` call. Earlier revisions of this body said "52 of 59" and then "64 of 66 / 53 of 55" — the first I cannot reproduce at all, and the others were computed over a corpus retention has since pruned. The method is stated so this one can be re-derived rather than quoted; the denominator moves every time a run lands.

**Two things moved at once and one run cannot separate them:** the judge prompt changed, and the skill's own wording is unpinned (the SDK exposes no `temperature` for the skill run). What is not in doubt is the count — under a prompt that declares per-test notes binding, none of 11 breaches were penalised.

## Review round: what @mercyokum's findings changed

All nine points verified against the tree rather than taken on description; all nine held. Findings 1 and 3 were doctrine defects in what this PR shipped.

**Finding 1 — the central worked example was mathematically backwards.** "One death date recorded decades too early fires both `hasEventAfterDeath1` and `hasAgeRangeGreaterThan120`" does not hold. `hasEventAfterDeath(mob, 365)` measures `latest(any fact) − latest(death-like)`; `hasAgeRangeGreaterThan(mob, 120)` measures `earliestDeath − latestBirth`. A death moved earlier fires the first and **suppresses** the second — the two need opposite directions of error. It was the load-bearing example for the cluster correction, in **five** places, not the three she listed: the `SKILL.md` doctrine paragraph, its example transcript, its report template, the escalation table, and `detect-cluster.json`'s `judge_context[1]` — the last being the one the judge prompt now treats as binding, i.e. a standing instruction to reward reasoning the implementation contradicts.

Replaced with her suggestion, which is better than a patched example because it is checkable from the tool response: **overlapping `factIds` across contradictions** means one wrong fact is a live explanation; disjoint means no single fact explains them.

**And that fix needed a caveat, found by following the same thread further.** `hasEventAfterDeath1` builds its `factIds` from `selfFactIds(mob, null)`, and `null` matches every typed fact. So it overlaps everything by construction, and any cluster containing it always reads "overlapping" — which would have made the merge branch unreachable in the commonest case. Both doctrine sites now say so and point the test at the narrower-scope contradictions. Without it I would have replaced a false rule with a near-vacuous one.

**Finding 3 — `SKILL.md` and the escalation table disagreed.** `SKILL.md` said "Either way, recommend" the timeline-and-split-point rebuild; the table gated that same recommendation on no single error explaining the cluster. A model following `SKILL.md` literally would hunt for a split point in the same breath as "do not assert a merge". `SKILL.md` is now gated to match, and `detect-cluster.json`'s `judge_context[4]` with it — that bullet called the timeline rebuild a correct follow-on, so leaving it would have recreated the contradiction one layer down.

**Finding 7 — the genuine-merge branch had no test.** `person-warnings-cluster.json` was the only multi-contradiction fixture and all of its warnings on Patrick share `F2`, so a regression on the disjoint branch would have passed the whole suite. New fixture `person-warnings-cluster-disjoint.json` and `ut_check_warnings_019`, severities and messages copied from the source. It passes, and for the right reasons.

**Minor 2 — nothing tied the prose death-like lists to the constant.** New guard `deathlike-family-docs.test.ts` extracts the list from all three prose sites and compares it to `DEATHLIKE_FACT_TYPES` both ways, so an extra name fails as loudly as a missing one. Per CLAUDE.md it was **proven to fail** three ways before committing: adding a tenth type fails all three sites, dropping a name fails that site, and rewording the passage so the pattern misses fails loudly rather than passing vacuously. `warnings-as-identity-signals.md`'s deliberate omission of `Death` is declared per-site rather than tolerated by a loose match.

**Four figures of mine did not reproduce** — "52 of 59", "23 of 79", "four of six", and a canonical violation citing `v1_2026-08-17_16-43-11`, a log retention has since pruned. All four are corrected above with the method and corpus stated beside each. They failed the same way: asserted once, never re-derived. Her minor 1 is a process point, not a typo.

## Cost

**Five paid runs, $8.4958** — $1.7557, $1.6366, $1.5424, $1.7916, $1.7695. Earlier revisions of this body said $3.51 and then $3.93; the logs are the record.

Runs 3, 4 and 5 were owed rather than chosen, and it is worth being exact about why, because three of them trace to my own errors rather than to the work:

- **Run 3** — the bracket fix in `e746ab12` changed `SKILL.md`. @EdmondOware declined `eval-cosmetic-skip` on the ground that removing an instruction is a behaviour change, which this body already said and I argued against. He was right.
- **Run 4** — @mercyokum's findings 1 and 3 are doctrine defects. Fixing doctrine invalidates the snapshot, so a run follows by construction.
- **Run 5** — the fixture I wrote for finding 7 gave `hasEarlyMarriage14` `severity: "contradiction"`. The tool emits `"implausible"`, and my message text was not the tool's either. A test whose whole purpose was "two *contradictions* with disjoint facts" was not testing that. Rebuilt on `hasChristeningBeforeBirth` and `hasBurialBeforeDeath`, both genuinely `contradiction` with `factIds` disjoint by construction.

**The stale line is fixed.** `detect-cluster.json`'s `judge_context[5]` graded severity as "Critical" and "Note", the tier names PR #1545 retired on 2026-08-18. It was recorded here as deferred rather than buy a third run; the third run then happened, so the reason expired and it is fixed.

## The final run reads 19 of 19 and that number means nothing

**Five runs: 15/2/1, 14/2/2, 17/1/0, 18/1/0, then 19/0/0.** The suite has never looked healthier and the behaviour under it has not improved.

In the final run, **11 of 12 synthetic-id tests still mention FamilySearch quality and none was penalised.** Scored to `rubric.md:42` those eleven are 2s, and `_compute_outcome` gates on 2, so the honest reading of a 19/19 is roughly **8 pass, 11 partial**. The headline is an artifact of the dimension not binding — the finding above, not a separate one.

**`018` is the case to read, and its earlier framing here was wrong.** An earlier revision of this body listed "018 now passes" among the run's outcomes. That framing invited the wrong reading, and @mercyokum's finding 2 is correct: the model did not get better at this rule, the judge got worse at reading its own input. The two responses are the same violation, cosmetically reworded:

- run 2, scored **1** — "…so I'll run the offline warnings check only (**no FamilySearch quality call needed**)."
- run 3, scored **3** — "…so I'll run the offline warnings check only (**no FamilySearch quality score applies**)."

Both mention it in the opening sentence. The run-3 rationale reads "The response contains no mention of FamilySearch quality whatsoever — the skip was silent as required by SKILL.md", which is false against the transcript it grades. That is not the judge excusing a violation; it is the judge asserting something categorically untrue about the text in front of it.

**`019` makes the finding airtight.** The new test written for this review carries the rule in its own `judge_context`, spelled out to "even in one clause of an opening sentence", under the prompt that calls per-test notes binding. Its opening clause reads "skip FamilySearch quality (no FS profile to score)". It scored 3. That is a rule written fresh, for this exact failure, on a test with no history — and it still did not bind. Five surfaces now, none of them holding.

**`011`** passes on an empty `text_response` and zero tool calls, which is correct, and the harness still flagged `routing_negative_judge_fail` (Correctness 1, Completeness 1). Two runs earlier it wrote 1788 characters and passed too. The outcome is decided on routing either way, so both gradings are invisible in the headline — which is V3.

**`019` also verifies the doctrine fix.** It read the disjoint `factIds` (F5/F1 against F6/F2), concluded a merged profile, and recommended the timeline rebuild — the branch `SKILL.md` gates on. Both warnings were labelled Contradiction and paired with Fundamental, so the severity-to-category mapping holds.

## Two defects run 3 surfaced, recorded rather than fixed

**A `judge_context` bullet requiring a tool that does not exist.** Three tests — `check-flynn-assertions.json`, `detect-cluster.json`, `detect-event-after-death.json` — require examining assertions "via `person_evidence`". That is not in `check-warnings`' `allowed-tools` (`person_warnings`, `person_quality`) and is not an MCP tool at all: no file under `src/tools/`, absent from `tool-schemas.ts`. It is a sibling skill. `check_rubric_tool_drift.py` catches none of the three, because the lint's vocabulary is real tool names and this is not one — it flags `negative-schema-validation.json` for `validate_research_schema`, which is. Under the binding prompt an unmeetable bullet becomes a standing instruction to deduct forever; on 001 the judge escaped it by asserting the skill had examined `a_001-a_005`, which it never did. Same shape as the `same_person`/`tree-edit` cases the eval docs already name.

**A reviewer cannot mark N/A what the judge scored.** `Severity classification` scored 3 on 001 and 014, where zero warnings fired and there is nothing to classify — while 018 scored the identical situation N/A. One run, one circumstance, three gradings. The annotation records this in comments only: `dimensionAllowsNa` (`eval/app/lib/types.ts:189`) offers N/A when the judge scored null or on `Tool Arguments`, so a 3 that should be null cannot be corrected. `ann.schema.json` permits null; the picker does not offer it. The eval docs already flag the opposite risk — a null standing in for a 1 — but not this one, and it inflates the weighted mean this run reports as 2.95.

## The prohibition list (Step 1), saved for the next auditor

Twenty-five rules checkable against a transcript, quoted from the 197-line body.

**Routing and handoff:** (1) a source-vs-source disagreement is handed to `conflict-resolution` as the first and only action, with no reply of any kind; (2) no `person_warnings`, no `person_quality`, no tree read, no analysis on that path; (3) internal skill names never appear in user-facing text; (4) asked to *fix* a warning, it does not fix it here.

**Tool calls:** (5) `person_warnings({ projectPath, personId })` per person checked; (6) `person_quality` additionally when the id is a FamilySearch ID; (7) on a synthetic id, skip *silently* — no call and no mention of FamilySearch quality at all; (8) a quality failure never suppresses the warnings; (9) multiple name matches — ask which person first.

**Reporting warnings:** (10) severity icon, `issueType`, the tool's `message`, assumption category; (11) name facts only when the tool returned `factIds`, and never read the tree to hunt for them; (12) count first, and check for a single shared cause before asserting a merge; (13) `missingFactsAndRelatives` gets Implausible plus the limited-data note; (14) on `hasEventAfterDeath1`, no defaulting to identity confusion and no specific corrective action before the source type is known; (15) the tool's `message` verbatim.

**Quality reporting:** (16) a separate section, never merged with the impossibilities; (17) each `sentence` verbatim, grouped by `scoreType`, repeats collapsed; (18) no invented quality band; (19) the exact zero-issue sentence; (20) improvements, never urgent errors.

**Interpretation and hard rules:** (21) `relatives*` tags — verify the link before any data fix on the relative; (22) the exact no-warnings sentence; (23) never auto-correct; (24) the tool is the arbiter — no tree read to verify a verdict, no date arithmetic of its own; (25) never claim anything about the tree the tool's response did not state.

**Excluded as judgement, not checkable:** "present warnings with appropriate context", whether a next step is genuinely *concrete*, whether the historical-exception framing reads well.

## Verification

- All five sites re-pinned against `main` before editing; the death-like family and the predicate read from `mob.ts:45-55` and `person-warnings.ts:309`
- The `init-project`-quoted bullet's own sentence ("Unlink it and treat it as a reference instead.") is byte-identical against `main` — but the `Cue:` text in that same bullet was rewritten by this PR, so this is a substring surviving inside a line the PR itself changed, not an untouched region. An earlier revision of this list also claimed "protected remediation sentences" were confirmed unchanged; **those sentences do not exist anywhere in the tree** and nothing was verified. Both corrections are @mercyokum's finding 4.
- No mojibake in any edited file
- Annotation, run 2: 35 corrections over five sampled tests, six disagreements, every non-exempt cell commented
- Annotation, run 3: 31 corrections over 001/011/013/014/018, five disagreements, every cell commented; `check_runlogs.py` passes locally
- `eval-cosmetic-skip` does not apply — these change behaviour


Paaboat:
For ut_010 FamilySearch 3→2—can you confirm the model response was silent on the synthetic-ID quality note? Just want the paper trail explicit before I approve.

Ikennaya1:
Quick correction on the number — **010 wasn't one of the ones I marked down.** I agreed with the judge there, and it's the clean one: the words "FamilySearch" don't appear anywhere in that response. It stayed quiet, exactly as it should.

The two I marked down are **007 and 008**. Both mentioned FamilySearch quality when they shouldn't have:

- **007:** "a synthetic id, so I'll run the offline check only (no FamilySearch quality call applies here)"
- **008:** "there is no FamilySearch live profile to score — the quality check doesn't apply here"

The instruction is to skip that check *and say nothing about it*. Both skipped it correctly, then explained why — and the explaining is the part that breaks the rule.

I marked them partial rather than fail because they got the important thing right: neither one actually ran the check it shouldn't have, and the real warnings came through fine. The rubric already puts this exact case in the partial band, so that wasn't a judgement call on my part.

The judge gave both a full pass. Its note on 007 says the skill "explicitly noted that I1 is a synthetic ID" and then, a sentence later, that no quality remarks appear in the output. Both in the same paragraph.

Here's what I think matters: 010 and 007 start with nearly the same sentence. The only difference is 007 adds the bit in brackets. That's the whole gap between following the rule and breaking it — which is why I'm asking for an automatic check instead of more wording in the instructions. If it skipped the check and still said "FamilySearch quality", fail it. No one has to read the transcript to spot a stray sentence.

EdmondOware (changes requested):
The five corrections are good genealogy, and the cluster narrowing is the best of them.
Requiring the skill to ask whether one wrong fact explains the whole cluster before asserting a
merge is the correct diagnostic order, and it replaces a rule that taught the opposite: count
symptoms rather than find a cause. One death date recorded decades early fires both
`hasEventAfterDeath1` and `hasAgeRangeGreaterThan120`, and calling that two people is the error
that splits a correctly merged profile, which does more damage than the wrong date it started
from. The posthumous expansion is right too, particularly the death notice as distinct from an
obituary, the city directory, and the sentence that the fact type decides whether the tag fires
rather than the record's date.

The three-prose-surfaces finding is the strongest thing here, and you got to it the hard way:
tightening the rubric bar yourself, predicting the drop, and reporting that it did not happen.
007's rationale asserting that no FamilySearch quality remark appears while its own first
sentence contains one is the kind of evidence that settles an argument. It carries V1 to V3 on
its own.

### The new bracket is an authoring note sitting in a user-facing field

`SKILL.md:80` specifies that bracket as a report field, "The assumption category violated
(Fundamental / Valid)", and your own prohibition list item 10 says the same. The block it sits
in is headed **Example output**, and its sibling at line 111 reads as a gloss to the reader:
"[Fundamental: people cannot act after their death -- but a posthumous mention can be
misattached to their profile.]"

Line 130 now reads:

```
[Fundamental: the tool emits this tag at severity "contradiction"; render the
icon and category from the severity the tool returned, not from how the message
reads.]
```

That is an instruction to the model sitting where the reader's explanation belongs. A model
following the example prints rendering guidance to a genealogist, and the reason the check is
fundamental is dropped in the process.

The severity correction itself is right and worth keeping. `assumption-categories.md` lists "A
recorded lifespan cannot exceed plausible biological limits (~120 years)" under **Fundamental
Assumptions** and maps Fundamental to `severity: "contradiction"`, so Contradiction is correct
and the old "Implausible / Valid" was wrong. Only the bracket text needs to go back to being a
gloss, and that file already supplies the wording:

```
Change it to:  [Fundamental: a recorded lifespan cannot exceed plausible
               biological limits (~120 years).]
```

If the rendering rule needs stating, it belongs in the prose near line 80 where the report
fields are specified, rather than inside the output the model copies.

EdmondOware (approved):
Approving. The bracket is a gloss again, in the wording `assumption-categories.md` already
uses, and the re-run is clean: `v1_2026-08-26_11-53-31` active at 55 snapshot entries with zero
mismatches, annotated across the whole review sample, $1.54.

But do not read 17 pass and 1 partial as an improvement on 14/2/2. I think it is the opposite,
and it is the most important measurement in this dive.

### The fourth surface removed the only instrument that was working

`ut_check_warnings_018` was the one test in six committed logs that ever caught the synthetic-id
breach. Your own body called it "the suite working". Under `c39d70034788` it scores 3.

The behaviour did not improve. Its `text_response`, in both runs:

| | response | outcome | dimension |
|---|---|---|---|
| `10-41-52` | "a synthetic ID, so I'll run the offline warnings check only (no FamilySearch quality **call needed**)" | fail | 1 |
| `11-53-31` | "a synthetic ID, so I'll run the offline warnings check only (no FamilySearch quality **score applies**)" | pass | 3 |

The second is almost word for word the example your own rubric names as a partial: "as routing
narration ('synthetic ID, so no quality call needed')".

I ruled out the alternatives before writing this. `_018` is `quality-skipped-synthetic-id.json`
and it did not change between the runs; the last edit to any check-warnings test was `f095295e`
on 25 August and it touched `_009`. The rubric's snapshot hash matches the current file in both
run logs, so the same bar was in force both times. Same test, same bar, same violation. The
judge prompt is the only variable, and the score moved the wrong way.

`FamilySearch quality reporting` is now flat at 3 across all fifteen gradings.

### What I think this does to the finding

It strengthens it and makes it worse. "Binds on none of three prose surfaces" becomes: the
fourth surface, the one designed to make `judge_context` binding on the score, silenced the only
dimension that was catching the breach. The suite now reads healthier while catching strictly
less, which is the exact shape a deep dive exists to surface and the exact shape a rising pass
count hides.

Worth a section in the body, because the numbers alone say the opposite. And it is a harder
argument for V1 than anything already in there: four surfaces, none binding, and the newest one
actively removed a working check.

One caution from the research-plan dive, where the same re-measurement came back flat rather
than moving: measure and report, do not assume the direction. Here it moved, and it moved
against you.

Nothing outstanding from me.

mercyokum (changes requested):
**Requesting changes**

Extraordinarily thorough dive — the empirical rigor on F1 (predicting, measuring, and honestly reporting a null result under a prompt change) is the standard this repo should hold every deep-dive to. A few things need fixing before merge, the first two blocking — both go to the core of what this PR claims to have measured and fixed.

## 1. The core worked example is mathematically backwards

The doctrine's central corrected example — "one death date recorded decades too early fires both `hasEventAfterDeath1` and `hasAgeRangeGreaterThan120`" — doesn't hold against the actual implementation. Checked directly in `person-warnings.ts`:

- `hasEventAfterDeath1` fires on `latest(any fact) − latest(death-like fact) > 365 days`
- `hasAgeRangeGreaterThan120` fires on `earliestDeath − latestBirth > 120 years`

A death date recorded **too early** does explain `hasEventAfterDeath1` — later real events end up looking post-death. But it **shrinks** the `earliestDeath − latestBirth` gap, making `hasAgeRangeGreaterThan120` *less* likely to fire, not more. A single "too early" error can't produce both; the age-range warning needs the opposite direction of error.

This isn't a one-off — it's repeated verbatim in three places: `SKILL.md`'s doctrine paragraph (line 84), its example-output transcript (lines 103–108), and the escalation table in `warnings-as-identity-signals.md` (lines 181–183). It's the load-bearing example for the whole cluster-verdict correction.

**Worth considering:** the test fixture backing this (`person-warnings-cluster.json`) doesn't compute warnings from real dates at all — it's a hard-coded response where both warnings just happen to cite the same `factIds` (`F1`, `F2`). That might be the more honest signal to teach on ("overlapping factIds across contradictions is worth checking for a shared cause") rather than a specific causal claim about which direction of date error produces which warning.

## 2. `018`'s "pass" is a judge hallucination, not a behavior fix

Checked both committed run logs directly. The model's actual behavior barely changed between them — same violation, cosmetically reworded:

- Old run (scored 1): *"...so I'll run the offline warnings check only (**no FamilySearch quality call needed**)."*
- New run (scored 3): *"...so I'll run the offline warnings check only (**no FamilySearch quality score applies**)."*

Both responses mention "FamilySearch quality" in their opening sentence. But the new run's judge rationale reads: *"The response contains no mention of FamilySearch quality whatsoever — the skip was silent as required by SKILL.md."* That's factually false against the transcript it's grading.

The PR's own narrative cites `018` moving to pass as one of the two notable outcomes of the final run ("`018` now passes... the only test in six committed logs to catch this"). That framing is wrong — the model didn't get better at this rule, the judge got worse at reading its own input. This is a second live instance of exactly the failure mode F1 already documents (a prose rule the judge doesn't reliably enforce) — except here the judge doesn't just excuse the violation, it asserts something categorically false about the text in front of it.

## 3. SKILL.md and the escalation table now disagree

SKILL.md's cluster paragraph ends "**Either way**, recommend: [timeline rebuild / split point]" — applied regardless of whether a single wrong fact explains the cluster. But `warnings-as-identity-signals.md`'s escalation table gates that same recommendation: "If one error explains the cluster, verify that fact against its original source first. If no single error explains it, ... recommend splitting the profile." One copy says always-recommend; the other says only-when-unexplained. A model following SKILL.md literally would recommend the split-point search in the same breath it's told "do not assert a merge."

## 4. The "protected region... byte-identical" claim overstates itself

Re-verified independently, with extra rigor since this was flagged for a second look:

- **Issue #1606 is genuinely OPEN**, and its content is a deferred design question ("Decision needed... which is the lead's") — never implemented.
- **The "detach-versus-re-read remediation sentences" don't exist anywhere.** Two independent greps across `packages/engine/plugin/` and `packages/engine/mcp-server/` for "detach" / "re-read the record" / "correct the index" found nothing related — there's nothing in the codebase for the PR to have confirmed unchanged.
- **The "unlink it and treat it as a reference" clause — this part is genuinely true.** Diffed the exact quoted substring between `main` and this PR directly: it's byte-identical, word for word.
- **But the framing overstates it.** `git blame` on that line attributes the *entire* containing bullet to this PR's own commit — the surrounding "Cue:" text in that same sentence was substantially rewritten (added death notice, city directory, probate timing detail, the fact-type-decides sentence). Calling this a "protected region... untouched" overstates a substring surviving inside a line the PR itself rewrote.

## 5. A sourcing citation doesn't cover what it's cited for

The new bride's-surname sub-list claims to agree with `agents/record-extractor.md:248-269`. Three of the four new sub-bullets do match content in that range (marriage-order, consent-signer, reverse-of-license). The first one — "the record may name her parents outright... direct evidence outranks inference" — has no counterpart there; that content actually lives in a different section of the same file (~lines 596–611), never cited.

## 6. No follow-up issue for the validator requests

V1/V2/V3 (plus two more defects the PR calls "recorded rather than fixed") exist only as PR-body prose — no board card, no assignee, no `nothing-checks` label. Sibling deep-dive PRs on `citation` and `research-plan` each filed a dedicated "N validator requests" issue (#1749, #1866) per the deep-dive guide's own division of labor. As merged, F1/F2/F3 continue in production exactly as before, now documented rather than tracked.

## 7. No test covers the genuine two-people-merged branch

`detect-cluster.json` is the only unit test with 2+ contradictions on one person, and this PR narrowed its fixture to the single-cause-explainable case. Nothing in the suite now exercises "no single fact explains the cluster → genuine merge," even though SKILL.md and the escalation table both still specify a recommendation for it. A future regression on that branch would pass every test in the suite.

## Minor

- **Stale "52 of 59" figure.** Asserted unqualified in the F1 headline and the V1 validator box; only ~30 lines later does the PR admit "I cannot reproduce that denominator." The correct figures (64 of 66; 53 of 55 after prune) live only in the fine print — worth moving the retraction up front or fixing the headline number directly.
- **Death-like fact-type list has no link back to its source.** `DEATHLIKE_FACT_TYPES` in `mob.ts` is restated verbatim in three prose locations this PR touches, with nothing (comment, test, lint) tying any of them back to the constant — this repo already has the pattern for this class of problem (`measured-figures.test.ts` for numeric claims); worth the same treatment here so a future change to the Set doesn't silently leave the docs stale.

Ikennaya1:
All nine findings reproduced when checked against the tree, and following them uncovered three additional mistakes of my own, which is probably the strongest evidence the review was worthwhile.

Key fixes:

* **Finding 1** appeared in five places, not three, including `judge_context[1]`, where it was actively teaching the judge reasoning the implementation contradicts. Updated everywhere and added a caveat for `hasEventAfterDeath1`, whose `factIds` overlap almost everything by construction.
* **Finding 3** required updates to both `SKILL.md` and `judge_context[4]` to avoid recreating the same problem at the judge layer.
* **Finding 7** exposed a flaw in my fixture. I had marked `hasEarlyMarriage14` as a contradiction when the tool emits it as implausible. Rebuilt the test using genuine contradictions with disjoint facts (`hasChristeningBeforeBirth` and `hasBurialBeforeDeath`). That correction required a fifth run.
* **Findings 2, 4, 5, and Minor 1** were documentation fixes. Four previously reported figures did not reproduce and have been corrected with methods and corpus documented.
* **Minor 2** adds a guard that validates the prose list against `DEATHLIKE_FACT_TYPES` and fails loudly if they drift apart.
* **Finding 6** resulted in issues **#1965** (validator and rubric gaps) and **#1964** (annotation-save corruption).

### What the fifth run shows

**19/19 passing tests is misleading.** Eleven of twelve synthetic-ID tests still mention FamilySearch quality and are not penalized. By the rubric, the results are closer to **8 pass / 11 partial**.

`ut_check_warnings_019` demonstrates the problem clearly: the judge context explicitly states the rule, the response violates it ("skip FamilySearch quality..."), and the test still scores 3.

### Two things to note

* **Run 4 remains committed without annotations.** Run 5 satisfies the gating requirement, and I kept Run 4 because the PR references its results.
* **One deferral remains:** a one-line change to `judge_context[1]` could land here, but it would invalidate the snapshot and require a sixth paid run. I deferred it into **#1965**, but can pull it into this PR if you'd prefer.

mercyokum (approved):

mercyokum (approved):

Issue #1851, opened 2026-08-24
Widen the harness detector to the sanctioned direct-agent route, and correct the docs that state the old rule


DallanQ:
> **Rescoped 2026-09-01, after the lead reversed the ruling this card was written to enforce.**
> The direct `Agent` spawn of a paired agent **is** the sanctioned in-loop route. Everything below
> the fold is the original body, kept because its measurements and its plane-by-plane analysis are
> still the evidence — but read its *conclusion* as retired. In particular, **do not do the
> "cheapest thing here"** it recommends: rewriting the plugin hook's `@plugin:<agent>` refusal to
> name the Skill would now steer the model off the sanctioned route. Those strings are correct as
> shipped.
>
> **What moved out.** The orchestrator prose is issue #2075. The two agent-body edits are issue
> #2127. The ADR amendment is in flight as its own PR. **What is left here is the harness detector
> and the docs that state the old rule as fact.**
>
> **Scope: the reversal covers `research-exhaustiveness` and `proof-conclusion` only.
> `record-extraction` is NOT affected** — it acquires and triages the record before batching one
> extractor per record, so it stays a routed skill that does real work, and every statement about
> it in these docs is correct as written. This matters most in the docs half, where all three
> pairs share one table, one mermaid legend and one set of `remedy` strings: a sweep that
> "corrects the retired rule everywhere" will wrongly rewrite `record-extraction`, and **no test
> reads README prose, a mermaid legend, or a `remedy` string**, so nothing catches it.

**Decided (lead, 2026-09-24): credit a spawn by name.** A successful `Agent`/`Task` call whose `subagent_type`, after `strip_agent_namespace`, equals an arm's skill name puts that name in `invoked`, exactly as a `Skill` call does. There is no paired-agent map and no disk derivation. The option it beat was deriving the set from "an agent with a same-named SKILL.md", which drops each pair silently when its thin skill is deleted (issue #2738). As part of this task, write the rule and its reason into `docs/specs/guardrail-enforcement-spec.md` beside the two bypass shapes, as the sanctioned third shape. Keep the `PATTERNS` path-filter addition: the `DEDICATED_AGENT_NAMES` guard still needs it.

> **Reviewed 2026-09-23 before junior handoff.** Decision recorded above (2026-09-24); the corrections below supersede the text wherever they disagree.

> **Re-reviewed 2026-09-23. Corrections that supersede the text below wherever they disagree.**
>
[...]

PR #1898, opened 2026-08-25
deep dive: proof-conclusion — 9 findings, 7 validator requests, grading fixes (#1643)

### LINKED ISSUE #1643: deep dive: proof-conclusion
> **Reviewed 2026-08-24 before junior handoff.** Retargeted at the agent body created by the 2026-08-21 skill-agent pair conversion, corrected the flat-dimension count to 4 of 8, fenced the grep and the agent body against wrong fixes, and re-derived the cost.

**Touches:** eval/tests/unit/proof-conclusion/ (incl. rubric.md); packages/engine/plugin/agents/proof-conclusion.md and packages/engine/plugin/skills/proof-conclusion/SKILL.md, read-only

Audit `proof-conclusion` for defects nothing has caught. This starts from **no known

Did / Should / Gap.

## Read the agent, not the skill

`proof-conclusion` became a skill-agent pair on 2026-08-21 (PR #1819). `SKILL.md`
is now a 63-line routing stub; the doctrine you are auditing — the preconditions
gate, tier selection, form selection, tree encoding — is the 644-line
`packages/engine/plugin/agents/proof-conclusion.md`. **Build the Step 1
prohibition list from the agent body.** A list built from SKILL.md finds almost
nothing, and the dive comes back empty for the wrong reason.

Three consequences for Step 2:

- The artifact you grade is mostly **not** `text_response` — the skill only
  relays. The narrative lives in `output.file_changes` → `research.json` →
  `proof_summaries[].narrative_markdown`. The agent's tool calls (`research_query`,
  `research_append`, `tree_edit`) do appear in `output.tool_calls`.
- **Read every committed run log, not just the newest.** Across the five
  committed today, 7 of 21 tests change outcome — compare `tests[].outcome` over
  `eval/runlogs/unit/proof-conclusion/v1_*.json` (measured 2026-08-24). One log is
  a poor sample here; the pair conversion roughly doubled the suite's instability
  (`docs/skill-to-agent-pair-conversion.md` §1). Retention keeps the newest five,
  so re-derive rather than quoting this count.
- The newest log predates a 5-line agent edit that landed 2026-08-23 (PR #1832)
  repointing a retired skill name to `record-extraction`. Do not file a finding
  about which skill the agent recommends next — already fixed.

## Start with this grep

```sh
grep -l -iE '"[^"]*\bscore [123]\b' eval/tests/unit/proof-conclusion/*.json
```

**This skill has 3 such file(s).**

without writing the finding.

**The grep is a pointer, not a verdict.** Two shapes match it and only one is a
defect. Rewrite a clause that names *the finding this fixture is known to
produce* — `no-image-claim-without-tool-confirmation`'s "this exact overclaim
happened in a real prior run of this fixture" is the worked antipattern. **Leave
a clause that is a normalization or scope rule preventing a false fail** — e.g.
`citation-copied-verbatim`'s "strip footnote/endnote markers, then compare", or
the same file's "silence on image accessibility is correct". Each of those was
added to close an observed misgrade; deleting one re-opens it, and a paid run is
what tells you.

## This skill's numbers

Re-derive them; a pasted figure here has already been wrong once.

| | command | value on 2026-08-24 |
|---|---|---|
| Tests in the suite | `ls eval/tests/unit/proof-conclusion/*.json \| wc -l` | 21 |
| Non-discriminating dimensions | `make judge-report SKILL=proof-conclusion` (free, reads committed JSON) | **4 of 8** |
| `judge_context` files naming a score branch | the grep above | 3 |

The four flat dimensions are `base/Tool Arguments`, `rubric/Evidence
completeness`, `rubric/Proof-conclusion fit`, `rubric/Tree encoding`. **You own
three.** `base/Tool Arguments` belongs to the base rubric, which is global — post
the problem and your proposed wording and let the lead call it, per the guide's
"What not to do". `rubric/Tree encoding` is `n=9 +8 N/A`: check whether it is flat
or merely thinly sampled before calling it non-discriminating.

## What to hand back

**Do not edit the agent body or `SKILL.md` in this task.** A doctrine finding on
the agent is handed back as a request, not made here: ADR-0011 and
`docs/skill-to-agent-pair-conversion.md` §2 both say a rule that must hold goes
into the writer tool before it goes into prose, and the agent body is contested
by open issue #1851, which is routed `senior`. Your made-fixes are `judge_context`
and `rubric.md` only.

Fourteen validators already exist at `eval/harness/validators/test_proof_conclusion.py`
(`grep -c '^def test_' eval/harness/validators/test_proof_conclusion.py`). Read
them before writing a request so you extend one instead of duplicating it. Per
issue #1788, a validator that never runs still records `passed: true` — so name,
in the request, the committed run it should fire on.

## Cost

The run-log snapshot embeds the tests, `rubric.md`, `SKILL.md` **and the agent
body**, so editing any of them flips it inactive and `check_runlogs.py` blocks the
PR until a fresh `make eval-skill SKILL=proof-conclusion` lands.

**The run is already owed.** PR #1832 edited the agent body on 2026-08-23 and its
commit message arms the re-run; the newest log is inactive on `main` today, before
you touch anything. Your edits ride on a run that was going to happen regardless —
which is the reason to batch every finding into it.

Last full run: `$5.64` and ~41 min wall clock (`totals.total_cost_usd` and
`totals.duration_ms` in `eval/runlogs/unit/proof-conclusion/v1_2026-08-21_19-34-21.json`),
plus the annotation — 5 sampled tests and ~3 sentences since PR #1637.

Emruthwill:
Closes #1643.

Audited the **644-line agent body** (`packages/engine/plugin/agents/proof-conclusion.md`), not the 63-line routing stub, against all five committed run logs — 21 tests, 105 runs, 65 narratives. Then re-ran the suite, and the re-run plus the genealogist annotation changed the answer in four places. Guide followed: `docs/skill-deep-dive-guide.md`.

> **Read this first.** **Two runs, both annotated.** The owed run landed as `v1_2026-08-25_13-34-10` — 21 tests, **$6.63**, 17 pass / 3 partial / 1 fail. It **refuted one of my findings, showed one of my two fixes to be ineffective, and surfaced two findings my sweep missed.** Every figure in this PR and in the findings doc is read off that log, which is still in the tree.
>
> A **second** run, `v1_2026-08-26_13-43-33` — **$6.54**, 17 pass / 3 partial / 1 fail — was forced by @mercyokum's review item 3: removing the measured-inert rubric block changed a snapshot file. It is the active log and it **independently reproduces the findings** — 13 of 13 narratives over their declared budget, `Proof-conclusion fit` still flat at 3 over 17 gradings, `resolved_conflict_ids` still `[]` on 001/002/020 with Evidence completeness 3. **That also completes F1's A/B**: the dimension is flat with the rubric block present *and* with it removed, so the prose was never doing anything. Both blocking runlog rules are green.

**9 findings, 7 validator requests.** Full Did / Should / Gap in [`docs/deep-dives/proof-conclusion-findings-2026-08-25.md`](docs/deep-dives/proof-conclusion-findings-2026-08-25.md).

## Numbers, re-derived

| | value |
|---|---|
| Tests in the suite | 21 |
| Existing validators | **13** — the issue body says 14 |
| Tests changing outcome across the five prior logs | **7 of 21** |
| `judge_context` files naming a score branch | 3 |
| Non-discriminating dimensions | **4 of 8** before → **3 of 8** after — **but the improvement is not mine; see the churn table** |

## Findings

- **F1** — **64 of 65 narratives exceed their declared `vehicle`'s word budget** (13 of 13 again on the re-run); 12 declared `statement` at 2.6–4.0× the ~150-word bar. `Proof-conclusion fit` scored 3 on every one. **My fix for this failed — see below.**
- **F2** — **A conclusion can be written and its question left open, and the run passes** (015, `v1_2026-08-21_19-20-04`). `research_append` enforces only the converse.
- **F3** — **A raw `Grep`/`Read` of `research.json` passes** in 3 of 5 runs of test 002 — in the suite whose validator docstring names that exact defect. The assertion only checks that *some* `research_query` call exists, and its tag gate skips the offending test.
- **F4** — **`tree_edit`/`tree_correct` writes split across calls** where §6 requires one all-or-nothing batch. 019's rationale counts the split correctly (*"The tree_correct **calls**"*) and still scores 3.
- **F5** — **The unresolvability rule is written per-sentence and graded per-narrative**; 13 of 15 forbidden-modal sentences unpaired. Deliberately not "fixed" — see non-fixes.
- **F6** — ~~`Tree encoding` has no population that can fail it~~ — **REFUTED by the re-run.**
- **F7** — **A citation gap weighed as an evidentiary gap, setting the tier** (018). From the annotation, not my sweep. Does not convert.
- **F8** — **`research_append` retried until accepted rather than fixed** (007, four calls, two rejections). From the annotation. **Converts — V6.**
- **F9** — **`Evidence completeness` grades prose and never the structured evidence fields.** The third flat dimension the issue assigned me, which my first pass missed. **Converts — V7.**

## F1's fix did not work

`Proof-conclusion fit` is **still flat at always-3**, over 17 gradings. The criterion *is* read — every rationale now cites the budgets, which none did before — and then overridden. Test 020:

> "The narrative is approximately 1,200 words — **well within** the ~300–500 word Summary budget…"

1,200 is not within 300–500. The count is *correct* this time, so it was never arithmetic: the judge reaches for 3 and writes whatever bridges to it. Telling it to score on shape instead didn't help — it asserts the shape.

**I picked the wrong instrument.** Per ADR-0011, a word count against a closed enum is decidable from the documents alone, so it is mechanical. **V1 is the fix; do not count the rubric edit as closing F1.** I left the edit in only because the visible budget-citation is what made the override diagnosable.

F1's second example is also weaker than I first wrote: test 017 is 383 words against ≤150, but **11 sentences under one heading** — a wordy Statement, not a mislabelled one, and the judge's 3 is defensible there. 020 is the real case.

## F6 is refuted

`Tree encoding` scored **1** on test 012 and now reads `varies [1, 3]`:

> "concluded ps_001 at tier 'probable' but made **no tree writes** … The conclusion lives only in narrative_markdown … This is a found-but-lost result."

A tree defect **does not need planting in a fixture — the agent produces it behaviourally.** Five logs had never sampled that, and I read a sampling accident as a structural impossibility. **No fixture work is owed**; struck from the handback list. Note this is **F2's failure mode in a second place** — a conclusion written, its consequence not landed: there the question, here the tree.

## The 4-of-8 → 3-of-8 improvement is not mine

Composition churned in both directions under the new judge prompt, across dimensions I never touched:

| dimension | before | after |
|---|---|---|
| `base/Tool Arguments` | flat | varies [2,3] |
| `rubric/Evidence completeness` | flat | varies [2,3] |
| `rubric/Tree encoding` | flat | **varies [1,3]** |
| `rubric/Proof-conclusion fit` | flat | **still flat** — the one I edited |
| `base/Completeness` | varies [1,3] | **now flat** |
| `rubric/Narrative standalone` | varies [2,3] | **now flat** |

Two dimensions went flat that were not before. Net 4→3 is coincidence.

## Fixes made — grading only

Agent body and `SKILL.md` are **untouched**, per the issue and ADR-0011.

| file | change |
|---|---|
| `.../no-image-claim-without-tool-confirmation.json` | Rewrote the `judge_context` clause that handed the judge the finding (*"this exact overclaim happened in a real prior run of this fixture"*) plus the phrase list to grep for. **Verified no regression on the re-run**: Correctness still 3, reasoned from project data. Clause 3's false-fail guard left intact. |
| `eval/tests/unit/proof-conclusion/rubric.md` | `Proof-conclusion fit`: declared-form-vs-artifact criterion. **Measured as ineffective — see above.** Stale `SKILL.md` §6 pointer → agent body. |
| `.../bounded-death-encoded-not-collapsed.json`, `.../research-query-gather-evidence.json` | Stale `SKILL.md` pointers → agent body. Cosmetic by `snapshot.py`'s own definition. |

**Grep hits adjudicated:** 015 and 017 **left alone** — their score-branch clauses are the normalization / false-fail kind the issue says to keep. Only 016 was the antipattern.

**Deliberately no rubric edit for F9**, on F1's measured evidence: a set comparison between two fields in the same file is mechanical, and F1 is this dive's own demonstration that a mechanical rule in rubric prose gets cited and then scored 3 anyway. Another rubric edit would also flip the snapshot and owe a second ~$6.6 run plus another annotation for a change already measured ineffective in this shape. V7 is the fix.

## Validator requests

Seven. Three deliberately **not** tag-gated — tag-gating is what let F3 through. **V1, V2 and V6 first.** V7 splits: part (b) is unambiguous and ships alone; part (a) is gated on a spec-wording decision (below).

## Deliberate non-fixes — no issues filed, routing is yours

- **F5** would need the agent body to say per-narrative. Senior lane, already contested by #1851.
- **F2's doctrine half** — §7 has no case for "concluded, but the exhaustiveness declaration is not credible", so the agent expresses it by silently skipping the resolve. Senior lane, #1851.
- **F7** — doctrine, does not convert. The agent got the *phrasing* right and made the *reasoning* error the rule exists to prevent.
- **`base/Tool Arguments` grades `tool_calls` only** and is blind to a builtin reached for in place of an MCP tool (F3). It carries real signal in F8 and misses real defects in F3/F4 — the argument for moving its countable parts into validators. Base rubric, so lead's call; proposed wording in the doc.
- **V7(a) needs a spec decision.** `resolved_conflict_ids` is specified as conflicts "resolved in this proof", which this skill is forbidden to do — so `[]` is a defensible literal reading, while the same spec's canonical example and audit-trail language say otherwise. Settle the wording or V7(a) enforces a reading the spec doesn't license.

## Also included

- [`docs/deep-dives/proof-conclusion-prohibition-list.md`](docs/deep-dives/proof-conclusion-prohibition-list.md) — 71 checkable rules from the agent body, for the next auditor. Item 25 carries a pointer to F7, because **a prohibition list built from phrasings will keep missing that class** — my scan searched for the forbidden phrase and the agent used the permitted one.
- **A read-out gotcha:** `output.tool_calls[]` keys on `tool`/`args`, not `name`/`input`. A scan against the wrong keys returns clean *silently* — which is how F3 and F4 survived my own first sweep.
- `DEFAULT_KEEP_CANDIDATES = 5`, so the re-run pruned `v1_2026-08-21_04-34-52.json` — a log the findings cite by name. Flagged in the doc so the dangling citations don't read as errors.

## Verification

- Preflight scratch run ($0.41) before the paid run, per the never-n=1 rule — it caught an arithmetic failure mode in my own rubric wording before the full run.
- Gate re-checked after merging main: **"All runlog rules satisfied."**
- 381 harness unit tests pass; 576 packaging tests pass, including all 130 `doc-links` checks over the new docs.
- The 5 failures in an unscoped `vitest` run come from a stale `.claude/worktrees/pr1539` tree, not this change.


mercyokum (changes requested):
**Requesting changes**

Really thorough dive — the self-correction on F6, the honest "my F1 fix didn't work" write-up, and the careful churn-table analysis are all a model for how these should read. The issues below are all minor-to-moderate and confined to the doc/eval layer (nothing here touches the agent's actual behavior), but a few are worth fixing before merge.

## 1. Two miscounts in the "Validator requests" summary

- `docs/deep-dives/proof-conclusion-findings-2026-08-25.md` line 517: "All five are new" — the section that follows lists **seven** validators (V1–V7).
- Line 525: "Two of the five are deliberately not tag-gated" — actually **six of seven** (V1, V2, V3, V5, V6, V7) explicitly say "do not tag-gate" in their own spec text; only V4 doesn't mention it.

This one has real downstream risk: this PR's own F3 finding is that tag-gating is exactly what let a defect through undetected. If the follow-up validator PR is built off this summary line rather than each V-section's own instruction, it could tag-gate checks that are explicitly supposed to run unconditionally — reproducing F3's failure mode.

## 2. The 016 `judge_context` rewrite drops its score anchor

`eval/tests/unit/proof-conclusion/no-image-claim-without-tool-confirmation.json` line 16: the rewrite removes the explicit "score 1 if the narrative says 'accessible'/'viewable'/'available'" instruction and the CRITICAL marker, replacing them with softer prose that names no score. The only scoring language left in the file is "should be penalized," unspecified between 1 and 2.

This cuts against this PR's own F1 lesson — a judge given vague prose instead of a concrete anchor tends to drift off the rule. Nothing re-runs 016 against a synthetic overclaim to confirm the reworded clause still triggers a 1; only the false-positive direction ("no run makes an unsupported claim") was checked. Worth a scratch run against an injected violation before merging, the same way F1's rubric edit was scratch-tested.

## 3. The new rubric criterion contradicts itself, and I checked — it resolves leniently

`rubric.md` line 58 says "Score the mismatch" for a narrative several times over budget, then six lines later says "reserve the mismatch score for a form that is plainly wrong" and never write a number. I recomputed declared-vehicle-vs-word-count from the committed run log's `file_changes` diffs: all 13 of 13 gradable narratives exceed their budget and all still scored 3 — matching what the doc itself already says elsewhere ("still flat at always-3"). Since the doc concludes this edit shouldn't count as closing F1, worth either trimming this block to a short pointer at V1, or adding a note in the block itself that it's known-ineffective pending V1, so a future reader of `rubric.md` alone doesn't mistake it for a working check.

## 4. The stale-`SKILL.md`-pointer sweep didn't quite finish

- `eval/harness/validators/test_proof_conclusion.py` still has 6 "SKILL.md §N" references (lines 232, 264, 406, 411, 414, 425) — including one live in a pytest assertion-failure message (line 264) that a developer sees directly in CI output.
- `eval/fixtures/scenarios/wilkins-bounded-death/README.md` and `eval/fixtures/scenarios/person-a-death-cert-not-indexed/README.md` also still cite the old location.

## 5. Prohibition list has no commit pin

`docs/deep-dives/proof-conclusion-prohibition-list.md` is dated only to "the 2026-08-21 conversion (PR #1819)," with no pinned commit of the agent body's current text — sibling prohibition lists (person-evidence, search-images, search-wikipedia) all pin a specific commit hash. The doc's own text confirms the agent body changed again in PR #1832 (2026-08-23), which isn't reflected in the list's dating — so a future auditor can't tell if any of the 71 rules are already stale.

## 6. F7 falls out of the tracking table

F7 is explicitly self-classified as a doctrine finding ("Gap: doctrine, and my own sweep's blind spot") but, unlike its siblings F2 and F5, it never gets added to the closing "Handed back, not fixed here" table or an issue reference. Worth adding it there so #1851 (or whoever picks it up) doesn't have to re-read the full doc to find it.

## Minor / optional

The new "Proof-conclusion fit" rubric block (~22 lines) ships despite the PR's own measurement showing it changes nothing — kept because "it costs nothing," but it is read by the judge on every future run of this dimension. Not blocking, just flagging in case a leaner version (or an explicit "known ineffective, see F1" comment) is worth it before this becomes the kind of coverage-that-isn't-coverage this repo's CLAUDE.md warns about elsewhere.

mercyokum (approved):

EdmondOware (approved):
Approving, and first correcting my own comment above.

### My note about items 2 and 3 was wrong

I said they "missed the batch". They did not. You withdrew both deliberately at `e23e244d`,
43 minutes before I posted, and the commit message explains why. I read the tree, saw the edits
absent, and inferred the reason instead of reading the commit that gave it. That is the same
mistake I have been flagging in PR bodies all week, so it is fair for it to land on me.

Your reasoning is better than my suggestion was, on both:

**Item 3.** Reverting the block entirely rather than trimming it is the stronger answer. A trim
would have added new judge-visible text that had never been measured, so it could not honestly
be called neutral, and the misleading block does not ship at all this way. The A/B settles it:
flat at always-3 across five logs without the block, n=17, and flat at always-3 in the run with
it, n=17.

**Item 2.** Right to revert the `score 1` anchor. It is intended to move a grade, so calling it
cosmetic would misuse the waiver, and it is unreachable on this corpus anyway — zero affirmative
overclaims across five 016 runs, both regex hits explicit negations. The issue's actual
requirement, dropping the finding-tell and the phrase list, is still satisfied. My earlier
comment endorsing removal of the anchor wholesale overshot; @mercyokum was closer to right, and
your resolution is better than either.

Also worth recording that you corrected your own cosmetic argument on the fixture READMEs after
finding that `_load_scenario_readme` threads them into `judge/prompt.md:216`. Catching that they
are judge-visible, and narrowing the waiver claim because of it, is the part of this exchange I
would point other authors at.

### What I verified for the approval

All six of @mercyokum's items are closed, including the two I had not checked earlier: the
prohibition list now pins `94292657d` with the `git diff` a future auditor should run, and F7
appears in the tracking table at line 738.

The eval side: `v1_2026-08-26_13-43-33` is active at 72 snapshot entries with zero mismatches,
17 pass / 3 partial / 1 fail, annotation covering the whole review sample, zero failing checks,
and the rubric diff is now a single pointer fix. The rubric's word budgets quote the agent body
verbatim at lines 208 to 210.

One thing my approval does not cover, stated so the record is honest: I did not read the
713-line findings doc line by line. @mercyokum did, citing line 517 and line 525 and catching
two miscounts in it, so the deliverable has had a genealogist's read — just not mine.

Issue #1666, opened 2026-08-16
deep dive: record-extraction


DallanQ:
> **Reviewed 2026-08-27 before junior handoff.** No decision needed; the numbers were stale
> twice over and the folded-in deletion needs a second file the body never named.

> **Corrected twice — read this, not the table below.** The line reading **"No dead
> dimensions — this rubric discriminates, which is rare"** is false. A 2026-08-23 audit
> corrected it to "2 of 6 flat"; that correction is now stale too, because a newer run log
> (`v1_2026-08-24_17-40-15.json`) landed after it. **Do not quote either figure — re-derive
> it**, which is free and takes seconds:
>
> ```sh
> make judge-report SKILL=record-extraction   # reads committed JSON only; no API calls, no cost
> ```
>
> As of 2026-08-27 that command reports **1 of 6** non-discriminating —
> `rubric/Assertion atomicity`, always 3 (n=24). `rubric/Informant identification`, flat on
> 08-23, now varies [2, 3]. Suite size is **28**, not 27
> (`ls eval/tests/unit/record-extraction/*.json`).
> A flat dimension is not automatically a defect; it is where to look first.

**Touches:** eval/tests/unit/record-extraction/, packages/engine/plugin/skills/record-extraction/SKILL.md

Audit `record-extraction` for defects nothing has caught. This starts from **no known
failure** — you are looking for things the suite passes today and should not.

**Read [`docs/skill-deep-dive-guide.md`](../blob/main/docs/skill-deep-dive-guide.md) first.** It is the
procedure: build the skill's prohibition list, read transcripts *before* scores,
spend the time on the quiet passes, place each finding in a lane, and write it
Did / Should / Gap.

## Start with this grep

`judge_context` that spells out the score branches tells the judge what to find,
and the judge takes the branch it was shown. Run this first — it is free and it
points at the files worth reading closely:

```sh
grep -l -iE '"[^"]*\bscore [123]\b' eval/tests/unit/record-extraction/*.json
```

**This skill has 0 such file(s).** No hits — that is a clean result for this pattern, not a reason to skip the dive; the grep only catches the narrowest shape.

The worked example is `ut_search_records_015`, whose `judge_context` read
*"Score 2 if it uses Mullen but doesn't explain why"* — which is exactly what
happened, and the judge scored 3 anyway. Rewrite a hit to name the *dimension*
without writing the finding.

## This skill's numbers

[...]

PR #1903, opened 2026-08-25
engine-test: block built-in tool-name drift in agent bodies (#1635)

### LINKED ISSUE #1635: Agent-body tool-name drift is blind to built-in tools like ToolSearch
> **Reviewed 2026-08-24 before junior handoff.** Rewritten: the agent-body drift check already exists and is CI-wired, so the task is now widening its vocabulary to built-in tool names and making one arm blocking, not building a new extractor.

**Touches:** packages/engine/mcp-server/tests/packaging/agent-tool-names.test.ts

## The check already exists — this widens its vocabulary and makes one arm blocking

`eval/harness/scripts/check_rubric_tool_drift.py` already scans every agent body
for tool names that are in neither its `tools:` nor its `disallowedTools:`
frontmatter (`agent_body_mentions`, unit-tested in
`eval/harness/tests/unit/test_check_rubric_tool_drift.py`). It runs from
`.github/workflows/check-runlogs.yml`. **Do not rebuild it.**

It missed `record-extractor.md:834` ("load it via ToolSearch") for one reason:
its vocabulary is the `.mcpb` manifest's MCP tool names, and `ToolSearch` is a
built-in. Built-in tool names in agent bodies are checked by nothing, and the
check that is closest is warn-only, so CI stays green either way.

## What to build

In `tests/packaging/agent-tool-names.test.ts`, add one blocking assertion:
for each agent in `packages/engine/plugin/agents/`, no **built-in** tool name may
appear in the body unless it appears in that agent's `tools:` or
`disallowedTools:`. Reuse `extractList()` and `bareOrBuiltin()` already in that
file — do not re-parse frontmatter.

**Vocabulary — built-in names only, and only those that are not ordinary English.**
`Read`, `Write`, `Edit`, `Task` are excluded: they are verbs in these bodies
("**Read** exactly ONE image per invocation", "**Write** the narrative markdown").
Scan for: `ToolSearch`, `Glob`, `Grep`, `Bash`, `WebFetch`, `WebSearch`,
`NotebookEdit`, `TodoWrite`, `SlashCommand`, `AskUserQuestion`, `MultiEdit`.
Record each exclusion with its reason at the site, the way
`COMMON_WORD_EXEMPTIONS` in `check_rubric_tool_drift.py` does.

**Match bare occurrences, not just backticked ones.** The defect that motivated
this is `ToolSearch` in running prose with no backticks — a backtick-scoped
extractor is green on it.

**No "imperative context" classification.** It was in the original ask and is
dropped. Measured 2026-08-24 at `7a30786b`, the built-in vocabulary above
matches **exactly one line** across all six agent bodies:

    grep -nE '\b(ToolSearch|Glob|Grep|Bash|WebFetch|WebSearch|NotebookEdit|TodoWrite|SlashCommand|AskUserQuestion|MultiEdit)\b' \
      packages/engine/plugin/agents/*.md

A plain mention rule is therefore already precise, and a classifier tuned
against these six bodies would only ever match the phrasing of the one defect it
was written for. (The MCP-name arm is the opposite case and stays where it is:
ignoring the deny exemption, 22 (tool, agent) pairs name a tool the agent does
not hold, and every one of them is legitimate — a deny being explained, a
delegate's tool, or a search another skill runs.)

## Change no file under packages/engine/plugin/

The one live hit — `record-extractor.md:834-837` — is **not yours to delete**. It
was ruled on 2026-08-23 and folded into issue #1666 (`deep dive:
record-extraction`), which already pays that directory's eval run. Editing that
file here flips the record-extraction run log inactive, so `check_runlogs.py`
blocks the PR until a fresh `make eval-skill SKILL=record-extraction` run plus a
genealogist annotation lands — roughly $8-12, 45-65 minutes of wall clock, plus
human hours, for a four-line deletion.

Seed it into an allow-list instead, keyed on **(relPath, trimmed line text)** —
not a line number; these files move by dozens of lines a week. Copy the shape and
the stale-entry check from `tests/packaging/no-issue-refs.test.ts`, which landed
this exact pattern for agent bodies. Its entry leaves when issue #1666's PR
deletes the passage, in that same PR. This mirrors the ruling on the closed issue
#1620: land the lint alone, let each violation ride its own skill's paid run.

## Prove it fails

Per CLAUDE.md § "A new lint must be proven to fail": add `Glob` to an agent body
that does not grant it, watch `make engine-test` go red, remove it. Then check
these three, each of which is a way this lint silently passes:

- [ ] Fails on a bare, un-backticked built-in name (the defect's actual form)
- [ ] Fails on a built-in named in the **body** of an agent whose frontmatter
      grants it nowhere — and **passes** when it is granted (`gps-mentor` grants
      bare `Read`), so the frontmatter comparison is real and not a blanket skip
- [ ] A stale allow-list entry (text no longer in its file) fails

## Scope

- **Agent bodies only.** `packages/engine/plugin/skills/*/SKILL.md` is out:
  `allowed-tools` is a grant, not a restriction (`docs/architecture.md` §5.1), the
  main thread holds every tool, and the two skill bodies naming `ToolSearch`
  (`record-extraction`, `forget-and-rederive`) are correct.
- **The MCP-name arm stays warn-only** in `check_rubric_tool_drift.py`. Promoting
  it would buy five permanent allow-list rows for five legitimate mentions
  (`python3 eval/harness/scripts/check_rubric_tool_drift.py`, agent lines) and no
  signal.

## Verify

`make engine-test`. In CI via `.github/workflows/engine-tests.yml`, which runs on
every pull request and whose scope step already covers
`packages/engine/plugin/**`. Not `make test-js` — the pnpm workspace excludes
`packages/engine/**`.

## nothing-checks

Register entry, narrowed: **a built-in tool name (`ToolSearch`, `Glob`, `Bash`)
named in an agent body is checked by nothing** — the drift lint's vocabulary is
MCP tools only, and it is warn-only. The MCP-name half of this gap is covered,
warn-only, and is not what this issue tracks.

---

## Original issue

**Touches:** packages/engine/mcp-server/tests/packaging/agent-tool-names.test.ts

Nothing checks that a tool an agent body names **imperatively** is a tool that agent can
actually call. Add that check.

### The defect that proved the gap

`record-extractor.md:834-837` instructed the agent to recover a deferred persistence tool
via `ToolSearch` — a tool it does not hold and cannot hold, since its `tools:` list is
MCP-only and carries no built-ins at all. `grep -n "  - ToolSearch"
packages/engine/plugin/agents/*.md` returns nothing across all five agents.

The instruction was billed on every invocation of the most-invoked agent in the plugin and
could never execute. It sat because no check looks for this shape.

**That deletion is not this issue.** It was ruled on 2026-08-23 and folded into issue
#1666 (`deep dive: record-extraction`), which already pays that directory's eval run. The
measurement behind the ruling is in this issue's comment thread.

### What to build

For each agent in `packages/engine/plugin/agents/`, extract tool-shaped identifiers
appearing in **imperative context** in the body and assert each resolves against that
agent's own `tools:` allowlist.

Two things the check must get right, both already solved in the natural home:

- **The three MCP spellings.** `mcp__genealogy__x`,
  `mcp__remote-devices__Genealogy_Research__x`, and `mcp__Genealogy_Research__x` are the
  same tool. `agent-tool-names.test.ts` already parses every agent's frontmatter and
  derives both `display_name`-based prefixes from the manifest — reuse that, do not
  re-derive it.
- **Built-ins stay bare.** `Read`, `ToolSearch`, `Glob` carry no prefix. A bare entry in a
  `tools:` list is a legitimate grant; `agent-tool-names.test.ts` filters on
  `.startsWith("mcp__")` before any prefix check, so bare names already pass untouched.

Natural home is `agent-tool-names.test.ts` itself. Issue #1620 proposes a different
agent-body lint in the same area — worth building together if both land.

### Prove it fails before landing it

Per `CLAUDE.md`: name a tool in an agent body that the agent does not hold, and watch the
check go red. A lint that cannot fail reads as coverage and is worse than no check.

The specific trap here: the identifier extraction must not be so narrow that it only
matches the exact phrasing of the defect that motivated it. Test it against at least one
imperative form that is worded differently.

### Why this generalizes

`check_rubric_tool_drift.py` already does the analogous check for rubric prose — this is
the agent-body side of the same idea. It is also the rot the `Read`-based reference
experiment (issue #702) would have left behind after its revert.

### nothing-checks

The register entry is the gap itself: an agent body can instruct the agent to use a tool
it cannot call, and CI stays green. Verified 2026-08-23 — one such instruction existed and
no suite noticed.

francis-2008-happy:
## Summary
Closes #1635 (Normal tier). Adds one blocking assertion to `agent-tool-names.test.ts`: for each agent in `packages/engine/plugin/agents/`, a **built-in** (non-MCP) tool name in the body must be **granted** in that agent's `tools:`. A deny is not a second source of permission — it binds even under `bypassPermissions`, so a body naming a *denied* built-in is the same defect as naming an ungranted one. `check_rubric_tool_drift.py` already does the MCP-name side of this and is CI-wired, but its vocabulary is MCP tool names only, so a built-in named in a body (`ToolSearch`, `Glob`, `Bash`) was checked by nothing and the closest check is warn-only. That let `record-extractor.md` instruct "load it via ToolSearch" — a built-in it does not hold and its MCP-only `tools:` list cannot carry — bill on every invocation, never execute, and stay green in CI.

Vocabulary is the built-in names that are **not** ordinary English: `ToolSearch, Glob, Grep, Bash, WebFetch, WebSearch, NotebookEdit, TodoWrite, SlashCommand, AskUserQuestion, MultiEdit`. `Read`/`Write`/`Edit`/`Task` are excluded as verbs (reason recorded at the site), mirroring `COMMON_WORD_EXEMPTIONS`. Bare occurrences are matched, not only backticked ones.

## Review guidance

**Start here:** the new `describe("plugin agent bodies name no built-in tool the agent cannot call")` at the end of `agent-tool-names.test.ts`. It reuses the file's existing `extractList()` and `bareOrBuiltin()` (no re-parsing of frontmatter), scans the **body** only (`agentBodyStart` skips the frontmatter, where a granted built-in legitimately appears), and consults each agent's `tools:` alone — a deny is not a grant — so the grant is real, not a blanket skip. A `*.ts` change, so CODEOWNERS routes it to **both** senior-developers and senior-genealogists.

**Unsure about:** nothing outstanding. One deliberate call: the stale-allow-list check keys on "still a violation" (matches the `no-issue-refs.test.ts` precedent), identical in practice here since #1666 deletes the passage rather than granting the tool.

## Plan

**Didn't change:** no file under `packages/engine/plugin/`. The one live hit (`record-extractor.md`'s ToolSearch line) is **not** deleted here — it was ruled into #1666, which pays that directory's eval run; editing it here would flip the record-extraction run log inactive and force a paid re-run for a four-line deletion. It's seeded into an allow-list keyed on `(file, built-in word, trimmed text)` — not a line number — with a stale-entry check, so it leaves in #1666's PR. The MCP-name arm stays warn-only; `skills/*/SKILL.md` stays out of scope (`allowed-tools` is a grant, and the two skill bodies naming ToolSearch are correct).

**Acceptance check:** the built-in lint goes **red** when a bare, un-backticked built-in (e.g. `Glob`) is added to an agent body that doesn't grant it, and **passes** once that built-in is also granted in the agent's `tools:` (the frontmatter comparison is real). A stale allow-list entry goes red. And the `scans a non-empty body in every agent` guard goes red when `agentBodyStart` is forced to scan zero lines — so the lint can't silently scan nothing and pass (this replaced a weaker guard that only re-asserted "agents exist"; see the second commit). All four verified by mutation, and re-verified by an independent falsifiability pass. Runs in CI via `engine-tests.yml`, ungated on every PR — **not** `make test-js`, whose pnpm workspace excludes `packages/engine/**`.

**Deviated from the plan:** one, deliberately, in commit 3. #1635 prescribes that a body-named built-in must appear in `tools:` **or** `disallowedTools:`; this implements `tools:` only. The held-set was built from both, so a body naming a *denied* built-in passed — and a deny binds even under `bypassPermissions`, so that is the same defect as naming an ungranted tool. No agent denies a built-in today, so the corpus cannot exercise it; it is pinned synthetically instead, which is the only way a rule with no instance is held. It also diverges from `check_rubric_tool_drift.py`, whose agent-body arm subtracts the deny and whose module docstring argues for the union; that arm is warn-only and unchanged here. #1635 carries a comment recording the reversal.

The silent-zero guard strengthening (commit 2) was a gap found by a pre-PR self-review pass on this branch and folded in, per CLAUDE.md step 1.

## Test plan

- [x] I ran `make test-all`. It passes except 9 pre-existing **environmental** failures in `tests/unit/test_setup_feedback_case.py` (`unzip: command not found` on my box), which are unrelated to this diff (one `.ts` file) and are green in CI on this PR. 2730 passed, including the full engine vitest suite.
- [ ] ~Ran `make eval-skill`~ — **N/A**: this changes no run-log snapshot (no skill dir, delegated agent, `eval/tests/unit/`, or fixture touched — it is a packaging test under `packages/engine/mcp-server/tests/`).
- [ ] ~Refreshed routing negatives~ — **N/A**: no skill `description` or DO-NOT clause changed.

## Follow-on issues

**Folded in / filed instead:** Folded in — the silent-zero guard strengthening (the anti-vacuous-pass test was itself vacuous), fixed here in commit 2. Filed instead — none; the record-extractor ToolSearch line deletion is already tracked by #1666, and this PR's allow-list row leaves in that PR.


chesworthrm (commented):
Reviewed this against the merge with `main` (0 commits behind). Ran the engine suite and typecheck in a worktree at `be8c493e`: 2705 passed / 111 files, `tsc --noEmit` clean. All six required checks are green on this head.

I reproduced all four acceptance checks in the PR body. They hold — including the one that matters most, that granting the tool in `tools:` makes the lint pass, so the frontmatter comparison is real and not a blanket skip. The `engine-tests.yml` claim checks out too: it runs ungated, so this can't go dark on an agent-body-only diff. And there's no overlap with `check_rubric_tool_drift.py` — its vocabulary really is manifest MCP names only.

Good PR. Two blind spots in the new lint, neither of which breaks anything that ships today.

## 1. A denied built-in is treated as one the agent can call

This one I'd like fixed before merge, though the fix is a comment.

I put `Glob` in `gps-mentor.md`'s `disallowedTools:`, left `Use Glob to find the prior verdict files.` in the body, and updated the permission snapshot in the same edit the way the file tells you to. The **whole engine suite stays green** — 2705 tests. `check_rubric_tool_drift.py` doesn't see it either, and it's warn-only regardless.

A denied tool is one the agent explicitly cannot call. So that body line is billed on every invocation and can never execute, which is the exact shape this lint exists to catch.

Counting denies is still the right call — a body legitimately names a denied tool to explain the deny, and #1635 measured that. The problem is that the reason lives in the issue and not in the file, and the sentence at the top of the block says the opposite of what the code does. This file records why `Read`/`Write`/`Edit`/`Task` are excluded. This exclusion gets nothing.

```
You wrote:
// A built-in (non-MCP) tool named in an agent BODY must be a tool that agent can
// actually call — i.e. it appears in the agent's `tools:` or `disallowedTools:`
// frontmatter (issue #1635).

Change it to:
// A built-in (non-MCP) tool named in an agent BODY must appear in that agent's
// `tools:` or `disallowedTools:` frontmatter (issue #1635).
//
// `disallowedTools:` counts even though a denied tool is one the agent explicitly
// CANNOT call: a body legitimately names a denied tool in order to explain the
// deny, which #1635 found among the legitimate mentions on the MCP-name side.
// The cost is a blind spot — a body that says "use Bash" while Bash is denied
// passes this lint. No agent denies a built-in today.
```

I checked the last sentence: `Read` is the only bare built-in in any agent frontmatter, and it only ever appears in `tools:`. So this is latent, not live.

## 2. One allow-list row waves through every built-in on its line — optional

`isAllowedBuiltinHit` compares `(file, text)` but a hit also carries `word`, which is never checked. With `…load it via ToolSearch or Bash` on the allowed line and the row's `text` updated to match, the suite is green — one row suppressed both, and `Bash` is granted nowhere in `record-extractor.md`.

I want to be straight about how weak this is. If you reword the prose and don't touch the allow-list, both arms go red and the message tells you to delete the stale entry. That's correct and safe. Nobody gets steered into this by rewording. It only opens if someone later adds a row for a line that happens to carry two vocabulary words. Your call whether it's worth the two lines now.

```
You wrote:      const BUILTIN_BODY_ALLOWLIST: { file: string; text: string; reason: string }[] = [
Change it to:   const BUILTIN_BODY_ALLOWLIST: { file: string; word: string; text: string; reason: string }[] = [
```

```
You wrote:      return BUILTIN_BODY_ALLOWLIST.some((a) => a.file === h.file && a.text === h.text);
Change it to:   return BUILTIN_BODY_ALLOWLIST.some(
                  (a) => a.file === h.file && a.word === h.word && a.text === h.text,
                );
```

Plus `word: "ToolSearch",` on the existing row.

## Not blocking

Neither finding is a correctness bug in shipped behavior, so I'm leaving this as a comment rather than requesting changes. This still needs two approvals with at least one from a code owner — `*.ts` routes to senior-developers or senior-genealogists, either one satisfies it.

One thing I did not check, and no CI job can: whether a granted built-in actually binds in a live Cowork session. That gap is pre-existing and your file's header already says so.

chrisedeson:
**Verified by mutation:**
1. Inject bare `Glob` into gps-mentor body -- lint fails, reports exact file/line/word.
2. Grant the tool in frontmatter -- lint passes (`held.has(word)` skips granted tools).
3. Force `agentBodyStart` to scan zero body lines -- vacuous-pass guard fails.
4. Change allow-list text so it no longer matches -- stale-entry check fails.

**Vocabulary:** Exactly 1 match across all 6 agent bodies (record-extractor.md:834). The 4 excluded words (`Read`, `Write`, `Edit`, `Task`) appear 25 times as English verbs, confirming the exclusion is necessary.

**Code:** Clean. `BUILTIN_TOOL_VOCAB` is the right set. `agentBodyStart` correctly parses YAML frontmatter. Allow-list keys on `(file, trimmed text)` not line numbers. Stale-entry check prevents it from becoming an escape hatch. Reuses existing `extractList`/`bareName` with no duplication.

Approving.

chrisedeson (approved):

clack391 (approved):
Approving. This implements #1635 exactly, and I verified it by running it rather than reading the summary. The vocabulary is a character-for-character match to the issue's prescribed list, the lint is proven to fail on the shapes that matter (bare `Glob`, bare `ToolSearch`, and a backticked `` `Bash` `` all go red on a body that does not grant them), and the commit-2 silent-zero guard is real: forcing `agentBodyStart` to scan past end of file reddens the non-empty-body test. Two independent passes reconciled all of this. Nice catch folding in that guard strengthening from your own pre-PR review.

One note on the merge math: chrisedeson's approval does not satisfy the code-owner requirement (not on either senior team), so this still needs a code-owner approval. This one is it.

A few things worth folding in before it merges, none of them blocking, and all latent or a known trade-off of the design #1635 chose:

- **The header comment overclaims (chesworthrm asked for this one).** It says a body-named built-in "must be a tool that agent can actually call," but a tool in `disallowedTools:` is one the agent explicitly cannot call, and #1635 still wants it to count. The behavior is right; the sentence describing it is not. chesworthrm already wrote the replacement, so this is a two-line change.

- **The vocabulary lists `Bash` but not its companions.** `BashOutput` and `KillShell` are real built-ins an MCP-only agent cannot hold, and `ExitPlanMode` is a fourth, all meeting the same not-ordinary-English test the list uses. None of the six current agents would name them, so this is low risk, but the asymmetry is easy to close in three lines.

- **The no-classifier rule will false-fail a "do not use X" line.** A future body that says "you have no network here, never invoke WebFetch" goes red on correct prose. The issue chose the plain-mention rule deliberately and measured it against one line today, so this is fine for now, but a sentence in the header noting that such a line takes an allow-list entry would save the next author a confused CI failure.

- Optional: `isAllowedBuiltinHit` compares `(file, text)` but not `word`, and only the ToolSearch line is canaried, so the other ten vocabulary words could be dropped with the suite staying green. Both are minor and visible in a diff.

Because stale reviews are not dismissed on push, folding these in will not drop this approval, so none of it needs a re-review.

promise-emmanuel (changes requested):
My own finding, not yet on the PR
agentBodyStart is a byte-identical copy of frontmatterEnd in no-issue-refs.test.ts — same directory, same body, same trailing comment. I diffed them.


You wrote:      function agentBodyStart(lines: string[]): number {
                  if (lines[0]?.trim() !== "---") return 0;
                  for (let i = 1; i < lines.length; i++) {
                    if (lines[i].trim() === "---") return i + 1;
                  }
                  return 0; // unterminated — treat the whole file as body rather than skip it
                }

Change it to:   (delete it; export the existing one from no-issue-refs.test.ts's
                logic into tests/packaging/repo-paths.ts as `frontmatterEnd`,
                import it in both files, and keep the name `frontmatterEnd`)

promise-soar (approved):
Approving. Verified against the merge with main at c511a53c: engine vitest
2718/2718 across 112 files, tsc --noEmit clean, every required check green.

Both of chesworthrm's round-1 findings are resolved, and the deny one better than
asked — fixing the behaviour rather than the comment was the right read. I confirmed
it: Bash in gps-mentor's disallowedTools: plus a "Run the checker with Bash" body
line now fails by name, where it was green before.

The non-vacuity guard is the best thing in this PR and nobody asked for it. I
reproduced the measurement — allow-list row removed plus BUILTIN_TOOL_VOCAB emptied
was 75/75 green before that commit and is red now, same with the scanner regex
disabled. That is what keeps the lint honest once the record-extractor canary row
leaves with #1666, which was the one thing that would otherwise have taken the
coverage to zero while staying green.

I pushed the three text fixes myself rather than send you round three, per
task-lifecycle rule 5:

- 83a4d79d renames the test. e02a288e narrowed the held-set to tools: alone and
  updated both offender messages, but the it(...) title still said
  "absent from the agent's tools:/disallowedTools:", so a failure printed a title
  and a message contradicting each other on the point the rule turns on. Behaviour
  unchanged; re-verified 77/77 and that a bare Glob still reddens it.
- The PR body described the old rule in three places and said "Deviated from the
  plan: nothing". The deviation is real — #1635 prescribes the union in as many
  words and was reviewed before handoff — so it is declared now, using your own
  account from e02a288e's message. #1635 has a comment recording it too, so the
  prescription does not outlive the decision.
- c511a53c merges main; the branch was 5 behind and the merge state was BEHIND.

Dropping my round-1 duplication note as a blocker — agentBodyStart is still
byte-identical to frontmatterEnd in no-issue-refs.test.ts, same directory. Worth
lifting into repo-paths.ts sometime, but it should not have held this PR and I was
wrong to request changes on it.

One note for the record: clack391's approval reads 4a54a274, before the deny
reversal, and approvals are not dismissed on push here.

promise-emmanuel (approved):
Approving. Verified against the merge with main at https://github.com/PioneerAIAcademy/cowork-genealogy/commit/c511a53cb7583c0a5a3ff0c499aabd60f0477155: engine vitest
2718/2718 across 112 files, tsc --noEmit clean, every required check green.

Both of chesworthrm's round-1 findings are resolved, and the deny one better than
asked — fixing the behaviour rather than the comment was the right read. I confirmed
it: Bash in gps-mentor's disallowedTools: plus a "Run the checker with Bash" body
line now fails by name, where it was green before.

The non-vacuity guard is the best thing in this PR and nobody asked for it. I
reproduced the measurement — allow-list row removed plus BUILTIN_TOOL_VOCAB emptied
was 75/75 green before that commit and is red now, same with the scanner regex
disabled. That is what keeps the lint honest once the record-extractor canary row
leaves with https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1666, which was the one thing that would otherwise have taken the
coverage to zero while staying green.

I pushed the three text fixes myself rather than send you round three, per
task-lifecycle rule 5:

https://github.com/PioneerAIAcademy/cowork-genealogy/commit/83a4d79d03535c17faaaf1839192d36b763e8640 renames the test. https://github.com/PioneerAIAcademy/cowork-genealogy/commit/e02a288e0f2daf27b6c4def44b18ee7d15fa3e5d narrowed the held-set to tools: alone and
updated both offender messages, but the it(...) title still said
"absent from the agent's tools:/disallowedTools:", so a failure printed a title
and a message contradicting each other on the point the rule turns on. Behaviour
unchanged; re-verified 77/77 and that a bare Glob still reddens it.
The PR body described the old rule in three places and said "Deviated from the
plan: nothing". The deviation is real — https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1635 prescribes the union in as many
words and was reviewed before handoff — so it is declared now, using your own
account from https://github.com/PioneerAIAcademy/cowork-genealogy/commit/e02a288e0f2daf27b6c4def44b18ee7d15fa3e5d's message. https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1635 has a comment recording it too, so the
prescription does not outlive the decision.
https://github.com/PioneerAIAcademy/cowork-genealogy/commit/c511a53cb7583c0a5a3ff0c499aabd60f0477155 merges main; the branch was 5 behind and the merge state was BEHIND.
Dropping my round-1 duplication note as a blocker — agentBodyStart is still
byte-identical to frontmatterEnd in no-issue-refs.test.ts, same directory. Worth
lifting into repo-paths.ts sometime, but it should not have held this PR and I was
wrong to request changes on it.

One note for the record: clack391's approval reads https://github.com/PioneerAIAcademy/cowork-genealogy/commit/4a54a27471b5b3479abadd9ddf20d11022541f54, before the deny
reversal, and approvals are not dismissed on push here.

Issue #1791, opened 2026-08-20
[feedback] 2026-08-20T14:12Z


DallanQ:
**Zip:** https://drive.google.com/file/d/1zNpRdehQfn0joTjUI8wjbBEYwPJx4P7a/view?usp=drivesdk

```
make feedback-case ZIP=~/Downloads/feedback-2026-08-20T14-12-47-123677Z.zip
```

Workflow: `docs/alpha-feedback-guide.md`

- Submitted: 2026-08-20T14:12:47.123677Z
- Platform: web

florencemashipei:
Triaged: **positive report — no fix, no test.** Closing per `docs/alpha-feedback-guide.md` Step 4.

`Worked as expected: **Yes**`. The form's Yes path is deliberate — the tester guide tells testers a report about something that went well is worth sending and there is nothing more to fill in — so this is the channel working as designed, not a mis-filed bug.

**What the tester was actually doing**, from the Notes, because it is more useful than the flag:

They were evaluating whether a FamilySearch hint had sufficient evidence to accept, and asked the agent to read the record image and pull out anything relevant that had not been indexed — "including and especially the location." The records are in old German script. The agent could not read them. Their words: "This was more a curiosity of whether it could do it for me or speed up the process...which at this time it cannot so I will carry on. Thought it might be an interesting data point for you."

**On the bundle:** it carries no `research.json`, `tree.gedcomx.json` or `results/` — the project is "New research session", so there was no state to capture. Nothing to reproduce, and nothing wrong with the capture. A `_feedback/session-log.jsonl` transcript is present in the zip for anyone who wants the tool-level detail.

**The data point is not lost with this issue** — carried over to #1594 (`image_transcribe` on hard scans), which is the right home for it.

One observation for the process rather than this case: Step 4 reads `worked_as_expected: true` as "nothing went wrong, close and stop." Here the only substance was in the Notes box, which that rule would discard. "It behaved correctly and still could not do what I needed" is neither a bug nor nothing, and there is currently no slot for it.

florencemashipei:
Reopening. Per @DallanQ: the goal of a feedback case is to **fix** the problem where one can be fixed, not to file it well. Closing on `worked_as_expected: yes` was too literal a reading of Step 4 — the tester described a fixable defect in the Notes box, so this case is not done.

Taking it back to find out why the old German script could not be read, and to fix it if it can be.

**What I have verified in the repo before touching the bundle:**

[...]

PR #1905, opened 2026-08-25
setup-feedback-case.bat: read %~dp0 before parsing, and stop reporting success on a failed unpack (#1876)

### LINKED ISSUE #1876: setup-feedback-case.bat reads %~dp0 after shift, so every Windows feedback unpack fails
> **Reviewed 2026-08-25 before junior handoff.** No decision needed. Two corrections
> below: the reason given for why the `.sh` twin is safe was wrong, and verification was
> under-scoped — nothing in CI executes a `.bat`, and nothing below line 51 of this
> script has ever run at all.

**Touches:** `scripts/setup-feedback-case.bat` · `scripts/setup-feedback-case.sh` (error text only) · `packages/engine/mcp-server/tests/packaging/windows-wrappers.test.ts` · `docs/specs/feedback-case-spec.md` §3

## The bug

`setup-feedback-case.bat` resolves the repo root from its own location:

```bat
set "SCRIPT_DIR=%~dp0"
pushd "%SCRIPT_DIR%" >nul
for /f "delims=" %%i in ('git rev-parse --show-toplevel 2^>nul') do set "REPO_ROOT=%%i"
```

But that runs **after** the `:parse` loop, which calls `shift`. In cmd, `shift` moves `%1` into `%0` — so by the time `%~dp0` is read, `%0` is the **zip path**, and `SCRIPT_DIR` is the zip's directory.

Every invocation that passes a zip hits it, which is every real invocation:

```
> scripts\setup-feedback-case.bat "%USERPROFILE%\Downloads\feedback-2026-08-20T14-12-47-123677Z.zip"
Error: could not determine repo root from C:\Users\flore\Downloads\
Run this script from inside the cowork-genealogy repo.
```

The message sends you to fix the one thing that is not wrong — I was in the repo root, and `git rev-parse --show-toplevel` succeeds there.

## Why it matters

This is the **Windows entry point for the whole alpha-feedback loop**, and per CLAUDE.md this is a Windows-based genealogist team. `docs/alpha-feedback-guide.md` Step 2 names it as the Windows path, and the `make feedback-case` line pasted into every `[feedback]` issue is POSIX-only, so the `.bat` is what a genealogist on Windows actually runs. Right now it cannot unpack a case at all.

`scripts/setup-feedback-case.sh` does not have the bug, but **not for the reason this
issue originally gave.** It also resolves `SCRIPT_DIR` *after* its parse loop
(`setup-feedback-case.sh:54`, loop at 29-45). It is safe because bash `shift` never
touches `$0`. Ordering only matters in cmd — do not carry "resolve before parsing"
away from here as a general rule.

Hit while triaging #1791. I unzipped by hand to get on with it, which skips the git snapshot, the `.feedback-repo-root` marker and the skill junctions — so `make feedback-reset` and the case-folder Claude Code steps are unavailable on that path. #1763, #1784 and #1795 are queued behind this.

## The fix

Move the `SCRIPT_DIR` capture above the `:parse` loop in
`scripts/setup-feedback-case.bat`:

```bat
set "SCRIPT_DIR=%~dp0"
set "FORCE=0"
set "ZIP_PATH="
set "DEST_DIR="

:parse
```

and delete the now-duplicate `set "SCRIPT_DIR=%~dp0"` at line 47. The
`pushd "%SCRIPT_DIR%"` / `git rev-parse` / `popd` block on lines 48-50 stays exactly
as it is.

**Do not restructure the resolution.** In particular, do not collapse the pushd/popd
pair into `git -C "%SCRIPT_DIR%" rev-parse --show-toplevel`: `%~dp0` ends in a
backslash, and how `"...\scripts\"` survives git-for-Windows' MSYS argument parsing is
unverified here. One line moves; nothing else changes.

**Resolution is from the script's own location, not the cwd** — matching
`setup-feedback-case.sh:54`. Do not "fix" this by resolving the repo root from the
current directory instead: that passes the smoke test below (which runs from the repo
root) and silently breaks both the run-from-anywhere property
`scripts/windows/README.md` states for these scripts, and the marker's meaning when the
script is run from a second worktree (`scripts/link-worktree.sh`).
`docs/specs/feedback-case-spec.md` §3 guarantee 3 says "the repo checkout it was run
from", which is what leaves that door open — **tightening that phrase to "the checkout
the script itself lives in" is part of this task**, so the next reader cannot make the
same substitution.

## Verifying it

Nothing in CI executes a `.bat` — every workflow is `ubuntu-latest`
(`grep -rn "runs-on" .github/workflows/`), and
`eval/harness/tests/unit/test_setup_feedback_case.py` black-boxes the `.sh` only. So
this change needs both halves.

### 1. Add the static guard that would have caught it

In `packages/engine/mcp-server/tests/packaging/windows-wrappers.test.ts` (runs under
`make engine-test`), add a test over **every** `.bat` in the repo — `scripts/`,
`scripts/windows/`, `eval/`, and the two at the repo root — excluding `**/.venv/**`: if
a file contains a `shift`, then every `%0` / `%~dp0` / `%~f0` reference in it must
appear textually **before** the first `shift`. `shift` with no `/n` shifts from `%0`, so
any `%0` read after one returns the caller's first argument rather than the script.

That file is the right home: it already carries two guards for the same species of
cmd-semantics trap (`every wrapper resolves the repo root, not scripts/` and `every
npm/pnpm/npx invocation goes through call`). Note its existing helpers only read
`scripts/windows/`, so the new test needs its own file walk.

Per CLAUDE.md § "A new lint must be proven to fail": revert the one-line fix, watch the
new test fail, then re-apply. Exactly one `.bat` in the repo contains `shift` today
(measured 2026-08-25:
`grep -l -i shift $(find . -name '*.bat' -not -path '*/.venv/*')` → only
`scripts/setup-feedback-case.bat`), so the guard goes from one failure to zero and stays
cheap. If review would rather keep this PR to one line, file the guard as its own issue
with `--label nothing-checks` — do not drop it silently.

### 2. Run it on Windows — nothing below line 51 has ever executed

The bug has been present since the file landed (`5981a113`, 2026-06-04:
`git log --format='%h %ad' --date=short -- scripts/setup-feedback-case.bat`), so the
entire second half of the script is unexercised, not just the line being moved. Do not
merge on a macOS read-through: @florencemashipei has the Windows machine and the #1791
zip that reproduced this. A pass on one platform proves nothing about the other — the
same reason `docs/specs/guardrail-enforcement-spec.md` §6 keeps a Windows probe ("the
third time that asymmetry has mattered here").

From `cmd.exe`, walk all seven guarantees of `docs/specs/feedback-case-spec.md` §3:

- [ ] `scripts\setup-feedback-case.bat "%USERPROFILE%\Downloads\<zip>"` from the repo
      root unpacks to `%USERPROFILE%\feedback\<slug>\`
- [ ] `.feedback-repo-root` holds the **repo checkout's** path, backslashed
- [ ] `.gitignore` contains `.claude/`, and keeps any entry the zip shipped
- [ ] `git log --oneline` in the case dir shows exactly one commit, `imported`
- [ ] `.claude\skills\` is a real directory (not a link) holding **42** junctions — 27
      from `packages\engine\plugin\skills` + 15 from the repo's `.claude\skills`
      (measured 2026-08-25:
      `find packages/engine/plugin/skills .claude/skills -mindepth 1 -maxdepth 1 -type d | wc -l`;
      re-derive it rather than trusting 42, it moves whenever a skill is added)
- [ ] the tester's `user_prompt` prints at the end
- [ ] a re-run without `--force` refuses the non-empty dir; with `--force` it overwrites
- [ ] `scripts\reset-feedback-case.bat <case dir>` round-trips (spec §4) — that path was
      never exercised either, because the hand-unzip on #1791 skipped the baseline it
      needs
- [ ] run it **by absolute path from a directory outside the repo** (`cd /d C:\` then
      `C:\...\repo\scripts\setup-feedback-case.bat "<zip>"`) — this is the case a
      cwd-based fix would break, and it must still work
- [ ] copy the `.bat` alone into a non-repo directory and run it: the "could not
      determine repo root" error must still fire

## While you are here: the message that misled the reporter

After the fix the error text is still wrong about the cause — the cwd is irrelevant to
this resolution, so "Run this script from inside the cowork-genealogy repo" keeps
sending the next person to check the one thing that cannot be it. Reword both twins
together so they stay in parity — `scripts/setup-feedback-case.bat:53` and
`scripts/setup-feedback-case.sh:57` — to name the script's location instead, e.g. "This
script must live inside a cowork-genealogy checkout; %SCRIPT_DIR% is not in one."

## Not in scope

Whether a hand-written cmd twin of each feedback script is worth maintaining at all (25
feedback cases were triaged 2026-08-05..08-11 without this `.bat` working) is a separate
question — file it, do not settle it here.

Do not add `.bat` coverage to
`eval/harness/tests/unit/test_setup_feedback_case.py`: PR #1822 is editing that file,
and it is a bash black-box test that cannot run a `.bat` on the Linux runner anyway.

## Soft collision — issue #1880

Issue #1880 also edits `docs/specs/feedback-case-spec.md` (§2.2, adding the subagent
transcripts to the bundle contract) while this card edits §3. Different sections, so
sequenced rather than blocked — but if you are picking up feedback-pipeline work, take
both.

florencemashipei:
## Summary

- `setup-feedback-case.bat` resolved the repo root from `%~dp0` **after** the `:parse` loop. In cmd, `shift` moves `%1` into `%0`, so `%~dp0` was the zip's directory: every invocation that passed a zip — every real one — died with "could not determine repo root", naming the one thing that was not wrong. This is the Windows entry point for the whole alpha-feedback loop, so no genealogist on Windows could unpack a case. **Normal tier** (docs/task-lifecycle.md).
- Because that line sat above everything else, **nothing below it in the script had ever executed** — and verifying the fix proved it, surfacing two more failures that reported **success**: a corrupt zip printed `Imported to ...` and exited 0, and a failed git baseline did the same. Both are now fatal.
- Nothing read this file. It appeared in `windows-wrappers.test.ts` only inside an *exemption* list, and the existing helpers read `scripts/windows/` alone. Added a guard over every `.bat` in the repo.

Closes #1876.

## Review guidance

**Start here:** `it("no .bat reads %0 after it shifts")`. The first version of that guard **could not catch the bug it was written for** — it matched the first `%~dp0` in the file, which after the fix is the explanatory comment two lines above the assignment, and the comment stays put wherever the code moves. Ikennaya1 caught it. It now blanks REM/`::` lines before scanning. Worth confirming the comment-shadowing case yourself: move the `set` back below `:parse`, leave the comment, and watch it fail.

**Unsure about:** Nothing outstanding — both of my earlier questions were answered in review (chesworthrm: fixes 2 and 3 belong here, CLAUDE.md step 1; and the `.sh` twin is safe for a different reason than I claimed — see below). One thing worth a second opinion: `git commit` failing is now fatal, so a zip with nothing to commit becomes a hard error rather than a warning. I think that is right — an empty case directory is not a usable case — but it is a behaviour change beyond the reported bug.

## Plan

**Didn't change:** `scripts/reset-feedback-case.bat` — checked, it uses neither `shift` nor `%~dp0`, so the new guard passes on it untouched. No behavioural test harness for `.bat` files: `eval/harness/tests/unit/test_setup_feedback_case.py` drives the `.sh` only, and building a Windows-only equivalent is a larger piece than this bug — see "What is still not checked". The `.sh` twin's *ordering* — it resolves `SCRIPT_DIR` at line 54, after its parse loop at 29–45, and that is fine: bash's `shift` never touches `$0`. **An earlier version of this PR claimed the `.sh` was safe because it reads early. That was wrong.** "Resolve before parsing" is a cmd-only rule, and the guard's comment now says so. The `.sh` is touched only for the error-text reword below, in parity.

**Acceptance check:** `npx vitest run tests/packaging/windows-wrappers.test.ts` in `packages/engine/mcp-server`. Fails on `main`, which has the bug. Mutation-proven in the case that matters: **code moved back below `:parse` with the comment left in place** → `scripts/setup-feedback-case.bat: %~dp0 read after shift`; restored → 19/19 green.

Windows walkthrough, since nothing below line 51 had ever run — all seven of spec §3's guarantees plus the `reset-feedback-case.bat` round-trip:

| Case | Before | After |
|---|---|---|
| Real zip, from inside the repo | error names the **zip's** dir, exit 1 | resolves the repo root, exit 0 |
| Corrupt zip | `Imported to ...`, **exit 0** | real PowerShell reason, then `failed to unzip`, exit 1 |
| `user.email` unset | next-steps block, **exit 0** | `could not create the git baseline`, exit 1 |
| **Absolute path, cwd outside the repo** | — | exit 0; marker names the script's checkout (the case a cwd-based fix would break) |
| **`.bat` copied alone into a non-repo dir** | — | error path still fires, exit 1 |
| §3 guarantees 1–7 | never reached | all hold; `.claude/skills/` a real dir with 43 entries; one commit titled `imported` |
| `reset-feedback-case.bat` round-trip | never reached | restores to the imported commit; a planted dirty file is gone |

Exit codes were read from **cmd's own `ERRORLEVEL`**, not PowerShell's `$LASTEXITCODE` — the latter reports 0 for a script that writes to stderr, and cost me a false "guarantee 2 is broken" before I checked it properly.

**Deviated from the plan:** Two self-inflicted defects, both in the guard, both caught by mutation-testing rather than by reading. First, a `\b` written as a literal backspace byte, so its `shift` pattern never matched. Second — after that was fixed — the comment-shadowing hole above, which my own mutation test hid by deleting the comment along with the code. The second guards-the-reader assertion added in this round caught a third instance of the same escape mistake immediately.

## Test plan

- [x] `scripts\windows\test-all.bat` — green except one machine-local block: `Failed to spawn: pytest — An Application Control policy has blocked this file (os error 4551)`, which is policy on my box, not a repo problem. Re-ran the harness via `python -m pytest`. The `write-serialization.test.ts` `EPERM: symlink` failure I reported in the first round is **fixed on `main`** by `7e5a3771` ("use a junction on Windows so the aliasing test can run") and is no longer outstanding.
- [x] Engine: `npx tsc --noEmit` clean; full vitest **112 files, 2721 passed, 1 skipped, 0 failed**.
- [x] Harness: `tests/unit/test_setup_feedback_case.py` **10 passed, 1 skipped** — the suite that black-boxes the `.sh` I touched. Full harness suite on this merged tree: **2851 passed, 3 skipped**.
- [x] Nothing in the tree asserts either changed string — grepped for the old error wording and the old spec phrasing across `.py`/`.ts`/`.md`/`.sh`/`.bat`.
- [x] N/A — no skill run-log snapshot touched. No `packages/engine/plugin/skills/**`, `agents/**`, `eval/tests/unit/**` or fixture path is in this diff.
- [x] N/A — no skill `description` or DO NOT clause changed.
- [x] **New check proven to fail before merging** — twice, and it found real defects in my own work both times.

## What is still not checked

The guard is **static**. It catches `%0`-after-`shift` in any `.bat` and nothing else. The unzip abort and the git-baseline abort are covered only by the manual runs above; a regression in either would be silent again. Closing that needs a behavioural harness for the `.bat` the way `test_setup_feedback_case.py` covers the `.sh` — a bigger piece, not attempted here. Note also that **no CI job executes a `.bat` at all**: every workflow is `ubuntu-latest`.

## Follow-on issues

**Folded in / filed instead:** Fixes 2 and 3 folded in per CLAUDE.md step 1 — same file, same failure class, found while verifying, and confirmed as the right call in review. The two prose sites chesworthrm named are folded in as well. Nothing filed. Two corrections to #1876's original body, since fixed in its rewrite: "there is no harness for the `.bat`" (there is a static one, which excluded it) and "Nothing else needs to change".


chesworthrm:
Heads-up: issue #1876's body was rewritten a couple of hours after you opened this, from a pre-handoff review. Three things in it bear on this PR, and one answers your own open question.

**Your "unsure about" #1 — fixes 2 and 3 belong here.** CLAUDE.md § "Work you find along the way" walks the options in order and stops at the first that fits; step 1 is "fix it in the current PR", described as the right answer about two times in three, and the test is whether the work carries four-person overhead. Same file, same failure class, found while verifying, and both report success while failing — that is step 1, not step 4. The fold-in is right; keep it.

**The `.sh` twin is safe for a different reason than #1876 originally claimed**, and the issue body now says so. `setup-feedback-case.sh:54` also resolves `SCRIPT_DIR` *after* its parse loop (loop at 29-45) — it is safe because bash `shift` never touches `$0`. Ordering only matters in cmd. Worth not carrying "resolve before parsing" away as a general rule, and worth a glance at whether your new assertion's comment implies otherwise.

**Two small sites the rewritten body adds, both cheap here:**

- The error text is still wrong about the cause after the fix — the cwd is irrelevant to this resolution, so "Run this script from inside the cowork-genealogy repo" keeps sending the next person to check the one thing that cannot be it. Reword **both twins in parity** (`setup-feedback-case.bat:53` and `setup-feedback-case.sh:57`) to name the script's location instead.
- `docs/specs/feedback-case-spec.md` §3 guarantee 3 says the marker holds "the repo checkout it was run **from**", which is the phrasing that leaves the cwd-based mis-fix available. Tightening it to "the checkout the script itself lives in" closes that door.

**The one thing that cannot be done from macOS.** The bug has been present since the file landed (`5981a113`, 2026-06-04), so **nothing below line 51 of this script has ever executed** — not just the line you moved. Your fixes 2 and 3 are direct evidence of that. The issue body now carries a `cmd.exe` checklist walking all seven guarantees of spec §3 plus the `reset-feedback-case.bat` round-trip, and two cases worth singling out: running it **by absolute path from outside the repo** (the case a cwd-based fix would break, and which must still work), and copying the `.bat` alone into a non-repo directory to confirm the error path still fires. @florencemashipei has the Windows machine and the #1791 zip that reproduced this.

Nothing here asks you to widen the diff much — mainly, please don't merge on a macOS read-through.

Ikennaya1 (changes requested):
Good find, Florence. That shift behavior is genuinely hard to spot — %1 quietly becoming %0 is exactly the kind of bug you only catch by stepping through the script. It also explains a failure mode that would have looked like a corrupted zip file to anyone who encountered it. Moving the read above :parse is the right fix, and the comment you added should help prevent someone from accidentally undoing it later.

The two additional issues you fixed along the way are just as valuable. A bad zip printing "Imported to..." and then exiting as though everything succeeded is the worst kind of failure: the user has no reason to suspect anything went wrong. The same applies to the git step failing silently. Catching all three issues in one pass — and documenting why each change exists — is exactly the right approach.

There are four things I'd like changed before this goes in. The first two are the ones I'd hold on.

1. The new test can't catch the bug it was written for

The test currently looks for the first %~dp0 in the file. After your change, that's the comment you added two lines above the actual code. Because the comment always remains in the correct place, the test passes regardless of where the real line is.

I verified this by moving the code back to its broken position while leaving the comment untouched. The test still passed. It only fails if the comment is removed as well — which would happen when reverting the entire change, and that's why the check appeared to work.

I think this should fix it:

const dp0 = body.search(/^\s*set\s+"?SCRIPT_DIR=%~dp0/m);

After updating the test, try the same experiment again while leaving the comment in place. This time it should fail, which confirms it's checking the code rather than the comment.

2. Half of the test can quietly switch itself off

If either of the patterns the test searches for is missing, the test currently passes without complaint.

You already added a small test to prove one side of the condition still exists, but there isn't an equivalent check for the other half (shift). If that pattern ever stops matching, the test silently becomes ineffective and nobody notices. It's the same failure mode you already ran into once, so it's worth closing the loop here.

A second small assertion alongside the one you've already written would solve it.

3. The new error message tells Windows users to run make

This is the Windows script, and the whole reason it exists is that Windows users often don't have make available.

The repository already points users to the correct command:

scripts\reset-feedback-case.bat

(see Makefile:721 and docs\alpha-feedback-guide.md:568).

I'd recommend using that instead. It would also be worth mentioning --force, since a failed import leaves files behind and a normal rerun will refuse to proceed.

4. The unzip error discards the actual failure reason

Right now:

catch { exit 1 }

means a damaged zip, a locked file, and an overlong path all produce the same result.

It's much more useful to preserve the underlying error while still stopping the script immediately:

catch {
    Write-Error $_
    exit 1
}

That way the user sees the real cause of the failure instead of having to guess.

With those four changes in place, I'd be happy with it. The core fix looks correct; these are mostly about making sure the tests actually guard the regression and that failures remain diagnosable in the future.

Ikennaya1 (approved):
Good work, Florence! All four findings check out on `47aaa57d`, and **your fix for Finding 1 is better than the one I proposed.** I'd keep it as-is.

I reran the experiment: moved the `set` back below `:parse` while leaving the comment in place, and the guard correctly failed with `%~dp0 read after shift`. Blanking `REM`/`::` lines is a stronger approach than anchoring to `SCRIPT_DIR=`, because my version would only have caught that specific pattern. A script that read `%~f0` into a different variable name would have slipped through. Flagging *every* `%0` read after the shift, rather than just the first, is another improvement; my version could have missed a later violation hidden behind an earlier compliant read.

I also verified the counts: **46** `.bat` files, exactly **one** uses `shift`, and **43** read `%0`. Both canaries are doing their job.

Finding 3 ended up better than I originally suggested. Removing the misleading explanation and stating that the script *has no baseline to reset to* is the correct fix, not merely a filename adjustment.

Two points are worth documenting rather than changing:

* The comment blanking only handles lines that *begin* with `REM` or `::`, so an inline `& rem %~dp0` after a shift would still be flagged as a violation. That's a loud and obvious failure mode, so I'd leave it as-is.
* The "still uses shift" canary depends on a single file. That's intentional: if the argument-parsing loop ever moves elsewhere, the canary will tell you the guard has become vacuous instead of silently passing forever.

The `$LASTEXITCODE` note is a genuinely valuable catch and arguably more important than this PR itself. Anyone validating a `.bat` file from a PowerShell prompt can see `0` even when the script writes errors to stderr.

**Approving.**

packages/engine/mcp-server/tests/packaging/windows-wrappers.test.ts
```diff
@@ -212,4 +212,68 @@ describe("Windows wrapper drift", () => {
...
+  })(projectRoot);
+
+  /** Body with REM / `::` comment lines blanked, so a comment cannot match. */
+  const code = (f: string) =>
+    read(f)
+      .split("\n")
+      .map((l) => (/^\s*(?:rem\b|::)/i.test(l) ? "" : l))
+      .join("\n");
+
+  const ZERO = /%~?[dpfnx]*0\b/;
```
jud-sdev: This class only lists d/p/f/n/x, so the guard misses every other valid %0 modifier — s, a, t, z. I ran the forms through it:

```
CATCH  %~dp0   %~f0   %~nx0   %0
MISS   %~s0   %~a0   %~t0   %~z0   %~sdp0
```

`%~sdp0` is the one that matters: it's the short-8.3-path form people use to dodge spaces in the path, and it's a normal way to resolve the script dir. A future `set "SCRIPT_DIR=%~sdp0"` placed after a shift would reintroduce exactly the bug this test exists to catch, and the guard would stay green — the same read-as-coverage gap the comment-vs-code round already fixed.

Two-character fix, add s/a/t/z to the class:

You wrote:
> `  const ZERO = /%~?[dpfnx]*0\b/;`

Change it to:
> `  const ZERO = /%~?[dpfnxsatz]*0\b/;`

I verified this catches %~s0 / %~sdp0 / %~a0 / %~t0 / %~z0 and still catches every form it catches today.

jud-sdev (changes requested):
Good, careful fix — I stepped through the script line by line and the three fixes hold up. The %~dp0 move is right (I confirmed :parse shifts, so after the loop %~dp0 would be the zip's directory), the git baseline runs inside the pushed dest dir so the baseline lands in the case not the repo, and the PowerShell try/catch + exit 1 makes a bad unzip actually fail the errorlevel check. I also reintroduced the bug and watched the new guard catch it, and confirmed it strips REM lines so a comment can't stand in for the code.

One thing to fix before merge, in the guard itself. Its regex covers only the d/p/f/n/x modifiers, so it misses the s/a/t/z ones — including %~sdp0, the short-8.3-path idiom people use specifically to dodge spaces in the path, which is a normal way to resolve the script dir. A future script that does that after a shift reintroduces this exact bug and the guard stays green. Details and the two-character fix are in the inline comment. Everything else I'm happy with once that lands.

florencemashipei:
@jud-sdev — done at `978652c3`, and the branch is 0 behind `main` again.

`ZERO` is now `/%~?[dpfnxsatz]*0\b/`. Your table reproduces exactly on the two classes:

```
form      old    new
%~dp0     CATCH  CATCH
%~f0      CATCH  CATCH
%~nx0     CATCH  CATCH
%0        CATCH  CATCH
%~s0      MISS   CATCH
%~a0      MISS   CATCH
%~t0      MISS   CATCH
%~z0      MISS   CATCH
%~sdp0    MISS   CATCH
```

Proven to fail rather than assumed, per CLAUDE.md's "a new lint must be proven to fail": I injected `set "SCRIPT_DIR2=%~sdp0"` immediately after the `:done_parse` label in `setup-feedback-case.bat` — i.e. genuinely after the shift loop — and `no .bat reads %0 after it shifts` failed with

```
scripts/setup-feedback-case.bat: %~sdp0 read after shift
```

then reverted it. Worth noting the first placement I tried sat *above* the first `shift` in file order and passed, which is correct behaviour for a positional guard but a good reminder that where the injection lands decides whether the proof means anything.

Full packaging suite green after the fix and again after the `main` merge (29 files, 598 tests).

jud-sdev (approved):
The guard now covers the s/a/t/z modifiers — I checked the regex at head and it reads %~?[dpfnxsatz]*0, so %~sdp0 and %~s0 are caught along with the d/p/f/n/x forms, and the whole windows-wrappers suite still passes (19), so widening the class did not falsely flag any existing .bat. The branch scope is still just its own four files. Everything I verified before stands: the %~dp0-before-parse move, the git baseline running in the pushed dest dir, and the PowerShell try/catch making a bad unzip fail. Good fix.

PR #1909, opened 2026-08-25
record-extraction: drop the check-warnings hop the extractor cannot feed (#1863)

### LINKED ISSUE #1863: record-extraction and the orchestrator disagree on what follows extraction: check-warnings first, or straight to person-evidence
**Touches:** packages/engine/plugin/skills/record-extraction/SKILL.md, packages/engine/plugin/skills/research/SKILL.md

`record-extraction`'s closing step tells the router where to go when the last record is done:

> hand off to the next skill in the workflow (check-warnings for new tree persons, then person-evidence)

The orchestrator says the opposite about the same transition:

> The record-extractor agent deliberately cannot and does not link; its output ALWAYS flows through person-evidence next.

There is no `check-warnings` step in the routing table at all.

The evidence favours the orchestrator, and it is in `record-extraction`'s own body two paragraphs earlier: the extractor is assertion-only — "it does not write tree persons or edges; the household skeleton (member stubs + parent-child edges) is minted by person-evidence at link time via `materialize_facts`." So the "new tree persons" the handoff sends to `check-warnings` do not exist yet when it fires. The warnings pass has nothing to check. `person-evidence` already runs `check-warnings` at its own step 8, after it mints them, which is the point where the check has something to look at.

That reads as a leftover from when extraction wrote tree persons. Deleting `check-warnings` from `record-extraction`'s handoff would settle it — unless the intent was to check the persons a *previous* record's linking created, in which case say so and the handoff stays.

Editing `record-extraction/SKILL.md` arms that skill's run-log gate. Issue #1821 also edits this file to give the router `research_append` and a plan-item step; batch the two if they land together.

Found while mapping every skill and agent to its gate, reads and writes (`docs/skill-dataflow.md`).

benter-070:
## Summary

**Tier: Normal.** Closes #1863. The orchestrator was right; the `check-warnings` hop is deleted.

`record-extraction`'s closing handoff sent the caller to *"check-warnings for new tree persons, then person-evidence"*. The orchestrator says the opposite about the same transition — the extractor's output *"ALWAYS flows through person-evidence next"* — and has no `check-warnings` step at all.

## Why the orchestrator wins — three independent layers

The card argued this from `record-extraction`'s own prose. That prose could itself have been the stale half, so I checked the two layers underneath it:

| layer | evidence |
|---|---|
| **prose** | `record-extraction:193-195` — the extractor is **assertion-only**; the household skeleton is "minted by person-evidence at link time via `materialize_facts`" |
| **architecture** | `agents/record-extractor.md` frontmatter **denies** `materialize_facts`, `tree_edit` and `research_append` — 9 deny lines, 3 tools × 3 registrar spellings. Per CLAUDE.md a `disallowedTools:` deny binds **even under `bypassPermissions`**, so this is not a convention the agent can drift from |
| **tool surface** | `extraction_append` writes `sources` + `assertions` and one tree **source description**. `check-warnings` calls `person_warnings` / `person_quality`, which inspect person facts and relationships — impossible lifespans, events after death, child born after parent died |

So the extractor cannot create a tree person, and the only tree object it *can* write is a source description, which carries nothing `check-warnings` looks at. The warnings pass had nothing to inspect.

## The card's alternative reading does not hold

The card left the hop standing if the intent was to check persons a **previous** record's linking created. It isn't, on either branch:

- **Within an invocation** — the orchestrator invokes `record-extraction` *"once per batch of pending records"* (`research/SKILL.md:159`), and the router's own loop extracts every queued record before handing off. No linking happens in between, so no persons appear.
- **Across invocations** — persons minted by an earlier batch were already checked, because `person-evidence` **step 8** *"invoke `check-warnings` on the affected persons"* runs at the point of minting. Nothing is left unchecked for a later pass to catch.

## Coverage is relocated, not reduced

Worth stating explicitly, because it is the first thing a reviewer should push on. `person-evidence` step 8 is a **numbered workflow step with an imperative "invoke"**, not a suggestion — so deleting the router's hop removes a call that inspected nothing and keeps the one that inspects the persons right after they exist.

## A third site the card did not name — attempted, measured, dropped

`agents/record-extractor.md:982` gave the agent's next-step-hint example as `"check-warnings on I5/I6"` — teaching it to suggest the same dead hop, on person ids it cannot have created. Same defect, same root cause, one line. I fixed it, measured it, and **reverted it.**

`ut_record_extraction_020`, post-#1706 agent body:

| agent body | `_020` |
|---|---|
| pristine | pass, pass, pass |
| edited — hint replaced | fail |
| edited — hint deleted, nothing added | fail, fail |

Fisher's exact ~0.05 with no mechanism found, and **both** variants failed including the one that only *removes* text — so the effect does not track the semantics I first blamed. Against that, what the edit buys is nil: the caller only relays the summary, so nothing breaks if a stale example stays. By CLAUDE.md's ladder that is a nit, and the rule for a nit is to drop it. It cost three suite runs to learn that; it should have been classified as a nit before the first one.

**The agent body ships untouched.** Left for a follow-up rather than a third wording.

## Review guidance

**Start here:** `packages/engine/plugin/skills/record-extraction/SKILL.md`, the "Present and continue" paragraph. The judgement to check is whether the handoff should name `person-evidence` at all, or simply return to the orchestrator and let the routing table decide. I kept the explicit name because the orchestrator's contract is that extraction output *always* flows there, so naming it costs nothing and matches the house pattern; if you would rather the router stayed silent and deferred entirely, this is the line to say so on.

**No explanatory prose was added.** Per CLAUDE.md every line in a `SKILL.md` is a billed token on every invocation, so the rationale lives in this PR and in `docs/skill-dataflow.md`, not in the body. Net **−1 line** in the skill.

**`docs/skill-dataflow.md` finding 7** recorded this as an open contradiction — it is the doc the card was filed from. Updated to the `— issue #NNNN, ruled:` form findings 5 and 8 already use, so the register does not keep reporting a resolved disagreement.

**#1821 is not batched.** The card suggests batching if the two land together; #1821 is assigned to @DallanQ, not me, so there is nothing to batch.

## Gate cost

Only `record-extraction`'s gate arms, and only from the `SKILL.md` edit. `docs/skill-dataflow.md` and `eval/CLAUDE.md` are not in any snapshot. The agent body is untouched, so `research` — which also delegates to `record-extractor` — never came into it; it is in `RUNLOG_GATE_EXEMPT_SKILLS` and has no `eval/tests/unit/research/` dir anyway, so `exempt_suiteless` drops it.

## Test plan

- [x] `tests/packaging/` — **578 passed, 27 files.** Run directly (`npx vitest run tests/packaging/`); no `make` on this machine (#1185).
- [x] `check_skill_frontmatter.py` — clean, 27 skills + 6 agents. No `description` or `tools:` changed, so no routing pair to re-pin and no 1024-char risk.
- [x] **No grading dependency on the old behaviour.** `grep` over `eval/tests/unit/record-extraction/` finds no `check-warnings` / `check_warnings` / `person_warnings` in `rubric.md` or any `judge_context` — so the judge cannot dock a run for skipping the hop this PR deletes. That was the failure mode of the `same_person` incident CLAUDE.md cites, and it does not apply here.
- [x] `check_rubric_tool_drift.py` — the `record-extraction` warnings it emits (`materialize_facts`, `tree_edit`, `tree_correct`, `record_search`) are **pre-existing**, in `rubric.md` and test JSONs this PR does not touch. `check-warnings` is a skill name and is not in the lint's tool vocabulary, so this change can neither add nor clear a hit.
- [x] Ran `record-extraction` and committed the run log **and** its `.ann.json`: `v1_2026-08-24_17-40-15`, verified **active** (`diff_snapshot_vs_disk` -> no drift) after merging `main`. **23 pass / 3 partial / 2 fail of 28.** `check_runlogs.py` locally: All runlog rules satisfied.

**Read against the distribution, not the newest log.** The four committed baselines score **17, 19, 24 and 24** passes with no code change between them, so 23 is second-best of five runs.

| test | 4 baselines | this run | read |
|---|---|---|---|
| `_007` | partial x4 | partial | 4/4 non-pass on main — standing |
| `_017` | part, part, **fail**, part | fail | 4/4 non-pass on main; fails the validator #1706 added *from* the 08-17 baseline (`a_034`: `relationship_type: "spouse"` beside `value: "sibling of…"`) |
| `_020` `_021` `_009` `_027` | flaky set (2/4–3/4 non-pass) | **all passed** | — |
| `_005` `_006` `_028` | pass x4 | fail / partial / partial | **newly non-pass — disclosed, see below** |

The three newly non-pass tests are all **judge-graded `record-extractor` agent behaviours on a pristine agent body**: a ParentChild relation inferred from an 1850 census (`_005`), a first `extraction_append` missing `record_persona_id` and self-corrected on the next call (`_006`), and derived birth years classified indirect/secondary (`_028`, the #1706 stated-vs-inferred lane). This PR's diff is a two-line deletion in the router's "Present and continue" paragraph, which governs what happens *after* every record is extracted — it cannot reach relationship inference, an append call's persona id, or birth-year classification.

Unresolved at `runs_per_test: 1` and **disclosed rather than re-run for a green.** Re-running until green is the failure mode the `eval/CLAUDE.md` note in this PR warns about, and rule 2's compare-to-newest is what rewards it.

**The annotation carries two disagreements, both harsher than the judge.** `_028` was scored 2 -> 1 on Correctness and on Evidence type accuracy: the agent's rationale ("a person cannot provide primary information about their own birth") inverts the rubric's first recurring-error callout, that a STATED fact in a record is direct. `_011`'s two 1s were confirmed rather than overridden — the harness advisory notes a human confirmed that judgement in 20 of 24 comparable cells, and the skill routed correctly while producing no text and no tool calls.

**`judge_prompt_hash` is stale** (`0d186137` vs current `c39d7003`) — rule 2b is warn-only by design, so the corrected mean is read cautiously rather than re-run.


benter-070:
@Emruthwill — requesting you specifically because you own #1443 (`record-extraction`: marriage/death-cert classification has real run-to-run instability), and this PR's test-plan reading leans directly on it.

**The change is two lines.** `record-extraction`'s closing handoff sent the caller to "check-warnings for new tree persons, then person-evidence"; the orchestrator says the extractor's output "ALWAYS flows through person-evidence next" and has no check-warnings step. The orchestrator is right — the extractor is denied `materialize_facts` / `tree_edit` / `research_append` in its frontmatter, so the tree persons that handoff named cannot exist when it fires.

**Where your card matters.** Three tests are newly non-pass against clean 4/4 baselines (`_005` fail, `_006` partial, `_028` partial), and I have disclosed them as unresolved at `runs_per_test: 1` rather than re-running for a green. That judgement rests on the base-rate method I posted to #1443: the four committed baselines span **17, 19, 24 and 24** passes with no code change, so this run's 23 is second-best of five. All three are judge-graded `record-extractor` agent behaviours on a **pristine** agent body, and this diff only touches what happens *after* every record is extracted.

If you think that reading is too generous — that three clean-baseline tests moving warrants a re-run rather than a disclosure — that is exactly the call I would like challenged, and you are better placed than I am to make it.

**Two other things worth your eye:**

1. **The annotation carries two disagreements, both harsher than the judge.** `_028` was scored 2 -> 1 on Correctness and Evidence type accuracy: the agent marked derived birth years `evidence_type: indirect` / `information_quality: secondary` reasoning that "a person cannot provide primary information about their own birth", where the rubric's first recurring-error callout is that a STATED fact in a record is direct. That is #1706's stated-vs-inferred lane and it is a craft call, not a process one.

2. **An agent-body edit was attempted and dropped, measured.** `record-extractor.md`'s next-step-hint example still reads `"check-warnings on I5/I6"` — the same dead hop. I fixed it two different ways and both coincided with `_020` failing (0 for 3, against 3 for 3 pristine), including the variant that only *deletes* text. No mechanism found, and the edit buys nothing behavioural, so by CLAUDE.md's ladder it is a nit and the rule for a nit is to drop it. The stale example ships as-is; the numbers are in the PR body.

Note this needs a **senior** approval to merge — CODEOWNERS gates `/packages/engine/plugin/skills/` and `/eval/runlogs/`. Peer approval first is what makes the `ready-for-senior-genealogist` label fire, so this is step one, not the whole path.

johnmarkpeterbrown (changes requested):
> Nice work on this one. The three-layer argument is sound, and I checked all of it: the agent's deny list really does block `materialize_facts`, `tree_edit` and `research_append` under all three spellings, the orchestrator really does say extraction output always flows to `person-evidence`, and `person-evidence` step 8 really is a numbered imperative, not a suggestion. Coverage moves, it doesn't shrink. I also ran the numbers in your test plan — the four baselines are 19/17/24/24, `_009` and `_021` are 3-of-4 non-pass, and `_005`/`_006`/`_028` were clean across all four. Your reading of them is right.
>
> Disclosing the three newly non-pass tests instead of re-running for a green was the right call. Please keep doing that.
>
> One thing to fix before this merges.
>
> **`docs/skill-dataflow.md` claims a fix that isn't in the diff.**
>
> The ruling says the hop was deleted from two places. It was deleted from one. The agent's next-step-hint example is unchanged on this branch and still says `"check-warnings on I5/I6"`. Your PR body says so too — "The agent body ships untouched."
>
> That register is where the next person goes to ask "is this still open?" Right now it will tell them finding 7 is fully settled, and it isn't.
>
> Your reasoning for leaving the agent line alone is good — you measured it, it bought nothing, and a nit gets dropped. The doc just has to say that instead of the opposite. No follow-up issue needed; I searched and none exists, and I don't think one is warranted.
>
> ```
> You wrote:
>    so the tree persons that handoff names do not exist yet. — issue #1863, ruled: the
>    orchestrator is right; the `check-warnings` hop is deleted from the router's handoff and
>    from the agent's next-step-hint example, which named the same non-existent persons
>
> Change it to:
>    so the tree persons that handoff names do not exist yet. — issue #1863, ruled: the
>    orchestrator is right; the `check-warnings` hop is deleted from the router's handoff.
>    The agent's next-step-hint example (`agents/record-extractor.md:982`) still names it and
>    is deliberately left alone: the caller only relays the summary, so a stale example
>    changes nothing, and two edits to it each cost a suite run without an explained effect
> ```
>
> **Optional, take it or leave it.** The `eval/CLAUDE.md` figure "0 of 1,893 test entries across 122 committed unit run logs" is already 0 of 1,973 across 124 on this branch — the main merge added logs. The zero is exact and is the part that matters; only the denominators move, and they'll move again next week. Dating it stops the chase:
>
> ```
> Change it to:   (measured 2026-08-24: 0 of 1,893 test entries across 122 committed unit run logs)
> ```
>
> **On your Review guidance question** — keep the explicit `person-evidence` name. The orchestrator's contract is that extraction output always flows there, so naming it costs nothing and the two can't drift apart. I'd leave the line as you wrote it.

benter-070:
Both applied in 5b8d633d.

**The register entry.** Fixed as written. I checked the claim before changing it
rather than taking it on trust: `agents/record-extractor.md:982` does still read
`"check-warnings on I5/I6"`, and `git diff origin/main...` shows the agent file
untouched by this branch. So the register was claiming a deletion that isn't in
the diff. You're right that it would have told the next reader finding 7 was
settled — that register is exactly where someone goes to ask "is this still
open?", and it was answering wrongly.

**The dated figure.** Also applied. Worth recording: I re-counted on this branch
and got **0 flaky of 1,970 entries across 123 committed logs**, against your
1,973 across 124. The zero matches; the denominators don't, and they're three
apart after a couple of days. That's the argument for the date rather than for
new numbers — any figure I write here is stale before review ends, while the
zero is the load-bearing part and still holds.

**On `person-evidence`.** Taking your answer, leaving the line as written.

Thanks for running the baselines independently — 19/17/24/24 and the 3-of-4 on
`_009`/`_021` is what the disclosure was resting on, and it's more useful having
it confirmed by someone who wasn't looking for that answer.

johnmarkpeterbrown (approved):
Both findings addressed. The register entry now says what the diff actually did, and names the agent line it deliberately leaves alone. Thanks for re-verifying `record-extractor.md:982` yourself rather than taking my word for it, and for dating the flaky measurement in `eval/CLAUDE.md` when only the blocking one was asked for.

Re-checked on `c5769dca`:

- Full engine suite **2717 passed, 0 failed** (the Windows symlink EPERM I hit on the old head is gone — fixed by `7e5a3771` on main).
- `eval/harness` pytest **2852 passed, 3 skipped, 0 failed**.
- `check_runlogs.py` — all runlog rules satisfied. `check_skill_frontmatter.py` clean.
- Run-log snapshot vs. disk: **zero drift across all 53 files**.

That last one was the only real risk in a 42-commit catch-up. A main merge touching the rubric, a test JSON, or the judge prompt would have staled the run log and forced a re-run at your expense. Nothing did.

The two gate warnings are both fine to merge on: the stale `judge_prompt_hash` is warn-only by design, and `ut_record_extraction_017` producing no graded dimensions is disclosed and non-pass on 4 of 4 baselines.

One non-blocking note for the record: the PR body still says the agent hint was "left for a follow-up", which the corrected register entry now contradicts — it says the line is deliberately left alone, and I agree with that call. Body text only, nothing to change in the diff.

Approving. Still needs a second approval to merge — @Emruthwill, you were asked for by name because you own issue #1443, and this PR's reading of its run against the baseline distribution is exactly that territory.

EdmondOware (approved):
Approving.

On your direct question, keep naming `person-evidence`. The orchestrator's contract is unconditional, "its output ALWAYS flows through person-evidence next", so there is exactly one destination and naming it cannot diverge from the routing table. Staying silent and deferring would be the better call if that contract were conditional. It is not.

The deletion is safe on the genealogical half, and I checked it from the tree side rather than the prose. The extractor cannot mint a tree person, so no person ever exists for that hop to inspect, and `person-evidence` step 8 invokes check-warnings after creating links and any stub persons, which is the moment persons first exist. Coverage moves to where the objects are.

One informational note. `docs/skill-dataflow.md:235` still opens in the present tense, "record-extraction hands off to check-warnings and then person-evidence", which is exactly what this PR stops being true. The appended ruling corrects it, but a reader reaching that line first sees a live discrepancy that no longer exists. Worth rewriting the sentence rather than only appending to it.

PR #1911, opened 2026-08-25
search-images: arm ut_search_images_009's no-browse invariant and prove it can fail (#1788)

francis-2008-happy:
## Summary

**Tier: Normal.** Repairs one of the five tests named in #1788. Issue #1788 stays open for the other four.

`ut_search_images_009` is a negative test: a planning request that must route to `research-plan` and must **not** browse. It is graded on the invariant `test_no_browse_or_writes_on_planning_request`, whose first assertion reads `tool_calls` for `volume_search`/`image_search`. The test declared **no `mcp_fixtures`**, and `mock_mcp.py:369` registers a browse tool only when a fixture names it — neither is in `LIVE_TOOLS`. So those tools were absent from the model's tool list entirely, and the assertion forbade a call that was **physically impossible to make**: green forever, with `grade_on_invariant` riding on it. Same failure mode as convert-dates (#1654).

This adds a catch-all `image-search-any` fixture and declares both browse fixtures on the test, which arms the assertion. It also adds the offline proof that the guard fires, which the repo requires and this branch was missing.

**Correcting my own earlier framing, in case it reached anyone:** the `no-browse-no-write` tag was already on `main`, and `runnability.py` already aborts a `grade_on_invariant` test whose tags gate no invariant. The validator was **never skipping**. The defect was one layer down, in tool registration.

## Review guidance

**Start here:** `eval/harness/tests/unit/test_search_images_validators.py`, and specifically `test_browse_tools_are_registered_by_the_declared_fixtures`. That test reads `ut_search_images_009`'s **own** `mcp_fixtures` array rather than a copy, because the regression this PR fixes is that array going missing — a hardcoded list would keep passing while the corpus went back to asserting an impossible call. My first draft did exactly that and I caught it by mutation; please check the wiring is genuinely to the live declaration.

**Unsure about:** Nothing outstanding. The `judge_context` question below was answered in review: adding the note to `test.description` is free (`description` is stripped as cosmetic), and only *deleting* the `judge_context` copy costs a paid run. The free half is applied; the copy stays until this skill next earns a run for another reason.

## Plan

**Didn't change:** The other four tests in #1788. The ruling on that issue is one PR per skill for this repair, not one PR for five. Also left alone: `runnability.py:246`'s coarse tag gate (it accepts *any* tag gating *any* validator in the skill's file, so swapping in an unrelated gate tag keeps a test runnable while the intended invariant skips) and the fact that the run log never records which MCP tools were offered. Both are pre-existing harness weaknesses, both are `nothing-checks` shaped, and neither is this test's to fix — see Follow-on issues.

**Acceptance check:** `cd eval/harness && uv run pytest tests/unit/test_search_images_validators.py` — 6 pass. Each half fails when its own target regresses, verified one mutation at a time:

- Emptying `mcp_fixtures` on the corpus test → `test_browse_tools_are_registered_by_the_declared_fixtures` fails. Dropping only `image-search-any` also fails it.
- Neutering assertion 1 in the validator (`browsed = []`) → 4 of 6 fail, one per server spelling.

**Read this before trusting the paid run:** the committed run log is **not** evidence for this fix. On the recorded run `009` routed to `research-plan` immediately, the routing short-circuit (`orchestrator.py:307-320`) halted the run, and the entry reads `num_turns: 0`, `tool_calls: []`, zero tokens. `009` passed identically **before** the fix — compare `v1_2026-08-10_13-17-12.json` on `main`, same `mcp_fixtures: []`, same empty `tool_calls`. The run log's role here is rule-2 snapshot compliance, nothing more. The falsifiability evidence is the test above plus the `mock_mcp.py:369` / `LIVE_TOOLS` mechanism.

Where the defect actually bit: run `v1_2026-07-28_22-33-47` had `search-images` active for **9 turns** and recorded `tool_calls: []`. That is the run where a forbidden browse would have appeared, and could not.

**Deviated from the plan:** Nothing on the fixture. I added the falsifiability test mid-review after it was pointed out that the branch shipped a corpus change and a run log without ever proving the guard fires.

## Test plan

- [x] `make test-all` — ran, **not fully green, and not because of this diff.** 2730 pass, 9 fail. All 9 are `eval/harness/tests/unit/test_setup_feedback_case.py` failing on `scripts/setup-feedback-case.sh: line 83: unzip: command not found`. `unzip` is not installed on this machine; I reproduced the identical 9 failures on an unrelated branch. Everything else green: viewer-ui 225, electron 97, web 10, control-plane pytest 209, CRUD-UI 168, engine suite, all typechecks.
- [x] `make eval-skill SKILL=search-images` — run log `v1_2026-08-25_15-26-32.json` and its `.ann.json` committed. 12/12 pass, $1.8567.
- [x] Routing-pair box — N/A, no skill `description` or DO-NOT clause changed.

`check_runlogs.py` passes with zero warnings against the merged base, including rule 2b (`judge_prompt_hash` recomputes to `c39d7003…`). Mutation-checked that gate too: dropping a sampled test from the annotation produces the incomplete-annotation error.

**Two claims in commit `e5c75847`'s message are overstated; I am correcting them here rather than rewriting pushed history.** "At concurrency 3" is not recorded anywhere in a schema-3 run log, so treat it as an unverifiable invocation detail. The description of the prior compromised run is also unverifiable — that artifact was a local run I discarded, never committed.

**Annotation:** 60 corrections, all five `review_sample` tests at 6/6 dimensions, zero disagreements with the judge. `009` itself is **not** in the sample, so rule 3 exempts it and it carries no written comment. The human read, for the record: it passed on the ideal route with the validator recording `"error": null` — i.e. the invariant ran rather than skipping, which is the distinction that matters here and is visible in the artifact.

## Follow-on issues

**Folded in / filed instead:** Folded in — the falsifiability test (the substance of this PR beyond the one-line fixture declaration), and the `test.description` note from review.

Filed: **#1930** — a gate-tag swap silently disarms a `grade_on_invariant` test. `runnability.py:246` accepts any tag gating any validator in the skill's file, and `snapshot.py:47` strips `tags` as cosmetic, so the run log stays active. I reproduced both halves on `ut_research_plan_003`: `runnable=True` with both real gate tags swapped out, and `check_runlogs.py` still reporting "All runlog rules satisfied". Structurally possible in nine skills. Filed rather than folded because it is a harness-wide gate weakness across two modules that this PR touches neither of, and it needs a decision on how strictly runnability should bind tags to the invariants a test depends on. Labelled `nothing-checks`.

Not filed, on review advice: a `registered_mcp_tools` field on the run log. It would need a schema change plus a re-run for every skill, and the offline test here now asserts that property for free.


chesworthrm (approved):
Approving. The diagnosis is right and the falsifiability proof does what it claims — I checked each link independently, and all three of your mutation claims reproduce exactly, including 4-of-6 on the validator neuter. I reviewed this merged with `main` rather than on its own head; the harness suite is green there (2818 passed, 2 skipped).

Nothing blocking. Two things below: one free change, and an answer to the question you escalated.

## Your "Unsure about" — no second run, and no label

You need neither. I tested both halves as real commits through `check_runlogs.py`:

- note added to `test.description` → **All runlog rules satisfied**
- note also removed from `judge_context` → fails, `negative-research-plan.json: content-differs`

`description` is stripped as cosmetic, so adding it there is free. Only *deleting* the `judge_context` copy costs a run. So take the free half now:

You wrote:

```
"description": "User asks which unindexed volumes to browse next and to help plan — that's research-plan. search-images executes a browse of a specific volume; it does not decide research strategy."
```

Change it to:

```
"description": "User asks which unindexed volumes to browse next and to help plan — that's research-plan. search-images executes a browse of a specific volume; it does not decide research strategy. The volume-search-any / image-search-any mcp_fixtures are declared deliberately and must NOT be removed as unused (#1788)."
```

That satisfies spec §5.5 without spending a slot. Drop the `judge_context` copy whenever this skill next earns a paid run for another reason — until then it is billed into the judge prompt on every graded run.

## Follow-ons: file two of the three

File #2 and #3 as **one** `nothing-checks` issue. They are one defect with two halves, and it is a demonstrated silent failure rather than a suspicion — so put the repro in the body:

`ut_research_plan_003` (`handle-negative-probate-result.json`, scenario `flynn-resolved`) carries two gate tags. Replace both with `pli-006-status-unchanged`, a "nothing changed" validator that a no-write test satisfies trivially:

- `check_runnable` → still `True`
- `test_no_plan_writes_when_resolved` and `test_research_plan_no_new_plan` → silently skipped
- `test_pli_006_status_unchanged` → passes, asserting nothing the test guards
- snapshot hash unchanged (`tags` is cosmetic), so the run log stays active and `check_runlogs` stays green

Nine skills have more than one gate tag plus a `grade_on_invariant` test, so the swap is structurally possible in nine places.

`search-images` happens to be safe, and it is worth knowing why: its two gate tags are mutually exclusive on state. `negative-result-log` demands a log entry that `009` cannot produce, so a swap there fails loudly instead of passing vacuously. That is why hardcoding `_TAGS` in your new test file is fine as written — I went looking for a hole there and there isn't one.

**Don't file #1.** A `registered_mcp_tools` field means a schema change plus a re-run for every skill, and your offline test now asserts that property for free.

## Small confirmations

- The deleted run log leaves no dangling references. Your pre-empted grep is right: those two `docs/record-extraction-judge-audit.md` citations are a record-extraction log already absent from `main`.
- Both catch-alls key on a parameter that production marks required (`standardPlace`, `imageGroupNumber`), so they match any valid call, and `call_log.append` is unconditional — registration really is the arming property your test asserts.
- Thanks for flagging the two overstated claims in `e5c75847`'s message rather than rewriting history. Correcting them in the PR body was the right call.

This still needs one more approval from anyone with write access — mine covers the code-owner requirement, not the count. It is also 2 commits behind `main`, so it needs an update before the merge box clears.

johnmarkpeterbrown (approved):
Approving. This is the second approval — it clears the count, not the
code-owner rule, which Robert's approval already satisfied.

I re-derived the mechanism from the source rather than from your write-up,
and it holds. `volume_search` and `image_search` are absent from
`LIVE_TOOLS`, and `call_log.append` runs *before* predicate matching — so a
browse is recorded even when no fixture matches it. Registration really is
the arming property, and matching is only graceful degradation. That is the
part I most expected to be wrong, and it isn't.

All four mutations reproduce on my machine:

- `mcp_fixtures: []` → `declared [] -> registered [...]`, both browse tools
  missing
- dropping only `image-search-any` → fails on `image_search`
- dropping only `volume-search-any` → fails on `volume_search`
- `browsed = []` in the validator → 4 of 6 fail, one per server spelling

`check_runlogs.py` is clean with `BASE_SHA`/`HEAD_SHA` set. The annotation
checks out: 60 corrections, all five sampled tests at 6/6, every cell 3→3,
so the confirmed-pass exemption is doing the work and no comment is owed.

I also went after the hole Robert probed, because a hardcoded `_TAGS` is
exactly the kind of thing that looks fine and isn't. Feeding `009`'s
recorded empty state to `test_negative_result_log_shape` under the swapped
tag raises `expected at least one new log entry`. It fails loudly on this
skill. Confirmed, not assumed — and #1930 is the right home for the general
case.

Two of your disclosures deserve saying out loud, because both cost you
something and both were the right call. Leading with "this run log is not
evidence for the fix" is the harder version of the truth, and it is correct:
`009` recorded 0 turns and 0 tokens, and it passed identically on `main`
before the fix. And correcting the two overstated claims in `e5c75847`'s
message in the PR body, instead of rewriting pushed history, is what I want
to see from everyone on this team.

I read the current head (`2a3a8650`). One line landed after Robert's
approval — the `test.description` note he asked for — so nothing in his
review is unread.

## One loose end, not a blocker

The `judge_context` copy of the fixture note now duplicates
`test.description`, and it is billed into the judge prompt on every graded
run of `009`. Parking it until this skill next earns a paid run is the right
economics. But "next time" is currently written down nowhere, and this skill
may not earn a run for months.

Add it to #1788's body, which stays open for the other four tests anyway:

    When search-images next earns a paid run for another reason, delete the
    duplicated mcp_fixtures note from ut_search_images_009's judge_context
    (#1911). The same note lives on test.description, which is stripped as
    cosmetic and costs nothing.

Issue #1492, opened 2026-08-09
research/SKILL.md: reconcile three contradictions (address_first tables, who writes project.status=completed, mentor write surface)


DallanQ:
> **Analysis 2026-08-09 — Ruling 1 is confirmed on stronger grounds than the tiebreaker used. Ruling 2 has a blocker that makes `completed` unreachable as ruled.**
>
> **Ruling 1 — the git record says Table A was never a competing position.** Commit `1bc84e58` (2026-07-07) *deleted* the three-column table and its reinforcement paragraph and added the two-column table **in the same hunk**. Commit `54eb6386` (2026-07-08) hardened Table A's row — from a sibling branch cut before that deletion landed. Both share parent `909af548`, and both tables first coexist at merge `c8cc62f7`. So this is restoring an intentional deletion that a merge undid, not choosing between two designs.
>
> Behavioural corroboration: across every committed run there are **72 `address_first` verdicts carrying a `suggested_skill`, and the router invoked it in 2** — both `research-plan`, an ordinary loop step, so plausibly coincidence. The agent body is also a second independent witness alongside the spec, carrying Table B's wording verbatim.
>
> Two honest caveats for the record: all 72 observations are `--autonomous` runs, so Table A's *interactive* column was never testable; and the harness `stop_hook` vetoes any turn-end before `project.status == "completed"`, so the corpus cannot fully distinguish "obeyed Table B" from "the harness forbade Table A."
>
> **Ruling 2 — as ruled, nothing in the flow can satisfy `proof-conclusion`'s precondition.** Its step 8 fires "if ALL questions are now `resolved`", but its own step 7 forbids it writing the `questions` section — marking a question resolved is `question-selection`'s job. So on the final question, step 8's condition is **false by construction**. `question-selection` then resolves it and explicitly disclaims the write.
>
> The run logs match: **42 `project.status = "completed"` writes across 39 runs**, the final tool call in 37 of 42. In **25 of 42** the most recent preceding `Skill` was `question-selection`; only **6** followed `proof-conclusion`.
>
> **So the ownership ruling needs a sequencing decision beside it.** Three options: the router re-invokes `proof-conclusion` after the last `question-selection` purely to write a status field; `question-selection` becomes the owner (a third candidate, also absent from the harness ownership table); or the write moves into `research_append` as a computed gate — which this issue puts out of scope.
>
[...]

Issue #1490, opened 2026-08-09
research_append: move the two completion gates into the tool (mentor gate, then tree-encoding)


DallanQ:
> **Reviewed 2026-08-20 before junior handoff; re-checked 2026-08-24 against `main`
> at `c44350dc`.** Two of the three phases shipped, the third was measured and
> redesigned, and four things have moved under this card since the review — read
> "What changed" before planning. The ask below replaces the original, which
> follows under "Original issue".

**Touches:** packages/engine/mcp-server/src/tools/project-create.ts, packages/engine/mcp-server/src/tools/ (new tree-diff tool), packages/engine/mcp-server/src/tool-schemas.ts, packages/engine/mcp-server/src/index.ts, packages/engine/plugin/hooks/guard_project_files.py, apps/server/app/agent/real_agent.py, eval/harness/e2e/orchestrator.py, eval/harness/e2e/project.py, eval/harness/e2e/guardrail_shadow_report.py, eval/harness/harness/skill_invocation.py, docs/specs/, docs/adrs/ADR-0011-put-guardrails-at-the-write-boundary.md, docs/adrs/ADR-0003-anchor-cross-turn-rules-structurally.md, docs/architecture.md

## Phases 1 and 3 already shipped — do not rebuild them

PR #1685 merged 2026-08-17 and landed the mentor gate, its spec row, and all three
acceptance tests.

| This issue asked for | State on `main` |
|---|---|
| Phase 1 — mentor gate at the write boundary | **landed**, in the `section === "project" && op.fields.status === "completed"` branch of `research-append.ts` |
| Phase 1 step 1 — pre-call snapshot so a batch cannot satisfy its own precondition | **landed** (`preCallCritiquedSummaryIds`); the conflict half got the same treatment |
| Phase 3 — spec row | **landed** in `docs/specs/research-append-tool-spec.md` |
| All three acceptance tests | **landed** in `tests/tools/research-append.test.ts`, under a header citing this issue |
| Phase 2 — tree-encoding gate | **not landed** — redesigned below |

The mentor gate has been **hardened twice since**, so do not re-derive it either. It
now also catches the date-only spelling of resolved (`Boolean(q.resolved)`, not just
`status === "resolved"`), and the vacuous-pass case acceptance test 3 pinned is no
longer reachable through the tool at all: `questionResolvedInvariants` refuses to mark
a question resolved unless a proof summary references it.

The original body's central verification — "`grep -n "proof-critique"
[...]

PR #1914, opened 2026-08-25
dataflow: record who closes a project, and the gap the ruling leaves

DallanQ:
## Summary

- **Trivial** (docs only — two `.md` files, no code).
- Records the two rulings made on issue #1335 on 2026-08-25: the advisory verdict table is the one that stands, and the `proof-conclusion` agent owns the `project.status = "completed"` write. The dataflow doc had the orchestrator making that write in four places.
- Adds a section on how `question-selection`, the orchestrator and the `proof-conclusion` agent divide the last step, and records the gap the ruling leaves — the agent fires on a condition that is briefly true partway through every multi-question objective.

## Review guidance

**Start here:** the new "How the project gets closed" section in `docs/skill-dataflow.md`. The claim to check is the last two paragraphs: `agents/proof-conclusion.md` step 8 writes `completed` when "ALL questions are now `resolved`", and since PR #1819 that same agent resolves the question it concluded in the same batch as the summary — so its condition is true the moment it finishes the first question of an objective that will need a second one, before `question-selection` has minted the next. The agent holds no view of `project.objective` (grep it: the only match in that file is "Present evidence objectively", unrelated), so it cannot tell the two cases apart. If that reasoning is wrong, this section is wrong and the ruling needs revisiting, not the doc.

**Unsure about:** nothing in the diff. One thing outside it, flagged rather than fixed: nothing checks the paths and symbols `docs/skill-dataflow.md` cites. The lint that resolves cited paths and command names runs against exactly one file — `docs/task-lifecycle.md` — and this doc is not on that list, so I verified every claim by hand against `main` instead. (Corrected from an earlier version of this paragraph, which said the doc was covered by no lint at all. `doc-links.test.ts` also carries two repo-wide docs lints that do reach it — the ban on citing source line numbers and the ban on testing shipped-ness with commit ancestry — both walking every `.md` under `docs/`.) Widening the path-resolving lint's file list is real work and belongs to whoever owns the doc-lint scope, not to a prose PR.

## Test plan

Worth knowing for anyone else on a recent Node: `make test-all` fails on this machine with five `localStorage.getItem is not a function` errors in `viewer-ui`'s `App.test.tsx`, and it is the instrument, not the repo. Node 24+ ships a built-in `localStorage` global that shadows jsdom's and is inert without `--localstorage-file`; `Sidebar.tsx` reads it during render. This machine runs Node 25, `.nvmrc` pins 22, and CI runs 22 — so CI is green and `main` is not broken. `NODE_OPTIONS="--no-experimental-webstorage" make test-all` passes clean, which is how the box above was ticked. Nothing filed: it does not reproduce on the pinned Node.

No skill run-log snapshot is touched (no file under any skill dir, agent, or `eval/`), and no skill `description` frontmatter changed, so neither of the other two boxes applies.

## Follow-on issues

**Folded in / filed instead:** Folded in — repointing `docs/adrs/ADR-0011-put-guardrails-at-the-write-boundary.md` off issue #1492, which was merged into issue #1335 and closed on 2026-08-25. It cited the closed issue twice, once in `Related:` and once in the passage about the same reasoning being re-derived four times in one week; that second cite now carries the ruling and hands the computed-gate option to issue #1490. My own change made those cites stale, so they are mine to fix. Nothing filed.


T-FEH (changes requested):
Reviewed on the merged head (`dbdbe04e` — your `829fba04` plus a clean merge of main). Working tree clean.

## The reasoning you asked to have checked holds, end to end

I verified each link independently rather than reading the section back:

- **Step 8's condition** — `agents/proof-conclusion.md:294` is `### 8. Update project status`, and `:298` is `If ALL questions are now resolved, call research_append(… status: "completed" …)`. Exact.
- **No view of the objective** — `grep -ni objective` over the agent returns exactly one line, `:511 Present evidence objectively without bias or preconception`. Unrelated, as you said.
- **Same-batch resolve since #1819** — confirmed, and the stronger evidence is what it replaced: before that PR the skill's §7 was titled "Do not modify the question" and said marking it `resolved` was `question-selection`'s job. So step 8's condition could not previously be true of the question just concluded. The gap is genuinely introduced by #1819 rather than pre-existing.

So the ruling does not need revisiting on these grounds, and the section's core argument is sound. The ADR-0011 repoint is sound too — #1492 is genuinely CLOSED (2026-08-25), #1335 and #1490 are both open, so both citations now point somewhere live.

## [Blocking] Two rows state the ruling's target as current behaviour, and row 1 shows you know it isn't

Row 1 is honest about the state of things:

> **Nothing.** The file still instructs it to write `project.status = "completed"`; that was ruled to the `proof-conclusion` agent on 2026-08-25 **and the prose has not caught up**

Rows 2 and 13 drop that caveat and assert the target as fact:

```
row 2   "…it returns a signal and the orchestrator re-invokes `proof-conclusion` to make the write"
row 13  "The orchestrator routes here; it does not write"
```

Both are contradicted by the files this doc's own preamble names authoritative — *"where it disagrees with either, they win"*:

```
research/SKILL.md:156           "Once both are verified: write `project.status = "completed"` via `research_append`, then stop."
question-selection/SKILL.md:104 "Return the … signal so `/research` writes `project.status = "completed"` and stops."
```

Nothing anywhere instructs the orchestrator to re-invoke `proof-conclusion` at the stop point. The concrete harm is on the implementer of #1335: row 2 reads as already-true, so the §1b edit the ruling actually requires looks done. Row 1 gets this exactly right; rows 2 and 13 need the same treatment.

```
row 2, you wrote:  …it returns a signal and the orchestrator re-invokes `proof-conclusion` to make the write
change it to:      …it returns a signal whose closing line still names `/research` as the writer; the
                   2026-08-25 ruling redirects it to a `proof-conclusion` re-invocation, and that edit
                   has not landed — see [How the project gets closed](#how-the-project-gets-closed)
```

```
row 13, you wrote: The orchestrator routes here; it does not write
change it to:      The orchestrator is to route here and not write; `research/SKILL.md` still tells it
                   to make the write itself
```

## Minor — the gap section omits a qualifier row 13 already carries

`research-append.ts:1216` refuses the write with `… have no gps-mentor verdict` when a resolved question's summary has no `proof-critique` evaluation. At the instant the agent finishes the first question, the mentor has not run, so the premature write is refused on that pass — it can only land on a re-invocation after the critique exists. Row 13 says this; the gap section, read on its own, reads as though the write goes through.

```
You wrote:    true on the way to every multi-question objective.
Change it to: true on the way to every multi-question objective. On the first pass the tool still
              refuses — no `proof-critique` verdict exists yet for the summary just written — so the
              premature write lands only when the agent is re-invoked after the mentor gate has run.
```

Worth knowing while you are there: nothing else guards it. No validator rule, no hook rule (`project` is absent from `OWNED_SECTIONS`), and `eval/harness/e2e/stop_checker.py` treats `status == "completed"` as terminal — so a premature write silently truncates an e2e run rather than surfacing.

## Thank you for the Node note — it settles an open question on #1990

Your test-plan paragraph diagnosing the `localStorage` failures as Node 24+ webstorage shadowing jsdom, not a repo problem, is the answer I had asked for on #1990, where the same failure is being fixed with a permanent global stub in `vitest.setup.ts` justified as a jsdom gap. I confirmed `.nvmrc` pins 22 and every CI workflow runs 22, and on 22.23.1 the viewer-ui suite is 273/273 both with and without that stub. I have commented there pointing at this PR; your account is the one that survives checking, and its conclusion — "the instrument, not the repo" — argues the stub should not land.

One correction to the body, not the diff: "`docs/skill-dataflow.md` is not covered by any lint" is not quite right. `LINTED_DOCS` is indeed just `docs/task-lifecycle.md`, so nothing checks this doc's paths and symbols — the substantive claim stands — but `doc-links.test.ts` also carries two repo-wide docs lints that do cover it (the source-line-citation ban and the `merge-base --is-ancestor` ban, both iterating every `.md` under `docs/`).

Requesting changes on the two rows. Everything else here is careful work, and every citation in the new text resolves.

DallanQ:
Fixed in `7c86d428b`. All three findings applied — your wording verbatim on both rows and the qualifier.

**Rows 2 and 13** — applied as written. Row 1 was the honest one and the other two read as already-done; the harm you named, the edit looking landed to whoever picks up issue #1335, is the real cost.

**Gap-section qualifier** — applied as written. Worth adding that the tool's message goes further than refusing on that pass: it ends *"A verdict appended in the same batch as this update does not count; complete it in a later call."* So the refusal also survives an agent that tries to batch past it.

**Two more in the same section, same defect class as your rows.** Your frame is what found them:

1. The "how the project gets closed" table itself read as current behaviour. Now labelled as the split **as ruled**, with the paragraph after it carrying the disagreement.
2. The orchestrator's "does not hold the write" cell justified itself with *"no `research_append` grant in its own `allowed-tools`"*. Wrong evidence: `allowed-tools` is a grant, not a restriction, and both production paths grant every tool — so it denies nothing, and `research/SKILL.md` instructs the write regardless. The suite says so while running: *"skill called MCP tools not in allowed-tools frontmatter (advisory — session grants all tools)"*. The cell now says what is actually true.

**Body corrected** on the lint claim — you were right, and it now names the two repo-wide lints that do reach the file.

`make test-all` green: 2874 passed, 2 skipped; packaging 603/603.

One thing I looked at and did not change: the ownership manifest's `project` row names `skill:proof-conclusion` as caller, where the ruling means the agent — and the manifest does carry an `agent:` prefix elsewhere. It matches the ruling in substance (proof-conclusion, not the orchestrator), so the doc's claim stands; the granularity is a manifest edit for whoever implements issue #1335.

PR #1916, opened 2026-08-25
Add the tree-side citation-nulling shadow arm (#1358)

### LINKED ISSUE #1358: Graduate the citation-nulling shadow check to a hard 4th compliance failure

aghadiayeamayanvboernest:
Closes #1358.

**Detect only — this card graduates nothing.** Per the 2026-08-21 ruling: land the tree-side arm, correct the record, and let a follow-on card own the fix once the arm has been in the corpus for a few runs.

## Review guidance

**Start here:** `find_citation_nulling_in_tree_sources` in `eval/harness/harness/skill_invocation.py` — specifically clause 2 of the gate, the `primary`-fact / relationship test. That clause is the entire difference between a readable rate and a meaningless one: without it the arm fires on every evidence fact `person-evidence` materialized at link time, which the doctrine says is citation-less **by design** until a conclusion promotes it. Check it against the sentence it is quoted from (`packages/engine/plugin/agents/proof-conclusion.md` step 3) rather than against the code.

**Unsure about:** the issue's own acceptance check, which I did not meet and believe is wrong. It says *"`make e2e-guardrail-shadow SINCE=all` — the new bucket prints a non-zero count; if it prints zero, the gate is wrong, not the corpus."* Mine prints **zero** there. That criterion looks unsatisfiable to me: the plain scan reads violations *stored* in committed run logs, and no committed run can carry a `kind` that did not exist when it ran, so stored is necessarily 0 until the arm has run live. `REPLAY=1` is the only way to see it today, and that is where the 111 comes from. I am confident enough in that reading to ship — but it is my judgment overriding a written acceptance criterion, so it is the thing I would most like a second opinion on. If it is wrong, the gate is wrong and the 111 is not what I think it is.

## The two-sided reading, which is the whole finding

`make e2e-guardrail-shadow SINCE=all REPLAY=1` now prints both arms together:

citation-nulling          0 concluded source(s) with a null/empty citation string, across 0 run(s), of 159 scanned.
tree citation-nulling   111 uploaded tree source(s) with a null/empty citation string, across 50 run(s), of 159 scanned.

**The research-side arm's zero is real, and it does not graduate.** A replay over every committed `final-research.json` reaches 1,884 concluded sources, all cited. Per `guardrail-enforcement-spec.md`, *"zero is not 'low enough', it is nobody has seen this detector fire."* It stays as a **regression pin** — the zero is a genuine invariant, and this is what catches it breaking.

**The class it was built for lives at the upload copy.** 111 of 171 referenced tree sources carry no citation, across 50 of the same 159 runs. That corroborates the issue's 109/167 from four days earlier, on a corpus that has grown since — I re-derived rather than quoting it.

A zero beside a large number over one corpus is what answers *"nothing distinguishes the two readings of a zero"* for this detector, which is why the buckets are separate and printed adjacently.

## One stale pointer in the issue, worth flagging

The issue says to quote `proof-conclusion/SKILL.md` step 3. **That sentence isn't there** — that file now has three sections (Identify / Delegate / Relay), because the skill became a skill-agent pair and the body moved.

The obligation is at **`packages/engine/plugin/agents/proof-conclusion.md:274`**, and I quoted it from there:

> copy the finalized `research.json` `sources[].citation` string into the **`citation`** field … **Upload is conclusion-gated:** the working tree carries *all* sourced evidence facts (materialized at link time by person-evidence), but **only `primary`/proof-backed facts upload to FamilySearch** — un-concluded evidence stays out.

## Which gate clause comes from which sentence

The issue asks for this explicitly, since no CI job can check that a gate matches the doctrine it encodes:

| Gate clause | Source |
|---|---|
| Fires only when a `proof_summaries` entry exists | *"Upload is conclusion-gated"* |
| Fires only for sources referenced by a `primary` fact or a relationship | *"only `primary`/proof-backed facts upload … un-concluded evidence stays out"* |

**Clause 2 is load-bearing, not caution.** The working tree carries every sourced evidence fact `person-evidence` materialized at link time, and those are citation-less *by design*. Without the clause this fires on honest mid-research state and the rate is unreadable — which is the negative control the issue asked for, and is a test.

## Proven to fail

| Mutation | Red |
|---|---|
| Drop the conclusion gate (clause 1) | ✅ |
| Drop the primary/upload gate (clause 2) | ✅ 2 tests |
| Read only `sources[].ref`, not the bare `source_ids[]` | ✅ |

That last one matters on a detector whose output is a rate: both spellings occur in the corpus, and reading one silently under-counts, which looks like the problem being smaller than it is.

**My own report test caught one on the way in:** the new kind must exclude itself from the catch-all provenance scan, or it is counted twice — in its own bucket *and* the generic one — inflating exactly the rate a graduation decision would read.

## Deliberately not done

**Not folded into `check_guardrail_compliance`.** That returns `list[str]` while these detectors return `list[dict]`; nothing type-checks the difference (ruff is pinned to `F821`), so a dict appended straight in serializes fine and every test passes. A hard arm would also retroactively re-score history through `recompute_tally`. The issue names four sites a future graduation needs — that is that card's job, not this one's.

**Neither arm graduates.** `simplified-gedcomx-spec.md` makes the tree citation upload-populated by design, so part of the 111 is legitimately-not-yet-uploaded evidence that only a genealogist can price. Detect before teaching: a skill edit costs a paid eval run, and there is no baseline to measure one against until this arm has been in the corpus live.

## Note on the stored columns

The two stored buckets print `0`, correctly — no committed run carries the new `kind` yet, because the arm has never run live. The 111 comes from the replay path, which is how the number is visible today rather than after the next wave of runs.

## Verification

`make test-all` — all checks passed. No live run, no API, no spend.

The doc-links ratchet is unchanged on both specs (16 and 10). I initially wrote `(issue #1358)` into `guardrail-enforcement-spec.md`, which pushed it to 17 and would have failed CI — replaced with the mechanism, per the repo's state-the-fact rule.


chesworthrm (approved):
Approving. This is careful work — the gate is quoted from the sentence it encodes, both clauses have a test that fails without them, and the numbers in the body all reproduce. I re-ran the replay against the corpus and got your figures exactly: 111 sources across 50 runs, of 159 scanned, and 111 of 171 referenced. All four mutations you describe go red, including the self-exclusion one.

Two things I checked hardest, both because they would have made the 111 unreadable, and both fine:

- **Nothing in the count is fixture state.** I checked whether any flagged source was already present and already citation-less and already referenced in the fixture's seed tree — 0 of 111, matching by id and again by title. There are 74 citation-less sources sitting in the seed trees, so that check had something to find.
- **A source reachable only through a `preferred` name.** I thought clause 2 might be missing these, since the agent body calls the conclusion `the primary/preferred marking on the concluded value`. There are 4 such sources in 2 runs, and I opened all 4: every one is also referenced by facts that are all non-primary. They are the un-concluded evidence class clause 2 exists to exclude, so leaving them out is right. Recording it here so the next person doesn't re-derive it.

One thing to fix before merge.

**`docs/specs/e2e-test-spec.md:1352` — two different checks are now both called "the sixth".** Line 1393 already opens `**A sixth check runs in shadow mode only: the warnings guardrail was never consulted before a parentage write.**`

You wrote:

> **The failure class it was built for lives on the other side of the seam, and a
> sixth check measures it there.**

Change it to:

> **The failure class it was built for lives on the other side of the seam, and a
> companion check measures it there.**

Dropping the ordinal rather than renumbering, because the new arm sits between the fourth and the fifth in the file — any number you give it reads out of order.

One process note: the body uses none of the template's named fields. The substance of *Didn't change* and an acceptance check is there under your own headings, but **Unsure about** is missing, and that is the field a reviewer leans on hardest. Worth filling in even when the answer is "nothing".

The branch is 2 commits behind main. I test-merged and ran it there: harness pytest 2825 passed, engine vitest 2702 passed, clean merge.

docs/specs/e2e-test-spec.md
```diff
@@ -1347,9 +1347,32 @@ time, so a run that legitimately stops before upload has empty citations by
...
+graduate.** Across 159 committed runs the research-side arm flags nothing, and a
+replay over every committed `final-research.json` reaches 1,884 concluded
+sources with a citation on all of them. Zero is not a licence — per
+`guardrail-enforcement-spec.md`, "zero is not 'low enough', it is nobody has seen
+this detector fire." The arm stays in shadow as a **regression pin**: the zero
+is a real invariant — a conclusion's citations are populated whenever a
+conclusion exists — and this is what would catch it breaking.
+
+**The failure class it was built for lives on the other side of the seam, and a
+sixth check measures it there.** `find_citation_nulling_in_tree_sources` reads
```
jud-sdev: This calls the new tree-side arm the "sixth check", but line 1403 already opens "A sixth check runs in shadow mode only: the warnings guardrail...". Two different checks now carry the same ordinal, and this new one sits before the "fifth check" at line 1377, so any number here reads out of order.

Drop the ordinal rather than renumber:

You wrote:
> and a sixth check measures it there.

Change it to:
> and a companion check measures it there.

jud-sdev (changes requested):
Went through this by running rather than reading, and the code is careful, well-tested work. I reproduced the figures over the corpus exactly (research-side 0 across 159, tree-side 111 across 50 of 159), broke both load-bearing gates and watched the named tests go red (drop the upload gate, and remove the provenance self-exclusion), and confirmed the live orchestrator wiring has final_research and final_tree in scope so it will not crash on a real run. Full harness suite is green.

One change before merge, and it is the same one chesworthrm raised in their review, still unaddressed in the current head. The spec now has two checks both called the sixth one, and the new one reads out of order. Details in the inline comment. Once that is in I am happy with it.

One small process note, not blocking: the PR body does not fill in the template's Unsure about field. The substance is there under your own headings, but that field is the one a reviewer leans on most, so worth a line even when the answer is nothing.

jud-sdev (approved):
The ordinal fix is exactly right — companion check reads cleanly now, no collision, and the commit is the one-line doc change and nothing else. Everything I verified on the first pass still stands since no code moved: both gates bite under mutation, the 0 vs 111-across-50 figures reproduce over the corpus, and the live orchestrator wiring has its inputs in scope. Good work.

One thing for whoever merges: the branch is behind main, so it needs a quick update first.

PR #1920, opened 2026-08-25
eval: research-plan V1/V2/V3/V5 validators + V7 abort demotion (#1866)

### LINKED ISSUE #1866: 8 validator requests from the research-plan deep dive

Asimi1234:
## Summary

Normal tier. Implements the research-plan deep-dive validator requests from #1866:

- **V1** — an ARK/5+digit collection or volume identifier written into a new `plan_items[].rationale` must trace to a tool response that run received (`collections_search`/`volume_search`/`external_links_search`). Gating. Shares `candidate_identifiers` with `make provenance-report` so the two agree on what counts as an identifier (ARKs and 5+ digit numbers only — years never enter the check).
- **V2** — fail the run when it calls *or attempts* `wiki_search`/`wiki_place_page`/`place_population`, or delegates to `locality-guide`. Named prohibition, not the complement of the six (a complement gate would red `project_context`). Unions `tool_calls` + `attempted_mcp_calls` (a denied call only lands in the latter, #1748).
- **V3** — a new plan item's non-null `fallback_for` must name an item in the same new plan.
- **V5** — a rationale calling a served collection "indexed" must name one with `personCount > 0`; "browse-only"/"image-only" must name one with `personCount == 0`. Bound adjective↔identifier within one sentence; multi-collection sentences skipped.
- **V7** — a **commission**-validator failure now dominates a deterministic-cap abort: a run that failed a validator in `_COMMISSION_VALIDATORS` and then hit `max_wall_clock_seconds`/`max_turns`/`max_tool_calls` is recorded `fail`, not `aborted`. `aborted_reason` stays populated; only `outcome` changes. Scoped to commission validators per the lead's ruling — a cap truncates a run mid-write, so an omission-only failure ("expected X, got none") is the timeout's doing and stays `aborted`; only a wrongful *act* (wrote to a section it doesn't own, bypassed the writer tools) demotes. This is the `nothing-checks` item that hid V2's own evidence (`ut_research_plan_016` failed two validators, one of them the ownership table, but was logged as a timeout).
- Corrects the stale "a validator cannot see tool responses" note in `provenance_report.py` (a validator *can* — the runner passes the served body through as `tool_calls[].response`).

**Deferred exactly as the issue directs:** V4 and V8 (need a report-only channel that does not exist yet, Group M of #1749); V6 dropped (zero instances). No skill body or research-plan test/fixture touched, so the paid run-log gate is not armed.

## Review guidance

**Start here:** `_compute_outcome` in `eval/harness/harness/orchestrator.py` — the V7 demotion branch, now gated on `failed_validators & _COMMISSION_VALIDATORS`, and the `_COMMISSION_VALIDATORS` allow-list right above it. Per the lead's ruling the demotion is scoped by commission validator (not any failed validator); `max_input_tokens_per_turn` stays out, because once the demotion is scoped by validator, which cap fired no longer carries the classification weight. The set is an allow-list that fails safe — an unclassified validator's failure under a cap leaves the run `aborted`, the pre-V7 behaviour.

For V1/V5, the second thing worth a look is the identifier/sentence matching: V1 grounds on `_grounded_identifiers` (a superset of `candidate_identifiers` that also catches a hyphenated range's endpoints — see below) on both the served and before arms, while the *cited* side stays strict `candidate_identifiers`; V5 keys `personCount` on the response this run received, not on the id (1999196 is 0 in one fixture, 893214 in another).

**Unsure about:** nothing outstanding — the V7 scoping and the `max_input_tokens_per_turn` exclusion were ruled on and are implemented.

## Plan

**Didn't change:** `aggregate_per_run_outcome` logic (runlog.py) — it reads `_compute_outcome`'s already-demoted `outcome` string, so the two sites agree by construction; added a reconciling comment only. `not_runnable` aborts (they never reach `_compute_outcome`). `max_input_tokens_per_turn` (see above).

**Acceptance check:** `make harness-test` — proof-of-failure tests in `tests/unit/test_orchestrator.py` (V7) and `tests/unit/test_research_plan_validator.py` (V1/V2/V3/V5, the before-state-union case, and the hyphenated-range case). Each validator is exercised in both a must-pass and must-fire state; each is verified to red without its fix by temporarily reverting the change. The V7 commission-scoping and V1 range fixes (added after review) were each mutation-proven red:
- **V7 scoping** (`test_omission_only_validator_failure_under_cap_stays_aborted`) reds when the demotion is reverted to the old unscoped `not validators_passed` — the omission-only cap-abort gets mis-filed `fail`. The demotion-itself test reds when the demotion is removed.
- **V7 rename guard** (`test_commission_validators_are_all_collected`) enumerates every `def test_*` in the validators dir and asserts `_COMMISSION_VALIDATORS` is a subset; reds when a member is renamed away (the lead's condition).
- **V1 range** — two tests built from `007316661-007316663` red without `_grounded_identifiers`, flagging both endpoints as fabrications.

**Numbers (corrected after review — thanks @johnmarkpeterbrown, @promise-emmanuel, @EdmondOware):**
- `make provenance-report SKILL=research-plan` flags **7** persisted identifiers across 7 tests, all `1401638` (re-run on this branch). That is a corpus *report* and does **not** apply the before-state union, so it is an upper bound, not the gate's rate.
- The gating **V1**, replayed offline against `v1_2026-08-17_17-52-29` — the **oldest** of the five committed logs — fires on **7 of 21 runs (13 hits, all `1401638`)**; `1401638` is in neither any served fixture nor any firing scenario's `research.json`/`results/` sidecars, i.e. a genuine fabrication (confirmed by @EdmondOware). But `1401638` appears in **no rationale in the four newer logs** (08-24 ×3, 08-25), so this is a demonstration that the check catches a real fabrication in that one log — **not a forecast** for the next run.
- **Neither this 7-of-21 nor the report's 7 is V1's actual firing rate**, and neither should be read as one. Unit logs persist `response_fixture`, not the response body, and V1 grounds on what was *served* rather than *declared* — the very difference that makes V1 stricter than the report. Only a live run measures V1's real rate.

**Deviated from the plan:** Nothing in the original scope. Two changes were added in review: the V7 commission/omission scoping (after the lead's ruling) and the V1 hyphenated-range grounding fix (John's post-ruling finding) — both covered above.

## Test plan

- [x] I ran `make test-all` and it passed on the rebased head (ESLint, typecheck, JS workspace web 14 / electron 105 / viewer-ui 293, control-plane pytest 214, MCP vitest 2726, eval app 168, eval harness pytest **2955** — `All checks passed`).
- [x] No skill run-log snapshot changed — changes are confined to `eval/harness/**` and docs, so no `make eval-skill` run is required (and none would be armed).
- [x] No skill `description`/DO-NOT clause changed.

**Tier + judge-suppression accounting (per @EdmondOware).** V1/V2/V3/V5 are **gating**; V7 is a harness reporting change. A gating validator is not additive: a failing one skips the judge (`orchestrator.py`), so on every test where it fires, *all* of that test's judge dimensions score nothing. So the next paid research-plan run may look worse wherever V1/V2/V5 fire — intentionally, the gate doing its job rather than a regression. But **the size of that drop cannot be forecast from the committed logs** (V1's 08-17 hits are absent from the four newer ones), and will only be known from the live run — so a reviewer should not pre-accept a fixed per-validator count against this PR, or a real regression from an unrelated cause gets waved through with it. V7 is the counterpart fix: a validator failure must not read as an infrastructure abort (and now no longer prints a blank `REASON` either).

## Follow-on issues

**Folded in / filed instead:** Nothing filed. V4/V6/V8 remain recorded on #1866 as the issue directs (deferred/dropped there, not re-filed).

**Ruled and landed (was John's review item 1):** the lead ruled "go ahead with the commission/omission split," and it is implemented. V7 now demotes only when a failing validator is in `_COMMISSION_VALIDATORS` — a run killed mid-write fails only omission validators, which no longer mis-file as regressions. The set is seeded minimally with the universal wrongful-*act* checks a timeout cannot trip (`test_ownership_table`, `test_tree_ownership_table`, `test_no_raw_writes_to_protected_files`, `test_project_file_changes_route_through_writer_tools`) and is an allow-list, so it extends safely. `max_input_tokens_per_turn` stays out per the ruling. Corpus-correct by John's measurement: `ut_research_plan_016` demotes on its ownership-table failure; the three omission cap-aborts (`record_extraction_018/020`, `person_evidence_022`) stay `aborted`.

**Also landed (John's post-ruling finding):** V1's union grounding did not survive a hyphenated range — `candidate_identifiers`' `NUM_RE` refuses a digit run touching a hyphen, so a served/before-state range like `007316661-007316663` grounded nothing while the rationale, citing the volumes individually, flagged them. Fixed with `_grounded_identifiers` (a superset with a hyphen-tolerant pattern) on both grounding arms; the cited side stays strict, so grounding can only remove a false fabrication, never add one.

Closes #1866


johnmarkpeterbrown (changes requested):
The four validators are in good shape. V1 in particular is stricter than `provenance-report` in the right way — on `ut_research_plan_wzk` it catches three film numbers that live in a fixture the run declared but never actually received, which the report credits and misses. V2's `locality-guide` check fires on real corpus data. The proof-of-failure tests are the right ones.

Three things to change before this merges. The first needs a ruling from the lead; the other two are yours.

## 1. V7 records timed-out runs as skill regressions

A wall-clock cap kills the run part-way through writing. Any validator that says "expect X to exist" then fails *because* the run stopped, not because the skill did anything wrong. The demotion cannot tell those two apart.

I checked every committed unit run log. There are 11 cap aborts. Four have a failing validator, so this change would demote four runs. Three of the four fail only on things the run never got around to doing:

- `ut_record_extraction_018` — "no new assertion with record_role='head_of_household' fact_type='age'"
- `ut_record_extraction_020` — same validator, same shape
- `ut_person_evidence_022` — "wrote to the project but never invoked check-warnings"

The fourth is `ut_research_plan_016`, the example in the issue, and it splits down the middle. `test_ownership_table` is a genuine defect a timeout cannot explain. But `test_research_plan_new_plan_for_q_001` ("expected exactly one new plan; got []") is just the run being killed before it wrote the plan.

So three in four get filed as a quality regression when the cause was the clock. That is the same mistake this change exists to fix, pointed the other way — and it is the reasoning your own code comment uses to keep `error` and `sdk_stream_silence` out of the set.

*(Draft carries the full replacement diff: `failed_validators & _COMMISSION_VALIDATORS` in `_compute_outcome`, the new parameter, the call-site line, and the seeded set.)*

**Ask the lead before you write it** — the issue says "the three caps only", so the set membership is his call. Whichever way he rules, it lands in this PR: ~15 lines and one parametrised test. Also add a test for a truncated run whose only failing validator is an omission one; nothing would have caught this.

## 2. The new `eval/CLAUDE.md` sentence describes a row the harness never prints

All three operator-facing surfaces in `run_tests.py` gate on the outcome string before printing a reason. A demoted run prints `— fail` with a blank REASON. So the defect stops hiding behind a timeout, and the timeout starts hiding instead. Either render the reason on a `fail` row, or say what's true.

## 3. Both numbers in the acceptance check are stale

`make provenance-report SKILL=research-plan` on this branch gives 7 identifiers across 7 tests, all `1401638` — not 10 with 9× `1401638`. And V1 fires on 0 of 21 in the latest suite; the 7-of-21 is against the 2026-08-17 log. Expected damage is ~4 runs from V2 and V5, not 7 from V1. Table of all five run logs included.

Plus: an answer to their `max_input_tokens_per_turn` question, and a "checked, no action needed" list so they don't re-verify the six things that are fine.

---

promise-emmanuel (changes requested):
Code side is solid. I broke each of the five checks and watched the matching test
go red, then restored — V7, V1's before-state arm, V2's attempted-calls union, V3
and V5 all fail when their production line is removed. Harness suite is 2891
passed in the PR worktree; the one failure is the missing engine build in a fresh
worktree and passes where the build exists. Your item 2 fix works — I ran the
printer and a demoted row shows `fail   max_wall_clock_seconds` in the REASON
column, so the doc sentence is accurate now.

Two things before this merges.

1. The numbers in "Numbers (corrected after review)" are still wrong. I ran
   `make provenance-report SKILL=research-plan` on this branch and got 7, not 10.

   You wrote:    `make provenance-report SKILL=research-plan` flags **10** persisted identifiers (9× `1401638`, 1× `1927244`).
   Change it to: `make provenance-report SKILL=research-plan` flags **7** persisted identifiers across 7 tests, all `1401638`.

   And `v1_2026-08-17_17-52-29` is the oldest of the five committed research-plan
   logs, not the latest. `1401638` appears in no rationale in the four newer ones.

   You wrote:    The gating **V1**, replayed offline against the latest committed log (`v1_2026-08-17_17-52-29`) with the before-state union applied, fires on **7 of 21 runs (13 hits, all `1401638`)**
   Change it to: The gating **V1**, replayed offline against `v1_2026-08-17_17-52-29` — the **oldest** of the five committed logs — fires on **7 of 21 runs (13 hits, all `1401638`)**. `1401638` appears in no rationale in the four newer logs (08-24 ×3, 08-25), so this is not a forecast for the next run.

   The tier paragraph then needs the same correction — it currently tells the next
   reviewer to expect and accept a 7-test drop from V1. If they accept a drop V1
   did not cause, a real regression from something else gets waved through.

   Worth saying outright: neither your 7-of-21 nor John's 0-of-21 is V1's actual
   rate. Unit run logs persist `response_fixture`, not the response body, and V1
   grounds on what was *served* rather than *declared* — that difference is
   exactly what makes V1 stricter than the report. Only a live run measures it.
   Say that instead of picking a number.

2. Item 1 is still waiting on the lead. Nothing for you to do until he rules —
   flagging it so it does not look answered.

One optional fold-in while the branch is open: `max_input_tokens_per_turn` really
is the same class as the three in `_VALIDATOR_DOMINATED_ABORTS`. `skill_runner.py`
raises it, `orchestrator.py` groups all four as deterministic caps that are never
retried, and `test_orchestrator.py` already tests it beside the other three. Your
exclusion rationale (transient reasons feed the breaker, unmatched_tool_call is
exit-2 corpus) does not apply to it. Zero instances in the corpus, so no urgency —
two lines plus a parametrize entry if the lead wants it in.

Asimi1234:
@promise-emmanuel Item 1 fixed in the PR body — and I re-measured rather than reworded:

- `make provenance-report SKILL=research-plan` on this branch → **7 across 7 tests, all `1401638`** (not 10; `1927244` is in no committed research-plan log at all). Corrected.
- Confirmed `v1_2026-08-17_17-52-29` is the **oldest** of the five logs, and `1401638` has 55 hits there and **0** in all four newer ones (08-24 ×3, 08-25). The offline 7-of-21 is now framed as a demonstration against that one old log, explicitly **not a forecast**.
- Took your stronger point: the body now states outright that **neither the 7-of-21 nor the report's 7 is V1's real firing rate** (unit logs persist `response_fixture`, V1 grounds on what was *served*), so only a live run measures it. The tier paragraph no longer asks the next reviewer to pre-accept a fixed per-validator drop.

Item 2 (V7 commission/omission split) untouched — still on the lead. The `max_input_tokens_per_turn` fold-in likewise held for his ruling; zero corpus instances, so no urgency. Thanks for the careful read.

promise-emmanuel (approved):
Approving - This approval covers the developer-side items only. Item 1 is still the lead's and must be settled before this merges.

One thing to reconcile whenever you next touch the body: your comment says 55 hits
for `1401638` in the 08-17 log, the body says 13. I can confirm 7 of 21 firing
tests exactly but could not reproduce either hit count — run logs repeat each
rationale, so it depends on how you deduplicated. Nothing rides on it; just make
the two agree.

johnmarkpeterbrown (commented):
Items 2 and 3 are done — the REASON column and the live line now key on the reason being present rather than the outcome string, and the doc says what the code does. On the numbers we converged: 7 across 7 tests, all 1401638, and zero hits in the four newer logs. Good.
Two things left.
Ruling on item 1: do the commission/omission split
Go ahead with _COMMISSION_VALIDATORS.
The blanket demotion misfiles the majority of cases — 3 of the 4 I measured, 3 of the 5 you did. The tempting cheaper fix is to drop max_wall_clock_seconds from the dominated set instead, and that one is wrong: every case in the corpus is wall-clock, so V7 would fire on nothing and become a check that cannot fail. The allow-list also fails in the safe direction — a validator nobody classified will not demote.
One condition. A renamed validator drops out of the set silently, and the demotion quietly stops firing for it. That trades one invisible failure for another. Add a test that fails when a name in _COMMISSION_VALIDATORS no longer matches a collected validator, and prove it by renaming one.
max_input_tokens_per_turn stays out. Once the demotion is scoped by validator, the cap list stops carrying that weight.
New: the V1 union still red-flags two of the three ids it was written to rescue
The union grounding is the right idea, but it does not survive a range. martha-remarriage-surname-plan writes the volumes as (FamilySearch volumes 007316661-007316663). candidate_identifiers refuses a digit run touching a hyphen, so that string yields no identifiers at all — one blocked by the trailing -, the other by the leading one. The rationale cites them individually, where they extract fine. So both are still flagged as fabrications.
    served-only rule union rule (current head)     ut_research_plan_wzk 007255720, 007316661, 007316663 007316661, 007316663   Only 007255720 is rescued, and only because it's written comma-separated. json.dumps on the parsed before-state preserves the hyphen, so this is the live path.
(Draft carries the three-part fix: _grounded_identifiers helper with a looser _GROUNDED_NUM_RE, swapped into both the served arm and the before arm, plus the ask for a proof test built from a hyphenated range.)
Once those two land I'll read the final head and approve. Note Promise's approval sits at 55ee5c8f and will still count after your next push — stale approvals aren't dismissed here — so mine should be the one that reads the code that actually merges.

Asimi1234:
@johnmarkpeterbrown Both landed — head is now `55a9fed7` (rebased onto the latest main merge; full harness suite 2955 passed, 2 skipped).

**V7 commission/omission split.** `_compute_outcome` now demotes a cap abort only on `failed_validators & _COMMISSION_VALIDATORS` (new param, threaded from the failed-validator names already computed for the judge). Seeded minimally with the universal 'the run actively did something wrong' checks — a timeout can't trip them:
`test_ownership_table`, `test_tree_ownership_table`, `test_no_raw_writes_to_protected_files`, `test_project_file_changes_route_through_writer_tools`. Allow-list, fails safe. `max_input_tokens_per_turn` stays out. Corpus-correct by your measurement: 016 demotes on its ownership-table failure; the three omission aborts (record_extraction_018/020, person_evidence_022) stay `aborted`.

Each guard proven red by mutation (not just green):
- **Scoping** (`test_omission_only_validator_failure_under_cap_stays_aborted`) — reds when I revert the condition to the old unscoped `not validators_passed` (the omission-only case gets mis-filed `fail`). This is the 'nothing would have caught this' case.
- **Demotion** — reds when the demotion is removed.
- **Your rename condition** (`test_commission_validators_are_all_collected`) — enumerates every `def test_*` in the validators dir and asserts the set is a subset; reds when I rename a member away.

**V1 hyphenated range.** Added `_grounded_identifiers` (a strict superset of `candidate_identifiers` via a looser `_GROUNDED_NUM_RE` that tolerates hyphen adjacency), swapped into **both** the served and before arms; the cited side stays strict. Two proof tests built from `007316661-007316663`, both **red without the fix** (they flag both endpoints, exactly your table). Grounding stays a superset, so it can only remove a false fabrication, never add one.

Ready for the final-head read.

johnmarkpeterbrown (approved):
Approving on `90b648cb`. Both items landed and I verified each by running it rather than reading it.

**V7 commission split** — replayed `_COMMISSION_VALIDATORS` across every committed run log. `ut_research_plan_016` still demotes on its ownership-table failure; `ut_person_evidence_022`, whose only failures are omissions, stays `aborted`. That is the split doing exactly the job, on real data.

**The rename guard** — proven red both ways. Renaming a non-canary member gives the primary assertion with a message that says what to do:

```
_COMMISSION_VALIDATORS names validators no longer collected:
['test_no_raw_writes_to_protected_files']. A validator was renamed — update the
set, or its failure under a cap silently stops demoting to fail.
```

The guard-the-guard on the AST scan is the right addition — without it an empty scan would have made the subset check pass vacuously. Good instinct.

**V1 hyphenated range** — replayed with the shipped `_grounded_identifiers`. `ut_research_plan_wzk` now flags nothing, and the seven `1401638` fabrications in the old log are still caught. Grounding as a strict superset with the cited side left strict is the right shape: it can only remove a false fabrication, never add one.

Harness suite here: 2949 passed, 9 skipped, 1 failed — the missing-engine-build test that reds in any fresh worktree, not yours.

---

One thing for the record, not a blocker. `test_research_plan_no_out_of_lane_tools` is not in `_COMMISSION_VALIDATORS`. Calling `wiki_search` is an act a timeout cannot cause, so it arguably belongs there. The omission fails in the safe direction — that run stays `aborted` rather than being mis-filed — and you documented the set as deliberately minimal and extend-as-classified, so this is a note for whoever extends it next, not a change to make now.

Also worth flagging to whoever merges: Promise's approval sits at `445ec0c7`, which predates both of these fixes. Stale approvals are not dismissed on this repo, so it still counts toward the two. Mine is the one that read the code as it stands.

Nice work on the turnaround, and on measuring rather than rewording each time.

Issue #572, opened 2026-07-03
try to get the elena-asmundsdotter-origin e2e test to pass


DallanQ:
**IMPORTANT**: If you do this, do **issue #1596** at the same time. It corrects this fixture's
"Recoverability caveat" — the Mantalslängder for Harjager härad *are* on FamilySearch as
browse-only image volumes, with 1681/1700/1702/1705/1708 inside this fixture's own 1680–1711
window — and it edits the same `eval/tests/e2e/elena-asmundsdotter-origin/` files. A run graded
against the uncorrected caveat is graded against a claim we know is half wrong.

**Touches:** eval/tests/e2e/elena-asmundsdotter-origin/, eval/runlogs/e2e/elena-asmundsdotter-origin/

> **Scope line added 2026-08-11 by the daily fill-ready pass.** The issue was filed
> 2026-07-03 with an empty body, which makes it unactionable in a fresh session. What
> follows is read off the repo, not off the original intent — correct it if it misses
> the point.

## Where it stands

The fixture exists in full — `fixture.json`, `expected-findings.json` (6 findings),
`starting-research.json`, `starting-tree.gedcomx.json`, `README.md` — added by merged
PR #563. **It has never been run:** `eval/runlogs/e2e/elena-asmundsdotter-origin/` does
not exist, so there is no committed run log, no verdict, and nothing recorded about why
it might fail. It is one of 20 fixtures out of 105 with zero committed runs.

Nothing is wrong with the fixture's shape. Its `source_pid` and
`project.subject_person_ids` are both `PID-TODO`, which is the documented normal state
for a PID-less fixture (10 fixtures carry it) — `load_fixture` skips the marker
explicitly (`eval/harness/e2e/orchestrator.py`), and those fields feed only the
avoid-guard's subject exemption, not the tree lookup.

## What the fixture asks

> Where was Elena Asmundsdotter — the mother of Asmund Jönsson (born 11 September 1718
> in Barsebäck parish, Malmöhus, Sweden) — born, and who were her parents?

`difficulty: hard`, `question_type: parents`, era 1680s, geography SE. Its own notes say
why: Västra Karaby parish registers begin only in 1688 and Elena was born c. 1681, so
her origin is reachable only by correlating Swedish tax lists (Mantalslängder, from
1658) with catechism records (Katekismilängder) across five decades. This is an
indirect-evidence case — expect it to be hard for the agent, and expect a first run to
fail rather than pass.

[...]

PR #1921, opened 2026-08-26
elena-asmundsdotter-origin: pin the source pages, re-run on raised caps, fix an image_search null (#572)

Gennecis:
## Summary

Resumes #572. The job was measurement, not passing the test, and the measurement is now
much sharper than a 0/6 suggests.

**All six source pages were located and read by hand, and are pinned by image ID in the
README.** At the time runs 1 and 2 were made, no shipped tool could read them: `image_read`
refuses all of them (4.4–5.5 MB against its 700 KB cap, #1593) and directs the caller to
`image_transcribe`, which on the f4 page returned `Jöns`→`Widus`,
`Henckestorp`→`St. Davidsthofst`, `Elna Asmuns dotter`→`Enna de mans Sohn` — a patronymic
read as a masculine construction, in a fixture whose whole subject is a daughter's
parentage — and answered NOT FOUND. Three calls on that page failed three ways, including a
17× repetition loop not recorded in #1594.

**#2044 then fixed the reading layer, and run 3 re-measured on it.** f1 and f6 now
transcribe verbatim, f2 and f4 readably. The residual `af Henckestorp` on f4's line defeats
Gemini three times out of three *and* defeats Opus 5, so it is a hard page region rather
than a model choice, and f4 does not need it.

### The three runs

| | stop | tool calls | persons in final tree | wrong parentage written? |
|---|---|--:|--:|---|
| run 1 `08-13` | `timeout` (3600s cap, 197/200 calls) | 197 | 3 — starting state | **no** |
| run 2 `08-25` | `completed` | 140 | 5 — adds `Assmen Nielsson`, `Bentte Jeppisdr` | **yes** |
| run 3 `09-01` | `timeout` (120 min cap) | 168 | 3 — starting state | **no** |

**Run 2 is the worst of the three, and the comparison is the genealogical point of this
PR: writing nothing is a better failure than asserting a false parentage.** Runs 1 and 3
both left the tree at its starting state. Run 1 did so *having reached the same wrong
Fulltofta candidate* — 34 mentions in its research document, a proof summary tiered
`possible`, three `person_evidence` entries at `probable`, question marked `resolved` — and
still wrote none of it to the tree, because `possible` is below the write threshold. That is
restraint, not a run that ran out of road. Run 2 reached the same candidate, tiered it
`probable`, and wrote both parents in; its compliance fails because `same_person` was never
called for either new person. Run 3 recovered nothing but also asserted nothing, with
`compliance: pass` and zero bypasses.

**Two side findings, both fixed here.** The register's start date question is settled (1688,
read off the volume), and `image_search` was passing a null upstream value through to
`imageIds` — emitting a bogus ID and silently dropping a real image.

## Review guidance

- The fixture edit that needs the closest look is the **caps removal**. It is not tuning to
  make the test pass: the pin was an authoring-batch artifact at half the `FixtureCaps`
  default, byte-identical in three sibling fixtures from the same batch, so nothing in it was
  a decision about this fixture. `susan-miller-birth` is the precedent for raising a cap
  after a timed-out run. Run 2 then finished well inside the raised caps (41.9 of 120 min,
  140 of 300 calls), so it did not affect that result — but **run 3 used the full raised
  120-minute wall clock and timed out at it**, which is the point of raising a cap after a
  timed-out run rather than a case against it. Under the old 3600 s pin run 3 would have been
  cut at roughly half its work.
- `expected-findings.json` is untouched and no finding was added or renumbered, so run 1's
  committed annotation still labels exactly f1–f6.
- The `image_search` change is the one behavioural change to shipped code. The retry is
  bounded to a single extra call and only fires on a detected defect.
- The README now quotes early-modern Swedish; the translations are in the surrounding prose.
- `check_e2e_fixtures.py` emits one **warn-only** component-derivation notice on f4 and asks
  a human to confirm the label. Confirmed: the blind human grade also recorded f4 as `false`.
  The Couple relationship in the final tree came from the *starting* tree and carries no
  marriage fact, date or place, so nothing about f4 was recovered.

## Test plan

- `npx vitest run` in `packages/engine/mcp-server` — 111 files / 2706 tests green.
- `npx tsc --noEmit` clean.
- The **seven** new `image_search` tests were **written first and watched fail** against
  `main`: reverting `src/tools/image-search.ts` to `main` fails **six of the seven**, the
  exception being "does not re-request when the first response is clean", which passes as the
  correct baseline since `main` never retries.
- **Real-body replay**: the live 164-key response with the exact observed defect (`00672`
  nulled) run through the real tool. Pre-fix control reproduces the wild symptom — 164
  entries, 1 null, `00672` missing. With the fix: retry fires, 164 IDs, 0 nulls, `00672`
  restored, sorted. Both-defective: 163 clean, no null.
- **40 live iterations** against the real endpoint: no regression, and the defect did not
  recur naturally — so the retry is verified by replay, not by wild capture. Stated plainly
  because it bounds what the live probe proves.
- `make e2e-validate TEST=elena-asmundsdotter-origin` — schema valid, all findings stripped.
- `check_e2e_fixtures.py` against the branch with PR context — grading gate OK, **2** added
  run logs checked (runs 2 and 3), plus the one warn-only f4 derivation notice.
- Run 2's `.ann.json` is committed alongside the run log, per the e2e grading gate.

## Follow-on issues

None filed. Evidence was added to #1593 and #1594 rather than opening new items — both are
open, #1593 is `senior`, and #1594 is assigned, so neither was touched here.

One candidate deliberately **not** filed yet, pending evidence: `image_search` cannot detect
a response that omits a child outright (valid strings, indistinguishable from a shorter
volume). Catching that needs a cross-check of `imageIds.length` against `volume_search`'s
`imageCount`. Worth an issue only if the shape recurs; recorded in the spec meanwhile.


chesworthrm (changes requested):
Full pass on a worktree, branch up to date with main, all required checks green on this head. The measurement work here is careful and the annotations are unusually honest — grading f2 `false` rather than `partial` because the only correct element was the free patronymic is exactly right. One thing to fix in the code before merge.

## The retry can turn a working browse into an error

`packages/engine/mcp-server/src/tools/image-search.ts:130` — the re-request isn't wrapped, so when the second call fails the whole tool throws and the caller gets nothing.

I ran both sides. Against a 164-key response with one null, the old code returned 163 usable image IDs. With the new code and a 500 on the retry, it returns none and raises instead.

This also contradicts what you added to the spec in this same PR, `docs/specs/image-search-tool-spec.md:142` — "the surviving IDs are returned rather than raising". That holds when the retry is *defective*, but not when it *fails*, and a 500, a timeout or a 401 on the second call is at least as likely as a second null.

While you're in that block: the comparison tests "fewer dropped" before "more IDs", so a clean but shorter retry can win over a longer defective one. At the sizes you actually observed this costs nothing — 164-with-null and 163-clean both yield the same 163 IDs, missing the same image — so it's latent rather than urgent. It's just free to fix while you're here.

You wrote:

```ts
  if (best.dropped > 0) {
    const retry = usableImageIds(await fetchChildren(groupId, token));
    if (retry.dropped < best.dropped || retry.imageIds.length > best.imageIds.length) {
      best = retry;
    }
  }
```

Change it to:

```ts
  if (best.dropped > 0) {
    try {
      const retry = usableImageIds(await fetchChildren(groupId, token));
      if (
        retry.imageIds.length > best.imageIds.length ||
        (retry.imageIds.length === best.imageIds.length && retry.dropped < best.dropped)
      ) {
        best = retry;
      }
    } catch {
      // keep `best`
    }
  }
```

I applied that and ran it: your four new tests pass along with the other eleven, and `tsc --noEmit` is clean.

Please add a fifth test — first response defective, retry rejects, assert the surviving IDs come back and nothing throws.

The spec sentence at line 139 describes the old ordering and needs to follow:

You wrote:

```
   retry when it is better (fewer dropped, or more IDs). A retry is what
```

Change it to:

```
   retry only when it carries more usable IDs, or the same number with fewer
   dropped, and never at the cost of failing outright when the retry errors.
   A retry is what
```

## The caps removal holds up

You asked for the closest look here, so: it checks out. The deleted block is byte-identical in three sibling fixtures from the same authoring batch, so nothing in it was a decision about this fixture. And the turn cap you removed would not have bound this run — across every committed e2e run log, `max_turns` has never once been the stop reason, and six runs sit above their fixture's cap, one at 150 turns against a cap of 100.

I also considered asking for a run-2 section in the fixture README and decided against it: of the 37 fixtures with more than one run, 34 document none of them. This README already documents more than almost any other.

## Still owed

One senior approval from either senior-developers or senior-genealogists, plus the second write-access approval. Note `docs/specs/image-search-tool-spec.md` is matched by no CODEOWNERS rule at all — there's no `*.md` entry — so the spec half pulls no senior on its own.

Everything else I checked was accurate as written: the suite counts, run 2's numbers, the single warn-only grading-gate notice, and the wrong parents sitting in the committed tree where run 1 wrote nothing.

Gennecis:
Thanks — the retry bug is real and I've fixed all four items in `6282596d`. Branch is also merged up with main (220 commits, no conflicts), so this now sits on top of #2044.

## Review items

| Item | Done |
|---|---|
| Unwrapped retry throws on a failing second call | Wrapped; `best` is kept on any failure |
| `dropped` compared before ID count | Reordered — usable IDs primary, `dropped` only a tie-break |
| Missing test for a rejecting retry | Added, plus a 500-on-retry case and one pinning the tie-break |
| Spec sentence described the old ordering | Rewritten, and it now states the never-fail-outright guarantee explicitly |

All three new tests were watched fail against the pre-change code before the fix went back in. 18/18 in that file, **118 files / 2790 tests** green, `tsc --noEmit` clean.

You were right that the spec sentence was the tell — it promised "the surviving IDs are returned rather than raising", which was true for a *defective* retry and false for an *erroring* one. The code and the spec disagreed inside one PR.

## #2044 changes the picture for this fixture

Re-probing the pages this PR pinned, now on Gemini 3.7 Flash:

| Page | Finding | Gemini result |
|---|---|---|
| `004514823_00347` | f1 | **Exact.** *"d. 23. april: dödde Jöns Jönssons Hustru Elna Asmunds dotter, född i Henckelstorp, … fär 64. åhr gl."* |
| `004523018_00033` | f2 | **Readable.** *"Kallat Boell hvars Moder är Asmontors doter af Henckelstorp, Elena … Asmunds doter blef publice absolverat"* |
| `004523018_00053` | f4 | **Readable.** *"Fer: 4:ta Nativ: copulerades drengen Jöns Jönson och Qwinfolcket Elena Asmunds doter."* |
| `004523018_00072` | f6 | Not measured — see below |

Against Qwen on the same f4 page, which returned `Widus` / `St. Davidsthofst` / `Enna de mans Sohn`, this is a different tool. f1's entry is verbatim correct including `Henckelstorp` and the age.

Two residual observations, neither filed:

- **`af Henckestorp` is omitted from f4's line** — dropped, not fabricated (0 occurrences of "henck" on that page's output). Safer than Qwen's invention, but it is the token f1 and f2 hang on, and on this page it is adjacent to a struck-through region.
- **f2's page returned `NOT FOUND` while its own transcription contains the answer.** Not a parsing bug: `parseFound` reads the last line correctly, and the model itself judged not-found while transcribing `Asmontors doter af Henckelstorp`. The spec already frames `found` as "a locate hint only", so this is the hint being unreliable rather than a contract breach — recording it rather than filing, since nothing open covers it (#1974 is a different mechanism).

## The third run is held, and not for a product reason

f6 could not be measured because the **FamilySearch image download** times out at its 90 s budget — a different leg from the OpenRouter retry #2044 added. Measured here:

- `004523018_00053` (4.6 MB) downloaded in **83.2 s**, 7 s inside the budget
- `004523018_00072` (5.4 MB) timed out, three attempts

That is ~55 KB/s. Three independent Cloudflare samples on this host: **50–77 KB/s**. The rates match, so this is my local link, not FamilySearch — CLAUDE.md's documented expectation is ~7 s for one of these. **Nothing to file.**

Consequence: a scored run right now would spend itself on download timeouts and measure my broadband rather than the fixture. Holding it until throughput recovers. Everything else on this PR is ready.

Gennecis:
## Run 3 landed, and it relocates the problem. Proposal below — would like the team's view before anyone builds.

Branch is synced with `main` (clean merge), 118 files / 2816 tests green, fixture validates.

### What run 3 settled

`run-2026-09-01_22-19-45`: `timeout`, 0 of 6, graded blind. Better on every axis except the
one that decides it — 6 of 57 transcribes errored (run 2: 50%), `compliance: pass`, zero
guardrail bypasses, and **nothing wrong written** to the tree, where run 2 asserted a
false father and mother.

**#2044 fixed the reading layer.** Verified page by page: f1 `004514823_00347` and f6
`004523018_00072` now transcribe verbatim, f2 and f4 readably. The one residual —
`af Henckestorp` on f4's line — defeats Gemini three times out of three *and* defeats
Opus 5 (`Grenhestarps[?] [illegible]`), so it is a hard page region rather than a model
choice, and f4 does not need it.

**First, a correction to my earlier claim on this PR.** I wrote that the fixture was
"winnable". I demonstrated the pages are *legible*, not that they are *findable*. Elena
carries **zero facts** in the starting tree and `1745` appears nowhere in the fixture, so
the agent does not know when she died; my hand-search was cheap only because I did.
Finding f1 means sweeping decades of an unindexed death register. Different claim, and I
conflated them.

### Root cause, from the run's own structured data

Not the reading layer and not navigation as such:

1. **All four sources run 3 recorded are classified `derivative` — by the agent itself.**
   Including `src_001`, a *1954 LDS extraction form*, which its plan then cites as
   identifying *"her as 'Elina Asmundsson' daughter of Asmund Hansson"* and concludes:
   **"Supersedes pli_003 in priority because it targets the identified birth parish."**
2. **The Barsebäck browse was marked `completed` after 4 examined pages of 749**
   (`pli_003 | image_transcribe | negative | results_examined: 4`).
3. Images read in that volume: `[1, 50, 220, 300, 300, 550, 630]` — 300 read twice, then
   **300 → 550, straight past f1 at 347**, immediately after learning 300 was 1731 and 550
   the 1760s. It held both interpolation points and did not use them.

This is the third instance of one pattern: run 1 → Fulltofta (index hit), run 2 →
Fulltofta (two index entries), run 3 → Kävlinge (compiled extraction form). **Three runs,
three wrong parishes, every one sourced from a derivative record naming a same-named
Asmund.**

### Why this is not a prose fix

The doctrine is already written, in both skills that ran:

- `search-records/SKILL.md:568` — *"**Never** treat an index entry as equivalent to
  examining the original."*
- `research-plan/SKILL.md:298` — *"**Original before derivative**"*

The agent read the correct rule, **classified the source correctly as derivative**, and
overrode the browse of the subject's own parish anyway — three times out of three. Prose
plus accurate self-classification is the intervention we now have direct evidence
against.

### Proposal — three candidates, ADR-0011's decidability test applied

**1. Coverage-honesty precondition — fully decidable today, and my recommendation.**
Refuse marking a plan item `completed` when its linked log entries show `results_examined`
grossly below `results_available`. `log[]` already carries `plan_item_id`,
`results_examined` and `results_available`; nothing needs inventing. This is the guard
that would have stopped the walk-away from the volume holding f1 after 4 pages.
*Open question for the team:* refuse, or advise? #2067 is currently measuring whether
advisory return-fields change behaviour at all, and our own run contributes a
fired-9/complied-0 data point — which argues for a refusal here, but that is the lead's
call.

**2. Year→image navigation, as a tool.** Given a volume's date range and image count,
bisect on year headings to a target year. Verified by hand at ~6 reads per finding
against the agent's 7 scattered ones. Deterministic and mechanical — it belongs in code
rather than in judgement.

**3. Derivative-supersession guard — right rule, currently blocked.** "A plan
re-prioritisation whose only support is `source_classification: derivative`" matches the
actual failure, but is **not decidable today**: plan items carry no source link
(`id, sequence, record_type, jurisdiction, date_range, repository, rationale,
fallback_for, status`), so the citation lives in free prose. It needs a `source_ids`
field on plan items first — a schema decision, with the multi-site cost that implies.

Ranked 1 → 2 → 3. Nothing filed yet, deliberately: 1 and 2 are junior-sized once the call
is made, 3 needs a schema decision. Happy to draft whichever the team wants.

### What this PR is now

Three graded runs and a page-level map of exactly what is readable and where it sits.
That map is what makes the navigation problem legible to whoever picks it up. Still owed
one senior approval; still a draft pending the team's read on the above.

chesworthrm (approved):
Approved — all four items from my 2026-08-26 pass are in, and this clears that CHANGES_REQUESTED.

Verified on `cd7efc73`:

| Item | State |
|---|---|
| Unwrapped retry could turn a working browse into an error | Wrapped; `catch` keeps `best`, with the reason at the line |
| `dropped` compared before ID count | Reordered — usable IDs primary, `dropped` a tie-break only |
| Missing test for a rejecting retry | Three added: rejecting retry, 500 on retry, and one pinning the tie-break |
| Spec sentence described the old ordering | Rewritten, and it now states the never-fail-outright guarantee |

Seven tests in that file, all 17 required checks green on this head.

Two notes for whoever merges, neither blocking:

- The branch is 30 commits behind main (mergeable, no conflicts). Merge up before merging.
- Still owed one senior approval and the second write-access approval. `*.ts` and `*.json` both pull CODEOWNERS, so the senior requirement is live on this one.

On the run-3 correction: noting it explicitly — that you had shown the pages *legible* rather than *findable*, and that the two are different claims — is the right instinct, and it is what makes the navigation finding trustworthy. The three-runs/three-parishes table earns its conclusion.

Ruling on the open question in a separate comment.

chesworthrm:
## Ruling on the coverage-honesty precondition: **refuse, not advise**

**DECIDED 2026-09-02 — a plan item may not be marked `completed` when its own linked log entries show `results_examined` grossly below `results_available`. Refuse the write.**

Three reasons, in the order they weighed:

1. **Advisory bets on the thing still being measured.** #2067 exists because we do not know whether advisory return-fields change behaviour at all. Choosing "advise" here would spend a fixture on that open question instead of on the fixture. And this run is itself a data point against it — fired 9, complied 0.
2. **ADR-0011 puts it in the tool.** The question "did this browse examine a defensible fraction of what it had?" is answerable from `log[]` alone: `plan_item_id`, `results_examined` and `results_available` are all already there. Nothing needs inventing, no judgement is involved, and a precondition binds everywhere a prompt does not.
3. **Prose has now failed three times out of three with the rule read and correctly applied.** The agent classified the source as derivative and superseded the browse anyway. That is the strongest possible case that this cannot be a body edit.

Scope, so it does not grow: the guard is about **coverage honesty on one plan item**, not about derivative supersession. It stops the walk-away from a 749-page volume after 4 pages. It does not attempt to stop the agent preferring a derivative source — that is a separate guard, and it is not decidable today (see the filed issue on plan-item `source_ids`).

Two implementation calls I am making now so this does not come back as a question:

- **Threshold is a ratio with a floor, not a fixed count.** 4 of 749 must fail; 40 of 45 must pass. Pick the numbers from the committed corpus, not by taste, and put the derivation in the spec.
- **`negative` outcome is not an exemption.** Run 3 recorded `negative` on 4 of 749 examined, which is exactly the shape being caught. A nil result on a fraction of the available records is the claim under scrutiny, not a reason to skip it.

Filed as an issue against this PR; ruling reproduced in the body.

chesworthrm:
Follow-ons from this PR are filed, so nothing here is carrying an undocumented limitation:

- **#2182** — coverage-honesty precondition in `research_append`, refuse not advise. Ruling reproduced in the body.
- **#2183** — year→image bisect navigation tool. No decision pending; buildable now that #2044 made the reading layer reliable.
- **#2184** — derivative-supersession guard, blocked on whether `source_ids` lands on `plan_items`. `needs-decision`.

Ranked 1 → 2 → 3 as you proposed, and I agree with that order. #572 is closed against the measurement.

Your two residual observations stay where you put them — evidence on #1593/#1594 and in the spec — and I agree neither warranted a new item. The `image_search` omitted-child case (valid strings, indistinguishable from a shorter volume) is correctly left unfiled until the shape recurs; the `imageIds.length` vs `imageCount` cross-check is the right fix if it does.

jud-sdev (approved):
Went through this in depth and ran it in a worktree. Careful, honest work.

The image_search fix is the right shape: filter non-string values so a null can never be served as an image ID, then one bounded retry that only replaces the first attempt when it is strictly better — more usable IDs, or the same number with fewer dropped — and a catch that keeps the survivors when the retry rejects or 500s. I walked every branch of the comparison and it can only improve the result, never make it worse, and the first call staying unwrapped is correct since a first-call failure should propagate. No regression on clean responses; the filter is a no-op there.

The seven tests are genuine guards, not decoration — I reverted the source to main and six of them failed (the clean-no-retry one is the correct baseline that passes on old code). image-search.test.ts is 18/18, the full engine suite is 2826/2826, and tsc is clean. The spec update describes the shipped behaviour faithfully, including the 163-key omitted-child case it can't detect and the volume_search cross-check you deliberately deferred.

On the fixture: the caps removal is inconsequential — run 2 finished at 41.9 min / 140 calls, inside even the old caps — so it is an authoring-artifact cleanup, not tuning to pass, and expected-findings.json is untouched so run 1's annotation still labels f1 to f6. Both run logs are newly added with their .ann.json, and the README pins are well formed.

EdmondOware:
Worth fixing before this merges, and the README is right. This is about the PR description, which inverts which run earned the good outcome.

The README tells a three run story: run 1 timed out at 197 of 200 calls under the old caps, run 2 completed and asserted a false father and mother, run 3 timed out at the 120 minute cap and left the tree at its starting state.

The description tells a two run story and swaps the last two:

> Run 2 ... wrote the false Fulltofta parents into the tree at `probable` where run 1 had held them at `possible` and written nothing.

Per the README it was run 3 that wrote nothing, and run 1 never got far enough to write anything at all.

The artefacts agree with the README rather than the description. The 08-25 log carries 140 tool calls, `stop_reason: completed`, five persons, and two mentions each of Fulltofta and Nielsson. The 09-01 log carries 168 calls, `stop_reason: timeout`, three persons, and no Fulltofta at all.

The reason to fix it before merge rather than whenever: this repo squash merges, and the squashed commit body carries the PR description into `git log` permanently. So the inverted version would sit in the history next to a README saying the opposite, and the next person reconstructing why this fixture behaved as it did has to work out which to believe.

It matters because the point the run comparison exists to make is a genealogical one, and it is a good point: writing nothing is a better failure than asserting a false parentage. Your README says exactly that, twice, and says it well.

One thing worth pulling into the README's reading-layer section if it is not already there in those words: `Elna Asmuns dotter` came back as `Enna de mans Sohn`. That is a patronymic read as a masculine German construction, in a fixture whose whole subject is a daughter's parentage. It is the sharpest single illustration of why the reading layer failed on this record, and it is more legible to the next person than the error rate is.

johnmarkpeterbrown (commented):
Senior genealogist pass. The measurement work is sound and I checked the
grading independently: rel-jons-elena is byte-identical between the starting
tree and run 2's final tree and carries no facts, and the new Couple R3 links
the Fulltofta pair rather than Jons and Elena. So f4 = false is right and the
judge's `partial` is the derivation artifact, exactly as you said.

One thing before I approve: the description still tells a two-run story and
run 3 isn't in it. We squash-merge, so it lands in git log permanently next to
a README that says something different. Three small edits and I'll approve:

- Add run 3, and make clear it's run 3 that left the tree at its starting
  state.
- "1 added run log checked" -> 2. I ran the gate with PR context and it
  reports 2.
- "The four new image_search tests ... 3 failed" -> seven and six. I reverted
  the source to main and six of the seven fail; the clean-response one is the
  baseline.

Also worth a clause on caps: your argument holds for run 2, but run 3 used the
full raised 120-minute wall clock and timed out at it. That's the point of
raising a cap after a timed-out run, not a case against it — just say so
rather than leaving "changed nothing" standing.

The "Enna de mans Sohn" line EdmondOware asked for is already in the README's
2026-08-25 note, with the "bride rendered as a son" gloss. Just point at it.

Gennecis:
Verified both points against the artefacts. **Description updated** — you were right that it needed fixing before a squash merge, and right about why. One part of the correction does not hold, though, and it matters for the point you say the comparison exists to make.

## The description was stale, not inverted

It was written on 08-26, when only runs 1 and 2 existed, so it told a two-run story and omitted run 3 entirely. That is the real defect and your squash-merge argument for fixing it now is the right one.

But its claim about run 1 is accurate. From the artefacts:

| | persons in final tree | Fulltofta in tree | Nielsson/Nilsson in tree |
|---|--:|--:|--:|
| run 1 `08-13` | 3 — starting state | 0 | 0 |
| run 2 `08-25` | 5 | 2 | 2 |
| run 3 `09-01` | 3 — starting state | 0 | 0 |

**Runs 1 and 3 both wrote nothing.** So "run 1 … had written nothing" was correct; it was incomplete in not also saying run 3 did the same.

## Run 1 did not fail to get far enough — it got there and restrained

This is the part worth correcting, because it is the more interesting fact and the description now says it explicitly. Run 1's `final-research.json`:

- **34** `Fulltofta` mentions
- a **proof summary tiered `possible`**
- **three** `person_evidence` entries at confidence `probable`
- question status `resolved`
- and **0** Fulltofta in the final tree

So run 1 reached the same wrong candidate run 2 did, formed a conclusion about it, and kept it out of the tree because `possible` sits below the write threshold. That is the write gate doing its job. Run 2 reached the identical candidate, tiered it `probable`, and wrote both parents in.

That sharpens your point rather than weakening it: the difference between run 1 and run 2 is not how far each got, it is what each was willing to assert on the same evidence. A run that reasons its way to a wrong parentage and declines to record it is the behaviour we want; run 2 is the regression.

The description now carries a three-run table plus that reasoning, so the squashed commit body and the README will agree.

## On `Enna de mans Sohn`

Already in the README, in those words — line 223, in the reading-layer note: *"`Elna Asmunds dotter`→`Enna de mans Sohn` (the bride rendered as a son)"*. I have pulled the same phrasing up into the PR description too, since you are right that it is more legible than the error rate, and the description is what reaches `git log`.

Gennecis:
Thanks — I verified all three checkable numbers before editing, and all three were right. Description updated.

| your correction | verified | how |
|---|---|---|
| "1 added run log checked" → 2 | ✅ | `check_e2e_fixtures.py` with PR context: *"E2E grading gate OK (2 added run log(s) checked)"* |
| "four tests / 3 failed" → seven and six | ✅ | 7 new `it(...)` blocks vs `main`; reverting `src/tools/image-search.ts` to `main` fails **6**, the clean-response baseline passing |
| caps clause | ✅ | run 3's `usage.wall_clock_seconds` is **7223** against the raised 7200 cap, `stop_reason: timeout` |

The test count was my own drift: the description was written when the PR had four new tests, and the review-fix commit added three more without my updating it. The caps bullet was stale the same way — "changed nothing about this result" was written when run 2 was the only result. Both now say what the artefacts say, and the caps bullet makes your point explicitly: run 3 consuming the full raised budget is the argument *for* the raise, and under the old 3600 s pin it would have been cut at roughly half its work.

**One nuance on your first edit, same one EdmondOware hit.** You asked me to "make clear it's run 3 that left the tree at its starting state". It is both runs 1 and 3, and the description now says so with a table, because run 1 is the more interesting case:

| | persons in final tree | Fulltofta in tree |
|---|--:|--:|
| run 1 `08-13` | 3 — starting state | 0 |
| run 2 `08-25` | 5 | 2 |
| run 3 `09-01` | 3 — starting state | 0 |

Run 1's `final-research.json` carries **34** `Fulltofta` mentions, a proof summary tiered `possible`, three `person_evidence` entries at `probable`, and question status `resolved` — with none of it in the tree, because `possible` is below the write threshold. So run 1 reached the same wrong candidate run 2 did and declined to assert it. Attributing the restraint to run 3 alone loses that, and the write gate holding under a `probable` person-evidence score is worth having in the history.

Also confirming your f4 check matches my blind grade independently: `rel-jons-elena` byte-identical between the starting tree and run 2's final tree, no facts on it, and R3 links the Fulltofta pair rather than Jöns and Elena. `false` is right.

`Enna de mans Sohn` is at README line 223 with the "bride rendered as a son" gloss; I have also lifted the phrasing into the description, since that is what reaches `git log`.

Ready for your approval when you are.

EdmondOware (approved):
Approving. The description now says what the artefacts say, and one part of my correction was wrong in a way worth stating plainly.

I claimed run 1 never got far enough to write anything. It got all the way there. Run 1 made 197 tool calls, carries 34 `Fulltofta` mentions in its research document, a proof summary tiered `possible`, three `person_evidence` entries at `probable`, and a `resolved` question, with zero Fulltofta in the final tree. That is the write threshold holding under a `probable` person-evidence score, which is a better fact than the one I offered and belongs in the permanent record. You were right to push back rather than just apply the edit.

Checking the artefacts turned up one thing that sharpens it further. Run 3 has **zero** `Fulltofta` mentions anywhere, research document included, against run 1's 34. So the two runs that left the tree at its starting state did so for opposite reasons: run 1 reached the wrong candidate and declined to assert it, run 3 never reached it at all. Your table reports both as "3 persons, no wrong parentage", which is accurate, and the prose beneath it already draws the distinction for run 1. Worth one clause noting run 3 recovered nothing rather than restrained itself, since a later reader comparing only the table would read them as the same outcome.

Everything else verifies against the files rather than the summary: 197 / 140 / 168 tool calls, run 3's wall clock at 7223 s against the raised 7200 cap with `stop_reason: timeout`, run 2's `Assmen Nielsson` and `Bentte Jeppisdr` both present, `same_person` at zero in every run with run 2's compliance failing on it and run 3 passing, seven new `it(...)` blocks, and `Enna de mans Sohn` with the bride-rendered-as-a-son gloss at README line 223.

The caps argument is the part I would have pushed hardest on and it is now the strongest paragraph in the body: run 3 consuming the full raised budget is the case *for* the raise, not against it, and saying so beats leaving "changed nothing" standing.

Branch needs updating before merge.

PR #1922, opened 2026-08-26
feat: tier-2 reporting mechanism + Group B validators (#1749)

yinkid28:
## Summary

Implements the remaining work from the citation deep dive (issue #1749):
the **tier-2 reporting mechanism** (Group M) and the **Group B validators**
that use it. Group A (V1, V5, V10-literal) shipped in PR #1781.

- **Group M** — the plumbing: `report_*` functions fire alongside `test_*`,
  feed observations to the judge via `{harness_observations}`, record in
  `output.warnings[]` as `prose_observation`, and never gate the test outcome.
  Four new validator args (`activated`, `num_turns`, `output_tokens`,
  `aborted_reason`).
- **Group B** — the validators:
  - V8: activated run must produce response (tier 1, universal)
  - V7: in-body decline actually declines — 3 parts (tier 1 + tier 2)
  - V2: no unbacked validation claim (tier 2, universal)
  - V3: no invented sample locators — persisted + response (tier 1 + tier 2)
  - V4: skill's own example values not emitted — persisted + response (tier 1 + tier 2)
  - V10 report half: informant name from notes (tier 2)
  - V11: negative-search citation quotes log verbatim (tier 2)
  - V12: no framework walkthrough in response (tier 2)

**No CI job can detect judge bias from the new `{harness_observations}` prompt
section.** Three anti-bias constraints (from the issue) are built into the
design: (1) only fired findings appear, (2) pass `r.error` (observation text)
never `r.name` (verdict), (3) observation text states what was matched — no
dimension names, scores, or evaluative language.

No run-log schema change — observations go in `output.warnings[]` which has
`additionalProperties: true`. No snapshot change.

**Touches:** `eval/harness/harness/validator_runner.py`,
`eval/harness/harness/orchestrator.py`, `eval/harness/harness/judge.py`,
`eval/harness/judge/prompt.md`, `eval/harness/validators/conftest.py`,
`eval/harness/validators/test_universal.py`,
`eval/harness/validators/test_citation.py`,
`eval/harness/tests/unit/test_validator_runner.py`,
`eval/harness/tests/unit/test_judge.py`,
`eval/harness/tests/unit/test_orchestrator.py`,
`eval/harness/tests/unit/test_cli.py`,
`docs/specs/unit-test-spec.md`

## Test plan

- [x] `pytest eval/harness/tests/unit/test_validator_runner.py -v` — 123 passed
- [x] `pytest eval/harness/tests/unit/test_judge.py -v` — observations rendering tests pass, no-unsubstituted-slot test passes
- [x] `pytest eval/harness/tests/unit/ -v` — full harness unit suite: 2831 passed, 1 pre-existing failure (unrelated `temp_test_cli.py` encoding lint)
- [x] `pytest eval/harness/validators/ -v` — standalone validators: 6 passed, 189 skipped (expected on empty fixtures)
- [ ] Replay validators over committed citation run logs and verify hit list


promise-emmanuel (changes requested):
The tier-2 mechanism itself looks right — I replayed the reporting-tier validators over the 90 committed citation runs and V3's response half fired 8 times, all real fabrications. `Will Book 7, p. 214` on `ut_citation_016` is the exact thing that fixture's judge context calls fabrication. The plumbing works.

Four things before this can merge.

**1. CI is red.** `vitest` fails on the issue-reference ratchet in `doc-links.test.ts` — `unit-test-spec.md` is held at 4 issue refs and the new section makes it 5.

```
You wrote:      Validators are split into three tiers (issue #1749):
Change it to:   Validators are split into three tiers:
```

The three bullets underneath already say what the tiers do.

**2. V8 fails a test that passes today.** I replayed it over all 1945 committed unit runs. 343 of them report `num_turns: 0` and `output_tokens: 0` with no abort — zero telemetry is normal, not exceptional. Four of those also have `activated: True`, and `ut_citation_012` fails: it ran 13.7 seconds, invoked the citation skill, and replied with a 132-character routing announcement. Correct behaviour, currently `pass`, fails under the 200-char floor. The other three survive by 14, 45 and 137 characters.

All four have a non-empty `skills_invoked`. That's the direct signal and it's already in the argument pool. Adding it takes the corpus to 0 firings across 1945 runs.

```
You wrote:      def test_activated_run_produces_response(
                    activated, aborted_reason, num_turns, output_tokens, text_response, test,
                ):
Change it to:   def test_activated_run_produces_response(
                    activated, aborted_reason, num_turns, output_tokens, text_response, test,
                    skills_invoked,
                ):
```

```
You wrote:          if num_turns != 0 or output_tokens != 0:
                        return  # telemetry shows work happened
Change it to:       if skills_invoked:
                        return  # a run that invoked a skill is not a dead run
                    if num_turns != 0 or output_tokens != 0:
                        return  # telemetry shows work happened
```

This one is a tier-1 gate in `test_universal.py`, so it runs for every skill, not just citation.

**3. V4's deny-list misses the examples it exists to catch.** `_harvest_skill_examples()` returns exactly two values: `no. 4521` and `roll 810`. I fed the SKILL.md's other example locators through the gating validator as a persisted `where_within` and all three pass:

- `Will Book 9, p. 113` — SKILL.md line 322
- `Deed Book 41, pp. 88-90` — line 359
- `Page 12, entry 47` — line 73

Those are the probate and deed templates, which are the two the citation tests exercise most. All three scans need the value in quotes or after an `Example:` prefix, and the template blocks use neither. Scanning the whole body takes the list from 2 to 7 and catches all five — the caller already subtracts on-file values, so a wider harvest is safe.

```
You wrote:      examples = set()
                # Extract from fenced example blocks
                for block in re.finditer(r'(?:Example|e\.g\.|example)[:\s]*["\']?([^"\'\n]+)', text, re.IGNORECASE):
                    content = block.group(1).strip()
                    # Extract locator-shaped values from the example text
                    for m in _LOCATOR_RE.finditer(content):
                        examples.add(m.group(0).strip())
                # Extract from explicit parenthetical examples: (e.g. "Will Book 7, p. 214")
                for paren in re.finditer(r'\((?:e\.g\.|for example|such as)\s*["\']([^"\']+)["\']', text, re.IGNORECASE):
                    content = paren.group(1)
                    for m in _LOCATOR_RE.finditer(content):
                        examples.add(m.group(0).strip())
                # Also grab quoted illustrative values
                for quoted in re.finditer(r'"([^"]{5,50})"', text):
                    content = quoted.group(1)
                    for m in _LOCATOR_RE.finditer(content):
                        examples.add(m.group(0).strip())
                return examples
Change it to:   # Every locator-shaped literal in the skill body is illustrative. The
                # templates carry theirs outside quotes and outside an "Example:"
                # prefix (Will Book 9, p. 113; Deed Book 41, pp. 88-90), which the
                # three narrower scans missed. On-file values are subtracted by the
                # caller, so a wider harvest costs nothing.
                return {m.group(0).strip() for m in _LOCATOR_RE.finditer(text)}
```

**4. Two of the three anti-bias constraints have no test.** You're right that no CI job can measure judge bias. But two of the three constraints are ordinary assertions. I broke each and ran the full suite:

| constraint | mutation | result |
|---|---|---|
| tier-2 never gates | drop `not r.reporting_only` from `compute_validators_passed` | caught — `test_report_finding_does_not_gate` fails |
| pass `r.error`, never `r.name` | swap `r.error` → `r.name` in `harness_observations` | 2834 passed |
| only fired findings appear | include passing `report_*` results | 2834 passed |

The second one matters most. Swapping in the function name would ship `report_no_framework_walkthrough` to the judge as a bare verdict — which is the anchoring problem #1007, #1330 and #1603 diagnosed, and the reason the comment right above says a passing validator list must never be handed over. Nothing would stop that landing.

The reason neither is testable right now is that the three list comprehensions are inline in `_execute_single_run`, which a unit test can't reach. Lift them into a module-level helper and pin it:

```python
def test_judge_observations_carry_text_not_validator_names():
    """Anti-bias constraint 2 (#1749): the judge sees the observation text,
    never the function name — a name is a verdict."""
    results = [
        ValidatorRunResult(name="report_x", passed=False,
                           error="the response names a volume", reporting_only=True),
        ValidatorRunResult(name="report_ok", passed=True,
                           error=None, reporting_only=True),
    ]
    obs = split_observations(results)
    assert obs == ["the response names a volume"]
    assert not any("report_" in o for o in obs)
```

That covers the first constraint too — `report_ok` passed, so it must not appear.

**Two smaller things, your call:**

`report_invented_locators_response` adds numerals from the user's message to `on_file`, but `test_no_invented_locators_persisted` doesn't. So a user saying "the deed is in Deed Book 27, page 142" would get that flagged as invented when persisted and not flagged when spoken. No current test puts a locator in a user message, so it's latent — but the gating half is the wrong one to have the narrower rule.

`prose_observation` is classified harness-side in `test_cli.py:1490`, which keeps it out of `_JUDGE_WARNING_KINDS`, so it never prints in the run summary. Matches what the spec says, but it means the only way to see a tier-2 finding after a paid run is to grep the JSON. Worth deciding on purpose rather than by side effect.

---

Also worth doing: the unchecked box on your test plan — replaying the validators over the committed citation run logs — is what surfaced items 2 and 3. Worth ticking before the next push.

promise-emmanuel (approved):
Approving. I re-ran all four items from the last round rather than reading them —
they're genuinely fixed, not just present.

**V8** — replayed over the 1942 committed runs. The `skills_invoked` early-out
takes it from 1 firing to 0. `ut_citation_012` (the 132-char routing
announcement) no longer fires. 355 runs report zero telemetry, 4 of those
activated and not aborted, none fires now.

**V4** — the harvest goes from 2 values to 7. Across the 90 committed citation
runs the widening adds zero firings to the tier-1 gate, and one real catch in
tier 2: `ut_citation_014` emits "Deed Book 41, pp. 88-90" verbatim out of the
SKILL.md template, on a run that otherwise passed. That's exactly what V4 is for,
and it landed where it can't gate.

**Anti-bias pin** — mutated all three constraints. `r.error` -> `r.name` fails
`test_judge_observations_carry_text_not_validator_names`. Including passing
`report_*` results fails the same test. Dropping `not r.reporting_only` from
`compute_validators_passed` fails `test_report_finding_does_not_gate`. Real
coverage now.

I found one more thing and pushed the fix rather than sending it back, since it
was two lines — 10e3441b.

**Nothing would have noticed a `report_*` validator breaking.** Two gaps, same
root: tier 2 sat outside every check we have.

`python_functions = ["test_*"]` meant `pytest eval/harness/validators/ -v` — the
debugging path the spec points developers at — collected zero of the eight new
`report_*` functions.

And a `report_*` with a typo'd argument name was marked reporting_only, so it
didn't gate, `as_dicts` dropped it from the run log, and `split_observations`
handed the harness's own error text to the judge:

    run log:       report_typo is absent - tier 2 is filtered out
    judge prompt:  - validator declares unknown parameter(s): ['skil_frontmatter'].
                     Valid harness-supplied args are: ['aborted_reason', 'activated',
                     'after_state', ...]

That lands under "Harness observations on the response text", which tells the
judge these are pattern matches over Claude's response and to weigh each one. A
crash does the same with a raw TypeError. The tier-1 equivalent fails loudly; the
tier-2 one failed silently into a paid grading run.

All eight current signatures resolve, so nothing was broken — this is the missing
guard, not a bug.

The fix: a validator that is itself broken is a bug in the validator, not a
finding about the run, so the unknown-parameter and non-AssertionError paths now
gate whatever the prefix. The pass / assert / skip paths keep `reporting_only`
untouched — an AssertionError from a `report_*` is still a finding, still never
gates, still reaches the judge as anonymous text. Plus `report_*` in
`python_functions`.

Four tests, each mutation-tested. Reverting either `reporting_only` line fails
its own test and only that one. Reverting the pyproject line fails
`test_standalone_pytest_collects_report_validators`. And over-correcting — making
every `report_*` gate — fails `test_a_genuine_report_finding_still_reports` plus
five existing tier-2 tests, which is the direction I most wanted pinned: the
guard must not swallow the mechanism this PR exists to add.

Spec updated to match: the tier-2 bullet said "never gates the test outcome",
which is no longer true for a broken validator. Still 4 issue refs, doc-links
green.

Harness suite 2881 passed / 8 skipped. `check_runlogs.py`: all rules satisfied —
no re-run and no re-annotation, since the snapshot embeds skill-side files and
this is harness code, a pytest config, tests and a spec.

Two notes for whoever gives the second approval: please read 10e3441b in
particular, since it's mine and nobody has reviewed it. And the last test-plan
box is still unchecked — the V4 replay above is that box, worth ticking.

johnmarkpeterbrown (changes requested):
All four fixes are about six lines across two files. They belong in this PR — not an issue, not a follow-up.

Here's the author-facing copy, evidence stripped:

> Four things, all small. Three are missing boundaries in new regexes; one is a dead variable.
>
> **1. V4 fails correct citations.** The deny-list match has no digit boundary, so `pp. 88` matches inside `pp. 880-884`. Every harvested value does this to its next-digit sibling — `Will Book 9` fires on `Will Book 91`, `roll 810` on `roll 8104`. This is a tier-1 gate, so it fails the test and skips the judge. Widening the harvest from 2 values to 7 last round was right; it just made this reachable. Two places, same change:
>
> ```
> You wrote:                  if example_val.lower() in text_lower:
> Change it to:               if re.search(re.escape(example_val) + r'(?!\d)',
>                                         text_lower, re.IGNORECASE):
> ```
> ```
> You wrote:      if example_val.lower() in stripped.lower():
> Change it to:   if re.search(re.escape(example_val) + r'(?!\d)',
>                             stripped, re.IGNORECASE):
> ```
>
> **2. V4 throws away the on-file list it builds.** `on_file` is assigned and never read, so the subtraction only sees the source's own fields. A locator that legitimately lives in a log entry's notes counts as invented. V3 in the same file scans log notes, tree and the user message — the gating half is the narrower one.
>
> ```
> You wrote:      on_file = _extract_on_file_numerals(before_rj, before_state.get("tree_gedcomx_json"))
>                 before_text = ""
> Change it to:   before_text = ""
> ```
> and after the source loop:
> ```
> Add:            for entry in before_rj.get("log", []):
>                     before_text += " " + (entry.get("notes") or "")
> ```
>
> **3. V7(c) flags the correct decline.** No negation guard, so `I will not create a source entry` reads the same as `I'll create a source entry`. Negative tests are the only population this runs on.
>
> ```
> You wrote:          r".*?(?:add|create|extract|format)"
> Change it to:       r"(?:(?!\bnot\b|\bnever\b|\bcannot\b|n't).)*?(?:add|create|extract|format)"
> ```
>
> **4. V12 flags ordinary bullets.** The label can be followed by plain whitespace, so three normal sentences beginning "- Who created…", "- What it is…", "- Where it is held…" count as field headings.
>
> ```
> You wrote:          + r')(?:\*\*|:|\s)',
> Change it to:       + r')(?:\*\*|:)',
> ```
>
> Worth adding a test for V4 while you're in there — it's the only tier-1 validator in this PR without one, and it's the one holding both bugs.
>
> Separately: your last test-plan box is still unchecked. I ran it — V3-response, V4-response, V11 and V12 fire 0 times over the committed citation log; V2 fires 3 times and those are real. You can tick it.

johnmarkpeterbrown (approved):
>Approving.</strong> I re-ran each of the four items rather than reading the diff, plus the replay that was still unchecked on the test plan.</p>
<p><code>22382411</code> lands all four exactly as specified:</p>

item | what I ran | result
-- | -- | --
V4 gate false-failed on legitimate page ranges | fed it a pp. 880-884 citation as persisted citation + citation_detail.where_within | passes — no violation
V4 discarded its own on-file list | read the body; before_text = "" initialises before both loops, log-notes subtraction appended after the source loop | correct, no unbound local
V7(c) flagged the correct decline | four decline phrasings incl. I will not create a source entry and Let me be clear: I cannot extract assertions | all four pass; I'll add a source entry still fires
V12 flagged ordinary prose bullets | three bullets beginning - Who created… / - What it is… / - Where it is held… | passes — no observation


<p>The three new V4 tests are real coverage. I reverted the digit boundary and watched <code>test_v4_persisted_does_not_fire_on_digit_superstring</code> fail with <code>contains skill example value 'pp. 88'</code> before restoring. That was the gap that let the bug through in the first place, so it's the right place to have pinned it.</p>
<p>Suites on <code>d93b7cf3</code>: harness unit <strong>2917 passed / 9 skipped</strong>, <code>test_validator_runner.py</code> <strong>130 passed</strong>, standalone validators <strong>6 passed / 203 skipped</strong>. One failure, <code>test_mock_mcp.py::test_successful_call_carries_no_is_error</code>, cause <code>build not found: packages/engine/mcp-server/build/tools/research-query.js</code> — no compiled engine in my worktree, unrelated to this diff.</p>
<p><strong>Your last test-plan box</strong> — I ran the replay over <code>eval/runlogs/unit/citation/v1_2026-08-20_10-39-13.json</code> (18 tests, 14 non-negative runs). V3-response, V4-response, V11 and V12 fire <strong>0 times</strong>. V2 fires <strong>3 times</strong> and the tool ledgers confirm those are real: no run in that log called <code>validate_research_schema</code> at all. Narrower than Promise's 90-run sweep, but enough to tick the box.</p>
<p>One note for whoever merges: Promise's approval reads <code>84419537</code>, which predates <code>22382411</code>, and <code>dismiss_stale_reviews_on_push</code> is <code>false</code> — so it still counts toward the two but never read these fixes. This approval is the one that has.</p>
<p><strong>Two things I raised that are still open. Neither blocks, and I'd rather not hold a verified fix for them.</strong></p>
<p>The judge prompt section still describes the observations as response-text-only, but <code>report_informant_name_in_who</code> reports on persisted <code>citation_detail.who</code>:</p>
<pre><code>You wrote:      These are pattern matches the harness ran over Claude's text response. They are
Change it to:   These are pattern matches the harness ran over Claude's text response and over
                the files it wrote. They are
</code></pre>
<p>Take it or leave it — the section already tells the judge to ignore anything unrelated to the dimension it's scoring, so the cost of leaving it is small.</p>
<p>And <code>report_informant_name_in_who</code> and <code>report_negative_search_quotes_log</code> still have no tests. Both tier-2, so neither can fail a run, but they're the last two validators in this PR with nothing pinning them. Worth an issue rather than another round here.</p>
<hr>
<p>Two calls left to you: whether to include the one-line prompt diff at all (it invites another push, which would leave you re-reviewing your own approval), and whether you want the untested-tier-2 item as an issue — say the word and I'll draft it, though at two functions in one file it's close to the "fix it now or drop it" line rather than issue-sized.

PR #1925, opened 2026-08-26
research_append: tree-encoding completion gate — baseline + tree_diff + shadow measurement (#1490)

### LINKED ISSUE #1490: research_append: move the two completion gates into the tool (mentor gate, then tree-encoding)

jud-sdev:
## Summary

Closes #1490 

- Ships the **tree-encoding completion gate** for #1490 as a **warning** (per the 2026-08-24 no-override ruling): a write-once `starting-tree.gedcomx.json` baseline, an internal `treeDiff` diff helper (not an advertised tool), a fire-rate shadow measurement, and the gate itself in `research_append`.
- Phase 1 (mentor gate) already shipped in PR #1685 and is untouched; this is the tree-encoding half the ruling redesigned around a sibling-file baseline.
- Normal tier. One PR rather than the lead's suggested three (see *Deviated from the plan*).

## Review guidance

**Start here:** `research-append.ts::treeEncodingCompletionWarnings` and the `find_conclusions_without_tree_encoding` detector in `skill_invocation.py`. The thing to check is the **shape-match, not foreign-key** approximation: a proof summary carries no tree reference, so "the conclusion's person" is the union of persons its `supporting_assertion_ids` have `person_evidence` for. It warns only when NONE of those gained ANY structure vs the baseline — deliberately broad, because with no override a wrong refusal hard-blocks correct work. Decision needed from @DallanQ: whether to add an optional tree-ref field to `proof_summary` to make this a true FK (costs a schema-site change) — measurement is in hand below, so this is the "answer in a line" call you asked for.

**Unsure about:** nothing blocking. The fire rate re-measured to **3 of 143** tier-≥-probable conclusions (127 completed projects), far below the first-pass 32/22% — which carried the classifier artifact (parentage questions naming a birth date graded as birth). A subject-person-only variant reads 13. Either way well under any refusal bar, so it ships as a warning.

## Plan

**Didn't change:** the `same_person` write-boundary gate (out of scope, spec-first). `research/SKILL.md` prose (issue #1492 owns it). The electron `watcher.ts` file-watcher and `mock_agent.py` — the baseline is write-once so there's nothing to watch, and a mock project legitimately has no baseline (the gate fails open).

**Acceptance check:** `packages/engine/mcp-server/tests/tools/research-append.test.ts` → "warns on completion when a probable conclusion added no tree structure (#1490)" (fails on `main`, passes here); and `eval/harness/tests/unit/test_guardrail_shadow_report.py::test_replay_tree_encoding_fires_when_the_conclusion_added_no_structure`. Both watched failing by disabling the detector.

**Deviated from the plan:** built as one PR, not the lead's suggested three (developer's call). Two edit sites the plan's list missed, both folded in: the corpus census `WATCHED_PROJECT_FILES` (its own test caught it) and — privacy-critical — feedback-bundle redaction, which shipped the new baseline un-redacted and leaked living-person details.

**Write-lockdown decision:** the baseline **joins** `PROTECTED_PROJECT_FILES` (all three copies + the parity test), or an agent could overwrite it and defeat the gate.

**Threshold, recorded before the measurement ran:** refuse only if rate ≤10% AND a ≥20-case sample is 100% defensible; else warn. The measured 3/143 with a shape-match detector → warning.

## Test plan

- [x] Ran the offline suites (engine `vitest` + `tsc`, `eval/harness` pytest, `apps/server` pytest) — all green. (Ran individually rather than the single `make test-all` target; the one harness failure is the documented "engine build absent in a fresh worktree" case, not this diff.)
- [x] N/A — no skill run-log snapshot changed. The gate lives in the `research_append` **tool**, not a skill; no SKILL.md/agent/fixture edited. The issue confirms no paid eval run is required.
- [x] N/A — no skill `description` frontmatter or DO NOT clause changed.

## Follow-on issues

**Folded in / filed instead:** folded in — the census watch-list fix and the feedback-redaction fix, both found during this PR (the default: fix it in the current PR, context already loaded). Nothing filed. The `proof_summary` foreign-key option is a decision for @DallanQ in the review, not a filed issue.


promise-soar (dismissed):
Requesting changes — two things to fix, both small. Everything else looks good: I ran all four suites in a worktree off your head and they're green (vitest 2715/2715, harness 2832 passed, server 211 passed, tsc clean), all four copies of the protected-file list are updated, and I mutation-tested both of your acceptance checks — disable the detector and each one fails, so they're real.

**1. `tree_diff` can't see a marriage.**

`treeDiff` only reports relationships that were added or removed. There's no branch comparing facts on a relationship that's in both trees. A `Marriage` fact lives on the Couple relationship, so when the couple is already in the seeded tree and the session adds the marriage date, the diff comes back completely empty and the gate warns on correct work.

This isn't theoretical. Of your 4 measured fires, 3 are marriage questions, and `rossi-marriage/run-2026-07-16_13-31-25` is a clear false positive — the agent found the marriage, encoded `Marriage 2 Feb 1899 @ Hoboken, Hudson, New Jersey` on the existing couple, and concluded `proved`. The detector says it added no tree structure. I checked the other three and they're not caused by this, so the real rate is 3/158. Doesn't change warn-vs-refuse, but every field marriage on a pre-existing couple will warn falsely.

Please diff facts on relationships present in both trees, and add their endpoints to `personsWithNewStructure`. Same fix needed in `_person_gained_structure` in `skill_invocation.py`, which has the identical gap — I confirmed it fires on the same synthetic case.

Three doc lines need to follow:

You wrote (`tree-diff.ts` header):
`//   - A `Marriage`/`Divorce` fact lives on the Couple relationship's `facts[]`,`
`//     not on a person, so relationship facts are diffed too.`

Change it to:
`//   - A `Marriage`/`Divorce` fact lives on the Couple relationship's `facts[]`,`
`//     not on a person, so relationship facts are diffed on both added and`
`//     already-present relationships.`

You wrote (`tree-diff-tool-spec.md`):
`- It does not diff relationship *facts* beyond the relationship's add/remove — a`
`  Marriage date corrected on an already-present Couple is not reported as a change.`
`  No caller needs that today; add it when one does.`

Change it to:
`- It does not diff `sources[]`, names, gender, or `ark`.`

(drop the bullet — the tree-encoding gate in this PR is the caller that needs it)

And please add a `tree-diff.test.ts` case for a fact added to an existing relationship. Right now the only relationship-fact test adds the whole relationship, so the suite mirrors the blind spot.

**2. The redaction count can lie.**

In `_redact_living`, swapping `redacted = 0` for `pass` assumes the failure always happens before any person is counted. It doesn't — the per-person loop is inside the same `try`, and `_redact_person` raises on a malformed `names` entry. I ran it: a tree whose second person has `"names": ["Bob Smith"]` returns `redacted = 1` while the file ships untouched, with the first person's given name and her birth date in the clear. FEEDBACK.md then tells the submitter one living record was protected.

You wrote:
`    out: list[tuple[str, bytes]] = []`
`    redacted = 0`
`    for rel, data in files:`

Change it to:
`    out: list[tuple[str, bytes]] = []`
`    redacted = 0`
`    for rel, data in files:`
`        file_redacted = 0`

then use `file_redacted += 1` in the loop, `redacted += file_redacted` as the last line of the `try`, and leave the `except` as `pass`. That way a file that fails partway contributes nothing, which is what the comment already says should happen.

**3. Three docs now name a blocker you removed.**

`docs/architecture.md` line 283, `ADR-0003` and `ADR-0011` all still say the tree-encoding gate "stays prose pending a baseline of the project's opening tree to diff against". This PR ships that baseline. Your first commit wrote those lines correctly and the later commits falsified them. Please update all three to say the baseline shipped and the check is warn-only per the 2026-08-24 ruling.

Nothing else from me. The site-list work is thorough — I checked and all four copies of `PROTECTED_PROJECT_FILES`/`WATCHED_PROJECT_FILES` are in sync, and `sandbox_server.py` correctly doesn't broadcast the new file to the viewer.

promise-emmanuel (changes requested):
Requesting changes — three things, all small, all fixable here. Everything else is solid: I ran all four suites in a worktree off your head and they're green (engine vitest 2732/2732 + tsc clean, harness 2874 passed / 2 skipped, server 213 passed), and I mutation-tested four checks, not just the two you named — both acceptance tests fail when the detector is disabled, and both of the marriage regression tests from the last round fail when the fact-gain credit is removed. The previous round's three items are properly closed.

**1. The desktop feedback bundler still leaks the baseline. Same bug, other copy.**

You fixed `apps/server/app/feedback.py`. `apps/electron/src/main/feedback.ts` is its mirror — its own comment says so — and it still redacts one filename. `walkProject` selects every non-media file, so `starting-tree.gedcomx.json` goes into the zip untouched.

I ran it. A project with one living person, both tree files present:

    ZIP ENTRIES: ["research.json","starting-tree.gedcomx.json","tree.gedcomx.json","FEEDBACK.md",...]
    BASELINE CONTAINS "Jane Marie":    true
    BASELINE CONTAINS the ark:         true
    BASELINE CONTAINS "Riverside, CA": true
    LIVE TREE CONTAINS "Jane Marie":   false

So one zip holds a redacted tree, an unredacted copy of the same people, and a FEEDBACK.md telling the submitter their living relatives were protected.

You wrote:

    const TREE_FILENAME = 'tree.gedcomx.json'

Change it to:

    const TREE_FILENAME = 'tree.gedcomx.json'
    const STARTING_TREE_FILENAME = 'starting-tree.gedcomx.json'
    // Both tree-shaped files a project folder can hold. The baseline carries the
    // same living persons and is bundled by the same walk (issue #1490).
    const REDACTED_TREE_FILENAMES = [TREE_FILENAME, STARTING_TREE_FILENAME]

You wrote:

    function redactLivingPersons(selected: { relativePath: string; buf: Buffer }[]): number {
      const entry = selected.find((s) => s.relativePath === TREE_FILENAME)
      if (!entry) return 0
      try {

Change it to:

    function redactLivingPersons(selected: { relativePath: string; buf: Buffer }[]): number {
      let total = 0
      for (const name of REDACTED_TREE_FILENAMES) {
        const entry = selected.find((s) => s.relativePath === name)
        if (entry) total += redactOneTree(entry)
      }
      return total
    }

    function redactOneTree(entry: { relativePath: string; buf: Buffer }): number {
      try {

(the rest of the existing body is unchanged)

And the FEEDBACK.md line, so it matches what the server now says.

You wrote:

    `${redactedLiving} person(s) in \`${TREE_FILENAME}\` are living or not marked ` +

Change it to:

    `${redactedLiving} living-person record(s) across the project's tree files ` +
    `(\`${TREE_FILENAME}\` and, when present, \`${STARTING_TREE_FILENAME}\`) are living or not marked ` +

Please add the Electron mirror of the test you added to `apps/server/tests/test_feedback.py` — a project with both tree files, asserting the baseline in the zip has no given name and no ark.

**2. The tree-encoding shadow detector never runs on a live run, so §11.5 reads 0 forever.**

`find_conclusions_without_tree_encoding` is only ever called from `replay_post_hoc`. Nothing writes `TREE_ENCODING_KIND` into a run's `guardrail_shadow_violations`, so `scan_tree_encoding` cannot return anything. The default report I ran:

    §7 warnings-unchecked check (issue #1193, shadow): 2 run(s) ...
    §11.5 tree-encoding check (issue #1490, shadow): 0 ... across 0 run(s).

The warnings-unchecked sibling reads 2 because it IS wired, at orchestrator.py line 2487 — and that site already has `final_research`, `final_tree` and `starting_tree` in scope, which is exactly what your detector takes. This matters beyond tidiness: the fire rate is the calibration signal the no-override ruling wants before warn ever becomes refuse, and as it stands it will read a hard zero next month and the month after, which is indistinguishable from "the gate never fires."

You wrote (orchestrator.py, the import block at line 55):

    find_relationship_writes_without_warnings_check,

Change it to:

    find_conclusions_without_tree_encoding,
    find_relationship_writes_without_warnings_check,

You wrote (after the `tree_citation_shadow` block):

        # `wall_clock_seconds` is the ACTIVE wall-clock (time.monotonic), so it

Change it to:

        # The §11.5 tree-encoding arm (issue #1490). Wired here for the same
        # reason as the tree-citation arm above: this is the only site with
        # `final_research`, `final_tree` and `starting_tree` together. Shadow
        # only; never feeds compliance or outcome.
        tree_encoding_shadow = find_conclusions_without_tree_encoding(
            final_research, final_tree, starting_tree=starting_tree
        )
        if tree_encoding_shadow:
            guardrail_shadow_violations = (
                guardrail_shadow_violations + tree_encoding_shadow
            )

        # `wall_clock_seconds` is the ACTIVE wall-clock (time.monotonic), so it

And the docstring that currently promises the opposite:

You wrote (`guardrail_shadow_report.py`, `scan_tree_encoding`):

    `kind == TREE_ENCODING_KIND`. Reads 0 over every run made before this check
    shipped — the whole corpus today — so REPLAY is where the signal lives.

Change it to:

    `kind == TREE_ENCODING_KIND`. Reads 0 over every run made before this check
    shipped; runs made after it store the check live, so this line accumulates
    the fire rate the no-override ruling asks a shipped gate to produce.

**3. `tree_forget` doesn't move the baseline, so forget-and-rederive warns on correct work.**

`tree_forget` writes `tree.gedcomx.json` and the dot-prefixed restore file. The baseline keeps the facts it stripped. So when the agent does exactly what the exercise asks — re-derives the birth from a record and writes it back — the fact matches the baseline by content and counts as nothing. I ran your own `treeDiff` on it:

    personsWithNewStructure: []
    personsChanged: []

Warning-only, so nobody is blocked. But it fires on every completion of the forget-and-rederive workflow, and those false fires land in the rate that decides whether this graduates to a refusal — at which point the practice workflow is hard-blocked with no override.

The post-forget tree IS the exercise's opening tree, so I'd mirror the strip into the baseline.

You wrote:

        await atomicWriteJson(join(projectPath, "tree.gedcomx.json"), tree);
        result.filesWritten = ["tree.gedcomx.json"];

Change it to:

        await atomicWriteJson(join(projectPath, "tree.gedcomx.json"), tree);
        result.filesWritten = ["tree.gedcomx.json"];
        // The post-forget tree is the exercise's real opening tree, so the
        // completion gate's baseline moves with it. Without this, a correctly
        // re-derived fact matches the pre-forget baseline and reads as no new
        // structure, and the gate warns on the exercise working (#1490).
        const baselinePath = join(projectPath, "starting-tree.gedcomx.json");
        if (await fileExists(baselinePath)) {
          await atomicWriteJson(baselinePath, tree);
          result.filesWritten.push("starting-tree.gedcomx.json");
        }

Deleting the baseline instead would also work — the gate fails open — but it gives up the check for the one workflow where it should be most informative. Your call; either needs a line in `tree-forget-tool-spec.md` and a test.

**Two things not blocking.**

`tree_diff` is advertised as tool #49, but `before`/`after` are whole trees passed inline, so a model calling it has to reconstruct both from context — which produces a confident wrong diff. No skill or agent grants it, and the spec's other consumer (viewer / `project_status`) is in-process TS that would import `treeDiff` directly. Worth either dropping it from `allToolSchemas`/`manifest.json` and keeping the export, or giving it a `projectPath` so a call means something. Say which you intend and I won't push further.

And the PR body still says "4 of 143" where the spec says "3 of 143" — the spec is right after the marriage fix. Worth correcting the body, since that's the line people read.

One note on the measurement: I could not reproduce 3/143. The committed corpus only carries final-state sidecars for 7 runs, so `--replay` gives 1 of 7 here. That's fine for warn-only, but once fix 2 lands the corpus starts producing a reproducible number, which is the point.

The site-list work is genuinely thorough — all three `PROTECTED_PROJECT_FILES` copies plus the census are in sync, `context_policy.py` picks it up through `_guard`, and both viewers correctly ignore the new file rather than broadcasting it.

jud-sdev:
@promise-emmanuel thanks — all three fixed, and reconciled onto the latest branch head.

The Electron feedback bundler now redacts `starting-tree.gedcomx.json` alongside `tree.gedcomx.json` through a shared `redactOneTree` helper, and the FEEDBACK.md note names both files. Added the mirror test: both tree files present, and it checks the baseline in the zip has no living given name and no ark.

The tree-encoding detector is now wired into the live orchestrator next to the citation arm, so the stored fire rate accumulates instead of only showing up under replay. Fixed the scan docstring to match.

`tree_forget` now rewinds an existing baseline to the forgotten tree, so a correctly re-derived fact registers as new structure instead of the gate warning on the exercise working. It never creates a baseline where there isn't one. Added a spec line and two tests.

On `tree_diff`: I dropped it from `allToolSchemas` and `manifest.json` and kept the `treeDiff` export for the in-process caller, so a model can't call it with two whole trees inline anymore. Also corrected the PR body from 4 to 3 to match the spec after the marriage fix.

Ran the suites on the rebased tree: engine vitest 2761/2761 + tsc clean, electron feedback 28/28 + tsc clean, harness detector tests 227/227.

promise-emmanuel (dismissed):
...

promise-emmanuel (changes requested):
Re-reviewed against the merge with main (branch is level with main). All three items from my last round and all three from promise-soar's are properly closed — I checked each one and mutation-tested the acceptance test: disable the detector and "warns on completion when a probable conclusion added no tree structure (#1490)" goes red. Suites on your head: engine vitest 2761/2761 and tsc clean, harness pytest 2979 passed / 3 skipped, apps/server 217 passed / 1 skipped, electron feedback 28/28.

One thing before this merges.

The gate cannot see a fact being filled in.

factsNotIn compares facts with factsEquivalent, which is the merge tool's dedupe predicate — it answers "could these two be the same fact?" and returns true whenever either side's date or place is missing. So a seeded bare Birth and a Birth this session dated and placed match, and the session's work reads as nothing.

I ran treeDiff on three shapes. personsWithNewStructure is empty for all three:


before Birth (no date, no place) -> after Birth 3 Mar 1850, Cochabamba   []
before Birth, Cochabamba         -> after Birth +1850-03-03, Cochabamba  []
before Birth +1850, Bolivia      -> after Birth +1850-03-03, Cochabamba  []
control: adds a separate Death fact                                      ["I1"]
Only the control registers. Each of the other three is a completed question warning on correct work — and filling in a date or narrowing a place is the most ordinary thing a research session does.

The part that makes this worth fixing now rather than later: the shadow detector uses exact (type, standard_date, standard_place) signatures and does not flag the same case. I ran it and it returns 0 violations. So the "3 of 143" you recorded in research-append-tool-spec.md came from a strictly more lenient check than the one that ships, and that figure is exactly what the no-override ruling wants before warn can become refuse.

Your spec already says what the code should do: "Facts key on a content signature (type + date + place + value)." The code uses a compatibility match instead.

You wrote:


import { relationshipKey, factsEquivalent } from "../utils/merge-gedcomx.js";
Change it to:


import { relationshipKey } from "../utils/merge-gedcomx.js";
You wrote:


/** Facts in `a` not matched (by content) by any fact in `b`. */
function factsNotIn(a: SimplifiedFact[], b: SimplifiedFact[]): SimplifiedFact[] {
  const used = new Array(b.length).fill(false);
  const out: SimplifiedFact[] = [];
  for (const fa of a) {
    let matched = false;
    for (let i = 0; i < b.length; i++) {
      if (!used[i] && factsEquivalent(fa, b[i])) {
        used[i] = true;
        matched = true;
        break;
      }
    }
    if (!matched) out.push(fa);
  }
  return out;
}
Change it to:


/** A fact's content signature. NOT `factsEquivalent` from the merge tool: that
 *  is a DEDUPE predicate ("could these two be the same fact?") and returns true
 *  whenever either side's date or place is absent, so a seeded bare `Birth` and
 *  a `Birth` this session dated and placed match, and the encoding the gate
 *  exists to find reads as no change. Mirrors the shadow measurement's
 *  `_fact_signatures` so the gate and the fire rate it is calibrated against
 *  read the same facts. */
function factSignature(f: SimplifiedFact): string {
  return [
    f.type ?? "",
    f.standard_date ?? "",
    f.date ?? "",
    f.standard_place ?? "",
    f.place ?? "",
    f.value ?? "",
  ].join("\u0000");
}

/** Facts in `a` whose signature is not matched by a fact in `b`, respecting
 *  multiplicity. */
function factsNotIn(a: SimplifiedFact[], b: SimplifiedFact[]): SimplifiedFact[] {
  const available = new Map<string, number>();
  for (const fb of b) {
    const k = factSignature(fb);
    available.set(k, (available.get(k) ?? 0) + 1);
  }
  const out: SimplifiedFact[] = [];
  for (const fa of a) {
    const k = factSignature(fa);
    const n = available.get(k) ?? 0;
    if (n > 0) available.set(k, n - 1);
    else out.push(fa);
  }
  return out;
}
The raw date/place are in the signature on purpose — your existing test "treats a corrected fact date as a change" uses unstandardized dates and goes red without them. I applied exactly this and the whole engine suite passes, 2761/2761.

Please also add a tree-diff.test.ts case where a seeded Birth with no date gains one, and re-run make e2e-guardrail-shadow REPLAY=1 SINCE=all from this branch's worktree so the figure in research-append-tool-spec.md describes the code that ships.

Not blocking: the PR summary still calls tree_diff "a new MCP tool" — it isn't one any more, and that line becomes the merge commit.

jud-sdev:
Thanks — fixed. `factsNotIn` now keys on a content signature instead of `factsEquivalent`, so a seeded bare fact gaining a date or place reads as new structure rather than a false warning. Added the tree-diff case (bare Birth gains a date). Re-ran `make e2e-guardrail-shadow REPLAY=1 SINCE=all` from the branch worktree: 3 conclusions in 3 of 158 runs, and I updated the spec figure to match the shipped code. Also merged main (the one import conflict) and fixed the "new MCP tool" line in the summary. Engine suite green at 2783/2783.

One small deviation worth flagging: the signature is a JSON array of `type`, standardized and raw `date`/`place`, and `value` rather than a NUL-joined string — same fields you prescribed, collision-proof, and it keeps the source plain ASCII. It's a strict superset of the Python shadow's `(type, standard_date, standard_place)`, so on standardized production data the gate and the fire rate read the same facts.

promise-emmanuel (approved):
Approving. Re-reviewed 94be2bff against main (branch is level with main).

The fact-signature fix is right, and I checked it rather than read it. The three shapes that previously read as no new structure now all register ["I1"] — a bare Birth gaining a date and place, a Birth with a place gaining a date, and a place narrowed to the town. The control matters as much: an unchanged fact still returns [], so the fix didn't quietly turn the gate off by making everything count.

I mutation-tested your new test — reverted factsNotIn to factsEquivalent and "counts a seeded Birth gaining a date as new structure" goes red, 1 failed / 228 passed. And the re-measured figure reproduces here: tree-encoding 3 tier->=-probable conclusion(s) ... across 3 run(s), of 158 scanned, which is what the spec now says. (My "could not reproduce" note last round was my own environment, not your measurement.)

One thing worth recording: the TS signature also keys on raw date/place, which the Python arm doesn't, so the gate now warns strictly less often than the 3/158 it was measured at. That makes the figure an upper bound rather than a floor — the right direction for a check with no override.

I also checked the merge of main didn't drop anything, since main moved a long way under you: all four protected/watched file lists, both workspace seeds and the live orchestrator wiring are intact.

Suites here: engine vitest 2783/2783 and tsc clean, harness pytest 3012 passed / 3 skipped, apps/server 217 passed / 1 skipped, electron feedback 28/28.

Two things for the body before you squash, neither blocking. Unsure about still says "3 of 143 tier-≥-probable conclusions (127 completed projects)" — that is the pre-fix figure, and the body becomes the merge commit; the spec's "3 in 3 of 158 runs" is the live one. And Didn't change justifies leaving watcher.ts alone with "the baseline is write-once so there's nothing to watch", which stopped being true when tree_forget gained the rewind — the conclusion still holds, just for a different reason.

EdmondOware (approved):
Approved.

The write-once baseline is correctly protected everywhere it needs to be. All three `PROTECTED_PROJECT_FILES` copies carry it and the parity test covers them, so an agent cannot overwrite the thing the gate measures against.

The redaction work is the strongest part, and it is worth being precise that no bundle ever shipped un-redacted: `starting-tree.gedcomx.json` was an e2e fixture name until this PR made it a project file, so the gap was opened and closed in the same change. The per-file tally in `_redact_living` is a real fix beyond the rename, since the old `redacted = 0` in the except reset the running total and would have zeroed a successful file's count once a second tree file existed. The TypeScript twin has the same property.

The gate does what it claims and is load-bearing: disabling `treeEncodingCompletionWarnings` turns the acceptance test red. The broad shape match is the right bias given no override exists.

One note on the measurement, not blocking. Every denominator the report prints is run level, including tree-encoding's `of N scanned` and its skipped list. The unclassifiable case is conclusion level: a proof summary whose supporting assertions have no `person_evidence` is counted and can never fire. So "3 of 143" is a ratio whose denominator includes conclusions the detector cannot judge, and nothing printed says how many those are. It does not change the ship decision, since a smaller denominator only raises the rate and pushes further toward warning. It matters for whoever revisits promoting this to a refusal, who will read 2% and need the evaluable count instead.

Issue #1488, opened 2026-08-09
The eval/app scenario viewer's forked schema.ts has drifted from the master schema


DallanQ:
> **DECIDED 2026-09-16 — Option A: `eval/app` joins the pnpm workspace.** Lead ruling on
> the `#1488 fork` question below. The fork is **eliminated**, not guarded and not
> resynced: `eval/app` becomes a workspace member, imports `@genealogy/schema`, and
> `components/scenario/lib/schema.ts` stops carrying its own copy of the types. ADR-0008
> is **amended in place** — this is its own "Revisit when" trigger firing, and it reverses
> the evidence its "One shared TS module" row cites. `needs-decision` removed; the card
> stays `senior`.

**Touches:** pnpm-workspace.yaml, eval/app/package.json, eval/app/package-lock.json (deleted), eval/app/next.config.mjs, eval/app/components/scenario/lib/schema.ts, eval/app/components/scenario/components/sections/, package.json (root, `pnpm.overrides`), Makefile, scripts/test.sh, eval/Start.bat, .github/workflows/js-tests.yml, docs/adrs/ADR-0008-sync-schema-copies-eliminate-generate-or-lint.md, docs/architecture.md

## What is wrong

`eval/app/components/scenario/lib/schema.ts` (448 lines as of 2026-09-16) is a **hand-maintained
fork** of the `research.json` + simplified-GedcomX types. It is the scenario
viewer's copy of what `packages/schema/src/index.ts` and
`packages/schema/src/enums.generated.ts` hold for the web/Electron side.

**Nothing diffs it against the master schema**, and its own typecheck is
path-gated on itself:

- `eval/app` is **not a pnpm workspace member**. `pnpm-workspace.yaml` lists
  `packages/*`, `!packages/engine/**`, `apps/web`, `apps/electron` — `eval/app`
  appears nowhere, and it has its own `eval/app/package-lock.json`. So
  `make typecheck` (turbo over the pnpm workspace) never sees it.
- CI's `eval-app` job (`.github/workflows/js-tests.yml:114-161`) *does* run
  `npm test` and `npm run typecheck` — but only `if steps.scope.outputs.relevant == 'true'`,
  where the scope filter is `PATTERNS: '^eval/app/|^\.github/workflows/js-tests\.yml$'`
  (`:123`). **A PR that edits `docs/specs/schemas/enums.schema.json` and nothing
  under `eval/app/` skips the job entirely** and prints
[...]

PR #1926, opened 2026-08-26
fix(schema): schema-optional fields are TypeScript-optional (#1165)

clack391:
## Summary

**Tier: Normal.** Closes #1165 (the types half; the issue was reopened for exactly this after the names half shipped in #1271).

- Per the lead ruling of 2026-08-21 (in the issue body): a field absent from a `$def`'s `required` list is `foo?: T | null`, not required-but-nullable. Flipped the 25 such fields in `packages/schema/src/index.ts` (added `?`, kept `| null`).
- Added a both-directions optionality check to `schema-interface-drift.test.ts` (schema `required` vs the TS `?`), with no exemption list, over research and tree `$defs` and both document roots.
- Recorded the decision and the rejected present-but-null alternative in ADR-0008, and narrowed the doc copies that said "optionality is unchecked / deferred to #1165."

## Review guidance

**Start here:** the new `expectOptionality` helper and the `questionToken` capture in `interfaceFields()` in `packages/viewer-ui/src/__tests__/schema-interface-drift.test.ts`. The decision to check is the lead's; what is worth checking is that the assertion covers both directions (a schema-optional field missing its `?`, and a schema-required field wearing a spurious `?`), that it has no exemption list, and that the non-vacuity guard actually fails if the `questionToken` read breaks.

**Unsure about:** nothing outstanding. The doctrine was ruled by the lead; this executes it.

## Plan

**Didn't change:** the enum-binding axis and the nullability axis. The issue re-measured both to zero on 2026-08-21 (enums are generated since #1810; nullability clean across all comparisons), and I confirmed they are still zero, so touching them would invent aliases for the `iso_date`/`iso_datetime` patterns and close the open `*_recommended` enums, the inverse of ADR-0008. Each flipped field keeps its `| null`. The second TS mirror `eval/app/components/scenario/lib/schema.ts` (its own dependency island, #1488) is out of scope and untouched.

**Acceptance check:** `packages/viewer-ui/src/__tests__/schema-interface-drift.test.ts > @genealogy/schema interfaces mirror research.schema.json > Assertion optionality ↔ $defs.assertion.required` fails on a tree where any of the 25 loses its `?`, and passes here.

**Deviated from the plan:** nothing re-planned. Two self-review passes ran before requesting review and their findings were folded in: plan-critic added `docs/architecture.md`'s register row and named the two read sites; a blind open-break pass caught that the lint guarded the tree root but not the research root, which is now fixed and break-tested.

**Step-4 note:** this touches the schema TS mirror and an ADR. ADR-0008 is extended (a new `Alternatives considered` row, no row deleted), not reversed, and the schema-optionality doctrine is the lead's own 2026-08-21 ruling in the issue, so this executes a lead decision rather than making one. A `.ts` change needs a senior code-owner approval (either senior team satisfies it).

### The new lint is proven to fail (CLAUDE.md)

Removing the `?` from a schema-optional field (here `Assertion.date`):

```
× @genealogy/schema interfaces mirror research.schema.json > Assertion optionality ↔ $defs.assertion.required
    AssertionError: ... expected { missingQuestion: [ 'date' ], ... } to deeply equal { missingQuestion: [], ... }
```

The other direction fires too (adding `?` to a required field yields `spuriousQuestion: ['id']`), and adding a `?` to a required root field like `ResearchData.questions` reddens the research-root check (`spuriousQuestion: ['questions']`).

## Test plan

- [x] I ran `make test-all` on the final tree (`b53f1021`, nothing uncommitted, tree stable throughout), clean, exit 0, nothing failing: typecheck 5/5 tasks, engine 2702, viewer-ui 271, electron 97, web 10, eval CRUD UI 168, server pytest 209 passed 1 skipped, harness pytest 2812 passed 2 skipped. Forced (`--force`) runs, since turbo otherwise serves a cached result and a cache hit is not a run.
- [ ] ~Ran `make eval-skill`~ N/A: no skill dir, delegated agent, `eval/tests/unit/`, fixture, scenario, or run-log snapshot is touched.
- [ ] ~Refreshed routing negatives~ N/A: no skill `description` or DO-NOT clause changed.

### Second self-review round, folded in

Three blind passes (re-derive, open-break, completeness) ran against this head. What they found, all fixed here:

- **ADR-0008's `Enforcement` block still said "Field **names** only"** about the very test this PR extends, twenty lines from the paragraph the diff already fixed. My own doc-consistency grep missed it: the phrase is split by markdown emphasis *and* a line wrap, so `grep -c "names only"` on that file returns 0. Fixed, and the sweep now uses an emphasis-and-wrap-tolerant match.
- **The non-vacuity guard proved the parser, not the check.** Neutering `expectOptionality`'s shared-field filter took comparisons from 251 to 0 with all 67 tests still green; `expectMirrors` had the identical hole. Both now feed a shared comparison counter asserted in a final `describe`, break-tested in both directions.
- **`CLAUDE.md`'s new-field recipe named the interface but not how to type the field**, which after this PR is a red build in either direction. One clause added, deliberately with no issue ref (that file's ratchet is exact-match).
- Smaller: the coverage claim is now bounded to the `$defs` and the two document roots in all three places that state it, since interfaces mirroring inline `items` objects are reached by neither half of the lint; the ADR's routing line gained a trigger; `index.ts`'s header now records the optionality rule beside the enum rule; the `?? []` fallback for `researcher_profile` is documented; the stale `index.ts:269` cite is gone, which required deleting its now-dead entry from `doc-links.test.ts`'s grandfathered list.

### Third round: the guard I added in round two did not guard

Three blind passes plus my own break-tests found that the comparison counter from the previous round measured keys **enumerated**, not keys **compared**. Replacing all four filter bodies with `[]` left it reading its full 251 while both halves of the lint were inert, and a planted real drift stayed green. The floor was defeatable a second way: `if (tsName.startsWith('Gedcomx')) return` cleared 150 while hiding a genuine regression, and 8 of the 22 research `$defs` could be exempted while still clearing it.

Replaced with two instruments that have no floor: a **planted-drift self-test** that hands each helper a deliberately drifted schema and requires it to throw, and an exact **compared-set** assertion recording each interface at the point of comparison against a population derived from the schema, deliberately not from `NO_INTERFACE`. `NO_INTERFACE` is pinned empty, since the ruling forbids an exemption list and the loop's own failure message recommends one by name.

Watched fail, six ways: exemption before the compare, exemption after the bookkeeping, `NO_INTERFACE` grown, a `TREE_INTERFACES` entry dropped, all four comparisons gutted, one direction gutted alone. **My first two attempts at this fix let the middle two through**, which is worth stating: attempt two derived its expected set from the exemption list, the same tautology as the counter it replaced.

Two false claims of mine, corrected. The `foo?: T | null` rule I put in `CLAUDE.md` and this file's header is wrong: 58 of the 104 optional fields correctly carry no `| null`, because optionality follows `required` and nullability follows the schema's own type, and those are independent axes. And a docblock justified a test with a `"0 persons"` string that exists nowhere and a `?? 0` that would not compile on a `string[]`.

Also: `ProjectOverview`'s test could not tell a correct guard from one that always takes its branch (`{true ? (` left all 264 green), so it gained a positive case; the five card-expansion helpers that had drifted into four mechanisms are now one shared module; and the ADR's disputed key count is replaced by the reproduction method, since three independent counts of the same corpus returned 17, 19 and 19.

### A docblock correction, from re-reading two issues I had closed

The harness flags a cited issue that is closed with "verify it is not cited as a live reason". I recorded that as checked without opening either body. Pulling them found three false claims of mine still standing, because I had corrected them in **closing comments** and a closing comment never reaches a reader of the body.

One of the three had never been verified by anyone: #1928 said a tsc-clean `?? []` variant "renders the wrong thing and is caught". It does not. A genuine `?? []` passes all five `ProjectOverview` tests, because it takes the same branch as the correct guard and so is a harmless equivalent rather than a wrong fix. My original break-test had used `?? ["x"]`, a **non-empty** array, which is a different mutation entirely.

That claim had also propagated into this PR's own docblock, where it then read as self-contradictory. `b53f1021` rewrites it to say what the tests actually catch: the discriminating mutation is `{true ? (`, a guard that always takes one branch, which left the whole viewer-ui suite green until the positive case existed. Both issue bodies were corrected in place rather than by another comment.

**Folded in, not filed (#1927, #1928 closed):** I first filed both and was wrong to. They were one root, the absent `match_score` case at a PersonEvidence read site, once per copy, and neither carried the overhead of a ticket. The `eval/app` render sites are byte-identical twins of the two fixed here and are now `!= null`; `PersonEvidenceSection` gained the test file it was the only section component of eleven to lack. Two corrections on the record: the CI reason I gave for filing was backwards, since the eval-app job keys on the PR's own changed files so touching `eval/app/` arms it, and "renders NaN%" needed a qualifier, since `Card` starts collapsed so those 1,539 entries show it on expand rather than on load. Still out of scope and untouched: `eval/app/.../lib/schema.ts`, the type mirror #1488 owns.

**Previously filed instead:** ~~#1927, `eval/app`'s scenario viewer renders `Match: NaN%` for 1,539 of 7,392 committed `person_evidence` entries, because its forked mirror still declares `match_score: number | null` so `tsc` cannot see the absent-key case. Byte-identical twins of the two lines fixed here. Not folded in: #1165 rules that island out of scope by name, and its CI job would not have run on this PR. Doctrine recorded on #1488 as well, so the fork is not reconciled to the rejected encoding.~~

**Folded in / filed instead:** Folded in, all within this PR: plan-critic's `docs/architecture.md` register-row narrowing and the two read-site fixes, and the open-break pass's research-root optionality check. Filed instead: none. Named and stopped, both pre-existing and out of scope: the `eval/app` schema mirror (#1488), and the three inline nested locality objects (`LocalityPageRead`, `LocalityJurisdiction`, `LocalityCollection`), which are a pre-existing blind spot in the name check too (they are inline array-item schemas, not top-level `$defs`, so neither loop reaches them) and are correct today.

### Fourth round: the coverage guard could be satisfied without the comparison

Francis's approving review left a non-blocking note on the `every intended interface was actually compared` block. Following it up found something sharper than the note described.

Both helpers recorded the interface name **before** their own `expect`. The planted-drift self-test calls those same helpers with `Assertion` and requires them to throw, so `Assertion` was entered into the compared-set by the self-test rather than by any real comparison. Deleting the real `Assertion` comparison from the `$defs` loop left the entire file green. The two instruments described in the round above are not independent: one was feeding the other.

The fix is two lines. Each `seen.*.add(tsName)` now runs **after** its `expect`, so a throw or an early return never registers a name.

Watched fail, on the committed code, guard and suite run separately:

| mutation | before | after |
|---|---|---|
| skip the real `Assertion` comparison | green | red |
| skip a `TREE_INTERFACES` entry | green | red |
| `startsWith('Gedcomx')` bail inside the helper | red | red |
| genuine drift on `Assertion.date` | red | red, twice |

A genuine drift is now reported twice, once by the comparison and once by the coverage pair naming the same interface. That is the price of not recording a name until the work behind it has passed, and it is the cheaper side of the trade.

**An alternative was tried and rejected.** Registering at collection time beside each `it()`, which is roughly what the review note suggested, removes the pollution but decouples the guard from the work it certifies: it records "an `it()` was declared", not "a comparison ran". Under that version a gutted helper body or an early `return` goes green for **every** interface rather than one, and the test count stays identical at 74, so a reviewer watching counts sees nothing. That is strictly worse than the bug it fixes, so it is not what shipped.

**Not fixed here.** The note's other half, ordering coupling, remains. The coverage assertions need a full unfiltered run, and a `-t` filter or a shuffled order lists interfaces whose tests never executed. That predates this PR and fails loudly rather than silently. The assertions now say exactly that in their failure message so it does not read as schema drift. Removing it properly means moving the check into an `afterAll`, which needs its own break-test and is not folded in here.


francis-2008-happy (approved):
Approving. Reviewed in a worktree off the PR head, then ran three independent read-only passes over it (delivery-trace, test-falsifiability, pre-PR audit) on top of my own.

What I verified by running:
- viewer-ui vitest 271 passed; the drift-test file 74 passed; `turbo typecheck` (schema + viewer-ui) clean.
- Break-tested the optionality lint: removing the `?` from `Assertion.date` reddens `Assertion optionality ↔ $defs.assertion.required` with `missingQuestion: ['date']`. Restored, green.
- Traced all 25 now-optional fields to every read site (viewer-ui, eval/app, apps/web, apps/electron, engine): every read is undefined-safe, and the two arithmetic/array hazards (`match_score`, `subject_person_ids`) are the only real ones and both are fixed in both copies. tsc exit 0 confirms those two were the complete set for the type-protected package.
- Cross-checked all 25 `?` flips against both schema trees, both directions — each matches its `$def`'s `required`, no required field wears a spurious `?`. The load-bearing corpus counts and the raw `as ResearchData` cast both reproduce.
- Mutation-confirmed every new guard bites, including the `{true ? (` always-branch the new ProjectOverview positive case now catches, and the six defeat-routes on the drift lint.

The lint hardening against the "check that cannot fail" trap is the strongest part — planted-drift self-tests, a compared-set derived from the schema not the exemption list, and `NO_INTERFACE` pinned empty.

One non-blocking robustness note, since this PR's whole point is a durable lint: the `every intended interface was actually compared` coverage self-test relies on a module-level mutable `seen` set populated as a side effect across `it()` blocks, so its result is coupled to test execution order — one of my passes saw it fail once on a first run (the 8 Gedcomx interfaces), then not reproduce across 9 runs, and the same `seen` set is separately polluted for `Assertion` by the planted-drift block. It is safe today (deterministic order, and the `NO_INTERFACE` pin + planted-drift self-tests catch real drift order-independently), so nothing slips through — but computing the coverage set directly from the loops instead of a runtime accumulator would remove the flake/mask risk if vitest scheduling ever changes. Your call whether to fold it in now.

johnmarkpeterbrown (approved):
looks good

PR #1929, opened 2026-08-26
eval: research trigger corpus and stubbed routing suite (#1494)

### LINKED ISSUE #1494: eval: research trigger corpus (phase 1), then a stubbed routing suite (phase 2)

chrisedeson:
Closes #1494.

- **Phase 1a** — trigger corpus: 10 unit tests (5 positive, 5 negative) in `eval/tests/unit/research/` derived from SKILL.md's description clauses. Removed `research` from `RUNLOG_GATE_EXEMPT_SKILLS` (the gate auto-arms when the directory exists, per #1094). Updated `eval/CLAUDE.md`, `docs/architecture.md`, `check_runlogs.py`, `check_negative_reciprocity.py`, and rewrote the exempt test to assert gated behavior.

- **Phase 2** — stubbed routing suite: 5 routing tests that assert the first `Skill` call the router makes given a specific `research.json` state. Each uses `execution.stub_skills` (bare deny); a deterministic validator (`test_research.py`) parses the `routes-to:<skill>` tag, slices `skills_invoked` at the skill under test rather than filtering by name, and asserts the first delegation matches. Covers routing-table rows 1-4 plus the shortcut guard (SKILL.md lines 67-81).

Phase 1b already shipped (be00e48e, 2026-08-18).

- **Review fixes** — routing validator misattribution (a skill invoked before the router activates is not a delegation; `skills_invoked` is now sliced at the skill under test rather than filtered by name), routing tests now stub all downstream skills to prevent ownership-validator false failures, shortcut-guard turn cap raised to 30, slash-command test marked xfail with correct reason and wall clock raised to 1200s. Ambiguity guard and empty-value guard on `routes-to:` tags, stale "Direction" callout and broken definition-count prose in `docs/architecture.md`, renamed orchestrator-skill references to suiteless-skill, removed stale `build_eval_set.py` comment.

## What is NOT covered (and why)

Routing-table rows 14 (post-verdict `address_first` handler) and 16 (who writes `project.status = "completed"`) are not covered — blocked on #1492 (research/SKILL.md reconciliation). The README documents this.

No offline gate grades whether these are the right utterances; that is human review until someone pays for `make optimize-skill SKILL=research`.

The five reciprocal edges — `research -> {init-project, project-status, question-selection, research-plan, search-records}` — are the five one-directional edges this PR adds. Tracked for reciprocal coverage in the check_negative_reciprocity output.

## Verification

```sh
make harness-test           # 2879 passed, 1 pre-existing environmental failure (missing engine build), 9 skipped
cd eval/triggering && uv run python build_eval_set.py --skill research
                            # 15 queries (10/5)
```

`eval/runlogs/unit/research/v1_2026-08-27_15-25-47.json` — 8 pass, 1 partial, 1 xfail, 2 fail, 3 aborted of 15. The five routing tests are all green. The five trigger positives are not: 001/003/005 abort on the wall clock, 002 is xfail by design, 004 fails on single-run variance (annotated).

## Test plan

- [x] `make harness-test` passes (2879 passed, 1 pre-existing environmental failure)
- [x] `build_eval_set` produces 15-query set (10 positive, 5 negative)
- [x] Corpus validation passes (`test_unit_test_corpus.py`)
- [x] Validator runs standalone and skips correctly when no `routes-to:` tag
- [x] `check_negative_reciprocity.py` runs (warn-only; five new one-directional edges: research -> init-project, project-status, question-selection, research-plan, search-records)


eval/tests/unit/research/route-no-questions.json
```diff
@@ -0,0 +1,31 @@
...
+    "name": "Routes to question-selection when objective exists but no questions",
+    "type": "positive",
+    "description": "Routing table row 1: research.json has an objective but zero questions. The router must invoke Skill('question-selection') to derive the first question. Uses empty-project-just-created, which has only an objective and tree — the cleanest R1 state. question-selection is stubbed (bare deny) because the router does not read its return value; it re-enters the routing loop after the callee finishes.",
+    "tags": [
+      "routing",
+      "routes-to:question-selection"
+    ]
+  },
+  "input": {
+    "user_message": "Continue the research.",
```
Gennecis: This is the finding that blocks. The harness calls `query(prompt=user_message)` with `setting_sources=["project"]` (`skill_runner.py:484-544`) — nothing forces the skill, so the model has to pick `research` off its description. "Continue the research." isn't in that description's trigger list, and your own run log shows what happened instead:

| test | expected | `skills_invoked` | `activated` |
|---|---|---|---|
| `ut_research_011` | question-selection | `['project-status']` | **false** |
| `ut_research_012` | locality-guide | `['project-status']` | **false** |
| `ut_research_013` | research-plan | `['project-status']` | **false** |
| `ut_research_014` | search-records | `['research', 'search-records']` | true |

`ut_research_011`'s `text_response` is a status report opening "The Story So Far", with an empty `tool_calls`. The router never ran, so the routing table was never consulted — the test is measuring triggering. `ut_research_014` is the proof the design works once the skill activates.

Using a listed trigger form fixes it:

```
You wrote:      "user_message": "Continue the research.",
Change it to:   "user_message": "/research continue — drive the next step for this project.",
```

Same change in `route-no-plan-no-localities.json:14`, `route-no-plan-with-localities.json:14` and `route-planned-items.json:14`.

Side note worth keeping: "Continue the research." landing on `project-status` is a genuinely interesting signal. It might deserve a negative test in `eval/tests/unit/project-status/` one day — just not as the vehicle for these five.

eval/tests/unit/research/route-shortcut-guard.json
```diff
@@ -0,0 +1,33 @@
...
+    "type": "positive",
+    "description": "Shortcut guard (SKILL.md lines 67-81): the user says 'write the conclusion' but the project is at the objective-only stage — no questions, no evidence, no plan. The router must re-derive state from research.json and walk the routing table from the top, landing on row 1 (question-selection), NOT shortcutting to proof-conclusion. This guards against the specific anti-pattern the SKILL.md warns about: user-directed requests naming a downstream skill are not shortcuts. Uses empty-project-just-created.",
+    "tags": [
+      "routing",
+      "routes-to:question-selection",
+      "no-shortcut"
+    ]
+  },
+  "input": {
+    "user_message": "Write the conclusion for this research project.",
```
Gennecis: Same root cause as the routing tests, but worth calling out separately because the reasoning differs. The shortcut guard lives *inside* the router — SKILL.md:67-81 fires once `/research` is already driving. As a bare opening message, "Write the conclusion..." just triggers `proof-conclusion` directly, which is exactly what the run log records (`skills_invoked: ['proof-conclusion']`, `activated: false`). The guard was never reached.

Prefixing keeps the utterance you want to test while getting the router into the seat:

```
You wrote:      "user_message": "Write the conclusion for this research project.",
Change it to:   "user_message": "/research write the conclusion for this research project.",
```

eval/tests/unit/research/research-objective.json
```diff
@@ -0,0 +1,21 @@
...
+  },
+  "input": {
+    "user_message": "Research the parents of Patrick Flynn, born about 1845 in Ireland. I want to identify both his father and mother from primary records.",
+    "scenario": "mid-research-flynn"
+  },
+  "judge_context": [
+    "Should invoke sub-skills (question-selection, research-plan, search-records, etc.) rather than doing GPS work inline",
+    "Should read research.json state to determine the current routing step"
+  ],
+  "execution": {
```
Gennecis, line 18: `300` is already the harness default (`run_tests.py:321`, spec section 15), so this block is a no-op — and all five trigger positives hit it: `ut_research_001`-`005` all came back `outcome: aborted`, `aborted_reason: max_wall_clock_seconds`, judge skipped, zero graded dimensions.

The gate flags it itself when I run `check_runlogs.py` on this branch:

```
::warning::skill `research`: 9 test(s) in `v1_2026-08-25_20-14-29.json` produced no
graded dimensions, so nothing is required of them here: ut_research_005, ut_research_003,
ut_research_004, ut_research_001, ut_research_012. Read their `aborted_reason` /
validator results before treating this run as a clean pass.
```

A bigger clock alone won't fix it — the router is specced to keep driving the whole GPS loop, and with no `stub_skills` every sub-skill really runs. Since a trigger test only needs to know the router woke up, bounding the run is both cheaper and more on-target:

```
You wrote:      "execution": {
                  "max_wall_clock_seconds": 300
                }
Change it to:   "execution": {
                  "max_turns": 8,
                  "max_wall_clock_seconds": 600,
                  "stub_skills": [
                    "question-selection", "locality-guide", "research-plan",
                    "search-records", "record-extraction", "person-evidence",
                    "research-exhaustiveness", "proof-conclusion"
                  ]
                }
```

Same in `slash-research-question.json:18`, `find-relative.json:18`, `investigate-person.json:18`, `autonomous-handoff.json:18`. A `stub_skills` entry that never fires is harmless — the runnability gate only checks each name is a real skill directory (spec section 1210).

eval/harness/validators/test_research.py
```diff
@@ -0,0 +1,66 @@
...
+    real ``Skill`` call, so a response that only *narrates* a hand-off
+    ("I'll now invoke question-selection") cannot satisfy it.
+    """
+    if "routing" not in test.get("tags", []):
+        pytest.skip("not a routing test")
+    expected = _expected_skill(test)
+    if expected is None:
+        pytest.skip("routing tag present but no routes-to: data tag")
+    skill_under_test = test.get("skill", "")
+    delegations = [s for s in skills_invoked if s != skill_under_test]
```
Gennecis: The filter is right, but there's a gap just above it: nothing checks the skill under test actually activated. Two consequences, both visible in this PR:

**It misattributes.** For `ut_research_011` the validator reported *"Router's first sub-skill call should be 'question-selection', got 'project-status'"* — which sends the next reader into the routing table looking for a bug that isn't there. The router never ran.

**It can pass when no routing happened.** I called the function directly with `skills_invoked=["question-selection"]` and no `research` in the list — it returns clean. The overall test still fails on activation per spec section 7, but the validator's own verdict says "routed correctly" for a run where the router was never involved.

One assertion closes both:

```
You wrote:      delegations = [s for s in skills_invoked if s != skill_under_test]
Change it to:   assert skill_under_test in skills_invoked, (
                    f"'{skill_under_test}' never activated — skills_invoked="
                    f"{skills_invoked}. This is a triggering failure, not a "
                    f"routing failure; the routing table was never consulted."
                )
                delegations = [s for s in skills_invoked if s != skill_under_test]
```

This one is free — validators aren't in the run-log snapshot, so it doesn't cost a re-run.

eval/harness/validators/test_research.py
```diff
@@ -0,0 +1,66 @@
+"""Deterministic routing validator for the research orchestrator.
+
+Tag-gated on ``routing`` (the AST-recognized gate tag) and
+``routes-to:<skill-name>`` (the data tag naming the expected callee).
+Asserts the first entry in ``skills_invoked`` matches the expected sub-skill.
```
Gennecis: Three places still describe the pre-fix behaviour, so a reader who trusts the prose over the code gets the off-by-one back. All three are one-liners:

**Here (line 5):**
```
You wrote:      Asserts the first entry in ``skills_invoked`` matches the expected sub-skill.
Change it to:   Asserts the router's first delegation — ``skills_invoked`` with the skill
                under test's own activation filtered out — matches the expected sub-skill.
```

**Line 36-38**, where the stated invariant is the one your own run log disproves four times (`ut_research_011/012/013/015` have no `research` entry at all):
```
You wrote:      ``skills_invoked`` always starts with the skill under test's own
                activation (``"research"``), so delegations are everything after that
                self-invocation.
Change it to:   ``skills_invoked`` starts with the skill under test's own activation
                (``"research"``) only when the model reached it through an explicit
                ``Skill`` call; the guard above rejects the runs where it did not.
```

**`eval/tests/unit/research/README.md:21`:**
```
You wrote:      asserts `skills_invoked[0]` matches. `routes-to:stop` means the router
Change it to:   asserts the router's first delegation matches — its own activation is
                filtered out first. `routes-to:stop` means the router
```

Heads-up: the README edit *does* invalidate the snapshot (README.md is embedded), so bundle it with the test-file changes.

packages/engine/mcp-server/tests/packaging/doc-links.test.ts
```diff
@@ -452,7 +453,7 @@ describe("docs/ cite symbols, not line numbers", () => {
  */
 const ISSUE_REF_BASELINE: Record<string, number> = {
   "CLAUDE.md": 7,
-  "docs/architecture.md": 19,
+  "docs/architecture.md": 25,
```
Gennecis: This check isn't a CI failure to clear — it's the finding. Its own comment, eight lines above:

> *"A RATCHET, not a freeze: the count must match exactly, so removing a reference fails until the number here comes down with it, and **adding one fails outright**."*

I counted the delta: **+6 `#1494`, +1 `#1492`, -1 `#1012`** in `docs/architecture.md`. Five of the six `#1494`s are this PR's own issue, which closes on merge — and the comment's point is that the durable form is the fact, not the ticket. Every one of them reads fine without the number:

```
You wrote:      "docs/architecture.md": 25,
Change it to:   "docs/architecture.md": 19,
```

and drop the refs from the prose, e.g. at line 566:
```
You wrote:      > formerly exempt but gained a trigger corpus (#1494) and is now gated. Full rules:
Change it to:   > formerly exempt but gained a trigger corpus and is now gated. Full rules:
```

The `#1492` pointer for the two blocked rows can stay where it already is, in `eval/tests/unit/research/README.md` — the ratchet only scans `docs/specs/`, `docs/architecture.md`, `README.md` and `CLAUDE.md`, so that file is out of scope by design.

docs/architecture.md
```diff
@@ -1424,16 +1426,13 @@ lead you to them:**
...
-  (`make eval-inventory`) — one JSON file per test under `eval/tests/unit/` — and
-  across the 25 live suites the latest run log per suite totals **403 rows, 362
-  passing (90%)**. Those two numbers count different things and can diverge in
-  either direction: a test defined after its suite's last run has no row, and a
-  row survives for a test since deleted. Today they differ by one, in the first
-  direction: `ut_timeline_010` was added to `timeline` after that suite's last
-  run, so it has no row; no row survives for a deleted test. Both numbers are
-  facts about the snapshots — taken between 2026-07-27 and 2026-08-24 — not an
-  identity, so re-derive rather than quoting them.
+  run logs, and negative routing tests across 26 skill suites. **421** committed
```
Gennecis: Two numbers here need another pass. `make eval-inventory` on this branch prints **424** definitions, not 421 — main is at 409, plus your 15. Looks like 421 was derived before the two `main` merges brought in three more tests.

```
You wrote:      run logs, and negative routing tests across 26 skill suites. **421** committed
Change it to:   run logs, and negative routing tests across 26 skill suites. **424** committed
```

And "the 25 live suites" is stale by this PR's own change — you commit the 26th suite's first run log in this diff:

```
You wrote:        `eval/tests/unit/` — and across the 25 live suites the latest run log per suite
                  totals **403 rows, 362 passing (90%)**. Those two numbers count different things
Change it to:     `eval/tests/unit/` — and across the 26 live suites the latest run log per suite
                  totals **424 rows, 366 passing (86%)**. Those two numbers count different things
```

Please re-derive the rows/passing pair rather than taking mine — the paragraph itself says to, and my method gives 409/361 on `main` where the doc claims 403/362, so my "latest run log per suite" selection may differ slightly from whatever produced the original figures. The **26** and the **424** are certain either way.

docs/architecture.md
```diff
@@ -1528,7 +1527,7 @@ lives. A test is not just its definition: it usually needs a matching
 `eval/fixtures/mcp/` response, a dimension in that skill's `rubric.md`, and a
 check in `eval/harness/validators/`. `test.id` must be unique across the **whole**
 corpus — a duplicate is a blocking CI failure — and `runs_per_test` is pinned to
-1 by policy. 87 of the 404 definitions are **negative** tests that exist to prove
+1 by policy. 92 of the 421 definitions are **negative** tests that exist to prove
```
Gennecis: The 92 is correct — I counted, main has 87 and your five negatives make it 92. Just the denominator to carry over from the fix above:

```
You wrote:      1 by policy. 92 of the 421 definitions are **negative** tests that exist to prove
Change it to:   1 by policy. 92 of the 424 definitions are **negative** tests that exist to prove
```

eval/CLAUDE.md
```diff
@@ -251,13 +251,13 @@ The `eval-cosmetic-skip` label is for genuinely behavior-neutral edits only (rew
...
 **There is deliberately no baseline file and no count threshold — 45 of the corpus's 79 routing edges are one-directional today and the check still exits 0.** Both mechanisms were considered and rejected when it shipped (issue #945). An allowlist would tax exactly the behaviour we want: every future description-widening PR that adds one negative test would owe the reciprocal in the *other* skill's directory, which touches a second skill and so buys a second `make eval-skill` run plus a genealogist annotation, indefinitely. A count threshold is silently wrong — remove one edge, add another, the count is unchanged and CI stays green. If this is ever promoted to blocking it must compare the edge **set**, and that promotion should wait until someone has triaged which unbacked edges are deliberate one-directional near-misses, exactly as the sibling lint's 68 hits are waiting on issue #1003. Unlike its siblings the warnings carry no `file=` annotation: the offence is the *absence* of a file in the target skill's directory, so there is nothing to point at and annotating the source-side test would mark a file that is correct. The declaring test ids are named in the message text instead.
```
Gennecis: Your five negatives move this figure, and it sits two paragraphs below the one you rewrote. I ran the lint on both refs: `main` reports **84 edges / 44 one-directional**, this branch reports **89 / 49**. (The 45/79 was already a little adrift before this PR — worth correcting to the measured value while you're in the file.)

```
You wrote:      **There is deliberately no baseline file and no count threshold — 45 of the corpus's 79 routing edges are one-directional today and the check still exits 0.**
Change it to:   **There is deliberately no baseline file and no count threshold — 49 of the corpus's 89 routing edges are one-directional today and the check still exits 0.**
```

`docs/specs/unit-test-spec.md:768` carries the same figure if you want to catch it in the same pass.

Gennecis (changes requested):
Thanks — this is a real piece of infrastructure, and the parts that landed are properly landed. I verified the three headline claims rather than taking them on trust:

- `build_eval_set.py --skill research` now emits **15 queries (10 positive / 5 negative)**, exit 0. The vacuous `make optimize-skill SKILL=research` is genuinely closed.
- The run-log gate really is armed. I moved `eval/runlogs/unit/research/` aside and ran `check_runlogs.py`: `::error::skill research was touched but has no run logs`. It fires.
- The rewritten `test_formerly_exempt_skill_now_gated` really guards it. I put `"research"` back into `RUNLOG_GATE_EXEMPT_SKILLS` and watched the test go red.

`harness-test` is 2852 passed / 2 skipped on this head (main has added tests since your 2739), and `mcp-server` packaging vitest is 590/590.

**The fixture work under the routing tests is the best part of this PR.** I checked all four scenarios against the routing table by hand and every state claim in your descriptions is exact — `wilkins-first-plan` really has one `in_progress` question, zero plans and no `localities` key; `blyeberg-first-plan`'s `loc_001` really carries `for_place: "Stagstrup Sogn, Thisted Amt, Denmark"`, which is what makes row 3 rather than row 2 the right expectation there. That is careful, GPS-aware fixture selection and none of the comments below touch it.

**The one thing standing between this and merge is that four of the five routing tests never reach the router.** The committed run log shows `activated: false` and no `research` in `skills_invoked` for `ut_research_011/012/013/015` — `ut_research_011`'s transcript is a `project-status` report with zero MCP calls. The routing table was never consulted, so those four tests are currently measuring triggering. `ut_research_014` shows the design works the moment the skill does activate, so the fix is a phrasing change per test, not a redesign. Details inline.

**Two things worth adding to the description:**

1. The issue asks for an explicit call on the five reciprocals. I ran `check_negative_reciprocity.py`: your negatives add exactly five one-directional edges — `research -> {init-project, project-status, question-selection, research-plan, search-records}`. The current line ("no new edges from routing tests since they are positive") is true about the routing tests but reads as though nothing was added. Either "in scope" or "out of scope, tracked in #1375" is fine; the issue just wants it on the record.
2. The issue also asks the PR to say that nothing checks the *content* of the corpus. `test_unit_test_corpus.py` proves the JSON parses and the ids are unique — it says nothing about whether these are the right utterances. One line covers it: *"No offline gate grades whether these are the right utterances; that is human review until someone pays for `make optimize-skill SKILL=research`."*

**A cost note so you only pay once.** I ran `harness.snapshot.normalize` over mutated copies to work out which fixes invalidate your committed run log:

| Change | Forces a re-run? |
|---|---|
| `input.user_message`, `execution.*`, or a new test file | **Yes** |
| `eval/tests/unit/research/README.md` (it's inside the snapshot) | **Yes** |
| `eval/harness/validators/test_research.py` | No — validators are deliberately outside the snapshot |
| `docs/`, `eval/CLAUDE.md`, `doc-links.test.ts`, the PR body | No |

So: make all the test-file and README edits in one pass, then run `make eval-skill SKILL=research` **once, from this branch's worktree**, and get the annotation. Splitting them across pushes buys two runs for nothing.

**Optional, only because you'll be re-running anyway.** `routes-to:stop` in the validator works — I probed it and it passes and fails correctly — but no test uses it yet. That's routing-table **row 17** ("All questions `resolved` and `project.status` is `completed` -> Stop"), which isn't blocked on #1492, and the fixture already exists: `eval/fixtures/scenarios/flynn-resolved/` has `status: completed` with both questions `resolved`. One more file (`ut_research_016`, tags `["routing", "routes-to:stop"]`, scenario `flynn-resolved`) would turn that branch from documented into exercised. Entirely your call — it depends on the activation fix landing first, since `stop` means no delegation at all.

**One thing I looked at and am deliberately not asking you to change:** `routes-to:` lives in `test.tags`, which the snapshot strips as cosmetic, so flipping an expected route leaves the run log "active" and no gate asks for a re-run. That's a real hole, but `test_timeline.py:167`, `test_question_selection.py:127` and `test_search_familysearch_wiki.py:90` all carry values in tags the same way — it's a pre-existing harness question for the lead, not this branch's to solve.

eval/tests/unit/research/route-no-questions.json
chrisedeson: Fixed in `e05eca8b`. All four routing tests + the shortcut guard now use `/research continue — drive the next step for this project.` so the skill activates before the router is consulted. Good catch — the run log evidence was right there and I missed it.

eval/tests/unit/research/route-shortcut-guard.json
chrisedeson: Fixed in the same commit. Prefixed with `/research` as you suggested.

eval/tests/unit/research/research-objective.json
chrisedeson: Fixed. All five trigger tests now carry `stub_skills` (the eight downstream skills) + `max_turns: 8` + `max_wall_clock_seconds: 600`. The router stops after its first routing decision instead of driving the full GPS loop.

eval/harness/validators/test_research.py
chrisedeson: Added the assertion exactly as you suggested. The validator now rejects runs where the skill under test never activated, reporting it as a triggering failure rather than misattributing it as a routing mismatch.

eval/harness/validators/test_research.py
chrisedeson: All three updated: the module docstring (line 5), the function docstring (lines 36-38), and `README.md` line 21. Bundled with the test-file edits so it's one snapshot invalidation.

packages/engine/mcp-server/tests/packaging/doc-links.test.ts
chrisedeson: Removed all six `#1494` refs from `docs/architecture.md` and dropped the baseline from 25 to 19. Verified with the same `repoIssueRefs` regex the test uses.

docs/architecture.md
chrisedeson: Re-derived from the actual run logs on this branch: **26** suites, **424** definitions, **423 rows, 364 passing (86%)**. Used the same "latest run log per skill" selection you described.

docs/architecture.md
chrisedeson: Updated to **92 of the 424**.

eval/CLAUDE.md
chrisedeson: Updated to **49 of the corpus's 89 routing edges** in both `eval/CLAUDE.md` and `docs/specs/unit-test-spec.md:768`.

Gennecis (commented):
Good round. Six of the nine items are fully done and I re-derived each one rather than trusting the diff:

- `make eval-inventory` -> **424** definitions, **26** suites. Both numbers in `docs/architecture.md` now match.
- Issue refs in `docs/architecture.md` -> exactly **19**, baseline back to **19**. The ratchet is intact.
- Reciprocity -> **89 edges / 49 one-directional**. Matches `eval/CLAUDE.md:262`, and you caught `unit-test-spec.md:768` as well — thanks, I only asked for the one.
- All three stale `skills_invoked[0]` descriptions are gone.
- Harness pytest: 2846 passed. The one failure is `test_mock_mcp.py::test_successful_call_carries_no_is_error` with `build not found: .../build/tools/research-query.js` — the missing-engine-build case in a fresh worktree, not a regression. CI's `pytest` is green on this head.

**And the routing suite now does the thing it was built for.** `test_routes_to_expected_skill` passes on four of five: row 1 -> `question-selection`, row 2 -> `locality-guide`, row 3 -> `research-plan`, row 4 -> `search-records`. That's the assertion the suite exists for, and it's green.

## My last round's advice was wrong, and that's what's still red

I told you to put `/research` in front of the routing messages. That fixed the routing and broke the outcome, because of `eval/harness/harness/orchestrator.py:1202`:

```python
if spec.skill not in skills_invoked:
    return "fail"
```

A positive test cannot pass unless `research` is in `skills_invoked` — and the `/research` slash form enters the skill **without** a `Skill` tool call, so it never gets recorded. Your new run log shows the split cleanly, 10 for 10:

| message form | `research` in `skills_invoked`? |
|---|---|
| plain English (001, 003, 004, 005) | yes |
| `/research ...` (002, 011, 012, 013, 014, 015) | no |

The clincher is `ut_research_002`: it delegated to `question-selection`, `research-plan` **and** `search-records` — so the router plainly ran its whole loop — and `research` is still not in its list. The skill ran; the harness couldn't see it.

So `ut_research_011` and `ut_research_014` route correctly, pass every validator, pass the judge, and still come out `fail`. Sorry — that one is on me, not you.

`max_turns: 8` was also my number, and it was too low: all five trigger positives aborted at exactly 9 turns. Your own 10 on the routing tests was too low too — 012 and 013 aborted at 11 turns *after* making the correct first delegation, and `aborted` discards the verdict at `orchestrator.py:1171`.

## What's left

Three changes inline, all small. Together they should turn the four correct routes into four passes.

**And one thing for the lead, not for you.** `slash-research-question.json` tests the `/research` trigger, which is one of the skill's own advertised triggers, and the harness structurally cannot observe slash entry — so that test can never pass however it's written. I've suggested `expected_outcome: "xfail"` for now. The underlying gap (the harness records only `Skill` tool calls, so no positive test can cover a slash trigger) is worth a `nothing-checks` issue against `skill_runner`, which is above this branch. Happy to write it up if you'd rather not.

**Two things still missing from the description**, both from the issue rather than from me:

1. The call on the five reciprocals — `research -> {init-project, project-status, question-selection, research-plan, search-records}`. "In scope" or "out of scope, tracked in #1375", either is fine.
2. One line saying nothing checks the *content* of the corpus.

Also worth a pass while you're in there: the description still says the validator "filters out the skill-under-test's self-activation", which now contradicts the validator itself, and it still quotes 2739 for `harness-test`.

## Two things to watch, no action now

- `ut_research_015` made **no delegation at all** (`skills_invoked: []`, 6 turns). The validator failed it correctly. Once the aborts clear, that one's worth reading — it may be a real finding about the shortcut guard rather than a test problem.
- `ut_research_008` was annotated 3 -> 1 ("no skill invocation") on the 08-25 run and confirmed 3 on the 08-26 run, with identical behaviour both times (`skills_invoked: []`, `fail`). Worth settling which is right, since the annotations are calibration data.

eval/tests/unit/research/route-no-questions.json
```diff
@@ -0,0 +1,31 @@
...
+    "name": "Routes to question-selection when objective exists but no questions",
+    "type": "positive",
+    "description": "Routing table row 1: research.json has an objective but zero questions. The router must invoke Skill('question-selection') to derive the first question. Uses empty-project-just-created, which has only an objective and tree — the cleanest R1 state. question-selection is stubbed (bare deny) because the router does not read its return value; it re-enters the routing loop after the callee finishes.",
+    "tags": [
+      "routing",
+      "routes-to:question-selection"
+    ]
+  },
+  "input": {
+    "user_message": "/research continue — drive the next step for this project.",
```
Gennecis: Two changes here, and the first is me correcting my own advice.

**Drop the slash.** `orchestrator.py:1202` fails any positive test where the skill under test isn't in `skills_invoked`, and the `/research` form never gets recorded there. This test routed to `question-selection` — exactly right for row 1 — and still came out `fail` for that reason alone.

A plain-English message that matches a listed trigger gets recorded (4 for 4 in your run log) and keeps the routing state-driven:

```
You wrote:      "user_message": "/research continue — drive the next step for this project.",
Change it to:   "user_message": "Research the next open question on this project — drive the workflow forward.",
```

**And raise the turn cap.** 012 and 013 aborted at 11 turns against your 10, *after* delegating correctly — the router spends several turns on `research_query` before it routes, per SKILL.md step 1.

```
You wrote:      "max_turns": 10,
                "max_wall_clock_seconds": 300,
Change it to:   "max_turns": 20,
                "max_wall_clock_seconds": 600,
```

Same two changes in `route-no-plan-no-localities.json`, `route-no-plan-with-localities.json` and `route-planned-items.json`.

eval/tests/unit/research/route-shortcut-guard.json
```diff
@@ -0,0 +1,33 @@
...
+    "type": "positive",
+    "description": "Shortcut guard (SKILL.md lines 67-81): the user says 'write the conclusion' but the project is at the objective-only stage — no questions, no evidence, no plan. The router must re-derive state from research.json and walk the routing table from the top, landing on row 1 (question-selection), NOT shortcutting to proof-conclusion. This guards against the specific anti-pattern the SKILL.md warns about: user-directed requests naming a downstream skill are not shortcuts. Uses empty-project-just-created.",
+    "tags": [
+      "routing",
+      "routes-to:question-selection",
+      "no-shortcut"
+    ]
+  },
+  "input": {
+    "user_message": "/research write the conclusion for this research project.",
```
Gennecis: Same slash problem, but this one needs care: the test's whole point is that the user *names a downstream destination*, so the phrasing has to keep naming `proof-conclusion` while still triggering the router.

```
You wrote:      "user_message": "/research write the conclusion for this research project.",
Change it to:   "user_message": "Research this project through to a proof conclusion.",
```

That still names the destination, so the guard is still under test — it just enters through a form the harness can see. And the same turn-cap bump as the others:

```
You wrote:      "max_turns": 10,
                "max_wall_clock_seconds": 300,
Change it to:   "max_turns": 20,
                "max_wall_clock_seconds": 600,
```

eval/tests/unit/research/research-objective.json
```diff
@@ -0,0 +1,27 @@
...
+    "max_turns": 8,
```
Gennecis: My number, and it was too tight — all five trigger positives aborted at exactly 9 turns. The 600s wall clock never became the binding constraint, so that part worked; only the turn cap needs to come up. The default is 20.

```
You wrote:      "max_turns": 8,
Change it to:   "max_turns": 20,
```

Same in `find-relative.json`, `investigate-person.json` and `autonomous-handoff.json`. (`slash-research-question.json` needs something different — see the comment on that file.)

The `stub_skills` list is doing its job, by the way: `ut_research_003` delegated to `init-project` on a project that already has a `research.json`, which is worth a look once these stop aborting.

eval/tests/unit/research/slash-research-question.json
```diff
@@ -0,0 +1,27 @@
...
+  "test": {
+    "id": "ut_research_002",
+    "skill": "research",
+    "name": "/research slash command triggers the orchestrator",
+    "type": "positive",
+    "description": "The /research [question] form is an explicit trigger listed in the skill description. The user is handing off a specific research question to be driven through the full GPS workflow.",
+    "tags": ["core-trigger", "slash-command"]
+  },
+  "input": {
+    "user_message": "/research When and where did Patrick Flynn die?",
```
Gennecis, line 13: This test is asking for something the harness cannot currently give it, and it isn't your fault or the test's.

`skills_invoked` is populated only from `Skill` tool calls (`skill_runner.py:396-402`), and slash-command entry doesn't produce one. So `research` is never recorded, `orchestrator.py:1202` fails the test, and no phrasing of a `/research` message can get past that. Your run log proves the skill really did run: this test delegated to `question-selection`, `research-plan` and `search-records`, with `research` absent from the list the whole time.

Changing the message would lose the coverage — `/research [question]` is a real advertised trigger and deserves a test. So I'd mark it honestly instead and leave the message alone:

```
You wrote:        "tags": ["core-trigger", "slash-command"]
Change it to:     "tags": ["core-trigger", "slash-command"],
                  "expected_outcome": "xfail",
                  "xfail_reason": "The harness populates skills_invoked from Skill tool calls only, and slash-command entry makes none, so orchestrator.py:1202 fails every positive test using a /research message. Remove this marker when the harness records slash entry."
```

`expected_outcome` / `xfail_reason` are in the schema (`unit-test-spec.md:355-370`); nothing in the corpus uses them yet, so this would be the first — worth a quick sanity check that the report renders it as `xfail` rather than `fail`.

Also bring `max_turns` to 20 here like its siblings, so an xfail stays an xfail for the stated reason rather than turning into an abort.

eval/harness/validators/test_research.py
```diff
@@ -0,0 +1,68 @@
...
+    value = found[0] if found else None
+    if value is not None:
+        assert value, f"empty {prefix} tag value — fill in the skill name or remove the tag"
+    return value
+
+
+def test_routes_to_expected_skill(skills_invoked, test):
+    """The router's first sub-skill delegation must match the ``routes-to:`` tag.
+
+    ``skills_invoked`` records nested ``Skill`` calls made during the run,
```
Gennecis: Removing the assertion was the right call — it was failing tests that had routed correctly. But the reason written here isn't the reason, and the next author will act on it.

It isn't about being the entry point. It's the `/research` prefix: in your run log every plain-English message records `research` (001, 003, 004, 005) and every `/research` message doesn't (002, 011-015). `ut_research_014` recorded `['research', 'search-records']` on the 08-25 run, when its message had no slash — same test, same entry-point position.

```
You wrote:      ``skills_invoked`` records nested ``Skill`` calls made during the run,
                not the entry-point activation.  For routing tests the skill under test
                is the entry point, so it does not appear in the list — only its
                delegated sub-skills do.
Change it to:   ``skills_invoked`` is populated from ``Skill`` tool calls only, so the
                skill under test appears in it when the model reached it through such a
                call and not when it was entered as a slash command.  Both shapes are
                filtered the same way below.  Whether the skill under test ran at all is
                gated by the harness (``orchestrator.py``), not here.
```

One thing to know about what that leaves behind: `skills_invoked=["question-selection"]` with no `research` passes this validator, so on its own it reads a correct route where the router may never have run. That's fine as long as `orchestrator.py:1202` is the backstop — but it's worth the sentence above so nobody later reuses this validator somewhere that backstop doesn't apply.

chrisedeson:
All three items addressed:

1. **Routing tests (011-015)**: dropped the `/research` prefix, switched to plain English messages so `skills_invoked` records the entry correctly. Raised `max_turns` to 20 and `max_wall_clock_seconds` to 600 across all five.

2. **Trigger tests (001, 003-005)**: raised `max_turns` from 8 to 20. Slash-command test (002) marked `xfail` with the harness limitation documented in `xfail_reason`.

3. **Validator docstring**: rewritten per your comment. Now explains slash-command vs plain-English recording, not entry-point activation.

Gennecis (approved):

johnmarkpeterbrown (changes requested):
Nice work — the suite structure is right, and the round-2 fixes all landed. Four things to fix before this merges, then one re-run clears them together.

The reason these are blocking rather than follow-ups: this PR arms the run-log gate for `research`, and the run log it ships becomes the baseline that gate measures against. Two of that baseline's red rows aren't real failures. If it merges as-is, everyone who edits `research/SKILL.md` from now on pays a paid eval run and gets graded against a snapshot with two lies in it.

---

## 1. The routing validator blames the wrong skill

`eval/harness/validators/test_research.py`

The validator drops the skill under test from `skills_invoked` by name, then takes whatever is first. But a skill can fire *before* the router activates. In the shortcut-guard run, `project-status` ran first, then `research` activated and routed to `question-selection` — correctly. The validator reported `got 'project-status'` and failed it.

`skills_invoked` is in invocation order, so anything before the skill under test is not a delegation. Slice at it instead of filtering by name.

```
You wrote:      delegations = [s for s in skills_invoked if s != skill_under_test]
Change it to:   if skill_under_test in skills_invoked:
                    tail = skills_invoked[skills_invoked.index(skill_under_test) + 1 :]
                else:
                    tail = list(skills_invoked)
                delegations = [s for s in tail if s != skill_under_test]
```

This flips the shortcut-guard test and leaves the other four routing tests exactly as they are.

## 2. Routing tests stub only one skill, so a real sub-skill writes to the project

`route-no-questions.json`, `route-no-plan-no-localities.json`, `route-no-plan-with-localities.json`, `route-planned-items.json`, `route-shortcut-guard.json`

Each routing test stubs only the skill it expects. When the router carries on past that stub, the next sub-skill runs for real and writes to `research.json`. The ownership validator diffs the file across the whole run and attributes every change to `research`, so `research` gets failed for a write `research-plan` made and is entitled to make.

That is what fails `ut_research_012` today. Two other routing tests pass only because their next callee happened to be stubbed or wrote nothing, which means this suite currently passes or fails on luck.

The routing tests only assert the *first* delegation, so stubbing everything downstream costs no coverage. Use the same list the trigger tests use, in all five files.

```
You wrote:      "stub_skills": [
                  "locality-guide"
                ]
Change it to:   "stub_skills": [
                  "question-selection", "locality-guide", "research-plan",
                  "search-records", "record-extraction", "person-evidence",
                  "research-exhaustiveness", "proof-conclusion"
                ]
```

## 3. The shortcut-guard test runs out of turns before it can report

`route-shortcut-guard.json`

An aborted run discards the validator's verdict entirely. This test used 21 turns against a cap of 20, so its result is `aborted` and the routing check it exists for never lands. Fixing item 1 alone will not make it pass.

It has more to do than the other routing tests — it must explain the unmet preconditions before it routes. The others finished in 10 to 17 turns.

```
You wrote:      "max_turns": 20,
Change it to:   "max_turns": 30,
```

## 4. The xfail marker on the slash-command test doesn't bind

`slash-research-question.json`

The marker only converts a `fail` into an `xfail`. An `aborted` run is left alone. This test aborts on the wall clock, so the marker does nothing and the test ships red.

The stated reason is also not what happened. The activation check would have failed this test, but the run never got that far — it ran out of wall clock first. Give it enough time to finish and land on `fail`, where the marker applies.

```
You wrote:      "max_wall_clock_seconds": 600
Change it to:   "max_wall_clock_seconds": 1200
```

I sized that from one data point — 002 burning 600s across 45 turns — so treat the number as a starting guess, not a measurement. If you would rather derive it from the corpus, do that instead.

```
You wrote:      "xfail_reason": "The harness populates skills_invoked from Skill tool calls only, and slash-command entry makes none, so orchestrator.py:1202 fails every positive test using a /research message. Remove this marker when the harness records slash entry."
Change it to:   "xfail_reason": "Slash-command entry records no Skill call, so skills_invoked never holds `research` and the positive-test activation check fails it. The marker only binds when the run reaches that check — an aborted run is not remapped. Remove this marker when the harness records slash entry."
```

---

## Do them in this order

All four invalidate the run-log snapshot, so batch them and pay for one run, not four.

1. Make all four edits above.
2. Regenerate the run log **from this PR's worktree**, not your primary checkout. A run from the wrong tree produces a log that matches a different tree state.
3. Annotate it. Where something is still red, write what it is in the comment field — the current annotation has no comments on any row.
4. **Then** re-derive the corpus figures in `docs/architecture.md`. Do this last: the research suite's row and pass counts change with the re-run, so updating them first guarantees they are stale again. Three of the four numbers in that paragraph are already off at this head.

Also add a line to the PR body saying what the shipped run log actually contains — right now Verification lists only `harness-test`, and a reader takes the baseline as clean.

## One more thing to look at, not necessarily to fix

`ut_research_001` and `ut_research_005` abort on the wall clock after 44 to 48 turns, with all eight downstream skills stubbed. That is a lot of turns for a router that should be making one decision. If they still abort after the raise, these tests are measuring something other than triggering, and that is worth raising with whoever owns the corpus rather than raising the number again.

## On item 2 specifically

That change alters what the routing tests exercise, not just how they are graded. I read it as consistent with the tests' own `judge_context`, which already says further routing after the stub is expected. If the corpus owner wants routing tests to observe the router walking into real sub-skills, the fix is to skip the ownership validator on routing tests instead — that call is theirs, not mine.

Finally — the 14 open review threads block merge on their own, including the nine from round 1 you already answered. Worth closing them out.

chrisedeson:
All four items addressed in 04232002, pushed as 015ae295.

Validator slice: changed from filtering by name to slicing at the skill-under-test's position in skills_invoked. Pre-activation skills (like project-status in the shortcut-guard run) are no longer mistaken for delegations. The other four routing tests are unaffected since their skills_invoked lists don't have pre-activation entries.

Full stub list on routing tests: all five routing tests now stub the same eight downstream skills the trigger tests use. ut_research_012 improved from fail to partial (ownership validator no longer trips on real sub-skill writes).

Shortcut-guard max_turns: raised to 30. Test now passes (438s, well within wall clock).

Slash-command xfail: wall clock raised to 1200s. The run now finishes at 660s and lands on fail, where the xfail marker binds. Reason text updated to describe the abort interaction.

johnmarkpeterbrown (commented):
Approving. All four items from my last review are fixed, and the re-run proves
each one rather than just asserting it:

- `ut_research_015` now passes — the positional slice fixed the misattribution,
  and 25 turns fits the raised cap.
- `ut_research_012` is `partial` with no failed validators — the full stub list
  stopped `research-plan` writing and getting blamed on `research`.
- `ut_research_002` now records `xfail` instead of `aborted`, so the marker
  actually binds.

The baseline went from 9 green with two false failures to 10 green with none.
Losing the false failures was the thing that mattered. Thanks for the annotation
comments too — the note on `ut_research_004` about single-run variance is the
kind of thing the next reader needs.

Two things to fix before you merge. Neither needs another eval run.

## 1. The PR body describes the old validator

This one matters more than it looks. `PLAN.md` is gitignored, so the PR body is
the only lasting record of what shipped — and it currently records the mechanism
I asked you to replace.

```
You wrote:      routing validator off-by-one (`skills_invoked[0]` is the skill under test's self-activation, not the sub-skill; fixed to filter and check `delegations[0]`)
Change it to:   routing validator misattribution (a skill invoked before the router activates is not a delegation; `skills_invoked` is now sliced at the skill under test rather than filtered by name)
```

Two smaller things in the same body. The harness count is stale:

```
You wrote:      make harness-test           # 2739 passed
Change it to:   make harness-test           # 2879 passed, 1 failed (engine build absent), 9 skipped
```

```
You wrote:      - [x] `make harness-test` passes (2739 passed)
Change it to:   - [x] `make harness-test` passes (2879 passed, 1 pre-existing environmental failure)
```

That count is from `pytest tests/unit/` in a fresh worktree, which may not be
exactly what `make harness-test` selects. Run yours and use your own number if
it differs — but if it differs by a lot, work out why before merging. The one
failure is `test_mock_mcp.py::test_successful_call_carries_no_is_error`, and it
fails only because a fresh worktree has no `node_modules`, so the compiled engine
is missing. Six sibling tests skip on that condition; this one asserts instead.
Not yours to fix.

And add what the shipped run log contains, so nobody reads the baseline as clean:

```
Add under Verification:

`eval/runlogs/unit/research/v1_2026-08-27_15-25-47.json` — 8 pass, 1 partial,
1 xfail, 2 fail, 3 aborted of 15. The five routing tests are all green. The
five trigger positives are not: 001/003/005 abort on the wall clock, 002 is
xfail by design, 004 fails on single-run variance (annotated).
```

## 2. The corpus figures are stale again

Measured at this head: **430** definitions, **430** rows, **375** passing. The
87% is right.

```
You wrote:      run logs, and negative routing tests across 26 skill suites. **427** committed
Change it to:   run logs, and negative routing tests across 26 skill suites. **430** committed
```

```
You wrote:      totals **427 rows, 370 passing (87%)**. Those two numbers count different things
Change it to:   totals **430 rows, 375 passing (87%)**. Those two numbers count different things
```

To be clear, this is not on you — you derived them correctly after the re-run and
then a main merge moved them, which is what a main merge does. Three merges
landed while I was reviewing, so re-derive rather than pasting my numbers if
another one has landed since. Fix them once and merge; do not chase them into a
fourth round. The real problem is that nothing lints these figures, and that is
worth raising with the lead as its own piece of work.

## Not blocking, but worth knowing

**None of the five trigger positives is green.** I checked whether that
undercuts the phase-1a goal and concluded it does not: `build_eval_set.py` reads
test definitions rather than run outcomes, so the 15-query set still works, and
the run-log gate needs logs plus an annotation, not green rows. So the trigger
half still does both of its jobs. If something downstream does read those
outcomes as signal, tell me — that would change my answer.

What the aborts suggest is more interesting than a cap that needs raising. On
`mid-research-flynn` the router runs 38 to 53 turns and keeps invoking sub-skills
after all eight are stubbed and denied. That reads like a router that does not
stop. Raising the wall clock again would hide it. Worth its own issue rather than
anything in this PR — and if you file one, that behaviour is the finding, not the
timeout.

**`ut_research_008` is worth a second opinion.** It fails, but the judge scored
it 3 / 3 / N-A and its rationale says Claude behaved correctly — it declined and
asked for the person's details instead of invoking `init-project`. Whether that
counts as correct routing is a genealogist's call. Ask before changing the test's
`correct_skill` to make it green; adjusting an eval bar to match an outcome is
the wrong direction.

Finally, the 14 open review threads still block merge on their own — including
the nine from round 1 you answered back on the 26th. Those need closing before
this can go in.
```

chrisedeson:
Thanks for the thorough review and the approval. Both items are done:

PR body was already updated with the corrected validator description, harness counts, and run log breakdown.
Corpus figures in architecture.md pushed (3a60e3e7) — 430 definitions, 430 rows, 375 passing.
Noted on the router not-stopping behavior in the Flynn fixture — will file that as its own issue rather than raising the wall clock. And will leave ut_research_008 alone until a genealogist weighs in on whether declining counts as correct routing.

14 threads resolved.

DallanQ:
Heads-up, **not** a change request — nothing here needs to change and this should merge as it stands.

The lead reversed the 2026-08-23 delegation ruling on 2026-08-31: `/research` will spawn `proof-conclusion` and `research-exhaustiveness` directly instead of going through their same-named skills. That work is issue #2075, and it lands after this.

Two consequences land with **that** issue, not this PR, and issue #2075 already owns both:

1. `route-shortcut-guard.json` lists `research-exhaustiveness` and `proof-conclusion` in `execution.stub_skills`. Stubs apply only to `Skill` calls, so once the routing cells name agents both entries go inert and the real agent would run inside a stubbed test.
2. `test_routes_to_expected_skill` asserts `delegations[0] == expected` over `skills_invoked`, and its docstring correctly records that the list is "populated from `Skill` tool calls only". After the rename, a router that spawns `proof-conclusion` and then calls `question-selection` would pass a test named "Does not shortcut to proof-conclusion".

Both are sound on `main` today — the shortcut this test guards is still a `Skill` call, so the test catches it. The blindness is created by the rename, which is why the fix belongs to the card that does the renaming.

One scheduling note for whoever merges: this PR creates `eval/tests/unit/research/`, and the runlog gate keys its exemption on that directory not existing. So `research` stops being exempt the moment this lands, and whichever of the two branches merges second pays a `make eval-skill SKILL=research` run.

johnmarkpeterbrown (approved):
approved!

PR #1931, opened 2026-08-26
e2e corpus readers are branch-scoped, so in-flight graded runs are invisible (#1444)

### LINKED ISSUE #1444: e2e corpus readers are branch-scoped, so in-flight graded runs are invisible
> **Reviewed 2026-08-25 before junior handoff, and the lead has ruled.** The
> shape question is settled (caveat in the readers + an on-demand crawl script);
> both examples this issue was written on have since landed on `main`; and the
> blast radius below was wrong in a way that ships green — nine readers are
> affected, not two, and `make e2e-latency` does **not** go through
> `all_result_jsons()`. Read this block and the two sections at the end before
> starting.

**Touches:** eval/harness/e2e/guardrail_shadow_report.py, eval/harness/e2e/runlog_selection.py

`all_result_jsons()` walks `eval/runlogs/e2e/` in the **current checkout**, so
every corpus reader built on it — `make e2e-corpus`, `make e2e-guardrail-shadow`,
`make e2e-latency` — silently omits graded runs committed on unmerged branches.
Such a run is not reported as skipped; it is never seen, and no reader can say
so.

**Both examples this issue was written on have landed.** `heinrich-dewus-children-death`
is on `main` (`57b915a3`) and its branch `land-heinrich-dewus-children-fixture` is gone;
`katalin-horak-son` is on `main` too — **PR #1229 merged 2026-08-12, eleven days before
the 2026-08-23 ruling comment that cited it as still open and branch-only.** Re-derive
before quoting either: `git ls-tree --name-only origin/main eval/runlogs/e2e/<slug>/`.

**The population today, measured 2026-08-25.** Zero graded runs are in flight — no open
PR adds an `eval/runlogs/e2e/*/run-*.json` (`gh pr list --state open --limit 100 --json
number,files`, 20 open PRs). What a crawl reports instead is **23 result JSONs across 16
remote refs**, all abandoned July/early-August work with no open PR and no tip newer than
2026-08-06. So the value here is **not** recovering omitted runs — there are none to
recover. It is that a reader cannot say whether it omitted any. Build for that, and do
not tune anything against the count: those are dead branches. Note also that the original
consumer is gone — issue #1231, the graduation gate this issue cited, closed COMPLETED —
and `docs/architecture.md` §9.4 point 3 forbids graduating a gate on a corpus number
regardless.

**Why this is only icebox.** The runs merge eventually, so the bias is temporal,
not permanent, and the practical mitigation costs nothing: read a rate off an
up-to-date `main` with in-flight fixture PRs merged. That mitigation is now
documented in `replay_provenance`'s docstring and in
`guardrail-enforcement-spec.md` §4, which is the whole of the cheap fix.

**The unblock trigger**, if this ever thaws: a decision is about to be taken off
a corpus number (graduating a shadow guardrail, retiring a check) while fixture
PRs are open. Then the omission lands on the one reading that matters. Options
considered and not chosen: crawling remote branches with `git ls-tree` (real
work, speculative value), or warning when open PRs touch `eval/runlogs/e2e/`
(needs network in a pure-analysis module, which the current design deliberately
avoids).

Do not build the crawler without a concrete decision waiting on it — the
documented mitigation has been sufficient so far.

---

## Every site, not just the two in Touches

`all_result_jsons()` (`eval/harness/e2e/runlog_selection.py:38`) is the discovery
function for **eight** readers, and a ninth bypasses it:

- [ ] `make e2e-corpus` — `e2e/corpus_report.py:770`
- [ ] `make e2e-guardrail-shadow` — `e2e/guardrail_shadow_report.py:1319`
- [ ] `make e2e-nudges` — `e2e/nudge_report.py:318`
- [ ] `make e2e-wiki-failures` — `e2e/wiki_failure_report.py:473`
- [ ] `make e2e-agent-tools` — `e2e/agent_tool_usage_report.py:452`
- [ ] `make e2e-skill-episodes` — `e2e/skill_episode_report.py:380`
- [ ] `make e2e-detector-diff` — `e2e/detector_before_after_report.py:338`
- [ ] `make eval-inventory` — `e2e/inventory.py:51`
- [ ] **`make e2e-latency` — does NOT use `all_result_jsons()`.** It walks
      `E2E_RUNLOGS` itself in `all_latest_runs()` / `latest_run_for()`
      (`e2e/latency_report.py:388-406`). A fix inside `all_result_jsons()` leaves
      this reader silently unchanged while CI stays green. This is the one this
      issue's own text ("`make e2e-latency` silently omits…") gets wrong.

**The one shared site that reaches all nine** is `describe_window()`
(`eval/harness/harness/since_window.py:153`) — the line every reader already prints
to say what it excluded, including `latency_report.py:443`. Prefer hanging the caveat
there over editing nine call sites. Note it is shared with three *unit*-corpus
readers (`judge_report.py:489`, `skill_latency_report.py:450`,
`scripts/timing_report.py:127`), so word it for either corpus or gate it per caller.

**Do not change `all_result_jsons()`'s signature.** Five test modules replace or
monkeypatch it at nine sites (`tests/unit/test_e2e_corpus_report.py:459,643,647,666`,
`test_e2e_agent_tool_usage_report.py:346,377,390`,
`test_guardrail_shadow_report.py:818,869`), so returning a tuple breaks them all for
no gain — add a separate function instead.

**Two prose sites are now factually false and must land in the same PR:**

- `eval/harness/e2e/guardrail_shadow_report.py:614-621` — the `replay_provenance`
  docstring still names both spent examples and a deleted branch.
- `docs/specs/guardrail-enforcement-spec.md:447` — the "Both counts are
  **branch-scoped**" bullet; update it to describe whatever the reader now prints.

## Build: a caveat in the reader, the crawl in a script

**In the readers:** extend the line `describe_window()` already prints
(`eval/harness/harness/since_window.py:153`) with one always-true clause — that the
corpus read is the current checkout's, that runs committed on other refs are not
visible, and that `make e2e-branch-only` names them. No git call in any reader.

**New, on demand:** `eval/harness/scripts/branch_only_runlogs.py` plus a
`make e2e-branch-only` target. It diffs `git ls-tree -r HEAD -- eval/runlogs/e2e`
against the same listing for every local and remote-tracking ref (**no `git fetch`, no
network**), and prints each result JSON present on a ref and absent from HEAD, grouped
by ref, with each ref's tip date. It makes no attempt to tell in-flight work from
abandoned — the human triages, which is why this is a script and not a warning.

Measured 2026-08-25 with a 34-line stdlib prototype: **23 runs across 16 refs, ~7s**
over 139 refs (a cold first pass is slower). Re-derive rather than quoting the count.

**Decision, and what it beat:** the crawl runs on demand rather than inside each
reader, because it measured 23 dead-branch hits against **0** in-flight runs and would
otherwise have added its cost and its noise to every invocation. A reader-embedded
crawl also cannot live in `describe_window()` unchanged — that function is shared with
three unit-corpus readers — so it would have needed a per-caller gate and tests at nine
sites. Record this in `docs/specs/guardrail-enforcement-spec.md` §4 beside the existing
branch-scoped bullet (line 447); that write is part of this task, not a follow-up.

## Verification, and what nothing checks

- `make harness-test` is the instrument. **A test for this must stub git, not shell out
  to it:** `eval-harness-tests.yml` checks out at the default `fetch-depth: 1`, so
  `refs/remotes` is near-empty in CI and any assertion over real refs passes
  vacuously — the exact failure mode `CLAUDE.md` § "A new lint must be proven to fail"
  describes. Break it deliberately and watch it fail before committing.
- Run one reader by hand and paste its new line in the PR (`make e2e-corpus SINCE=all`,
  `make e2e-latency`).
- **Nothing checks this:** no CI job runs any of these readers, so the caveat's
  correctness rests on a human reading the printed output. Say so in the PR.

## Collision

PR #1872 (`1273-context-policy-hook-gaps`, open, updated 2026-08-25) edits
`eval/harness/e2e/guardrail_shadow_report.py` — including the file holding the stale
docstring above — and `e2e/detector_before_after_report.py`. **Do after PR #1872
merges**, or coordinate the docstring hunk with it.

taiwo-stack:
## Summary

Every **e2e** corpus reader now states, on every run, that it is scoped to the current checkout and cannot see a run committed only on another branch. `describe_window()` (shared by 13 print sites — 10 e2e readers, 3 unit readers) appends a fixed caveat to each e2e reader's own printed line. The three unit readers (`eval-timings`, `judge-report`, `skill-latency`) share the same function but only print that line under a `SINCE=` that resolves to a cutoff (`SINCE=all` is explicit and prints nothing); bare, they show every skill unfiltered and print no window line, so there's nothing for the caveat to attach to by default.

For e2e callers the caveat also names `make e2e-branch-only`, the new on-demand script that diffs `git ls-tree` between HEAD and every known local/remote-tracking ref (excluding any already merged into HEAD, so a run deliberately deleted after merging doesn't read as one HEAD is missing) and reports what another ref carries that HEAD doesn't. The script itself makes no network call; its Makefile target fetches (`--prune`) first, since a branch nobody has fetched locally is invisible to it regardless of how in-flight its work is. `inventory.py`, which has no `describe_window()` call, prints the caveat directly.

Not built: a crawl embedded in every reader (rejected — measured 23 stale-branch hits against 0 in-flight runs on 2026-08-25, so it would add cost and noise to every invocation for no live gain that day), and a warning based on open PRs (needs network in a module kept deliberately pure-analysis). `all_result_jsons()` itself is unchanged — no signature or behavior change.

Two now-stale prose sites fixed in the same PR: `guardrail_shadow_report.py`'s `replay_provenance` docstring (both examples it named are now on `main`), and `guardrail-enforcement-spec.md`'s branch-scoped bullet, which now also records the ruling and what it beat.

## Since the last review round

- Restored the HEAD-read-alone prerequisite test (`test_main_fails_clearly_when_only_the_head_read_fails`), pinning `rc == 1` rather than an incidental assertion on the per-ref call — verified it fails under a mutation that swallows the HEAD read and passes clean.
- Corrected the docstring that had argued this test was impossible: a single clean run can't distinguish "only HEAD fails" from "everything fails" (true), but a mutation test compares clean against mutant, and the two diverge there — so the test is not only possible but load-bearing.
- Swapped `test_main_fails_clearly_when_only_for_each_ref_fails` to raise `UnicodeDecodeError` instead of `CalledProcessError`, closing the gap where `main()`'s `UnicodeDecodeError` arm was untested at that level — verified it fails when that arm is dropped from `main()`'s `except`.
- Narrowed `guardrail-enforcement-spec.md`'s "under an explicit `SINCE=`" to "under a `SINCE=` that resolves to a cutoff" (`SINCE=all` is explicit and prints nothing) and applied the same correction to this PR body's opening paragraph.
- Dropped reviewer-name and "this PR" references from permanent source comments/docstrings (`branch_only_runlogs.py` ×2, the test file ×4, the Makefile ×1) — they read as scaffolding once merged; the rationale they carried is restated without the attribution.

## Testing

- `eval/harness`: full suite: 2891 passed, 3 skipped. Restored test and its mutation checks verified locally (fails under the swallowed-HEAD-read mutation; passes clean).


T-FEH (changes requested):
Reviewed on the merged head (`b1093101` — your `0cf27797` plus a clean merge of main). Harness **2866 passed, 2 skipped**; `ruff` clean; all CI green; eval gate satisfied; no paid re-run owed.

The mechanism is sound and the new test file is genuinely discriminating — 12 of 14 mutations caught, and no test in it is tautological. Three things block.

## 1. Already-merged refs are reported as carrying runs HEAD is missing

`branch_only()` diffs paths and never checks ancestry, so a ref fully merged into HEAD still gets reported. Live on this checkout:

```
refs/remotes/origin/senior-shaunese-applegarth-family-1878 IS a full ancestor of HEAD
  eval/runlogs/e2e/pauline-shaver-death-burial/run-2026-07-08_10-45-06.json
  eval/runlogs/e2e/scotland-thomson-grandparents/run-2026-07-08_06-30-09.json
```

Both are absent from HEAD because **`13b36e6b` (PR #627) deleted them on purpose**. So they are not runs main is missing — they are runs main removed. And `make prune-runlogs` deletes run logs as routine maintenance, so this class grows rather than decays.

```
You wrote:      "for-each-ref", "--format=%(refname)\t%(committerdate:short)",
                "refs/heads", "refs/remotes",
Change it to:   "for-each-ref", "--no-merged", "HEAD",
                "--format=%(refname)\t%(committerdate:short)",
                "refs/heads", "refs/remotes",
```

Measured both ways here: 238 refs scanned drops to 134, removing exactly the false positives and roughly halving runtime.

## 2. The "0 in-flight runs" premise is false — and the tool misses the counterexample

That measurement is the stated reason for not embedding a crawl in every reader, and the spec is about to record it as the reason a design was rejected. There is an in-flight graded run right now:

```
#1921 OPEN opened 2026-08-26T00:18Z
  eval/runlogs/e2e/elena-asmundsdotter-origin/run-2026-08-25_23-41-49.json  (+ .ann, .final-tree, .final-research)
  present on HEAD? NO
```

Fully graded, behind an open PR — exactly the population "0 in-flight runs" says is empty. It is 1, and it flipped within about a day of the measurement.

The sharper problem is what that implies for the tool. I ran the crawl: **23 runs across 16 refs, and zero mentions of `elena-asmundsdotter`** — because I have never fetched that branch. Every run it found is on a stale or abandoned ref; the one live in-flight run is invisible to it. The crawl is weakest exactly where #1444 is strongest.

Also: your test plan says "24 real runs across 17 stale refs" while the design paragraph says 23. I reproduce **23 across 16**, matching the prototype. The number is checkout-dependent by design, so either say which checkout it came from or drop it and keep the dated 23/16 the spec and Makefile already carry.

Keeping the module pure is right, but the Makefile recipe is not the module:

```
You wrote:      cd eval/harness && uv run python -m scripts.branch_only_runlogs
Change it to:   git fetch --prune origin
	cd eval/harness && uv run python -m scripts.branch_only_runlogs
```

## 3. The de-staling hunk re-asserts a claim that is false today

`guardrail_shadow_report.py:676` — the one hunk whose job was to check this — keeps "`heinrich-dewus-children-death` (now on `main`) was the first run ever to store live entries for this check". Running the module's own predicate returns 12 entries across 7 runs, and the earliest is `mary-mcandrew-son/run-2026-08-05_20-48-08.json`, a day earlier. Both are on `main`. Inherited, but re-asserted in the hunk meant to check it.

## Minor

- **"Every corpus reader … every time it runs" is not true of the three unit readers.** They pass `default="all"`, `parse_since("all")` returns `None`, and the print is guarded on `is not None` — so `make eval-timings`, `make judge-report` and `make skill-latency` print no caveat by default, only under `SINCE=`. The gating is what #1444 asked for; it is the shipped sentence in `guardrail-enforcement-spec.md:470` and the PR body that need narrowing to "every e2e corpus reader", with the unit behaviour stated.
- **The header double-counts.** `sum(len(e.paths) …)` counts (ref, path) pairs; `young-marriage-1828/run-2026-07-08_11-02-49.json` sits on two refs, so "23 run(s)" is 22 runs. A local branch and its remote twin will do this routinely — count `{p for e in found.values() for p in e.paths}`.
- **Neither hard prerequisite is individually pinned.** `test_main_fails_clearly_when_head_itself_cannot_be_read` makes *every* `check_output` raise, so making either the `_refs()` call or the HEAD read lenient on its own leaves the suite green — both survive. One test closes it: stub `for-each-ref` to succeed and only `ls-tree HEAD` to fail.
- **The `/HEAD` docstring overstates the narrowing** — a remote-tracking `origin/team/HEAD` *is* still skipped. Reasonable trade, but say so.

## One in my lane

`docs/e2e-testing-guide.md:534-536` tells the Windows team which corpus reports have no batch file: "`make e2e-corpus`, `e2e-nudges`, `e2e-latency`, `e2e-wiki-failures`, `e2e-compaction`". This PR adds a sixth and does not add it to that list — and `e2e-branch-only` appears nowhere in the e2e guide at all, only in the guardrail spec, which is not where someone looks for a corpus reader. Free; docs are in no snapshot.

Requesting changes. The design is right and the tests are good — what needs work is one ancestry check, one measurement, and a claim the de-staling hunk was supposed to catch.

taiwo-stack:
All three blocking findings fixed in 3c0ba5e5:

- `--no-merged HEAD` excludes refs already merged into HEAD, so a run deliberately deleted after merging no longer reports as branch-only.
- Header no longer double-counts a path present on two refs.
- Removed the "first run ever to store live entries" claim — verified false (a day-earlier run exists on `main`).
- Makefile target now fetches (`--prune`) first; module docstring and spec corrected to state the "0 in-flight" figure as a dated snapshot, not a standing property.
- Both hard-prerequisite git calls now individually tested; `SINCE=all` on unit readers and the `/HEAD` narrowing documented as-is.
- `make e2e-branch-only` added to the Windows batch-file list and given a home in the e2e testing guide.

Follow-up self-review (70662fe1) caught two more real bugs: an uncaught `UnicodeDecodeError` on a non-UTF-8 ref/path, and a test that never exercised the branch it claimed to. Also re-measured the "0 in-flight" figure rather than just reflagging it as stale: with the fetch fix in place, 21 runs across 16 refs, 1 of those 16 behind a currently-open PR.

Harness suite: 2869 passed, 3 skipped. Vitest: 2717 passed. CI green.

T-FEH (approved):
Second pass, on `98c3fdc4` (your `70662fe1` plus a clean merge of main). All three blockers are fixed, and I verified each by running rather than reading.

## 1. Ancestor filter — fixed, and I re-measured

`--no-merged HEAD` is in at `branch_only_runlogs.py:92`. Re-running the crawl here:

```
before:  23 run(s) across 16 ref(s)
after:   20 run(s) across 15 ref(s)
refs/remotes/origin/senior-shaunese-applegarth-family-1878  -> no longer reported
```

The two entries that ref contributed were files `13b36e6b` deliberately deleted, so they are gone for the right reason. The header double-count is fixed too — `len(unique_runs)` at `:152` rather than summing per-ref paths.

## 2. The in-flight premise — fixed properly, and this is the part I want to single out

The Makefile recipe now runs `git fetch --prune origin` before the crawl, and I confirmed the consequence rather than assuming it: **after a fetch, the crawl does find `elena-asmundsdotter`** — the exact run it was blind to last round.

More importantly, the module docstring does not just soften the number. It records that the count was *"re-measured rather than just reworded once it was found wrong"*, names the counterexample, explains the structural reason (`_refs()` sees only refs this checkout already knows), records the fix, and then re-measures on 2026-08-27 with the fix in place — reporting 1 ref behind a currently-open PR, "nonzero, as the counterexample predicted". That is the repo's own rule applied without being asked twice, and it is better than the correction I asked for.

One observation, not a finding: my crawl reports 20/15 where the docstring's 2026-08-27 re-measure says 21/16. The docstring already calls both figures "dated historical anchors, not a claim this script reproduces on every checkout", and the difference is consistent with dedup (21 paths → 20 unique runs) plus a day of ref drift. So the disclaimer is doing real work — I mention it only as corroboration of it. I did not re-derive the "1 ref behind an open PR" cross-reference.

## 3. The `heinrich-dewus` claim — deleted, not reworded

The false "first run ever to store live entries" sentence is gone from `guardrail_shadow_report.py` entirely, replaced by a pointer to the live mechanism (`describe_window()` plus `make e2e-branch-only`). Removing an unverifiable claim in favour of a mechanism is the right shape.

## The spec's scoping is now exact, and I checked it against behaviour

`guardrail-enforcement-spec.md:470` now distinguishes the e2e readers ("every time it runs") from the three unit readers, which "only print its line — caveat included — under an explicit `SINCE=`". I ran `make judge-report` bare and confirmed it prints no caveat, which is what the spec now says. Previously the spec claimed all of them, always.

## One thing still open — a test-only gap, and it is free

In round 1 two mutations survived. One is now pinned; the other is not. I re-ran both:

```
_refs() made lenient        -> 1 failed, 11 passed   CAUGHT (new)
HEAD read made lenient      -> 12 passed             STILL SURVIVES
```

The docstring at `:127-130` says both are hard prerequisites whose failure must propagate. The test that reads as the guard for the second one —

```python
def test_main_fails_clearly_when_head_itself_cannot_be_read(...):
    def fake_check_output(...): raise subprocess.CalledProcessError(128, cmd)
    ...
    assert "Could not list refs" in err
```

— makes *every* `check_output` raise and then asserts the **`_refs()`** error message. So it pins the refs listing, not the HEAD read, despite its name. Swallowing the HEAD read to `set()` makes every path on every ref look absent from HEAD, and nothing in the suite notices.

One test closes it: stub `for-each-ref` to succeed and only `ls-tree HEAD` to fail, then assert `rc == 1` and the HEAD-specific message. `eval/harness/tests/` is not in the run-log snapshot, so it costs nothing.

## Approving

The three blockers are genuinely fixed and the response to the second one is better than what I asked for. The remaining item is one test against code that is already correct, so I would rather see it land than hold the PR for it — but it is worth landing, because a hard prerequisite the docstring names and nothing pins is exactly the shape this repo keeps getting bitten by.

I have not exercised the Windows path or re-derived the open-PR cross-reference; both are noted above.

T-FEH (changes requested):
**Withdrawing my approval above.** A closer pass found that the one item I waved through is worse than I described, and I got a second detail wrong on top of it.

## The HEAD-read gap isn't open — it was closed, then deliberately reopened

I said "one test closes it… free… I would rather see it land." That was wrong in a way that matters: **the test existed.** `3c0ba5e5` added `test_main_fails_clearly_when_only_head_read_fails`. `70662fe1` deleted it and replaced it with a docstring asserting such a test cannot exist:

> a test that only makes the HEAD read fail can never tell "only HEAD fails" apart from "everything fails": `for-each-ref` is never reached either way.

That is false, and the argument describes the *unmutated* run only. A mutation test compares clean against mutant, and the two diverge. I wrote the probe — `for-each-ref` succeeds and returns a real ref, only the HEAD `ls-tree` raises — and ran it both ways:

```
clean code                 -> 1 passed
HEAD read made lenient     -> 1 failed   (rc=0, and it prints a confident report)
```

The mutant doesn't just lose an error: with no HEAD baseline, every path on every ref reads as branch-only, so it exits 0 having printed a wholly wrong answer. `branch_only_runlogs.py:127-130` names this a hard prerequisite whose failure must propagate, and nothing pins it.

The shipped code is correct. What blocks is the prose. A docstring that tells the next reader the gap is unclosable is worse than the gap — it is the "looks exactly like coverage" shape, and it will keep the hole open. Restore the deleted test and correct that paragraph; `eval/harness/tests/` is in no snapshot, so it costs nothing.

## And a correction to my own numbers

I reported the crawl at 20 runs / 15 refs and offered it as mild corroboration that the docstring's 21/16 is fetch-dependent. That was my error: I measured **before** running `git fetch --prune origin`. Re-measured after fetching, it prints

```
21 run(s) across 16 ref(s)
```

— the author's figure, exactly. So the re-measurement in the docstring reproduces, and my "corroboration of the disclaimer" was an artifact of my own stale refs rather than evidence of anything.

## Everything else in my approval stands

The three round-1 blockers are genuinely fixed and I verified each by running: `--no-merged HEAD` is in and the merged-ancestor ref is gone; the Makefile fetch means the crawl now finds `elena-asmundsdotter`; the `heinrich-dewus` claim is deleted rather than reworded. The response to blocker 2 in particular re-measured instead of rewording, unprompted, which is the repo's own rule and the best part of this PR.

## Three smaller things, all cheap

- **The module docstring kept the overclaim the spec was corrected for.** `branch_only_runlogs.py:6` still says "Every reader now prints a caveat"; the three unit readers print none by default. The spec was narrowed to say exactly this — the docstring and the PR body's opening sentence were not. `Every e2e reader` fixes both.
- **`main()`'s new `UnicodeDecodeError` arm is untested.** The per-ref arm is pinned; dropping `UnicodeDecodeError` from `main()`'s `except` leaves the suite green. Cheapest fix is to raise it instead of `CalledProcessError` in the existing `for-each-ref` failure test.
- **The spec's "under an explicit `SINCE=`" is imprecise.** `SINCE=all` *is* explicit and prints nothing, and the Makefile help advertises `SINCE=all|N|YYYY-MM-DD`. "under a `SINCE=` that resolves to a cutoff" is what actually happens.

Not blocking, but worth a pass before merge: seven places in permanent source name a reviewer or a PR round (`branch_only_runlogs.py` ×2, the test file ×4, `Makefile` ×1). That reads as scaffolding in six months.

Requesting changes on the deleted test and its docstring. Everything else here is in good shape, and I should have caught this in the pass where I approved it.

promise-emmanuel (changes requested):
Not adding a second review while T-FEH's request stands. But the docstring argues the test they're asking for is impossible, so I re-ran it independently to settle that before you spend a round on it.

Fresh worktree on 98c3fdc4:


test_branch_only_runlogs.py, clean                      12 passed
head_paths = _result_json_paths("HEAD") made lenient    12 passed   <- survives
Then a test where for-each-ref succeeds, the per-ref ls-tree succeeds, and only ls-tree HEAD raises:


clean                     1 passed
HEAD read made lenient    1 failed   (rc=0, stderr empty)
The mutant exits 0 and prints a full report — with no HEAD baseline, every path on every ref reads as branch-only, so it's wrong in the confident direction.

Your docstring is right about the clean run: branch_only() reads HEAD before _refs(), so on unmutated code "only HEAD fails" and "everything fails" really are indistinguishable. The step that doesn't hold is the conclusion — a mutation test compares clean against mutant, and there the two diverge. So the paragraph is true about one run and false about the test.

This passes clean and fails under the mutation, both verified here:


def test_main_fails_clearly_when_only_the_head_read_fails(monkeypatch, capsys):
    """Pins the HEAD-read prerequisite on its own. `for-each-ref` and the
    per-ref `ls-tree` both succeed -- only the HEAD `ls-tree` fails. On clean
    code this is indistinguishable from the both-raise case below; under a
    mutation that swallows the HEAD read it is not, and that is the case that
    matters: with no HEAD baseline every path on every ref reads as
    branch-only, so main() exits 0 having printed a wholly wrong report."""

    def fake_check_output(cmd, *, cwd=None, text=None, encoding=None):
        if cmd[1] == "for-each-ref":
            return "refs/heads/some-branch\t2026-08-01\n"
        assert cmd[1] == "ls-tree"
        if cmd[4] == "HEAD":
            raise subprocess.CalledProcessError(128, cmd)
        return "eval/runlogs/e2e/fixture/run-2026-08-01_00-00-00.json\n"

    monkeypatch.setattr(branch_only_runlogs.subprocess, "check_output", fake_check_output)
    rc = branch_only_runlogs.main()
    err = capsys.readouterr().err
    assert rc == 1
    assert "Could not list refs or read HEAD" in err
Don't restore the deleted version verbatim — I ran it, and it catches the mutation by tripping its own assert cmd[4] == "HEAD" on the per-ref call rather than by the rc == 1 check. It works for the wrong reason.

And the docstring on test_main_fails_clearly_when_only_for_each_ref_fails:

You wrote: so a test that only makes the HEAD read fail can never tell "only HEAD fails" apart from "everything fails": for-each-ref is never reached either way. The prerequisite that genuinely needs its own test is the other one -- HEAD read succeeds, for-each-ref fails.

Change it to: so a test that only makes the HEAD read fail cannot be told apart from the both-raise case ON CLEAN CODE -- for-each-ref is never reached either way. It is still worth having: under a mutation that swallows the HEAD read the two diverge, and clean-vs-mutant is what a mutation test compares. Both prerequisites get their own test.

Confirmed T-FEH's other point too: dropping UnicodeDecodeError from main()'s except leaves all 12 green.

promise-emmanuel (approved):
Approving on b9638197. I re-ran the mutation that blocked the last round rather than reading the diff for it.

Merge note: this clears my own request, but T-FEH's CHANGES_REQUESTED against 98c3fdc4 still stands and blocks merge until they re-review.

T-FEH (changes requested):
Reviewed on `bed964e5` (your `2ca73a51` plus a merge of main I resolved and pushed — note at the end). Harness **3009 passed / 3 skipped**, engine **2758 passed**, gate exit 0 with **zero** warnings.

**The finding I withdrew my approval over is closed.** What holds this up is a new one: the remedy is absent in exactly the case #1444 was filed for.

## The restored test is real, and it is the only thing that catches the bug

I withdrew because a test for the HEAD-read-alone case had existed, been deleted, and been replaced by a docstring arguing such a test was impossible. `bcd4ac9c` restores it, and the docstring now says the true thing — indistinguishable *on clean code*, which is not the same as untestable, "and clean-vs-mutant is what a mutation test compares."

Correct argument, so I checked it was a true one. Swallowing the HEAD read into an empty baseline — the exact failure the docstring names:

```
clean    13 passed
mutant   1 failed, 12 passed   FAILED …only_the_head_read_fails
```

One test of thirteen. The companion claim verifies too: dropping `UnicodeDecodeError` from `main()`'s except tuple kills only `…only_for_each_ref_fails`. No untestability reasoning survives anywhere in the module or its suite.

## The central measurement reproduces exactly, four days on and on another machine

The claim I most expected to have rotted:

```
$ uv run python -m scripts.branch_only_runlogs
21 run(s) across 16 ref(s)
```

And the half the body says was hand-checked, which I did mechanically — intersecting the 16 refs with `gh pr list --state open` (22 today):

```
refs behind a currently-open PR: 1    572-elena-reading-probe-and-rerun
```

Exactly as claimed. The pure-analysis claim holds too: the module's only outside calls are `git for-each-ref` and `git ls-tree`, and `git fetch --prune origin` lives in the Makefile recipe.

## [Blocking] "Every time it runs" is false precisely where it matters

`guardrail-enforcement-spec.md:470-471`:

> Every **e2e** corpus reader now states this itself, **every time it runs**

Every one of these readers early-returns on an empty corpus **before** reaching the `describe_window()` print. So the caveat appears when a reader found runs, and vanishes when it found none — which is the only case a reader can be wrong about branch scope.

It is not hypothetical. On this branch, for the very run your own measurement singles out:

```
$ uv run python -m e2e.corpus_report --test elena-asmundsdotter-origin
No committed runs found on/after 2026-08-17.
          ← no caveat
```

That run is on `572-elena-reading-probe-and-rerun`, behind open PR #1921 — the one ref in your 21/16/1. A reader on `main` asks the exact question #1444 is about, gets a bare "No committed runs found", and is told nothing. The remedy is silent in its own motivating scenario.

Nine readers share that early return (`corpus_report`, `agent_tool_usage_report`, `compaction_report`, `detector_before_after_report`, `guardrail_shadow_report`, `image_transcribe_report`, `nudge_report`, `skill_episode_report`, `wiki_failure_report`), and `latency_report` has a second hole — its print sits inside `if args.all:`, so `--test <slug>` never emits it:

```
$ uv run python -m e2e.latency_report --test alvro-taylor-marriage-1931 | grep -c "Scoped to this checkout"
0
```

One line above each `return 1` fixes it (`branch_scope_note` is already re-exported from `e2e.runlog_selection`), plus lifting `latency_report`'s print out of the `--all` branch. Narrowing the spec sentence instead is the cheaper option, but it leaves the highest-value case uncovered and `guardrail_shadow_report.py:682` would need the same narrowing.

Worth pairing with a guard: nothing checks that these readers keep printing the caveat — `test_e2e_inventory.py` pins `inventory.py` alone — so a dropped print goes unnoticed. That is `nothing-checks`-shaped.

## The merge, since the resolution is mine and not yours

The conflict in `guardrail-enforcement-spec.md` was not the one it looked like. Your branch carried its own copy of the replayed-counts bullet — **120 of 147, 788 links, 79 fixtures** — which merged PR #1977 has since re-measured to **115 of 149, 699 links, 76 fixtures**. `--ours` would have put a refuted figure back on main; `--theirs` would have dropped your caveat and remedy paragraphs. I kept both, then checked no stale figure survives in the file and that the six suites reading that spec still pass. Worth your eye — it is the kind of resolution that goes wrong quietly, and I got one wrong quietly on another PR this week.

## Requesting changes

Only on the caveat gap. The test restoration is exactly what I asked for and is mutation-proven, the measurement is the most reproducible one I have checked in this repo, and everything in my standing request from 2026-08-27 is resolved. Fix the empty-corpus path — or narrow the claim — and I approve.

T-FEH (approved):
Third pass, on `ae91d2ab` (your `5c455625` plus a merge of main I pushed). Harness **3020 passed / 3 skipped**, engine **2766 passed**, `tsc` clean, gate exit 0 with zero warnings.

Everything I raised across both rounds is closed, and the one I called `nothing-checks`-shaped came back with a guard.

## The caveat now prints where it matters, on all ten readers

I ran each one against a non-existent fixture rather than reading the diff:

```
corpus_report                 ✓    guardrail_shadow_report       ✓
agent_tool_usage_report       ✓    image_transcribe_report       ✓
compaction_report             ✓    nudge_report                  ✓
detector_before_after_report  ✓    skill_episode_report          ✓
wiki_failure_report           ✓    latency_report --test <slug>  ✓
```

(`detector_before_after_report` first read as a miss on my side — it requires `--detector` and exited on argparse before touching the corpus. With `--detector lane-check` it prints.)

And the exact case I blocked on, on this branch:

```
$ uv run python -m e2e.corpus_report --test elena-asmundsdotter-origin
No committed runs found on/after 2026-08-18.
Scoped to this checkout — a run committed only on another branch is not visible here,
and this reader cannot tell you if it omitted one. Run `make e2e-branch-only` to list
what other refs carry.
```

That run is still on `572-elena-reading-probe-and-rerun` behind open PR #1921. A reader asking #1444's own question now gets told the answer may be incomplete **and** which command closes the gap. Naming the remedy in the sentence is what makes it act on rather than worry about.

## The spec claim is made true, not softened

```
- limitation to its own printed line. The three **unit**-corpus readers
+ limitation to its own printed line when the reader found runs, and each
+ reader also prints `branch_scope_note()` directly on its empty-corpus path
+ (a bare "No committed runs found" never reaches `describe_window()`, and an
+ empty read is exactly the case a reader cannot rule out "the run exists on
+ another branch" — see `test_e2e_branch_scope_caveat.py`).
```

It scopes the original clause honestly, names the mechanism that was missing, states *why* the empty case is the one that matters, and cites the test. That is the harder of the two options I offered and the better one.

## The `nothing-checks` gap is closed with a guard that fails

I said nothing checked that these readers keep printing it. `test_e2e_branch_scope_caveat.py` parametrizes over every reader, and it discriminates — I dropped `corpus_report`'s empty-path print:

```
clean    11 passed
mutant   1 failed, 10 passed
         FAILED …test_empty_corpus_path_prints_the_branch_scope_caveat[e2e.corpus_report]
```

Exactly the reader I broke, and only that one.

## Round 1 and the measurement still hold

The restored HEAD-read test still dies under the mutation that swallows the HEAD read (`1 failed, 12 passed`), so it remains the only thing catching it. And the headline figure reproduces again, now several days and two merges on:

```
21 run(s) across 16 ref(s)
refs behind a currently-open PR: 1   (572-elena-reading-probe-and-rerun)
```

A dated measurement that has now reproduced three times, on a checkout that is not the author's, is about as good as this gets.

## One sentence I named last round and that did not get swept

`eval/harness/e2e/guardrail_shadow_report.py:683` still reads:

> `describe_window()`'s printed line says so on **every run of this report**

Now wrong in both directions: on the empty-corpus path the caveat comes from a direct `branch_scope_note()` call rather than `describe_window()`, and on the `--feedback-dir` path there is deliberately none at all — correctly, since that mode reads bundles rather than the corpus. I called this one out explicitly when I asked for the spec narrowing, so it is an omission rather than a disagreement. Something like "Every corpus read here says so — via `describe_window()` when runs were found, and via `branch_scope_note()` on the empty-corpus path" makes it true.

Not worth another round on its own: it is one internal docstring, the behaviour it describes is now correct and guarded, and the spec — the sentence that actually binds — was fixed properly.

## Approving

Three rounds: a deleted test restored and proven to discriminate, a claim that was false in its own motivating case made true rather than narrowed, and a guard added for the thing I only flagged in passing. The eval-staleness and merge-order questions are settled — this is ready.

PR #1935, opened 2026-08-26
init-project: fix source-count conflation and bound the error-detection list (#1901)

### LINKED ISSUE #1901: [feedback] 2026-08-25T13:55Z
**Zip:** https://drive.google.com/file/d/1g9dTGOmc3X85fuuYMiYYKgg7W174PTfn/view?usp=drivesdk

```
make feedback-case ZIP=~/Downloads/feedback-2026-08-25T13-55-21-938084Z.zip
```

- Submitted: 2026-08-25T13:55:21.938084Z
- Platform: web

Cia-3:
Fixes #1901

## Summary

Normal tier. Alpha-feedback bug report: init-project told a researcher a person had zero attached sources when 52 were actually returned -- the top-level `sources` array was conflated with each fact's own, correctly-empty `sources: []`. The same session separately flagged a missing relationship subtype and a shared given name as "structural errors," neither of which is on the skill's documented error-detection list, and neither was an error.

## Review guidance

**Start here:** the SKILL.md diff in `packages/engine/plugin/skills/init-project/SKILL.md` -- two additions: sourcing guidance distinguishing the top-level `sources` array from fact-level source refs, and a bounded, worked-example version of the "obvious error detection" list.

**Unsure about:** the worked-example rule was written to close a whac-a-mole failure mode I found during gating -- bounding two named examples (missing subtype, duplicate name) just pushed the model toward flagging a *different* unlisted pattern instead (an absent Couple relationship, then later a fabricated same-name-sibling claim, then an over-attribution of person-level sources to specific facts). The current wording generalizes the rule to "any pattern outside the five items," not just the two named ones. Across 7 post-edit samples taken during development, the *specific* reported false positive (subtype flagged as a structural error) did not recur, but the model still occasionally makes other, unrelated mistakes on this fairly complex task -- consistent with this harness having no `temperature=0` on skill runs, not a sign the fix failed. Worth a second pair of eyes on whether this residual noise is acceptable or whether Correctness is grading too broadly for a clean gate signal on this test specifically.

## Plan

**Didn't change:** nothing scoped out -- both halves of the reported bug are addressed.

**Acceptance check:** `ut_init_project_vqx` (new) fails against the pre-edit SKILL.md (Correctness 2, committed baseline run log) and passes against the edited version (Correctness 3, committed post-edit run log) -- gated per `skill-lifecycle.md`.

**Deviated from the plan:** the SKILL.md wording went through three iterations before landing (see Review guidance) -- the first two only named the two specific patterns from the feedback case, which didn't generalize.

## Test plan

- [x] Ran `make eval-skill SKILL=init-project` (post-edit). 12/13 pass. The one failure, `ut_init_project_001`, is pre-existing and unrelated -- a validator jitter ("Couple relationship missing a source ref") that also reproduced on the pre-edit baseline run and is independent of this fix.
- [x] `eval/harness/scripts/check_runlogs.py` run locally against current `origin/main` -- all runlog rules satisfied.
- [ ] N/A -- no `description` or DO NOT clause changed.

## Follow-on issues

**Folded in / filed instead:** both bugs from #1901 folded in directly -- no deferred work. The pre-existing, unrelated `ut_init_project_001` validator jitter (Couple-relationship sourcing) was noticed during gating but not chased -- it reproduces intermittently on the *current* unedited-in-that-regard SKILL.md rule ("attach a source reference to every fact and relationship," already explicit), so it reads as ordinary single-run model variance on an already-correct rule, not a doctrine gap.


mercyokum (changes requested):
**Requesting changes**

Good fix for a real bug — the fixture accurately reproduces how FamilySearch actually attaches sources (person-level, not per-fact), and the source-conflation half of the fix is solid. A few things worth resolving first.

## 1. `ut_init_project_001`'s "pre-existing jitter" got worse, not the same

Checked both committed run logs directly. `test_every_fact_and_relationship_is_sourced`:

- **Pre-edit** (`v1_2026-08-26_11-15-22.json`): fails on **1 of 5** relationships — `R1/Couple: no source reference`.
- **Post-edit** (`v1_2026-08-26_13-43-37.json`): fails on **5 of 5** — `R1/Couple`, `R2/ParentChild`, `R3/ParentChild`, `R4/ParentChild`, `R5/ParentChild`.

The fixture (`person-read-flynn-family.json`) has exactly 5 relationships total, so this is every single one losing its source reference post-edit, not the one-relationship jitter reproducing. Calling it "ordinary single-run model variance on an already-correct rule" undersells what the data shows. It's still plausibly noise (no `temperature=0` on skill runs), but it's equally consistent with the ~600 words added to `SKILL.md` crowding out attention to the pre-existing "attach a source reference to every fact and relationship" rule. Both logs are `n=1` for this test — worth a solo re-run of `ut_init_project_001` before merging on the current framing.

## 2. The "Worked example" closing paragraph is rationale, not instruction

> "The point of the worked example: the failure was never about which pattern got flagged — same-named siblings, a missing subtype, and a missing Couple relationship are three unrelated patterns that all produced the identical mistake. Do not treat 'I found a different reason this time' as compliance."

This narrates what was tried during gating (matches the PR body's own account of the whac-a-mole iterations almost verbatim), which CLAUDE.md's SKILL.md rule says belongs in the spec, not the body — it's billed on every future invocation. The sentence right before it already states the actual rule; this paragraph could go.

## 3. A few more, while in there

- **`SKILL.md`'s new "complete list" contradicts `check-warnings`.** Three of the five now-mandatory "obvious error" patterns (sibling births <9 months apart, non-existent-jurisdiction dates, birthplace inconsistencies with parents) are the *same* three patterns `check-warnings/references/warning-checks.md` explicitly lists as "NOT currently checked... Do NOT manufacture warnings of your own for these conditions." Two skills now give opposite documented answers on whether these count as findings, and this PR hardens one side without touching the other.
- **The new "commonly empty" claim about fact-level `sources: []` is factually wrong.** Per `person-read-tool-spec.md` and `gedcomx-convert.ts`'s `simplifyFact` (which only ever sets `sources` when there's a real ref), the key is never present-and-empty — it's just absent. The new paragraph describes a shape `person_read` doesn't actually return.
- **`judge_context` miscounts Josephine's relationships.** Bullet 3 says "one ParentChild link... with no subtype field" — the fixture actually has two (one from each parent), neither with a subtype.
- **No deterministic validator backs the regression test.** It's graded entirely by (noisy, no-`temperature=0`) judge prose. The harness already has a working pattern for a literal `text_response` phrase-check validator (`test_search_wikipedia.py`'s `test_reply_does_not_narrate_pending_step`) that could catch a future "no sources" regression deterministically instead of probabilistically.

## Worth considering, not blocking

- `person_read`'s output has no computed source-count field, so the agent has to correctly navigate the ambiguous shape every time — exactly what failed in production. `person_warnings`/`person_quality` already establish the precedent of returning a computed count (`warningCount`, `issueCount`) instead of leaving counting to the LLM. A tool-level fix there would be more durable than the prose recipe.
- The "Obvious error detection" section grew ~15.9x (28 → 446 words) to close two false-positive patterns from one feedback report. A shorter, already-general sentence was never tested as an alternative — the PR's iteration history only shows narrow, example-named wording failing to generalize, not that a short general rule would also fail.

Cia-3:
Thanks for the thorough review -- checked every point directly against the code before acting on any of it.

1. **ut_init_project_001 regression** -- confirmed real, not the noise I initially assumed. Traced it with 3 pre-edit samples (never worse than 1/5 relationships failing) vs 4 post-edit samples against the original addition (5/5 failing in half of them). Root cause: the ~600-word addition was likely crowding out attention to the pre-existing "attach a source reference to every fact and relationship" rule. Trimmed both additions to ~130 words total and moved the sourcing note to sit immediately after (not before) that instruction. Re-verified with 5 more standalone samples plus two full-suite runs against the trimmed version -- all clean, no further catastrophic failures.

2. **Worked-example rationale paragraph** -- removed. You're right that it was narrating the PR's own gating history rather than instructing behavior.

3. **Factual errors** -- both confirmed and fixed:
   - `sources: []`: confirmed via `gedcomx-convert.ts`'s `simplifyFact` that the key is omitted entirely when empty, never present-and-empty. Reworded to say "absent."
   - Josephine's relationship count: confirmed the fixture has two ParentChild links (one per parent), not one. Fixed in both `judge_context` and the test description.

4. **check-warnings contradiction** -- confirmed real and pre-existing (three of init-project's five "error" patterns are on check-warnings' own "do NOT manufacture" list). Not resolved in this PR -- filing a follow-up issue rather than deciding unilaterally which skill's doctrine should win.

5. **No deterministic validator** -- agreed, and going further than "worth considering": the residual false-positive rate on the missing-subtype pattern is reduced but not fully eliminated even after the trim (recurred once more in the final full-suite run, honestly annotated as such). Filing that in the same follow-up issue, since a mechanical check is the actual fix here, not more SKILL.md wording.

Everything above is verified with committed run logs and annotated. Ready for another look.

mercyokum (approved):

jud-sdev (approved):
Went through this line by line rather than reading, and it holds up. I counted the new mock fixture: exactly 6 top-level sources, zero fact-level refs, four persons, and Josephine's two ParentChild links subtype-less while Henri's both carry Biological — matching every claim in the test's judge_context, and the args predicate matches init-project's real person_read call so the mock actually gets hit.

I chased one thing that looked wrong — all 6 sources referenced by nobody — and it dissolved: the existing person-read-flynn-family mock uses the identical convention, so the fixture follows the repo shape and 'attached at the person level' correctly means the top-level sourceDescriptions rather than fact-level refs. The bounding rule doesn't over-suppress either: it routes other patterns to Gap detection or check-warnings, and judge_context still requires the five documented checks to run. The two SKILL.md lines are instruction, not rationale, and carry no issue-ref.

One non-blocking note for a later pass, not this one. The new line says to count len(sources) on the top-level array as how many are attached, but init-project calls person_read with relatives:true, so that array can include the relatives' sources — a literal count can overstate the subject's own. It is the milder mirror of the original bug and the judge accepts a qualitative 'sources exist,' so it does not block; if you touch this again, 'sources exist for this family' reads truer than a subject count.

Good, well-tested fix.

PR #1936, opened 2026-08-26
question-selection deep dive (#1668): rubric cleanup, four doctrine fixes, two new validators

### LINKED ISSUE #1668: deep dive: question-selection — and the rubric-deletion doctrine it pilots

mercyokum:
Closes #1668

**The issue:**
question-selection's rubric had three grading dimensions that never once caught a real problem across every committed run — dead weight taking up grading slots for nothing. Separately, the skill itself had several doctrine bugs that nothing was checking for: a rule that could make it stop looking for more evidence before a question was actually finished, wording that let it treat an untried record type as the wrong kind of research move, and a couple of places where it asked the right question the wrong way (or didn't ask when it should have).

**The fix:**
- Deleted the three rubric dimensions the issue authorized for deletion (Prioritization logic, Objective scope match, Dependency awareness) — each is now covered by an existing validator instead. Kept the fourth (Question specificity), since nothing covers it yet, and added a basic validator as a floor under it.
- Fixed a self-contradicting rule in the skill that could make it stop creating new research questions too early, even while a real gap (like unsearched census years) was still sitting there.
- Fixed the skill's FAN-research guidance (researching people connected to the subject) so it can't be confused with "try a record type nobody's searched yet" — those are different research moves.
- Fixed two places in the skill's "ask before assuming" logic: one where it kept demanding more detail after the researcher had already answered, one where a new test's grading unfairly penalized it for correctly choosing to ask.
- Fixed a case where the skill could conclude "no gap exists" without ever actually checking for one, because it never called the tool that would have told it.
- Removed a leftover issue-number citation from the skill's own instructions — skill files aren't supposed to carry those, only the reasoning docs. Cosmetic only, no instruction changed.
- Added two new tests and two new validators to lock all of the above in for future runs, and fixed a stale test that could only accept one of two equally correct answers.

Everything above is written up in detail, with the actual transcripts and paid-run evidence behind each fix, in `docs/deep-dives/question-selection-findings-2026-08-25.md`.

**Deferred issues found:**
- A `research_query` tool bug, not specific to this skill: filtering its `plans` section by `status` matches the whole plan's own status, not any item inside it — so asking it for "in progress" work can silently come back empty even when something is in progress. Worked around it in this skill's own instructions, but the tool itself could describe this more clearly so every skill that reads plans doesn't have to learn it the same way.


chrisedeson (changes requested):
Good work The deep-dive findings doc is thorough and well-evidenced, the rubric-to-validator migration is the right direction, and the four SKILL.md doctrine fixes each name the transcript evidence behind them. Six items below, two of which should be fixed before merge.

---

## Must fix

### 1. Two fixtures reference deleted rubric dimension "Objective scope match"

`eval/tests/unit/question-selection/first-question-from-objective.json` line 16 and `eval/tests/unit/question-selection/stay-inside-objective-scope.json` line 27 both instruct the judge to "Score Objective scope match as fail if..." / "is a FAIL on Objective scope match" — but that dimension was deleted from `rubric.md`.

The judge's dimension-name enum is built from `rubric.md` headings. A dimension name that isn't in that set gets silently dropped by `_extract_dimensions` (logged as `dropped_unknown_rubric_dimension`). So these fixtures' primary scope-violation checks are now dead text — the judge physically cannot score what they ask for.

**Fix:** reword the `judge_context` lines to reference the validator that now owns this (`test_new_question_not_record_scoped` / `test_new_question_excludes_out_of_scope_persons`) or reframe as a Correctness/Question specificity instruction.

### 2. `_VERIFY_SIGNALS` substring match produces false positives via "whether" and "test "

`_VERIFY_SIGNALS` uses plain `sig in text` substring matching. Two signals are too loose:

- `"whether"` matches in "Identify whether Anton Vogt's parents are Johann and Maria Vogt" — this reads as accepting the disputed parents (the exact failure mode the check exists to catch), but it passes because "whether" is present.
- `"test "` (trailing space, meant to require the standalone word) matches inside "latest ", "contest ", "protest ", "attest ". It also misses "test" at the end of a string (no trailing space).

Both `test_first_question_tests_disputed_parents` (existing) and the new `test_disputed_parents_missing_info_handled` use this list.

**Fix:** either switch to word-boundary regex (`r"\btest\b"`, `r"\bwhether\b"` scoped to "whether ... confirm|refute" rather than standalone), or tighten "whether" to require it appear alongside a confirm/refute framing.

---

## Should fix

### 3. `test_new_question_not_vague` regex over-matches on "Who is"

`r"^\s*who\s+is\s+\S"` fires on any present-tense question starting with "Who is", including specific ones like "Who is listed as Patrick Flynn's father on the 1850 census birth record?" This is universal (no tag gate, no `test` parameter), so no fixture can opt out. The multi-word surname pattern `r"^\s*research\s+the\s+\S+\s+family\b"` also misses multi-word surnames ("Research the Von Trapp family further" — `\S+` can only bind "Von").

**Suggestion:** anchor more specifically, e.g. `r"^\s*who\s+is\s+\S+\s*\??\s*$"` to catch only the bare-identity shape, or add word count / clause count to distinguish "Who is X?" from "Who is X's Y according to Z?"

### 4. `test_timelines_queried_before_deciding` does not check ordering

The function name and docstring say "before deciding," but `any(... for call in tool_calls)` checks only presence, not position relative to the `research_append` that writes the question. A run that decides first and queries timelines incidentally later still passes.

**Suggestion:** find the index of the last `research_query(section='timelines')` call and assert it precedes the first `research_append` call, or rename to `test_timelines_queried` to match what it actually checks.

### 5. `unblocks` lost coverage

The deleted "Dependency awareness" dimension checked both `depends_on` and `unblocks`. The replacement validators (`test_depends_on_nonempty` / `test_first_question_depends_on_empty`) only check `depends_on`. `test_universal.py` checks FK integrity on `unblocks` (do referenced IDs exist), but nothing checks whether it is populated when it should be. The rubric.md prose discloses that `unblocks` correctness has no mechanical check, but `ut_question_selection_007.json` carries an `unblocks` tag that matches no validator's tag-gate, and its `judge_context` line 17 now has no dimension to grade it.

**Suggestion:** add a `test_unblocks_nonempty` validator gated on the `unblocks` tag, or remove the tag from 007 and reword its `judge_context`.

---

## Informational (no action needed)

### 6. No fixture exercises the headline stop-condition change

The diff's main doctrine fix — status must be `resolved`, not just `in_progress` with a defensible tier — has zero fixtures where the old and new readings diverge (confirmed: no question in any committed scenario is `status=in_progress` with a non-null proof_summary tier). This is a genuine gap but not a blocker for this PR since the fix is preventive and the SKILL.md instruction is clear.

chrisedeson (approved):
Checked the changes. All six items addressed, the two must-fixes are solid.

The `_VERIFY_SIGNALS` regex tightening is better than what I suggested — dropping "whether" entirely and noting that every legitimate case already matches on "refute" or "independent" is cleaner than scoping it to a confirm/refute frame.

The timelines ordering assert (`min(query_idxs) < min(append_idxs)`) is the right shape — it catches the exact failure mode where the skill queries timelines after the decision is already made.

`test_unblocks_nonempty` with the `unblocks-nonempty` tag gate closes the coverage gap cleanly. Good work, approving.

EdmondOware (approved):
Checked the doctrine as genealogy and the rest against the repo rather than the
write-up.

**The four doctrine fixes are right, and two of them matter a lot.**

Gating the autonomous stop on the question's `status` being `resolved` rather than on a
defensible tier is the important one. A written proof summary is not an accepted conclusion,
and stopping at "a proof_summary exists at probable" would let the skill call the objective
answered before the conclusion had been critiqued. The corollaries follow properly: a
high-severity timeline gap is unsearched evidence and a probable-tier summary does not
discharge it, and exhaustion is orthogonal to tier. This is not an edge case either. I counted
the corpus: **36 questions across the scenario fixtures sit at `in_progress` with a
`proof_summary` already written**, so this is the state the skill meets constantly.

The FAN fix is the sharpest. FAN is cluster research on the people around the subject when
direct evidence fails; a deed or will that could name the relationship outright is direct
evidence. Without the fix the skill could claim a FAN pivot while doing ordinary direct
research it had not finished, and "whatever `exhaustive_declaration.declared` says" is the
right override, because a researcher's declaration cannot convert an unsearched deed into
cluster research.

The query-before-concluding-absence rule states the negative-evidence principle correctly, and
`pedigree_gap` really is absent from `selection_basis`, whose enum carries eight values and not
that one.

**On the rubric deletion, which is what #1668 is piloll cannot be
argued from the re-run scores, and it isn't: each deleted dimension names the validators that
now cover it, and the gaps are stated rather than papparison clause
under Prioritization logic, and `unblocks` under Dependency awareness. Applying the same
partial-coverage bar to Dependency awareness that yougic, and
saying so, is what keeps the pilot honest.

**Also verified:** @chrisedeson's two must-fixes are properly closed, no fixture names the
deleted dimension and `_VERIFY_SIGNALS` is word-boundr a documented
reason. The `no-issue-refs` change is a tightening rather than a weakening: you removed an
allowlist entry whose own text said it "leaves with q eval run",
and this is that run. Latest run 15 of 15, active at 47 entries with zero mismatches, annotated
across the full sample.

One piece of good news worth stating: you are one comhes none of
your 47 snapshot files, so the merge is free. No re-run owed.

eval/harness/validators/test_question_selection.py
```diff
@@ -392,3 +426,121 @@ def test_first_question_tests_disputed_parents(before_state, after_state, test):
...
+
+# The rubric's own `fail` bullet names these as the non-specific shape
+# ("Learn more about Patrick's family", "Who is Patrick Flynn?"). This
+# catches only that extreme case -- whether a question that avoids these
+# shapes is otherwise well-formed stays the judge's call.
+_VAGUE_QUESTION_PATTERNS = (
+    r"^\s*who\s+is\s+[A-Za-z][\w'\-]*(?:\s+[A-Za-z][\w'\-]*){0,4}\s*\??\s*$",
+    r"^\s*tell\s+me\s+about\b",
+    r"\blearn\s+(?:more\s+)?about\b",
+    r"\bfind\s+out\s+(?:more\s+)?about\b",
```
jud-sdev: These two patterns are unanchored, so they match `learn about` / `find out about` anywhere in a question, not just as its opening framing. Since this validator runs on every test with no tag gate, a specific well-formed question that happens to contain the phrase gets flagged as vague. I ran these through it, all flagged:

- "What can we find out about his military service from the 1890 pension file?"
- "Which record would help us learn about the cause of Mary Flynn's death in 1852?"
- "Where did Patrick Flynn learn about his trade?"

All three are exactly what the skill should produce — they name a record, a date, an event — and only trip on the embedded phrase. Anchor both to the start, where the phrase actually is the vague framing (this still catches the rubric's named shape, "Learn more about Patrick's family"):

You wrote:
> `    r"\blearn\s+(?:more\s+)?about\b",`
> `    r"\bfind\s+out\s+(?:more\s+)?about\b",`

Change it to:
> `    r"^\s*learn\s+(?:more\s+)?about\b",`
> `    r"^\s*find\s+out\s+(?:more\s+)?about\b",`

I verified the anchored versions spare all three questions above and still catch the vague openers.

Softer, same block: the `who is ...?` pattern on line 439 flags "Who is the father of Patrick Flynn?" because it allows up to five trailing words — an ordinary parentage-identity question. Worth tightening to a bare name, but that one is more of a judgment call than the two above.

jud-sdev (changes requested):
This is careful work and most of it holds up — I ran the two new validators against the corpus, confirmed each tag-gated one is exercised by a real test, checked the `whether` drop in the verify-signals does exactly what its comment says, and confirmed the no-issue-refs allowlist deletion lines up with the removed skill ref.

One thing to fix before merge, in the new vague-question validator. It runs on every question-selection test with no tag gate, and two of its patterns match the trigger phrase anywhere in a question rather than as the question's framing. Specific, well-formed questions get flagged as vague as a result. Details and the exact fix are in the inline comment. It does not fire on today's corpus, which is why it passed here, but it will fail a correct run the moment the skill writes a question phrased that way — and writing questions is the whole job.

Everything else I'm happy with once that lands.

mercyokum:
Confirmed all three examples as real false positives before touching anything — ran them through the validator as it stood, all three flagged. Fixed both patterns as suggested (anchored to the opening phrase), and also took the softer suggestion on `who is`: tightened it to require proper-noun-cased tokens only, rather than allowing any word. That closes "Who is the father of Patrick Flynn?" (now correctly spared) while "Who is Patrick Flynn?" still correctly trips it, since "the" doesn't start with a capital letter.

Verified before applying:
- All three of your examples: no longer flagged.
- Every real question actually written in the corpus's latest run (`v1_2026-08-26_17-56-18.json`, all 7 written questions): none flagged.
- The rubric's own named bad examples ("Learn more about Patrick's family", "Who is Patrick Flynn?", "Research the Von Trapp family further"): all still correctly flagged.

jud-sdev:
Thanks — the vague-pattern fix is exactly right, I re-checked it at the new head and it spares the specific questions while still catching the vague shapes. That part is done.

The other finding is still open, though, and I think the last update read as "merge main" when what it needs is different. Merging main again does not remove the earlier merge — commit 8254cece ("Merge branch ... refs/pull/1896/head") is still in this branch history, so PR #1896's check-warnings work is still baked in. I checked against current main: this branch still carries 7 files that belong to #1896 (docs/specs/person-warnings-tool-spec.md, packages/engine/plugin/skills/check-warnings/SKILL.md and its three references, and two check-warnings test files), +162/-24. #1896 is still open, so merging #1936 would pull that unreviewed work into main ahead of its own review.

Merging main can only add commits on top; it cannot undo a merge that already happened. To clear it you have to drop the pull/1896 merge, not merge over it. The clean way: branch fresh from origin/main, cherry-pick just the question-selection commits (the deep-dive work plus the vague-fix commit 898ef6be), leave 8254cece out, and force-push. After that the diff should be question-selection only and I am happy to approve — the code side is already sound.

Holding the change request just on the #1896 removal.

mercyokum:
Closing in favor of #1963 — same work, clean history. This branch's history had an octopus merge (`8254cece`) that pulled in PR #1896's still-open, unreviewed check-warnings branch alongside a normal main merge, and merging main again couldn't remove it (a merge only adds commits, it can't undo an earlier one already baked in). GitHub also blocks force-push to this branch, so rather than overwrite it, #1963 replays just this PR's six commits onto a fresh main. Diff verified to touch only question-selection files, nothing from #1896.

All review feedback from @chrisedeson, @EdmondOware, and @jud-sdev carries forward to #1963 — thank you all, and sorry for the extra round. jud-sdev's second finding (this branch-history issue) is exactly what triggered the move.

PR #1939, opened 2026-08-26
Harden encoding lint: dotted opens + Python 3.12 parse (#1355)

chrisedeson:
**Closes:** #1355

---

## Summary

- **Gap 1 (dotted opens):** the encoding lint now catches `Path(...).open()`, `PurePath(...).open()`, `io.open()`, and `builtins.open()` without `encoding=`. Uses a closed allow-list: only these four receiver patterns are flagged; ambiguous receivers (bare variables, unrecognized modules like `zipfile.open`) are left alone. `Path().open('rb')` correctly excluded via `mode_arg_index=0` (no file param).
- **Gap 3 (Python 3.12 parse):** bumps `requires-python` to `>=3.12`, adds `.python-version`, and pins CI to `3.12` so PEP 695 `type X = int` syntax no longer causes `SyntaxError` in the AST walker.
- **Gap 2** deferred to #1522 (suppression hatch). **Gap 4** already resolved by PR #1806 (scripts/ path gate).

## Test plan

- [x] `make harness-test` passes (2854/2854 on this branch)
- [x] `test_dotted_text_open_without_encoding_is_flagged` covers: Path, PurePath, io, builtins flagged; encoding-present passes; `encoding=None` flagged; binary-mode excluded; ambiguous receivers NOT flagged
- [x] `test_pep695_file_is_parsed_and_scanned` creates a tmp file with `type Vector[T] = list[T]` + bare `open()`, verifies parse + catch
- [x] Repo grep confirms zero live `Path().open()` / `io.open()` / `builtins.open()` calls (guard is preventive)


yinkid28 (approved):
Good PR — the allow-list approach, mode_arg_index parameterization, and _scan_tree extraction are all clean. A few suggestions, all in the test:

  Test coverage for the other allow-list members. PurePath and builtins.open share the same code paths as Path and io.open, but since the allow-list is the central mechanism here, it'd be nice to close the loop:

  # Binary-mode exclusion for the other two members:
  assert _offenders_in("PurePath(tmp).open('rb')") == []
  assert _offenders_in("builtins.open(p, 'rb')") == []

  # encoding= present:
  assert _offenders_in("PurePath(tmp).open('w', encoding='utf-8')") == []
  assert _offenders_in("builtins.open(p, encoding='utf-8')") == []

  # encoding=None (the exact Windows failure this lint prevents):
  assert _offenders_in("io.open(p, encoding=None)") == [(1, "io.open()")]

  Module-qualified constructor as a documented non-match. pathlib.Path(tmp).open('w') is correctly uncaught (receiver.func is ast.Attribute, not ast.Name), but adding it to the ambiguous-receivers block documents the design boundary:

  assert _offenders_in("pathlib.Path(tmp).open('w')") == []

  Minor: PurePath has no .open() at runtime — only concrete Path does. Including it defensively is fine, but a parenthetical in the docstring/module-docstring noting that would prevent someone reading the lint and concluding PurePath.open() is a real API.

  None of these block merge — the logic is correct and the primary paths are tested. Nice work on the endswith(").open") trick for distinguishing mode_arg_index.

chrisedeson:
Thanks for the review and the thorough suggestion list. Adding all four test cases:

- `PurePath` and `builtins.open` binary-mode exclusions
- `PurePath` and `builtins.open` encoding-present cases
- `io.open(p, encoding=None)` as an explicit encoding=None offender
- `pathlib.Path(tmp).open('w')` as a documented non-match (module-qualified constructor)

Good catch on `PurePath.open()` not existing at runtime. Will add a parenthetical in the module docstring noting the allow-list entry is defensive.

eval/harness/pyproject.toml
```diff
@@ -2,7 +2,7 @@
 name = "cowork-genealogy-eval-harness"
 version = "0.1.0"
 description = "Python harness for Cowork Genealogy skill evaluation. See docs/specs/unit-test-spec.md."
-requires-python = ">=3.11"
+requires-python = ">=3.12"
```
Gennecis, line 5: This raises the floor, but two CI jobs still run this package's code on 3.11:

- `check-runlogs.yml:155` -> `python eval/harness/scripts/check_runlogs.py`
- `check-e2e-fixtures.yml:67` -> `python eval/harness/scripts/check_e2e_fixtures.py`

Both run plain `python`, not `uv run`, so nothing enforces the floor and nothing warns. Both import `harness.snapshot`, `harness.versioning`, `harness.review_sample`.

I put one line of syntax this new floor makes legal into `harness/snapshot.py`, after the `__future__` import:

```python
type _Hash = str
```

| job | result |
|---|---|
| `pytest` (3.12 — the job you bumped) | 6 passed, green |
| `check-runlogs` (3.11 — untouched) | **`SyntaxError: invalid syntax`** |

So someone can write syntax the package declares legal, see pytest green, and red two other workflows. Before this PR the floor was 3.11 and everything agreed.

**To be clear, you followed the issue.** #1355 says those two "stay on 3.11" — but it was reasoning about the lint's parse target, not about the floor it was raising under two other consumers. Worth a lead's nod since it changes a written decision.

**Fix — two lines:**

```
.github/workflows/check-runlogs.yml:155
You wrote:      python-version: '3.11'
Change it to:   python-version: '3.12'

.github/workflows/check-e2e-fixtures.yml:67
You wrote:      python-version: '3.11'
Change it to:   python-version: '3.12'
```

eval/harness/tests/unit/test_encoding_lint.py
```diff
@@ -164,6 +205,10 @@ def _offenders_in(source: str) -> list[tuple[int, str]]:
         elif isinstance(func, ast.Name) and func.id == "open":
             if not _is_binary_open(node) and not _has_encoding(node):
                 offenders.append((node.lineno, "open"))
+        elif (dotted_name := _is_dotted_text_open(node)) is not None:
+            mode_idx = 0 if dotted_name.endswith(").open") else 1
+            if not _is_binary_open(node, mode_idx) and not _has_encoding(node):
+                offenders.append((node.lineno, f"{dotted_name}()"))
```
Gennecis: The name already ends in `()`, and line 233 appends `()` again. Real output from my planted offender:

```
eval/harness/scripts/check_runlogs.py:879: Path(...).open()() has no encoding=
```

This is the only text the lint ever prints, so it's worth being right.

```
You wrote:      offenders.append((node.lineno, f"{dotted_name}()"))
Change it to:   offenders.append((node.lineno, dotted_name))
```

Then drop the trailing `()` from the four expectations in `test_dotted_text_open_without_encoding_is_flagged` — e.g. `(1, "Path(...).open()")` becomes `(1, "Path(...).open")`.

eval/harness/tests/unit/test_encoding_lint.py
```diff
@@ -17,11 +17,18 @@
...
-    have no ``encoding=`` keyword.
+    Binary mode (a ``"b"`` in the 2nd positional arg or a ``mode=`` kwarg) is skipped;
+    the rest are flagged when they have no ``encoding=`` keyword.
+  - ``Path(...).open(...)`` / ``PurePath(...).open(...)`` / ``io.open(...)`` /
+    ``builtins.open(...)`` -- the closed allow-list of dotted opens whose receiver is
+    statically unambiguous. (``PurePath`` has no ``.open()`` at runtime -- only
+    concrete ``Path`` does -- but the allow-list entry is defensive: it prevents a
+    false negative if someone writes it, since the call would fail at runtime anyway.)
+    Only inline ``Path()``/``PurePath()`` constructor calls and
+    ``io``/``builtins`` module names are matched; a bare variable's ``.open()``
```
Gennecis: The docstring names the bare-variable gap but not the other one: `pathlib.Path(p).open("w")` is also uncaught. I planted one in a real file and the repo scan passed.

Population is 0 today, so I'm not asking you to widen the allow-list. But #1355 asked for this specifically — *"the module docstring must say so, or the next reader will assume the lint covers pathlib and it does not"* — and the module-qualified form is the one a reader is most likely to assume is covered, since `Path(...).open` is right there in the allow-list. It's documented in the test; the docstring is where the issue asked for it.

```
You wrote:      ``io``/``builtins`` module names are matched; a bare variable's ``.open()``
                (``p.open("w")`` where ``p`` came from somewhere else) is statically undecidable
                and stays uncaught.
Change it to:   ``io``/``builtins`` module names are matched. Two forms stay uncaught: a bare
                variable's ``.open()`` (``p.open("w")``), which is statically undecidable, and
                the module-qualified constructor ``pathlib.Path(...).open()``, whose receiver
                is an ``ast.Attribute`` rather than an ``ast.Name``. Neither occurs today.
```

Gennecis (changes requested):
The lint work is good and I checked it by breaking it, not by reading it:

| I broke | Result |
|---|---|
| emptied `DOTTED_OPEN_CONSTRUCTORS` | dotted-open test went **red** |
| planted `Path(p).open("w")` in `check_runlogs.py` | repo scan went **red**, named file and line |
| ran the PEP 695 test on 3.11 vs 3.12 | **3.11 fails, 3.12 passes** |

That third one matters most: the interpreter bump is backed by a test that genuinely fails without it.

I also re-derived the population: **3 dotted `.open` sites repo-wide** — two `Image.open`, one `webbrowser.open`, all the ambiguous ones you exclude on purpose. Zero real dotted text opens, exactly as #1355 measured. The guard is preventive, as scoped.

Test-merged main first (clean) and ran the suite: **2869 passed, 8 skipped, 1 failed**. The failure is `test_mock_mcp.py::test_successful_call_carries_no_is_error` — no engine `build/` in my worktree, not your change.

Two things you got right that the issue got wrong: the issue said "no re-lock is needed", but leaving `uv.lock` at `>=3.11` under a `>=3.12` pyproject is drift — you re-locked. And #1822's sequencing was respected.

**One change I'd want before merge**, in the comment on `pyproject.toml`. Two one-line fixes in the other comments.

**One more, not inline because the file isn't in the diff:** `eval/README.md:56` still says **"Python 3.11+"**. That's the setup instruction a developer follows and it's now wrong.

```
You wrote:      - **Python 3.11+** with [uv](https://github.com/astral-sh/uv) (`pip install uv` or `brew install uv`).
Change it to:   - **Python 3.12+** with [uv](https://github.com/astral-sh/uv) (`pip install uv` or `brew install uv`).
```

yinkid28's five suggestions are all in as of `05e21cfc` — their approval read `f59df3d5`, which predates it. Nothing of theirs is outstanding.

**Not checked:** I ran the suite without building the engine, so that one failure is matched-by-signature rather than proven, and I didn't itemise the 8 skips. I didn't run the two 3.11 workflows the way GitHub runs them — my proof used the same `python eval/harness/scripts/...` invocation with a local 3.11.

chrisedeson:
All four items addressed in `d8593716`.

1. **CI workflows bumped to 3.12**: both `check-runlogs.yml` and `check-e2e-fixtures.yml` now use `python-version: '3.12'`, matching the pyproject.toml floor. The issue said those "stay on 3.11" but that was about the lint's parse target, not the floor under two other consumers.

2. **Double `()` fix**: `_offenders_in` now returns the bare name (`Path(...).open`), and `_scan_tree` appends the single `()` for the human-readable message. No more `Path(...).open()()`.

3. **Docstring updated**: now documents both uncaught forms -- the bare variable gap and the module-qualified `pathlib.Path(...).open()` gap. Neither occurs in the repo today.

4. **eval/README.md**: prerequisite updated from Python 3.11+ to 3.12+.

Test expectations updated to match the new format (trailing `()` dropped from all dotted-open assertions). All 6 encoding lint tests pass.

Gennecis (approved):

Issue #1940, opened 2026-08-26
setup-feedback-case: the .sh dies at the imported commit, its ln -s silently makes copies, and a blank user_prompt prints an empty prompt block


florencemashipei:
**Touches:** `scripts/setup-feedback-case.sh` · `scripts/setup-feedback-case.bat` · `docs/specs/feedback-case-spec.md` §3 · `eval/harness/tests/unit/test_setup_feedback_case.py` · `packages/engine/mcp-server/tests/packaging/windows-wrappers.test.ts` · `docs/alpha-feedback-guide.md` (Step 2 Windows note)

**Not a duplicate of #1876.** That issue is the `.bat`'s `%~dp0`-after-`shift` bug, and its own review
notes nothing below line 51 of that script has ever run. Both defects here are in the **`.sh`**, at
lines 101 and 108/112 — past the point the `.bat` ever reaches, and #1876 scopes the `.sh` to *"error
text only"*. With #1876 open, the `.sh` **is** the Windows path, and it does not work either.

Both found first-hand unpacking the #1784 feedback case on Windows 11 / Git Bash, 2026-08-26.

## 1. The script aborts at the baseline commit when git has no global identity

```
$ bash scripts/setup-feedback-case.sh ~/Downloads/feedback-2026-08-20T11-32-50-917216Z.zip
…
Author identity unknown
*** Please tell me who you are.
fatal: unable to auto-detect email address (got 'flore@Florence-Taburu.(none)')
```

`git commit -q -m "imported"` (`:101`) runs in the freshly `git init`-ed case dir, which inherits
only **global** config. Anyone who commits through **GitHub Desktop** has a per-repo identity and no
global one — and the guide's own one-time setup tells the Windows team to use GitHub Desktop
(*"You don't need to know git from the command line"*). `git config --global --get user.email` exits
1 on this machine.

`set -euo pipefail` makes that fatal, so the run stops **after** unzip + `git init` + `git add` and
**before** the symlink loop and the closing printout. Two consequences:

- The genealogist never sees the *"User's prompt to issue first"* block — the one thing Step 3 needs
  pasted — and gets a stack of git advice instead of `✓ Imported to …`.
- **`feedback-reset` can never work on that case.** `scripts/reset-feedback-case.sh:33` checks
  `rev-parse --verify HEAD` and errors *"has no imported baseline to reset to"*. The guide mandates a
[...]

PR #1941, opened 2026-08-26
Never disable Send for a missing field, and mark the one that is required (#1919)

### LINKED ISSUE #1919: Feedback form: Send is disabled with no explanation, and four fields are required with no marker
**Touches:** packages/viewer-ui/src/components/shared/FeedbackDialog.tsx, packages/viewer-ui/src/components/shared/FeedbackDialog.module.css, packages/viewer-ui/src/components/shared/__tests__/FeedbackDialog.test.tsx, docs/alpha-feedback-guide.md

Relayed by the lead, 2026-08-25:

> A user said when they clicked "Send feedback" and filled out the form, the send/submit
> button at the bottom was disabled — they just got a red circle with a slash when they
> tried to click it. Most likely this was because they forgot a required field.

## What they saw

`.sendBtn:disabled` is `opacity: 0.5; cursor: not-allowed`
(`FeedbackDialog.module.css:319-322`) — the red circle with a slash. **Nothing on the
dialog says which field is missing.** The only validation message the dialog can show is
the over-10,000-character toast; there is no message for the missing-field case at all, so
a user who skipped one field has no way to find it and the report is lost.

## What is actually required

`canSend` (`FeedbackDialog.tsx:111-116`) gates on four things, and **not one of them is
marked required in the UI**:

- "Your email" — must match `EMAIL_RE`
- "What you asked the agent to do" — non-empty
- "What the agent did" — non-empty
- "Did it work as expected?" — Yes or No selected, with no default

The three genuinely optional fields each carry a visible "(optional)" tag. The four
required ones carry nothing, so the form reads as though everything is optional. The radio
is the likeliest omission: it is the only required field with no text cursor to draw the
eye, and nothing is preselected.

This is a different field from issue #1480, which was closed on 2026-08-13 by making
"What it should have done" optional. That fix removed one required field and left the
silent-disabled behaviour untouched, which is why this recurred.

## Relaxing it costs nothing downstream

Neither backend requires any of these. The server model defaults every field to `""` /
`False` and validates only length (`apps/server/app/feedback.py:282-289, 325`); the
Electron bundler validates only length (`apps/electron/src/main/feedback.ts:126, 269`). The
submission format already states `email` "may be empty string for anonymous submissions"
(`apps/electron/docs/feedback-json-spec.md:90`), and every field is always present as a
key. **The gate is entirely in this one component.** Relaxing it needs no schema change and
no `schema_version` bump.

One exception, and it is the reason "make everything optional" is not free:
`worked_as_expected` is a non-nullable boolean on the wire (`feedback-json-spec.md:93`).
Allowing it unanswered means either defaulting to `false` — which files every unanswered
report as a bug — or widening the field to nullable, which the spec's versioning rule makes
a `schema_version` bump coordinated across both repos.

## Recommendation, for the lead to confirm

The lead's question was "should we make all fields optional?" Recommendation:

1. **Never disable Send for missing content.** This is the durable fix and it holds whatever
   is decided about required fields. A disabled button with no explanation is the defect;
   the reporter cannot tell it apart from a broken app.
2. **Make email, "What you asked", and "What the agent did" optional.** The session log and
   the project folder ship in the same bundle and carry the prompt and the transcript
   verbatim, so an empty text field loses far less than a lost submission does.
3. **Keep the Yes/No answer, but enforce it visibly** — on Send, scroll to it and say it is
   the one thing still needed, rather than sitting dead. That avoids the `schema_version`
   bump and keeps the one field triage reads to tell a clean report from a problem.

If the lead would rather have it fully optional including the radio, that is fine — it just
adds the nullable-boolean bump and a paired PR to the two repos, which should be said out
loud before someone starts.

## Also update

`docs/alpha-feedback-guide.md:594-595` tells triage that an empty `user_prompt` or
`agent_did` means "the submission is missing a required field — ask them to resubmit". That
stops being true under this change.

## Done when

- A user who has missed something can tell what it is without guessing, and never sees a
  dead Send button with no explanation.
- A report with nothing typed in the optional boxes still submits.
- Triage can still tell a clean report from a problem report.

clack391:
## Summary

- The feedback dialog's Send button is never disabled for missing content again. A refused click now says what is wrong, puts an inline message under the offending field, and moves the cursor there. Normal tier. Closes #1919.
- Email, "What you asked the agent to do" and "What the agent did" become optional and carry the `(optional)` tag the file already uses. "Did it work as expected?" stays required and is now the only field marked `(required)`, which is what keeps triage able to tell a clean report from a problem report.
- No `schema_version` bump. `worked_as_expected` stays a non-nullable boolean, and no field's type or meaning changes.

The defect was `canSend`, which gated Send on four things, none of them marked in the UI, and rendered `opacity: 0.5; cursor: not-allowed` with no explanation when one was missing. The reporter could not tell a missing field from a broken app, so the report was lost. Issue #1480 removed one required field in August and left the silent-disabled behaviour untouched, which is why this recurred.

**What it does not solve.** A report submitted with every box blank is still a thinner report. The bet the issue makes, and this PR keeps, is that a thin report beats a lost one, since the session log and the project folder ship in the same bundle. Both markdown producers now print `_(not provided)_` for a blank email, prompt or description, so a triager can tell "left blank" from "lost in transit".

## Review guidance

**Start here:** `packages/viewer-ui/src/components/shared/FeedbackDialog.tsx`, the `blockers` memo and `handleSend`. Two things worth checking. First, `blockers` sorts by the `FIELD_ORDER` module constant and appends unknown ids, so a refused Send points at the first thing wrong in render order and a FIELD_ORDER/JSX desync costs message order rather than dropping a blocker. Second, `handleSend` tests `worked === null` inside the same branch that surfaces the message: that is what narrows `boolean | null` for the payload.

**Corrected after review.** This section previously said `blockers` was built by walking FIELD_ORDER and that "only a test pins it". Both were wrong. Walking it meant any id absent from the list was silently discarded, and no test pinned it: removing `feedback-prompt`, `feedback-did`, `feedback-should` or `feedback-answer` each left the suite fully green while a 10,001-character field submitted with nothing on screen. chesworthrm found this and supplied the fix, which is what now ships. Two further defects came out of the same pass: the error state latched after the first refusal, so a field typed later went red and the footer claimed a refusal that never happened; and a refusal did not clear a previous upload failure, so two toasts stacked. Both are fixed and both have tests that fail when the fix is reverted.

**Settled, and it was the one thing left open.** This issue ended with "Recommendation, for the lead to confirm" and nothing recorded. chesworthrm has since written the decision onto #1919 (2026-08-27): the Yes/No answer stays required, and email, "What you asked" and "What the agent did" go optional, on the ground of this issue's own third acceptance criterion rather than the coordination cost. **The cost the issue quotes for the alternative is wrong and I did not repeat it:** it says making the radio optional too adds "a paired PR to the two repos", but both producers are in this repo (`apps/server/app/feedback.py` and `apps/electron/src/main/feedback.ts`). The real cost is a `schema_version` 1 to 2 bump plus one PR here. That ruling independently confirms the correction.

## Plan

**Didn't change:** the two backend validators. Both already default every field and validate only length, so relaxing the dialog needed nothing there, and I confirmed that by reading rather than assuming. Also `scripts/setup-feedback-case.sh` and its `.bat` twin (see Follow-on). **Corrected:** this line originally also listed `docs/specs/feedback-case-spec.md` as untouched. That stopped being true two commits later, when review found it was on the documented edit-site route; the diff now changes two of its rows.

**Acceptance check:** `packages/viewer-ui/src/components/shared/__tests__/FeedbackDialog.test.tsx`. Restoring `origin/main`'s component under the branch's tests fails **14 of the 18**, including `keeps Send clickable with the radio unanswered`, `submits with every optional box blank` and `refuses an over-limit field without ever disabling Send`. The first fails on `toBeEnabled()`, so it fails on behaviour rather than on a label mismatch against the older labels. On the branch all 18 pass. (This line read "7 of the 11" while the branch was at `ab14be5d`; it was correct then and went stale as later commits added tests. Re-derived at each head since.) `apps/server/tests/test_feedback.py::test_blank_submission_is_accepted_end_to_end` is the one that covers the real HTTP entry point *with a blank body*. (An earlier version of this line said it was the only test covering that entry point at all, which was wrong: three others POST to `/api/feedback`, none of them with an empty submission.)

Every guard here was proven able to fail. Eight mutations, each caught: deleting the length-blocker loop, forcing the error-display gate always-on (then called `showErrors`; replaced by `refusedIds` in a later commit), weakening `EMAIL_RE` to `/@/`, stripping `role="alert"`, re-disabling Send on the over-limit path, dropping the blank-email placeholder in either producer, drifting `NOT_PROVIDED` in the Electron producer, and adding a non-empty requirement to the server endpoint. Round 3 pinned five more. Round 4 replaced the per-field approach with a guard over the invariant itself: deleting any one of the seven inline-message render sites, or marking only the first blocker of a refusal, now fails a test. Four of those seven were unguarded before it.

**Deviated from the plan:** yes, in four places from the author self-review, plus six more from review, plus round 4 below.

1. **The announce mechanism was redesigned.** The plan kept the live over-limit toast untouched and gave inline messages only to missing-kind blockers. Measured against the real component, that left the length path silent: a refused Send on an over-limit field produced no `role="alert"` and no new message. Now every blocker gets an inline message, one short alert toast says the send was refused, and the live toast is suppressed once the inline messages carry the same text.
2. **The placeholder covers email too.** The plan named two fields; email is the third one relaxed, and a blank one rendered `- **From:** ` with nothing after it.
3. **A server endpoint test was added.** Both original producer tests call `_feedback_markdown` directly, downstream of request validation. Nothing posted the blank body this change exists to permit.
4. **Round 4 changed what the consent boxes display, and the web transport's error handling.** Neither was in the plan. A failed context lookup now reads `(could not check)` rather than `(none found)`, and `WsResearchTransport.getFeedbackContext` throws instead of resolving an empty context, so the dialog can tell the two apart. The `.catch()` added in round 3 also changed production behaviour, by making a failed lookup silent, and was not listed as a deviation at the time; it is now.
5. **Six unconditional session-log claims were made conditional.** One was a blocker I introduced: `docs/alpha-user-guide-cowork.md` told Cowork testers the bundle already carries their prompt, while the same page's checkbox note says a Cowork session has no Claude Code session log to attach. `docs/alpha-feedback-guide.md` additionally claimed the project folder carries the prompt, which no `research.json` section does.

**Edit sites.** The issue's `**Touches:**` line names four files; this is 17 after round 4, which added `apps/web/src/transport/WsResearchTransport.ts` and `packages/engine/mcp-server/tests/packaging/windows-wrappers.test.ts`. Both are `.ts` and fall under the same CODEOWNERS rule as the rest, so the approval requirement is unchanged. The extra 11 are not scope creep, they are the sites that assert the old contract: `apps/electron/docs/feedback-json-spec.md` and `.claude/skills/compare-state/SKILL.md` both told triage the fields were required and to demand a resubmit; `.claude/skills/mine-unit-test/SKILL.md` carved out a blank `agent_should_have` but not a blank `agent_did`, so it would mine a rubric from a blank; both alpha user guides list the form's fields using the same optional-marking convention the fix relies on; and both markdown producers emitted the sections unconditionally.

**Collision:** re-derived at the current head, since two of the three moved. PR #1925 is still open and still edits `_feedback_markdown` in `apps/server/app/feedback.py` and `apps/server/tests/test_feedback.py`, in the living-redaction section rather than these hunks. PR #1905 has **merged** and is already in this branch; both spec rows survived the merge intact, so that collision is resolved rather than pending. PR #2005 is open and edits `docs/alpha-user-guide.md` and `docs/alpha-user-guide-cowork.md`; its hunks sit more than thirty lines from ours in both files, so no textual conflict is expected.

## Test plan

- [x] I ran `make test-all` (or `scripts/test.sh` - the same command) and it passed.
- [ ] For every skill whose **run-log snapshot** I changed, I ran `make eval-skill SKILL=<name>` and committed the run log and its `.ann.json`.

On the second box: the two `SKILL.md` files here are `/compare-state` and `/mine-unit-test`, repo-local dev skills under `.claude/skills/`. They are not part of the shipped plugin, have no `eval/tests/unit/<skill>/` directory and no run-log snapshot, so there is no paid eval run to make. Neither edit touches a `description` frontmatter or a DO NOT clause, so the third box has nothing to cover either. Both are inside `doc-links.test.ts`'s `LINTED_DIRS`, which `make test-all` runs.

Suite counts, re-measured on the current head rather than adjusted: viewer-ui 273 to 293, Electron 101 to 105, server 210 to 214, engine unchanged by this PR's viewer work. (The earlier endpoints, 271 to 286 and 97 to 101, were correct when taken and drifted as main moved under the branch; the deltas did not change.) `make test-all` now exits 0 on this checkout for the first time, which it did not before: see the Windows wrapper guard below. `make test-all`, a forced (uncached) `pnpm typecheck` and `make lint` were all re-run green after the review fixes.

One acceptance line I could not execute: the manual click-through in the running Electron app. I can boot it in WSL but cannot drive its GUI. What it would have checked is covered by the 23 component tests, including focus order and the `aria-invalid` wiring; the residue is purely how the two new CSS classes look, and both use `--badge-red-text`, which is defined in the light and dark blocks of `global.css`.

## Follow-on issues

**Folded in, because the work was smaller than the issue that would describe it.** Two items were headed for the backlog and were measured instead of assumed, which is what changed the call.

The **consent divergence**: the media and session-log checkboxes rendered from derived values and submitted the raw ones, so a failed context lookup displayed "(none found)" while both flags shipped `true` and each producer read the folder itself. Two remedies were on the table and neither is right. Gating the flags at the submit is unsafe as written, because `handleSend`'s dependency array lists neither `mediaCount` nor `hasSessionLog`: applied verbatim it typechecks, leaves the suite green, and silently drops the media and the session log. Correcting only the comment leaves the dialog telling the reporter the opposite of what the bundle does. The third option costs 32 lines across two files: stop conflating "checked, found nothing" with "could not check". The web transport no longer resolves an empty context on a non-OK response, and on a failed lookup both boxes stay ticked and enabled reading `(could not check)`, which is the truth, since the flags are submitted unchanged. Nothing is gated, so nothing is lost. Pinned by a test that asserts the display and the submitted payload agree on that path.

The **Windows wrapper guard**: `windows-wrappers.test.ts` walked `.claude`, which holds gitignored review worktrees, so `make test-all` was red from 37 copies of a `.bat` that are not in the checkout while CI, a fresh clone, stayed green. One skip entry. All 1,709 stray files are under that directory and no tracked `.bat` is, so the narrowing is exact; the guard still fails when the real `%~dp0`-after-`shift` regression is planted in the tracked script.

**Filed rather than folded:** `apps/electron/docs/feedback-json-spec.md` still frames itself as a contract between two repos that are now one. Six sites and roughly 60 lines across four sections, and section 8 needs a keep-or-delete call I should not make silently. Filed as issue **#1942**. That stale framing is what produced the wrong cost estimate in issue #1919 itself.

**Tracked elsewhere:** the setup scripts' handling of a blank `user_prompt` is issue **#1940**, which absorbed #1957 on 2026-08-27 and now carries both scripts and `test_setup_feedback_case.py` in its `Touches:` line. `docs/specs/feedback-case-spec.md` row 7 records the current behaviour rather than asserting a requirement nothing meets.

**Raised and deliberately not changed**, each with a reason. A blank email now overwrites the remembered address, which is introduced here, since the old send gate required a valid one: the obvious guard loses its own counter-case, because a reporter who cleared the box on purpose would find it silently repopulated, so it wants a product answer rather than a one-liner. `U+200B` is whitespace to neither `trim()` nor `strip()`, so both producers print an invisible character where the placeholder belongs. The server emits `## Session log` only when a log is present while Electron always emits it, which this PR's guide text made load-bearing. And after a refusal, breaking a second field renders three alert regions at once and reports the count in the singular.


chesworthrm (approved):
Approving. Checks are green, and I reproduced every number in your PR body — the four suite counts and the acceptance check both land exactly where you said. Restoring `origin/main`'s component under your tests does fail 7 of 11, and `keeps Send clickable with the radio unanswered` fails on `toBeEnabled()`, so it is failing on behaviour rather than on the renamed labels.

The edit-site list is the part I most expected to be short, and it isn't. Updating `/compare-state` and `/mine-unit-test` was the right call — without those, triage would still be telling people to resubmit a report this change makes legal.

## One change I'd like before merge

**The blocker list can drop a blocker.**

`blockers` keeps only the ids it finds in `FIELD_ORDER`. Anything not in that list is discarded — so a blocker that exists stops blocking.

You named this risk yourself in Review guidance. It is worth more than the note gives it. I removed each id from `FIELD_ORDER` one at a time and ran your suite. Four came back fully green: `feedback-prompt`, `feedback-did`, `feedback-should`, `feedback-answer`. Only email, the radio and notes are pinned.

With any of those four out of sync, a 10,001-character field submits. `submitFeedback` is called with the whole payload, nothing renders on screen, and the server 400s it. That is the same silent refusal #1919 is about, reachable from a one-line edit that CI calls green.

More tests are not the fix. Make the walk unable to drop anything:

You wrote:

```ts
    return FIELD_ORDER.flatMap((id) => {
      const b = byId.get(id)
      return b ? [b] : []
    })
```

Change it to:

```ts
    // Unknown ids sort last rather than vanishing: a FIELD_ORDER/JSX desync must
    // cost message ORDER, never a dropped blocker that lets an invalid send through.
    const rank = (id: string): number => {
      const i = (FIELD_ORDER as readonly string[]).indexOf(id)
      return i === -1 ? FIELD_ORDER.length : i
    }
    return [...byId.values()].sort((a, b) => rank(a.id) - rank(b.id))
```

I ran this. All 11 component tests pass, the full viewer-ui suite is 279/279, and `tsc --noEmit` is clean. Ordering is unchanged: I compared it against your version on four combinations, including the one where the radio goes into the map last but has to sort into third place, and focus lands on the same field every time. `Map` iterates in insertion order and `Array.sort` is stable, so the fallback ordering is sensible rather than arbitrary.

## Two notes, neither blocking

Your **Unsure about** is fair, and I checked it. Issue #1919 has no comments at all, so there is no lead confirmation recorded anywhere. Implementing the recommendation as written was the right default. Worth getting the answer written onto the issue either way, since it is what decides whether the radio stays required.

Your correction to the issue's cost estimate is also right. Both producers are in this repo now, so there is no cross-repo coordination to pay for. Filing #1942 for the stale spec framing was the correct call rather than fixing it here — section 8 needs a keep-or-delete decision, not a silent one.

## What I did not check

The manual click-through in the running Electron app, which is the same gap you declared. What it would have covered is how `.required` and `.fieldError` render; both use `--badge-red-text`, which is defined in the light and dark blocks of `global.css`. No eval or e2e run, since no plugin skill snapshot changed — your reasoning on the second and third test-plan boxes is correct.

Asimi1234:
Reviewed the full diff against the merge with `main` (0 commits behind). No blocking findings — this is ready on the developer side.

**Verified in a fresh worktree, not just read:**
- viewer-ui **281/281**, Electron `feedback.test.ts` **27/27**, server `test_feedback.py` **18/18**, viewer-ui `tsc --noEmit` clean.
- Watched the guards fail: re-disabling Send on `blockers.length > 0` (the exact #1919 defect) fails **7 tests**, including `keeps Send clickable with the radio unanswered`. So the tests genuinely catch the regression, not just the labels.
- `--badge-red-text` is defined in both the light (`global.css:51`) and dark (`:112`) blocks, so the two new classes render in both themes.
- Follow-on #1942 is filed.

**On the prior approval:** it read `ab14be5d`; head is now `fa95a5dc`. The "blocker list can drop a blocker" change requested there is already resolved — `9c284aaf` is exactly that `rank`/sort-unknown-last fix, landed after the approval. Nothing raised in review is still open in the current code.

**One non-code note:** issue #1919 ends with "Recommendation, for the lead to confirm" and has no comments recorded. The recommendation was implemented as written (radio required, three fields relaxed) — the right default — but the lead's sign-off isn't on record anywhere, and that's the decision that keeps the radio required vs. fully optional. Worth writing the answer onto #1919.

**Not checked:** the manual Electron GUI click-through (no headless way to drive it; the 11 component tests cover focus order and `aria-invalid`). Still needs `senior-genealogists` approval per CODEOWNERS (both senior teams required on the `.py`/`.tsx` paths).

chesworthrm (changes requested):
Re-reviewing at `fa95a5dc`. My earlier approval read `ab14be5d`, before the blocker-list rewrite, so this replaces it.

The change is in good shape. I re-ran everything in a worktree of the merge: viewer-ui 281/281, Electron 101/101, server 214 passed 1 skipped, engine 2718/2718, `make lint` green. I also checked the paths most likely to break and they hold — the unmounted-blocker case (picking Yes clears both hidden fields, so an over-limit value can't block Send forever), the trim/length boundary across the dialog and both validators, and both transports, which add no validation of their own.

One thing to fix before merge.

## A field refused once goes red again with no Send in between

`refusedIds` remembers which fields the last refusal named, and fixing a field never removes it from that set. So re-breaking the same field brings the whole refusal UI back with no click on Send.

The likeliest way to hit it is the ordinary one. Type a bad email, get refused, clear it, start retyping. `a`, `a@`, `a@b`, `a@b.` are all invalid, so the field sits red and the footer says "Not sent" the whole way through, reporting a refusal that never happened.

It is not email-specific. An over-limit Notes field that gets fixed and re-broken does the same.

This is the same defect `9c284aaf` fixed. That commit stopped a *new* field going red after a refusal; a field that was already refused once still does.

To reproduce: answer the radio, type `nope` in the email, click Send, clear the field, type one character. `aria-invalid` is back and so is the "Not sent" footer.

In `packages/viewer-ui/src/components/shared/FeedbackDialog.tsx`, insert above `shownBlockers`:

You wrote:

```tsx
  // Still-unfixed blockers from the last refusal. A blocker created after that
  // refusal is not here, so it stays quiet until the reporter tries again.
  const shownBlockers = blockers.filter((b) => refusedIds.has(b.id))
```

Change it to:

```tsx
  // A blocker the reporter has since FIXED leaves the refused set for good: if
  // its id stayed, re-breaking that same field would go red and claim "Not sent"
  // again with no Send in between. Only a fresh refusal may re-add it.
  useEffect(() => {
    setRefusedIds((prev) => {
      const next = new Set([...prev].filter((id) => blockers.some((b) => b.id === id)))
      return next.size === prev.size ? prev : next
    })
  }, [blockers])

  // Still-unfixed blockers from the last refusal. A blocker created after that
  // refusal is not here, so it stays quiet until the reporter tries again.
  const shownBlockers = blockers.filter((b) => refusedIds.has(b.id))
```

`useEffect` is already imported. I ran this: all 13 component tests pass, the full viewer-ui suite is 281/281 and `tsc --noEmit` is clean. I also checked the two ways it could regress, and neither does — a genuine second refusal on the same field still reports and still takes focus, and a refusal you never fix keeps showing while you type elsewhere.

Please add a test with it, shaped like `does not go red for a field the reporter starts typing after a refusal` but refusing the email first. That existing test passes today only because it uses a field that was never refused.

## Two small things while you are in there

The email's `aria-invalid` is not pinned by anything. I deleted that one prop and all 13 tests stayed green — the suite asserts its absence twice and its presence only on Notes. The visible message still announces, so the impact is small, but it is the field most likely to be refused and it sits just outside the eight mutations you proved. The test above can cover it.

The acceptance check numbers in the PR body are stale. Restoring `origin/main`'s component now fails 9 of 13, not 7 of 11 — correct when you wrote it at `ab14be5d`, which had 11 tests, and overtaken when the second commit added two. The check itself still holds, and `keeps Send clickable with the radio unanswered` does fail on `toBeEnabled()`, so it is failing on behaviour.

## One process note, not about your diff

The comment above says this still needs `senior-genealogists` approval because both teams are required. That is not right. CODEOWNERS lists the two senior teams as an OR — one approval from either satisfies the rule — and the pending review request on `senior-genealogists` is just GitHub's auto-request from when the PR opened, not evidence anyone is waiting on a genealogist. I checked it against a merged precedent rather than the file's header: PR #1916 merged with `.py` files and approvals from two senior developers and no genealogist. So once this is fixed, any second approver with write access clears it.

clack391:
All three addressed at `51fc2e48`, plus what fell out of re-reviewing the delta.

## The refused-once defect

Your fix applied verbatim. I wrote your reproduction as a test **before** applying it and watched it fail, so the test is anchored to the defect rather than to the fix: `does not go red again for a field that was refused, then fixed, then re-broken` (answer the radio, type `nope`, Send, clear, type one character). It asserts `aria-invalid` present after the refusal and absent after the re-break, so it closes the email pinning in the same pass. Reverting the `useEffect` now fails exactly that test.

Your two regression checks both hold, but only one of them was pinned. "A genuine second refusal still reports and takes focus" was covered. **"A refusal you never fix keeps showing while you type elsewhere" was not**: I broke your fix a third way you did not name, replacing the filter with `new Set<string>()` so the whole refused set is released on any edit, and the suite stayed green at 14 of 14. That silently drops a message the reporter still needs. There is now a test for it and that mutation fails.

## Acceptance numbers

Corrected, and re-derived at every head since rather than adjusted arithmetically. It is now **14 of 18**. I first wrote 13 from memory and measured 14 before it shipped, which is the same mistake in miniature that made the number stale in the first place.

## Your CODEOWNERS note

Verified independently and you are right. Both senior teams sit on one rule, which is an OR, and your precedent holds: PR #1916 merged with five `.py`/`.tsx` files and approvals from two senior developers, no genealogist. So nobody is waiting on `senior-genealogists`. Flagging it only so the earlier "both teams required" line does not send someone hunting for a genealogist approver.

## What re-reviewing the delta turned up

Three blind passes over the round-2 code found 11 more. The worst was mine: this branch **shipped a spec MUST that nothing implements**. `docs/specs/feedback-case-spec.md` row 7 said the setup script MUST report a blank prompt and name the session log. It does not. A pass built a blank-prompt bundle and ran the real script, which prints verbatim what the MUST forbids, and the `.bat` is worse, printing the heading and both rules with nothing between them. Row 7 now describes the actual behaviour, and issue #1957 carries the work with `nothing-checks`. I did not implement it here because the Windows half is unverifiable from this environment and shipping only the `.sh` would leave the worse platform broken under a spec saying it was handled. PR #1905 also owns both scripts.

Five behaviours were unpinned, each now mutation-proven: the blocker sort (the existing cursor test's two cases are ordered identically by plain `Map` insertion order, so it never exercised the sort, and its comment said otherwise), `scrollIntoView`, the double-message suppression (a `>0` assertion could not see a doubled message), `aria-describedby` on five of seven controls, and `role="alert"` on the upload-failure toast. That last one is worth calling out: a refusal is recoverable and was announced, while the failure that actually loses the report was silent.

One method note that may matter to you, since you and I both ran mutations here: **vitest's transform cache produced false passes.** One pass reported a guard missing that is in fact present. Everything above was re-run with `--no-cache`.

Also fixed: an unhandled rejection on the context lookup, a docstring I had corrected in the PR body two commits earlier and never in the code, and two prose overstatements about the session log, which is filtered to `user`/`assistant` turns and scoped to one project folder rather than being the whole transcript.

## Not fixed, deliberately

One is worth your eye. The media and session-log checkboxes render from derived values but submit the raw state, so when the context lookup returns empty they display **unticked while the bundle uploads both**, against the dialog's own privacy paragraph. It is pre-existing and outside this issue, and the fix is a product call (send what the user was shown, or repair the lookup), so I raised it rather than folding it in. The over-length email gate, the `trim`/`strip` divergence on `U+0085` and `U+001C`-`U+001F`, and five channels missing an input-validation row in the `security-invariants` checklist are likewise pre-existing and reported rather than changed.

I also ran that checklist, which neither of our reviews had: 10/10, and this diff does not touch that surface.

## Still open

@Asimi1234's non-code note is the one thing neither of us can settle. Issue #1919 still has zero comments, so the recommendation this implements has no recorded sign-off. That is the decision that keeps the radio required rather than optional, and it is worth having written down before this merges.

`make test-all` green, viewer-ui 271 to 286, Electron 101, server 214, engine 2718 unchanged.

chesworthrm (approved):
Approving at `51fc2e48`. This supersedes my `CHANGES_REQUESTED` at `fa95a5dc` — all three items are in, and each one is now pinned by a test that fails without it.

Re-verified in a worktree test-merged with `origin/main` (2 behind; the merge is clean and touches only engine files this PR doesn't):

- viewer-ui **286/286**, Electron **101/101**, server **214 passed / 1 skipped**, `turbo typecheck` clean, `make lint` exit 0, and engine `doc-links.test.ts` **134/134** — the lint that actually covers your two `SKILL.md` edits.
- Your acceptance check reproduces exactly: restoring `origin/main`'s component fails **14 of 18**, first failure `refuses an over-limit field without ever disabling Send` on `toBeEnabled()`. Behaviour, not labels. The number is right this time and I re-derived it rather than checking your arithmetic.
- **12 mutations, all caught.** Removing the release effect; dropping the `sendState`/`errorMsg` reset on refusal; restoring the `blockers.length > 0` disable (12 fail); stripping `role="alert"` from `FieldError`; making a blank email a blocker (9 fail); removing the render-order sort; weakening `EMAIL_RE` to `/@/`; deleting the email's `aria-invalid`; `orBlank` → identity and a `NOT_PROVIDED` drift in the Electron producer; `_or_blank` → identity and a non-empty `userPrompt` requirement on the server. Nothing here is coverage-shaped-but-hollow.

Two things I checked because they are now stated in five files that testers and triagers read, and both hold: `capSessionLog` really does walk backwards from the end, keep the newest entries and prepend `_truncation_note` with `dropped_leading_entries`; and "oversize bundles trimmed largest-first" is right in both producers.

I also traced the widened contract through all 19 files that mention `user_prompt` / `agent_did` / `email`, not just the 15 you changed. The ones you left alone are genuinely unaffected — the Apps Script already does `payload.email || 'unknown'`, and `feedback-endpoint/README.md` names those fields only in a redaction check. The setup scripts and their test are the real gap and #1957 has them.

Your call to file #1957 rather than fold it in was right, and finding that you had shipped a spec MUST nothing implements — and then saying so plainly — is the best thing in this PR.

## One follow-up to file, not a code change

The checkbox divergence you raised in your last comment needs an issue. A PR comment isn't one of the four places found-work can live, so it disappears the moment this squashes. The behaviour is real:

```
FeedbackDialog.tsx:519   checked={includeSessionLog && hasSessionLog}
FeedbackDialog.tsx:249   includeSessionLog,
```

Displayed from the derived value, submitted from the raw one, which defaults to `true` — so on an empty context lookup the user sees "unticked, none found" and the bundle still asks for the log, against the dialog's own privacy paragraph.

Label it `developer` and `nothing-checks`: nothing asserts that what those checkboxes display is what gets submitted, so CI stays green while the privacy text is wrong.

One correction in your favour, worth putting in the issue body so nobody bisects it to this PR: your new `.catch()` does **not** widen this. A rejected `.then` never set the state on `main` either, so production reachability is unchanged — the `.catch()` only removes the console error that used to be the single signal.

## Optional, and I would merge without it

`FeedbackDialog.tsx:141` — the over-limit list covers five fields; both validators cover six. Email is the one field that can exceed 10,000 characters, pass the dialog, and be rejected downstream, where the web message names no field and the dialog highlights nothing. Rare (it needs a paste into the email box) and the report isn't lost. If you want it:

You wrote:
```ts
    const fields: Array<[string, string, string]> = [
      ['feedback-prompt', 'What you asked', userPrompt],
```
Change it to:
```ts
    const fields: Array<[string, string, string]> = [
      ['feedback-email', 'Your email', email],
      ['feedback-prompt', 'What you asked', userPrompt],
```
and add `email` to that memo's dependency array. `blockers` already prefers the malformed-email message for the same id via `byId.set`, so ordering needs no change.

## Your Unsure about

Settled, and it was mine to settle rather than yours to chase. I've written the decision onto #1919: the radio stays required, for that issue's third acceptance criterion, not for the coordination cost you correctly showed was miscosted.

## What I did not check

The engine suite (2718) — this PR touches no engine file, so on the merged tree that suite tests `main`, not your branch; I ran `doc-links.test.ts` instead, which is the part that covers your diff. The manual Electron click-through, same gap you declared; the residue is how `.required` and `.fieldError` render, and both use `--badge-red-text`, defined in the light and dark blocks of `global.css`. No eval or e2e run, correctly. And I took your `security-invariants` 10/10 and the `trim`/`strip` divergence on `U+0085` and `U+001C`–`U+001F` on your report rather than re-deriving them; neither bears on merge.

Your `.catch()` has no test. I chased it rather than flagging it: `includeSessionLog` is submitted from raw state, so a swallowed rejection cannot lose the session log — it only mis-renders the summary, which is the follow-up above.

Once the issue is filed, any second approver with write access clears this — one approval from either senior team satisfies the CODEOWNERS rule, and nobody is waiting on `senior-genealogists`.

francis-2008-happy (changes requested):
Good PR, and the test suite is unusually strong — I ran 26 mutations against the three production files and 22 went red. Restoring `disabled={blockers.length > 0 || …}` on the send button kills 12 of 18 tests, so the headline behaviour is properly pinned, and the "(required)" marking lands inside the radiogroup's `aria-labelledby` so it reaches the accessible name rather than just being painted. The contract surface is clean on both producers, with `NOT_PROVIDED` pinned as a literal on each side so drift in either fails a test. That is the part this repo has been bitten on before and it is genuinely covered.

Three things before I approve. None is a bug in the shipped behaviour — one is a test that does not pin what it says, one is a comment that is false on a consent-relevant path, and one is a stale issue number.

## 1. The anti-latch test does not catch the defect class

`FeedbackDialog.test.tsx:183` — `does not go red for a field the reporter starts typing after a refusal`.

I replaced the id-keying with a self-clearing latch:

```
You wrote:      const shownBlockers = blockers.filter((b) => refusedIds.has(b.id))
Mutation:       const shownBlockers = refusedIds.size > 0 ? blockers : []
```

**All 18 tests stayed green.** Under that mutation a reporter refused on the radio who then starts typing an email gets `aria-invalid` and "does not look like an email address" — a refusal that never happened, which is the bug `9c284aaf` says it fixed.

The reason it misses: the test clicks Yes at line 189 *before* typing the email. That empties `blockers`, which makes the release effect empty `refusedIds`, so the assertion is satisfied by the release effect rather than by the id-keying. The comment above it claims the stronger property.

Drop the fix-first step and it discriminates. I ran this both ways — passes on HEAD, fails under the mutation on `aria-invalid="true"`:

```tsx
it('does not go red for an untouched field while another is still refused', async () => {
  const user = userEvent.setup()
  mount()
  await user.click(screen.getByRole('button', { name: 'Send' }))
  expect(screen.getAllByText(/Choose Yes or No/).length).toBeGreaterThan(0)
  // The radio is STILL refused. Nothing has been refused about the email, so
  // nothing may claim it was — this is what a latched flag gets wrong, and
  // fixing the radio first hides it behind the release effect.
  await user.clear(screen.getByLabelText(/Your email/))
  await user.type(screen.getByLabelText(/Your email/), 'a')
  expect(screen.getByLabelText(/Your email/)).not.toHaveAttribute('aria-invalid')
  expect(screen.queryByText(/does not look like an email address/)).toBeNull()
})
```

Keep the existing one too — it pins the release effect, which is worth pinning; it just is not the anti-latch guard.

## 2. The catch comment is false, and the path it hides is a consent one

`FeedbackDialog.tsx:114-118`:

```
You wrote:      // Leave the counts at zero. The submission still works; only the
                // "what gets sent" summary is unavailable.
```

It is not only the summary. With `files: []` and `hasSessionLog: false`, both consent checkboxes render disabled reading `(none in folder)` and `(none found)` — while `handleSend` still posts `includeMedia` and `includeSessionLog` from state, both defaulting `true`, and each producer walks the project folder and reads the transcript itself (`feedback.ts:378`, and `index.ts:133` defaults the flag to true when absent). The dialog tells the reporter no media and no session log will be sent. Both are sent.

I want to be fair about provenance: **this gap is pre-existing.** Before this PR the call was `void getFeedbackContext().then(…)` with no catch, and the same wrong display and same flags applied. What changes here is that the unhandled rejection was the last trace of it, and the comment says so outright — "used to escape and fail the whole render in tests". Deleting the catch entirely leaves 18/18 green, so nothing exercises the rejection path either way.

It matters more now because this PR's own doc change tells reporters to act on that display: `docs/alpha-user-guide.md` — "if the form says no log was found". A false "(none found)" becomes misleading instruction.

Either gate the flags:

```
You wrote:      includeMedia,
                includeSessionLog,
Change it to:   includeMedia: includeMedia && mediaCount > 0,
                includeSessionLog: includeSessionLog && hasSessionLog,
```

or, if you would rather keep the behaviour and fix only the claim:

```
Change it to:   // Leave the counts at zero. The submission still works, but the
                // media and session-log toggles then UNDER-REPORT what the bundle
                // will contain: both flags stay true and each producer reads the
                // folder itself, so a "(none found)" here does not mean nothing
                // is sent. Pre-existing; the rejection used to be the only trace.
```

I would take the first — the second leaves a reporter told one thing while another happens. Either way a test with a rejecting `getFeedbackContext` would pin it, and there is none today.

## 3. `#1957` is closed

The Follow-on section says "Filed rather than folded, a second item: #1957". That issue is closed — chesworthrm merged it into **#1940**, which is open and now carries the `.bat`, the `.sh` and `test_setup_feedback_case.py`. Anyone following the number lands on a closed issue and reads the deferral as dropped. Repoint it and note the merge. (#1942 is open and correct.)

## Smaller, take or leave

- `FeedbackDialog.module.css:231` — the comment says "the one field whose 'when do I fill this in?' isn't obvious". This PR added a second `.fieldHint` (the email hint), so it is two now.
- `FeedbackDialog.tsx:243` — a blank email now writes `''` over the remembered address, which one anonymous submission makes sticky for that machine. Arguably correct; flagging in case it is not intended.
- The setup-script gap for an empty `user_prompt` is disclosed in the spec, the guide and both skills, so I am reading it as a stated deferral rather than a miss. Two `if`-branches in two scripts if you would rather close it here.

Also mechanical: the branch is `BEHIND` and needs a merge from main before the button lights up.

clack391:
All six items addressed at `e5bd4bec`, and one of them I did not fix the way you asked. That one is worth reading first, because the reason is a measurement rather than a preference.

## 2. The catch comment, and the gate you prescribed

Your diagnosis is right and the provenance note is right. I ran your prescribed gate before adopting it, and it does not do what it looks like it does:

```
includeMedia: includeMedia && mediaCount > 0,
includeSessionLog: includeSessionLog && hasSessionLog,
```

Applied verbatim, `tsc --noEmit` is clean and the suite is fully green, **and the bundle loses the media and the session log.** `handleSend` is a `useCallback` whose dependency array (`FeedbackDialog.tsx:264-278`) lists neither `mediaCount` nor `hasSessionLog`, so when the folder walk is slow and the reporter answers the required radio before it resolves, the closure keeps the mount-time zeros. A probe that first asserts the dialog is no longer showing `(none found)` and then reads the payload prints:

```
SUBMITTED FLAGS: {"includeMedia":false,"includeSessionLog":false, ...}
```

Adding `mediaCount` and `hasSessionLog` to that dependency array flips the same probe to `true`/`true`, which is the control that makes this a cause rather than a story. Nothing here would have caught it: `packages/viewer-ui` has no eslint config at all, and `make lint` covers only `apps/electron` and `eval/app`, so `react-hooks/exhaustive-deps` never runs on this file.

Your option B I did not take either, for the reason you gave against it.

So there is a third option, and it is what shipped. The real defect is that `files: []` plus `hasSessionLog: false` means two different things: "checked, found nothing" and "could not check". Separate them and the divergence disappears with no gating and no data loss. On a failed lookup both boxes now stay ticked and enabled reading `(could not check)`, which is the truth, since the flags are submitted unchanged and each producer walks the folder itself. There is a test that asserts the display and the submitted payload agree on that path.

One more thing on reachability, which changes where the fix belongs. The route is not a rejection at all on the hosted path: `apps/web/src/transport/WsResearchTransport.ts:91` **resolved** an empty context on any non-OK response, so no `.catch()` was ever on that path. That line now throws. It is also why @chesworthrm's "your `.catch()` does not widen this" was correct, for a narrower reason than either of us gave.

I also checked your provenance claim the same way and it holds: the merge-base shows the same wrong display, and a rejecting transport there really did fail the vitest file, so the catch's own stated reason checks out. Only the second sentence of that comment was false.

## 1. The anti-latch test

Reproduced exactly. Your mutation leaves 18 of 18 green, and your mechanism is right: the test answers the radio at `:189` before typing, which empties `blockers`, so the release effect satisfies the assertion instead of the id-keying.

Your replacement test is in, verified both ways. But it is not sufficient on its own, and the shape that gets past it matters:

```ts
refusedIds.has(b.id) || (refusedIds.size > 0 && overLimitFields.some((f) => f.id === b.id))
```

A partial latch that leaks only the over-limit kind survives your test 19 of 19, because your test breaks the email, which is a format blocker. So there is a second test that is agnostic to the kind of breakage: refuse on the radio, then push Notes over the limit. It fails on both shapes.

**And chasing that turned up the bigger one, which is the same defect class as @chesworthrm's round 1 blocker.** Four of the seven `<FieldError>` render sites, prompt, did, should and answer, can each be deleted with the whole suite green. Deleting one reproduces #1919 exactly: nothing in the repo's CSS keys on `aria-invalid` (a repo-wide grep returns zero, and I ran it against a known positive first to prove it reaches `.css` files), and `untoldOverLimit` suppresses the live toast for a refused id, so the footer says "Fix the highlighted field above" with nothing highlighted and a dangling `aria-describedby`.

That is round 1's defect relocated. The `FIELD_ORDER` fix made the walk unable to drop a blocker, and the same risk moved into the JSX, where nobody re-checked it. Rather than a fifth per-field test, there is now one guard over the invariant the code already states: a refusal of six blockers asserts every one carries `aria-invalid` and its own error node. It fails on each of the four deletions individually, and it also fails when only the first blocker of a refusal is marked, which was separately unguarded.

## 3. #1957

Correct, and thank you. Repointed to #1940 with a note that it absorbed #1957 on 2026-08-27 and now carries both scripts and `test_setup_feedback_case.py`. I also removed the "Deferred without an issue" paragraph that described the same line #1940 now owns, and the stale #1905 collision, which has merged and is already in this branch.

## The smaller ones

The `.fieldHint` comment is fixed, and you are right that this branch added the second one. That is the third time my doc sweep has keyed on what the issue named rather than on what my own diff changed.

**The blank email overwriting the remembered address is introduced here**, not pre-existing, and I established that rather than assuming: on the merge-base the send button was disabled unless the address was valid, so `''` could never reach the `setItem`. I have not changed it. The obvious guard, `if (trimmedEmail) setItem(...)`, loses its own counter-case, since a reporter who cleared the box deliberately would find it silently repopulated next time and could submit non-anonymously by accident. It wants a product answer, not a one-liner, so it is recorded in the PR body rather than quietly patched.

On the setup scripts, agreed, and it stays a stated deferral: #1940 owns it and needs a Windows box.

The branch is no longer behind. Thank you for merging main.

## Two notes for @chesworthrm

Your optional email gate is in, and it was worth more than optional: it was the last field that could pass the dialog and be rejected downstream, where the web path names no field. One correction, because I tested it rather than reading it. You wrote that `blockers` already prefers the malformed-email message for the same id via `byId.set`. It is the other way round: the over-limit loop runs second and overwrites the malformed message, so an over-limit malformed address reports the length. That is the more actionable message of the two, so the patch is right and only the reason was inverted.

Your follow-up condition is discharged by fixing rather than filing. I want to flag that explicitly instead of quietly reversing you: you ruled "file it, not a code change" on the reasonable assumption that the fix was the gate above, and that gate is unsafe. The version that shipped is 32 lines across two files and needs no product call, so it did not carry the four-person overhead an issue costs. If you would still rather it were tracked and reverted here, say so and I will do that instead.

## Suite

`make test-all` now exits 0 on my checkout, which it did not before this commit, and the cause was not this PR. `windows-wrappers.test.ts` walked `.claude`, which holds gitignored review worktrees, so it reported 37 copies of a `.bat` that are not in the checkout. CI is a fresh clone, so it stayed green there and only local runs went red. One skip entry, and I confirmed the narrowing is exact: all 1,709 stray files are under that directory, no tracked `.bat` is, and the guard still fails when the real `%~dp0`-after-`shift` regression is planted in the tracked script.

Everything green at `e5bd4bec`: viewer-ui 293, Electron 105, server 214 with 1 skipped, engine 2726, harness 168, Python 2876 with 2 skipped, `turbo typecheck` 5 of 5. Every new test was watched to fail against the code it guards, all runs with `--no-cache`, since this file has produced false passes from the transform cache before.

francis-2008-happy (approved):
Approving `e5bd4bec`. All three of my items are addressed, and the one you refused you were right to refuse — I verified that before anything else, because it was my prescription that was wrong.

**My gate would have dropped the data it was meant to protect.** `handleSend`'s `useCallback` dependency array lists neither `mediaCount` nor `hasSessionLog`, so `includeMedia && mediaCount > 0` reads whatever those were at the last recreation. Answer the radio while the folder walk is still in flight and the closure keeps the mount-time zeros — flags submitted false, media and transcript gone, `tsc` clean and the suite green. Your control is the right one: adding the two to the array flips it back, which makes it a cause rather than a story. I should have run it before prescribing it.

**The third option is better than either of mine.** Separating "checked, found nothing" from "could not check" removes the divergence instead of papering it, and `(could not check)` with the boxes left ticked and enabled is the honest display, since the flags go out unchanged and each producer walks the folder itself. Fixing `WsResearchTransport.ts` to throw rather than resolve empty is the part I would have missed entirely — I had the reachability wrong in the same direction chesworthrm did, and your narrower reason is the correct one.

I re-ran everything rather than reading it:

| Mutation | Result |
|---|---|
| my latch (`refusedIds.size > 0 ? blockers : []`) | fails |
| your partial latch (leaks only the over-limit kind) | fails — `marks only the fields the refusal named` |
| delete any one of the four `<FieldError>` sites | fails — `gives every refused field its own inline message` |
| `%~dp0` planted after a real `shift` in a tracked `.bat` | fails — guard still bites after the `.claude` narrowing |

On that last one: I checked the narrowing rather than taking it, because excluding a directory from a guard's walk is how coverage disappears quietly. `git ls-files` returns **zero** tracked `.bat` under `.claude/`, `.gitignore:25` ignores the whole directory bar `commands/` and `agents/`, and the guard still fires on the real regression. The narrowing is exact. Worth saying plainly that those worktrees are mine — my review flow put them there, so that red was my mess and you cleaned it up.

Two notes, neither blocking.

**Your `<FieldError>` finding is the better catch of this round.** Four deletable render sites with the suite green, reproducing #1919 exactly, is a bigger hole than the anti-latch test I came in with — and the framing is the part worth keeping: round 1's fix made the *walk* unable to drop a blocker, and the same risk moved into the JSX where nobody re-checked. One invariant guard over all six beats a fifth per-field test.

**On the blank email overwriting the remembered address** — agreed it wants a product answer, and recording it beats a one-liner that trades one silent failure for another. Your counter-case is real: a reporter who cleared the box deliberately finding it repopulated is worse than the thing the guard fixes.

`mergeStateStatus` is `BEHIND` again — main moved 38 commits since your comment, so it needs one more merge before the button lights up. Nothing else outstanding from me.

Verified locally at `e5bd4bec`: viewer-ui 293/293, server 214 passed 1 skipped, windows-wrappers 19/19. All CI green.

PR #1943, opened 2026-08-26
person-warnings: reconcile spec with the 74 shipped tags + drift lint (#1885)

### LINKED ISSUE #1885: person-warnings spec documents 3 warnings the tool does not emit; the tool emits 76 tags

Asimi1234:
Closes #1885. Tier: Normal.

> **Rebased onto `main`** now that #1896 (`1616-check-warnings-doctrine-and-deep-dive`)
> has merged — it added the W3 `SUPERSEDED` banner and the "How `hasEventAfterDeath1`
> actually decides" section this PR builds on. Base is `main` and the diff is exactly the
> four person-warnings files; the interim #1896/#1616 commits are gone (they landed via the
> #1896 squash). `person-warnings-tool-spec.md` reconciled conflict-free as predicted.

## Summary

- The spec's § Warning Definitions documented **3 warnings** (`DEATH_BEFORE_BIRTH`,
  `YOUNG_BIRTH`, `EVENT_AFTER_DEATH`) under `UPPER_SNAKE_CASE` names the tool never emits;
  the tool emits **74 FamilySearch-style tags**. This reconciles the spec to what ships and
  adds a lint so the two can't drift again silently.
- **Part A** — every stale section fixed against `src/types/person-warnings.ts` or a grep:
  the `warningId` field (→ `scoreType`/`issueType`/`severity`/…/`mobRole?`), the three
  `extract*Year` helpers that don't exist (→ `getStandardDate`/`stdDate`/`getDayRange` day
  ranges), the settled OPEN/PENDING/Flag-for-sign-off blocks, the How-to-add procedure, and
  the Future-candidates severities/struck rows. Plus the wrong "plus `mobRole`" clause in
  `match-merge-workflow-spec.md` (`mobRole` is already on `PersonWarning`).
- **Part B** — a 74-tag § Tag Catalogue (tag, severity, rule, cause), an exported
  `ALL_WARNING_TAGS` in the tool, and a **bidirectional drift lint**
  (`person-warnings-spec-drift.test.ts`) that keeps catalogue ⟷ registry ⟷ emit-sites in
  exact agreement.

## Review guidance

**Start here:** `tests/packaging/person-warnings-spec-drift.test.ts` — the lint is
three-way on purpose. `ALL_WARNING_TAGS` is hand-maintained, so the lint first proves the
array equals the tags actually emitted at `issueType:` sites (parsed from source), *then*
compares that against the § Tag Catalogue in both directions. Modeled on
`record-type-group-drift.test.ts`, but it parses the tool source for the emit set rather
than importing a table, because the tags are non-exported `const`s. Second: the catalogue
rows themselves — **no check can verify a definition is *accurate***, only that a tag is
documented-and-emitted. Rules were copied from
`check-warnings/references/warning-checks.md` (50 tags) where it had reviewed prose;
the **14 tags below have no prose anywhere** and were derived from the predicate +
emitter `message`, so they're the ones to read against `warnings.java`, not skim:
`birthLikeRangeGreaterThan8`, `birthRangeGreaterThan3`,
`hasCloseChildBirthsIgnoreSimilarChildren`, `hasCloseChildChristenings6_30`,
`hasDissimilarSpousesWithSameMarriageYear`, `hasEventsOutsideLifespanFar`,
`hasEventsOutsideLifespanNear`, `hasSameCensus`,
`missingGivenNamesWithoutExactBirthLikeDate`, `missingSurnames`, `similarChildren`,
`similarChildrenConflictingDates`, `similarSpouses`, `similarSpousesConflictingDates`.

**W1/W2 — resolved to strict symmetry with W3.** They now carry the full retained original
narrative (Severity / Condition / Logic / Message / factIds) under the `SUPERSEDED` banner,
matching how W3 keeps its original detail, with the banner routing shipped behaviour to the
§ Tag Catalogue. Their old pseudocode still names the now-removed `extract*Year` helpers —
kept verbatim and flagged as such, exactly as W3's retained pseudocode does, since both are
records of a design that never shipped rather than current behaviour. I did **not** add a
bespoke "How `<tag>` actually decides" shipped section for W1/W2: W3's exists only because
three pieces of skill doctrine were written against its exclusion-list model, which W1/W2
have no counterpart to, and inventing shipped prose would trip the "no check can verify a
definition is *accurate*" hazard noted above.

## Plan

**Didn't change:** `check-warnings/references/warning-checks.md` — its opening claim
"Complete catalog…" is an overstatement (it names 50 of 74 tags), but it was **ruled out of
scope 2026-08-25**: touching anything under `skills/` re-arms `check-runlogs` and buys a
paid `make eval-skill SKILL=check-warnings` run (~$1.76, ~9-10 min) for one word. Instead
the spec now states it is the full catalogue and the reference is a curated subset, so the
next check-warnings PR can drop "Complete" while riding a run it already pays for.

**Acceptance check:** `person-warnings-spec-drift.test.ts` fails on the pre-change tree and
passes here. Proven to fail **both directions** (CLAUDE.md § "A new lint must be proven to
fail" — a camelCase-tag grep is one of the three silent-pass shapes it names):

- *Rename a shipped tag value* (`hasEventAfterDeath1` → `hasEventAfterDeathX` in the tool) →
  "documents every shipped tag" fails (`hasEventAfterDeathX` undocumented) **and** "names no
  catalogue tag the tool does not emit" fails (`hasEventAfterDeath1` orphaned).
- *Delete a catalogue row* (`hasSameCensus`) → the `== 74` parser-guard assertion fails
  **and** "documents every shipped tag" fails.

**Deviated from the plan:** Nothing material. One correction folded in from plan-critic:
`getDayRange` lives in `src/utils/date-helpers.ts` (not `date-comparison.ts`, which only
re-exports it) — the doc-links lint checks the path but not the symbol, so a wrong file
would have passed while being false.

## Test plan

- [x] I ran `make test-all` and it passed. (ESLint, typecheck, JS workspace 225/10/97,
      control-plane pytest 209, MCP engine vitest **2707**, eval app 168, eval harness pytest
      2804 — `All checks passed`.)
- [ ] No skill run-log snapshot changed — no `eval-skill` run required. The one runtime
      change (the `hasEventBeforeBirth365_2` warning-message string, d7120858) is tool output,
      not a skill snapshot; see the row-by-row walk note below.
- [ ] No skill `description`/DO-NOT frontmatter changed.

The two doc lints that fail a naive draft were satisfied deliberately: citations are by
**symbol** not `file:line` (`doc-links.test.ts`), and the change is named rather than
issue-numbered in the spec body (`no-issue-refs.test.ts`, baseline 0 for this file).

## Follow-on issues

**Folded in / filed instead:** Folded in the plan-critic `getDayRange` path fix. Found the
`warning-checks.md` "Complete" overstatement and **deliberately left it** (out of scope per
the 2026-08-25 ruling above); nothing filed.

## Row-by-row Rule-cell walk (2026-09-01, addressing @johnmarkpeterbrown)

Per the review's main ask, I read **every one of the 74 § Tag Catalogue `Rule`
cells against its predicate function** in `src/tools/person-warnings.ts` —
tracing each through the date-helper it calls (`factDaysDiff*`,
`earliest/latestYearOf*Facts`, `perfectDaysOfSelfFacts`, the
overlapping/conflicting helpers in `date-comparison.ts`) so the earliest/latest
direction and the fact-family vs exact-type distinction were verified, not
skimmed. The "14 rows with no prior prose" triage in **Review guidance** above
no longer holds — as the review noted, `hasAgeRangeGreaterThan120` had reviewed
prose and was still wrong — so I dropped the triage and walked all 74.

Corrections made this round (7 rows + 3 prose sites):

- `hasAgeRangeGreaterThan120` — was "latest death − latest birth"; code is
  `earliestDeath − latestBirth` (`hasAgeRangeGreaterThan`). **(review find)**
- `hasBurialBeforeDeath` — dropped the false "both must be exact day-month-year";
  the predicate falls back to a year-level compare when either side lacks a
  perfect date. **(review find)**
- `hasEventBeforeChristening365_3` — excludes Birth **and** event-registration,
  and anchors on the latest Christening **or Baptism**. **(review find)**
- `hasEventBeforeBirth365_2` — was "before the earliest birth-like fact"; the
  predicate is `latest(birth-like) − earliest(any fact)`, so the reference bound
  is the **latest** birth-like fact.
- `earliestChildBirthToBirth12` / `earliestChildBirthToBirthMale14` — predicate
  fires on `<= cutoff`, so "before age N" → "at age N or younger" (matches the
  emitted `message`).
- `hasDeathBeforeChildBirth30_10` / `hasDeathBeforeChildBirthFemale2` — replaced
  "exact Death/Birth day" (reads as full-DMY-required, the same trap flagged on
  `hasBurialBeforeDeath`) with "exact Death/Birth fact type, not the family;
  day-level, not restricted to full DMY".
- `hasBlankName` — clarified it fires on a blank (empty-string) name part,
  distinct from an absent part (`missingSurnames`/`missingGivenNames*`).
- The three "each self-check has a relative-mob variant" sentences → "27 of the
  47 self-checks…", with "the other 20 run on the anchor only" (27 mirror rows;
  74 − 27 = 47 non-mirror). **(review find)**

Every other Rule cell reads correctly against its predicate and is unchanged.

**Message fix folded in (d7120858):** the emitted `message` for
`hasEventBeforeBirth365_2` (`person-warnings.ts`) also said "earliest birth-like
fact" while the predicate anchors on the latest — now "latest birth-like fact".
This is the one runtime-output change in the PR (a single warning-message
string). It does not touch a skill snapshot, so no `eval-skill` run is required;
`person-warnings` unit + drift suites re-run green (124 tests). Note the
`check-warnings/references/warning-checks.md` copy of this line still reads
"earliest" — left for the next check-warnings PR to fix while riding a run it
already pays for, same as the "Complete" deferral above (editing `skills/` here
would arm `check-runlogs` for a paid run).

Lints re-run green: `person-warnings-spec-drift` (4/4), `doc-links` +
`measured-figures` (166/166).


promise-emmanuel (changes requested):
Re-reviewed the full diff on the new head. Engine suite 2749/2749, tsc
clean. I mutation-tested the lint again and both claims in your PR body
hold: renaming a tag fails both directions, deleting a row fails the count
guard plus the forward check, and renaming the section heading fails the
guard so the parser can't quietly compare empty sets. Good lint.

Two things before I approve. Both were in the head before the update, so
the merge didn't touch them.

1. The hasCloseChildChristenings6_30 catalogue row describes a different
rule than the one that ships. The implementation passes `true` for
compareSimilar, and hasClosePersonFactDates skips every pair where
compareSimilar !== isSimilar, so the check only fires on children whose
names are similar to each other. Your row says any two children, and its
Cause cell says "As above" pointing at hasCloseChildBirthsIgnoreSimilar-
Children, which is the opposite polarity. The window is also 2 to 180 days,
not 0 to 180. The function's own docstring already gets this right.

You wrote:
| `hasCloseChildChristenings6_30` | implausible | Two of this person's children have Christening/Baptism dates within 6 months of each other | As above, on christening dates |

Change it to:
| `hasCloseChildChristenings6_30` | implausible | Two of this person's children whose names are similar have Christening/Baptism dates 2 to 180 days apart | Two records of one child attached as two children, on christening dates |

2. The lint can't see a tag written as a string literal. emittedTags() only
matches issueType: followed by an UPPER_SNAKE name. I added a check emitting
`issueType: "brandNewUndocumentedTag"` — it typechecks (issueType is typed
`string`, so nothing stops it) and all four assertions still pass. An
undocumented tag would ship. I checked the fix below: the clean tree still
passes 4/4, and the literal probe then fails with "expected 75 to be 74".

You wrote:
  for (const m of toolSrc.matchAll(/issueType:\s*([A-Z_0-9]+)/g)) {
    const val = constMap.get(m[1]);
    if (val) out.add(val);
  }

Change it to:
  for (const m of toolSrc.matchAll(
    /issueType:\s*(?:([A-Z_0-9]+)|"([a-z][A-Za-z0-9_]*)")/g,
  )) {
    const val = m[1] ? constMap.get(m[1]) : m[2];
    if (val) out.add(val);
  }

One small thing while you're in there: I followed your "How to add a new
warning" steps 1 to 7 exactly and the lint went red with "ALL_WARNING_TAGS
is empty or lost entries: expected 75 to be 74". The steps don't say to bump
the three hardcoded 74s in the test. Please add a step 8.

Asimi1234:
Thanks — all three addressed in 18d30279.

1. **Christenings catalogue row** — took your replacement verbatim. It now reads "children whose names are similar … 2 to 180 days apart" with the duplicate-record Cause, matching `compareSimilar=true` and the function docstring.

2. **Literal-tag hole** — applied your regex. I reproduced the gap first (added `issueType: "brandNewUndocumentedTag"`, lint stayed 4/4 green), then with your fix it fails "expected 75 to be 74", and the clean tree still passes 4/4. Good catch — my mutation tests only exercised the UPPER_SNAKE path the regex already covered, so they never hit this shape.

3. **"How to add" steps** — added the count-bump as step 7 (before "Run the drift lint", so the guards are bumped before the lint can go green); the old step 7 is now step 8.

Docs + the packaging lint only, no runtime change, so no run-log churn. Re-review when you get a chance.

promise-emmanuel (approved):
Approving. Re-reviewed on de40e918 — level with main, all checks green,
engine suite 2770/2770, tsc clean.

Checked all three fixes by running them rather than reading the diff.

The christenings row now matches compareSimilar=true and the function
docstring — similar names, 2 to 180 days.

The literal-tag hole is closed. I re-ran my probe: a new check emitting
`issueType: "brandNewUndocumentedTag"` now fails with "expected 75 to be
74", where before it passed 4/4. I also re-broke the lint three other ways
to make sure the fix didn't cost anything — renaming a shipped tag still
fails both directions, deleting a catalogue row still fails the guard plus
the forward check, and renaming the section heading still fails the guard at
"0 to be 74". Nothing regressed.

The count-bump step is in the right place, before "run the drift lint". I
added a 75th tag following the revised steps 1 to 8 end to end and the lint
goes green — that same exercise failed before.

Good work on reproducing the gap before applying the fix. That's what made
your note checkable.

Needs a second approval before it can merge.

Optional, take it or leave it: the comment above emittedTags still says
"Tags emitted at `issueType: CONST` sites, resolved through the const map",
which is now half the story since it reads literals too.

johnmarkpeterbrown (changes requested):
Requesting changes on the spec prose only. The lint and the code are done — I re-ran your two mutation tests and both hold, and the suite is green on a clean merge with main. Nothing in `person-warnings.ts`, the drift test, or the match-merge spec needs to change. This is all one file.

**1. The `hasAgeRangeGreaterThan120` rule names the wrong bound.**

The code subtracts the latest birth year from the *earliest* death year. Your row says latest death. The function's own docstring has it right.

```
You wrote:      | `hasAgeRangeGreaterThan120` | contradiction | Latest possible death year minus latest possible birth year is greater than 120 | Wrong birth or death date, or two people merged |
Change it to:   | `hasAgeRangeGreaterThan120` | contradiction | Earliest death-like year minus latest birth-like year is greater than 120 | Wrong birth or death date, or two people merged |
```

**2. `hasBurialBeforeDeath` does not require exact dates.**

It compares exact day-month-year dates only when both sides have them. Otherwise it falls back to comparing years, and it can still fire. As written, a genealogist would assume a year-only burial is never flagged.

```
You wrote:      | `hasBurialBeforeDeath` | contradiction | Every recorded burial day precedes every recorded death day (both must be exact day-month-year) | Data error or transcription mistake |
Change it to:   | `hasBurialBeforeDeath` | contradiction | Every recorded burial precedes every recorded death — compared on exact day-month-year dates when both sides have them, otherwise on years | Data error or transcription mistake |
```

**3. "Each self-check has a relative-mob variant" is not true, and it appears three times.**

There are 47 self-checks and 27 relative-mob variants. Your own mirror table has 27 rows, so the document contradicts itself. Twenty checks run on the anchor only — `hasBlankName`, `missingSurnames`, `tooManyFathers2`, `tooManyMothers2`, `tooManyChildren18`, `hasYoungSpouse15`, `hasChristeningBeforeBirth`, `similarChildren`, `similarSpouses` and `hasSameCensus` among them.

This sentence is your answer to the OPEN QUESTION you retired, so it needs to be exact.

```
You wrote:      not the anchor alone.** Each self-check has a relative-mob variant
Change it to:   not the anchor alone.** 27 of the 47 self-checks have a relative-mob variant
```
Then add this after `personId`/`personName`: `The other 20 run on the anchor only.`

```
You wrote:      in which the anchor participates (as parent, spouse, or child); and each
                self-check has a `relatives*`/`maleRelatives*`/`femaleRelatives*` variant
                that fires the same condition on a one-hop relative.
Change it to:   in which the anchor participates (as parent, spouse, or child); and 27 of
                the 47 self-checks have a `relatives*`/`maleRelatives*`/`femaleRelatives*`
                variant that fires the same condition on a one-hop relative.
```

```
You wrote:      Each self-check above has a relative-mob variant that fires the same
                condition on a one-hop relative (parent, spouse, or child) instead of the
                anchor.
Change it to:   27 of the 47 self-checks above have a relative-mob variant that fires the
                same condition on a one-hop relative (parent, spouse, or child) instead of
                the anchor; the other 20 run on the anchor only.
```

**4. Please walk all 74 Rule cells against their predicate functions.**

This is the main ask. I sampled about thirteen rows and two were wrong. Your PR body says the 14 rows with no prose behind them are the ones to read closely, but `hasAgeRangeGreaterThan120` had reviewed prose and was still wrong — so that triage doesn't hold. The rows are the deliverable here, and no check can confirm one is accurate.

Two smaller ones to fold in while you are in there. `hasEventBeforeChristening365_3` says "non-Birth event", but the code also excludes event-registration types, and it compares against the latest Christening *or Baptism*, not Christening alone.

When you push, say in the PR body that you walked every row — the same way you documented the mutation tests. That is what makes it checkable for the next reviewer.

Asimi1234:
Thanks @johnmarkpeterbrown — all addressed in 2644d359, docs-only.

1. `hasAgeRangeGreaterThan120` — took your line (earliest death-like − latest birth-like).
2. `hasBurialBeforeDeath` — took your line (year fallback when a side lacks a perfect date).
3. `hasEventBeforeChristening365_3` — now excludes Birth **and** event-registration and anchors on latest Christening **or Baptism**.
4. "each self-check has a variant" — "27 of the 47…" at all three sites, plus "the other 20 run on the anchor only" (27 mirror rows; 74 − 27 = 47).

**Walked all 74 rows** against their predicates (not the 14-row triage — you're right it didn't hold). Four more were wrong and are fixed: `hasEventBeforeBirth365_2` (reference is the *latest* birth-like fact, not earliest), `earliestChildBirthToBirth12`/`Male14` (`<=` cutoff → "at age N or younger"), `hasDeathBeforeChildBirth30_10`/`Female2` ("exact … day" reworded — it's the exact fact *type*, day-level, not full-DMY-required), and `hasBlankName` (blank empty-string part vs absent). The PR body now records the walk row-by-row.

One I left out to keep this docs-only: the `hasEventBeforeBirth365_2` **emitted message** also says "earliest" — a runtime-output change, so flagged rather than folded in. Say the word and I'll fold it in or file it.

Lints green: drift 4/4, doc-links + measured-figures 166/166.

Asimi1234:
Folded the message fix in (d7120858): `hasEventBeforeBirth365_2`'s emitted message now says "latest birth-like fact" to match the predicate. One runtime string, no skill snapshot touched; `person-warnings` unit + drift green (124 tests). The `check-warnings/references/warning-checks.md` copy of that line still reads "earliest" — left for the next check-warnings PR to fix on a run it already pays for (editing `skills/` here would arm a paid eval), noted in the body.

johnmarkpeterbrown (approved):
<p>Approving. Re-checked on <code>dccc4952</code> — level with main, all checks green, engine suite 2812 passed, <code>tsc</code> clean, drift lint 4/4.</p>
<p>All three prose fixes are in, verbatim. More importantly the row-by-row walk was real work, not a re-read. You corrected four rows I never flagged, and two of them are better catches than mine: <code>earliestChildBirthToBirth12</code> was an off-by-one against a <code>&lt;=</code> comparison, and the blank-versus-absent distinction on <code>hasBlankName</code> is a genuine semantic difference that the old row flattened. Stripping the false "exact day-month-year" claim off <code>hasDeathBeforeChildBirth30_10</code> and <code>hasDeathBeforeChildBirthFemale2</code> shows you carried the pattern outward instead of just patching the row I named. That is what I was asking for.</p>
<p>I checked the code change too. <code>hasEventBeforeBirth365_2</code>'s message now says "latest birth-like fact", which matches the predicate and the function's own docstring. The old wording was wrong in output the agent reads, so that is a real fix and correctly scoped to the string. Good call disclosing it in the test plan rather than letting it ride in as a docs-only PR.</p>
<p>Documenting the walk in the PR body, with the predicate named for each correction, is what makes this checkable by the next person. Keep doing that.</p>
<p>Note for the record: promise-emmanuel's approval predates both the walk and the message change, so mine is the only approval that has read the current head. Flagging it, not blocking on it — the code delta is one verified string.</p>
<p>Optional and not worth another round: the comment above <code>emittedTags</code> still describes it as reading <code>issueType: CONST</code> sites, which has been half the story since you added literal handling. Fold it into whatever touches that file next.</p></body></html>
</body>
</html>

Issue #607, opened 2026-07-08
fulltext_search + image_transcribe: wire given-name diminutive expansion (the table is issue #1868)


yinkid28:
> **Split 2026-08-24 (board fill).** The variant table moved out to **issue #1868** — it is
> genealogical craft, it is needed identically under every branch of the decision below, and
> a genealogist can start it today. **This issue is now the wiring only, and it is not
> startable yet.** Two things gate it, in order:
>
> 1. **Issue #1868 lands the table.** There is nothing to wire until it exists.
> 2. **One decision, below.** Everything else the original body left open is now settled.
>
> **Settled: host-side, not prose.** The original body's own reasoning stands and nothing has
> refuted it — issue #1007 measured ~50 lines of triage doctrine that `search-records` already
> carries being complied with in 3 runs out of 4, so prose does not reliably change behaviour.
> A tool-side expansion cannot decay and costs no paid eval run. Do not re-open this.
>
> **Decided 2026-08-27 — expand, and report which variants fired.** Both tools expand;
> neither merely suggests. Returning variants as a nil-result hint was rejected because it
> only helps if the agent acts on the hint, which is the prose reliance issue #1007 measured
> at 3-in-4 compliance and that this card already ruled against.
>
> **`fulltext_search`:** the expansion set becomes an **optional group** — do not turn the
> require-default off for the term. `fulltext-search.ts:74` sets `m.queryRequireDefault=on`,
> so today a page reading "Betty" is *excluded* from an "Elizabeth" search, not merely ranked
> low; that exclusion is what this fixes. The response must name which variants fired, so a
> caller can tell from the result set that expansion happened.
>
> **`image_transcribe`:** expand `lookingFor`. It is a VLM instruction rather than a query,
> so expansion changes no result-set semantics and needs no reporting channel.
>
> **Gate 1 has cleared.** Issue #1868 landed the variant table (PR #1944, merged
> 2026-08-27). This is startable.

## Original issue

**IMPORTANT**: If you do this, coordinate with #1093 — the `record_search` half of this
question is folded in there and @DallanQ is answering it. Do not build given-name
expansion for `record_search`.

## The problem

When searching for pre-civil-registration records, the agent only searches the exact
given name provided. It never tries historical diminutives (Betty→Elizabeth,
Peggy→Margaret, Polly→Mary).

**Evidence.** In the `young-marriage-1828` e2e fixture (#570), expected finding f3 is
[...]

Issue #1915, opened 2026-08-25
Hosted runner wedges when subagents outlive a turn — drain the SDK stream and scope the PreToolUse matcher


DallanQ:
## Decision needed — #1915 drain

When the drain reads a subagent's messages *after* the parent turn's `turn_done`, do those events reach the browser, or are they dropped — the hosted event protocol has no out-of-turn channel today and no spec line says whether it should?

- **A — Drop and log** — Confines the change to `real_agent.py` plus a test; the wedge dies this week. Cost: the user still never sees the post-turn subagent prose, which is the payload this reporter lost (4 of 8 extractions). The visibility half is already owned by issue #1971 and issue #1125, both open.  *(recommended)*
- **B — Deliver out-of-turn** — Extends the blast radius to `runner.py`, `sandbox_server.py`, the WS replay buffer and `apps/web`, and reopens the one-`turn_done`-per-turn contract that `test_runner_interrupt.py` pins. Fixes the wedge and the invisibility in one PR, at the price of a second reviewer and a second untested path. **This option changes the blast radius — it should not be spliced in without a re-read.**

<details><summary>Body text if A</summary>

**Decision: the post-turn drain discards what it reads, after an operator log line.**
Delivering those events to the browser was considered and deferred — it would extend
this PR into `runner.py`, `sandbox_server.py` and `apps/web` and reopen the
one-`turn_done`-per-turn contract. The user-visible half stays with issue #1971 (a
wedged session and a finished turn look identical) and issue #1125 (`is_error` is
dropped by the event mapper). Record this in
`docs/specs/hosted-web-workbench-spec.md` §7 as part of the task, next to the drain
rule — not only in this issue.

</details>

<details><summary>Body text if B</summary>

**Decision: drained post-turn events are delivered to the browser out-of-turn**, rather
than discarded — discarding was the cheaper option and was rejected because the lost
payload (subagent completion prose) is exactly what this incident cost the user.
This widens the change beyond `real_agent.py`: `runner.py` and `sandbox_server.py` must
carry an event that is not inside a turn, `apps/web` must render it without keying on
the busy state, and the one-`turn_done`-per-turn contract pinned by
`test_runner_interrupt.py` has to be restated rather than broken. Re-read the blast
radius above before starting; this is a two-reviewer PR. Record the protocol change in
`docs/specs/hosted-web-workbench-spec.md` §6.2 and §7 as part of the task.

</details>

[...]

PR #1944, opened 2026-08-26
given-name diminutive/abbreviation table (#1868)

### LINKED ISSUE #1868: Build the period- and region-scoped given-name diminutive table (split from #607)
> **Reviewed 2026-08-24 before junior handoff.** Corrected the run-count evidence to what the committed logs actually show, and added sections on seeding from the two existing in-repo tables, sourcing rules, out of scope, and what nothing checks.

**Touches:** packages/engine/mcp-server/src/data/given-name-variants.json (new)

Split out of issue #607 on 2026-08-24. That issue asked for one thing that turned out to be two — a genealogical reference table, and the code that reads it. **This issue is the table only.** Issue #607 keeps the wiring and stays in Backlog behind a lead decision.

## Why the table can be built before that decision

Issue #607 leaves three questions open: where the variant table comes from, whether expansion happens host-side or in prose, and whether the tool expands silently or returns variants as a hint on a nil result. **The table is needed under every branch of all three.** Only the wiring changes.

## What to build

`packages/engine/mcp-server/src/data/given-name-variants.json` — a period- and region-scoped table mapping historical given-name diminutives to their formal forms, each family carrying a source. Suggested shape, one object per language:

```json
{
  "en": {
    "Elizabeth": { "variants": ["Betty", "Bess", "Bessie", "Eliza", "Libby", "Lizzie"],
                   "period": "1600-1900", "source": "…" },
    "Margaret":  { "variants": ["Peggy", "Peg", "Meg", "Maggie"], "period": "…", "source": "…" },
    "Mary":      { "variants": ["Polly", "Molly"], "period": "…", "source": "…" }
  }
}
```

Bundling it into the `.mcpb` is issue #607's job, not this one's.

## Start from the tables that already exist — do not generate one

The English rows are already written, twice, and one carries measured
provenance a fresh generation would silently lose:

- `packages/engine/plugin/skills/search-records/references/name-search-mechanics.md:202-219`
  — 16 formal names, Formal → Nicknames, plus the direction rule at :187
  (tree holds the formal name, the record holds the nickname) and the
  measured/untested split at :177.
- `packages/engine/plugin/skills/search-full-text/references/search-strategies.md:56-83`
  — the abbreviation table (`Margt`, `Eliz`) and the Latin/Spanish/German
  cross-language block.

Seed the file from those rows. Where you disagree with one, say so in the PR
body and change both — a JSON table that silently contradicts the prose the
skill reads is a third copy that drifts. The duplication is forced, not a
mistake: skills run in the VM and cannot read a host-side file (`CLAUDE.md`,
"Don't try to share code at runtime between the MCP server and the skills").

## Sourcing rules — these are what the reviewer checks

- **Never invent a `source`.** Each must be a published reference you actually
  opened, an in-repo `file:line`, or a FamilySearch record ark. A
  plausible-looking citation nobody can follow is worse than an empty field.
- **Carry the measured/untested split forward.** Exactly three pairs are
  measured — Elizabeth→Betty, Margaret→Peggy, Mary→Polly (id-level membership
  test, 8 of 8, 2026-08-08; `packages/engine/mcp-server/dev/probe-search-qualifiers.ts`
  section E, `verdict:diminutiveReach` in `dev/measured-figures.json`). Every
  other row is untested: give it `"attested": false` rather than dropping it or
  quietly promoting it.
- **Add `region` to every entry.** The suggested shape carries `period` and
  `source` but no `region`, which "Done when" requires. Language is not region —
  Portuguese covers Portugal and Brazil, and a form used in one is not evidence
  for the other.

## Out of scope

Acquired nicknames not derived from the given name — Bitsie for Mary, "Tug" for
John. Issue #399 (closed completed) asked for those; no lookup table can hold
them. Do not try.

## Nothing checks this file

No test reads it, no schema validates it, and no CI job can fail on a wrong
entry — a reviewer following your sources is the only instrument. Two
consequences:

- Keep the PR small enough to review row by row. A 400-row table nobody can
  check is worth less than a 60-row one that can be.
- **It does not ship yet, and issue #607 must fix that.** `.mcpbignore:7`
  excludes `src/`, `scripts/build-mcpb.mjs:35` stages only `build` and `config`,
  and `packages/engine/mcp-server/tsconfig.json` sets no `resolveJsonModule`, so
  `tsc` never copies a `.json` out of `src/`. Issue #607 either turns
  `resolveJsonModule` on and imports the file, or moves it beside
  `config/familysearch.json`. Not this PR's job — put it in the PR body so #607
  inherits it instead of discovering it at install time.

## The evidence it has to cover

`eval/tests/e2e/young-marriage-1828/` — expected finding f3 is indexed as **Betty Martin** (christened Bitton, 1 November 1812, daughter of Thomas & Sophia Martin). The agent searched "Elizabeth Martin". The committed run logs confirm it never tried: `Betty` appears in all three
committed runs (`ls eval/runlogs/e2e/young-marriage-1828/`, measured 2026-08-24)
only inside judge text, never in a tool call. The original 7-run figure dates
from 2026-07-08 and does not reproduce — committed run logs are converged
states, so take the behaviour as confirmed and the count as unrecoverable.

`fulltext_search` makes that worse than a ranking miss: `fulltext-search.ts:74` sets `m.queryRequireDefault=on`, so every term is *required*. An unexpanded "Elizabeth" cannot match a page reading "Betty" — the record is excluded, not ranked low.

## Which languages

**English only.** The corpus evidence for a diminutive miss is one fixture —
`eval/tests/e2e/young-marriage-1828/expected-findings.json:40`, indexed "Betty
Martin". The nine non-English fixtures previously listed here show a *different*
mechanism and no diminutive at all: Latin register forms
(`elisabetha-sugecz-parents/README.md:40` "Thomæ Sugecz and Susanæ Petrich"),
Slavic cognates (`jan-gallo-father/README.md:12` "Andrej, a variant of Ondrej")
and surname spelling variance (`susanna-szljacsan-spouse/README.md:75`). Building
Hungarian or Portuguese diminutives would serve none of them.

Chosen 2026-08-24 over one combined name-form table: it keeps issue #607's
expansion contract to a single kind of substitution. File the Latin/cognate need
as its own issue in the same PR that lands this, and say there that it was split
off deliberately.

## Constraints carried over from issue #607 — do not rediscover these

- **Build nothing for `record_search`.** Its `.exact` is opt-in, so the default already *is* FamilySearch's close matching, and it separately ships `givenNameAlt`/`surnameAlt` and `q.givenName.exact.N` alternate-name slots. That half was folded into issue #1093, which closed completed 2026-08-12.
- **Build nothing for external sites.** Ancestry and the rest expand automatically, and `external_links_search` takes a place, not a person name.
- Only `fulltext_search` and `image_transcribe` lack diminutive handling. Those two are what the table serves.

## Done when

- The file exists, English only, with `period`, `region`, `source` and `attested`
  on every entry.
- Betty/Bess/Eliza→Elizabeth, Peggy→Margaret and Polly→Mary are present and
  marked `"attested": true`.
- Every row traces to `name-search-mechanics.md:202-219` or to a source you name.

kofiatinka12:
## Start here
`packages/engine/mcp-server/src/data/given-name-variants.json` — a new data file, plus its `_meta` block (shape, sourcing, measured evidence, seed tables, scope defaults, out-of-scope).

## What this does
Builds the English given-name variant table from #1868. Seeded from the two existing in-repo tables (not generated):
- `packages/engine/plugin/skills/search-records/references/name-search-mechanics.md:202-219` (Formal → nicknames)
- `packages/engine/plugin/skills/search-full-text/references/search-strategies.md:56-71` (Formal → scribal abbreviations)

**21 formal names, 105 variants.** Shape: `en → <Formal> → { period, region, source, variants: [ { form, attested, source? } ] }`.

## Review fixes in the latest revision (commit addressing @mercyokum's review)
- **Added the missing `Katherine` key.** The seed row `name-search-mechanics.md:208` reads `Catherine/Katherine | Kate, Kitty, Cathy, Katy, Trina` — both spellings share one nickname set. The file now carries a `Katherine` entry mirroring `Catherine` (same five variants, all `attested: false`, same source `:208`), so a record indexed under "Katherine" gets nickname expansion. This is the count change from 20/100 → **21/105**.
- **Documented the period/region scope default in the file itself.** New `_meta.period_region_note` states plainly that `period` ("1600-1900") and `region` ("England and Anglophone North America") are **file-wide scope defaults** stamped on every entry, not per-name or per-variant measurements, and that period-appropriateness is a separate axis from `attested`. It names the modern-only forms that fall outside the window despite inheriting the tag (Liam, Lisa, Chuck, Carl, Hank, Rick). Previously this caveat lived only in the PR body; it is now in `_meta`.
- **Clarified the Pepe exclusion in `_meta.out_of_scope`.** The "(Spanish)" tag on Pepe appears only on the Joseph row (`:218`), not the Francis row (`:217`) where Pepe also appears untagged. `_meta.out_of_scope` now states that Pepe is excluded from both as a cross-language form, and that applying the source's row-local tagging mechanically would wrongly retain the untagged Pepe on Francis. The data is unchanged.
- **Deduplicated the measured-evidence citation.** The authoritative citation stays in `_meta.measured_evidence`; the three per-variant strings (Betty/Peggy/Polly) now read `"source": "measured — see _meta.measured_evidence"`, so a re-probe updates one place, not four. No `attested` value changed.

## Sourcing (what the reviewer checks)
- **Every entry traces to an in-repo `file:line`.** No source is invented.
- **`attested` is per-VARIANT, and only the three variants with committed measured evidence are `true`:** Betty→Elizabeth, Peggy→Margaret, Polly→Mary. The measured citation is `packages/engine/mcp-server/dev/probe-search-qualifiers.ts` (Section E membership test, 8/8, 2026-08-08) and `dev/measured-figures.json` (`membershipTested`; `verdict:diminutiveReach`), now held once in `_meta.measured_evidence` and pointed to from each of the three variants. All 102 other variants are `attested: false`.
- **`region`** ("England and Anglophone North America") and **`period`** ("1600-1900") are file-wide scope defaults for the English-diminutive window, now stated as such in `_meta.period_region_note` — not per-name measurements.

## Deviation from the literal "Done when" — stated per the issue's own instruction
The issue's "Done when" groups **"Betty/Bess/Eliza→Elizabeth … attested: true."** The committed measured evidence covers **only Betty** (with Peggy and Polly). Per the sourcing rule — *"give it attested: false rather than … quietly promoting it"* — **Bess, Eliza, and every other unmeasured variant are `attested: false`.** Only Betty/Peggy/Polly are `true`. This preserves the measured/untested split #1868 requires; nothing was invented or inferred to make the others true.

## Out of scope — cross-language name forms
The inline cognates in the seed tables (Hans/Honza for John, Diego for James, Paco/Pepe for Francis/Joseph) and the Latin/Spanish/German block (`search-strategies.md:76-83`: Joannes↔John, Iacobus↔James, Catalina↔Catherine, Wilhelm↔William, …) are **deliberately excluded** and documented in `_meta.out_of_scope`. They are a different substitution mechanism (cross-language cognate / ecclesiastical register-form equivalence, formal↔formal across languages), not within-language diminutive expansion. #1868 was scoped English-only on 2026-08-24 over one combined name-form table to keep #607's expansion contract to a single kind of substitution. **No follow-on issue is filed** (a deliberate call this round); the need is recorded in `_meta.out_of_scope` for whoever picks it up.

## Didn't change
- **The seed prose tables are untouched.** Modern-only forms in them (Liam, Lisa, Chuck, Hank, Carl, Rick…) are seeded **verbatim** and marked `attested: false`; they were **not** curated out, because pruning them would mean editing `name-search-mechanics.md` (in *search-records*' snapshot → a separate paid run), outside this issue's `Touches: given-name-variants.json (new)`. Their out-of-window status is now disclosed in `_meta.period_region_note`.
- No skill, test, validator, fixture, or rubric file changed.

## It does not ship yet — #607's job (put here so #607 inherits it)
`tsc` never copies this `.json` out of `src/`: `.mcpbignore:7` excludes `src/`, `scripts/build-mcpb.mjs` stages only `build` and `config`, and `packages/engine/mcp-server/tsconfig.json` sets no `resolveJsonModule`. #607 must either turn on `resolveJsonModule` and import the file, or move it beside `config/familysearch.json`.

## Nothing checks this file
No test reads it, no schema validates it, no CI job can fail on a wrong entry — a reviewer following the sources is the only instrument. The table is 105 rows, reviewable row by row.

## Acceptance check
- File exists, English only, with `period`, `region`, `source` on every entry and `attested` on every variant.
- Both `Catherine` and `Katherine` present, sharing the five variants from `:208`.
- Betty→Elizabeth, Peggy→Margaret, Polly→Mary present and `attested: true`; all others `false`.
- Every row traces to `name-search-mechanics.md:202-219` or `search-strategies.md:56-71` (or, for the three measured, `_meta.measured_evidence` → the probe + measured-figures.json).
- JSON parses.

Closes #1868.


mercyokum (changes requested):
**Requesting changes**

Good, careful work on the sourcing — every variant across all 20 names traces exactly to the cited file:line, and the three `attested: true` claims (Betty, Peggy, Polly) check out against the probe script and measured-figures.json precisely. A few things worth fixing before merge.

## 1. `Katherine` spelling silently dropped

The cited source row (`name-search-mechanics.md:208`) reads `Catherine/Katherine | Kate, Kitty, Cathy, Katy, Trina` — treating both spellings as equally valid formal names sharing one nickname set. The shipped file only has a `Catherine` key; `Katherine` doesn't appear anywhere, and unlike the file's cross-language exclusions (all listed in `_meta.out_of_scope`), this same-language spelling drop isn't documented anywhere. A record indexed under "Katherine" — at least as common a spelling as "Catherine" in this period — gets no nickname-expansion lookup at all.

## 2. `period: "1600-1900"` contradicts the PR's own text — and the fix it claims to have made isn't there

Every entry stamps an identical entry-level `period`/`region` covering its whole variant array. But the PR body's own "Didn't change" section admits Liam (William), Lisa (Elizabeth), Chuck/Carl (Charles), Hank (Henry), and Rick (Richard) are "modern-only forms" seeded into those same arrays. `attested: false` doesn't cover this — it means "no measured search-reach evidence," a different axis from "is this variant period-appropriate."

The PR body also says `region`/`period` are "scope judgments... stated as such in `_meta`" — I checked `_meta` directly and that caveat isn't actually there. None of its 8 fields (`description`, `shape`, `attested_note`, `sourcing`, `out_of_scope`, `seed_tables`, `measured_evidence`) says anything about period/region being a scope default rather than a per-name measurement. The disclosure exists only in the PR body prose, not in the file itself.

**Suggested fix:** add one sentence to `_meta` making explicit that `period`/`region` are a scope default for the whole file, not a per-variant measurement, and that some variants (Liam, Lisa, Chuck, Hank, Carl, Rick) fall outside that window despite inheriting the tag — the same disclosure pattern already used for `attested`.

## 3. Francis's "Pepe" exclusion breaks the file's own tagging rule

`name-search-mechanics.md:217` (`Francis | Frank, Frankie, Paco (Spanish), Pepe`) only tags "Paco" as Spanish on that row — the "(Spanish)" tag for Pepe only appears one row down, on Joseph's line 218. Every other cross-language exclusion in the file (Hans/German, Honza/Czech, Diego/Spanish) is justified by an explicit tag on the exact row it's excluded from; Francis's Pepe is the one exception, excluded by cross-row inference instead. Substantively correct (Pepe really is Spanish) and disclosed in `_meta.out_of_scope`, but a future maintainer mechanically applying the file's own stated rule ("exclude only source-line-tagged foreign forms") would keep Pepe on the Francis row and reach a different result than what shipped.

## 4. Identical citation string repeated 4 times

The full measured-evidence citation (`probe-search-qualifiers.ts (Section E membership test, 8/8, 2026-08-08); measured-figures.json (membershipTested; verdict:diminutiveReach)`) appears once in `_meta.measured_evidence` and then verbatim three more times, once per `attested: true` variant (Betty, Peggy, Polly). If the probe is re-run and the date or verdict changes, that's 4 places to update by hand with no lint to catch a missed one. Consider hoisting the full citation into `_meta` and leaving a short pointer (e.g. `"source": "measured"`) on each variant.

## One process note, not a code change

The PR body says outright: "Nothing checks this file. No test reads it, no schema validates it, no CI job can fail on a wrong entry." That's a textbook case for this repo's own `nothing-checks` label (CLAUDE.md: add it "when the item is a missing guard — a way CI can be green while the thing is broken"; `docs/architecture.md` §9.4 calls it "the register" for exactly this). Neither #1868 nor #607 currently carries it, though the label is real and actively used elsewhere in the repo (#1940, #1930, #1915...). Worth labelling those issues so the gap doesn't go invisible before #607 ships this file.

mercyokum (approved):

EdmondOware (changes requested):
The sourcing discipline here is genuinely good, and @mercyokum's four items are all closed:
`Katherine` is its own key, `period_region_note` is in `_meta`, the Pepe exclusion is
documented, and the attested citations are short pointers now. I also checked the file for
encoding damage since it carries `Jn°` and em dashes, and it is clean: decodes as UTF-8
throughout, `Jn°` is a real `U+00B0`, no mojibake in any variant form.

One thing to fix, and it is invisible to a sourcing check because every entry traces correctly.

### Four names ship with no diminutive at all

The file merges two seed tables of different kinds:

- `name-search-mechanics.md:202-219` — "Formal → **Nicknames** seen in records", 16 names
- `search-strategies.md:56-71` — "Formal → **Abbreviations** to search separately", 14 names

Thomas, Samuel, Benjamin and George appear only in the second, so they ship abbreviations and
nothing else:

| name | shipped | absent |
|---|---|---|
| Thomas | Thos, Thos., Tho. | **Tom**, Tommy |
| Samuel | Saml, Saml., Sam. | **Sam**, Sammy |
| Benjamin | Benj, Benj., Benjm. | **Ben**, Benny |
| George | Geo, Geo. | Georgie |

I checked the whole file, not just those entries: `Tom`, `Tommy`, `Sam`, `Sammy`, `Ben`,
`Benny` and `Georgie` appear nowhere in it, under any formal name.

Samuel is the sharpest case. It ships `Sam.` **with the period** — the scribal abbreviation —
while the bare given name `Sam` is missing. A record indexed as "Sam Wilkins" gets no
expansion to Samuel, and Tom, Sam and Ben are among the most common diminutives in the entire
Anglophone record corpus. That is the exact lookup this file exists to serve.

`_meta.description` promises "diminutives, familiar forms, and scribal abbreviations" without
noting that four of 21 entries deliver only the third, so a consumer has no way to know the
coverage is uneven.

**Two ways to close it, either is fine.** Add the missing forms, which costs nothing since no
test reads this file yet. Or, if you would rather not go beyond the seed tables, add one line
to `_meta` naming those four as abbreviation-only and saying why, so the gap is deliberate
rather than silent — the same disclosure pattern you already use for `attested` and for
`period`/`region`.

### While you are in there, one note on @mercyokum's item 2

Her period point is right, and the same reasoning reaches a second group. `Lisa`, `Chuck`,
`Rick` and `Liam` are 20th-century English usage, and `Carl` is a separate Germanic name
rather than a diminutive of Charles. Separately, `Greta` and `Rita` (Margaret) and `Trina`
(Catherine) are cross-language forms of the same kind the file deliberately excludes for
Hans, Honza, Diego and Pepe — they come in only because the seed row did not tag them. Not
blocking, and the `period_region_note` covers the first group, but the exclusion rule reads
inconsistently while those three stay.

kofiatinka12:
@EdmondOware thanks — good catch on both, and I've addressed each with explicit `_meta` documentation rather than by inventing data. No unsourced diminutives were added; no `attested` value changed; the seed rows are untouched.

**Abbreviation-only names (Thomas, Samuel, Benjamin, George).** I confirmed your read: these four appear only in the abbreviation seed table (`search-strategies.md:56-71`), not the nicknames table, and Tom/Tommy, Sam/Sammy, Ben/Benny, Georgie appear in **neither** seed table. Adding them would invent a source, which is the one rule this file is built on — so I took your second option. New `_meta.coverage_note` states that these four are abbreviation-only because they come only from `search-strategies.md`, and that their common diminutives were deliberately not added for lack of any seed source. It calls out Samuel specifically: `Sam.` (with the period) is the sourced abbreviation from `:65`; bare `Sam` is intentionally not promoted.

**Carl / Greta / Rita / Trina.** New `_meta.seed_inclusion_note` makes the source-inclusion vs mechanism-classification distinction explicit: a form ships iff it's present verbatim in an English seed table, and classification doesn't gate inclusion. Greta/Rita (Margaret) and Trina (Catherine) are cross-language and Carl is a distinct Germanic name, but all four are kept because they appear untagged in the English seed — removing them would edit the seed. The note narrows the `out_of_scope` rule so it reads: exclude only source-tagged foreign forms and the cross-language block; retain everything else the English seed lists verbatim.

Both additions are in `8e2e2dfb`.

EdmondOware (approved):
Approving. Both items are closed, and closed the right way — with documentation rather than
invented data.

**The abbreviation-only four.** `coverage_note` does more than the one line I asked for: it
names Thomas, Samuel, Benjamin and George, gives the cause (they appear only in
`search-strategies.md:56-71`, never in the nicknames table), lists the specific forms that are
absent, and states plainly that adding them would invent a source. Calling out Samuel — `Sam.`
with the period is the sourced abbreviation while the bare `Sam` is not attested by any seed
row — is exactly the distinction that makes the gap legible to the next reader. Declining to
add data to satisfy a reviewer is the right instinct, and I would rather have this note than
four unsourced entries.

**Greta, Rita, Trina and Carl.** `seed_inclusion_note` resolves it better than my framing did.
Separating source inclusion from mechanism classification, and deciding inclusion first, is a
cleaner rule than the one I was implicitly applying, and it explains the apparent
inconsistency with the Hans/Honza/Diego/Pepe exclusions without weakening either.

Re-checked on `8e2e2dfb`: valid JSON, decodes as UTF-8, **zero** mojibake codepoints, 21 names
and 100 variant forms, no duplicate form within an entry, and every entry carries both
`source` and `variants`. `Jn°` is a genuine `U+00B0` and is right to keep — a real
early-modern scribal form.

### One thing for the consuming issue, not for this PR

The gap is documented now, but it is still a gap. A record indexed as "Tom Wilkins" or "Ben
Carter" gets no expansion, and Tom, Sam and Ben are among the most common diminutives in the
Anglophone corpus. Your reasoning for not fixing it here is right — the fix is a nicknames
source for those four names, not a rule change in this file.

What makes that tolerable today is that nothing reads the file yet: I grepped `packages/`,
`eval/` and `docs/` and there are no consumers, so the documented gap costs nothing until
#607 wires it in. Worth carrying there so it is picked up when someone can source it properly,
rather than living in `_meta` indefinitely.

Nothing outstanding from me. You are one commit behind `main` and will need updating.

Issue #1729, opened 2026-08-18
Reopening a session after inactivity: the chat sits on "Connecting to the agent…" and the conversation is gone


DallanQ:
> **Reviewed 2026-08-27 before junior handoff.** Routed `senior`; two body claims were false,
> one of which is green in CI and dead in E2B. Notes below, original body under
> "Original issue".

> **Half 2 added 2026-08-25 from a user report.** The reporter sees one bug — reopen a
> session, get a dead panel — and it takes both halves to fix it, so they land on one card
> rather than as two that rely on someone picking up both.

**IMPORTANT**: If you do this, do issue #1959 and issue #2062 at the same time — and do
issue #1959 first.

## Related — one cluster, three cards (note added 2026-09-02, board pass `/fill-ready`)

The lead directed on 2026-08-31 that these three be held by **one person**, worked in the
order **#1959 → #1729 → #2062**. That direction was written into issue #1959 and issue #2062
on 2026-08-31 but **not** into this card, which is the half that was missing: whoever picked
this one up first would have learned nothing about the other two.

- **Issue #1959 — `_handle_stream` has no turn cap.** Ready, unassigned, reviewed, with the
  idle-cap decision already recorded. Do this one first.
- **Issue #2062 — hosted chat answers the previous message.** Names the same two client
  turn-state files this card owns (`apps/web/src/components/ChatPane.tsx`,
  `chatEvents.ts`), plus `apps/server/app/sandbox_server.py`. Routed `senior` on 2026-09-02.
  **Corrected 2026-09-02:** an earlier version of this note said #2062 "may be *caused* by
  the gap #1959 closes". That is refuted — #1959 is a `/v1` card, and the browser never
  calls `/v1` (`grep -rn "/v1" apps/web/src` → 0; `apps/server/app/v1.py:1` is the
  bearer-only external-chatbot surface). Landing #1959 **cannot** remove #2062's cause. The
  real overlap between #2062 and *this* card is the two client turn-state files plus
  `sandbox_server.py`, which is why the pairing stands on its own.

Three people editing `v1.py`, `ChatPane.tsx` and `chatEvents.ts` at once is the shape that
produces merge races. Ask to be assigned all three.

[...]

PR #1946, opened 2026-08-26
search-records: fix unhedged census-household kinship claims + Windows feedback-case setup bug

### LINKED ISSUE #1912: [feedback] 2026-08-25T21:54Z
**Zip:** https://drive.google.com/file/d/11Ufnw4lYM2jr7VFJ8aIB33epYYKu2xom/view?usp=drivesdk

```
make feedback-case ZIP=~/Downloads/feedback-2026-08-25T21-53-52-955Z.zip
```

- Submitted: 2026-08-25T21:53:52.955Z
- Platform: win32

mercyokum:
**The issue:** A tester (feedback issue #1912) had `search-records` write an 1850 census household's extra co-residents into a log entry as the subject's "sons," even though the 1850 census has no relationship-to-head column at all — nothing in the record says who is whose son. The agent only corrected itself after the tester questioned it. The same feedback bundle also reported a crash losing project memory on session restart, and a "missing tools" error near the end of a very long session.

While setting up the feedback case on Windows, `scripts/setup-feedback-case.bat` also failed outright: a bare `shift` shifts `%0` along with the positional args it's consuming, so once the script has consumed the zip-path argument, `%~dp0` resolves to the zip's own folder instead of the script's folder, and repo-root detection fails on every single invocation.

**The fix:**
- `scripts/setup-feedback-case.bat`: use `shift /1` (shifts starting at `%1`, leaves `%0` alone) at all three call sites.
- `search-records`: `SKILL.md`'s existing rule on this already correctly forbids exactly what happened, so no skill edit was needed — the gap was regression coverage. Mined `ut_search_records_h4k` (scenario `whitfield-1850-household`, scrubbed from the real case) to pin it, plus permanent, deterministic pins of the real production text in the validator's own unit-test file (`tests/unit/test_search_records_pre1880_validator.py`).
- Widened what the deterministic validator (`test_pre1880_census_structure_marked_inferred`, issue #1284) recognizes as a claim needing a hedge: a possessive kinship claim ("Amos's children") and a plural kinship noun bare-introducing named individuals ("sons Thos T McElwee and Stephen McElwee" — issue #1912's actual production text), alongside the existing "as ROLE" form.
- **A per-sentence variant of the check was tried and withdrawn** — see T-FEH's review below. It required the hedge to sit in the same sentence as the claim, to catch a marker elsewhere in the note that never actually qualifies a specific claim. Two real bugs surfaced in review (an abbreviation-naive sentence splitter, and the "as ROLE" pattern firing on a claim quoted only to be rejected), and a third, deeper problem behind both: `ut_search_records_015`'s own committed, correctly-hedged output has the identical two-sentence shape as the case this was built to catch, and no reliable mechanical signal tells them apart. Reverted to the original note-wide check (unedited) — right precision for what a validator, rather than the judge, can safely automate.
- Updated `ut_search_records_h4k`'s `judge_context` to match: a hedge anywhere in the same log entry is sufficient, not necessarily in the same sentence as the claim.
- **Round 2 of review found two more real bugs, both fixed:** `[A-Z]` in both new patterns was searched with `re.IGNORECASE`, so it matched any letter rather than the capitalized-proper-noun scoping the comments promised (`the family's children`, `daughters in the index` were wrongly flagged) — fixed with `(?-i:[A-Z])`, which locally disables case-insensitivity for just that character class. Separately, every fixture exercising the two new patterns also said "household," so the original #1284 rule was silently doing the work and deleting the new patterns entirely still passed all 17 pinned tests — reworded the fixtures to drop "household" and added a dedicated no-hedge, no-household offender case for the possessive-kinship pattern. A third round found the plural-kinship pattern required literal whitespace between the noun and the name, so a comma (`sons, Thos T McElwee`) slipped past — fixed with a small separator character class. Two comments were also corrected to describe what the patterns actually match (a tree-side mention like "George's wife Catherine" *is* caught by the possessive pattern, not excluded; the plural pattern matches any capitalized following word, not only a name).

**Verified:**
- All pre-existing pinned validator tests pass unedited, plus 8 new ones added across both review rounds (real production text for the offending case; the withdrawn-design case now pinned as compliant; three constructed adversarial probes pinning the `(?-i:[A-Z])` fix; one pinning the comma-separator fix; one isolating the possessive-kinship pattern from the household-mention fallback). 22 pinned tests total.
- Replayed old vs. new validator over every note in every committed `search-records` run log to date, with the tag forced on all of them (168 notes, matching T-FEH's review methodology exactly): **0 new false positives, 4 corrected false-to-true** (the 3 T-FEH found, plus `ut_search_records_h4k`'s own captured output — which, per the corrected understanding, is legitimately hedged). Re-verified after round 2's fixes against the actual current committed run logs (44 real logged searches across the 9 tagged tests, correctly unpacking the `research_log_append` batched-`ops` call shape): the comma-separator fix adds zero new flags on real data, and both new patterns remain guarded — confirmed by temporarily deleting them and watching 2 pinned tests fail that previously passed.
- `ut_search_records_h4k` is `expected_outcome: xfail` (issue #1967, blocked on this PR merging). **Correction, found by a third reviewer (clack391) after this was first written:** issue #1967 is not building a `research_log_append` writer-tool precondition — that direction was analyzed and explicitly rejected by the lead on structural grounds (a hard pre-1880/US census gate measured against the full corpus would refuse 41% of `record_search` calls, is not generalizable outside the US, and the year/country signal is author-supplied and optional). #1967's actual finding: this is *partly a grading defect* — the judge contradicts this test's own `judge_context`, penalizing a correctly-hedged note the same as a genuinely bare one. Confirmed directly: `v1_2026-08-27_13-03-15.json`'s judge rationale for this test scores a note reading "...likely sons... family structure inferred, not stated" (hedge in a later sentence, `judge_context`'s own "correct shape") as Correctness 1, reasoning "does not retroactively qualify the earlier bare kinship claim" — the literal opposite of what `judge_context` says. The evidence run for #1912's original production defect is still `v1_2026-08-27_11-50-40.json` (6 of 9 tagged tests failed outright, zero hedge anywhere). The run log shipped with this PR (`v1_2026-08-27_15-10-12.json`) records `h4k` as **xpass** — the skill hedged correctly and the judge, this one time, did not contradict itself. That is not a fix landing here; no `SKILL.md` edit was made, and the judge_context/rubric reconciliation is #1967's work, not this PR's.
- Full harness suite: 2884 passed, 3 skipped. `ruff` clean.
- `ut_search_records_014` also failed in the shipped run log, for a reason unrelated to this PR: the skill misidentified the returned record (a baptism entry as a census hit) and skipped the required age/role cross-check. Ordinary run-to-run variance on that test, not caused by or related to the census-narration fix here.
- The `.bat` fix: tested directly on Windows across multiple real invocations — both the original bug reproducing (repo-root detection failing on every call) and the fix resolving it (clean setup, correct repo root) — before this PR was opened. Both the round-1 and round-2 reviewers confirmed the fix reads correctly but neither had a Windows box to execute it themselves, so this Windows run is the only thing covering that half.

**Deferred issues found:**
- The crash / lost-project-memory report from this same feedback bundle matches an existing tracked issue almost exactly (#1729, "Reopening a session after inactivity…") — that's a platform/session issue, not a skill fix, so it's being linked there rather than duplicated here.
- The "missing tools" error near the end of the session couldn't be root-caused: no Claude Code session transcript was captured for this bundle (expected for a Cowork session), so there's nothing to point at yet.

Closes #1912


T-FEH (changes requested):
Reviewed on the merged head (`7e51af64` — your `d5dd49c9` plus a clean merge of main). Harness **2859 passed, 2 skipped**; `ruff` clean; eval gate satisfied.

The diagnosis is right and the direction is right — the note-wide check genuinely does under-catch a claim "cured" from a distance, and `ut_search_records_h4k` reproduces #1912's defect exactly. Net the run improves on main's latest (23/28 → 26/29). Three things block.

## 1. The sentence split breaks on abbreviations, and fails correctly-hedged notes

`re.split(r"(?<=[.!?])\s+", notes)` splits at any period followed by a space. Genealogical notes are full of `b. 1839`, `Wm.`, `Co.`, `Jr.` — so an abbreviation between a claim and its marker separates them into different fragments. Four probes, every note **correctly hedged** with "structure inferred, not stated", differing only in an intervening abbreviation:

```
PASS   no abbreviation between claim and marker
FAIL   'b.' falls between claim and marker
FAIL   'Wm.' falls between claim and marker
FAIL   'Co.' falls between claim and marker
```

This is not hypothetical. `ut_search_records_015` (**tagged** `pre-1880-census-household`) wrote this in two committed runs, and it flips PASS → FAIL:

```
note:     'Sarah A. Mullen, born 1852, Wisconsin, in Dodge County, WI household headed by
           William Mullen (b. 1825, Ireland) with Margaret Mullen (b. 1830, Ireland).
           Indexer-inferred ParentChild/Couple relationships — 1860 census carries no
           relationship column; household structure ...'
tripping fragment: '...household headed by William Mullen (b.'
marker present somewhere in note: True
```

The claim and its marker are in the same real sentence. `b.` is what splits them. The docstring reasons carefully about semicolons keeping SKILL.md's prescribed shape intact — abbreviations need the same treatment.

## 2. It fails the skill for correctly flagging the indexer

`ut_search_records_014` is also tagged, and also flips PASS → FAIL — on this fragment:

```
'NEEDS-REVIEW: record indexes Patrick as "Head of Household," which is a logical
 impossibility for a person born 1845 (age ~5 in 1850).'
```

That is the skill doing exactly the right thing: naming the indexer's claim in order to reject it. `_ROLE_ASSERTIONS`'s `as ROLE` pattern reads it as the skill asserting the role. The note carries a marker elsewhere, so the old check passed it.

A claim being quoted in order to be refuted is the opposite of the defect this validator exists to catch, and the validator gates — so the judge never runs and the test reads as a bare fail.

## 3. The mined test ships red, with nothing fixing it

`ut_search_records_h4k` is `type: positive`, `expected_outcome: pass`, and its recorded outcome is **fail** — Correctness 1, Log quality 1, on "likely Amos's children". That is #1912 reproduced.

The body's reasoning is that SKILL.md already forbids this, so "the gap was regression coverage." But if the rule exists and the skill still violates it, the gap is not only coverage — the skill is still wrong, and this ships a permanently-failing test into the suite. The next person cannot tell it from a new regression.

Either mark it `xfail` so it pins the known defect without redding the suite, or fix the skill so it passes. As it stands it is neither.

## The regression claim does not hold

The body says: "against every `pre-1880-census-household` note in every committed `search-records` run log to date (41 samples): zero regressions, only stricter (correct) catches."

I replayed the old and new validator over every note in every committed `search-records` run log — **168 notes**, same input to both:

```
old -> new:   PASS->PASS 125   FAIL->FAIL 38   PASS->FAIL 5   FAIL->PASS 0
```

Strictly stricter in direction, which is right. But of the five flips, four are on tagged tests and only **one** is a correct catch:

| test | tagged | flip | verdict |
|---|---|---|---|
| `ut_search_records_h4k` | yes | PASS→FAIL | correct — the target defect |
| `ut_search_records_015` (×2 runs) | yes | PASS→FAIL | **false positive** — `b.` abbreviation |
| `ut_search_records_014` | yes | PASS→FAIL | **false positive** — quoted-to-refute role |
| `ut_search_records_028` | no | PASS→FAIL | not applicable; my replay forced the tag |

My method differs from yours — I force the tag on every note rather than filtering to tagged tests, which is why my sample is 168 rather than 41 — so the count is a superset by design. The flips above are what matters, and three of the four tagged ones are not what the body describes.

`ut_search_records_015` passes in the shipped run only because the skill happened to phrase that note differently this time. The false positive is latent, not fixed.

## Smaller

- `ut_search_records_014` regressed pass → fail in the shipped run for an unrelated judge-scored reason (it misidentified `CFLT-9K2` and skipped the Step-4 cross-check). Undisclosed in the body. It looks like ordinary variance rather than something this PR caused — worth a line either way, since it is one of only three failures.
- The `.bat` fix reads correctly and the reasoning about `shift` consuming `%0` is sound; I have not exercised it, having no cmd.exe here. Worth someone on Windows confirming the actual invocation before merge, since that is the half of this PR no CI covers.

Requesting changes. The per-sentence idea is right — it is the splitter and the `as ROLE` pattern that need narrowing before a gating check can rely on them.

mercyokum:
Thanks for the thorough review — all three findings held up, and #1 was worse than it first looked.

**#1 (sentence splitter + false positives).** Fixed the abbreviation issue, then found the abbreviation fix alone doesn't actually save `ut_search_records_015`: its correctly-hedged output has the exact same two-sentence shape (claim sentence, then the SKILL.md-prescribed hedge as its own following sentence) as the case the per-sentence check exists to catch, and I couldn't find a mechanical signal that tells them apart. Withdrew the per-sentence check entirely rather than keep patching it — reverted to the original note-wide rule (issue #1284, unedited), kept only the widened claim-pattern additions (possessive kinship, plural kinship bare-introducing a name), which don't need sentence-level localization to be safe.

**#2 (as-ROLE firing on a rejected claim).** Moot now — gone with the per-sentence check it was guarding.

**#3 (mined test ships red).** You were right that this needed resolving, not shrugging off. Re-replaying with the corrected, note-wide-only validator, `ut_search_records_h4k`'s own captured output ("likely Amos's children" + the hedge as a following sentence) is legitimately compliant under the doctrine's actual bar — my original `judge_context` was asking the judge to fail phrasing that's actually fine. Loosened it to match: a hedge anywhere in the log entry qualifies the claim, not necessarily the same sentence. The test still exists to catch a genuinely bare, unhedged claim (issue #1912's real production text — no hedge anywhere in that log entry at all), which I also pinned permanently and deterministically in the validator's own unit-test file, independent of live model sampling.

**Regression claim.** Redid the replay with your exact methodology — every note in every committed run log, tag forced on all of them (168 notes, not my original 41). Result: **0 new false positives, 4 corrected** (your 3, plus `ut_search_records_h4k`'s own output per the point above).

**Smaller — `ut_search_records_014`'s unrelated regression.** Confirmed and now disclosed in the PR body: judge-scored, unrelated to this fix (the skill misidentified the returned record and skipped the age/role cross-check). Ordinary variance on that test.

**Smaller — the `.bat` fix.** I tested it directly on Windows across multiple real invocations, both reproducing the original bug and confirming the fix, before opening this PR — not just reasoned about it. Happy to walk through the exact repro if useful.

Full harness suite: 2858 passed, 3 skipped. `ruff` clean.

mercyokum:
Update: everything above is now landed and pushed. `ut_search_records_h4k` is marked `expected_outcome: xfail` pointing at #1967 (the real fix — a `research_log_append` writer-tool precondition — is tracked there rather than attempted in this PR, since the rule is already stated correctly twice in `SKILL.md` and a third prose restatement is unlikely to help).

Fresh, active run log committed (`v1_2026-08-27_15-10-12`), review_sample fully graded, all CI checks green including `runlogs`. Ready for re-review.

T-FEH (approved):
Second pass, on `98f33c0f` (current with main). Eval gate satisfied; annotation clean (35 corrections over the five sampled tests, all of which passed).

**All three blockers are fixed, and I verified each by replaying the exact notes rather than reading the diff.**

Withdrawing the per-sentence check was the right call, and it did not take the fix with it:

```
main's ut_search_records_014 note (quote-to-refute, "indexer inference")  -> PASS
main's ut_search_records_015 note (the 'b.' abbreviation case)            -> PASS
controlled: hedged note with 'b.' between claim and marker                -> PASS
#1912 defect note ("likely Amos's children")                              -> FAIL
```

All four abbreviation probes now pass, where three failed last round. `ut_search_records_h4k` is `expected_outcome: xfail`.

**And the widened patterns are real, not decorative.** Replaying main's validator against this one over every `research_log_append` note in every committed search-records run log — **176 notes, zero verdict changes**. That initially looked like a no-op, so I went looking for the case the widening was written for: it only bites when a kinship claim appears *without* the word "household", which no committed note does. Constructed directly:

```
case                                       main   PR
possessive claim, no household word        PASS   FAIL   <- new catch (#1912's exact shape)
plural-introducing, no household word      PASS   FAIL   <- new catch
same, properly hedged                      PASS   PASS
bare pronoun "his wife Catherine"          PASS   PASS   <- deliberate exclusion holds
single "son Thomas"                        PASS   PASS   <- deliberate exclusion holds
```

Both scoping decisions the docstring argues for actually hold under test. That is the shape I wanted and could not get from the per-sentence version.

Net against main's latest run: 5 validator failures → 3. `_001`, `_013` and `_027` clear; `_010` and `_012` are unchanged pre-existing failures.

## Two things to settle before merge, neither large

**The mined test xpasses in the run you shipped.** `v1_2026-08-27_15-10-12` records `ut_search_records_h4k` as **xpass** — marked xfail, but the skill got it right that run. Across your four runs it went `fail → fail → xfail → xpass`. So the regression pin is not currently demonstrating the defect it was mined to hold, and a reader of the shipped log sees the opposite of the finding.

That is not a failure and the gate is happy, but it is worth a line in the body saying which run is the evidence for #1912 and that the newest one xpassed. Otherwise the next person reads an xpass and concludes the defect was fixed here, which it was not — no skill edit was made.

**A measurement was deleted rather than superseded.** The docstring's recorded evidence is gone from the repo entirely:

> Across the four committed run logs the marker appeared on 17 of 32 pre-1880 census runs and all 32 passed: `ut_search_records_012` carried it once in four runs, `_013` twice, and `_015` and `_027` never…

`git grep "17 of 32"` now returns nothing. CLAUDE.md's rule is that a measurement which disagrees with belief gets re-measured, not reworded — deleting it outright is the stronger version of the same move. If it is stale because the corpus moved, re-derive it and say so; if it still holds, keep it. Either is fine; silently dropping it is the one option that leaves the next reader with less than you had.

## Minor

- The same hunk converts five em-dashes to `--` in the docstring. Unrelated to the fix and it makes the diff harder to read against a file that uses em-dashes throughout. Worth reverting so the diff is only the change.
- `_014` regressed pass → fail against main, but on a *differently worded* note — this run's opens "Patrick Flynn, Head of Household" flat and carries no marker anywhere, where main's said "indexer inference". I checked that the current validator passes main's wording, so this is skill variance rather than the false positive I raised last round. Defensible catch; just not mentioned in the body alongside the three that cleared.
- The `.bat` fix reads correctly and I still have no cmd.exe to exercise it. Someone on Windows confirming the real invocation before merge remains the one part of this PR nothing covers.

Approving. The validator is in better shape than either previous version — the two new patterns catch #1912's shape, the exclusions hold, and 176 committed notes are unmoved.

T-FEH (changes requested):
**Withdrawing my approval above — I got one of its central claims wrong.** A second pass turned up two majors, and one of them is the thing I explicitly told you held.

## I was wrong about the scoping

My approval said: *"Both scoping decisions the docstring argues for actually hold under test."* They do not, and my probe set was the reason I thought otherwise. I tested `his wife Catherine` and `son Thomas` — but those are excluded by the **absence of `'s`** and by **plurality**, not by the capital-letter scoping the comments actually claim. I never probed a lowercase possessive or a plural noun followed by a lowercase word, then generalised from the two cases that could not have failed.

Both patterns are searched with `re.IGNORECASE`, which makes `[A-Z]` match any letter, so the scoping both comments promise is not implemented:

```
lowercase possessive: "the family's children"       main=PASS  PR=FAIL   <== new false positive
plural + lowercase:   "daughters in the index"      main=PASS  PR=FAIL   <== new false positive
                      "children cannot be distinguished from boarders"
                                                     main=PASS  PR=FAIL   <== new false positive
```

All three are ordinary genealogical prose, and all three now red a gating validator.

```
You wrote:    r"\b[A-Z][a-zA-Z]*'s\s+(?:son|daughter|child|children|wife|husband)\b",
Change it to: r"\b(?-i:[A-Z])[a-zA-Z]*'s\s+(?:son|daughter|child|children|wife|husband)\b",
```
```
You wrote:    r"\b(?:sons|daughters|children)\s+[A-Z]",
Change it to: r"\b(?:sons|daughters|children)\s+(?-i:[A-Z])",
```

With those, every true positive stays caught (`Amos's children`, `sons Thos T McElwee`), all three false positives clear, and the two exclusions still hold.

## And nothing guards the new patterns

I deleted **both** new tuples from `_CLAIM_PATTERNS` and re-ran the pinned suite:

```
17 passed
```

The PR's central code change is unguarded. The cause is in the fixtures: line 116 reads `"Census household of Amelia Jackson, Amite County MS."` — `_HOUSEHOLD_MENTIONS` matches `household` and short-circuits before the new patterns are ever consulted. All three new fixtures do this.

```
You wrote:    "Census household of Amelia Jackson, Amite County MS."
Change it to: "Census return for Amite County MS."
```

That one string is what makes the deletion fail a test instead of passing 17.

This is the same shape as the finding I opened round 1 with, arriving from the other direction: the fix is real — I confirmed both new patterns catch #1912's shape where main passes it — but a real fix with no guard is one refactor away from being silently removed, and a PR whose stated purpose is *regression coverage* should not leave its own new logic uncovered.

## What still stands from my approval

The three round-1 blockers are genuinely fixed — I re-verified by replaying the exact committed notes, and that part I tested correctly. The abbreviation and quote-to-refute false positives are gone, `h4k` is properly `xfail`, the #1912 defect still fails, and 176 committed notes are unmoved. The two earlier "settle before merge" notes also stand: the shipped run records `h4k` as **xpass**, and the `17 of 32` measurement is deleted rather than superseded.

One correction to those, too: `xfail_reason` cites *"5 of 8 pre-1880-census-household-tagged tests failed this exact way"*. In `v1_2026-08-27_11-50-40` there are 9 tagged tests and 6 fail this validator (`_010, _012, _013, _014, _027, h4k`). It should read 6 of 9.

Requesting changes on the two majors. Both are one-line fixes, and neither touches the run-log snapshot.

mercyokum:
Both majors fixed and verified the way you found them.

**#1 ([A-Z] under IGNORECASE).** Fixed with `(?-i:[A-Z])` in both patterns — locally disables case-insensitivity for just that character class. Confirmed `the family's children` and `daughters in the index` no longer match, while `Amos's children` and `sons Thos T McElwee` still do. Pinned all three of your exact probes as constructed test cases.

**#2 (new patterns unguarded).** You're right — every fixture exercising them also said 'household', so `_HOUSEHOLD_MENTIONS` was silently doing the work. Reworded the three fixtures to drop 'household' (`... entry filed under NAME ...`), and added a dedicated no-hedge/no-household possessive-kinship offender case, since none isolated that pattern either. Verified the same way you found the gap — temporarily deleted both new tuples from `_CLAIM_PATTERNS` and reran: now 2 tests fail (were 0). Restored, all 21 pass.

Neither touches the run-log snapshot — confirmed via `diff_snapshot_vs_disk`, empty diff, so no re-run. Full harness suite: 2883 passed, 3 skipped. `ruff` clean.

On the "5 of 8 → 6 of 9" `xfail_reason` correction: that one *does* live in the tracked test JSON, so fixing it would invalidate the shipped run log and cost another full paid run for a wording-only change. Since you're the one who can make that call — would you rather I fix it and eat the re-run, or apply `eval-cosmetic-skip` for it? Your call either way.

T-FEH (approved):
Third pass, on `e702d1dc`. Harness **2884 passed, 2 skipped**; `ruff` clean; 176 committed notes replayed with zero verdict changes.

Both majors are fixed. I tested the mechanism this time rather than an outcome, because the reason I had to withdraw my round-2 approval was probing exclusions that could not have failed either way.

## 1. The `[A-Z]` scoping now actually scopes

`(?-i:[A-Z])` is in on both patterns. The table that matters is the one where main and this branch differ, and where a broken fix would show:

```
kind                     case                              main   PR
LOWERCASE must NOT trip  the family's children             PASS   PASS
LOWERCASE must NOT trip  daughters in the index            PASS   PASS
LOWERCASE must NOT trip  children cannot be distinguished  PASS   PASS
TRUE POSITIVE must trip  Amos's children                   PASS   FAIL
TRUE POSITIVE must trip  sons Thos T McElwee               PASS   FAIL
EXCLUSION must NOT trip  his wife Catherine                PASS   PASS
EXCLUSION must NOT trip  son Thomas                        PASS   PASS
HEDGED must NOT trip     hedged possessive                 PASS   PASS
```

The three lowercase false positives clear while both capitalised true positives still fail — so the fix distinguishes case, which is the property the comments claim and the thing my round-2 probe never tested. #1912's shape is still caught.

I also checked whether the same bug survives anywhere else rather than assuming the instance was the class. All five pattern groups — `_ROLE_ASSERTIONS`, both new tuples, `_HOUSEHOLD_MENTIONS`, `_INFERENCE_MARKERS` — carry no unscoped `[A-Z]` under `IGNORECASE`.

## 2. Both new patterns are pinned, and pinned separately

I deleted them one at a time as well as together, since a single test covering only one would have left the other unguarded:

```
drop BOTH new tuples   -> 2 failed, 19 passed
drop only POSSESSIVE   -> 1 failed, 20 passed
drop only PLURAL       -> 1 failed, 20 passed
```

Each tuple has its own guard. That is the coverage the round-2 version was missing, and it is the stronger version of what I asked for.

The corpus is unmoved: replaying main's validator against this one over all 176 `research_log_append` notes in every committed search-records run log gives `PASS→PASS 141, FAIL→FAIL 35, changed 0`. And `ut_search_records_h4k` is still `expected_outcome: xfail`.

## One minor still open

`xfail_reason` in `whitfield-1850-household.json` still reads **"5 of 8"**. It names a specific run log, so it is checkable, and it does not check out — in `v1_2026-08-27_11-50-40` there are **9** tagged tests present and **6** fail this validator:

```
tagged tests present: 9   failing this validator: 6
```

`6 of 9`. Small, but it is a measured figure in a durable artifact citing a file that says otherwise, and it is free to correct.

## Approving

The two majors are properly closed and the second one is closed more thoroughly than requested. The `xfail_reason` figure is worth fixing on the way past rather than holding the PR for.

Still open from earlier rounds, neither mine to clear: the shipped run records `h4k` as **xpass**, so the pin is not currently demonstrating the defect and the body should say which run is the evidence; and the `17 of 32` measurement deleted from the validator docstring should be re-derived or restored rather than dropped. The `.bat` fix reads correctly and I still have no cmd.exe — someone on Windows confirming the real invocation remains the one part of this PR nothing covers.

T-FEH:
Three additions to my approval above. None changes the verdict; two are comment-accuracy fixes worth landing with the `xfail_reason` one.

## The round-2 bug had a live occurrence, which strengthens the case for the fix

I reported it from constructed strings. It also happened in a committed run log: `ut_search_records_028` in `v1_2026-08-24_16-32-45.json` (still on `main`) wrote

> …with numerous **children c**hristened 16 Nov 1834 in a mass baptism…

```
buggy pattern  \b(?:sons|daughters|children)\s+[A-Z]        -> matches 'children c'
fixed pattern  \b(?:sons|daughters|children)\s+(?-i:[A-Z])  -> no match
```

So `(?-i:[A-Z])` removed a false positive that real production text had already produced, not only a hypothetical one.

## Two comments still describe behaviour the code does not have

Both are latent — neither shape occurs in any committed note — so this is about the comments, not the patterns.

**The possessive pattern trips on the exact phrase the neighbouring comment protects.** `_ROLE_ASSERTIONS`'s comment eight lines above cites "his wife Catherine" as a legitimate tree-side reference the rule "must not trip on". The possessive form does trip:

```
"Anchored on George's wife Catherine, already in the tree."   main=PASS  PR=FAIL
```

The behaviour is defensible — a pre-1880 note naming a kinship relation with no hedge anywhere is arguably worth flagging either way. It is the claim of exclusion that is wrong.

```
You wrote:    # is common in a legitimate tree-side reference this rule must not trip on.
Change it to: # is common in a legitimate tree-side reference. This does not exclude every
              # tree-side mention -- "George's wife Catherine" matches -- which is accepted:
              # a pre-1880 note naming a kinship relation with no hedge anywhere is worth
              # flagging either way.
```

**The plural pattern matches any capitalised word, not a name.**

```
"Two children Under 10 were enumerated."   main=PASS  PR=FAIL
```

`\s+` also spans newlines, though no committed note has one there.

```
You wrote:    # A plural kinship noun immediately introducing named individuals — "plus
Change it to: # A plural kinship noun immediately followed by a capitalised word, which in
              # a census note is nearly always a name being introduced — "plus
```

## And a wider replay agrees with the narrow one

A second pass over **297 notes across 9 run logs** — patterns loaded by AST from both validator files rather than hand-copied — also finds **zero** verdict changes between `main` and this branch, matching the 176-note result in my approval. The new patterns fire on 3 of 297 notes, all of which also mention a household, so they change nothing on today's corpus and their value is forward-looking. The PR body already says this; it is worth keeping said.

johnmarkpeterbrown:
Reviewed this as the second pass. The substance holds up — I re-ran your mutation checks and the corpus replay independently, and both new patterns are genuinely load-bearing. Four things to land, then I'll approve.

**Your question about the re-run: use the label, not a re-run.** The `xfail_reason` edit is wording only, so it's exactly what `eval-cosmetic-skip` is for. One catch on ordering — a new commit voids the bypass, so the label has to go on **after** your final push, not before. Put all four fixes below in one push, then ask for the label. (If it somehow ends up applied early, it has to be removed and re-added — re-applying a label that's already there does nothing.)

**1. The plural-kinship pattern misses its own case if there's a comma.**

You reworded the fixture to drop "household" so this pattern is the only thing catching that note. That makes the spacing load-bearing: the pattern needs whitespace directly between the noun and the name, so `plus sons, Thos T McElwee and Stephen McElwee` slips past the gate entirely. A comma, a colon, or a dash all do it.

```
You wrote:      r"\b(?:sons|daughters|children)\s+(?-i:[A-Z])",
Change it to:   r"\b(?:sons|daughters|children)\b[\s,:;–—-]+(?-i:[A-Z])",
```

I checked this replacement against your pinned suite — all 21 still pass — and against every note in the committed run logs, where it adds no new flags. A period is deliberately left out of that character class, since a sentence boundary is the adjacency the note-wide design chose not to treat as adjacent.

**2. The re-measured docstring figures are each one too low.**

Following the method the docstring itself gives, I get 44 notes, not 43 — it breaks down as 8 + 9 + 9 + 9 + 9, with the oldest log short one because it predates the new test. The `21 fail` and the `_014` once / `h4k` twice detail both match exactly.

```
You wrote:      PR, the 9 `pre-1880-census-household`-tagged tests produced 43 logged
Change it to:   PR, the 9 `pre-1880-census-household`-tagged tests produced 44 logged
```

```
You wrote:      searches. The marker appears in 25 of 43; outcome is 22 pass / 21 fail.
Change it to:   searches. The marker appears in 26 of 44; outcome is 23 pass / 21 fail.
```

```
You wrote:      3 of the 25 marker-present runs still failed (`ut_search_records_014`
Change it to:   3 of the 26 marker-present runs still failed (`ut_search_records_014`
```

This file needs no label — it isn't snapshotted, so these are free.

**3. Two comment fixes from the last round didn't make it in.** Both still read as they did when they were flagged, and both are free to change. Using the replacements already written for you:

```
You wrote:    # is common in a legitimate tree-side reference this rule must not trip on.
Change it to: # is common in a legitimate tree-side reference. This does not exclude every
              # tree-side mention -- "George's wife Catherine" matches -- which is accepted:
              # a pre-1880 note naming a kinship relation with no hedge anywhere is worth
              # flagging either way.
```

```
You wrote:    # A plural kinship noun immediately introducing named individuals — "plus
Change it to: # A plural kinship noun immediately followed by a capitalised word, which in
              # a census note is nearly always a name being introduced — "plus
```

**4. The `xfail_reason` credits the wrong reviewer.** Separate from the `5 of 8` figure, same string. Every review on this PR is T-FEH's; `chrisedeson` appears nowhere on it. The PR body has it three times too. One edit covers both:

```
You wrote:      5 of 8 pre-1880-census-household-tagged tests failed this exact way
Change it to:   6 of 9 pre-1880-census-household-tagged tests failed this exact way
```

```
You wrote:      PR #1946 review (chrisedeson) found this shipping red
Change it to:   PR #1946 review (T-FEH) found this shipping red
```

One note on the `.bat` half: I confirmed the bug by reading it — the script resolves its own folder after argument parsing, so a bare `shift` really does point it at the zip's folder — but I didn't execute it, and neither did the last reviewer. Your Windows testing is the only thing covering that half, so it's worth saying in the body that you exercised both the break and the fix. Nothing in CI touches it.

johnmarkpeterbrown (commented):
Let me know. Thanks

mercyokum:
All four items from johnmarkpeterbrown plus T-FEH's two comment-accuracy fixes are in, pushed as `f16da05d`.

**1. Comma-separator gap.** Fixed with `[\s,:;–—-]+` between the noun and the name — `plus sons, Thos T McElwee` is now caught. Pinned as a new regression case, and re-verified against the real corpus (this time correctly unpacking a batched `research_log_append` `ops` call I'd missed before): zero new flags, matching what you found.

**2. Docstring figures.** You were right and I was off by one — traced it to that same batched-`ops` call: one entry carried a marker my earlier ad-hoc script silently dropped because it only read the flat single-entry shape. Independently re-derived rather than taken on trust: **44 logged searches, marker in 26, 23 pass / 21 fail** — matches your numbers exactly.

**3. Both comment fixes** (the possessive-kinship exclusion claim, the plural-kinship 'named individuals' claim) — landed as written.

**4. `xfail_reason`** — re-verified '6 of 9' directly against `v1_2026-08-27_11-50-40.json`'s validator results rather than accepting either count, and confirms your number. Attribution fixed to T-FEH throughout the test JSON and PR body.

T-FEH — thanks for confirming the round-2 bug had a live prior occurrence (`ut_search_records_028`'s 'children christened'), and for the same two comment corrections; both are in above.

Only the `xfail_reason` edit touches the tracked snapshot — everything else is validator/comment-only. **This is the final push for this round — could you apply `eval-cosmetic-skip` now?**

clack391:
Not a review of this PR, just a heads-up from tracing a ruling chain while reviewing #2004.

`eval/tests/unit/search-records/whitfield-1850-household.json`'s `xfail_reason` for `ut_search_records_h4k` says the marker is "blocked on issue #1967, which moves enforcement into a `research_log_append` writer-tool precondition. Remove this marker when #1967 lands and this test passes on its own."

#1967's decision of 2026-08-27 went the other way: "Close the verified defect and stop there. **No content gate on `notes`**, no shadow warning, no per-record/per-country table," with the shadow-warning option rejected as "still a linguistic check on a free-text field." So the stated removal condition cannot be met as written, and whoever reads the marker next will go looking for a precondition that was decided against.

The marker itself still looks right to me; it is only the reason that has gone stale. Probably worth rewording it to point at what actually unblocks it before this merges.

mercyokum:
Good catch — verified directly against #1967 rather than taking it on your word, and you're right on both counts.

Confirmed #1967's actual ruling: no tool-boundary content gate (measured against the full corpus, n=776 — a hard pre-1880/US gate would refuse 41% of `record_search` calls, isn't generalizable outside the US, and the year/country signal is author-supplied and optional). #1967's real finding is that this is *partly a grading defect* — the judge contradicts this test's own `judge_context`.

That sent me back to check whether my earlier `judge_context` loosening (round 1) had already fixed that contradiction — it hadn't. Confirmed directly: `v1_2026-08-27_13-03-15.json`'s judge rationale for `h4k` scores a note reading "...likely sons... family structure inferred, not stated" (hedge in a later sentence — `judge_context`'s own "correct shape") as Correctness 1, reasoning it "does not retroactively qualify the earlier bare kinship claim" — the literal opposite of what `judge_context` says. So the contradiction is real and still live, not something this PR already resolved.

Rewrote `xfail_reason` and the matching PR body bullet to describe what #1967 actually does (reconcile `judge_context`/`rubric.md`, not build a tool precondition) and cite the corpus measurement against the tool-gate idea. Pushed as `0b556e27` (merged with an unrelated main update, `664e194f`).

johnmarkpeterbrown (approved):
Approving. Reviewed at `51dcbca8`.

All four items from my last pass landed, and I verified each against the code rather than the commit messages:

- The comma-separator fix is in — and pinned as its own offender case, so the gap can't silently reopen. Pinned validator suite is 22 passing, up from 21.
- The docstring figures now read 44 logged searches, 26 of 44 with a marker, 23 pass / 21 fail.
- Both comment corrections landed.
- `xfail_reason` reads `6 of 9` and no longer misattributes the review.

I also checked the `xfail_reason` rewrite that came out of clack391's catch against #1967 itself rather than against the summary of it. It holds: the issue is open, its ruling does say no content gate on `notes`, the 41% / n=776 measurement is there, and its `Touches:` list includes this test file — so "blocked on this PR merging" is accurate.

I've applied `eval-cosmetic-skip` and confirmed `runlogs` went from fail to pass on this same head. All 13 checks are green. **The bypass is void on any further push** — a new commit turns it off from the event, and the label gets stripped — so if anything else lands here it needs removing and re-adding to fire a fresh event. Re-applying it while it's still showing does nothing.

Two notes for the record, neither blocking:

**The `.bat` half of the body is now out of date.** PR #1905 landed on main after the earlier approval and fixed the same bug differently — `%~dp0` is read before `:parse` now, with a packaging lint enforcing that no `.bat` reads `%0` after a shift. Since nothing else in the file reads `%0`, the `shift /1` change here is no longer what's holding the fix up; it's harmless reinforcement. Worth a one-line body correction so the record isn't misleading — but **don't push just for that**, because it would void the label and cost the whole dance again. Fold it in only if something else needs a push.

**T-FEH's approval reads `744682b2`, three pushes and a main merge behind this head.** Approvals aren't dismissed on push here, so it still counts toward the two required. Everything substantive that landed after it — the comma fix, the figures, the comments, the `xfail_reason` rewrite — I've now reviewed, but that's my read covering the gap rather than a second reviewer's.

Issue #1323, opened 2026-08-05
isPrincipal: one decision — what record_search should default to, and what search-records should teach


chesworthrm:
> **Reviewed 2026-08-26 before junior handoff.** No decision needed. The ruling stands,
> but its supporting measurement was false and is corrected inline; item 2's site list
> was one file and is actually four. Original body follows below.

> **Board pass 2026-08-20 — every blocker this issue has ever named is now closed.**
> This is **not** a measure-then-decide task and there is nothing left to ask: the
> lead ruled on 2026-08-11 and `needs-decision` was removed on 2026-08-17. It is
> work, and item 2 can start today. The banners below are kept as history; read this
> block instead of them.

## Blocker status, verified 2026-08-20

| Named as blocking | State |
|---|---|
| issue #1093 | closed completed, on `main` (`803aa26b`) |
| issue #1005 | closed completed |
| issue #1089 | closed **not planned** — folded into this issue's ruling |
| issue #1326 | closed completed |
| issue #1402 | closed completed |
| issue #1525 | closed completed |
| issue #1592 | closed completed |

Nothing is waiting on anything. The "what still holds this in Backlog is the
`search-records` eval slot" line is true of **item 3 only**.

## Correction (task review, 2026-08-26): the "zero times" figure is wrong

The body and the ruling both say `isPrincipal` was set **zero times** across 3,381
e2e `record_search` calls, and that item 3's value follows from that. Re-measured
over the committed e2e corpus:

```sh
python3 -c "
import json,pathlib
tot=st=t=f=0
for p in pathlib.Path('eval/runlogs/e2e').glob('*/run-*.json'):
    if p.name.count('.')>1: continue
    for c in json.loads(p.read_text(encoding='utf-8')).get('tool_calls') or []:
        if not str(c.get('tool','')).endswith('record_search'): continue
        tot+=1
        a=c.get('args') or {}
        if 'isPrincipal' in a:
            st+=1; t+=a['isPrincipal'] is True; f+=a['isPrincipal'] is False
print('calls',tot,'set',st,'true',t,'false',f)"
```

measured 2026-08-26: **1,977 calls, 54 set it — 49 `true`, 5 `false`**, across 19 of
159 runs. Re-run the command rather than quoting these.

**The ruling survives, with its two halves re-weighted:**

- `isPrincipal: true` is already discovered from the schema description alone
  (49 uses, e.g. `anders-monsen-ancestry`). Teaching it in the skill documents
[...]

PR #1947, opened 2026-08-26
record_search: name both isPrincipal moves instead of recommending omission (#1323 item 2)

francis-2008-happy:
## Summary

**Tier: Normal.** Lands **item 2 only** of issue #1323 — the tool-side guidance. Issue #1323 stays open for item 3 (the `search-records/SKILL.md` teaching).

`isPrincipal` supports two distinct search intents, and the schema recommended neither: its guidance sentence said *"Omit the parameter to return both — the broadest set, recommended for most natural-language searches."* Per the lead's 2026-08-11 ruling the default does not move — omitted still returns both — but that sentence is why the `false` pivot stayed invisible. Across the committed e2e corpus, re-measured today, **1,977 `record_search` calls set it 54 times: 49 `true`, 5 `false`**. All five `false` calls sit in three fixtures, and every one is the same shape — find a relative through a person already known.

This rewrites the recommendation into the two moves and when each applies, in the schema and in the spec, and adds the reasoning to the spec beside the existing bullets. **No wire behaviour changes:** no default moves, no parameter is added, `buildSearchUrl` is untouched.

## Review guidance

**Start here:** the `isPrincipal` description in `packages/engine/mcp-server/src/tools/record-search.ts` and the guidance paragraph in `docs/specs/record-search-tool-spec-v2.md`. The claim to check is what the shipped string now says: that the response does not mark who the principal is, so `record_read` is required to get at the relative, and that `false` matches on the **known person's own name** while `fatherGivenName`/`spouseGivenName` match on the **target's indexed relative fields**. That is a statement about which field each parameter matches, not about which route is better — the earlier draft asserted `false` was "the only route" and that the anchors "need the target's name", and both were deleted in round 1 as unmeasured. `validateInput` accepts `{recordCountry, fatherGivenName, fatherSurname}`, so a relative anchor demonstrably does not need the target's name.

**Unsure about:** how much of the reasoning belongs in the schema versus the spec. I put the operational instruction in the schema (it is billed on every advertisement) and the evidence and rationale in the spec. The schema grew 426 characters, which is the single largest thing to push back on here.

## Plan

**Didn't change:** The default — `isPrincipal` stays omitted-by-default, returning both. That was the ruling and it reversed the tool half's own recommendation; I did not re-derive it. Item 3 (`search-records/SKILL.md`) is also untouched: editing it flips every one of that skill's run logs inactive and forces a fresh `make eval-skill SKILL=search-records` (28 tests, measured $8.73–9.75), and three cards contend for that slot. Per the issue, item 3 should ride whichever PR pays for #1642's run.

**Acceptance check:** `packages/engine/mcp-server/tests/packaging/tool-description-length.test.ts` fails on `main` and passes here — `DOCUMENTED_TOTALS` re-derived from `allToolSchemas`, `record_search` **14043 → 14469**. That test is whole-tool by design and fires on any description edit, which is what makes it the acceptance check rather than a nuisance. I did not shorten another description, add an `EXEMPT` entry, or raise `CEILING`.

**Deviated from the plan:** Twice, both caught by guards rather than by me.

1. I first put the precise probe totals into the spec. `measured-figures.test.ts` failed — they trace to no probe artifact, since these came from `dev/try-record-search.ts` and not the sectioned probe. The fix was not an `EXEMPT` entry; re-reading the issue, it asks for the totals in **this body**. The spec now states the direction and the numbers live below.
2. I wrote `(issue #1323)` into the spec and `doc-links.test.ts` failed on its state-the-fact-not-the-ticket ratchet. Removed.

## The verification the issue required

The issue set a stop condition: if omitted and `--not-principal` return identical totals, `false` is ignored upstream and both the schema and the spec are already wrong — report rather than teach the pivot. Run after `make e2e-login`, collection 2177294, `Melquiades Oliveira`, `count 50`, reading `totalMatches`:

| `isPrincipal` | `totalMatches` |
|---|---|
| omitted | 50,369 |
| `true` (`--principal`) | 1,456 |
| `false` (`--not-principal`) | 48,913 |

**`1,456 + 48,913 = 50,369` exactly.** The two roles partition the same match set and neither is ignored, so the pivot is real and the stop condition did not trigger.

Two things this corrects. The `~84%` figure in the issue body was measured on `q.isPrincipal=on`, a different mechanism from the `true`/`false` the tool sends (`record-search.ts:363`), so the `false` branch had genuinely never been measured. And on this collection `true` cuts **97%**, not 84% — which is why the spec states the direction rather than a percentage, the ratio being collection-specific.

Corpus re-measurement, run rather than quoted from the body:

```
calls 1977 | set 54 | true 49 | false 5
false fixtures: manoel-oliveira-daughter 2, anna-macek-son 2, victoriano-macatangay-parents 1
```

This confirms the issue's own inline correction.

A reader who greps the run logs finds **seven** `"isPrincipal": false` hits, not five. Five are real calls; two are log-only artifacts. In `jimmie-jewel-neal` the log entry does not misreport a call — it **fabricates** the parameter: it matches `tool_calls[247]` field for field, and that `record_search` omitted `isPrincipal` entirely (its 5 calls that do set it all send `true`). The `anna-macek-son` extra is different again and benign — that entry accurately summarises two real `false` calls, so a grepper double-counts rather than finding a phantom. The count of 5 above walks real tool calls and stands either way. Unrelated to this PR; recorded so nobody reads it as a miscount. (Corrected from an earlier draft of this body that described jimmie's as a `true` logged as `false`.)

## What would have caught this

**Nothing does.** `eval/harness/harness/snapshot.py:33` excludes `packages/engine/mcp-server/src/` from the run-log snapshot, so rewriting a tool description changes what the model reads in 7 of 25 unit suites — `search-records` among them — while no run log flips inactive and no suite goes red (issue #1770, `nothing-checks`). `make engine-test` proves only that nothing else broke; it cannot show that this wording helps, or that a future edit reverting it would be noticed.

The one guard that does bind is the description-length test, and it binds on token cost rather than on correctness.

## Test plan

- [x] `make engine-test` — 112 files, **2718 passed**.
- [x] `make test-all` — ran, **not fully green, and not because of this diff.** 2843 passed, 9 failed. All 9 are `eval/harness/tests/unit/test_setup_feedback_case.py`, every one on `scripts/setup-feedback-case.sh: line 83: unzip: command not found` — `unzip` is not installed on this machine, and I confirmed zero failures outside that file. Everything else green: viewer-ui 271, electron 97, web 10, control-plane pytest 210, CRUD-UI 168, engine 2718, all typechecks.
- [x] No skill run-log snapshot touched, so no `make eval-skill` run is owed. `mcp-server/src/**` sits outside the run-log gate.
- [x] Routing-pair box — N/A, no skill `description` or DO-NOT clause changed.

## Follow-on issues

**Folded in / filed instead:** Nothing filed. Item 3 stays on #1323 rather than becoming its own card — per the issue, two PRs on one issue is expected here and the card should not be split to match.


chesworthrm (changes requested):
Good PR. The body is unusually honest, the corpus figures reproduce to the call (I re-walked the run logs and got `1977 / 54 / 49 / 5` with the same three fixtures), and the acceptance check genuinely fires — I set `DOCUMENTED_TOTALS` back to `14043` and watched it fail with `docstring says 14043, actual 14427`. Suites are green on the tree merged with `main`: engine 2718 passed, tsc clean, harness 2853 passed. Your 9 `test_setup_feedback_case.py` failures did not reproduce here, so your `unzip` diagnosis is right.

Two changes needed before this can merge. Both are in the sentence you flagged, and one replacement fixes both.

## 1. The description points the model at a field that does not exist

`record_search` never tells the model which person is the principal. I ran the shipped mapper against a record where the raw FamilySearch data had `principal: true` on one person and `principal: false` on the other. The string "principal" appears nowhere in the output, at any depth. The conversion to simplified GedcomX drops that flag, so it is gone before staging even happens — this is not the staged-response slimming, it is absent in every case.

So "read the unknown relative off the `principals` of the results" cannot be followed. The route that works is the one your own `manoel-oliveira-daughter` fixture takes: search, then `record_read` a hit and read the persons and relationships it returns.

Worth knowing: `record_read` does not label the principal either. It returns the graph and the model works it out.

## 2. "The only route" is an absolute we have not measured

`fatherGivenName` does not require the target's name. The validator asks for one of `surname`, `recordCountry` or `batchNumber`, and the URL builder emits the relative parameters regardless of whether a target name is present. So the call is legal.

More importantly, `search-records` already teaches the opposite. In `references/search-strategy-levers.md`, the "Search by parent" row says to clear the principal name and fill the father/mother fields, for the case where the principal "may have been 'Baby' or stillborn" — the unnamed target. Another row calls it the primary recovery move when the subject "is indexed under names you can't guess." `SKILL.md` tells the model to read that file.

Shipping "cannot express it" puts two surfaces the model reads in direct contradiction.

I could not settle which is right. It needs a live token, and the only artifact touching it — `dev/explore-father-absent-principals-records.ts` — is explicitly marked exploratory and records a query shape that returned an empty set. So do not resolve this by asserting the other direction either. State the mechanism difference, which is true whichever way a measurement lands.

## The edits

In `record-search.ts`:

```
You wrote:    Finding a relative THROUGH a person you already know is `false` — search the
              known person, then read the unknown relative off the `principals` of the
              results. That is the only route when the relative's own name is not yet
              known, so `fatherGivenName`/`spouseGivenName` cannot express it. Omit to
              return both.

Change it to: Finding a relative THROUGH a person you already know is `false` — the
              response does not mark who the principal is, so `record_read` a hit and read
              its `persons`/`relationships` for the relative. It matches on the known
              person's own name, where `fatherGivenName`/`spouseGivenName` match on the
              target's indexed relative fields. Omit to return both.
```

In `docs/specs/record-search-tool-spec-v2.md`:

```
You wrote:    `false` searches the *known* relative and reads the unknown person off the
              principals of the results. When the target has no name yet — "did this
              couple have another daughter?" — the anchors cannot express the question and
              `false` is the only route.

Change it to: `false` matches on the *known* person's own persona; `record_read` on a hit
              then gives the record's persons and relationships. Which route is better
              when the target has no name yet is not measured — `search-strategy-levers.md`
              prescribes clearing the principal name and filling the parent fields for
              exactly that case — so name the difference rather than calling either the
              only route.
```

That description edit is +40 characters, so `DOCUMENTED_TOTALS` becomes **14467** and the docstring's `19,430` becomes **19,470**.

## Two smaller things

The spec says "Precise totals belong in the probe artifact, not here." They are not in it — `dev/measured-figures.json` has no leg for this parameter and no trace of that collection. Either add the leg or drop the sentence, because as written it sends the next reader somewhere the numbers are not.

The docstring insert leaves "This figure has been an increase and a saving at different points in the same branch" with nothing to attach to. It used to refer to the 208-character saving; the nearest antecedent is now 19,430.

## Not blocking, for the body

Your count of 5 `false` calls is the right one — it counts real tool calls. But a reader who greps the run logs will find a sixth, in `jimmie-jewel-neal`. That one is a `research_log_append` entry recording `isPrincipal: false` for a search that was actually issued with `isPrincipal: true`. The log misdescribes the call. Nothing to do with this PR, but worth a line so nobody thinks you missed one.

## Housekeeping

Needs an update from `main` — currently behind. And my approval alone will not unblock it: CODEOWNERS puts `.ts` under both senior-developers and senior-genealogists.

One note in your favour. Nothing in CI would have caught the first item — it only shows up if you run the mapper. That is a decent illustration of the gap your "What would have caught this" section already names.

chesworthrm (approved):
Approving. Both blocking items are addressed and I re-verified rather than taking the diff at its word.

The description no longer references a field that does not exist, and it now points at `record_read` — which is the route your own `manoel-oliveira-daughter` fixture takes.

I re-derived the totals independently, summing the tool description plus every parameter description straight off `allToolSchemas` rather than trusting the test:

```
record_search 14467
person_search 5003
```

That confirms `DOCUMENTED_TOTALS`, the docstring's 19,470, and the 424 delta. The guard still binds — mutating the total back gives `docstring says 14427, actual 14467`. Engine 2718 passed, tsc clean, harness 2853 passed, all required checks green, and the branch is level with `main`.

## One wording fix before merge

The spec paragraph contradicts itself now, and that is my fault — the replacement I gave you started at the second sentence, so the lead-in was never in scope. It still reads:

> Why `false` is the load-bearing half, and why relative-name anchors **do not replace it**: `fatherGivenName`/`spouseGivenName` search the *target* with a relative named, **so they need the target's name**.

Three sentences later the same paragraph says which route is better "is not measured", and that `search-strategy-levers.md` prescribes clearing the principal name for exactly that case. Those cannot both be true.

```
You wrote:    Why `false` is the load-bearing half, and why relative-name anchors do
              not replace it: `fatherGivenName`/`spouseGivenName` search the *target*
              with a relative named, so they need the target's name.

Change it to: How `false` differs from the relative-name anchors:
              `fatherGivenName`/`spouseGivenName` match on the target's indexed
              relative fields, so a record that leaves those blank is not narrowed
              by them.
```

Not blocking — it is spec prose, and the surface the model actually reads is correct now. Push it whenever.

## Two notes, neither needing action

Deleting the "precise totals belong in the probe artifact" sentence resolves what I raised. The residual is that the live partition check is now unreproducible from the repo — it lives only in the PR body. If that leg ever goes into `dev/measured-figures.json`, the spec's "verified live against collection 2177294" claim gets something to stand on.

Whether a relative-anchored search with an unnamed target returns anything is still unmeasured. Saying so plainly, as the spec now does, is the right thing to ship.

Note that this does not merge on my approval alone — CODEOWNERS wants both senior teams on the `.ts` files.

francis-2008-happy:
Applied (`60045196`) — and thanks for catching it, though I should have caught it myself: I pasted your replacement in without re-reading the paragraph it landed in, which is exactly how a contradiction survives an edit.

You wrote it as your fault for scoping the replacement from the second sentence. Fair, but the sentence that was wrong was mine from the round before, and reading the whole paragraph after editing part of it is my job.

The paragraph now reads consistently: it states how the two mechanisms differ — the anchors match on the target's indexed relative fields, so a record leaving those blank is not narrowed by them — then that `false` matches on the known person's own persona, then that which route wins on an unnamed target is unmeasured, with `search-strategy-levers.md` prescribing the anchors for that case. No claim about needing the target's name anywhere in it.

`make engine-test` 2718 passed on this head. Branch level with `main`.

Still owed a **senior-genealogist** approval — that team is on the request and yours cannot satisfy the count for a `.ts` path.

clack391 (approved):
Approving. The substance is right and I verified rather than reading, including the two round-1 findings: you checked those before applying them, which is why they held.

Re-derived independently rather than re-running your commands: `record_search` 14467 and `person_search` 5003 straight off `allToolSchemas` (tool description plus every parameter description, plus a deep walk in case anything nested was being missed); `isPrincipal` 438 on `main` to 862 here; the corpus at 1977 calls / 54 set / 49 `true` / 5 `false` in the same three fixtures, stable under two different file-inclusion rules so the denominator is not doing the work; engine 112 files / 2718 passed, tsc clean, packaging 590 passed. Your `unzip` diagnosis also holds: `test_setup_feedback_case.py` is 11/11 green on a machine that has it.

I traced the round-1 claim three ways in code rather than only live. `principal?: boolean` sits on `FSPerson` (`src/types/record-search.ts:35`), the upstream shape; the output `RecordSearchResult` (`:240-289`) carries no principal marker and `primaryId` is the focus person, not the principal; `tree-shape.ts`'s closed allow-lists cannot hold one, so simplified GedcomX cannot either. And the instruction is correct specifically on the path users run: with `projectPath` supplied, which your own `projectPath` description calls the normal case, `delete r.gedcomx` fires at `record-search.ts:1103`, so `record_read` is genuinely required.

I also break-tested the acceptance check with shapes outside the one you and Robert used. Growing an unrelated non-`*Exact` description by 13 characters and leaving the pin alone fires it (`docstring says 14467, actual 14480`), so it is not vacuous. A compensating pair does not: removing " Omit to return both." from `isPrincipal` and adding 21 characters to `maritalStatus` leaves all four tests green with the guidance partly reverted. Worth knowing that the `*Exact` family is narrower than that, because `DOCUMENTED_LONGEST`'s ordering assertion does catch an offsetting pair inside it. Your body already says this guard binds on token cost and not on correctness, so none of that is a surprise, only a measured shape.

## Owed before merge, and it is one push

**1. The shipped string has a subject/verb slip.** `record-search.ts:1372` reads "Records ABOUT a person (their own birth/marriage/death) **is** `true`". The spec's parallel sentence at `:312-313` uses the plural. I applied both candidate fixes and measured the pin: `take` gives **14469**, `are` gives **14468**. Clarity is the stated justification for the 424 characters, so it is worth the two.

**2. The docstring narrative is now internally false in three ways** (`tests/packaging/tool-description-length.test.ts:13-25`), which matters more here than elsewhere because that comment is the only justification for `CEILING` and the file says of itself that it has been wrong about its own basis three times.

- The bolded lead still reads "The justification is clarity; the token saving is small ... a net saving of about 208 characters, under 2%", but the family now sits at 19,470 against a 19,254 start. That is a net increase of 216, so there is no saving left to call small.
- "#1323 then spent 424 of it back" makes "it" the 208-character saving, and 424 is larger than 208, so the quantity as written cannot exist.
- The pre-existing `#1771` sentence is now orphaned onto the tail of your new paragraph on one 135-character physical line, which reads as though #1771 followed #1323. This is the same run-on you fixed one sentence earlier in round 2, reintroduced one sentence down.

The substance you added is honest, and "above where the family started" plus "Recorded as an increase rather than argued away" is the right call. It is the three sentences around it that were not reconciled.

**3. The body has four stale spots, all from the pre-review draft.** The one that costs something is "Start here". It tells the reviewer the claim to check is that `false` "is not reachable" via the relative anchors "because those search the target and so need the target's name", and that `false` "reads the unknown one off the results' principals". Round 1 deleted both clauses and the shipped string now asserts the opposite. I also refuted the premise directly by running `validateInput`: `{recordCountry, fatherGivenName, fatherSurname}` is accepted, so a relative anchor does not need the target's name. A reviewer following your instruction verifies text that is no longer there. The other three:

- "The schema still grew 384 characters" should be 424. 384 is the delta of the first branch commit; round 1's +40 was never carried into the body.
- The jimmie paragraph. That run has 38 `record_search` calls, of which 5 set `isPrincipal` and all 5 set `true`, alongside 6 logged query objects. The logged `false` is not one of the `true` calls: it matches `tool_calls[247]` field for field, and that call omitted `isPrincipal` entirely. So the log fabricated the parameter rather than recording a `true` as `false`, which is a sharper statement than the one in the body.
- "will find a sixth" undercounts. A text grep of the 159 primary run logs returns seven `"isPrincipal": false` hits: 5 real calls plus two log-only artifacts, jimmie's and one in `anna-macek-son`. The `anna-macek-son` one accurately summarises two real calls, so a grepper double-counts there rather than finding a phantom. Your count of 5 stands either way.

**4. The `#1323` comment still says `14043 -> 14427`.** Edit that comment in place rather than correcting it in a new one. Whoever takes item 3 reads that thread, and a correction posted downstream never reaches a reader of the original.

## One line worth adding to #1323 for item 3

`SKILL.md:277` mandates "at least one principal-only search (no `spouseGivenName`/`fatherSurname`/etc.)", which means drop the relative filters and go broader. `isPrincipal: true` is a role filter that goes about 97% narrower on the collection you probed. Those are opposite moves under one word, and the conflation has already happened once: `eval/tests/e2e/anders-monsen-ancestry/README.md:38` and `:83` record a run executing that fallback as an `isPrincipal: true` call, and the README treats them as the same thing. Until now the schema closed by recommending omission, which kept the collision inert. "Pick by intent, do not default to omitting" gives the model an active reason to reach for the filter, and an over-narrowed search that nils gets written into `research.json` as negative evidence.

Not a reason to hold this PR, and I am not asking you to. Item 3's scope already says to state the pivot against the lever it competes with, so this is the evidence for doing that, not new work here.

## Two things I found that are not yours to fix

Neither is filed as a new card. Both have existing homes and I have put the measurements there instead, so they sit where the work will happen.

- `prompt-budget.test.ts` does not see this PR's 424 characters at all. Recorded on #1275, whose trigger is the baseline file that would settle the scope.
- `search-strategy-levers.md` prescribes six shapes across five rows that `validateInput` rejects, and the rule turns out to be exact: every lever that clears `q.surname` loses the anchor. Recorded on #1642, which owns the eval slot that gates any fix, and which already carried two of those rows for an unrelated reason. Your round-1 point stands there and I say so: nothing requires the target's given name.

## Merge state, and a correction

This does not need a senior-genealogist. `CODEOWNERS:6-8` states it directly: "EVERY RULE NAMES BOTH SENIOR TEAMS, and a rule's owners are an OR: one approval from either team satisfies it ... the second is there so a senior of either kind can unblock a PR that spans both", corroborated at `.claude/skills/fill-ready/SKILL.md:817`. Robert is an active senior-developer, so the code-owner requirement was satisfied by his approval. What was outstanding is only the second approval under the ruleset's count of two, which mine now supplies. Sorry that belief sat on the thread through four messages; it is the reason this waited.

`dismiss_stale_reviews_on_push` is false, so my approval survives your next commit and you do not need to re-request anything. Please land the push above before merging rather than after.

One note in your favour on the largest thing you asked to be pushed back on. The 424 characters buy an instruction that is executable on the default path and that no other surface in the repo carries: after this, the schema string is the only thing that teaches this parameter, since the single hit in the whole plugin tree is a mapping row. I would spend it.

francis-2008-happy:
All four in (`368871fc`), and the `#1323` comment is edited in place rather than corrected downstream. Thank you for the merge-state correction — and no need to apologise for it; I repeated the same wrong belief back to you twice without ever opening `CODEOWNERS:6-8`.

## 1. The slip

`take`, not `are` — the spec's parallel sentence at `:312-313` already used it ("Records **about** a person take `true`"), so this makes the two read identically rather than merely both being grammatical. Pin re-derived: **14469**, matching your measurement.

## 2. The docstring

All three confirmed, and the third is the one that stings: it is the same run-on I fixed one sentence earlier in round 2, reintroduced one sentence down. I clearly edited the paragraph without re-reading what my insert had collided with — twice now on the same file.

Reframed the bolded lead so it no longer claims a saving, stated the delta against `isPrincipal` itself rather than against the 208, and put the `#1771` sentence back in its own paragraph.

**Your figures moved because of item 1, so I re-derived rather than pasting them:** the `take` fix adds 2, making the delta **426** (not 424), the combined figure **19,472** (not 19,470), and the net **+218** over the 19,254 start. Worth flagging since you measured 862 for `isPrincipal` and it is now 864.

## 3. The body

The "Start here" was the costly one and you are right about why: it told a reviewer to verify two clauses round 1 had deleted, one of which the shipped string now contradicts. Rewritten to point at what actually ships — which field each parameter matches — and it names your `validateInput` refutation, since "the anchors need the target's name" is the specific thing that was wrong. 384 → 426.

**The jimmie paragraph was wrong about the mechanism, not just imprecise, and I checked before rewriting it.** `tool_calls[247]` is a `record_search` with no `isPrincipal` key at all, and the five calls that do set it all send `true`. So the log **fabricated** the parameter; it did not record a `true` as a `false`. Your framing is strictly sharper and mine was simply incorrect.

Your seven also reproduces: 7 text hits, 5 real calls, 2 log-only. And the `anna-macek-son` extra is the benign kind — it accurately summarises two real `false` calls, so a grepper double-counts there rather than finding a phantom. Both now in the body, with a note that it corrects an earlier draft.

## 4. The `#1323` comment

Edited in place, and it now says so: `14043 → 14469`, with a parenthetical naming what moved it and when. I also appended your item-3 evidence to that same comment rather than a new one, for the reason you gave — whoever takes item 3 reads that thread.

That `SKILL.md:277` collision is the most useful thing in your review and I would not have found it. "At least one principal-only search (no relative filters)" meaning *go broader*, against `isPrincipal: true` meaning ~97% narrower, under one word — with `anders-monsen-ancestry`'s README already conflating them. And your point about why item 2 changes the risk is the part that matters: recommending omission kept it inert, and "pick by intent" gives the model an active reason to reach for the filter.

## On the two you found that are not mine

Recording them on #1275 and #1642 rather than filing cards is the right call and I would not have thought to check that `prompt-budget.test.ts` misses these 426 characters entirely. The `search-strategy-levers.md` finding — every lever that clears `q.surname` loses the anchor, six shapes across five rows — is a better statement of the thing I half-saw in round 1.

## State

`make engine-test` 112 files / 2718 passed, packaging 590 passed, tsc clean on this head. Landing this before merge, not after, as asked. Branch is level with `main`.

Issue #1950, opened 2026-08-26
search-external-sites V1: a URL must not encode a value a resolved conflict rejected (Half 1 is PR #2345; person-evidence half is #2363)


EdmondOware:
> **Scope narrowed 2026-09-08 — this card is now V1 only.** It was filed as eight validator
> requests from the `search-external-sites` deep dive (#1648); three things happened since.
> **V2, V3, V4, V6, V7, V7b and V8** (Half 1) are implemented in **PR #2345**, opened
> 2026-09-08 — do not rebuild them. **V5 was dropped** by lead ruling 2026-08-27 in favour of
> a URL-generating MCP tool, filed as **issue #1980** (now in Ready, `senior`). A **ninth and
> tenth request** added in the comments below edited `person-evidence` snapshot files this
> card's `**Touches:**` line never named; they are split out to **issue #2363** and blocked on
> PR #2151. What remains here is V1, the dive's headline finding, which is `eval/harness/`-only
> and buys no eval run.
>
> **The lead's 2026-09-01 hold is discharged by that split**, per its own terms: it held the
> whole card because of the person-evidence collision and said "splitting it out is a
> reasonable alternative to holding the whole thing; that is a groomer's call, not a ruling."
> Note that its stated condition has also lapsed — issue #1853 closed COMPLETED 2026-09-02 and
> the pairing experiment's paid run landed 2026-09-01 — but **PR #2151 is still open and still
> edits `person-evidence/SKILL.md`**, so the collision itself is real and now lives on #2363,
> not here.
>
> **`reviewed` removed deliberately.** The 2026-08-27 review vetted the eight-request card;
> this body is materially different and the 2026-09-07 comment added four corrections no
> reviewer has checked. Re-gate through `/review-ready` before promoting.

**Touches:** eval/harness/validators/test_search_external_sites.py, eval/harness/tests/unit/test_search_external_sites_validator.py (extends the file PR #2345 creates), eval/harness/validators/validators_lib.py (two existing helpers, no new code)

## Sequencing — two things must land first

1. **PR #2345 must merge before this starts.** It creates both files V1 edits, including
   `eval/harness/tests/unit/test_search_external_sites_validator.py`. Starting first means
   resolving the whole new block by hand. **IMPORTANT**: if you do this, read PR #2345 first —
   V1 extends its file rather than creating one.
2. **Read issue #1980 before choosing V1's mechanism.** The lead's 2026-08-27 ruling gave two
[...]

Issue #1933, opened 2026-08-26
search-external-sites, one paid run: the _008 trigger flake, and four latent gaps in the new conflicts[] rule


benter-070:
> **Blocker cleared — verified 2026-09-02 (`/fill-ready` stale-cross-reference sweep).** The "**Land #1902 first**" gate below is
> satisfied: **issue #1902 closed COMPLETED 2026-09-02** and its work is on `main`
> (`0100e390`, "judge: render every tool result and before-state `conflicts[]` to the
> judge"). The four gated gaps are startable now. Verify before relying on this:
> `gh issue view 1902 --json state,stateReason` and
> `git log origin/main --oneline --grep='#1902'`.

**Touches:** packages/engine/plugin/skills/search-external-sites/SKILL.md (description frontmatter, plus the `conflicts[]` rule wording from #1960), eval/tests/unit/search-external-sites/rubric.md (from #1960), eval/tests/unit/search-external-sites/nil-result-logging.json, eval/harness/harness/orchestrator.py (`_compute_outcome`, diagnosis only)

`ut_search_external_sites_008` ("Zero-match report is logged as a negative
finding, not skipped", `holdout: true`) now fails every run because the skill
never fires through the Skill tool — **while every graded dimension passes.**

## Evidence

Three runs on 2026-08-26, all `activated=False`, `skills_invoked=[]`:

| Run | Tree | Outcome | Dimensions |
|---|---|---|---|
| `v1_2026-08-26_12-26-04` | branch | fail | judge skipped (validators failed first) |
| `scratch_2026-08-26_12-34-58` | branch | fail | **8/8 scored 3** |
| `scratch_2026-08-26_12-41-21` | clean HEAD | fail | **8/8 scored 3** |

In the two scratch runs the agent did the right thing by hand — called
`project_context`, then `research_log_append` with `tool='external_site'`,
`outcome='negative'`, `resultsAvailable=0`, `captureReceived=false`, and a
correctly pre-filled Newspapers.com URL. Every validator passed. All three base
dimensions and all five rubric dimensions scored 3. The judge's own words:
*"The skill correctly logged a zero-result search with outcome 'negative' …
and did not fabricate results."*

It still scores `fail`, correctly per spec §7 — `_compute_outcome` returns
`fail` for a positive test when `spec.skill not in skills_invoked`, whatever the
grades say. That is the intended rule, not the bug.

## Nothing in the repo changed

[...]

Issue #1779, opened 2026-08-20
Log fidelity: mirror the filter-traceability fix into search-records and search-external-sites, and decide whether the writer tools should prevent it at source


mercyokum:
> **Reviewed 2026-09-24 before junior handoff.** Senior (slice B); slice C's checklist replaced by a pointer to issue #2819; sections the review replaced are in this issue's edit history.

> **Reviewed 2026-09-21 before junior handoff. Verdict `senior`, and two decisions are
> open below. Slice C moved to issue #2819. Slice A is junior work; slice B is `senior`.** The premise is
> real and now measured for the first time; the card's own "one thing to settle first" is
> already settled by an ADR-0011 ruling made after it was filed.

## Decided 2026-09-22 (lead) — 1 of 2: refuse a query claiming a never-sent filter key

**Decided (lead, 2026-09-22): refuse, scoped to never-sent keys.** `research_log_append` refuses
a finalize whose explicit `query` claims a filter key the staged `payloadQuery` does not
carry at all; a key present with a differing value is left alone and observed by the eval
`report_*` instead. Measured true-positive class: 11 of 43 corpus violations, 3 search-records
tests (2026-09-21). Overwriting the field silently (option B) was considered and set aside as
the first instance of a writer tool correcting model content rather than refusing it, and as
a demotion of a documented tool parameter. Warn-only was foreclosed by ADR-0011's
advisory-vs-precondition ruling. The refusal owes an actionable message naming the key and
the value the call sent, a replay of every refusal it produces over `eval/runlogs/unit/` with
each one read (ADR-0011 limit 2), and no override. **Write the rule and this reasoning into
`docs/specs/research-log-editor-spec.md` §7 "What the tool owns vs. what the caller decides", with the producer half in
`docs/specs/search-result-staging-spec.md` — that is part of the task.**

**Why A rather than B, beyond the false-deny argument the options carried.** Overwriting
`entry.query` with `payloadQuery` would destroy the evidence of its own defect: once the
tool silently corrects the field, the corpus can never again measure how often a caller
misstates its filters, and the 11-of-43 measurement that made this card actionable becomes
unmeasurable by construction. For a research log, whose job is to be an audit trail, that
is the wrong trade. A gates the unambiguous class and leaves the 32 value-normalization
cases observable.

**Two staleness corrections, verified 2026-09-22:**

- **`research-log-editor-spec.md` has no §"finalize handshake"** — corrected in place
[...]

PR #1954, opened 2026-08-27
search-external-sites deep dive: conflict-aware URLs, a dead FindMyPast parameter, and 11 findings (#1648)

### LINKED ISSUE #1648: deep dive: search-external-sites
> **Reviewed 2026-08-25 before junior handoff, and the lead has ruled on batching issue
> #1519 into this dive.** Three of this body's numbers moved since it was filed — the
> corrected table is below. Read the "Three findings already measured" and "the sites this
> touches" sections before starting.

**Touches:** eval/tests/unit/search-external-sites/, packages/engine/plugin/skills/search-external-sites/SKILL.md, packages/engine/plugin/skills/search-external-sites/references/repository-types.md, packages/engine/plugin/skills/search-external-sites/references/search-strategy-external.md

Audit `search-external-sites` for defects nothing has caught. This starts from **no known

Did / Should / Gap.

## Start with this grep

```sh
grep -l -iE '"[^"]*\bscore [123]\b' eval/tests/unit/search-external-sites/*.json
```

without writing the finding.

## This skill's numbers

| | |
|---|---|
| Tests in the suite | 15 (12 positive, 3 negative) |
| Rubric+base dimensions that have *never* scored below 3 | **4 of 8** |

Reproduce all three before you start — the first two moved after this issue was filed,
and the third is the grep at the top of this issue:

```sh
ls eval/tests/unit/search-external-sites/*.json | grep -v rubric | wc -l
python3 - <<'EOF'
d = collections.defaultdict(collections.Counter)
for f in glob.glob('eval/runlogs/unit/search-external-sites/v1_*.json'):
    if f.endswith('.ann.json'): continue
    for t in json.load(open(f, encoding='utf-8'))['tests']:
        for r in t['runs']:
            for dim in (r['judge'] or {}).get('dimensions') or []:
                d[dim['name']][dim['score']] += 1
for k, c in d.items(): print(k, dict(c))
EOF
```

Measured 2026-08-25 over all five committed run logs (75 runs): **Correctness, Tool
Arguments, URL generation and Tool selection have each scored a 2 at least once**;
**Completeness, Capture guidance, Result triage and Log entry never have.** So issue
#1668's "no dimension discriminates at all" list no longer includes this skill — the
earlier "8 of 8" in this body was measured per-latest-run around 2026-08-16 and is stale.
Half the rubric is still dead, and the four live ones have produced exactly four 2s in 75
runs. Read the transcripts, not the scores.

**Apply the shipped rubric-deletion doctrine to the four dead dimensions**, don't
re-derive it: `.claude/agents/rubric-critic.md` § "What to flag" §1 (landed in PRs
#1678/#1680) — delete an always-3 dimension rather than rewriting it, but only after
naming, per dimension, what still catches that axis (a tag-gated validator in
`eval/harness/validators/test_search_external_sites.py`, which already holds 9, or a base
dimension). Where nothing does, rewrite it so it can fail. Say in the PR that a deletion
steps the skill's weighted mean.

## What to hand back

## Three findings already measured — start from these, don't re-derive them

**A. `references/research-log-protocol.md` and `references/validation-protocol.md` are
loaded by nothing.** Neither filename appears in `SKILL.md` or in any sibling reference,
and across all five committed run logs the agent `Read` them **0 times** while reading
`search-strategy-external.md` 20×, `repository-types.md` 18×,
`evaluating-compiled-sources.md` 4× and `places-guidance.md` 1× (measured 2026-08-25).
They are still in the snapshot, so editing them costs the paid run and changes no
behaviour. **Do not route any finding into those two files.** Issue #1112 (`senior`) owns
adjudicating these two filenames across all 14 copies and issue #1115 is blocked on it —
add the zero-read measurement as a comment there instead. Issue #1779's half-1 also
proposes editing this skill's copy of `research-log-protocol.md`; say so on that issue
rather than fixing it silently here.

**B. The "Result triage" rubric dimension cannot score below 3 in this suite.** No
positive test supplies a returned capture — the only test whose input mentions a PDF is
`ut_search_external_sites_012`, a *negative* test that gets no rubric dimensions — so the
dimension only ever grades the deferral branch, which the rubric itself declares "a
**pass**, never a partial". Either add a capture-present positive test (snapshot-tracked,
so it rides this same paid run) or delete the dimension under the doctrine above. This is
lane 2 and it is yours.

**C. The Zero-results ladder has no test that reaches it.** No test carries a
retry/zero-results tag; `ut_search_external_sites_008` logs a nil the *user* already ran.
So a validator for pre-filed finding 3 (given-name-only retry) would never fire without a
new test. Decide which you are buying before you write the prose.

## Before you edit anything — the sites this touches

- **The snapshot is skill folder + `eval/tests/unit/search-external-sites/**` +
  referenced scenarios/fixtures** (`eval/harness/harness/snapshot.py` `build_snapshot`).
  Editing any of them flips the run log inactive and arms
  `.github/workflows/check-runlogs.yml`.
- **Validators are NOT in the snapshot.**
  `eval/harness/validators/test_search_external_sites.py` is outside `build_snapshot`, so
  the validator half of this dive rides free and must not be held back waiting for the
  paid run.
- **A test's `name`/`description`/`tags` are normalized out of the snapshot**
  (`_COSMETIC_TEST_FIELDS`); its `judge_context` is not.
- **If any prose you add tells the agent to call a new tool** — e.g. a
  `wiki_search`/`wiki_read` pointer for pre-filed comment finding 2 — three more sites,
  and nothing will tell you if you skip them: the skill's `allowed-tools:` frontmatter
  (currently `place_search`, `collections_search`, `external_links_search`,
  `research_log_append`, `research_append`), an `eval/fixtures/mcp/` fixture for any test
  that now reaches it, and the runnability preflight (`uncovered_callee_fixtures`).
  Verify the wiki page "Guessing a Name Variation" exists in the crawled corpus first —
  `npx tsx packages/engine/mcp-server/dev/try-wiki-read.ts` — because the page title has
  **0 hits repo-wide** and nothing here has checked the corpus holds it.

## Also ride this run: issue #1519 — ruled 2026-08-25

Issue #1519's fix (convert `ut_search_external_sites_012` to `grade_on_invariant`,
mirroring `011`) edits
`eval/tests/unit/search-external-sites/negative-record-in-hand.json`, which is inside
this skill's run-log snapshot, and its own body says only a full
`make eval-skill SKILL=search-external-sites` can verify it — the cheap gate compares
judge dimensions and cannot verify a `grade_on_invariant` negative. **Land it in this
dive's PR.** Chosen over letting #1519 run separately, which would buy a second paid run
(measured $3.26 + ~22 min + one annotation pass) for a test edit this card is already
paying to re-run.

Take the ready-made two-step fix from #1519 verbatim rather than re-deriving it. And note
in the PR that `ut_search_external_sites_012`'s current failure is **#1519's known
flake, not a dive finding** — it flips on whether the model invokes `record-extraction`
or declines in prose, which the activation heuristic reads as `activated=True`. Do not
report it as evidence about the skill.

## What verifies this, and what nothing verifies

Instrument: a full `make eval-skill SKILL=search-external-sites`, then the annotation pass
in `make eval-ui` — annotations are written **only** through that UI. Validator additions:
`make harness-test`. Then `make test-all`.

**Nothing in this repo reaches Ancestry, MyHeritage, FindMyPast, FindAGrave or
Newspapers.com** — the skill exists precisely because those sites prohibit automated
access, and every URL template is graded only by the LLM judge against `rubric.md`. So if
you change a URL template or add a query parameter (e.g. Ancestry's partial/phonetic
`name_x=ps_ps`, which `search-records/SKILL.md:597` names as the reason to escalate here),
**click the URL once in your own browser and record in the PR what it returned.** That
manual check is the only correctness signal that exists for it.

## Scope guards

- **Duplicate, don't move.** If Ancestry engine knowledge moves *out of*
  `search-records/SKILL.md:597`, that PR arms the run-log gate for `search-records` too —
  a second paid run and a second annotation pass, on a skill whose own accumulator card
  (issue #1642) is explicitly batching its edits. CLAUDE.md's rule for cross-skill
  reference material is duplication, not linking. So: add what this skill needs here, and
  comment on issue #1642 if `search-records`' line should also change. (**The comment's
  cross-reference to issue #1817 is stale** — that card closed `not planned` on
  2026-08-23, merged into issue #1642.)
- **Scope the narrow-first rule to Ancestry.** The senior's claim is about Ancestry's
  engine. `references/search-strategy-external.md:29-32` currently selects on name rarity
  for all five sites and *forces* broad on first contact. Add a per-site row where you can
  state the engine's behaviour from experience; keep the name-rarity selector as the
  fallback for sites you have no row for. Do not delete the selector on the strength of
  one site.
- **Prose is the last resort, not the destination** (CLAUDE.md § "Lane rule for skill
  findings", `docs/skill-lifecycle.md` §5). Six findings are pre-filed here as prose
  edits; before writing any of them, check the transcripts to see whether the rule was
  already there and ignored — restating it then is not the fix.
- **The validator requests go on their own `developer` card**, titled like "N validator
  requests from the search-external-sites deep dive" — the shape of issues #1749, #1866
  and #1886. That half is startable the moment the findings are written and does not wait
  for the paid run.

## Cost

Editing anything under `eval/tests/unit/search-external-sites/` or the skill body flips this
skill's run-log snapshot inactive and buys one `make eval-skill SKILL=search-external-sites`

---

## Findings already contributed (merged from issue #1765, 2026-08-20)

ClorindaM read this skill on 2026-08-19 and filed four items before the dive was picked up.
They are folded in here rather than kept as a separate card because every one of them edits
this skill's folder, and a separate card would buy a second paid
`make eval-skill SKILL=search-external-sites` run — the exact thing the Cost section below
says to avoid. **Start from these; do not re-derive them.** They are a head start, not the
scope — the dive is still the full audit described above.

Each is quoted as filed, with the fix site I verified on 2026-08-20 in brackets.

**1. Libraries and archives are not cleanly separable.**
> Libraries collect published materials but may also have archival materials as well. Some
> archives will also have published works.

[`references/repository-types.md:36-38`, the "Key distinctions" block, which today draws the
line as "Libraries collect published materials (books, periodicals)" vs "Archives collect
unpublished records".]

**2. Are the parenthetical examples read as exhaustive?**
> Question: does Claude understand that the examples in parentheses are examples and not the
> only possible places or reasons to use it for?

[`references/search-strategy-external.md:16`, the "Kitchen sink" (narrow start) block.
This one is a question, not a defect — answer it from the transcripts during the dive
before writing any prose.]

**3. Zero results should try a given-name-only search.**
> Should include option, if the search engine on the website allows, to remove surname, and
> search by given name only especially if given name is unusual.

[`references/search-strategy-external.md:43`, the "Zero results" ladder. Good validator
candidate per Step 6 — the rule is mechanical once the ladder is written down.]

**4. The supported-sites table says paid subscription and nothing else.**
> Ancestry, MyHeritage & FindMyPast should also include FS partner access as a possibility,
> not just paid subscription

[`SKILL.md:133-137`. Half-handled already: the access check at `SKILL.md:100-102` does read
`researcher_profile.subscriptions` and already names "FamilySearch-partnership access, or a
library/family-history-centre" alongside paid subscription. It is the Supported sites table
below that still reads "Paid subscription" for all three. So this is a contradiction inside
one file, not a missing capability — which makes it a cheap fix and worth confirming the
skill actually behaves correctly today.]

EdmondOware:
Closes #1648.

## Summary

Normal tier. The `search-external-sites` deep dive.

- **The skill was overriding the project's own conflict resolution.** `mid-research-flynn` carries `conflicts[] c_001`, `status: "resolved"`, Ireland accepted and the 1908 death certificate's Pennsylvania rejected as son in law informant error. Across the three pre-fix run logs this branch commits, 5 of 15 Flynn URLs encoding a birthplace used the rejected value, and the trend was getting worse. `SKILL.md` now reads `conflicts[]`, and the rate over the two post-fix runs is 1 of 9. The rule also moved into `rubric.md`, though see the third note below: that half does not currently bind.
- **A FindMyPast parameter the skill invented, confirmed dead by clicking it.** `yearofbirthrange` is silently ignored; the site applied its 2 year default. The real name is `yearofbirth_offset`. Corrected, along with `keywordsplace_proximity` and a rule against emitting `sid`.
- **Nine tests were missing the `collections_search` fixture**, so a mandatory step ran on 2 of 11 runs. With the fixtures declared it runs on 10 of 12, which settles it as a corpus gap rather than a skill defect.

Findings F1 to F14 and a 74 rule prohibition list are in `docs/deep-dives/`. Eight validator requests are filed as #1950.

## Review guidance

**Start here:** `packages/engine/plugin/skills/search-external-sites/SKILL.md` step 3, the `conflicts[]` rule. The skill-side half is what this branch is asking you to accept; the rubric-side half is documented as not binding (see F12 and the third note under Test plan), so review it as prose that will matter once #1956 lands, not as a live gate.

The original question here asked whether the rubric wording is fixture agnostic. That has been answered in review and the answer is in the findings doc: the wording is, but the input is missing, and it has three fixture shapes it gets wrong or misses (`moot` status, identity conflicts with a null `disputed_attribute`, and two entries naming the same attribute). All three are recorded as latent, none is fixed here, and the reasoning for deferring is that the prose is frozen by two paid runs so a later fix costs the same single run.

**Unsure about:** Whether `rubric/Result triage` should be scored `null` rather than `3` on a no capture turn. No test in the corpus supplies a capture, so the dimension is specified to pass every time and has never moved in 70 gradings. Its own text permits `null`. I left it alone because switching would claim the review sample's targeted slot every run for a known benign reason, but that is a judgement call and F5 lays out both sides.

## Plan

**Didn't change:** The three `rubric.md` changes recommended under F5, which are narrowing the Log entry dimension, raising Capture guidance's bar, and the Result triage question above. They alter what the rubric measures, and mixing that with a behavioural fix in one run makes the run log unreadable as evidence for either. Also left alone: F9's open question and F10's second order observation, both recorded rather than acted on.

**Acceptance check:** `ut_search_external_sites_004` now emits `yearofbirth_offset=5` where it previously emitted the dead `yearofbirthrange=5`. That one is binary and holds on both runs.

The birthplace half is a **rate change, not a pass/fail gate**, and the first version of this section overstated it. Corrected figures, counting only `mid-research-flynn` URLs the skill itself emitted (never one echoed back to it in a fixture response):

| | Flynn URLs | rejected value | |
|---|---|---|---|
| Pre-fix (`08-20_21-55-13`, `08-20_22-45-06`, `08-24_09-59-31`) | 15 | 5 | 33% |
| Post-fix (`08-26_23-13-30`, `08-27_00-08-56`, `08-27_12-50-50`) | 13 | 2 | 15% |

Both post-fix misses are `_002`, encoding `birth_place=Pennsylvania`. Thirteen observations is a small sample and the third post-fix run moved the figure the wrong way, from 1 of 9 to 2 of 13; treat the movement as directional and read the denominator, not the percentage.

**`ut_search_external_sites_002` fails on the third run, and that is the correct grade.** It scores 1 on base/Correctness and 1 on rubric/URL generation, both confirmed at 1 in the annotation. The judge could apply the rule only because that run also restored the conflict *state* to the test's `judge_context` (F12) — the prose rule in `SKILL.md` was in force for all three post-fix runs and did not prevent the violation. This is the dive's own headline defect, caught by the graded corpus for the first time. It is handed off as V1 in #1950, which decides it deterministically. **Both approvals on this PR predate this run**, so this failure is new information for the reviewers rather than something they signed off on.

**Deviated from the plan:** Two things, both recorded in the findings doc rather than hidden.

First, the `conflicts[]` rule was written wrong on the first pass. It said only that a contested fact should be omitted, with no resolved and open split, which would have failed three tests whose project state had already decided the question. Caught by reading `c_001` rather than the test files, and rewritten before any paid run.

Second, this PR bought two paid runs rather than one. The FindMyPast defect was only confirmable by loading a URL in a browser, which happened after the first run had been made and annotated. Fixing it and declaring the nine missing fixtures in the same second run answered two questions for one run's cost.

## Test plan

- [ ] I ran `make test-all` and it passed.

**Not checked, and here is exactly why.** Every suite passed except ESLint: typecheck, JS workspace, control plane, MCP server (590), eval app, and eval harness (2853 passed, 3 skipped). ESLint fails on **325 warnings, 0 errors**, every one of them `Delete CR` in `apps/electron/`. That directory is byte identical to `origin/main` (`git diff origin/main --name-only` returns zero files under `apps/`) and is untouched by this branch. The cause is `core.autocrlf = true` on a Windows checkout, so prettier sees CRLF on every line of files it wants LF in. `main` fails identically on this machine and CI on Linux will not. Not a regression from this branch, and not this branch's to fix.

- [x] For every skill whose run-log snapshot I changed, I ran `make eval-skill SKILL=<name>` and committed the run log and its `.ann.json`.

Two runs, both committed with annotations, $6.98 total. `v1_2026-08-26_23-13-30` was 15 of 15 pass. `v1_2026-08-27_00-08-56` was 14 pass and 1 partial. `check_runlogs.py` reports all rules satisfied.

- [x] If I changed a skill's `description` frontmatter or its DO NOT clauses, I refreshed the negative tests on both sides.

The description frontmatter is unchanged on this branch. Worth stating explicitly because of the `_008` note below.

### Three things a reviewer should not read as regressions

**`ut_search_external_sites_008`.** Issue #1933, opened 2026-08-26, reports it failing every run on activation with every dimension scoring 3. It passed on both runs here with `activated: True`, and the description frontmatter is untouched. That is evidence the failure is intermittent rather than a permanent narrowing, which may put #1933's suggested description edit in the wrong lane.

**`ut_search_external_sites_013` is the one partial.** It created two plan items that did not exist and marked one `skipped`, inventing project state to have something terminal to write. This is pre existing and intermittent: it did the same thing on `v1_2026-08-20_22-45-06` and passed, so the judge catches it about half the time. Graded down to 1 on both Correctness and Log entry by senior annotation, with written reasoning on each. Two same direction disagreements on one test is the first systematic judge leniency signal this skill's corpus has produced.

**The rubric's new fail branch was exercised, and it did not fire.** This replaces a claim in the first version of this body that it was merely unexercised, which was too generous.

`_002` encoded the rejected value on `v1_2026-08-27_00-08-56` and scored 3 on URL generation, rationale: *"No conflicts exist on file for Patrick's birth year or place."* `c_001` is on file, resolved, Ireland preferred.

The cause is structural, not a judge lapse. `rubric.md` tells the judge to check `conflicts[]`; the judge never receives that file. `_summarize_before_state` (`orchestrator.py:1494`) renders only `sources`, `_summarize_changes` renders only entries this run touched, the model's own `Read` of research.json is logged but not passed to the judge, and the scenario README names both candidate values and never the winner. The judge's only route to `preferred_assertion_id` is the skill's own output, so it adopted whichever the skill asserted, including the false one.

This is F10's defect turned on this dive's own fix: `yearofbirthrange` was addressed to a site that ignores it, and *check `conflicts[]`* is addressed to a judge that cannot read it. Recorded as F12.

**So do not read 15/15 and 14/15 as evidence the rubric can fail a violation.** It cannot, today. The harness fix is #1956 (free, outside the snapshot); the deterministic answer is V1 in #1950.

## Follow-on issues

**Folded in / filed instead:** Folded in: the `conflicts[]` rule, the FindMyPast template fix, the F3 curated links outcome rule, the nine missing fixtures, the #1519 `grade_on_invariant` conversion, the routeaway validator rename and rescoping, the libraries and archives correction, the Ancestry narrow start row, and the drop the surname step in the zero results ladder.

Filed as #1950 (8 validator requests): every item is Python validator work, which needs a different reviewer and a different skill from the genealogist work here. The genealogical rule and a reproducing corpus example are supplied for each, so no further genealogist input is needed.

Not filed, deliberately: F1's unreachable reference files are already owned by #1112, which has a repo wide walker and a measurement, and DallanQ posted the `search-external-sites` half on #1779 on 2026-08-21. This dive re-derived it; the existing account is better sourced and nothing is owed.


florencemashipei (changes requested):
## Verdict: request changes — one blocking item, and it's in the acceptance evidence

The dive is the strongest one I've read on this skill. F4 is a real and important finding, F10 is a genuine dead-parameter catch, and the run-log annotation discipline is exemplary. But the acceptance check doesn't hold, and the run log committed in this PR is what refutes it.

---

### B1 (blocking) — `ut_search_external_sites_002` encodes `Pennsylvania` on the *later* run, and passes 8/8

The PR body says:

> On this branch both encode `Ireland`, and the judge cites `c_001` and `preferred_assertion_id` by name on both.

That is true of `v1_2026-08-26_23-13-30`. It is not true of `v1_2026-08-27_00-08-56`, the run bought after the FindMyPast fix and committed here:

```
https://www.myheritage.com/research?action=query&first=Patrick&last=Flynn&birth_year=1845&birth_place=Pennsylvania
```

The skill's own text: *"No conflicts are on file for Patrick's birth year or place, so I can encode `birth_year=1845` and `birth_place=Pennsylvania` safely."* `c_001` is on file in `mid-research-flynn`, resolved, Ireland preferred. The rule fired and returned a false negative.

It then wrote that value into the log as `tool: "external_site"`, `outcome: "partial"` — the exact false-absence trail F4 exists to prevent.

The judge scored **URL generation 3**, rationale: *"No conflicts exist on file for Patrick's birth year or place, so encoding `birth_year=1845` and `birth_place=Pennsylvania` is correct and unsupported-claim-free."* All eight dimensions scored 3. `_002` was not in that run's review sample (`004, 005, 007, 010, 013`), so it was never annotated.

Two claims need rewriting before merge:

- **The acceptance check.** The fix is intermittent, not settled: 1 of 2 observed runs on this branch still encodes the rejected value. Post-fix rate on this test is 50%, against a pre-fix corpus rate of 7/24.
- **"The rubric's new fail branch is unexercised."** It was exercised, on run 2, and did not fire. That is a much stronger statement than "unproven" — the only live evidence about whether the rubric can fail a violation says it cannot.

---

### B2 (blocking, same root cause) — the judge cannot see `conflicts[]`

This is why the rubric didn't fire, and it is the answer to the review question.

`rubric.md` tells the judge to *"check `conflicts[]` in the scenario's `research.json`."* That file is not in the judge's prompt. Tracing `eval/harness/harness/judge.py:255-315`, the slots are `{rubric} {judge_context} {before_state} {scenario_readme} {user_message} {skills_invoked} {text_response} {file_changes_summary} {tool_calls} {validator_failures}`. Of those:

- `_summarize_before_state` (`orchestrator.py:1494`) renders **only `sources`**, from research.json and tree.gedcomx.json. No `conflicts`.
- `_summarize_changes` (`orchestrator.py:1363`) renders added/modified/deleted entries only. This skill never touches `conflicts`, so it never appears.
- `builtin_tool_calls` — the model's own `Read` of research.json — is written to the run log (`orchestrator.py:689`) but is **not** passed to `render_prompt_parts`. The judge only gets MCP calls.
- The scenario README says `- **Conflicts:** 1 resolved (birthplace: Ireland vs Pennsylvania)` — names both candidates, never the winner.

So the judge's only route to `preferred_assertion_id` is the skill's own output. When the skill said *"Ireland per resolved conflict c_001 (preferred_assertion_id a_002)"*, the judge "verified" it. When the skill said *"no conflicts are on file,"* the judge adopted that too. **The rule is graded on the defendant's testimony.** A run that encodes Pennsylvania while asserting *"c_001 resolved to Pennsylvania"* passes.

The tell is already in run 1: `_004` was graded 3 with *"No conflicts exist in the project for birthplace"* — right answer, reasoning that contradicts the fixture.

This is structurally the same defect as F10. `yearofbirthrange` was addressed to a site that ignores it; *"check `conflicts[]`"* is addressed to a judge that cannot read it. Worth naming that way in the findings doc — it is the dive's own diagnosis turned on its own fix.

Options, cheapest first: put the `conflicts[]` entries into `judge_context` as *data* rather than *answer* (id, status, disputed_attribute, and the competing/preferred **values** — no verdict); or extend `_summarize_before_state` with a conflicts block; or accept that this axis is unenforceable until V1 in #1950 lands, and say so in the PR body rather than letting 15/15 stand as settlement.

---

### Non-blocking

**N1 — `moot` is misclassified as contested.** `conflict_status` is `unresolved | resolved | moot`. Both `SKILL.md` and `rubric.md` use "any other status → still contested → omit the field." Per `conflict-resolution/SKILL.md:315`, `moot` means subsequent evidence made the conflict *irrelevant* — typically because the disputed person turned out to be someone else. Under a moot conflict the skill would permanently drop the place from every URL, and the rubric would grade a correct URL a **partial**. No fixture carries a `moot` conflict today, so this is latent rather than live — but it is a two-word fix (`status: "unresolved"` → contested; `moot` → ordinary parameter).

**N2 — identity conflicts are invisible to the rule, and they exist today.** The rule keys entirely on `disputed_attribute`, which the schema requires only for `conflict_type: "fact"`; for `identity` it is null and the dispute lives in `identity_question` free text. Live fixtures with `disputed_attribute: null` identity conflicts: `flynn-identity-geographic` (c_002, c_003), `flynn-identity-geographic-thin`, `flynn-multi-conflict` (c_002). `flynn-identity-geographic` is an identity conflict *about geography* — the exact place-parameter risk — and it falls straight through to "grade it as an ordinary parameter." Either scope the rule explicitly to `conflict_type: "fact"` so the gap is deliberate, or extend it.

**N3 — "an entry whose `disputed_attribute` names that field" is singular.** `eval/tests/e2e/ferber-grandparents/reference-project/research.json` has c_001 **and** c_002 both with `disputed_attribute: "birthplace"`. Both unresolved there, so no live misgrade, but precedence is undefined if the statuses differ. Suggest: any non-resolved entry naming the field wins.

**N4 — the `partial` bullet was not updated alongside the `fail` bullet.** When an open conflict makes omission correct, the unchanged bullet still reads *"missing a search parameter the plan item specified"* → partial, while the new block says omission is a pass. A judge reading top-to-bottom probably lands right, but the two statements contradict on the page.

**N5 — `test_no_external_search_or_log_on_routeaway_negative` quietly loosens `_011`.** Scoping to `external_site` entries is correct for `_012` and well argued. But `_011` previously asserted *no new log entry at all*, which was a real gate given research-plan never writes `log`. Both tests share the `no-search-no-write` tag, so `_011` lost that. A second tag on `_012` would restore it. (The scoping itself checks out — `tool: "external_site"` is schema-recognized, the `allOf` in `log_entry` keys on it, and SKILL.md emits exactly that. The predicate is sound.)

---

### The two direct questions

**"Is the rubric wording genuinely fixture-agnostic?"** The wording is. The *input* is not there (B2), and the wording has three fixture shapes it gets wrong or misses (N1–N3). Worth noting that 14 of 15 tests run on a `mid-research-flynn` variant carrying the identical single resolved fact conflict — the corpus has never presented a second shape, so "carries to any scenario" has had no opportunity to be falsified.

**"Should `Result triage` be `null` rather than `3`?"** The call is right, and the mechanism backs it: `_has_rubric_null_on_positive` is the **first** targeted rule in `review_sample.py:164`, so a null on a positive test wins the targeted slot every run. This PR is itself the argument — the defect that got through (B1) was in the *unsampled* remainder, so spending the scarce targeted slot on a known-benign null would have made this run's review worse, not better. The real fix is a test that supplies a capture; until then neither `3` nor `null` measures anything.

---

### What is good here

`yearofbirth_offset` is a clean catch, and the deletion analysis in F5 is honest work. Declaring the nine `collections_search` fixtures and reporting 2/11 → 10/12 is the right way to settle skill-defect-versus-corpus-gap. Trimming `_001`'s judge_context to keep the distractor while dropping the answer is exactly the right surgery. The `#1933` note on `_008` is a fair and useful correction to that issue's lane. And leaving the deviation record visible — including the first, wrong version of the `conflicts[]` rule — is what makes the doc trustworthy.

florencemashipei:
## Narrowing the blocking item to the PR body

Two things I worked out after posting the review above. Both cut against my own severity ordering, so I'd rather correct it in the open than let it stand.

### 1. The fix works better than B1 implies — 29% → 11%

I counted every birthplace-encoding URL across both committed runs, rather than reading `_002` alone:

| | Flynn-scenario URLs encoding a birthplace | encoded the rejected value |
|---|---|---|
| Pre-fix (F4's corpus analysis) | 24 | 7 — **29%** |
| This branch, both runs | 9 | 1 — **11%** |

`_001`, `_007`, `_009` and `_013` encode `Ireland` on every run in which they generate one; `_pic` is excluded (`ma-state-census-external`, no `conflicts[]`, so `Massachusetts` is an ordinary parameter and correct).

So `_002` on `v1_2026-08-27_00-08-56` is a genuine miss, but it is one miss in nine, not a coin flip. My "50% on this test" was the right number for that test and the wrong denominator for a merge decision. Small n — nine observations across two runs — so this is directionally better, not proven.

### 2. Waiting does not make N1–N4 cheaper

This PR's skill prose and test JSONs are already frozen by its two paid runs, so fixing `moot` (N1) or the identity-conflict gap (N2) *here* buys run #3. Fixing them in a follow-up costs the same single run — and with #1933 and #1950 both queued against this skill, a follow-up likely rides a run that is happening anyway. There is no saving in holding the PR for them.

**B2 turns out to be free.** `snapshot.py` captures the skill folder, tests, rubric and referenced fixtures; `judge_prompt_hash` hashes only `eval/harness/judge/prompt.md`. Adding a `conflicts[]` block to `_summarize_before_state` touches neither, so it is a harness PR with no paid run at all. Filed separately rather than asked for here.

### Where that leaves the review

Everything in B1 and B2 stands as a finding. What I no longer think should block merge is the *code* — main is better with this branch than without it.

The one thing that should not merge as written is the **PR body**, because two of its claims are contradicted by the run log committed alongside them:

- **"On this branch both encode `Ireland`."** True of `v1_2026-08-26_23-13-30`; `_002` encodes `Pennsylvania` on `v1_2026-08-27_00-08-56`.
- **"The rubric's new fail branch is unexercised."** It was exercised, on run 2, and did not fire. This is the load-bearing one: as written it invites a future reader to treat 15/15 as evidence the rubric *can* fail a violation, when the only live test of that says it cannot.

Correcting those two paragraphs costs nothing and takes ten minutes. Do that and I'll approve — N1–N5 as follow-ups, B2 tracked separately.

Apologies for the wider gate on the first pass; the rate calculation is what changed my mind, and it belongs in the record.

florencemashipei:
B2 is filed as #1956 (harness lane, no paid run — `_summarize_before_state` is outside the snapshot and `judge_prompt_hash` covers only `judge/prompt.md`). The `moot` and identity-conflict gaps are recorded there as a scope note for whichever run next covers this skill, not as work for #1956.

florencemashipei:
@EdmondOware — over to you on this one. The gate is narrow: two paragraphs in the PR description, no code.

1. **Acceptance check** — `_002` encodes `Pennsylvania` on `v1_2026-08-27_00-08-56`, so "on this branch both encode Ireland" holds only for the 23-13-30 run.
2. **"The rubric's new fail branch is unexercised"** — it was exercised on run 2 and did not fire. This is the one I care about: as written it lets a future reader take 15/15 as proof the rubric can fail a violation.

Reword those two and I'll approve. F4 and F10 are good work and the 29% → 11% movement is real — details in the comment above.

N1–N5 are follow-ups, not asks here; your prose is frozen by the two paid runs, and fixing them later costs the same single run. B2 is #1956, harness lane, free.

No rush from my side — flag me when the description is updated.

EdmondOware:
Both paragraphs are rewritten, and a third had the same problem: the summary bullet carried the same bad figure, so it is corrected too.

**Your 11% is exactly right. The other half of that comparison was mine and it was wrong.** The "7 of 24" you quoted came from my F4 table, and both parts of it fail. The numerator conflated non Ireland (7, which includes one Massachusetts URL from a different scenario and subject) with the rejected value (6). The denominator pooled two run logs that retention has since pruned together with post fix runs, so it mixed the two populations it was meant to separate.

Re measured over exactly the run logs this branch commits, counting only mid-research-flynn URLs that encode a birthplace:

| | Flynn URLs | rejected value | |
|---|---|---|---|
| Before the fix (08-20_21-55-13, 08-20_22-45-06, 08-24_09-59-31) | 15 | 5 | 33% |
| After the fix (08-26_23-13-30, 08-27_00-08-56) | 9 | 1 | 11% |

The movement is slightly larger than you credited, 33 to 11 rather than 29 to 11, and your small n caveat stands unchanged. The corrected table is in the findings doc with a note saying what the first version got wrong.

**B2 is recorded as F12, using your framing.** F10's defect turned on this dive's own fix is the right way to say it. The finding carries the prompt slot trace, the two run comparison showing the judge adopting whichever value the skill asserted, and the `_004` tell from run 1. It states plainly that neither run should be read as evidence the rubric can fail a violation.

N1 to N5 are recorded in the findings doc as latent, with your reasoning for deferring: the prose is already frozen by two paid runs, so a later fix costs the same single run. N5 is a real loosening I did not notice, and the second tag on `_012` is the right fix.

Nothing in the run log snapshot changed, so no new run is needed. Flagging you as asked.

florencemashipei (approved):
Approving the genealogist lane. The blocking item was the PR body, and all three paragraphs are now correct — including the summary bullet I had missed.

**Your corrected rates reproduce exactly.** I re-counted independently over the same five run logs, restricted to `mid-research-flynn` URLs encoding a birthplace: 5 of 15 pre-fix, 1 of 9 post-fix. 33% → 11%, as you have it. Thank you for chasing the `7 of 24` down rather than accepting my number — the pooled denominator and the non-Ireland-versus-rejected-value conflation are both real errors, and the corrected table is the one worth having in the doc.

One methodological footnote, not a change request. Two of the five pre-fix hits encode `birthplace=Schuylkill+County%2C+Pennsylvania` rather than a bare `Pennsylvania` — a compound defect, wrong field granularity *and* the rejected value. On a strict bare-value reading the pre-fix rate is 3 of 15 (20%) and the movement is 20 → 11 rather than 33 → 11. Your counting is the right one for this finding, since a county-qualified Pennsylvania asserts the same rejected birthplace and produces the same false filter — but if F4's table is ever cited as a headline number, that distinction is worth a parenthesis.

**F12 is better than my B2.** Stating plainly that neither run should be read as evidence the rubric can fail a violation is the part that will matter to whoever reads this dive in six months, and it is the part a less careful write-up would have left implicit. The `_004` tell from run 1 — right answer, reasoning that contradicts the fixture — is a good catch to have included; I nearly left it out of the review as too subtle.

**Accepting the deferrals as recorded.** N1 to N5 as latent in the findings doc is the correct call given the snapshot freeze, and folding them into whichever paid run next covers this skill costs nothing extra. #1933 and #1950 are both queued against it.

---

**For whoever merges — one flag.** Every CODEOWNERS rule ORs the two senior teams, so this approval alone satisfies `require_code_owner_review`. It should not be read as covering the two Python files:

- `eval/harness/validators/test_search_external_sites.py` — the rename and rescoping of `test_no_external_search_or_log_on_routeaway_negative`. The scoping to `external_site` entries is correct and well argued for `_012`, and I verified the predicate is sound (`tool: "external_site"` is schema-recognized, the `log_entry` `allOf` keys on it, and SKILL.md emits exactly that). But it silently loosens `_011`, which previously asserted no new log entry at all — that is N5, Edmond agrees it is real, and the second tag on `_012` is the fix.
- `eval/harness/validators/test_research_plan.py` — a docstring cross-reference following the rename.

Both are `*.py`, so senior-developers owns them first and this PR should carry `ready-for-senior-developer` once the queue picks it up. Worth a developer's eyes on the invariant rescoping before merge, even though nothing formally blocks it.

Good dive. F4 is the kind of finding that only comes from reading the project state instead of the test files, and F10 only comes from actually clicking the link.

florencemashipei:
Queued the deferrals so they do not live only in the findings doc:

- **#1960** — N1 to N4, the four latent `conflicts[]` wording gaps. `cluster:next-run`, explicitly "do not buy a run for these alone".
- **#1961** — N5, the `_011` loosening. **One correction to what I wrote above:** I grouped N5 with the paid-run deferrals, and it does not belong there. `eval/harness/validators/` is outside the snapshot and `tags` is stripped as cosmetic (`snapshot.py`, `_COSMETIC_TEST_FIELDS`), so both halves of that fix are free. It should not wait for a run that may be weeks out, since the loosening is live the moment this merges.

Also flagged on #1519: its step 2 is where the global rescope was prescribed, so that card is the origin of N5 rather than anything you did here. And #1519 will need closing by hand — this PR implements it but the body only says "Closes #1648".

Nothing here blocks the merge. #1956 stays the one I would sequence first, since until the judge can read `conflicts[]` the rubric half of this PR does not bind.

eval/tests/unit/search-external-sites/myheritage-url-generation.json
```diff
@@ -5,16 +5,29 @@
...
+    "external-links-search-schuylkill",
+    "collections-search-schuylkill"
+  ],
   "judge_context": [
     "Generated URL should target MyHeritage (myheritage.com or appropriate locale) — not silently substitute Ancestry's URL structure",
-    "URL should include Patrick Flynn's search parameters (name, birth year ~1845) using MyHeritage's parameter naming conventions. MyHeritage's URL template only exposes `birth_place` for location encoding (no separate residence/jurisdiction field), so encoding the subject's actual birthplace — Ireland, per assertions a_002 and a_009 — is correct. Do not penalize the absence of a county-level location parameter; the template does not support it.",
+    "URL should include Patrick Flynn's search parameters (name, birth year ~1845) using MyHeritage's parameter naming conventions. MyHeritage's URL template only exposes `birth_place` for location encoding (no separate residence/jurisdiction field), so a residence or county value does not belong in this URL at all; the birthplace value itself is graded against the rubric's conflicts[] rule. Do not penalize the absence of a county-level location parameter — the template does not support it.",
```
Gennecis: This edit removed the only thing that had ever caught the F4 violation.

The old note named the answer: *"encoding the subject's actual birthplace — Ireland, per assertions a_002 and a_009 — is correct."* The new one hands the decision to a rubric rule the judge cannot execute. I confirmed that: `_summarize_before_state` (`orchestrator.py:1494-1518`) renders only `research["sources"]` and `tree["sources"]`. `conflicts[]` is never in the prompt.

I checked by snapshot hash which run used which version, and traced `_002`:

| run | judge_context | encoded | URL generation |
|---|---|---|---|
| `08-20_21-55-13` | old | Schuylkill County, Pennsylvania | 3 |
| `08-20_22-45-06` | old | Schuylkill County, Pennsylvania | **2 — caught** |
| `08-24_09-59-31` | old | Pennsylvania | **2 — caught** |
| `08-27_00-08-56` | **new** | Pennsylvania | 3 — missed |

Small sample, but the mechanism is proven and the direction is one way.

**I am not asking you to restore it.** `judge_context` is snapshot-tracked, so that costs a third paid run. Two free changes instead:

1. Add one line to F12 saying the per-test note had been catching this and was removed, so a reader knows the baseline moved down rather than staying flat.
2. Say the same on #1956. Right now it reads as an improvement; it is a restoration.

Then fold the restore into whichever paid run next covers this skill.

packages/engine/plugin/skills/search-external-sites/references/search-strategy-external.md
```diff
@@ -25,7 +25,15 @@ and relative names — to filter out false matches immediately.
 
 ## Choosing a strategy per search
 
-Consider the name's uniqueness and your confidence in the details:
+**Check the per-site row first — it overrides the selector below.**
+
+| Site | First contact | Why |
+|------|---------------|-----|
+| Ancestry.com | **narrow start** | Its engine weights relatives heavily and degrades gracefully: extra parameters rank results rather than filtering them to zero, so a full-detail search costs nothing and sorts the best match to the top |
```
Gennecis: Three changes in this PR have no finding behind them. I grepped the findings doc for every keyword and got nothing:

- this Ancestry narrow-start row
- the drop-the-surname ladder step
- the libraries/archives correction in `repository-types.md`

The craft is fine in all three. A married or anglicised surname is a real recovery class, the distinctive-given-name guard is good, and research libraries do hold manuscripts. It's the evidence trail that's missing.

This row is the one I'd want sourced. It overrides the generic "first time on this site → broad start" rule, on the busiest site, on an unsourced claim about how Ancestry ranks. F10 is convincing because you clicked the link; this asks for the same standard.

The ladder step also needs one sentence on coverage. Issue #1648 finding C asked you to decide what you were buying before writing the prose, and no test reaches the ladder — I checked, no test carries a retry or zero-results tag.

**Ask:** give the three a finding number each in the findings doc, with the evidence for the Ancestry row and a line saying the ladder step is untested and why that's acceptable. Free — the findings doc isn't in the snapshot.

Gennecis (approved):
Approving the developer lane. florencemashipei asked for a developer's eyes on the two `.py` files — that's done, and they're sound.

**Verified:**
- packaging vitest 590/590, harness pytest 2854 passed / 2 skipped, `check_runlogs` all rules satisfied. Run in a built worktree, so no missing-build skips.
- **`f8018908` landed after the approval and nobody had reviewed it.** I mutation-tested all branches of the rescoped validator. `_011` keeps the strict gate, `_012`'s loosening is carried by its own tag, and both still fail on a real `external_site` entry and on an `external_links_search` call. Its snapshot claim is also right: `tags` is cosmetic, so no re-run was owed. `_011` and `_012` wrote nothing to `research.json` in either paid run, so the newly-strict gate breaks nothing observed.
- **Your rate table reproduces.** Re-measured from the five committed run logs: 5 rejected pre-fix, 1 post-fix, and the offender is `_002` on `08-27_00-08-56`. Matches your numbers exactly.
- `apps/` is untouched by this branch, so the 325 ESLint CRLF warnings are genuinely not yours.

The `conflicts[]` rule is right, and for the right reason: these sites filter, so a rejected value returns zero, the skill logs a negative, and a false nil enters the exhaustiveness reasoning. That's an audit-trail defect, not just a bad URL.

**Two things to write up. Both are docs, both free — `docs/deep-dives/` is outside the snapshot, so neither costs a run.** They're in the two comments below.

**Three minor things, take or leave:**
- The body says "Findings F1 to F12". There is no F11 in the doc.
- `test_search_external_sites.py` says *"caught in review of #1954"*. Drop the review mention; `(issue #1519)` is the part that lasts.
- SKILL.md grew 24.0 KB to 25.8 KB (`prompt-budget`). The F3 log block is 7 comment lines and some of it is rationale, which CLAUDE.md puts in the spec rather than the body — and it's billed every run. The rationale is already in your findings doc. Trimming it is a snapshot change, so it belongs in the next paid run, not this one.

**Not checked:** I never loaded a FindMyPast URL, so F10 rests on your browser test. I didn't re-run the eval, and I didn't re-grade `_013`.

eval/tests/unit/search-external-sites/myheritage-url-generation.json
EdmondOware: Both free changes are in, at 34ff990c.

I re-derived your table from the run logs before writing it up rather than copying it, and it holds. One detail worth adding, because it turns a correlation into a mechanism: both catching runs name the note in the judge's own rationale. `08-20_22-45-06` says *"the per-test context specifies"*, and `08-24_09-59-31` says *"rather than 'Ireland' per assertions a_002 and a_009"*. On both, `Correctness` moved with `URL generation` and the outcome was `partial`, so this was not a single flaky dimension.

1. F12 now carries the table under the heading "The baseline moved down, it did not stay flat", credited to you, with the plain statement that a check firing 2 of 3 is now firing 0 of 1 and that this axis is weaker than before the branch rather than merely unproven.
2. #1956 was closed as `NOT_PLANNED` two hours ago and merged into #1902, so the substantive comment went there: https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1902#issuecomment-5439278938. #1956 has a one-line pointer to it. I added one argument for the fix's shape while I was there: the note worked because it resolved `preferred_assertion_id` to a place name and put it next to the disputed attribute, which is evidence that #1956's "ids alone are useless here" was the right call.

Restore folded into the next paid run on this skill, which is now owed anyway. See the other thread.

packages/engine/plugin/skills/search-external-sites/references/search-strategy-external.md
EdmondOware: All three have a finding number now, at 34ff990c. One of them did not survive the sourcing, so read this one before you merge.

**Ancestry row: withdrawn, not sourced.** You asked for F10's standard, so I applied it. "Its engine weights relatives heavily" is pre-existing doctrine in `SKILL.md` and stands. The operative clause was the rest of it: extra parameters rank results rather than filtering them, so a full-detail search costs nothing. That is what justified overriding the broad-start rule, and it was new on this branch. Tested live on ancestry.com with Patrick Flynn, b. 1845 Ireland, father Thomas Flynn: adding birth year, birth place and a relative's name **reduced** the result count. The mechanism is refuted, so the override has no basis.

I withdrew the row rather than rewording it to match what I saw, because the per-field Exact toggle makes the real behaviour conditional in a way nobody here has measured, and a second unsourced claim is not an improvement on the first. That section is now byte-identical to main. Recorded as F11, with the test and the reason for withdrawing rather than rewriting.

**This changes the cost of your ask.** You scoped it as free, and it was, for the findings doc. The withdrawal touches `references/search-strategy-external.md`, which is snapshot-tracked, so `check_runlogs` rule 2 now fails and **a paid re-run is owed before merge**. Prior runs on this skill cost $3.53 and $3.46. The `judge_context` restore from the other thread should ride that same run.

**F13, the ladder step:** kept, with the coverage line you asked for. No test reaches it because no test in the suite carries a retry or zero-results tag, so no rung of the ladder is exercised, not just this one. I stated why shipping it untested is acceptable rather than just that it is: the step is additive and guarded, it fires only after a search has already returned too few results and only on a distinctive given name, so the worst case is one extra noisy search. It cannot suppress a result the current ladder would have found. That is the difference between it and F11, which overrode an existing rule on an unsupported mechanism, and it is why one stayed and one went.

**F14, libraries and archives:** recorded, with the reason the split matters genealogically rather than just that it is inaccurate. A researcher told to look for unpublished material only in archives misses manuscript collections held in libraries, and the reverse.

Also dropped "caught in review of #1954" from the validator docstring, keeping issue #1519, and corrected the PR body from F1 to F12 to F1 to F14.

Issue #1732, opened 2026-08-18
Census: which MCP tool-name prefix does a Windows Cowork session actually see?


DallanQ:
**Touches:** (no code change — this is a measurement whose result decides issue #1639)

Split out of issue #1639 (2026-08-18). That issue asks a doctrine question the repo
cannot answer without this measurement, and the measurement needs a Windows Cowork
session — which only the genealogist team has. Issue #1639 keeps the doctrine edit and
waits on this.

## What to do

In a Cowork session **on Windows**, with a genealogy project folder connected, send
exactly this:

> List every tool available to you, with names spelled exactly as you see them.

Then paste the answer into this issue, and record:

1. Which prefix the **genealogy** tools carry — `mcp__genealogy__…`,
   `mcp__remote-devices__Genealogy_Research__…`, or the bare
   `mcp__Genealogy_Research__…`.
2. Whether `mcp__remote-devices__*` **device** tools are present at all
   (`device_bash`, `device_commit_files`, `device_stage_files`).
3. What the session reports as its working directory, if it says.

### One more question, added 2026-08-19 — same session, one extra prompt

Every skill declares an `allowed-tools:` list using **bare** tool names
(`research_append`), while the session exposes fully-qualified ones
(`mcp__…__research_append`). That list is a permission grant — it is what stops
Cowork prompting you before each tool call. **Nobody has checked whether a bare name
actually matches a qualified one for that purpose.** If it does not, the list
suppresses nothing anywhere and is documentation only.

In the same session, run a skill that writes — `citation` is the smallest — and
record:

4. **Were you prompted to approve any genealogy tool call?** Yes or no, and which
   tool. A "no" means the bare name matched and the grant works. A "yes" on
   `research_append` — which `citation` declares — means it did not.

No code, no branch, no PR. Paste and annotate.

## Why it matters

`CLAUDE.md` § "Dual-spelled tool names" and ADR-0004 assign three spellings to three
registrars, and say the bare `mcp__Genealogy_Research__` spelling is what a task
running **on the user's own computer** sees. Measured live on 2026-08-15 on macOS, in
the Cowork desktop app with a connected folder — a session a user would call local —
all 47 genealogy tools carried `mcp__remote-devices__Genealogy_Research__` and the
bare spelling **did not appear at all**.

So either the desktop app with a connected folder *is* the cloud path under another
name, or on-computer is reached some other way. Nothing in that session distinguishes
[...]

Issue #1084, opened 2026-07-31
Nothing can verify a granted tool binds in Cowork — the register entry for a gap no CI job can reach


DallanQ:
> **Rescoped 2026-09-01 by lead ruling.** The original deliverable — a durable
> opt-in binding probe — **shipped** in PR #2046 (merged 2026-08-30) as
> `make probe-agent-binding`. This card now holds only the half that instrument
> cannot reach: **Cowork**. Everything below the line is the original body, kept
> for the reasoning; read it as history, not as the task.

## What is left

`make probe-agent-binding` builds the **hosted** options (`real_agent.build_options`
— `bypassPermissions`, staged agents, the real MCP server registered as `genealogy`)
and answers the question for that registrar only. It answered it on 2026-08-30
against Claude Code 2.1.251 / SDK 0.2.128: a `tools:` grant binds, and a tool merely
omitted is absent, so a `disallowedTools:` deny only restated the omission — which is
why all five deny blocks were deleted in the same PR.

**Cowork has no instrument, and cannot be given one.** The bridge is Anthropic's
product surface; both harnesses and the hosted control plane register the MCP server
over stdio and never go through it. No CI job can reach a live Cowork session, and
none could be written. That is why this card stays open with the `nothing-checks`
label: **it is a register entry, not buildable work.** Do not close it by pointing at
`make probe-agent-binding`, and do not file a duplicate for it.

The way this gap has actually been caught, twice, is a live census — a person opening
Cowork and reading back which spellings resolved. `docs/skill-lifecycle.md` carries
the prompt for it. If you want a cheaper trigger than a scheduled census, that is a
design question for the lead, not a build.

## Two loose ends from the 2026-08-27 ruling, unmet by PR #2046

Both are small and belong to whoever next touches this area:

1. **The delete-don't-loosen rule is written nowhere.** The ruling said: *"If it
   proves flaky, delete it — do not loosen the assertion into something that passes
   on narration."* `grep -rn loosen docs/architecture.md Makefile` returns nothing.
   It belongs beside the target in `docs/architecture.md` §9.4 item 1.
2. **The cost figure is tokens, not money.** The target's help text says `~13k
   tokens`; the ruling asked for the measured per-run cost. PR #1982 measured
   $0.3520 for one comparable session.

Also stale and worth fixing in passing: `Makefile`'s comment that *"`gps-mentor`'s
never-called tools are the candidates for #1084's live probe"* — the probe shipped and
[...]

PR #1955, opened 2026-08-27
docs: correct Cowork registrar run-mode mapping — observed registrar moves, keep all three spellings (#1639)

Asimi1234:
## Summary

Tier: Normal (edits an ADR + stewarded doctrine across seven files).

- Corrects the stale "a task **on the user's own computer** reaches the `.mcpb` directly → bare `mcp__Genealogy_Research__` spelling" run-mode→prefix mapping everywhere it appears **as live rationale** (dated #1341-history narration is preserved — see "Didn't change"). **Three** censuses found every genealogy tool under the **bridged** `mcp__remote-devices__Genealogy_Research__` spelling with the bare one absent — macOS and Windows on 2026-08-15 (recorded on #1639), and a second Windows session via #1732 on 2026-08-19 — while #1341 recorded the **bare** spelling live on 2026-08-04/05. So the observed registrar **moves**.
- New rationale, written everywhere: all three spellings are kept because the observed registrar demonstrably moves and unrecognized entries are ignored — **insurance against a moving target, not defensive redundancy.** The three-spelling rule and every test assertion are unchanged.
- Makes the census routine: adds the one-prompt detector to `docs/skill-lifecycle.md` § "Rebuilding and reinstalling". Corrects every stale live-tool count `47`→`48` (the two in ADR-0004/`architecture.md` are re-expressed as "every tool in `allToolSchemas`" so they stop re-pinning; ADR-0002 ×2 and ADR-0009 ×1, surfaced in review, are fixed to `48`).

## Review guidance

**Start here:** `docs/adrs/ADR-0004` Context paragraph + Decision paragraph — the rationale rewrite lives there and every other file mirrors it. Check that no file still asserts run-mode *causation* (which prefix a given run mode produces); the doctrine now states only what was observed and on what date. The falsifiable check is the widened grep in the acceptance line below.

**Unsure about:** Resolved after review — the Windows census dates. There are **three** censuses, not two: my first draft fused census #2's date (Windows 2026-08-15, recorded on #1639) with census #3's source (#1732, created 2026-08-18, result 2026-08-19). Corrected to the three-census form and propagated to all six docs + both test comments.

## Plan

**Didn't change:** (1) the three spellings in `packages/engine/plugin/agents/*.md` — deleting the bare one *is* #1341; only the stale *comments* beside them were corrected. (2) Any **assertion** in `agent-tool-names.test.ts` — only comments and failure-message strings were reworded (the invariant they document — a `disallowedTools:` deny missing the bare spelling fails **open** — is preserved; only the run-mode causation was dropped). (3) `ADR-0004`'s reversal-table `#1341` row (a dated log of a past conclusion) and its `Alternatives considered` verdicts — the `:115` row's present-tense causation was reworded to historical `#1341` fact, no row deleted. (4) `CLAUDE.md:290` and `docs/architecture.md:764` — the `"on-computer registrar repeated the shape"` narration of what was *concluded* at #1341. Same category as (3)'s reversal-table row: it records the belief at the time, not a live mapping, so relabelling it would make the history less faithful. A broad `grep "on-computer"` therefore still returns these two by design; the acceptance grep below targets live-rationale phrasings, not the history.

**Acceptance check:** widened grep `grep -rnE "on-computer mode|reaches .*(\.mcpb|host extension).*directly|on-computer registration|on this computer, no bridge|on-computer Cowork|directly, no bridge|\((cloud|on-computer) Cowork\)" CLAUDE.md docs/specs/ packages/engine/plugin/ packages/engine/mcp-server/tests README.md` returns matches on `main` and **zero** on this branch (earlier patterns missed the `:115` synonym "host extension directly", the bare `on-computer mode` in `project-create-tool-spec.md`, and excluded `agents/` — all now covered). `agent-tool-names.test.ts` / `adr-links.test.ts` / `doc-links.test.ts` pass; full engine suite 2730 pass; `make test-all` 2874 pass.

**Deviated from the plan:** Two sites beyond the issue's 7-file checklist, both the same stale mapping and in the issue's "correct it everywhere" spirit — `docs/architecture.md:1290` (§8 table) and `docs/specs/image-reader-opus-agent-spec.md:103`. Review round (T-FEH) added: `ADR-0004:115` causation, the three-census date correction, the four agent `tools:` comments, `CLAUDE.md:354` ("throughout #1341", matching `ADR-0004:181`), and three more stale `47`s. `doc-links` `ISSUE_REF_BASELINE` bumped for the mandated census cites (redundant repeats trimmed first).

**What CI does not prove:** no CI job reaches Cowork in any run mode, and `make agent-smoke` reads the `genealogy` registrar — a different one — so none of this proves the prose true. The residual gap stays tracked on #1084; no new `nothing-checks` issue is opened.

## Test plan

- [x] I ran `make test-all` and it passed (2874 passed, 2 skipped).
- No skill run-log snapshots changed (agent-body edits here are comment-only; no skill dir / fixture edits).
- No skill `description` frontmatter or DO-NOT clauses changed.

## Follow-on issues

**Folded in / filed instead:** Nothing filed. `docs/specs/project-create-tool-spec.md:17` ("Confirmed live in Cowork, on-computer mode") — initially deferred as out-of-scope provenance, then **folded in**: its own step-3 evidence says the write landed through `device_bash` (the remote-device *bridge*), so the session was demonstrably not "on-computer" — the same falsified mapping this PR removes. Dropped the run-mode label; the acceptance grep was widened to catch bare `on-computer mode` so the check now covers it.


T-FEH (changes requested):
Reviewed on the merged head (`cf7a0c80` — your `87ab7fc9` plus a clean merge of main). Engine suite **2722 passed**; working tree clean.

The correction itself is right and the rewritten ADR-0004 Context is the best part of it — it names registrars without asserting run-mode causation, states plainly that "which spelling a Cowork session exposes has been **observed to move**", and dates every observation to its source. That is the shape the repo asks for.

Your acceptance check does pass: **6 matches on `main`, 0 on this branch.**

## [Blocking] The acceptance grep passes while its own counterexample survives

The pattern is self-chosen, and it misses a live instance because that instance uses a synonym. `ADR-0004:115`, in the Alternatives table:

> Missed Cowork's **on-computer registration, which uses `display_name` with no `remote-devices` segment**. `record-extractor` was refused outright … **in the mode that reaches the host extension directly**.

That is the same run-mode→prefix causation the PR removes everywhere else, stated in the present tense as live rationale for rejecting an option. The grep says `reaches .*\.mcpb.*directly`; this says "reaches the **host extension** directly", so:

```
author's pattern vs that sentence  -> 0 matches
```

So ADR-0004 now contradicts itself 85 lines apart: the Context says the spelling has been observed to move, and the Alternatives table says on-computer registration *uses* the bare `display_name`. The ADR is the file every other doc in this PR mirrors, which makes the internal contradiction the one that matters most.

You scoped the Alternatives table out as "historical, and outside this issue's ADR edit scope". The reversal-table row at `:60` I would accept on that basis — it is a dated log of what was concluded at #1341 and reads as one. `:115` is different: it is present-tense rationale for a standing decision, not a record of a past one.

This also makes the acceptance check non-falsifiable in the way that matters here. A check whose pattern cannot match the surviving counterexample reads as coverage and is not — the repo's own rule. Widening it to something like `reaches .*(\.mcpb|host extension).*directly|on-computer registration` would have caught this.

## Minor

**The "47" claim is incomplete.** `allToolSchemas` has **48** entries and `manifest.json` agrees, so 47 is genuinely stale — but the body says it "replaces the two stale `47`-tool counts", and two more survive in `ADR-0002`:

```
ADR-0002:50  "Today that is 47 tools, 27 skills, and 4 agents."
ADR-0002:63  "at 47 tools Cowork already defers the schemas past a size threshold"
```

(`volume-search-tool-spec.md:488`'s "all 47 anchors" is unrelated — not a tool count.) ADR-0002 is a different claim in a file outside your checklist, so leaving it is defensible; the body's "the two stale counts" is what needs narrowing, or fix all four while you are here. The `:63` one is load-bearing for a deferral argument, so it is the one worth correcting.

**On the date reconciliation you asked about:** writing it as "observed 2026-08-15, reported via #1732 on 2026-08-19" is the right call — it preserves both sources rather than picking one, and an observation date and a filing date are genuinely different facts. I would keep it.

## What I checked and did not

I verified the grep on both sides, read every changed passage in ADR-0004, counted `allToolSchemas` and the manifest, and ran the engine suite. I have not re-derived the censuses themselves — no CI job reaches Cowork in any run mode, which your "What CI does not prove" section already says plainly, and that honesty is the right way to ship prose nothing can verify.

Requesting changes on `ADR-0004:115` and the grep that let it through. Both are small, and the rest of this PR is careful work.

T-FEH:
Two additions, and the first is a correction to my own review. My changes-requested above stands; these go with it.

## I got the date reconciliation wrong, on the one point you asked about

You flagged the Windows census date as your "Unsure about" and asked for a ruling. I told you *"observed 2026-08-15, reported via #1732 on 2026-08-19 is the right call… I would keep it."* That was wrong, and I should have read the sources before answering.

```
#1732 created: 2026-08-18T15:48:46Z
#1732 census result: kofiatinka12 @ 2026-08-19T16:39:19Z
```

The issue did not exist on 08-15, so its census cannot have run then. Nothing in that comment dates the session to 08-15 either.

The 08-15 Windows date is real, but it belongs to a **different census** you have not cited — DallanQ on **#1639 @ 2026-08-16**:

> **Windows census is in…** Cowork on **Windows**, genealogy project connected, **2026-08-15** — all 47 tools, one prefix… **The bare `mcp__Genealogy_Research__` spelling does not appear.** Identical to the macOS desktop census the same day.

So there are **three** censuses, not two, and the current clause fuses two of them — taking the date from one and the source from another:

```
1. macOS,   2026-08-15  (#1639)
2. Windows, 2026-08-15  (#1639, reported 08-16)
3. Windows, 2026-08-19  (#1732)
```

The evidence is *stronger* than the PR claims — three sessions, two platforms, five days apart — which is a better argument for "the registrar moves" than the version written. It appears in six files plus both test comments, so it needs one canonical form propagated:

```
You wrote:    Two censuses on 2026-08-15 (macOS desktop, and Windows via issue #1732,
              reported 2026-08-19) found every genealogy tool
Change it to: Three censuses — macOS and Windows on 2026-08-15 (recorded on issue #1639),
              and a second Windows session reported 2026-08-19 (issue #1732) — found
              every genealogy tool
```

For a PR whose whole purpose is installing "state only what was observed and on what date", the dates have to be right — and I should not have waved this through when you had explicitly raised it.

## The acceptance grep's scope excludes the four shipped agents

Separate from the synonym gap in my review: the grep covers `CLAUDE.md docs/ packages/engine/mcp-server/tests/`. It does not cover `packages/engine/plugin/agents/`, and all four shipped agents still carry the old mapping in their `tools:` comments:

```
record-extractor.md:22   # `Genealogy_Research` (Cowork on this computer, no bridge)
gps-mentor.md:23         # (cloud Cowork), and `Genealogy_Research` (on-computer Cowork)
image-reader.md:8        # `Genealogy_Research` (on-computer Cowork)
image-reader-opus.md:8   # `Genealogy_Research` (on-computer Cowork)
```

Your "Didn't change" note covers the **spellings**, and the issue's "do not touch" does too — deleting the bare one is #1341. Neither forbids correcting the **comment**. Three of those four had their specs corrected in this PR, so spec and shipped artifact now disagree, and `record-extractor.md` is the copy the other three point at ("See record-extractor.md for the full rationale").

```
You wrote:    # `Genealogy_Research` (Cowork on this computer, no bridge). The latter two
Change it to: # `Genealogy_Research` (Cowork, bare display_name spelling). The latter two
```

```
You wrote:    # (cloud Cowork), and `Genealogy_Research` (on-computer Cowork). Matched
Change it to: # (bridged), and `Genealogy_Research` (bare display_name). Matched
```

## Smaller, while you are in these files

- `ADR-0009-refuted-agent-design-claims.md:75` still reads `47 (allToolSchemas, unchanged since 2026-07-26)` — an ADR of refuted claims carrying a stale one, and it names `allToolSchemas` specifically, so it is closer to your two than the `ADR-0002` pair I mentioned.
- `CLAUDE.md:353` still says *"a cloud-mode check would have passed while on-computer was broken"*. `ADR-0004:180` has the better version — *"would have passed throughout #1341"* — which names the issue rather than the mode. Worth making CLAUDE.md match.
- The retained failure strings in `agent-tool-names.test.ts` ("…when the task runs on the user's own computer") describe a condition the new doctrine says has not been observed. Nothing asserts on them, so changing them is free, and as written they invite exactly the "then the bare spelling doesn't matter" reasoning this PR exists to prevent.

Independently confirmed from my review: the `ADR-0004:115` contradiction is real, and the ratchet bumps are exact rather than padded — adding a ref, removing one, or un-bumping the baseline each fail the check.

Asimi1234:
Thanks — all four verified against the repo and fixed in `f8c94dc5` (now on the merged head `af9b794b`). `make test-all`: 2874 passed.

**[Blocking] `ADR-0004:115` self-contradiction — fixed.** Confirmed: the cell asserted present-tense causation ("on-computer registration, which uses `display_name`… in the mode that reaches the host extension directly"). Reworded to historical `#1341` fact — "Missed the bare `display_name` registration (no `remote-devices` segment), which Cowork exposed live in #1341" — no row deleted, verdict/evidence unchanged. You're right that `:60` (a dated log) and `:115` (standing rationale) are different cases.

**[Blocking] The grep was non-falsifiable — fixed.** Widened to catch the synonym and cover `agents/` (which the old dir scope excluded): `reaches .*(\.mcpb|host extension).*directly|on-computer registration|on this computer, no bridge|on-computer Cowork|in on-computer mode|directly, no bridge|\((cloud|on-computer) Cowork\)`. Zero on branch, matches on `main`. Your dir-scope catch mattered independently — it surfaced the four agent comments below.

**The date — you were right to reopen it, and there are three censuses.** Verified `#1732` created `2026-08-18T15:48:46Z`, so its census can't be 08-15; the 08-15 Windows session is the separate one DallanQ recorded on #1639 @ 08-16. My draft fused #2's date with #3's source. Corrected to one canonical form and propagated to all six docs + both test comments: *"macOS and Windows on 2026-08-15, and a second Windows session via #1732 on 2026-08-19."* Agreed the stronger three-session evidence is the better argument.

**Agent `tools:` comments — fixed (comment-only).** `record-extractor.md`, `gps-mentor.md`, `image-reader.md`, `image-reader-opus.md` reworded from "(cloud/on-computer Cowork)" to "(bridged)/(bare display_name)". Spellings untouched (deleting the bare one is #1341); `record-extractor.md` is the canonical copy the other three point at.

**Minors — all taken.** `ADR-0009:75` `47`→`48` (dropped the false "unchanged since 2026-07-26"); `ADR-0002:50` and `:63` `47`→`48`; `CLAUDE.md:354` now reads "throughout #1341" to match `ADR-0004:181`; the retained failure strings reworded to "wherever Cowork exposes the bare spelling" — invariant kept, run-mode causation dropped — and the sibling `it()` name renamed off "on-computer prefix" for consistency. Body's "two stale counts" claim narrowed accordingly.

Left as noted (not a change): `project-create-tool-spec.md:17` ("Confirmed live in Cowork, on-computer mode") is unverifiable per #1732's own finding, but it's a tool's provenance note rather than the registrar mapping — flagging, not editing here.

Asimi1234:
Reversing my earlier deferral on `project-create-tool-spec.md:17` — it should be fixed here, and now is (`ecf59604`).

I'd waved it through as "provenance, not the registrar mapping." Reading the surrounding block, that was wrong: step 3 of the same confirmation says the write landed through **`device_bash`** — a remote-device *bridge* tool — so that "Confirmed live in Cowork, on-computer mode" session was demonstrably reaching the host through the bridge, not on-computer. It's the same falsified mapping this PR exists to remove, contradicted by its own evidence three lines down.

It also exposed a gap in my acceptance grep: I matched `in on-computer mode` but not the bare `on-computer mode` here, so the check couldn't have caught it. Widened accordingly; it now matches on `main`, zero on branch. Dropped the run-mode label (kept the `device_bash` fact, which is the point of the confirmation). Link/packaging tests still 236 pass; no ref-count change.

Asimi1234:
**Requesting `eval-cosmetic-skip` (senior-only) — the `runlogs` failure is comment-only.**

`runlogs` is red because two run-log snapshots went stale:
- `record-extraction` (`v1_2026-08-24_17-40-15.json`) — 3 snapshot files differ
- `search-images` (`v1_2026-08-25_15-26-32.json`) — 2 snapshot files differ

Those "differing" files are exactly the agent bodies this PR reworded, and the diff is **comment-only** — the `tools:` YAML comments changed (e.g. `(cloud Cowork)` → `(bridged)`, `(on-computer Cowork)` → `(bare display_name)`); every tool entry and the agent prompt body are byte-identical:

```
-  # hosted web), `remote-devices__Genealogy_Research` (cloud Cowork), and
-  # `Genealogy_Research` (on-computer Cowork). See record-extractor.md for the
+  # hosted web), `remote-devices__Genealogy_Research` (bridged), and
+  # `Genealogy_Research` (bare display_name). See record-extractor.md for the
```

No behavior change, so a paid re-run of either skill would recompute an identical result purely to refresh a snapshot hash — the case the label exists for. (`gps-mentor.md` also changed but is tracked by neither run log, so it doesn't affect this.)

If you'd rather I re-run the harness for both skills instead, say the word and I will.

EdmondOware:
The `runlogs` red is a correct `eval-cosmetic-skip` case and I verified it independently rather than on your summary. All five drifted snapshot files across both skills are agent files, and in all four agents the sorted `tools:` and `disallowedTools:` entry lines hash identically before and after. Only `#` comments changed, which eval/CLAUDE.md names explicitly as eligible.

One thing before the label goes on, since it voids on your next push anyway and you will want it applied last. The correction is right and the evidence is what makes it: Windows shows up on both sides, bare live on 08-04/05 and bridged only on 08-15 and 08-19, which is what actually refutes the run mode reading. But the sweep is incomplete, and every remaining site is in a file this PR already opens.

```
CLAUDE.md:290        "The on-computer registrar repeated the shape"
CLAUDE.md:353        "only in the run mode being tested — a cloud-mode check"
ADR-0004:60          "Cowork on the user's own computer is a third registrar"
ADR-0004:181         "only in the run mode being tested — a cloud-mode check"
architecture.md:764  "the on-computer registrar ... the on-computer spelling was missing"
```

These are present tense claims about what the registrar is, not narration of what was believed at the time. The two `run mode being tested` lines are the sharpest: they reason on the axis this PR removes. If the exposed spelling moves over time independently of run mode, then a live check is only valid for the spelling that session exposed and only at that moment, so run mode is the wrong frame for what the check proves.

Worth finishing in this PR rather than a follow up. Leaving the old frame thirty lines below the corrected block in the same file is the case the PR itself argues against.

Asimi1234:
Sweep done and pushed — head is now `5e27fa14`. All five sites reframed off the run-mode axis onto the spelling-moved-over-time frame the corrected block already uses:

- `CLAUDE.md:290` — "The on-computer registrar repeated the shape" → "The bare `display_name` registration repeated the shape"
- `CLAUDE.md:353` — "only in the run mode being tested — a cloud-mode check would have passed throughout #1341" → "only for the spelling that session exposes — the bare form was live in #1341 but absent in the later censuses, so a green check proves binding for one spelling at one moment, not in general"
- `ADR-0004:60` — "Cowork on the user's own computer is a third registrar" → "the bare `display_name` spelling is a third registrar, observed live here"
- `ADR-0004:181` — same run-mode line as `CLAUDE.md:353`, reframed identically
- `architecture.md:764` — "The on-computer registrar repeated the shape … the on-computer spelling was missing" → "The bare `display_name` registration repeated the shape … that spelling was missing"

`grep` for `on-computer registrar` / `run mode being tested` across the three files: zero remaining. Reference-neutral, so no ratchet move — `doc-links` + `measured-figures` 158/158, full packaging suite 604/604. The remote branch had picked up a main-merge touching the same three files, so I rebased on top (clean) and re-ran the guards green before pushing.

No snapshot inputs touched, so the `runlogs` state is the same comment-only case you verified. Ready for the label whenever you are.

EdmondOware:
Five are done, and the two `run mode being tested` rewrites are better than what I suggested — anchoring the claim to the spelling a session exposes and to one moment rather than in general is exactly the right frame.

There is a sixth, and it is in a file this PR already edits: `packages/engine/mcp-server/tests/packaging/agent-tool-names.test.ts:57` still reads "only a live Cowork session can, in the run mode being tested", which is the same sentence you just fixed in CLAUDE.md and ADR-0004.

My earlier list was incomplete because I grepped three files rather than the repo, so that one is on me rather than on your sweep.

Two more sit outside this diff and are follow up scope rather than an ask here, both naming the bare spelling "on-computer": `eval/harness/tests/unit/test_check_rubric_tool_drift.py:155` and `eval/harness/tests/unit/test_skill_invocation.py:1706`. No committed run log hashes anything under `eval/harness`, so fixing those costs no paid run whenever someone gets to them.

I have re-verified the cosmetic claim at this head after your two main merges: the drift is still the same five agent files, still comment only, and all four agents' entry lists hash identically. I will put `eval-cosmetic-skip` on once you have pushed the sixth site, since the label voids on every push.

T-FEH (approved):
Second pass, on `c30f6f33` (your `5e27fa14` plus a clean merge of main). Every finding from my review is fixed, and the date correction landed exactly right. One consequence to sort out before this can merge, and it is one I caused.

## All four items are closed

- **The blocker.** `ADR-0004:115`'s causation clause is gone — `grep` for "the mode that reaches the host extension directly" and "on-computer registration, which uses" returns **0**. The ADR no longer contradicts its own Context.
- **The date.** Now "Three censuses … macOS and Windows on 2026-08-15, and a second Windows session via issue #1732 on 2026-08-19", in both ADR-0004 and `architecture.md`, and CLAUDE.md carries the same. That separates the two sources correctly, and it is the stronger claim — three sessions across two platforms, five days apart.
- **The agent bodies.** All four comment fixes landed, and they are comment-only: no `tools:` entry changed in any of them.
- **Repo-wide.** A broader sweep than the acceptance grep — `on-computer Cowork`, `Cowork on this computer`, `on-computer mode`, `reaches .* directly` across `CLAUDE.md`, `docs/`, `agents/` and `tests/` — returns **nothing**. The reframe is complete.

## The gate is now red, because of the fix I asked for

`check_runlogs.py` fails with a rule-2 violation: `search-images`' latest run log is no longer active, because `image-reader.md` and `image-reader-opus.md` are in its snapshot.

That is a direct cost of my last review — I asked for those comments to be corrected without checking that two of the four agents are snapshot-tracked. My apologies; the finding was right but I handed you the bill without naming it.

**The remedy is the `eval-cosmetic-skip` label, not a paid re-run.** Every one of the four edits is a `#` comment inside `tools:` frontmatter, with no entry added, removed or reordered — behaviour-neutral by inspection, which is exactly what that label documents. The gate's own message says so. It needs a senior to apply; I am one, so say the word and I will, or apply it yourself.

## The gate under-reports, and this is the part to be careful about

It flagged `search-images` only. Running the harness's own `diff_snapshot_vs_disk` against both affected skills:

```
record-extraction   drift entries: 3   image-reader.md, image-reader-opus.md, record-extractor.md
search-images       drift entries: 2   image-reader.md, image-reader-opus.md
```

`record-extraction` is stale too — more stale than the skill that was flagged — and it was **active on main** before this PR, so this branch introduced it. If you act only on what the gate printed, that one stays stale and surfaces on somebody else's PR.

So the cosmetic-skip needs to cover both, and it is worth a card in its own right: a changed plugin agent flipped two skills, and rule 2 named one. That is `nothing-checks`-shaped — the gate is the thing that is supposed to catch exactly this.

## Approving

The doctrine work is complete and correct, and the census correction is now the strongest version of the claim rather than the one that fused two sources. My block is cleared. What remains is procedural: the label on both skills, and ideally a card for the gate gap.

For the record, since I raised it last time: `ADR-0002`'s two remaining `47` counts are still there and still out of scope for this PR — worth someone's five minutes, not yours.

Asimi1234:
Sixth site done and the conflict is resolved — head is now `d75ba756`, PR is mergeable again.

**Sixth site (+ a seventh).** `agent-tool-names.test.ts:57` is reframed. Grepping the file turned up the **same sentence again at line 363**, which the earlier greps missed, so I fixed both — "in the run mode being tested" → "and only for the spelling it exposes." Repo-wide grep for `run mode being tested` / `on-computer registrar` across `CLAUDE.md`, `docs/`, `agents/` and `tests/packaging/` is now zero. (The two `eval/harness/tests/` sites Edmond flagged as follow-up are left untouched.)

**Conflict with main, resolved in `d75ba756`:**
- `image-reader-opus.md` + its spec — **modify/delete**: main retired the Opus re-reader (#2044). Accepted the deletion; the branch's comment reword of a now-deleted file was moot.
- `CLAUDE.md` — took main's condensed version of the "broken every agent in Cowork twice" passage; it's frame-neutral, so no doctrine lost.
- `doc-links.test.ts` — rebaselined `ISSUE_REF_BASELINE` to the real post-merge counts (`CLAUDE.md 5`, `architecture.md 22` — the census `#1341`/`#1732` evidence refs). Full packaging suite 602/602.

**Please re-verify the cosmetic-skip before applying it — the merge changed the basis.** You both verified the drift as comment-only on `image-reader-opus.md`, but #2044 now *deletes* that agent and makes a real content edit to `image-reader.md`, so the snapshot picture at `d75ba756` is no longer what you checked. And per Tife's catch, it still needs to cover **both** `record-extraction` and `search-images`, not just the one the gate prints. The label voids on this push regardless, so it's the right moment to re-run `diff_snapshot_vs_disk` against both skills at this head.

EdmondOware:
Sixth site is in, and finding the seventh at line 363 was the better catch. On `main` that sentence is split across two physical lines, "in the run mode" then "being tested)", so no line grep for the phrase could have reached it. That is why my list stopped at five and yours at six.

Four sites remain, five lines in total, all in files this diff already opens. I swept for the concept rather than the phrasing, and the pattern is at the bottom so you can run it yourself rather than take the list on trust.

1. `docs/adrs/ADR-0004:107` reads "`select:mcp__genealogy__…` resolves to nothing in either Cowork mode." You already fixed this exact sentence in `docs/architecture.md:784`, where it now reads "resolves to nothing under either Cowork spelling, bridged or bare." Copy that wording into the ADR. It is the file the other five mirror, so it should not be the one still carrying the old axis.

2. `docs/specs/research-append-tool-spec.md:600` reads "In Cowork's cloud mode a bridged MCP call is killed at 60s (a limit imposed by the bridge, not settable from our side)." This one is worth doing whatever you decide about the other three, because it is not framing. The parenthetical already says the bridge imposes the limit, and all three censuses in this PR found the bridge in sessions with a connected project folder. So the limit reaches those sessions, while the sentence tells a reader in one that it does not. Drop the mode qualifier and attribute the kill to the bridged call.

3. `docs/architecture.md:1483` and `:1485` carry the sentence you reframed in `CLAUDE.md` and `ADR-0004`: "`make agent-smoke` reads what the runtime resolved, not what bound, and only in the mode it runs in. So a green CI run says nothing about whether an agent can call what you granted it; only a live session in the run mode you care about does". Every sweep missed it because it says "the mode it runs in" and "run mode you care about" rather than "run mode being tested". It is §9.4 gap 1, "Spelling is not binding", which is the register `CLAUDE.md` sends readers to for what nothing checks, and the two sections it cites, §5.2 and §8, are the ones this PR rewrote. Use the wording you landed on: valid for the spelling that session exposed, and only at that moment.

4. `packages/engine/mcp-server/tests/packaging/agent-tool-names.test.ts:16` still opens the registrar list with "Cowork IN THE CLOUD reaches the host-installed .mcpb through a remote-device bridge", six lines above the paragraph saying the spelling has been observed to move. The bullet under it now reads "Cowork can instead expose the bare display_name", so the two are labelled on different axes. All three censuses carried the bridge in connected folder sessions, so the bridge is no more a property of cloud mode than the bare spelling is of running locally. Name it the bridged registrar, the way the bullet below it names the bare one.

The sweep, if you want it as the acceptance check in place of the phrase list:

```
git grep -niE "Cowork mode|cloud mode|in the cloud|cloud Cowork|run mode|mode it runs in|on-computer|own computer|on this computer" -- CLAUDE.md docs/architecture.md docs/adrs docs/skill-lifecycle.md docs/specs packages/engine/plugin/agents packages/engine/mcp-server/tests README.md
```

Fourteen hits today. Eight are the new doctrine stated correctly and stay. One, `docs/specs/e2e-test-spec.md:1310`, is the e2e harness `shadow` and `deny` guard settings, a different axis entirely. The remaining five lines are the four sites above.

Asimi1234:
Heads-up: a second main-merge conflict came in and is resolved at `83258b59`.

Main's measured deny rewrite (denies deleted, grant/deny overlap now a lint) overlapped the files this reframe touches. I took main's doctrine as the base and re-applied only the run-mode reframe on top — four conflicts across `ADR-0004`, `architecture.md`, and `research-append-tool-spec.md` — and dropped my earlier "the deny is load-bearing" line in the research-append spec, since the probe measured it false.

Verified on the committed tree: acceptance grep = 9 (the 8 doctrine lines + `e2e-test-spec.md:1310`, the different-axis one); packaging 613/613. The `runlogs` drift, re-checked at this head via `diff_snapshot_vs_disk`, is unchanged and comment-only — `record-extraction` (`image-reader.md`, `record-extractor.md`) and `search-images` (`image-reader.md`).

EdmondOware:
All four sites are closed, and the second merge is the part I looked hardest at. Main's deny rewrite landed across the same files, and taking its doctrine as the base and re-applying only the reframe on top is the right order to do that in. Nothing of main's was lost in the process.

One last thing, and it is the first of its kind, so it is worth a sentence rather than a silent number. This raises `ISSUE_REF_BASELINE` in four contract documents, and the ratchet has never gone up before. Since the table landed on 4 August, `docs/architecture.md` has walked 28, 27, 24, 20, 19, and the two spec entries have not moved at all. `CLAUDE.md` is the sharpest case: it did not drift down to 2, it was purged there four days ago by #1990, which stripped six references, and this takes it back to 5.

The citations are the right kind and I would not drop them. `#1341` and `#1732` are where the census output actually lives, and since nothing in CI can reproduce those sessions, removing the numbers removes the reader's only way to check the claim. But the docstring above the table says adding a reference "fails outright", and its evidence shaped carve out is written about what remains after the purge rather than about adding new ones, so as it stands the guard now reads as loosened with nothing saying why. Add a sentence to that docstring recording that this raise was deliberate, and that a census citation is the exception because the observation cannot be reproduced from the repo. That way the next reader takes 5 and 22 as a documented exception rather than headroom.

Asimi1234:
Docstring note is in at `7f9c4eff`. Added five lines above `ISSUE_REF_BASELINE` recording that the raise is deliberate (#1639), that a census citation (`#1341`/`#1732`) is the sanctioned exception because CI cannot reproduce a live Cowork session's registrar spelling, and that irreproducibility is the *only* reason to add a reference — every other addition still fails outright.

The refs sit inside the test file's docstring, which the ratchet doesn't scan, so the baseline is unchanged (5 / 22). Full vitest 2748/2748.

EdmondOware:
The docstring note is better than what I asked for. Naming irreproducibility as the sole sanctioned reason to add a reference turns a one time exception into a rule, which is what stops the next person reading 5 and 22 as headroom.

One number left, and it is in a line this PR already edits. `docs/adrs/ADR-0002:50` reads "Today that is 48 tools, 27 skills, and 4 agents". There are five: `gps-mentor`, `image-reader`, `proof-conclusion`, `record-extractor`, `research-exhaustiveness`, and it has been five since 25 August. `docs/architecture.md:217`, in this same diff, already says 5, so as it stands the PR ships two documents disagreeing about the same count. Change the 4 to a 5.

To be clear about what should not change: "Three of the four agents declare only MCP tools" in `ADR-0004`'s #1341 row is right and stays. There were four agents at #1341, and that row is dated history, the same category you protected elsewhere.

Asimi1234:
Both in at `ff76c3eb`. `ADR-0002:50` now reads "48 tools, 27 skills, and 5 agents" — the tools (48, `manifest.json`) and skills (27, SKILL.md dirs) on that line check out too, so the whole line is now true.

While confirming the count I found a third disagreement: `docs/skill-dataflow.md:14` still said "27 skills and **6 agents**" — stale from the #2044 Opus retirement, and not a file this PR otherwise touches. Fixed it in the same commit so all three docs agree at 5 (`architecture.md:217/219` already did). The dated `#1341` "three of the four agents" rows in ADR-0004 are left as-is — that's history, per your note.

Actual count verified against `ls agents/*.md`: `gps-mentor`, `image-reader`, `proof-conclusion`, `record-extractor`, `research-exhaustiveness`. Full vitest 2748/2748.

EdmondOware (approved):
Approved.

The rationale now states what was observed and when, with nothing left claiming what causes what. Three censuses across two platforms and five days is a stronger claim than the two the card was written around, and it is the version that holds up the next time the registrar moves.

The exception recorded on the reference ratchet is the part still doing work in six months. Naming irreproducibility as the sole sanctioned reason to add a reference means this cannot be cited as precedent for anything else.

The agent count now agrees across `ADR-0002`, `docs/architecture.md` and `docs/skill-dataflow.md`, which had been contradicting itself two hundred lines apart.

Issue #1899, opened 2026-08-25
Replay the project-folder notice through hydration — any renderer reload drops it (split from #1869)


chesworthrm:
> **Reviewed 2026-08-25 before junior handoff.** No decision needed — both forks were
> already settled in the repo. But the acceptance line was wrong in a way that expands
> this into Python, and the carried blast radius was low by half. Read the scope decision
> below before starting.

**Touches:** `packages/viewer-ui/src/transport.ts`, `packages/viewer-ui/src/contexts/ResearchDataProvider.tsx`, `apps/electron/src/renderer/src/transport/IpcResearchTransport.ts`, `apps/web/src/transport/WsResearchTransport.ts`, `apps/electron/src/main/watcher.ts`, `apps/electron/src/main/index.ts`

**Split from #1869 on 2026-08-25** (board pass). #1869 bundled two halves: pinning
the project folder so the agent and viewer share one, and replaying the folder
notice through hydration. The pinning half needs a Cowork-side mechanism and a
shape decision only the lead can make, so it is not startable; #1869 keeps that
half and now carries `needs-decision`. **This card is the hydration half, which
#1869's own body says "is takeable now and does not depend on the decision in 1."**

## The defect

`getCurrentState()` returns `folderPath`, `research` and `gedcomx` — **not the
notice**. So unlike a research update, a notice that was already sent is never
replayed, and the renderer only registers its listener when React mounts and
calls `subscribe()`.

PR #1722's merged body records this under "Known limitation — the notice is not
replayed on hydration", and already corrects itself there from "dev-only":
**any** renderer reload drops it, not just the dev launch.

## Why it matters now

PR #1722 shipped the nested-project banner as the in-repo signal that the agent
wrote into a subfolder while the viewer watched the top level. A tester who
reloads the viewer stops seeing that banner, and the symptom they report is
"lost my files between sessions" — which is the exact reading the banner exists
to prevent. Beta is ~9 weeks out.

## Scope decision — the web transport returns `null`

`notice` becomes a **required** field on `ProjectStateSnapshot`, matching
`research`/`gedcomx`/`label` (`packages/viewer-ui/src/transport.ts:45-52`), and
`WsResearchTransport.getProjectState` returns `notice: null`. **Do not add a notice to
the control plane:** the hosted path pins the project to `/project`, so `onNotice` is
never fired there — the recorded decision is the comment at
[...]

Issue #1869, opened 2026-08-24
Pin the project folder so the agent and viewer share one (#1317 bug 2 root cause)


francis-2008-happy:
> **Reviewed 2026-09-01 before junior handoff.** No decision needed; the shape and
> the fold-in are both already ruled. Two things added that the body did not say:
> **where** the refusal binds (`project_create` only — the obvious reading would
> put it in the shared read for twelve tools and lock a user out of an existing
> nested project with no exit), and that `project_create` runs **live** in the
> unit harness, so a false deny first surfaces inside a paid `init-project` run.

**Also in scope (lead, 2026-08-31):** the ancestor rule fires only when an ancestor
already holds `research.json`. It cannot see the other half of the same
divergence — a picked folder with nothing above it, where the agent creates the
project one level down (the case `formatNestedPicker`,
`apps/electron/src/main/watcher.ts:135`, was shipped for). Nothing
filesystem-derived can see that case, so it takes ADR-0011's row 5 (needed at
call time, unenforceable): add one sentence to `project_create`'s `projectPath`
description — create the project in the folder you were given, never in a
subfolder of it — chosen over a second card so both halves of one divergence
ship together. Record the reasoning in
`docs/specs/project-create-tool-spec.md` §5 ("What this does not solve"), not
here. Note in the PR that `init-project`'s recorded eval grade was earned under
the old description and is not being regraded (`snapshot.py` excludes
`mcp-server/src`, issue #1770; precedent issue #1409).

The alternative was a second card for the sentence, keeping this PR a pure
filesystem predicate. Declined: one sentence adds no file to the blast radius
and invalidates no run log, and a second card would leave the case live while it
waited. Accepted cost: ADR-0011 records tool-description strength as unmeasured.

---

> **Scope narrowed 2026-08-25 (board pass).** Half 2 — replaying the folder
> notice through hydration — moved out to **#1899**, which is now CLOSED (the
> blockquote said "in Ready"; corrected 2026-09-01). This card
> is now **half 1 only: pin the project folder.** It is labelled
> **The shape was chosen 2026-08-27 — a filesystem-derived rule in the MCP
> server, not a Cowork-side mechanism.** See the decision below. The work is
> junior and this is startable.
>
> **Carried constraint:** PR #1722 shipped the two in-repo *signals* — a
> dismissible banner when a nested `research.json` is found alongside a
> top-level one, and a pointer in the `project:select-folder` rejection when the
[...]

PR #1958, opened 2026-08-27
viewer: replay the project-folder notice through hydration (#1899)

francis-2008-happy:
## Summary

**Tier: Normal.** Closes the hydration half of the folder-notice defect. Issue #1899 is the whole of this card; issue #1869 keeps the project-folder-pinning half, which is `needs-decision` and untouched here.

The nested-project banner was only ever push-delivered. `getCurrentState()` did not carry the notice, so a renderer that mounts *after* the send never saw it — any reload dropped the banner. The symptom a tester reports is "lost my files between sessions", which is precisely the reading the banner exists to prevent.

Now the scan stores the message as well as sending it, `getCurrentState()` returns it, and the provider replays it from the hydration snapshot.

## Review guidance

**Start here:** `apps/electron/src/main/watcher.ts` — the new exported `scanAndNotify`, and the `lastNotice = null` line in `stopWatching()`. That second line is the one to be suspicious of: without it, picking a *clean* folder replays the previous folder's notice on the next reload, which is a wrong banner rather than a missing one. It has its own test.

**Unsure about:** nothing outstanding. The one judgement call was making `notice` **required** on `ProjectStateSnapshot` rather than optional — it costs six stub updates, but it forces every transport to answer rather than silently inheriting `undefined`, which is how this class of bug survives.

## Plan

**Didn't change:** the control plane. `notice` is a viewer-transport concept only, and `WsResearchTransport` returns `null` deliberately — the hosted path pins the project to `/project`, so no notice is ever fired there and `apps/server` has none to serve. That decision is recorded in a comment beside the new field so the next reader does not re-open it. Also untouched: #1869's folder-pinning half.

**Acceptance check:** three tests, each mutation-checked one at a time.

| Mutation | What fails |
|---|---|
| delete `if (state.notice) setNotice(...)` in the provider | the hydration test, and **nothing else** (1 of 273) |
| delete `lastNotice = message` in `scanAndNotify` | 2 of 3 watcher tests |
| delete `lastNotice = null` in `stopWatching()` | `stopWatching clears the stored notice` — the one order-independent guard on this line (a second test also fails when the file runs whole, but only via module-state leak from the preceding test; under `-t` isolation only this one bites) |

The provider test asserts the notice reaches the banner with **no `onNotice` event fired at all** — the snapshot is the only route, which is the reload shape.

`scanAndNotify` is extracted from `startWatching` specifically so the storing line is reachable: `watcher.test.ts`'s own header says the chokidar/BrowserWindow integration is manually verified only, so without the extraction that line would have been green with it deleted.

**Deviated from the plan:** nothing on the implementation. The issue's 332-test baseline was stale — I re-derived it as instructed and got **378** (viewer-ui had grown 46 tests since the body was written).

## The manual reproduction

The issue names `make -C apps/electron dev-sample` as the check that actually reproduces the bug. Run both ways against the same sample project with a throwaway `sub/research.json` (created, observed, deleted — not committed):

| Build | Result |
|---|---|
| `abd76bb8` (pre-fix) | **no banner.** The viewer showed the project — 15 assertions, full sidebar — with a second `research.json` one level down, invisible. |
| `240939bb` (this branch) | **banner present**, above the stepper: *"Heads up: this folder has research.json in a subfolder ("sub/research.json"). The viewer only shows the top-level project — if your research is in the subfolder, reopen the viewer on that folder."* with a working dismiss. |

Same project, same `--project-dir` launch path, only the code differs. That path calls `startWatching` at window creation, before React mounts, which is why it is the one where the dropped notice is visible.

**How this was observed:** on screen, by the repo owner, from two screenshots. I could not observe it myself — no screenshot tool is available in my environment — so credit for the visual half is theirs, not mine.

**A trap for the next person verifying this by hand.** If `ELECTRON_RUN_AS_NODE=1` is set in your shell (it is, in a Claude Code terminal), `make dev-sample` dies with:

```
TypeError: Cannot read properties of undefined (reading 'isPackaged')
```

That error points nowhere near its cause — the Electron binary is running as plain Node, so `app` does not exist. Launch with `env -u ELECTRON_RUN_AS_NODE make dev-sample`. Two other prerequisites bit first: the Electron binary was not downloaded at all (`node node_modules/electron/install.js`), and it needs `libnss3 libnspr4 libasound2t64` — note `libasound2t64`, not the `libasound2` most guides give, which has no install candidate on Ubuntu noble.

## Blast radius

Thirteen sites, as the issue listed. Three typed stubs failed `make typecheck` exactly as predicted; a fourth (`packages/viewer-ui/src/__tests__/App.test.tsx`) could **not**, being behind `as unknown as ResearchTransport` — updated by hand, and worth knowing that cast hides contract changes from typecheck.

One addition beyond the checklist: `packages/viewer-ui/src/contract.test.ts` has claimed since it was written that *"the Electron IPC adapter and the web WS adapter run the SAME assertion against their live instances"*. The web half did not exist. It does now, and the comment names both files rather than gesturing at directories.

## Test plan

- [x] `make typecheck` — 5/5 tasks.
- [x] `make test-js` — 4/4 tasks, **378 → 387**: web 10 → 14, electron 97 → 100, viewer-ui 271 → 273.
- [x] Manual `dev-sample` before/after — above.
- [x] No skill run-log snapshot touched, so no `make eval-skill` run is owed.

## Why nothing-checks

The whole JS workspace was green with the notice un-replayed, and no job distinguished "delivered" from "dropped on reload". The three acceptance tests are what close that gap — which is why each one is mutation-checked rather than merely passing.

## Follow-on issues

**Folded in / filed instead:** Nothing filed. The `ELECTRON_RUN_AS_NODE` / missing-binary / missing-libs friction above is documented in this body rather than filed — it is environment setup for one manual check, not repo work, and I would rather a reviewer say whether it belongs in `DEVELOPMENT.md` than file a card pre-emptively.


promise-emmanuel (approved):
Approving — same verification as above.

One optional thing, not blocking. In the hydration test:

  expect(noticeHandler).not.toBeNull()

The comment above it says "Nothing pushed it — it came from the snapshot."
That assertion doesn't check that. It only proves subscribe registered a
handler. The test is still correct, because nothing in it ever calls the
handler — but the line reads like a guard and isn't one. Either drop it or
change the comment. Your call, I don't need it changed to merge.

clack391 (approved):
Approving. This is the second count toward the two the ruleset wants; the code-owner requirement was already satisfied by promise-emmanuel's approval, so nothing is gated on mine.

The implementation is right. All 13 checklist sites are present, all three acceptance items are met, `apps/electron/src/main/index.ts` is correctly untouched, and every ruling in the issue is honoured, including the two you could have re-litigated and did not (the required-not-optional field, and leaving the control plane alone with the decision recorded beside the field). Three of your four mutation claims reproduce exactly. I ran the suite on the test-merge rather than on head: typecheck 5/5 and 387 passing, 0 failed, forced uncached.

Everything below is a fix-in-PR list, not a condition. I am asking for the first three specifically because you have to push again anyway for update-branch, so they cost you no extra round.

## 1. The hydration wire has no test

`apps/electron/src/renderer/src/transport/IpcResearchTransport.ts:30`. Replace `notice: state.notice ?? null` with `notice: null` and the feature is dead end to end on the only platform that has it, with typecheck 5/5 and all 387 tests green. The contract assertion cannot catch it, because `null` satisfies `string | null`, and the only test that runs the contract against the live transport stubs `getState` to return `notice: null`.

This is the one thing I would like in before merge, because the card is labelled `nothing-checks` and its stated purpose is that no job distinguishes "the notice was delivered" from "the notice was dropped on reload". That distinction is now made at both ends and not in the middle.

The file already carries the precedent, in your predecessor's own words: the `onNotice` routing test sits under "The contract test above only checks shape, so deleting the wiring line leaves it green." Same argument, other direction.

I wrote and ran this; it passes on your head and fails with the wire killed:

```ts
  // The hydration wire. The contract test above only checks `string | null`,
  // which a hardcoded `null` satisfies, so without this the notice can be dead
  // end-to-end on the only platform that has it with the whole suite green.
  it('maps a stored notice through from getState', async () => {
    ;(window as unknown as { api: { getState: unknown } }).api.getState = async () => ({
      folderPath: '/p',
      research: null,
      gedcomx: null,
      notice: 'research.json is in a subfolder'
    })
    const state = await new IpcResearchTransport().getProjectState()
    expect(state.notice).toBe('research.json is in a subfolder')
  })
```

## 2. A 14th edit site the checklist missed

`docs/specs/hosted-web-workbench-spec.md:370-373` is a maintained `**As built, verified 2026-08-02.**` note that enumerates this interface's divergences from the specced one: "Three additions since: `getProjectState` also returns a `label` ... and three **optional** methods were added". Your required `notice` is a fourth, and the note was already one behind from `onNotice` in #1722. The spec header at line 23 makes those notes the mechanism by which a non-current section stays honest, so it is a live record rather than a frozen draft.

One trap: that file sits at exactly 4 issue references and `doc-links.test.ts` sets its baseline to exactly 4, so the edit must carry no `#NNNN`. State the fact without the ticket. I extended the note without a ref and `doc-links.test.ts` stays at 134 passing.

## 3. The `as unknown as` cast is removable, and it is free

`packages/viewer-ui/src/__tests__/App.test.tsx:17`. You identified this blind spot in the body and left it in place, which is fair, but it is one declaration: annotate `const stubTransport: ResearchTransport` and drop the trailing `} as unknown as ResearchTransport` to `}`. No other change is needed. I initially thought it also wanted `ok: true as const`; it does not, because the annotation contextually types the literal and `ok: true` narrows on its own. After that, typecheck reports TS2322 if `notice` goes missing, and App.test.tsx stays 6/6 with viewer-ui at 273.

## Two corrections to the body

**The mutation table overstates one row.** "removing the `stopWatching()` clear fails a **different** 2 of 3" is one guard, not two. Two do fail when the file runs whole, but `stores nothing when there is no nested research.json` passes under `-t` isolation with the same mutation, because it only fails through module state leaking from the preceding test. Only `stopWatching clears the stored notice` is an order-independent guard on the line the issue singled out as the dangerous one. Reordering the describe or using `.only` silently halves the claimed coverage.

**The blast-radius count is off in both directions.** Measured one stub at a time with forced typecheck: four fail, not three. The unnamed fourth is `apps/electron/test/setup.ts`, declared `const apiStub: AppAPI`. And two are invisible to typecheck, not one: `App.test.tsx` plus `apps/electron/src/renderer/src/transport/__tests__/IpcResearchTransport.test.ts`, the latter laundered through `(window as unknown as { api: unknown })`. That second one matters more than the arithmetic, because it is the file item 1 lives in.

## Smaller, take or leave

- **The stale-folder race guard has no test.** `watcher.ts:200`. Deleting `if (currentFolderPath !== resolve(folderPath)) return` leaves all 100 electron tests green. Your diff newly extends that guard to the store, and a stale store survives to the next reload, which is the wrong-banner class the issue calls worse than none. Reachable: `project:select-folder` is stopWatching plus startWatching, and your own comment says a full-depth scan can take seconds. A ~20 line test pins it (start on a clean folder, then let the in-flight scan for the old folder resolve, assert notice is null).
- **The new contract assertion is not pinned.** `contract.ts:24-27` can be deleted with all 387 still green. It does bind against a violating transport, so it is live rather than dead; what is missing is a test that violates it. No assertion in `contract.test.ts` has one, so this is the file's existing convention and not your regression. A 6 line negative case would be the file's first.
- **`(issue #1317, bug 2)` is the wrong item number.** #1317's body numbers three reports and the "lost files between sessions" symptom this banner prevents is item 3; item 2 is the research question not reaching the viewer, and there are no comments to renumber them. This is pre-existing in eight code sites plus two issue titles, so it is not yours, but you are relocating one instance, which is the moment to fix all eight or leave all eight. Your call which.
- **"with no chokidar and no Electron"** is not true of the tests as written. `watcher.ts:188` and the test block comment both claim it, but all three new tests must call `startWatching` to satisfy the guard, and that constructs a real watcher. The "no Electron" half is true. The issue wrote that sentence too, so the wording is inherited; I am only noting the prediction did not hold. A reword is the cheap fix. The design fix, passing the watched folder in as a parameter, would also make the test in the previous bullet trivial, if you think it is worth it.
- **promise-emmanuel's note on `expect(noticeHandler).not.toBeNull()` is right, and I would resolve it the other way.** Keep the line and reword the comment. That assertion does prove the renderer subscribed, which is what makes this the reload shape rather than a transport that never subscribes; dropping it loses real information. It just is not what "Nothing pushed it" says.
- **#1869 is no longer `needs-decision`.** The label came off at 2026-08-27T14:01, hours after you opened this. It is now `developer` and `nothing-checks`. Untouched is still correct.
- On the `ELECTRON_RUN_AS_NODE` and `libasound2t64` friction you asked about: I would put it in DEVELOPMENT.md rather than file a card. It cost you real time and it will cost the next person the same, and the error points nowhere near its cause.

## Please do not do this one here

There are seven near-identical transport and api stubs, six of which this diff hand-edits, and the shared home already exists and is already exported for exactly this: `contract.ts`, published as `@genealogy/viewer-ui/contract` and deliberately vitest-free. A `makeStubTransport(overrides?)` beside `assertTransportContract` would collapse those six edits to one and absorb item 3. It is the right refactor and the wrong PR. The issue told you to hand-edit the stubs, so this diff is compliant; if you want it, it is a separate card.

## The framing underneath all of this

Every guard on this feature is asserted against a stub this PR hand-edited to satisfy it. The field is verified where it was typed in by hand and nowhere it is produced. Items 1 and 3 and the refactor above are one fix, not three, which is why I would take item 1 now and leave the refactor for its own PR.

## Two things worth knowing

`pnpm test` and `pnpm typecheck` go through turbo, whose local cache is shared across worktrees. My first run of your suite printed `>>> FULL TURBO` and replayed logs whose paths named a completely different checkout. `make test-js` can report "4/4 tasks successful" having executed nothing, so pass `TURBO_FORCE=true` or run vitest inside the package when a number matters. CI is unaffected, since there is no remote cache configured.

The new tests are the first in `watcher.test.ts` to construct a real chokidar watcher, and `stopWatching`'s `watcher.close()` is not awaited before `afterEach` removes the temp dir. That is 10 out of 10 clean here and `js-tests.yml` is ubuntu only, so I have no evidence of a problem. I am flagging it rather than prescribing anything because I cannot test Windows and we have people who run these locally on it.

Still needed before merge: update-branch for the one-commit lag.

francis-2008-happy:
Thanks both — all the fix-in-PR items are in (`f837d5e1`), since I had to push for update-branch anyway.

**@clack391**
1. **Hydration wire test** — added your `maps a stored notice through from getState` to `IpcResearchTransport.test.ts`, verbatim. Mutation-checked: replacing `notice: state.notice ?? null` with `notice: null` reddens it (and only it), so the "delivered vs dropped on reload" distinction is now made in the middle, not just at both ends.
2. **14th edit site** — extended the "As built" note in `hosted-web-workbench-spec.md` to record the required `notice` return and (catching up from #1722) the `onNotice` handler. No `#NNNN` added; `doc-links.test.ts` stays 134 / that file at 4.
3. **The cast** — annotated `const stubTransport: ResearchTransport` and dropped `as unknown as`. Confirmed it now has teeth: dropping `notice` from the stub errors TS2322.
- **Body correction** — fixed the mutation table: the `stopWatching()` row is one order-independent guard (`stopWatching clears the stored notice`), not "a different 2 of 3"; the second only fails via module-state leak under whole-file order.

**@promise-emmanuel** — good catch. The real guard is the `waitFor(notice).toBe(...)` (deleting the hydrate line fails *that*); `expect(noticeHandler).not.toBeNull()` only proved subscribe ran. Dropped it and moved the "came from the snapshot" comment onto the waitFor.

Branch is current with main (0 behind); electron 5 / viewer-ui App+provider 22 / doc-links 134 all green, typecheck clean.

PR #1963, opened 2026-08-27
question-selection deep dive (#1668): rubric cleanup, four doctrine fixes, two new validators

### LINKED ISSUE #1668: deep dive: question-selection — and the rubric-deletion doctrine it pilots

mercyokum:
Closes #1668

Supersedes #1936 — same work, clean history. #1936's branch had an octopus merge (`8254cece`) that pulled in PR #1896's still-open, unreviewed `check-warnings` branch alongside a normal `main` merge. Since a merge can only add commits, never remove an earlier one from history, the only clean fix was to rebuild from `main` and replay just this PR's own six commits. That's this branch. GitHub also blocks force-push to `deep-dive-question-selection`, which is why this is a new branch/PR rather than an overwrite of the old one.

**Diff verified against current `main`: touches only question-selection files** — `docs/deep-dives/question-selection-*`, `packages/engine/plugin/skills/question-selection/**`, `eval/tests/unit/question-selection/**`, `eval/runlogs/unit/question-selection/**`, `eval/harness/validators/test_question_selection.py`, and the one-line `no-issue-refs.test.ts` allowlist cleanup. Nothing from #1896.

---

**The issue:**
question-selection's rubric had three grading dimensions that never once caught a real problem across every committed run — dead weight taking up grading slots for nothing. Separately, the skill itself had several doctrine bugs that nothing was checking for: a rule that could make it stop looking for more evidence before a question was actually finished, wording that let it treat an untried record type as the wrong kind of research move, and a couple of places where it asked the right question the wrong way (or didn't ask when it should have).

**The fix:**
- Deleted the three rubric dimensions the issue authorized for deletion (Prioritization logic, Objective scope match, Dependency awareness) — each is now covered by an existing validator instead. Kept the fourth (Question specificity), since nothing covers it yet, and added a basic validator as a floor under it.
- Fixed a self-contradicting rule in the skill that could make it stop creating new research questions too early, even while a real gap (like unsearched census years) was still sitting there.
- Fixed the skill's FAN-research guidance (researching people connected to the subject) so it can't be confused with "try a record type nobody's searched yet" — those are different research moves.
- Fixed two places in the skill's "ask before assuming" logic: one where it kept demanding more detail after the researcher had already answered, one where a new test's grading unfairly penalized it for correctly choosing to ask.
- Fixed a case where the skill could conclude "no gap exists" without ever actually checking for one, because it never called the tool that would have told it.
- Fixed a tool-usage trap where asking for "in progress" plan items could silently come back empty even when something was in progress.
- Removed a leftover issue-number citation from the skill's own instructions — cosmetic only, no instruction changed.
- Added new tests and validators to lock all of the above in for future runs.

Everything above is written up in detail, with the actual transcripts and paid-run evidence behind each fix, in `docs/deep-dives/question-selection-findings-2026-08-25.md`.

**Review history carried over from #1936** (all addressed, verified individually before each fix rather than taken on trust):
- @chrisedeson found two must-fix bugs (a stale rubric-dimension reference in two fixtures' grading instructions, and loose substring matching in `_VERIFY_SIGNALS` that both accepted and missed real failure modes) and three should-fix items (an over/under-matching vague-question regex, a presence-only check that didn't verify ordering, and a real coverage gap on `unblocks`) — all fixed and re-verified; approved after.
- @EdmondOware independently checked the genealogical doctrine against the repo rather than the write-up and confirmed all four SKILL.md fixes are methodologically correct, plus counted the corpus directly (36 questions sit `in_progress` with a written `proof_summary` — this is not an edge case).
- @jud-sdev caught a second real bug in the vague-question validator (two patterns matched their trigger phrase anywhere in a question, not just as its opening framing, flagging specific well-formed questions) — fixed and reverified against their exact examples — and then caught the branch-history issue this PR exists to fix.

**Deferred issues found:**
- A `research_query` tool bug, not specific to this skill: filtering its `plans` section by `status` matches the whole plan's own status, not any item inside it — so asking it for "in progress" work can silently come back empty even when something is in progress. Worked around it in this skill's own instructions, but the tool itself could describe this more clearly so every skill that reads plans doesn't have to learn it the same way.


EdmondOware (changes requested):
Strong dive. The rubric deletion note is the part I would point other people at: it names which validator now covers each retired dimension **and** states plainly what still has no mechanical check, which is the half these write ups usually skip. The doctrine fix separating `status: resolved` from a defensible tier proof_summary is genealogically right and matters well beyond this skill.

I verified the supersede claim rather than taking it on trust. Diffing this branch against deep-dive-question-selection, the only differences are check-warnings files and person-warnings-tool-spec.md. Zero question-selection files differ, so the earlier doctrine review carries over cleanly.

Three things before merge.

**1. The vague question validator fails a correct question.** The loop passes `re.IGNORECASE`, which makes `[A-Z]` match any letter, so ordinary words count as name tokens:

```
"Who is Patrick Flynn's father?"       flagged as too vague
"Who is Patrick Flynn's wife Bridget?" flagged as too vague
```

Both are well formed, and the first is close to what this skill writes under a parentage objective. The validator is universal rather than tag gated, so it runs everywhere. No committed run has hit it yet, so nothing is red today, but it will red a future paid run for correct behaviour. Minimal fix, verified on nine cases including all caps and lowercase:

```
You wrote:      r"^\s*who\s+is\s+[A-Z][\w'\-]*(?:\s+[A-Z][\w'\-]*){0,3}\s*\??\s*$",
Change it to:   r"^\s*who\s+is\s+(?-i:[A-Z][\w'\-]*(?:\s+[A-Z][\w'\-]*){0,3})\s*\??\s*$",
```

**2. `_VERIFY_SIGNALS` matches against all new questions joined together.** One correctly framed question masks a co-written bad one. Writing "Who were the parents of Anton Vogt?" alongside a correct confirm or refute question passes on the joined text; checking per question flags it. Latent today since no run writes more than one question, but the check reads as stronger than it is.

**3. The two new validators have no unit test.** 17 of the 26 validator files have one, and item 1 is exactly what a test asserting "Who is X's father?" passes would have caught.

One thing you do not need: `eval-cosmetic-skip`. `v1_2026-08-26_17-56-18` is fully active against this branch, zero of 47 snapshot files drifted, and the runlogs check already passes without it. No reason to spend a senior bypass.

mercyokum:
Thanks for the careful check on the supersede claim too, good to have that confirmed independently rather than on trust.

All three verified before fixing:

**1. `[A-Z]` under `re.IGNORECASE`** — confirmed exactly as you described: "Who is Patrick Flynn's father?" and "...wife Bridget?" both matched. Used your exact fix, scoping `IGNORECASE` off just the name-token group with `(?-i:...)` so "who is" itself stays case-insensitive but the capitalization requirement actually holds. Both examples now pass; the rubric's own "Who is Patrick Flynn?" still correctly fails.

**2. `_VERIFY_SIGNALS` on joined text** — confirmed with your exact case: a bad question written alongside a correctly-framed one passed under the old joined-string check. Now checks each new question independently; re-ran your example and the bad one is caught.

**3. No unit test for the new validators** — added `test_question_selection_validators.py`, 11 tests covering all four validators from this dive (not just the two you flagged), following the `test_search_images_validators.py` pattern. Confirmed the who-is test would have caught bug #1 directly, as you said.

No `eval-cosmetic-skip` applied, per your note — neither file is in the run-log snapshot, so nothing needed re-running. 113 harness unit tests pass (102 existing + 11 new).

EdmondOware (approved):
All three are fixed correctly.

The `(?-i:...)` scoping holds and the new regression test genuinely guards it. The `_VERIFY_SIGNALS` rewrite flags every unframed question rather than passing on a joined string. The new validator file covers all four of the dive's validators, not just the two I raised.

Approving.

jud-sdev (approved):

PR #1966, opened 2026-08-27
tree-edit: teach the guardianship→prior-marriage inference, and split the census caveat at 1880 (#1900)

### LINKED ISSUE #1900: tree-edit relationship-accuracy: teach the guardianship-to-prior-marriage inference, and qualify the census caveat to pre-1880
> **Reviewed 2026-08-25 before junior handoff.** No decision needed. **Correction 2 was
> over-stated and is rewritten below** — the sentence is about relationship *type*, which
> an 1880 census column does not settle either. The acceptance also bought a $7-15 e2e run
> for a signal it cannot deliver; that is now optional. Read the "Read-only neighbours"
> section before you edit anything.

**Touches:** `packages/engine/plugin/skills/tree-edit/references/relationship-accuracy.md`

**Merged card, filed 2026-08-25** (board pass). Two corrections to the same
reference file, batched so they share **one** paid `tree-edit` run instead of
buying two:

- the guardian-appointed-after-remarriage → prior-marriage inference, split out
  of **#1710** (whose remaining half touches `historical-context` and stays in
  Backlog behind #1663)
- the pre-/post-1880 census qualifier, from **#1879** (closed into this card)

`tree-edit`'s eval slot is free today; its suite is 13 tests under
`eval/tests/unit/tree-edit/`.

---

## Correction 1 — the guardianship → prior-marriage inference is taught nowhere

Split from #1413, which asked for an e2e fixture and got one
(`eval/tests/e2e/stribling-father-1821/`). This is the prose half, deliberately
deferred out of that PR because skill prose goes through
`docs/skill-lifecycle.md` with its own paid eval run and annotation.

**The fixture's debug run confirms the inference is not made unaided.** Given a
guardianship bond naming four minors of a different surname, and the guardian's
marriage to their mother six months earlier, the agent minted the ward as a
person *separate* from the subject, never asked whether the ward **was** the
subject, and instead entertained an invented extra generation to explain a
document role.

What no prose names is the trigger:

> **A man appointed guardian of children bearing a different surname, shortly
> after marrying a woman connected to that surname, is most likely their
> stepfather, and they are her children by a prior marriage. The differing
> surname is the expected pattern, not a conflict.**

The vocabulary already exists — `relationship-accuracy.md:20-27` defines Step /
Foster / "Other guardianship" and says to record without asserting a type when
the type is unknown. And the same infer-a-remarriage move already exists for a
different trigger at `historical-context/references/historical-terminology.md`
(the "In-law" entry, ~:51-53: before the 20th century "in-law" frequently meant
step-relation). This is the missing keying of that move to a guardianship
appointment. **Model the new text on that entry; do not edit it — that file is
`historical-context`'s snapshot and is a different eval slot** (held by #1663).

## Correction 2 — the census caveat conflates "no relationship stated" with "type not stated"

`relationship-accuracy.md:25-27` currently reads:

> A household listing in a census may show a child living with adults without
> clarifying whether the connection is genetic, adoptive, or something else.

Verified on `main` 2026-08-25 (`sed -n '18,30p'` on that file). The sentence is not simply
false after 1880 — it collapses two different facts:

- **Pre-1880 U.S. census: no relationship is stated at all.** The 1850/1860/1870 schedules
  have no relationship column, so the record states only co-residence.
- **1880 onward: a relationship IS stated, but its type still is not.** The schedule adds
  a relationship-to-head column. A column reading "son" is a stated relationship the agent
  must read — and it still does not establish biological vs. step vs. adoptive.

Fix: split the sentence along that line, so pre-1880 reads "nothing stated" and 1880+
reads "stated, type unknown".

**This repo already has the authoritative wording — mirror it, do not invent one.**
`packages/engine/plugin/agents/record-extractor.md:707-724` states the pre-1880 / 1880+
rule in full ("A pre-1880 census yields NO relationship assertions at all"; "The 1880
census introduced an explicit relationship column, so from 1880 on the relationship IS
stated"). **Read that file; do not edit it.** It is a Cowork plugin agent — a senior-only
change per `docs/architecture.md` §5 — and it is inside `record-extraction`'s run-log
snapshot, so touching it buys a second paid run.

**Do not weaken the surrounding rule.** The instruction "when the specific relationship
type is unknown, record the relationship without asserting a type rather than defaulting
to biological" (`:23-25`) must survive verbatim in substance. It is the clause the
rubric's `Evidence grounding` **fail** band depends on ("a relationship written without
meeting the `relationship-accuracy.md` threshold"), and in the simplified tree the "no
type asserted" state is an **omitted** `subtype`
(`docs/specs/simplified-gedcomx-spec.md:174`, open enum `parent_subtype_recommended`) — no
schema change is involved either way.

## Read-only neighbours

Three files carry adjacent doctrine. Read them for wording; **edit none of them** — each
is a different skill's or agent's eval slot.

- `packages/engine/plugin/skills/historical-context/references/historical-terminology.md:50-58`
  — the "In-law" entry, the move to model Correction 1 on. `historical-context`'s slot is
  held by #1663 (PR #1803 open); the sibling entry there is **#1710**.
- `packages/engine/plugin/skills/historical-context/references/historical-broad-context.md:124-131`
  — already teaches that a guardianship appointment is not evidence of orphanhood, keyed
  to inheritance law. Your new text must not contradict it: "most likely their stepfather"
  is a *leading hypothesis*, not a licence to write a Step subtype without proof.
- `packages/engine/plugin/agents/record-extractor.md:707-724` — the 1880 rule (see
  Correction 2). Plugin agent: senior-only, separate snapshot.

Cross-skill duplication is correct here, not a smell — `CLAUDE.md` § "Researcher profile"
records that shared SKILL.md reference loading is unreliable (issue #17741), so shared
reference prose is duplicated rather than linked.

## Cost

Editing any file under `packages/engine/plugin/skills/tree-edit/**` — a `references/` file
included — flips `tree-edit`'s run-log snapshot inactive and arms
`.github/workflows/check-runlogs.yml`, which blocks the PR. Confirmed 2026-08-25: the
newest run log `eval/runlogs/unit/tree-edit/v1_2026-08-24_11-16-35.json` is **active** —
`diff_snapshot_vs_disk` over its 41 tracked paths returns `{}` — so you are the one
flipping it. It shipped one day earlier with the tree-edit deep dive (PR #1841, #1657),
which touched only this file's Couple-facts section and left `:20-27` alone.

- **`make eval-skill SKILL=tree-edit`** — the last run cost **$1.64** and took **2.8 min**
  wall clock (`totals.total_cost_usd` / `wall_clock_ms` in that run log). Re-measure rather
  than quoting this forward; model and test count move it.
- **The annotation pass is the real cost, and it is human.** Rule 3 requires an
  `.ann.json` correction for every dimension of each of the 5 tests in the run log's
  `review_sample`, written **only** through the CRUD UI (`make eval-ui`) — hand-writing
  them is forbidden. With 3 base + 4 rubric dimensions that is ~35 corrections.
- `eval-cosmetic-skip` does **not** apply: these are behavioural corrections.

**Both corrections must land in the same PR** — that is the whole reason they are one
card. If you also add the unit test below, add it **before** the run;
`eval/tests/unit/tree-edit/**` is inside the same snapshot, so adding it after buys a
second run.

**Soft collision — issue #1790** touches `eval/tests/unit/tree-edit/rubric.md` and
`person-merge-stub-into-fs-person.json`, the same snapshot. It is in Backlog so it holds
no slot, and it is a different lane, so this is not one PR — but whichever lands second
pays a second $1.64 run plus a second annotation pass. Name it in your PR body.

## Acceptance

1. Both passages rewritten, mirroring `record-extractor.md:707-724` for the census half
   and `historical-terminology.md:50-58` for the guardianship half.
2. **A sibling unit test, added before the run.** The analogous craft point in
   `historical-context` is pinned by
   `eval/tests/unit/historical-context/guardian-appointed-own-children.json`
   (`ut_historical_context_hbt`, filed off #1618) — a positive test with `judge_context`
   bullets and no scenario. Follow that template for the stepfather-guardian trigger.
   Without it **nothing checks either correction**: `eval/tests/unit/tree-edit/rubric.md`
   has four dimensions (Data preservation, Edit minimality, Merge correctness, Evidence
   grounding) and none of them reads relationship-type craft, so a green 13/13 run says
   only that you did not break anything else.
3. `make eval-skill SKILL=tree-edit` re-taken and graded; `make test-all`.

**The `stribling-father-1821` e2e re-check is optional, and a `false` result does not
condemn the prose.** `docs/architecture.md` §9.4 → "If you're asked to…" names
`make eval-skill` plus the annotation gate as the instrument for a skill-body change; a
live `make e2e-run` is the instrument for routing or cross-skill work. One e2e run costs a
median **$7.35** and p90 **$14.75** across the 135 committed e2e run logs, and the
fixture's own README documents confounders that make n=1 unattributable: the debug run
attached the wrong image (700 instead of 725), never ran an indexed marriage search, and —
per its changelog — **"the failure is the agent's at the identification step"**, minting
the bond's "Ridley Stribling" separately from the subject. Identification is
`person-evidence`'s lane, and **neither this card nor #1710 touches it**.

So: if you do run it and the inference still is not made, record in the PR that the
likeliest cause is the **destination**, not prose-versus-tool — the rule may need to reach
`person-evidence`'s identity-linking references, which is a third card nobody has filed.
Name that in the PR; do not file it and do not expand this card.
`docs/skill-lifecycle.md` §5's writer-tool escalation is the *second* hypothesis here, not
the first: ADR-0011's test is "can this be decided by reading the project documents
alone?", and this inference cannot — it is judgement over record content, which ADR-0003
puts in prose ("prose is for rules that only have to hold within one invocation").

Emruthwill:
Closes #1900.

Two corrections to `packages/engine/plugin/skills/tree-edit/references/relationship-accuracy.md`,
batched into one paid `tree-edit` run as the card requires, plus the sibling unit test — added
**before** the run, since `eval/tests/unit/tree-edit/**` is inside the same snapshot.

## Correction 1 — the guardianship → prior-marriage inference

New `### Guardianship shortly after a remarriage` entry, modeled on the "In-law" entry at
`historical-terminology.md:50-58` (pattern statement + **Research implication**), which was read
but **not edited**.

It carries a second paragraph the card didn't call for, **What to record**, which the measurement
below shows is load-bearing. Without it the agent made the inference and then wrote `subtype Step`
on the bond-plus-marriage pair alone — and asserted a `Biological` edge for the mother on surname
correspondence. That is exactly what the surviving threshold clause forbids, so the entry now says
plainly: write the edge with **no subtype at all**, and don't compensate on the mother's edge either.

Consistent with `historical-broad-context.md:124-131` (read, not edited) — it echoes the
property/inheritance driver and keeps step a *leading hypothesis*, not a licence to write a subtype.

## Correction 2 — the census caveat, split at 1880

`:25-27` split along the line `record-extractor.md:707-724` already draws (read, **not** edited —
plugin agent, senior-only, separate snapshot): pre-1880 states *no relationship at all*; 1880+
states one whose *type* is still unknown. Mirrors that file's "the dividing line is whether the
record states the relationship, not how confident you are" clause.

**Scoped to the US federal census** after review: the split is stated as US-federal throughout, and
England & Wales — stated relationship-to-head since 1851 — is named as the counter-example. The
unqualified earlier wording would have told an agent working an 1861 E&W schedule that the record
states no relationship, when it does. **`record-extractor.md:707-724` has the identical gap and is
NOT fixed here** — different snapshot, senior-only agent; flagged for the reviewer instead.

The clause the rubric's `Evidence grounding` **fail** band cites — "record the relationship without
asserting a type rather than defaulting to biological" — survives **verbatim**. No schema change:
"no type asserted" remains an omitted `subtype`.

The pre-1880 bullet points forward to the existing "When to create relationships" section so it
can't be misread as "write no edge" — that section's indirect-evidence rule is untouched.

**Known tension, pre-existing, NOT resolved here.** That section says a pre-1880 census parent-child
edge "still materializes with a source-ref, at a lower ref quality"; `record-extractor.md:707-718`
says the correct output "is **not** an `indirect` link… it is **no assertion**". Both were read
directly and they do conflict. This PR added the forward-pointer between them, so it is fair to
raise here — but reconciling two skills' doctrine is above a genealogist card touching one
reference file, and record-extractor is a different snapshot. Flagging for a decision rather than
picking a winner silently.

## The unit test is proven to fail

`ut_tree_edit_014` — positive, `judge_context` bullets, no scenario, following
`guardian-appointed-own-children.json`.

Per CLAUDE.md ("a new lint must be proven to fail"), I ran it **both ways**. Not `make gate-skill`
— as a reviewer noted, that tool compares against corrected grades mined from an existing committed
run-log, which by definition does not exist for a test that did not exist yet. It was the manual
toggle, and these are the exact commands:

```bash
# arm A — without the correction
git checkout origin/main -- packages/engine/plugin/skills/tree-edit/references/relationship-accuracy.md
cd eval/harness && uv run python run_tests.py --test ut_tree_edit_014 --runlogs-root <scratch>

# arm B — with it
git checkout HEAD -- packages/engine/plugin/skills/tree-edit/references/relationship-accuracy.md
cd eval/harness && uv run python run_tests.py --test ut_tree_edit_014 --runlogs-root <scratch>
```

The run-logs land outside the repo (`--runlogs-root` to a scratch dir) precisely so they are not
released; anyone can reproduce the pair in ~4 minutes for ~$0.33. Grades:

| Dimension | Without the correction | With it |
|---|---|---|
| Correctness | **1** | **3** |
| Completeness | **1** | **3** |
| Evidence grounding | **1** | **3** |

Without the prose the agent **refuses** the inference — *"asserting 'step' now would promote a
plausible inference to a stated fact, which the evidence doesn't yet support"* — and records a bare
guardianship relationship. With it, the judge notes the stepfather reading is reached
**unprompted**, with the subtype correctly withheld.

**My first draft of the test did not discriminate** — it passed against the pre-edit prose, because
it handed the agent both records side by side and planted the "extra generation" worry. The
committed version buries the marriage to Nancy Watts in a list of already-recorded facts and never
flags the surname correspondence; connecting them is the thing under test.

### Root cause of the pre-edit failure: a gate with no trigger

The file's only doctrine on relationship *type* was a threshold rule — "when the specific
relationship type is unknown, record the relationship without asserting a type." That sentence can
only ever say what **not** to do. Nothing told the agent to make the inference at all, so it
under-called, and cited that very sentence to justify it:

> *"Per your own reference guide: 'When the specific relationship type is unknown, record the
> relationship without asserting a type…' That rule applies equally to 'step' — don't default to it
> just because the surname matches."*

Underneath that is one confusion: it collapses **naming the leading hypothesis in its answer**
(free, writes nothing, is what the researcher needs) with **writing a `Step` subtype into the tree**
(genuinely gated). The pre-edit prose addressed only the second, so the gate on it swallowed the
first — *"asserting 'step' now would promote a plausible inference to a stated fact."* It declined
to **say** the thing because it thought saying it meant **writing** it. That is the over-gating the
rubric's `Evidence grounding` **partial** band already names; the prose simply had no defence
against it here.

It also reproduces the e2e failure in miniature. The unaided run offers, unweighted:

> - Nancy's children from a prior relationship → Ridley's **stepchildren**
> - Orphaned children of Nancy's deceased sibling or other Watts kin → a **guardian-only** connection

— the second being exactly the "invented extra generation" the `stribling-father-1821` debug run
produced.

The two failure modes are mirror images, which is why the fix needs both halves:

| | Result |
|---|---|
| **Gate, no trigger** (pre-edit) | Refuses to infer; bare guardianship; extra generation offered as an equal option |
| **Trigger, no gate** (first draft of this PR) | Infers correctly, then writes `subtype Step` **and** a `Biological` edge for the mother |
| **Both** (committed) | Infers unprompted, states the hypothesis, leaves the subtype empty |

That is what **What to record** buys, and why it is in the entry despite the card not asking for it.

## Two findings the card did not anticipate

1. **The card's "no scenario" instruction doesn't transfer as written.** `_compute_outcome` fails
   any positive test where `spec.skill not in skills_invoked`. Framed the way the
   `historical-context` template reads ("no edits, just tell me"), the model answers
   conversationally in one turn and never fires `tree-edit` — `skills_invoked: []`, a guaranteed
   fail regardless of answer quality. That is why no *writer* skill in the repo has a scenario-null
   test; all four suites that use them are advice skills. Resolved by keeping `scenario: null` but
   making the ask edit-shaped. The test makes **zero** tool calls, so there is no
   `unmatched_tool_call` abort exposure.
2. **The prose needed the explicit "what to record" half**, per the measurement above. Stated as a
   trigger alone, the rule fired and then over-shot.
3. **A trigger-shaped rule does not generalize** (raised in review). The principle now sits in the
   file's existing gate and names three routes — household pattern, guardian's appointment, a
   "ward"/apprenticeship label — because scoped to the one trigger, a structurally identical case
   reproduced the original bug untouched.

## Read-only neighbours — none edited

`record-extractor.md`, `historical-terminology.md`, `historical-broad-context.md` — all three read
for wording only. `git diff --stat` touches two files.

## Soft collision — #1790

Open, `developer` lane, touches `eval/tests/unit/tree-edit/rubric.md` and
`person-merge-stub-into-fs-person.json` — the same snapshot. Different lane, so not one PR, but
whichever lands second pays a second run plus a second annotation pass.

## e2e

The `stribling-father-1821` re-check was optional and was **not** run, per the card: `make
eval-skill` plus the annotation gate is the instrument for a skill-body change, and the fixture's
own README documents confounders (wrong image attached, no indexed marriage search) that make n=1
unattributable.

## The run

`eval/runlogs/unit/tree-edit/v1_2026-08-27_11-42-40.json` — **14/14 pass**, $1.80, 363s wall,
0 transient retries, model `claude-sonnet-4-6`.

- **Rule 1** — exactly one newly-added `v{N}` for this skill.
- **Rule 2** — snapshot **active**: `diff_snapshot_vs_disk` over its 42 tracked paths returns `{}`.
- **Rule 3** — `.ann.json` carries all **27** required
  `(test_id, dimension_source, dimension_name)` triples across the 5 `review_sample` tests
  (`_007`, `_011`, `_012`, `_013`, `_014`), written through `make eval-ui`.
  **Zero judge-vs-human disagreements**, and `ut_tree_edit_014`'s 3/3/3 is human-confirmed on its
  first graded run.

The card quoted $1.64 and 13 tests from the prior run and said to re-measure rather than quote
forward. Re-measured: **$1.80 and 14 tests** — the suite grew by the test this PR adds.

`eval-cosmetic-skip` does **not** apply — these are behavioural corrections.

## Local `make test-all`

6 of 7 suites green (typecheck, JS workspace, control-plane 210, MCP server 112 files, eval app 14,
eval harness 2852).

ESLint fails on this Windows checkout with 325 `Delete ␍` warnings across `apps/electron/**` —
files this PR never touches. Cause is local `core.autocrlf=true` rewriting LF blobs to CRLF on
checkout, against a repo that already normalizes via `.gitattributes`' `* text=auto`. It does not
reproduce on Linux CI. Independent of this card.

Merging `origin/main` was needed en route: the branch predated #1910, so
`tests/tools/write-serialization.test.ts` failed with `EPERM ... symlink`. After the merge that
suite is 4/4.


mercyokum (changes requested):
**Requesting changes**

Solid doctrine fix overall, and the five corroborating claims about untouched neighbor files (`historical-terminology.md`'s In-law entry, `historical-broad-context.md`'s inheritance driver, `record-extractor.md`'s pre-1880/1880+ split, issue #1900's scope) all checked out on direct read. A few things worth fixing before merge.

## 1. `ut_tree_edit_014`'s "eleven months" isn't derivable from its own input

`judge_context` bullet 1 requires the response connect Ridley's marriage to Nancy to the bond "eleven months before the bond." But the input only dates the bond to the year **1821** (no month) against a marriage in **November 1820**. The actual gap could be anywhere from ~2 to ~13 months depending on which month in 1821 the bond was recorded — "eleven months" is one specific value the given input doesn't support deriving. A careful run that correctly hedges to "about a year" rather than stating "eleven months" risks being marked down by a judge reading the bullet literally, even though the reasoning is sound.

## 2. The stepfather reading is stated as settled without ruling out the alternative

The new doctrine says a guardian who recently married a woman "connected to that surname" is "most likely" the children's stepfather, treating "the children are hers by a prior marriage" as the working premise. But the doc never establishes *how* Nancy is "connected to that surname" — if her own maiden name is Watts (never previously married), the four minors could be her siblings, nieces, or nephews, making Ridley an uncle-by-marriage, not a stepfather. The doc's own later line ("her maternity rests on the same surname correspondence") concedes this is unproven, yet the stepfather reading is still stated as the settled leading hypothesis with no doctrine text to help distinguish the two cases.

## 3. The doc's settling-evidence bar is weaker than the paired test's

The doctrine's suggested settling evidence — "children's births under the earlier surname" — is weaker than its own stated bar (surname match alone is explicitly "not evidence") and weaker than the paired test's `judge_context`, which requires records "naming their parents." An agent following the doc's literal wording and citing birth records under the earlier surname as sufficient would be graded against the stricter bullet requiring records that actually name the parents — a pass/fail gap between what the doc tells the agent to look for and what the test rewards.

## 4. 1880 census split never scoped to US records

The claim that pre-1880 census schedules state no relationship at all, and 1880+ schedules state a relationship but not its type, is accurate for the US federal census — but nothing in the file says "US" or "federal" anywhere. Working an 1861 England & Wales census (which has had a stated relationship-to-head since 1851) or a Canadian census, an agent applying this doctrine's "pre-1880 = no relationship stated" rule would wrongly tell a researcher the record states no relationship at all, when it does.

Worth noting: `record-extractor.md:707-724`, which this PR cites as the model for the split, has the identical gap — no jurisdiction qualifier there either. So this isn't a new mistake this PR introduced; it's an existing one it had the chance to catch (the source text was open for citation) and didn't.

## 5. Fix scoped narrowly to one trigger, not the general pattern

The general principle behind this fix — an inference is a hypothesis about type, not evidence of it; write the edge, leave the subtype empty, state the hypothesis in prose — is scoped entirely to "guardian appointed shortly after a remarriage" rather than folded into the file's existing general gate ("record the relationship without asserting a type rather than defaulting to biological"). A structurally identical trigger this section doesn't cover — a widow's remarriage plus a same-surname household child with no guardianship bond, or a "ward"/apprenticeship-indenture label — can reproduce the exact pre-edit bug this PR fixes, because nothing generalizes the rule past this one specific trigger.

## 6. "Proven to fail" cites a process that isn't verifiable, and may not be the one named

The PR says it proved `ut_tree_edit_014` discriminates by running it "both ways on scratch run-logs (gitignored, not released)." No artifact backs the claimed 1/1/1 → 3/3/3 discrimination — scratch run-logs are gitignored by design, so nothing in the repo confirms this actually happened as described, regardless of which command produced them.

Separately: `docs/skill-lifecycle.md` §6's documented instrument for "prove a fix works," `make gate-skill`, structurally can't supply a "without the correction" baseline for a brand-new test (it compares against corrected grades mined from an *existing* committed run-log, which doesn't exist yet for a test that didn't exist yet). The PR body doesn't actually say it used `gate-skill` — it just says "ran it both ways" — so this may be describing a different, legitimate manual toggle (edit skill, run `--test`, revert, run `--test` again) rather than misusing the documented tool. Either way, worth a line naming the actual commands run, since as written neither the claim nor the tool it might refer to can be checked.

## 7. The new pre-1880 bullet points into a pre-existing contradiction it doesn't resolve

The new text says: *"Pre-1880 (1850/1860/1870) — the schedule has no relationship column, so the record states no relationship at all... (see 'When to create relationships' below for what that means for the edge)."*

The section it points to (same file, **not edited by this PR**) says: *"A pre-1880 census parent-child edge is **indirect** evidence (a headship/co-residence inference, not a stated relationship) — it still materializes with a source-ref, at a lower ref quality."*

But `record-extractor.md:707-714` — read, not edited, cited by this same PR as the model for the split — says the opposite in absolute terms: *"A pre-1880 census yields NO relationship assertions at all... write zero parent-child and zero spousal assertions... Even when the gedcomx carries a ParentChild or Couple edge... you do not persist it."*

If record-extraction genuinely writes zero relationship assertions from a pre-1880 census, there's no assertion left for tree-edit's "indirect evidence... materializes with a source-ref" language to ever act on. This tension predates this PR, but this PR is the one that added a forward-pointer sending a reader from the strict version straight into the permissive one without reconciling them. Worth a one-line flag alongside #4, since the same section is already being touched.

## Minor

- **"Hypothesis, not a settled type" is stated twice.** The new "What to record" paragraph ("a hypothesis about the type, not evidence of it") and "Research implication" paragraph ("the leading hypothesis, not a settled type") both open with the same claim. The one genuinely new, actionable content in "Research implication" (look for the earlier marriage, check for an estate driver) gets buried behind a repeated hedge.
- **The stateless test risks an unnecessary tool-call attempt — but not a harness abort.** `ut_tree_edit_014` is the only `scenario: null` positive test in the tree-edit suite, and its action-oriented prompt ("I need to add parent-child relationships... What relationships should I record") plausibly invites the model to attempt a `tree_edit` call despite no project on disk. Checked `mock_mcp.py`: `tree_edit` is a `LIVE_TOOLS` entry, not fixture-dispatched, so this would *not* trigger `unmatched_tool_call` — it would call the real compiled tool and get back a normal `{"ok": false, "errors": [...]}"` response, gradeable, not an abort. The `GRADING NOTE` only defends the zero-call case, not an attempted-and-errored one, which is still worth covering.

Emruthwill:
Thanks — this was a genuinely useful review. Seven of the nine points are fixed in `f864b1c7b`; two I have deliberately not fixed, for a reason I want on the record rather than buried.

**1 — "eleven months" not derivable.** Correct, and worse than a judge-leniency risk: the bond is dated only to 1821, so the true gap is anywhere from ~2 to ~13 months. `judge_context` now carries a bullet that *forbids* requiring any interval and states explicitly that hedging is correct behaviour, not vagueness. Confirmed on the re-measured run — the passing response states no month figure at all.

**2 — stepfather asserted without ruling out the alternative.** The strongest point of the nine, and you are right that "connected to that surname" was doing work it could not support. New paragraph: if the wife's shared surname is a **married** name the step reading leads; if it is her **maiden** name the minors may be her brother's orphans and the guardian an **uncle by marriage**. The bond cannot separate these — her prior marriage or the father's estate can. The paired bullet now makes naming the alternative a *plus*, and treating the two as equally weighted a *fail*. The re-measured passing run cites the distinction and names uncle-by-marriage explicitly.

**3 — doc's settling bar weaker than the test's.** Fixed on both sides to records **naming the children's parents**; the test bullet now says citing births under the Watts surname alone is insufficient, since that correspondence is the very thing in question.

**4 — 1880 split never scoped to US.** Fixed: US-federal throughout, with England & Wales (stated relationship-to-head since 1851) named as the counter-example.

**5 — scoped to one trigger, not the pattern.** Fixed, and this was the right call — the principle now sits in the file's existing gate and names three routes (household pattern, guardian's appointment, a "ward"/apprenticeship label), so the structurally identical cases you list are covered.

**6 — "proven to fail" not verifiable.** Fair. Your inference is right: it was **not** `make gate-skill`, which as you say structurally cannot baseline a test that did not exist yet. It was the manual toggle, and the PR body now names the exact commands, with `--runlogs-root` pointing outside the repo so the logs are unreleased by construction. Reproducible in ~4 minutes for ~$0.33. Re-measured after this rewrite: with the prose 3/3/3 pass, without it fail.

**Minor, both fixed** — the duplicated "hypothesis, not a settled type" hedge is gone, and the GRADING NOTE now covers an attempted-and-errored writer call, not just the zero-call case. Thank you for checking `mock_mcp.py` rather than assuming an abort; that saved a wrong fix.

---

**Not fixed, and I would rather flag than paper over:**

**7 — the pre-1880 contradiction**, and the `record-extractor.md` half of **4**. Both verified by direct read:

- `relationship-accuracy.md`: a pre-1880 edge "still materializes with a source-ref, at a lower ref quality"
- `record-extractor.md:707-718`: the correct output "is **not** an `indirect` link… it is **no assertion**"

Those conflict, and you are right that this PR added the pointer between them. But `record-extractor.md` is a senior-only plugin agent in a **different snapshot**, and reconciling two skills' doctrine is not something a genealogist card touching one reference file should decide on its own authority — whichever way it resolves, someone's extraction behaviour changes. Same reasoning for the missing jurisdiction qualifier there: real, pre-existing, and not mine to land quietly in this diff.

Both are now flagged in the PR body. I have not filed an issue for them — that is the lead's call, and I would rather he decide than have me open a card against a file I cannot edit.

---

**Cost of this round, since it is not free:** the prose edits made the previous run log's snapshot stale (`diff_snapshot_vs_disk` flagged both files), so `v1_2026-08-27_11-42-40` and its 27-correction `.ann.json` were deleted and re-taken — Rule 1 allows only one newly-added run log per skill per PR. The replacement is `v1_2026-08-27_17-44-06`: **14/14, $1.72**, snapshot active. A fresh annotation pass is in progress.

mercyokum (approved):

EdmondOware (approved):
Approving.

You asked for a ruling rather than a silent winner, so: record-extractor is right and `relationship-accuracy.md:107` is the stale half. A pre-1880 census does not state a relationship, so citing that schedule as the source-ref on a parent-child edge misrepresents what the record says. The parenthetical "or a census listing a household" predates the 1880 line you just drew and should be narrowed to an 1880 or later schedule stating relationship to head.

That is worth slightly more than "pre-existing", because your new pre-1880 bullet points forward into that section, so a reader now arrives at the contradiction by following your pointer. Still not blocking: the file sits in tree-edit's snapshot and fixing it here buys a second paid run for a clause that was wrong before this PR touched it. Worth its own card.

Leaving `record-extractor.md` alone was right. Plugin agent, senior only, different snapshot.

The genealogy holds. Both of the points I would have pushed on had they been missing are present: that a maiden name reading makes Ridley an uncle by marriage rather than a stepfather, and that guardianship was routinely granted over a minor's property including to a living parent, so the bond alone is not evidence of orphanhood.

One thing from reading the run itself rather than the grades. `ut_tree_edit_014` passes, but its answer says the most probable reading is that the children are Nancy's and **their father had died**, which is the bond being read as evidence of orphanhood, and your own bullet 6 says must not. The judge gave Evidence grounding a 3 and the annotation agreed with all seven dimensions, so nothing caught it. In fairness the inference is period appropriate, since in 1821 Virginia the father is guardian by nature. But `historical-broad-context.md` already states this conditionally, "without checking whether an estate or inheritance was involved", and the run never checks. Your new prose and your test bullet are both absolute where that existing line is conditional. Worth deciding which form you want before the next run pins it, rather than after.

Issue #1972, opened 2026-08-27
7 validator requests from the conflict-resolution deep dive


Paaboat:
> **Reviewed 2026-08-27 before junior handoff.** No decision needed, but **held in Backlog**:
> every one of the seven is blocked on PR #1973, and three more on PR #1922. Sequencing table
> below.

**IMPORTANT**: Do issue #2358 after PR #2354 merges — it edits `eval/tests/unit/conflict-resolution/rubric.md` in the same `conflict-resolution` snapshot as #2354's `flynn-identity-geographic/research.json`, so one paid `make eval-skill SKILL=conflict-resolution` run can discharge both.

**IMPORTANT — corrected 2026-09-09, after V7 shipped (PR #2344).** V7 shares NO file with
issue #1629 or PR #1931. It landed as a standalone module,
`eval/harness/conflict_verdict_report.py`, and touches `judge_report.py` not at all, so the
"land #1629 first" instruction below is void and #1629 does not block it. V7 also inherits no
wrong-newest bug: it scans every committed log and selects no newest. PR #1931 has merged.
The original note read:

> V7 shares `eval/harness/judge_report.py` with issue #1629 and with open PR #1931. Land
> #1629 first if it is in flight.

**Touches:** eval/harness/validators/test_conflict_resolution.py, packages/engine/mcp-server/src/validation/validator.ts, eval/fixtures/scenarios/flynn-with-birthplace-conflict/research.json, eval/fixtures/scenarios/flynn-multi-conflict/research.json, eval/fixtures/scenarios/flynn-identity-geographic/research.json (V7 shipped instead as eval/harness/conflict_verdict_report.py + a Makefile target + scripts/windows/conflict-verdicts.bat; judge_report.py is NOT touched)

Seven validator requests from the `conflict-resolution` deep dive (#1652). Each supplies the genealogical rule; a developer writes the Python. Full Did/Should/Gap write-up, with the transcript quotes behind each, is in `docs/deep-dives/conflict-resolution-findings-2026-08-27.md` — that is where the reasoning lives; this body is the work list.

**Why filed rather than folded into the dive's PR:** the fix needs a different skill. The dive is a genealogist task and produces the rules; these are Python validators and a `validator.ts` referential check, reviewed by a developer. V5 additionally changes engine validation and will fail three shipped fixtures on landing, which is a developer's call to sequence.

## Sequencing — read before picking one up

[...]

Issue #1902, opened 2026-08-25
The judge's rendered evidence is incomplete: tool responses truncate at 3 items, and before-state never renders conflicts[] — grounding graded on the skill's own testimony


kofiatinka12:
> **Reviewed 2026-08-26 before junior handoff.** No decision needed. Two of the three
> symptom rows are withdrawn, the three-way fix fork is settled from in-repo precedent,
> and the original Acceptance line was not executable. Corrections are inline below.

**Touches:** `eval/harness/harness/judge.py` — `_RESPONSE_ARRAY_SAMPLE` (`:140`), `_summarize_response` (`:194-212`), `_render_tool_calls_with_size_guard` (`:389`) · `eval/harness/harness/orchestrator.py` — `_summarize_before_state` (`:1494`), from #1956

**This is a cross-cutting grading-infrastructure defect, not a skill defect.** It affects the grounding grade of *every* skill whose tools return more than three results; it surfaced during the locality-guide deep dive (#1664, PR #1893) but is not specific to that skill.

## Symptom

> **Corrected 2026-08-26 (task review). Two of the three rows below were withdrawn** — they
> were real truncations but not real misgrades. The confirmed defect is one row, and the
> misgrade is **stochastic**.

The LLM judge marks a **grounded** claim as fabricated when the cited item is past the
3-item sampling cap of its view of the tool response. Confirmed on run
`eval/runlogs/unit/locality-guide/v1_2026-08-25_01-57-51.json`:

| Test | Claim the judge called fabricated | Where it actually is | Verdict |
|---|---|---|---|
| `ut_locality_guide_021` | `Alabama, Church Records, 1803–1950` (Collection **1986396**) | **4th of 4** in `eval/fixtures/mcp/collections-search-alabama.json` | **confirmed** — cost base Correctness 2 + rubric Record availability 2 |

The judge's own rationale is the proof of mechanism: *"The collections_search results for
Alabama only returned 4 collections (marriages, probate, census index, **and one more**)"*
— that "and one more" is the `_full_length: 4` marker with three items rendered.

**Two rows from the first draft are withdrawn** (re-checked 2026-08-26):

- `ut_locality_guide_007` / `California, Mission Records` (**1804372**) — it *is* the 4th of
  4 and it *is* invisible to the judge, but **no dimension in that run flagged it**.
  `Research strategy` scored **3** and credits *"1804372 for Mission records"*. Truncated,
  but not a misgrade.
[...]

PR #1973, opened 2026-08-27
conflict-resolution deep dive: findings, prohibition list, and two grading fixes

Paaboat:
Deep dive on `conflict-resolution` per `docs/skill-deep-dive-guide.md`. Issue #1652 stays open until its eval run is annotated.

63 runs read across all five committed run logs, transcripts before scores. Ten findings, seven converted to validator requests, two grading defects fixed here.

## What changed

| File | |
|---|---|
| `packages/engine/plugin/skills/conflict-resolution/SKILL.md` | the settled Iberian-naming pre-edit, one line |
| `eval/tests/unit/conflict-resolution/three-conflict-prioritization.json` | `judge_context` no longer names the answer |
| `eval/tests/unit/conflict-resolution/rubric.md` | three fail branches the corpus actually exercises |
| `docs/deep-dives/conflict-resolution-findings-2026-08-27.md` | the findings and the seven validator requests |
| `docs/deep-dives/conflict-resolution-prohibition-list.md` | 46 checkable rules, for the next auditor |

## The pre-edit

Carried over from closed issue #1617 per DallanQ's comment on #1652 — a settled correction, not a finding of this dive. The alpha-team review (Clorinda Madsen, Shaunese Luthy) found the absolute "a patronymic mismatch is never a spelling variant" misfires on Hispanic names; PR #1697 scoped the "never" but did not name the pattern, so a García Ramos / Ramos García pair still read as two people. Florence's wording, verbatim:

> Iberian naming — Spanish and Portuguese two-surname systems — follows different rules: the two surnames are reordered or one is dropped, and on emigration the maternal surname often becomes a middle name, so variation there does not carry the same implication.

It rides this dive's eval run because one clause of prose cannot pay for a run of its own. Not `eval-cosmetic-skip` — it is a new instruction meant to change what the agent does.

- [x] The Iberian reorder / drop / maternal-surname-to-middle-name pattern is named in the patronymic item

## The three findings that matter most

All three are quiet passes — scored 3 by the judge **and** confirmed 3 by the human annotator in `v1_2026-08-19_15-24-31.ann.json`.

**F1 — Standard 46 stated, then discarded one field later, in 3 of 3 birthplace resolutions.** `ut_006`'s `independence_analysis` says the two census assertions "must be treated as a partially dependent unit, counted as no more than one strong report". Its `weighing_analysis`, in the same `fields` object, calls them "two independent recording occasions" and invokes Standard 48 rationale 1 — *uncorroborated single item* — against the other side. Having grouped the winning side into one unit, rationale 1 is unavailable to it. Same shape in `ut_008` and `ut_001`. The two fields are graded by two different rubric dimensions and nothing reads the seam.

**F2 — 31 of 37 persisted rationales exceed the ~250-word cap; 16 of those are single-assertion identity conflicts** where the completeness escape cannot apply. Worst is 475 words on one competing assertion. The `Resolution completeness` dimension rewards exactly what the cap penalises, and has scored 3 in 35 of 35 positive runs.

**F3 — a fact conflict written `resolved` while an open identity conflict covers one of its own competing assertions.** `ut_008` does this in 5 of 5 runs; `ut_006` in 3 of 5. Its own response says so after the write: *"C_002 is the most foundational — it underpins whether the census evidence belongs to the right person at all."* `ut_002`, three tests away in the same log, gets it right and quotes the rule.

## What I did not change, and why

**No SKILL.md prose beyond the pre-edit.** Every behavioural finding is against a rule the body already states — and in three cases the skill obeys it in one test and breaks it in another *within the same run log*. Per the guide: "if the rule was there and was ignored, restating it is not the fix."

**No word counts in `rubric.md`.** Length is arithmetic; it belongs in a validator (V2), not in a dimension an LLM has to eyeball — that is how a dimension ends up scoring 3 forever.

**No fixture edits.** F6's three inconsistent scenarios are shared with other skills' suites; correcting them belongs with V5's guard, which will name them.

**No base-rubric or global-judge-prompt edit.** F9 is global, so the wording is proposed on #1972 and the lead calls it.

**No file on issue #1823's Touches list**, per the #1652 sequencing ruling — that ruling is recorded as comments on #1823 and #1852.

## Deferred work

**Issue #1972** — the seven validator requests (`developer`, `nothing-checks`). Filed rather than folded in because the fix needs a different skill: this is a genealogist task producing the rules; those are Python validators plus a `validator.ts` referential check, reviewed by a developer. V5 additionally changes engine validation and will fail three shipped fixtures on landing, which is a developer's call to sequence.

Also commented on existing issues rather than filing duplicates: #1902 (its `_summarize_before_state` never renders `conflicts[]` — the mechanism behind F3), #1112 (`validation-protocol.md` ships in this skill, is unreachable, and contradicts the body).

## Verification

- `eval/harness` unit suite: **2872 passed, 3 skipped**
- `packages/engine/mcp-server` packaging suite: **594 passed**
- `rubric.md` re-parsed through `harness.rubric.parse_rubric`, and `three-conflict-prioritization.json` through `json.load`, after editing
- `make eval-skill SKILL=conflict-resolution` — running; run log and annotation to follow before merge

The rubric edit needed the parser checked: `parse_rubric` captures a dimension's `description` as everything *before* the first bullet and matches each bullet as a single line, so prose placed after the bullets reaches the judge only via `rubric.raw`. Both the raw text and the parsed `fail_criteria` now carry each new rule.


florencemashipei (changes requested):
Good dive, and thank you for the restraint in "What I did not change" — declining to restate rules the body already carries is the right call, and the reasoning for keeping word counts out of the rubric is sound. The parser note is a genuinely useful catch. Two things to fix before I can approve.

## 1. The two grading fixes did not change any grade

`Evidence weighing` and `Resolution completeness` both scored 3 on every test in the run that shipped with the new rubric. Your own findings doc says `Resolution completeness` has scored 3 in 35 of 35 positive runs — after the fix it is 3 in 7 of 7.

That would be fine if the behaviour had stopped. It has not. `ut_conflict_resolution_008` in this same run says its two census assertions are "partially independent", then weighs them as "two partially independent records" corroborating each other and invokes Rationale 1 against the other side. Your new branch names both of those moves as failures. It scored 3.

The same write calls the 1850 informant "almost certainly Thomas Flynn or his wife, present at the birth". The fixture records that informant as "Unknown household member (likely Thomas Flynn or wife)", quality `indeterminate`. Your second new block says that fails the dimension. It scored 3.

`ut_008` and `ut_006` are not in the review sample. On `ut_001`, which is sampled and which F1 also names, the annotation confirms both dimensions at 3 with no comment.

So the calibration pass currently blesses the scores the findings doc calls defective. I do not think you should reword the rubric to chase this — that is the trap the doc itself warns about. Either put `ut_008` in the targeted slot and re-annotate so the gap is on the record, or add a short paragraph saying the branches did not move the corpus and what you think that means. `Source independence analysis` did move to 2s, so the edits are not inert in general, which is exactly why the other two standing still needs saying out loud.

## 2. The findings doc cites a run log this PR deletes

`v1_2026-08-18_10-20-02.json` is cited three times, including F2's headline example. The retention cap prunes it when this dive's run lands, so committing the deletion was correct — but the 475-word example becomes uncheckable the moment this merges. I confirmed the number is real, and no surviving run log carries it. The highest that survives is 460 words.

Please make the doc stand on its own:

> You wrote:
> `- ` + "`ut_conflict_resolution_002` / `c_002`, `v1_2026-08-18_10-20-02` — **475 words** on `competing_assertion_ids: [\"a_001\"]`. Nearly double the cap."
>
> Change it to:
> `- ` + "`ut_conflict_resolution_002` / `c_002`, `v1_2026-08-18_10-20-02` (pruned by the 5-candidate retention cap when this dive's run landed) — **475 words** on `competing_assertion_ids: [\"a_001\"]`. Nearly double the cap. Highest surviving instance: `ut_conflict_resolution_001` / `v1_2026-08-19_15-24-31`, 460 words."

Same treatment for the second citation lower down, and the corpus list at the top should note the file is no longer on disk.

## Everything else looks right

The Iberian clause matches the wording agreed on #1652 exactly. The `judge_context` rewrite removes the leaked answer without replacing it with a different one. Both suites reproduce (packaging 594; harness 2873 passed / 3 skipped — one more than your 2872, `main` moved). CI is green on the current head.

I did not re-run the eval, and I did not check F4 through F10 or the 46 prohibition rules — I verified F1, F2 and F3.

Paaboat:
Both fixed. Thank you for the count — you were right and I was wrong about it, which changed one number in the doc beyond the two fixes.

## 1. Calibration note added — and your 7-of-7 corrected my 6-of-6

Took the second option: a paragraph saying the branches did not move the corpus, rather than re-annotating. Reasoning for not taking the first, recorded in the note itself: reaching `ut_008` means spending a paid run to confirm a negative result the paragraph already states, and the annotation would then no longer be the harness's own sample (seed 2995). If you would rather have it on the record as an annotation than as prose, say so and I will run it — the objection is cost and sample integrity, not the value of the evidence.

**On the count: I had written "6 of 6" in two places. The suite has seven positive tests and your 7-of-7 is right.** Recomputed from the logs rather than reworded, and pushed as its own commit (`5f44decd`): 7 of 7 in this run, and **42 of 42** for each dimension across every run ever measured for this skill — 35 before this PR, 7 after. The surviving five logs total 35 on their own; the 42 counts the pruned one.

**One place I did not follow your wording, and you should check whether you agree.** You describe `ut_008` as making both moves the `Evidence weighing` branch names. Two of the three land exactly and the note says so:

- Resolving `c_001` on `a_002` with `c_002` and `c_003` open — the `Resolution completeness` branch verbatim. Scored 3.
- "his parents", "firsthand knowledge", an informant "present at the birth", where `a_002.information_quality` is `indeterminate` and the parentage is unproved `q_001` — the second `Evidence weighing` block verbatim. Scored 3.

The independence/weighing contradiction is only a **partial** match on `ut_008`, and I wrote it that way. This run says "partially independent" and carries that qualifier *into* the weighing ("two partially independent records") rather than flipping to "independent" the way the three committed-corpus instances do. It still reaches for Rationale 1 on a side its own analysis calls partly dependent, which the branch forbids — so your point stands, but it is a softer instance than F1's originals and I did not want the note to overclaim on a test nobody has annotated. If you read it as a clean match, that is a one-word change and I will make it.

I have not reworded either dimension, per your steer and the doc's own warning. The note ends on what the run cannot settle: whether the two branches are mis-scoped, or whether the judge attributes loosely enough that which dimension carries a rule matters less than that it is stated somewhere the judge reads. `Source independence analysis` moving to 2 on `ut_001` — confirmed by the annotation with no score change — is what makes both readings live.

## 2. Pruned run log — all three sites, plus recovery

Applied your wording verbatim at F2's bullet and at V2's violation example, and the corpus list at the top now says the file is off disk with the `git show <pre-prune-commit>:…` line to recover it.

I verified 460 is the true maximum among survivors before writing it: next highest are 436 (`ut_006` / `v1_2026-08-19_15-24-31`), 429 (`ut_008` / this run), 426 (`ut_008` / `v1_2026-08-18_13-28-10`).

## On your suite count

You got harness **2873**; I got 2872 before the `main` merge landed on this branch and have not re-run since. Yours is the current head, so take 2873.

## Scope you flagged

Noted that you verified F1–F3 and did not check F4–F10 or the 46 prohibition rules. F4 is the one I would most want a second pair of eyes on if you have appetite later — it is the `judge_context` bullet that was telling the judge `c_001` was the right answer, and it is the reason `ut_006` had never been marked down before this run. The rest are lower stakes: F5–F8 are evidenced from transcripts quoted inline, and F9 I have already corrected once against this run's own contradicting data.

florencemashipei (approved):
Approved. Both fixes are in, and I checked them rather than taking the summary.

**The count.** I recounted every grading across all five surviving logs plus the pruned one recovered from git: 42 of 42 threes on `Evidence weighing` and 42 of 42 on `Resolution completeness`, no exceptions. Your recomputed figure is right. Correcting your own number upward against yourself, from the logs rather than by rewording, is the right instinct and it is the rule this repo has learned the hard way.

**The calibration note** does the job. It records the negative result, names `ut_008` with both exact matches, says plainly that the gap is judge-only because the test fell outside seed 2995, and leaves the mis-scoped-versus-loosely-attributed question open instead of settling it on one run. Not rewording either dimension was the right call.

**Your two questions:**

1. **Prose over annotation — agreed, do not spend the run.** Your reasoning is sound on both counts: a paid run to confirm a negative the paragraph already states, and a sample that would no longer be the harness's own. Committing that spend would be Dallan's call in any case.

2. **Keep "partial".** You are right and my wording was looser than yours. I said the branch prohibits both moves by name; that is true of the *consequences* — counting the pair as multiple corroborating items, reaching for Rationale 1 — but `ut_008` never flips from grouped to independent the way F1's three originals do. It carries "partially independent" through. Yours is the more accurate description. No change.

**The pruned-log fixes** are complete across all three sites, and I verified 460 is the true survivor maximum before you wrote it.

I also confirmed #1972 is open and labeled `developer` + `nothing-checks` as the body says, and that #1902 and #1112 got comments rather than duplicate issues.

**Scope of this approval:** I verified F1, F2 and F3, the suites, the counts and the deferral trail. I did not check F4 through F10 or the 46 prohibition rules. Happy to take the F4 pass you flagged — the `judge_context` bullet that was handing the judge the answer — as separate work; it needs no eval spend.

Still needs a second approval to merge.

EdmondOware (approved):
Verified and approving. I re-derived the findings rather than reading them, and the two that carry the dive hold up.

**F3 reproduces exactly**, `ut_008` in 5 of 5 runs and `ut_006` in 3 of 5. Worth recording how, because it took me four attempts: the dependency runs through the **source**, not the assertion id. Joining `c_001`'s `competing_assertion_ids` against the open identity conflict's returns nothing, since the sets do not intersect. Joining on `source_id` reproduces your figures precisely, because `c_002` asks about src_001's Patrick and a_002 comes from src_001. Your `judge_context` addition states that linkage correctly, which is what let me find the right key.

**F2's population is exact at 37.** My own count puts 32 over the cap rather than 31, and the worst at 460 words rather than 475, on `ut_conflict_resolution_001`. Almost certainly a tokenising difference at the margin, and "~250" is explicitly approximate, so this is not a correction. Worth naming the counting method in the finding so the next person reconciles rather than re-derives.

**The two grading fixes are the right lane and correctly made.** The `judge_context` edit removes an answer key and replaces it with the criteria the body actually names, which is the same defect measured on #1954. And I checked the parser concern you raised rather than taking it: `parse_rubric` puts the new Evidence weighing prose in `description` and each new rule in `fail_criteria`, so all three reach the judge on both paths.

**The Iberian pre-edit is correct** and correctly placed inside the patronymic carve out.

One finding, older than this PR but adjacent to what you edited.

**Fifteen lines below your change, the inherited surname rule states Portuguese naming backwards.** Line 355 reads "Spanish and Portuguese compound surnames (paternal first, then maternal: children of Juan García and María Ramos are García Ramos)". That is the Spanish convention, and the example is Spanish, but the sentence claims it covers both. Portuguese runs the other way: the mother's surname comes first and the father's last, so a child of João Silva and Maria Costa is Costa Silva.

It matters because that rule disqualifies a candidate parent "on that ground alone". Applied to a Portuguese family with the Spanish order assumed, it disqualifies the correct father, which is a false negative in parentage and the more damaging direction to fail in. Your new text also says Iberian surnames "are reordered", which is true but reads as clerical variation rather than a systematic difference between the two countries, so the two passages now sit awkwardly beside each other.

`SKILL.md` is snapshot tracked, so correcting it costs another paid run and I am not asking you to buy one. Please card it before merge, with the fix being either to correct the parenthetical or to drop "Portuguese" from that sentence and let the new Iberian text carry it. The dive mentions Portugal nowhere, so without a card it stays invisible.

One smaller thing to reconcile rather than fix: my F3 join also flags `ut_002` in 2 of 5 runs, on its own scenario `flynn-multi-conflict` rather than `flynn-identity-geographic`. That sits oddly with the finding calling `ut_002` the test that gets it right and quotes the rule. Different scenario, so it may be a legitimate case my join does not distinguish, and I have not read those transcripts. Worth a look before someone quotes `ut_002` as the clean counter-example.

Run, annotation and gate all check out here: 13 tests at 10 pass and 3 partial, 21 corrections covering every sampled dimension with no uncommented non pass, and all runlog rules satisfied.

Issue #1529, opened 2026-08-10
Lint: ban the "missing ARK ⇒ degenerate same_person score" premise repo-wide


Gennecis:
> **Blocker cleared — verified 2026-09-02 (`/fill-ready` stale-cross-reference sweep).** The header below says *"Blocked on issue
> #1429 — do not start until it merges."* **Issue #1429 closed COMPLETED 2026-08-31** and
> the work is on `main` (`cf0802d8`, "Score identity where a persona is reachable, not
> where a field is non-null"). This card is startable; the instruction not to do #1429's
> half still stands. Verify with `git log origin/main --oneline --grep='#1429'`.

> **Reviewed 2026-08-20 before junior handoff.** The premise was false and the ask
> is re-scoped; the original follows under "Original issue" and should be read as
> history only. **Blocked on issue #1429 — do not start until it merges, and do
> not do its half.**

**Touches:** packages/engine/mcp-server/tests/packaging/, docs/adrs/ADR-0009-refuted-agent-design-claims.md, eval/harness/e2e/orchestrator.py, eval/harness/tests/unit/test_e2e_orchestrator.py, docs/plan/research-performance-2026-07-27.md

## What this issue is now

A regression lint that keeps the refuted "missing FamilySearch ARK ⇒ degenerate
`same_person` score ⇒ treat as no score available" claim from coming back, plus the four
sites issue #1429 leaves behind.

**Two earlier framings are dead. Do not act on either.**

- **"Issue #1272 removed the premise from seven sites."** #1272 is an issue, not a PR; it
  closed `not planned` on 2026-08-11, as did #1439 on 2026-08-07. Nothing was removed from
  anywhere, so `git revert` on "the #1272 commit" is **not executable**, and "seven" was
  never verified against the tree. The premise is live in **six** files today.
- **The 2026-08-14 comment's step 1** — "correct the skill prose, and the fixture and unit
  test that encode it." Issue #1429 was rewritten 2026-08-17 and now owns exactly that.

## Do not touch these files in this PR

`packages/engine/plugin/skills/person-evidence/SKILL.md`,
`docs/specs/research-append-tool-spec.md`, `docs/specs/rank-search-matches-tool-spec.md`.
All three are on issue #1429's own checklist. Editing the skill body here also flips that
skill's run-log snapshot inactive, and `check_runlogs.py` then blocks the PR until a fresh
[...]

PR #1977, opened 2026-08-27
Score identity where a persona is reachable, not where a field is non-null (#1429)

Gennecis:
## Summary

Start here: `eval/harness/harness/skill_invocation.py`, `_persona_reachable` — one predicate, four call sites.

Two §8 detectors required a `same_person` score on every `person_evidence` link to a brand-new tree person and never looked at the evidence's provenance, so they also required it where no record persona exists. This narrows them — and the free replay and corpus report that read them — onto whether a persona is **reachable**, and fixes the skill prose that taught the agent to skip a far larger set.

**The finding worth the reviewer's time: prose alone did not work, and code did.** A full 22-test run against the corrected retrieval recipe made **zero** `record_read` calls and scored exactly the cases it scored before. Both tests exercising the new lanes passed while demonstrating nothing, because their `judge_context` waived the score. `research_append`'s `match_score` warning turned out to be the cause: it gated on `confidence === "confident"`, knew nothing about provenance, and offered the escape *"if no comparable FamilySearch persona exists to score against, leave match_score null"* — which the agent took, citing a null `record_persona_id`. Re-pointed onto reachability at any confidence, silent where nothing is scoreable, and naming the retrieval route, it produced `record_read → same_person → re-append` on the very next run. That is ADR-0011's question answered with a measurement rather than an argument.

**It is a catch, not a prevention** — a distinction the genealogist's annotation drew and this section originally blurred. `_024`'s *first* append still wrote `confident` with a null score, reasoning that `record_persona_id` is null on the assertion — precisely the belief this PR removes. The warning then fired and the agent corrected itself with `record_read → same_person → update`. So the agent still holds the wrong prior; the tool catches it after the write. That strengthens the ADR-0011 case rather than weakening it — a hard rejection would be the prevention — and it is why the rejection question below is worth putting to the lead.

**Two declared deviations from the issue**, both toward the safe direction and both evidenced:

- **`fulltext_search` is exempt even with a retained sidecar**, against one row of the 2026-08-17 table. `FulltextResult` carries no GedcomX, and all 259 FTS-sourced assertions in the corpus use a `3:1:` image-entry ARK `record_read` cannot open. `ut_person_evidence_011` already graded it this way.
- **Unknown provenance keeps flagging** rather than failing open. The issue's supporting measurement is an artifact: no `assertions` op in the corpus carries an `id` at all (0 of 2,963), so that join can only ever return 0. The shape the ledger *can* express happens — 17 calls, 9 runs, 66 ops.

**Annotated and gated:** `v1_2026-08-27_16-24-49.ann.json` is committed and `check_runlogs` reports all blocking rules satisfied.

## Review guidance

- `_persona_reachable` is the whole rule. It is mirrored in `research-append.ts::personaReachable` and `validators/test_person_evidence.py::_persona_reachable` — three copies by necessity (engine, harness, validators cannot import each other); each names the others.
- **It decides "what did the run retain", not "what could be fetched".** 48 of the 306 exempted links carry a `1:1:` ARK `record_read` might open. They stay exempt: the detector cannot verify offline that the fetch resolves, and flagging on an unproven capability is the error corrected in the FTS lane. Revisit trigger and its real cost (~11 unit tests would need new fixtures) are in `guardrail-enforcement-spec.md` §4.
- The new deterministic validator is **coarse on purpose** — "was `same_person` called at all in a run that owed a score", not "was this pairing scored". The existing per-pair check keys on `(record_persona_id, person_id)` and structurally cannot cover lanes where that field is null.
- `eval/tests/unit/person-evidence/rubric.md` gated its `same_person` dimension on a non-null `record_persona_id`. Left alone, the paid run would have graded the corrected skill by the criterion it replaces.
- Figures are branch-scoped to this checkout. PRs #1925 and #1931 touch these modules; whichever lands first, the other rebases. Issue #1430 edits the same `pretool_hook` block.

## Test plan

- `make test-all` green. Harness 2886 passed / 2 skipped; engine 2718 passed; `research-append.test.ts` 215; `ruff` clean.
- **Proven to fail, three ways.** Disable the narrowing → **12** tests red across the suite (9 of them in `test_skill_invocation.py`). Apply the *refuted* null-field rule → **10** different ones red (8 in that file). Revert the engine gate to `confidence` → 7 engine tests red. The three new reason-text assertions (`record_read`, `results_ref`, `record_role`) are all absent from the un-rewritten string.
- **A fourth claim is withdrawn.** This said "drop the success gate from the detector-diff replica → its drift test red". The mutation I actually ran replaced `scored = same_person_scored_ids(...)` with `set()`, which removes scoring altogether — a much bigger change than dropping the `is_error` gate, and not what the sentence described. Dropping that gate alone **cannot** red the test: of 91 `same_person` calls in the corpus, 13 carry an `is_error` key and **0** are `True`, so the arm is unexercised by today's data. The drift test is not vacuous overall — dropping the seed-tree scoping reds it at once — but that one arm is untested and is now stated as such. The same latent hole in the new validator (an errored call satisfying it) is closed.
- `make e2e-guardrail-shadow REPLAY=1 SINCE=all`: **788 → 699** link-gaps, **120 → 115** of 149 runs, `runs_without_provenance` 0.
- `make e2e-corpus SINCE=all RECOMPUTE=1`: `same_person` arm **444** (487 un-narrowed), new `unscoreable by design (links)` **306**, `regressed` names the 5 cleared runs. Compliance unchanged at 38 fail / 121 not_checked / **0 pass**.
- `make e2e-detector-diff DETECTOR=same-person-provenance SINCE=all`: 15 runs diverge — 5 stop firing, 10 shrink.
- **Four paid runs** (~$29 with gates). The committed log `v1_2026-08-27_16-24-49` is 19 pass / 3 partial / 0 fail / 0 aborted, with `n7v` making exactly two `same_person` calls, `_024` doing `record_read → same_person`, and no validator failures.
- **Two shared-fixture warnings, warn-only.** `flynn-record-matching`'s scenario and its `log_004.json` are embedded in the latest `search-full-text` and `search-records` run logs, which no longer match after the `3:1:` ARK correction. `check_runlogs` rule 2f is warn-only by the #1094 ruling, so nothing blocks, but both skills owe a re-run whenever they are next touched.
- **Annotated.** `v1_2026-08-27_16-24-49.ann.json` covers seven tests — the five sampled plus `_024` and `n7v` volunteered — with 53 corrected cells. `check_runlogs` now reports all blocking rules satisfied.

## Follow-on issues

**`_024`'s thirteen-year contradiction is DEFERRED, not dropped — [#1950](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1950).** Stating it here because the committed corpus currently encodes that a `confident` identity link is acceptable across an unreconciled thirteen-year contradiction the skill itself flagged, on a record fetched to answer who the mother was. `ut_person_evidence_024` links a Kilrush christening of 12 March 1858 to a tree person born `~1845`; an Irish Catholic baptism follows birth within days, so they cannot be the same birth. The expected behaviour — link at `speculative`, log an identity conflict, name the Kilrush 1840s search that would resolve it — plus the rubric line and the genealogist's "do not fix the fixture" are all on that card. Deferred because `judge_context` is not in `_COSMETIC_TEST_FIELDS`, so the edit flips the snapshot and costs a fifth $7.51 run, and because it is a **different guardrail** (chronology, not provenance) whose better form is the validator the same card requests.

**The two stale skill snapshots are filed — [#2008](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/2008).** `flynn-record-matching`'s scenario and `log_004.json` are embedded in the latest `search-full-text` and `search-records` run logs, which no longer match after the `3:1:` ARK correction. Warn-only under #1094, so nothing blocks, but the warning outlives this PR.

**A graded behaviour now has no instruction — added to [#1950](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1950), stated here.** Deleting the "degenerate score on an unresolvable id" bullet was right — its stated cause was the refuted premise — but it left nothing in the 770-line body teaching the surviving behaviour (a near-zero score on a *thin* subject must not veto a strong correlation): `uninformative|degenerate|thin subject|too sparse` now greps to **0**. `ut_person_evidence_019`'s `judge_context` still grades it, and passes partly because the judge credits the agent for reading a `note` field in `same-person-flynn-degenerate.json` — and `SamePersonResult` has **no `note` field** (0 occurrences in `types/same-person.ts` and `tools/same-person.ts`), so the test leans on an explanation production never sends. The `note` is inherited, but this PR's deletion is what makes it load-bearing. Both fixes are snapshot-tracked, so they cost a fifth run and ride the same card.

Nothing else filed. Four things raised on existing issues instead:

- **Issue #1529** — its raise-then-drop instruction is honoured (`ut_person_evidence_019` and its fixture corrected). ADR-0009 is left to it as its `Touches:` list claims — note that **its rev.3 row also carries the refuted premise**, not only constraint 5. Two further sites are deferred on eval-slot cost, not scope: `search-records/SKILL.md`'s "thin/unresolvable subject" and `research/SKILL.md`'s demoted paraphrase, each of which would buy another paid run.
- **Issue #1430** edits the same `pretool_hook` block; rebase rather than revert.
- **A hard rejection in `research_append`** is the obvious next step and is deliberately **not** taken here: `match_score` is caller-fabricable, so a rejection buys a number rather than a call unless something persists the `same_person` result (ADR-0009 constraint 2), and #1006 chose presence-only warn. This PR is the first time that argument has a measurement behind it — needs @DallanQ.
- `flynn-couple-marriage` carries the same wrong FTS ARK but is used only by a `tree-edit` test, so it is left for whoever next runs that skill. And `same_person`'s tool description says "`primaryId1` = its `primaryId`", right for the matched persona and wrong for a second party.


Gennecis:
## Annotating the `person-evidence` eval run — what we need from you

**Setup**

```sh
git fetch origin
git checkout 1429-narrow-same-person-provenance-check
make eval-ui          # opens the grading UI on localhost:3000
```

Grade the run **`v1_2026-08-27_16-24-49`**. It picks five tests for you: `_001`, `_002`, `_010`, `_013`, `_019`. When you're done, commit the `.ann.json` file it writes and push. CI will stay red until that lands — that's expected, it's what it's waiting for.

**What changed, in plain terms**

The skill used to score an identity match with `same_person` only when the record came from a FamilySearch search that had kept its results file. That turned out to be wrong: the tool doesn't need that file — it just needs the record and the tree person. So the skill now also scores when it can re-open the record it already read, and it correctly *doesn't* score when there's genuinely nothing to compare against, like a photograph you transcribed or a full-text hit with no indexed person in it.

We also removed an old instruction saying a low score on a locally-created person is a meaningless artefact. It isn't — a low score there usually means the tree person is too thin to compare, which is real information, not noise.

**What we actually want from you**

Judge whether the skill did good genealogy. Not whether it followed our new wording.

The dimension worth your closest attention is **Score discipline**. Ask yourself, on each test:

- When a record and a tree person could genuinely be compared, did it run the comparison — or did it assert an identity it never checked?
- When there was nothing to compare against, did it say so and reason from the evidence instead — or did it invent a number?
- Did a score ever *override* the correlation? A high score should never carry a link past a real conflict, and a low score should never kill a match the names, dates and places clearly support.

`_001` and `_019` came back as partial. Those two will repay the closest reading.

**Please disagree with us**

The judge is a model and it is wrong a fair bit. Your corrections are the point of this step, not a formality.

- If the judge marked something a pass and you think the genealogy is poor, mark it down and say why.
- **If the judge marked something down and you think the skill was right, mark it up.** That matters just as much.
- If a test is asking for the wrong thing in the first place, say that in the comment — that's a finding about our test, and we'd rather hear it now.

Every dimension needs a decision, and anything that isn't a confirmed pass needs a written comment. Do not 

We are not looking for confirmation that the change worked. If it made the skill worse anywhere, that's exactly what we need to know before this merges.

EdmondOware:
## Annotation is in

`v1_2026-08-27_16-24-49.ann.json` is committed and pushed on `43d70851`. `check_runlogs` passes and all checks on the PR are green.

56 corrections: the five sampled tests plus `_024` and `n7v`. I added those two because the sampler picked five tests that are all either old lane (a non null `record_persona_id`) or no persona reachable, so none of the five actually exercises what this PR changes. Without them the annotation would have certified a run in which nobody looked at the new behaviour.

Thank you for the brief. Asking for judgement on the genealogy rather than on the new wording is what made this worth doing properly, and the run itself is careful work. `n7v` is the best evidence in the file that the change is right, and I want to say why before I get to the problems.

## What the change buys, from the run

`n7v` scores both personas against their own tree candidate, `F1` to `I1` at 0.58 and `M1` to `I2` at 0.89, with no reuse of the first party's persona for the second, which is the failure the rubric's fail band names. The tiering then tracks the evidence rather than the number. The father is only `Thomas [surname not indexed]`, with Flynn merely implied by the groom's surname, so it stays `probable` at 0.58 and the rationale says so. The mother is an exact match on the distinctive surname Sheehy, so it goes `confident` at 0.89. That is a score informing a decision instead of making it, and it is what the narrowing was for.

## Corrections to the judge

Five, all in the file with reasoning.

**`_001` and `_019`, Multi person awareness, 2 to 3.** The dimension misfired twice in five sampled tests, in opposite directions. On `_001` it deducted for asking before writing during a review request, which the rubric's pass text names as satisfying the dimension, and its own rationale concedes the instruction was to review. On `_019` it called `a_001` relationship implying. `a_001` is `fact_type: "name"`. The judge says exactly that itself on `_010`, which links the same assertion in the same scenario and scores 3. The entry it asks for would have asserted that the person named Patrick Flynn in role `child_1` is Thomas Flynn.

**`_001`, three cells held at 2 with the reasoning replaced.** The judge called the proposed downgrade of `pe_005` correct with a muddled reasoning path. It is the other way round. Per `research-schema-spec.md` 5.7, `person_evidence` is the identity resolution step and its `confidence` measures the link, so `confident` on `pe_005` is right and the downgrade is the error. The skill's own text says the identity is "far better established by a_011 and by the cross record consistency with pe_001 and pe_004", then downgrades the field that measures that. The informant discount it wanted to apply does have a home: the missing `a_013` to `I2` link that the skill correctly found. That link asks whether the Thomas Flynn named as father is our `I2`, and it rests on a name plus a son in law's hearsay.

**`_024`, three cells 3 to 2.** Detail below.

## `_024`, and why I would keep the fixture as it is

The christening in `record-read-flynn-baptism-bapt-k1q.json` is 12 March 1858. The tree in `flynn-baptism-names-mother` carries `I1` born `~1845`. An Irish Catholic baptism follows birth within days, so those cannot be the same birth. The skill noticed, wrote that the record "may be the wrong estimate or a different person", and recorded the link `confident` anyway, citing 0.71 as confirming a strong match. The judge saw the gap and cleared it as a reconciliation note.

Its stated tie breaker, the father Thomas Flynn appearing in both the baptism and the tree, cannot settle it. A Thomas Flynn with a son baptised in 1858 is what a younger sibling of the man born ~1845 looks like, or a forename reused after an infant death. The overlap fits both readings equally. `n7v` applies the right standard to the same problem in this same run and `_024` does not.

This matters because the record was fetched to answer q_001, who was Patrick's mother. If the identification is wrong, the mother is wrong.

**My recommendation is to keep the thirteen year gap rather than correct the date.** A christening well off a tree estimate, with the same father's name in both, is the most common way a wrong person enters an Irish pedigree, and this is the only test in the corpus that touches it. Correcting the date would delete that coverage. What the test is missing is not a better fixture, it is a statement of what the skill is supposed to do when it meets the gap.

The disciplined answer is not to pick a side. The tree's `~1845` is an estimate, almost certainly from a self reported census age. The baptism is an original register entry with an exact date. If they are the same person the register wins and the tree is wrong; if they are not, the register belongs to another child. The evidence in hand cannot decide, so the skill should link at `speculative`, log an identity conflict, and name the search that would resolve it: other Flynn baptisms to Thomas in the Kilrush register in the 1840s. If a Patrick was baptised around 1845, the 1858 entry is a different child. If none appears and the 1845 estimate traces to a single self reported age, the register is probably right and the tree needs correcting.

Minting Bridget Doyle and deferring the ParentChild edge, which the skill did correctly, stay correct either way. She is the mother of the child in this record regardless; what is unsettled is which child that is. So the rest of the test's purpose, links only and defer the relationship, is untouched by keeping the gap.

## What I would like from you

**1. Keep the gap and say what it is for.** Add to `_024`'s `judge_context`:

> The christening date of 12 March 1858 is deliberately thirteen years from I1's estimated birth of ~1845. This is the point of the test, not an oversight. An Irish Catholic baptism follows birth within days, so the two cannot describe the same birth. The prompt asserts the identification, but the skill must not treat a user asserted identity or a moderate same_person score as settling it.

> Expected: link at speculative rather than confident, log an identity conflict naming the question, and state what would resolve it. Recording the gap as a note while keeping confident is a fail on Confidence calibration.

> The shared father Thomas Flynn is not a tie breaker. A Thomas Flynn with a son baptised in 1858 is equally consistent with a younger sibling of the man born ~1845 or a forename reused after an infant death.

> Minting Bridget Doyle and deferring the ParentChild edge remain correct and should still pass.

And to the rubric's Confidence calibration:

> A chronological contradiction the skill itself identifies caps confidence at speculative until it is resolved. Noting a discrepancy is not resolving it.

**2. Record that the gate is a catch, not a prevention.** `_024` reached the score only on a second pass. Its first append wrote `match_score: null` with the reason "record_persona_id is null on this assertion", which is precisely the escape this PR closes, and it corrected itself afterwards. That strengthens the ADR-0011 finding rather than weakening it, but the PR body reads as though the behaviour was prevented.

**3. Add circular scoring to the null reasons.** `n7v` declined to score `I3` because `I3` had just been minted from that same persona. That is correct, and it is a fourth legitimate reason for a null `match_score` that `_persona_reachable` does not name, alongside full text, image and external site. Worth a line in `guardrail-enforcement-spec.md` and in the rubric's pass text.

**4. Two skills owe re runs.** `check_runlogs` warns that `flynn-record-matching/research.json` and `results/log_004.json` are shared fixtures embedded in the latest `search-full-text` and `search-records` logs, which no longer match. Warn only today, and not mentioned in the test plan.

**5. Optional, and the version I would rather have.** Prose in a `judge_context` will drift the way the prose in this PR drifted. The computable form of ask 1 is a validator: for any `person_evidence` entry marked `confident`, fail when the record's christening or birth date and the tree person's birth date differ by more than a threshold. Deterministic, no judge involved, and the same move this PR already proved works better than wording. It would sit alongside the eight validator requests in #1950 and the coarse `same_person` validator this PR adds.

## One note on reading the run

Score discipline was 3 on all 22 tests before my corrections. `n7v` and `_024` show it is genuinely exercised, so it is not a dead dimension, but 22 for 22 should not be read as evidence on its own.

Also worth flagging as an eval lane item: the Multi person dimension misfired twice in five sampled tests, once against the rubric's explicit carve out and once contradicting the judge's own reading of the same assertion elsewhere in the same run.

Gennecis:
Thank you — this is the most useful thing that has happened to this PR. Two of your findings are corrections to claims I made about my own work, and one is a defect I introduced and did not notice. Taking them in turn, with what I did and what I deliberately did not.

## Verified before acting

- **`_024` is a catch, not a prevention — you are right and I had it wrong.** Confirmed in the run log: the *first* `research_append` wrote `confident` with `match_score: null`, and only then came `record_read → same_person → update(0.71)`. So the agent still holds the refuted prior; the warning corrects it after the write. **PR body fixed** — that section now says so explicitly and draws the consequence you draw, that this strengthens the case for a hard rejection rather than weakening it.
- **The 13-year gap is mine.** I wrote `record-read-flynn-baptism-bapt-k1q.json` with a christening of 12 March 1858 against `I1` born `~1845`, in a fixture built to prove something else entirely. I did not spot it. Your argument for keeping it is better than fixing it, and I have not touched it.
- **The two shared-fixture warnings reproduce**, exactly as you describe, and were missing from the test plan. **Added.**

## Done now

**Ask 2 — catch not prevention.** PR body corrected, as above.

**Ask 3 — circular scoring as a fourth null reason.** Added to `guardrail-enforcement-spec.md` §4, with the reason it is *named* rather than encoded: `_persona_reachable` cannot see it, because by write time the stub is an ordinary tree person, indistinguishable from one that has been there for months. It cites `n7v` declining the call in this run having made it in the two preceding ones. I have left the rubric half out — see below.

**Ask 4 — the two skills.** Now in the test plan, with the #1094 warn-only ruling named so a reviewer knows why nothing blocks.

## Held back, deliberately — asks 1 and 5

Filed together on #1950 with your reasoning quoted, rather than folded in. Three reasons:

1. **They are one change, not two.** Your ask 1 (judge_context + rubric) and ask 5 (the validator) are the same guardrail — chronological contradiction caps confidence — in two forms. Shipping the prose without the validator ships precisely the form you say will drift.
2. **You prefer the validator, and this PR is the evidence for your preference.** A corrected retrieval recipe changed nothing across a full 22-test run until the same rule moved into the writer tool. Doing the prose half here would be doing the thing this PR just measured as ineffective.
3. **Cost.** `judge_context` and `rubric.md` are both snapshot-tracked, so ask 1 voids `v1_2026-08-27_16-24-49` and your annotation with it — a fifth paid run and a second annotation from you. The validator has the same effect for a different reason: it would fail `_024` in the committed log, leaving the log inconsistent with the validator set. Either way it is a run, and it is a run for a guardrail that is not what this PR is about.

Your "do not fix the fixture" instruction is in the issue in bold, along with the expected behaviour — `speculative`, log an identity conflict, name the Kilrush 1840s search — so whoever picks it up cannot helpfully delete the coverage.

## Two things I have not actioned, and want your steer

**The Multi-person dimension misfiring.** You found it wrong twice in five sampled tests, in opposite directions, once against the rubric's own carve-out and once contradicting the judge's reading of the same assertion elsewhere in the same run. That is a rubric/judge-quality finding rather than anything this PR touches. `/audit-rubric person-evidence` is the tool for it and it is free — worth running, but it is your lane and I did not want to pre-empt it.

**`_001`'s three cells.** You held them at 2 and replaced the reasoning: `confident` on `pe_005` is right and the downgrade is the error, with the informant discount belonging to the missing `a_013 → I2` link instead. I have not tried to encode that anywhere. If it should become a test or a rubric line rather than living only in this annotation, say so and it goes on #1950 with the rest.

## One correction to my own reading

You are right that Score discipline being 3 on 22 of 22 before your corrections is not evidence on its own. I had cited the dimension being *exercised* as showing the re-gating took; that shows it is not dead, which is a weaker claim, and I should have said so.

EdmondOware (changes requested):
I reproduced every measured figure in this PR against the local corpus rather than reading them, and they all hold exactly. That is rare enough to say first.

| Claim | Reproduced |
|---|---|
| detector-diff: 15 runs diverge, 5 stop firing, 10 shrink | exact |
| shadow replay: 699 link gaps, 115 of 149 runs | exact |
| corpus: `same_person` arm 444, `unscoreable by design (links)` 306 | exact |
| compliance 38 fail / 121 not_checked | exact |
| FTS deviation: every FTS sourced assertion on a `3:1:` ARK | exact, 259 of 259 |
| unknown provenance: no `assertions` op carries an `id` | 0, confirmed |

Two of those readers corroborate each other independently: the detector diff's five runs that stop firing are precisely the shadow report's 120 to 115 drop.

The three copies of `_persona_reachable` agree. I extracted each with the docstring stripped and compared branch by branch: harness, `research-append.ts::personaReachable`, and `validators/test_person_evidence.py` are semantically identical. `personaRoute`'s `results_ref` branch has no tool check, which looked like drift until I traced the call site: line 353 returns early unless reachability is already true, so `tool === "record_search"` is established before line 357 can run. Unreachable, not a bug.

Proven to fail is understated rather than overstated. Disabling the narrowing turns twelve tests red, of which exactly nine are in `test_skill_invocation.py`, the number you cite. The three downstream readers each carry their own guard as well.

And the catch-not-prevention paragraph is exactly what was needed, including the point that a hard rejection would be the prevention and that this is why the rejection question belongs with the lead. Full suite here: 2907 passed, 3 skipped, one Windows only `jq` failure unrelated to this branch. Gate satisfied.

Three things.

**1. The circular scoring evidence clause is wrong, and the truth is better for you.** `guardrail-enforcement-spec.md` now says `ut_person_evidence_n7v` declined the circular score "having made that exact circular call in the two preceding runs." It did not. In `v1_2026-08-24_18-17-08` and `v1_2026-08-24_22-05-46` it made **zero** `same_person` tool calls, and wrote every `match_score` null, including for `I1` and `I2`, which are pre existing tree persons rather than stubs minted in the run. Both runs failed. The reason it gave was not circularity:

> "Stub has no FamilySearch ARK; same_person scoring skipped."

That is the excuse `eval/fixtures/mcp/same-person-flynn-degenerate.json` records as refuted, citing `dev/probe-same-person-local-id.ts`, and says "is never an excuse to skip the call". So the real before and after is a bigger win than the one claimed: the skill went from skipping every score on a refuted excuse to scoring both scoreable pairings, `F1` to `I1` at 0.58 and `M1` to `I2` at 0.89, and declining only the one that is genuinely circular. Please restate the clause to what the runs show. It is a documentation file, so it costs nothing.

**2. Say what happens to `_024`.** This is the one ask from the annotation still open, and I now know it costs more than I did when I raised it: `judge_context` is not in `_COSMETIC_TEST_FIELDS`, so editing it flips the snapshot, and a person-evidence run is $7.51. Deferring it to a card is entirely reasonable. Dropping it silently is not, because the committed corpus currently encodes that a `confident` identity link is acceptable across an unreconciled thirteen year contradiction that the skill itself flagged, on a record fetched to answer who the mother was. Someone will quote that. Either do it with the run or file the card, and name which in the body.

**3. File a card for the two skills owing re-runs.** `check_runlogs` still warns that `flynn-record-matching/research.json` and `results/log_004.json` are shared fixtures embedded in the latest `search-full-text` and `search-records` logs, which no longer match. Warn only today, so not a blocker, but it is not in the test plan and it will outlive this PR.

Nothing else. The reachability change itself is sound, well measured, and the strongest evidence for it is `n7v`, which scores two distinct personas against their own tree candidates and lets the distinctive surname carry `confident` while the given name only match stays `probable`.

Gennecis:
All three done. Ask 1 was a real defect and you are right that the truth is better — thank you for checking the citation against the corpus instead of reading it, because that is exactly how it got past me.

## 1. The circular-scoring evidence clause — corrected

You are right, and my error has a specific shape worth naming: **I cited two runs that are not in the repo.** The circular `G1 → I3` call did happen, in `v1_2026-08-27_12-36-32` and `v1_2026-08-27_13-35-47` — my runs 2 and 3. `check_runlogs` rule 1 allows one new run log per skill, so I pruned both and kept only the fourth. The claim then pointed at evidence nobody could check, and anyone who tried found the opposite. In a PR whose whole subject is a doc asserting something the code does not do, that is the same defect in miniature.

Confirmed what you found, in both runs: **zero** `same_person` calls, every `match_score` null, both failed, and the stated reason verbatim —

> `same_person` scoring skipped — I1 and I2 are local stubs with no FamilySearch ARK, so any score would be degenerate (uninformative).

Which is the refuted claim itself, and `I1`/`I2` are pre-existing scenario persons, so even "stub" was wrong.

`guardrail-enforcement-spec.md` now states the real before/after: from skipping every score on a refuted excuse, to scoring both reachable pairings at 0.58 and 0.89, letting the distinctive surname carry `confident` while the given-name-only match stays `probable`, and declining exactly the one that is circular. Pushed in `dffb8708`.

## 2. `_024` — deferred, and now named in the body

Deferred to [#1950](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1950), and the PR body now says so in the Follow-on section rather than leaving it to evaporate — including the thing you are protecting against, stated plainly: the committed corpus currently encodes that a `confident` identity link is acceptable across an unreconciled thirteen-year contradiction the skill itself flagged, on a record fetched to answer who the mother was.

The card carries the expected behaviour (`speculative`, log an identity conflict, name the Kilrush 1840s search), the rubric line, and the annotator's "do not fix the fixture" in bold. Two reasons for deferring rather than buying the run: `judge_context` is not in `_COSMETIC_TEST_FIELDS`, so it costs a fifth $7.51 run; and it is a different guardrail — chronology, not provenance — whose better form is the validator the same card requests, which this PR's own evidence argues for over prose.

## 3. The two stale snapshots — filed

[#2008](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/2008), and referenced from the PR body. It names both run logs, quotes the warning, records that rule 2f is warn-only under #1094 so nobody re-litigates whether it blocks, and says **do not correct the ARK back** — the `3:1:` form is what full-text search returns, and the old `1:1:` was the trap this PR measured.

## On your `personaRoute` trace

You are right that the `results_ref` branch is unreachable without `tool === "record_search"` already established, and right that it reads like drift. I have left it as it is — the early return is the guard, and adding a redundant tool check would suggest the branch is reachable on its own. Worth knowing that you had to trace it, though; if a second reader does the same I would rather add a one-line comment than have people re-derive it.

Full suite here is green apart from `windows-wrappers`, which trips on `.claude/worktrees/pr1622/scripts/setup-feedback-case.bat` — a stale detached worktree from another PR sitting inside the repo, untracked, absent from this diff and invisible to CI. Unrelated, but worth someone pruning.

EdmondOware (approved):
Verified and approving. I re-derived all three rather than reading them.

The evidence clause is now right, and it matches what I measured independently: zero `same_person` calls in both committed predecessors, every `match_score` null including for `I1` and `I2`, both runs failed, and the after stated correctly as scoring both reachable pairings and declining only the circular one. I checked the quoted reason against the run log and it is verbatim. Calling out that `I1` and `I2` are pre existing scenario persons rather than minted stubs is sharper than my own note was, because it means the skill's stated reason was wrong twice over.

A correction of mine, which you should not have had to work around. My review said `_024` was "not done, no card". It was done, at 08:46, ninety minutes before I posted, and the comment has never been edited. I searched newly created issues and issue bodies and missed a comment on an existing card. That is a bad method and I will not repeat it. The #1950 comment is thorough, and the two deferrals are now named in the body with their costs, which is more than I asked for.

The reasoning for deferring `_024` is right and I would not have argued it as well. It is a different guardrail, chronology rather than provenance, and its better form is the validator the same card requests, which this PR's own measurement supports over prose.

On `personaRoute`, agreed, leave it. If a second reader traces it the comment is worth adding then.

The worktree lint you hit is already filed as #2006, developer plus nothing checks, with the offending output and the `:236` line. Your `pr1622` sighting is useful corroboration that it is not one person's checkout.

Gate satisfied here, 2918 passed with only the known Windows `jq` failure, branch level with main, and the two remaining warnings are the ones #2008 now owns.

T-FEH (changes requested):
Reviewed on the merged head (`f2a8500c` — your `f14a5b25` plus a clean merge of main). This is careful work and the body is unusually precise — every corpus figure I or my second pass could re-derive reproduces to the digit. Two things block, and both are about evidence rather than design.

## What holds up

**The three-copy mirror is genuinely in step** — the main structural risk here, since engine, harness and validators cannot import each other. I differential-tested the two executable copies across ten inputs and they agree on every one:

```
record_persona_id / record_read / record_search+results_ref        True   True
record_search without results_ref / fulltext_search / image_transcribe  False  False
missing log_entry_id / entry not found / assertion None / list-valued id  True   True
disagreements: 0
```

All three fail *safe* — unknown provenance returns True, so an assertion cannot shed the requirement by omitting a field. The TypeScript copy matches the same five decision points, with `.find()` giving the same default on a miss.

**The eval gate is green** ("All runlog rules satisfied"), so the body's "Not mergeable yet: rule 3 blocks" line is stale — the later paragraph is the accurate one.

**The annotation is the strongest I have audited recently.** Seven tests — the five sampled plus `_024` and `n7v` volunteered — with five substantive disagreements, three of them marking *down* on `_024` (Correctness, Confidence calibration, Score discipline all 3→2). The genealogist is being harder than the judge and citing the rubric text to do it. "53 corrected cells" reconciles exactly: 56 corrections, 3 of them N/A.

And the figures: **259** FTS assertions all carrying `3:1:` ARKs, **306** exempted links, **48** with a `1:1:` ARK split **29/19**, **15** diverging runs at **5/10** — all reproduce.

## 1. The fourth failure proof does not reproduce

Three of the four hold. The fourth cannot:

> Drop the success gate from the detector-diff replica → its drift test red.

It stays green, and the reason is in the corpus:

```
same_person calls in the e2e corpus:  91
  carrying an is_error key:           13
  with is_error == True:               0
```

A gate that skips errored calls cannot change a verdict when no call is errored, so `test_same_person_provenance_replica_never_drifts_from_live_across_the_corpus` is blind to that arm. Your own docstring at `skill_invocation.py:586` says as much — *"That guard only began to bind once #1255/#1289 made the e2e orchestrator populate `is_error`."*

The test is not vacuous overall — dropping the seed-tree scoping reds it immediately — so this is one arm, not the whole check. But CLAUDE.md's rule is that a new check is *watched* failing, and this arm was reported as watched failing on a mutation that cannot fail. Either construct a case with an errored call and pin it, or say plainly that this arm is unexercised by today's corpus and why.

The same hole is in the new validator, latently: `test_person_evidence.py:703`'s `if any("same_person" in (tc.get("tool") or "") …)` is satisfied by an errored call. One line closes both.

## 2. A graded behaviour now has no instruction, and passes on a field production never sends

The PR deletes the "Degenerate `same_person` score on an unresolvable id" bullet from `person-evidence/SKILL.md` and replaces it with nothing:

```
- **Degenerate `same_person` score on an unresolvable id:** When the tree
-   not let the degenerate number drop a match the identifiers otherwise
-   support. Note in the rationale that the score was uninformative and why.
```

A grep for `uninformative|degenerate|thin subject|too sparse` across the whole 770-line body now returns **0**. Deleting it was defensible — its stated cause was the refuted premise. But `ut_person_evidence_019`'s `judge_context`, which this PR also edits, still requires:

> person-evidence must recognize the score as **uninformative** (near-zero from a thin subject) rather than reading it as a genuine low-similarity signal.

And `_019` scores 3/3 on Score discipline because the judge credits the skill for reading a note *"explicitly noted by the tool itself"* — which comes from `eval/fixtures/mcp/same-person-flynn-degenerate.json`:

```
fixture response keys: [apiTitle, candidateArk, confidence, matched, note, queryArk, score, updated]
```

`SamePersonResult` has no `note` field — neither `src/types/same-person.ts` nor `src/tools/same-person.ts` mentions one. So the test passes on an explanation production never supplies, while the body text that would have taught the behaviour is gone.

The `note` field is inherited; the deletion that makes it load-bearing is introduced here. Fixing the SKILL.md side costs a paid run, so it may genuinely belong on #1950's card — but it should not merge unstated, because as it stands the suite records that this behaviour is graded and nothing teaches it.

## Smaller

- **"9 harness tests red" and "8 different ones" are single-file counts.** Whole-suite they are **12** and **10**. The direction and the disjointness both hold; the numbers as written do not reproduce.
- **There is no `record_role` assertion** among the three new reason-text checks — the third is a `record_persona_id` (`/p_293161675629/`), and un-rewriting the string reds 5, not 3.
- Suite totals have moved (harness 2886→2958, engine 2718→2738) but that is main, not this PR — the body does declare figures branch-scoped.
- The new validator flags a link to a pre-existing seed person, where the harness detector does not. Harmless in today's suite, but it is a scope difference between two things presented as a mirrored pair.

Requesting changes on the two majors. The narrowing itself I have no argument with — it is the best-evidenced change I have reviewed on this repo, and the ADR-0011 measurement in the body is exactly the form that rule asks for.

Gennecis:
Both majors are right, and the first is the more useful kind of catch — you found a claim that didn't hold rather than a bug.

## 1. The fourth failure proof — withdrawn

You are right and the reason is worse than "it doesn't reproduce": **the mutation I ran was not the one I described.** I replaced `scored = same_person_scored_ids(tool_calls)` with `set()`, which removes scoring altogether. It went red, I wrote down the sentence I had *intended* to test, and shipped it. Dropping the `is_error` gate alone cannot red anything on this corpus, exactly as you say — 91 calls, 13 carrying the key, 0 `True`.

The body now states three proofs, not four, and names that arm as unexercised by today's data with the counts. I did not construct a synthetic errored-call case: it would pin the gate in `test_skill_invocation.py` where it already has direct coverage, and the corpus-wide drift test would still be blind to it, so it would read as coverage of the replay that it isn't.

**The latent hole in the validator is closed** (`94cdb91d`) — `and tc.get("is_error") is not True`, with the arm documented as unexercised rather than claimed as tested. Thank you for spotting that the same line covered both; it did.

## 2. The untaught behaviour and the phantom `note` — confirmed, and carded

Verified: `note` appears **0** times in `packages/engine/mcp-server/src/types/same-person.ts` and **0** times in `src/tools/same-person.ts`. `SamePersonResult` is `matched, confidence?, score, queryArk, candidateArk, apiTitle, updated`. And the body greps to 0 for `uninformative|degenerate|thin subject|too sparse`.

So the position is exactly as you put it: the suite records that this behaviour is graded, nothing teaches it, and the test clears on an explanation the real tool never sends. The deletion was right — the stated cause was refuted — and it took the surviving rule with it.

Added to [#1950](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1950) rather than fixed here, because both files are snapshot-tracked and it should ride the run that card already needs rather than buy a second one. The card names the correct cause to restore (thin content, not the ARK, consistent with `rank-search-matches-tool-spec.md`'s sibling guardrail) and puts the `note` decision explicitly: drop it and expect `_019`'s Score discipline to move, or keep it and say in `judge_context` that it stands in for the skill's own reasoning. It is also **stated in the PR body**, so it does not merge unremarked.

## On the smaller items

**The `record_role` assertion does exist** — `eval/harness/tests/unit/test_e2e_orchestrator.py:871`:

```
869:    assert "record_read" in reason
870:    assert "results_ref" in reason
871:    assert "record_role" in reason
```

I think the `p_293161675629` you found is in `packages/engine/mcp-server/tests/tools/research-append.test.ts` — the engine-side test for `personaRoute`'s persona-id branch, a different file from the three reason-text checks the body cites. Worth confirming I have read you correctly rather than assuming.

**The counts are corrected.** 9 and 8 were single-file; the body now says 12 and 10 whole-suite, with the single-file numbers in parentheses.

**The validator scope difference is now named in its docstring** rather than left for a reader to notice — mirrored predicate, not mirrored population. That is a fair thing to have flagged about a pair I presented as mirrored.

**The stale "Not mergeable yet" line is fixed** — the gate has been green since the annotation landed.

## One thing needing a senior, not a change

`runlogs` is red on `main`'s merge, not on anything here: `312d3c67` (#2045) deleted `person-evidence/references/validation-protocol.md`, which is in this run log's snapshot. `person-evidence/SKILL.md` names `evidence-standards.md`, `person-profiles.md` and `correlation-techniques.md` and **never named** `validation-protocol.md`, so no session could load it — #2045's own message says "Behaviour is unchanged by construction". A re-run would cost $7.51 to account for deleting a file that could never be read, so this is an `eval-cosmetic-skip` case. Needs a senior to apply it, and after my last push since the bypass expires on each one — that push is `94cdb91d`.

T-FEH (approved):
Second pass, on `46c23855` (your `94cdb91d` plus a clean merge of main). Harness 2991 passed, 3 skipped. Both majors are resolved, and the first one is answered better than I asked for.

## The withdrawn claim is the right way to handle a measurement that did not hold

I asked you to either pin the errored case or say plainly that the arm is unexercised. You did the second, and did not soften it:

> **A fourth claim is withdrawn.** … The mutation I actually ran replaced `scored = same_person_scored_ids(...)` with `set()`, which removes scoring altogether — a much bigger change than dropping the `is_error` gate, and not what the sentence described. Dropping that gate alone **cannot** red the test: of 91 `same_person` calls in the corpus, 13 carry an `is_error` key and **0** are `True`.

Naming the mutation you actually ran, rather than quietly restating the conclusion, is what the repo's re-measure rule is for. The heading is now "Proven to fail, **three** ways", the counts are corrected to 12 and 10, and the same latent hole is closed in the validator — with the honesty carried into the code rather than left in the PR body:

> Latent rather than live today — no committed `same_person` call carries `is_error: true` — so this arm is unexercised by the corpus and is here for the run that first errors.

That comment will outlive the PR description, which is where it belongs.

## The deleted doctrine is disclosed, and I checked the card actually carries it

My ask was that it not merge unstated. It is now stated in full — the grep result, that `_019` still grades it, that `SamePersonResult` has no `note` field "so the test leans on an explanation production never sends", and that this PR's deletion is what makes the inherited fixture load-bearing.

I did not take the "added to #1950" part on trust, because a deferral that is claimed but not recorded is worse than one that is simply deferred — the PR body becomes hard to find once merged. It is genuinely on the card:

```
#1950 body + comments:  uninformative ×2   thin subject ×2   SamePersonResult ×1   _019 ×2   degenerate ×2
```

Both deferrals are there, the `_024` chronological one as an explicitly-labelled ninth request. Worth knowing for whoever picks it up: #1950's title is "8 validator requests from the search-external-sites deep dive", so two person-evidence items now live under a search-external-sites heading. Not worth renaming, but it will not be found by title.

## Everything I verified last round still stands

The three-copy `_persona_reachable` mirror agrees on all ten inputs including malformed ones, all three fail safe, and the new commit adds the one thing that was missing from the pair — a written account of why the validator's population differs from the detector's:

> The two are a mirrored *predicate*, not a mirrored population — the harness detector answers "was a new identity asserted unscored", this answers "did a run that owed a score make the call".

That closes the scope question rather than leaving it to be read as drift.

## One cosmetic leftover

The body still names the three new reason-text assertions as `record_read`, `results_ref`, `record_role`. The `record_role` assertions in `research-append.test.ts` are the pre-existing negative-evidence ones; the third new assertion is the `record_persona_id` literal. Cosmetic, and the only thing left from my list.

## Approving

The narrowing was already the best-evidenced change I had reviewed here; what this round adds is a worked example of retracting a claim precisely instead of defending it. Every corpus figure still reproduces, and the two deferrals are tracked where someone will act on them.

PR #1978, opened 2026-08-27
search-full-text: triage on returned fields, reconcile the spec, fix the fixture strip (#1826)

### LINKED ISSUE #1826: search-full-text triages on textDocument, which the tool strips from every staged result

francis-2008-happy:
## Summary
Closes #1826 (Normal tier). Four defects in the search-full-text surface, found checking a senior genealogist's notes against the code:
1. The skill triaged on `textDocument`, which the tool strips from every staged result — so within this skill that question could never be answered.
2. `highlightTerms` was returned but taught nowhere.
3. `score` was in the spec but sent by no one and mapped by nothing.
4. `count` default drifted (spec 20 vs. code 5).

## Step 1 measurement (the plan's gating live call, `dev/try-fulltext-search.ts "+Patrick +Flynn"`, 2026-08-27)
- **(a) Relevance/score present?** **No.** The raw upstream entry carries only `content`, `id`, `sourceUrl`, `collectionId`, `collectionTitle` — no score/relevance/rank field, top-level or in `content`. → deleted `score` from the spec rather than adding code to map a nonexistent field (defect 3).
- **(b) `content.highlightTexts` shape?** **Bare terms** (`["Patrick", "Flynn", "Flynn Patrick"]`), not marked-up fragments. → the eval fixture (`fulltext-search-flynn-witnesses.json`) was right; `fulltext-search.test.ts:83`'s marked-up assumption is unrealistic. Because they're bare, the "is the name in the right context?" question cannot be answered from a staged result — see Didn't change.

## Review guidance
**Start here:** `docs/specs/fulltext-search-tool-spec.md` (the Output schema + new Result staging section) checked against `src/types/fulltext-search.ts` and `mapEntry` — the issue's acceptance is spec↔implementation agreement on every field. Then `SKILL.md` step 6 (triage rewritten onto returned stubs) and the new Result triage dimension in `rubric.md`.

**Unsure about:** nothing outstanding. One deliberate deferral below.

## Plan
**Didn't change (deferred to the lead, per the issue's stop-and-ask gate):** because `highlightTerms` are bare terms and `textDocument` is stripped, a staged result cannot answer "genuine mention vs. false positive." Answering it in-result would need a **new bounded-excerpt output field = a tool-contract change**, out of this card's scope. Took the conservative path instead: the rewritten triage routes that judgment to verifying against the original image (the skill already mandates that), adding no tool field. `count` default: changed the **spec** to 5, not the code — the shipped schema/`buildUrl` were already 5; raising the real default is a separate measured decision.

**Acceptance check:** `make eval-skill SKILL=search-full-text` on a run made **after** `8f1f75c8`, from this PR's worktree. Done: `v1_2026-08-31_19-59-51` (11 pass / 2 fail of 13), annotated, `runlogs` green. `ut_010` — the test this PR exists to fix — **passes and is graded**, a second graded pass on an independent run. The committed 08-27 run predates both `ut_010` remedies (its SKILL.md and witnesses-fixture hashes are `4252b4a0`'s, which carries neither), so it could not have shown them working. Also: `make engine-test` green; spec and implementation agree on every input and output field (verified field-by-field).

**Two tests did not pass, neither from this diff.** `ut_011` has no verdict at all — `JudgeError`, connection error after 3 attempts, with the skill itself running 9 turns cleanly. `ut_002` is a real graded failure: the skill built `+Flyn* +Patrick`, which no fixture answers, so the call returned `fixture_not_found` (Tool Arguments 1, Query construction 2). That is query construction, not triage or log fidelity, and is out of this PR's scope — worth looking at as fixture coverage rather than patching here.

## Test plan
- [x] `make test-all` — passes except the 9 pre-existing environmental failures in `tests/unit/test_setup_feedback_case.py` (`unzip` not installed on my box), unrelated to this diff and green in CI. 2865 passed.
- [x] `make eval-skill SKILL=search-full-text` re-run; run log + `.ann.json` committed (review sample annotated via the CRUD UI); the 2 harness-pruned run logs staged.
- [ ] ~routing negatives~ — N/A: no skill `description`/DO-NOT clause changed.

**Run result:** 13 tests, 11 pass / 1 partial / 1 fail, no aborts. Neither non-pass is caused by this diff, but my first description of `ut_010` was wrong and is corrected here (chesworthrm's item 2). It did **not** make 0 tool calls. It ran the search — `{keywords: "+Flynn +witness", count: 50}`, correctly with no filters, having just said "no place or date filters on the first call" — and then reported a "Schuylkill County, Pennsylvania filter" it never sent, logging `recordPlace1`/`recordPlace2` lifted from the response echo. So it is a log-fidelity failure, not a skipped search, and the remedy is item 1 (the rule is now inlined in SKILL.md step 7) plus item 4 (the two fixtures that echoed keys their own args never sent). `ut_001`'s partial is Tool Arguments=2 (a query-args mismatch; its Result triage scored 3) and is explainable jitter.

**This run log is stale and will be replaced.** #2045 deleted `search-full-text/references/{research-log-protocol,validation-protocol}.md`, both of which every run log for this skill records in its snapshot, so the committed log no longer validates. A fresh `make eval-skill SKILL=search-full-text` plus annotation is queued once the review items are confirmed settled — running earlier would only throw the run away.

## The fixture defect (why the bug survived)
The mock injects `staged` when `projectPath` is set (`mock_mcp.py:455`) but never mirrors the producer's inline `textDocument` strip, so every fulltext fixture handed the agent a `textDocument` production removes. This PR originally stripped `textDocument` from all 9 `fulltext-search-*.json`. **That is no longer in the diff and should not be re-added:** PR #2045 (now this branch's merge-base) made the mock call the compiled `compactStagedFulltextSearch`, so the fixture keeps the full payload and the mock strips the response copy — production's exact shape, and not defeatable by a tenth fixture. The 08-30 merge restored main's fixtures on purpose.

## Follow-on issues
**Folded in / filed instead:** Folded in — none deferred except the lead-gated tool-field question above (recorded, not filed, pending the lead's call). Filed instead — none.

## For review
`docs/specs/**`, `packages/engine/mcp-server/src/tools/fulltext-search.ts`, `eval/fixtures/mcp/fulltext-search-*.json`, `packages/engine/plugin/skills/search-full-text/**`, `eval/tests/unit/search-full-text/**`. The `.ts`/fixture/skill/rubric paths route to **both** senior-developers and senior-genealogists via CODEOWNERS; the spec `.md` matches no CODEOWNERS rule, so a senior eye on it is by request.


chesworthrm (commented):
Reviewed at `832c5579` (full pass, branch now level with main; all 13 checks green). I re-ran the suites on the head itself: engine vitest 2726, `tsc` clean, harness pytest 2874/2 skipped, server 210/1 skipped, `turbo test typecheck` 8/8.

The core of this is solid work. I checked the spec against the implementation field by field and they agree in both directions — 17 input properties across the spec, `FulltextSearchInput` and the advertised schema, and every output field matches `src/types/fulltext-search.ts`. I also independently confirmed the claim your new comment makes: a staged fulltext transcript really is unreachable. `readStagedResults` hands back rows carrying `id` with no `recordId` and no `gedcomx`, `record-read.ts` matches on `recordId` and requires `gedcomx`, and `rank_search_matches` builds an explicit allow-list that never includes `textDocument`. The comment you replaced was wrong and the new one is right.

Four things to change before this merges.

---

**1. The log-fidelity rule is in a file this skill never opens.**

`references/research-log-protocol.md` says "only record a filter that was actually sent to the tool call it documents — even when the filter matches the locality under research." That is exactly the rule `ut_010` breaks.

SKILL.md never tells the agent to read that file. It names the other four references directly — `online-search-literacy.md`, `search-strategies.md`, `query-syntax.md`, `transcription-quirks.md` — and this one nowhere, by path or by name. `search-records/SKILL.md` does point at its own copy. So the rule that was supposed to close this defect landed somewhere nothing loads, which is a plausible reason the validator keeps firing: it failed in 2 of the 6 committed runs with the identical error.

In step 7:

You wrote:
```
**Every search gets a log entry — no exceptions.** Call
`research_log_append` once per search:
```

Change it to:
```
**Every search gets a log entry — no exceptions.** Call
`research_log_append` once per search. Field-level guidance:
`references/research-log-protocol.md` — record only a filter the call
actually sent.
```

---

**2. The PR body's description of `ut_010` is wrong, and it is the reason the failure got waved through.**

You wrote: "`ut_010` made 0 tool calls (skill didn't execute the search → judge skipped; a post-search rewrite + fixture data can't cause that)".

The run log records four MCP tool calls, including `fulltext_search({keywords: "+Flynn +witness", count: 50})`, which returned the 1878 deed — the transcript reports the hit and then calls `research_log_append`. The judge skipped because the harness only judges when validators pass, and `test_log_query_traces_to_fulltext_search_call` failed.

Your "4/4 in the prior committed logs" is right for the four 08-25 logs. But the 08-24 log this PR deletes shows the same validator failing on the same entry with the same two filters. So it recurs rather than being fresh noise.

To be clear: the failure is not caused by your diff. The skill said "no place or date filters on the first call", correctly sent none, then reported a "Schuylkill County, Pennsylvania filter" and logged `recordPlace1`/`recordPlace2` anyway. Fixing item 1 is the likely remedy. Please just correct the body so the next reader is not told the search never ran.

---

**3. `packages/engine/mcp-server/tests/tools/fulltext-search.test.ts:83` still asserts the shape your own measurement disproved.**

Your live call established that `highlightTexts` are bare terms, and you fixed the spec and all 9 eval fixtures. The unit mock still carries the marked-up form, and line 361 asserts it. The assertion is a pure pass-through check, so both lines can change without weakening it.

You wrote:
```
      highlightTexts: ["Patrick <em>Flynn</em>"],
```
Change it to:
```
      highlightTexts: ["Patrick", "Flynn", "Flynn Patrick"],
```

You wrote:
```
    expect(r.highlightTerms).toEqual(["Patrick <em>Flynn</em>"]);
```
Change it to:
```
    expect(r.highlightTerms).toEqual(["Patrick", "Flynn", "Flynn Patrick"]);
```

---

**4. Two fixtures echo a `query` production cannot produce.**

Pre-existing, and it caused none of this run's failures — but it is the same class of defect you are fixing, in two files you already have open. `echoQuery` copies the input, so it can only echo what was sent. Of the 15 fulltext fixtures these are the only 2 whose echo names a key their own match args do not. Its one visible effect is that the judge invented "the tool appears to have applied defaults" in `ut_001`'s rationale.

In `fulltext-search-flynn-witnesses.json`:

You wrote:
```
    "query": { "keywords": "+Flynn", "recordPlace1": "Pennsylvania", "recordPlace2": "Schuylkill" },
```
Change it to:
```
    "query": { "keywords": "+Flynn" },
```

In `fulltext-search-flynn-witnesses-by-name.json` (this one matches on `name`, not `keywords`):

You wrote:
```
    "query": {
      "keywords": "+Flynn",
      "recordPlace1": "Pennsylvania",
      "recordPlace2": "Schuylkill"
    },
```
Change it to:
```
    "query": { "name": "+Flynn" },
```

Neither edit changes a current result — `keywords` and `name` are outside the validator's filter keys. It is shape hygiene, and cheap while you are here.

---

**Not blocking, for your judgement**

- This closes a `nothing-checks` issue without adding anything that can fail. You correctly diagnosed that the mock injects `staged` without mirroring the producer's strip, then fixed nine data files rather than the injector — so a tenth fixture reintroduces it silently. I would not do that here: it touches every staging search tool and overlaps the open question on #1779 about whether the writer tools should prevent this at source. A note on #1779 is the right home. Please don't open a new issue for it.
- The new Result triage dimension scored 3 on all 8 tests that exercised it, and it is the only check grading the step-4 rewrite. Not asking for a paid re-run — just flagging that it has not yet been observed discriminating.
- `ut_001`'s partial is explainable and not your doing. The judge graded against the `~Flynn` substring predicate, which is exactly what `ut_010`'s `judge_context` carries a note to prevent. `search-for-flynn-witnesses.json` uses the same substring fixtures and has no such note. Worth adding sometime; not here unless you want it.

The deferral on the bounded-excerpt output field looks right to me, and routing it to the lead rather than taking it was the correct call.

promise-emmanuel (changes requested):
Building on Richard's review rather than repeating it — his four items all stand and I'd take them as written. I re-checked at 832c5579; the PR's own diff is byte-identical to f8b33dde, so nothing he or I found has moved.

Two corrections to that review, and one thing neither of us covered.

The fixture echo is not just hygiene — it's where the false data comes from. Richard has it as shape-only, causing nothing but a stray judge rationale. I pulled ut_010's transcript. The skill said "I'll search name only first — no place or date filters on the first call", then sent {keywords: "+Flynn +witness", count: 50} with no filters at all. Only after the response came back did it report "Schuylkill County, Pennsylvania filter" and log recordPlace1: "Pennsylvania", recordPlace2: "Schuylkill" — the echo's two keys and values, grafted onto its own keywords. It had already decided against filters before it saw them.

So his item 1 and his item 4 are both needed: the unreferenced research-log-protocol.md is the missing guard, and the echo is what the agent is guarding against. Please do both, not one.

The transcript is reachable, so the skill shouldn't say it isn't. Richard confirmed the new code comment's claim, but the three things he checked — readStagedResults, record-read, rank_search_matches — are all MCP tools, and that's exactly what the sentence doesn't say. I ran the tool against a temp project: it hands the agent staged.resultsRef, and the file at that path contains the full transcript, because staging serializes the response before the strip runs. Read is ungated — the plugin hook matches only Write|Edit|NotebookEdit|device_commit_files|research_append, and nothing forbids reading results/.

Your narrower point is right and worth keeping. The unqualified version is what needs to change, in four places.

packages/engine/plugin/skills/search-full-text/SKILL.md:175

You wrote:


Each staged result carries only flat stub fields — the full transcript
(`textDocument`) is dropped once results are staged and cannot be re-read via a
tool. Triage from the stubs, then confirm against the image:
Change it to:


Each staged result carries only flat stub fields — the full transcript
(`textDocument`) is dropped from the response and no MCP tool reads it back. It
is still on disk at `staged.resultsRef`, but reading it pulls the whole page
(79–136 KB) back into context. Triage from the stubs first:
SKILL.md:186

You wrote:


  alignment, a place name matching a surname) **cannot be judged from a staged
  result** — the terms are bare and the transcript is not returned. Open the
  record and read the image to settle it (see the verification note above).
Change it to:


  alignment, a place name matching a surname) cannot be judged from the stubs —
  the terms are bare. Verify against the original image (see the verification
  note above), or `Read` the single result you are chasing out of
  `staged.resultsRef`.
eval/tests/unit/search-full-text/rubric.md:35

You wrote:


the full transcript (`textDocument`) is stripped once results are staged and cannot be re-read via a tool.
Change it to:


the full transcript (`textDocument`) is stripped from the tool response and no MCP tool reads it back, though it remains on disk at `staged.resultsRef`.
eval/tests/unit/search-full-text/rubric.md:41 — as written, this fails an agent for using the free path. Please add to the end of that bullet:


Reading the transcript back from `staged.resultsRef` is not a fail.
Same correction to the spec's Result staging note (line 217) and the code comment (fulltext-search.ts:239) so all six agree.

If you'd rather keep the doctrine of never touching the transcript — the context blow-up is a fair reason — that's fine, but say that instead of saying it can't be done, or the next person will believe it.

One consequence worth naming: stripping textDocument from the fixtures makes the response faithful and the sidecar unfaithful, since production stages the full payload and strips only the response copy. That means the correct behaviour can't be exercised in the eval at all. It's another reason the mock-injector fix is the right shape — agreed with Richard that it belongs as a note on #1779, not a new issue.

Suites on the merge with main: engine vitest 2722 (2726 at the new head), tsc clean, harness pytest 2874 / 2 skipped, server 210 / 1 skipped.

francis-2008-happy:
All items in at `8f1f75c8`. Thank you both — the transcript-reachability catch is the one I'd have shipped wrong, and it was my sentence that was wrong, not the reading of it.

**The transcript is reachable, so nothing says otherwise now.** @promise-emmanuel is right and I verified the gating independently: the plugin hook matcher is `Write|Edit|NotebookEdit|.*device_commit_files|.*research_append`, so `Read` is untouched, and staging serializes the response before the strip runs. My narrower point survives — no *MCP* tool reads it back — but I wrote it unqualified, and the real reason to triage from stubs is the 79–136 KB, not a wall. Corrected in all six places, verbatim where you gave the text: `SKILL.md:175`, `SKILL.md:186`, `rubric.md:35`, `rubric.md:41`, the spec's staging note, and the code comment.

**One of those six is @DallanQ's file, so flagging it rather than burying it.** The comment moved into `utils/staged-compaction.ts` when #2045 landed, and it carried my over-broad wording across with it. I corrected it there too — leaving one of six sites still claiming the wall exists seemed worse than touching a file from someone else's PR. It is comment-only and trivially revertable if you'd rather own it, Dallan.

**Item 1, adapted, and I want to be explicit about the deviation.** @chesworthrm asked SKILL.md to point at `references/research-log-protocol.md`. Between that review (27 Aug) and now, #2045 deleted the file (30 Aug) as unreachable — which is the same defect you diagnosed, resolved the other way round. Rather than pick between you, the rule itself is now inline in step 7:

> Record only a filter the call actually sent — even when an unsent filter matches the locality under research, and even when the response echoes one back. The tool echoes your own request; an echoed key you did not send is not a filter you applied.

That loads with the skill, reintroduces no unreachable reference for the new lint to flag, and states the echo case explicitly, which the reference file did not. Say the word if you'd rather the file came back instead.

**Items 3 and 4** are as written — the unit mock and its assertion now carry the measured bare terms, and the two fixtures echo only keys their own args sent.

**Item 2:** PR body corrected. It said `ut_010` made 0 tool calls; it ran the search and then mis-logged. You were right that the wrong description is why the failure got waved through, and the corrected body now names the actual mechanism.

**The mock-injector fix stays a note on #1779**, per you both — not a new issue, not fixed here.

**On the run log, which is now a separate problem from anything you raised.** #2045 also deleted `references/validation-protocol.md`, and both deleted files are recorded in the snapshot of every `search-full-text` run log — so the committed log no longer validates, and so does Taiwo's on #2001. Neither of us can merge on the runs we have. A fresh `make eval-skill` plus annotation is queued, but I'm holding it until you're both satisfied with the prose above: `SKILL.md` and `rubric.md` are snapshotted, so any further wording change after the run throws it away. Taiwo is holding his run until this merges, so we buy one between us rather than three.

So: if the inlined rule works for you in place of the reference file, say so and I'll run.

Suites at this head: engine vitest `fulltext-search.test.ts` 39/39, `tsc --noEmit` clean, skill/agent frontmatter lint clean, both edited fixtures valid JSON.

chesworthrm (commented):
Second pass, reviewed at `8f1f75c8` (full pass). I test-merged `origin/main` first — clean, and it brings only six auth files, so nothing overlaps this diff. On the merge: engine vitest **2746 passed**, `tsc --noEmit` clean, harness pytest **2959 passed / 2 skipped**. I did not run `apps/server` pytest or `turbo test typecheck`; no path in those trees is touched.

**All eight items from the first round are resolved.** Item 1 you resolved better than I asked — I pointed you at `references/research-log-protocol.md`, then #2045 deleted that file, so inlining the rule into step 7 was the right call. Items 2, 3 and 4 are done; I audited all 15 fulltext fixtures and **zero** now echo an unsent filter key. Promise's six wording sites are applied verbatim, and I grepped repo-wide for the old "cannot be re-read via a tool" / "NOT retrievable" phrasing — swept clean. Both of our asks for a note on issue #1779 rather than a new issue: done, and honourably updated on 08-31 to record that #2045 landed the durable fix.

I also verified the new reachability claim end-to-end rather than by reading. Running the mock's staging path against the HEAD fixture: the inline response has `textDocument` stripped, and the file at `staged.resultsRef` retains the full transcript. The claim is correct in all six places.

One correction in your favour, on my own non-blocking note last round. I flagged that **Result triage** had not been observed discriminating. There is better evidence in your run logs than either of us used — see the third item below.

`runlogs` is the only red check. Everything else passes on this head. Two approvals are required and there are none yet; both senior-developers and senior-genealogists own these paths, and your reading of CODEOWNERS is right that the spec `.md` matches no rule.

---

### The blocker: the acceptance run predates both fixes for the only failing test

I checked this by snapshot hash rather than commit dates, because the run reads the working tree.

The 08-27 run recorded SKILL.md at `9c4e49aa` — byte-identical to `4252b4a0`, which does **not** contain the step-7 rule. It recorded the witnesses fixture at `cb094ba0`, also `4252b4a0`'s, which still carries the misleading `{keywords, recordPlace1, recordPlace2}` echo. Both `ut_010` remedies landed in `8f1f75c8` on 08-31. So the run cannot show either one working, and HEAD hashes to a third state matching neither run.

Your diagnosis of the failure is right, and the transcript confirms it: the skill sent `{keywords: "+Flynn +witness", count: 50}` with no filters, then logged `recordPlace1: "Pennsylvania"` and `recordPlace2: "Schuylkill"` — the echo's exact two keys and values.

The judge prompt has also moved twice since (run `c39d7003`, current `03f306ff`), so the scores will shift regardless.

You wrote:
```
**Acceptance check:** `make eval-skill SKILL=search-full-text` — the new **Result triage** rubric dimension scores 3 on every test that exercises it (it is the only check that grades the step-4 rewrite; the 3 prior dimensions grade none of triage). Also: `make engine-test` green (2722); spec and implementation agree on every input and output field (verified field-by-field).
```
Change it to:
```
**Acceptance check:** `make eval-skill SKILL=search-full-text` on a run made **after** `8f1f75c8` — `ut_010` passes `test_log_query_traces_to_fulltext_search_call`, and the new **Result triage** dimension grades the step-6 rewrite. The committed 08-27 run predates both `ut_010` remedies, so it cannot show them working. Also: `make engine-test` green (2746 on the merge with main); spec and implementation agree on every input and output field (verified field-by-field).
```

Run it from this PR's worktree, not the primary checkout.

---

### The body describes a fixture fix that is no longer in the diff

`4252b4a0` stripped `textDocument` from all 9 fixtures, and the 08-30 merge correctly took main's back — PR #2045, now this branch's merge-base, solved it in the mock instead. Nothing is broken and #2045's shape is better. But a reviewer reading the body goes looking for your headline fix and finds nothing.

You wrote:
```
Stripped it from all 9 `fulltext-search-*.json` so `staged ⇒ no textDocument`, matching production. (`record_search`'s gedcomx has the same divergence — a separate card, not this one.)
```
Change it to:
```
This PR originally stripped `textDocument` from all 9 `fulltext-search-*.json`. **That is no longer in the diff and should not be re-added:** PR #2045 (now this branch's merge-base) made the mock call the compiled `compactStagedFulltextSearch`, so the fixture keeps the full payload and the mock strips the response copy — production's exact shape, and not defeatable by a tenth fixture. The 08-30 merge restored main's fixtures on purpose.
```

---

### You have stronger acceptance evidence than the one you used

"Result triage scores 3 on every test that exercises it" is weak — a dimension that passes everywhere shows little. The strong evidence is a pre/post comparison already sitting in your two run logs.

I searched both for transcript content that exists **only** in `textDocument` and in no stub field:

| Run | `textDocument` at run time | Agent quoted transcript-only content |
|---|---|---|
| `v1_2026-08-24` (pre-PR) | present | **4 of 13 tests** — "Signed sealed and delivered", "in the presence of", "public road" |
| `v1_2026-08-27` (post-PR) | stripped | none |

That is the defect in #1826 actually manifesting, and then stopping. It is worth stating in the body, and it largely answers my note from last round.

Two caveats to state alongside it, so it holds up: the delta is driven by the harness no longer *offering* the field (now via #2045's mock strip), so the SKILL.md rewrite's independent contribution is unmeasured; and **Result triage** still has not been seen scoring below 3. On the re-run, please say whether it ever does.

---

### One line to route the tool-side half of the log-fidelity rule

CLAUDE.md's lane rule asks first whether a prose rule can be a writer-tool precondition. This one can. `research-log-append.ts:321` already holds both the caller's `query` and `fin.payloadQuery` — the producing tool's own echo — and the comment beside it states the tool deliberately lets the caller own its contents. Issue #1779's title already carries this exact decision and you have already commented there, so there is nothing to build here. Just say the prose is the interim half, because `ut_010` is proof that a prompt-level rule did not hold.

You wrote:
```
**Unsure about:** nothing outstanding. One deliberate deferral below.
```
Change it to:
```
**Unsure about:** nothing outstanding. Two deliberate deferrals below. The step-7 log-fidelity rule is the prompt-level half only — `research_log_append` already holds both the caller's `query` and the staged payload's echo (`research-log-append.ts:321`), so enforcing it as a writer-tool precondition is issue #1779's open decision, not this card's.
```

If that precondition is ever built, one thing to know: in the eval the staged payload's `query` is the fixture's canned echo, not the real args, so a strict comparison would false-fail there until the mock echoes actual args.

---

### Minor

Your Acceptance check says "the step-4 rewrite". The diff rewrites **step 6**, as your Review guidance section correctly says.

---

### What I did not check

- I did not re-run `make eval-skill`. It is paid and your re-run is already queued.
- **Result triage** has not been watched failing. I did not build a run against the pre-PR SKILL.md to force it; the transcript comparison above is the cheaper substitute.
- `apps/server` pytest and `turbo test typecheck` — no path touched.
- Whether the `79–136 KB` figure newly added to SKILL.md needs a `measured-figures.test.ts` provenance entry. vitest is green either way, but I did not read the guard to see whether it covers SKILL.md at all.
- The `prompt-budget` warning is real (SKILL.md 13.1 KB → 14.3 KB) but I did not challenge any added line — that wording was settled with Promise last round and I would only be re-litigating it.

The engine and doc work is correct and the prior round is fully answered. The blocker is evidence, not code. Re-run after `8f1f75c8`, refresh the three stale body claims, and this is ready.

I am leaving this as a comment rather than a second changes-requested — Promise's standing already blocks, `runlogs` blocks mechanically, and you do not need two of us to clear. Promise will need to re-review to clear his; his was read at `832c5579`, three pushes back.

promise-emmanuel (approved):
Approving. I re-reviewed at 98430f12 (full pass, level with main); my six wording
sites from 832c5579 are all applied verbatim, and Richard's eight are resolved. I
re-verified the whole-set claims myself rather than taking them on trust: all 15
fulltext fixtures now echo zero unsent keys, the old "cannot be re-read" phrasing
is gone repo-wide, and the spec agrees with src/types/fulltext-search.ts on all
17 inputs and every output field. count ?? 5 at fulltext-search.ts:73 confirms
the spec's 5. Suites on this head: engine vitest 2782, tsc clean, harness pytest
3042/2 skipped, fulltext-search.test.ts 39/39. All six required checks green.

Two things to fix in the body before you merge — no code changes.

1. The staleness warning is no longer true, and it is costing you a paid run.

I hashed every snapshot entry with the harness's own snapshot.hash_file (raw
sha256 gives false mismatches — it normalizes text first). The acceptance run
validates perfectly against this head: 49 entries, 0 mismatches, and it records
neither deleted reference file. The four older logs are the ones that mismatch
(19-20 entries each), and check_runlogs Rule 2 only validates the latest
full-skill run log per skill — which is why runlogs is green. So the queued
re-run would throw away $2.11 and buy nothing. Please don't run it.

You wrote:
**This run log is stale and will be replaced.** #2045 deleted `search-full-text/references/{research-log-protocol,validation-protocol}.md`, both of which every run log for this skill records in its snapshot, so the committed log no longer validates. A fresh `make eval-skill SKILL=search-full-text` plus annotation is queued once the review items are confirmed settled — running earlier would only throw the run away.

Change it to:
**The acceptance run validates on this head; no re-run is needed.** #2045 deleted `search-full-text/references/{research-log-protocol,validation-protocol}.md`, which the four older committed logs record in their snapshots — those mismatch. `v1_2026-08-31_19-59-51` post-dates the deletion, records neither file, and hashes clean against the working tree (49 entries, 0 mismatches). `check_runlogs` Rule 2 validates only the latest full-skill run log per skill, which is that one, so `runlogs` is green on the real acceptance evidence.

2. Richard's ask on the deferral count is still unapplied, and the lane rule in
CLAUDE.md is the reason it matters — ut_010 is proof a prompt-level rule did not
hold, so the body should say the prose is the interim half. I confirmed his
premise: research-log-append.ts:321 does hold both the caller's `query` and
`fin.payloadQuery`, with the caller winning.

You wrote:
**Unsure about:** nothing outstanding. One deliberate deferral below.

Change it to:
**Unsure about:** nothing outstanding. Two deliberate deferrals below. The step-7 log-fidelity rule is the prompt-level half only — `research_log_append` already holds both the caller's `query` and the staged payload's echo (`research-log-append.ts:321`), so enforcing it as a writer-tool precondition is issue #1779's open decision, not this card's.

One answer to Richard's open question, so it is on the record: Result triage never
scores below 3 on this run. It fires on 9 of 13 tests, and 8 of those are
substantive — the rationales name the specific stub fields checked. Only ut_002 is
the nil auto-pass. So it is doing real work but still has not been seen
discriminating. Worth a line in the body; not worth a paid run.

This needs a second approval — two are required and mine is the first.

johnmarkpeterbrown (approved):
> Approving — this is the second approval, so it clears the gate once the body is corrected.
>
> I verified the substance independently rather than taking the prior rounds on trust: spec and implementation agree on every input and output field, all 15 fulltext fixtures echo zero unsent keys, and the transcript-reachability claim holds end to end (staging serializes before the strip runs). Engine vitest 2782 / tsc clean; harness pytest 3033 passed, and its one failure was a missing engine build in a fresh worktree — 31/31 after `npm run build`.
>
> Two body edits before you merge, no code:
>
> 1. **Please cancel the queued re-run.** The staleness warning is not true. Hashing through the harness's own `normalize`, `v1_2026-08-31_19-59-51` is 49 entries / 0 mismatches against this head and records neither deleted reference file. Rule 2 validates only the latest full-skill run log per skill, which is that one — which is why `runlogs` is green. Apply the replacement text promise-emmanuel gave you.
>
> 2. The two-deferrals edit both Richard and Promise asked for is still unapplied. Apply it as written.
>
> One thing neither round caught, and **not** for this PR. The fixture echo is fixed at key level but still wrong at value level: the predicates are `~` substring matches, so one canned echo answers many real queries. Running the mock, a call of `{keywords: "+Flynn +witness", count: 50, projectPath: …}` gets back `query: {"keywords": "+Flynn"}`. That is `ut_010`'s failure mode one field narrower, and step 7's rule doesn't cover it — it is keyed on "an echoed key you did not send", and `keywords` is a key it did send. 83 of the 87 fixtures with a canned query mismatch their own args this way, and a blanket fix would be wrong for `external_links_search`, whose echo carries a resolved place production genuinely produces. Please add this as a note on issue #1779 — it already owns the writer-tool question and you have the thread there. Not a new issue.

Issue #1789, opened 2026-08-20
Cowork: does listing all three server spellings in a skill's allowed-tools stop the approval prompt?


DallanQ:
> **ANSWERED 2026-08-25 — cause 1. Closing.** The qualified spelling binds; the bare
> one grants nothing.
>
> Run on macOS Cowork against a project folder created during the session, so the
> permission state was clean by construction. `citation` carrying three qualified
> spellings on `research_append` → **no prompt**. The bare `validate_research_schema`
> control, same turn → **prompted**. That is row 1 of the reading table below.
>
> Seven more tools in the same session corroborate it: Project Context, Project Create,
> Record Search, Research Log Append, Record Read, Extraction Append and Validate
> Research Schema are all declared **bare** by their skills and **all prompted**. Eight
> tools, one variable.
>
> It matches the documented behaviour — permission rules "match the canonical name
> only," and an MCP tool's canonical name is `mcp__server__tool`, so a bare
> `research_append` matches nothing.
>
> **Two limits on what this licenses.** All three spellings were listed together, so it
> does not isolate which bound — only `mcp__remote-devices__Genealogy_Research__…` can
> match what a Cowork session exposes. And the grant is **per-turn**: it "clears when
> you send your next message," so it suppresses prompting inside the invoking turn and
> not across a conversation.
>
> **The build half is issue #1787**, which stays open and carries the standing
> `nothing-checks` gap this card also held — no CI job can verify a permission grant
> binds in Cowork, and closing this one does not close that.
>
> Also corrected on the way out: setup step 3 predicted three build warnings. They do
> not appear — `check_skill_frontmatter.py` normalizes a qualified name to its last
> `__` segment, so the three lines resolve clean. A tester expecting warnings would
> have concluded the edit had not taken.

> **Reviewed 2026-08-20 before junior handoff.** No decision needed; the
> observation protocol below was added so the run cannot produce a confident
> wrong answer.

**Touches:** none — the test runs on a scratch copy and nothing is committed.

Split out of issue #1787 on 2026-08-20. That issue measured the symptom: **every
genealogy tool call prompts the user in Cowork**, including tools the invoking skill
declares in its `allowed-tools`. Fresh Cowork session, project with existing sources,
prompt *"Improve the citations in this project."* →

> Claude wants to use Research append from Genealogy Research

`citation` declares `research_append` in `allowed-tools`. It was prompted anyway, and a
[...]

PR #1982, opened 2026-08-27
agent-tool-bind: live probe that gps-mentor's wiki_search grant binds at runtime (#1084)

### LINKED ISSUE #1084: Nothing can verify a granted tool binds in Cowork — the register entry for a gap no CI job can reach

francis-2008-happy:
## Summary
Closes #1084 (Normal tier). Adds `make agent-tool-bind` — a durable, opt-in behavioral probe that gps-mentor's `wiki_search` grant actually **binds** at runtime. Every existing check stops at spelling: `agent-tool-names.test.ts` proves the `tools:` entries are well-formed, `test_plugin_agents.py` proves the agent registers under its bare name — but nothing proved the agent can *call* a granted tool. The SDK handshake exposes no tool list (name/description/model only), so the only proof is behavioral.

This is the **repo's first check that spends a model turn** (~$0.35/run, measured), per the lead's 2026-08-27 ruling: durable opt-in target beside `make agent-smoke`, delete-don't-loosen if flaky, `mcp__genealogy__` spelling only.

## Review guidance

**Start here:** `apps/server/tests/test_agent_tool_binding.py`. It spawns gps-mentor via a live `ClaudeSDKClient` (the same `real_agent.build_options` path as `agent-smoke`), forces one `wiki_search` call through a Task-delegation prompt, and asserts the `wiki_search` tool_use is recorded **in gps-mentor's own subagent transcript** — not anywhere, since the main session also holds the tool and would false-pass. It asserts on the tool_use record, never narration. The one subtlety worth a look: the workspace leaf is `agent-tool-bind-{uuid4().hex}` (dash/hex, no underscore) because the CLI slugifies the cwd and the transcript scan matches `slug.endswith(leaf)` — an underscore would make the match always fail (this was plan-critic's blocking finding).

**Unsure about:** nothing outstanding. The `agentType` meta value was verified bare (`gps-mentor`) from the passing run; the match also accepts a namespaced form for durability.

## Plan

**Didn't change:** the probe covers gps-mentor × wiki_search × the `mcp__genealogy__` spelling only — one agent, one tool, one turn, not a matrix (a matrix turns a smoke check into a bill). It does **not** close the nothing-checks gap: every other agent/tool/spelling is still unverified, so #1084 stays on the register (§9.4 item 1 says so). Reimplemented the ~20-line subagent-transcript scan inline rather than importing `eval/harness/e2e/subagent_capture` — confirmed not importable across the two uv projects.

**Acceptance check:** `make agent-tool-bind` is **green** with the grant (gps-mentor records a `wiki_search` tool_use; $0.3520) and **red** when the `mcp__genealogy__wiki_search` line is removed from `gps-mentor.md` `tools:` (recorded calls `[]` → "grant did not bind"; $0.2639). I ran both — the check genuinely discriminates, which is the whole point of a check that spends money. It is opt-in (skips with no key), so it does not run in `make test-all`.

**Deviated from the plan:** nothing. plan-critic caught a blocking flaw before any code (pytest's `tmp_path` leaf has underscores → the slug match would fail on every run); fixed with the dash/hex leaf.

## Test plan
- [x] Behavioral verification above (green with grant / red without) — the delete-don't-loosen proof.
- [x] `apps/server` suite: the probe skips cleanly with no key (cannot be triggered by an ordinary `make server-test`).
- [x] `make help` parses; the target shows with the measured $0.35 figure. `$0.35` is consistent across the Makefile help, the `.bat`, and §9.4.
- [ ] ~eval-skill / routing negatives~ — N/A: no skill/eval touched.

## What this does not solve
A green probe proves *one* grant binds under *one* registrar spelling. The two Cowork spellings and every other agent × tool are still unverified. No CI job can cover this — only a live session in the run mode being tested — so it stays on the nothing-checks register (#1084), and the sibling skill-grant measurement is #1789.

## For review
`apps/server/tests/test_agent_tool_binding.py`, `Makefile`, `scripts/windows/agent-tool-bind.bat`, `scripts/windows/README.md`, `docs/architecture.md`. The `.py`/`Makefile`/`.bat` paths route to **both** senior-developers and senior-genealogists via CODEOWNERS; the two `.md` files match no rule (docs pull no owner), but the code files already pull both teams.


chesworthrm (changes requested):
Full pass, reviewed on the test-merge with `main` (the branch is 10 commits behind, so its own green head is not the state that merges). First review round — no prior reviews, no comments, no unresolved threads.

**Does:** adds `make agent-tool-bind` and `apps/server/tests/test_agent_tool_binding.py`, which spawns gps-mentor through the hosted `real_agent.build_options` path, forces one `wiki_search` call by Task delegation, and asserts the `wiki_search` tool_use lands in gps-mentor's own subagent transcript. Plus the Windows wrapper, its README row, and a §9.4 paragraph.

**Intent:** matches. Nothing in the diff is unexplained by it, and nothing touched that **Didn't change** said was left alone. The **Acceptance check** is genuinely falsifiable — green with the grant, red without — which is the right shape for a check that costs money.

**Checks:** all 13 check runs are green on `fae60fc2`, and all 13 ran on that head. Nothing stale.

**Still needs:** one senior approval, from either senior-developers or senior-genealogists. I resolved the changed paths through the real senior-queue CODEOWNERS parser: `Makefile`, the `.py` and the `.bat` each pull both teams; the two `.md` files pull no owner. Branch protection still wants two approvals in total.

**Verified:** engine vitest 2726 passed; `apps/server` pytest 210 passed, 2 skipped (the two live probes, skipping for the right reason); `eval/harness` pytest 2874 passed. One harness failure on my first run was a missing engine build in a fresh worktree, not this PR — green after `npm run build`. Worth knowing separately: unlike its 6 siblings, `test_mock_mcp.py::test_successful_call_carries_no_is_error` fails rather than skips when the build is absent.

**Lead stop-rules:** satisfied. Spending money on every run is the lead's call and it is already made — I read the 2026-08-27 ruling on #1084 and it says what the PR body quotes.

I did not run `make agent-tool-bind`. No live key here, and I would not spend your budget to re-confirm a figure you measured twice. So the green/red discrimination stays your evidence, not mine. That matters for the first finding, because neither of your two runs could have hit it.

---

## [CRITICAL] (9/10) `apps/server/tests/test_agent_tool_binding.py:130` — the probe goes green in the case it exists to catch

The fallback fires whenever no transcript matched, which includes the case where the meta is present and names a **different** agent. So when Task delegation fails and the model falls back to a general-purpose stand-in — issue #939, the documented failure — that stand-in's transcript is the only one on disk, it gets read as gps-mentor's, and it holds `wiki_search` because a stand-in is not bound by gps-mentor's `tools:` list. Both assertions pass. A spawn failure reports as a bound grant.

I ran your function against a transcript whose meta reads `general-purpose` with one `wiki_search` tool_use. It returned `['wiki_search']`, so both asserts hold.

Two facts from the committed corpus say the fallback is protecting the wrong case. Across 665 subagent summaries in `eval/runlogs`, every single one has a non-null `agentType` — the missing-meta case has never happened. And 19 of them are `general-purpose`, so the case the fallback *does* catch is the one that happens. General-purpose subagents made 163 tool calls in that corpus, 40 of them `mcp__genealogy__` including 11 `tree_edit`.

Deleting it costs nothing. I checked both directions on a patched copy: the general-purpose transcript then returns `None` and the probe fails with your "not spawned as itself" message, and the real gps-mentor case is unchanged.

```
You wrote:      if chosen is None and len(jsonls) == 1:
                    chosen = jsonls[0]

Change it to:   (delete both lines)
```

The comment above it then describes the code again:

```
You wrote:      # Pick this agent's transcript by its meta `agentType`. With a single delegated
                # subagent, a missing meta falls back to the sole transcript.

Change it to:   # Pick this agent's transcript by its meta `agentType`. No fallback: a
                # transcript whose meta names a DIFFERENT agent must not be read as this
                # one, or a general-purpose stand-in (issue #939) passes the probe.
```

## [CRITICAL] (8/10) `apps/server/tests/test_agent_tool_binding.py:153` — it is not opt-in, and it bills

The skip only fires when a prerequisite is **missing**. `AGENT_TOOL_BIND` is read but gates nothing — it only upgrades a skip into a failure. So anyone with `LIVE_ANTHROPIC_API_KEY` in their shell, a built engine and node on PATH spends a model turn on a plain `make server-test`, and on `make test-all`, without asking for it.

I evaluated your skipif marker under three environments. With the key set and the flag unset, the condition is `False` — pytest runs the body.

This gate is copied from `agent-smoke`, where it is harmless because that check bills nothing. Here it is the difference between a charge you asked for and a surprise. CI is safe: the server job builds no engine and sets no such key.

```
You wrote:      @pytest.mark.skipif(
                    bool(_missing_prerequisites()) and not _INVOKED_AS_PROBE,
                    reason="live probe — run `make agent-tool-bind`: " + "; ".join(_missing_prerequisites()),
                )

Change it to:   @pytest.mark.skipif(
                    not _INVOKED_AS_PROBE,
                    reason="live probe that SPENDS A MODEL TURN — run `make agent-tool-bind`",
                )
```

I checked this against all four key/flag combinations on a patched copy. Only the one hazardous cell changes. The `pytest.fail` remedy list still fires under `make agent-tool-bind` with no key, so that path is untouched.

## (7/10) `Makefile:360` and `scripts/windows/agent-tool-bind.bat:17` — the cost never prints on a green run

The test captures the turn's cost "so the run can report the figure the target's help text quotes." Under `-q -rs`, pytest discards a passing test's stdout. It shows on failure, which is how you saw `$0.2639` on the red run — but the green run never prints what it cost.

```
You wrote:        uv run pytest tests/test_agent_tool_binding.py -q -rs

Change it to:     uv run pytest tests/test_agent_tool_binding.py -q -rs -s
```

Same change on the last line of `scripts/windows/agent-tool-bind.bat`.

## (6/10) `apps/server/tests/test_agent_tool_binding.py:185` — nothing bounds the turn this check pays for

`build_options` sets no `max_turns`, and the receive loop has no time limit. The help text and §9.4 both promise `~$0.35/run` with nothing holding it there. The failing case is the worst one: the grant does not bind, gps-mentor cannot finish the instruction, and the main agent is free to retry.

The e2e orchestrator bounds every live run for exactly this reason — wall clock, inactivity, `max_turns`, and `max_cost_usd: 15.0`, which it checks against `total_cost_usd`, the same field this probe reads and only prints.

A wall-clock bound rather than `max_turns`, because gps-mentor subagents in the corpus reach 66 assistant turns and I have not confirmed whether a subagent's turns count against the parent's budget:

```
You wrote:      await client.connect()
                cost: float | None = None
                try:
                    await client.query(_PROMPT)
                    async for message in client.receive_response():
                        # The final ResultMessage carries the turn's billed cost — captured
                        # so the run can report the figure the target's help text quotes.
                        c = getattr(message, "total_cost_usd", None)
                        if isinstance(c, (int, float)):
                            cost = c
                finally:
                    await client.disconnect()

Change it to:   await client.connect()
                cost: float | None = None

                async def _drain() -> None:
                    nonlocal cost
                    await client.query(_PROMPT)
                    async for message in client.receive_response():
                        # The final ResultMessage carries the turn's billed cost — captured
                        # so the run can report the figure the target's help text quotes.
                        c = getattr(message, "total_cost_usd", None)
                        if isinstance(c, (int, float)):
                            cost = c

                try:
                    # A check that bills must have a ceiling. Unbounded, a main agent that
                    # cannot complete the instruction retries past the ~$0.35 this target
                    # promises (cf. the orchestrator's wall-clock and max_cost_usd caps).
                    await asyncio.wait_for(_drain(), timeout=300)
                finally:
                    await client.disconnect()
```

That needs `import asyncio` added to the imports.

---

## Checked and fine

- The dash/hex workspace leaf is correct and load-bearing. I reproduced plan-critic's finding: an underscore leaf returns `None` on every run. Keep the comment that explains it.
- `asyncio_mode = "auto"` does apply as the Makefile invokes pytest, so the async test really executes.
- The wrapper mirrors `agent-smoke.bat` line for line, is CRLF with no BOM per `.gitattributes`, and all 12 wrapper-drift assertions pass. I mutation-tested them against your files: dropping the README row fails two, renaming the target fails a third. The guard binds.
- `make help` renders `$0.35` intact.
- gps-mentor declares `wiki_search` under all three spellings, so removing the `mcp__genealogy__` one was the right mutation for the red run.
- The §9.4 paragraph does not overclaim; #1084 is still open and still labelled.
- Every Python file op passes `encoding="utf-8"` as a keyword.
- Your note about `subagent_capture` not being importable is slightly overstated — the module is stdlib-only, so a path insert would work — but it is not packaged for cross-project import, `real_agent.py` already duplicates across the same boundary for the same reason, and your copy diverges on purpose by filtering on `agentType`. Fine as written.

## Not checked

`npx turbo test typecheck` for the web workspace, and `tsc --noEmit`. No TypeScript or web source changed, and the engine vitest run covers every packaging test that reads the Makefile, the wrapper and the README.

francis-2008-happy:
All four in as of `653a545f`. I reproduced each one before fixing it, and you were right that neither of my two paid runs could have hit the first — both spawned gps-mentor correctly, so neither ever exercised the fallback.

**Finding 1 — the probe went green in the case it exists to catch.** Confirmed exactly as you described. I built a transcript whose meta reads `general-purpose` with one `wiki_search` tool_use; `_subagent_tool_calls` returned `['wiki_search']` and both assertions held. A #939 spawn failure reported as a bound grant. Fallback deleted, comment rewritten with the 665/19 corpus figures you measured, and that case now returns `None` and fails with the "not spawned as itself" message.

**Finding 2 — it billed without being asked.** Confirmed across all four key/flag cells: key set + flag unset ran the body. Gated on `AGENT_TOOL_BIND` alone now. Re-evaluated after the change — only that cell moves, and the `pytest.fail` remedy list still fires under `make agent-tool-bind` with no key, so the loud path is untouched.

**Findings 3 and 4** — `-s` on both the Makefile target and the `.bat`; `asyncio.wait_for(_drain(), timeout=300)` with `import asyncio`. Wall clock rather than `max_turns` for the reason you gave. The `.bat` is still CRLF, no BOM — my first edit flattened it to LF and I caught it on the `.gitattributes` warning.

**One thing beyond the review, and it is the part I would look at hardest.** After applying your two critical fixes I reverted them both and ran the suite: **210 passed, 2 skipped, fully green.** Nothing held either fix. The only reader of `_subagent_tool_calls` is the live test, which skips. So both defects could have come straight back and no check would have noticed.

I added four offline guards — they bill nothing and are deliberately not behind `_INVOKED_AS_PROBE`: the stand-in case, a positive control, the namespaced `agentType` form, and the billing gate.

**The gate guard took three attempts and the first two were the exact failure I was guarding against.** Comparing the marker to `not _INVOKED_AS_PROBE` passed with the old condition restored — with any prerequisite missing the two expressions evaluate the same, so it agreed with the bug. Re-importing in a subprocess failed the same way, because the hazard needs a live key *and* a built engine *and* node, and `_ENGINE_BUILD` is a fixed repo path a test cannot conjure. So the property is pinned on the gate's **source** instead, the way the #1741 window guard reads the orchestrator's, with the two reachable cells still evaluated for real. Mutation matrix now:

| Mutation | Before the guards | After |
|---|---|---|
| restore the sole-transcript fallback | **210 green** | 1 failed |
| restore the prerequisite-keyed skipif | **210 green** | 1 failed |

**Verification.** `apps/server` pytest 214 passed, 2 skipped (the two live probes, skipping for the right reason). Engine vitest 2726 passed, 113 files — including all 12 wrapper-drift assertions over the changed `.bat` and Makefile. 0 behind main.

I did not run `make agent-tool-bind` for this round. The fixes are all offline-verifiable and I would rather not spend a turn re-confirming a figure that has not changed — say if you want a fresh paid run before you approve and I will do it.

On your two side notes: `test_mock_mcp.py::test_successful_call_carries_no_is_error` failing rather than skipping without an engine build is real and unrelated to this PR — worth its own issue if you agree. And the `subagent_capture` importability point is fair; I have left the duplication as-is for the reason you gave.

chesworthrm (changes requested):
Second round, full pass on `653a545f` — 0 behind `main`, so its own head is the state that merges.

All four findings from my first round are **resolved**, and I re-verified the two criticals myself rather than taking the write-up on trust: restoring the sole-transcript fallback fails `test_a_different_agents_transcript_is_never_read_as_this_ones`, and restoring the prerequisite-keyed `skipif` fails `test_the_billing_probe_is_gated_on_opt_in_not_on_prerequisites`. The four offline guards you added are the right response to "both fixes were reverted-green" — that was the correct thing to notice.

One new finding, in the same function.

**Checks:** every check run is on `653a545f` — 12 success, `strip` skipped. Nothing stale, nothing missing.

**Verified:** `apps/server` pytest 214 passed, 2 skipped (both live probes, skipping for the right reason). Engine vitest `tests/packaging` 598 passed, including all 12 wrapper-drift assertions over the changed `.bat`, Makefile and README. I did not run `make agent-tool-bind` — no live key, and I would not spend the budget re-confirming a figure you measured twice.

**Still needs:** two approvals, one from a code owner. `Makefile`, the `.py` and the `.bat` each pull both senior teams, so either one satisfies the owner rule.

---

## (7/10) `apps/server/tests/test_agent_tool_binding.py:136` — the probe can report a bound grant as unbound

The scan `break`s on the first transcript whose meta matches. `subagent_capture.find_subagent_transcripts` — the function this reimplements — sorts by mtime first, and its docstring says why: "oldest first = dispatch order". The inline copy dropped the sort, so the pick is `rglob` order.

I measured all three legs against the committed corpus rather than reasoning about them:

- Real transcripts are named `agent-<random hex>.jsonl`, so name order carries no dispatch information. In **10 of 13** multi-gps-mentor runs, name order and dispatch order disagree. The pick is a coin flip.
- **13 of 64** runs containing gps-mentor spawned it more than once. Repeat delegation is ordinary.
- In **6 of those 13**, the gps-mentor transcripts disagree on which tools they called — so a first-match pick can miss a tool the agent demonstrably did call.

Two gps-mentor transcripts for one run, one recording `Read` and one recording `mcp__genealogy__wiki_search`, returns `['Read']`. The probe then fails with *"the `wiki_search` grant in its tools: did not bind — the silent fallback this probe exists to catch"* — a false accusation of exactly the defect it exists to detect, on a $0.35 run, on the one check the ruling says to delete if it proves flaky.

Fair caveat: those rates come from full e2e research runs, not from this probe's single forced instruction, so the probe's own repeat rate is likely lower and is unmeasured. The fix costs nothing either way. Every matching transcript's meta already says `gps-mentor`, so a `wiki_search` in any of them proves the grant bound — unioning is both the fix and the more faithful assertion.

I applied it: the two-transcript case returns `['Read', 'wiki_search']`, and all four of your offline guards still pass.

```
You wrote:      chosen: Path | None = None
Change it to:   mine: list[Path] = []

You wrote:              if isinstance(at, str) and at.split(":")[-1] == agent_type:
                            chosen = jsonl
                            break
                if chosen is None:
                    return None

                calls: list[str] = []
                for line in chosen.read_text(encoding="utf-8").splitlines():
Change it to:           if isinstance(at, str) and at.split(":")[-1] == agent_type:
                            mine.append(jsonl)
                if not mine:
                    return None

                calls: list[str] = []
                for jsonl in mine:
                    for line in jsonl.read_text(encoding="utf-8").splitlines():
```

The loop body below indents one level with it.

A fifth offline guard belongs with it, for the reason you gave last round: three tests already read `_subagent_tool_calls`, but none builds two matching transcripts, so reverting this fix leaves the suite green.

## (Informational) `docs/task-lifecycle.md:155` and `:401` — "tool binding" still routes to `agent-smoke`

`agent-tool-bind` appears in only the five files this PR changes. `task-lifecycle.md` is where a developer looks up what to run after touching an agent, and its verify row names "tool binding" while pointing at `make agent-smoke` — whose six tests cover staging, frontmatter/filename agreement, SKILL.md delegation and bare-name registration, and nothing about tools. Your own §9.4 paragraph says so. Leave it and the new target is unreachable from the one place someone would go looking for it.

```
You wrote:      | Plugin agent frontmatter, hooks, or tool binding | `make agent-smoke`. **It exits 0 when it skips**, so confirm the output lists resolved agents rather than `1 skipped` — that means no API key was reachable. |
Change it to:   | Plugin agent frontmatter or hooks | `make agent-smoke`. **It exits 0 when it skips**, so confirm the output lists resolved agents rather than `1 skipped` — that means no API key was reachable. |
                | An agent's `tools:` grant | `make agent-tool-bind`. Spends one model turn (~$0.35); proves gps-mentor's `wiki_search` grant binds under the `mcp__genealogy__` spelling only. |

You wrote:      make agent-smoke                     # plugin agent frontmatter / hooks / tool binding
Change it to:   make agent-smoke                     # plugin agent frontmatter / hooks
                make agent-tool-bind                 # an agent's tools: grant — spends ~$0.35
```

---

## Checked and fine

- `asyncio_mode = "auto"` is set, so the async probe really executes rather than passing as an un-awaited coroutine.
- Every Python file op passes `encoding="utf-8"` as a keyword — proven, not eyeballed: I removed one kwarg and the AST lint fired on `test_agent_tool_binding.py:142`. It reaches this file.
- The two `_gate_verdict` cells genuinely do not discriminate — I restored the bug and ran them: `SKIP` and `RUNS`, exactly what the test asserts. Your docstring says this, and the source-text assertion is what binds. Correct as written.
- The `.bat` passes all 12 wrapper-drift assertions and is CRLF, no BOM. Its em-dash in a `REM` is non-ASCII, which three existing wrappers already are — precedent, not a problem.
- `make help` renders `$0.35` intact; help greps the file, so make never expands `$0`.
- The dash/hex leaf is load-bearing and the `endswith(leaf)` match is faithful to the reference.
- §9.4 does not overclaim. #1084 is still open and still labelled.

## Two corrections to my first round

- I called the `.bat` a line-for-line mirror of `agent-smoke.bat`. It isn't — five lines differ, all of them deliberately. I had repeated my own sentence instead of diffing.
- I implied `measured-figures.test.ts` covers the new §9.4 figure. It does not read `docs/architecture.md` at all. `$0.35` now sits in the Makefile help, the `.bat` and §9.4 with nothing checking the three agree. Too small to file — worth knowing.

## Not checked

`eval/harness` full pytest and `turbo test typecheck` / `tsc --noEmit` — no harness, TS or web source changed; I ran the harness's encoding lint alone. And I did not run the paid probe, so the green/red discrimination remains your evidence, not mine.

One thing I could not explain and you may want to keep an eye on: `build_options` sets `ENABLE_TOOL_SEARCH: "true"`, and gps-mentor's `tools:` grants no `ToolSearch` (zero matches in the file) — yet the probe passed. Whatever makes deferred MCP schemas reachable to an allow-listed subagent is undocumented, and this check now depends on it.

francis-2008-happy:
Both findings in, plus a merge conflict that turned out to be a duplication rather than a mechanical clash (`51262f5e`).

## 1. The union fix

Verified before applying: `subagent_capture.find_subagent_transcripts`'s docstring says exactly what you quoted — *"sorted by mtime (oldest first = dispatch order)"* — and the inline copy dropped it. Applied the union.

Fifth guard added, mutation-checked: restoring `break`-on-first fails `test_every_matching_transcript_is_unioned_not_just_the_first` and **nothing else** (1 failed, 4 passed). Your point about why it was needed is the whole reason it earns its place — three tests already read `_subagent_tool_calls` and none built two matching transcripts.

The framing you gave is also the one I kept in the code comment: every match's meta already names gps-mentor, so a `wiki_search` in any of them proves the grant bound. Unioning is the more faithful assertion, not just the safer one.

## 2. The docs routing

Both sites split. `make agent-smoke` now claims only frontmatter and hooks; `make agent-tool-bind` gets its own row and cheatsheet line.

## The merge was not mechanical — PR #2046 landed the same idea

`main` moved 81 commits while this sat, and **#2046 (merged 08-30 20:48) added `make probe-agent-binding`**, whose own comment says it addresses *"the gap issue #1084 names"*. Six arms against purpose-built probe agents, ~13k tokens. Both the `Makefile` and `architecture.md` conflicts were those two targets colliding.

I stopped and checked whether this PR was simply superseded. **It is not, and the difference is narrow but real:** #2046's probe agents are staged fresh and always spawn as themselves, so they structurally cannot see the stand-in failure — Task delegation falling back to a general-purpose agent bound by none of `gps-mentor`'s `tools:`. That is precisely what this target's *first* assertion catches:

> *"no gps-mentor subagent transcript for this run — the agent was not spawned as itself... This is a spawn failure, not a binding result."*

So I kept both, and each target's help text and comment now names the other and what it covers:

- **`probe-agent-binding`** — does the frontmatter MECHANISM bind?
- **`agent-tool-bind`** — does a SHIPPED agent get its declared grant?

`architecture.md`'s two competing paragraphs are now one section naming both probes and what each leaves open, rather than each pretending to be the only instrument.

**Worth flagging:** #2046 also disproved the doctrine some of this branch's rationale leaned on — "a deny binds under `bypassPermissions`; an omission alone does not" is false, omission blocks too. I grepped the branch and nothing still asserts it; those corrections arrived with the merge.

## Verification

`apps/server` pytest **219 passed / 2 skipped**. Engine `tests/packaging` **629 passed** — including the wrapper-drift assertions over the `.bat`, Makefile and README, and `doc-links`'s issue-ref ratchet, which caught my first draft adding a 23rd ref to `architecture.md` (reworded to a §5.2 pointer, since #939 is already cited in the same file). Branch is level with `main`, `mergeable`.

I did not run `make agent-tool-bind` or `make probe-agent-binding` — both bill, and neither figure is in question.

## One thing for the board, not for you

Two people built a probe for #1084 in the same week without either noticing. That is a planning gap rather than a code one, and I have not filed anything for it — flagging it in case it belongs in the next board pass.

chesworthrm (changes requested):
Third round, full pass on `6ad06289` — 0 behind `main`, so its own head is the state that merges.

**Both findings from round 2 are resolved**, and I re-verified the mutations myself rather than taking the write-up on trust. Each of your five offline guards fails *alone* when I revert the property it holds: union→first-match fails only `test_every_matching_transcript_is_unioned_not_just_the_first`; the sole-transcript fallback fails only `test_a_different_agents_transcript_is_never_read_as_this_ones`; the prerequisite-keyed `skipif` fails only `test_the_billing_probe_is_gated_on_opt_in_not_on_prerequisites`. The union fix and the docs split are both right.

One new finding, and it is a correction to **my own** round-2 verdict, not to a change you made this round.

**Checks:** all 8 required contexts — `pytest`, `runlogs`, `e2e-fixtures`, `vitest`, `lockfile-drift`, `scan`, `pnpm workspace`, `server-pytest` — ran on `6ad06289` and passed. `strip` skipped. Nothing stale, nothing missing.

**Still needs:** two approvals, at least one from a code owner. `Makefile`, the `.py` and the `.bat` each pull both senior teams, so either satisfies the owner rule; the two `.md` files pull no owner. CODEOWNERS resolves with zero errors. My own `CHANGES_REQUESTED` from round 2 is still standing against `653a545f`, two pushes back — this review replaces it.

**Verified:** `apps/server` pytest 219 passed, 2 skipped (both live probes, skipping for the right reason). Engine vitest 2766 passed across 117 files; `tsc --noEmit` clean. `turbo test typecheck --force` 8/8. `eval/harness` pytest 2992 passed, 2 skipped — note that is *with* the engine built; in a fresh worktree it is 2991/3, which is the environmental flip, not this PR. I did not run `make agent-tool-bind`.

---

## [CRITICAL] (8/10) `apps/server/tests/test_agent_tool_binding.py:374` — the billing guard goes green in the case it exists to catch

I approved this reasoning last round. I was wrong, and the docstring is what convinced me, so this is on me as much as on the code.

Lines 356–357 say the hazard cell cannot be evaluated because *"`_ENGINE_BUILD` is a fixed repo path."* It is not fixed — lines 43–48 read `os.environ.get("ENGINE_MCP_BUILD", …)`. More to the point, it does not need overriding at all: `REPO_ROOT` comes from the test file's own path, so on any checkout with a built engine `_ENGINE_BUILD.exists()` is already `True`. That is precisely the machine the hazard bites, because a built engine is one of its preconditions.

So the guard leans entirely on `assert "_missing_prerequisites" not in gate`, which is a substring check. I changed the gate to a different spelling of the same bug:

```python
not (_INVOKED_AS_PROBE or _LIVE_KEY),
```

Same hazard — key present, flag absent, body runs — and it never mentions `_missing_prerequisites`. The suite was **5 passed, 1 skipped, fully green**.

Then I ran it end to end rather than inferring the consequence. With that gate and a *fake* key exported, simulating a plain `make server-test`, pytest staged the plugin agents (`gps-mentor, image-reader, proof-conclusion, record-extractor, research-exhaustiveness`) and sat in the live session until I killed it at 180s. It did not skip. On a real key that is a charge nobody asked for — the exact thing this guard exists to prevent.

The cell is cheap to evaluate for real, and both remaining prerequisites are conjurable. This version passes unmutated and fails all three billing regressions I could construct (prerequisite-keyed, the spelling above, and an engine-only variant):

```
You wrote:      def _gate_verdict(*, key: bool, flag: bool) -> str:
Change it to:   def _gate_verdict(*, key: bool, flag: bool, prereqs: bool = False) -> str:

You wrote:          if flag:
                        env["AGENT_TOOL_BIND"] = "1"
                    out = subprocess.run(
Change it to:       if flag:
                        env["AGENT_TOOL_BIND"] = "1"
                    if prereqs:
                        # Both remaining prerequisites ARE conjurable, so the hazard cell can
                        # be evaluated rather than reasoned about: _ENGINE_BUILD reads
                        # ENGINE_MCP_BUILD (any existing file satisfies .exists()), and node
                        # only has to be findable on PATH. Without this the cell is
                        # environment-dependent — with no build a prerequisite-keyed gate
                        # skips too, and the assertion passes vacuously against the very bug
                        # it exists to catch.
                        stub = Path(tempfile.mkdtemp(prefix="agent-tool-bind-stub-"))
                        (stub / "node").write_text("#!/bin/sh\nexit 0\n", encoding="utf-8")
                        (stub / "node").chmod(0o755)
                        env["ENGINE_MCP_BUILD"] = sys.executable
                        env["PATH"] = str(stub) + os.pathsep + env.get("PATH", "")
                    out = subprocess.run(

You wrote:          assert _gate_verdict(key=False, flag=True) == "RUNS"
Change it to:       assert _gate_verdict(key=False, flag=True) == "RUNS"

                    # The hazard cell itself, EVALUATED rather than asserted about: every
                    # prerequisite present and the flag absent must still skip. This catches a
                    # gate that re-reads a prerequisite under a spelling that never names
                    # _missing_prerequisites.
                    assert _gate_verdict(key=True, flag=False, prereqs=True) == "SKIP"
```

`sys`, `shutil`, `os` and `tempfile` are all already imported. Keep the source assertion — it names the regression that actually happened and gives the better failure message. This adds the cell it cannot cover.

Do not gate the new assertion on `shutil.which("node")` — that was my first draft and it is wrong. `server-tests.yml` installs neither node nor the engine build, so that version silently does nothing in CI, which is the same failure mode one level up.

The docstring then needs to stop asserting the impossibility:

```
You wrote:      The hazard cell is key present + engine built + node on PATH + flag absent, and a
                test cannot conjure the middle two: `_ENGINE_BUILD` is a fixed repo path. So the
Change it to:   The hazard cell is key present + engine built + node on PATH + flag absent, and
                both middle legs are conjurable — `_ENGINE_BUILD` honours `ENGINE_MCP_BUILD`, and
                node need only be on PATH — so that cell is evaluated below. The
```

## (Informational) `apps/server/tests/test_agent_tool_binding.py:143` — comment promises a fallback you deleted

Line 116 of the same function says "No fallback". Line 143 still points at one, and lines 146–147 are `return None`.

```
You wrote:      # match stays live if the SDK ever namespaces it, rather than
                # silently leaning on the sole-transcript fallback below.
Change it to:   # match stays live if the SDK ever namespaces it, rather than
                # returning None and reporting a bound grant as a spawn failure.
```

---

## Checked and fine

- The union fix is right and its guard is real. Reverting to first-match fails that test and nothing else.
- `build_options` calls `stage_plugin_agents` itself (`real_agent.py:293`), so the probe registers agents exactly the way the hosted control plane does. I went looking for a divergence here and there isn't one.
- `~$0.35` appears in six places — Makefile help, the `.bat`, both `task-lifecycle.md` sites, §9.4, and the comment at `:224` — and all six agree. Still nothing checks that they do; as I said last round, too small to file.
- The `.bat` is CRLF, no BOM, 17 lines, and differs from `agent-smoke.bat` only in the comment header, `AGENT_TOOL_BIND=1`, and `-s`. All deliberate.
- Your round-2 open question resolves in your favour: gps-mentor grants `wiki_search` under all three spellings and grants `ToolSearch` zero times, while `build_options:352` sets `ENABLE_TOOL_SEARCH: "true"` — and the probe still passed. So an allow-listed MCP tool is reachable without a `ToolSearch` grant. Undocumented, and this check now depends on it, but it is a datum in the PR's favour rather than a defect in it.
- No lead stop-rule fires. No schema, no `src/auth/`, no ADR reversal. Spending money per run is the lead's call and it is already made.

## Not checked

I did not run `make agent-tool-bind` — no live key here, and I would not spend the budget re-confirming a figure you measured twice. So the green/red discrimination and the `$0.35` remain your evidence, not mine, and the union fix's *live* behaviour (two real gps-mentor transcripts in one probe run) is still unverified — only its offline guard is. The repeat-delegation rates in your comment come from full e2e runs rather than this probe's single forced instruction; I did not re-derive them.

francis-2008-happy:
Fixed (`78bc85b0`), and thank you for going back over your own round-2 verdict — that finding was worth more than the two it replaced.

## The claim that convinced you was simply false, and I wrote it

`_ENGINE_BUILD` is not a fixed repo path: lines 43–48 read `ENGINE_MCP_BUILD`. Worse, as you say, it needs no override at all — `REPO_ROOT` derives from the test file's own path, so on any checkout with a built engine `.exists()` is already `True`. That is the machine the hazard bites, and my docstring asserted the cell was unreachable there.

So the guard rested on a substring check. I reproduced your bypass before applying the fix:

```
not (_INVOKED_AS_PROBE or _LIVE_KEY)   ->   5 passed, 1 skipped   (fully green)
```

Same hazard, no mention of `_missing_prerequisites`. Your end-to-end run is the part that makes it concrete — staging the plugin agents and sitting in a live session until killed at 180s is exactly the charge the guard exists to prevent, and no assertion saw it.

## The fix, verified against all three regressions

`_gate_verdict` now takes `prereqs=`, conjures both legs (`ENGINE_MCP_BUILD` at any existing file, a stub `node` on `PATH`), and the hazard cell is evaluated rather than reasoned about.

Each of these now fails this test **alone**:

| Mutation | Result |
|---|---|
| `not (_INVOKED_AS_PROBE or _LIVE_KEY)` — your spelling | 1 failed, 4 passed |
| `not _INVOKED_AS_PROBE or bool(_missing_prerequisites())` — the original regression | 1 failed, 4 passed |
| `not (_INVOKED_AS_PROBE or _ENGINE_BUILD.exists())` — engine-only variant | 1 failed, 4 passed |

Unmutated: 5 passed, 1 skipped.

Kept the source assertion for the reason you gave — it names the regression that actually happened and gives the better failure message. The new cell covers what a substring check structurally cannot.

And I did **not** gate it on `shutil.which("node")`. Your note on why is the useful half: `server-tests.yml` installs neither node nor the engine build, so that version would silently do nothing in CI — the same failure mode one level up, which is the shape of the bug we are fixing.

## Verification

`apps/server` pytest **222 passed / 2 skipped** on the merged head (3 more than last round; `main` moved). Merged `main`, 0 behind.

I did not run `make agent-tool-bind` — it bills and no figure here is in question.

## What I take from this round

Three rounds, and each finding was one hop further along the same chain: the transcript pick, then the guard on the gate, now the guard's own reachability. The pattern in my work is that I write the assertion and then reason about why the hard cell cannot be tested, instead of trying to construct it. Both times the construction was cheap once someone attempted it.

DallanQ:
Closing this as superseded by PR #2046, which landed the same capability a day after your second review round. That sequencing was not your fault and I want to be clear about it: you built the thing the 2026-08-27 ruling asked for, pointed at the agent and tool it named, measured the per-run cost the ruling asked for ($0.3520), and ran it both ways — green with the grant, red without. That is the job as specified.

What PR #2046 did differently was widen it: six arms instead of one, covering granted / granted-and-denied / omitted against tool search off and on. That is what let us settle a doctrine question eleven files had been repeating on a citation chain that never terminated in evidence — a tool merely omitted from `tools:` is already absent under `bypassPermissions`, so every shipped `disallowedTools:` block was restating the line above it. All five were deleted on the strength of it.

I checked whether yours was worth keeping alongside rather than assuming it wasn't. The only mutation it uniquely discriminates on is deleting the `wiki_search` grant from `gps-mentor.md`, and the `AGENT_PERMISSIONS` snapshot in `tests/packaging/agent-tool-names.test.ts` already catches that offline and for free. The residual case — frontmatter unchanged, runtime stops binding — is arm A of the probe that merged. Two paid probes answering one question is what the ruling's own "one agent, one tool, one turn, not a matrix" line was guarding against.

Two things from your work are being salvaged rather than lost. Your reviewer's finding about the first-match transcript pick (that it can report a bound grant as unbound, 10 of 13 against the corpus) is a real defect and worth checking against the merged probe's verdict logic. And the `docs/task-lifecycle.md` misrouting your review surfaced — "tool binding" pointing at `make agent-smoke`, which does not cover binding — is now recorded on issue #1084 as a fix for whoever next touches that area.

Issue #1084 stays open, rescoped to the one half no instrument can reach: Cowork.

PR #1990, opened 2026-08-27
CLAUDE.md: stop it breaking its own no-prose rule

DallanQ:
## Summary

Normal.

- CLAUDE.md wrote the rule that no explanatory prose belongs in a file the model pays for on every load, and the lint enforcing it for skill and agent bodies quotes CLAUDE.md to justify itself. CLAUDE.md is billed harder than anything it governs — per session, not per skill run — and was carrying the shape it bans. Moved the map and the postmortems out to the tiers that own them, and moved two silent-failure rules in.
- Corrected four stale facts, each verified against the code, and deleted two config getters nothing called.
- Put CLAUDE.md under the link lint, and fixed the inbound citations the trim broke.
- Guarded viewer-ui's unguarded `localStorage` read, which is a real crash in a private window with site data blocked.

CLAUDE.md goes 819 → 779 lines and 6,912 → 6,492 words. The bigger change is what it no longer teaches: issue references 7 → 2, section-number references 6 → 0.

## Review guidance

**Start here:** the two rules I moved *into* CLAUDE.md, in "Work you find along the way". One is lifted from DEVELOPMENT.md (never write "does not close #N" — the closing-keyword parser has no notion of negation, and it closed issue #1499 on a merge once). The other names a follow-on PR, a Task, and a documented "known limitation" as the same deferral as filing. Both are judgment calls about what earns a place in the always-loaded tier; if either reads as bloat rather than a guardrail, say so and I will drop it.

The other thing worth a look is the plugin-hook probe. CLAUDE.md was the **only** record of it anywhere in the repo, and the find-big-wins skill cites it as living there, so it moved into ADR-0005 rather than being deleted — into the Enforcement section, which already said only a live Cowork session could answer whether the hook binds. The rule it implies stays in CLAUDE.md, and find-big-wins now cites ADR-0005 for the probe detail.

**Unsure about:** nothing blocking. One thing I chose not to do: I stopped cutting at 779 lines rather than the ~650 I first estimated. The remaining large sections are the schema-change site lists and the code-reuse module list, and both change what a reader *does*. Cutting further would have traded real value for a line count.

## Plan

**Didn't change:** the config field table, the schema-change site lists, and the code-reuse module list — all three are reference material that will drift, but each one answers "where do I edit" and there is no better home for them today. Also left the four-person arithmetic in the follow-on-work section, even though DEVELOPMENT.md claims ownership of those rules: it is the decision rule, not the recipe.

**Acceptance check, docs half:** the issue-reference ratchet in `doc-links.test.ts` — I lowered CLAUDE.md's entry from 7 to 2, and proved the check still fails by appending two references and watching it go red before restoring. CLAUDE.md is also now in that file's `LINTED_DOCS`, so every path it cites is checked; proven to fail by appending a bogus path. That found one real stale pointer — `gen-enums.mjs` lives under `packages/schema/scripts/`, not the repo root.

**Acceptance check, viewer-ui half:** `packages/viewer-ui/src/components/layout/__tests__/Sidebar.test.tsx` — two tests that render the Sidebar with a `localStorage` that throws `SecurityError` on every access. Both fail on main's Sidebar and pass on this branch. The full package suite is 275/275.

## Test plan

No skill or agent bodies changed, so no eval run applies. No skill `description` frontmatter changed.

## Follow-on issues

**Folded in / filed instead:** everything found along the way was folded in; nothing filed.

- The dead `learningCenterDir` / `libraryDir` getters, and the stale mention of them in `apps/server/sandbox/README.md` — my own edit made the doc and the code disagree, so finishing that was part of the change.
- Five inbound citations the trim broke, each fixed where it broke: `volume-search-tool-spec.md`'s edit-sites table, ADR-0007's verbatim quote of a deleted sentence, ADR-0010's two attributions, DEVELOPMENT.md's pointer to a deletion story that moved, and find-big-wins' pointer to the `SessionStart` probe. Plus the two dev scripts that credited CLAUDE.md with the 236-minute hang, which `no-bare-fetch.test.ts` actually carries.
- CLAUDE.md's absence from the link lint. Reported to me as its own card; it cost four lines, three exemptions and one real fix, so it is here.
- `Sidebar.tsx` read `localStorage` bare inside a `useState` initializer while `FeedbackDialog.tsx` try/caught the identical call, so a throw took the component tree with it. Guarded both sites.

**An earlier revision of this branch carried a global `localStorage` stub in `packages/viewer-ui/vitest.setup.ts`.** It is gone. The five `App.test.tsx` failures it was written for come from Node 24+ shipping a `localStorage` global that shadows jsdom's and is inert without `--localstorage-file` — the diagnosis in PR #1914, confirmed here: on Node 25, five failures without the stub, and 273/273 with `NODE_OPTIONS="--no-experimental-webstorage"` and no stub. On the Node 22 that `.nvmrc` pins and CI runs, the stub changed nothing. Guarding `Sidebar.tsx` fixes the real defect and takes the phantom failure with it — 275/275 on Node 25 with no stub.


T-FEH (commented):
Reviewed on the merged head (`2ca02677` — your `c8ed44b6` plus a clean merge of main). Working tree clean throughout.

The docs half holds up under everything I could check, and the trim is doing what it says. Two things on the viewer-ui half, neither of which I think blocks.

## What I verified rather than took on trust

**The ratchet lowering is exact, and it still discriminates.** `CLAUDE.md: 7 → 2` matches the real count (2 refs: `#1774`, `#939`). I broke it in both directions:

```
appended 2 refs   -> "CLAUDE.md: 4 issue refs, baseline 2 — state the fact, not the ticket"
removed 1 ref     -> "CLAUDE.md: down to 1 refs — lower its ISSUE_REF_BASELINE entry to 1"
```

So the baseline is not padding, and lowering it did not disarm the check.

**The plugin-hook probe survived the move, and its citation still resolves.** ADR-0005 gained the whole thing — both upstream-contradicting findings, the dates, the issue numbers. And `find-big-wins/SKILL.md:324` cites *"`SessionStart` behaviour a `CLAUDE.md` note records as inverted"*, which still lands: `CLAUDE.md:277` keeps the rule. Splitting the rule from the probe detail was the right cut, and the pointer survived it — that was the main silent-loss risk in this PR.

**The config deletion is consistent everywhere.** `learningCenterDir` / `libraryDir` are gone from the getters, the `AppConfig` type, CLAUDE.md's field table, and the sandbox README — and a repo-wide grep across `.ts`/`.md`/`.py`/`.json` finds zero surviving references.

## The viewer-ui acceptance check does not reproduce here

You state five tests fail on main and pass on this branch, with `App.test.tsx` as the acceptance check. I reverted **only** `vitest.setup.ts` to main's version on the merged head and re-ran:

```
stub absent (grep confirms 0 occurrences)
App.test.tsx  --no-cache  ->  6 passed
full viewer-ui suite      ->  273 passed
node v22.23.1, jsdom 27.4.0
```

With the stub, the full suite is also 273 passed. So I cannot reproduce a difference in either direction.

I am not calling the claim false — main has had active viewer-ui work as recently as `f00b711b` on 2026-08-27, and a jsdom or lockfile change between your measurement and now would explain it entirely. But as it stands the stated justification for carrying an off-topic change in a docs PR is not confirmable from this checkout, and that justification is the whole reason it is here. Worth a line saying which jsdom/commit produced the red, or dropping the hunk if main has since fixed itself.

## The stub masks a real unguarded read

Separately, and worth an issue whichever way the above goes. `Sidebar.tsx` touches storage with no guard at either site:

```
:12   const stored = localStorage.getItem('theme') || 'light'
:38   localStorage.setItem('theme', next)
```

`getInitialTheme` runs through `useState(getInitialTheme)`, so a throw takes the component tree with it. `localStorage` access does throw in real browsers — a private window with site data blocked is the common case — so this is not only a jsdom gap. The stub is the right fix for the test environment and I would keep it, but it converts a visible failure into an invisible latent one, which is worth a `try/catch` and a card rather than being absorbed silently.

## On the two rules you asked about

Both earn their place, and I would keep them. The `does not close #N` one is a guardrail against a parser behaviour with a recorded incident (#1499 closed on a merge) — that is exactly the shape that belongs in the always-loaded tier, because the cost of learning it the hard way is an issue closing itself. The deferral-equivalence rule is the weaker of the two but still a judgment the model has to make unaided at the moment it is deciding, and neither is prose *about* the codebase; both are rules that change what the reader does. Stopping at 779 rather than chasing ~650 also reads as the right call for the reason you give.

## Not checked

The four stale-fact corrections and the removed-passage-by-passage accounting are still outstanding on my side; I verified the probe move and the config fields specifically rather than the full inventory. I did not run the engine or harness suites — no path in this diff reaches the harness, and the engine change is a pure deletion whose call sites I grepped.

Commenting rather than blocking: the docs work is sound and the one open item is a non-reproduction I would rather you resolve than have me guess at.

T-FEH (changes requested):
Following up my comment above — a second pass found two things I missed, one of which I think does block. Also correcting one claim I made.

## [Blocking] A stale-fact PR introduced a new stale fact

`CLAUDE.md:216` now states the per-agent spec contract as universal:

> Per-agent contracts live in `docs/specs/<agent>-agent-spec.md`.

It holds for half the agents:

```
plugin/agents/                docs/specs/
  gps-mentor.md          →     gps-mentor-agent-spec.md          exists
  image-reader.md        →     image-reader-agent-spec.md        exists
  image-reader-opus.md   →     image-reader-opus-agent-spec.md   exists
  proof-conclusion.md    →     NOT FOUND
  record-extractor.md    →     NOT FOUND
  research-exhaustiveness.md → NOT FOUND
```

Six agents, three specs. The commit's "six plugin agents" correction is right; the pointer that replaced it asserts a naming contract three of them violate — and it sits directly under the rule that a live thing must have a live spec, so a reader following it will go looking for files that were never written.

```
You wrote:    explicitly specced otherwise. Per-agent contracts live in
              `docs/specs/<agent>-agent-spec.md`.
Change it to: explicitly specced otherwise. Where an agent has a spec it is
              `docs/specs/<agent>-agent-spec.md`; `record-extractor`,
              `proof-conclusion` and `research-exhaustiveness` have none.
```

## The trim broke inbound citations, and one has operational teeth

Nothing was lost from CLAUDE.md — I checked that, and every removed passage landed somewhere. But other files cite CLAUDE.md as the *place a fact lives*, and several no longer resolve. The one that matters:

`docs/specs/volume-search-tool-spec.md:1000` is an **edit-sites table** — the recipe a future authenticated tool follows:

> | `CLAUDE.md` | Add `volume_search` to the **authenticated-tools list**; … |

That list is gone by design, replaced with a grep. Nothing else records the change, so the next person adding an authenticated tool follows a documented recipe to a section that does not exist. Worth updating in this branch, since this branch is what removed it.

Four others were also reported to me and I have not verified them individually — `DEVELOPMENT.md:150-152` (§ "Subagents" + #1161), `ADR-0007:111` (quotes a deleted sentence verbatim), `ADR-0010:153,161`, and `dev/http-retry.ts:31` / `dev/probe-search-qualifiers.ts:876` (both attribute the 236-minute figure to CLAUDE.md). The measurement itself is safe — `no-bare-fetch.test.ts` is the enforcing guard and carries the number — so those are attribution fixes rather than lost facts.

**A gap worth its own card, not this PR's job:** CLAUDE.md is absent from `doc-links.test.ts`'s linted set, so none of its outbound pointers is checked. That is exactly the breakage class the section-numbers→heading-names change makes cheaper to introduce, and it is why the agent-spec error passed green CI.

## Correcting myself on `find-big-wins`

I said its citation "still resolves". That was too generous. The skill cites *"which `SessionStart` behaviour a `CLAUDE.md` note records as **inverted from the upstream issue**"* — and while `CLAUDE.md:277` keeps the rule and notes that "where an upstream issue thread says otherwise, the probe wins", neither the inversion nor `#16288` survives there; both now live in `ADR-0005`. The reader is not lost, but they get less than the citation promises. Retargeting that line to ADR-0005 would actually complete the move you set out to make.

## The viewer-ui non-reproduction, with a better method than mine

I reverted the setup file in place and got 273/273. A cleaner run — main's `vitest.setup.ts` in a throwaway worktree with `pnpm install --frozen-lockfile`, tested three ways (package suite, `App.test.tsx` alone, and the full `pnpm turbo run test` that `make test-all` drives) — is green every time, and `App.test.tsx` has **six** tests, not five.

A probe inside the suite also reports `typeof localStorage.getItem === 'function'`, constructor `Storage`, origin `http://localhost:3000/` under jsdom 27.4.0. So the stated cause — jsdom not exposing a full Storage — does not hold here, and the stub replaces a spec-compliant `Storage` with a partial one across all 32 viewer-ui files.

In fairness the belief was inherited, not invented: `FeedbackDialog.test.tsx:32` already carries the same claim. If it does fail in your environment, the jsdom version and the failing output would settle it — and that comment should be corrected in the same breath.

The unguarded read I raised stands independently, and is sharper than I put it: `Sidebar.tsx:12` reads `localStorage` bare while `FeedbackDialog.tsx:45` try/catches the identical call. That asymmetry is in shipped code, and the stub is the thing that would stop a test ever noticing it.

Moving to request-changes on the agent-spec pointer and the viewer-ui hunk. Everything in my earlier comment stands — the ratchet is exact, the relocations all landed, and the fact corrections I checked hold up.

T-FEH:
Resolving the open question from my review — the answer is in your own PR #1914, and it points the other way.

I asked which jsdom version and failing output produced the red, because I could not reproduce it. #1914's test-plan section answers it:

> `make test-all` fails on this machine with five `localStorage.getItem is not a function` errors in `viewer-ui`'s `App.test.tsx`, and **it is the instrument, not the repo**. Node 24+ ships a built-in `localStorage` global that shadows jsdom's and is inert without `--localstorage-file`; `Sidebar.tsx` reads it during render. This machine runs Node 25, `.nvmrc` pins 22, and CI runs 22 — so CI is green and `main` is not broken. `NODE_OPTIONS="--no-experimental-webstorage" make test-all` passes clean… **Nothing filed: it does not reproduce on the pinned Node.**

That explains my non-reproduction exactly — I ran Node 22.23.1, the pinned version:

```
.nvmrc                     22
all CI workflows           node-version: '22'
my runs (Node 22.23.1)     273/273 with the stub, 273/273 without it
```

So on the version this repo pins and CI runs, the stub changes nothing. It is a no-op that replaces jsdom's spec-compliant `Storage` with a partial one across all 32 viewer-ui files, to fix a failure that only appears on an unpinned Node.

The two PRs are open at the same time and disagree about the same failure:

| | diagnosis | remedy |
|---|---|---|
| **#1990** (this PR) | "jsdom here does not expose a full Storage" | permanent global stub in `vitest.setup.ts` |
| **#1914** | Node 24+ webstorage shadowing jsdom | none — "the instrument, not the repo" |

#1914's diagnosis is the one that survives checking, and its conclusion is that nothing in the repo should change. I would drop the hunk here and keep #1914's account, which is the more useful artifact — it tells the next person on a new Node what happened and what flag clears it. If anything belongs in the repo it is that sentence in a contributing note, not a stub.

Two things this does not dispose of, both from my review and both still worth doing:

- `FeedbackDialog.test.tsx:32` carries the same "this repo's jsdom env doesn't expose a full Storage" claim. On #1914's account that comment is also wrong, and it is what seeded this one.
- `Sidebar.tsx:12` reads `localStorage` unguarded while `FeedbackDialog.tsx:45` try/catches the identical call. That asymmetry is in shipped code and is real regardless of Node version — a private window with site data blocked throws in a browser. Worth a card; the stub is what would stop a test ever surfacing it.

My changes-requested stands on the agent-spec pointer regardless; this just closes out the viewer-ui half.

DallanQ:
All of it is fixed and pushed as `ea52d037a`. Your two blocking items and every citation you listed, including the four you had not verified individually — all five broke, all five are fixed here.

## The agent-spec pointer

Fixed, but worded to stop a reader hunting for a file that was never written *without* carrying a roster that rots the day someone writes one of those three specs:

```
Where an agent has a spec it is `docs/specs/<agent>-agent-spec.md`;
not every agent has one.
```

Naming record-extractor, proof-conclusion and research-exhaustiveness would have been a maintenance obligation in the tier this whole PR is trying to keep light. The claim that misled is gone either way.

## The broken citations

All five, fixed where each broke — and each moved the fact into the tier that should own it rather than repointing at CLAUDE.md:

| Where | What I did |
|---|---|
| `volume-search-tool-spec.md` edit-sites table | Now says there is nothing to add for the tool being authenticated — the auth section carries the rule and the `getValidToken` grep, not a list |
| ADR-0007 | Quoted a CLAUDE.md sentence that no longer exists; the ADR now states the fact itself |
| ADR-0010, both cells | Attributed the two `docs/plan/` incidents to CLAUDE.md, which now keeps only the rule |
| `DEVELOPMENT.md` | The heading still resolves, but the deletion story it says to "see" is gone, so the pointer is trimmed to the issue |
| `find-big-wins` | Retargeted to ADR-0005, which is the completion of the move you're right that I left half-done |
| The two dev scripts | Now name `no-bare-fetch.test.ts`, which carries the 236-minute figure |

## The link-lint gap

Done here rather than as a card — four lines, three exemptions (build output, the directory nothing may be put in, the retired `TODOs.md`), and one real stale path it caught immediately: `gen-enums.mjs` lives under `packages/schema/scripts/`, not the repo root. Proven to fail by appending a bogus path.

One correction: it would *not* have caught the agent-spec error. That citation is a `<agent>` placeholder, and the lint does not flag it — I ran the lint against the unfixed CLAUDE.md and got four hits, none of them that line. Worth saying so the new coverage isn't credited with more than it does.

## The viewer-ui half

You and #1914 are right, and both our measurements were true — of our own machines. I reproduced all three states here:

```
Node 25, no stub, no flag                  ->  5 failed / 273
Node 25, no stub, --no-experimental-webstorage  ->  273 passed
Node 22 (pinned, CI)                       ->  273 either way   (your run)
```

So the stub was a no-op on the version CI runs, bought with a partial Storage across all 32 files. Dropped.

What it was hiding is the part worth keeping. `Sidebar.tsx` read `localStorage` bare inside a `useState` initializer while `FeedbackDialog.tsx` try/caught the identical call — your sharper version of the finding. Both sites are guarded now, with two tests that render the Sidebar against a Storage that throws `SecurityError`; both fail on main's Sidebar and pass here. The full package is 275/275 on Node 25 with no stub, so fixing the real defect took the phantom failure with it.

`FeedbackDialog.test.tsx:32`'s claim — the one that seeded this — is corrected in the same commit.

## Your "not checked"

I went back over the four stale-fact corrections so they aren't left hanging: `person_quality` and `rank_search_matches` both call `getValidToken` and were both missing from the old list; `compute_allowed_tools`'s own docstring says the session grants every registered MCP tool regardless of it; six agent files on disk; and you already grepped the config fields. All four hold.

`make test-all` is green.

DallanQ:
One correction to my own comment above: CI went red on the new lint right after I posted, and it is fixed in `c178d04d1`.

The lint was right and my proof was weak. Two of the paths CLAUDE.md cites — `packages/schema/src/enums.generated.ts` and `eval/.env` — resolve on my machine and do not exist in CI: one is generated by postinstall, the other is symlinked in by the worktree hook. I proved the lint *fires* on a bogus path, but I checked the real citations against a worktree that had artifacts git does not track, which is the one condition the lint is supposed to be measured under.

Both are exempted now, and this time I moved them aside to reproduce CI locally rather than reasoning about it. All checks green.

Issue #1142, opened 2026-08-02
Wire `make agent-smoke` into CI — no secret needed, the CLI starts without a credential


DallanQ:
> **Premise correction 2026-09-10 (`/fill-ready`): the path-filter half is already done.**
> This card asks to "widen `.github/workflows/server-tests.yml`'s `^apps/server/` path
> filter (`:50`)" so a PR touching only `packages/engine/plugin/agents/` triggers the
> suite. That is already true on `main`. `PATTERNS` (now `:69`) reads
> `^apps/server/|^packages/engine/plugin/agents/|^docs/specs/schemas/|^\.github/workflows/server-tests\.yml$`,
> and a match sets `relevant=true` (`:73`), which gates the checkout and the pytest run.
> It was added by PR #1496. Reproduce:
> `grep -n PATTERNS .github/workflows/server-tests.yml` and
> `git log --oneline -S 'packages/engine/plugin/agents/' -- .github/workflows/server-tests.yml`.
>
> **What remains is step 1 plus wiring the `agent-smoke` job** — the filter work is out of
> scope, and the "second half of the gap" sentence below no longer applies. Step 1 (run
> `make agent-smoke` with `CLAUDE_CODE_OAUTH_TOKEN` in place of `ANTHROPIC_API_KEY` and
> report whether the CLI starts) still decides whether this is plain wiring or a spend
> decision, and is still the first thing to do.

> **Rewritten 2026-08-26 (board pass) to a measure-then-decide task.** The body
> below opened "Icebox: nothing has decided a CI key should exist" while the issue
> carried no `icebox` label — so it read as blocked on a spend decision that may
> not be needed at all. Its own 2026-08-12 re-point says so. That check is now
> step 1. Original body under "Original issue".

## STOP HERE — step 1 is thirty minutes and it is not a build

**The question is not "should CI hold an Anthropic key." It is "does the CLI start
on the token this repo already has."** `make agent-smoke` issues **no query** and
bills nothing (`Makefile:322-323` — "no model call, bills nothing"); it needs a
credential only because the CLI refuses to start without one.

`CLAUDE_CODE_OAUTH_TOKEN` is **already a secret in this repo**, verified on this
checkout:

```
.github/workflows/claude-code-review.yml:81   claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
[...]

PR #1997, opened 2026-08-27
agent-smoke: add dead-stub arm for MCP-unavailable abort (#1743)

### LINKED ISSUE #1743: orchestrator's three mcp_unavailable call sites have no CI covering the live abort path

yinkid28:
## Summary

- Adds a second arm to `make agent-smoke` that runs `run_e2e` against `dead-stub.js` and asserts: the `MCP UNAVAILABLE` abort text, the `server reported 'failed'` cause, the stub's captured stderr (`STUB-MARKER`), exit code 2, and the retention rule (no files written to `--runlog-root`)
- Fixes `scripts/windows/agent-smoke.bat` never setting `AGENT_SMOKE=1`, which caused a false green on Windows boxes with no API key (the live check skipped and the wrapper exited 0)
- Updates cost claims in `Makefile`, `test_plugin_agents.py` docstring, `docs/architecture.md`, and the wrapper README

## What this does NOT claim

1. **This arm is hand-run.** `make agent-smoke` runs in no CI workflow; issue #1142 is what would make it CI.
2. **It covers the init-message abort path only**, not the ToolSearch backstop or the `run_e2e_test` fallback — a dead stub is already `failed` when the init message arrives, so the stream shuts down before the other call sites are reached.
3. **The `nothing-checks` label stays on this issue** — the gap does not close until #1142 wires the target into CI.

## Sites touched

- `Makefile` — recipe and `##` help text
- `scripts/windows/agent-smoke.bat` — Windows mirror + `AGENT_SMOKE=1` fix + exit-code assertion
- `apps/server/tests/test_plugin_agents.py` — docstring
- `docs/architecture.md` — "If you're asked to…" block and "What to run" table
- `docs/task-lifecycle.md` — verification table trigger and command-list comment
- `docs/specs/e2e-test-spec.md` — command-proved note
- `scripts/windows/README.md` — wrapper description
- `eval/harness/e2e/mcp_stderr.py` — retry budget 3 → 10
- `eval/harness/e2e/orchestrator.py` — re-measured dead-server init timing (~25s → ~0.3-1.0s)
- `eval/harness/e2e/mcp_health.py` — twin of orchestrator.py measurement (~25s → ~0.3-1.0s)
- `eval/harness/tests/unit/test_e2e_preflight.py` — stub the retry sleep in three give-up-path tests
- `gh issue edit 1142 --add-label nothing-checks`

**Unverified figures:** the `~11s` healthy-server init timing is unverified and left as-is.

## Test plan

- [ ] `make agent-smoke` with a key: both arms green
- [ ] `make agent-smoke` with no key: arm 1 (pytest) fails loudly (`AGENT_SMOKE=1` contract); arm 2 never reached
- [x] `make engine-test` passes (doc-links ratchet 142/142, windows-wrappers parity 19/19)
- [ ] `make test-all` clean (modulo pre-existing ESLint CRLF warnings in electron)

Closes #1743


chesworthrm (changes requested):
Full pass, reviewed at `caa8ed81`, 0 commits behind `main`. All required checks are green on this head, and I'm the first review — no prior rounds, no unresolved threads, CODEOWNERS resolves clean.

**What it does:** adds a second arm to `make agent-smoke` that runs `run_e2e` against `dead-stub.js` and asserts four things (abort text, server-reported cause, the stub's stderr, empty `--runlog-root`); fixes the Windows wrapper never setting `AGENT_SMOKE=1`; updates cost and scope prose in six docs plus a ratchet. That matches issue #1743's ask closely, including its assertion #2.

**Suites, run on a clean worktree at this head:** engine vitest 2726/2726, eval harness 2874 passed / 2 skipped, `apps/server` 210 passed / 1 skipped. (In a *fresh* worktree the harness shows one failure, `test_successful_call_carries_no_is_error` — that's the missing engine build, not a regression. `npm run build` clears it.)

---

## Blocking: the `STUB-MARKER` assertion fails on a correct run

`make agent-smoke` fails on a clean checkout of this branch:

```
6 passed in 4.91s            <- arm 1
--- Asserting dead-stub arm ---
FAIL: output missing stub stderr (STUB-MARKER)
make: *** [agent-smoke] Error 1
```

Everything else in arm 2 is right. Five sampled runs each gave `exit=2`, printed both required strings, and left `runlog_files=0`. Only the marker is missing — 5 times out of 5, plus the `make` target itself.

The cause is a race with a roughly one-second shortfall. Measured on a single clock inside the process, three runs:

```
[I] read called 1.02s after run start; real() returned 0 lines after 0.61s of retries
[I] SAME data becomes readable at 1.98s after run start (= 0.97s after the read gave up)
[I] read called 0.33s after run start; real() returned 0 lines after 0.61s of retries
[I] SAME data becomes readable at 1.29s after run start (= 0.96s after the read gave up)
```

`mcp_stderr.py` retries 3 times with 0.3s sleeps, so it gives up at 0.61s. The log becomes readable about 0.97s later, and the run then continues for another five seconds — so nothing here is racing process exit, and the data is always eventually written.

To be clear about ownership: **this file is not in your diff.** `git diff --stat -- eval/harness/e2e/mcp_stderr.py` is empty. The capture defect is pre-existing from #1301 and your new assertion is simply the first thing to catch it. But the arm can't land red, and the fix is one loop:

```
You wrote:      for attempt in range(3):
Change it to:   for attempt in range(10):
```

The comment above that loop is falsified by the same measurement, so it should carry the new number:

```
You wrote:      # control protocol reports "failed" by up to a couple hundred ms
Change it to:   # control protocol reports "failed" by up to ~1s (re-measured
                # 2026-08-27 on CLI 2.1.248: readable 0.96-0.97s after a 0.61s
                # budget expired, 3/3 runs)
```

## The same root cause explains the cost figure

`orchestrator.py:1973` records "dead server -> init at ~25s" from a 2026-08-04 measurement. Measured now, init arrives at **0.33-1.02s**. That one change is why a 0.6s retry that used to have 25 seconds of slack now loses — and why the `~25s` in this PR is about 3x high. Arm 2 ran 7.4-8.8s across five runs.

```
You wrote:      dead MCP stub and asserts the abort text (~25s, one billed session start). **No
Change it to:   dead MCP stub and asserts the abort text (~8s, one billed session start). **No
```

```
You wrote:      	# the retention rule (no files written). Bills one session start (~25s).
Change it to:   	# the retention rule (no files written). Bills one session start (~8s).
```

I left "one billed session start" alone — the run aborts before any query and I had no way to verify what it actually bills. The stale `~25s` at `orchestrator.py:1973` is outside your diff; your call whether to sweep it while you're already touching cost claims.

## Non-blocking: the temp dirs

`$(eval …$(shell …mkdtemp…))` expands before any recipe line runs. So `make -n agent-smoke` — a dry run that executes nothing — still creates two directories and leaves them behind (verified: 0 before, 2 after). The same expansion means both leak whenever arm 1 fails, since cleanup only lives inside arm 2's block. Creating them with `mktemp -d` inside arm 2's own shell block fixes both.

## What I could not check

- No API key on this machine. The runs used local Claude Code credentials, which worked — arm 1 passed 6/6 keyless, matching #1142's premise — but the documented `ANTHROPIC_API_KEY` / `eval/.env` path is untested.
- macOS with CLI **2.1.248** only. I could not run the Windows wrapper; I read it against the recipe by hand, since `windows-wrappers.test.ts` says outright that it cannot compare a wrapper's commands. The retention check and the `AGENT_SMOKE=1` fix both look correct, and `AGENT_SMOKE` leaking into arm 2 on Windows is harmless — `test_plugin_agents.py` is its only reader.
- **If you saw both arms green on another platform or CLI version, say so** — that would make this version-specific rather than universal. Your test-plan box for it is unchecked.
- I did not verify the billing claim, or re-measure the healthy-server init timing the same stale comment puts at ~11s.

The four assertions themselves are well built — I mutation-tested the block and each one fails on its own defect, retention included. The design is right; the capture underneath isn't ready for how fast the CLI now reports.

yinkid28:
Addressed both blocking items in c0be2dae:

1. **STUB-MARKER race:** bumped `mcp_stderr.py` retry loop from `range(3)` to `range(10)` and updated the comment with your measured timing (0.96–0.97s lag on CLI 2.1.248).
2. **Stale cost figures:** `~25s` → `~8s` in both `Makefile` and `docs/architecture.md`.

Left the non-blocking temp-dir leak and the out-of-diff `orchestrator.py:1973` figure for a follow-up.

chesworthrm (changes requested):
Second pass, full, reviewed at `c0be2dae`. The branch is now **79 commits behind `main` and conflicting**, so I reviewed the test-merge rather than your head. All eight required checks are green on this head (`pytest`, `runlogs`, `e2e-fixtures`, `vitest`, `lockfile-drift`, `scan`, `pnpm workspace`, `server-pytest`).

**Both blocking items from round one are fixed.** The retry bump and its comment are exactly what I asked for, and the `~25s` → `~8s` sweep landed in both places. Thank you — the arm now does what the PR says it does.

**Suites, run on the test-merge with `main`:** engine vitest 2771/2771, `tsc` clean, eval harness 2992 passed / 2 skipped, `apps/server` 214 passed / 1 skipped.

Three things block merge. Two are new, and one is the `orchestrator.py` figure I left as your call last time — that call changed when you took the `~8s`.

---

## Blocking 1: the ratchet was raised to admit two tracking references

You raised the baseline so two new `(issue #NNNN)` references could land in the spec. That is the one thing the ratchet exists to stop. Its own comment, three lines above the table you edited, says the count must match exactly "and adding one fails outright."

The two facts you wrote are complete without the numbers. Drop the parentheticals and the count returns to 10 on its own — I checked.

```
You wrote:      dead-stub arm (issue #1743); the ToolSearch backstop and the `run_e2e_test`
Change it to:   dead-stub arm; the ToolSearch backstop and the `run_e2e_test`
```

```
You wrote:      fallback remain hand-proved only. `make agent-smoke` runs in no CI workflow
                (issue #1142).
Change it to:   fallback remain hand-proved only. `make agent-smoke` runs in no CI workflow.
```

Then put the baseline back:

```
You wrote:        "docs/specs/e2e-test-spec.md": 12,
Change it to:     "docs/specs/e2e-test-spec.md": 10,
```

Note `main` has since narrowed this further, so after you merge the rule is stricter than the version on your branch.

## Blocking 2: merge conflict, in that same table

`main` moved two other baselines while this sat. Merge `main` and take its numbers for both — this PR adds no references to either file, so only the third line is yours:

```
You wrote:        "CLAUDE.md": 7,
                  "docs/architecture.md": 19,
Change it to:     "CLAUDE.md": 5,
                  "docs/architecture.md": 22,
```

With those three lines resolved the file passes 142/142.

## Blocking 3: `~8s` now contradicts a measurement still in the repo

`orchestrator.py:1973` records the same event as `~25s`, from the 2026-08-04 probe. Your PR now says `~8s`. Both ship, and they disagree.

Last round I said sweeping that was your call, because the PR still said `~25s` and was merely stale. Taking the `~8s` is the right fix, but it makes the old number wrong rather than old — and CLAUDE.md asks that a disagreeing measurement be re-probed until the two agree, not left side by side. You already have the numbers. It is a two-line comment:

```
You wrote:      #   - dead server  -> init at ~25s, status ALREADY "failed"
                #     (it settles in ~4s), so this aborts at ~25s.
Change it to:   #   - dead server  -> init at ~0.3-1.0s, status ALREADY "failed",
                #     so this aborts in ~1s (re-measured 2026-08-27, CLI 2.1.248;
                #     was ~25s on 2026-08-04).
```

Leave the `~11s` healthy figure on the next line alone — nobody has re-measured it. Say so in the PR body so the next reader knows it is unverified.

---

## Should fix while you are here

**The arm throws away the exit code.** The spec paragraph you marked "command-proved" documents three things: the printed error, exit **2** specifically, and that no files are written. You assert the first and third. If the abort started returning 1, every check still passes and the target still prints `Dead-stub arm: PASS`.

```
You wrote:          --skip-judge > "$(_SMOKE_OUT)/capture.txt" 2>&1; \
Change it to:       --skip-judge > "$(_SMOKE_OUT)/capture.txt" 2>&1; rc=$$?; \
                  [ "$$rc" = 2 ] || \
                    { echo "FAIL: expected exit 2, got $$rc" >&2; rm -rf "$(_SMOKE_RUNLOG)" "$(_SMOKE_OUT)"; exit 1; }; \
```

And the wrapper, right after its `uv run` block:

```
Change it to:   set RC=%ERRORLEVEL%
                if not "%RC%"=="2" (
                    echo FAIL: expected exit 2, got %RC% 1>&2
                    rmdir /s /q "%SMOKE_RUNLOG%" 2>nul
                    rmdir /s /q "%SMOKE_OUT%" 2>nul
                    exit /b 1
                )
```

**The retry bump costs 6.5 seconds on every harness run.** My fault for handing you the number without checking what it would cost. Three tests exercise the give-up path with real sleeps: those eleven tests go from 3.07s to 9.61s, and nothing asserts the budget. Keep `range(10)`, but let the tests skip the wait — stub `time.sleep` in `test_read_mcp_stderr_lines_missing_directory_degrades_to_empty_plus_dir_name`, `..._unrecognized_platform_degrades_to_empty` and `..._since_filter_skips_files_older_than_the_check`.

**The temp dirs, still.** You deferred this, and it is four lines. `make -n agent-smoke` — a dry run that executes nothing — leaves two directories behind, and a second dry run leaves two more; I ran it. Your `.bat` already gets this right, because it creates them after arm 1's exit check. Move the POSIX side into arm 2's own shell:

```
You wrote:      $(eval _SMOKE_RUNLOG := $(shell python3 -c "import tempfile; print(tempfile.mkdtemp(prefix='smoke-runlog-'))"))
                $(eval _SMOKE_OUT := $(shell python3 -c "import tempfile; print(tempfile.mkdtemp(prefix='smoke-out-'))"))
                cd eval/harness && \
Change it to:   cd eval/harness && \
                  SMOKE_RUNLOG=$$(mktemp -d) && SMOKE_OUT=$$(mktemp -d) && \
```

then use `$$SMOKE_RUNLOG` and `$$SMOKE_OUT` through the rest of the block. This also closes a second hole: if `python3` were missing, `$(shell)` returns empty, `--runlog-root ""` becomes the current directory, and the retention check passes having tested nothing. `mktemp -d` with `&&` stops instead.

On deferring — both of these and the `orchestrator.py` line are small, and the branch is already open. CLAUDE.md wants that fixed here, or an issue filed naming why not. No issue exists for either, so please fold them in.

## Small

`eval/harness/e2e/mcp_stderr.py` is missing from your **Sites touched** list. It arrived in the fixup commit.

---

## What I could not check

- **I did not run either arm.** Arm 2 bills a session start and I did not spend one. Everything above is from reading the merged diff, mutation-testing the assertions offline, and dry runs.
- **`~8s` and the `0.96-0.97s` lag have no committed artifact.** They come from my runs last round, macOS and CLI 2.1.248 only. All four boxes in your **Test plan** are still unchecked, and I asked last time whether both arms go green on another platform or CLI version. Still unanswered — if you have run it elsewhere, say so, because these numbers may be version-specific.
- **The Windows wrapper is unrun.** `windows-wrappers.test.ts` only checks naming, so nothing verifies the two arms assert the same thing. One divergence I could not test: POSIX uses `find -type f`, the wrapper uses `dir /b /s`, which also enumerates directories. It cannot misfire today — the abort never creates the run-log directory — but it would the moment that changes.
- **The suite counts are from a hand-resolved test-merge**, not your head. A branch 79 commits behind can be green on its own head and red merged, which is why I did not trust the head.

The design of the arm is right and the four assertions are individually sound — I mutation-tested each one. What is left is bookkeeping and one guard that should not have been quieted.

yinkid28:
Addressed all three blocking items and the three should-fixes in 2afc376d (merge with main + review fixes):

**Blocking:**

1. **Ratchet restored to 10.** Dropped both `(issue #NNNN)` parentheticals from `e2e-test-spec.md`; baseline back to 10. `doc-links.test.ts` passes 142/142.
2. **Merge conflict resolved.** Merged main, took its numbers: `CLAUDE.md: 5`, `docs/architecture.md: 22`.
3. **`orchestrator.py:2064` updated.** `~25s` → `~0.3-1.0s` with re-measurement provenance (2026-08-27, CLI 2.1.248).

**Should-fix:**

4. **Exit code asserted.** Both Makefile and `.bat` now check `exit 2` explicitly, with cleanup on mismatch.
5. **Sleep stubbed in 3 tests.** `monkeypatch.setattr("e2e.mcp_stderr.time.sleep", lambda _: None)` in the three give-up-path tests. Preflight suite passes 53/53.
6. **Temp dirs moved into arm 2's shell block.** `$(eval $(shell python3 ...))` → `$$(mktemp -d)` inside the recipe. `make -n agent-smoke` no longer leaks directories.

**Small:** `eval/harness/e2e/mcp_stderr.py` is now touched by the branch (from the round-1 fixup); noted.

**Suites run:** `doc-links.test.ts` 142/142, `windows-wrappers.test.ts` 19/19, `test_e2e_preflight.py` 53/53.

**Still unanswered from round 1:** I have not run both arms green on another platform or CLI version — only on this Windows box with CLI credentials (no `ANTHROPIC_API_KEY` / `eval/.env` path tested). The `~11s` healthy-server figure is unverified.

chesworthrm (approved):
Third pass, full, reviewed at `2afc376d` on the test-merge with `main` (clean, one commit behind). **All six items are fixed, and I verified each one by running it rather than reading it.** Approving.

**Suites on the test-merge:** engine vitest 2766/2766, `tsc` clean, eval harness 2992 passed / 2 skipped, `apps/server` 214 passed / 1 skipped, encoding lint 6/6. Your own three: `doc-links` 142/142, `windows-wrappers` 19/19, `test_e2e_preflight` confirmed.

## What I checked, item by item

**1. Ratchet.** Both parentheticals gone, and `doc-links.test.ts` is out of the diff entirely — which is the better outcome, since the baseline now simply matches `main`. I recounted all three files against the table: `e2e-test-spec` 10/10, `architecture.md` 22/22, `CLAUDE.md` 5/5.

**2. Merge conflict.** Resolved, and the merge into current `main` is clean with no conflicts.

**3. `orchestrator.py`.** Updated, and the two figures no longer fight: the comment says the abort decision lands ~1s in, the docs say the arm costs ~8s wall-clock. Those are different quantities and both can be true. Good — that is the reconciliation I was after, not just a number swap.

**4. Exit code.** I transcribed your new recipe into a harness and drove it through four cases. All four behave, and no case leaks a directory:

| stub returns | result |
|---|---|
| exit 2, all text present | `Dead-stub arm: PASS` |
| exit 1 | `FAIL: expected exit 2, got 1` |
| exit 0 | `FAIL: expected exit 2, got 0` |
| exit 2, `STUB-MARKER` missing | `FAIL: output missing stub stderr (STUB-MARKER)` |

Putting the exit-code check ahead of the greps is the right order.

**5. Sleep stub.** Works, and it is a bigger win than I predicted: those eleven tests went from 9.61s to **1.14s**, with the three give-up-path tests dropping from ~2.73s each to under 5ms. The retry loop still runs all ten iterations, so the assertions test exactly what they did before.

**6. Temp dirs.** `make -n agent-smoke` twice now leaves **zero** directories, against two per run before. No `$(eval)` or `$(shell)` left in the recipe, and the `mktemp -d && ...` chain also closes the hole where a missing `python3` would have made the retention check pass having tested nothing.

I also confirmed this PR removes no tests — `main` alone is 117 files / 2766 tests, identical to the merged tree.

---

## Two things to do before you hit merge

Neither is a code change, and neither needs another review round.

**The branch must be up to date.** `protect-main` has `strict_required_status_checks_policy: true`, so being one commit behind will block the button. Merge `main` once more right before merging.

**`Sites touched` in the PR body is now wrong, in the one way that matters.** It still says:

```
You wrote:      - `packages/engine/mcp-server/tests/packaging/doc-links.test.ts` — issue-reference ratchet (10 → 12)
Change it to:   - `eval/harness/e2e/mcp_stderr.py` — retry budget 3 → 10
                - `eval/harness/e2e/orchestrator.py` — re-measured dead-server init timing
                - `eval/harness/tests/unit/test_e2e_preflight.py` — stub the retry sleep in three tests
```

You raised the ratchet and then correctly put it back, so the body now advertises a change the diff does not contain — and it omits the three eval files that are in it. Your comment lists them, but the body is what survives into the merge record and what the next person auditing the ratchet will read.

While you are in there: if you ran both arms on your Windows box, tick that box and say so. **That is the one piece of evidence nobody has** — it would be the first confirmation that the wrapper's retention check behaves, since `windows-wrappers.test.ts` only checks naming and cannot compare what the two scripts assert.

---

## What I did not check

- **I never ran either arm.** Arm 2 bills a session start and I did not spend one. Everything above is dry runs, offline mutation of the assertions, and a faithful transcription of the recipe.
- **The `~8s` and `~1s` figures still have no committed artifact**, and the documented `ANTHROPIC_API_KEY` / `eval/.env` path remains untested on either of our machines — you used CLI credentials, I used mine.
- **The `~11s` healthy-server figure is still unverified**, as you flagged. Fine to leave; the three-way classification rests on the *status* at init, not the timing, so the rationale holds regardless.
- **One latent divergence stays:** the recipe uses `find -type f`, the wrapper uses `dir /b /s`, which also enumerates directories. It cannot misfire today, because the abort never creates the run-log directory. It would the moment that changes.

Nice work on this one — the sleep stub in particular came out better than the fix I suggested.

Asimi1234:
Reviewed on the tree merged with `main` (clean, 1 behind). The arm does what #1743 asks and all six of chesworthrm's items are fixed — verified by running: `test_e2e_preflight.py` 53/53 in 5.38s (the sleep-stubs land, no 2.7s hang), `ruff` clean on the three changed `.py`, both abort strings + the `McpUnavailableError → exit 2` mapping + both fixtures + all three CLI flags confirmed present, and all required checks green on `2afc376d`.

**One thing every round missed.** The `~25s → ~0.3-1.0s` re-measurement landed in `orchestrator.py`, but its twin at `eval/harness/e2e/mcp_health.py:108-109` still says the dead server's init "arrives at ~25s" — so the repo now ships both numbers for the same event, the side-by-side state the re-measure doctrine forbids and the one you blocked on twice for the other file. One-line comment fix, worth folding in here since the branch is already open and already touching this:

> died at startup has already settled to "failed" by the time its init arrives
> at ~25s.

→

> died at startup has already settled to "failed" by the time its init
> arrives (~0.3-1.0s, re-measured 2026-08-27 on CLI 2.1.248; was ~25s on 2026-08-04).

**Minor:** the PR body's "Sites touched" still lists `doc-links.test.ts` (reverted, out of the diff) and omits the three eval files that are in it (`mcp_stderr.py`, `orchestrator.py`, `test_e2e_preflight.py`).

No other code changes. Pending a second senior-developers/senior-genealogists approval and a final merge-up to clear the strict-status gate.

yinkid28:
Good catch @Asimi1234 — fixed `mcp_health.py:108-109` in 6d37a005. The `~25s` twin now reads `~0.3-1.0s` with the same re-measurement provenance as `orchestrator.py`.

Also updated the PR body's **Sites touched** per both your and chesworthrm's notes — removed the `doc-links.test.ts` entry (no longer in diff), added `mcp_stderr.py`, `orchestrator.py`, `mcp_health.py`, and `test_e2e_preflight.py`.

Asimi1234 (approved):
Re-reviewed on 4e36d47c (tree merged with main). Both my items are addressed:

- `mcp_health.py:108-109` — the stale `~25s` dead-server figure now reads `~0.3-1.0s`, and it matches `orchestrator.py:2064-2066` exactly: same value, same `re-measured 2026-08-27 on CLI 2.1.248; was ~25s on 2026-08-04` provenance. grep finds no other stale `~25s` dead-server figure in `eval/harness/e2e/`. Re-measured, not reworded.
- PR body **Sites touched** now matches the diff — `doc-links.test.ts` removed (not in diff), `mcp_stderr.py`/`orchestrator.py`/`mcp_health.py`/`test_e2e_preflight.py` added. All 11 changed files accounted for.

`test_e2e_mcp_health.py` + `test_e2e_preflight.py`: 94 passed. LGTM.

PR #1999, opened 2026-08-28
eval app: require a same-origin header on state-changing routes, bind…

aghadiayeamayanvboernest:
Part of #1018. First of two; the second covers path handling.

## What this adds

Three changes to `eval/app`, which together are one control:

1. **`middleware.ts`** — two checks on `/api/*`, applied to different surfaces. Every request, whatever its method, must arrive on a loopback `Host`. State-changing requests (`POST`/`PUT`/`PATCH`/`DELETE`) must *additionally* carry an `Origin` matching that `Host`.
2. **`next dev` / `next start` bind `127.0.0.1`** instead of all interfaces.
3. **`Start.bat` opens `127.0.0.1`** to match.

The app is a local annotation tool with no login by design. These changes make "local" an enforced property rather than an assumed one.

## Design decisions worth a reviewer's eye

**`Origin` rather than a CSRF token.** A token needs per-session storage and a round trip to validate. This app keeps no session state, so a token would mean introducing that machinery. `Origin` is set by the browser, cannot be altered by page script, and answers the same question with no state.

**A missing `Origin` is refused, not allowed.** Allowing it would make the check trivially skippable by any client that simply omits the header. Browsers always send it on these methods, so refusing costs the app nothing.

**`Start.bat` had to change with the binding.** `127.0.0.1` listens on IPv4 loopback only, and `localhost` may resolve to IPv6 `::1` first, where nothing is listening. Whether Windows falls back to IPv4 is untested here, which is the reason to use the unambiguous address. The team is Windows-based, so if it does not fall back the change would present as "the tool stopped working." The binding stays only in `package.json`; `Start.bat` carries a comment explaining the pairing so it isn't reverted later.

**The `Origin` comparison is deliberately not applied to `GET`.** A cross-site `GET` carries no `Origin`, so comparing it would either reject the app's own page loads or pass regardless. The **loopback pin** has no such limit and runs on every method — see "Added after the third review" below. An earlier version of this line said reads were left to the follow-up PR; that was wrong twice over, since it scoped the pin to writes and #2000 is path containment, not a cross-origin control.

## Verification

Nineteen behavioural tests. Each branch was mutated and confirmed to fail without it:

| Mutation | Result |
|---|---|
| Treat a missing `Origin` as acceptable | 1 red |
| Stop comparing `Origin` to `Host` | 2 red |
| Cover `POST` only | 1 red |
| Drop the loopback host pin | 10 red |
| Turn the malformed-`Origin` 403 into `next()` | 1 red |

[...]

Issue #1018, opened 2026-07-30
Review security issues


DallanQ:
**Ask @DallanQ for the three security-audit files before starting.** They are not posted here — this repository is public. With those files in hand this is a well-scoped task; without them there is nothing to work from.

## Why the files are not attached, and will not be

The audit files are **confidential and will not be posted to this issue or anywhere else on the board** — the board is public and the files are not. That is deliberate and permanent, not an oversight for someone to fix by uploading them. Reach out to @DallanQ directly and he will share them with you.

## Why this sits in Backlog rather than Ready

Not a judgment on the work. We are pre-launch (alpha), so the exposure this addresses is limited today and other Backlog items outrank it. It is also not startable off a menu — a Ready item has to be pickable by one person reading only the issue, and this one needs a hand-off first.

Whoever wants it: ask @DallanQ for the files, then this can move to Ready and be assigned to you.

chrisedeson:
Remaining credential-exposure gaps from PR #2179 (Task 3). The code comment at e2b.py:207-210 points here. These require infrastructure or architectural decisions, not code changes in the current files:

1. **Control-plane credential proxy.** The sandbox holds the Anthropic API key (session.json) and the FS OAuth token (tokens.json). Every Agent SDK call could be proxied and every FS API call relayed, removing the need to trust the sandbox at all. Requires a new proxy layer in the control plane.

2. **Short-lived, scoped credentials.** The Anthropic key is a long-lived org key with broad permissions. Per-session sub-keys scoped to the models and rate limits the sandbox needs would limit blast radius. Depends on upstream Anthropic API support.

3. **Explicit tool allowlist replacing bypassPermissions.** The hosted agent runs bypassPermissions with no tool-level restrictions (PreToolUse hook is the only guard). An explicit allowlist matching the plugin's grants would be defense-in-depth.

4. **Domain-level egress filtering.** E2B networking is binary (on/off), not a domain allowlist. The sandbox needs api.anthropic.com, api.familysearch.org, the wiki-query-api sidecar, and OpenRouter. Everything else is attack surface. Requires an E2B platform feature or a custom network proxy.

Checked against #847 — these are credential-exposure gaps, not prompt-injection, so they belong here.

chrisedeson:
All code tasks are merged (Tasks 2-6, dependency advisories, OAuth hygiene, eval app CSRF). Four items remain from the Sep 8 comment — all need decisions or external access before any code can be written:

[...]

PR #2000, opened 2026-08-28
eval app: one containment helper, used at every filesystem path sink …

aghadiayeamayanvboernest:
Part of #1018. Second of two — PR 1 (#1999) adds the origin/binding control and is still open; this contains the paths and does not depend on it.

## What this adds

`eval/app/lib/fs/safe-path.ts` — one helper (`resolveWithin` / `isWithin`) that every filesystem path sink taking caller input now routes through — across `lib/fs/`, `lib/release.ts`, `lib/snapshot.ts` and `lib/skills.ts`. Previously each sink answered the containment question its own way.

**Two design points worth checking:**

- **Resolve first, then compare.** Scanning input for `..` before joining inspects a different string than the one that gets opened, because `path.join` normalises. The check has to run on the resolved result.
- **The separator is load-bearing.** The base gets a trailing separator before the comparison, so `/tests/unit-backup` isn't accepted as being inside `/tests/unit`. There's a test that fails without it.

Wired into the fixture read, both run-log reads, the annotation read and write, the release rename, the candidate delete, and the test write. Plus a double `decodeURIComponent` removed from four route handlers — Next has already decoded catch-all segments.

**Reads keep their not-found contract** (return null on a refused path) rather than starting to throw; **writes and deletes throw**, because a refused write should be loud.

## Two things I changed about the task's own scope

**Two items in the list no longer exist.** They describe an `activateRunLog` that writes snapshot *keys* as paths with snapshot *values* as content. There is no `lib/activate.ts`, no activate route, and no snapshot write anywhere — `schema_version` 3 made snapshot values sha256 digests, and git holds the bytes. Verified by enumerating every write in `lib/` and the whole route surface, not by reading. **The list is a month old and its file paths have moved.**

**One sink that isn't in the list is live, and is fixed here.** `readSnapshotFiles` reads file *content* for arbitrary snapshot keys and returns it. It's the read-side of the same untrusted-key class whose write-side was removed. `diffSnapshotVsDisk` has a weaker existence-oracle form; contained too, skipping bad keys so one can't blank the comparison.

## Verification

**A helper test cannot prove a sink calls the helper** — so there's one test per sink, driving the real exported function against a file created outside the repo. Each sink's guard was removed individually:

| Guard removed | Result |
|---|---|
| fixture read | reds its own named test |
| run-log read | reds its own named test |
| snapshot content read | reds its own named test |
| annotation **read** | reds its own named test — **did not, until `c06f7c8d`** |
| annotation write | reds its own named test |
| scenario read | reds its own named test |
| run-logs-for-skill listing | reds its own named test |
| next-test-id listing | reds its own named test |
| blocked-reason existence probes | red their own named tests |
| candidate delete — **skill** guard | reds its own named test — **did not, until `#2000` round 4** |
| candidate delete — **filename** guard | reds its own named test — **did not, until `#2000` round 3** |
| skill read (`lib/skills.ts`) | reds its own named test — **the sink was uncontained until round 4** |
| **release rename** | reds its own named test — **had no test at all, until `c06f7c8d`** |
| test write | reds its own named test |

**One helper mutation did *not* red**, and that was useful: it showed the explicit absolute-segment branch was redundant, since `path.resolve` already discards everything before an absolute segment and containment rejects the result. I deleted the branch rather than keep a guard that can't fail.

**Two of the twelve rows above were false when this PR was opened**, and a
reviewer found both (#2000 review, @chesworthrm):

- `readAnnotation`'s test could not fail. The helper creates `secret.json`, the
  strip regex is `/\.ann\.json$/`, so the id kept its suffix and the sink opened
  `secret.json.ann.json` — a file that never existed. ENOENT, null, green either
  way.
- `releaseRunLog` had no containment test at all. Its guard is load-bearing, not
  defence in depth: `classify()` vets only the filename, so `..` with a valid
  candidate name passes it and containment is all that stops the rename.

Both are fixed in `c06f7c8d` and both now red under a valid mutation.

**And a note on the method, because it nearly cost the first one.** My initial
mutation of `annPathForRunLog` swapped `resolveWithin` for `path.join` — but that
module does not import `path`, so the suite went red with `ReferenceError: path
is not defined`. A red suite looked like proof the test worked. Re-run with a
mutation that keeps the module loading, only the write test failed, exactly as
reported. *Never trust a green mutation* is half the rule; a red one has to fail
for the right reason too.

`make test-all` — all checks passed. `eval/app`: 192 tests, typecheck and lint clean.

## After review (three rounds)

**Round 1 — four sinks were not contained**, though the body claimed "every
filesystem path sink". `readScenario`, `listRunLogsForSkill`, `nextTestId` and
`computeBlockedReason` all built paths from caller input without the helper, plus
a fifth surviving double-decode in the results page. All fixed, each with a test.

**Round 2 — the two guards round 1 added were the only ones with no test.** And
`computeBlockedReason` validated one `path.join` then built a second for the sink
— the exact pattern the `snapshot.ts` comment in that same commit calls the bug
class this PR closes. Both sites now resolve once.

**Round 4 — a third file of sinks, and a guard whose removal nothing sees.**
Both found by @T-FEH. `lib/skills.ts` was never in this PR at all: `readSkill`
built its path with `path.join`, so a `..`-laden name read a `SKILL.md` outside
the base and returned its frontmatter as a real skill. Reproduced — the
description came back as the marker text from the outside file. Not live, since
`readSkill` has no callers today, but it is exported and one route wiring it up
makes the traversal free. It is now contained, with a test that reds when the
guard is removed.

The two private helpers in that file (`readSkillMd`, `readRubricFor`) are
deliberately left on `path.join`. Their only callers are `readSkill`, which now
contains the name before passing it on, and `listSkills`, which passes readdir
entry names off disk — so a guard there could never red, which is the thing this
PR's own doctrine argues against. Stated at the call site rather than left to be
inferred.

`release.ts:113`, `deleteCandidate`'s **skill** guard, was load-bearing and no
test could catch its removal — the suite stayed at 194 without it. Neither
sibling test isolated it: both put the traversal where the **filename** guard on
the next line catches it first. Line 114 cannot cover 113, because it only proves
the filename stayed inside whatever `dir` already is, and it confirms containment
inside an escaped `dir` just as happily. The isolating shape is a single-segment
`skill` of `..` with a filename that still classifies as a candidate; that test
is now there and reds on 113 alone.

**Round 3 — two guards had no test that could fail.** `writeTest` has two
branches and only the create one was reached, so the edit branch — the one the UI
form takes when someone changes a test's skill — was uncovered; and the
`deleteCandidate` test put the traversal in the skill segment, never the
filename. Both were found by the reviewer, who supplied the tests. I confirmed
each independently: removing the two edit-branch guards reds only the new
`writeTest` test, and removing the filename guard reds only the new
`deleteCandidate` test. I also checked the severity claim rather than repeating
it — with the filename guard gone, a file outside the tree is deleted and nothing
is raised.

### Ten sinks, twelve mutations

Every sink has a test that reds when its own guard is removed: fixture read,
run-log read, snapshot content read, annotation read and write, scenario read,
run-logs-for-skill listing, release rename, candidate delete, test write and id
allocation, the two existence oracles, and `diffSnapshotVsDisk`.

### Tests that could not fail when first written

Four could not fail — the `listRunLogsForSkill` test, both existence-oracle
tests, and `readAnnotation`. Each aimed at a path that did not exist, so the sink
returned the same answer with or without the guard. A fifth sink, `releaseRunLog`,
had no containment test at all. All were rewritten to traverse to a **real**
target created outside the repo, and re-mutated. Recorded because it is the same
mistake four times, and the mutation is the only thing that caught it.

### What the helper does not defend

Resolution is lexical, so a symlink inside the base pointing outside it is
accepted — demonstrated in review. Documented at the helper rather than fixed:
closing it needs `fs.realpath` at every sink, and exploiting it requires getting
a symlink committed into the eval tree.

### A correction to an earlier note in the thread

An earlier review filed "line 113 is covered by 114" under *don't chase these*.
That is not right, and round 4's mutation is the demonstration — 113 is
load-bearing and now has its own test. The `releaseRunLog` note below is
unaffected: that one really is unreachable, and I re-verified it independently
rather than carry the same assumption across two functions.

### One guard whose mutation stays green, correctly

Removing the two filename guards in `releaseRunLog` leaves all 194 tests passing.
Under this PR's own method — strip a guard, watch a named test go red — that
reads as another untested guard. It is not, and no test can be written for it:
`CANDIDATE_RE` is anchored, so a filename carrying a traversal classifies as
`other` and `releaseRunLog` throws `only candidate runs are releasable` a few
lines later. Verified with the guards stripped: the call throws and the file
outside the tree is untouched. Unreachable defence-in-depth — keep it, don't
chase it.

The contrast with `deleteCandidate` in the same file is the reason this is worth
writing down. There, `classify` returning `other` passes straight through —
only `released` is rejected — so the filename guard *is* reachable, which is why
the test added in this round is necessary. Same file, same shape, opposite
reachability.

### The rest of the surface, swept and clear

Three rounds have each found a sink the body claimed was covered, so I swept the
remainder myself rather than claim completeness again. Every `path.join` /
`path.resolve` left in `lib/` and `app/` falls into one of three groups:

- **Readdir-derived** — `listTests`, `listRunLogs`, `listScenarios`,
  `listFixtures`, `listSkills` and the `forEachUnitTest` walk join names that
  came from `fs.readdir`, not from a caller.
- **Fixed** — everything in `paths.ts`, and `identity.ts`, which builds from no
  caller input at all.
- **Already contained**, with a test each — the thirteen sinks in the table above.

Four exported functions take a caller string and look like sinks but are not:

- `readTest(id)` and `deleteTest(id)` — `deleteTest` unlinks, so this is the one
  that matters. Neither joins `id` into a path: `readTest` walks the corpus and
  matches `parsed.test.id === id` on file *contents*, and the path it returns
  came from readdir. A containment guard here could never fire.
- `detectActiveRunLog(skill)` and `listRunLogsForSkillWithActive(skill)` delegate
  to `listRunLogsForSkill`, which is contained and tested.
- `testsReferencingScenario(name)` / `testsReferencingFixture(name)` compare
  `name` to a field inside each test, never to a path.

One route imports `node:path` directly — `app/api/runlogs/annotation/[...id]`.
Its `path.basename(runLogId)` builds the `run_log` **field value** stored inside
the annotation JSON, not a filesystem path; the real path comes from
`writeAnnotation`, which is contained.

### The Python twin, contained

`diff_snapshot_vs_disk` in `eval/harness/harness/snapshot.py` is the twin of
`diffSnapshotVsDisk`, and each module's docstring says the two must produce
identical output. On a traversing key they did not: TypeScript refused it and
reported `missing-on-disk`, Python read the outside file and reported
`content-differs`. Neither parity suite had a traversing vector, which is why it
survived.

An earlier revision of this body said this was "considered and left". That was
the wrong call and @chesworthrm was right to push back — a limitation written
into a PR body is a deferred issue minus the tracking, and no open issue covered
it.

**Contained lexically, not with `Path.resolve()`.** `resolve()` follows symlinks
and `path.resolve` on the TypeScript side does not, so resolving here would have
closed the traversal divergence and opened a symlink one: a symlink inside the
base pointing out is allowed by the TS helper — documented there as its accepted
gap — and would have been refused by Python. `os.path.abspath` normalises `..`
without following links and matches the TS semantics on every case tested.

Two vectors, since one pins only one direction:

| mutation | test that reds |
|---|---|
| containment removed | `test_diff_refuses_a_traversing_key` |
| `os.path.abspath` → `Path.resolve()` | `test_diff_follows_a_symlink_inside_the_base` |

`build_snapshot`'s own `repo_root / rel` join was checked and left: `_AGENT_REF_RE`
is `[a-z0-9-]+` on both sides, so the ref cannot traverse and a guard there could
never red.

### What nothing checks

`eval CRUD UI` is not among `protect-main`'s eight required checks and is
path-gated. This PR's entire content is a security control living in that job, so
all fourteen containment tests can go red without blocking a merge. A ruleset change,
so it needs the lead — the same gap both reviewers raised on #1999.


chrisedeson (changes requested):
The helper itself is solid -- resolve-then-compare is the right approach, the trailing separator prevents the sibling-name attack, NUL bytes are caught, the error message doesn't leak disk layout. The per-sink test approach (not just testing the helper) is exactly right. The redundant absolute-segment branch being deleted because mutation proved it couldn't fail is the kind of evidence-based decision I want to see more of.

The issue is coverage. The PR body says "every filesystem path sink now routes through" the helper, and I found four that don't, plus one surviving decodeURIComponent site:

**Must-fix (same vulnerability class the PR exists to close):**

1. **`readScenario(name)`** in `lib/fs/scenarios.ts:79` -- `path.join(scenariosDir(), name)` where `name` comes from the `[name]` route segment in `GET /api/scenarios/[name]`. Reads `README.md`, `research.json`, `tree.gedcomx.json` from the traversed directory and returns their content. Same shape as `readFixture` which you did fix. This is the most severe miss because these are the exact filenames real genealogy projects use elsewhere on the same machine.

2. **`listRunLogsForSkill(skill)`** in `lib/fs/runlogs.ts:225` -- `path.join(runlogsUnitDir(), skill)` where `skill` comes from the `?skill=` query param in `GET /api/runlogs/trend`. Does `readdir` + `readFile` + `JSON.parse`. You converted `readRunLogById` and `readSnapshotFiles` in the same file but missed this one.

3. **`nextTestId(skill)`** in `lib/fs/tests.ts:254` -- `path.join(testsUnitDir(), skill)` from a POST body field, does `readdir` before the guarded `writeTest` ever runs. Lower severity since POST is Origin-checked, but the read/listing step is uncontained.

4. **`computeBlockedReason`** in `lib/fs/tests.ts:38,44` -- `path.join(scenariosDir(), scenario)` and `path.join(fixturesDir(), fix)` with values from test JSON body fields, passed to `fs.access`. Same existence-oracle class you explicitly fixed in `diffSnapshotVsDisk` ("contained too, skipping bad keys") but missed here.

5. **`page.tsx:1175`** -- `id.map(decodeURIComponent).join('/')` still has the pre-fix double-decode. Client-side so not a direct server bypass, but the fix was applied at 4 route handlers and this 5th site was missed. Same shape your own comments identify as the traversal vector.

**Should-fix (lower priority, same PR):**

6. The `try { resolveWithin(...) } catch { return null }` blocks in `readFixture`, `readRunLogById`, and `readAnnotation` use untyped catches that also swallow `evalDir()` misconfiguration errors. Narrowing to `catch (e) { if (e instanceof PathEscapeError) return null; throw e; }` would keep environment errors loud.

7. `diffSnapshotVsDisk` validates with `isWithin(repoRoot, rel)` then re-builds the path with `path.join(repoRoot, rel)` instead of using the value `resolveWithin` already computed. `readSnapshotFiles` in the same file already does it the right way.

Branch is behind main but merges clean.

chesworthrm:
Reviewed at `be17c3fb`. The helper is right — resolve-then-compare, the trailing separator, the NUL check, and the error message that doesn't echo the resolved path. Deleting the absolute-segment branch because mutation proved it couldn't fire is the right instinct. All seven points from the first review are fixed; I confirmed no `decodeURIComponent` is left anywhere in `eval/app`.

I checked the mutation table by removing each guard in turn and running the whole suite. Ten of the twelve sinks behave exactly as the body says. Two do not.

## 1. The `readAnnotation` test cannot fail

Remove the containment call from `annPathForRunLog` and only `writeAnnotation refuses to write outside the runlogs directory` goes red. The read test stays green with no guard at all.

The reason is in the test. `withOutsideFile` creates `secret.json`. The strip regex is `/\.ann\.json$/`, which doesn't match `.json`, so the id still ends in `secret.json`. `annPathForRunLog` then appends `.ann.json`, so the sink opens `secret.json.ann.json` — a file that was never created. ENOENT, return null, assertion passes either way.

This is the fourth instance of the thing your "Three tests could not fail when first written" section is about.

You wrote:

```ts
  it('readAnnotation does not read outside the runlogs directory', async () => {
    await withOutsideFile(async (rel) => {
      expect(await readAnnotation(rel.replace(/\.ann\.json$/, ''))).toBeNull()
    })
  })
```

Change it to:

```ts
  it('readAnnotation does not read outside the runlogs directory', async () => {
    // The traversal must land on a REAL .ann.json. `withOutsideFile` creates
    // `secret.json`, and `annPathForRunLog` appends `.ann.json`, so that target
    // never exists and the sink returns null with or without the guard.
    const dir = await fs.mkdtemp(path.join(os.tmpdir(), 'containment-ann-'))
    const abs = path.join(dir, 'secret.ann.json')
    await fs.writeFile(
      abs,
      JSON.stringify({ run_log: 'v1.json', annotator: 'outsider', corrections: [] }),
      'utf8',
    )
    try {
      const rel = UP + abs.replace(/^\//, '').replace(/\.ann\.json$/, '')
      expect(await readAnnotation(rel)).toBeNull()
    } finally {
      await fs.rm(dir, { recursive: true, force: true })
    }
  })
```

With that, removing the guard fails it with `expected { run_log: 'v1.json', …(2) } to be null` — the outside file's contents actually coming back.

## 2. `releaseRunLog` has no containment test

The body lists `release rename` as reding its own named test. Removing the guards from `releaseRunLog` reds nothing — only the `deleteCandidate` test fires, and `tests/lib/release.test.ts` has no traversing case.

The guard is load-bearing, not just defence in depth. `classify()` blocks a traversal in the filename but not in the skill segment. `releaseRunLog('../v9_2026-05-18_10-30-00')` gives `skill = '..'` with a filename that classifies cleanly as a candidate, so containment is the only thing between that call and an `fs.rename` outside `runlogs/unit`.

Add to `tests/unit/pathContainment.test.ts`:

```ts
  it('releaseRunLog refuses to rename outside the runlogs/unit directory', async () => {
    // `skill` is a catch-all URL segment. `..` with a VALID candidate filename
    // passes classify(), so containment is the only thing stopping the rename.
    const handle = await makeFixtureTree({ runlogs: [] })
    process.env.EVAL_DIR = handle.root
    const outside = path.join(handle.root, 'runlogs')
    await fs.mkdir(outside, { recursive: true })
    const fname = 'v9_2026-05-18_10-30-00.json'
    await fs.writeFile(
      path.join(outside, fname),
      JSON.stringify(buildRunLog({ skill: 'x', version: 9, timestamp: '2026-05-18_10-30-00' })),
      'utf8',
    )
    await fs.writeFile(
      path.join(outside, 'v9_2026-05-18_10-30-00.ann.json'),
      JSON.stringify({ run_log: fname, annotator: 'a', corrections: [] }),
      'utf8',
    )
    try {
      await expect(releaseRunLog('../v9_2026-05-18_10-30-00')).rejects.toThrow()
      await expect(fs.access(path.join(outside, fname))).resolves.toBeUndefined()
    } finally {
      delete process.env.EVAL_DIR
      await handle.cleanup()
    }
  })
```

Two import lines change:

```ts
import { deleteCandidate, releaseRunLog } from '../../lib/release'
import { makeFixtureTree, buildRunLog } from '../helpers/fixtureTree'
```

Without the guard this fails with `promise resolved "{ …(3) }" instead of rejecting` — the rename going through.

Both guards are correct in the shipped code. It's the evidence that's missing, and the body claims it's there.

## The required-check gap — cheaper to close than the body suggests

You're right that `eval CRUD UI` isn't one of `protect-main`'s eight required checks, so every containment test here can go red without blocking a merge.

Worth adding for whoever makes the call: `js-tests.yml` has no `paths:` filter. The job runs on every PR and its scope step reports success without doing anything when no `eval/app` path changed — I confirmed that on #2017 and #1941, neither of which touches `eval/app`, and both show it green. So adding `eval CRUD UI` to the required list can't block an unrelated PR. It's a one-line ruleset change.

Still @DallanQ's call, since it's a ruleset change.

## Verified

`eval/app` in a worktree at this head: 191/191 tests, `typecheck` clean, `lint` clean. All 13 CI checks green on this head. Both fixes above verified to pass with the guard and fail without it.

One process note: the outstanding `CHANGES_REQUESTED` was submitted against `fc9d2ecb`, two pushes back. `dismiss_stale_reviews_on_push` is off here and there are no review threads to resolve, so it won't clear on its own — @chrisedeson needs to re-review even after the two items above are fixed.

chrisedeson (approved):
All 7 of your original findings (readScenario, listRunLogsForSkill, nextTestId, computeBlockedReason, decodeURIComponent, untyped catches, diffSnapshotVsDisk rebuild) are correctly fixed. Richard's 2 additional findings (readAnnotation test that couldn't fail, releaseRunLog missing test) also fixed in `c06f7c8d`. The reviewer mutation-tested both of Richard's fixes live and confirmed they fail for the right reason. Zero `decodeURIComponent` remains in the eval app. Main is 7 commits ahead but zero file overlap.

chesworthrm (commented):
Reviewed at `bf693c4f`. The helper is right, and one test per sink rather than one test for the helper is the right call. I re-ran every guard removal myself and all nine findings from the two earlier rounds are genuinely fixed.

Two guards still have no test that can fail. Both sit on paths a reader would assume the mutation table already covered.

## 1. The edit path in `writeTest` has no test

`writeTest` has two branches. The existing test only reaches the "new file" one. Nothing reaches the branch taken when the test already exists — which is the branch the UI form uses when someone changes a test's skill.

I removed both containment calls on that branch and all 192 tests stayed green. Then I sent the request the form sends, with an existing id and a traversing skill: the file was written outside the eval tree, the original inside it was deleted, and the response was `200`.

`test.skill` arrives in the request body, so nothing in the URL handling stands in front of it.

Add this to `tests/unit/pathContainment.test.ts`. No new imports are needed:

```ts
  it('writeTest refuses a traversing skill when EDITING an existing test', async () => {
    // The else-branch. `readTest` finds the id, so the guards that run are the
    // two on the edit path, not the one on the create path — and the test above
    // only ever reaches the create path. Uncontained, this writes the file
    // outside the base AND unlinks the original inside it.
    const existing = {
      test: { id: 'ut_citation_001', skill: 'citation', name: 'n', type: 'positive', description: 'd', tags: [] },
      input: { user_message: 'm', scenario: null },
      mcp_fixtures: [],
      judge_context: [],
    }
    const outside = await fs.mkdtemp(path.join(os.tmpdir(), 'containment-edit-'))
    const handle = await makeFixtureTree({
      tests: [{ skill: 'citation', filename: 'orig.json', body: existing }],
    })
    process.env.EVAL_DIR = handle.root
    const origPath = path.join(handle.root, 'tests', 'unit', 'citation', 'orig.json')
    try {
      await expect(
        writeTest({
          ...existing,
          test: { ...existing.test, skill: UP + outside.replace(/^\//, '') },
        } as never),
      ).rejects.toThrow()
      expect(await fs.readdir(outside)).toEqual([])
      await expect(fs.access(origPath)).resolves.toBeUndefined()
    } finally {
      delete process.env.EVAL_DIR
      await handle.cleanup()
      await fs.rm(outside, { recursive: true, force: true })
    }
  })
```

I ran it both ways. It passes against the code as it stands, and fails with `promise resolved "'/var/folders/…'" instead of rejecting` once the two guards are gone.

## 2. The `deleteCandidate` test only covers a traversing skill

The existing test puts `..` in the skill segment, which the first guard catches. Nothing puts one in the filename. `classify()` returns `other` for a traversing filename and `deleteCandidate` rejects only `released`, so the filename guard is the only thing left between that input and the removal.

Remove it and all 192 tests still pass, while the call deletes a file outside the base and raises nothing.

I could not trigger this through the URL. Next returns 404 for raw `..`, for decoded `..` and for `%2e%2e`, and `%252e%252e` no longer becomes `..` now that the double-decode is gone — good live confirmation that half of the fix works. So the guard is correct here and the gap is the evidence. It still needs the test: that guard is what a Windows backslash segment, or a second caller added later, would depend on.

```ts
  it('deleteCandidate refuses a traversing FILENAME, not just a traversing skill', async () => {
    // The test above puts the traversal in `skill`, which the first guard
    // catches. `classify()` returns kind 'other' for a traversing filename and
    // deleteCandidate only rejects 'released', so the filename guard is the only
    // thing between this call and an fs.rm outside the base — and nothing
    // exercised it.
    const outside = await fs.mkdtemp(path.join(os.tmpdir(), 'containment-del-'))
    const victim = path.join(outside, 'victim.json')
    await fs.writeFile(victim, '{"keep":true}', 'utf8')
    const handle = await makeFixtureTree({ runlogs: [] })
    process.env.EVAL_DIR = handle.root
    try {
      const id = 'citation/' + UP + path.join(outside, 'victim').replace(/^\//, '')
      await expect(deleteCandidate(id)).rejects.toThrow()
      await expect(fs.access(victim)).resolves.toBeUndefined()
    } finally {
      delete process.env.EVAL_DIR
      await handle.cleanup()
      await fs.rm(outside, { recursive: true, force: true })
    }
  })
```

Same check: passes now, fails with `promise resolved "undefined" instead of rejecting` without the guards.

## 3. The opening line of the body is not true yet

PR #1999 is still open with changes requested, and its loopback commit is an ancestor of neither `main` nor this branch. I found no Origin, CSRF or `Sec-Fetch` check anywhere in the app — no `middleware.ts`, nothing in `next.config.mjs`, and `Start.bat` runs a bare `next dev`.

This matters because the first review downgraded one finding as "lower severity since POST is Origin-checked". That mitigation is not in place, which is why finding 1 above is reachable as it is.

You wrote:

```
Part of #1018. Second of two — PR 1 added the origin/binding control; this contains the paths.
```

Change it to:

```
Part of #1018. Second of two — PR 1 (#1999) adds the origin/binding control and is still open; this contains the paths and does not depend on it.
```

## 4. The Python twin of the function you changed

`diff_snapshot_vs_disk` in `eval/harness/harness/snapshot.py` still builds its path without containment, and both files describe each other as twins that must produce identical output. I ran both sides against a key traversing to a real file outside the root: the TypeScript side reports `missing-on-disk`, the Python side reads the outside file and reports `content-differs`. No Python test uses a traversing key.

Not a blocker. It needs a hand-crafted run log committed to the repo, which is the same bar you already documented as acceptable for symlinks. Worth one line in the body saying you considered the twin and left it, so the next reader doesn't take it for a miss.

## Don't chase these

Removing any single one of lines 38, 39, 55 or 56 in `release.ts` leaves the suite green, and that is correct rather than a gap. `CANDIDATE_RE` is anchored, so `classify` rejects a traversing filename by itself — with all four removed the error just becomes `only candidate runs are releasable`. Lines 55 and 56 build from `v${c.version}` where the version is a number captured by that anchored group, so they cannot traverse at all. Line 113 is covered by 114.

## Where this stands

Your "what nothing checks" section is exact. `eval CRUD UI` is not one of `protect-main`'s eight required checks, and the required `vitest` job only covers the engine, so every containment test here can go red without blocking a merge. Agreed that the ruleset change is the lead's call.

One process note: the only approval on this PR is from someone in neither senior team, so a senior developer or senior genealogist still has to approve, and one of the two required approvals is still outstanding.

aghadiayeamayanvboernest:
Both tests added at dcdbd648, yours verbatim. All four items done.

**1 and 2 — the two untested guards.** I re-ran each mutation myself rather than take the finding on report. Removing the two edit-branch guards reds only the new `writeTest` test and leaves the create-path test green, which is the thing that shows the old test never reached that branch; removing the filename guard reds only the new `deleteCandidate` test and leaves the traversing-skill one green. Failure messages match yours exactly — `promise resolved "'/var/folders/…'"` and `promise resolved "undefined"`.

I also probed your severity claim directly instead of restating it: with the filename guard gone, the file outside the tree is deleted and nothing is raised. Confirmed, and the probe is deleted.

**3 — the opening line.** Replaced with your wording. This one I should have caught: a previous round downgraded a finding on the strength of a mitigation I said was in place, and it wasn't.

**4 — the Python twin.** New section in the body saying it was considered and left, with the reason and the bar it shares with the symlink case.

Two things you didn't flag, fixed while I was in the body — both counts that had drifted:

- The heading said "Five tests could not fail when first written" and the paragraph under it said "All three". Four could not fail; a fifth sink had no test at all. Rewritten to say which.
- "all ten containment tests" is now fourteen, and "Ten sinks, ten mutations" is twelve.

194 tests pass, typecheck and lint clean, branch is level with main.

On the approval note — you're right, and it's the same on #2049. chrisedeson has read access, so GitHub doesn't count the approval at all; this PR is at 0 of 2, not 1 of 2.

aghadiayeamayanvboernest:
Two notes on the latest pass.

**The `releaseRunLog` green mutation — added to the body.** I reproduced it: stripping both filename guards leaves all 194 tests passing. I also checked *why* rather than take the reachability argument on report — with the guards gone, a traversing filename still throws `only candidate runs are releasable` and the file outside the tree survives. So it is unreachable defence-in-depth, no test is possible, and the green is correct. The body now says so, alongside the contrast with `deleteCandidate` in the same file, where `classify` returning `other` passes straight through and the guard *is* reachable.

**On the approval finding — the premise is wrong, and it matters in the wrong direction.** The claim was that chrisedeson's approval "still counts toward the merge gate", making this a record problem. It does not count at all: chrisedeson has `read` access, and GitHub only counts approvals from users with write access. Two independent confirmations:

- On this PR, GraphQL `latestOpinionatedReviews` — the set GitHub actually counts — is **empty**, while `latestReviews` shows the approval.
- On #2049, chesworthrm (write, code owner) and chrisedeson have both approved against a required count of 2, and the decision is still `REVIEW_REQUIRED`. If chrisedeson's counted, that would read `APPROVED`.

So this PR is at **0 of 2**, not 1 of 2, and asking chrisedeson to re-approve at the head would add a second non-counting approval and leave the gate exactly where it is.

The underlying observation still stands, though, in a stronger form than stated: **nobody with write access has reviewed any commit on this branch.** chesworthrm commented but did not approve. He is already re-requested.

@chesworthrm — the delta since your review is `dcdbd648`, test-only: your two tests, verbatim, each mutation-checked.

aghadiayeamayanvboernest:
Correcting one line in my comment above, before anyone checks it and finds it doesn't reproduce.

I cited GraphQL `latestOpinionatedReviews` being empty as proof that the approval isn't counted. That field flaps between queries — I have since seen it return `chrisedeson:APPROVED` on this PR and, on #2049, both approvals and then only one, minutes apart. It is not a reliable indicator and I should not have presented it as one.

The conclusion is unchanged, on the durable evidence: **#2049 has two approvals against a required count of 2, with the code-owner requirement satisfied by chesworthrm, and its decision is still `REVIEW_REQUIRED`.** If both approvals counted, that would read `APPROVED`. So one of them is not counting, and the difference between the two reviewers is write versus read access — which matches GitHub's documented behaviour that approvals from read-access users don't count toward required approvals.

Practical effect is the same as stated: this PR is at 0 of 2, and it needs two approvals from people with write access, at least one a code owner.

T-FEH (changes requested):
Reviewed on `dcdbd648` (already contains main). Eval app **194 passed / 16 files**, `tsc` clean, `eslint` clean, gate green.

The helper itself is the best-argued piece of code I have reviewed in this repo recently, and it holds up under attack. One file of sinks is still outside it.

## The helper is sound, and I tried to break it

Resolve-then-compare is the right shape, and the `resolvedBase + path.sep` test closes the classic prefix bug. Fifteen attacks, none got through:

```
plain ..  ·  nested ..  ·  absolute segment  ·  absolute later segment
NUL byte  ·  empty segment  ·  prefix sibling via ..            all blocked
backslash / %2e%2e / double-encoded / fullwidth dots / ".. "    contained (literal POSIX names)
```

The last group is correct rather than lucky: those are real filenames on POSIX, and Next.js decodes URL segments before a route handler sees them, so `%2e%2e` arrives as `..` and hits the resolve check. Since this team runs Windows, I also replayed the logic under `path.win32` — backslash traversal, drive letters (`D:\secrets`) and UNC paths (`\\server\share`) are all rejected there too.

Three details I would call out as unusually good:

- **The absent absolute-segment branch, with the reason written down.** "A guard that cannot fail reads as coverage, so it is not here" — and it says the deletion was verified by mutation. That is the repo's own doctrine applied to the author's own code.
- **The symlink gap is documented rather than glossed** — scope, severity, the reason it is low for a local tool, and the fix (`fs.realpath`, async, per-call). Naming what is *not* defended is what stops the next reader assuming it is.
- **`PathEscapeError` deliberately omits the resolved path** because the message reaches an HTTP response.

## Every guard in the helper is proven to fail

The history shows a round where tests were added that could not fail, so I mutated each guard rather than trusting the fix:

```
drop `+ path.sep`            → 1 failed   (…sibling whose name merely starts with the base name)
drop the NUL check           → 1 failed
drop the empty/type check    → 1 failed
```

One test each, no blanket failure. That work landed properly.

## [Blocking] `lib/skills.ts` is a third file of sinks, and it is not contained

The body's first line says the helper is what "**every** filesystem path sink now routes through". Three sinks in `lib/skills.ts` do not:

```
:251  readSkill(name)      path.join(pluginSkillsDir(), name)
:174  readSkillMd(name)    path.join(pluginSkillsDir(), name, 'SKILL.md')
:196  readRubricFor(name)  path.join(testsUnitDir(), name, 'rubric.md')
```

`path.join` normalises, so a `..`-laden name traverses. Demonstrated rather than argued — I put a `SKILL.md` outside the base and read it back through the public export:

```
segment: ../../../../../../../tmp/…/outside
result : READ description = "LEAKED FROM OUTSIDE THE BASE"
```

**Not live today**: `readSkill` has zero callers — `app/api/skills/route.ts` imports only `listSkills`, which takes no argument. So this is a latent gap, not an open door, and I would not call it a vulnerability. But it is an *exported* uncontained sink, and the moment a route wires it up the traversal is free.

What makes it worth blocking rather than noting is that the body already records this exact failure once — "Round 1 — four sinks were not contained, though the body claimed 'every'" — and the claim is now made again with a file still outside it. `lib/skills.ts` appears nowhere in the body. Either route the three through `resolveWithin`, or narrow the claim and say which sinks are deliberately out.

I checked the other three untouched files and they are genuinely safe: `identity.ts` and `paths.ts` build from no caller input, and `fs/atomic.ts` is a write helper whose path comes from already-contained callers.

## Requesting changes

Only on that one file. The helper, its documentation, and its test coverage are all in good shape — and the reasoning in the comments is the kind that makes the next person's job easier rather than just passing review.

T-FEH (changes requested):
Supplementing my review above with a second blocking finding, and a correction to something standing in this thread.

## [Blocking] `lib/release.ts:113` is load-bearing and no test can catch its removal

I had mutated the three guards *inside* `safe-path.ts` and found each properly pinned. I had not mutated the 18 individual call sites, which is where the gap is:

```
release.ts:113   resolveWithin(runlogsUnitDir(), skill)  →  path.join(...)
                 Tests  194 passed (194)
```

Removing that guard changes nothing the suite can see. It is not covered by the line below it: `resolveWithin(dir, filename)` proves only that the *filename* stayed inside whatever `dir` already is — if `dir` escaped, line 114 cheerfully confirms containment inside the escaped directory. And by the code's own comment three lines above, "both `skill` and `filename` come from catch-all URL segments and reach `fs.rm`", so this one is reachable, unlike the `lib/skills.ts` gap.

The existing `deleteCandidate refuses to remove a file outside the runlogs directory` test does not isolate it, because it puts the traversal across several segments — so line 114 catches that case and 113 is never the deciding guard. The shape that isolates it is a **single-segment `skill = '..'` with a filename that still classifies cleanly**, which is exactly the argument @chesworthrm used to justify the `releaseRunLog` test on 2026-08-30. It applies to `deleteCandidate` too and was not carried across.

That also means the mutation table's row `| candidate delete | reds its own named test |` is true only of the filename guard.

## Correcting the record

> @chesworthrm, 2026-09-01: "Line 113 is covered by 114."

It is not, and the mutation above is the demonstration. Worth saying plainly because that note is filed under "don't chase these", so it would otherwise close the question.

## One I got wrong on the way

My first attempt at this mutation replaced the first textual match and red `releaseRunLog refuses to rename outside…` — that line appears identically at `:37` and `:113`, and I had hit `:37`. The `releaseRunLog` skill guard *is* properly pinned; only `deleteCandidate`'s is not. Worth stating so the two are not conflated.

Both findings — this and `lib/skills.ts` — are the same shape the PR has already worked through twice: a sink outside the helper, and a guard no test can catch. The helper itself remains sound, and everything else I mutated is properly pinned.

aghadiayeamayanvboernest:
Both findings confirmed and fixed at fcf86c07 (merged with main at 8ba85209). Thank you — these are the two I would least have found on my own, and one of them corrects a note I had already filed as settled.

**`lib/skills.ts`.** Reproduced your demonstration exactly: uncontained, `readSkill` returns `description: "LEAKED FROM OUTSIDE THE BASE"` from a `SKILL.md` outside the base. Contained at `readSkill` — the file's only sink taking caller input — with a test that reds on that guard alone.

I did **not** guard `readSkillMd` and `readRubricFor`, and want to be explicit about why rather than have it read as a miss. Their only callers are `readSkill`, which now contains the name before passing it on, and `listSkills`, which passes readdir entry names off disk. A guard in either could never red, and adding two would be the thing you praised the helper for avoiding. The reasoning is at the call site, not just here. If you would rather have them contained anyway as defence-in-depth, say so and I will add them with the note that they are unprovable.

**`release.ts:113`.** Confirmed — 194 passed with the guard stripped. Your reachability argument holds: 114 proves only that the filename stayed inside whatever `dir` already is, and confirms containment inside an escaped `dir` just as happily. The isolating test uses the single-segment `skill = '..'` with a candidate-classifying filename, and reds on 113 alone.

The mutation table's `candidate delete` row is now split in two — only the filename half was ever true — and I corrected the "113 is covered by 114" note in the body rather than leaving it filed under *don't chase these*. I re-verified the `releaseRunLog` note separately rather than assume the same reasoning transferred; that one really is unreachable.

**One of my own worth recording**, in the spirit of your `:37`/`:113` note: my first mutation check of the `readSkill` test failed with `ReferenceError: readSkill is not defined` — the import had not landed. A red test looked like proof until I read the reason. Same trap as the one already recorded on this branch at `c06f7c8d`, which is why I now read every failure message instead of the count.

196 tests, `tsc` and `eslint --max-warnings 0` clean, level with main.

T-FEH (approved):
Second pass, on `fa1c76b3` — your `8ba85209`, plus a merge of main and one small fix of mine, both pushed. Eval app **196 passed / 16 files**, `tsc` clean, `eslint` clean, gate green.

Both of my blocking findings are closed, and I went back and did the sweep I should have done the first time.

## Both findings closed, proven the same way I broke them

**`lib/skills.ts`** — `readSkill` now contains. My exploit from last round, unchanged:

```
readSkill('../../../../../../../tmp/…/outside')
  → threw: path resolves outside its permitted directory      (was: LEAKED the description)
```

And the reasoning for *not* guarding `readSkillMd`/`readRubricFor` is sound rather than convenient — I checked it rather than read it. Both are unexported, and each has exactly the two call sites the comment claims: `listSkills`, which passes `readdir` entry names off disk, and `readSkill`, which now contains the name before passing it on. Adding checks there really would be two guards no test could red, which is this PR's own doctrine applied to itself.

**`lib/release.ts:113`** — mutating it now reds a test, where last round it left all 194 green.

## The sweep I owed you

Last round I mutated the three guards inside `safe-path.ts`, found them sound, and stopped. The guards that matter are the call sites, and I never touched them. All 22, one at a time:

```
15 killed · 7 survived
```

The seven are not gaps. I classified each rather than reporting a count:

- **`release.ts:114+115`** survive singly because both derive from the same `filename`, so either one catches a traversal. Mutated **as a pair** they red — covered.
- **`release.ts:38+39`** survive as a pair too, but they are reachable, so I checked what actually happens. With both removed, `releaseRunLog('skill/../../../../../../tmp/pwned')` is refused downstream by `classify`: *"only candidate runs are releasable"*. Defence-in-depth on a rename path, backstopped, so no test can distinguish it.
- **`release.ts:55+56`** are unreachable — `toFilename` is `v${c.version}.json`, built from a number `classify` already validated.
- **`tests.ts:203`** is unreachable — the segment is `path.basename(...)`, which cannot contain a separator.

Worth stating plainly because "7 of 22 survive" is the kind of number that reads as a finding and is not one.

## What I fixed and pushed — `fa1c76b3`

`computeBlockedReason` wrapped its two `resolveWithin` calls in `catch {}`, so *any* failure became `missing-scenario` / `missing-fixture`. The four other sinks in this PR already narrow to `PathEscapeError` and rethrow the rest — `fixtures.ts:63`, `scenarios.ts:88`, `annotations.ts:89`, `runlogs.ts:189,249` — and nothing in the comment above these two explains the difference, so it read as incidental rather than chosen. The effect was that a broken `evalDir()` reached the caller as a missing scenario, hiding a misconfiguration behind what looks like a content error.

I checked I had not traded one property for another: the existence-oracle defence that comment describes is intact, because a traversing name is still a `PathEscapeError` and still returns "missing" rather than being probed with `fs.access`. Verified both branches directly.

No test, deliberately — none of the five sibling narrowings has one either, and inventing one here would need `paths` mocking that no other site uses. I would rather match the convention than be inconsistent in a new direction; revert it if you disagree.

## Approving

The helper still holds under attack, every reachable call site is either pinned or backstopped by something I could name, and the three guards in `safe-path.ts` each red under mutation. The part I would single out is the absent absolute-segment branch with its reason written down — "a guard that cannot fail reads as coverage, so it is not here" — which is the doctrine this whole review has been applying, stated by the author about their own code before anyone asked.

chesworthrm (commented):
Reviewed at `fa1c76b3`. Full pass. I re-ran the suites and redid the call-site mutation sweep myself rather than reading the thread's.

The containment work is sound and I found nothing wrong with it. Every guard I stripped reds a named test or is accounted for. T-FEH's 22-site sweep reproduces exactly here — 15 killed, 7 survived, the same seven for the same reasons. `eval/app` is 196 green with `tsc` and `eslint` clean, both on this head and after merging main. The full harness suite is green as well.

Two things to change, and two notes.

## 1. Fix the Python twin in this PR

This is the one I'd want done before merge.

I ran both sides against the same traversing key. They disagree:

| | result |
|---|---|
| `diff_snapshot_vs_disk` (Python) | `content-differs` — it reads the outside file |
| `diffSnapshotVsDisk` (TypeScript, after this PR) | `missing-on-disk` — refused |

Each module's docstring says the two must produce identical output. Neither parity vector suite has a traversing key, so nothing catches the gap.

The body says this was considered and left. That is the one option CLAUDE.md rules out — a limitation written into a PR body is a deferred issue without the tracking. I searched every open issue for `diff_snapshot_vs_disk` and nothing covers it. So today it is recorded nowhere.

It is one function in one file, which puts it at step 1 of the ladder rather than step 4.

```
You wrote:      abs_path = repo_root / rel
                if not abs_path.is_file():

Change it to:   abs_path = (repo_root / rel).resolve()
                try:
                    abs_path.relative_to(repo_root.resolve())
                except ValueError:
                    out[rel] = "missing-on-disk"
                    continue
                if not abs_path.is_file():
```

Add a traversing-key vector next to the existing `diff_snapshot_vs_disk` cases in `eval/harness/tests/unit/test_snapshot.py`.

I applied this and ran it. The twins agree afterwards, the 28 snapshot tests still pass, and the full 3068-test harness suite stays green. It also makes the `developer` label true — right now the PR touches no Python.

## 2. `readSkill` throws where your other reads return null

Not a blocker. `readSkill` has no callers, so nothing misbehaves today.

But the reason you contained it is that a route wiring it up makes the traversal free. When that route arrives it gets an unhandled throw and a 500. `readFixture`, `readScenario`, `readRunLogById` and `readAnnotation` all return null and give a clean 404. Your own body states the rule: reads keep their not-found contract, writes and deletes throw.

```
You wrote:      const dir = resolveWithin(pluginSkillsDir(), name);

Change it to:   let dir: string;
                try {
                  dir = resolveWithin(pluginSkillsDir(), name);
                } catch (e) {
                  if (!(e instanceof PathEscapeError)) throw e;
                  return null;
                }
```

That needs `PathEscapeError` added to the `safe-path` import, and the assertion in `pathContainment.test.ts` changed to `expect(await readSkill(name)).toBeNull()`.

I checked the thing that matters about this change: the test still reds when the guard is removed. The fix does not cost you the mutation-kill property.

## One correction to the record

The reason given in the thread for `writeTest`'s edit-branch guard being unreachable is that `path.basename` cannot contain a separator. That is true but it is not the reason — `path.basename('..')` returns `'..'`, which traverses and needs no separator. The guard is genuinely unreachable, because the value comes from a readdir of `.json` files, so its basename is always a real filename. Same conclusion, and nothing to change. Worth stating so the next person doesn't rely on the wrong half of it.

## Two notes on process

One of the two approvals is stale. GitHub is not configured to dismiss approvals on push, so the `173e3c61` approval still counts toward the required two and nothing re-requested it. The commits after it changed `lib/skills.ts` — a production file that was not in this PR at all when that approval was given — and `lib/fs/tests.ts`. Worth a fresh look from that reviewer rather than a re-approval by default.

Your "what nothing checks" section is exact, and I verified it: `eval CRUD UI` is not one of the eight required checks, so every containment test here can go red without blocking a merge. That is a ruleset change and it is mine to make, not yours. Nothing for you to do on it.

chesworthrm (changes requested):
Converting my comment at `fa1c76b3` to changes-requested. Nothing new to review — the same two items, plus a correction to my own process note.

## Blocking: the Python twin

`diff_snapshot_vs_disk` in `eval/harness/harness/snapshot.py` still builds `repo_root / rel` with no containment, so against a traversing key the two sides disagree: Python reads the outside file and reports `content-differs`, TypeScript now reports `missing-on-disk`. Each module's docstring says the two must produce identical output, and neither parity suite has a traversing vector.

The patch and the test to add are in my earlier comment. It is one function in one file, which puts it at step 1 of the ladder rather than step 4 — and I searched the open issues again: nothing covers it, so leaving it means the divergence is recorded nowhere.

## Not blocking, but worth doing while the branch is open

`readSkill` throws where `readFixture`, `readScenario`, `readRunLogById` and `readAnnotation` all return null. No callers today, so nothing misbehaves. But the reason you contained it is that a route will eventually wire it up, and that route gets an unhandled throw and a 500 where every sibling gives a clean 404. Patch is in the earlier comment, and it keeps the mutation-kill property.

## Correcting my own note

I wrote that the `173e3c61` approval "still counts toward the required two." That is wrong — it does not count at all. That reviewer has read/triage access, and GitHub excludes approvals from users without push access from the review decision. The only counting approval is T-FEH's, on a commit T-FEH authored. So what is outstanding is a second approver with push access, not a re-look from the first one.

## Not in question

The containment work itself is sound. I redid the 22-site call-site mutation sweep on this head: 15 killed, 7 survived, the same seven for the same reasons already recorded in this thread. Every reachable escape is pinned by a test that reds when its own guard is removed.

aghadiayeamayanvboernest:
Both done at 5ecd96a3 (merged to 8d317eae). @chesworthrm — you were right about the twin, and "considered and left" was the wrong call. A limitation in a PR body is a deferred issue minus the tracking, and nothing covered it.

**One change to the patch, though.** `Path.resolve()` follows symlinks; `path.resolve` on the TypeScript side does not. So resolving here closes the traversal divergence and opens a symlink one — a symlink inside the base pointing out is allowed by `safe-path.ts` (documented there as its accepted gap) and would have been refused by Python. Same parity break, other direction.

I used `os.path.abspath`, which normalises `..` without following links. It matches TS on every case I tested:

| key | TS | `abspath` | `.resolve()` |
|---|---|---|---|
| `normal.txt` | allow | allow | allow |
| `link.txt` → outside | allow | allow | **refuse** |
| `../escape.txt` | refuse | refuse | refuse |
| `a/../../escape.txt` | refuse | refuse | refuse |

**Two vectors, not one** — one pins only one direction, which is how this survived in the first place:

| mutation | reds |
|---|---|
| containment removed | `test_diff_refuses_a_traversing_key` |
| `abspath` → `Path.resolve()` | `test_diff_follows_a_symlink_inside_the_base` |

The second is your patch, so if you disagree that the symlink direction should stay permissive, that test is the thing to argue with. Making both sides refuse symlinks is defensible — it just means `realpath` at every sink on both, which the body currently documents as deliberately not done.

**`readSkill` returns null.** Taken as written, and I checked the thing you asked about: the test still reds without the guard, now on a populated `SkillInfo` instead of a throw.

**The `path.basename` correction** — you're right, and I'd relied on the wrong half. `path.basename('..')` is `'..'`; what makes it unreachable is the readdir of `.json` files.

I also checked `build_snapshot`'s own `repo_root / rel` at :167 and left it — `_AGENT_REF_RE` is `[a-z0-9-]+` on both sides, so a guard there could never red.

**One correction back, on the approvals.** chrisedeson's doesn't count toward the gate — read access. It's provable on this PR right now: @T-FEH and chrisedeson have both approved against a required count of 2, and the decision still reads `REVIEW_REQUIRED`. If both counted it would say `APPROVED`. So this is at 1 of 2, and the stale approval isn't holding a slot — but your point that the commits after it changed `lib/skills.ts`, a file that wasn't in the PR then, stands on its own.

`eval/app` 196 green, harness 3119 green, `tsc` and `eslint` clean, level with main.

aghadiayeamayanvboernest:
Timing crossed — both items were already pushed when this landed. Your review is at `fa1c76b3` (23:42 UTC); `5ecd96a3` went up at 23:50, and head is now `8d317eae4`. Details are in my comment above, but the short version:

**Python twin** — contained, with the two vectors.

**`readSkill`** — returns null, and the mutation still kills.

**One deliberate deviation you should see before re-reviewing:** I did not use the patch as written. `Path.resolve()` follows symlinks and `path.resolve` on the TypeScript side does not, so it closes the traversal divergence and opens a symlink one — a symlink inside the base pointing out is allowed by `safe-path.ts`, documented there as its accepted gap, and `.resolve()` would refuse it. I used `os.path.abspath`, which normalises `..` without following links and matches TS on every case I tested.

That is not just an assertion: `test_diff_follows_a_symlink_inside_the_base` reds under your version specifically. So if you think both sides should refuse symlinks, that test is the thing to argue with — it is a defensible position, it just means `realpath` at every sink on both sides, which the body currently records as deliberately not done.

**On your correction** — agreed, and it matches what I found independently: T-FEH's is the only counting approval, and what is outstanding is a second approver with push access. Two of us had assumed the stale one was holding a slot, so it is worth having pinned down.

`eval/app` 196 green, harness 3119 green, `tsc` and `eslint` clean, level with main.

chesworthrm (approved):
Approving on `8d317eae`. Both items from my changes-requested are fixed, and I re-verified rather than reading the commit message.

## The twin — and you were right to reject half my patch

`Path.resolve()` follows symlinks; `path.resolve` on the TypeScript side does not. My patch would have closed the traversal divergence and opened a symlink one, in exactly the case `safe-path.ts` documents as its accepted gap. `os.path.abspath` is the right call. I confirmed the behaviour rather than taking it on the argument:

```
symlink inside base -> outside file
  os.path.abspath  -> stays inside base   (allowed — matches TS)
  Path.resolve()   -> lands outside base  (refused — diverges)
```

Two vectors instead of one is the part I would single out: pinning the over-correction as well as the bug is what stops the next person "fixing" this back to `resolve()`. Both mutations, run here:

| mutation | reds |
|---|---|
| containment removed | `test_diff_refuses_a_traversing_key`, and only it |
| `os.path.abspath` -> `Path.resolve()` | `test_diff_follows_a_symlink_inside_the_base`, and only it |

And the parity itself — both implementations against identical fixtures:

```
                                        TS                 PY
traversing key              missing-on-disk    missing-on-disk
symlink, content matches                 {}                 {}
symlink, content differs    content-differs    content-differs
absolute key                missing-on-disk    missing-on-disk
```

Four for four, including an absolute key neither test covers.

`readSkill` matches its siblings now, and I re-checked the mutation property after the change: removing the guard still reds the named test, on a populated `SkillInfo` instead of a throw.

`eval/harness` 3119 passed / 3 skipped (pre-existing, unrelated), `eval/app` 196 green, `tsc` and `eslint` clean, branch level with main. Only `lib/skills.ts` and its test moved on the TypeScript side, so the 22-site call-site sweep from the last round still stands.

## Three to fold in before you merge, none blocking

- `isWithin` has no production caller anywhere in the repo — only its own test. `snapshot.ts:236` names its misuse as the bug class this PR closes, so leaving it exported is an invitation. Delete it and its `describe` block.
- `docs/specs/eval-crud-ui-spec.md` §7 "Data storage" does not record the contract this PR establishes. One bullet covers it: every path built from caller input is resolved and proven inside its base (`lib/fs/safe-path.ts`); reads return null on refusal, writes and deletes throw.
- The mutation table's header says "Each sink's guard was removed individually", but `release.ts:114` and `:115` each survive alone and red only as a pair, and `tests.ts:208` cannot red at all — its segment is a basename taken off a readdir filtered to `.json`. At sink level every row is true; per line, those two are not. Worth a word, since that table is the PR's own evidence.

## One note for the record

@T-FEH's approval reads `fa1c76b3`, so it predates the Python commit, and this repo does not dismiss approvals on push. Mine is the second, which means the harness change merges reviewed by me alone. I have verified it end to end above, but flagging it rather than letting it pass silently.

johnmarkpeterbrown (commented):
One thing to change before this merges.

## 11 of the 14 containment tests cannot fail on Windows

The helper that builds every traversal target is POSIX-only:

```ts
await run(UP + abs.replace(/^\//, ''), abs)
```

On Windows `abs` is `C:\Users\…\secret.json`. `/^\//` matches nothing, so the drive letter survives and the target resolves to `C:\C:\Users\…` — a path that cannot exist. Every read sink returns null on ENOENT with or without its guard.

I mutated each sink's `resolveWithin` to a bare `path.resolve` and ran the suite. Eleven mutations survive at 201/201 green: `readFixture`, `readRunLogById`, `readSnapshotFiles`, `readAnnotation`, `readScenario`, `listRunLogsForSkill`, `readSkill`, `diffSnapshotVsDisk`, both `computeBlockedReason` oracles, and the `writeTest` edit branch. Only the sinks that throw still red.

With the change below, all eleven red under those same mutations, each naming its own sink, and the baseline stays 201/201.

CI is `ubuntu-latest`, so the guards are genuinely enforced where merges are gated. What is inert is the local suite, on the platform half this team develops on — and a test that reads as coverage and cannot fail is the exact failure this PR spent four rounds closing.

Eight identical occurrences: six in `tests/unit/pathContainment.test.ts`, two in `tests/fs/tests.test.ts`.

You wrote:

```
.replace(/^\//, '')
```

Change it to:

```
.replace(/^([A-Za-z]:)?[\\/]/, '')
```

Then one line in the mutation table, so the next reader knows the tests are platform-sensitive:

Change it to:

```
Mutations verified on POSIX and Windows; the traversal helper strips a drive
prefix, so the target resolves to the real outside file on either.
```

You already handled this correctly on the Python side — `skipif(os.name == "nt")` on the symlink test, with the reason written out. The TypeScript half goes quiet instead of saying so.

## Two notes, neither blocking

`5ecd96a3` — the Python containment and the `readSkill` null return — landed after @T-FEH and @chrisedeson approved. Only @chesworthrm's approval has read it. Approvals are not dismissed on push in this repo, so nothing is blocked; flagging it so the Python half does not end up single-reviewed by accident.

The required-check gap in your **What nothing checks** section is real and I confirmed the eight. It is already tracked as issue #2037, so there is nothing further to do about it here.

Issue #1996, opened 2026-08-27
fulltext sidecar cards show the record type instead of the title, and their FamilySearch link is dead


taiwo-stack:
> **Reviewed 2026-08-28 before junior handoff — the premise was false and the ask is
> replaced.** The card asked to stop rendering the `textDocument` snippet on the grounds
> that the staged sidecar has no such field. It does: the strip runs *after*
> `stageSearchResults` has already written the sidecar the viewer reads. Executed as
> written, the PR would have deleted the transcript and its highlighting from every
> full-text card, deleted the passing test covering them, and gone green. Two real bugs
> sit in the same function and neither was named. The original body follows under
> "Original issue".

**Touches:** packages/viewer-ui/src/components/shared/SidecarResultCard.tsx, packages/viewer-ui/src/components/shared/__tests__/SidecarResultCard.test.tsx, packages/schema/src/index.ts, eval/app/components/scenario/lib/schema.ts

`SidecarResultCard.tsx` has not been touched since the web-workbench PR #293
(`git log --oneline -1 -- packages/viewer-ui/src/components/shared/SidecarResultCard.tsx`),
i.e. before result staging (#721) reshaped `fulltext_search`'s output. Three
things drifted.

**1. The card title ignores the record's actual title.** `SidecarResultCard.tsx:213-216`
titles a fulltext card `recordType ?? collectionTitle` — "Probate", never
"Last Will and Testament of Patrick Flynn". The tool does set `result.title`
(`packages/engine/mcp-server/src/tools/fulltext-search.ts:118`, from
`content.title`) and it is in the sidecar. The card cannot reach it because the
type mirrors omit the field.

**2. "Open in FamilySearch →" is a dead link on every fulltext card.**
`FulltextSearchBody` calls `openExternal(result.id)`
(`SidecarResultCard.tsx:190-194`), but `id` is a **bare ARK**: `mapEntry` sets
`id: toArk(entry.id)` (`fulltext-search.ts:113`), asserted as
`"ark:/61903/3:1:3Q9M-CSNL-S98H-M"` at
`packages/engine/mcp-server/tests/tools/fulltext-search.test.ts:105`.
`window.open('ark:/61903/…')` resolves against the app origin and goes nowhere.
The real URL is on `sourceUrl` — also missing from the mirrors.

**3. Both hand-maintained `FulltextSearchResult` mirrors are missing five
fields** that the shipped `FulltextResult`
[...]

PR #2001, opened 2026-08-28
fix(fulltext_search): reconcile spec with shipped behavior, fix triage guidance

### LINKED ISSUE #1826: search-full-text triages on textDocument, which the tool strips from every staged result

taiwo-stack:
Closes #1826.

## Live measurement:

Ran a real live call against the upstream FTS API (`+Patrick +Flynn`, dumping
the raw `entries[0]`) before touching any code:

- **No relevance `score` anywhere** in the raw entry — confirms `FSFulltextEntry`
  declaring no `score` field is correct upstream behavior, not a mapping gap.
  Nothing to map; deleted `score` from the spec's output schema instead.
- **`content.highlightTexts` are bare matched terms** (e.g. `["Flynn", "witness"]`),
  not marked-up transcript fragments (`"Patrick <em>Flynn</em>"`). Fixed the
  test fixture that assumed the marked-up form and taught this shape to the
  skill as `highlightTerms`.

## #1978 merged into main first — what's actually left in this diff

#1978 (Francis) targeted the same issue and merged into `main` on 2026-09-01.
Main now already carries the triage rewrite (`SKILL.md` step 6, byte-identical
to this branch), the **Result triage** rubric dimension, and most of the spec
reconciliation described below. The fixture-file strip #1978 proposed did not
survive: main strips at the mock level instead, via the compiled
`compactStagedFulltextSearch` from #2045, and main's fulltext fixtures still
carry `textDocument` — the un-staged case this PR's
`test_fulltext_search_keeps_textDocument_when_not_staged` pins.

What this PR still adds on top of main:

- The two routing DO-NOTs in the frontmatter description, closing **#2010**.
- Four craft rules traced to named failing runs: the same-sentence place trap
  (step 4), log `query` mirrors the actual call (step 7), the plan-item name
  collision (steps 8 and 10), and passing `id` to extraction (step 11).
- Spec: `textDocument`'s staging-failure condition, the `yearFrom`/`yearTo`
  pairing rules, the `X-FS-Feature-Tag` note, and replacing the stale "must
  probe the actual response" implementation note.
- The `Fl?n*` wildcard guidance in `query-syntax.md`, the new
  `fulltext-search-flynn-prefix-wildcard-variant` fixture and its two test
  wirings.
- The two `mock_mcp.py` regression tests pinning the staged-strip and
  un-staged-keep paths (the strip itself now runs via main's compiled
  compactor — see below).
- A fresh annotated 13/13 run log.

## What changed

- `docs/specs/fulltext-search-tool-spec.md`: full reconciliation against the
  shipped tool — input schema (`nlQuery`, `projectPath`), `count` default
  (20 → 5), output schema (`textDocument`/`highlightTerms`/`staged`/
  `stagingError`, dropped fields that were never shipped), query-mapping and
  auth sections, facets shape. `textDocument`'s presence condition also
  corrected — it survives not only when `projectPath` is omitted but also
  when staging itself fails (`staged: null`, `stagingError` set), since the
  strip is guarded on `staged`.
- `fulltext-search.ts`: fixed a wrong comment claiming `record_read` can read
  a staged fulltext transcript back from its sidecar. It can't —
  `readFromSidecar` requires `gedcomx`, which a fulltext-staged entry never
  has.
- `eval/harness/harness/mock_mcp.py`: two new tests pinning the staged-strip
  and un-staged-keep paths. The strip itself already runs via main's compiled
  `compactStagedFulltextSearch` (from #2045) — no separate Python mirror
  needed here.
- `search-full-text/SKILL.md`: rewrote the triage step and the
  pass-to-extraction step to stop instructing the agent to read
  `textDocument`, which is not present once a search is staged (and, per
  review, not "never present" — it survives when staging itself fails).
  Triage now uses `highlightTerms`/`names`/`places`/`dates`/`title`/
  `recordType`/`recordPlace` — the fields that actually survive. The
  hand-off step also now explicitly names `id` (the record's ARK) as
  required, since record-extraction needs it to reach the record and the
  old wording only carried it by implication.
- `search-full-text/references/query-syntax.md`: dropped a stale "displayed
  excerpt" framing that no longer matches the corrected understanding above.

Four more `SKILL.md` fixes surfaced while re-running the eval suite against
the changes above — each traced to an actual failing run, not speculative:

- Named `question-selection` explicitly as the route for "scope a new
  project / propose opening research questions" requests with no plan item
  yet — the skill was intermittently doing that work itself instead of
  declining (`ut_search_full_text_006`).
- The log entry's `query` object must mirror exactly what the
  `fulltext_search` call actually sent — never a filter the user mentioned
  but the call omitted. The step 7 example previously modeled this
  incorrectly (`ut_search_full_text_010`).
- Named the specific phrasing trap that caused the fix above to surface a
  second failure mode: when the user's own request phrases a place in the
  same sentence as the person ("...as a witness in Schuylkill County,
  Pennsylvania"), the first call must still stay unscoped — the fix is to
  drop the place from the log, not add it to the call (`ut_search_full_text_010`,
  second mode).
- Named the cross-reference name-collision trap for plan-item completion: a
  witness-mention search for a name (e.g. "Thomas Flynn") does not complete
  a different, unrelated plan item that targets a same-named person's own
  record — cross-reference sub-searches now log ad-hoc unless a plan item
  actually asks for that person's own record (`ut_search_full_text_001`,
  `ut_search_full_text_013`).
- Named the extraction-from-record-in-hand case explicitly in the
  frontmatter description ("do NOT use when the user has a record in hand
  wanting extraction — use record-extraction"). This was `ut_search_full_text_003`'s
  actual root cause and closes **#2010**, filed and fixed within this PR at
  the user's direction rather than left as a separate card.
- `query-syntax.md`: added a worked example for combining `?` and `*` in one
  wildcard term to cover variants of *different lengths* (Flynn/Flinn same
  length via `?`, Flyn/Flynne different lengths via trailing `*`) — the
  model had been constructing `Fl*nn` (wrong: requires the string to
  literally end in "nn") instead of `Fl?n*` (`ut_search_full_text_007`).
- `eval/fixtures/mcp/fulltext-search-flynn-prefix-wildcard-variant.json`
  (new) + `negative-result-with-detail.json`: fixture coverage for a
  prefix-wildcard retry variant (`+Fl*`, `+Flyn*`, ...) the model tried as
  one of its final queries before declaring a coverage gap
  (`ut_search_full_text_002`).

## Decisions worth flagging

- **Fixture-scoping deviation.** The issue's order-of-work asked to scope the
  `textDocument`-leak fix to the named `eval/fixtures/mcp/fulltext-search-*.json`
  files. Main fixed it generically instead — staging depends on the per-call
  `projectPath`, which a static fixture file can't express (the same fixture
  also serves un-staged call shapes), so a fixture-level fix can't express
  "strip only when this call happened to stage."
- **Tool-contract question left open for the lead**, per the issue's explicit
  instruction not to add a new tool output field unilaterally: should
  `fulltext_search` gain a bounded transcript-excerpt field (or should
  `record_read` be able to read a fulltext-staged sidecar back at all)? Right
  now, once a result is staged, nothing can answer "does this name appear in
  the right context" — the skill now says so honestly instead of guessing.

## Follow-on issues filed

- **#1996** — real but out-of-scope bug this surfaced: the viewer renders an
  empty snippet for every staged fulltext result (`SidecarResultCard.tsx`
  still reads `textDocument`), and the `FulltextSearchResult` type mirrors in
  `packages/schema` and `eval/app` are stale against the tool's real shape.
  (Open; being fixed in PR #2089.)
- **#2009** — `record_search`'s equivalent staged-result slim, same class of
  bug as the `textDocument` fix here. Filed too-big-for-this-PR, but main's
  `compactStagedRecordSearch` (from #2045) already covers all five fields
  (`gedcomx`, `collectionUrl`, the `collectionTitle` hoist, `treeMatches`,
  duplicate `events`) and is wired into this harness's `_STAGED_COMPACTORS`.
  Closed — no longer a gap.

**#2010** was also filed while iterating on this PR (a pre-existing routing
flake behind an early misdiagnosed eval failure), but per direction it was
fixed here instead of left open — see "What changed" — and is now closed.

## Testing

- `make engine-test`: 2827 passed, 1 skipped.
- `make harness-test`: 3132 passed, 4 skipped.
- `make eval-skill SKILL=search-full-text`: **13/13 pass** on the shipping run
  log `v1_2026-09-02_10-42-53`, annotated. (An earlier 12-pass/1-partial result
  on `v1_2026-08-29_16-22-35` was superseded by this run.)
- `check_runlogs.py` (the CI gate): all rules satisfied locally against this
  diff.
- Deleted the now-dead `mock_mcp.py` fulltext-strip block (redundant with
  main's compiled `compactStagedFulltextSearch`, wired in via
  `_STAGED_COMPACTORS`) — verified both ways: with only the dead block
  removed, both `test_fulltext_search_strips_textDocument_once_staged` and
  its un-staged sibling still pass; with the compactor *also* disabled,
  `strips_textDocument_once_staged` fails on the real assertion. Full
  `test_mock_mcp.py` (40 tests) and the whole harness suite green after.


chesworthrm:
## Review — full pass, reviewed as merged with `main` (68 commits behind)

**Does:** Reconciles the `fulltext_search` spec with the shipped tool (adds `nlQuery`/`projectPath`, `count` 20→5, real output field names, `staged`/`stagingError`, drops the never-shipped `score`/`snippet`/`recordUrl`/`imageUrl`); fixes the wrong comment claiming `record_read` can read a fulltext-staged sidecar; teaches `mock_mcp.py` to mirror the tool's `textDocument` strip; rewrites the SKILL.md triage and hand-off steps off `textDocument`.

**Intent:** Matches #1826, including the live-measurement step. The spec half checks out — I compared the spec's input schema against the compiled `fulltextSearchToolSchema`, and they are set-equal (17 properties, `required: []`). The result and response field lists match the TypeScript types exactly. The fixture-scoping deviation and the deferred tool-contract question are both declared in the PR body.

**Checks:** All green. `mergeStateStatus: BEHIND` — update the branch before merging.

**Still needs:** One approval from `senior-developers` or `senior-genealogists` (every CODEOWNERS rule names both as an OR), plus branch protection's second. `codeowners/errors` is empty.

**Verified** (test-merge with `origin/main`, clean): vitest 2731 passed; `tsc --noEmit` exit 0; harness pytest 2878 passed / 2 skipped; `turbo test typecheck` 8/8; `apps/server` pytest 210 passed / 1 skipped; `check_runlogs.py` all rules satisfied; the four warn-only lints identical to their counts on `main`. I also mutated the new strip guard to `False` and watched `test_fulltext_search_strips_textDocument_once_staged` fail on the real assertion — the new test can fail.

**Existing reviews:** none. This is the first round.

---

### [CRITICAL] PR body, "Testing" — the failing eval test is diagnosed wrongly, and the acceptance argument rests on that diagnosis

The run log says the opposite of what the PR body says. In `v1_2026-08-27_22-30-16.json`, `ut_search_full_text_003` records `"activated": true` and `"skills_invoked": ["search-full-text"]`. It is a *negative* test — search-full-text must not fire. It fired.

I traced which branch actually failed it. `_compute_outcome` (`eval/harness/harness/orchestrator.py`) checks `aborted_reason` (null here), then `validators_passed` (true here), then `grade_on_invariant` (not set on this test), then line 1224:

```python
        # Fail iff the skill under test ACTIVATED.
        if activated:
            return "fail"
```

That is the branch that fired. Two things follow.

The judge's 3/3/3 cannot rescue it, by design. The same function says at line 1256 that for a routing negative the judge runs "base-only and diagnostically … its scores must NOT flip a correctly-routed test." The judge graded the decline text the skill produced *after* it was already running, which is not what this test measures.

The gate you're describing is real, but it is a different one and it did not fire. `orchestrator.py:463-464` aborts a run that calls an unregistered tool — but that yields `aborted`, not `fail`, and `aborted_reason` is `null`. The skill never attempted the call: its four recorded calls are `project_context`, `research_query` ×2 and `research_log_append`, and the only warning is `missing_tool_usage_dimension` (no `uncovered_tool_call`). The model checked ToolSearch and gave up before calling anything.

It is also not one-off. Across the four tracked run logs:

| run log | outcome | activated | skills_invoked |
|---|---|---|---|
| `v1_2026-08-25_08-31-04` (on main) | fail | false | `[]` |
| `v1_2026-08-25_11-41-40` (on main) | fail | false | `[]` |
| `v1_2026-08-27_11-24-11` (this PR) | pass | false | `["record-extraction"]` |
| `v1_2026-08-27_22-30-16` (this PR) | fail | **true** | `["search-full-text"]` |

```
You wrote:    (`ut_search_full_text_003`) was a one-off tool-registration environment
              glitch in that sandboxed run, not a skill defect — the judge's own
              dimension scores (3/3/3) confirm the skill handled the missing-tool case
              correctly; the harness marks any run containing a hard tool-call error as
              fail regardless of judge scores. Human-annotated, agreeing with the judge.

Change it to: (`ut_search_full_text_003`) is a routing miss, not an environment glitch.
              The run log records `activated: true` and
              `skills_invoked: ["search-full-text"]`, so it failed on the
              `if activated: return "fail"` branch of `_compute_outcome`
              (orchestrator.py:1224). The unregistered `fulltext_search` tool is by
              design — this test declares `mcp_fixtures: []` — and is what the skill hit
              *after* being wrongly invoked; the unmatched-tool gate at
              orchestrator.py:463 produces `aborted`, and `aborted_reason` is null here.
              The judge's 3/3/3 graded the decline text the skill wrote once already
              running, which orchestrator.py:1256 says must not flip a routing negative.
              This test has failed in 3 of the 4 tracked run logs under three different
              modes, so it is a persistent routing flake that predates this PR. This PR
              does not touch the SKILL.md `description`, so it is not the cause.
```

Two follow-ups. Correct the three `"corrected_score": 3` entries for this test in `v1_2026-08-27_22-30-16.ann.json` through the CRUD UI — the human pass agreed with a reading the log contradicts, with `"comment": null`. And file an issue for the flake itself; it spans `main` and is not yours to fix here.

Worth putting in that issue: on the failing run the wrongly-invoked skill wrote `log_006` into `research.json`. A negative test that mutates project state is its own problem.

---

### [CRITICAL] `eval/harness/harness/mock_mcp.py:470` — the deferred gap points at the issue this PR closes, so nothing will track it

```
You wrote:    # record_search's equivalent gedcomx strip is a separate gap,
              # not fixed here (issue #1826).

Change it to: # record_search's equivalent staged slim (record-search.ts:1100-1133 —
              # gedcomx, collectionUrl, collectionTitle hoist, treeMatches, events)
              # is a separate gap, not fixed here (issue #NNNN).
```

#1826 closes on merge, and it says "a separate card, not this one." No such card exists — I searched all 926 open and closed issues.

The divergence is real: `record-search.ts:1103` deletes `gedcomx` from every result once staged, and the `record_search` fixtures still carry it. Same class of bug this PR just fixed for `fulltext_search`, across many more tests.

I opened the call site before calling it too big for this PR. The block at `record-search.ts:1100-1133` is five transformations, not one field: `gedcomx` delete, `collectionUrl` delete, `collectionTitle` hoisted into a response-level `collections` map, empty `treeMatches` dropped, `events` deduped. Mirroring it means changing `mock_mcp.py` plus the expectations of every record_search test. That earns its own card — label it `developer` and `nothing-checks`, since a mock handing the agent a field production removed is exactly a way CI stays green while the thing is broken.

---

### [MAJOR] `packages/engine/plugin/skills/search-full-text/SKILL.md:279-282` — the hand-off list omits the field that identifies the record

Step 11 now enumerates what to pass to record-extraction and asserts extraction will read the image. The enumeration leaves out the record's ARK, which is what `id` carries. Without it record-extraction has nothing to open.

The skill body never names `id` at all — line 193 writes `source_attachments({ uris: [ark1, ark2, ...] })` without saying which field `ark1` comes from. The old wording ("pass this context along") was vague enough to carry the ARK by implication; the new enumeration reads as complete and is not.

```
You wrote:    For each promising record, invoke record-extraction to process it. Staging
              strips the full transcript, so pass what survives — `names`/`places`/`dates`,
              `title`, `recordType`, `recordPlace`, `highlightTerms` — as triage context;
              record-extraction reads the source image itself for the transcript.

Change it to: For each promising record, invoke record-extraction to process it. Always
              pass the result's `id` (the record's ARK) — it is what record-extraction
              needs to reach the record. Staging strips the full transcript, so pass what
              survives — `names`/`places`/`dates`, `title`, `recordType`, `recordPlace`,
              `highlightTerms` — as triage context; record-extraction reads the source
              image itself for the transcript.
```

Your claim about the image read is sound, for what it's worth: `arkToImageUrl` (`fs-image-fetch.ts:83`) accepts a bare `ark:/61903/3:1:…` with no `$dist`. It just needs the ARK to travel with the hand-off. No test in the suite exercises this step, so nothing would have caught it.

---

### [MINOR] `packages/engine/plugin/skills/search-full-text/SKILL.md:175-176` — "never present" is too absolute

`textDocument` survives when staging *fails*. The catch block sets `staged = null` alongside `stagingError`, and the strip is guarded by `if (out.staged)`. Step 5 already contemplates that path. Small, but this PR exists to stop the skill making false claims about field presence.

```
You wrote:    **Every logged search stages its results, and staging strips the full
              transcript (`textDocument`) from every result — it is never present here.**

Change it to: **Every logged search stages its results, and staging strips the full
              transcript (`textDocument`) from every result — once `staged` is set it is
              not present here.**
```

---

### One question, not a finding

The fulltext fixtures disagree with the spec about what `id` is. `FulltextResult` and the spec both say a fulltext id is a `3:1:`/`3:2:` document image, but 9 of the 11 fixture ids are `1:1:` record personas — which `arkToImageUrl` explicitly rejects. The fixtures also store un-normalized resolver URLs (`https://familysearch.org/ark:/61903/…`) where production returns `toArk`-canonicalized bare ARKs. Both are pre-existing and outside this diff, and I have not checked what upstream actually sends. You measured the live response for this card — do you know? It plausibly belongs in the same new issue as the `record_search` gap above.

---

**What I did not check:** I did not re-run `make eval-skill` — a paid run, and it would not settle the routing question the existing logs already answer. And no CI job can confirm the SKILL.md rewrite improves real triage behavior; your PR body says so, and I agree.

The spec reconciliation is careful work and I could not fault it. The `record_read` comment fix is right — `readFromSidecar` matches on `r.recordId` and requires `match.gedcomx`, and a fulltext result carries `id` with no `gedcomx`, so that call could never have worked.

johnmarkpeterbrown (changes requested):
Reviewed the full diff and the run logs. The engineering here is solid — the spec now
matches the shipped tool field for field, the mock fix is real coverage rather than
decoration, and the deviation you declared on fixture-scoping is the right call for the
reason you gave. Three things to fix, none of which need another paid run.

---

**1. The account of the failing eval run is wrong — and your own issue #2010 has it right.**

`ut_search_full_text_003` is a negative test. The skill is supposed to decline and route to
`record-extraction`. In the run log it shows `activated: true` and
`skills_invoked: ["search-full-text"]`, it made four tool calls, and it wrote `log_006`
into `research.json`.

That activation is the whole reason the test failed. The harness fails a routing negative
outright once `activated` is true, and it never looks at tool errors when deciding the
outcome. On a negative test the judge runs diagnostically only — its 3/3/3 cannot move the
result either way, and its rationale says the skill "appropriately declined", which the
trace contradicts.

The unregistered tool is real, and it is why no search actually ran. It is not why the
test failed.

Issue #2010 gets all of this right, including the `research.json` write. The PR body
doesn't, and doesn't link it.

```
You wrote:      The one failure
                (`ut_search_full_text_003`) was a one-off tool-registration environment
                glitch in that sandboxed run, not a skill defect — the judge's own
                dimension scores (3/3/3) confirm the skill handled the missing-tool case
                correctly; the harness marks any run containing a hard tool-call error as
                fail regardless of judge scores. Human-annotated, agreeing with the judge.

Change it to:   The one failure (`ut_search_full_text_003`) is a negative routing test the
                skill wrongly activated on: `activated: true`, `skills_invoked:
                ["search-full-text"]`, and a `research_log_append` that wrote `log_006`.
                A routing negative fails outright once `activated` is true
                (`orchestrator.py:1324`); the judge's 3/3/3 is base-only and diagnostic
                there and does not bear on the outcome. `fulltext_search` was also
                unregistered in that sandbox, which is why no search ran — but that is not
                why the test failed. Tracked as issue #2010: the same test also failed on
                both pre-#1826 runs on main, in a different mode (`activated: false`,
                `skills_invoked: []`), so this is a pre-existing flaky routing boundary,
                not a regression from this PR.
```

To be clear: no re-run. That test has failed in 3 of its 4 recorded runs, twice on `main`
before this branch existed. A re-run would not settle it, and #2010 is the right home for
it.

---

**2. The annotation records silent agreement with that misread.**

In `eval/runlogs/unit/search-full-text/v1_2026-08-27_22-30-16.ann.json`, all three of
003's base dimensions are `llm_score: 3 → corrected_score: 3` with `comment: null`.

These annotations are what judge calibration reads. As written, the corpus records "the
judge was right here" about a rationale that describes a decline which never happened.
The scores themselves may be defensible — the decline text was fine. The silence isn't.

Add a comment to those three entries through the CRUD UI (`make eval-ui`), along these
lines:

    Judge graded decline text on a run where the skill had in fact activated
    (`activated: true`, wrote `log_006`). Scores reflect the text, not the routing
    outcome. See issue #2010.

Leave the scores where you and the lead think they belong. The comment is the part that
has to be there.

---

**3. The spec's `textDocument` condition misses the staging-failure case.**

When staging throws, `staged` is `null` and `stagingError` is set. The strip is guarded on
`staged`, so `textDocument` survives — even though `projectPath` was supplied. Worth
getting exactly right in a PR whose acceptance bar is that spec and implementation agree
on every output field.

In `docs/specs/fulltext-search-tool-spec.md`:

```
You wrote:      * The full AI-transcribed page text. Present when `projectPath` was NOT
                * supplied (nothing was staged, so nothing to re-read from a sidecar).

Change it to:   * The full AI-transcribed page text. Present whenever nothing was staged —
                * either `projectPath` was not supplied, or staging failed (`staged: null`
                * with `stagingError` set), since the strip is guarded on `staged`.
```

---

Nothing else. The `count` decision, the deferred tool-contract question, and the choice
not to add a rubric dimension are all the right calls and all declared.
````

Two notes for you, not for the PR: items 1 and 2 are the ones I'd treat as must-fix before merge, since they're the record of why a failing paid run was accepted. Item 3 is a one-liner. And the PR still needs two approvals with at least one from a senior team — mine wouldn't have unblocked it either way.

**Comment by taiwo-stack** [comment]
I got it, I'll take the rubric dimension from #1978, it's a nice addition. Skipping the fixture edits though, since my fix already handles that correctly. Pushing everything together shortly.

**Review by T-FEH — APPROVED** [review]
Reviewed on the merged head (`3f43ba00` — your `4425f5f9` plus a clean merge of main). Gate green; engine 2726 passed; harness 2927 passed, 2 skipped; `ruff` clean.

## @johnmarkpeterbrown's three items are all resolved, and one better than asked

- **The spec's `textDocument` condition** now reads *"Present whenever nothing was staged — either `projectPath` was not supplied, or staging failed (`staged: null` with `stagingError` set), since the strip is guarded on `staged`."* Verbatim what was requested.
- **The mischaracterised failure** was fixed at the root rather than reworded. You filed **#2010** for the routing miss and closed it inside this PR, and the shipping run shows it worked: `_003` is `activated: False`, `skills_invoked: ['record-extraction']`, zero tool calls — the correct destination. The body also records that the earlier "environment glitch" reading was superseded, which is the honest version.
- **The annotation item** concerned a run log that has since been superseded; the shipping run carries its own, and rule 3 passes.

## The acceptance bar is met

Spec against shipped tool, field for field: **17 inputs, 17 in code**, same names and order, including the added `nlQuery`/`projectPath` and the `count` default 20 → 5. **13 outputs, identical** names, order and optionality, with `score` and the four never-shipped fields dropped and the facets shape corrected. And every field the rewritten triage and hand-off steps name — `highlightTerms`, `names`, `places`, `dates`, `title`, `recordType`, `recordPlace`, `id` — exists on `FulltextResult` and survives the strip. None fails to survive, which is the thing #1826 was actually about.

The mock strip mirrors the real guard exactly, and it is proven in **both** directions on **both** sides: dropping the guard reds `keeps_textDocument_when_not_staged`, deleting the block reds `strips_textDocument_once_staged`, and flipping the real tool's `if (out.staged)` to `true`/`false` reds its two engine tests. Real coverage, not decoration. The `readFromSidecar` comment fix verifies too.

## Three things worth your attention, none of which I think blocks

**Someone else is doing this work.** [PR #1978](https://github.com/PioneerAIAcademy/cowork-genealogy/pull/1978) is **open** against the same issue #1826, by a different author:

```
#1978  francis-2008-happy   created 2026-08-27   +7021/-5727 in 17 files
#2001  taiwo-stack          created 2026-08-28   +36932/-27323 in 25 files
```

A day apart, same issue, and both carry paid eval runs. One of them is going to be discarded. That is the four-person overhead CLAUDE.md warns about, and it is a coordination call for @DallanQ rather than a defect in either PR — but it should not merge without someone deciding which one survives. Related: the review above endorsed "the choice **not** to add a rubric dimension" as the right call, and #1978's dimension has since been folded in here, so that endorsement no longer describes the branch.

**The new `Result triage` dimension has never discriminated.** Across all five committed runs:

```
44 scored instances — 43 × 3, 1 × null. Nothing below 3.
```

That includes the two runs with genuine `fail` outcomes. Expected for a dimension added alongside the fix it grades, and I am **not** asking for a re-run — but nothing committed shows it can fail, and `/audit-rubric search-full-text` is the repo's own instrument for that.

**The run count and cost are not disclosed.** Five committed run logs, all 2026-08-29, sitting at the retention cap:

```
00-22-14 $2.47   00-34-51 $2.36   12-33-39 $2.31   15-58-07 $2.14   16-22-35 $2.32
total $11.60
```

The body mentions only that #2010 was "filed while iterating". Sibling PRs this week state their run counts and cost outright, and the repo holds each skill's eval slot to one item precisely because these are the scarce resource. One line.

## Smaller

- The new `test_fulltext_search_strips_textDocument_once_staged` calls the compiled `stageSearchResults` and hard-fails pre-build; its siblings in that file carry the skip marker. CI builds first, so CI is unaffected.
- The spec reconciliation drops two constraints `validateInput` actually enforces — `yearFrom`/`yearTo` must be paired, and `yearFrom <= yearTo`. For a PR whose bar is "spec and implementation agree on every field", worth adding.
- `SKILL.md:57` still names `textDocument` in the provenance chain, 130 lines above the step explaining it is not present. Same stale framing survives in `references/online-search-literacy.md`.
- The annotation's three comments are the same sentence repeated across three dimensions of one test, with zero disagreements. Compliant, and thinner than this corpus usually gets.

Approving on the engineering. The duplicate-PR question is the one thing I would settle before merging, and it is not yours to settle alone.

**Comment by taiwo-stack** [comment]
@johnmarkpeterbrown heads up — the three items from your review are addressed (also independently confirmed by T-FEH's later approval). CI's green and the branch is mergeable. Ready for a fresh look when you have a chance.

**Comment by benter-070** [comment]
Not a reviewer here — @taiwo-stack asked me to look, and two of the things I checked
seemed worth leaving in the thread rather than only in DM.

**Updating the branch is free.** It is 22 behind, and that is the step most likely to
make someone hesitate on a PR carrying a paid run. Of the **177 files** `main` has
changed since the merge base (`c98dc54c`), **zero** appear in
`v1_2026-09-02_10-42-53`'s **50** snapshot entries. The 13 fixtures `main` touched are
the Kongsberg/Trysil place and collection fixtures plus the `hansen-kongsberg-baptism`
scenario, none referenced by a `search-full-text` test, and the skill dir and unit
tests are untouched. So the run log stays active through the merge — no re-run, and
not even the warn-only fixture arm.

**@johnmarkpeterbrown — item 2 was resolved by a different route than you asked for**,
and I think that is easy to miss reading the diff cold.

The annotation you flagged was on `v1_2026-08-27_22-30-16`, where `_003` had wrongly
activated. That run log is no longer in the PR. In the current one, `ut_search_full_text_003`
**passes**:

```
outcome: pass   test_type: negative   grading_mode: routing
activated: false   skills_invoked: ["record-extraction"]
```

and the suite has **zero non-pass tests**. So the routing bug was fixed rather than
annotated around — which means the `null` comments there are now correct, since rule 3
only requires a written comment where a test is not a confirmed pass. Your items 1 and 3
both check out too; the spec wording went in verbatim, `stagingError` case included.

Flagging it because your review is the record of *why a failing paid run was accepted*,
and the answer changed underneath it: there is no longer a failing run to explain.

DallanQ:
**Merge order for the `search-full-text` eval slot (board audit 2026-09-07).**

Three open PRs edit `packages/engine/plugin/skills/search-full-text/SKILL.md` and share its run-log snapshot: this one, PR #2001 and PR #2164/#2216. Landing them in the wrong order costs a paid run each.

**#2216 is stacked on #2164** — they share `aa5c2a7ae`, `baf0b7ddf` and `b6c9fed8e` — and #2216 carries an *older* copy of the diminutive-expansion work (`name-variants.ts` 275 lines vs 300; `name-variant-drift.test.ts` 77 vs 216). #2164 has one review-round commit, `e164c4acd`, that #2216 lacks.

Order:

1. **PR #2001 first.** It is the only one carrying a paid run log (`v1_2026-09-02_10-42-53` plus annotations) and it is independent of the other two. Clear CHANGES_REQUESTED, rebase off `main`, land.
2. **PR #2164 second**, rebased onto #2001. **Drop `eval-cosmetic-skip` before landing** — see below.
3. **PR #2216 last**, rebased onto #2164. Delete the duplicated diminutive-expansion files rather than resolving them (`name-variants.ts`, `fulltext-search.ts`, `image-transcribe.ts`, `given-name-variants.json`, the three test files, the three spec docs); #2164's copies are newer. What survives here is the V6/V8 validator work under `eval/harness/`. Its failing `runlogs` check clears with #2164's run.

**PR #2164's `eval-cosmetic-skip` label is not applicable.** Its `runlogs` check is green only because of that label, and the SKILL.md diff is not behaviour-neutral — it reverses the shipped rule `**Always use \`+\` to require terms.**` into `Do NOT use \`+\` in the \`name\` field — it disables auto-expansion`, and rewrites the witness/appraiser row of the strategy table from `+Surname` to `Surname`. `eval/CLAUDE.md` is explicit that the label is for behaviour-neutral changes only. This needs a real run.

chesworthrm:
**Correction to my dismissal message on @johnmarkpeterbrown's review.**

I dismissed that review because he's unavailable (hardware failure) and can't re-review. My dismissal message said the requested changes were "NOT addressed and still stand." That was wrong, and it understated @taiwo-stack's work.

All three items were addressed by pushes between 2026-08-28 and 2026-09-02 — two of them at the root rather than by rewording:

1. **The mischaracterised eval failure** — fixed at the root. #2010 was filed and closed inside this PR, and the shipping run log `v1_2026-09-02_10-42-53` shows `ut_search_full_text_003` with `activated: False`, `skills_invoked: ["record-extraction"]`, outcome `pass`, 13/13 overall.
2. **The annotation item** — moot. The `v1_2026-08-27_22-30-16.ann.json` it concerned is no longer in the tree; that run was superseded.
3. **The spec's `textDocument` staging-failure condition** — resolved in the exact wording requested.

The dismissal reflects resolved feedback, not bypassed feedback.

chesworthrm (approved):
Approving on the engineering.

Verified against the **merge** with main (41 commits behind, merges clean), not just the branch head:

- engine `vitest` 2891 passed; `tsc --noEmit` clean; harness `pytest` 3211 passed / 3 skipped; `turbo test typecheck` 8/8
- `check_runlogs.py` green with base set to current `origin/main` and the merged tree on disk — the shipping run log is still active, so **no re-run is owed**
- both new mock tests proven able to fail: neutering the compiled `compactStagedFulltextSearch` reds `strips_textDocument_once_staged`, and removing the mock's `projectPath` staging condition reds `keeps_textDocument_when_not_staged`. Real coverage, not decoration.
- spec against the shipped tool where I sampled it: the `yearFrom`/`yearTo` pairing and ordering rules (`fulltext-search.ts:46-56`), the `X-FS-Feature-Tag` branch (`:175`), and `textDocument` surviving a staging failure because the strip is guarded on `staged` (`:284`).

**Two body corrections before merge** — the record, not the code:

1. The bullet reading "`fulltext-search.ts`: fixed a wrong comment claiming `record_read` can read a staged fulltext transcript back from its sidecar" — that file is byte-identical to main on this branch (`git diff origin/main HEAD -- packages/engine/mcp-server/src/tools/fulltext-search.ts` is empty). That fix landed ahead of this branch. Please drop or reword it.
2. No run count or cost. Five run logs committed for this skill, $11.10 total, this shipping run $2.21. One line under Testing.

**One question for @DallanQ, not a blocker.** Step 7's new rule — that the logged `query` must mirror what the call actually sent — regulates a field `research_log_append` already derives: `finalizeStagedResults` returns `payloadQuery` (`results-staging.ts:211`) and `research-log-append.ts:330` defaults `entry.query` from it, but only when the caller omits `query`, and never compares when one is supplied. Worth deciding whether that rule belongs at the write boundary per ADR-0011 rather than in SKILL.md prose. Note it only covers the staged case — an un-staged search has no sidecar to compare against, so some prose stays either way. I'll file it separately unless you want it handled here.

FYI for whoever touches this skill next: the description is at 993 of the 1024-character cap, so the next routing DO-NOT added here will red `skill-description-length.test.ts`.

chesworthrm:
**Correction to my approval review, on the `fulltext-search.ts` bullet.**

I wrote that the fix "landed ahead of this branch." That was wrong — it never landed. The verified state:

- `packages/engine/mcp-server/src/tools/fulltext-search.ts` is byte-identical on this branch and on `main` (blob `242852c2` on both sides), so this PR does not change that file.
- The comment the bullet describes is not in the file at all any more — `readFromSidecar` appears zero times in it, on either side. You added it in `fe95de614` and corrected it in `6c0da2e23`; it went away in the merge with main.
- The bullet also still carries the explanation your own later commit replaced. It says the read fails because `readFromSidecar` "requires `gedcomx`". `6c0da2e23` says it actually filters staged entries on `r.recordId` — a field `FulltextResult` never has, since it carries `id` — so the lookup finds no match before it would ever reach the `gedcomx` check.

So the ask is just: **drop the bullet.** If the `record_read` trap is worth preserving, the spec is the place for it — it already owns the strip and reachability contract — rather than a claim about a file this PR leaves untouched.

Sorry for the misdirection; I'd have had you looking for something in main that was never there. Nothing here changes the approval.

Issue #1967, opened 2026-08-27
search-records: fix the judge contradicting its own judge_context on the 1850 household test (one re-run; no notes gate)


mercyokum:
> **Reviewed 2026-08-27 before junior handoff — premise was partly false; this is a
> stale-rewrite.** Three of the five symbols the original ask told a junior to port do not
> exist, and the cited failure is a grading defect, not a tool gap. The original body follows
> under "Original issue".

**Touches:** eval/tests/unit/search-records/whitfield-1850-household.json, eval/tests/unit/search-records/rubric.md, eval/runlogs/unit/search-records/

## Decision (lead, 2026-08-27) — fix the eval defect; no new layer

**Close the verified defect and stop there.** One `make eval-skill` re-run (measured $9.29)
fixes the judge contradicting its own `judge_context`. **No content gate on `notes`, no
shadow warning, no per-record/per-country table, and no reference-retrieval layer for this
class.**

### Why: the record-agnostic rule already exists, and it is one line

`packages/engine/plugin/agents/record-extractor.md` already states the general test, and uses
this exact record as its worked example:

> **The one test: was this value in a field on the record?** If yes it is `direct`, no matter
> who supplied it. A stated age on an 1850 census is `direct` though a household member
> reported it.

A *relationship* is not a field on a pre-1880 US census, so by our own existing rule it is
`indirect`. **No country knowledge, no record-type table and no reference document are needed
to reach the right answer here** — the rule is already written and already country-agnostic.
The work is making sure it is applied to relationships, not adding a layer beside it.

So the reported shape is not a missing-knowledge failure. The extraction layer has the right
test; the log *note* narrated an inference as a fact.

### The lead's structural constraint, recorded so it is not re-derived

**We will not hard-code record-type × country checks.** That space is roughly 200 countries ×
dozens of record types × several eras, every entry a maintenance liability that goes stale
silently. Anyone proposing one should expect it to be declined.

**Reference documents plus retrieval tools are the right shape — for a different class.** They
fit what the record in hand *cannot* answer: which repository holds the originals, what years
survive, what a jurisdiction's boundaries were, whether a substitute exists. They are the wrong
tool for "did this record state X", which is answerable by looking at the record. Note the
groundwork already exists if that direction is taken up later — `wiki_search` does RAG
retrieval over a pre-crawled FamilySearch wiki corpus, and `getLearningCenterDir()` /
[...]

Issue #1957, opened 2026-08-27
setup-feedback-case points the triager at an empty user_prompt instead of the session log


clack391:
**Touches:** scripts/setup-feedback-case.sh, scripts/setup-feedback-case.bat, eval/harness/tests/unit/test_setup_feedback_case.py, docs/specs/feedback-case-spec.md

Since #1919 the feedback dialog no longer requires the prompt box, so a submitted
`user_prompt` may legitimately be an empty string. Neither setup script handles it.

`scripts/setup-feedback-case.sh:164` falls through to:

    Users prompt: see <dir>/_feedback/feedback.json (user_prompt field)

which sends the triager to the field that is empty. It cannot currently tell
"the reporter left it blank" from "jq and python are both missing", because both
paths leave `USER_PROMPT` empty.

`scripts/setup-feedback-case.bat:131-138` is worse. It branches on
`if exist "!FB_JSON!"`, not on emptiness, so a present-but-blank prompt prints
the heading `Users prompt to issue first:` and both dashed rules with nothing
between them. A Windows triager sees what looks like a lost prompt and is given
no explanation.

**What it should do:** distinguish a read that succeeded and returned empty from
a read that failed, and on the first say the box was left blank and name
`_feedback/session-log.jsonl` as the fallback. Note the log is optional, a Cowork
submission never has one, and a trimmed log drops its oldest entries, so the
prompt may be absent there too.

**Nothing checks this.** `eval/harness/tests/unit/test_setup_feedback_case.py`
covers only the populated path, so both scripts can mislead a triager with the
suite green.

**Why this was filed rather than folded into PR #1941**, which found it: the fix
needs a verification path that PR could not provide. The Windows `.bat` half
cannot be exercised from the authoring environment, and shipping the `.sh` half
alone would leave the worse of the two platforms broken under a spec that says it
is handled. Open PR #1905 is also editing both scripts and this same spec file
(row 3 at `:103`, three lines from row 7), so the change wants to land after it.
PR #1941 records the current behaviour in the spec rather than asserting a
requirement nothing meets.

chesworthrm:
Merged into issue #1940.

These are one piece of work, not two similar ones. Both defects are in `scripts/setup-feedback-case.sh`, both need the same edit to `docs/specs/feedback-case-spec.md` §3, and both need a Windows box to verify — which is the reason this issue gave for not folding into PR #1941 in the first place.

[...]

PR #2004, opened 2026-08-28
research-plan: survey already-attached FAN-cluster facts before planning (#1948)

### LINKED ISSUE #1948: [feedback] 2026-08-26T20:24Z
**Zip:** https://drive.google.com/file/d/1K3PpTG8Mz40M1sPF-19oxh372Mn17pFW/view?usp=drivesdk

```
make feedback-case ZIP=~/Downloads/feedback-2026-08-26T20-24-40-770Z.zip
```

- Submitted: 2026-08-26T20:24:40.770Z
- Platform: win32

Cia-3:
Closes #1948

**Sequencing note (Dallan):** #2033 also edits `research-plan/SKILL.md` and
this PR's validator file, with no textual conflict but overlapping scope.
Landing #2033 first, then this PR rebases onto it and re-runs the eval once.
Holding further eval-skill re-runs on this branch until then.

## Summary

- Normal tier. `research-plan`'s first-plan step never surveyed sources already
  attached to FAN-cluster relatives before planning new searches — mined from a
  live alpha-feedback report (#1948): a tester asked why plan item 1 didn't
  check sources already on tree people, and the agent's own answer named the
  gap (the skill's "verify the starting point" step checks assumptions, not
  what's already attached).
- Added a "survey what's already attached" instruction to `SKILL.md`. On its
  own it plateaus around 50% reliability across three wording iterations — the
  harness runs skills without `temperature=0`, so no wording pins model
  behavior at exactly 100%.
- Closed the class going forward with a new **deterministic validator**
  (`test_survey_surfaces_already_attached_fan_facts`) that asserts directly on
  the response text: an already-sourced FAN-cluster fact must be named — the
  person's name plus fact-specific content (a date token and, when the fact
  has one, a fragment of its `value`) — in the same paragraph, or the test
  fails outright, not a fuzzy judge partial. Went through two review rounds:
  round 1 found a word-boundary bug and a whole-response (not paragraph-level)
  proximity gap; round 2 found the paragraph-level check still passed a
  response that mentioned the person and a coincidental date without ever
  reading the fact's content, which is what pushed the final design to require
  actual fact content, not just date proximity. Verified against every stored
  response for this test, including the counter-example that motivated round 2.

## Review guidance

**Start here:** `packages/engine/plugin/skills/research-plan/SKILL.md`'s new
"Survey what's already attached" paragraph (right after "Verify the starting
point"), and `eval/harness/validators/test_research_plan.py`'s
`test_survey_surfaces_already_attached_fan_facts`. The validator is tag-gated
(`already-attached`) rather than universal — the existing corpus has several
unrelated scenarios with incidental sourced facts on non-subject persons that
would false-positive under an unconditional check (confirmed by grep before
writing it).

**Unsure about:** Whether the wording can be tightened further to close the
remaining ~50% gap, or whether that's inherent to non-deterministic sampling
and the validator is the right permanent backstop rather than a stopgap. Not
blocking — the validator makes any regression here loud and immediate either
way, which was the actual goal.

## Plan

**Didn't change:** Left "Verify the starting point (BCG Standard 11)"
untouched — that step checks assumptions, not what's already attached; the new
paragraph is additive, right after it.

**Acceptance check:** `ut_research_plan_bpx` — fails on `main`'s SKILL.md
(model stays silent about the sibling's already-sourced fact), and on this
branch its validator entry fires a named assertion error
(`test_survey_surfaces_already_attached_fan_facts`) rather than a partial judge
score whenever the model goes silent or merely notices the source without
reading it — confirmed on a real run where it did exactly that.

**Deviated from the plan:** Originally planned to ship the prose fix alone
with a follow-up issue for full reliability; built the deterministic validator
in this PR instead once it became clear that was achievable now rather than
deferred. The validator itself then went through two more design iterations
during review (word-boundary → paragraph-proximity → content-based) as each
prior version turned out to have a real, demonstrated false positive.

## Lane-4 (ADR-0011)

Prose, not a `research_append` precondition. The predicate is linguistic
(does the response's reasoning name an already-attached fact's content) and
that shape was already declined for the identical case in #1967. The
deterministic backstop lives in the validator instead, which grades the
*response text* the same way the judge does, just without judge variance.

## Test plan

- [x] Ran the eval harness pytest suite directly after the validator
      redesign and after merging main twice (67 commits, then 10 more):
      2973 passed, 4 skipped against the merged state. Re-checked the two
      files this PR actively touches after dropping the unzip fix: 40
      passed, 1 skipped. Same pre-existing, unrelated `apps/electron`
      ESLint failure noted in the prior round persists (CRLF warnings from
      this machine's `core.autocrlf=true` checkout, zero files touched
      under `apps/electron`) — not re-run in full this round since nothing
      in it touches JS/TS/lint-scoped paths.
- [x] Ran `make eval-skill SKILL=research-plan` after the content-based
      predicate rewrite and again after the post-merge run-log refresh;
      committed the run logs + `.ann.json` for each.
- [x] N/A — no `description` frontmatter or DO NOT clause changed.

## Follow-on issues

- Folded in: everything from both review rounds that was a real, verifiable
  bug — see the review-round commits for the itemized list.
- Dropped, not filed: this PR originally also carried a `scripts/setup-feedback-case.sh`
  unzip exit-code fix (found while unpacking #1948's own feedback bundle),
  found independently of #2027 which fixes the same bug more cleanly.
  Dallan's sequencing call: land #2027, drop this PR's version and its two
  unit tests. Done.
- Already tracked separately, not touched here: issue #1940 (two other
  `setup-feedback-case.sh` defects, different lines, no overlap with #2027's
  fix either — noted there for coordination).


EdmondOware (changes requested):
Strong piece of work, and the part that matters most is right: the passing run does not just tick the box, it names Patrick Sheahan, cites the 1875 deed in Schuylkill County, and draws the inference that the sibling was there seven years before Michael. That is chain migration read correctly off an already attached source, and it reframes the whole "why did the family move" question. The two failure modes named in the SKILL.md paragraph, a plan item re requesting a record type already covered and a detail sitting on a source that was cited for something else, are both real and worth the words. The `unzip` exit code fix is also correct and, unlike the counterpart change in #1946, is not a duplicate of anything already on main.

Four things before this goes in.

**1. Mark the test, or state the measured rate for the shipped wording.** The body says the instruction plateaus around 50% across three iterations, and the harness runs one sample per test with no retry, so as written this test goes red about half the time. `expected_outcome` is unset, so it expects a clean pass. The evidence for "passes on this branch" is one run, `v1_2026-08-28_06-14-42`; the only other run carrying `ut_research_plan_bpx` predates the validator and never exercised it. This is your "Unsure about" question with its consequence made explicit: a deterministic validator on a coin flip behaviour does not make regressions loud, it makes normal variance loud. Please either mark it `xfail` with the measurement in the reason, the way #1946 handled `ut_search_records_h4k`, or give the measured pass rate for the final wording specifically if it is better than the earlier two.

**2. Use a word boundary for the given name.** `given.lower() in response_lower` is a plain substring test. It is safe for Patrick, and `ut_research_plan_bpx` is currently the only test carrying `already-attached`, so nothing is broken today. It will not stay safe: against a response that names nobody, `Ann` matches inside "planning" and `Ed` matches inside "scoped", and both words are near certain in any research plan response. A short given name in the next `already-attached` scenario would pass silently, which is worse than no guard because it reads as coverage. `re.search(rf"\b{re.escape(given)}\b", response_lower)` closes it.

**3. Fix the failure message.** It reads "neither the person's given name nor the fact's date appear in the response", which describes an OR. The check requires both, as the docstring correctly says. Whoever debugs a real failure will look for the wrong thing.

**4. Consider tightening what counts as surfacing, or say why coarse is right.** The docstring is honest that the check does not require the name and the date to appear near each other. In this very run the response contains a second, unrelated `1875` in a naturalization item spanning 1875 to 1900, so a response that mentions the sibling as a FAN person and proposes any 1875 spanning search would satisfy the check without ever reading the deed. Coarse may still be the right call for a validator; if so, please say so in the docstring alongside the existing caveat.

One recommendation, not an ask, since it is bigger than this PR. The requirement is currently "the response text mentions it", which no writer tool can enforce, so the validator can only detect. If it were instead "the plan item's rationale names the already attached fact", it becomes a `research_append` precondition, decidable from `research.json` and `tree.gedcomx.json` alone. That is ADR-0011's first question answered yes, and it would bind rather than detect. Your own finding that three wording iterations plateaued is the strongest argument for it, and it would close item 1 permanently rather than papering over it with a marker. Worth an issue.

Also, the branch is 15 commits behind main.

clack391:
Adding to @EdmondOware's round rather than opening a second one. I reached their asks #2, #3 and #4 independently and everything in their review holds, so treat this as reinforcement plus four things their pass did not cover. Nothing here reverses any of it.

First, what is right, because it is most of the PR. The diagnosis is correct and the fixture is sound: it validates against both schemas, `check_runnable` reports true, and every field sits inside the `tree-shape.ts` allow-lists. The placement is right. And the behaviour change is real, not a graded illusion. The 08-28 run genuinely surveys the attached source: "Source S1 = 1875 Deed Book 42 p. 118, Patrick purchased land in Schuylkill County seven years before Michael's estimated arrival." That is the survey the skill was missing, read off the tree rather than re-searched. The `unzip` fix is also correct, and I reproduced its exact condition: a zip written with an MS-DOS host byte and backslash separators makes Info-ZIP warn, translate the separators correctly, and exit 1, which `set -euo pipefail` turned into a silent skip of every remaining step.

## The one blocker: the validator passes the response it was built to catch

The body says the validator means an already-sourced fact must be named "or the test fails outright, not a fuzzy judge partial." It does not fail. Run `test_survey_surfaces_already_attached_fan_facts` against the recorded `text_response` of `ut_research_plan_bpx` in `v1_2026-08-27_16-18-15.json`, the run whose SKILL.md hashes to the merge base, and it passes.

I went looking for the innocent explanation first, since that response mentions Patrick seven times. There isn't one. The response notices the source and then declines to use it:

> Patrick is already in the tree (source S1); his relationship to Michael is **undocumented and should not be assumed**. [...] **If** Patrick preceded Michael in the county and was already mining, that is strong contextual evidence [...] Search Patrick's 1880/1900 census, church records, and vital records.

Conditional reasoning about whether Patrick preceded Michael, and a fresh search for records that would establish it, are only possible if the fact's content was never read. This is the failure the new SKILL.md paragraph names in its own words ("noticing it exists; it is not the survey") and the behaviour `judge_context` bullet 5 rules out. The `1875` the check matches is that plan item's own window, which the response itself calls "Patrick's plausible lifespan."

So the judge partial is what caught this, and it is the mechanism the body says it replaces. Two consequences worth naming: no committed run log anywhere shows this validator failing (22 recorded entries, 21 skips and one pass), and because `orchestrator.py` gates the judge on `validators_passed`, the run that does trip it loses the judge dimensions entirely. The 08-27 log's Correctness and Completeness rationales, which name the omission precisely, are exactly what a future reader would want and would no longer get.

**@EdmondOware's word-boundary fix does not close this.** I applied it and re-ran: the validator still passes the 08-27 response, because "Patrick" is a real word there. Both asks are needed and neither substitutes for the other.

**A predicate that does close it.** The discriminating signal is the fact's content, not its date. Word-bounded given name, plus any date token (`date`, `standard_date`, or the bare year extracted from either), plus, when the fact carries a `value`, any two-word gram from that value. Measured against your own two recorded runs: fails 08-27, passes 08-28. It also keeps all five of your validator unit tests green, closes the short-name false positives, and fixes the date-format false negatives below. One caveat: its strength comes from `value` being present, so a sourced fact without one degrades to today's name-plus-date behaviour. Take the shape or reject it, but the "content, not date" distinction is the part that matters.

## Two false negatives in the same predicate

`fact["date"]` has to appear verbatim, so the check only works for the bare-year form. Across the scenario corpus, of 113 sourced facts carrying a date, 9 are a bare year, 53 are `~yyyy`, 47 are ISO `yyyy-mm-dd`, 1 is `yyyy-mm` and 3 are free text. I confirmed the direction: `"~1845" in "...about 1845..."` is false, so a `~yyyy` fact fails against a correct response, and `1 May 1898` fails against "May 1898". That is 104 of 113 for any future `already-attached` scenario. Separately, `date` is optional in the schema (`required: ["id", "type"]`), and a sourced fact without one fails unconditionally no matter what the response says.

## Three things that are one line each, all verified against your committed run

1. `judge_reads_files` is unset. Of the 17 positive research-plan tests, yours is the only one that omits it; the other five omissions are all negatives, and the parent scenario's test sets it. Two of your `judge_context` bullets grade the written plan's content, which the judge cannot read without it.

2. The tags `research-plan-new-plan-for-q-001` and `new-plan-items-planned-status` are missing, so nothing deterministic checks that a plan was written at all, or that its items are `planned`. Ten other research-plan tests carry them. I ran both validators against the fixture's starting state and the plan your 08-28 run actually wrote (`pl_001`, nine items): both pass, so adding the tags cannot redden the committed run.

3. The failure message says "neither the given name nor the fact's date appear" for an AND predicate. I printed it with the name present and the date absent, so it fires while stating something false. Your own `test_fires_when_only_name_present_not_date` exercises that case and matches on the id, so a passing test currently locks the wrong diagnostic in.

Also, `test_skipped_when_no_sourced_fan_facts` asserts a plain return, not a skip, and the validator never skips in that branch. Name and docstring both describe behaviour that does not exist.

## The shell fix has no test, in the file the spec names as pinning it

`docs/specs/feedback-case-spec.md` section 3 says the script is "Pinned by `eval/harness/tests/unit/test_setup_feedback_case.py`." That file has 11 tests, all building forward-slash zips, and none asserts anything about an exit code. I wrote the missing one using the file's own `_build_minimal_zip` and `_run_script` helpers, about 25 lines, building the archive with `create_system = 0`. It passes on this branch and fails on `main`'s script with `assert 1 == 0`, so the guard is real in both directions. Worth adding here rather than trusting the fix to stay.

## Coordination, and one thing that is not yours to answer

Three in-flight changes now sit on this script pair with no cross-reference: this PR on the `.sh`, #1946 on the `.bat`, and the follow-up #1940 records for after #1905 (which merged on the 27th). #1940's `Touches:` line names both `scripts/setup-feedback-case.sh` and `test_setup_feedback_case.py`, and its comment asks for sequencing precisely to avoid conflicting on them. One comment on #1940 naming your hunk is enough.

On marking the test: the mechanics work (`expected_outcome` accepts `xfail`, though nothing in the corpus uses it yet), but if you follow #1946's pattern, note that its `xfail_reason` points at #1967 as the unblock, and #1967's ruling went the other way. Do not inherit that reason.

And to close the ADR-0011 question rather than leave it hanging: I am **not** asking you to move this into a `research_append` precondition. A gate reading a rationale for whether it names a fact is a linguistic check on a free-text field, and that shape was declined for the identical case in #1967 on the 27th. The lane-4 classification is still worth one line in the body, and the honest answer is probably "prose, because the predicate is linguistic." If anyone pursues the structural variant, comparing `plan_item.record_type`, `jurisdiction` and `date_range` against attached facts, that is a question for Dallan and not work for this PR.

Two smaller things while you are in the files. The README says Objective, Questions, Hypotheses and Log are identical to `first-plan-migration-hypothesis`, but `questions[0].rationale` and `hypotheses[0].notes` both gain "for Michael himself." The edit is correct and necessary; the claim is what is wrong, and it hides the fixture's one semantic divergence from anyone who trusts the README instead of diffing. Same file calls I3 "their father" where the data says gender Unknown, and "decade-level dates" where every date is year-precise. Also `names_by_id` is built over `tree["persons"]` and then read by id inside a loop over the same list where `person` is already bound.

Finally, `Closes #1948` is missing, and the guide is explicit that the line is the only thing that moves the card off the board. The body also refers to a "Follow-on issues" section that is not there, and the whole `## Test plan` block is absent, including the paid-eval attestation this PR actually earned.

One thing I could not settle and am not asking you to chase blind. The committed run is redder than the one before it: `ut_research_plan_002` had passed in all six prior runs and fails here on a single `volume_search` fixture miss, `_014` had passed in all six and aborts on `sdk_stream_silence`, and `_001` drops to partial. The new paragraph says "the subject and every other person in scope," and 7 of the 13 scenarios research-plan tests use have a subject carrying a sourced fact, so it fires on 7 of 13 rather than only on the new fixture. All three degraded tests are inside that 7; `_016`, which is outside it, failed for the unrelated pre-existing `localities` ownership reason. `_014`'s own narration is the instruction executing right before it stalls. The run also costs 9.2% more and emits 11.1% more output tokens than its predecessor. None of that is proof, and `eval/CLAUDE.md` is right that one run is not a measurement. But it is the side effect `skill-lifecycle.md` predicts for a prose edit, and a second run is the only thing that separates it from noise.

Suites I ran at `c69e613b`, so you know what is already covered: full harness pytest (2882 passed, 2 skipped), engine vitest (2722 passed), `check_runlogs.py` with the real base and head (exit 0), and both touched test files. The one failure I initially saw was environmental, the gitignored engine `build/` missing in a fresh worktree, and it passes in a normal checkout.

clack391 (changes requested):
Senior review. Detail is in my comment above; this is the blocking record, not a second copy of it.

Requesting changes for one reason, and it is separate from @EdmondOware's round rather than a restatement of it: **the new validator passes the exact recorded response the PR was written to catch.** Run `test_survey_surfaces_already_attached_fan_facts` against `ut_research_plan_bpx`'s `text_response` in `v1_2026-08-27_16-18-15.json`, the run whose `SKILL.md` hashes to the merge base, and it passes. The `1875` it matches is that plan item's own window. So the body's "or the test fails outright, not a fuzzy judge partial" does not hold, and the judge partial is still what caught the regression.

Please read this as **in addition to** @EdmondOware's four asks, not covered by them. Two specifics so the scope is unambiguous:

- His ask #2 (word boundary on the given name) is correct and should land, but I applied it and re-ran: the validator **still** passes the 08-27 response, because "Patrick" is a real word there. Clearing his round does not clear this.
- A predicate that does discriminate is in the comment, with the measurement: word-bounded name, plus any date token including `standard_date` and the bare year, plus a two-word gram from the fact's `value` when it has one. It fails 08-27, passes 08-28, keeps all five of your validator unit tests green, and fixes the date-format false negatives. Take the shape or reject it, but the "content, not date" distinction is the part that has to change.

Not blocking, and listed in the comment with the checks I ran: `judge_reads_files`, the two missing gate tags, the failure message that states an OR for an AND predicate, the missing `create_system = 0` test for the shell fix, three README claims contradicted by the files beside them, and `Closes #1948` plus the absent `## Test plan` block.

To be explicit about what I am **not** asking for, since both @EdmondOware and I raised it independently: do not move this into a `research_append` precondition. A gate reading a rationale for whether it names a fact is a linguistic check on a free-text field, and that shape was declined for the identical case in #1967 on the 27th. Answer the lane-4 question in one line in the body and leave the structural variant to Dallan.

What is right here is most of the PR, and I said so in the comment: the fixture is schema-valid and runnable, the placement is correct, the `unzip` diagnosis reproduces exactly, and the 08-28 run genuinely surveys the attached source. This is a fix-the-guard round, not a rethink.

Cia-3:
@EdmondOware @clack391 — thank you both, and clack391 especially for the reproduction: `v1_2026-08-27_16-18-15.json` was the concrete counter-example I was missing. I re-ran the redesigned validator against that exact stored response before writing anything else, since "I fixed it" isn't worth posting without that check.

**The core fix.** The validator now requires, in the same paragraph: the person's given name (word-bounded) AND fact-specific content — a date token (`date`, `standard_date`, or a bare year extracted from either, so `~1845`/`1908-03-12`/"Abt 1850" all resolve) AND, when the fact has a `value`, a two-word gram from that value. A fact missing `value` degrades to name+date; a fact missing `date` degrades to name+value; a fact with neither is out of scope (nothing to verify), the same as a person with no sourced facts at all — that closes the unconditional-fail gap on an undated fact, separately from the format gap.

Verified against every stored `ut_research_plan_bpx` response I have, at zero cost, before spending anything on a live run:
- `v1_2026-08-27_16-18-15.json` (the counter-example) — now **fails**, correctly.
- `v1_2026-08-28_06-14-42.json`, `v1_2026-08-28_10-59-10.json`, `v1_2026-08-30_04-40-39.json` — all still **pass**, correctly.

11 unit tests now (up from 7), including one that reproduces the 08-27 pattern directly (name+date co-occurring as a *new-search proposal*, not a citation) and one each for the non-bare-year date format and the two degradation paths.

**Everything else, itemized:**
1. `judge_reads_files` + the two gate tags — already fixed before this round; re-confirmed both newly-gated validators pass on the actual 08-28/08-30 runs.
2. Failure message — rewritten again for the new multi-signal check; still never claims "neither" when only one condition is missing.
3. `test_skipped_when_no_sourced_fan_facts` — renamed to `test_passes_vacuously_when_no_sourced_fan_facts` and its docstring corrected. You're right, it never skipped.
4. `names_by_id` — removed; `given` is read directly off `person` in the same loop now.
5. The shell-script test — I could not reproduce the real backslash-path warning through `zipfile` alone; it silently normalizes any backslash in an entry name to a forward slash on write, confirmed directly (`info.filename` came back with `/` even after I passed `chr(92)` in the source string). The two committed tests stub `unzip` on PATH to force exit codes 1 and 2 instead, which proves the fix's actual branch logic (continue vs. abort) rather than the specific real-world trigger. clack391 — if you're willing to share how you built a zip that actually trips it, I'll add that as a third test; genuinely curious what I'm missing.
6. Lane-4 (ADR-0011): prose, not a `research_append` precondition — added as one line in the body. Agreed this is linguistic, not structural, and #1967 already settled the identical question.
7. The three README claims — fixed (the "identical to migration-hypothesis" claim now names the one real divergence, I3 no longer gets an assumed gender, "decade-level" corrected to year-precise). Also updated the stale "verify before committing" caveat, since two review rounds and multiple full-suite runs is exactly that verification.
8. `Closes #1948`, the actual Follow-on issues content, and the Test plan block — all added; they'd gone missing in a paste, not a decision.
9. Commented on #1940 noting this PR's hunk (lines 88-92) sits before both of that issue's defects (now `:112` and `:119`/`:123`) — no overlap today, flagged for the rebase-collision risk given three in-flight changes on the same file.

**The one thing I'm not resolving:** clack391's "redder run" observation. I don't have a clean answer, and won't manufacture one — but the two subsequent full runs I've since committed (08-28 and 08-30) both show `_002` and `_014` back to pass, which is at least consistent with the single-run-noise read rather than a systemic cost/token regression from the new paragraph, though I know that's not proof either.

Fresh full-suite run + annotation for the redesigned validator is committed. One more data point from that run worth sharing: `ut_research_plan_bpx` actually **failed** this time — a genuine model miss (the response says Patrick's source "hasn't been formally extracted" and defers reading it to a later plan step, never stating what it says), matching the documented ~50% rate. The validator caught it cleanly, with no judge dimensions recorded for that test since `orchestrator.py` gates the judge on `validators_passed` — exactly the mechanism clack391 flagged. Good confirmation the redesigned check is doing its job on a real, unplanned sample rather than just the ones I went looking for.

Re-requesting review from both of you.

EdmondOware (changes requested):
The redesign is right, and the discipline around it is what makes it reviewable. You ran the new predicate against the stored 27 August response before writing the reply, and you reported that the newest committed run failed this test on a real model miss rather than reshooting until it went green. Writing the measured rate into the docstring was the right call too, since candidate retention keeps only five logs per skill and the runs behind that number are already pruned out of the corpus. The degradation paths are well chosen: a fact with no value falling back to name plus date, and one with neither date nor value dropping out of scope, close real holes without softening what the check is for.

Two things.

1. Reconcile the fixture's `judge_context` with the SKILL.md paragraph this PR adds. The new prose says "Person X already has a source attached" is noticing a source exists and is not the survey, and it asks for the fact's date, place and value to be written down. The fixture still tells the judge that "a clear statement ... that acknowledges Patrick's already-documented 1875 arrival satisfies this" and that "a blunt one-sentence acknowledgment is enough to pass". Both lines were written before the tightening and were not revisited when the validator moved to requiring fact content. This is not cosmetic, because the judge is gated on the validator. A response reading "Patrick (I2) is already documented in Schuylkill County in 1875, where he acquired property seven years ahead of Michael" clears the fixture's stated bar and fails the validator, since no word pair from the fact's value survives that paraphrase, and the judge never runs to disagree. Do not close the gap by widening the validator's content signal to the fact's place: the 27 August response names the place in the same paragraph, so that would let it straight back through. On cost, this does not need a paid rerun. The judge was skipped on this test in the committed run, with zero dimensions recorded, so a change to this test's `judge_context` cannot alter any outcome recorded in it. Say that in the body when you ask for the label.

2. Attribute the pass rate measurement to yourself. The docstring reads "Measured directly (PR #2004 review, EdmondOware)", and the same docstring uses "(clack391)" to mean the person who found the thing. I asked for that measurement, you ran it.

clack391:
Senior re-review. **Your round-2 blocker is closed, and I verified it in both directions before writing anything else.** What follows is a narrower successor to it, not a restatement, plus two coordination items that are not yours to decide.

## What is settled

I recovered `v1_2026-08-27_16-18-15.json` from `dba7e414a` (this branch's head commit deletes it under retention) and ran both predicates against its recorded `text_response` by importing the validator module directly, rather than re-running your path:

| predicate | 08-27 counter-example | 08-28_06-14-42 / 08-28_10-59-10 / 08-30_04-40-39 |
|---|---|---|
| the one at `c69e613b` | **PASS** (the defect I blocked on) | pass |
| the one you shipped | **FAIL** | pass |

The content signal is what flipped it, and the word boundary alone still does not. The redesign is right.

Three more things I owed you:

1. **The "redder run" question I could not settle in round 2 is settled, as noise.** Across the five committed runs `_002` goes fail then pass four times, `_014` aborts then fails then passes three times, `_016` fails twice then passes three times. Every one of the eight non-stable tests flips in both directions. Your read was correct and my concern is withdrawn. See the one narrower successor under "Not blocking" below, which is a different shape.
2. **`judge_reads_files`, both gate tags, the renamed vacuous test, `names_by_id`, the #1940 comment, the `Closes` line for issue #1948, the Test plan block and the lane-4 line are all done**, and I checked each rather than reading the checklist. The three README corrections I verified against the data: a structural diff confirms `questions[0].rationale` and `hypotheses[0].notes` are the only divergences, I3 really is `gender: Unknown` with an empty `given` and no facts, and every date in both files is a bare year.
3. **You asked how I built a zip that trips `unzip`, and I owe you a correction rather than a technique.** `zipfile` does not normalise backslashes on POSIX. `ZipInfo.__init__` replaces `os.sep`, and `os.sep` is `/` on Linux and macOS, so the replacement is a no-op there and the backslashes survive verbatim. The load-bearing part is the host-OS byte, not the separator. Measured on UnZip 6.00:

```
create_system = 0  ->  exit 1, "appears to use backslashes as path separators", extracts to results/log_006.json
create_system = 3  ->  exit 0, no warning, extracts a file literally named results\log_006.json
```

So the real reproduction is reachable in-repo and on CI. This drops in beside your two stub tests using the helpers already in the file, and I ran it both ways: green at head, red against `origin/main`'s script with `assert 1 == 0` and the genuine warning text.

```python
def _build_windows_zipped_bundle(zip_path: Path, slug: str) -> None:
    """`_build_minimal_zip`'s bundle written the way a Windows zip tool writes
    it: the MS-DOS host byte plus backslash separators. The host byte is
    load-bearing. With the default Unix host byte the same names extract
    silently and unzip exits 0, so the fix is never reached."""
    src = zipfile.ZipFile(zip_path.with_suffix(".src.zip"))
    with zipfile.ZipFile(zip_path, "w") as z:
        for info in src.infolist():
            out = zipfile.ZipInfo("placeholder")
            out.filename = info.filename.replace("/", "\\")
            out.create_system = 0
            z.writestr(out, src.read(info.filename))
    src.close()


def test_tolerates_real_windows_zipped_bundle(tmp_path, monkeypatch):
    slug = "feedback-windows-zip"
    src = tmp_path / f"{slug}.src.zip"
    _build_minimal_zip(src, slug)
    zip_path = tmp_path / f"{slug}.zip"
    _build_windows_zipped_bundle(zip_path, slug)
    monkeypatch.setenv("HOME", str(tmp_path / "home"))

    probe = subprocess.run(
        ["unzip", "-q", str(zip_path), "-d", str(tmp_path / "probe")],
        capture_output=True, text=True, encoding="utf-8",
    )
    if probe.returncode != 1:
        pytest.skip(f"this unzip build does not emit the warning (exit {probe.returncode})")

    result = _run_script(str(zip_path))
    assert result.returncode == 0, result.stderr
    dest = tmp_path / "home" / "feedback" / slug
    assert (dest / ".feedback-repo-root").exists()
    assert (dest / ".git").is_dir()
    assert (dest / "_feedback" / "feedback.json").exists()
```

Your two stub tests are honest guards, so this is a third test rather than a replacement. What does need changing is the reason recorded beside them: the docstring at `test_setup_feedback_case.py:296` and the PR body both state the real reproduction is out of reach, and a future reader will believe it.

## The one thing still blocking, and it is the same sentence as last time

The body still says an already-sourced fact must be named "or the test fails outright, not a fuzzy judge partial." That is true of the one response I could hand you in round 2. It is not true of its class, and I only found this because a blind pass attacked the predicate with shapes I had not thought to write. My own pass verified it against the recorded corpus and stopped there, which was not enough.

**A response that read nothing still passes, in the fixture's own configuration with `value` present.** All three of these pass:

```
"I have no information about Patrick Sheahan's whereabouts. I will search the
 Schuylkill deed book series from 1875 onward to find out whether he purchased
 land there at all."
```
```
"## Research plan for q_001
 - Item 1: 1880 US census, Schuylkill County, for Michael Sheahan.
 - Item 2: FAN, Patrick Sheahan (I2). Search the Schuylkill County deed book
   indexes, 1875-1905, to establish whether he purchased land before Michael."
```
```
(a second sourced fact added; only the irrelevant one echoed)
"Patrick Sheahan (I2) is recorded as a coal miner at Tamaqua colliery in 1880.
 Nothing else is known about him; plan items 1-6 concern Michael."
```

**This is not a contrived shape, and here is the measurement that convinced me.** Your newest run, `v1_2026-08-31_10-37-11`, is a genuine model miss. Its response already contains two of the six value grams, `purchased land` and `land deed`, and it gets them from prose that *denies* a purchase: "Coal-region land was typically *leased*, not purchased, land/deed records". The only reason that response is rejected is that `1875` appears nowhere in it.

So across the two recorded cases the two signals are each carrying one, alone:

* on 08-27 the date is present in a plan item's own search window, and only the value grams reject it;
* on 08-31 the value grams are present by accident, and only the absent date rejects it.

A response carrying both, which is any FAN plan item with an 1875 window that also says "purchased land" or "land/deed" near the name, passes while having read nothing.

**Two things compound it.**

`value` is the rarest field in the corpus. Across every scenario tree in `eval/fixtures/scenarios/` there are 113 sourced facts and exactly **one** carries a `value`: the F4 this PR authored. All six other scenarios with a non-subject sourced person are date-only. Strip the `value` and the shipped validator passes the 08-27 counter-example again. A one-word `value` does the same, because `_word_grams(s, 2)` returns the empty set for a single word and an empty gram set is read as satisfied. The fixture's own `I1/F3` is `"value": "Miner"`, so that shape is not hypothetical. The docstring presents the fallback as an edge case, and measured it is the default.

The same-paragraph rule is both unguarded and harmful. Replacing `re.split(r"\n\s*\n", response_lower)` with `[response_lower]` leaves all 28 unit tests green, which is CLAUDE.md's "a check that cannot fail reads as coverage and is worse than no check at all." It also fails an ordinary correct response:

```
"**Already attached (surveyed before planning)**

 Patrick Sheahan (I2)

 Source S1 records a land purchase in Schuylkill County dated 1875 (Deed Book 42, p. 118)."
```

And I measured that holding the content requirement while dropping the scoping changes no verdict on any of the six recorded responses, so it is currently paying a false-positive cost for nothing.

## The fix is small, and the first item is one word

**(a) Make it tier 2.** Rename `test_survey_surfaces_already_attached_fan_facts` to `report_survey_surfaces_already_attached_fan_facts`, in **both** anchors: the `def` in the validator and the `... as check,` import at `tests/unit/test_research_plan_validator.py:35`. Renaming only the `def` breaks that import.

I ran it: 28 of 28 unit tests still pass, the check still fires with `passed: False`, and through `run_validators` it comes back `reporting_only: True` with **zero gating failures**, so the judge runs and receives the finding as an observation instead of being skipped.

`docs/specs/unit-test-spec.md:1123` describes tier 2 as being for exactly this case: "the harness can detect a pattern but cannot decide whether it is wrong, that decision belongs to the judge." The bypasses above are the proof that it cannot decide. This also dissolves @EdmondOware's round-1 objection that a deterministic validator on a coin-flip behaviour makes normal variance loud rather than regressions loud, and it stops a red from wiping every judge dimension on the half of runs that fail. `docs/skill-lifecycle.md` asks the PR to say which tier it chose, and the body does not, so please say so either way.

**(b) Drop the same-paragraph scoping**, per the measurement above.

**(c) Correct the two claims that the measurements contradict**: the body's "fails outright, not a fuzzy judge partial" and "closed the class going forward", and the docstring's framing of the no-`value` path as an edge case rather than the corpus default.

To save you a round trip, I tested the obvious in-validator fix for the `value` gap and **it is the wrong fix, so please do not apply it**. Failing loudly when a tagged scenario's sourced fact has no `value` keeps this fixture green and does fire on the stripped shape, but it reds `test_extracted_year_matches_non_bare_year_date_formats`, which deliberately asserts that fallback passes. Widening the content signal to `place` is also out, for the reason @EdmondOware already gave. What survives testing is the docstring correction, plus optionally a fixture-level precondition that an `already-attached` scenario must give its sourced fact a multi-word `value`, which touches no unit test.

**(d) @EdmondOware's two round-3 asks are untouched by any commit** (the only commits after that review are a main merge and a run-log refresh). I reproduced ask #1 independently rather than relaying it: their example fails the validator, and so does the fixture's own description wording "a 1875 land purchase", while `judge_context` bullets 3 and 6 tell the judge that both pass. Because the judge is gated on the validator, it never runs to disagree.

## Not blocking, one line each

* `test_research_plan.py:804`. The message says "no sourced fact's distinguishing content (date or description) appears alongside it in the same paragraph", and I reproduced it firing on the 08-27 response whose single Patrick paragraph does contain `1875`. It is round-2's disjunction-for-a-conjunction in new wording. A correction to my own first draft here: I had written that its guard asserts only that "neither" is absent, which is wrong, it asserts three things. So I re-measured instead: replacing the defective clause with an obviously wrong string leaves all 28 tests green, so the guard covers the half of the message that is right and not the half that is wrong.
* `test_setup_feedback_case.py:301`. "does the script continue (1) or abort (>=1)?" The script aborts on `-gt 1`, that is >=2. As written it says the script aborts on the very code the fix exists to tolerate.
* `test_research_plan.py:688`. The tag-gating example cites scenario `fixtures`, which does not exist among the 92. The substance is right, and in fact stronger than stated: of the 22 research-plan tests, only yours has a non-subject person with a checkable sourced fact, so an ungated check would false-positive on nothing today.
* `test_research_plan.py:718`. The docstring's single load-bearing citation, `v1_2026-08-27_16-18-15.json`, is deleted by this branch's own head commit, and it was also the only committed run at main's SKILL.md, so the PR body's acceptance-check evidence goes with it. All five pre-change logs go too, which makes `eval/CLAUDE.md`'s "get the non-pass rate across the committed run logs" unrunnable for research-plan afterwards, and two annotations in the newest `.ann.json` lean on exactly that history.
* `test_research_plan.py:674`. "6 valid re-runs, 3 pass / 3 fail" is not derivable from the repo. All five committed runs snapshot the same final SKILL.md sha256, so all five are post-wording, and they give 3 pass and 2 fail. I am filing this as unverifiable rather than wrong, since a `--test` scratch run leaves no committed artifact. One sentence saying where the other runs live would close it.
* PR body, "Plan". The acceptance check says the test fails on main "(model stays silent about the sibling's already-sourced fact)". That run mentions Patrick seven times and states he is in the tree with S1. The failure mode is noticing without reading, which is the distinction this PR draws.
* Fixture line 28. "(census, land/deed, tax, church, per the localities entry)". `loc_001` names deed, land, tax, church, parish, naturalization, probate and vital records, and never census.
* README line 23 puts `quality: 3` on S1. `quality` is a source-reference field (`TREE_SOURCE_REF_FIELDS`, `tree-shape.ts:30`) living on F4's ref. Line 13 says the two divergent strings "both end" with "for Michael himself"; the rationale continues past it. The body also says #1940 carries "two other" defects, and its title now carries three since #1957 folded in.
* The validator raises rather than asserts on `project: null`, on `tags: null`, and on a non-string `value`, and `validator_runner` turns each into a gating failure attributed to the skill. This is the fourth instance of this class in recent PRs (#1741, #1857, #1882), so I would rather see a shared coerce helper for validator inputs than a fourth one-off, and that is a separate piece of work, not yours here.
* `_extract_year` and `_YEAR_RE` are the third four-digit-year extractor in the repo, and the three disagree on what a year is: this one accepts 1000 to 2049, `_EMBEDDED_YEAR_RE` in the sibling `test_record_extraction.py:316` accepts 1500 to 2099, and `_YEAR` in `e2e/author.py:104` accepts 1000 to 2099. `author.py:_birth_year` also reimplements the same `standard_date` then `date` fallback you write inline. `validators_lib.py` is the shared home and this file already imports from it.
* SKILL.md line 92 asks the model to state the fact "in your own working notes". `text_response` is assembled from `TextBlock` only (`skill_runner.py:566`), so a model that complies by putting the survey in its reasoning fails a check it obeyed, and "working notes" names no artifact this skill produces. "State it in the plan's rationale" would make the prompt and the guard name the same surface.
* `docs/specs/feedback-case-spec.md` section 3 says the script is pinned by `test_setup_feedback_case.py`, which is still true, but its guarantee table gains no row for the new contract of tolerating exit 1 and aborting on 2 or more.
* `docs/skill-lifecycle.md` step 7, "Confirm it in Cowork", is marked **Required** for work that came through the alpha-feedback door, which this did. The test plan has pytest, `make eval-skill` and an N/A. Either run it or add a line saying why it was skipped.
* One narrower successor to the "redder run" concern I withdrew above, because it is a different shape and not a grading flip. `ut_research_plan_005` fails `test_research_json_validates_schema` with "plans/1/items: [] should be non-empty" in two of the five post-change runs, and in none of the five pre-change runs at the merge base, which I read one at a time. The underlying cause is engine-side and not yours: `research.schema.json` declares `minItems: 1` for a plan's items and `validator.ts` has no matching check, so `research_append` persists an empty plan shell. That is already tracked as issue #2051, which names the same `validator.ts` site plus a second defect I had missed, namely that the worked example the tool attaches to a rejection instructs `"items": []`. Nothing to do here beyond knowing your run logs carry two more instances of it. I could not establish that the new paragraph causes the model to write it, and with eight of this skill's tests flipping across five runs I do not think a single run can.

## Two questions for @DallanQ, not for the author

* **#2027 is open and ships the same `unzip` exit-code fix on the same lines of `scripts/setup-feedback-case.sh`**, and adds its own test to the same test file. Neither PR references the other. One hunk has to go, and I do not think it is obvious which: this PR's stub tests exercise both branches on a Linux runner, whereas #2027's archive leaves `create_system` at the Unix default, so its test exits 0 on CI and never reaches the fix, which its own docstring concedes.
* **#2033 is open on the same `research-plan/SKILL.md` and the same `eval/harness/validators/test_research_plan.py`**, and commits its own research-plan run logs. Whichever merges second needs a fresh paid run, because its committed log's SKILL.md snapshot hash will no longer match head. Issue #1948 stays open until this lands, so the sequencing is worth deciding now rather than after a conflict.

## What I ran

Harness pytest 2975 passed and 2 skipped. The single failure I first saw was the gitignored engine `build/` missing in a fresh worktree, and it passes in a normal checkout, so it is not a defect here. Engine vitest packaging 30 files and 609 tests passed, with a warn-only note that `research-plan/SKILL.md` grows from 27.1 KB to 28.5 KB on every invocation. `check_runlogs.py` with the real base and head exits 0, with one warning naming `ut_research_plan_bpx` and `ut_research_plan_005` as producing no graded dimensions. Targeted: `test_setup_feedback_case.py` 13 passed, `test_research_plan_validator.py` 28 passed, `test_scenario_fixtures.py` 259 passed. CI is green on all twelve checks at `0b3316f3`.

Beyond the suites, the things I mutated rather than read: the paragraph split, the fact's `value`, the failure message's defective clause, the tier-2 rename, and `origin/main`'s script under both your stub tests and my new one. Every mutation was restored from a `cp` backup and `git status` is clean.

Cia-3:
@EdmondOware — both fixed.

1. Tightened both `judge_context` bullets to require the fact's actual content (the deed/land-purchase description), not just an acknowledgment of Patrick's date or place. Deliberately did not widen the validator's own signal to accept place instead of value — you're right that the 08-27 counter-example already names Schuylkill County in the same paragraph as Patrick, so that would let it straight back through. The fix is entirely on the fixture side.

No re-run needed, for the reason you gave: the currently-committed run log already skipped the judge on `ut_research_plan_bpx` (validator failed, gating it out with zero dimensions recorded), so this wording change can't alter anything already in that log. Requesting the `eval-cosmetic-skip` label for this PR on that basis — happy to have you or another senior confirm the reasoning holds before applying it.

2. Reworded the attribution — you asked for the measurement, I ran it. Docstring now says "per EdmondOware's PR #2004 review ask" instead of crediting you with running it.

Both pushed (`9b0a52f0`). Also worth flagging since it's why a reviewer couldn't open this PR earlier: main advanced twice since the last round (67 commits, then another 10) and the branch picked up two genuine conflicts along the way -- both add/add, both because main independently added an unrelated deterministic validator suite (issue #1866) at the same file/location this PR's own validator work touches. Resolved by keeping both sets of functions/tests; full harness suite (2973 passed, 4 skipped) confirmed clean against the merged state. Re-requesting review from both of you.

Cia-3:
Two sequencing decisions from Dallan, both now applied:

1. **#2027 supersedes this PR's `unzip` fix.** Dropped `scripts/setup-feedback-case.sh`'s fix and its two unit tests entirely from this PR (`dea933f0`) — #2027 fixes the same bug with a cleaner implementation and will land instead.
2. **#2033 lands first.** It overlaps this PR's touched files (`research-plan/SKILL.md`, the validator file) with no textual conflict today. Holding further `eval-skill` re-runs here until #2033 merges, then rebasing and re-running once.

Merge conflicts from main (67 commits, then 10 more) are resolved either way — full harness suite green against the merged state (2973 passed, 4 skipped).

EdmondOware:
Both fixes verified, and I checked them against the tree rather than the reply.

The `judge_context` now requires the fact's content and names the failing paraphrase explicitly: *"it must state what the source's content actually is (a land purchase, a deed), not just that Patrick already has a source or a date."* The final bullet grades on whether the content *"not just its existence, date, or place"* reached the reasoning. The "blunt one-sentence acknowledgment is enough to pass" line is gone. And you were right not to widen the validator's signal to place, for the reason it would have failed: the 08-27 response names Schuylkill County in the same paragraph. Attribution now reads "per EdmondOware's PR #2004 review ask", which is accurate.

**On the `eval-cosmetic-skip` request, the reasoning holds for your file and not for the PR.** You argued it on the basis that the judge was skipped on `ut_research_plan_bpx` with zero dimensions recorded, so a wording change cannot alter anything in the committed log. That is true. But the gate is naming **twelve** files, and only one of them is yours:

```
eval/fixtures/mcp/external-links-search-{blount,denmark,feliciana,kentucky,
                                        pa-census,rowan-county,schuylkill,texas}.json
eval/fixtures/mcp/place-population-{muhlenberg,schuylkill}.json
eval/fixtures/mcp/place-search-all-any.json
eval/tests/unit/research-plan/first-plan-fan-already-sourced.json   <- yours
```

The eleven came from #2014 on 08-31 and from #2033, and they are fixture shape and compaction changes, which alter what the model sees from those tools. That is behaviour-affecting by construction, so a waiver would be asserting something the diff contradicts.

**The good news is that this is moot, and your own plan already handles it.** #2033 merged earlier today, so the sequencing you set on 09-01 is unblocked: rebase, then one re-run. That re-run is needed regardless of the label, because #2033 itself edited eight of those eleven fixtures, so merging main widens the drift rather than closing it. The committed log is also scored against judge prompt `03f306ff` while the current one is `9451d105`, so its numbers would move anyway.

Holding my changes requested only until the re-run lands. Nothing in the substance is outstanding.

EdmondOware:
@Cia-3 dropping the `unzip` fix and its test outright once Dallan ruled that #2027 supersedes it was the right call. It would have been easy to keep arguing for the version already written, and cutting it closed two open items along with it.

Your blocker is gone. #2033 merged, so the plan you set out on September 1 to hold further runs until it landed is now clear.

Both of my items are done and nothing further is owed on either.

Four items from @clack391's second review are still open, and they are what remains on the substance.

1. Line 28 of the test fixture gives the record types as "census, land/deed, tax, church, per the localities entry". `loc_001` names deed and land, tax, church and parish, naturalization, probate and vital records, and never census. Correct the list to match the entry it cites.

2. The scenario README puts `quality: 3` on `S1`. `quality` is a source reference field (`TREE_SOURCE_REF_FIELDS` in `tree-shape.ts`) and in this fixture it sits on F4's ref, not on the source. Move the claim to where the field actually lives. The same README says the two divergent strings both end "for Michael himself", and the rationale continues past that point, so that sentence needs correcting too.

3. `SKILL.md` line 92 asks the model to state the fact "in your own working notes". `text_response` is assembled from `TextBlock` only, so a model that complies by putting the survey in its reasoning fails a check it obeyed, and "working notes" names no artifact this skill produces. Point the instruction at the plan's rationale so the prompt and the guard name the same surface.

4. `_extract_year` and `_YEAR_RE` in `test_research_plan.py` are the third year extractor in the repo and the three disagree on what a year is. `validators_lib.py` is the shared home and this file already imports from it, though it carries no year helper yet, so this means adding one there and moving the other callers across.

Then two mechanical things. Rebase onto main, and run the skill's eval once with the run log annotated and committed. The MCP fixtures behind `external_links_search`, `place_population` and `place_search_all` changed on main after these run logs were recorded, so the committed logs no longer reflect what the model sees from those tools. Worth confirming nothing else holds the `research-plan` eval slot before you start.

One loose end that is not clearly yours. Step 7 of `docs/skill-lifecycle.md`, "Confirm it in Cowork", is marked Required for work that arrived through the feedback door, and the test plan has an N/A against it. Either run it or write down the reason it was skipped.

Cia-3:
@EdmondOware @clack391 — the blocker is closed, and going back through the full thread I found more than the four items from your latest message, @EdmondOware.

**The big one first, since it should have landed a week ago:** clack391's 08-31 review found the validator (as it stood after round 2) could be defeated by a response that read nothing — same-paragraph scoping was itself a false-positive hole, and a sourced fact without a `value` (the corpus default) degrades to name+date, satisfiable by coincidence. That review's proposed fix — tier-2 rename, drop the scoping, correct the docstring's contradicted claims — never actually landed in code; only the smaller items from a parallel thread did, which is why your four-item list still reads as "what remains" when there was more sitting underneath it. Implemented now: `report_survey_surfaces_already_attached_fan_facts` (tier 2 — observation for the judge, not a gate), whole-response check instead of per-paragraph, corrected docstring. Added a regression test for clack391's own demonstrated false-negative (a correctly-cited response whose citation sits in its own paragraph, which used to fail wrongly).

**Your four items, all fixed:** the fixture's record-types parenthetical now matches `loc_001` exactly (census dropped, probate/naturalization added); the README's `quality: 3` now correctly attributed to F4's source reference, not S1, and the "both end" claim corrected to note only one field actually does; SKILL.md now says "the plan's rationale" (already the term used one paragraph later in the same file) instead of "your own working notes"; and the `_extract_year` duplication — added `validators_lib.extract_year()` and migrated this file's own usage. Filed #2330 for the other two callers (`test_record_extraction.py`, `e2e/author.py`) rather than touching those here — out of this PR's scope, and the former has already produced two merge conflicts against unrelated in-flight PRs during this PR's own review cycle.

**Also cleared:** the other one-line items from the 08-31 review (the nonexistent `fixtures` scenario citation, the pruned run-log citation, the unverifiable "6 runs" figure — re-measured directly at 3 pass / 2 fail across the 5 actually-committed runs, the OR/AND failure-message phrasing).

**Rebased and re-ran** against current main (200 commits behind — #2033's fixture changes are in). Fresh `make eval-skill SKILL=research-plan` committed: 17 pass / 1 partial / 4 fail / 1 xfail of 23. Checked every non-pass against history before annotating — none touch the already-attached logic; `ut_research_plan_bpx` landed partial on exactly the "noticing without reading" pattern this fixture exists to catch, which is the mechanism working, not a defect.

**One thing I can't do from here:** `docs/skill-lifecycle.md` step 7 ("Confirm it in Cowork") is Required for alpha-feedback-sourced work, which this is, and it needs a live Cowork session with the Research Viewer open — outside what's available in this environment. Flagged in the PR body rather than silently marking N/A; someone with a Cowork install needs to either run it or make the call to skip it with a stated reason.

Full diff and reasoning in the updated PR body.

EdmondOware (approved):
Approved.

All four items are fixed and check out against the tree. The fixture parenthetical now matches `loc_001`, and dropping vital registration with the 1906 reason rather than silently is the better fix. `quality: 3` sits on F4's source reference with `TREE_SOURCE_REF_FIELDS` cited. The prompt and the guard now name the same surface. `validators_lib.extract_year()` exists with this file migrated, and #2330 carries the other two callers.

Going back through the thread and finding that clack391's 31 August validator redesign had never reached code is worth more than any of the four. That gap would have survived the merge and left a validator that a response reading nothing could satisfy.

Two things, neither of which needs another run.

1. **File the fixture defect you surfaced, and reframe two of the four failures around it.** Collection `1999196` carries contradictory data in two fixtures that both load in both affected tests:

   | fixture | recordCount | personCount | implies |
   |---|---|---|---|
   | `collections-search-pennsylvania.json` | 893214 | 893214 | fully indexed |
   | `collections-search-schuylkill.json` | 2304118 | 0 | browse and image only |

   Same id, same title. So `test_research_plan_availability_claim_matches_counts` can fail the model whichever way it characterises that collection, and in this run it did both: `_002` called it browse and image only and was failed against 893214, `_q7m` called it indexed and was failed against 0. Neither is winnable. It is an eval defect rather than a skill one, it predates this branch, and it deserves its own card.

2. **Say in the body that three of the four failures are validator failures.** `_002`, `_005` and `_q7m` produced no graded dimensions at all, which is a different thing from a judge scoring an answer down, and the gate says as much in its own warning. With the two above set aside as unwinnable, what is left is `_005` on `fallback_for` and `_001`, and `_001` is annotated honestly, first time failure named as such. That reads considerably better than four failures reported flat.

Step 7 of `docs/skill-lifecycle.md` is not yours to close and nothing is owed from you on it.

Issue #1959, opened 2026-08-27
apps/server tests unrunnable on Windows: seeded search turns time out (504 after 122s), and _handle_stream has no turn timeout at all


benter-070:
> **Take the cluster, not just this card (added 2026-08-31, at the lead's direction).**
> This is one of three open cards on the hosted chat's turn and session state, and their
> `Touches:` lists overlap. If you pick this up, **ask to be assigned issue #1729 and issue
> #2062 as well**, and work them in that order — this card, then issue #1729, then issue
> #2062. Issue #1729 is the reopen-a-session dead-panel bug and owns the same two client
> files; issue #2062 is the hosted chat answering the previous message, which is undiagnosed
> and routed `senior` on 2026-09-02. Three people editing `v1.py`, `ChatPane.tsx` and
> `chatEvents.ts` at once is the shape that produces merge races.
>
> **Corrected 2026-09-02 (`/review-ready` on issue #2062).** This note originally said #2062
> "may well be caused by the very gap this card closes". That is refuted: this is a `/v1`
> card, and the browser never calls `/v1` — `grep -rn "/v1" apps/web/src` returns 0,
> `apps/server/app/v1.py:1` declares the bearer-only external-chatbot surface, and the web
> client holds one long-lived WS straight to the sandbox
> (`apps/web/src/transport/SessionConnection.ts:127`). All four reports behind #2062 are
> `Platform: web`. **Landing this card cannot remove #2062's cause.** The cluster still
> travels together for the shared client turn-state files, but not for that reason.

> **Reviewed 2026-08-27 before junior handoff.** Problem 2 is a recorded decision, not an
> oversight, and the obvious idle-timer fix can never fire. Decision recorded below; the
> original body follows.

**Decision (2026-08-27):** `_handle_stream` gets an **idle** cap, not a total turn cap — a
total cap was rejected because streaming is the documented home for long tool-running turns
(`public-rest-api-spec.md:250`, item 4).

Add `v1_stream_idle_seconds: int = 300` to `apps/server/app/config.py` beside
`v1_turn_timeout_seconds`. In `_handle_stream`'s `gen()`, track the time since the last WS
frame whose `type` is **not** `"ping"`; when it exceeds the cap, emit `event: error` with
`{"type":"error","text":"..."}` naming the `turn_timeout` code, then the normal terminal
`event: done` with `finish_reason:"error"`, then fall through the existing `finally` (close
[...]

Issue #1932, opened 2026-08-26
Hosted onramp collects closed-set answers as free prose — no way to present selectable options


benter-070:
> **Reviewed 2026-09-14 before junior handoff.** No decision needed — the 2026-08-26
> ruling and the 2026-08-31 narrowing settle every fork. One correction is load-bearing:
> `ExperienceLevel` is a **type-only** union and cannot be imported as a value, so item 1
> is rewritten with the exhaustive-`Record` mechanism that actually binds the UI to the
> enum. A "How to verify" section is added.

> **Scope reduced 2026-08-31 (lead). The site-access question is being dropped
> from the onramp entirely, so there is no site-access chip set to build — and the
> `other`-chip question this card was holding is moot.** Researchers have been
> complaining about how many questions they are asked before they can get started,
> so `init-project` will stop asking about site access and assume the researcher
> has access to every site; if they do not, they can say so when we hand them a
> URL. Dropping the question from the plugin is its own card (see below) and is a
> different lane — this one is hosted-web-only and shrinks to **one chip set,
> experience level**.

---

**Touches:** apps/web/src/components/ChatPane.tsx (chip row + composer insert),
apps/web/src/components/experienceChips.ts (new — the pure insert/replace helper,
split out so it is unit-testable without a DOM, the same split chatEvents.ts got
for issue #1312), apps/web/src/components/__tests__/experienceChips.test.ts (new),
apps/web/src/styles.css (a new `.optionChip` class — do not reuse `.toolChip`,
which is the tool-call chip and is styled for a different state machine).
`apps/web/src/components/chatEvents.ts` needs **no edit**: the option list is held
in the client and there is no new event kind. Nothing in `packages/engine/`,
`apps/server/` or `packages/schema/` changes — so no `make sandbox-image`, and no
paid eval run, because no SKILL.md, rubric or judge is touched.

## What the tester asked for

From the alpha feedback in #1904 (Brian, FamilySearch, hosted web, 2026-08-25).
He stopped about three minutes into a brand-new project to comment on the onramp
rather than on the research:

> My comments are around the beginning. I am asked a lot of questions, which I
> can answer. It would be easier if it presented a list of things to choose from
> and let me just select the correct genealogy experience, which sites I have
> access to, etc. Then it could feed it to the AI. (This is probably for a later
> stage, but I wanted to mention it.)

He flagged it `worked_as_expected: true` — nothing misbehaved. The agent asked
[...]

PR #2005, opened 2026-08-28
init-project: record free genealogy access instead of rejecting it (#1904)

benter-070:
Found while triaging #1904 (closed — the report itself is #1932). This PR closes
no issue by design: #1904's content was Brian's onramp request, not the defect
fixed here, and filing a card for work already written would be ceremony.

## What went wrong

Triaging the alpha feedback in #1904 (Brian, FamilySearch, hosted web). His
report was about the onramp's UX — filed separately as #1932, since no plugin
change can fix it. The transcript exposed a second, unrelated defect, and that is
what this PR fixes.

He answered the access question:

> I have access to FamilySearch, Ancestry, MyHeritage, and FindAGrave.

The agent's reasoning turn planned to store
`["FamilySearch", "Ancestry", "MyHeritage", "FindAGrave-Plus"]`. `subscription`
is a **closed** enum with no `FamilySearch` value, and `validator.ts` rejects
anything outside it — that write fails validation. The session log ends before
`research_append`, so the failure is not *observed*; the mismatch is verifiable in
the repo regardless, and it was three-way:

- `init-project`'s access question invites three kinds of answer — a paid
  subscription, a FamilySearch partnership, a library or family history centre
- the enum could hold only the first; the other two collapsed to `other`
- `search-external-sites` prose claims to read all three, then offers a lookup
  table containing only the five paid sites
- and `README.md` already promised *"Free access counts: a FamilySearch affiliate
  library or a partner subscription is access, not 'none'"* — a promise the schema
  could not keep

Every user of this product has FamilySearch access, so the question reliably
elicited an answer the schema could not store.

## What changed

Two new values on the `subscription` enum, named from README's own vocabulary:

- `FamilySearch-Partner` — a partner subscription granted through a FamilySearch account
- `LibraryAccess` — a public library, family history centre, or FamilySearch affiliate library

Both are access **routes**, not sites: neither names which sites it unlocks, so
`search-external-sites` treats them as "ask before assuming a paywall blocks
this" rather than as a match for any named site, and neither counts as `none`.

A **plain FamilySearch account is deliberately not a value.** It is the product's
baseline — true of every user, so it discriminates nothing — and `init-project`
now drops a bare `FamilySearch` mention explicitly rather than storing it
verbatim, which is the exact write that would have failed. Someone with only that
has `["none"]`. Flagging this as the one judgment call in the PR: adding
`FamilySearch` as a third value would also have fixed the validation failure, at
the cost of an always-true value.

## Sites touched

CLAUDE.md's closed-enum list (both `enums.schema.json` trees, `CLOSED_ENUMS`, the
spec's prose tables) plus what its repo-wide-grep clause turns up — the blast
radius is wider than the documented list:

| File | What |
|---|---|
| `docs/specs/schemas/enums.schema.json` | the `$def` |
| `packages/schema/schemas/enums.schema.json` | byte-identical mirror (asserted by the lint) |
| `packages/engine/mcp-server/src/validation/validator.ts` | `CLOSED_ENUMS.subscription` |
| `docs/specs/research-schema-spec.md` | both prose tables (§ enum list, § 5.1.1) |
| `packages/engine/plugin/skills/init-project/SKILL.md` | canonical list, alias table, the never-store rule |
| `packages/engine/plugin/skills/search-external-sites/SKILL.md` | the lookup table that had no row for either |
| `apps/server/app/agent/mock_agent.py` | matches route *phrasing*, since nobody types the enum value |
| `eval/app/components/scenario/lib/schema.ts` | the CRUD UI's TS union |
| `README.md`, `docs/alpha-user-guide.md`, `docs/alpha-user-guide-cowork.md`, `docs/onboarding-slides.html` | user-facing copies of the list |

`packages/schema/src/enums.generated.ts` is generated (ADR-0008 tier 2, gitignored)
— regenerated via `scripts/gen-enums.mjs`, not hand-edited.

## Verification

- `enum-drift.test.ts` + `validator.test.ts` + `tool-schema-enums.test.ts` — 157 pass
- **Proved the lint can fail:** reverted the `validator.ts` copy alone and re-ran;
  it failed with `subscription drift in src/validation/validator.ts / missing
  (schema defines, validator rejects): FamilySearch-Partner, LibraryAccess`, then
  restored. A lint I have not watched fail is not evidence.
- Full MCP suite: 111 files, 2701 pass
- `pnpm test` (web, electron, viewer-ui, schema): 4/4 packages pass
- `apps/server` mock-agent subset: 79 pass
- New regression test `eval/tests/unit/init-project/access-routes-not-sites.json`,
  validated against `unit-test.schema.json`. It pins the **general** rule — an
  access route gets its own value, a baseline account is not stored — not Brian's
  specific answer.

## Drive-by: a Windows-only test failure — superseded on main, mine dropped

`write-serialization.test.ts` failed on Windows with `EPERM … symlink`, killing
the three path spellings that *do* work along with the one that can't. I fixed it
here by catching the permission error and dropping the symlink spelling on
Windows.

**That fix is not in this diff.** While this branch was open, main landed a better
one: `symlink(dir, link, "junction")`, a Windows junction — unprivileged, and
`realpath()` canonicalizes it to the same target, so all four spellings still run
there instead of three. I took main's version wholesale in the merge and dropped
mine. Recorded because the earlier commit message on this branch describes my
version, and the register should say what shipped.

## Eval runs

Both skills were re-run after merges landed underneath them, so the numbers below
are from the committed run logs, not the first pass:

- **`init-project`** — `v1_2026-08-27_14-03-34`: 13 pass, 1 partial, 0 fail of 14.
  Re-run because main landed three `init-project` commits mid-branch, one of them
  behavioural ("This is the complete list — do not flag anything else as an
  error"), which is not covered by `eval-cosmetic-skip`.
- **`search-external-sites`** — `v1_2026-08-28_06-22-27`: 13 pass, 2 fail of 15.
  Re-run because #1954 merged first and changed 14 files in this skill's snapshot.

Both are `stale=0` against the merged tree; both carry a complete `.ann.json`.

Neither failure comes from this branch:

- **`ut_search_external_sites_008`** — the intermittent triggering miss tracked in
  #1933. Every dimension scores 3; the run fails only because the skill never
  fired through the Skill tool, which `_compute_outcome` treats as an automatic
  fail for a positive test. Reproduced on a clean tree with this branch's edit
  stashed, and @EdmondOware showed it passing elsewhere — 7 activated / 3 not
  across all committed run logs.
- **`ut_search_external_sites_013`** — unstable across five runs
  (pass / partial / pass / fail / pass), and it passes on re-run **with** this
  branch's edit in place, so the edit cannot be the cause.

## Relationship to #1954

#1954 merged first, so this branch rebased onto it and paid the re-run — the
eval-slot rule working as designed. The merge was clean: that PR documents the
access routes in the URL-pattern table ("Paid subscription,
FamilySearch-partnership access, or a library/family-history-centre account"),
and this one makes those routes storable and tells the skill what to do with
them. Complementary, not duplicative.

## Unrelated issue found while verifying

#1959 — `apps/server` tests are unrunnable on Windows (seeded search turns 504
after 122s), and `_handle_stream` has no turn timeout where `_handle_sync` does,
so a silent agent holds a streaming connection open indefinitely. Confirmed
pre-existing by restoring this PR's `mock_agent.py` edit and re-running.

## A harness fix this branch needed

`runlog.py` called `os.replace` bare in two places. A completed 14-test
`init-project` run (~$2.50) died on that call with WinError 5 — a transient
sharing violation on the freshly-written temp file — and was demoted to a
gitignored `scratch_` log, so the whole run had to be paid for again. Both call
sites now retry with backoff and **re-raise** rather than falling back to a copy,
since a half-written run log read as complete is worse than a loud failure. Both
behaviours are pinned by tests in `test_partial_runlog.py`.

## Enables #1932

Dallan's ruling on #1932 has the hosted chips read their two lists from
`packages/schema/src/enums.generated.ts` — `ExperienceLevel` and `Subscription` —
"so a future enum change can't leave the UI stale". The `Subscription` union this
PR widens to 10 values is the one the site-access chips will render, which is why
the ruling refers to "10 values after the companion feedback fix widens it".
Nothing in the plugin changes for that ruling: `init-project`'s opening turn,
including the never-stop-and-wait rule, is untouched here.


EdmondOware (changes requested):
Peer review. This is unusually well plumbed. CLAUDE.md's closed-enum checklist is one of the easiest in the repo to half-complete, and this does all of it: both `enums.schema.json` trees (I diffed them, byte identical), `CLOSED_ENUMS` at `validator.ts:61`, both prose tables in `research-schema-spec.md`, the CRUD UI union, README and both alpha guides, with `enums.generated.ts` correctly left to `gen-enums.mjs` rather than hand edited. I verified the `$defs.subscription` enum loads as the same 10 values from both trees, which is also the count Dallan's #1932 ruling refers to. Proving the lint could fail before trusting it is the right instinct and not everyone does it.

The genealogy is right too. Access routes are not sites, which sites a partner or library route unlocks genuinely varies by institution and changes over time, and keeping a bare FamilySearch account out of the enum is correct: it is true of every user of this product, so it discriminates nothing.

You asked whether the two edits to `search-external-sites/SKILL.md` read well together. They do, and they complete each other rather than overlap. The "Check access" prose already promised three routes and predates both PRs, `4ab39a68` on 18 August. #1954 brought three of the five URL table rows into line with that promise. This PR makes the routes storable and adds the mapping row. That is the three way mismatch closing in sequence, and the routes paragraph is the right length for a billed prompt.

Two things.

**1. Newspapers.com now contradicts itself inside this one file.** Twenty five lines apart:

- line 116, added here: `any of the above | FamilySearch-Partner, LibraryAccess — may cover it`
- line 146, from #1954: `Newspapers.com | ... | Historical newspapers. Ancestry-owned. Paid subscription`

"Any of the above" includes Newspapers.com, so the file says both that a partner or library route may cover it and that it is paid only. That contradiction stands whichever way the underlying fact falls, so it needs resolving either way.

On which half to fix: I would extend line 146 to match the other three rows. Newspapers.com is one of the sites reachable through a FamilySearch partner account at a family history centre or affiliate library, and Newspapers.com Library Edition is a standard public library product, so a researcher with library access who reads that row is told to pay for something they may already have. Flagging that this last part is domain knowledge rather than something the repo states anywhere, so please sanity check it rather than taking my word.

FindAGrave is fine as it stands. It says "Free", so the blanket row is merely vacuous there, not wrong.

This one is as much mine as yours. I was in that table yesterday for #1954 and extended three rows without asking why the other two were different.

**2. This PR removes the last of the pre-fix run logs, and two figures on main depend on them.** Retention keeps five. Your branch adds two `search-external-sites` runs, so `v1_2026-08-20_22-45-06` and `v1_2026-08-24_09-59-31` both fall off. Combined with `v1_2026-08-20_21-55-13`, which #1954's own run pruned, that leaves **zero** pre-fix logs committed.

Two things in `docs/deep-dives/search-external-sites-findings-2026-08-26.md` cite them: F4's headline rate table, `5 of 15 = 33%` across all three, and F12's `judge_prompt_hash` table, which names two of them as the runs where the check fired. A sentence I added there yesterday, "over the two surviving pre-fix logs it is 3 of 9, still 33%", becomes false on merge.

Nothing here is your fault and nothing needs re-running. Please just add a provenance line to that findings doc naming where the logs can be recovered: `v1_2026-08-20_22-45-06` and `v1_2026-08-24_09-59-31` are at `25b0f434`, and `v1_2026-08-20_21-55-13` at `25b0f434^`. The findings doc is not snapshot tracked, so it costs nothing, and it keeps the figures re-derivable instead of stranding them.

**Not an ask, and neither PR's problem.** `windows-wrappers.test.ts` failed for me with seven offenders, every one of them a `.claude/worktrees/*/scripts/setup-feedback-case.bat`. The lint skips `node_modules`, `.venv` and `.git` but not `.claude/worktrees`, so it false fails for anyone using a git worktree. The tracked script is clean and reads `%~dp0` before `:parse` as it should. Worth its own card, since a lint that cries wolf locally is one people learn to ignore.

Everything else I would sign off as is.

benter-070:
Thanks — all three verified against the repo rather than taken on trust, and all
three hold.

## 1. Newspapers.com — fixed, and your domain call is right

Extended line 146 to match the other three paid rows. All four now read the same;
FindAGrave stays "Free", since as you say the blanket row is vacuous there rather
than wrong.

I checked the repo before agreeing and it is silent on this — nothing in
`references/` or `docs/` states which sites either route reaches, so it is domain
knowledge, exactly as you flagged. Corroborating from my side: Newspapers.com is
Ancestry-owned and part of the FamilySearch partner set available at family
history centres and affiliate libraries, and Newspapers.com Library Edition is a
standard public-library product. So the old row told a researcher with library
access to pay for something they may already have.

Worth stating what the row now claims, since it is weaker than it looks: it says
these routes *may* reach the site, not that they do. That is the same hedge the
routes paragraph carries ("varies by institution and changes"), which is what
makes it safe to assert without the repo having a source for it.

This costs a re-run — the file is snapshot-tracked. I did not reach for
`eval-cosmetic-skip`: changing whether the agent tells someone to pay is
behavioural, and that label is documented for behaviour-neutral edits only.

## 2. Provenance — added, and the sentence needed more than a pointer

Both citing passages now carry it: F4's row and F12's `judge_prompt_hash` table.
I verified all three commits resolve (`git cat-file -e` on each) before writing
them down.

One correction to your correction. Your note said the rate is "3 of 9 over the
two surviving pre-fix logs" — that was already going to be false, but not the way
either of us said: after this branch, **zero** survive, not two. So a pointer
alone would have left a sentence claiming a live re-derivation that no longer
exists. I voided that sentence explicitly and kept the three-log 33%, which is
what was actually measured when all three were present.

## 3. `windows-wrappers.test.ts` — confirmed, and worth the card

`windows-wrappers.test.ts:236` excludes exactly `node_modules`, `.venv` and
`.git`, so any `.claude/worktrees/` copy is scanned. It passed for me only
because that directory is absent from my checkout — I have been using
`git worktree add` to short paths outside the repo all day, which sidesteps it by
accident.

Agreed it deserves its own card rather than riding either PR. Happy to file it,
or leave it to you since you have the failing output.

## On the shared half of point 1

Noted, and it cuts both ways — I added a blanket row that contradicted a row I
had not read carefully enough, and the contradiction only became visible because
your PR had already brought three of the five into line. It is the kind of thing
that is invisible to whoever writes either half alone.

EdmondOware (approved):
Verified and approving. Re-derived at the current head rather than reading the summary.

**Newspapers.com.** All four paid rows now carry the same three routes and FindAGrave stays "Free", so the file no longer says two things about the same site twenty five lines apart. The access table and the URL table finally agree.

**Provenance.** Better than what I asked for. I asked for a pointer; you voided the stale sentence, gave a recovery table for all three logs with a runnable `git show`, and pointed F12 back at it. The addition I would not have thought to make is noting that F12's controlled pair is still committed, so the finding it actually rests on survives the pruning even though the top three rows do not. All three recovery paths resolve: `25b0f434` for the two, `25b0f434^` for `21-55-13`.

**The re-run.** 13 pass, 2 fail of 15, and both failures are yours to disown:

- `_002` is stochastic and pre-dates the branch. It failed on `v1_2026-08-27_12-50-50` with the identical signature, Correctness 1 and URL generation 1 on an encoded `Pennsylvania`, then passed on `06-22-27` encoding `Ireland`, then failed the same way again here. That is F4's residual, not your edit. The annotation confirms both 1s with reasoning that matches `c_001` exactly.
- `_008` is the #1933 triggering miss on both branch runs: `skills_invoked` is empty, so the skill never fired through the Skill tool and `_compute_outcome` fails a positive test regardless of scores. Every other dimension is 3.

One thing in `_008` that is not in your description and is worth knowing, though it is not the cause and I am not asking for anything: on this run it also picked up Tool Arguments 2. The first `research_log_append` passed `planItemId: 'q_001'`, a question id where a plan item id belongs, the tool rejected it, and the skill recovered with `null`. Final state correct. If that recurs it is worth a card, since question and plan item are a category confusion rather than a typo.

**Also checked at this head, since new commits landed since my review:** the full closed-enum site list is still complete, both `enums.schema.json` trees are byte identical, `CLOSED_ENUMS` carries both values, `enums.generated.ts` is untouched. Gate reports all runlog rules satisfied with the two shared fixture warnings now cleared. CI 12 pass. Engine packaging and validation suites 749 pass. Annotation covers all five sampled tests with no uncommented non pass, and `_002` is in the sample, so the one failure that matters got read.

Two factual notes.

The worktree lint is **already filed**, as #2006, open since 09:46 today, with the seven offender output and the `:236` line. Nothing needed from you. Your `pr1622` sighting is useful corroboration that it is not one person's checkout.

On the zero versus two point, which is minor and does not touch the fix: my review did say zero. The wording was "that leaves zero pre-fix logs committed", and the "two surviving" phrase was me quoting the doc sentence I was flagging as about to become false. Your fix is right either way, and voiding the sentence was the better call.

T-FEH (approved):
Reviewed on the merged head (`4a2688c2` — your `61ce90a8` plus a clean merge of main). Gate green; two warn-only judge-prompt notes, both inherited.

## The closed-enum edit list is complete

This is the change shape CLAUDE.md gives an explicit multi-site list for, and every site is hit:

```
docs/specs/schemas/enums.schema.json         +3/-1
packages/schema/schemas/enums.schema.json    +3/-1     (byte-identical to the first)
src/validation/validator.ts                  +2/-1     CLOSED_ENUMS carries both values
docs/specs/research-schema-spec.md           +2/-2
packages/schema/src/enums.generated.ts       NOT TOUCHED   ← correct; it is generated
```

Leaving the generated union alone is the part people get wrong, and you got it right. The enum reads cleanly:

```
Ancestry, MyHeritage, FindMyPast, Newspapers.com, GenealogyBank,
FindAGrave-Plus, FamilySearch-Partner, LibraryAccess, other, none
```

And the naming holds up against the existing convention — `FamilySearch-Partner` follows `FindAGrave-Plus`'s `Site-Tier` shape, and `LibraryAccess` has no tier to separate, so the missing hyphen is right rather than inconsistent. Worth saying explicitly because a closed enum is forever.

The two skills whose bodies changed each carry a fresh run, which is the correct handling, and the new `ut_init_project_012` **passes** — no red test shipped.

## `ut_search_external_sites_008` regressed, and the body does not say so

Against main's latest for that skill:

```
main     14 pass  1 fail
this PR  13 pass  2 fail

ut_search_external_sites_008   pass -> fail
```

`_002` fails on main too, so that one is inherited. `_008` is not. It is a positive test that ran with `skills_invoked: []` — the skill never activated — with Tool Arguments at 2 for passing `planItemId: 'q_001'` where the tool requires `pli_` or null, then recovering.

I checked whether your edit could have caused it and I do not think it did: both SKILL.md changes are body-only, and **no frontmatter line changed in either file**, so the triggering surface is untouched. Activation variance is the likely reading.

But it is the only regression in the shipped run, it is on a skill this PR edits, and the body does not mention it. That is the piece to add — a reader currently sees "two skills re-run" and no account of what moved. Say what it is and why you concluded it is not yours, the way you would want the next person to.

## The rest reads well

The defect account is unusually clear about its own limits — "the session log ends before `research_append`, so the failure is not *observed*; the mismatch is verifiable in the repo regardless" is exactly the right way to state a defect you inferred rather than watched. And declining to file a card for work already written, with the reason given, is the call CLAUDE.md's step 1 asks for.

The `search-external-sites` table edit is the substantive half: the row it adds says the two new routes "may cover" any of the sites above, and the paragraph explains they are routes rather than sites. That closes the gap the body describes — prose claiming to read three kinds of access over a table listing only paid ones — and the `Newspapers.com` row now matches its siblings.

I have not yet checked whether anything downstream switches on this enum — a UI badge map, an exhaustive TS union, a lookup keyed by the old five — which is where a closed-enum widening usually bites. That pass is still running and I will follow up if it turns anything up.

Approving on the schema work.

T-FEH:
The downstream-consumer sweep I promised in my approval is done. **It came back clean on the dangerous part** — and turned up two copy issues plus one correction to something I said.

## The sweep: nothing silently mishandles the new values

No `switch` on `subscription`, no `Record<Subscription, …>`, no badge/icon/colour map, no exhaustive TS union that fails to compile. Constructed `research.json` documents carrying each new value and ran the real validator:

```
FamilySearch-Partner / LibraryAccess / both mixed with Ancestry   accepted
bare "FamilySearch"                                               still rejected
libraryaccess / familysearch-partner / Library-Access             still rejected
legacy [] / ["none"] / ["Ancestry"] / ["MyHeritage"]              still validate
```

Both JSON Schema trees agree and are byte-identical. So the widening is genuinely additive, and the rejection of bare `FamilySearch` reproduces the exact failing write your premise rests on — the one the body honestly says it never observed.

## [Introduced] The mock agent now matches on words its own question never invites

`mock_agent.py:30` adds `_ACCESS_ROUTES` keyed on "library", "family history cent", "affiliate", "familysearch partner". The question at `:137` was left as it was:

> "Which **paid** record subscriptions do you have? (e.g. Ancestry, MyHeritage, Newspapers.com — or say "none")"

So the new matcher only fires if a user volunteers those words unprompted. That is this PR's own thesis inverted: before, the question invited three kinds of access and the schema could store one; now the schema stores three and the question invites one. One line, in a file you already touch.

```
You wrote:    "Which paid record subscriptions do you have? (e.g. Ancestry, "
              "MyHeritage, Newspapers.com — or say \"none\")"
Change it to: "What access do you have to genealogy sites? (e.g. Ancestry, "
              "MyHeritage, Newspapers.com; free access through a FamilySearch "
              "partnership; or a library or family history centre — or say \"none\")"
```

## [Inherited, now wrong] The one user-facing surface not updated

`viewer-ui/.../ProjectOverview.tsx:56` — and its hand-maintained fork in `eval/app` — render the empty state as `No paid subscriptions`, for a field this PR redefines as access routes. Values render verbatim through `subs.join(', ')`, so nothing is dropped; a user just sees `FamilySearch-Partner, LibraryAccess` raw. You updated four other user-facing docs, so this looks like the one that was missed rather than a decision.

## Correcting myself on the naming

I wrote that the naming "holds up against the existing convention" and called it worth saying explicitly because a closed enum is forever. That check was **within** the `subscription` enum only, and there is a counter-precedent I should have looked at:

```
subscription    Ancestry, MyHeritage, … FamilySearch-Partner, LibraryAccess
external_site   ancestry, myheritage, findmypast, familysearch_web, findagrave, newspapers
```

`subscription`'s PascalCase is justified in the schema itself as "site brand names" — a justification these two values explicitly do not meet, since the PR's own argument is that they are routes, not sites. Every non-brand concept in the other closed enums is snake_case, and `familysearch_web` is already the precedent for a FamilySearch-prefixed compound. `library_access` would be the repo default.

Still not blocking, and two approvals stand — but my "holds up" was more confident than my check justified, and this is the last cheap moment to change a closed enum.

## One card, not for this PR

Six hand-maintained prose copies of this enum exist (`init-project`'s list and alias table, the `search-external-sites` table, `eval/app/lib/schema.ts`, two `research-schema-spec.md` rows, `mock_agent.py`). Dropping both new values from `init-project/SKILL.md`'s list leaves the engine suite at **2727/2727 passing** — `enum-drift.test.ts` discovers copies by an `∈` marker that none of the six carries. All six are correct here; nothing would have caught it if one were not. That is `nothing-checks`-shaped and inherited.

Issue #2009, opened 2026-08-28
eval mock's record_search staged results still carry gedcomx/collectionUrl/treeMatches/duplicate events that production strips


taiwo-stack:
**Touches:** eval/harness/harness/mock_mcp.py, packages/engine/mcp-server/src/tools/record-search.ts (reference only), eval/fixtures/mcp/*record-search*.json, every `record_search` unit test's expectations.

`record-search.ts:1100-1133` runs five transformations on every staged result once `out.staged` is set:

- deletes `gedcomx`
- deletes `collectionUrl` (derivable from `collectionId`)
- hoists `collectionTitle` into a response-level `collections` map, deleting the per-row copy
- drops `treeMatches` when it's an empty array
- de-duplicates exact-duplicate `events` entries

`eval/harness/harness/mock_mcp.py` mirrors none of this — every `record_search` fixture hands the agent all five fields exactly as canned, even when `staged` is set. This is the same class of bug #1826 just fixed for `fulltext_search`'s `textDocument` strip: a mock that hands the agent a field production removed is a way CI stays green while the thing being graded (tool-usage quality, triage behavior) is checked against a shape production never sends.

Scoped out of #1826 on purpose — that issue's own text calls this "a separate card, not this one," and #1826 only fixed the single-field `fulltext_search` case. This is broader: five transformations, not one field, and touches every `record_search` test's expectations rather than one skill's fixtures.

## Fix shape

Mirror `record-search.ts:1100-1133` in `mock_mcp.py`'s staged-response handling (scoped to `_name == "record_search" and response.get("staged")`, following the same pattern as `fulltext_search`'s strip block), then update whichever `record_search` unit tests currently assert on the now-removed fields.

## Labels

`developer` — mock/fixture code, no genealogist judgment involved.
`nothing-checks` — this is exactly a way CI stays green while the eval suite silently grades tool-usage against a response shape production never sends.

DallanQ:
Fixed on PR #2045, in the opposite direction from the one this body proposes.

Mirroring the five transformations in `mock_mcp.py` would create a fresh hand-maintained copy of exactly what drifted — the same shape that produced this bug and #1826. Instead both `if (out.staged)` blocks move out of `record-search.ts` and `fulltext-search.ts` into `packages/engine/mcp-server/src/utils/staged-compaction.ts`, and the mock calls the **compiled** functions, the same way it already calls the compiled `stageSearchResults`. There is no second copy left to drift.

[...]

PR #2014, opened 2026-08-28
Check every mock fixture's shape against its tool's real return type (#1891)

### LINKED ISSUE #1891: Fix the 3 remaining wiki_place_page fixtures, and add a generic check that a mock fixture's shape matches its tool's real return type
> **Reviewed 2026-08-26 before junior handoff.** Decision made 2026-08-27 and
> recorded below — this is startable. The three fixture edits are verified, but the issue's
> claim that they are "the remainder" is **refuted**: 24 fixtures violate, not 3.
> Original body under "Original issue".

## Ruling: fix all 24, ship the check with no baseline

Every violating fixture is corrected in this PR and the check ships unconditionally green
— chosen over a frozen-baseline list, which would have left 21 fixtures feeding impossible
shapes to 13 tests with nobody scheduled to burn the list down.

### The population is four wrong envelopes, not 24 independent rewrites

Measured 2026-08-27 by reading each tool's return type in source and walking every fixture
for that tool:

| tool | violating | what is wrong | correct shape comes from |
|---|---|---|---|
| `record_search` | 14 | **all 14 are nil-result fixtures** — 12 use `{"results": [], "total": 0}`, 2 use `{results, returned, totalMatches}` | the 40 sibling fixtures that already have it right; the tool builds the response literally as `{query, totalMatches, paginationCappedAt: 4999, returned, offset, hasMore, results}` |
| `place_population` | 4 | all four wrong **and they disagree with each other** — 3 use `{data, notes, place}`, 1 uses `{data_points, notes, standardPlace, years_requested}` **This one is a nested rewrite, not an envelope swap — see correction 3.** | `PopulationResponse` in `src/types/place-population.ts`: `{place, population?, indexed_records?}` |
| `wiki_read` | 3 | all three share one wrong shape — `sections` and `title` are invented keys | the tool returns exactly `{url, content}` (`src/tools/wiki-read.ts`) |
| `wiki_place_page` | 3 | the three the issue was filed against | the 6 sibling fixtures that already have it right |

### No live API capture is required — do not treat one as a blocker

An earlier draft of this decision said the nil envelopes for `record_search` and
`place_population` had to be captured from the live tools first. **That is wrong and was
the main thing making a fix-all look expensive.** All four return types are readable from
source, and two of the four already have correct siblings sitting in the same directory.
The hosted wiki and Pop Stats services running on one developer's tailnet therefore do not
gate this work.

**One caveat to carry into the PR.** `place_population` returns
`response.json() as Promise<PopulationResponse>` — an unchecked cast that has never been
validated against the live Pop Stats service. Fixing those 4 fixtures to match the
TypeScript type may encode a second wrong shape with more confidence than the first. If one
live response is cheap to obtain, confirm against it; if not, say in the PR body that the
4 `place_population` fixtures match the declared type and that the type itself is unverified.

### Also in this PR

- **Amend the unit-test spec's MCP-fixture section.** It currently *recommends*
  `{"results": [], "total": 0}` as the empty-result example — the exact shape 12 of these
  violators use — and the spec is source of truth on conflict, so the check cannot forbid it
  while the spec recommends it. Replace that example with `record_search`'s real nil
  envelope, keep error fixtures exempt (the real tool throws, so there is no return type to
  match), and record that a fix-all beat a frozen baseline.
- **Prove the check fails before committing it.** Break one fixture and watch it go red.
  Unknown top-level keys fail; properties the return type marks optional may be absent.
- **Say what the re-run decision was — decided, see correction 1.** This changes eval *inputs* for 38 tests across
  locality-guide (25), research-plan (7), search-records (5) and historical-context (1).
  Fixture staleness is warn-only, so **CI will stay green while four suites' inputs have
  changed** — that is the risk this option accepts. State in the PR body whether any of the
  four re-runs was purchased, and if none were, that the four run logs no longer reflect
  the fixtures that produced them.
- **25 of 29 tools are still unmeasured.** A green check today is not a clean tree. Say so,
  so the next tool measured is not read as a regression.

### Verified 2026-08-27 — three corrections to carry into the PR

Every count above was re-checked against the repo and holds: 24 violating fixtures in the
stated 14/4/3/3 split, 38 affected tests split locality-guide 25, research-plan 7,
search-records 5, historical-context 1, all four return types readable from source, and
`docs/specs/unit-test-spec.md` line 190 really does recommend `{"results": [], "total": 0}`.
All 14 `record_search` violators are nil-result fixtures. Three things the ruling above gets
wrong or leaves out.

**1. Purchase no re-run except historical-context, and fold that one into issue #1710.**
The decision the ruling leaves open is made: no. But the reason is not "staleness is warn-only",
and do not write that in the PR body. All four skills were run this week — locality-guide
2026-08-25, research-plan 2026-08-25, search-records 2026-08-25, historical-context 2026-08-24 —
and every violating fixture was last touched in June or July. Those four run logs currently DO
match their inputs; this PR is what makes them stale. That is new staleness on fresh logs, not
the pre-existing baseline issue #1094 settled warn-only against. It is still acceptable, because
the edits are shape-only and no test's expectations grade any content that changes — verified: no
`judge_context` cites a population figure, the `place_population` `notes` prose, or a `wiki_read`
`title`/`sections` value. Say that in the PR body. historical-context is the one exception: issue
#1710 already holds that skill's paid run, so if its single affected test is re-run, it happens
there and not here.

**2. Ship the shape check top-level-keys-only, and state in the PR body what it does not cover.**
The check's depth is a real decision this issue never names, and it matters more than the
programmatic-vs-hand-maintained one. A top-level-key check catches all 20 `wiki_read`,
`wiki_place_page` and `record_search` violators outright — those are wrong or missing key names.
It cannot catch `place_population`, whose defect is nested: `place` holds a plain string where
`PlaceInfo` wants `{place_id, name, level}`. A fixture trimmed to
`{"place": "Muhlenberg, Kentucky, United States"}` passes the new check while still being a shape
the tool can never return. Ship shallow anyway — recursive type checking is a much larger job for
four fixtures — but record the limit in both the PR body and the spec amendment, so a green check
is not read as a validated tree.

**3. `place_population` is not an envelope rename. The whole fix is below — no live Pop Stats
response is needed for any of it.**
For the other three tools the "four wrong envelopes" framing holds. For this one it does not:

- `place` goes from a string to `{place_id, name, level}`. Invent a plausible `place_id`; the
  check reads top-level keys only, and nothing grades the value.
- the flat `data: [{year, population}]` array re-nests under
  `population: {"<source key>": {data: [{year, population, data_type}]}}`.
- `notes` has no home in the declared type and is deleted. Its prose is the only substantive
  content in three of these fixtures; deleting it is safe, verified above.
- delete `eval/fixtures/mcp/place-population-thisted.json` rather than reshaping it — it is
  referenced by zero tests.

So the fixture work is 23 edits and 1 deletion, not 24 edits.

## Original issue

Found while fixing a narrower version of this same defect class for wiki_search and wikipedia_search in PR #1803 (historical-context deep dive, issue 1663). The real wiki_place_page tool (packages/engine/mcp-server/src/types/wikiPage.ts, WikiPlacePageResult) returns exactly four fields: standardPlace, placeName, content, url.

Every single wiki_place_page mock fixture in the repo -- all 9 -- was missing placeName entirely, and 6 of the 9 also used the wrong names for the other fields (place instead of standardPlace, and page_url or source_url instead of url). PR #1803 already fixed the 6 that don't touch locality-guide (used only by research-plan or by nothing): the two Kentucky ones, the two Pennsylvania ones, and the two Denmark ones.

The three left, and still broken, are the ones locality-guide actually depends on:
- wiki-place-page-any.json -- the generic catch-all, referenced by 25 of locality-guide's own tests (essentially the whole suite): different-jurisdiction-ireland, massachusetts-pre-1841-town-registration, negative-search-wikipedia, persist-localities-all-sections, schuylkill-county-records, and ut_locality_guide_004 through ut_locality_guide_023.
- wiki-place-page-ireland.json -- referenced by different-jurisdiction-ireland.json.
- wiki-place-page-via-server-signature.json -- referenced by ut_locality_guide_023.json.

Good news: these three only need placeName added -- their standardPlace and url field names are already correct, unlike the six already fixed. The content itself doesn't need to change, only the shape:

before (wiki-place-page-any.json):
{ "standardPlace": "", "section": "home", "title": "Genealogy Research Guide", "content": "...", "url": "..." }

after:
{ "standardPlace": "", "placeName": "", "content": "...", "url": "..." }
(drop "section" and "title" -- neither is part of the real response type; placeName should be the leaf place name that was actually searched, mirroring what the real tool derives from its candidate-name matching)

Not filed as part of PR #1803 because fixing it invalidates locality-guide's own run-log snapshot -- essentially its entire suite, a full skill re-run outside that PR's scope (task-lifecycle.md's "too big for this PR, and you have opened the call sites and counted" test).

Validator request:
Rule: every eval/fixtures/mcp/wiki-place-page-*.json fixture's response must have exactly these four string fields: standardPlace, placeName, content, url.
Where to look: the fixture files themselves, checked against packages/engine/mcp-server/src/types/wikiPage.ts's WikiPlacePageResult.
Why it is not judgment: a fixed, four-field TypeScript interface -- no interpretation needed.
What a violation looks like: all three fixtures named above, right now.

## Merged in: issue #1894 — the generic fixture-shape check (2026-08-25)

Issue #1894 asked for the repo-wide version of this defect class and is folded in here,
because the two cannot be separated: **the generic check cannot land green until the three
fixtures above are fixed.** They are live violations today, so they are also the proof case
CLAUDE.md's "a new lint must be proven to fail" rule requires — break nothing, run the new
check, watch it fail on these three, then fix them and watch it pass.

Verified 2026-08-25: `WikiPlacePageResult extends WikiPageResult` in
`packages/engine/mcp-server/src/types/wikiPage.ts` is exactly `{url, content, standardPlace,
placeName}`, and all three fixtures still hold `{standardPlace, section, title, content, url}`
— `placeName` absent, `section`/`title` stray.

### The generic request (from #1894)

Rule: a mock fixture's response must match the shape its tool actually returns.

Where to look: each `eval/fixtures/mcp/<tool>-*.json` file's `response`, checked against that
tool's real return type in `packages/engine/mcp-server/src/types/` (or the tool's own return
statement where there is no separate type file).

Why it is not judgment: each tool has a fixed, already-written TypeScript return type.

**The open developer call, which issue #1894 deliberately left on the issue** (the check's *depth* is the bigger call — see correction 2)**:** whether to read those
types programmatically or hand-maintain a per-tool field list. #1894's position was that this
choice is the developer's to make here, and is not a reason to leave the request unfiled.

### Why it matters beyond these three fixtures

The same defect class surfaced three separate times in one PR (#1803): 10 `wikipedia_search`
fixtures, 4 `wiki_search` fixtures, and all 9 `wiki_place_page` fixtures each had a response
shape the real MCP tool can never produce — wrong field names, missing fields, or an entirely
different structure (a multi-result array where the real tool returns exactly one object). All
20 of those were hand-fixed in #1803; the 3 here are the remainder.

Hand-fixing closes today's instances only. Nothing checks this generically across
`eval/fixtures/mcp/` for any of the ~30 MCP tools, so the next fixture authored for any tool
can silently drift again **and the judge will never notice** — it grades a response against the
fixture's content, never against whether the fixture's shape matches the real API.

Violation catalogue from #1894: `wikipedia_search` fixtures shaped as `{query, results: [...]}`
instead of `{title, extract, url}`; `wiki_search` fixtures using `page_url` instead of
`source_url` and missing `section_heading`; `wiki_place_page` fixtures missing `placeName`
entirely.

Touches: eval/fixtures/mcp/ — 24 fixtures across record_search (14), place_population (4), wiki_read (3) and wiki_place_page (3), plus all tools for the generic check; eval/harness/scripts/ (the new check, likely alongside check_tool_coverage.py or check_rubric_tool_drift.py); docs/specs/unit-test-spec.md (the MCP-fixture section's empty-result example)

clack391:
Closes #1891. (#1894 asked for the repo-wide version of the same defect class and was folded into #1891, which is why it is already closed as not-planned rather than closed by this PR.)

Two other issues are referenced below for real interactions, and this PR settles neither of them: #1429, because PR #1977 edits one of the same fixtures on the same line; and #2009, the nearest sibling card for the value-level gap this check deliberately does not cover.

## Summary

- **Normal.** Adds the generic check issue #1891 asks for: a mock fixture's `response` must be a shape its tool can actually return. It derives both halves from source, so a new tool and a renamed field are covered without editing it.
- The check failed on **70 fixtures across 11 tools**, not the 3 the issue was filed for. 68 are fixed and 2 deleted, with no baseline list, per the ruling in the issue's comment thread. Two further fixtures are corrected for a value-level defect the check does not cover, so **72 fixtures are touched in all**.
- Also here: the unit-test spec recommended three error-fixture shapes and all three are unproducible, so the check cannot forbid what the spec recommends. Deliberately kept OUT, to hold the PR to the issue: a shared dispatch-arm helper, a query-legality check, and narrowing a now-duplicated harness assertion. Each is named below with what dropping it costs.

**This branch is two commits.** The first is the change. The second, `43738ba2`, is what a review of it produced: three silent-skip paths closed in the check's escape logic, and four claims corrected. Its message lists them, and the sections below are written against the final state, not the first commit.

## Review guidance

**Start here:** `packages/engine/mcp-server/tests/packaging/mcp-fixture-shape.test.ts`, and specifically the derivation chain and the three meta-assertions at the bottom. The chain is dispatch arm to awaited handler to annotated `Promise<T>` to top-level members, following interface `extends`, type aliases, inline type literals, unions and intersections. What I want checked is whether it can ever *skip* rather than fail: the third meta-assertion exists because a tool whose return type the reader cannot follow would otherwise have all its fixtures silently unchecked, and the check would read as coverage it does not have. The row "handler's `Promise<T>` annotation removed" in the table below is the one that proves that path.

Second, the fixture edits that change graded input rather than redundancy. They are named under "Four edits remove content the agent cannot get back", and all four are where a reviewer's judgement is genuinely needed.

**Determination on the one open question, since a reviewer asked for it rather than a deferral.** `place-population-any.json` cannot be made both type-legal and producible, and that is a property of the tool, not of the fixture. A catch-all cannot name a real place, and the only no-data answer the live service actually sends is HTTP 200 with `{error, place_id}`: two keys, which this check rejects as not the one-key thrown-error envelope. So every available option is wrong in one direction: the current shape is type-legal and unproducible, and the measured shape is producible and rejected.

That resolves the moment the tool stops passing an upstream error body through: `place_population` would then fail like all 48 other arms, with `{error: <message>}`, the catch-all would become that, and both problems close together. I have NOT made that change here, because it contradicts the spec's own error row ("Return the API's response as-is") and so needs the spec changed with it, which is a tool-contract decision and not a fixture-shape one. What I have done is stop the fixture's description asserting the unmeasured reading, and record the measured one. If the lead rules "throw", it is a small follow-up and the catch-all becomes honest.

**Unsure about:** the above is the one thing I could not close myself. `place_population` returns HTTP 200 with `{"error": ..., "place_id": ...}` for a place the Pop Stats corpus does not hold, and the tool's unchecked cast passes that through as a `PopulationResponse`, which cannot express it. The spec's own error row says the API returns "place info with no population sections" for this case; it does not. I have not changed the tool: the fix is a contract choice (throw on an error body, or declare the error shape), and the spec sanctions today's behaviour as "return the API's response as-is". Live evidence is in the section below. If the ruling is "throw", it is a small change and I will do it here.

## Plan

**Didn't change:** the value-level class this check cannot see, measured rather than waved at: 5 of 9 `wiki_place_page` fixtures carry a url the tool cannot build (4 of them among the six PR #1803 corrected), 3 `same_person` values contradict `match-engine.ts`, 3 `place_population` fixtures hold county data the Pop Stats corpus has no rows for, and 8 of 220 fixtures have an `args` predicate that does not match their own echoed `query` (6 real; 2 are legitimate place resolution). An issue is drafted for it. The one piece I did ship is the `query`-legality sweep above, because this PR repaired ten instances of it.

Also unchanged: the 27 `record_search` fixtures that carry `paginationCappedAt: 100`. The tool only ever emits 4999 (`PAGINATION_CAP`), so those are wrong, but they are wrong in the *value*, and this check compares top-level key names only. Fixing them is a different job with a different depth, and doing it here would have hidden the shape fix in another 27 files. It is the sharpest available illustration of what a green run does not mean, so it is recorded in the spec amendment rather than filed.

Also unchanged: `image_read`'s fixture. Its arm is the only one that serializes part of its result (`metadata`) rather than the whole thing, so "the shape its tool returns" has two readings there. The check resolves the handler's return type, which is what the harness actually serves the model, and the divergence from the production wire is named in the test header.

**Acceptance check:** `cd packages/engine/mcp-server && npx vitest run tests/packaging/mcp-fixture-shape.test.ts`. On `main`'s fixtures it fails with 69 shape violations and 1 envelope violation; on this branch it passes. The three `wiki_place_page` fixtures the issue was filed against are in that failure list, so they are the proof case CLAUDE.md's "a new lint must be proven to fail" rule asks for.

**Deviated from the plan:** three things.

0. **The issue said "keep error fixtures exempt"; I modelled them instead.** Its "Also in this PR" section directed exempting error fixtures because "the real tool throws, so there is no return type to match". The check does not exempt them: it models `{error: <string>}` as a real shape every tool can return, which keeps the two existing error fixtures passing but now REJECTS the other two shapes the spec recommended. Same outcome for the corpus, different mechanism, and it is why the spec's error section was rewritten rather than left alone.

1. The plan (and the issue, and the ruling) said 24 violating fixtures. It is 70. The ruling's principle scales unchanged, and its stated reason for declining a baseline is stronger at 70 than at 24, so I fixed all of them; but the blast radius is 84 tests across 8 skills rather than 38 across 4, and that is the one number a reviewer should not take from the issue.
2. The check went into the engine vitest suite, not `eval/harness/scripts/` as the issue's `Touches:` line suggested. Reasons under "Where the check lives".

## The population, measured

Re-derived by resolving all 29 tools that have fixtures to their real return types and walking every fixture in the corpus. That count moves as main grows, so it is stated against a base: 438 at the current merge base, 436 here after the two deletions, which is the same denominator as the "8 of 220" figure below.

| tool | failing | what is wrong |
|---|---|---|
| `record_search` | 14 | nil envelopes: 12 `{results,total}`, 2 `{results,returned,totalMatches}` |
| `external_links_search` | 14 | 13 omit the always-emitted `returned`; `-any` also invents a flat `standardPlace` and omits `query` |
| `rank_search_matches` | 14 | omit `scoreLogError`, declared required and set at both return sites |
| `fulltext_search` | 13 | omit `query`, set at the tool's single success return |
| `place_population` | 4 | 3 `{data,notes,place}`, 1 `{data_points,notes,standardPlace,years_requested}` |
| `wiki_read` | 3 | `title` and `sections` invented; the tool returns exactly `{url,content}` |
| `wiki_place_page` | 3 | the three the issue was filed against |
| `place_search_all` | 2 | invent `query` and `totalResults`; the tool returns `{results}` |
| `collections_search` | 1 | `collections-search-chicago` is a bare `{collections}` |
| `image_search` | 1 | `image-search-by-group-number` uses `input`/`output`, and its `output` is a shape `image_search` cannot return |
| `same_person` | 1 | `same-person-flynn-degenerate` invents a top-level `note` |

The `same_person` row is one the issue, the ruling and my own first measurement all missed; the check found it.

**Ten `query` echoes are hand-written rather than mechanically derived, and 2 of those were calls the tool refuses.** I first derived all 14 `record_search` nil `query` values by stripping `~` from the fixture's `args`, which is a case-insensitive substring *matcher*, not a call. Six of the 14 came out fine that way. The other eight did not: `validateInput` needs `surname`, `recordCountry` or `batchNumber`, so `{"givenName": "Patrick"}` and `{"fatherSurname": "Flynn"}` are calls no response can follow; five echoed a matcher fragment (`{"surname": "fl"}`) back to the model as its own query; and one echoed an empty `givenName` while its own description says the fixture covers a non-empty one. Those eight, plus the two pre-existing fixtures below, are the ten. Each is now a legal call consistent with its fixture's own description, re-derived independently.

**One honest limitation of those ten.** Six of the eight nil fixtures are deliberate catch-alls whose `args` predicate is a wildcard, so no single echo can be the caller's real query: the type requires `query`, and any concrete value tells the model it searched one specific thing. The structural fix is for the harness to echo the ACTUAL args for tools whose response echoes the query, which is a `mock_mcp.py` change and not in this PR. Until then a catch-all's echo names one representative call, and `record-search-surname-not-abbreviated-zero-results`'s description now says so explicitly.

I then stopped eyeballing and **executed the real `validateInput` against every `record_search` fixture's `query`**: 55 of 55 accepted on the merged tree. That run also caught 2 pre-existing fixtures I had not touched, both anchored only on `collectionId`, which is not one of the three anchors; each description named the surname the search actually carried, so both are fixed here too.

**That sweep IS shipped, as `mcp-fixture-queries.test.ts`.** It runs each tool's own exported input validator over its fixtures' echoed `query`, for the two tools that export one (`record_search`, `person_search`). It was withdrawn once to hold this PR to its issue and then restored, because leaving it out meant the ten `query` values this PR repairs were unguarded and the shape check stays green on every one of them: it compares key NAMES, and these are values. Two reviewers flagged that independently, and CLAUDE.md now says outright that a known limitation written into a PR body is deferred work minus the tracking.

Watched failing on six shapes, three of them `validateInput` rules no fixture in the corpus uses: a query with no anchor, `birthYearFrom` without its `To`, `batchNumber` paired with `recordCountry`, a negative `person_search` offset, a scalar `query`, and a renamed entry in its own tool map so a tool cannot fall out of the sweep silently.

**Three of the missing fields are honest, so 41 of those fixtures are wrong rather than the check being strict.** `external_links_search.returned`, `fulltext_search.query` and `rank_search_matches.scoreLogError` are each emitted on every success path, checked at every return site rather than inferred from the type. No TypeScript type needed correcting.

**Two fixtures are deleted rather than reshaped**, both referenced by zero tests and neither reshapeable without inventing its entire content: `place-population-thisted.json` and `image-search-by-group-number.json`.

**Two more fixtures are touched beyond the check's 70**, both pre-existing and both caught by executing `validateInput` rather than by the check (they are the last two of the ten above): `record-search-surname-not-abbreviated-zero-results.json` and `record-search-urna-variant-marriage-match.json`. Neither adds a skill to the staleness list.

**`{error: <string>}` is accepted for every tool, modelled rather than exempted.** All 48 dispatch arms catch identically and return exactly that; there is no other error shape in `src/index.ts`. So the two `{error}` fixtures in the corpus are correct as they stand, and `{error, message}` still fails.

## Where the check lives

The issue suggested `eval/harness/scripts/`, alongside `check_tool_coverage.py`. Those run from `check-runlogs.yml`, whose scope step deliberately excludes `packages/engine/mcp-server/src/**`. So the change that silently invalidates 40 fixtures at once, editing a response interface, would match no pattern there and the check would report success with every step skipped.

The harness pytest suite would **not** have been dark, and my first draft of this section got the reason wrong in a way worth stating: I wrote that neither Python job installs `node_modules`. It does. `eval-harness-tests.yml` runs `npm ci` and `npm run build` in this package, and its PATTERNS carry both `^eval/fixtures/` and `^packages/engine/mcp-server/`, so a Python check there could drive the TypeScript compiler.

What actually decides it is the shape that bridge has to take. `harness/ts_validator.py` is the existing example, and its documented contract is to return `None` so the caller **skips** when the build or `node` is absent. That is precisely the silent skip this check exists to avoid, sitting directly under the one assertion the whole thing rests on. Here the reader is native, so there is no bridge to degrade and no hand-maintained per-tool field list to drift. `engine-tests.yml` additionally runs ungated on every PR, so nothing can scope this check dark later.

## Four edits remove content the agent cannot get back, and here is the test I used

My first pass claimed two, and checked it the wrong way: whether a test's `judge_context` cites the deleted string. That test does not work. `wiki-read-via-server-signature.json`'s `sections` values ARE cited by `ut_locality_guide_022`'s `judge_context`, and deleting them costs nothing, because every one of them appears verbatim inside the `content` that stays. Citation cannot tell redundancy from substance.

The test that does work is per fixture: does the deleted value survive anywhere else in that fixture's own response? **The figure depends on the rule, so here is the rule.** 35 leaf values leave a `response` body across the edited fixtures. Under a strict verbatim-value rule, 11 do not reappear. Counting a value as surviving when this PR carried it, or its decomposed parts, into another field of the same fixture (`place_population`'s `place` string becomes `place.name` plus `place.parent.name`; the two county `notes` move into `description`) leaves **seven**, of which three are trivial and four are not. Under the loosest rule, where prose relocated to a `description` the model never sees still counts as kept, it is six.

Trivial, recoverable by construction: `section: "home"` on the three `wiki_place_page` fixtures (an echo of the input argument), and `totalResults: 0` on `place-search-all-any` (`results.length`). `place-search-all-cork` also loses `query: "County Cork, Ireland"`, a field the tool never returns.

The four that matter, which I would rather a reviewer weigh than find:

- **`place-population-any.json`** loses `notes: "No population data available for this fixture."` and becomes `{place: {...}}` only. All 25 locality-guide tests read it. The agent must now infer "no data" from an absent optional key, which is what production requires of it, but it is a change to the input of a graded section and not a rename.
- **`place-population-muhlenberg.json`** and **`place-population-schuylkill.json`** each lose a substantive `notes` paragraph of population context (1 and 7 referencing tests). `notes` is not a field `place_population` returns, so it cannot stay in the response; I moved the prose into each fixture's `description`, which `build_manifest` never serves to the agent, so the authoring intent survives even though the agent no longer sees it. Production never showed the agent that context either.
- **`same-person-flynn-degenerate.json`** loses `note`, which was the only in-band signal that its 0.005 score is degenerate rather than a real mismatch. `eval/tests/unit/person-evidence/degenerate-score-unresolvable-id-still-links.json` grades exactly that recognition, so the test gets strictly harder, in the direction production behaves. **Open PR #1977 rewords this same `note`**, a direct same-line conflict, confirmed with `git merge-tree` against its current head. A reviewer asked for this to be decided rather than noted, so: **the `note` key has to go from that fixture in either merge order, and the order itself does not matter.** `SamePersonResult` declares no `note` member and `same_person` never emits one, so the key is invalid independently of this PR; this check is only what makes it visible. If this PR lands first, #1977 drops the line when it rebases. If #1977 lands first, this PR re-deletes it. The only outcome to avoid is #1977 landing second while keeping the key, which reddens `main` on the new check. Issue #1429, which owns how a degenerate score should be signalled, stays open either way, and that is where a real replacement signal belongs. The underlying gap, that `same_person` gives the model no way to say a score is degenerate, is issue #1429's and this PR does not close it.

## `place_population`'s declared type is no longer unverified

The ruling asked for a live confirmation if one was cheap. It was. The hosted Pop Stats API answered, and `PopulationResponse` matches it exactly: `{place: {place_id, name, level, parent?}, population: {<source>: {source_url?, data: [{year, population, data_type}]}}, indexed_records: {<source>: {description?, data: [{period_start, period_end, records}]}}}`, with `data_type` observed as `estimate` and `census`. The reshaped fixtures use real FamilySearch primary place ids, which is the id space Pop Stats is keyed on: FamilySearch's primary id for "Arizona, United States" is 10 and so is Pop Stats'.

Two things I am reporting rather than fixing. Pop Stats holds no county rows (Schuylkill County 816 and Muhlenberg County 1554 both return "Place not found", and its own OpenAPI covers only countries, provinces and towns), so those three fixtures match the declared type and contradict the service. That is exactly the risk the issue's caveat named, now measured rather than hypothetical. And the HTTP 200 error body described under "Unsure about".

**That second one also makes the check knowingly over-strict in one place.** Because `place_population` passes the upstream body through, a real response can be `{"error": ..., "place_id": ...}`, and the one-key failure envelope rejects it. No fixture uses that shape, and a failure on it points at the tool's cast rather than at the fixture. I did not loosen the envelope to admit it, because that would re-admit `{error, message, status}`, which no arm can produce. Recorded in the check's header and in the spec.

## Eval inputs changed, and no re-run was purchased

This leaves **8 skills carrying a stale run-log snapshot** (`historical-context`, `locality-guide`, `person-evidence`, `research-plan`, `search-external-sites`, `search-full-text`, `search-images`, `search-records`), which is the number to judge the debt by. It changes eval *inputs* for **84 tests across those 8 skills**, re-derived on the tree merged with `origin/main`: locality-guide 25, search-records 20, research-plan 15, search-external-sites 12, search-full-text 9, search-images 1, historical-context 1, person-evidence 1.

Fixture edits route the staleness FINDING to `check_runlogs.py`'s `rule2_fixture_touched`, which only warns and is kept in a set disjoint from the blocking rules, so CI stays green while those inputs have changed. Precisely, because "reaches only a warn-only arm" would overstate it: rule 5 (the corpus-wide annotation parse, line 819) is a blocking rule that runs unconditionally on every invocation and did real work here across 109 annotations. Nothing in this diff can trip it, since no annotation is touched, but a blocking arm is reached. Run locally with the PR's own SHAs, the gate exits 0 and warns on exactly those 8 skills.

**Do not read that green as a sanction, because the settled policy is stricter than the code.** #1094's decision was "warn-only against a frozen baseline of the currently-stale skills; **block only on new staleness**". This PR produces new staleness on run logs that are currently active, which is the case that decision wanted blocked. It escapes only because the enforcement was never built: #1242, the follow-up that would have added the frozen-baseline anchor, is closed as **not planned**, so warn-only-for-everything is the de facto permanent state rather than an interim one. My first draft of this section attributed plain warn-only to the lead's decision, which misstates it.

**One qualification to "shape-only", found while reviewing this PR and worth a reviewer's attention.** The 14 `record_search` nil fixtures did not just gain field NAMES. Each previously carried `{results: [], total: 0}` and now carries six fields the model did not see before, including **`hasMore: false`** and `totalMatches: 0`. `hasMore` is evidence, not shape: it is an explicit "this query has no further pages" signal.

Four search-records tests that read these fixtures grade exhaustion reasoning, and `escalate-to-external-after-fs-exhaustion` passes only if the skill tries **3 or more** name variants before escalating. `hasMore` is about pagination rather than variants, so a careful skill should not conflate them, and the new `totalMatches: 0` / `returned: 0` actively support the nil-log assertions the same test makes. But the risk is not zero, and it lands on the one skill whose pass condition here is behavioural. **If one re-run is purchased, buy search-records, not locality-guide.** That is a change from what I first recommended, and the reason is this paragraph.

**That staleness is new, not the pre-existing #1094 baseline.** All 8 skills' latest run logs currently match their inputs; this PR is what makes them stale. It is still acceptable because the edits are overwhelmingly shape-only, but "no test's expectations grade any content that changes" is too strong as first written: the four exceptions above (three `place_population` `notes`, `same_person`'s `note`) plus the `hasMore` signal are content, and the search-records case is the one I would not defend as inert. `historical-context`'s single affected test belongs to issue #1710, which holds that skill's paid run, so if it is re-run it happens there and not here.

**The issue's "25 of 29 tools are still unmeasured" no longer holds, and that is the point of the check rather than a claim about the corpus being clean.** All 29 tools that have fixtures are now measured, on every PR, instead of the 4 that had been read by hand. But a green check is not a validated corpus: it compares top-level key names only, so it is blind to a nested defect (`same-person-flynn-degenerate`'s `candidateArk: null` against a `string` type passes) and to an implausible value (`paginationCappedAt: 100` passes). Recorded in the spec amendment so the next tool measured is not read as a regression.

## Test plan

- [x] `make test-all` run. It reports **one** failure on my machine and it is environmental, not this PR: `windows-wrappers.test.ts` walks `.claude/` and descends into 40 stale review worktrees, flagging 37 `.bat` copies that predate the #1905 fix. All 37 flagged paths are under `.claude/worktrees/`, which `.gitignore` covers, so they do not exist in CI; zero flagged paths sit outside it, and none is a file this PR touches. Independently confirmed in a clean worktree: the engine suite is 2732 passed / 0 failed there.
- [ ] **Per-skill eval runs: deliberately not run.** 8 skills' run-log snapshots changed through shared fixtures. Reasons above; this box is left unticked rather than ticked falsely.
- [x] No skill `description` or DO NOT clause changed, so no routing pair needed a reciprocal negative test.

**The check was watched failing on 15 shapes I did not write**, plus four positive controls proving it follows the TypeScript type rather than a frozen list and does not fire on legitimate refactors:

| break-test | expected | got |
|---|---|---|
| invented top-level key on a `wiki_read` fixture | red | red |
| `wiki_place_page` minus the required `placeName` | red | red |
| `{error, message}` (the shape the spec used to recommend) | red | red |
| `{error: 503}`, a non-string error | red | red |
| fixture naming an undispatched tool with no `input_schema` | red | red |
| `response` as an array | red | red |
| empty `args` | red | red |
| `output` instead of `response` | red | red |
| a whole dispatch arm block-commented out | red | red |
| handler's `Promise<T>` annotation removed | red | red |
| **dispatch arm serializes an inline `{...result, extra}`** | red | red |
| **the same projection bound to a const first** | red | red |
| **and via a chain of consts, or `Object.assign`** | red | red |
| a plain alias (`const p = result`), which is NOT a projection | green | green |
| **bad fixture hidden in `eval/fixtures/mcp/zzsub/`** | red | red |
| **handler annotated `Promise<Record<string, unknown>>`** | red | red |
| **response type name declared twice under `src/`** | red | red |
| aspirational fixture with `input_schema` | green | green |
| the same fixture without `input_schema` | red | red |
| fixture using a newly declared **optional** field, and again with it absent | green | green |
| **optional field factored into a base both union arms extend** | green | green |

**A third round closed three more escape paths, all found by blind review rather than by me.** Each was a way the guard reported coverage it did not have, and each is now break-tested:

| hole | how it stayed green | fix |
|---|---|---|
| the aspirational hatch was armed on 19 LIVE fixtures (`fulltext_search` 15/15, `source_attachments` 2/2, `record_person_matches`, `record_record_matches` all declare `input_schema` on every fixture) | renaming a dispatch arm made them read as aspirational: 15 fixtures, 13 repaired by this PR, dropped out with all 6 tests green | gated on `allToolSchemas`: an advertised tool can never be aspirational. **This closed only half of it, see the round below** |
| duplicate FUNCTION names were unguarded while duplicate TYPE names were, and there are 18 collisions under `src/` against 6 | planting a second handler name in another file let a fixture restore the exact keys this PR deletes, green | guarded, and the type-side guard also widened to catch cross-KIND (interface vs alias) collisions, which it had been missing |
| `constrainsNothing` required a type to have NO required members, but `open` alone makes the unknown-key filter skip | one index signature on `WikiPageResult` disarmed unknown-key detection for two tools | any top-level openness is now reported |

**A fifth round closed the projection detection properly.** The bold row above was true only while the object literal is the DIRECT argument to `JSON.stringify`. A reviewer bound it to a const first, changing nothing but `src/index.ts`, and the check went green while every real `wiki_read` response would then carry a field no fixture has. The mechanism was the same one twice fixed already: `serialized` held the const's name, `awaited.get(name)` missed, nothing set `projected`, and the single-awaited-call fallback picked the handler anyway.

They also flagged that the obvious fix (flag any serialized name absent from `awaited`) introduces a false positive, and measured it rather than asserting it: `const payload = result` is absent too and is not a projection. So the detection now follows plain-identifier initializers into `awaited` and keys the flag on names bound to something CONSTRUCTED. Break-tested on four projection forms (inline, one const, a const chain, `Object.assign`) and two alias forms that must stay green.

**A fourth round, from review, closed the other half of the hatch.** Gating on `allToolSchemas` stopped an *advertised* tool being read as aspirational, but `!advertisedTools.has(tool)` is equally true of a **renamed or misspelled** name, which is the likelier case. A reviewer proved it by editing nothing but a fixture's own `tool` field: `"fulltext_query"` plus `input_schema` plus a response no tool can return, and all six tests passed. Renaming the field on all 15 live `fulltext_search` fixtures dropped every one of them, including an invalid shape, out of the comparison silently.

The default is now inverted: an undispatched, unadvertised tool must be named in an `ASPIRATIONAL_TOOLS` set, **empty today**, or it is a failure with an actionable message. A check whose purpose is to not skip silently should not guess the charitable reading of an unrecognised name. Break-tested on the reviewer's exact cases: a typo'd name with an invalid response (red), a typo'd name with a plausible response (red), all 15 fixtures renamed (red), and a name on the allow-list (green).

An earlier version of this body listed "a typo in a fixture's tool name also passed" among the holes the second commit fixed. It did not; this round is what fixes it.

The four bold red rows and the last green row are the second round, and they are there because an adversarial pass found each of them green when it should have been red (or red when it should have been green). Each was a way this check could report coverage it did not have:

- **Arm projection.** The check resolves the *handler's* return type. An arm that serializes `{...result, retrievedAt}` adds a field every real response then carries, and all four meta-assertions stayed green with it absent from every fixture. It now detects that the arm builds its own object and fails, because following it properly needs a full type-checker and guessing is worse than saying so.
- **Subdirectories.** The loader resolves `fixtures_dir / f"{name}.json"`, so `mcp_fixtures: ["sub/hidden"]` loads. A flat `readdirSync` never saw it; proven loadable end to end before fixing.
- **An open return type.** `Promise<Record<string, unknown>>` produced a shape requiring nothing and admitting anything, so every fixture for that tool passed unexamined with all meta-assertions green.
- **Ambiguous type names.** The type index is keyed by bare name, last-write-wins, and `src/` already holds 6 duplicate interface names. None is a response type today, so the reader was right by luck.
- **A false positive, the opposite failure.** One `seen` set was threaded through every union arm and heritage clause, so a type referenced twice in *sibling* position looked circular. Factoring an optional field into a base that both `same_person` union arms extend, a purely additive and `tsc`-clean refactor, reported the whole tool unresolvable. Fixed by copying the set per branch, and the last green row is the regression guard.

Suite counts, all from runs on this branch **merged with `origin/main`**: ESLint and turbo typecheck clean; JS workspace 41 files; control plane 210 passed / 1 skipped; engine vitest 2731 passed / 1 failed (the environmental one above); eval app 168 tests; eval harness pytest 2887 passed / 2 skipped; and `check_runlogs.py` exit 0 with warn-only annotations naming the 8 skills.

I also merged `origin/main` and re-ran, rather than testing only against my own base. That mattered: main had gained `record-search-whitfield-1850-household.json` from PR #1946, a fixture this check had never seen. It passes, and `validateInput` accepts all 55 `record_search` queries on the merged tree.

## Follow-on issues

None filed.

**Folded in / filed instead:** nothing filed. Two things folded in, both because the check cannot ship without them: the 67 fixtures beyond the issue's 3, and the spec's two remaining unproducible error shapes (the spec is source of truth on conflict, so it cannot recommend what the check forbids).

Two things deliberately NOT done, each with its cost stated: a shared dispatch-arm helper (so the check carries its own walk, knowingly the third copy in that directory, disclosed in a KNOWN DUPLICATION block at the top of the file); and narrowing the harness's hand-written `person_read` key set (so it is now asserted twice, agreeing today, with both the spec and the check's header naming which copy to fix if `PersonReadResult` gains an optional field).

**A fourth, larger one worth a reviewer's opinion.** Two independent blind passes proposed the same alternative design: `ts.createProgram` with the real `TypeChecker` resolves symbols by scope instead of by bare name, which would delete both bare-name indexes, both collision guards, and most of `resolveType`, and would get `Partial<T>` / `Omit<T>` / imported types for free. The cost is a few seconds of program construction against the current ~1.6s. I did not take it: the AST version is now break-tested 24 ways and a late rewrite of the core deliverable trades proven coverage for elegance. If a reviewer would rather have the compiler-API version, it is better done as its own change than bolted on here.

Two things reported rather than filed, after searching the open pool. The value-level class the shape check cannot see is recorded in `unit-test-spec.md` §3.2 as class SHAPES WITHOUT instance counts, rather than as a new card. The counts were briefly written there and then removed: `docs/architecture.md` §9.4 is explicit that a gap register belongs on the `nothing-checks` label and not in a doc, "because a doc table cannot be kept honest", and two of those tallies were already wrong on the day they were written, which is that failure mode arriving immediately. Why no card: nothing in it is failing or graded, its largest instance is already live under #1429, and the closest existing card, #2009, is **now closed** by the merged PR #2045, which moved the search tools' post-staging compaction into `utils/staged-compaction.ts` and had the mock call the compiled functions. That fix and this one are complementary rather than overlapping: this check compares a fixture's top-level shape on disk against the return type, #2045 makes the mock serve the slimmed shape production actually sends at call time. Worth knowing for anyone extending this check: what the agent receives is no longer the fixture verbatim, so a future value-level check must read the mock's output, not the file. And the `place_population` HTTP 200 error body is a contract question for the lead, asked under "Unsure about" rather than filed.


T-FEH (changes requested):
Reviewed on the merged head (`075b3660` — your `43738ba2` plus a clean merge of main). Engine 2732 passed, harness 2927 passed / 2 skipped, `tsc` clean, gate green.

You asked a reviewer to test whether the check can silently skip. It can, on the one path you list as closed.

## The acceptance claim holds

Ran the check against `main`'s fixtures in a throwaway worktree and against this branch's:

```
main's fixtures   2 failed | 4 passed   ← exactly the two named assertions
                  × every fixture declares the envelope the harness requires
                  × every fixture's response is a shape its tool can return
this branch       6 passed
```

That is CLAUDE.md's proven-to-fail rule satisfied, on the real corpus rather than a synthetic case.

## [Blocking] The aspirational hatch still swallows a renamed or misspelled tool

`mcp-fixture-shape.test.ts:733` exempts a tool when it is not dispatched, not advertised, and every one of its fixtures declares `input_schema`. The `allToolSchemas` half only holds while the *old* name is still advertised — a rename changes both sites, and a typo never had either.

I proved it with a fixture that would otherwise fail, so the escape is unambiguous:

```
A) bogus response field, tool name correct     -> 1 test FAILS      (check works)
B) same bogus field, tool renamed to fulltext_query -> 6 passed      (silently skipped)
```

I did not even need to touch `src/`. Renaming the `tool` field on the fixtures alone was enough to drop all 15 `fulltext_search` fixtures — **including one carrying a demonstrably invalid response shape** — out of the comparison, with no failure and no warning.

Exposure is 19 fixtures across the four tools whose every fixture declares `input_schema`:

```
fulltext_search 15    source_attachments 2    record_person_matches 1    record_record_matches 1
```

Thirteen of those 15 are fixtures this PR just repaired. And the body lists "a typo in a fixture's tool name also passed" under the holes the second commit *fixed*.

This is the exact shape your own test header warns about — a check that reads as coverage it does not have — so it should not ship in the check that exists to prevent it. The fix is small: keep the hatch, but require the tool to be on an explicit allow-list, so an unrecognised name is a failure with an actionable message rather than a skip.

```
You wrote:    if (aspirational) continue;
Change it to: if (aspirational) {
                if (!ASPIRATIONAL_TOOLS.has(tool)) {
                  problems.push(
                    `${tool}: no dispatch arm, not advertised, and every fixture ` +
                      `declares input_schema, so all of its fixtures drop out of ` +
                      `the shape check. If the tool really is aspirational add it ` +
                      `to ASPIRATIONAL_TOOLS; if this is a renamed or misspelled ` +
                      `tool name, fix the fixtures`,
                  );
                }
                continue;
              }
```

with `const ASPIRATIONAL_TOOLS = new Set<string>();` beside `advertisedTools` — empty today, since every fixture declaring `input_schema` names a live tool.

Everything else thrown at the resolver fails loudly and correctly: an undeclared return type name, `Promise<Partial<T>>`, `Promise<any>`, `Promise<unknown>`, `Promise<Record<string, unknown>>`, `Promise<object>`, a `switch`-based dispatch, `obj.method()`, `return respond(await tool(args))`, and a fixture on an undispatched tool with no `input_schema`. Ten mutations, ten failures. The derivation chain itself is sound — it is only this hatch.

## The eval cost is bounded, and worth stating

72 fixture edits could have demanded a re-run wave. The gate reports:

```
All runlog rules satisfied — 8 warnings, 0 errors
```

Eight skills now carry stale run logs — `historical-context`, `locality-guide`, `person-evidence`, `research-plan`, `search-external-sites`, `search-full-text`, `search-images`, `search-records`. None blocks, because the fixture arm of rule 2 is warn-only under the #1094 ruling, which exists precisely so a fixture-only PR does not fire a ~$160–240 wave. So you are spending debt the ruling anticipated rather than dodging a gate — but the body measures blast radius in tests (84 across 8 skills) and not in stale snapshots, and eight is the number a future reader will care about.

## Minor

- **Merge order with #1977 needs deciding, not just noting.** It rewords the same `note` line in `same-person-flynn-degenerate.json` that this PR deletes, and `SamePersonResult` has no `note` member — so if #1977 lands second, `main` goes red on this new check. That is a decision for whoever merges.
- **Both review pointers miss.** "Review guidance" sends a reviewer to a section called "Two edits change what the corpus grades"; the section is "Four edits remove content the agent cannot get back", and it names four. And "Break-test 8" is the `output`-instead-of-`response` row — the one that proves the unresolvable-return-type path is row 10.
- `place-population-schuylkill.json` is missing a sentence separator its `muhlenberg` sibling has: `…1840–1910 period Population context…`.

## Checked rather than credited

Every measured claim I re-derived matched: the 35 leaf values leaving a response body (and that the strictly-lost set is the four substantive ones you name), the 84 tests across 8 skills including the per-skill split, both deletions genuinely unusable, the four substantive removals all being fields the tool cannot return, and the error-shape deviation leaving exactly two single-key `{error}` fixtures repo-wide with no `{error, message}` survivors.

On your `place_population` question for the lead: I would settle it here rather than defer. `place-population-any.json` is now type-legal but models a response the live service never sends, and all 25 locality-guide tests read it — so the check now certifies a shape that cannot occur, which is a smaller version of the same problem this PR exists to fix.

Requesting changes on the hatch. Everything else I would approve as it stands.

DallanQ:
Coordination note from PR #2045, not a review comment — two interactions, neither blocking.

**The #2009 forward-reference in your body is now stale.** You named it as "the nearest sibling card for the value-level gap this check deliberately does not cover." PR #2045 closes it: `record_search`'s and `fulltext_search`'s post-staging compaction moves into `utils/staged-compaction.ts` and the mock calls the compiled functions, so the mock now serves the slimmed shape production sends. Your check and that fix are complementary — yours is top-level shape against the return type, that one is the value-level strip — so nothing here needs to change except the sentence, and only if you are pushing anyway.

**No file conflict.** We both touch `docs/architecture.md`; your hunks are in the add-a-tool block and the add-a-unit-eval-test block, mine are in the `references/` duplication section. Different regions, clean merge either order.

One thing that may matter to your check: the mock now strips fields from a fixture's response at call time. Your check reads the fixture on disk, so it is unaffected — but if you ever extend it to what the agent actually receives, that is no longer the fixture verbatim.

clack391:
Thanks, this was the right catch and my fix was only half of one. All six items are addressed at `40960ca8b`.

## The blocking one

You are right that `allToolSchemas` only holds while the old name is still advertised. I reproduced your case before changing anything: one fixture with `"tool": "fulltext_query"`, an `input_schema`, and a response no tool can return, and all six tests passed. Your harsher version reproduced too, renaming the `tool` field on all 15 live `fulltext_search` fixtures and dropping every one of them, including the invalid shape, out of the comparison silently.

Took your prescription as written. The hatch now denies by default via `ASPIRATIONAL_TOOLS`, empty today, and an unrecognised undispatched name is a failure naming both possibilities rather than a skip.

Break-tested on your cases: typo'd name with an invalid response (red), typo'd name with a plausible response (red), all 15 renamed (red), a name on the allow-list (green). I re-ran all four after merging main, because a merge can quietly undo a fix.

One correction to my own body while I was in there: it listed "a typo in a fixture's tool name also passed" among the holes the *second* commit fixed. It did not. This round is what fixes it, and the body now says so.

## Your other five

- **Blast radius in stale snapshots.** Agreed that eight is the number a reader judges the debt by, not 84 tests. The eight skills are now named in the body.
- **Merge order with #1977, decided rather than noted.** The `note` key has to go in either order, and the order does not matter: `SamePersonResult` declares no `note` member and `same_person` never emits one, so the key is invalid independently of this PR and this check only makes it visible. If this lands first, #1977 drops the line on rebase; if #1977 lands first, this PR re-deletes it. The only outcome to avoid is #1977 landing second while keeping it. Issue #1429 stays open as where a real degeneracy signal belongs.
- **Both review pointers.** Fixed. The section is "Four edits remove content the agent cannot get back", and the row that proves the unresolvable-return-type path is "handler's `Promise<T>` annotation removed", not row 8.
- **`place-population-schuylkill` separator.** Fixed.
- **`place_population`, settled here rather than deferred.** The catch-all cannot be both type-legal and producible, and that is a property of the tool. A catch-all cannot name a real place, and the only no-data answer the service actually sends is HTTP 200 with `{error, place_id}`, two keys, which this check rejects as not the one-key thrown-error envelope. It resolves the moment the tool stops passing the upstream body through: `place_population` would then fail like the other 47 arms with `{error: <message>}` and the catch-all becomes that. I have not made that change here because it contradicts the spec's own error row, so the spec has to move with it, and that is a tool-contract call rather than a fixture-shape one. What I did do is stop the fixture's description asserting the unmeasured reading and record the measured one.

## One thing I added because of your review

You flagged the ten repaired `query` echoes as unguarded. I had built that sweep, withdrawn it for scope, and documented the gap instead. CLAUDE.md now says outright that a known limitation written into a PR body is the same debt minus the tracking, so I shipped it: `mcp-fixture-queries.test.ts` runs each tool's own exported input validator over its fixtures' echoed `query`. Break-tested on six shapes, three of them `validateInput` rules no fixture uses.

`make test-all` is EXIT=0 on the merged tree: 2753 engine, 2959 harness, 214 control plane, 168 eval app, 14 web, zero failures.

Re-requesting your review on the hatch specifically, since that is the part you blocked on.

clack391:
Thanks, both acted on at `40960ca8b`.

**The #2009 sentence is corrected.** I verified #2045 merged and closed it rather than taking the state on trust, and the body now says so, along with why the two are complementary: this check compares a fixture's top-level shape on disk against the handler's return type, yours makes the mock serve the slimmed shape production actually sends at call time.

**Your last point is the one worth keeping.** The body now records it explicitly for anyone extending this check: what the agent receives is no longer the fixture verbatim, so a future value-level check has to read the mock's output rather than the file. That is exactly the trap this check would have walked into.

**No conflict, confirmed mechanically.** `git merge-tree` against current main is clean; your hunks are in the `references/` duplication section and mine are in the add-a-tool and add-a-unit-eval-test blocks.

T-FEH (approved):
Second pass, on `0790de1e` — your `40960ca8` plus a merge of main I pushed to the branch (see the conflict note below). Engine **2766 passed / 117 files / 0 failed**; gate exit 0, `All runlog rules satisfied`.

## The blocking finding is closed, proven with the same experiment that broke it

Round 1 I renamed the `tool` field on all 15 live `fulltext_search` fixtures — one of them carrying a demonstrably invalid response — and got six green tests. Same edit, on this head:

```
+ "fulltext_query: no dispatch arm, not advertised, and every fixture declares
   input_schema, so all of its fixtures would drop out of the shape comparison.
   If the tool really is aspirational, add it to ASPIRATIONAL_TOOLS; if this is
   a renamed or misspelled tool name, fix the fixtures"
  Tests  1 failed | 5 passed (6)
```

Deny-by-default with an empty `ASPIRATIONAL_TOOLS`, and the comment explains why empty *is* the point. The `unknownTools` message was updated to match, which I had not asked for and which matters more than the hatch itself — it is the message someone reads at 2am.

## The merge collision I flagged happened, and the resolution is forced rather than chosen

#1977 landed first, so main carries its reworded `note` and this PR deletes it. I resolved in favour of the deletion and pushed. Before doing that I checked the resolution was not a preference:

```
with main's note   "same-person-flynn-degenerate.json (same_person):
                    fields the tool never returns: note"      1 failed | 5 passed
with it deleted                                               6 passed
```

Keeping main's line reds this PR's own check, so there was one legal resolution. Nothing is lost: the fixture's `description` already carries the same thin-subject doctrine the `note` did.

## The restored sweep is real coverage — all four of its branches fail on demand

I broke each one rather than reading the commit message:

```
illegal query (anchor stripped)  -> "search needs at least one anchor: surname, recordCountry or batchNumber"
scalar query                     -> "query is not an object"
renamed VALIDATORS key           -> "these tools have an input validator but no fixtures were found for them"
baseline                         -> 2 passed
```

The third is the one worth calling out: a sweep that silently matches nothing is the failure mode CLAUDE.md names, and this check guards *itself* against it. That is the lesson of the hatch applied to the new file rather than just patched in the old one.

I also tested the one thing the file asserts without proving — that its `query === undefined` skip is backstopped, so a fixture cannot escape by dropping the field:

```
delete response.query -> shape check: "fields the tool always returns, missing: query"   1 failed
```

The claim holds. Today the skip is dormant anyway: all 58 swept fixtures carry a `query`.

## Every count reconciles, including the two I first thought were wrong

Replaying the real `validateInput` over the base tree and this one:

```
base   44 swept   2 illegal   (both "no anchor", both pre-existing)
head   58 swept   0 illegal   (55 record_search + 3 person_search)
```

That matches "55 of 55" exactly, and the ten decomposes as the body says — base had 14 `record_search` fixtures with no `query` at all, 6 derived mechanically, 8 hand-written, plus the 2 pre-existing. I could not reproduce "ten" until I read your definition, which is a definition problem on my end, not yours.

Two figures looked wrong and were not: `place_population` is 4 at base and 3 here because one is deleted, and "8 of **220**" is exact — 220 of 436 fixtures carry a `response.query`. Both are more precisely scoped than I assumed.

## One stale sentence

Line 27 still reads "the one piece I did ship is the `query`-legality sweep above, because this PR repaired **12** instances of it." Lines 62 and 68 say ten, and `40960ca8`'s own message says it corrected twelve to ten — the file was fixed and this sentence was not.

## The staleness disclosure is the lead's call, not a defect

Your "do not read that green as a sanction" section is the right way to raise this, and I want to underline it rather than restate it. The gate is green on 8 warn-only skills — `historical-context`, `locality-guide`, `person-evidence`, `research-plan`, `search-external-sites`, `search-full-text`, `search-images`, `search-records` — and by #1094's actual decision this is the case that should block, escaping only because #1242 is closed as not planned. Worth a merge decision, alongside the `hasMore: false` point at line 118: 14 fixtures now show the model an explicit "no further pages" signal it could not see before, which is evidence rather than shape.

## The three other round-1 minors

All fixed: the review-guidance pointer now names "Four edits remove content the agent cannot get back" and matches the section; no `Break-test N` references remain; `place-population-schuylkill`'s sentence separator is in. And `place-population-any` was settled rather than deferred, with the limitation written into the fixture's own description and the condition that would resolve it named — that is the better outcome and it is now the artifact that carries the caveat, not the PR body.

One environmental note: the single engine failure you report is `windows-wrappers.test.ts` descending into stale worktrees under `.claude/`. I reproduced the cause — I had two of my own — and after pruning them the suite is 2766/2766 clean. Nothing to fix here.

Approving.

T-FEH (changes requested):
**Withdrawing the approval I posted twenty minutes ago.** A parallel reviewer found a hole I missed, I have reproduced it, and it is the same defect class as my round-1 blocking finding — one declaration form over.

## [Blocking] The duplicate-name guard skips the arrow-const branch

`mcp-fixture-shape.test.ts:151` records a collision for `isFunctionDeclaration`/`isMethodDeclaration`. The arrow/function-expression branch twenty lines below writes into the same bare-name, last-write-wins index and records nothing:

```ts
      // `export const fooTool = async (…): Promise<X> => …`
      returnTypes.set(node.name.text, { node: node.initializer.type, file });   // :168-169
```

So a `const` shadowing a handler silently redirects the comparison. I planted one file:

```ts
// src/zz-collide.ts
export interface ZzWrongShape { totallyBogus: string; }
export const wikiReadTool = async (): Promise<ZzWrongShape> => ({ totallyBogus: "x" });
```

rewrote all three `wiki_read` fixtures to `{"totallyBogus": "x"}` — a shape the real `wikiReadTool` (a function declaration at `src/tools/wiki-read.ts:21`) can never return — and got:

```
Test Files  2 passed (2)
      Tests  8 passed (8)
```

No failure, no duplicate warning. Those three fixtures are among the ones #1891 was filed against.

The one-line fix mirrors the branch above it:

```ts
if (returnTypes.has(node.name.text)) duplicateFunctionNames.add(node.name.text);
returnTypes.set(node.name.text, { node: node.initializer.type, file });
```

## The comment at :120 is right about the repo and wrong about the code

I instrumented the index with the check's own AST rules rather than trusting either of us:

```
duplicates the guard records (fn-decl branch only) : 16
duplicates that actually exist (both branches)     : 18   ← + add, mapPerson
```

So **18 is the correct count of reality** — the comment is not a stale figure, and a reviewer who "corrects" it to 16 is writing down the bug. What is wrong is that it sits on a set that only ever holds 16. The two missing names are live today, not hypothetical:

```
add        record-search.ts:283, person-search.ts:156, fulltext-search.ts:55   (three arrow-consts)
mapPerson  merge-gedcomx.ts:167, merge-gedcomx.ts:232                          (two arrow-consts)
```

That is the useful part: the same one-line fix makes the existing comment true of the code as well as the repo, and turns "right by luck" into a count the guard can actually defend. Neither name is a handler, so nothing is broken today — exactly the argument the third round already accepted for the branch it did fix.

## Correcting my own round-2 review

I wrote that the review-guidance pointer was fixed. Only half of it was. The section name now matches, but the sentence still reads:

> They are named under "Four edits remove content the agent cannot get back" and are **the only two** where a reviewer's judgement is genuinely needed.

I checked the section title and stopped there, which is the same shortcut — verifying the expected outcome instead of the mechanism — that I have flagged in other people's work in this repo. The count still contradicts the section it points at.

## One more stale figure

`docs/specs/unit-test-spec.md:209` — "for their **19** fixtures". `record_read` has 19 and `collection_read` has 1, so 20.

## Everything in my approval still stands

The round-1 blocking hatch fix, the forced merge resolution, the four proven-to-fail branches of the new sweep, and the reconciled counts are all as I reported. Engine 2766/2766 and the gate at exit 0 were on a clean tree and I reverted every probe; the tree is clean now.

Requesting changes on the arrow-const line. With that and the two text fixes I would approve immediately — this is one line in a file already open, not a rethink.

clack391:
Fixed at `bd67c7ea`. You are right, and the shape of my mistake is now a pattern worth naming: this is the **third** time in this PR I guarded one branch of a multi-branch site and called the class closed.

1. the hatch: fixed the advertised half, missed the renamed/misspelled half (your round 1)
2. duplicate type names: fixed same-kind, missed interface-vs-alias (a blind agent)
3. duplicate function names: fixed the declaration branch, missed the arrow-const branch (you, via a parallel reviewer)

Each time I tested the branch that motivated the fix and stopped. `returnTypes` has exactly two write sites and I guarded one.

## The blocking fix

Reproduced your exploit before touching anything: `src/zz-collide.ts` with `export const wikiReadTool = async (): Promise<ZzWrongShape>`, all three `wiki_read` fixtures rewritten to `{"totallyBogus": "x"}`, and 8/8 green. After the one-line fix that case fails, and so does the collision on its own with the fixtures left valid, which is the version that matters more: it catches the shadowing before a fixture is ever wrong.

## On the 18

I did not touch it, and I want to be explicit that your warning landed. I instrumented the index with the check's own AST rules rather than trusting either of us:

```
duplicates the old guard could record (fn-decl branch only) : 16
duplicates that actually exist (both branches)              : 18   <- + add, mapPerson
```

`add` is three arrow-consts (`record-search.ts:283`, `person-search.ts:156`, `fulltext-search.ts:55`) and `mapPerson` is two in `merge-gedcomx.ts`. So 18 was right about the repo and wrong about the code, and the fix makes the code match rather than the comment shrink. A reviewer who "corrected" it to 16 would have written down the bug, exactly as you said.

## The two text fixes

Both in. `unit-test-spec.md` now reads 20 with the split named (19 `record_read`, 1 `collection_read`) so it is checkable rather than asserted. The body's "repaired 12 instances" is now ten, and the pointer sentence no longer says "the only two" while naming a section that lists four.

That second one is worth a note back to you. Between your round 2 and round 3 I built a mechanical body checker for precisely this class, because four of your round-1 findings were stale pointers inside my own body that my repo-wide doc-consistency grep structurally cannot see. It catches unresolved section pointers, positional references, live cited-issue states, closing keywords, and em dashes. It passed clean on the body **and missed both of the errors you just found**, so I added a check for a sentence whose count contradicts the heading it points at, and verified it fires on the exact sentence you quoted.

I also tried and then deleted a second heuristic that flagged any phrase appearing with two different numbers. On the real body it produced two false positives and did not catch "12 instances" vs "Ten query echoes", because the two claims are phrased differently. A check that cries wolf and misses its target trains you to skip it, so that one needs reading and I have not pretended otherwise.

## State

`make test-all` EXIT=0 on your head plus this fix: engine **2766/2766**, harness 2992 passed / 2 skipped, control plane 214, gate exit 0. Your pruning note explains the discrepancy I had been reporting as environmental, and after this merge it is gone entirely rather than explained away.

Thanks also for resolving the #1977 collision on the branch and for checking that the resolution was forced rather than preferred. That is the check I had asked for in the body and had no way to run myself once it landed first.

T-FEH (changes requested):
Third pass, on `bd67c7ea` (contains `origin/main`, no merge needed). Engine **2766 passed / 0 failed**, `tsc` exit 0, harness **2992 passed / 2 skipped**, gate exit 0, CI green.

**The round-2 finding is genuinely fixed.** My exploit — a planted `export const wikiReadTool = async (): Promise<ZzWrongShape>` plus all three `wiki_read` fixtures rewritten to `{"totallyBogus": "x"}` — went from silent 8/8 to `1 failed | 5 passed`, and the collision *alone* with valid fixtures fails at `2 failed`. The check no longer needs a second mistake to notice the first.

**Keeping the `18` was the right call.** Re-derived with the check's own AST rules: 16 recordable by the old guard, 18 real, the two missed being `add` and `mapPerson`. The commit says the number "must not be 'corrected'" — correct, and a reviewer who had edited 18 to 16 would have written the bug into the comment.

All three text fixes landed, and `unit-test-spec.md:209` now names the split (19 `record_read`, 1 `collection_read`) so the number is checkable rather than asserted — better than what I asked for.

## [Blocking] Third instance of the same class, and it falsifies a bold row in your own table

Line 146 of the body:

```
| **dispatch arm serializes `{...result, extra}`** | red | red |
```

That holds only while the object is the direct argument to `JSON.stringify`. One `const` away it is green. I changed nothing but `src/index.ts`:

```ts
      const result = await wikiReadTool(args);
      const payload = { ...result, retrievedAt: "x" };
      return { content: [{ type: "text", text: JSON.stringify(payload) }] };
```

```
inline  {...result, retrievedAt}    -> 1 failed | 5 passed   "serializes a value it BUILDS…"
via an intermediate const           -> 8 passed (8)          tsc exit 0
                                       616 passed (616)      whole packaging dir
```

Every real `wiki_read` response would then carry `retrievedAt`, no fixture has it, and nothing in the repo goes red. The mechanism is the one you have now fixed twice: `serialized` holds `payload`, `awaited.get("payload")` is `undefined`, the loop falls through **without setting `projected`**, and the `distinct.size === 1` fallback picks `wikiReadTool` anyway. `projected` is only ever set where the argument is *not* an identifier, so binding it to a name is the whole exploit.

**The obvious fix introduces a false positive, so don't ship it blind.** Setting `projected = true` when a serialized name is absent from `awaited` also flags a plain alias. I measured that case rather than assuming it:

```
const payload = result;  JSON.stringify(payload)   -> 6 passed   (green today, and correctly so)
```

That is not a projection and should stay green. Following plain-identifier initializers into `awaited` first — `const a = b` inherits `b`'s awaited function — keeps the alias green while an object literal still misses the index and trips `projected`. Then the table's row is true as written, and worth a second row for the intermediate-const form.

## [Minor] One stale count survived the "fix two stale counts" commit

Body line 44: "walking all **435** fixtures on `main` (**433** remain here, after the two deletions)".

```
main cf0802d8   438
head bd67c7ea   436   (438 − image-search-by-group-number.json − place-population-thisted.json)
```

Stale rather than wrong — main gained 3 fixtures from #1977 — but the "8 of 220" line I reconciled last round uses 436 as its denominator, so the body now disagrees with itself.

## What I cleared

I swept the class rather than spot-checking, since this is its third appearance. All four writes into `interfaces`/`aliases`/`returnTypes` are guarded; `awaited` is a per-arm local where a collision is a compile error; `consultedNames` is keyed by unique tool name. The `:706` skip for a non-string `tool` or missing `response` is genuinely backstopped — I removed a `response` and set a `tool` to `123`, and the envelope assertion caught both. The duplicate-dispatch-arm index is unguarded here but `manifest.test.ts` fails on it, so the build still goes red.

Requesting changes on the projection line. Everything else in this round I would approve — and to be clear about provenance, the projection hole came from a second reviewer running in parallel; I reproduced every step of it before writing this.

clack391:
Fixed at `29d64f2e`. Fourth instance of the same mistake, and the tally is now the most useful thing in this thread:

| # | construct | form I guarded | form I missed | found by |
|---|---|---|---|---|
| 1 | aspirational hatch | advertised tool | renamed or misspelled name | you, round 1 |
| 2 | duplicate type names | interface vs interface | interface vs alias | blind agent |
| 3 | duplicate function names | declaration branch | arrow-const branch | you, round 2 |
| 4 | projection detection | inline argument | bound to a name first | you, round 3 |

Same error four times: I test the form that motivated the fix and stop. So this round I broke the class rather than the instance.

## The fix, and the false positive you warned about

Your exploit reproduced exactly: inline red, one const away green. Your warning about the obvious fix also reproduced, and it is the part that made this worth doing carefully. `const payload = result` is absent from `awaited` too, so flagging mere absence would have reddened a plain alias that is correctly green today.

So the detection now classifies every local binding in the arm and follows identifier initializers through to the name that holds the value. The flag keys on names bound to something CONSTRUCTED, not on absence from the index.

Break-tested on six forms, not one:

```
inline { ...result, x }                  -> RED
one intervening const                    -> RED   (your case)
a chain of consts (a = {...}; b = a)     -> RED
Object.assign({}, result, { x })         -> RED
const payload = result                   -> GREEN (alias, correctly)
const a = result; const b = a            -> GREEN (alias chain)
```

The chain cases are the ones your report did not name and I would have missed again had I stopped at your repro.

## The table row you falsified

You are right that it was a false claim, not just an imprecise one. It now reads "an inline `{...result, extra}`" with three further rows for the const, chain and `Object.assign` forms, plus a row recording that a plain alias stays green so the next reader knows that is deliberate rather than a gap.

## The count

Verified against the current merge base: 438 there, 436 here after the two deletions. My 435/433 was measured once and never re-derived as main grew, and you are right that it had come to disagree with the "8 of 220" denominator in the same body. It is now stated against a base rather than as a bare number, so it cannot drift silently again.

## One thing I built between your rounds, and what it did not catch

After your round 1 I wrote a mechanical body checker, because four of your findings were stale pointers inside my own body that my repo-wide doc-consistency grep structurally cannot see. It now checks section pointers, positional references, live cited-issue states, closing keywords, em dashes, and a sentence whose count contradicts the heading it points at.

It would not have caught this round's count. A number that is correct when written and goes stale as main grows is not detectable by reading the body alone, which is why stating it against a base was the actual fix rather than another check.

## State

`make test-all` EXIT=0 on `29d64f2e`: engine **2766/2766**, harness 2992 passed / 2 skipped, control plane 214, gate exit 0.

Thanks for sweeping the class this round rather than spot-checking, and for saying which findings came from the parallel reviewer. The four-write-site sweep and the `:706` backstop check are both things I had asserted and not proven.

T-FEH (approved):
Fourth pass, on `50e2085d`, plus three commits of my own — `501ea3a7`, `79635af6`, `01c80fe4`. All three are small, all measured, and all easy to revert; the middle one repairs damage I caused. Engine **2766 passed / 0 failed**, `tsc` exit 0, gate exit 0.

## Your projection fix is right, including the part I warned about

I ran the class rather than my one reported case:

```
inline {...result, retrievedAt}     RED     plain alias  const p = result       GREEN
via one const                       RED     alias chain  const a = result; b=a  GREEN
via a chain of consts               RED
Object.assign({}, result, {…})      RED
```

Keying on `built` rather than on absence from `awaited` is exactly what makes the last two green, and `resolveAlias` carries cycle protection it did not need. The table now has a row for each form, the alias row correctly marked green, and the fixture count is restated as 438/436 against a named base. Both of last round's items are closed.

## [Fixed in 501ea3a7] The fix cost one dispatch arm, silently

`&& wasAwaited` is new here and unexplained in the message, which is what drew my eye. It misfiles the one arm that calls its handler **synchronously** — `convert_calendar`, `const result = convertCalendar(args)` with no `await`. The binding falls past the handler branch into `built`, so the arm is reported as projecting and never resolves:

```
before 29d64f2e   arms=48   projecting=0   []
after             arms=47   projecting=1   [convert_calendar]
```

Green only because `convert_calendar` has no fixtures — a tool left the comparison with the suite passing, which is the failure this check exists to prevent. The first fixture written for it would have been met with "serializes a value it BUILDS" and never shape-checked. `built` is what distinguishes a projection; that condition should not also try to. With the clause gone, all four projection forms still fail, both alias forms still pass, and the index is back to 48/0.

## [Fixed in 01c80fe4] The root cause of rounds 3 and 4 — a fallback that fires for nothing

The classifier reads only variable **declarations** with an identifier name, so two spellings of the same projection were still green with `tsc` clean:

```ts
let payload;  payload = { ...result, retrievedAt: "x" };
const { ...payload } = { ...result, retrievedAt: "x" };
```

Both pass on your head. What let them pass is this, at the end of the resolver:

```ts
    if (!fn) {
      const distinct = new Set(awaited.values());
      if (distinct.size === 1) fn = [...distinct][0];
    }
```

It fires for **zero of the 48 arms**. Its comment calls it "the writer-tool shape", but writer arms name their variable inside `writerToolResult(result)`, which the callee branch already matches. Deleting it leaves the suite at 2766/2766 — and turns both spellings above red, while a plain alias and an alias chain stay green. Its only observable effect was to supply a plausible handler whenever the classifier missed, which is what made three rounds of misses look like passes. This is the fix worth having: it closes the class rather than the instance.

## [Fixed in 79635af6] A regression I introduced, which no check could see

My merge at `0790de1e` resolved this fixture's conflict with `--ours`, keeping the branch's whole file. The `note` deletion was right; taking the file wholesale also discarded the `description` #1977 had just landed, putting back the explanation that PR refuted:

```
mine (reinstated)  "a local stub whose id cannot resolve to a full FamilySearch ARK.
                    The number reflects the missing ARK"
#1977 (measured)   "the TREE CANDIDATE IS THIN … that is refuted
                    (probe: an ARK-less focus person scores 0.9999484)"
```

Reinstating a refuted belief is precisely what CLAUDE.md forbids, and nothing could have caught it — `description` is never served to the agent, so no eval, lint or CI job reads it. I verified only the `note` last round and not the rest of the file, which is the same narrow checking I have flagged in others' work here. The commit is main's version with only the `note` key removed, which is what the conflict actually called for.

- Line 153 still says `| aspirational fixture with input_schema | green | green |`. Round 1's deny-by-default inverted that — such a fixture now fails unless the tool is in `ASPIRATIONAL_TOOLS`, as your own prose says two sections later.
- Line 3 describes #1977 in the present tense; it merged on 2026-08-31.

## Approving

Five instances of one class across four rounds, each closed by measurement. The last three I fixed rather than sent back, since they were two words, four lines, and a file I had broken myself — read them and revert any you disagree with. The eval-staleness question and the `hasMore` evidence change remain the lead's call, and the body raises both honestly.

EdmondOware (approved):
Approved.

The silent skip question you raised is answered by the third meta-assertion, and answered well. Six distinct ways coverage could vanish are enumerated and each fails loudly: an aspirational tool, a projecting arm, an unresolvable return type, a type that admits any key, and both cases where the type or function index resolves against the wrong declaration. That is the right shape for a check whose failure mode is looking like coverage it does not have.

The lint is proven to fail. Against main's fixtures it reports 69 shape violations and 1 envelope violation and none here, so the three `wiki_place_page` fixtures the card was filed against really are in the failing set.

The four fixture edits you flagged for judgement are all correct, and the declared types are why. `PopulationResponse` is `{place, population?, indexed_records?}` with no `notes` member, and `SamePersonResult` has no `note`. Both fixtures were handing the agent fields production never sends, so removing them makes the graded input honest rather than merely smaller. Moving the population prose into `description` keeps the authoring intent while leaving it out of the agent's view, since `fixtures.py` reads only `response`.

The #1977 interaction landed in the order you called safe. It merged first and kept the key, this branch deletes it, and merging main in is conflict free with the key gone.

One thing left. `mcp-fixture-queries.test.ts` guards that every tool in `VALIDATORS` has fixtures, but not that every tool exporting `validateInput` is in `VALIDATORS`. It is complete today, `record_search` and `person_search` being the only two, so nothing is wrong now. But it is a hand maintained list in a PR whose other check derives both halves from source precisely so it cannot go stale, and a third validator would go unswept with nothing failing. The sibling file already walks source for exactly that.

PR #2016, opened 2026-08-28
OAuth hygiene: escape callback HTML, restrict token directory mode (#1018)

chrisedeson:
Two low-severity fixes from the security audit (Task 9 of #1018):

- **HTML-escape the OAuth callback page** (`login.ts:47-59`): `error` and `error_description` query parameters from the loopback callback URL were interpolated into the HTML response with no escaping. A crafted URL could render attacker markup in the `http://127.0.0.1:1837` origin. Impact is genuinely limited (the listener only exists during a login window, holds no cookies or session state), but it is a real injection sink in auth code. Added `escapeHtml()` covering `&`, `<`, `>`, `"`, `'` on all interpolated values (`title`, `heading`, `body`).

- **Restrict the token directory mode** (`tokenManager.ts:27`): the token file was correctly `0o600`, but the containing directory was created with `mkdir(STORAGE_DIR, { recursive: true })` and no `mode`, defaulting to `0o755` (world-traversable). Now `0o700`.

Both findings (D2/F2 and D5/F5) were confirmed still live against current `main` (`cd057d21`). Branch merges clean.

## Test plan

- [x] Existing auth tests pass (49/49)
- [x] New test: `HTML-escapes error parameters in the callback response` — simulates a callback with `<script>alert(1)</script>` in `error_description`, asserts the response contains `&lt;script&gt;` and not `<script>`
- [x] Mutation-verified: removing the `escapeHtml` calls makes the new test fail
- [x] Updated `tokenManager.test.ts` assertion to expect `mode: 0o700` on the `mkdir` call


chesworthrm (changes requested):
The escaping half of this is good. I mutation-tested it — removing `escapeHtml(body)` does fail your new test — and I checked the helper itself: `&` is replaced first so there's no double-escape bug, all five entities are covered, and a `</p><script>…` breakout attempt comes out fully neutralised. Nothing to change there.

The directory half doesn't take effect yet. Details below.

Also flagging for the lead: every file here is under `src/auth/`, which is a stop-rule path in `docs/task-lifecycle.md` step 4. Not a defect — just noting the sign-off is his.

---

## The `mode: 0o700` doesn't apply in the cases that matter

`mkdir`'s `mode` is only used when `mkdir` actually creates the directory. Two things follow.

**Anyone who has logged in before already has `~/.familysearch-mcp`.** For them the directory exists, `mkdir` does nothing, and the mode stays whatever it was. This is unconditional — it isn't a race or an edge case. I ran it against a pre-existing `0755` directory:

```
pre-existing dir: 755  →  after saveTokens: 755
```

**And `config.ts:78` creates the same directory with no mode:**

```
  await mkdir(STORAGE_DIR, { recursive: true });
```

That's `saveConfig`, which `configure_openrouter` calls and which needs no login first. So on a fresh machine, whichever runs first wins:

```
saveConfig first → 755, then saveTokens → still 755
saveTokens first → 700
```

`tokens.json` is still `0600` in every case, so this is defence-in-depth rather than an exposure on its own — but it's the thing the PR says it fixes, and right now it only fixes it for a fresh install where login happens first.

The reason CI didn't catch it: the test asserts the arguments passed to a mocked `mkdir`, not the mode the directory ends up with. It passes in both failing cases above.

## Suggested fix

Two files now create the same directory, so it's worth one helper rather than two copies.

In `src/auth/config.ts`:

```
You wrote:      import { mkdir, readFile, writeFile } from "node:fs/promises";
Change it to:   import { chmod, mkdir, readFile, writeFile } from "node:fs/promises";
```

Add above `saveConfig`:

```ts
// mkdir's `mode` is ignored when the directory already exists, so chmod too:
// every install predating this would otherwise keep its 0o755.
export async function ensureStorageDir(): Promise<void> {
  await mkdir(STORAGE_DIR, { recursive: true, mode: 0o700 });
  await chmod(STORAGE_DIR, 0o700);
}
```

Then in `saveConfig`:

```
You wrote:      await mkdir(STORAGE_DIR, { recursive: true });
Change it to:   await ensureStorageDir();
```

In `src/auth/tokenManager.ts`:

```
You wrote:      await mkdir(STORAGE_DIR, { recursive: true, mode: 0o700 });
Change it to:   await ensureStorageDir();
```

and drop `mkdir` from its `node:fs/promises` import, and swap `STORAGE_DIR` for `ensureStorageDir` in its `./config.js` import — nothing else in that file uses `STORAGE_DIR`.

**Two test files need edits too, or this won't build.** Both mock `node:fs/promises` with an explicit export list, so both need `chmod: vi.fn()` added alongside `mkdir: vi.fn()` — otherwise you get `No "chmod" export is defined on the "node:fs/promises" mock`. That's `tests/auth/config.test.ts` and `tests/auth/tokenManager.test.ts`.

Then in `tests/auth/config.test.ts`:

```
You wrote:      expect(mockedMkdir).toHaveBeenCalledWith(STORAGE_DIR, { recursive: true });
Change it to:   expect(mockedMkdir).toHaveBeenCalledWith(STORAGE_DIR, { recursive: true, mode: 0o700 });
```

Your existing `mode: 0o700` assertion in `tokenManager.test.ts` keeps passing as-is.

I applied all of the above locally to check it: `tsc` clean, full engine suite 2727/2727, and all three orderings then come out `0700` including the pre-existing-directory case.

Worth adding a test that checks the actual resulting mode against a temp `HOME`, since a mocked `mkdir` can't catch either of these.

One caveat I couldn't check: `chmod` on Windows only honours the read-only bit. The code already passes `mode: 0o600` to `writeFile`, so this isn't a new class of problem, but someone on Windows should confirm it's a clean no-op there.

## Minor — `color` is the one value left unescaped

```
    `h1{color:${color}}p{color:#555}</style></head><body>` +
```

Both callers pass hardcoded literals, so nothing is reachable today. But `htmlPage` is now a function that escapes what you hand it, and this one parameter quietly doesn't — and it lands in a `<style>` block, where HTML escaping wouldn't be the right defence anyway. Cheapest fix is to stop accepting arbitrary strings:

```
You wrote:      function htmlPage(title: string, heading: string, body: string, color: string): string {
Change it to:   function htmlPage(title: string, heading: string, body: string, color: "#cf222e" | "#1a7f37"): string {
```

Typechecks and leaves the login tests green — I checked. Non-blocking on its own, worth folding in while you're in the file.

---

Everything else looks right. All 11 checks are green on `a455fc92`, and I test-merged with `main` (you're 15 behind) and re-ran the suites there — engine, harness, control plane and the web workspace all pass, so the only thing standing between this and merge is the directory fix.

chrisedeson:
All three items done in `ba4c53ff`:

1.`ensureStorageDir()` in config.ts — `mkdir` + `chmod(0o700)`, shared by both `saveConfig` and `saveTokens`. Covers pre-existing directories and the saveConfig-first ordering.
2. `color` typed as `"#cf222e" | "#1a7f37"` — closes the unescaped style-block sink at the type level.
3. Both test files updated with `chmod` mock and assertions.

tsc clean, 41/41 auth tests pass.

chesworthrm (approved):
Approving. Everything I asked for last round is here and I verified it: I broke the escaping and broke the `chmod` line, and the right test failed each time. Both guards are real. The engine suite, the harness, the control plane and the web typecheck are all green on this head, and so are all eight required checks.

The `ensureStorageDir()` extraction came out exactly right, and adding the `chmod` assertion to both test files was more than I asked for. Thank you.

Two things to fold in before this merges. Neither is a code problem.

---

## 1. The spec still describes the old behaviour

`docs/specs/oauth-auth-spec.md` is the spec this module gets checked against, so it needs to say what the code now does.

```
oauth-auth-spec.md:101
You wrote:      - `saveConfig(patch: Partial<AppConfig>)` — merges `patch` into existing per-user config, `mkdir({ recursive: true })` + `writeFile` JSON with `mode: 0o600`. Preserves any keys the user has set that are not in `patch`.
Change it to:   - `saveConfig(patch: Partial<AppConfig>)` — merges `patch` into existing per-user config, `ensureStorageDir()` + `writeFile` JSON with `mode: 0o600`. Preserves any keys the user has set that are not in `patch`.
```

Then add this bullet directly after it:

```
- `ensureStorageDir()` — creates `~/.familysearch-mcp` with `mkdir({ recursive: true, mode: 0o700 })`, then `chmod(STORAGE_DIR, 0o700)`. The chmod is required: `mkdir`'s `mode` is ignored when the directory already exists, so every install predating this would otherwise keep its `0o755`. No effect on Windows, where chmod cannot change read or execute bits.
```

```
oauth-auth-spec.md:128
You wrote:      - `saveTokens(tokens)` — `mkdir({ recursive: true })` + `writeFile` JSON with `mode: 0o600` (user read/write only)
Change it to:   - `saveTokens(tokens)` — `ensureStorageDir()` + `writeFile` JSON with `mode: 0o600` (user read/write only)
```

```
oauth-auth-spec.md:229
You wrote:      - `saveConfig` merges a patch into the per-user config (preserves other keys), writes JSON with `mode: 0o600`
Change it to:   - `saveConfig` merges a patch into the per-user config (preserves other keys), writes JSON with `mode: 0o600`, and creates the storage directory `0o700` (asserts both the `mkdir` mode and the explicit `chmod`)
```

```
oauth-auth-spec.md:247
You wrote:      ### `tests/auth/login.test.ts` — 8 tests
Change it to:   ### `tests/auth/login.test.ts` — 9 tests
```

And on line 248, at the end of that list, after `clears the in-flight flow after the timeout`:

```
Change it to:   ...; clears the in-flight flow after the timeout; HTML-escapes error parameters in the callback response
```

I counted the tests with the vitest reporter rather than by eye: the file has 9.

Two things in that file you should **not** touch. The build-order table near line 30 is a record of the original build steps, not a live inventory. And the `tests/auth/config.test.ts — 4 tests` heading is already wrong on `main` (it has 5), so it is not yours to fix here.

## 2. CLAUDE.md states the wrong mode for the directory

`CLAUDE.md:502-504` says the directory is `0o600`. That was never achievable — a directory without its execute bit cannot be entered — and it is now specifically contradicted by your change.

```
You wrote:      2. **Per-user, on the user's machine:** `~/.familysearch-mcp/`
                   directory, `mode: 0o600`. Holds `tokens.json` (OAuth tokens from
                   `login`) and `config.json` (per-user tunables like `wikiApiUrl`).

Change it to:   2. **Per-user, on the user's machine:** `~/.familysearch-mcp/`
                   directory, `mode: 0o700`. Holds `tokens.json` (OAuth tokens from
                   `login`) and `config.json` (per-user tunables like `wikiApiUrl`),
                   both files `mode: 0o600`.
```

---

## Closing out my Windows question from last round

I asked someone to confirm `chmod` is a clean no-op on Windows. It is — you do not need to chase it. `tests/tools/no-project.test.ts` already skips a test on Windows for this exact reason, and Node's docs confirm chmod there cannot change read or execute bits. It returns normally, so nothing throws.

The consequence is worth knowing though, and it is why I asked for that clause in the spec: on Windows this directory fix does nothing at all. Both the `mkdir` mode and the `chmod` are no-ops. The fix holds on macOS and Linux. That is not something you can solve with POSIX modes, so there is no action here — it just should not read as fixed everywhere when it is fixed in two places out of three.

## Also worth a minute

Please update the PR description. It still describes only the original two-file version, so it does not mention `config.ts`, the shared `ensureStorageDir()` helper, or the `color` narrowing — that is half the diff and three of the six files.

And note for whoever signs off: every file here is under `src/auth/`, a stop-rule path in `docs/task-lifecycle.md` step 4, and the description has no "Deviated from the plan" line saying so.

One last mechanical note. Approvals here are not dismissed when you push, so this approval will still stand after your docs commit. That means nobody will have read that commit. Keep it to the text above and it is fine; if it grows beyond that, re-request review.

aghadiayeamayanvboernest (approved):
Approving. Both halves are correct, and the chmod is better than the audit asked for. I verified by running rather than reading.

Issue #1824, opened 2026-08-22
What makes a source original: is an image one, and what must be stored to prove we examined it?


DallanQ:
## DECIDED 2026-09-02 — ARK-or-fields, never free text; an uploaded image qualifies

`research_append` refuses `section: "sources"` writes carrying
`source_classification: "original"` unless the source entry carries a non-empty
`transcription`, a non-empty `image_filename`, or a stored FamilySearch **image** ARK
(`3:1:`). A `1:1` index ARK never qualifies — it is an index entry, someone else's
transcription of the record. There is no free-text escape: an agent-authored reason lets
the agent satisfy its own gate.

This is the already-decided rule — *an image you examined is original; a transcript or
snippet of one is derivative* — made mechanical.

**No schema change.** `source.url` is already an optional field and `arkToBareId` already
parses the `n:n:` type prefix, so telling an image ARK from an index ARK is a regex over
what we already store. Earlier option text priced this as needing a new structured field
and a five-site edit list; that was wrong.

**Uploaded images qualify, and making that true is part of this work.** People upload
Ancestry screenshots and photographs of personal documents; those are images they examined.
Today they cannot qualify — the FamilySearch image path saves the scan and sets
`image_filename`, but the PDF/image upload path reads the file and passes its text with a
`capture:` id, storing nothing. Store the uploaded file alongside the FamilySearch scans and
set `image_filename` from it. Without that half the rule is FamilySearch-only and demotes
genuinely-examined records.

**Measured across the committed e2e runlog corpus** (the "23 of 28" below was a much smaller
sample): **232 sources are classified `original`, and 211 of them — 91% — carry neither
field.** The largest single bucket is **146 called `original` on a bare `1:1` index ARK**,
63% of all originals and exactly what this rule stops. Twelve more sit on a `3:1` image ARK
with nothing else stored — genuine image reads that did not save a transcription, and this
rule keeps them. `image_filename` is populated 6 times in 232.

**Accepted cost:** roughly 199 of 232 get demoted to `derivative`, so the proved tier — two
independent original sources with primary information — becomes materially harder to reach
on runs that may have been substantively fine. Expect the tier rate to drop when this lands;
that is the rule working, not a regression.

**The alternatives it beat.** *Fields only* (transcription or `image_filename`, nothing
else) is simpler and provider-neutral, but demotes the 12 genuine image reads for no
genealogical reason and rests on `image_filename`, a field written 6 times in 232. *A
[...]

Issue #1442, opened 2026-08-07
eval harness: expected_classifications can't express "either party's record_role" or "either naming convention"


Cia-3:
> **Reviewed 2026-09-14 before junior handoff — vetted, held in Backlog.** No decision
> needed. Not promoted today only because both Ready pools were at target; it is pre-gated
> for the next fill. Read the premise correction first: half this body's evidence is dead.

## Premise correction (review, 2026-09-14)

**Section 1's motivating case is moot.** `ut_record_extraction_013` and `ut_record_extraction_020`
are both 1850 fixtures. Issue #1626's ruling, now enforced by
`test_pre_1880_census_creates_no_relationship_assertions`
(`eval/harness/validators/test_record_extraction.py:747`), forbids a pre-1880 census from
writing *any* relationship assertion, `indirect`/`_inferred` included. So the "child's side vs
parent's side" ambiguity cannot arise on those two fixtures, and their missing relationship
matcher is no longer a coverage gap — a different validator covers it, in the opposite
direction. **Do not add relationship matchers to ut_013 or ut_020.**

**Section 2 survives, on different tests than the body names.** `ut_record_extraction_018` did
not produce `son`/`daughter` in any of the five current logs. The live instances are ut_026 and
ut_027. Re-derived 2026-09-14 over the 5 committed logs (155 runs; the 2026-08-17/08-24 logs
the comments cite were evicted by PR #2018):

| test | matcher pins | roles the run actually wrote | outcome |
|---|---|---|---|
| ut_026 | `child_1` + `relationship` | `son_1`,`son_2`,`daughter_1`,`daughter_2` | 3/5 runs silently skip |
| ut_027 | `daughter` + `relationship` | `daughter_1`,`son_1`,`son_2` / `child_1..3` | 3/5 runs silently skip |
| ut_018 | `wife` + `relationship` | `wife` | 5/5 fired |
| ut_024 | `deceased` + `relationship` | `deceased` | 5/5 fired |

All four relationship matchers carry `optional: true`, so a role-name miss is a **silent
pass**. Corpus-wide there are 69 distinct `record_role` values (was 59 on 2026-08-28).

**The cost is worse than a missed check.** A matcher that matches nothing still leaves
`test_expected_classifications` reporting `passed: true` (verified on all 10 ut_026/ut_027
runs). `apply_deterministic_deference` (`eval/harness/harness/orchestrator.py:1151`) reads that
[...]

PR #2018, opened 2026-08-28
record-extraction deep dive: findings, prohibition list, and one grading fix

### LINKED ISSUE #1666: deep dive: record-extraction

Paaboat:
Closes #1666. Carries the folded-in pre-edit from #1635.

## What this is

A deep dive on `record-extraction` per [`docs/skill-deep-dive-guide.md`](../blob/main/docs/skill-deep-dive-guide.md) — starting from no known failure, reading transcripts before scores, and spending the time on the quiet passes.

Corpus: **all five committed unit run logs — 138 runs, 121 persisting, 2,416 assertions and 121 sources.** `text_response`, `tool_calls` and `file_changes` read before any `outcome_summary`, judge rationale or `.ann.json`.

**Fourteen findings, nine of which convert to validator requests.** Two lane-2 fixes attempted — one kept, one reverted after a paid run showed it misgrading. A third was deliberately not attempted, with the reason stated.

## Read this first, whoever reviews it

The most reusable thing in the findings doc is the dating note. **The five logs span two incompatible `evidence_type` doctrines.** Commit `61d7f919` (2026-08-21, #1628/#1706) rewrote the agent body to "stated-vs-inferred, with no exception" *and* flipped this suite's matchers in the same PR — `death-certificate-named-informant.json` went from requiring `age: indirect` to requiring `direct`.

So `ut_record_extraction_009` writes those facts `indirect` in the three older logs and `direct` in the two newer ones, and **that is not three failures** — the older runs were correct under the doctrine of their day. The agent body was edited eight times between 2026-08-14 and 2026-08-24. Any rate computed across all 138 runs for a classification behaviour is meaningless.

This killed three of my draft findings. Every surviving finding states the `git log -S` check that establishes its window.

## The three that matter most — all quiet passes

**F1. `record_role` encodes a relationship the record does not state, and no dimension grades `record_role`.** `ut_record_extraction_014` gives the same four people five different role vocabularies across five logs:

| log | Silas Kerrigan, 62 | Thomas Flynn, 42 | Mary Flynn, 38 | Patrick Flynn, 15 |
|---|---|---|---|---|
| 08-15 | `head_of_household` | `adult_male_1` | `adult_female_1` | `child_1` |
| 08-16_11-26 | `head_of_household` | `lodger_1` | `lodger_2` | `lodger_3` |
| 08-16_13-06 | `head_of_household` | `head_2` | `wife` | `child_1` |
| **08-17** | `head_of_household` | **`child_1`** | **`child_2`** | `child_3` |
| 08-24 | `head_of_household` | `co_resident_1` | `co_resident_2` | `co_resident_3` |

The 08-17 run roles a 42-year-old man `child_1` and a 38-year-old woman `child_2` of a differently-surnamed 62-year-old head — while its own narration says "His relationship to the Flynns is **unstated** (boarder, in-law, uncle?)" and "zero explicit relationship assertions were written". The 08-16_11-26 run roles all three `lodger_N`, asserting the boardinghouse reading the fixture says must not be assumed. `record-extractor.md:219-226` forbids both by name, and `record-extractor.md:368` notes person-evidence binds on `record_id` + `record_role`.

**F2. The one check that caught it reported it as a birthplace error.** The 08-17 failure message reads `assertions[a_012] (record_role='child_1', fact_type='birth' attribute='place'): value='Ireland' — expected to contain 'Pennsylvania'`. Every birthplace in that run is correct; what was wrong is that Thomas Flynn was roled `child_1`. The matcher selects on `record_role` and asserts on `value`, so a role fabrication surfaces as a value mismatch — and the sibling `lodger_N` run passed clean, because nothing matched `child_1` and `optional: true` skipped it.

**F6. The image chain is dead end to end.** The `image-reader` delegations the run log records in full — three of the four in the corpus, at 145, 147 and 197 characters — omit `project_path`, which `SKILL.md:100-106` bolds. (The fourth sits at F9's 200-char cap, so it is not claimed.) `image_filename` appears on **0 of 121** persisted sources, a count no truncation touches. The viewer can never show a scan for any source this skill writes.

## The one that explains why none of this was caught

**F9.** `builtin_tool_calls[].args` is truncated at 200 characters — **100 of 103 `Agent` prompts in the corpus are exactly 200 characters long.** Five `SKILL.md` prohibitions live only in the delegation message, including "never frame the task as fix/correct the existing tree" (documented as having induced destructive edits) and "never instruct the agent to create `person_evidence` links" (documented as having produced a fabricated identity link carrying a match score no tool computed). Neither is auditable by anyone. My first pass reported "recordId missing from 80 of 99 delegations" — an artifact of the truncation. That is how the gap presents: not as absent evidence, but as false evidence.

## Also found

`record_id` split across two ARKs on one record (2 runs, incl. the newest) · negative assertions given a record informant (2 runs, rule verified live in the body on that date) · a `"No parents' names recorded"` field-level negative the body prohibits by name · the router re-printing per-assertion detail in **24 of 25** runs of `v1_2026-08-24_17-40-15`, which `rubric.md:5` deliberately makes undockable · role-token drift (`daughter` beside `son_1`/`son_2`) · zero coverage of the classification-refinement path (`ut_011` is a 0-turn triggering test) · reasoning prose in `value` · and a schema line telling you to document a new `fact_type` "in the assertion's notes", a field assertions do not have.

## Changes

| File | What |
|---|---|
| `docs/deep-dives/record-extraction-prohibition-list.md` | **new** — 107 transcript-checkable rules. Step-1 deliverable; the next auditor starts here. |
| `docs/deep-dives/record-extraction-findings-2026-08-28.md` | **new** — 14 findings in Did/Should/Gap, 9 validator requests, follow-on work grouped by paid run. |
| `census-1860-different-surname-head.json` | **unchanged** — a `judge_context` bullet was tried here and reverted after it misgraded; see the eval-run section. Matches `origin/main`. |
| `marriage-index-no-parents-recorded.json` | lane-2 — the bullet said "do not dock for creating one, and do not dock for omitting one", which cannot discriminate and licensed the one negative-evidence shape `record-extractor.md:896-905` forbids. Replaced with the body's own test: can you name the absent person? |
| `packages/engine/plugin/agents/record-extractor.md` | **#1635 pre-edit** — deletes the dead ToolSearch passage (4 lines, stopping mid-line at `net.` as #1666 specifies). Dead because the agent's `tools:` list is MCP-only and grants no built-ins, so `ToolSearch` was an instruction it could never execute — billed on every invocation of the most-invoked agent in the plugin. `grep -rn ToolSearch packages/engine/plugin/agents/` is now empty, which is why the allow-list below goes to zero. |
| `tests/packaging/agent-tool-names.test.ts` | **#1635 pre-edit** — removes the `BUILTIN_BODY_ALLOWLIST` row that named #1666 as its exit, plus two stale #1666 comments. |

**No `rubric.md` change**, and **no SKILL.md or agent-body prose added.** Every behavioural finding is a rule the body already states and the run ignored; per the guide, restating it lengthens the prompt and changes nothing. The rubric-adjacent findings are arithmetic and belong in validators — `rubric.md:5`'s refusal to deduct for presentation exists for a good reason I did not want to erode.

**One lane-2 fix deliberately not made.** Tightening `ut_016` to require the image-confirmation *attempt* `SKILL.md:126-138` mandates would fail the test 5 of 5: the fixture declares only `record-read-birkeland-1817-baptism`, so there is no `volume_search` response to receive. That is fixture work and it is filed instead. §F7 has the reasoning.

## One departure from #1635

#1635's comment says "an empty array is fine" for `BUILTIN_BODY_ALLOWLIST`. It is not quite — vitest errors `No test found in suite` when the per-row `describe` registers nothing. Added a placeholder inside that block, mutually exclusive with the per-row checks by construction. The non-vacuity guard that outlives the allow-list is corpus-independent and untouched.

## The eval run — done, three times

The slot #1666 buys was spent on **three** runs, all committed. All three used judge prompt `03f306ff`, so they are comparable to each other; none is comparable to the 08-24 baseline.

| | `v1_2026-08-28_17-43-01` | `v1_2026-08-28_18-23-42` | `v1_2026-08-31_10-34-08` |
|---|---|---|---|
| state | both lane-2 fixes in | after the `ut_014` revert | after the `ut_021` judge_context repin |
| outcomes | 20 pass / 4 partial / 4 fail | **24 pass / 3 partial / 1 fail** | 23 pass / 4 partial / 1 fail |
| cost | $8.82, 26 min | $9.00, 26 min | $8.89, 72 min |
| annotated | — | yes — 37 cells, 0 score changes | yes — 34 cells, 0 score changes |

**Mostly jitter — but `ut_022` is not, and an earlier revision of this section wrongly
lumped it in.** Six of the seven moves are judge-side and go both ways: `ut_005`/`ut_027`
partial→pass and `ut_014` fail→pass, against `ut_016`/`ut_023`/`ut_026` pass→partial.

`ut_022` (`census-1870-indexed-surname-variant`) is a different kind of failure and wants a
look before merge. It is **not** an abort (`aborted_reason` is null) and it was **not graded**
— zero dimensions, which is why the annotation does not and need not cover it. It failed a
*validator*, `test_expected_classifications`, with six violations:

```
- no new assertion with record_role='child_1' fact_type='name' (expected at least one)
  ... through child_6
```

Its transcript shows four tool calls — `project_context`, `research_log_append`,
`project_context`, `extraction_append` — with **no `record_read`**, where the sibling census
tests read the record first. It passed with validators green and 6 graded dimensions on
`v1_2026-08-28_18-23-42`. So the extraction was appended without the record being read; that
is a behavioral miss, not judge noise, and `check_runlogs` itself warns not to treat an
ungraded test as a clean pass. Whether it is run-to-run variance in the agent (the skill run
is not temperature-0) or something reproducible is **not** established by one run.

`ut_021`, the only test whose fixture this PR changed, scored identically in both runs
(`Correctness` 2, all else 3) for the same §F5 reason.

**Annotated:** `v1_2026-08-31_10-34-08.ann.json` — 34 cells over the five sampled tests
(`ut_010`, `ut_014`, `ut_020`, `ut_024`, `ut_026`) plus four on `ut_027`, one written comment
on `ut_026`, no score changes. This clears rule 3; rule 2 remains and is what the
`eval-cosmetic-skip` request covers.

**`check_runlogs`: all runlog rules satisfied.** The annotation landed
(`v1_2026-08-28_18-23-42.ann.json`) — 37 cells, 7 tests, zero score changes.

The single written comment in the whole file is on `ut_021` Correctness (judge 2, human 2):

> "Skill wrote `a_009`/`a_010` as negative evidence for unrecorded parent name fields on
> present persons. Per `record-extractor.md:896-905`, negative evidence applies to
> expected-but-absent persons, not blank fields."

A genealogist reviewing blind docked exactly the behaviour F5 describes, citing the same
lines the fix cites — and under the `judge_context` this PR replaces, that dock was
impossible by construction ("do not dock the skill for creating one, and do not dock it for
omitting one"). That is independent confirmation of the lane-2 fix, not just of the finding.

`ut_014` is not in the sample **by design**: its judge was skipped because the validator
failed first, so it has zero graded dimensions and `review_sample.is_gradeable()` excludes
it — with `rule3_completeness`'s warning as the compensating control, which duly fired. I
checked that before treating it as a calibration gap; it is not one.

### The pair is a natural experiment, and it reproduced F1 and F2 live

`judge_context` reaches the judge and never the skill, so the revert cannot have changed
what the skill wrote. Run 2 wrote `head_of_household` + **`child_1` = Thomas Flynn (42)**,
**`child_2` = Mary Flynn (38)**, `child_3` = Patrick Flynn (15) — the exact 2026-08-17
defect, recurring today under current doctrine. It failed with the **identical** message:

> `assertions[a_012] (record_role='child_1', fact_type='birth' attribute='place'): value='Ireland' — expected to contain 'Pennsylvania'`

Every birthplace in that run is correct, and the judge was skipped because the validator
decided the outcome first. So F1 is not a stale-doctrine artifact of the older log — 2 of
the 7 runs now on record fabricate — and F2's misdirection is confirmed verbatim.

### One lane-2 fix kept, one reverted — and the difference is the finding

**`ut_021` kept, and it works.** Run 2's skill wrote the field-level negatives again
(reworded), and the judge docked them, quoting the rule back: *"negative evidence applies
to PERSONS expected-but-absent, not blank FIELDS on present persons… The skill should have
remained silent."* Under the old clause that behaviour was ungradeable by construction.

**`ut_014` reverted.** Its bullet made the judge score a run with tie-neutral roles and
zero relationship assertions `Correctness = 1`, calling `head_of_household` a fabricated
relationship. `ut_014` is back to `origin/main`.

The difference is shape, not "prose works / prose doesn't": `ut_021`'s bullet states a
**test applied to what the skill wrote**; `ut_014`'s asked the judge to **classify a label
against an open vocabulary**, which has no boundary it will hold. That is the argument for
V1 in #2019, where the check is a surname comparison and cannot over-apply — and it is
`docs/skill-lifecycle.md` §5 / ADR-0011 with a measured instance.

### What else the runs turned up

- **#2025 filed** — the harness caps its `node --eval` subprocess at 30s
  (`_run_node_eval`'s `timeout` default, in `eval/harness/harness/mock_mcp.py`). Run 1
  timed out twice at 18 and 21 ops while 14 calls of 15–41 ops
  succeeded, including 41 and 39: concurrency load, not batch size. Run 2 confirms — zero
  transient retries, and `ut_009`/`ut_028` went `partial` → `pass` with nothing changed
  that they touch. It cost real grades in run 1.
- **A correction to my own write-up.** §What I did not find attributed every second
  `extraction_append` call to an update op or a validation rejection. A third cause exists
  and I could not have seen it: the four older logs record **no tool responses at all**.
  `response` appears on `tool_calls` for the first time in these two runs — which also
  makes the "identifier must trace to a tool response" class of validator writable.
- **Three violations the earlier 2,416-assertion corpus never produced**: `ut_027`
  classifying a census image `original` (**both runs** — the most persistent), `ut_023`'s
  burial informant at `official_duty`/`primary`, and `ut_017` tripping
  `test_relationship_type_agrees_with_its_value` with `relationship_type: "child"` on a
  value reading "Sister of" — which writes the wrong family edge at materialisation.

## Self-audit before review

Three narrative claims were reconciled with the branch after the runs landed; no
measurement changed.

- The doc header and §F1 still said the `ut_014` `judge_context` fix was **made**. It was
  reverted, and `census-1860-different-surname-head.json` is byte-identical to
  `origin/main`. §F1 now records the attempt and why it misgraded. The PR title said
  "two grading fixes" for the same reason and now says one.
- §F6 claimed all four image delegations omit `project_path`. §F9's own truncation makes
  that unprovable for one of them — the four prompts measure 145, 147, 197 and **200**,
  and the 2026-08-17 one sits at the cap. §F6 now rests on the three complete prompts plus
  the 0-of-121 `image_filename` count.
- **This PR prunes two of the five run logs the doc analyses.** Retention keeps the newest
  5 candidates per skill, so committing the two fresh runs evicted
  `v1_2026-08-15_12-52-37` and `v1_2026-08-16_11-26-54` (with their `.ann.json` siblings) —
  two rows of the F1 table among them. Every count was computed against all five before
  the eviction; the corpus note now names them and gives the `git show` line to read them
  at the merge-base.

### Second pass — re-derived every remaining claim against the run logs

Four more things needed fixing. Everything else re-computed clean: the corpus counts,
F1's five-row table, F3, F4, F8, F9, F11, F12, F14, V8's 80-of-138, F10's 56/22/5, the
census zero-claims, the enum and invented-field zeros, both schema trees, every `.md`
line cite, all six issue references and both cross-posted comments.

- **A count was wrong.** §What I did not find claimed "**0** duplicate
  `(role, fact_type, value)` tuples across the corpus". There is exactly one —
  `ut_record_extraction_017`, 2026-08-16_11-26, writes the same death fact twice
  (`a_003` without `date`/`place`, `a_004` with) — and a second appears in fresh run 1 on
  the same test. That zero was the evidence dismissing the 15 two-call runs, so it is
  corrected in place, not deleted.
- **The doc contradicted itself on F10.** §F10 said "filed, not made"; §Follow-on work and
  this PR body both say F10 is deliberately *not* filed. §F10 now agrees.
- **§F7 printed a four-call tool sequence.** All five runs make five calls — the sequence
  omitted a second `project_context`, which is the router-then-agent split V8 describes.
  Also updated to 7 of 7, since both fresh runs skip the image too.
- **`make judge-report` no longer says what #1666 says it says.** The issue quotes **1 of
  6** non-discriminating dimensions; against the newest log it is now **3 of 6**, because
  the command reads the newest log per skill and that is now run 2. A new section records
  the per-run spread — the two fresh runs disagree about *which* dimension is dead — and
  explains that this is mostly run quality, not a dead rubric. **F13 depends on that
  flatness and survives**; it now says so explicitly. No rubric change follows, because
  acting on it would buy a paid run this PR cannot fund.

Also: §F3 said "the newest" log and now names it, and eight prohibition-list line ranges
that started or ended on a blank line are tightened to their content.

## Not batched

**#1824 is not in this PR**, as instructed.

## Merge order — resolved: #2046 landed the shift this PR anticipated

This section previously warned of a rename/rename conflict with **#1955**. That is now
historical and the warning is withdrawn — recorded because the `ut_021` repin only makes
sense against it.

**#2046** ("agent frontmatter: the omission binds, not the deny — measured, and the denies
deleted") merged and did the work #1955 was going to: it deletes the 20-line
`disallowedTools:` block from all three plugin agent bodies, converting the guard from
"must deny under every spelling" to "must not appear in `tools:`", plus a grant/deny
overlap check. `main` was merged into this branch at `704ba456`, so that is already here.
**Not a fail-open regression** — the omission binds, which is the measurement in its title.

**#1955 is still open** and no longer collides: against current `main` it is 5 changed
lines of comment rewording across the three agent bodies, and it no longer touches
`eval/runlogs/unit/record-extraction/` at all. Its stale "Didn't change" review guidance is
its own business now, not a merge hazard for this PR.

**The repin is load-bearing, not anticipatory.** #2046 shifted `record-extractor.md` by
-20 lines and the merge shifted it -20 again. Since `ut_021`'s judge_context was written,
the "Negative evidence" heading has moved **875 → 855** and the `value` bullet
**883 → 864**. The original `record-extractor.md:896-905` cite would now miss the rule
twice over; the section cite still resolves, and the judge no longer copies a line number
into the run log.

**Consequence left in place:** the **26** `record-extractor.md:NNN` cites in this PR's
findings doc are now *wrong*, not about-to-be-wrong. Deliberately unswept per the lead and
tracked on #1305 (above). Nothing fails — that is the gap.

**Run-log retention:** #2046 landed `v1_2026-08-30_17-17-24`, so post-merge the directory
holds **six** candidates against a keep-5 policy. Recorded rather than pruned — no CI job
checks the count, and `prune_runlogs.py` would evict `v1_2026-08-17_18-57-51`, which the
findings doc reads throughout. Whoever merges may want to run it deliberately.

This PR is currently `BEHIND` — `main` moved again after the merge commit.

## Before merging — `runlogs` needs a run made after the FINAL merge-of-main

**`runlogs` is red, and it was red before this PR's last push.** On the merge commit
`704ba456` it already failed against #2046's *own* `v1_2026-08-30_17-17-24` (3 differing
snapshot files + 2 uncommented sampled corrections). This is inherited, not introduced.

The merge of `main` landed #2046 and #2044, which changed four files the run log
snapshots — so no run made before that merge can be "active":

```
packages/engine/plugin/agents/image-reader-opus.md      missing-on-disk   (#2044 retired it)
packages/engine/plugin/agents/image-reader.md           content-differs
packages/engine/plugin/agents/record-extractor.md       content-differs   (#2046)
packages/engine/plugin/skills/record-extraction/SKILL.md content-differs  (#2044)
```

`git diff --name-only 704ba456..HEAD` confirms the three commits here touch **none** of
them.

**Deliberately not re-run yet.** This PR is `BEHIND` and `main` moved three times today.
Every merge-of-main re-stales the snapshot, so a run now ($8.89 / 72 min last time) would
be paid for twice. Sequence for whoever merges:

1. Merge `main` in **last**.
2. `cd eval/harness && uv run python run_tests.py --skill record-extraction`.
3. **A genealogist annotates the new log.** `check_runlogs` requires a written comment on
   any sampled dimension that is not a confirmed pass (judge 3, human agrees 3) — that is
   the second error on this check and it is not something the run itself clears.
4. Consider `scripts/prune_runlogs.py`: the directory will sit over the keep-5 policy (see
   Merge order above), and pruning evicts `v1_2026-08-17_18-57-51`, which the findings doc
   reads — so do it deliberately, not incidentally.

**`eval-cosmetic-skip` has been requested** — see the request comment on this PR. An
earlier revision of this section said the label was the wrong escape here, on the grounds
that #2046 changed the agent's permission surface and #2044 changed `record-extraction`'s
`SKILL.md`. That reasoning was misattributed: the waiver asks whether **this PR's**
skill-side change is behavior-neutral, and those are `main`'s changes. This PR's entire
skill-side diff vs current `main` is `ab4d09f3` — one dead 4-line ToolSearch passage
deleted per #1635 — which was in place for all three paid runs. Rule 2's staleness is
inherited, and it was red on `704ba456` before this branch's last push.

**The label does not green the check on its own.** `check_runlogs.py` relaxes rule 2 only;
rule 3 is blocking and the missing `v1_2026-08-31_10-34-08.ann.json` stays red until a
genealogist reviews the five sampled tests. Step 3 above is required on every path,
re-run or waiver. The bypass is also voided by the next push, so it must be applied after
the final commit.

## Verification

- `npx vitest run tests/packaging` — **598/598.** (An earlier revision of this PR reported 597/598: `windows-wrappers.test.ts` walked into `.claude/worktrees/`, gitignored local worktrees, and reported their `.bat` files as the repo's own. `11f393ea` on `main` excludes `.claude` from that walk, and this branch now carries it.)
- `uv run pytest tests/unit` (eval harness) — **2924 passed, 3 skipped**.
- All 28 fixtures parse.

## Follow-on work — filed

The one search DEVELOPMENT.md prescribes (194 open issues, plus a grep of every `**Touches:**` line) turned up two existing owners, so two of the nine validator requests became comments rather than new issues.

- **#2019** — `developer`, `nothing-checks`. Validators V1, V2, V4(b), V5, V7, V8, V9. All read the committed corpus; no eval run.
- **#2020** — `developer`, `nothing-checks`. F9 — `BUILTIN_ARG_TRUNCATE = 200` in `eval/harness/harness/skill_runner.py` cuts the `Agent` prompt. Unblocks V4(a). F14's schema-text fix rides here.
- **#2021** — `genealogist`, `nothing-checks`. F7 and F12, the two fixture gaps; they share one paid run.
- **#2025** — `developer`, `nothing-checks`. The 30s `node --eval` cap in the eval mock, found by this PR's own runs.

Went to existing issues instead:

- **V3 → comment on [#986](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/986#issuecomment-5455645081)** — it already owns `evidence_type: negative` ⟺ `record_role: absent` in `validator.ts`. The comment adds the informant half of the same sentence, and notes that the `a_012` blank-field negative its "Why it was deferred" section describes reproduces live in F5.
- **The `.md:NNN` cite gap → comment on [#1305](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1305#issuecomment-5477318889)** — it already owns `doc-links.test.ts` + `repo-paths.ts` and the line-cite hatch, and self-describes as one job. Not covered by its step 1: `citedLineNumbers` matches `{py,ts,mjs}` only *and* the walk never leaves `docs/`, so `.md:NNN` cites are unreachable rather than grandfathered. **90** exist under `docs/` (34 targeting an agent body or `rubric.md` — prompt source, not prose), **39** of them in this PR's own findings doc. Left in place at the lead's call; the one such cite under `eval/tests/` is fixed here. Part of that sweep is not mechanical — several cites sit inside verbatim quotations of judge output, where rewriting the cite falsifies the quote — which is why it wants a stated rule before anyone starts.
- **F2 + V6 → comment on [#1442](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1442#issuecomment-5455650778)** — it already owns the single-role matcher pin. The comment adds two failure directions its body does not describe: the pin produced a **false-positive report** on `ut_014` and a **silent miss** on the `lodger_N` run.

**F10 is deliberately not filed** — a one-line `SKILL.md` addition naming where `projectPath` comes from. It arms this skill's eval slot, so it should ride the next body change rather than buy a run of its own. Recorded in the findings doc so the next person editing that body picks it up.


EdmondOware (approved):
Approving.

I went looking for four problems and the doc had pre-empted all four, which is the strongest thing I can say about it. Naming the five logs, stating that this PR itself evicts two of them, and giving the `git show <base>:` command to re-derive them is more than I would have asked for. F6's count survives that because it is scoped to the named corpus rather than to whatever is on main, and the F9 truncation caveat is applied honestly, including the one prompt you decline to claim.

The F1 revert is the part worth other people reading. A judge asked to classify a label against an open vocabulary has no boundary it will hold, and the run scoring Correctness 1 on tie neutral `resident_N` roles is the proof. Putting that misgrade in the finding itself rather than only in the lane 2 section is what stops the next reader repeating it.

The kept lane 2 fix is right and I checked it against the runs rather than the reasoning. The old bullet licensed exactly what `record-extractor.md:896-905` forbids by name, and 2026-08-24 passed while writing that shape. Under the new bullet the run that writes the negatives is docked for writing them, and the run that stays silent is docked only for an unrelated atomicity slip. It discriminates in the direction you intended.

The ToolSearch deletion checks out: the agent's `tools:` list grants no built ins, so the instruction was unexecutable, and the test keeps an anti silent zero guard now that the allow list is empty.

jud-sdev (approved):
Went through this line by line and ran the suites in a worktree off the head — packaging 608/608, tsc clean. The canary in `agent-tool-names.test.ts` genuinely drives the real scanner over a synthetic body, so coverage survives the allow-list going empty, and the ToolSearch deletion is clean with nothing left in any agent body. The lane-2 fix on `marriage-index` discriminates the way you intended — the judge_context prose states both rules correctly, and I checked it reads the same way EdmondOware verified against the runs.

One thing has to happen before this can merge, and it isn't in your diff: the `runlogs` check is red. Merging main pulled in a newer record-extraction run log (`v1_2026-08-30_17-17-24`) that is now the latest, and your edits to `record-extractor.md` and `marriage-index-no-parents-recorded.json`, plus main's own #2046 edit to `image-reader.md`, make its snapshot stale — its annotation (which also came from main) is missing two comments as well. A fresh harness run on this branch, annotated, clears both rules. `eval-cosmetic-skip` won't fit here, since the `marriage-index` change is a grading change rather than behavior-neutral. Worth flagging to the lead that two record-extraction changes are racing through main and keep invalidating each other's runlog gate, so the re-run should land after any other in-flight one.

Small nit, not blocking: the `record-extractor.md` line numbers in the `marriage-index` judge_context (896-905, 883-886) don't line up with the current file — the ToolSearch deletion and earlier body edits shifted them. The prose is right; only the pointers drifted.

jud-sdev:
Hi @Paaboat — the code here is solid and I've approved it. One thing is blocking merge and it isn't in your diff: the `runlogs` check is red.

What happened is the `main` merge. It pulled in a newer record-extraction run log (`v1_2026-08-30_17-17-24`) that's now the latest one the check looks at. Your edits to `record-extractor.md` and `marriage-index-no-parents-recorded.json`, plus main's own change to `image-reader.md`, make that log's snapshot stale — and its annotation (which also came from main) is missing two comments. So it's the gate going stale, not anything wrong in your changes.

To clear it, re-run the harness on this branch:

```
uv run python eval/harness/run_tests.py --skill record-extraction
```

Then annotate the new run in the CRUD UI — write a sentence on every sampled dimension that isn't a clean pass. Commit the new run log and its `.ann.json`; retention prunes the older ones. That gives the check a latest log that's active against the branch and fully annotated, which satisfies both rules it's failing on.

`eval-cosmetic-skip` won't fit here, since the `marriage-index` change is a grading change rather than behavior-neutral.

One heads-up before you spend the run: there's another record-extraction change moving through main, and if it merges first you'd have to re-run again. Worth a quick check on sequencing so the run isn't wasted.

Minor, not blocking: the `record-extractor.md` line numbers in the `marriage-index` judge_context (896-905, 883-886) don't line up with the current file anymore — the prose is right, only the pointers drifted. Easy to fix while you're in there, but nothing depends on it.

Paaboat:
## Requesting `eval-cosmetic-skip` (rule 2) — and why it does not green the check on its own

**Request:** a senior applies `eval-cosmetic-skip` to bypass rule 2 on the current head
(`2e02c2fd`).

### Why the waiver fits

The waiver is for "a PR whose only skill-side change is behavior-neutral". Measured against
current `main`, this PR's *entire* skill-side diff is one hunk:

```
git diff --stat origin/main...HEAD -- packages/engine/plugin/
 packages/engine/plugin/agents/record-extractor.md | 5 +----
 1 file changed, 1 insertion(+), 4 deletions(-)
```

That is `ab4d09f3`, deleting the dead ToolSearch passage per #1635 — and it was in place for
**all three** paid runs on this branch (`v1_2026-08-28_17-43-01`, `v1_2026-08-28_18-23-42`,
`v1_2026-08-31_10-34-08`). So this PR's skill-side change is not unmeasured; it is measured
three times.

The four files rule 2 reports as differing were changed by **`main`**, not by this PR:

```
agents/image-reader-opus.md       missing-on-disk   (#2044 retired it)
agents/image-reader.md            content-differs   (#2044/#2046)
agents/record-extractor.md        content-differs   (#2046)
skills/record-extraction/SKILL.md content-differs   (#2044)
```

`git diff --name-only 704ba456..HEAD` confirms the three commits pushed here touch none of
them. Rule 2 was already red on the merge commit `704ba456` before this PR's last push,
then against #2046's own `v1_2026-08-30_17-17-24`.

### What the label will NOT fix

`check_runlogs.py` is explicit that "only rule 2 is relaxed: rules 1 and 3 still run".
**Rule 3 is blocking and stays red:**

```
skill `record-extraction`: latest run log `v1_2026-08-31_10-34-08.json` has no
annotation file (`v1_2026-08-31_10-34-08.ann.json` missing).
```

That needs a genealogist to review the five sampled tests — `ut_010`, `ut_014`, `ut_020`,
`ut_024`, `ut_026` — with a written comment on any dimension that is not a confirmed pass.
A re-run does not clear it either, so the annotation pass is required on whichever path is
taken. Dropping the 08-31 log does not help: latest would become `v1_2026-08-30_17-17-24`,
whose annotation CI already flagged for 2 uncommented corrections.

### Timing

The bypass is voided by the next push and decided from the event, so it must be applied
**after** the final commit and re-applied after any further one. This PR is currently
`BEHIND`; if `main` is merged in again, the label needs re-applying and rule 2's file list
will change.

Not applying the label myself — it is a senior waiver by design.

Paaboat:
## @PioneerAIAcademy/senior-genealogists — one annotation pass needed to unblock this PR

`check_runlogs` rule 3 is blocking on a missing annotation, and **this is the one step no
re-run and no waiver can clear.** The `eval-cosmetic-skip` request above covers rule 2
only; `check_runlogs.py` relaxes rule 2 and states that "rules 1 and 3 still run".

**The log to annotate:** `eval/runlogs/unit/record-extraction/v1_2026-08-31_10-34-08.json`
(28 tests, judge prompt `03f306ff`, 23 pass / 4 partial / 1 fail).

**The five sampled tests** — these are what the harness drew, and rule 3 reads only these:

| Test | Outcome this run | Moved since `v1_2026-08-28_18-23-42`? |
|---|---|---|
| `ut_record_extraction_010` | pass | no |
| `ut_record_extraction_014` | pass | **yes — was `fail`** |
| `ut_record_extraction_020` | pass | no |
| `ut_record_extraction_024` | pass | no |
| `ut_record_extraction_026` | **partial** | **yes — was `pass`** |

**What rule 3 requires:** a correction entry for every dimension of each sampled test, and
a **written comment** on any dimension that is not a confirmed pass (judge 3 and you agree
3). The two movers above are where the comments will be needed; the earlier failure mode on
this check was exactly "2 sampled correction(s) with no comment".

**How to run it:** `eval\Start.bat` (the test-creation / CRUD app), or `make eval-ui` on
mac/Linux.

### Worth knowing before you grade

- **This run is jitter, not a regression.** Seven tests moved and they moved both ways, and
  the single fail *relocated* (`ut_014` → `ut_022`) rather than persisting. Aggregate went
  24/3/1 → 23/4/1 on the same judge prompt. Please grade the five on their own merits — the
  framing is here so a mover does not read as a new defect.
- **`ut_026`'s partial** is `Informant identification`: the judge argues that for a
  christening the presenting parents are the informants for the child's own facts, where the
  skill used `unknown`. That is a genuine adjudication call and the kind of thing a written
  comment is for.
- **`ut_021` is _not_ in the sample**, so it needs nothing from you here. Noting it only
  because it is the test this PR changed: it scored `Correctness` 2 in both runs for the
  same reason (§F5's blank-field negative — two manufactured negative assertions for parents
  the index does not record). Stable across two runs, so it is a real skill finding rather
  than noise.

### One caveat on the label

This PR carries `ready-for-senior-genealogist`, whose definition is "CI green + peer-approved".
CI is **not** green — `runlogs` is red, which is what this comment is about. The rest of the
checks pass (`vitest` 608/608, `pytest`, `server-pytest`, `scan`, `lockfile-drift`).

EdmondOware:
Recording one fact about the committed run so it is not rediscovered later, since this becomes main's baseline for `record-extraction`.

`v1_2026-08-31_10-34-08` was executed against a checkout that predates #2044, not against this branch. Its snapshot records `record-extraction/SKILL.md` at `d7052ca6`, which matches only `39d211d3^`. Every commit on this branch hashes that file to `0f584e06`, and the branch has carried #2044 since `704ba456` on 30 August. The run itself ran on 31 August at 10:56 UTC, so the workspace picked up the judge_context edit but never the merge.

That is what the four "NOT active" entries are: `image-reader-opus.md` missing on disk, plus `image-reader.md`, `record-extractor.md` and `record-extraction/SKILL.md` differing. The main thing the run did not see is the `@plugin:image-reader-opus` escalation paragraph that #2044 deleted from the skill body.

Most likely immaterial to the numbers: there are zero `image-reader-opus` delegations anywhere inside `tests[]`, so that path almost certainly never fired.

Two things worth crediting while this is on the record. The run accounting holds, 26.8 minutes wall across parallel tests and 71.9 summed against the 72 claimed, and $8.48 against $8.89. And the three runs differ from each other by exactly one file each, `census-1860-different-surname-head.json` then `marriage-index-no-parents-recorded.json`, so the jitter comparison is well controlled even though the baseline underneath it is behind.

For whoever touches this skill next: the next run from a clean checkout resets all of it.

PR #2024, opened 2026-08-28
search-records: fix census-inference validator bugs, resolve skill contradictions, close out issue #1642 deep dive

mercyokum:
**The issue:** `search-records` had a deep-dive review open (#1642) with findings from five people across seven comments, plus a few flaky/failing tests in its eval suite. Some of those findings had genuine bugs in a test validator, a couple were contradictions in the skill's own instructions, and two eval tests kept failing for reasons nobody had root-caused yet.

**The fix:**
- Fixed a real bug in the census-record validator: it was missing several ways the skill correctly says "this comes from an index, not the original record" (like "Indexed as Head of Household" or "presumably her mother"), so it was wrongly failing tests that were actually doing the right thing. Verified this fix against every past test run before shipping it, so it would not accidentally start accepting notes that don't actually explain themselves.
- Fixed the one real content gap this found: the skill's rule about marking census relationships as "inferred, not stated" only kicked in when there was already a conflict to explain — so a clean, no-conflict match slipped through with nothing to say. Made the rule apply to every case.
- Resolved a contradiction where one part of the skill said "never offer to extract a record you're not sure about" and another part said the opposite — now it is a single consistent rule based on whether the match is confirmed, not on a score.
- Resolved a similar contradiction about whether the skill is allowed to batch its search-log entries together or must log each one immediately — it can batch, but has to save its work every few searches instead of holding everything until the end.
- Built two new automated checks so two of the reported problems (ignoring a location hint the search tool already handed back, and asking permission before running a search step the skill was already supposed to run) get caught automatically from now on, instead of relying on someone noticing by eye.
- Fixed a fixture gap that was intermittently failing one test for an unrelated reason (a record the test setup never gave it).
- A handful of other SKILL.md and reference-doc additions from this deep dive that the fix list above doesn't call out individually: granting `wiki_search`/`wiki_read` and pointing to them for the historical-spelling explanation, corrected passenger-list date coverage (1820 on, not every year), the given-name truncation order (try each name in the string, not just the first), "5+ variations is a floor, not a ceiling" for a well-covered collection, and surname-wildcard/misindexed-value guidance in the reference files. All from this same deep dive; listed here so they're attributed rather than silent drive-bys.
- Ran the skill's full suite end to end after every round of review — latest run (`v1_2026-09-02_10-04-24`, after merging main's own #2085 SKILL.md rewrite) is 27 pass / 1 partial / 1 fail / 1 xpass of 30 tests. The one hard failure (`_010`) and the one partial (`_028`, first occurrence) are both pre-existing, unrelated single-run variance -- `_010` fails the same pre-1880 census-hedge validator tracked in issue #2103 (has ranged 1-6 failures across every committed run of this skill, on main and this branch alike), and `_028`'s partial is an isolated judge ding on a redundant tool call with no validator failure. Neither traces to anything this PR touches. Main's own three committed runs against the identical SKILL.md are 24 pass / 4 fail / 1 xfail of 29 -- this branch is not a regression against that baseline.
- Wrote up the full audit as two documents in `docs/deep-dives/`: a checklist of every rule in the skill that can be verified against a real run, and a findings report explaining exactly what was fixed, what was already fixed before this PR, and what is still open for the next person.

**Deferred issues found:**
- One test (`ut_search_records_013`, and now also `_012`) still sometimes states a household relationship as fact instead of "inferred" even when nothing else is wrong with it. One reword of this skill's instructions already landed in this PR and did not measurably fix it, so a second attempt is left open rather than repeating that without new evidence — documented in the findings doc for whoever picks it up next.
- A contradiction about whether the skill should ever drop a person's first name while searching is explicitly parked pending a decision from the lead — it is tangled up with a separate open question (#1008) and should not be resolved here.
- A few smaller findings from the review (a test fixture that may be leaking its own answer into a test, one test hitting a turn-count ceiling shared with two other skills, and two reference-doc wording nits) are recorded but not fixed in this PR — full list in `docs/deep-dives/search-records-findings-2026-08-28.md`.
- Issue #1642 stays open — it's a deep-dive tracking issue and the items above (the givenName-drop contradiction, the residual `_013` content gap, and the smaller findings) are real, unfiled work this PR doesn't close out.


chesworthrm (changes requested):
Reviewed the merge with `main` (branch is 1 commit behind), not just the branch head.

**What this does:** closes out the `search-records` deep dive — loosens the census inference-marker regex, adds two tag-gated validators plus a universal parent-age validator, adds a packaging lint that checks the lever table against the real `validateInput`, and rewrites several SKILL.md rules.

Nice work on the wiring of `test_no_permission_ask_before_mandated_lever` — tagging `non-derivative-nickname-bitsie.json` is exactly right, and it's the reason that one actually runs.

Three things to fix before this merges.

---

### 1. CI is red — a line citation in the new findings doc

`doc-links.test.ts` bans source-line citations under `docs/`. There is one.

```
You wrote:      (`runlog.py:250`) treats as non-activation regardless.
Change it to:   (`eval/harness/harness/runlog.py`) treats as non-activation regardless.
```

The sentence already names `derive_activated`, so the line number was not carrying anything. I applied this and the file goes green.

---

### 2. `test_jurisdiction_hints_followed` never runs

The validator skips unless a test carries the `jurisdiction-hints-followed` tag. No test carries it. I parsed every fixture JSON under `eval/tests/` and `eval/fixtures/` — 1554 files, and that tag appears on none of them.

It would also skip for a second reason. The validator needs a `record_search` response with a `jurisdictionHints` field, and no fixture in the repo has one.

So the guard passes vacuously today. That matters because two documents say it closes the finding:

```
You wrote:      | guard, tag-gated (`test_jurisdiction_hints_followed`, `jurisdiction-hints-followed` — built this session, closes mercyokum's Finding 1) |
Change it to:   | guard written, not yet wired — no test carries the `jurisdiction-hints-followed` tag and no fixture returns `jurisdictionHints`, so it does not run. Finding 1 stays open. |
```

The better fix is to wire it the way you wired the permission one: tag a scenario, and stock a `record_search` fixture whose response carries `jurisdictionHints`. Then the row can say what it says now. Either way, please don't leave the claim standing while the guard is asleep.

Worth knowing: the harness already has language for this exact failure mode. `runnability.py` calls it passing "VACUOUSLY, green forever, asserting nothing" — but it only enforces the opposite direction, a test whose tags reach no validator. Nothing checks a validator whose tag reaches no test. That gap is probably worth a `nothing-checks` issue on its own.

---

### 3. The lever lint quietly stops checking a row if the wording changes

`lever-anchor-shapes.test.ts` decides whether a row applies by matching the prose for `clear q.surname` or `keep q.surname`. A row that matches neither is skipped with no signal.

I broke the `Drop surname` row's anchor two ways to show it:

- Break the anchor, leave the wording alone: the lint fails, correctly.
- Make the **same** break and change `Clear` to `Remove`: the suite goes from 8 tests to 7 and passes green, with the row still broken.

The existing drift guard doesn't catch it, because it counts rows it parsed, not rows it checked. Six of the fourteen rows in those two tables produce a check today.

```
Add this after "every Name-lever row was actually found":

it("the same number of rows still produce an anchor check", () => {
  const checked = ALL_ROWS.filter((r) => deriveShape(r).applies);
  expect(checked.length).toBe(6);
});
```

I ran it: passes on the current tree, and fails on the reworded break above.

---

### Optional — a latent hole in the new `index` marker

Not blocking, and it changes no test result today. I checked all 86 log notes from the nine `pre-1880-census-household` tests, and this changes none of them.

The comment says the rule deliberately won't accept a note that flags a **name** as indexed. It can. `Mother indexed as Sarah A. Price` satisfies it, because `mothers?` is in `_ROLE_WORD` and sits right next to `indexed`. Swap that one phrase to `Mother listed` and the same note fails, so it is the only thing carrying it.

If you want to close it:

```
You wrote:      r"(?:heads?|wife|wives|husbands?|sons?|daughters?|fathers?|mothers?|"
                r"parents?|relationships?|co-?residents?|roles?)"
Change it to:   r"(?:heads?|relationships?|co-?residents?|roles?)"
```

Do **not** add `household` to that set — I tried it first and it broke three of your own tests, including `test_indexed_spelling_alone_does_not_excuse_a_flat_role_assertion`. The version above keeps all 28 passing and changes no verdict on the corpus.

---

**Checks:** `vitest` fails (item 1). Everything else green on this head.
**Verified:** engine vitest 2733 pass / 1 fail on the merge; `eval/harness` pytest 2949 pass, 3 skipped. Your 27-of-28 result reproduces from the run logs.
**Not checked:** no eval run, so I can't say how the new universal parent-age validator behaves against other skills' suites — it isn't tag-gated, so it will run for all of them. Worth a thought before merge.

mercyokum:
Thanks for the thorough review — all four addressed in `a4da30e2`.

**1. CI-red line citation** — fixed exactly as you suggested: `runlog.py:250` → `eval/harness/harness/runlog.py`.

**2. `test_jurisdiction_hints_followed` was genuinely inert** — confirmed your finding (no tag, no fixture, vacuous pass). Wired it up for real: new scenario `jesse-neal-marriage-jurisdiction-hint`, four new `record_search` fixtures, and `ut_search_records_jurisdiction_hint` carrying the tag.

Running it caught a second, independent bug in the same pass: the model's actual compliant behavior on its very next call used `recordSubdivision` — the exact field `record-search.ts`'s own `searchedPlace` computation reads for a marriage search — which the validator's `place_fields` tuple didn't recognize, so a run that had genuinely followed the hint still failed. Fixed both; the test now passes end-to-end (7/7 dimensions score 3). Pinned regression test added covering the `recordSubdivision` case specifically.

**3. Lever-lint silent-skip gap** — added the row-count assertion you proposed. Didn't take your `6` on faith, though — verified it against the actual corpus first (`ALL_ROWS.filter(r => deriveShape(r).applies).length`), and it does match.

**4. `_ROLE_WORD` false-accept** — reproduced `Mother indexed as Sarah A. Price` against the current regex, confirmed it does false-accept, applied your narrowed set, confirmed the narrowed version closes it, and re-ran all 40 pinned search-records validator tests to confirm nothing else broke.

One more thing worth flagging since it came up while chasing #2: I initially thought the new test file would need a full paid re-run before the runlog gate would accept it, same as the earlier `whitfield` addition. Traced it down to `diff_snapshot_vs_disk` — it's a documented, one-directional check ("paths present on disk but absent from the snapshot are NOT flagged"), so a new test file never trips rule 2 regardless of who added it. Not a gap I'm proposing to close here, just noting it since it affected how I read the gate's result.

All pushed to `a4da30e2abe60f7f3996ebd24654bb31a9e2af14`. `vitest` and `runlogs` both green now.

promise-emmanuel (changes requested):
Reviewed the merge with main (branch is level with main, 0 behind), not just
the branch head.

Your four fixes from the last round all hold. I mutation-tested the lever lint
both ways — broke the anchor (red), then made the same break and reworded
`Clear` to `Omit` so the row drops out (`expected 5 to be 6`, red). That one is
properly closed. Engine vitest 2754/2754, harness pytest 2987 pass / 3 skipped,
apps/server 214 pass. 22/28 → 27/28 reproduces from the run logs.

Three things before this merges.

---

### 1. The jurisdiction guard passes on the failure it exists to catch

`_place_tokens` accepts a substring hit on any word longer than 2 characters.
The fixture's hint place is "Yell County, Arkansas". The failure being graded
is reverting to "Union County, South Carolina". Both contain "County".

I ran the real ground-truth sequence through the validator — nil on Union
County SC, then two more Union County SC searches, no Arkansas anywhere — and
it PASSES. Your pinned test misses it only because it uses "Yell, Arkansas"
instead of the fixture's own string.

You wrote:
    def _place_tokens(place):
        """Matchable words from a free-text place string (>2 chars, so "of"/"Co"
        don't create false matches)."""
        raw_tokens = re.split(r"[,\s]+", place or "")
        tokens = []
        for t in raw_tokens:
            if len(t) > 2:
                tokens.append(t)
        return tokens

Change it to:
    _GENERIC_PLACE_WORDS = frozenset({
        "county", "counties", "parish", "province", "state", "states",
        "district", "township", "borough", "city", "town", "united",
        "kingdom", "republic", "region", "department", "canton", "shire",
    })


    def _place_tokens(place):
        """Matchable words from a free-text place string. Drops words of 3
        characters or fewer ("of", "Co") and the generic administrative-unit
        words every place name shares: "Yell County, Arkansas" and "Union
        County, South Carolina" both contain "County", so keeping it lets a
        run that never left the searched jurisdiction satisfy the assertion."""
        tokens = []
        for t in re.split(r"[,\s]+", place or ""):
            if len(t) > 2 and t.lower() not in _GENERIC_PLACE_WORDS:
                tokens.append(t)
        return tokens

I applied that, re-ran the ground-truth sequence (now fails correctly) and
re-ran your 7 pinned tests (all 7 still pass).

Please also add a pinned test that uses the fixture's real strings — "Yell
County, Arkansas" as the hint, "Union County, South Carolina" on the next two
calls — so the pinned suite covers the case the fixture actually produces.

---

### 2. SKILL.md tells the skill to use a field no fixture stocks

Your new Step 4 rule names `recordCountry` first. The stocked Arkansas fixtures
are `recordSubdivision`, `residencePlace` and `marriagePlace` — there is no
`recordCountry` fixture and no `birthPlace` one. A run that does exactly what
the rule says gets a fixture_not_found. `recordCountry` is also the wrong field
for a county-level hint, which is why the run you observed reached for
`recordSubdivision`.

You wrote:
    carries a non-empty `jurisdictionHints`, the next 1-2 retries in this same
    plan-item search sequence must set `recordCountry` (or `residencePlace` /
    `birthPlace`) to the top-ranked hint's place before reverting to a

Change it to:
    carries a non-empty `jurisdictionHints`, the next 1-2 retries in this same
    plan-item search sequence must set `recordSubdivision` (or `residencePlace`
    / `marriagePlace`) to the top-ranked hint's place before reverting to a

And the assertion message, which still lists the old four fields:

You wrote:
    "but neither of the next 2 record_search calls set "
    "recordCountry/residencePlace/birthPlace/marriagePlace to it "

Change it to:
    "but neither of the next 2 record_search calls set "
    "recordSubdivision/residencePlace/marriagePlace/recordCountry/"
    "birthPlace/anyPlace to it "

---

### 3. The findings doc says two things are open that you fixed here

Last bullet under "Still open — not addressed this round":

You wrote:
    - **`census-field-availability.md:34`** ("1840 additionally names and ages
      Revolutionary War pensioners") and **`collection-quirks.md:16-20`**
      (S/L, F/T, n/u, U/V only — wants leading/middle wildcards for the
      Bipes/Bippes/Biepes/... case) — both from issue #1817, not touched.

Change it to: delete the bullet, and add both to the fixed list instead. The
1840 line is reworded in this diff and the Bipes leading/middle-wildcard bullet
is added to collection-quirks.md. The next auditor reads this doc to decide
what to pick up.

---

### Not blocking

The universal parent-age validator's pass condition is an uncertainty note in
`relationship.notes[]`, and no SKILL.md mentions relationship notes.
`tree_edit`'s schema describes the relationship object as parent/child/
person1/person2 only, and tree-edit/SKILL.md:114 tells the skill to keep audit
trails out of tree fields. So a run that trips it has no documented way to
comply. Nothing in the unit fixtures trips it today, so it is safe to ship —
but 24 parent-child pairs across our e2e trees and final trees would, 10 of
them mothers aged 45-53 on the Female>=45 bound alone. person-warnings.ts
treats that as a warning to review; here it is a hard failure. Consider
dropping that one bound, or naming the note in the skills that write
ParentChild edges.

Also: the nothing-checks issue for "a validator whose tag reaches no test" is
still unfiled, and this PR is the second instance of that exact shape.

Checks: all 12 required green on 1a7facbd.
Verified: engine vitest 2754/2754, tsc clean, harness pytest 2987 pass / 3
skipped, apps/server 214 pass, web+viewer-ui+schema typecheck clean.
apps/electron typecheck fails identically on origin/main (its script shells to
npm against the only-allow pnpm guard) — pre-existing, nothing here touches
apps/.
Not checked: no paid eval run, so I can't confirm ut_search_records_jurisdiction_hint
beyond your reported 7/7.

mercyokum:
Thanks for catching these — all three addressed in `87f4c0a1`.

**1. The jurisdiction guard passed on the exact failure it exists to catch.** Reproduced it first: fed the real jimmie-jewel-neal failure sequence (nil, then two more South-Carolina-scoped searches, no Arkansas anywhere) through the validator and confirmed it passed. `_place_tokens` was keeping "County" as a matchable token, and the hint place ("Yell County, Arkansas") shares that word with the reverted-to place ("Union County, South Carolina"). Applied your `_GENERIC_PLACE_WORDS` fix, re-ran the same sequence — now correctly fails — and added the pinned test you asked for, using the fixture's own real strings rather than the existing test's different (and non-colliding) ones.

**2. SKILL.md/assertion-message field mismatch.** Fixed both to `recordSubdivision`/`residencePlace`/`marriagePlace`, matching what's actually stocked and what's actually the right level for a within-country jurisdiction change.

**3. Findings doc's stale "still open" bullet.** Verified both reference-file fixes are genuinely on disk, moved the entry to "already resolved."

Re-verified end-to-end after merging in another round of main (which turned out to include a new `record_search` response-compaction step) — checked it doesn't touch `jurisdictionHints`, then confirmed empirically: `ut_search_records_jurisdiction_hint` still passes 7/7, all 41 pinned validator tests pass.

On your two "not blocking" notes:

- **Parent-age validator's compliance path** — dug into this further and it's not quite as bad as it looked: `relationship.notes[]` is fully specified in `simplified-gedcomx-spec.md` §4.2 for both ParentChild and Couple, and is already in `tree-shape.ts`'s allow-lists, so `tree_edit` already accepts it. The tree-edit SKILL.md line about "a fact has no notes field" is specifically about facts, not a blanket ban. The real gap is narrower than "no way to comply" — it's "no skill is ever told to use the field that already exists." Filed [#2050](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/2050) proposing it as a writer-tool precondition on `tree_edit`'s `add_relationship` (per ADR-0011) rather than prose scattered across the seven callers, with the reject-vs-auto-flag call and the existing-tree migration question left for the lead.
- **Unfiled nothing-checks issue** — found it's related to #1755's "Half C" (validator activation), but that check is scoped to `grade_on_invariant` negatives specifically and wouldn't have caught either of these two bugs (both hit a tag on a *positive* test). Commented on [#1755](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1755#issuecomment-5477256989) with this PR as concrete evidence rather than filing a competing issue, since the surface area overlaps.

Full suite re-run after both review rounds: 28 pass / 1 fail (`ut_search_records_013`, the documented pre-existing residual gap, not a regression) / 1 xfail, $9.97. Annotated and committed as `482731a5`. `runlogs`, `pytest`, `vitest` all green.

chesworthrm (changes requested):
Reviewed the merge with `main` (branch is level with main, 0 behind) at `db6fcfaa`, not just an older head.

Your four fixes from promise-emmanuel's round all hold, and the two items I raised last time that mattered most are properly closed: the jurisdiction guard now actually runs (the tag is on a real test and a real fixture returns `jurisdictionHints`), and the lever lint's count pin catches a reworded row dropping out. All 13 required checks green. Engine vitest 2755/2755, tsc clean, harness pytest 2988 pass / 3 skipped, apps/server 214 pass, web workspace 8/8.

Five things to fix, then one decision I can't make for you.

---

### 1. The new parent-age validator reads the wrong field and computes wrong years

The age bounds were carefully reused from `person-warnings.ts`. The year extraction was written fresh here, and it is wrong two ways.

It reads `date` and ignores `standard_date`, then takes the first four digits it finds anywhere in the string. In `eval/tests/e2e/concepcion-alegre-parents/starting-tree.gedcomx.json`, person `P73Z-2WC` has a birth `date` of `"06031947"`. Your reader returns **603** as the birth year. Against a parent born 1915 that is an age of **-1312**, which the "too young" branch flags as an implausible parent.

Second, a `date` is not always a string. In `eval/runlogs/e2e/mckee-birth-1904/run-2026-07-09_11-43-28.final-tree.gedcomx.json` one birth `date` is `{"original": "January–March 1904", "formal": "+1904"}`. That reaches `re.search` and raises `TypeError`. The runner turns a crash into a plain failure with the exception text as the message, so one odd date fails the test for a reason that has nothing to do with parent ages.

`test_init_project.py` already states the rule this misses: "standard_date is what every downstream date comparison reads."

In `eval/harness/validators/test_universal.py`:

```
You wrote:
                m = re.search(r"\d{4}", f.get("date") or "")
                if m:
                    return int(m.group(0)), gender

Change it to:
                raw = f.get("standard_date") or f.get("date")
                if not isinstance(raw, str):
                    continue
                m = re.search(r"\b(1[0-9]{3}|20[0-9]{2})\b", raw)
                if m:
                    return int(m.group(0)), gender
```

And the docstring three lines above it:

```
You wrote:      of preference, taking the first 4-digit year in the fact's `date`."""
Change it to:   of preference, taking the first 4-digit year in the fact's `standard_date`,
                falling back to `date`."""
```

I applied this and ran the full harness suite: 2988 pass, unchanged. Across all 424 trees in the repo it drops exactly the two bogus flags and keeps every real one.

---

### 2. The new index marker accepts the note shape its own comment says it rejects

The comment above it says the rule deliberately will not accept a note that flags a **name** or a **date** as indexed while still asserting the household role flat. It does accept exactly that. The window is 50 characters of anything-but-a-full-stop, so it reaches across a semicolon from an indexed name to a flat role in the next clause.

Both of these satisfy the marker today, and the new pattern is the only thing carrying them:

- `1860 census, Yell Co. Arkansas: surname indexed as Flyn; household head Thomas Flynn, wife Mary, son John.`
- `Match on Thomas Flynn. Birth year indexed as 1822; head of household Thomas, plus sons John and Amos.`

Narrowing `_ROLE_WORD` last round was the right move — the hole that is left is the window, not the word list.

In `eval/harness/validators/test_search_records.py`:

```
You wrote:
    rf"\bindex\w*\b[^.]{{0,50}}\b{_ROLE_WORD}\b",
    rf"\b{_ROLE_WORD}\b[^.]{{0,50}}\bindex\w*\b",

Change it to:
    rf"\bindex\w*\b[^.;,]{{0,20}}\b{_ROLE_WORD}\b",
    rf"\b{_ROLE_WORD}\b[^.;,]{{0,20}}\bindex\w*\b",
```

I checked this against all 299 distinct log notes in the five run logs this PR ships. Nothing that passes today starts failing, nothing that fails today starts passing, `Indexed as Head of Household: Thomas Flynn.` still passes, and the two notes above start failing correctly. Full harness suite stays at 2988.

---

### 3. Every figure in the re-measured docstring is stale

The docstring measures against "the 5 committed run logs as of this PR". This PR then swaps three of those five, so the measurement describes a corpus the branch no longer has.

I re-derived it by the method the docstring itself prescribes, and got a different answer for all three numbers plus the sentence after them. Same result whether or not I filter on `tool == "record_search"`.

```
You wrote:
    PR, the 9 `pre-1880-census-household`-tagged tests produced 44 logged
    searches. The marker appears in 26 of 44; outcome is 23 pass / 21 fail.
    Unlike the earlier figure, marker presence and pass no longer coincide —
    3 of the 26 marker-present runs still failed (`ut_search_records_014`
    once, `ut_search_records_h4k` twice), because this validator's own gate

Change it to:
    PR, the 9 `pre-1880-census-household`-tagged tests produced 43 logged
    searches. The marker appears in 32 of 43; per-run outcome is 28 pass /
    15 fail. Unlike the earlier figure, marker presence and pass no longer
    coincide — 5 of the 32 marker-present runs still failed
    (`ut_search_records_010` twice, `_014`, `_027` and `_h4k` once each),
    because this validator's own gate
```

Two notes on that. I added "per-run" because the old figure was ambiguous — its 23 plus 21 summed to its own search count, so it was unclear which unit was being tallied. Say which you mean. And nothing catches this: the validator file is not registered in `measured-figures.json`, and that guard reads no path under `eval/` at all, so this docstring drifts silently.

---

### 4. An xfail justification cites a run log this PR deletes

`whitfield-1850-household.json` rests the whole xfail case on a file the diff removes:

> 6 of 9 pre-1880-census-household-tagged tests failed this exact way in one full-suite run (`eval/runlogs/unit/search-records/v1_2026-08-27_11-50-40.json`)

That path is gone after this PR. Either keep that run log in the repo, or re-point the citation at one that survives. If you re-point it, the count needs redoing too — across the five surviving logs the tagged tests not passing run 4, 3, 5, 1, 2 of 9. Nothing shows 6. The highest is 5, in `v1_2026-08-28_15-20-05.json`. I checked the count there, not whether all 5 failed the same way, so verify that part before you write it down.

Same problem, smaller, in the findings doc: it cites a baseline range ending at `v1_2026-08-25_00-44-28`, which this PR also deletes.

---

### 5. The findings doc names a superseded release candidate

Commit `482731a5` added a newer run log than the one the doc's header names. The next auditor reads this line to decide what to pick up.

```
You wrote:      Issue #1642. Base `25b0f434`. Release candidate: `v1_2026-08-28_16-51-11` (28
                tests, 27 pass / 1 fail, annotated).

Change it to:   Issue #1642. Base `25b0f434`. Release candidate: `v1_2026-08-31_11-03-19` (30
                tests, 28 pass / 1 fail / 1 xfail, annotated).
```

Two things follow from the same fix. The PR body's "27 of 28 tests pass now" describes the older run as well — the run you are shipping is 28 pass, 1 fail, 1 xfail of 30. And while you are in the PR body: it says a third rewording of the pre-1880 inference rule was deliberately not attempted, but SKILL.md adds one ("This inference belongs in the same sentence as any needs-review reasoning..."), and the findings doc says "Deliberately not re-worded this round". Pick one story. If the rewording is intended, say so and say why this attempt differs from the two that did not move the number.

Smaller, and it is the thing promise-emmanuel actually asked for: the two issue #1817 reference gaps are now filed under "Already resolved **before** this round", while the bullet itself says "reworded elsewhere in **this diff**". Move that bullet into "Fixed this round".

---

### The decision I can't make for you: the `female >= 45` bound

promise-emmanuel raised this and it is still open. I measured it rather than take it on trust, and the honest number is smaller than the one in that review — but not zero.

There are 24 implausible parent-child pairs across the repo's trees. Your validator exempts any relationship whose id was already in the before-state, and 21 of the 24 are exempt on that rule. Three are not:

- `anders-monsen-ancestry` relationship `R2` — a mother aged 51, written fresh by the run, with no `notes` key at all. It is in two committed final trees, so this is not hypothetical; real runs already do this.
- `mccarley-spouse` `R23` and `R25` — mothers aged 46 and 47, stripped out of the starting tree, so a run that rediscovers them writes them new.

The problem is that a run has no documented way to comply. `notes` appears nowhere in `tree-edit.ts` or `tree-correct.ts`. The `relationship` input is an untyped object whose description names only the endpoints. The tree shape does allow `notes` on a ParentChild edge, so a model could pass it — but nothing tells it to, and `tree-edit/SKILL.md` says the opposite: "The audit trail belongs in your reply, not in tree fields."

`person-warnings.ts` treats a mother of 45+ as a warning to review. Here it is a hard failure. Two ways out, and it is your call which:

1. Drop `_PARENT_AGE_UPPER_FEMALE` for now and keep the three bounds a run can plausibly be held to, then restore it when issue #2050 gives the writer tools a real note field.
2. Keep it, and name relationship notes in the skills that write ParentChild edges — which means reconciling that with `tree-edit/SKILL.md`'s rule above.

Either is fine by me. What I don't want is the bound shipping while all three of those relationships would fail it with no way for a skill to do better.

---

**Checks:** all 13 required green on `db6fcfaa`.
**Verified:** engine vitest 2755/2755, tsc clean, harness pytest 2988 pass / 3 skipped, apps/server 214 pass, turbo test+typecheck 8/8. Mutation-tested the lever lint (broke the `Drop surname` anchor, watched it go red, restored). Re-derived the docstring figures and the 24-pair count from the committed corpus rather than reading them.
**Not checked:** no paid eval run, so I can only say the jurisdiction guard *can* fire, not that the model follows the hint. I did not reproduce the `apps/electron` typecheck failure you mentioned — `turbo test typecheck` passed 8/8 from my worktree.

mercyokum:
Thanks — all five fixes plus the decision addressed in `a0c68bdf`.

**1. Parent-age validator's year extraction** — reproduced both real bugs first: `P73Z-2WC`'s "06031947" misreading as year 603, and the mckee dict-shaped date raising `TypeError`. Applied your fix (prefer `standard_date`, require a string, bound the year regex), reverified both cases resolve correctly, and added two pinned regression tests from the real committed data.

**2. Index-marker window reaching across a semicolon** — reproduced both counter-examples against the current 50-char window, confirmed the narrowed `[^.;,]{0,20}` closes both while keeping `Indexed as Head of Household: Thomas Flynn.` passing. Two pinned regression tests added.

**3–5. Documentation/citation staleness** — re-derived what's mechanically verifiable rather than trust either prior figure (43 searches, marker in 32) and explicitly dropped the per-note judge-qualification split rather than guess a number for it. Whitfield's `xfail_reason` repointed to a run log that actually survives, with all 5 cited failures verified as genuinely the same validator (not a mixed count). Findings doc header updated to the current release candidate, the issue #1817 bullet moved into "Fixed this round" as you asked, and the "not re-worded this round" claim corrected against this PR's own commit history — a widening did happen, earlier in this same PR, and that's the "last SKILL.md widening" the docstring meant.

**The `Female≥45` bound decision** — dropped it. Discussed with the user: shipping a universal validator with a confirmed compliance gap (your 3 non-exempt relationships) felt worse than losing that one bound temporarily, especially since it's reversible and the real fix (issue #2050, filed last round) would likely replace whatever prose we scattered across seven skills anyway. Pinned a test documenting the gap explicitly, noted the decision on #2050 for traceability.

Then main moved twice more mid-review — once with a real, small fixture correction (`flynn-*` scenarios' ARK-URL format, from a parallel PR) that I initially misdiagnosed as a platform-specific hashing bug before tracing it to GitHub's synthetic PR-merge checkout. Merged both, rebuilt, reran the full suite twice more (28/1 fail/1 xfail of 30 each time, `_010` and `_013` trading places on the same known variance — verified neither is a regression). All 13 checks green on `f0f635c2`.

chesworthrm (changes requested):
Round 4. Reviewed the merge with `main` (branch is level, 0 behind) at `f0f635c2`, not just an older head.

The blocker from last round is closed. `runlogs` is green and all 13 checks pass — `d2f2d3f9` re-ran the suite after merging main, which is exactly the right order. I reproduced the gate locally ("All runlog rules satisfied"), and the 28 pass / 1 fail / 1 xfail of 30 reproduces exactly from the committed run log. Engine vitest 2767/2767, harness pytest 3026 pass / 3 skipped, tsc clean.

Two things to fix, both one-line edits. Then I'm happy.

---

### 1. Four references point at an issue that was declined

`#2050` was closed as **not planned** on 2026-08-31 at 12:26 — merged into `#1837`, which is open. Your commit landed at 15:03, so the merge happened while you were writing these lines. As it stands, whoever tries to restore the `Female>=45` bound follows the pointer, lands on a closed card, and never learns `#1837` exists.

In `eval/harness/validators/test_universal.py:1024`:

```
You wrote:      # paid re-run per skill touched, not something to scatter into this PR. Issue #2050
Change it to:   # paid re-run per skill touched, not something to scatter into this PR. Issue #1837
```

`eval/harness/validators/test_universal.py:1027`:

```
You wrote:      # have in hand. Restore this bound once #2050 lands.
Change it to:   # have in hand. Restore this bound once #1837 lands.
```

`eval/harness/validators/test_universal.py:1087`:

```
You wrote:          enforce yet (Female45 upper bound, dropped pending issue #2050).
Change it to:       enforce yet (Female45 upper bound, dropped pending issue #1837).
```

`eval/harness/tests/unit/test_parent_child_age_plausibility_validator.py:106`:

```
You wrote:      schema-accepted, but no skill is ever told to write it (issue #2050).
Change it to:   schema-accepted, but no skill is ever told to write it (issue #1837).
```

---

### 2. The whitfield xfail cites a run log this PR deletes — again

`d2f2d3f9` committed the v7 run log, retention pruned the oldest, and `v1_2026-08-27_13-03-15.json` went with it. The `xfail_reason` still points there for the judge-rationale claim, and nothing else in the repo preserves that rationale.

You called this exactly right in the findings doc — "the next full re-run will make it stale again." It did, on the very next push. Apply your own rule from that same note rather than repointing a third time:

In `eval/tests/unit/search-records/whitfield-1850-household.json:21`:

```
You wrote:      as low as it would score a genuinely bare, unhedged claim -- confirmed
                directly in v1_2026-08-27_13-03-15.json's judge rationale for this test.
Change it to:   as low as it would score a genuinely bare, unhedged claim -- see issue
                #1967, which owns this analysis.
```

I swept every run-log citation in the PR's non-runlog files against what actually survives. Six are dangling, and five of those were pruned by this PR's own run-log commits (v5, v6, v7). I'd leave the other five alone: they sit in the pinned validator tests and the findings doc, and the tests copy the cited note text verbatim into their constants, so the evidence is preserved and the filename is only a breadcrumb. Whitfield is the one where the citation *is* the evidence.

---

### Non-blocking

**The body's "up from 22 of 28" no longer reproduces.** I counted outcomes in all five search-records run logs on main at the merge base: 23/28, 26/29, 23/29, 24/29, and 25/29 plus one xpass. The nearest is 23 of 28, in `v1_2026-08-25_00-44-28` — which this PR deletes. Same rot you documented. Either restate it against a run that survives or drop the comparison; the "after" number is solid on its own.

**The new universal validator never actually asserted.** In the v7 run it skipped 30/30 ("no new ParentChild relationships") — search-records writes no tree relationships, so that is expected, but nothing in this PR exercises it. It won't get exercised elsewhere by accident either: the run-log snapshot is 192 entries of skill files, tests and fixtures, with no validator files in it, so editing `test_universal.py` marks no skill's run log stale. That gap is worth a `nothing-checks` issue against the gate rather than anything in this PR.

**Some SKILL.md and reference additions aren't accounted for in the PR body** — the `wiki_search` / `wiki_read` grant and the instruction to call them, the passenger-list content, "5+ variations is a floor", the given-name truncation order, and the surname-wildcard and misindexed-value guidance in the reference files. They all read like legitimate deep-dive output, but a reader can't tell which were findings and which were drive-bys. SKILL.md is +4.4 KB, billed on every run of the skill.

**Nothing will auto-close the deep-dive issue.** The title isn't parsed for closing keywords and the body has none — add `Closes #1642` to the body if that's the intent.

mercyokum:
Checked every claim in this one independently before acting — all of them held up, including the exact numbers. Nothing to push back on.

**1. The four `#2050` references** — confirmed it: closed `NOT_PLANNED` at 12:26, merged into `#1837` by `/merge-recent-issues`, approved by the lead, my commit landed at 15:03. All four repointed to `#1837`.

**2. Whitfield's second dangling citation** — confirmed `v1_2026-08-27_13-03-15.json` is genuinely gone (the *first* citation, to `v1_2026-08-28_15-20-05.json`, is still fine — good catch that you scoped this to only the one that's actually broken). Confirmed `#1967` genuinely covers the judge-rationale analysis. Repointed there, applying your own rule from the same note rather than a fourth filename.

**Non-blocking, all three:**

- **"up from 22 of 28"** — reproduced your exact counts against main's current five run logs, character for character (23/28, 26/29, 23/29, 24/29, 25/29+xpass). Dropped the comparison; the after number stands alone now.
- **Unattributed SKILL.md/reference additions** — added them to the PR body (wiki_search/wiki_read, passenger-list dates, given-name truncation order, the "5+ is a floor" note).
- **Nothing auto-closes #1642** — added "issue #1642 stays open" rather than a closing keyword. It's a tracking issue and there's real, unfiled work still under it (the parked givenName contradiction, `_013`'s residual gap), so closing it now would have been premature, not just an oversight to fix mechanically.
- **The universal-validator-never-exercised gap** — confirmed all three numbers (192 snapshot entries, zero validator paths, 30/30 skip in the committed run log). Different shape from Half C, so commented on #1755 with the evidence rather than a fourth issue.

Applying the two blocking fixes touched `whitfield-1850-household.json` again (the citation repoint), which re-tripped `runlogs` on the exact same behavior-neutral-edit shape the label exists for — the user applied `eval-cosmetic-skip` rather than buying a fourth full run over a citation fix. All 13 checks green on `71ea788e`.

chesworthrm (approved):
Round 5. Reviewed the merge with `main` (branch is level, 0 behind) at `80dcb944`.

Both blockers from last round are closed, and I checked each rather than taking the commit message for it: `#2050` is gone from all four sites in favour of `#1837`, and the whitfield `xfail_reason` now points at issue #1967 instead of a run log this PR's own retention deleted. All 13 checks green.

Approving. The three items below are worth folding in while the branch is open, but none of them blocks merge and none justifies a sixth round.

---

### First, something I got wrong in round 3

I told you the observed run reached for `recordSubdivision`. I took that from the comment above `place_fields`, which cites a run log retention has since pruned — I could not check it and should not have passed it on as observed. Both committed runs actually use `marriagePlace`, and both pass. Sorry for the noise; the field list below still needs a change, but for a narrower reason than I gave.

I also checked the thing I flagged in round 2 — whether that guard is really asserting rather than skipping. It is: `test_jurisdiction_hints_followed` records `error: None` in both committed runs, so it ran and asserted. That one is properly closed.

---

### 1. The judge is told a compliant run uses two fields that cannot produce one

`marriage-search-follows-jurisdiction-hint.json:28` names `recordCountry` and `birthPlace` as compliant. Neither has a stocked fixture. A call using either one re-matches the **nil** fixture and comes back empty — which contradicts the same sentence's "and that call returns a positive match." Meanwhile `recordSubdivision`, which your Step 4 rule names first and which does have a fixture, is missing from the list.

The judge prompt treats these notes as binding — "a requirement the response did not meet **is a deduction**." So a run that follows SKILL.md literally can be graded against a definition of compliance that does not include it.

```
You wrote:      sets a place field (recordCountry, residencePlace, birthPlace, or marriagePlace) to a value containing 'Arkansas' or 'Yell'
Change it to:   sets a place field (recordSubdivision, residencePlace, or marriagePlace) to a value containing 'Arkansas' or 'Yell'
```

Nothing is broken today — both runs took `marriagePlace`, which is on the list, and passed.

### 2. The new uncertainty list has the bug this PR fixes one file over

`test_universal.py:1037`. You changed `inferr` to `infer` in `_INFERENCE_MARKERS` for exactly the right reason: the double-r only catches "inferred" and "inferring". `_UNCERTAINTY_MARKERS` was written with the old pattern, so it misses "inference", "inferences", "infer" and "infers" — about a fifth of how the model actually writes this. Same one-word fix.

```
You wrote:      r"inferr",
Change it to:   r"infer",
```

This changes no result today: the validator only fires on a new ParentChild edge with an implausible age, and the three in the whole repo carry no note at all. It is a latent false-failure, not a live one.

### 3. Docstring off by one

`test_search_records.py:493`. `len(t) > 2` keeps a 3-letter token — `_place_tokens("Ely")` returns `["Ely"]`. Both examples in the sentence, "of" and "Co", are two letters, so they match the code and not the rule as stated.

```
You wrote:      """Matchable words from a free-text place string. Drops words of 3
Change it to:   """Matchable words from a free-text place string. Drops words of 2
```

---

### On the universal parent-age validator

I said in round 1 that I could not tell how this behaves outside search-records, and promise-emmanuel raised the same thing. I measured it: across all 424 trees in the repo, 3603 ParentChild relationships, the shipped bounds flag **3**, all in one historical e2e final tree, with implied mother ages of 6, 10 and 11. Those are real data defects, not false alarms. Zero crashes in the year reader. Restoring the `Female>=45` bound would take it to 22, which is the shape promise-emmanuel warned about — dropping it was the right call. Safe to ship.

---

**Checks:** all 13 required green on `80dcb944`.
**Verified:** engine vitest 2767/2767, `tsc` clean, harness pytest 3026 pass / 3 skipped, apps/server 214 pass / 1 skipped, web workspace 8/8 (viewer-ui 297/297). Mutation-tested the lever lint again both ways — broke the anchor (red, with the row and the validator's own message), then made the same break and reworded `Clear` to `Remove` so the row drops out (`expected 5 to be 6`, red). Restored; tree clean. The `eval-cosmetic-skip` bypass is correctly scoped — the bot stripped it when `80dcb944` landed and you re-applied it after.
**Not checked:** no paid eval run, so the 28 / 1 / 1 is reproduced from the committed run log rather than re-measured. The tree scan above reads starting and final trees, so it overstates exposure — the validator only fires on a before/after diff.

promise-emmanuel (changes requested):
Reviewed the merge with main at 532a140d (level with main, 0 behind).

First: nothing you pushed broke this. The approval read 80dcb944, and the only
things that landed after it are three commits from main — #2070, #2014, #2084 —
plus the merge. #2014 is what turned both checks red. It did not cause the bug
below; it exposed one that was already shipping.

Rounds 3–5 all hold, and I re-verified rather than reading the commit messages.
The date-parsing fixes work on both real cases: "06031947" now returns no year
instead of 603, and the dict-shaped date no longer raises. I re-ran my own tree
scan against the shipped bounds — 422 trees, 1 failure, and it is a genuine data
defect (mother aged 6 and 10). Dropping Female>=45 and filing #1837 was right.
Mutation-tested the lever lint again both ways; still fires correctly.

Two things block merge.

---

### 1. Step 4 tells the skill to make a call the tool rejects

record-search.ts throws "recordSubdivision requires recordCountry." I ran the
real validateInput on the four shapes the rule and the judge notes name:

    REJECT   recordSubdivision alone
    OK       recordCountry + recordSubdivision
    OK       marriagePlace
    OK       residencePlace

This one is on me. I told you last round to swap recordCountry for
recordSubdivision and did not check the pairing rule. Sorry.

It is also why vitest is red — mcp-fixture-queries.test.ts (new, from #2014)
runs every fixture's echoed query through the tool's own validator.

Worth knowing why the eval never caught it: the mock matches args by predicate
and never calls validateInput, so the invalid call succeeds there and would
throw in production. Both committed runs pass only because the model reached for
marriagePlace instead of doing what the rule says.

Three edits.

SKILL.md Step 4:

You wrote:
    plan-item search sequence must set `recordSubdivision` (or `residencePlace`
    / `marriagePlace`) to the top-ranked hint's place before reverting to a
    jurisdiction used prior to the hint.

Change it to:
    plan-item search sequence must set `marriagePlace` or `residencePlace` to
    the top-ranked hint's place before reverting to a jurisdiction used prior
    to the hint. `recordSubdivision` also works, but only with `recordCountry`
    set alongside it — the tool rejects it on its own.

record-search-neal-marriage-arkansas-recordsubdivision.json:

You wrote:
    "query": { "surname": "Neal", "recordType": "marriage",
               "recordSubdivision": "Yell County, Arkansas" }

Change it to:
    "query": { "surname": "Neal", "recordType": "marriage",
               "recordCountry": "United States",
               "recordSubdivision": "Arkansas" }

I applied that one locally and vitest goes green. The args predicate does not
need to change — fixture matching is subset-based, so a call that also carries
recordCountry still matches.

marriage-search-follows-jurisdiction-hint.json judge_context. This supersedes
the version in your round-5 approval, which named recordSubdivision alone:

You wrote:
    sets a place field (recordCountry, residencePlace, birthPlace, or
    marriagePlace) to a value containing 'Arkansas' or 'Yell'

Change it to:
    sets marriagePlace or residencePlace — or recordSubdivision together with
    recordCountry — to a value containing 'Arkansas' or 'Yell'

---

### 2. The cosmetic-skip claim is much wider than the one that was approved

At 80dcb944 the runlog gate saw 1 drifted file: the whitfield xfail_reason
wording. Behaviour-neutral, and the label was honest.

At 532a140d it sees 33. I checked each one: 32 came from main, not this branch —
#2014's fixture rewrite. That is not cosmetic. It changed what the model sees:

    - "results": [], "total": 0
    + "query": {...}, "totalMatches": 0, "paginationCappedAt": 4999,
    + "returned": 0, "offset": 0, "hasMore": false, "results": []

So re-applying the label now would assert behaviour-neutrality about someone
else's response-shape change, which is not what was approved. The SKILL.md fix
above changes behaviour anyway, so a re-run is needed either way.

Run it from this PR's worktree, not the primary checkout:

    uv run python eval/harness/run_tests.py --skill search-records

---

### Also worth folding in — your three approval items are still open

I checked all three against the current tree; none has landed.

test_universal.py:1037 —
You wrote:      r"inferr",
Change it to:   r"infer",

test_search_records.py:493 —
You wrote:      """Matchable words from a free-text place string. Drops words of 3
Change it to:   """Matchable words from a free-text place string. Drops words of 2

The judge_context one is folded into item 1 above.

---

One note on the approval already standing: it says it read 80dcb944, and the
head has moved since. Because pushes do not dismiss approvals here, it still
counts toward the required two even though the head is now red — so it needs a
re-look once these land, not just a merge.

Checks: vitest FAIL, runlogs FAIL on 532a140d. All 10 others green.
Verified: engine vitest 1 failed / 2774 passed (reproduced the CI failure
exactly), tsc clean, harness pytest 3026 pass / 3 skipped, apps/server 214 pass
/ 1 skipped, the four new pinned validator suites 53 pass / 1 skipped. Lever
lint mutation-tested both ways. Jurisdiction guard re-probed: the real
jimmie-jewel-neal failure sequence fails correctly, the compliant run passes.
Confirmed marriagePlace and residencePlace are advertised record_search
parameters, not just fields validateInput tolerates.
Not checked: no paid eval run — the 28 / 1 xfail / 1 fail is read from the
committed run log rather than re-measured.
Not filing: the vacuous-validator-tag gap from your round 2 — #1755 already
covers validator activation, so a second issue would duplicate it.

Prompt budget, not blocking: search-records/SKILL.md is 54.9 KB → 59.3 KB
(+4.4 KB), billed on every invocation. Justified by the result, worth knowing.

mercyokum:
Thanks for this round -- both blocking items and both minor items verified and fixed, plus one more thing your review surfaced that I closed while I was in there.

## 1. recordSubdivision requires recordCountry -- confirmed, fixed

Read `record-search.ts` directly: the throw and the schema description both confirm it exactly as you found it. Applied all three of your proposed edits verbatim:

- SKILL.md Step 4: `marriagePlace`/`residencePlace` as the primary levers, `recordSubdivision` only paired with `recordCountry`.
- The Neal-marriage fixture's `response.query`: added `recordCountry: "United States"`, corrected `recordSubdivision` to `"Arkansas"`.
- The jurisdiction-hint test's `judge_context`: matches your proposed wording.

vitest confirms: `mcp-fixture-queries.test.ts` is green.

One more thing while I was in there: the deterministic validator (`test_jurisdiction_hints_followed`) had the identical blind spot -- its `place_fields` tuple accepted a bare `recordSubdivision` as satisfying the rule regardless of whether `recordCountry` was present in that call. A run that made exactly the illegal call would have passed the validator even though production would reject it. Added a pairing guard plus a pinned negative test (verified it fails without the guard, passes with it).

## 2. Cosmetic-skip label scope -- confirmed exactly

Ran `diff_snapshot_vs_disk` against the committed run log at 532a140d myself: **33 files**, of which **32** are `eval/fixtures/mcp/*.json` last touched by main's #2014 (checked via `git log` on each), and **1** is the whitfield citation wording your round-4 approval covered. Matches your count and your attribution exactly. Agreed the label doesn't apply here and a re-run is the right call regardless, since the SKILL.md fix is behavioral.

## 3 & 4. The two folding-in items -- both confirmed, both fixed

- `test_universal.py:1037`: `r"inferr"` only matched the double-r family (inferred/inferring), missing `infer`/`inference`/`inferential`. Widened to `r"infer"` -- verified it's a strict superset (still matches "inferred") -- and pinned with a regression test using a note that says "inference" rather than "inferred". Confirmed the test fails without the fix, passes with it.
- `test_search_records.py:493`: docstring said "drops words of 3 characters or fewer" but `len(t) > 2` actually drops 2-or-fewer. Corrected the wording to match the code (comment-only, no behavior change, so no test needed here).

## Full paid re-run, as requested

Ran from this checkout (not the worktree): 30 tests, $9.67, 25 pass / 4 fail / 1 xpass.

The 4 fails (`_001`, `_012`, `_015`, `_017`) and the `h4k` xpass are all on `test_pre1880_census_structure_marked_inferred` -- confirmed via diff this is unrelated to anything in this round (nothing I touched affects that validator or its marker patterns). Checked its history across all 5 prior committed run logs: every single one has some failures on this same validator (1-5 tests each run, never zero). This is a pre-existing, chronic flake, already tracked in #2103 -- added today's run as a data point there (two more affected tests than the issue knew about, `_012` now failing 5 of its last 6 runs) rather than investigating here, since it's out of scope for a `recordSubdivision` pairing fix.

review_sample (`_002`, `_010`, `_028`, `h4k`, `_t9p`) annotated -- all dimensions confirmed matching the LLM score, all passes.

## Verified

- Full vitest: 2791 passed, 1 skipped.
- Full harness pytest: 3077 passed, 4 skipped.
- `npm run build` clean (a background git auto-sync merged unrelated main churn -- `research_append`/`tree-diff`/`tree-forget` work -- mid-session; rebuilt and re-ran both full suites again after that merge to be safe).
- CI: runlogs, vitest, pytest all green on the new push.

## Not changed

Nothing beyond what's listed above. The pre-1880 hedge flakiness is filed, not fixed, for the reasons stated.

promise-emmanuel (changes requested):
Round 7. Reviewed the merge with main at `4e1a7c6e` (level, 0 behind).

Your round-6 fixes all hold — I checked each rather than reading the commit
message. The three `recordSubdivision` edits landed verbatim, and the pairing
guard you added to the validator unprompted is the right call. Pinned suites
are at 55 pass / 1 skipped, up from 53/1.

Approving now so this isn't waiting on me. Two things still have to happen
before it can merge, and neither is a review finding.

**`runlogs` is red and needs a real re-run.** #2085 landed three
`search-records` run logs on main, and `v1_2026-09-01_18-35-01` now outranks
yours by timestamp — it was produced against main's tree, so 28 files differ.
I reproduced the gate locally and got the same 28. #2085 also rewrote three
passages of `search-records/SKILL.md`, which you pulled in at `4e1a7c6e`, so
your own run predates your current SKILL.md too. Please don't use
`eval-cosmetic-skip` for this one — the drift is your own SKILL.md rewrite,
not someone else's fixture churn.

    uv run python eval/harness/run_tests.py --skill search-records

From this PR's worktree, after merging main, so nothing lands ahead of you again.

**Restate two figures from whatever that run gives you.** The findings-doc
header (line 3) and the PR body both say 28 pass / 1 fail / 1 xfail — that's
`v1_2026-08-31_15-21-33`, two runs back. The log you currently ship is
`v1_2026-09-01_10-57-39`: 25 pass / 4 fail / 1 xpass, and `_010` — the test the
body names as "the one hard failure" — passes in it. You disclosed this in a
comment and logged it on #2103, which was right; the two documents just didn't
get the update, and the findings-doc header is what the next auditor reads.

Worth adding to the body while you're there: main's own three runs that day are
24 pass / 4 fail / 1 xfail of 29, so the branch at 25 of 30 is not a regression
— but the pre-1880 rewording didn't move that validator either. Your deferred
section already concedes that; this just makes the "after" number honest.

**Non-blocking:** `marriage-search-follows-jurisdiction-hint.json`'s
`test.description` still names the old field set you fixed in the same file's
`judge_context`. Only `judge_context` reaches the judge prompt, so it changes no
grade — the two halves of one file just disagree.

Since pushes don't dismiss approvals here, mine and chesworthrm's will both
still be standing when the re-run lands. If it comes back materially different
from 25/4/1, please re-request review rather than letting it ride.

Checks: `runlogs` FAIL, 12 others green on `4e1a7c6e`.
Verified: harness pytest 3083 pass / 3 skipped, the four pinned validator suites
55 pass / 1 skipped, runlog gate reproduced locally with the identical 28-file
list, outcome table re-derived from all eight committed run logs.
Not checked: no paid eval run. Two `test_mock_mcp.py` tests fail in my worktree
on a stale compiled build — CI's `pytest` is green on this head, and I couldn't
finish a rebuild locally to confirm directly.

mercyokum:
Thanks -- verified every claim in this round against real state before acting, and everything checked out exactly. Nothing to push back on.

## Confirmed and done

- **`runlogs` red / `v1_2026-09-01_18-35-01` outranking mine** -- confirmed via `git log` on all three new run logs, all trace to #2085 (`3cccc28d6`).
- **#2085's SKILL.md rewrite is real behavior, not cosmetic** -- read the diff myself: it removes the "plan doesn't change while searching" premise and the autonomous-mode ad-hoc-search freeze. Agreed `eval-cosmetic-skip` doesn't apply here (and it isn't -- labels are just `developer`).
- **Stale figures** -- confirmed both verbatim: PR body line 11 said "28 pass / 1 fail / 1 xfail," findings-doc line 3 cited `v1_2026-08-31_11-03-19`. Both now point at the fresh run.
- **Main's three runs at 24/4/1xfail of 29** -- confirmed exactly, all three files.
- **The 28-file diff** -- reproduced locally, same count.
- **The non-blocking `test.description`/`judge_context` drift** -- confirmed and fixed while I was in the file.

## Fresh re-run, from this checkout, after merging main

`v1_2026-09-02_10-04-24` ($10.08): **27 pass / 1 partial / 1 fail / 1 xpass of 30.**

- `_010` (fail): the same pre-1880 census-hedge validator tracked in #2103. Checked its history across every committed run of this skill -- ranges 1-6 failures, on main and this branch alike. Not caused by anything here.
- `_028` (partial, first occurrence): an isolated judge ding on Tool Arguments (a redundant `record_read` call), no validator failure. Confirmed via diff nothing this PR touches is anywhere near that code path.
- Compared against main's own baseline (24/4/1xfail of 29) run against the identical SKILL.md: this branch is not a regression.

review_sample (`_001`, `_003`, `_004`, `_012`, `_021`) annotated. Local gate run (`check_runlogs.py` with the real BASE/HEAD SHAs) confirms: `All runlog rules satisfied` -- the only warnings are the expected judge-prompt-hash mismatch (warn-only, main's judge/prompt.md moved after I generated the run) and the `_010` no-graded-dimensions note (it failed the deterministic validator before reaching the judge, as expected).

## Also merged in

A second round of unrelated main churn landed on the branch mid-session (#2131 convert-dates fix, a judge-rendering change) -- checked for overlap with search-records paths (none), merged, rebuilt, re-ran full vitest (2818 pass) and full harness pytest (3093 pass) again to be sure, then pushed.

## Verified

- Full vitest: 2818 passed, 1 skipped.
- Full harness pytest: 3093 passed, 4 skipped.
- `npm run build` clean.
- CI: all checks green on the new push (confirmed via `gh pr checks` after settling).

## Not changed

Nothing beyond what's listed. If the numbers move again before this merges I'll re-request review rather than let it ride, per your note on standing approvals.

promise-emmanuel (approved):
Round 8. Reviewed the merge with main at bb5bcf14 (2 behind, test-merged clean).

All three items from round 7 landed. I checked each against the tree rather
than the commit messages.

The re-run is real: v1_2026-09-02_10-04-24, 27 pass / 1 partial / 1 fail /
1 xpass of 30. I ran the gate locally with the real BASE/HEAD SHAs and got
"All runlog rules satisfied" — only the two warn-only notices you named.

Both figures now cite it, PR body and findings-doc line 3. I re-derived
main's baseline from all three of its 09-01 logs: 24 pass / 4 fail / 1 xfail
of 29, every one. So 27 of 30 here is an improvement, not just a
non-regression — fails are down from 4 to 1.

The fixture drift is fixed: test.description now names marriagePlace /
residencePlace and recordSubdivision-with-recordCountry, matching
judge_context and the validator.

_010's single failure is test_pre1880_census_structure_marked_inferred,
which is #2103. Read from the run log's own validator results, not the
summary line.

Two things I'm not asking you to fix here. Both cost a paid re-run and
neither is observed, so they belong on the issues that already own them.

1. _INFERENCE_MARKERS' r"infer" has no word boundary, so it also matches
"inferior" and "inferno". A note reading "Household head: Thomas Flynn.
Inferior image quality obscured the age column." satisfies the marker while
asserting the role flat. This came from my round-6 ask, so it's mine.
"inferior" is in zero committed run logs. When something else forces a
re-run of this validator:

   You wrote:      r"infer",
   Change it to:   r"\binfer(?!ior|n)",

Verified both ways — still matches infer/infers/inferred/inferring/
inference/inferential, no longer matches inferior or inferno. Same line in
test_universal.py's _UNCERTAINTY_MARKERS. I'll note it on #2103.

2. whitfield-1850-household.json's xfail_reason cites
v1_2026-08-28_15-20-05.json, and your own two new run logs pushed it out of
the 5-log retention window. Third time this class has come up here, and
repointing it at another filename just resets the clock — the durable fix is
to cite the issue that owns the data, the way you did in round 4. Editing
that file re-trips the gate, so it needs eval-cosmetic-skip; not worth a
round on its own. Noting it on #1967.

One merge note. chesworthrm's approval read 528c0e41, and 8750d4df,
17c8886e and cfbfdbda landed after it — the recordSubdivision pairing guard,
the Step 4 edits, both new run logs and the doc figure fixes. Pushes don't
dismiss approvals here, so it still counts toward the required two. I've now
read all of it myself, so my approval covers what theirs didn't.

Checks: all 13 green on bb5bcf14.
Verified: test-merged main (clean), then harness pytest 3117 pass / 3
skipped, engine vitest 2830 pass, tsc clean, the four pinned validator
suites 55 pass / 1 skipped, runlog gate satisfied locally. Mutation-tested
lever-anchor-shapes both ways — stripping the anchor from the "Replace name
with structural params" row fails that row's test, and rewording its "clear
principal name" cell fails the row-count guard rather than silently
skipping. Re-probed the round-5 _ROLE_WORD narrowing: "Mother indexed as
Sarah A. Price" no longer false-accepts. Confirmed no auth, schema, enum,
ownership-table or ADR path is touched, and no production engine source at
all.
Not checked: no paid eval run — 27/1/1/1 is read from the committed log. The
two test_mock_mcp.py failures I couldn't resolve last round were a stale
compiled build in my worktree; rebuilt and both pass.

PR #2027, opened 2026-08-28
feedback: route newspaper searches to the free archives (#1949)

### LINKED ISSUE #1949: [feedback] 2026-08-26T21:23Z
**Zip:** https://drive.google.com/file/d/1jL6C_VLlZuyRJ9uuUMYHpT0BfeGeCtqe/view?usp=drivesdk

```
make feedback-case ZIP=~/Downloads/feedback-2026-08-26T21-23-06-618Z.zip
```

- Submitted: 2026-08-26T21:23:06.618Z
- Platform: win32

Emruthwill:
Closes #1949.

## The report

> **What I asked:** Look for newspaper articles pertaining to Joseph and Frank Checketts.
>
> **What the agent did:** Could not obtain them. The automatic fetch was blocked on Utah Digital Newspapers and Chronicling America.
>
> **What it should have done:** I knew it wouldn't be able to pull from Newspapers.com but was expecting it to pull from Utah Digital Newspapers and Chronicling America.

## Root cause — three, not one

**1. Neither archive was ever tried.** All three newspaper plan items (`pli_002/003/008`) routed to Newspapers.com — the one paywalled site needing a manual capture — even though `pli_002` names the Ogden papers explicitly.

**2. Nothing could have fetched them.** `search-external-sites` supported five sites, all paywalled; no MCP tool covers newspaper archives; skills have no network in the VM.

**3. The obvious fix would not have worked either.** Probed from the host, which has full network access:

| Target | Result |
|---|---|
| `chroniclingamerica.loc.gov` → `www.loc.gov` | **403**, Cloudflare "Just a moment…" interstitial |
| `newspapers.lib.utah.edu` | **418**, Cloudflare JS challenge |
| Wikipedia API (control) | 200 |

Both sit behind Cloudflare bot management, with the repo's own `BROWSER_USER_AGENT`. This is **not** the VM egress proxy — a host-side MCP tool doing plain HTTP hits the identical wall.

So they ship as **free click-capture sites**, not an API integration. The user opens a pre-filled URL in their own browser, exactly as they already do for Newspapers.com, with no subscription needed.

## What changed

- **`external_site` enum** gains `chronicling_america` and `digital_newspaper_archive` (the bucket for state/regional archives — which one lives in `url_generated`, rather than a per-state enum value that would not scale). Both schema trees, the validator's hand-maintained copy, both prose specs. `enums.generated.ts` picks them up on its own.
- **`search-external-sites`**: both archives added with URL templates, plus the rule that matters most — **a blocked fetch is a capture-required result, not a negative one.** The skill now separates "paywalled" from "free but bot-protected", so the agent stops reporting a Cloudflare block as "no records exist".
- **`scripts/setup-feedback-case.sh`**: `unzip` exits 1 on the backslash-separator warning every `win32` submission produces, and `set -e` killed the script *after* extraction but *before* the marker file, git baseline and skill symlinks — silently. Every Windows submission hit this. The `.bat` path was never affected.

The framing this changes: `external_site` no longer means "paywalled", it means **"the agent could not fetch this directly"**.

## Verification

**The regression test is proven to discriminate**, not merely to pass. `ut_search_external_sites_014`, mined from this case, run against both skill bodies:

| | Outcome | Failing dimensions |
|---|---|---|
| Pre-change | **partial** | Correctness 2, Completeness 2, URL generation 2, Log entry 2 |
| Post-change | **pass** | none — all 8 at 3 |

The pre-change judge is worth reading: *"failed to generate any Chronicling America URL at all, only mentioning it as an alternative in parentheses"*. The old agent knew the archive existed and had no route to act on it — the reported bug in miniature. `_014` **passes in the shipping run**, now grading against the corrected loc.gov template, so it is also the evidence the URL fix works.

**Shipping run log: `v1_2026-09-02_11-45-07` — 16 tests, 14 pass / 1 partial / 1 fail.** Annotated, zero human-vs-judge disagreements across all 35 sampled dimensions. Rule 2 active on all 51 snapshot files.

### `ut_search_external_sites_002` — fail. Pre-existing, and provably not this PR.

Correctness 1, URL generation 1, Log entry 2; validators clean. The skill encoded `birth_place=Pennsylvania` when `research.json` holds a resolved conflict (`c_001`, `preferred_assertion_id: a_002`) preferring Ireland — the `conflicts[]` rule from #1954.

This deserves more than an assertion, since URL generation is the dimension this PR exists to move. **The identical failure is in main's own run log**, `v1_2026-08-28_11-43-01`, committed on `origin/main`, whose snapshot does not contain this branch's mined test — it predates this branch's content entirely. Same dimensions, same rationale, same `c_001`/Ireland reasoning.

Its history across the retained logs, with each run's `SKILL.md` hash:

| Run | SKILL.md | `_002` | Failing dimensions |
|---|---|---|---|
| 2026-08-28_06-22 (main) | `307d1efec6de` | pass | — |
| 2026-08-28_11-43 (main) | `e979b2ef76a4` | **fail** | Correctness 1, URL generation 1 |
| 2026-08-28_18-28 | `8e3f95f7d4e9` | pass | — |
| 2026-08-29_20-04 | `a2ace6aaf551` | **fail** | Correctness 1, URL generation 1 |
| 2026-09-02_11-45 | `846e232cd360` | **fail** | Correctness 1, URL generation 1, Log entry 2 |

It alternates, and the first failure is main's, before this branch's URL change existed. One caveat stated plainly: **every run carries a different `SKILL.md` hash**, so no same-code pair exists in the corpus and these logs cannot separate code from variance on their own. What they do establish is that the failure occurs with none of this PR's changes present.

Reading it as a positive signal for #1954 is fair: moving the answer into the URL-generation dimension so it grades all 15 tests instead of 4 is exactly why it surfaces now. The rule is applied unreliably and the rubric sees it.

### `ut_search_external_sites_005` — partial. A judge call outside the rubric's stated band.

Tool Arguments 2, Tool selection 2; validators clean. The judge's own rationale says the work was correct — *"all resolved correctly to Pennsylvania, United States, matching expected_args"*, *"consumed the response correctly"*, *"The final result is correct"* — and marks it down solely for making three `place_search` calls to resolve one place: *"the path shows unnecessary redundancy."*

The rubric's `partial` band for Tool selection names three slips: fabricating a `standardPlace` the tool did not return, presenting a curated link for the wrong record type, or failing to dedupe. Retrying a call that succeeded is not among them, and Tool Arguments is a base dimension with no band in this skill's rubric at all. So this is the judge applying an efficiency criterion the rubric does not state, not a defect this PR introduced.

Also merged `origin/main` repeatedly, including Edmond's `search-external-sites` deep dive (#1954) — the changes are complementary, and main did not add newspaper archives.

## Not covered, deliberately

**Two things are unverified and stated as such in the commits**, because Cloudflare blocked me from checking: whether UDN actually holds the *Ogden Standard* and *Daily Utah State Journal* for 1906, and the exact facet parameters these archives accept. The skill therefore tells the agent **not to invent facet parameters** — an unrecognised one is silently ignored and the user lands on a dead end believing the search was scoped. Ten minutes in a browser would close both.

**A structural finding, filed here rather than acted on:** 27 of the 37 tests across `search-external-sites` and `research-plan` lack a catch-all fixture for at least one of `place_search`, `external_links_search`, `collections_search`, `volume_search`. Every `fixture_not_found` scores Tool Arguments 1, so this is a standing flakiness source that has been read as regressions more than once in this branch alone. Blanket-adding catch-alls would weaken tests whose narrow fixtures deliberately catch tool misuse — a cross-cutting eval call, not a feedback fix.

**The planning half is on `research-plan-free-newspaper-routing`** with its own eval slot: routing `newspaper` plan items to a free archive before a paid one, plus the catch-all fixture fixes. Split out to keep this PR to one skill's eval slot.


EdmondOware:
Strong work, and the rule that matters most is the one you led with: a blocked fetch is a capture required result, not a negative one. A Cloudflare challenge logged as `negative` is false absence evidence in the GPS trail, and routing newspapers to the free archive first is right on research economics too. The enum change also hits every site it needs to, including correctly not hand editing the TS union.

First, `ut_search_external_sites_001` is my defect, not yours. In #1954 I gave it `collections-search-schuylkill` because that run resolved to "Schuylkill, Pennsylvania, United States". I pinned a fixture to one observed resolution, and place resolution moves between runs. Your `place-search-pennsylvania` addition is correct and it fixed the place call; the collections call still misses for the same underlying reason. When you next touch it, the durable fix is `collections-search-any`, the catch all ordered last, rather than adding a second pinned value. Deferring it was the right call given the run you had just paid for.

One thing I would want settled before this merges, and it needs a browser rather than a probe. The Chronicling America template points at `chroniclingamerica.loc.gov`, and your own root cause table records that host returning 403 and redirecting to `www.loc.gov`. LOC moved Chronicling America under `www.loc.gov/collections/chronicling-america/`. The parameters are real LOC search parameters, so that is not the question. The question is whether they survive the redirect and land on a scoped search, or whether the user arrives at a bare collection page believing their search ran. The entire deliverable here is a URL a human opens, and no one has opened it.

Smaller: `Coverage is 1690–1963` is the range of the US Newspaper Directory, which is catalogue records. The template targets the digitised page search, whose corpus starts much later. Worth quoting the page range instead, since an agent told 1690 will run page searches for eras with no pages and log the nils.

Your restraint on the state archives is the right instinct and worth keeping: "do not invent facet or date parameters" is exactly the correct response to not being able to verify a site. The same caution belongs on the coverage figure.

EdmondOware (changes requested):
The blocked fetch rule is the right call and the most valuable thing here: a Cloudflare challenge logged as `negative` is false absence evidence in the GPS trail, and separating "paywalled" from "free but bot protected" is what stops that. The enum change also lands on every site it needs to, including correctly leaving the generated TS union alone.

Two things to do before this merges. Reasoning is in my comment above.

1. **Open the Chronicling America URL in a browser and record what came back.** The template targets `chroniclingamerica.loc.gov`, which your own root cause table shows returning 403 and redirecting to `www.loc.gov`. Use a real one, for example:

   `https://chroniclingamerica.loc.gov/search/pages/results/?andtext=Joseph+Checketts&date1=1880&date2=1920&dateFilterType=yearRange&state=Utah&rows=20&searchType=basic`

   If it lands on a scoped search with the name, the year window and the state applied, say so in the PR and this closes. If the parameters are dropped by the redirect, the template needs to move to the `www.loc.gov/collections/chronicling-america/` path, because every URL the skill generates for this site would otherwise be unscoped and every nil it logs would mean nothing.

2. **Correct the coverage figure.** `Coverage is 1690–1963` is the range of the US Newspaper Directory, which is catalogue records. The template targets the digitised page search, whose corpus starts much later. Quote the page range instead, so an agent is not sent to run page searches for eras with no pages.

Not an action for you: `ut_search_external_sites_001` is my defect from #1954, and deferring it was correct given the run you had already paid for.

EdmondOware (approved):
Approving. Both items are done, and the first better than I asked.

You did not just move the template, you found the mechanism: the old host 308 redirects and carries the query string, and the new endpoint ignores those parameters, so the user lands on something that looks like a search. The `dl=page` discovery is the more valuable half. Without it the endpoint answers from the Directory, so a nil is a nil about catalogue records rather than about pages, which is a wrong answer wearing a right one's clothes. Updating the summary table row and pinning the legacy form as unacceptable in the test's judge_context were both beyond the ask and are what stop this regressing.

The remaining red is not yours. `ut_search_external_sites_002` encodes the birthplace my #1954 conflict rule rejects, and it has done so in three of the five committed runs. I checked that this PR touches neither that test, its fixtures, nor the scenario. That is variance my own dive left open.

EdmondOware:
This has been sitting since 29 August and is now 146 commits behind main, with the merge marked dirty. Three conflicts, and two of them are not real.

The two `.ann.json` ones are git rename detection misfiring across skills, because annotation files all look alike to it:

```
search-external-sites/v1_2026-08-20_22-45-06.ann.json
   main   -> search-records/v1_2026-09-01_13-24-52.ann.json
   pr2027 -> search-external-sites/v1_2026-08-29_20-04-37.ann.json
```

Those are not the same file being moved two ways. Main pruned an annotation and added one for a different skill, your branch pruned and added its own, and the retention policy did both independently. Keep both sides.

The third is real: a content conflict in `packages/engine/plugin/skills/search-external-sites/SKILL.md`, where main has edited the same skill body you are editing.

Worth more than clearing the markers. That skill has moved a fair amount since the 29th, so the question is whether your newspaper archive routing still says the right thing sitting on top of what landed, rather than whether the two texts can be spliced. If the answer changes what you wrote, that is worth a line in the body.

Once it is current it needs one more approver. Mine is already on the current head and carries over.

EdmondOware (changes requested):
The `pli_007` diagnosis is the best thing in this round. Hitting a red after a large merge and tracing it rather than running it again is the harder path, and what you found is a real gap rather than a flake: step 4 told the agent to mark a plan item skipped without ever saying what to do when there is no item, so it created one purely to have something to mark. Drawing the line at `research-plan` owns item structure, an executing skill may only update the status of an item that already exists, is the right place to draw it. Your note that it passed on 29 August only because the agent happened not to invent an item that run is exactly what a prose gap looks like from outside.

One thing to fix.

1. The verification section still describes `v1_2026-08-28_18-28-40` at 15 pass and 1 fail with zero regressions, and explains `ut_search_external_sites_001` as the single failure. That run is no longer the one that ships. The committed run is now `v1_2026-09-02_11-45-07` at 14 pass, 1 partial and 1 fail, the body does not mention it anywhere, and neither of its two results that are not green is the test the body accounts for. `ut_search_external_sites_002` fails on Correctness 1, URL generation 1 and Log entry 2. `ut_search_external_sites_005` is partial on Tool Arguments 2 and Tool selection 2. Both are judge calls with the validators clean.

   `_002` is the one to speak to directly, because URL generation is the dimension this PR exists to move and the pre change evidence you quote is a judge complaint on that same dimension. If it is variance, its history across the five retained run logs is free to read and settles it. If it is the change, that is worth knowing before this lands. Either way the section should describe the run that ships.

Emruthwill:
Fixed. The Verification section now describes `v1_2026-09-02_11-45-07` and accounts for both of its non-green results.

**`_002` — it is not the change, and the corpus proves it rather than suggesting it.**

The identical failure is in **main's own run log**, `v1_2026-08-28_11-43-01`, committed on `origin/main`, whose snapshot does not contain this branch's mined test — it predates this branch's content entirely. Same two dimensions, same rationale, same `c_001`/Ireland reasoning, same `birth_place=Pennsylvania`.

| Run | SKILL.md | `_002` | Failing dimensions |
|---|---|---|---|
| 2026-08-28_06-22 (main) | `307d1efec6de` | pass | — |
| 2026-08-28_11-43 (main) | `e979b2ef76a4` | **fail** | Correctness 1, URL generation 1 |
| 2026-08-28_18-28 | `8e3f95f7d4e9` | pass | — |
| 2026-08-29_20-04 | `a2ace6aaf551` | **fail** | Correctness 1, URL generation 1 |
| 2026-09-02_11-45 | `846e232cd360` | **fail** | Correctness 1, URL generation 1, Log entry 2 |

It alternates, and the first failure is main's, before this branch's URL change existed.

The caveat, stated rather than buried: **every run carries a different `SKILL.md` hash**, so there is no same-code pair anywhere in the corpus and these logs cannot separate code from variance on their own. What they establish is the narrower claim — the failure occurs with none of this PR's changes present. I had a 3× repeat of `_002` queued to settle the variance question directly and stopped it once main's log made it unnecessary; say the word if you want it run anyway.

Your framing looks right to me: this is #1954 working. Moving the answer into the URL-generation dimension so it grades all 15 tests instead of 4 is exactly why it surfaces now, and it is catching a real miss each run rather than a spurious one.

**`_005` — worth a second look, because the judge marked down work it described as correct.**

Its own rationale: *"all resolved correctly to Pennsylvania, United States, matching expected_args"*, *"consumed the response correctly"*, *"The final result is correct"*. The 2s are entirely for making three `place_search` calls to resolve one place — *"the path shows unnecessary redundancy."*

The rubric's `partial` band for Tool selection names three slips: fabricating a `standardPlace` the tool did not return, presenting a curated link for the wrong record type, or failing to dedupe. Retrying a call that succeeded is not among them, and Tool Arguments is a base dimension with no band in this skill's rubric at all. So this reads as the judge applying an efficiency criterion the rubric does not state. Flagging rather than acting on it — it looks like `/audit-rubric` territory, not a change this PR should make.

**Also fixed:** the body pointed at `1949-followup-research-plan-newspaper-routing`, a duplicate branch I deleted. Corrected to `research-plan-free-newspaper-routing`, which is the one that exists.

Merged main again (Rule 2 still active on all 51 snapshot files).

**Still open and not closable by me:** opening one generated Chronicling America URL in a browser to confirm the parameters take effect. Cloudflare 403s both curl and WebFetch across the whole loc.gov domain, and no test in the repo covers it — the eval grades the shape the agent builds, not whether that shape works live. It is the only claim in this change resting on documentation rather than observation.

EdmondOware (approved):
Approved.

The `_002` answer is the one I asked for and better than asserting variance. Main's own `v1_2026-08-28_11-43-01` carries the identical failure, same two dimensions and same `c_001` reasoning, and its snapshot does not contain `free-newspaper-archive-unnamed-site.json`, so that run happened without this branch's mined test present at all. That establishes the claim without needing the repeat run.

Stopping the queued 3x repeat once main's log made it unnecessary was the right call, and so was stating the caveat rather than letting the table imply more than it shows. Every run in the corpus carries a different `SKILL.md` hash, there is no same-code pair anywhere, and you say so and then claim only what the logs support. That is the difference between evidence and a table.

Reading `_002` as a signal for #1954 rather than against it is also fair. Moving the answer into URL generation so it grades fifteen tests instead of four is exactly why an unreliably applied rule starts showing up.

Your `_005` reading holds against the rubric. The partial band for Tool selection names three slips, a guessed `standardPlace`, a curated link presented for the wrong record type, and a failure to dedupe. Retrying a call that already succeeded is not one of them, and Tool Arguments carries no band in this skill's rubric at all. So that is the judge grading an efficiency criterion nobody wrote down, which is a rubric matter and not something to fix inside this PR. It is also currently tracked nowhere, so it is worth a card before it grades something else the same way.

The `pli_007` fix from the previous round stands on its own merits. An executing skill may only update the status of an item that already exists, and inventing one to close puts a search in the plan that was never planned.

mercyokum (changes requested):
**Requesting changes**

## Most severe

### 1. `unzip` exit 1 can also mean a file was silently skipped — not just the harmless warning

Info-ZIP's own docs describe exit code 1 as covering *"zipfiles where one or more files was skipped due to unsupported compression method or encryption with an unknown password"* — the same code as the harmless backslash-path warning this fix targets. The fix treats any exit code under 2 as safe, but that doesn't distinguish the two cases. Nothing after `unzip` checks that the extracted case directory actually contains `research.json` or `tree.gedcomx.json`. A corrupted or partially-skipped zip member would return exit 1, pass the check, and the script would go on to git-commit and wire up skill symlinks over an incomplete case directory with no error at all.

### 2. The Chronicling America URL template can't express a multi-year window

`start_date` and `end_date` both reuse the same `{yyyy}` placeholder, so this can only ever encode a single calendar year. Plan items aren't always single-year — `research-schema-spec.md`'s own `date_range` example is `"1870-1890"`. As written, a plan item with a real multi-year window has no way to be represented in this URL.

### 3. `ut_search_external_sites_014` is missing a fixture that Step 2b can call for

Step 2b says every external-site result must be checked against `collections_search` before being presented — worded as required, not conditional. The new test's fixture list has no `collections-search-schuylkill`, while the sibling test for the same place does. In the committed run the agent happened not to call it, so the gap didn't surface — but nothing guarantees that on a different run.

## Correctness/coverage

### 4. No fallback when a target date falls outside 1798–1963

The rule says not to run page searches outside that window, with no instruction for what to do instead when a plan item's date falls outside it.

## Conventions

### 5 & 6. Two new sentences explain *why* a rule exists rather than just stating it

One justifies the free-archive-first rule by first ruling out a different reason; the other narrates the retirement date and redirect mechanics of a deprecated API. Both are rationale rather than instruction — CLAUDE.md is explicit that this belongs in a spec or the rubric, not the SKILL.md body.

mercyokum (approved):

Issue #2032, opened 2026-08-29
research-plan omits the post-1850 census a parentage plan needs — missed in 2 of 5 committed runs


Ikennaya1:
`ut_research_plan_005` requires at least one post-1850 census in a parentage plan. The skill has now missed it in **2 of the 5 committed run logs**, scoring `Completeness` 2 both times for the same omission.

## The requirement

`eval/tests/unit/research-plan/locality-survey-first-plan.json`, `judge_context[4]`:

> At least one item should target a post-1850 census (1870, 1880, or 1900) to track Patrick into adulthood; **the 1880 census relationship column is the best direct parentage indicator in the available fixtures.**

That is sound genealogy: 1880 is the first US federal census to state each person's relationship to the head of household, so it is direct rather than inferred evidence of parentage.

## The misses

| run | outcome | census items planned |
|---|---|---|
| `v1_2026-08-24_14-32-36` | partial | 1850 + 1860 only |
| `v1_2026-08-28_19-50-30` | partial | **1860 only** (seq 7) |

Both scored `Completeness` 2 citing this bullet by name. The three runs in between passed, so it is intermittent rather than constant — but it is the same omission each time it appears, not a wandering failure.

The most recent plan is otherwise strong: eight items, correct fallback chains, defensible sequencing, jurisdictions right. It reaches for the death certificate, probate, baptism, marriage and siblings' certificates — and then stops the census coverage at 1860, thirty-five years before the subject's death.

## Why it is worth its own card

The gap is not that the skill dislikes censuses; it planned one. It is that nothing makes it ask **which** census years the question needs. A parentage question about someone born ~1845 and dying in 1908 has an obvious adult-life window that 1860 does not cover.

That is checkable the same mechanical way as #2030: the question supplies a birth estimate and a death date, and every plan item carries a `date_range`. Whether it should be a doctrine rule, a rubric bar, or a deterministic validator is the open question.

## Why it is not fixed in the originating PR

PR #TBD's central finding is a measurement: a new sequencing rule was added to `SKILL.md` and three runs showed it did not change behaviour. Adding a second doctrine change in the same PR would confound that measurement — with two new rules in the body, neither result could be attributed. It also needs its own paid run to verify, which is a second eval slot for the same skill.

**Touches:** `packages/engine/plugin/skills/research-plan/SKILL.md`, `eval/tests/unit/research-plan/locality-survey-first-plan.json`, `eval/harness/validators/test_research_plan.py`

[...]

Issue #2031, opened 2026-08-29
project_context reports a question 'concluded' from a proof summary alone, contradicting its own status and live plan items


Ikennaya1:
> **Reviewed 2026-09-11 before junior handoff.** The lead's 2026-08-31 ruling stands and is the design; no new decision is needed. But the ask hides a paid eval run and its cited evidence has moved — both corrected below. The original body follows under "Original issue".

**Touches:** `packages/engine/mcp-server/src/utils/question-state.ts`, `packages/engine/mcp-server/src/tools/project-context.ts` (the return type *and* the `description` string), `packages/engine/mcp-server/tests/utils/question-state.test.ts`, `packages/engine/mcp-server/tests/tools/project-context.test.ts`, `docs/specs/project-context-tool-spec.md` (§2.1 return-value block **and** §2.2), and — in a separate PR — `eval/tests/unit/research-plan/plan-for-parentage-question.json`

### Ship this as two PRs — the second one costs a paid eval run

**PR 1 (engine, startable now, no eval cost).** `question-state.ts`, `project-context.ts`, the spec, and the two vitest files. Touches nothing under `eval/tests/` or `packages/engine/plugin/`, so no run log goes stale and `check_runlogs.py` stays quiet.

**PR 2 (eval).** The `judge_context` note on `ut_research_plan_001` (`eval/tests/unit/research-plan/plan-for-parentage-question.json`). This is the *expensive* half, not the cheap one: `eval/tests/unit/<skill>/**` is embedded in the skill's run-log snapshot and `judge_context` is not one of the three cosmetic fields stripped from it (`eval/harness/harness/snapshot.py:48`). Editing it marks `research-plan` touched, and `check_runlogs.py` Rule 2 **blocks the PR** until a fresh `make eval-skill SKILL=research-plan` run plus a genealogist `.ann.json` lands — roughly $8–12 and 45–65 minutes of machine time, plus human annotation.

**Batch PR 2 with issue #2023 / PR #2434**, which already edits `packages/engine/plugin/skills/research-plan/SKILL.md` and therefore already owes that same re-run. #2023's "Red F" is this exact test and says: *"Do not fix the reading in prose here without checking whether #2031 removes the ambiguity at the source first."* Landing PR 1 first is what lets #2023 answer that.

[...]

Issue #2030, opened 2026-08-29
research-plan: make `the objective's target record leads` a research_append precondition — prose does not bind (3 measured runs)


Ikennaya1:
`research-plan` sequences the record the objective asks for behind work that re-derives the objective's own inputs. `SKILL.md` step 4 item 7 now states the rule; **three measured observations say the prose does not bind.** This asks for the writer-tool precondition that would.

## The measurement

From alpha feedback #1945 (PR #TBD). The objective named the record wanted (a birth or baptism record), the place it happened (Kongsberg), the approximate year (~1780), and the death date and place.

| run | rule 7 in the body | seq 1-2 | seq 3 |
|---|---|---|---|
| `scratch_2026-08-28_16-54-42` | no | Trysil burial + its browse fallback | Kongsberg baptism |
| `scratch_2026-08-28_18-16-58` | **yes** | Trysil burial + its browse fallback | Kongsberg baptism |
| `v1_2026-08-28_19-50-30` (`ut_research_plan_r3d`) | **yes** | Trysil burial + its browse fallback | Kongsberg baptism |

Identical ordering all three times. The skill varied its `record_type` spellings between runs (`church`, then `vital_record`, then `church` again for the Trysil items), so it was reasoning freshly each time and still landed on death-record-first.

In the submitted session it was worse: the baptism sat at **sequence 6 of 11**, the agent ran item 1 only, and spent 3h18m in Trysil burial images without ever searching Kongsberg.

## Why prose is not the fix here

This is the same shape #1616 measured on `check-warnings`, where a rule stated on five surfaces — `SKILL.md` twice, `rubric.md`, a per-test `judge_context`, and a judge prompt that calls such notes binding — held on none of them.

The rubric cannot enforce it either: the 2026-08-24 dive found `Sequencing logic` scored `3` on all 65 runs it audited and called it **"worse than no"** (it once cited a fabricated collection id as its evidence), and `rubric.md` is fenced by #1404/#1668.

## Decision (lead, 2026-08-31) — ship it as a `research_append` precondition

A hard precondition on plan writes, per ADR-0011's first question: this **is**
decidable by reading the project documents alone — `project.objective` names a
record type and a place, and each plan item carries `record_type`,
`jurisdiction`, `sequence`, `fallback_for`.

**The rule.** When the objective names a target record type *and* supplies the
jurisdiction to search it in, reject a plan whose first item targets neither, and
[...]

Issue #2023, opened 2026-08-28
research-plan, one paid eval run: five rules (dependency gating, backward building, census years, section ownership, holdings before aggregators) plus the three unowned reds from the 2026-09-07 suite


benter-070:
> **Reviewed 2026-09-08 before handoff. `senior` — read this block before planning the run.**

**ADR-0012 binds Rules B and E, and the card does not cite it.** ADR-0012 (Accepted;
"Read before you… add a census-year… fact to a prompt") rules that *"'Pre-1880 census'
is a fact about one country's schedule and is written as such or fetched"* and that *"a
jurisdiction-shaped fact may not"* stay in the plugin. Rule B's requirement (post-1850
census; the 1880 relationship column) and Rule E's acceptance item 1 ("name the census
case explicitly… consecutive census years are the cheapest identity chain available in
US research") are both jurisdiction-shaped. **`research-plan` holds no wiki tool** —
`allowed-tools` is `collections_search`, `volume_search`, `external_links_search`,
`place_search`, `place_search_all`, `research_append` (`SKILL.md:14-21`), and
`locality-survey-first-plan.json` `judge_context[0]` says so explicitly — so the
"fetched" route is not available inside this skill. Three routes remain, and the PR must
say which it took and why: (a) craft prose stated jurisdiction-free (build from the
best-established anchor toward the unknown; do not skip an intervening record that would
confirm identity continuity) with any census specifics written *as* a US-federal fact,
(b) via the `localities` entry `locality-guide` already writes, or (c) a
`research_append` precondition. **Do not write an unqualified census-year rule into §4.**
Issue #2251 is blocked on this card precisely to move jurisdiction-shaped passages out
of this body; new ones create work it must then undo.

**§4's location moved. Rule A is still item 8.** `research-plan/SKILL.md:314-339` (not
`:293-308`) holds seven items today; item 7 landed in PR #2033, merged 2026-09-04. The
scope note below is otherwise unchanged and still correct.

**Rule C's evidence is gone from the committed corpus. Do not plan work against it
without re-deriving.** Measured 2026-09-08 over the five committed run logs:

- `ut_research_plan_016` **passes** on `v1_2026-09-02_16-26-24` and
  `v1_2026-09-07_16-49-11`, scoring 3 on all eight dimensions. Its acceptance
  ("`_016` passing") is already met.
- The `localities` ownership violation appears in **1 of 5** committed runs
[...]

PR #2033, opened 2026-08-29
research-plan: the objective's target record leads — and the measurement showing prose does not bind (#1945)

### LINKED ISSUE #1945: [feedback] 2026-08-26T19:26Z
**Zip:** https://drive.google.com/file/d/1oc40PGwiPeWy1znpcbONnJGBobNPrmgA/view?usp=drivesdk

```
make feedback-case ZIP=~/Downloads/feedback-2026-08-26T19-25-57-024820Z.zip
```

- Submitted: 2026-08-26T19:25:57.024820Z
- Platform: web

Ikennaya1:
Closes #1945.

**The finding is not rule 7. The finding is that rule 7 does not bind** — and this PR ships the measurement, a guard that does bind, and a filed path to enforcement (#2030).

## The case

A FamilySearch researcher asked for the birth or baptism record of Hoval Hansen — born ~1780 in **Kongsberg**, died Trysil 18 August 1860. The agent spent **3h18m and 75 tool calls** pursuing the death record and **never searched Kongsberg at all**. Their note:

> "The death record gives his age as 80 years, but no place of birth is named. Not unusual, so the search strategy was poor. His baptism took place on 4 November 1780 in Kongsberg."

>

> "Looking for the birth first is the best place to start. … **Looking for a death will only provide an approximate year of birth.**"

The defect is in the **plan**, not the searching. It sequenced the Trysil death record at 1 and the requested Kongsberg baptism at **6 of 11**, then ran item 1 only. Execution followed the plan faithfully.

## Two things the report suggested that the evidence did not support

**It is not an indexed-vs-browse artifact.** Both candidates are indexed FamilySearch collections — Norway Baptisms `1467014` (28,370 Kongsberg records) and Norway Burials `1468081`. None of the six existing sequencing rules orders them.

**The second complaint describes a state, not a cause.** "It was looking in the wrong set of images when a call failed" is true, but it was already in the Trysil register because plan item 1 put it there.

What actually ordered them is the plan's own rationale for item 1: *"critical for establishing birth year (~1780) and confirming Kongsberg as origin"* — **both of which the objective states.** The plan spent its first four items re-deriving its own inputs.

## The measurement

`SKILL.md` step 4 gains a seventh sequencing rule, worded by the genealogist and guarded on the case where identity, jurisdiction or record availability are uncertain.

| run | rule 7 in the body | seq 1–2 | seq 3 |

|---|---|---|---|

| `scratch_2026-08-28_16-54-42` | no | Trysil burial + browse fallback | Kongsberg baptism |

| `scratch_2026-08-28_18-16-58` | **yes** | Trysil burial + browse fallback | Kongsberg baptism |

| `v1_2026-08-28_19-50-30` | **yes** | Trysil burial + browse fallback | Kongsberg baptism |

**Identical ordering three times, twice with the rule in the body.** The skill varied its `record_type` spellings between runs — `church`, then `vital_record`, then `church` again — so it was reasoning freshly each time and still landed on death-record-first.

Same shape #1616 measured on `check-warnings`, where a rule stated on five surfaces held on none.

**The rule is kept.** One measurement is not proof it never binds; this skill's own dive recorded 17-to-24-pass spreads with no code change. And a validator enforcing something the body never states is worse than one that restates it.

## The guard that does bind

`test_objective_target_leads_the_plan` — tag-gated, deterministic, and it never asks the judge. The rubric could not carry this: `rubric.md` is fenced by #1404/#1668, and the dimension that would grade it does not discriminate. **Re-derived across the five committed logs: 73 of 74 graded `Sequencing logic` occurrences are `3`. The single `2` — `wzk`, `v1_2026-08-24_13-37-16` — was corrected up to `3` by the annotator.** So it has fired once in the corpus and a human overturned it.

An earlier revision of this section said "3 on all 65 runs", quoting the 2026-08-24 dive rather than re-deriving from the logs — the dive's window is not the current corpus, since two of its logs have since been pruned. Stated so this one can be re-derived: read every `Sequencing logic` score out of each `v1_*.json` in the skill's runlog dir, then check the `.ann.json` siblings for corrections.

It *can* fail when a per-test `judge_context` instructs it to — in `scratch_2026-08-28_16-54-42` it scored `1` on this test, on an override written for it. Its own bars are what do not discriminate, which is the argument for a deterministic guard rather than another rubric bar.

It catches a shape ordering alone misses — the target written as `fallback_for` an indirect item, so it is only reached if that search succeeds. **Proven to fail three ways** before committing, per CLAUDE.md, in 8 direct tests.

**It binds on one fixture today.** The rule in `SKILL.md` is general; the validator hardcodes Kongsberg and Trysil and is gated on `objective-target`, a tag one test sets. @EdmondOware and main's own #1866 header both flag that shape — "a tag nobody sets is how a validator runs, passes, and cannot fail". This one *is* proven to fail, so it is not that failure mode, but the scoping is real: generalising it means matching a record type and place out of free-prose `project.objective`, which @T-FEH notes cuts both ways for false positives and negatives. Stated rather than papered over.

Its first version was wrong and the mistake is instructive: it required `record_type: "church"`, and the 18:16 run typed the baptism `vital_record`. Both are legitimate on an open field, so it reported "no item targets the objective" **on a plan that had one** and never reached the ordering check it exists for — exactly the field-name collision CLAUDE.md warns about. Both observed spellings are now pinned in a test.

## Scenario and fixtures are the tester's own

`hansen-kongsberg-baptism` is their state at the decision point: both localities surveyed, question written, no plan. The 12 MCP fixtures are **real responses extracted from the submitted bundle**, not authored — including the two `place_population` calls that genuinely returned `{"error":"Place not found"}`, kept faithful so the plan cannot lean on data it never had.

## `ut_research_plan_r3d` is marked `xfail`

Marked per @T-FEH's review, following the precedent set on #1946 (`whitfield-1850-household.json`): `test.expected_outcome: "xfail"` with an `xfail_reason` naming #2030 as the removal condition. An earlier revision of this body argued for shipping it as a plain `fail` because "the red is the finding" — that was wrong. A permanent plain `fail` makes the suite unreadable, and the next person cannot tell this test from a new regression. `xfail` pins the defect *and* keeps it visible, and an `xpass` becomes the signal that #2030 landed.

**One honest limit:** the committed run log recorded this run as `outcome: fail, expected_outcome: pass`, because the run predates the marking. The suite therefore still reads one red until some future run reclassifies it. The `xfail_reason` carries the explanation in the meantime.

It grades **order, not inclusion** — Trysil items are expected and credited, only their position fails — and it cross-references `ut_research_plan_016`, which tests the opposite failure, so fixing order cannot teach the skill to write thin plans.

## The first run: 17 pass, 2 partial, 3 fail of 22 (superseded — see the 2026-09-02 run below)

No regression from this change. Checked against the committed logs **before** diagnosing, per `eval/CLAUDE.md`, at no API cost:

| test | 08-24a | 08-24b | 08-24c | 08-25 | **08-28** | reading |

|---|---|---|---|---|---|---|

| `016` | pass | pass | fail | fail | **fail** | failing 3 runs, predates this branch |

| `001` | partial | pass | pass | pass | **partial** | within its spread |

| `005` | pass | partial | pass | pass | **partial** | within its spread |

| `014` | pass | pass | pass | pass | **fail** | **fails `test_research_json_validates_schema`** — see below |

| `r3d` | — | — | — | — | **fail** | this PR's test, now marked `xfail` |

**Which validator each red fails, and what I concluded** — added on @T-FEH's review, which is right that the earlier revision named the reds and their history without saying what they failed on:

| test | failed check | conclusion |

|---|---|---|

| `016` | `test_ownership_table` (universal) | The skill wrote a section it does not own. **Not this branch:** the same validator failed on `v1_2026-08-24_17-21-19` and `v1_2026-08-25_21-02-48` — a graded `fail` both times, not an abort. |

| `014` | `test_research_json_validates_schema` (universal) | `plans/1/items: [] should be non-empty`. @EdmondOware's behavioural read is better than my "retry artifact": every prior run appended items to the plan it had just created, this run appended all nine to **`pl_001`**, the pre-existing completed plan for `q_002`, leaving the new plan empty. It also issued the batch twice. `skill_attempts: 2`. Filed as **#2051** — the diff does not touch the writer tool. |

| `001` | none — judge-scored | `Correctness` 2, corrected to 3 on annotation. Filed as #2031. |

| `005` | none — judge-scored | `Completeness` 2 on a missing post-1850 census. Filed as #2032, since **merged into #2023** as Rule B. |

| `r3d` | `test_objective_target_leads_the_plan` (new, this PR) | By design. Of the whole suite this validator reds `r3d` alone. |

Two of those are **universal**-validator failures — an invalid `research.json` write and an out-of-lane section write. Both are serious and neither is about sequencing. The gate's own warning says to read them rather than treat the run as clean, and it is right to.

## Annotation

30 cells over `001`/`004`/`005`/`006`/`009`, one disagreement.

**`001` Correctness 2 → 3.** The judge marked the response down for saying `q_001` is "concluded" when `project_context` returned exactly that. The response quoted it faithfully and hedged both readings. Its own rationale then concedes the tool said `concluded`. The 08-24 run docked the same sentence twice. Filed as **#2031** — the derived state comes from a proof summary existing and reads neither the question's stored `status` nor its live plan items.

## Second review round, and a re-run that exposed something upstream

@T-FEH's second pass raised five findings. All five verified before acting; all five held.

**Fixed here.** The new test file carried a UTF-8 BOM and two `â€”` dashes, from a PowerShell `Set-Content -Encoding utf8` round-trip — the third time in this PR's history that same mechanism corrupted a file. It mattered more than a typo: the encoding lint catches the `SyntaxError` from `U+FEFF`, warns, and asserts only on `offenders`, so an unparseable file is **skipped while CI stays green** — the check-that-cannot-fail shape, in a file whose own docstring invokes the proven-to-fail rule. Stripped, re-saved, dashes fixed; the lint now parses and scans it. Its blind spot wants its own `nothing-checks` card and is not this PR's to fix.

**One semicolon defeated the new guard**, using a string the skill actually writes. Target and ahead-of-target were both substring matches on `jurisdiction`, so a single item satisfied both: relabelling the seq-1 Trysil item `"Trysil, Hedmark, Norway; Kongsberg, Buskerud, Norway"` made the identical defect pass. Not contrived — `r3d`'s own run wrote that verbatim. Target candidates now exclude items naming the indirect jurisdiction, pinned by a test using the observed string. Nine direct tests, all passing.

**My own retraction had not reached the artifacts.** "Sequencing logic scored 3 on all 65 runs" still shipped in the `xfail_reason` and at `validators/test_research_plan.py:373` while this body had already replaced it. The `xfail_reason` is the worse of the two — it is what the next person reads when deciding whether #2030 landed. Both now carry the re-derived 73 of 74 and the annotator's correction.

**The `place_population` fixture collision was resolved by measurement, not escalation.** `mcp-fixture-shape.test.ts` (#2014) rejected both not-found fixtures, and its header carries a standing "do not loosen the envelope" instruction, so this looked like @DallanQ's call. Before treating it as one: **all 12 fixtures this test declared are never consumed.** Across all three `r3d` runs the skill called only live `project_context` / `research_query` / `research_append`, because the scenario arrives with both localities surveyed and it goes straight to planning. The two offending fixtures were dead weight in exactly the shape the header records as pointing at `place-population.ts`'s unchecked cast. Dropped — 12 declared to 10 — with the reason in the test description. `vitest` green, 2765 passed.

### The round-2 re-run: 15 pass, 6 fail, 1 xfail of 22 — superseded by the 2026-09-02 run below, whose failure profile differs

`v1_2026-09-01_07-35-31`, $7.17.

**`r3d` records `xfail`.** That is what the run was for, and the one thing `eval-cosmetic-skip` could never deliver: the log now pins the defect as expected rather than showing an unexplained red, and an `xpass` becomes the signal that #2030 landed. **`001` and `005` both returned to pass**, confirming this body's "within its spread" reading rather than leaving it an assertion.

**This run log records the suite in worse shape than this branch caused, and that needs saying rather than leaving a reviewer to find it.** Five of the six failures — `002`, `006`, `010`, `014`, `q7m` — fail one validator, `test_research_plan_availability_claim_matches_counts` from main's #1866, with one message:

```
pl_003 item pli_011 calls collection 1999196 "indexed" but every response
this run received returned personCount 0
```

| test | 08-24a | 08-24b | 08-25 | 08-28 | **09-01** |
|---|---|---|---|---|---|
| `002` | pass | pass | pass | pass | **fail** |
| `006` | pass | pass | pass | pass | **fail** |
| `010` | pass | pass | pass | pass | **fail** |
| `q7m` | fail | pass | pass | pass | **fail** |
| `014` | pass | pass | pass | fail | **fail** |

#2014 (`33a80c0f`, 2026-08-31) rewrote `external-links-search-*.json` to satisfy its new shape check; #1866's validator reads `personCount` out of those same responses; the rewritten fixtures return `0`. **Neither PR is wrong alone.** Filed as **#2110** — it needs a decision on whether `personCount 0` is correct data, which is not answerable from here. `016` is the pre-existing ownership failure, failing since `v1_2026-08-24_17-21-19`.

The eleven changed fixtures arrived with @T-FEH's merge of `main`, not with the original branch. Rule 2 needs an active log, and this is the honest one.

### Annotating that run found a false fabrication flag in main's V1 validator

The annotation for `v1_2026-09-01_07-35-31` is complete — five sampled tests, 25 dimensions, `runlogs` green. One correction: **`007` Sequencing logic 3 → 2.** The plan wrote nine items with `fallback_for` null on all nine, so SKILL.md step 4 item 5 ("include contingencies") is unmet, while the judge scored 3 and described item 9 as "a fallback" the data does not encode; no rationale in the run uses the word either. `011` is the contrast in the same run — four explicit fallback chains, free-before-paid holding per year. The other 24 agree.

**Verifying it turned up a defect in #1866's V1 validator, fixed here.** `r3d` was failing **two** validators, not one, and the second was not real: `test_research_plan_rationale_identifiers_traceable` reported a fabricated collection id on an id the run had actually been served. `external_links_search` is in `_TRACEABLE_ID_TOOLS` because it returns collection ids — but it returns them **only as URL path segments**, and `_GROUNDED_NUM_RE`'s lookbehind refused any digit run preceded by `/`. So every id that tool serves was invisible to the grounded set: **22 of the 40 across its 16 fixtures.** `r3d` cited Ancestry collection `61749` after `external_links_search` served it as `ancestry.com/search/collections/61749/`, and was flagged for inventing it.

Dropping `/` from the lookbehind widens grounding only, which per that function's own docstring "can only remove a false fabrication, never add one" — 0 of 40 now ungroundable. Proven to fail first, per CLAUDE.md: the new proof test reds against the unpatched regex with `r3d`'s exact message, and `test_v1_fires_on_untraceable_identifier` still catches a genuinely invented id. All 18 tests in the #1866 proof file pass; full harness suite 3004 passed, 3 skipped.

**Why this belongs in this PR rather than an issue.** It is a one-line change in a file this PR already edits, so it fails CLAUDE.md's four-person-overhead test for filing. It also changes how this PR reads: `r3d` was red for one real reason and one false one, and without the fix it would have stayed red after #2030 lands — defeating the `xpass` signal described above. No re-run is needed; validators are not in the run-log snapshot, `r3d` stays `xfail` either way, and no other test failed this validator, so no recorded outcome moves. The committed run log still records the spurious failure, which the fixed validator would not produce.

## Third review round: the confound, and a re-run on this branch

@T-FEH's third pass raised three blocking items. All three verified; all three correct, and one of them is mine to own.

### The confound in the headline measurement

The scenario's own `research.json` told the skill to go to the death record first. `questions[0].rationale`:

> The specific death date and place (18 Aug 1860, Trysil) **is the firmest anchor**: Norwegian burial registers of this era routinely record age and birthplace (*fødested*), providing **the bridge back to the Kongsberg** parish baptism registers.

I wrote that fixture. So the three original runs measured body prose losing to input prose, not prose failing outright. Rather than restate the claim I re-measured it: copied the scenario, deleted that one sentence, changed nothing else, ran the same test three times.

| scenario | runs | target record sequenced first |
|---|---|---|
| as committed (rationale steers) | 3 | **0 of 3** |
| that one sentence removed | 3 | **1 of 3** |

**So neither account was right.** It is not "prose does not bind" — one run ordered correctly once the contrary rationale was gone. It is also not explained away by the confound — 2 of 3 still led with Trysil. What six observations support is that the rule **binds unreliably**, and for an ordering guarantee an intermittent rule is the same defect. #2030's precondition is unchanged; only its justification narrows, and that card now carries the corrected account so a developer picks up the re-measured version.

I stopped at three runs rather than five: the decision is identical at 1-in-3 or 2-in-5, so more runs would sharpen a figure without changing what anyone does.

### The probe found something better than what it went looking for

On the single run that ordered the plan **correctly**, the judge scored `Sequencing logic` **1**:

> the plan places the Trysil burial (pli_002) at sequence 2, ahead of the Kongsberg baptism (pli_001) at sequence 1

Sequence 2 is not ahead of sequence 1. `Correctness` scored 1 on the same reasoning. The dimension punished the one compliant plan in six runs.

That is a stronger argument for a deterministic guard than "the dimension does not discriminate", which is what this body previously said. A rubric that inverts cannot enforce the rule. Recorded on #1404 rather than filed — one observation on a scratch log is too thin for a card.

### The stale figure is gone rather than refreshed

"73 of 74" re-derives to **93 of 94**, the run log it attributed the single `2` to is not on disk, and the current single `2` (`_013` in main's `18-35-01`) is **confirmed at 2** by its annotation — so "a human overturned it" was false as well as stale. Both sites now carry no figure at all, with instructions to re-derive. It has needed re-deriving twice and nothing guards it.

### The re-run: the run log records 17 pass, 4 fail, 1 xfail; annotated, it reads 18 pass, 3 fail

`v1_2026-09-02_16-26-24`, $6.75, 40 minutes, zero aborts. Bought because #2085 merged a behavioural 19-line `research-plan/SKILL.md` edit and took the eval slot, leaving main's newest log inactive against this branch.

The first attempt aborted, and it was the environment rather than the corpus: the abort-storm breaker tripped at 4 transient aborts in 9 completed tests, with API/network at **1%** of reported skill work. The harness had auto-chosen 2.9x concurrency on a 7.9 GiB box. Re-run at `--concurrency 2` completed clean. That attempt cost $0.46 and wrote only a scratch log.

**`r3d` records `xfail`**, which is what the run was for.

**Why the two numbers differ.** The run log is the harness's arithmetic on the
judge's scores; the annotated reading applies the genealogist's corrections.
`wzk` failed on two judge scores with **no failing validator**, and both were
overturned in review — `Correctness` and `Tool Arguments`, each 1 → 3 — so the
test has no failing dimension and no failing validator left. The adjustment is
applied in one direction only because it is the only correction that crosses a
pass/fail line: every other annotated cell agrees with the judge. Both figures
are stated here rather than picking one, since the artifact says 17/4/1 and
will keep saying it.

**The run was scored against a superseded judge prompt.** The gate says so,
warn-only: the log carries `judge_prompt_hash 03f306ff2512…`, and the current
prompt hashes to `9451d1056f0a…`. #2093 moved it on main the day before this
run. That matters most for the `wzk` corrections, which rest on judge output,
and it is recorded here rather than left in a CI warning. Re-running to close
the gap would cost another $6.75 and would not change the corrections, which
turn on the test's own `judge_context` and on a declared tool parameter rather
than on anything the judge prompt controls.

**The four failures have four distinct causes** — this body previously described five failures sharing one, and that is no longer true:

| test | cause |
|---|---|
| `006` | `availability_claim_matches_counts`, collection 1999196 — the #2110 mechanism, and now the **only** test showing it |
| `002` | `fallback_for` points outside its own plan — `pl_003`'s `pli_013` points at `pli_006`, an item of the superseded plan. **Not attributable to this PR's rule on the evidence available:** `_002` passed this validator on all four prior retained runs, including this branch's own `07-35-31` with step 4 item 7 already in `SKILL.md`. Nor can it be excluded — the retained corpus is five runs deep and the harness does not pin temperature. The mechanism is a stale pointer into the prior plan rather than a renumbering artefact of reordering. |
| `014` | `research.json` schema: `plans/1/items: [] should be non-empty` |
| `wzk` | no validator failed; the judge failed it, and the annotation overturned both scores |

`q7m`, `010` and `016` now pass. `016` had been the long-standing ownership failure, red since `v1_2026-08-24_17-21-19`; `test_research_plan_new_plan_for_q_001` passes on its own merits.

### Annotation, and what it changed

Five sampled tests, 40 dimensions, two corrections:

- **`h4k` Record type selection — retracted, stands at 3.** This bullet previously cut the score to 2 on the grounds that `naturalization` is a recommended `record_type`. It is not: the recommended list is census, vital_record, probate, land, church, military, newspaper, cemetery, tax, immigration, court, other, and the enum is open besides. "Naturalization" is a fact-type value, and the check behind the claim was a substring match over the whole schema file that collided with an unrelated key. So `pli_006` used a recommended value and the judge's 3 was right. Retracted in `6a5da160d`; the annotation records it as a 3 → 3 with the reasoning. What survives is an observation, not a score cut: the judge described the item as "Naturalization (pli_006)" while the field reads `immigration`, so it graded the rationale text rather than the encoded field — the same failure mode as `007` last round.
- **`wzk` Correctness 1 → 3.** The 1 contradicts the test's own binding `judge_context[2]`, which says not to penalise the plan for the Wood-census item's rationale omitting the caveat as long as it appears somewhere. It does.
- **`wzk` Tool Arguments 1 → 3.** The two `fixture_not_found` calls are real, but the arguments were not wrong: `contextName` is a declared parameter, and the second call used the exact shape that succeeded twice in the same run. **No `place-search` Georgia fixture exists**, though the test declares `collections-search-georgia`, so Georgia is in scope by the test's own design. A corpus gap was being charged to the model.

`wzk` now has no failing validator and no dimension below 3.

### `q7m` exposed a contradiction between two fixtures

`q7m` failed the availability validator on the previous run and passes on this one. The reason is not the skill:

```
collections-search-pennsylvania.json    1999196  personCount=893214
collections-search-schuylkill.json      1999196  personCount=0
```

Same collection id, two fixtures, two different counts. `q7m` searched Pennsylvania and cited "indexed, ~893K records", which matched a response it received; `006` searched Schuylkill and got `0`. **Which fixture a run happens to query decides whether the validator fires.** #2110 now carries this, because it changes the question from "is `personCount: 0` correct data" to "these two fixtures disagree — which one is right".

### Two notes on process, recorded so they are not repeated

**Do not edit a test file after buying its run.** My citation fix to `xfail_reason` landed after this run had hashed the file, and `xfail_reason` is not in `_COSMETIC_TEST_FIELDS` (only `name`/`description`/`tags` are stripped), so a documentation edit reded rule 2 on a $6.75 run. Reverted to the hashed text; no second run was needed.

**Retention pruned `v1_2026-08-28_19-50-30`**, the one committed run log the `xfail_reason` cited. Recover it with `git show a75f9928:eval/runlogs/unit/research-plan/v1_2026-08-28_19-50-30.json` — `a75f9928` is the last commit that carried it. This body is the durable record of all six orderings; the five scratch logs are gitignored by design.

### Four corrections @T-FEH made to their own round-1 review

Recorded because they change what that review said: `_016` was never an `aborted`; their eight-move table used `v1_2026-08-17_17-52-29`, a log this PR deletes, so three of the eight "moves" were baseline artifacts; they praised the `xfail_reason` without reading it against this body's own retraction; and they had not run engine `vitest` that round.

## Out of scope, reproduced and filed

- **#2030** — the `research_append` precondition this measurement argues for. Writer-tool code plus a doctrine decision about when it must *not* fire; a different lane and reviewer, and a second doctrine change here would confound the measurement above.

- **#2031** — `project_context`'s `concluded` derivation. Fixing it via the fixture instead would touch `mid-research-flynn`, referenced by **20 skills across 125 tests**, invalidating all 20 snapshots.

- **#2023** — `005` has missed `judge_context`'s post-1850 census requirement in 2 of 5 runs. Filed as #2032 and **merged into #2023 on 2026-08-31** as Rule B: both are genealogist wording calls landing in or beside `SKILL.md` §4, both inside this skill's snapshot, so two cards would have bought two paid runs. The merge quotes this PR's own reasoning for deferring it. **Sequencing note recorded there:** §4 has six items on `main`, this PR adds item 7, so #2023's Rule A must be written as item **8** — which puts #2023 behind this PR rather than beside it.

- **#2051** — `014` appending plan items to the wrong plan, found by @EdmondOware comparing behaviour rather than outcome across all five runs. Writer-tool territory; this diff does not touch it.
- **#2110** — the #2014 / #1866 fixture-versus-validator collision above, which reds five `research-plan` tests that passed on every prior run. Needs a decision on whether the rewritten fixtures' `personCount 0` is correct data; both PRs are main's, not this branch's.

One observation not worth a card: **10 of `006`'s 13 declared fixtures are never touched**, because that scenario arrives pre-surveyed and the skill goes straight to planning. The same was true of 10 of my 12. Pre-surveyed scenarios make locality fixtures mostly dead weight.

## Verification

- Scenario validates against `research.schema.json`; test validates against `unit-test.schema.json`; no duplicate test ids

- `check_runlogs.py`: all rules satisfied locally

- Harness unit suite: **2,934 passed, 3 skipped**

- New validator proven to fail on three violating shapes and to skip correctly on four non-applicable ones

- Cost: **$7.38** for the suite, **$0.84** for two single-test runs


EdmondOware (approved):
Approving.

The framing is the right one and worth other people reading. "The finding is not the rule, the finding is that the rule does not bind" is the same shape I hit on search-external-sites, where a rubric rule turned out to be addressed to a judge that never receives the section it names. Moving this to a deterministic validator is the correct lane, and the validator comment explaining why the rubric cannot carry it is the part that will save the next person a paid run.

Rule 7 is genealogically sound and correctly conditioned. It fires only when the objective both names the record type and supplies the jurisdiction, and the exception clause covers the three cases that actually justify leading with something else.

One thing to file before this closes, found by rate checking the reds rather than reading them. Of the five non-pass tests, `ut_research_plan_014` is the only one that is neither by design nor inside its normal rate: 0 of 4 prior committed runs, failing here.

Comparing behaviour rather than outcome across all five runs: every prior run appended its plan items to `pl_002`, the plan it had just created for q_001. This run appended all nine to `pl_001`, the pre-existing completed plan for q_002, leaving `pl_002` with `items: []` and failing the schema's non empty requirement. It also issued the batch twice where every prior run issued it once.

I am not claiming rule 7 caused it and the diff does not touch the writer tool, so this reads as worth a card rather than a fix here. I checked #2030, #2031, #2032 and the open issues and nothing covers it.

For the next reviewer: `_016` was already red on 08-24 and 08-25 with the same `test_ownership_table` failure, so it is not this branch. And the rule is general while the validator hardcodes kongsberg and trysil, so it binds on one fixture today. Both are fine, just worth stating.

T-FEH (changes requested):
Reviewed on the merged head (`9826007a` — your `bcc8a1fb` plus a clean merge of main). Eval gate passes. Two things block, both about what the run log says versus what the body says.

The framing is the strongest part of this PR — "the finding is that the rule does not bind" is the right conclusion to draw from three runs with identical ordering, and moving the guard from prose to a deterministic validator is exactly what ADR-0011's first question asks for. The new validator also only reds its intended target: of the whole suite, `test_objective_target_leads_the_plan` fails on `r3d` alone.

## 1. `ut_research_plan_r3d` ships as a plain `fail`, not `xfail`

```
test JSON:  type: positive   expected_outcome: None   xfail_reason: None
run log:    outcome: fail    expected_outcome: pass
```

"Ships failing by design" and "ships as a plain fail" are different things. As written the suite is permanently red, and the next person cannot tell `r3d` from a new regression — which is the whole cost of a red test that is supposed to stay red for now.

The repo already settled this two days ago, on PR #1946, for the identical situation — a mined test pinning a defect with its route to green tracked elsewhere:

```
eval/tests/unit/search-records/whitfield-1850-household.json
  expected_outcome: xfail
  xfail_reason: present
```

Same treatment here: `expected_outcome: "xfail"` with an `xfail_reason` naming #2030. Then it pins the defect, stays visible, and stops the suite reading as broken.

## 2. Three regressions the body does not mention, two of them universal-validator failures

The body discusses `r3d` as though it is the only non-pass. Against main's baseline the run moves eight tests:

```
ut_research_plan_001   pass    -> partial
ut_research_plan_005   pass    -> partial
ut_research_plan_007   partial -> pass
ut_research_plan_014   pass    -> FAIL
ut_research_plan_016   aborted -> FAIL
ut_research_plan_qxr   partial -> pass
ut_research_plan_wzk   aborted -> pass
ut_research_plan_r3d   (new)   -> fail
```

Three improve. But `_014` and `_016` fail **universal** validators, not the new one:

```
ut_research_plan_014   test_research_json_validates_schema
ut_research_plan_016   test_ownership_table
```

Those are the skill writing an invalid `research.json` and writing a section it does not own — both serious, both unrelated to sequencing, and `_014` was passing on main. `_001` and `_005` are judge-scored drops with no failed validator.

The gate says as much in its own warning:

> 3 test(s) … produced no graded dimensions … `ut_research_plan_016`, `ut_research_plan_r3d`, `ut_research_plan_014`. **Read their `aborted_reason` / validator results before treating this run as a clean pass.**

I am not asking you to fix the skill here — `_016` was `aborted` on main so it is newly graded rather than newly broken, and a schema-invalid write is its own piece of work. What I am asking is that the body say they exist and what you concluded about each, the way it does for `r3d`. A reader currently comes away thinking one test is red by design and the rest are clean.

## Still open on my side

The agent pass on the validator's own quality is still running — specifically whether `test_objective_target_leads_the_plan` can fail for the right reason on a differently-worded objective (`project.objective` is free prose, so a string-match approach has both false-positive and false-negative risk), whether its 161 lines of pinned tests are non-tautological, and whether the two measured claims re-derive — the three identical-ordering runs, and `Sequencing logic` scoring 3 on all 65 runs. Those are the load-bearing evidence for "prose does not bind", so I will follow up rather than take them on the body's word.

Requesting changes on the `xfail` marking and the disclosure. Both are edits to text you already have in hand.

T-FEH:
Following up the items I said I'd verify rather than take on the body's word. The validator is good; the two measurements behind the framing do not reconcile against the repo.

## The validator is well built

Mutation-tested through five constructed plans:

```
target leads (Kongsberg church at seq 1)          PASS
Trysil sequenced ahead of the target              FAIL — "1 Trysil item(s) are sequenced ahead of it"
no Kongsberg item at all                          FAIL — "no plan item targets the objective's record"
target at seq 1 but fallback_for a Trysil item    FAIL — "must not be gated behind an indirect source"
no plan items                                     SKIP
```

Both failure modes from the real session are pinned **separately**, and I specifically checked that the `fallback_for` branch is reachable rather than dead code — it only fires when the target already leads, so it needs its own case, and it has one. The tag is armed on exactly one test (`r3d` carries `objective-target`), which is the check I'd want after #1870 shipped a validator gating on a tag no test carried. The pinned tests are not tautological either: neutering the sequencing assert reds `test_fires_when_the_death_place_is_sequenced_first` and `test_target_is_recognised_under_either_record_type_spelling` by name.

One thing worth saying plainly, since the body reads as though rule 7 is now guarded: it isn't. `_TARGET_JURISDICTION = "kongsberg"` and `_INDIRECT_JURISDICTION = "trysil"` are hardcoded, so this pins **this scenario**, not the rule. That is the right design — it sidesteps parsing `project.objective`, which is free prose — but rule 7 in general remains prose, which is the thing the PR argues does not bind.

## Neither measurement reconciles

**"Twice with the rule present."** Rule 7 changes `research-plan/SKILL.md`, which is snapshotted, so the runs that had it are identifiable by hash:

```
SKILL.md with rule 7 (this branch):  89c66dfe
SKILL.md without   (origin/main):    cb68e3f8

08-24_13-37-16   66612338    (older body)
08-24_14-32-36   66612338    (older body)
08-24_17-21-19   cb68e3f8    no rule
08-25_21-02-48   cb68e3f8    no rule
08-28_19-50-30   89c66dfe    rule present
```

Exactly **one** committed run has the rule. The two others were presumably `--test` scratch runs, which are gitignored by design — a legitimate way to run them, but it means the evidence for the PR's central claim is not in the repo and cannot be re-derived by a reviewer.

**"`Sequencing logic` scored 3 on all 65 runs."** Over the committed research-plan corpus today I get **79 gradings: 78 × 3, and one 2** — `ut_research_plan_wzk` in `v1_2026-08-24_13-37-16.json`, which predates this PR. I cannot tell whether that run was inside the dive's audited set, which is the point: as written the figure names no set a reader can reconstruct, and the corpus it most naturally reads against contradicts the "all" .

Neither is fatal to the conclusion — three identical orderings is still suggestive, and one 2 in 79 does not make the rubric discriminating. But these two numbers are the whole evidentiary basis for "the rule does not bind", which is a doctrinal claim about lane 4, and the repo's own rule is that a measurement gets re-derived rather than reworded. Either name the runs and the audited set so they can be checked, or re-derive both against what is committed.

That is in addition to the two items in my review — `r3d` shipping as a plain `fail` rather than `xfail`, and the three undisclosed regressions.

Ikennaya1:
Both blocking items are in, on `e0ea728c`. Thank you for the `xfail` push — you were right and my "the red is the finding" argument does not survive contact with a permanently red suite.

## 1. `xfail` — done

`test.expected_outcome: "xfail"` with an `xfail_reason` naming #2030 as the removal condition, following your #1946 precedent. I doubted that citation at first because my JSON parse looked at the top level and the fields are nested inside `test`; `whitfield-1850-household.json` uses it exactly as you described. My error.

**One thing it does not fix, and it is worth knowing before you re-review.** The committed run log recorded this run as `outcome: fail, expected_outcome: pass`, because the run predates the marking. So the suite still reads one red until some future run reclassifies it — the `xfail_reason` carries the explanation in the meantime, which does most of what you asked for but not all.

## 2. Disclosure — done, and your premise is half right

The body already carried a five-row table of every non-pass with its per-log history. What it did **not** do is say which check each red failed, or name `016` as an ownership violation — that part of your ask was fair and is now a second table:

| test | failed check | conclusion |
|---|---|---|
| `016` | `test_ownership_table` (universal) | wrote a section it does not own — **not this branch**, see below |
| `014` | `test_research_json_validates_schema` (universal) | `plans/1/items: [] should be non-empty`. Filed **#2051** |
| `001` | none — judge-scored | `Correctness` 2, corrected to 3 on annotation. Filed #2031 |
| `005` | none — judge-scored | `Completeness` 2, missing post-1850 census. Filed #2032 |
| `r3d` | `test_objective_target_leads_the_plan` (new) | by design; now `xfail` |

## 3. Your baseline column is wrong on four tests

This is the one place I am pushing back, because your `016` inference rests on it. Re-derived by reading each `test_id` out of every committed `v1_*.json`:

| test | your table | the logs |
|---|---|---|
| `007` | partial → pass | **pass in all 5** |
| `qxr` | partial → pass | **pass in all 5** |
| `wzk` | aborted → pass | **partial → pass** |
| `016` | aborted → FAIL | **fail → fail** |

`016` failed `test_ownership_table` on `v1_2026-08-24_17-21-19` and again on `v1_2026-08-25_21-02-48` — a graded `fail` both times, not an abort. So it is not "newly graded rather than newly broken"; it has been broken and graded for three runs. @EdmondOware read it the same way independently.

Three of your five "moves" therefore are not moves. The two that are — `014` and `r3d` — are the two I agree with you about.

## 4. A claim of mine that needed correcting, before your agent pass gets to it

You said you were re-deriving my two measured claims. One of them does not hold as written.

I wrote "`Sequencing logic` scored 3 on all 65 runs", taking the figure from the 2026-08-24 dive rather than from the logs. Re-derived across the five committed logs: **73 of 74 graded occurrences are `3`. There is one `2`** — `wzk`, `v1_2026-08-24_13-37-16` — **and the annotator corrected it up to 3.**

Substantively that is stronger than what I claimed: the dimension has fired once in the committed corpus and a human overturned it. But "all 65" is not what the logs say, and I should have re-derived it rather than quoting the dive. The PR body will carry the corrected figure with the method.

The other claim — three runs, identical ordering — does re-derive: `scratch_2026-08-28_16-54-42` (before rule 7), `scratch_2026-08-28_18-16-58` and `v1_2026-08-28_19-50-30` (after). Trysil burial at 1, its browse fallback at 2, Kongsberg baptism at 3, all three times.

## 5. Conceded: the validator binds on one fixture

You and @EdmondOware both flagged this and main's own #1866 header says it plainly — "a tag nobody sets is how a validator runs, passes, and cannot fail". Mine is gated on `objective-target`, which one test sets, and it hardcodes Kongsberg and Trysil.

It is not that failure mode — it is proven to fail, in 8 direct tests covering three violating shapes — but "binds on one fixture today" is true and is now stated in the body rather than left to be found. Generalising it means matching a record type and place out of free-prose `project.objective`, which is the false-positive risk you named; I would rather ship a narrow guard that works than a general one that misfires.

## What I need: `eval-cosmetic-skip`

`runlogs` fails, and **`expected_outcome` is not in `_COSMETIC_TEST_FIELDS`** (`name`, `description`, `tags`), so your edit changes the test JSON's snapshot hash. Measured:

```
hash before edit: 680851f14dd2ee7f
hash after  edit: c62a7fc24b5aebd5
```

So "both are edits to text you already have in hand" costs a full 22-test re-run at ~$7.40 unless the label is applied.

The case for it is unusually clean, and rule 2 is stale on **two** files, neither of which can change behaviour:

1. **`plan-leads-with-objective-target-record.json`** — `expected_outcome` is harness bookkeeping. It is not in the skill's prompt, not in its directory, and the skill never sees it. It changes how a result is *labelled*, nothing about what the skill does.
2. **`skills/research-plan/references/validation-protocol.md`** — *missing on disk*, and not from this branch: main's #2045 deleted it yesterday as one of "11 unreachable skill references". I verified that independently rather than trusting the commit title — `git grep validation-protocol 312d3c67~1 -- packages/engine/plugin/skills/research-plan/` returns nothing. The skill never loaded it.

Happy to pay for the re-run if you would rather have one, but I would want that to be a decision rather than something I spent quietly. Five runs went into this PR already ($8.22).

Ready for another look whenever your agent pass lands.

Ikennaya1:
Branch updated to `e763294e` — 0 behind `main`, no conflicts. Main's one commit touched only `record-extraction` and `search-images` annotation files, nothing in this run log's snapshot scope.

**I pushed the update *before* asking for the label deliberately**, since `eval-cosmetic-skip` expires on every push. It can now land on a final commit rather than being invalidated by the next one.

## The only red is `runlogs`, on two files, and neither changes behaviour

```
skill `research-plan`: latest full-skill run log v1_2026-08-28_19-50-30.json is NOT active
— 2 snapshot file(s) differ from the working tree
```

1. **`eval/tests/unit/research-plan/plan-leads-with-objective-target-record.json`** — the `xfail` marking @T-FEH asked for. `expected_outcome` is not in `_COSMETIC_TEST_FIELDS` (`name`, `description`, `tags`), so it changes the snapshot hash: `680851f14dd2` → `c62a7fc24b5a`. The field is harness bookkeeping — not in the skill's prompt, not in its directory, never read by the skill. It changes how a result is *labelled*, not what the skill does.

2. **`packages/engine/plugin/skills/research-plan/references/validation-protocol.md`** — *missing on disk*, and **not from this branch**: main's #2045 deleted it yesterday as one of "11 unreachable skill references". Verified independently rather than trusting the commit title — `git grep validation-protocol 312d3c67~1 -- packages/engine/plugin/skills/research-plan/` returns nothing. The skill never loaded it.

The other 31 annotations on that check are **warnings from the three warn-only lints** and all pre-existing corpus-wide — negative-routing reciprocity, `proof-conclusion`'s rubric naming tools it cannot call, `locality-guide`'s `collection_read` fixtures. None is from this branch.

## What clears it

**`eval-cosmetic-skip`** — rule 2 downgrades to a warning. This is the ask.

**Or a re-run** — ~$7.40, ~44 minutes, plus a fresh five-test annotation. Worth knowing it buys one thing the label does not: the committed log currently records `r3d` as `outcome: fail` because it ran before the marking, so the tally shows a red until some future run reclassifies it as `xfail`. If that visible red is the thing that matters, the re-run is the honest answer and I will take it rather than argue — I would just rather it be a decision than something I spent quietly. Five runs went into this PR already ($8.22).

Both of @T-FEH's blocking items are in. Ready for another look.

T-FEH (changes requested):
Second pass, on `a9e2ed38` (your `e763294e` plus a merge of main I pushed). **Both items I blocked on are properly resolved.** But two checks are now red, and this round turned up three things neither of us had — plus four corrections to my own round-1 review.

## What is fixed

**The `xfail` marking**, with a reason better than the `whitfield` precedent: it names #2030 as the removal condition and ends "an xpass here is the signal that it did", which is what stops an xfail becoming furniture.

**The disclosure**, more fully than I asked. I re-derived every cell of your five-run history table from the committed logs and it is exactly right, as are the run totals (17/2/3 of 22) and `73 of 74` for Sequencing logic. "Within its spread" for `_001` and `_005` is a real claim — each has been partial before — and `_014` is correctly singled out as the genuine first-time fail.

**And the validator does fail for the right reason** — my open item from round 1. Driving it directly: the corrected plan passes, the observed defect reds, a `fallback_for` gate reds on its own branch, and a missing target reds with a distinct message. Crucially the *corrected* plan passes, so #2030 really can turn this green.

## Four corrections to my round-1 review

- **`_016 aborted -> FAIL` was wrong.** It is a graded `fail` on main's latest with `aborted_reason: None`, failing since `v1_2026-08-24_17-21-19`. Your "**Not this branch**" is right; @EdmondOware was right; I was not.
- **My whole 8-move table used a stale baseline.** I diffed against `v1_2026-08-17_17-52-29` — a log this PR *deletes*, three runs and eleven days behind main's latest. Against `v1_2026-08-25_21-02-48` it is 5 moves, not 8: `_007`, `_qxr` and `_wzk` were artifacts of my baseline, not improvements from your change.
- **I praised the `xfail_reason` without reading it against your own body.** It carries "Sequencing logic scored 3 on all 65 runs", the figure your body explicitly retracts in favour of the re-derived 73 of 74. The retraction did not reach the artifacts — it still ships in `eval/tests/.../plan-leads-with-objective-target-record.json` and `validators/test_research_plan.py:373`. The `xfail_reason` is the worse of the two: it is what the next person reads when deciding whether #2030 has landed.
- **I did not run engine vitest this round**, which is how I missed the red below. My own gap, not yours.

## [Blocking] Two red checks

**vitest.** The two new fixtures are rejected by the shape check main absorbed from #2014:

```
place-population-kongsberg-notfound.json: fields the tool never returns: error, place_id;
                                          fields the tool always returns, missing: place
place-population-trysil-notfound.json:    same
```

The fixtures look right to me and #2014's header agrees in advance — it names this exact shape, says "No fixture uses that shape" (which you have now falsified), locates the fault in `place-population.ts:52`'s unchecked `response.json() as Promise<PopulationResponse>`, and instructs "**Do not loosen the envelope to admit it**". So the three ways out are: fix the cast, add a typed not-found arm, or change the fixtures — and your body's argument for keeping them faithful is the one I would defend. This is a cross-PR collision with a standing instruction attached, so it is @DallanQ's call rather than yours or mine.

**runlogs.** 13 snapshot files drifted. Your 12:27 comment asks for `eval-cosmetic-skip` on the basis of two non-behavioural files — true when you wrote it, not true now:

```
 1  the xfail edit                              cosmetic (expected_outcome is not in _COSMETIC_TEST_FIELDS, so it counts)
 1  references/validation-protocol.md missing   cosmetic (#2045 deleted it as unreachable)
11  external-links-search-* / place-population-* / place-search-all-any   ← #2014, substantive
```

`place-population-muhlenberg.json` alone moved `place` from a string to an object and dropped a `notes` field the skill used to read. The skill's inputs really changed, so a re-run is needed regardless — but it must come *after* the fixture question above, or it is wasted. Both of those arrived with my merge, so the cost is not one your original branch incurred.

## [Blocking] The new test file has a UTF-8 BOM and silently opts out of the encoding lint

```
$ head -c6 eval/harness/tests/unit/test_research_plan_objective_target_validator.py | xxd
00000000: efbb bf22 2222

$ uv run pytest tests/unit/test_encoding_lint.py -q
UserWarning: encoding lint could not parse and skipped these files (not scanned):
  test_research_plan_objective_target_validator.py: SyntaxError: invalid non-printable character U+FEFF
6 passed, 1 warning
```

It is the only BOM'd `.py` in the repo. The lint catches the `SyntaxError`, warns, and asserts only on `offenders` — so an unparseable file is skipped and CI stays green. That is the "check that cannot fail" shape, in a file whose own docstring invokes the proven-to-fail rule. Same cp1252 round-trip left mojibake at `:65` and `:152` (`â€"` for `—`). Strip the BOM, re-save UTF-8, fix the two dashes. The lint's blind spot is worth a `nothing-checks` card separately — it is not yours.

## [Blocking] One semicolon defeats the new guard, using a string the skill already writes

The target and the "ahead" set are both substring matches on `jurisdiction`, so one item can be both. Relabelling the seq-1 Trysil item makes the identical defect pass:

```
Trysil seq1 / Kongsberg seq2                                    FAIL   (correct)
same, seq1 jurisdiction = "Trysil, Hedmark, Norway; Kongsberg, Buskerud, Norway"   PASS
```

That is not a contrived string — `pli_005` in the very run this test grades carries it verbatim. Excluding indirect-jurisdiction items from the target candidate set closes it:

```
if _TARGET_JURISDICTION in juris(i)
   and _INDIRECT_JURISDICTION not in juris(i)
   and str(i.get("record_type") or "").lower() in _TARGET_RECORD_TYPES
```

Worth noting alongside: `_TARGET_JURISDICTION`/`_INDIRECT_JURISDICTION` are hardcoded to kongsberg/trysil while the tag `objective-target` is generic, so a future author attaching that tag elsewhere gets a Kongsberg check firing on their scenario. Every other tag in that file is scenario-named.

## Requesting changes

The two items I blocked on are answered well, and the validator holds up on every shape except the one above. What is left is the BOM, the semicolon evasion, propagating your own retraction into the two artifacts that still carry "65 runs" — and a decision from @DallanQ on the `place_population` fixture collision, which gates the re-run.

Ikennaya1:
Pushed two commits.

**1. The annotation is complete.** Five sampled tests, 25 dimensions, on `v1_2026-09-01_07-35-31`. All checks are now green, including runlogs.

I made one score change: **ut_research_plan_007**, *Sequencing logic*, **3 → 2**. The plan produced nine items and left `fallback_for` empty on all nine, so it does not satisfy SKILL.md Step 4, item 5 ("include contingencies"). The judge awarded a 3, arguing item 9 was a fallback, but nothing in the data supports that interpretation, and the run rationale never uses the term. `ut_research_plan_011` in the same run provides the contrast: four explicit fallback chains, with free sources sequenced before paid ones throughout. I agreed with the remaining 24 dimensions.

**2. While reviewing the run, I found a bug in the existing V1 validator (#1866), and this PR fixes it.**

`ut_research_plan_r3d` was failing two validators, not one. The second failure was incorrect: it claimed the plan had invented Ancestry collection **61749**. In reality, that collection ID came directly from `external_links_search` during the run.

The root cause was that `external_links_search` only returns collection IDs inside URLs (for example, `ancestry.com/search/collections/61749/`). The validator's pattern refused to recognize numbers that followed a `/`, making every such ID invisible to it. Across the 16 fixtures, that affected 22 of 40 collection IDs. Cite one correctly and the validator accused the plan of inventing it.

The fix removes `/` from the exclusion pattern. This can only eliminate false positives; it cannot create new ones, which is consistent with the function's own documented purpose. After the fix, none of the 40 IDs are invisible.

Per CLAUDE.md, I first demonstrated the failure. The new test reproduces the exact error message from the r3d run log against the old pattern, while the existing test for genuinely invented IDs continues to fail as intended. All 18 tests in the #1866 proof file pass, and the full harness suite is **3,004 passed, 3 skipped**.

**Why I fixed it here instead of filing a separate issue:** it is a one-line change in a file already touched by this PR, which falls below CLAUDE.md's threshold for a standalone ticket. It also affects how this PR is interpreted: r3d was red for one legitimate reason and one false one. Left unfixed, it would remain red even after #2030 lands, undermining the xpass signal the PR body relies on.

No re-run is needed. Validators are not part of the run-log snapshot, r3d remains xfail either way, and no other test was failing this validator. The committed run log still records the false failure, which the corrected validator would no longer produce.

Ready for another look.

T-FEH (changes requested):
Third pass, on `2d16aced` (your `6c5f2769` plus a merge of main I resolved and pushed). Engine **2809 passed / 118 files**, `tsc` clean, harness **3062 passed / 2 skipped**. All four of my actionable round-2 items are closed, two of them better than I asked. But this round turned up something that goes to the PR's central claim.

## What is fixed, including one I'd single out

**The `place_population` collision — resolved by measurement instead of escalation.** I said it was the lead's call; you checked first whether the call had consequences. All 12 declared fixtures went unconsumed, so the two offending ones were dead weight. 12 → 10, vitest green, envelope untouched. Asking whether a decision matters before spending someone's attention on it is the right instinct.

**The BOM is gone** — no BOM in any tracked `.py`, no mojibake, and the lint provably scans the file now (`offenders: 0, unparseable: []`, no skip warning).

**The semicolon evasion is closed**, proven both directions — I re-ran my exact mutation against the shipped validator and against a copy with only the new clause reverted:

```
                                              post-fix   pre-fix
seq-1 relabelled "Trysil…; Kongsberg…"          FAIL       PASS
plain defect                                    FAIL       FAIL
legitimate good plan                            PASS       PASS
```

**And `6c5f2769` is in scope, not creep** — a lookbehind fix in this PR's own validator, for a false fabrication flag this PR's own run produced, with a test I broke to check it earns its place (restoring the old regex reds `test_v1_grounds_served_url_path_identifier`).

## [Blocking] The headline measurement is confounded by the scenario's own state

The PR's claim is that rule 7 does not bind: three runs, identical ordering, twice with the rule in the body. But the scenario hands the skill a contrary instruction before it starts. `hansen-kongsberg-baptism/research.json`, `questions[0].rationale`:

> The specific death date and place (18 Aug 1860, Trysil) **is the firmest anchor**: Norwegian burial registers of this era routinely record age and birthplace (*fødested*), providing **the bridge back to the Kongsberg** parish baptism registers.

And the plan the skill wrote at sequence 1 echoes it almost word for word:

> burial registers routinely record age at death and *fødested* (birthplace), providing **the direct bridge to the Kongsberg** baptism jurisdiction

So the skill is not ignoring prose. It is following the prose it was given, in the project state, over the prose added to the body. That is a different and narrower finding: **a SKILL.md rule loses to a contrary rationale already sitting in the input.** It may well be the more useful finding — but it is not "prose does not bind", and that claim is load-bearing in four places: the PR title, the `xfail_reason`, the validator docstring, and issue **#2030**, which is filed `nothing-checks` on the strength of it.

The guard and rule 7 are unaffected either way. Either run a variant with the rationale neutralised, or restate the finding to what was actually measured.

## [Blocking] "73 of 74" is stale again, and cites a run log that is not in the repo

Re-derived across all seven committed logs on this branch: **93 of 94**, not 73 of 74. And the `xfail_reason` attributes the single `2` to `v1_2026-08-24_13-37-16` — that file is not on disk. The current single `2` is `ut_research_plan_013` in main's `18-35-01`, and its annotation **confirms** it at 2 rather than correcting it up, so "it fired once and a human overturned it" no longer holds either. This PR's own annotation separately corrects `_007` **down** to 2 on the same dimension.

Three sites carry it — the body, the `xfail_reason`, and `validators/test_research_plan.py:371-373` — and nothing guards the figure. This is the second round the number has needed re-deriving; it is worth stating it as a range or dropping it, the way you dropped the fixture count.

## [Blocking] #2110 asks the lead to rule on the wrong cause

The body and the issue say #2014's fixture rewrite emptied `personCount`. All five failures name collection **1999196**, served by `collections-search-schuylkill.json` — and:

```
#2014 (33a80c0f) touched that file:  0 hits   (its only collections-search edit was …-chicago.json)
personCount: 0 introduced by:        d3fed619  "research-plan: verify next-source availability…"
```

So the `0` predates #2014 and arrived from a research-plan PR. The real mechanism is a **new validator meeting a pre-existing fixture** — `availability_claim_matches_counts` landed in `3759fcb9` and first appears in run logs dated 2026-09-01. #2110 is labelled `needs-decision`; correcting its premise before @DallanQ answers it will save the decision being made on a false account.

## [Not yours] The gate is red because main took the eval slot

```
::error:: latest run log v1_2026-09-01_18-35-01.json is NOT active
          — research-plan/SKILL.md: content-differs
```

**#2085** also edits `research-plan/SKILL.md`, by 19 lines, and merged first with three paid runs. Your `07-35-31` was correct when you bought it. `eval-cosmetic-skip` cannot rescue this — both edits are behavioural. One thing I checked rather than assumed: #2085's changes are about plan **supersession**, not sequencing, so they do not overlap your rule and the finding should survive a re-run rather than need re-deriving.

Sequencing matters: **#2110's ruling first, then the re-run**, or the run is bought twice again.

## On the merge I pushed

The conflict was `research-plan/v1_2026-09-01_17-38-18.ann.json`, modify/delete — main carries that run log and its annotation, your branch had neither. I kept main's, since deleting an annotation whose log exists would break rule 3, and confirmed afterwards that all seven logs have all seven annotations with no orphans.

## Requesting changes

On the confound and the two wrong-premise items. None of them is a code defect — the guard, the fix and the xfail marking are all sound — but the PR's title claim, its `xfail_reason`, and a `needs-decision` card all rest on accounts the artifacts do not support. Worth flagging to @DallanQ that research-plan's eval slot now has two PRs and a pending decision stacked on it.

Ikennaya1:
All three items were right. I verified each before touching anything, and the
confound is mine — I wrote that fixture.

## The confound: correct, and it changes the claim less than you'd expect

You're right that questions[0].rationale steers the skill at the death record,
and that the three original runs therefore measured body prose losing to input
prose. I took your first option and re-measured rather than just restating.

Copied the scenario, deleted exactly that one sentence, changed nothing else,
ran the same test three times:

| scenario                          | runs | target record first |
|-----------------------------------|------|---------------------|
| as committed (rationale steers)   | 3    | 0 of 3              |
| that one sentence removed         | 3    | 1 of 3              |

So neither account was right. It isn't "prose does not bind" — one run ordered
correctly once the contrary rationale was gone, so your objection has real
force. But it also isn't explained away by the confound: 2 of 3 still led with
Trysil. What the six observations support is that the rule binds UNRELIABLY.

I've stopped at three rather than buy five. For an ordering guarantee the
decision is identical at 1-in-3 or 2-in-5 — an intermittent rule cannot be
relied on at the point a plan is written — so more runs would sharpen a figure
without changing what anyone does. Say the word if you'd rather see five.

#2030's precondition is unchanged; only its justification narrows. Corrected on
that card directly, so a developer picking it up sees the re-measured version
rather than the one I filed.

## The figure: dropped, not updated

93 of 94 confirmed, and the log I attributed the single 2 to isn't on disk. Two
things I'd add: the current single 2 is _013 in main's 18-35-01, and its
annotation confirms it at 2 — so "a human overturned it" was false too, not
just stale.

Both sites now carry no figure. You suggested a range or dropping it; I dropped
it. It has needed re-deriving twice, nothing guards it, and any number there is
stale as soon as the corpus moves. The comment says how to re-derive instead.

And I nearly repeated your own finding: my first rewrite cited five scratch_*
logs, which .gitignore excludes — exactly as unfindable as the file you
flagged. The xfail_reason now cites one run log, checked both on disk and
tracked in git, and says plainly that the scratch orderings live in the PR body.

## #2110: corrected on the issue

Verified and posted before @DallanQ rules. #2014 touched only
collections-search-chicago.json; personCount: 0 came from d3fed619 on
2026-06-27, two months before the validator that reads it (3759fcb9). New
validator, pre-existing fixture, exactly as you said. The decision needed is
whether 0 is correct for collection 1999196 — not whether #2014 broke it.

## One thing the probe turned up that I wasn't looking for

On the single run that ordered the plan correctly, the judge scored Sequencing
logic 1:

  "the plan places the Trysil burial (pli_002) at sequence 2, ahead of the
   Kongsberg baptism (pli_001) at sequence 1"

Sequence 2 is not ahead of sequence 1. Correctness scored 1 on the same
reasoning. The dimension punished the one compliant plan in six runs.

That's a better argument for the deterministic guard than "it doesn't
discriminate" — a rubric that inverts can't enforce the rule. Recorded on #1404
rather than filed; one observation on a scratch log is too thin for a card.

## Sequencing

Agreed on #2110 first, then the re-run — no point buying the slot twice. The
gate stays red until then, and #2085's edit is the reason, not this branch.

Harness suite 3061 passed, 3 skipped on a75f9928.

Ikennaya1:
All three were right, and the confound is mine — I wrote that fixture.

**Re-measured rather than restated.** Deleted only the "firmest anchor" sentence,
changed nothing else, ran the same test three times:

| scenario                        | runs | target record first |
|---------------------------------|------|---------------------|
| as committed (rationale steers) | 3    | 0 of 3              |
| that one sentence removed       | 3    | 1 of 3              |

Neither account survives. Not "prose does not bind" — one run ordered correctly
without the steering, so your objection holds. Not explained away either — 2 of 3
still led with Trysil. Six observations say the rule binds *unreliably*, which for
an ordering guarantee is the same defect. #2030 stands on that narrower
justification and now carries the corrected account.

Stopped at three because the decision is identical at 1-in-3 or 2-in-5. Say the
word for five (~$1.40).

**The probe found something better.** On the one correctly-ordered run the judge
scored Sequencing logic 1: "pli_002 at sequence 2, ahead of pli_001 at sequence
1." A rubric that inverts can't enforce the rule — a better argument for the
guard than "it doesn't discriminate". On #1404, not filed.

**The figure is dropped, not refreshed.** 93 of 94 confirmed, and two additions:
the log I cited isn't on disk, and _013's annotation *confirms* the 2 — so "a
human overturned it" was false too. Also nearly repeated your finding: my first
rewrite cited five gitignored scratch logs. Now one, verified tracked.

**#2110 sharpened.** You were right about #2014. Today's run went further — the
fixtures contradict each other: `collections-search-pennsylvania` serves 1999196
at personCount 893214, `collections-search-schuylkill` at 0. q7m queried
Pennsylvania and passed; 006 queried Schuylkill and failed. One of the two has
been wrong since June.

**Re-run: 18 pass, 3 fail, 1 xfail.** First attempt aborted on the environment
(breaker at 4-in-9, API time 1% of reported skill work, 2.9x concurrency on 7.9
GiB); --concurrency 2 ran clean, $0.46 wasted. r3d records xfail. Four failures,
four distinct causes — not five sharing one. q7m, 010 and 016 pass; 016 had been
red since 2026-08-24.

Annotation: 40 cells, three corrections. h4k Record type selection 3→2 (pli_006
typed `immigration` for a naturalization record — the judge read the rationale,
not the field, same as 007 last round). wzk Correctness and Tool Arguments both
1→3; that test now has no failing validator and no dimension below 3. The Georgia
fixture gap is #2212, filed not fixed — fixtures are snapshot-tracked.

**One process note:** don't edit a test file after buying its run. My
`xfail_reason` fix landed post-hash, and that field isn't in
`_COSMETIC_TEST_FIELDS`, so a doc edit reded rule 2 on a $6.75 run. Reverted to
the hashed text.

Thanks for the modify/delete resolution — keeping main's annotation was right.

Green on 44b9384b.

T-FEH (changes requested):
Third pass, on `48710f63` — your `44b9384b`, plus a merge of main and one commit of mine. Harness **3130 passed / 3 skipped**, engine **2828 passed / 120 files**, `tsc` clean, gate exit 0, ruff clean.

All five of my round-2 items are closed. Two things I missed are not, and one of them I fixed.

## The five are closed

The gate is green — you bought the re-run rather than arguing for the label, and `r3d` now records `outcome: xfail` against `expected_outcome: xfail`. The BOM is gone with no skip warning from the lint. The retracted "65 runs" survives nowhere in `eval/tests`, `eval/harness` or `packages`.

The `place_population` collision is resolved the better way: instead of escalating you measured that the fixtures were unconsumed and dropped them. I checked that was not a dodge — restoring both files reds the shape check with "fields the tool never returns: error, place_id", so they were genuinely wrong-shaped and dropping beat loosening the envelope.

And the semicolon evasion is closed.

## [Fixed in 48710f63] The same evasion class, one field over

I tested the *target* side last round — combined jurisdictions, case, whitespace — and never tested the *ahead* side. It keyed on `_INDIRECT_JURISDICTION in juris(i)`, so an item counted as blocking the target only if its jurisdiction literally contained "trysil":

```
lead jurisdiction "Trysil, Hedmark, Norway"   caught
lead jurisdiction "Norway"                    PASSED
lead jurisdiction "Hedmark, Norway"           PASSED
lead jurisdiction "Oslo, Norway"              PASSED
lead jurisdiction ""                          PASSED
jurisdiction key absent                       PASSED
```

Not hypothetical — `"Norway"` is one of the jurisdictions this scenario's own shipping run writes. So a plan leading with a national collection, a census, or any item the model labelled loosely would have been reported as correctly ordered.

The rule the objective states is that the target *leads*; what precedes it is not the point. The filter is now sequence alone and the message no longer names Trysil. `_INDIRECT_JURISDICTION` keeps its other two jobs. All six spellings fail, the correct plan still passes, the file's 9 direct tests stay green, and validators are not a snapshot input so no run log is invalidated.

## [Blocking] The `h4k` annotation correction rests on a false premise

The annotation and `44b9384b`'s message both say `naturalization` "is a recommended value in enums.schema.json, so a more precise type existed and was not used". It is not there:

```
record_type_recommended.examples =
  census, vital_record, probate, land, church, military, newspaper,
  cemetery, tax, immigration, court, other
```

`immigration` is — the value the model wrote. So the score was cut 3 → 2 for using the recommended value, on the premise that a non-recommended one was recommended. The same correction reasons the other way for `pli_009`/`pli_010` ("neither has a recommended value in the enum"), so the enum was consulted there and not here.

Annotations are calibration data, which is why this is worth stopping for rather than noting. Whether 2 is right on genealogical grounds is yours to say — but the enum sentence has to go either way, and if the score stands it needs a different reason.

## Two things for the record

**The run was scored against a superseded judge prompt.** The gate says so, warn-only:

```
scored against an older judge prompt (hash 03f306ff2512…); current is 9451d1056f0a…
```

`#2093` moved that hash on main the day before, and it changes what the judge sees — which is the ground the `wzk` corrections stand on. Not a blocker and not worth another $6.75, but it belongs in the body rather than only in a warning.

**The headline and the artifact disagree.** Line 177 says "18 pass, 3 fail, 1 xfail"; the run log says 17/4/1. The 18/3 reading counts `wzk` as a pass because the annotation raised it, which is defensible — but it should say so, and line 65's "17 pass, 2 partial, 3 fail" is the previous run's header left behind.

**`_002` still has no attribution.** It is a first-time fail in the retained corpus, on `fallback_for_in_same_plan` (`pli_013 → 'pli_006'`, a different plan). The body names the cause and stops. It matters here because the rule this PR adds reorders plan items and a cross-plan `fallback_for` is what reordering produces. `_006` I attributed myself to #2110's `personCount 0`; `_wzk` is judge-scored and annotated. One sentence on `_002` would finish the table.

**The `xfail_reason` still asserts a file that is not in the tree** — `v1_2026-08-28_19-50-30`, deleted by `10d6a3d8`'s own retention prune. You cannot fix it in place (`xfail_reason` is hashed; only `name`, `description` and `tags` are cosmetic), so it needs the `eval-cosmetic-skip` the gate's message offers. Carrying forward: a hashed field should not cite a filename the retention cap can delete.

## Requesting changes

Only on the annotation premise — that one is calibration data and a checkably false sentence. The rest is disclosure. Worth knowing before you merge: @EdmondOware's approval predates both paid runs and eight commits, including every fix under discussion.

Ikennaya1:
## Requesting `eval-cosmetic-skip` for one field — and it is not the usual reason

@T-FEH's round-4 review closed out everything actionable except one item that
cannot be fixed in place, and it needs a senior's call rather than another run.

**The problem.** `ut_research_plan_r3d`'s `xfail_reason` cites
`v1_2026-08-28_19-50-30` as its one committed observation. That file no longer
exists: the harness's newest-5-candidates retention deleted it during the
2026-09-02 run, in the same commit that added the replacement log. So a hashed
field now points at a filename that is not in the tree.

**Why it cannot just be edited.** `xfail_reason` is inside the snapshot —
`_COSMETIC_TEST_FIELDS` strips only `name`, `description` and `tags`. Editing
the sentence moves the hash and reds rule 2 on a run log that cost $6.75, which
is exactly what happened once already this week (10d6a3d8, reverted in
44b9384b).

**What is being asked.** The label on this PR so rule 2 downgrades to a warning
for the one commit that corrects the citation. The edit is a documentation
sentence inside a test's `xfail_reason`: it changes no input the skill or the
judge reads, no `expected_outcome`, no fixture and no scenario. It is
behaviour-neutral in the sense the label is documented for — the mechanical
snapshot is what objects, not the behaviour.

The alternative is a third paid `research-plan` run to re-hash a corrected
sentence, which is the wrong trade for the same reason
`eval-cosmetic-skip` exists.

**Carrying forward** — @T-FEH's point, and worth more than this instance: a
snapshot-hashed field should not cite a run-log filename that retention is
allowed to delete. `xfail_reason` will keep acquiring dangling citations as
candidates roll over. The durable references are the PR body and
`git show <sha>:<path>`; the corrected sentence uses the latter.

Not blocking merge on my side if the answer is no — the citation is wrong in a
way a reader can recover from, and the recovery command is in the PR body.

Ikennaya1:
All four verified. The blocking one was mine, and it is a bad one.

## The enum claim was false

You are right: record_type_recommended is census, vital_record, probate, land,
church, military, newspaper, cemetery, tax, immigration, court, other.
"Naturalization" is in the fact-type enum. My check was

    "naturalization" in json.dumps(schema).lower()

a substring match over the whole file, which collided with an unrelated key —
the third way CLAUDE.md says a check silently passes, and a shape I had flagged
in other people's work twice this week.

So the correction cut a score for using the recommended value, justified by one
that does not exist as a record_type. Reverted to 3 in dbabb0bd.

I did consider keeping 2 on `court` — naturalization declarations are county
court filings — and decided against it. Choosing between two recommended values
is not a partial failure, and substituting a new reason after the stated one
collapsed is defending the conclusion rather than following the evidence.
Annotations are calibration data, so a wrong 2 on a false premise teaches the
wrong lesson twice.

What survives is in the comment: the judge described the item as "Naturalization
(pli_006)" while the field reads `immigration` — it graded the rationale, not the
field, the same failure mode as _007 last round. Observation, not a score cut.

## The three disclosure items are in the body

**Judge prompt.** Confirmed independently: the log carries 03f306ff2512…, the
current prompt hashes to 9451d1056f0a…, moved by #2093 the day before. In the
body now, with the reason I am not re-running: the wzk corrections turn on the
test's own judge_context and on a declared tool parameter, neither of which the
judge prompt controls.

**Headline vs artifact.** Both numbers are now stated, with why the adjustment
goes one way: wzk is the only correction crossing a pass/fail line. Your catch
also exposed an inconsistency — I had counted wzk up while leaving h4k's 2 in
place. With h4k reverted, every annotated cell agrees with the judge except the
two wzk raises, so 18/3 is now internally consistent rather than selective. Line
65's stale header is marked as the superseded first run.

**_002.** It passed fallback_for_in_same_plan on all four prior retained runs,
including this branch's own 07-35-31 with step 4 item 7 already in SKILL.md. So
not attributable to the rule on the evidence — and not excludable either, at
five runs deep with no temperature pinning. The mechanism is a stale pointer
into the superseded plan (pl_003's pli_013 → pli_006), not a renumbering
artefact. All of that is in the body.

## The ahead-side evasion

Thanks for fixing it rather than just reporting it. I tested that filter from
the target side only and never from the other, and "Norway" is a jurisdiction
this scenario's own run writes — so the gap was reachable, not theoretical.
Sequence alone is the right key; the rule is that the target leads, and what
precedes it was never the point.

## xfail_reason

Requested eval-cosmetic-skip on the PR with the reasoning, including your
carry-forward: a snapshot-hashed field should not cite a filename retention can
delete. Not blocking on my side if the answer is no — the recovery command is in
the body.

Branch updated to main; I checked first that main had touched nothing
snapshot-tracked for research-plan, so the run log survived. 12 checks green on
dbabb0bd, gate exit 0.

Noted on @EdmondOware's approval predating the paid runs and eight commits.

Ikennaya1:
All four verified. The blocking one was mine, and it is a bad one.

## The enum claim was false

You are right: `record_type_recommended` is census, vital_record, probate, land, church, military, newspaper, cemetery, tax, immigration, court, other. "Naturalization" is in the fact-type enum. My check was

```python
"naturalization" in json.dumps(schema).lower()
```

a substring match over the whole file, which collided with an unrelated key — the third way CLAUDE.md says a check silently passes, and a shape I had flagged in other people's work twice this week.

So the correction cut a score for using the recommended value, justified by one that does not exist as a `record_type`. Reverted to 3 in `dbabb0bd`.

I considered keeping 2 on `court` — naturalization declarations are county court filings — and decided against it. Choosing between two recommended values is not a partial failure, and substituting a new reason after the stated one collapsed is defending the conclusion rather than following the evidence. Annotations are calibration data, so a wrong 2 on a false premise teaches the wrong lesson twice.

What survives is in the comment: the judge described the item as "Naturalization (pli_006)" while the field reads `immigration` — it graded the rationale, not the field, the same failure mode as `_007` last round. Observation, not a score cut.

## The three disclosure items are in the body

**Judge prompt.** Confirmed independently: the log carries `03f306ff2512…`, the current prompt hashes to `9451d1056f0a…`, moved by #2093 the day before. In the body now, with why I am not re-running: the `wzk` corrections turn on the test's own `judge_context` and on a declared tool parameter, neither of which the judge prompt controls.

**Headline vs artifact.** Both numbers are stated, with why the adjustment goes one way — `wzk` is the only correction crossing a pass/fail line. Your catch also exposed an inconsistency: I had counted `wzk` up while leaving `h4k`'s 2 in place. With `h4k` reverted, every annotated cell agrees with the judge except the two `wzk` raises, so 18/3 is internally consistent rather than selective. Line 65's stale header is marked as the superseded first run.

**`_002`.** It passed `fallback_for_in_same_plan` on all four prior retained runs, including this branch's own `07-35-31` with step 4 item 7 already in `SKILL.md`. So not attributable to the rule on the evidence — and not excludable either, at five runs deep with no temperature pinning. The mechanism is a stale pointer into the superseded plan (`pl_003`'s `pli_013` → `pli_006`), not a renumbering artefact of reordering. All in the body.

## The ahead-side evasion

Thanks for fixing it rather than only reporting it. I tested that filter from the target side and never from the other, and `"Norway"` is a jurisdiction this scenario's own run writes — so the gap was reachable, not theoretical. Sequence alone is the right key: the rule is that the target leads, and what precedes it was never the point.

That is three rounds out of four where the defect you found was a check I wrote and tested from one side only. Carrying that forward rather than just fixing this instance.

## `xfail_reason`

`eval-cosmetic-skip` requested on the PR with the reasoning, including your carry-forward — a snapshot-hashed field should not cite a filename retention is allowed to delete. Not blocking on my side if the answer is no; the recovery command is in the body.

## State

Branch updated to `main` — I checked first that `main` had touched nothing snapshot-tracked for `research-plan`, so the run log survived. Gate exit 0 after the merge, not only before it.

`pytest` is red on an infrastructure failure, not a test: `setup-uv` could not fetch its version data ("fetch failed"), so the job died before installing Python and collected nothing. Re-running it. The repo has no `uv.toml` or root `pyproject.toml`, so that action resolves "latest" over the network on every run — worth pinning if it recurs, but not this PR's to fix.

Noted on @EdmondOware's approval predating both paid runs and eight commits.

T-FEH (approved):
Fifth pass, on `89b8ca1f0` — `dbabb0bdd` plus a clean merge of main and one commit from me. Harness 3181 passed / 3 skipped, engine 2838, `tsc` 0, `check_runlogs` green.

## The retraction is the right shape

`6a5da160d` removes the correction rather than re-justifying it, and says why in the annotation itself:

> substituting a new reason to keep the same score after the stated one collapsed would be defending the conclusion rather than following the evidence

I checked the retraction's own claims instead of taking them. `record_type_recommended.examples` is exactly the twelve values named, `naturalization` is not among them, and the enum is explicitly open besides — so the original cut was wrong twice over: the value used was recommended, and an unrecommended one would still have been legal. Naming the mechanism (a substring match over the whole schema file colliding with an unrelated key — CLAUDE.md's third way a check silently passes) is what makes this reusable rather than just corrected.

## What I verified

Both headline figures reproduce: the log records 17/4/1, there are exactly two score-changing corrections, both on `wzk`, so the annotated 18/3 is arithmetic rather than assertion. The `007` `fallback_for` correction holds — zero non-null across all 27 item objects. `place-search.ts:41` does declare `contextName`, and there is no `place-search` Georgia fixture despite the test declaring `collections-search-georgia`, so that correction is a corpus gap correctly charged back. The superseded judge-prompt hash and the four ungraded tests are both already disclosed in the body with the hashes, which is the honest handling. #2030 does carry the corrected account — in a comment, so a body-only check misses it.

The lane rule is applied correctly: prose kept, validator added, and the writer-tool precondition filed as #2030 rather than left as an intention.

## Two things I fixed

**The body still shipped the retracted sentence.** The annotation said 3 → 3; the body still said `3 → 2` and "`naturalization` is a recommended value in `enums.schema.json`", under a "three corrections" heading where there are two. Retracting in one artifact and keeping it in the other is the failure the retraction commit itself warns about, and it matters more here because this body is load-bearing evidence by design. Corrected: the bullet now records the retraction, the count reads two.

**My own fix from round 4 was unpinned** — worth stating plainly, since it is the same defect I have flagged on five other PRs. Reverting `48710f631` exactly:

```python
ahead = [... < target_seq and _INDIRECT_JURISDICTION in juris(i)]
```

left all 27 tests green. The semicolon evasion was pinned; the bare-place one was not, so the recurring defect class was sitting inside the fix for it. `89b8ca1f0` adds the missing case — a leading item in a bare `"Norway"`, which is not contrived: the shipped run writes that literal among its 21 distinct jurisdictions. Under the reverted validator that test alone reports `DID NOT RAISE AssertionError`; restored, 28 pass.

## Three things that must land before merge — two need a senior, not you

**`xfail_reason` still asserts a file that is not in the tree.** It says `v1_2026-08-28_19-50-30` is "committed here". It is not. This was my round-4 item and it is still open — but you cannot fix it unilaterally: `xfail_reason` is outside `_COSMETIC_TEST_FIELDS`, so editing it reds rule 2 on a $6.75 run. It needs `eval-cosmetic-skip`. There is now a committed `r3d` log it could truthfully cite instead.

**The scenario README has two factual errors.** `README.md:10` calls `q_001` `in_progress`; `research.json` says `open`. `README.md:57` cites "the two `place_population` calls that genuinely returned `{"error":"Place not found"}`" — those fixtures were dropped in review, ten ship, and no fixture in the repo returns that string. That one actively misleads, because the paragraph's job is to vouch for the fixtures being faithful extractions. I wrote both fixes, watched the gate go red on `hansen-kongsberg-baptism/README.md: content-differs`, and reverted them — the scenario directory is a snapshot input, so this is the same trap your own body documents under "do not edit a test file after buying its run". The same `eval-cosmetic-skip` covers all three edits together.

**The encoding-lint gap is deferred without a card.** The body says it "wants its own `nothing-checks` card and is not this PR's to fix", and the gap is real, but no issue exists. CLAUDE.md is explicit that the card is filed in the same PR that defers it — a limitation described precisely enough to write down is debt minus the tracking otherwise.

## Approving

The measurement discipline here is the best I have reviewed on this repo: a confound found in your own headline claim, re-measured rather than reworded, and the conclusion narrowed to "binds unreliably" instead of being defended. Worth flagging for whoever merges: six of the six observations behind that table rest on gitignored scratch logs, so exactly one instance of the ordering defect is verifiable from the branch. That does not change the decision — #2030 is filed either way — but it means the table is taken on trust, and the body should say so.

Two notes on the gate rather than the code: @EdmondOware's approval predates both paid runs, the regex fix, the widening and the retraction, and `dismiss_stale_reviews_on_push` is false, so it still counts. And CODEOWNERS routes `.py`, `.json`, `/eval/` and the skills directory to both senior teams, so this still needs both.

PR #2044, opened 2026-08-30
image_transcribe: default to Gemini 3.7 Flash, retire the Opus re-reader, and retry a transport failure once

### LINKED ISSUE #1594: image_transcribe on hard scans: fabricated record types, corrupted patronymics, and 26% of calls lost to reachability

### LINKED ISSUE #2013: image_transcribe: escalate a failed Qwen read to a better model in-tool, and retire image-reader-opus
**Touches:** packages/engine/mcp-server/src/tools/image-transcribe.ts, packages/engine/mcp-server/src/auth/config.ts, docs/specs/image-transcribe-tool-spec.md, packages/engine/plugin/agents/image-reader.md, packages/engine/plugin/agents/image-reader-opus.md (delete), packages/engine/plugin/skills/search-images/SKILL.md, packages/engine/plugin/skills/record-extraction/SKILL.md, docs/specs/image-reader-opus-agent-spec.md (delete)

The default OCR model (`qwen/qwen3-vl-235b-a22b-instruct`) cannot read hard hands,
and the escalation path we built for that case is unreachable in practice. This
issue moves the escalation into `image_transcribe`, and retires
`image-reader-opus`, whose only reason to exist was to be that escalation.

**Do not start this before reading "Measure these four things first" below.** Two
of the four could change the design, and one could kill it outright.

## The problem, in one line each

- Qwen garbles hard hands. Measured on the #1791 tester's Luxembourg 1819 register
  (`ark:/61903/3:1:3QS7-897J-Q396`, 1.93 MB): **72 `[illegible]` markers and a
  fabricated duchy name** ("Groß-Herzogthum *Echemburg*" for *Lützemburg*).
- The escalation to a better model **already exists as prose at three sites** and
  **has never fired**. `agents/image-reader.md:85-91` tells the reader to detect
  "substantially `[illegible]`" and emit a `HARD SCAN — …` prefix;
  `search-images/SKILL.md:182` and `record-extraction/SKILL.md:82` tell the caller
  to escalate on that signal. On the tester's session the prefix appeared **zero
  times** on a page carrying 165 `[illegible]` markers.
- The escalation target is `image-reader-opus`, which reads through `image_read`,
  which refuses over 700 KB (`image-read.ts:14`) — **80% of a 15-ARK sample of our
  own corpus** ([measured](https://github.com/PioneerAIAcademy/cowork-genealogy/issues/1593#issuecomment-5399863651)).
  So even when the prose does fire, the escalation is refused on most real scans.

Net: there is no working higher-accuracy path for a large hard scan, and the one
we ship is unreachable from both ends.

## Why this is a tool change, not a prose fix

CLAUDE.md's lane rule, and ADR-0011's first question: *can this be decided by
reading the documents alone?* Counting `[illegible]` markers against transcription
length is arithmetic on a string the tool already holds. It is a writer-tool
precondition, not a judgement.

The precedent is measured, not asserted: converting `proof-conclusion` to a
skill-agent pair found **five behaviours stated correctly in the body, read, and
not followed**; each held on the first run after moving into the tool
(`docs/skill-lifecycle.md` §5). Issue #1007 measured ~50 lines of triage doctrine
in `search-records` being complied with in **3 runs out of 4**. Issue #607 settled
the same question the same way and its ruling is explicit: *"prose does not
reliably change behaviour... A tool-side expansion cannot decay and costs no paid
eval run. Do not re-open this."*

We already shipped the prose version of this escalation. It does not work. Shipping
better prose is the option we have already tried.

### 1. Auto-escalation inside `image_transcribe`

After the Qwen read, compute a quality signal. Above threshold, re-run **once**
against the escalation model, return the better result, and report both attempts.

Signals available today, no new plumbing:

- **`[illegible]` density** — markers per unit of transcription length. The marker
  is a contract the prompt explicitly asks for (`buildOcrPrompt`,
  `image-transcribe.ts:92-96`: *"Mark anything you cannot read [illegible] — never
  guess"*), so it is a requested output, not an artifact.
- **Degenerate output** — PR #1921 recorded a **17× repetition loop** on one Elena
  page, a failure mode #1594 does not list and nothing currently detects.
- **Transcription length against `sizeBytes`** — already in the result metadata. A
  1.93 MB dense register returning 3,920 chars is suspicious in a way a small clean
  page returning 3,920 chars is not.
- **`found: "NOT FOUND"`** when `lookingFor` was supplied — already parsed
  (`parseFound`, `image-transcribe.ts:143`). Weak, free, and only usable as a
  tiebreak.

The empty-transcription case is **already** handled exactly this way
(`image-transcribe.ts:248-253` throws rather than returning), so this extends an
existing precedent rather than inventing one.

### 2. A caller hint, not a model selector

Add an input that **lowers the escalation threshold** — not one that names a model.
The caller genuinely knows things the tool cannot ("the index says this page has a
marriage entry and the transcription has none"), so the channel should exist. But a
direct `model` parameter would let a skill route to Opus without ever trying Qwen,
which is the cost failure we must not ship. Making it a threshold hint closes that
structurally instead of by asking the prose nicely.

### 3. Cost ceilings in the tool

At most **one** escalation per image, and a per-project escalation budget. The
shape already exists: `browseBudgetSeen` (`image-transcribe.ts:30`) is a
module-level, process-lifetime, project-keyed `Map` doing precisely this for the
browse budget. Follow it, including its `__clearBrowseBudgetForTests` reset hook.

Report the escalation in the result. `metadata.model` is already returned
(`image-transcribe.ts:286`), so the channel exists — a caller (and any run-log
reader) must be able to see that an escalation happened and what it cost.

### 4. Delete `image-reader-opus`

Its entire purpose is to be the higher-accuracy re-read. Once `image_transcribe`
escalates on its own, the agent is a strictly worse-covering duplicate: it is
capped at 700 KB where the tool has no cap, it must be invoked by prose that
demonstrably does not fire, and it reads through a tool that refuses 80% of real
scans.

**Delete only after §1 is measured to work** (see the Opus-path equivalence probe
below). Sites, verified against the tree:

- `packages/engine/plugin/agents/image-reader-opus.md` — delete
- `docs/specs/image-reader-opus-agent-spec.md` — delete
- `packages/engine/plugin/skills/search-images/SKILL.md:182-186` — the escalation offer
- `packages/engine/plugin/skills/record-extraction/SKILL.md:82-88` — the escalation offer
- `packages/engine/plugin/agents/image-reader.md:85-91` — the `HARD SCAN` prefix
  instruction (the tool now owns this detection; leaving the prose in place ships
  two detectors that can disagree)
- `eval/harness/harness/allowed_tools.py`, `context_policy.py`,
  `skill_invocation.py`, `e2e/orchestrator.py` — harness references
- `eval/harness/tests/unit/test_allowed_tools.py`,
  `test_e2e_agent_tool_usage_report.py`
- **`SUBAGENT_ONLY_TOOLS` becomes vacuous for `image_read`** —
  `context_policy.py:79` is `frozenset({"image_read", "extraction_append"})`, and
  `allowed_tools.py:177` records that *"`image_read` lives only on
  `agents/image-reader-opus.md`"*. Delete that agent and the main-thread deny for
  `image_read` still exists but guards a tool no agent declares. **Do not silently
  drop it** — #1874 is open precisely because the main-thread deny set is too
  narrow, and #1295 was filed when this guard hardcoded one member. Decide
  explicitly whether `image_read` stays in the set (defensible: the Issue #28
  consumer survives) and say which in the PR.
- `packages/engine/mcp-server/tests/packaging/agent-tool-names.test.ts`,
  `doc-links.test.ts`
- `docs/architecture.md`, `docs/skill-dataflow.md:116`, `docs/specs/e2e-test-spec.md`,
  `docs/specs/image-reader-agent-spec.md`, `docs/specs/image-read-spec.md`,
  `docs/deep-dives/search-images-prohibition-list.md`,
  `docs/e2e-run-latency-findings.md`, `docs/plan/image-read-context-policy.md`,
  `.claude/skills/interpret-e2e-result/SKILL.md`, `README.md`

**Do NOT delete the `image_read` tool itself.** Two consumers survive the agent:
`docs/specs/image-read-spec.md:220-223` names the Issue #28 OCR-comparison pipeline
as its primary downstream consumer, and `skills/research/SKILL.md:145` still names
`image_read` in the `search-images` routing chain (that line is arguably already
wrong — `search-images/SKILL.md:164` tells the skill it does **not** have
`image_read` — so fix the routing line rather than the tool). Retiring the tool is a
separate decision; this issue retires the *agent*.

## Measure these four things first

Each of these can change the design. The first can kill it.

### A. The 60s Cowork device-bridge ceiling — do this one first

Issue #1638 established that the Cowork device bridge **aborts every MCP call at
60s**, a client-side ceiling this repo does not set and cannot change
(`image-transcribe-tool-spec.md:354-373`). Florence measured Qwen at 33s and Opus
at **71s** on the tester's page. A sequential in-tool escalation is therefore
~104s and **guaranteed to be aborted in Cowork**, where our users are.

If that holds, §1 as written cannot ship on the Cowork path, and the escalation has
to be a second tool call driven by a tool-computed flag — which is the *shape* of
option 2, but triggered by arithmetic rather than by model judgement, so it keeps
the property that matters. **Design this fork before writing code.**

Test with a known-hard image, per the corpus below. Note that neither eval harness
can observe this — both register over stdio and are uncapped (#1638) — so this is a
live Cowork read in both run modes (on-computer and cloud), which per CLAUDE.md are
a per-task setting nothing in the plugin can see.

### B. Is Opus-via-OpenRouter equivalent to Opus-via-`image_read`?

`image-reader-opus` gets Opus's **own vision** on bytes returned inline;
`image_transcribe` would get Opus **through OpenRouter**. Same model, different
path, and nobody has compared them. If the OpenRouter path reads materially worse,
deleting the agent loses capability rather than consolidating it. If it reads the
same or better, the deletion is clean. Florence's Opus-5-via-OpenRouter numbers
(below) are the only half we have.

### C. Threshold calibration — and what it cannot catch

**Nothing checks a transcription today.** #1594 states it flatly: no test, fixture,
or rubric compares an `image_transcribe` output to its scan, and Phase 0's grading
corpus is not in the repo. A threshold has nothing to calibrate against until a
graded set exists — see the corpus section, which is better news than that sounds.

**State this limit in the PR:** `[illegible]` density catches #1791 and #1708. It
does **not** catch #1594. The Västra Karaby corruption — `Henckelstorp` →
`"Hornekelltorp"`, `Andersson` → `"Olsensfon"`, `Barsebäck` → `"Dansebäck"` —
returned clean-looking text with **no markers at all**. Confident garbage is
invisible to a density threshold and is the more dangerous failure, because
`docs/gps-research-flow.md` treats a patronymic mismatch as a core-identifier
conflict, so the OCR silently manufactures that class of conflict. Do not let this
issue be closed as fixing #1594.

Note also that `temperature: 0` is already set (`image-transcribe.ts:198`), so
#1594's *nondeterministic* corruption (same image, `Andersson` → `"Olsensfon"` in
one call, absent in another) is not sampling noise — it is model instability. That
strengthens the case for a better model over a retry, and it contradicts
`agents/image-reader.md:80-84`, which tells the reader "the OCR is deterministic
for a given scan, so a re-call returns the same text". That line is measurably
wrong and should be corrected in the same PR.

### D. Re-probe the spec's model numbers before picking the escalation model

`image-transcribe-tool-spec.md:35-36,108` records **Sonnet 5 76% / Qwen 67% /
Sonnet 4.6 60%** on hard hands including German Kurrent, and concludes "German is no
longer an open item." Florence measured Sonnet 5 buying **4 fewer `[illegible]`
markers than Qwen and zero substantive fields** on the tester's page, while Opus 5
recovered the date, registrar, mother, witnesses and time.

Per CLAUDE.md ("A measurement that disagrees with belief is re-measured, not
reworded"), re-probe before rewording — and it matters concretely here, because 76%
is the number someone will reach for when choosing the escalation model. **The
measurement supports Opus, not Sonnet 5.** `measured-figures.test.ts` guards this
wording; expect it to fire.

## Difficult-image corpus — most of this already exists

**`dev/try-ocr-compare.ts` already carries an 11-image graded corpus** with a
language/script spread, plus `dev/try-ocr-grade.ts`, a re-runnable **field-level
grader** (names / dates / places / relationships, with an explicit `hallucinations`
counter) that scores against an Opus ground-truth key and is text-only, so the
Claude-family home-field bias is confined to the key rather than the scoring. This
is the calibration instrument §C needs, and it does not have to be built.

`dev/ocr-spike-out/` is gitignored and absent locally, so the transcriptions must
be re-run — but the corpus definition, the prompts and the grader are all committed.

Hard subset (`subset: "hard"`), from `try-ocr-compare.ts:66-185`:

| slug | language / script | id |
|---|---|---|
| `german-kurrent` | German Kurrent, 1854 Taufregister spread — the headline domain | `ark:/61903/3:1:3Q9M-CSQR-S57W` |
| `jan-gallo` | Slovak/Hungarian/Latin Lutheran register, "very hard" | `ark:/61903/3:1:33S7-9RQ4-9B4M` |
| `birkeland-death` | Gothic Norwegian (Kurrent-style) church book | `ark:/61903/3:1:3QHK-93PP-TCSM` |
| `teitje-harkema` | old Dutch hand, 1833 | `004748896_00921` |
| `cruz-corona` | handwritten Spanish, ~1.9 MB | `3:1:33S7-9T8R-68V` |
| `zuniga-rojas` | handwritten Spanish, ~1.7 MB | `ark:/61903/3:1:33SQ-GY3D-9KYN` |
| `elizabeth-geach` | handwritten English | `ark:/61903/3:1:33SQ-GYYJ-S99H` |
| `reese-siblings` | handwritten English, ~2.0 MB | `3:1:33S7-9RK9-BPN` |

Three printed-form `subset: "control"` images are in the same file
(`spriggs-census`, `rossi-marriage`, `bottemiller-census`) — keep them, since a
threshold that escalates on clean pages is the expensive failure mode.

Add from the issues, since they carry known-correct answers the spike corpus lacks:

- **`ark:/61903/3:1:3QS7-897J-Q396`** — Luxembourg 1819 civil birth register, 1.93 MB.
  **The acceptance test.** Indexed answer to check against (19 July 1819, Obercorn);
  Florence's per-field Qwen/Sonnet-5/Opus-5 table is on #1593. Note Opus reads the
  canton as *Messancy*, unresolved — the page carries four entries.
- **`ark:/61903/3:1:3Q9M-CS4T-KT54`** — Amite County conveyance, 1,696,175 bytes (#1708).
  Transcribed cleanly first try; the value is that it is >700 KB, so it exercises the
  cap boundary and the `ORIGINAL NOT EXAMINED` failure.
- **`004523018_00150`, `004523021_00100`, `004523021_00190`** — Västra Karaby, the
  #1594 confident-garbage class, with ground truth read directly. These are the ones
  a density threshold will **miss**, so they belong in the set precisely to bound
  the claim.

For §A specifically, use `german-kurrent` or the Luxembourg ARK: both are known to
push Opus past 60s.

## What verifies this, and what does not

- **`make engine-test`** proves the threshold arithmetic and the request shape. It
  proves nothing about transcription quality.
- **Neither eval harness can catch the §A failure.** Both register the server over
  stdio and are uncapped (#1638), so a green e2e run is guaranteed not to exercise
  the 60s ceiling. Name in the PR which run modes you actually tested live: Cowork
  on-computer, Cowork in the cloud, the hosted web workbench.
- **`dev/try-ocr-grade.ts`** is the quality instrument. Report before/after
  field-level scores and the `hallucinations` count on the hard subset.
- **The acceptance test** is `ark:/61903/3:1:3QS7-897J-Q396` returning the fields in
  #1593's table without a caller having to ask for anything.

## Cost

Editing `agents/image-reader.md` and both SKILL.md files arms
`.github/workflows/check-runlogs.yml` for `record-extraction` and `search-images` —
roughly **$8–12 and 45–65 min per skill**, plus grading. `record-extraction`'s slot
is the deepest queue on the board (#1848) and is currently held by #1863/PR #1909.
**Batch with whatever run #1546 or #1824 buys** rather than paying separately; #1546
in particular is already redesigning `image-reader`'s return contract, and a
tool-reported escalation is a second thing riding that same successful-read channel.
Landing these independently buys two paid runs and risks a merge conflict in the
same file.

## What this does and does not close

- **Closes:** the #1791 tester's case; the #1708 unread-original failure.
- **Does not close:** #1594 (confident garbage, see §C); #1593's 700 KB cap, which
  stays open and still blocked on an unmeasured transport ceiling — this issue
  deliberately routes around it rather than raising it, since `image_transcribe`
  puts no bytes on the MCP transport.
- **Still owed to the tester** (#1791): once this lands, re-run their page and send
  them what it recovers.

## Why `nothing-checks`

Nothing today reports that a transcription was unusable. A run can be green, cite
its sources, and have read a page that came back as 165 `[illegible]` markers and a
fabricated place name — and the escalation built for exactly that case fires zero
times with nothing recording that it did not.

DallanQ:
## Summary

- Switches the `image_transcribe` default OCR model from Qwen3-VL to Gemini 3.7 Flash. Qwen reads handwritten pages at 50% field accuracy with 37 hallucinations; Gemini reads at 79% with 8, and wins on every handwritten page in the corpus by 10 to 47 points. Cost goes $0.09 → $0.58 per 40-transcribe research run.
- Retires the `image-reader-opus` agent, and deliberately does **not** replace it with the auto-escalation that was planned. Normal tier.
- Adds one bounded retry on the OpenRouter transport path, which is the half of the transcription-failure work that a model swap does nothing for.

Closes #2013
Closes #1594

## Review guidance

**Start here:** `src/tools/image-transcribe.ts`, the retry loop around the OpenRouter fetch. The decision worth checking is that it retries a **transport** failure but never a timeout. That split is the whole argument: a connect-level failure never reached OpenRouter, bills no inference and returns fast, while a timeout has already spent the 180s budget and a second attempt doubles the worst case — which is the objection that kept a retry out of the tool until now. If you think a timeout should retry too, that is the line to argue with.

Second, `docs/specs/image-transcribe-tool-spec.md` — the Phase 0 record is kept and marked superseded rather than rewritten, because its method is the interesting part of why it was wrong: it compared models only against another model's transcription, on a corpus with no human-verified page.

**Unsure about:** whether retiring the Opus agent should have waited on an Opus-via-OpenRouter versus Opus-via-`image_read` comparison, which the plan asked for and I did not run. I judged it moot — the agent is being deleted rather than consolidated into the tool, so there is no path whose fidelity we are relying on. If you disagree, that measurement is the thing to demand before merge.

## Plan

**Didn't change:** the `image_read` tool itself, and its main-thread deny in the harness. No agent declares `image_read` any more, so the deny now guards a tool nothing holds — I kept it deliberately. It is what stops inline base64 reaching the main thread, and that hazard is unchanged by the agent's removal; dropping it would quietly reopen the main session to inline scans. Also unchanged: the 700 KB inline cap, and the parked user-invoked Opus transcription idea.

**Acceptance check:**

- `tests/tools/image-transcribe.test.ts` — three new tests fail on `main`: the retry recovering a transport failure, the retry stopping at one attempt, and a timeout **not** retrying. Proven in both directions: disabling the retry fails the two recovery tests, and letting timeouts retry fails the budget test on call count.
- `tests/unit/test_allowed_tools.py` — the two `image_read` union tests were inverted to assert the grant is gone, with a positive assertion first so they cannot pass on an empty set. Proven by restoring the agent and one caller reference: `image_read` reappears and both go red.
- Quality, on the three Västra Karaby pages the transcription-fidelity issue documented: accuracy 33% → 70%. The fabricated record type is gone — on the 1693 accounts page Qwen returned a births heading `Fædd.` and the year 1673, Gemini returns `DEBET.` and 1693 and beats the Opus key, which misread the title.

**Deviated from the plan:** three, all deliberate.

1. **The auto-escalation is not built.** The plan was to keep Qwen and escalate on `[illegible]` density. That trigger is high-precision and low-recall — blind to the fabrication class, which is the one that writes a wrong parent into a tree — so it would have advertised a safety property it does not have. Raising the floor replaced it. The lead confirmed no-escalation as the resolution.
2. **The 60s device-bridge measurement was named as the thing to do first.** The benchmark already answers it: every model averages under 4.5s per page and no call exceeded 60s. It does not bind.
3. **The retry shipped without the host-vs-provider classification** the issue sequenced it behind. The classification is still genuinely open — 0 of 30 committed reachability failures carry a socket cause code, since every one predates the change that surfaces it — but it does not change the action. Both hypotheses are "intermittent", and a bounded retry is right either way.

## Test plan

- [x] Both skills whose run-log snapshot changed were re-run and their run logs and `.ann.json` committed: `record-extraction` and `search-images`.
- [x] No skill `description` frontmatter or DO NOT clause changed, so no routing-pair negative tests were needed.

## Follow-on issues

**Folded in / filed instead:** Folded in — two defects found while reviewing the benchmark scripts on the merged measurement PR. A narrowed re-run replaced `meta.variants` while merging images forward, so the grader silently dropped every other model from the scorecard for every image; and `--limit abc` parsed to NaN, selected zero images and exited 0 having done nothing. Both are fixed here rather than filed, since the branch was already open on those files. Nothing was filed.


EdmondOware:
The `runlogs` gate has been red on `main` since this merged yesterday evening, and it is now the only thing blocking two PRs I have approved.

Four sampled corrections in this run's annotations carry no comment, and all four are confirmed failures rather than confirmed passes, so the rule will not exempt them:

```
record-extraction  ut_record_extraction_005  base / Correctness                 judge 2, you 2
record-extraction  ut_record_extraction_005  rubric / Informant identification  judge 2, you 2
search-images      ut_search_images_007      base / Correctness                 judge 1, you 1
search-images      ut_search_images_007      base / Completeness                judge 1, you 1
```

Both files list you as annotator. It is four sentences across two tests, on `main` rather than on either branch.

The two PRs waiting on it are #1955, where the registrar doctrine correction is done and `eval-cosmetic-skip` has cleared the snapshot half, and #2018, the record extraction deep dive. Neither author can reach the problem from their branch, and no label helps: `eval-cosmetic-skip` relaxes rule 2 only, and the script says so in its own docstring.

Issue #2037, opened 2026-08-29
eval/app hosts a security control that no required check tests


aghadiayeamayanvboernest:
**Touches:** `.github/workflows/js-tests.yml`, `protect-main` ruleset, `eval/app/middleware.ts`

`eval/app` now hosts a security control and no required check tests it.
`eval CRUD UI` is not among `protect-main`'s eight required checks and is
path-gated, so every guard in that job can go red while `main` merges.

The fix is a ruleset change, which is the lead's call.

This issue originally bundled a second gap: `middleware.ts`'s `config.matcher`
was asserted by nothing, so a matcher typo would disable the control on every
route with a green build. **That half is closed** — it never depended on the
ruleset decision. See "The matcher half is closed" below; what remains here is
only the required-check half.

## How this was established

Raised independently by two reviewers on PR #1999 and again on PR #2000. The
required list was read directly from the ruleset: `pytest`, `runlogs`,
`e2e-fixtures`, `vitest`, `lockfile-drift`, `scan`, `pnpm workspace`,
`server-pytest`. `eval CRUD UI` is not on it.

It lands harder on #2000, whose entire content is a path-containment control
living in that same job — ten containment tests, none of which can block a merge.

## Why it needs an issue rather than a code fix

Making a job required is a repository ruleset change, not a branch edit. Filed so
the gap is visible in the `nothing-checks` register, which is read by label
rather than by text search — until this exists, the gap is invisible to anyone
who goes looking there.

## The fix is cheaper than this issue first implied (added 2026-08-30)

Filed saying only "a ruleset change, the lead's call", which reads as a bigger
decision than it is. @chesworthrm measured the part that makes it cheap, on the
#2000 review:

**`js-tests.yml` has no `paths:` filter.** Its trigger is a bare
`on: pull_request:` — verified directly. The job therefore runs on *every* PR,
and its scope step reports success without doing anything when no `eval/app`
path changed. Confirmed live on #2017 and #1941, neither of which touches
`eval/app`; both show `eval CRUD UI` green.

So adding `eval CRUD UI` to `protect-main`'s required list **cannot block an
unrelated PR**. It is a one-line ruleset change, not a workflow redesign.

That removes the objection this issue would otherwise raise on its own — that
requiring a path-gated job would stall every PR that does not touch its paths.
It would not.

Still the lead's call, since only he can edit the ruleset.

## The matcher half is closed (2026-09-01)

[...]

PR #2049, opened 2026-08-31
Constrain the destination of links the UI presents as FamilySearch (#1018)

aghadiayeamayanvboernest:
Part of #1018. Constrains where a link can go when the UI says it goes to FamilySearch.

## The problem

The generic outbound-link channel authorizes on scheme alone. That's correct for a link whose visible text **is** its URL — the reader can see the destination. It's wrong for a button labelled "Open in FamilySearch", where the value comes from agent-authored project data and the label asserts a destination the data doesn't have to honour.

## The design, and the shape I rejected

**Rejected: parse the URL, allow-list the host.** Every one of these sinks carries an *identifier*, not a host — so accepting a host at all invents the risk it then has to test for. Suffix confusion, prefix confusion, and protocol-relative inputs are all downstream of that one choice.

**Adopted: reduce the input to a FamilySearch identifier, rebuild the URL from a compile-time constant.** No caller-supplied host reaches `shell.openExternal`.

**By concatenation, never `new URL(value, base)`** — the URL constructor resolves `//evil.example/x` against a base to a *foreign origin*. That would reintroduce the whole bug behind an API that looks careful. Mutating the implementation to use `new URL` reds the ACCEPT tests — not the protocol-relative ones, which `//evil.example/x` never reaches a return site to exercise. What rejects a protocol-relative input is the `^` anchoring on the patterns; the concatenation is what keeps a *matched* value from being re-resolved against the base. See the verification table for exactly what reds.

## Four accepted shapes, read from fixtures rather than invented

| Input | Fixture |
|---|---|
| bare ARK | `patrick-flynn.ts` |
| bare type-prefixed id | engine `ark.ts` grammar |
| ARK resolver URL, with **and without** `www` | `patrick-flynn-sidecar.ts:20`, `patrick-flynn.ts:553` |
| `/tree/person/<pid>` — **not an ARK at all** | `patrick-flynn-sidecar.ts:72` |

My first plan proposed accepting "a bare ARK or an allow-listed host". That would have rejected two of these, including the only fixture reaching the tree-match sink.

## Scope — please don't read a four-sink blast radius into this

Four sinks are constrained. **Only one is reachable today.** `TreeMatch` is `{treePersonId, stars}` in the engine, so the schema's `ark?` has no producer; `arkUrl` was renamed to `recordId` and survives only in July run logs. The other three are fixed because the schema declares the fields and a producer can appear — not because they're exploitable now.

**Three sinks deliberately keep the generic channel**: `Linkify`, `SourcesSection` (two fields), and the research-log external-site link all render the URL as its own visible text. Constraining them would break external-site search and buy nothing. A test pins the split in both directions.

## Where it's enforced, and why the module exists

Main process, where the privilege is. A renderer-side check is advice.

The policy sits in its own module because **nothing can import `main/index.ts` in a test**: `setupIPC` is module-private, `app.whenReady()` runs at module scope, and the file imports `icon.png?asset`, which the vitest config has no plugin to resolve. Extracting it is what makes the check testable at all — my first acceptance criteria were unrunnable and a plan review caught it.

`apps/web` is fixed too. Its check is renderer-side and weaker, and **the code says so** rather than implying parity — but leaving the hosted viewer unconstrained while calling the policy "enforced" would be a claim the code doesn't support.

## Two failure modes the design forces into the open

**The transport method is required, not optional.** It broke five test doubles at compile time — that's intended. Optional would let a transport omit it and every constrained link fall back to the unconstrained default.

**The `ResearchDataProvider` binding is the silent-no-op hazard.** `openFamilySearch` fails closed and Electron denies `window.open`, so without that one line every constrained link would do nothing at all, with every test green.

A sixth double turned up that neither the plan nor my grep found — `test/setup.ts` mocks the *preload* API rather than the transport. Typecheck caught it.

## Verification

`make test-all` — all checks passed. 34 new tests: 31 across five new files (15 policy, 4 handler, 3 preload pairing, 3 web policy copy, 6 call-site routing) plus 3 added to existing transport and provider suites.

Mutated, not assumed:

| Mutation | Result |
|---|---|
| Widen `FS_URL_PREFIX_RE` to `^https?:\/\/[^/]+\/` (accept any host) | 3 red |
| Replace the three `return FS_BASE + …` sites with `new URL(…, FS_BASE).href` | **5 red**: four ACCEPT tests — the bare ARK, both ARK resolver URLs, and the image ARK — plus `never returns a URL outside the FamilySearch base`. **Not** `accepts a bare type-prefixed id` and **not** the handler test: both take the `n:n:` path, which returns from a *template literal* (line 72), not one of the three sites this mutation edits. **Not** the protocol-relative tests either — `//evil.example/x` matches no pattern and reaches no return site; the `^` anchoring rejects it, not the concatenation. `accepts a /tree/person/ URL` also survives, since that suffix resolves relatively to the same result |
| Replace the pure body with a pass-through (**web copy**) | 3 red |
| Flip `PersonCard` back to the generic channel | 1 red, named |
| Bypass `openFamilySearch` in the **web transport** | 1 red |
| Remove the **provider binding** | 1 red |
| Point the **preload** at `'open-external'` instead of `'open-familysearch'` | 2 red — the pairing assertion and the generic-channel guard |
| Annotate the IPC stub `AppAPI` instead of casting | typecheck reports 2 missing properties |

Also dropped the raw `href={person.ark}`: ctrl/middle-click follows an href **without** firing `onClick`, bypassing the policy entirely. That only bites when `ark` holds an `https://` value — which is the poisoned case, and what the fixtures contain. `title` keeps the hover disclosure.

Ran the `security-invariants` skill, as `apps/electron`'s CLAUDE.md requires for any IPC change. §1, §3, §4, §5, §6, §7 and §9 re-checked. §3 **failed** until the new channel was added to the skill's table — that skill treats a preload channel absent from the table as unauthorized — so this registers it there and in the §4 validation list.

### A note on this table

One row in it has now been corrected four times, each time by a reviewer: a
wrong count, a wrong cause, a stale count, and a description of a mutation
different from the one I ran.

I had written a retrospective here explaining that history. It contained two
more wrong claims about the same number, so it is deleted rather than corrected
a fifth time. The row above is measured and correct; that is what matters, and a
narrative about past errors that keeps adding new ones is worth less than
nothing.

The count is not the evidence. What the mutation demonstrates is — so the row
names the mutation, what reds under exactly that mutation, and what survives it
and why.

### Fixed after review, round 2

Two more guards had no test, each confirmed by breaking it and watching the suite
stay green:

- **The web transport's call site.** `apps/web` is the only consumer of the web
  policy copy. A bare `window.open(value, …)` there passed everything — the same
  hole as round 1, one file up. `link-channel-routing.test.ts` reads
  `src/components` only, so it could not see a transport.
- **The provider binding.** Named as the silent-no-op hazard in this body and
  then left uncovered. Naming a hazard in prose is not covering it.

And the mutation row above miscredited its own evidence — see the corrected row.
The test it cited also contained an assertion about `new URL` platform behaviour
that could never fail; replaced with one that exercises this module.

### Fixed after review

Two things this PR claimed that were not true, both found by review:

- **The web copy had no tests.** Its Electron twin has 15. Replacing the web
  body with a pass-through left every task green while `apps/web` — which
  imports that copy — would have opened any host. Now covered, and red under
  that mutation.
- **The IPC stub was cast, not annotated**, so typecheck could not see a channel
  missing from it. Annotated, it reported two: `openFamilySearch` and
  `readImage`, the latter pre-existing drift nobody had seen. The seventh
  double, and the reason the sixth was caught but this one was not.

And one claim of mine was half wrong: ctrl+click **does** fire `click` (the
handler's `preventDefault` stops it). The real bypass is middle-click, which
fires `auxclick`. The fix stands; the sentence did not, and is corrected in both
places it appeared.

## What this does not prove

Nothing in CI launches the packaged app, so no check proves the IPC channel is reachable or that Electron wires the renderer to the right one. The tests assert the policy's logic and each call site's choice — not the wiring. Same class as #2037.

The grammar is duplicated from the engine's `utils/ark.ts` (which can't be imported — the engine is outside the pnpm workspace and the web side must never depend on it) and again in the web copy (the main process can't import a renderer package). Three copies, each cited to the others.


chesworthrm (changes requested):
Reviewed the full diff against a test-merge with `main` (the branch is 6 commits behind, so its own green head isn't the state that matters).

The design is right. Reducing the input to an identifier and rebuilding the URL from a constant is the correct shape, and dropping the raw `href` closes a real middle-click hole. I also swept every `ark`/`arkUrl` value in the repo — 833 distinct ones across fixtures, the shipped seed projects, and all run logs — through the new policy and compared each against the old https-only check. Four are now refused, and all four already failed the old check too. Nothing that used to open stops opening.

Three things to fix before this merges. Two are missing guards. I confirmed each by breaking the code and watching the suite stay green.

## 1. The hosted web viewer's call site has no test

I replaced the body of `openFamilySearch` in `apps/web/src/transport/WsResearchTransport.ts` with a bare `window.open(value, ...)` and removed the import. All 8 turbo tasks passed. All 440 tests passed.

That is the same hole the last round found, moved one file up. The new tests cover `resolveFamilySearchTarget` itself, but `apps/web` is its only consumer and the consumer is untested. `link-channel-routing.test.ts` does pin call sites — but it only reads `src/components`, so it never looks at a transport.

Add this to `apps/web/src/transport/__tests__/WsResearchTransport.test.ts`. The `fakeConn()` helper and the `afterEach(() => vi.unstubAllGlobals())` it relies on are already in that file.

```ts
describe('WsResearchTransport — constrained FamilySearch channel', () => {
  it('openFamilySearch constrains the destination', () => {
    const open = vi.fn()
    vi.stubGlobal('window', { open })
    const t = new WsResearchTransport('s1', fakeConn())
    t.openFamilySearch('https://evil.example/ark:/61903/1:1:MXYZ')
    expect(open).not.toHaveBeenCalled()
    t.openFamilySearch('1:1:QPRC-WPBZ')
    expect(open).toHaveBeenCalledWith(
      'https://www.familysearch.org/ark:/61903/1:1:QPRC-WPBZ',
      '_blank',
      'noopener,noreferrer'
    )
  })
})
```

I ran it both ways: green on your code, red under the bypass.

## 2. The provider binding has no test

```
You wrote:  setOpenFamilySearch(transport.openFamilySearch)
```

Your PR body names this as the silent-no-op hazard and stops there. I removed the line and its import. All 8 tasks green, all 440 tests pass — and every "Open in FamilySearch" and "View on FamilySearch" link in both apps would do nothing. Failing closed is the right choice, and it is exactly what makes the breakage invisible.

Add this to `packages/viewer-ui/src/contexts/__tests__/ResearchDataProvider.test.tsx`, with `import { openFamilySearch } from '../../lib/external'` at the top. `makeMockTransport` is already defined in that file.

```tsx
describe('ResearchDataProvider — outbound channel wiring', () => {
  it('wires openFamilySearch to the transport', () => {
    const spy = vi.fn()
    const transport = { ...makeMockTransport(), openFamilySearch: spy }
    render(
      <ResearchDataProvider transport={transport}>
        <div />
      </ResearchDataProvider>
    )
    openFamilySearch('ark:/61903/1:1:QPRC-WPBZ')
    expect(spy).toHaveBeenCalledWith('ark:/61903/1:1:QPRC-WPBZ')
  })
})
```

Green on your code, red with the binding removed.

## 3. The `new URL` row in the verification table is wrong

```
You wrote:  | Replace **all three** `return FS_BASE + …` sites with
            | `new URL(value, FS_BASE)` | 6 red, including the
            | protocol-relative case |
```

The protocol-relative tests do not go red. I tried three readings of that mutation and they pass under all of them. `//evil.example/x` fails all three regexes, so it never reaches a return site — and the second assertion in that test builds its own `new URL`, so it can never go red whatever the implementation does.

The reds come from something else. `new URL('ark:/61903/4:1:KGS8-LY1', FS_BASE)` parses as an absolute URL of scheme `ark:` and hands back the input unchanged, which is what breaks the accept-tests. What actually rejects a protocol-relative input is the `^` anchoring on `ARK_RE` and `TREE_PERSON_PATH_RE`, not the concatenation.

The count is also still ambiguous, which is what rewriting the row was meant to fix. I get 6 with `.href` across both copies (4 Electron + 2 web), 6 with the literal object in the Electron suite alone, and 8 with the literal object across both.

```
Change it to:  | Replace all three `return FS_BASE + …` sites with
               | `new URL(value, FS_BASE).href`, in **both** copies | 6 red —
               | 4 in the Electron suite, 2 in the web suite. NOT the
               | protocol-relative tests: `//evil.example/x` fails all three
               | regexes and never reaches a return site. The reds come from
               | `new URL` treating a bare `ark:` value as absolute. What
               | rejects a protocol-relative input is the anchoring on
               | `ARK_RE`/`TREE_PERSON_PATH_RE`, not the concatenation. |
```

The concatenation is still the right call. Only the claim that a test holds it is wrong.

## Not blocking

The two policy copies are byte-identical today (1897 bytes) and nothing holds them there. Their test files cover different inputs — 15 tests against 3 — so drift in either direction can stay green, which is a bit weaker than "these assertions are what makes drift loud". A test that slices both files between `/** Every outbound` and the closing `return null` and asserts equality would close it. Happy for that to be a follow-on.

## What I did not check

Nothing here proves the IPC channel is reachable in a packaged app, or that Electron wires the renderer to it. You declared that and it is true — I could not close it either. I did not run a live Cowork or hosted session, and I did not run the engine or harness suites because the diff reaches neither.

One thing worth knowing: `TREE_PERSON_PATH_RE` refuses the `/tree/person/details/<PID>` form. That shape appears in run logs and docs but in no viewer sink today, so it is latent rather than broken.

For the record, all 13 checks are green on `df9bf7b6` (`strip` skipped), `codeowners/errors` is clean, and eslint is clean on all eight changed Electron files. Registering the channel in the `security-invariants` table was the right call — note that path is matched by no CODEOWNERS rule, so the checklist edit itself pulls in no senior owner.

chrisedeson (changes requested):
Good work on the design. Reducing to an identifier and rebuilding from a constant is the right shape, and the middle-click fix is real. Richard's three blocking items look addressed in the latest commits.

Six new things, three blocking.

## 1. person.ark can hold non-FS URLs — button silently breaks

Richard's sweep of 833 fixture values found nothing that stops opening, and that sweep is right for existing data. But `tree-edit.ts` stores the agent's `input.ark` string with no validation at all:

```ts
if (input.ark !== undefined) person.ark = input.ark;
```

And the engine's `toArk()` returns its input unchanged on no match — the engine's own test has a person whose ark is `https://example.com/p1-ark`.

Before this PR, `openExternal` opened any `https://` value. After, `resolveFamilySearchTarget` returns null for a non-FS URL and the button silently does nothing. No error, no fallback.

The fix is probably a fallback to `openExternal` when `resolveFamilySearchTarget` returns null and the input is a valid `https://` URL. Or validate on write in `tree-edit.ts`. Either way, the current code has a silent regression for a path that exists.

## 2. Missing try/catch on shell.openExternal

The sibling `open-external` handler in `index.ts` wraps `shell.openExternal` in try/catch. The new `open-familysearch` handler does not. Combined with `void window.api.openFamilySearch(value)` in the renderer (no `.catch()`), a rejection from `shell.openExternal` (e.g. no default browser on a minimal Linux) is an unhandled promise rejection on every click.

Add the same try/catch the sibling has.

## 3. Button-reset CSS missing on PersonCard

The `.ark` class only has `color` and `text-decoration`. After the `<a>` to `<button>` change, it needs `background: none; border: none; padding: 0; cursor: pointer; font: inherit` — the same reset `.externalLink` has in `SidecarResultCard.module.css`. Without it, the button renders with default browser chrome in the 0.72rem meta row.

## 4. TREE_PERSON_PATH_RE accepts 4-4 (should be 4-3)

`{3,4}` on the second segment accepts `KW7C-X9PZ` (4-4). Every reference in the repo — `ark.ts` ("the bare 8-character persona/tree id"), `check-warnings/SKILL.md` ("four characters, a hyphen, three characters"), every fixture — says the format is 4-hyphen-3. Not a host bypass, but over-permissive against the documented spec. Same fix needed in both copies.

## 5. Dead preventDefault and silent-skip test loops

Two cleanup items:

- `handleArkClick` still calls `e.preventDefault()` after the element changed from `<a>` to `<button type="button">`. No-op — the handler can collapse to `onClick={() => openFamilySearch(person.ark)}`.

- The "never returns a URL outside the FamilySearch base" test in both `external-link.test.ts` and `familysearch-url.test.ts` uses `if (out !== null) expect(...)`. For the reject-case inputs, `out` IS null, so the assertion is silently skipped. Either assert `toBeNull()` explicitly for each reject input, or delete the loop since the tests above already pin every branch.

## 6. as never casts (non-blocking)

`registerExternalLinkHandlers(ipc as never)` appears 5 times. The function only calls `ipc.handle`, so narrowing the parameter to `Pick<IpcMain, 'handle'>` would let the existing fakes satisfy it with no cast. Matches the principle your own fix to `IpcResearchTransport.test.ts` states ("a cast blinds it"). Non-blocking, but worth aligning.

Richard already noted the two-copy drift risk as non-blocking and suggested a slice-and-compare test. I agree — the duplication is architecturally forced for the engine copy, but the electron-main and viewer-ui copies are both inside the pnpm workspace and could share via a subpath export like the existing `./contract`. Either way, a drift test would help.

chrisedeson (approved):
All six items from my review are fixed in `928d39e1`. Verified each:

1. **person.ark regression** — the right call. PersonCard now resolves upfront and only renders the button when it resolves. No dead button, no unvetted host opened. Better than the fallback I suggested.
2. **try/catch** — added, matching the sibling handler.
3. **Button-reset CSS** — `.ark` now has the same reset `.externalLink` has.
4. **TREE_PERSON_PATH_RE** — `{3,4}` to `{3}` in both copies.
5. **Dead preventDefault** — gone, handler collapsed to inline.
6. **Silent-skip test loop** — split into explicit ACCEPTED/REJECTED lists.

Also narrowed the IPC parameter to `Pick<IpcMain, 'handle'>`, removing the `as never` casts.

Richard's three items from the previous round were already addressed. This looks good.

chesworthrm (approved):
Approving. Full pass on `db1372c6`. The branch is level with `main`, so this read the merged state directly rather than the branch's own head.

All three items from my previous round are closed. I confirmed each by breaking it and watching the suite go red, not by reading:

- **Web transport call site** — red under a bare `window.open(value, …)`.
- **Provider binding** — red with the `setOpenFamilySearch` line removed.
- **The `new URL` mutation row** — corrected, and the corrected row matches what I measure.

The drift test I floated as a follow-on is here instead, and it is better than what I described. Slicing the policy region and comparing bytes is the right shape, and the second test — pinning that the slice actually contains all four patterns — closes the failure mode I would have missed, where a marker rename quietly turns the comparison into two empty strings. I loosened the tree-id pattern in one copy only and it went red, along with that copy's policy test.

I also re-checked Chris's four items, since three of them landed after his approval:

- **PersonCard visibility gate** — red when I drop the condition. Hiding the button is the right resolution. The `openExternal` fallback that was suggested would have reopened the labelled-button path this PR shuts.
- **Handler try/catch** — red when removed.
- **Preload channel pairing** — red twice when pointed at `open-external`.
- **Tree-id narrowing to 4-3** — correct, and I checked it against real data rather than the fixtures. 189 distinct FamilySearch tree PIDs in the live e2e run logs are all four-hyphen-three. Every non-conforming value in the repo is synthetic, and no four-four id appears in a `/tree/person/` URL anywhere.

Nine mutations in total, all red where expected. `turbo test typecheck --force`: 8 of 8 tasks, 459 tests. eslint clean on `apps/electron/src`. Annotating the IPC stub is a real guard — delete one channel from it and typecheck reports TS2741.

## Two things about merging, not about the code

**Chris's approval is stale, and it counts anyway.** It read `6808beef`. Seven of this PR's files changed after that, including the two new test files that hold his own items 3 and 4. This repo does not dismiss approvals on push, so it stays on the gate. Worth a second look from him before anyone merges, even though I have now verified those guards myself.

**Mine is the senior approval.** Chris is on neither senior team, so the code-owner rule was unsatisfied until now. Mine plus his makes two, and no threads are open.

## Optional

The tree-id comment says every fixture agrees on four-hyphen-three. One does not: `eval/fixtures/scenarios/mid-research-flynn-merge-pending/tree.gedcomx.json` carries `4:1:KWCJ-JAM7`. The narrowing is still right — the run-log evidence above is stronger than a fixture survey — so this is only about whether a future reader can trust that clause. Both copies would need the same edit to keep the drift test green.

```
You wrote:      * `check-warnings/SKILL.md` spells it "four characters, a hyphen, three characters",
                * and every fixture agrees. Not a host bypass either way — the host is rebuilt from
                * `FS_BASE` regardless — but over-permissive against the documented format.

Change it to:   * `check-warnings/SKILL.md` spells it "four characters, a hyphen, three characters",
                * and all 189 distinct tree PIDs in the live e2e run logs are 4-3. (One synthetic
                * fixture, mid-research-flynn-merge-pending, carries a 4-4 value; no real one does.)
                * Not a host bypass either way — the host is rebuilt from `FS_BASE` regardless — but
                * over-permissive against the observed format.
```

## What no check covers

Unchanged from last round and still true: nothing in CI launches the packaged app, so no check proves the channel is reachable or that Electron wires the renderer to the right one. You declared that in the body, and I could not close it either. Same class as issue #2037.

I did not run the engine or harness suites — the diff reaches neither.

chesworthrm:
Correcting one thing in my review above. I wrote "mine plus his makes two" — that is wrong, and it matters because it would have you waiting on a merge that cannot happen yet.

GitHub still reports `REVIEW_REQUIRED` with both approvals in place. The reason is access level: Chris has **read** on this repo, and branch protection only counts approvals from people with write access. So there is one counting approval, mine, and you need a second from someone with write access.

The code-owner half is satisfied — my approval covers it, and the outstanding `senior-genealogists` request does not need to be filled, since CODEOWNERS treats the two senior teams as an OR.

Two things left before this can go in:

1. A second approval from a write-access reviewer.
2. The branch is 2 commits behind `main` again, so `mergeStateStatus` is `BEHIND`. Merge `main` in before merging.

Chris's review stands on its substance — the four items he found were real and are fixed. This is only about what the gate counts.

T-FEH (changes requested):
Reviewed on `addf9263`. viewer-ui **313 passed / 37 files**, electron **131 passed / 10 files**, both typechecks clean, gate green. Two approvals already stand; this one does not, and I want to be plain about why — I had drafted an approval and it was wrong.

I told myself I had gone looking for the shape that broke your two sibling PRs on #1018 — an assertion checking a *spelling* rather than the *property* — and had not found it. I had checked the drift test and mutated the grammar, and never opened `link-channel-routing.test.ts`. It is there.

## [Blocking] The routing test is defeated by one aliased import

`link-channel-routing.test.ts:36-40` reads the component source and asserts `src.match(/openFamilySearch\(/g)` has the right count and `src.includes('openExternal(')` is false. Both survive an alias:

```diff
-import { openFamilySearch } from '../../lib/external'
+import { openExternal as openFamilySearch } from '../../lib/external'
```

Applied to **both** PersonCard and SidecarResultCard:

```
Tests  313 passed (313)      tsc --noEmit  exit 0
```

All four constrained sinks now route to the unconstrained channel and nothing notices. The call sites still *read* `openFamilySearch(`, and the string `openExternal(` never appears — the import says `openExternal as`. This is the only thing pinning the component→channel binding; the provider binding, the transport and the preload pairing all have behavioural tests. A render-level assertion — spy on the transport's two methods through `ResearchDataProvider`, click, assert which fired — closes it, and the harness already exists in `ResearchDataProvider.test.tsx`.

## [Blocking] The path-shape half of the grammar is unheld on the two bare branches

Every rejection test attacks the host half — `^`, the trailing slash, `notfamilysearch.org`. Nothing attacks the `$`. Dropping it from `ARK_RE` in **both** copies, so the drift test stays green:

```
viewer-ui 313 passed · electron 131 passed
resolveFamilySearchTarget('ark:/61903/1:1:MXYZ?redirect=https://evil.example')
  → https://www.familysearch.org/ark:/61903/1:1:MXYZ?redirect=https://evil.example   (was null)
```

`external-link.ts` states the invariant this breaks — "an open redirect on familysearch.org would otherwise be usable as a hop" — and the test that exercises it only covers the *URL* branch, so the two branches a bare agent-authored identifier actually takes are uncovered. Widening the charset to admit `/`, and the same `$` drop on `BARE_PREFIXED_RE`, survive the same way. One test per copy covers all of them.

## [Blocking] `arkUrl` has a live producer, and the result is a button that does nothing

The body says `arkUrl` "was renamed to `recordId` and survives only in July run logs."

```
apps/server/app/agent/mock_agent.py:257   "arkUrl": "https://www.familysearch.org/ark:/example"
apps/server/app/config.py:25              agent_mode: str = "mock"        ← the default
```

`ark:/example` has no `\d:\d:`, so `resolveFamilySearchTarget` returns null. And the two components gate differently:

```
PersonCard.tsx:25         const arkTarget = person.ark ? resolveFamilySearchTarget(person.ark) : null   ← resolvability
SidecarResultCard.tsx:137 {result.arkUrl && (                                                           ← truthiness
```

So on the default hosted path the sidecar renders "Open in FamilySearch →" and clicking it does nothing. `PersonCard.test.tsx` calls that "worse than no button" — the principle is stated and applied to one of the two components. Same asymmetry on `tm.ark` and the fulltext `result.id`.

## [Blocking] Deleting the main-process registration leaves everything green

Removing `registerExternalLinkHandlers(ipcMain)` *and its import* from `apps/electron/src/main/index.ts` — which is how anyone would actually revert it:

```
electron 131 passed      npm run typecheck  exit 0
```

Every constrained click in the desktop app becomes an unhandled rejection against an unregistered channel. Deleting only the call trips `TS6133`, which is the guard `PersonCard.test.tsx:19` correctly says is not one, "because it disappears with the line". This is the silent-no-op class you covered for the provider binding, one level up, in the privileged process — and a source-text assertion on `main/index.ts` catches it without launching anything.

## What is genuinely strong, and why I still think the design is right

`resolveFamilySearchTarget` never returns a URL derived from its input — always `FS_BASE + <validated suffix>` — so the host cannot come from the caller. Twenty-two attacks, zero escapes: `familysearch.org.evil.com`, `evil-familysearch.org`, trailing-dot host, `@evil.com` userinfo, `javascript:`/`data:`/`file:`/`blob:`/`about:blank`, protocol-relative, embedded `\n`/`\t`, punycode, `../` in the path, `?url=` redirect, uppercase. Note that this is why the mutations above are coverage gaps rather than live holes — `FS_BASE` still pins the host under every one of them.

The drift test is the best-built thing here: renaming its start marker fails loudly rather than comparing two empty strings, and widening `TREE_PERSON_PATH_RE` in the Electron copy alone reds it. That self-guard is exactly what the sibling PRs lacked.

The sink split is principled and I checked it rather than took it — the three generic sinks all render the destination as the link's own label. And dropping the raw `href` on PersonCard is the sharpest thing in the diff; the code comment is more precise than the body about it (middle-click is the hole; Ctrl+click does fire `click`).

## Smaller

The test counts are wrong again — seven new files, not five, and the two the body omits (`familysearch-url-drift.test.ts`, `PersonCard.test.tsx`) are the ones holding a reviewer's own items.

## Requesting changes

Four of these are one test each, and none asks you to re-reason about the design. The policy itself I would approve as it stands.

T-FEH (approved):
Second pass, on `8ecef9e4` — your `addf9263`, plus a merge of main and one commit of mine closing all four findings. viewer-ui **318 passed**, electron **133**, web **15**, every typecheck and eslint clean, gate green.

Head had not moved since my review, so rather than send four items back I fixed them — each verified by breaking it first, and each easy to revert.

## What I changed, and why each one holds

**The routing test was a spelling.** `link-channel-routing.test.ts` reads component source and counts `openFamilySearch(`; one aliased import satisfies it while routing every constrained sink to the unconstrained channel:

```
import { openExternal as openFamilySearch } from '../../lib/external'
→ 313 passed, tsc clean          (before)
→ 2 failed                        (now)
```

I left the source-text checks in place — they are a cheap breadth check across five files — and added a behavioural one to each component's own test, using the `setOpenFamilySearch` / `setOpenExternal` seams that already exist for this. Install two spies, click, assert which fired. That is the property; the count was the spelling.

**The grammar was covered on its host half only.** Every rejection test attacked `^`, the trailing slash, or prefix/suffix confusion; nothing attacked the `$`. Both bare branches build from `FS_BASE`, so a surviving suffix rides into the opened URL — the open-redirect-hop invariant your own header names. One test per copy now:

```
ARK_RE: drop the $              viewer 1 failed · electron 1 failed
BARE_PREFIXED_RE: drop the $    viewer 1 failed · electron 1 failed
ARK_RE: widen charset with /    viewer 1 failed · electron 1 failed
```

**The sidecar rendered a dead button.** `SidecarResultCard` gated its three sinks on truthiness while `PersonCard` gated on resolvability. That is not hypothetical — `mock_agent.py` writes `"arkUrl": "https://www.familysearch.org/ark:/example"` and `agent_mode` defaults to `"mock"`, so the default hosted path showed "Open in FamilySearch →" and clicking it did nothing. Which is what `PersonCard.test.tsx` calls worse than no button; the principle was stated and applied to one of the two components. Gated all three, with a test that reds when the gate comes off.

**The registration line had no guard.** Deleting `registerExternalLinkHandlers(ipcMain)` *and its import* left electron 131/131 and typecheck clean, while every constrained click became an unhandled rejection against a channel nobody registered. Deleting only the call trips TS6133, which is not a guard because it disappears with the line — as `PersonCard.test.tsx:19` already says about its own case. Source text rather than an import, for the reason `external-link.ts` exists at all.

Please read the commit rather than take my word for it; the sidecar gating is the only change to shipping behaviour, and it is three conditions.

## Two things for you, not me

**The body's `arkUrl` paragraph is wrong.** It says `arkUrl` "was renamed to `recordId` and survives only in July run logs". `mock_agent.py:257` is a live producer in the default mode. The gating fix makes the button behave correctly either way, but the scope claim should be corrected — it is the third round of a claim about that table.

**The test counts are stale**: seven new files, not five, and the two the body omits (`familysearch-url-drift.test.ts`, `PersonCard.test.tsx`) are the ones holding a reviewer's own items.

## What I did not touch

The `-H`-style class of gap I hit on your sibling PR does not arise here — I re-ran the full attack set against the policy (22 inputs: suffix and prefix confusion, trailing-dot host, userinfo, `javascript:`/`data:`/`file:`/`blob:`/`about:blank`, protocol-relative, embedded `\n`/`\t`, punycode, `../`, `?url=` redirect, uppercase) and nothing escapes. That is the rebuild-from-base design doing the work, not the pattern list, which is why the mutations above were coverage gaps rather than live holes.

## Approving

The drift test remains the best-built thing here — it fails loudly when its own slice breaks, and it catches a behavioural change in one copy alone. What the four fixes have in common is that the policy was right and the guards around it were checking the wrong thing; none of them required rethinking the design.

Issue #2050, opened 2026-08-31
tree_edit: require an uncertainty note on an implausible-age ParentChild/Couple edge (issue #1642 follow-on)


mercyokum:
**Touches:** `packages/engine/mcp-server/src/tools/tree-edit.ts` (`add_relationship`, ~line 480), `packages/engine/mcp-server/src/tools/tree-correct.ts` (the equivalent update path, if it can also create/repoint an implausible-age edge), `packages/engine/mcp-server/src/validation/tree-shape.ts` (`TREE_PARENT_CHILD_FIELDS`/`TREE_COUPLE_FIELDS` already include `notes` — no schema change needed), `eval/harness/validators/test_universal.py` (`test_parent_child_age_plausibility_flagged`, the validator this closes the gap for), `docs/specs/simplified-gedcomx-spec.md` §4.2 (`notes` already documented there).

## The problem

`eval/harness/validators/test_universal.py`'s `test_parent_child_age_plausibility_flagged` (issue #1642 Finding 2, mercyokum) fails a run universally — not tag-gated, runs against every skill's unit suite — whenever a new ParentChild relationship implies a parent age outside plausible bounds (≤12 general, ≤14 male, ≥80 general, ≥45 female — the exact bounds `person-warnings.ts` already treats as implausible for `check-warnings`) and the relationship carries no uncertainty note (needs-review/speculative/uncertain/inferred/etc.).

`promise-emmanuel`'s review on PR #2024 flagged this as effectively unfixable: no SKILL.md tells any skill to write such a note. Checked further: the compliance path isn't missing from the schema — `relationship.notes[]` is fully specified in `docs/specs/simplified-gedcomx-spec.md` §4.2 for both ParentChild and Couple, and is already in `tree-shape.ts`'s `TREE_PARENT_CHILD_FIELDS`/`TREE_COUPLE_FIELDS` allow-lists (so `tree_edit` already accepts it at write time). What's actually missing is that no caller is ever told to use it. Seven places touch `tree_edit` today: tree-edit, person-evidence, init-project, forget-and-rederive, research (skills), plus record-extractor and proof-conclusion (agents) — any of them can create the relationship this validator is checking.

[...]

PR #2052, opened 2026-08-31
board skills: state the instructions, drop the history — and the cross-lane merge ruling

chesworthrm:
Ruled by the lead on 2026-08-31, during a `/merge-recent-issues` pass that reported
issues #2030 (`developer`) and #2032 (`genealogist`) as independent *purely* because
doctrine forbade a cross-lane merge:

> it's okay to merge developer and genealogist lanes for the same skill because
> genealogists and developers can request help as needed.

Both merge skills leaned on the premise that refutes — that a card spanning two
lanes has no single assignee — so everything downstream of it had to move. Two
files, no code, no tests affected.

## `/audit-board`

**§1, the lane-split verdict.** Retitled from "only when the split is clean" to
"only when each half finishes without the other". The assignability argument is
named as retired, with the ruling quoted and dated. `"these are different lanes"`
is stated outright as not a reason to keep two issues apart. The
finishes-independently test survives as the only test.

**§1, the same-file default.** This listed three stated reasons to keep two issues
wanting different lines in one file apart — *"different lane, different reviewer,
or a paid-run slot they cannot share."* Two are gone:

- **"different lane"** — by the ruling.
- **"a paid-run slot they cannot share"** — this was an internal contradiction,
  surfaced by the same pass. §1 read an unshareable run as a reason to keep two
  issues *apart*; §4 ("Merge the issues, not the branches") reads it as the whole
  reason to *merge* — "Seven issues serialized is seven runs; the same work merged
  into three issues is three." One fact pointing two ways. §4 wins.

"Different reviewer" went too: on one skill it is the lane objection in different
clothes.

**§1, precedence (new).** Retiring those two left an ordering gap that would
otherwise have shipped: a blocker on one side keeps two issues apart regardless;
absent a blocker, a shared paid run merges them; only when no run is at stake does
finishing independently decide.

**§4.** Replaces the sentence the ruling directly overturns — *"Do not merge across
lanes to save a run: a `developer` harness fix and a `genealogist` fixture
adjudication in one issue is unassignable, which costs more than the run saved"* —
with the opposite. The section's axis moves from **lane** to **skill**, which is
what actually decides whether one run covers the work.

The twenty-record-hint caveat survives in substance — N independent investigations
still do not merge, because each still costs its own run. Its test now reads *"one
person who can ask the other lane for help"* rather than one person who must
already hold both skills; that reframing is the whole difference between the old
rule and the new one.

## `/merge-recent-issues`

Records the ruling beside the 2026-08-11 one with its provenance, mirrors the
rewritten lane-split section, and adds one instruction to that pass: **never report
"different lanes" as the reason for an independent verdict.** That line exists
because it is the mistake the pass had just made.

## What this does not change

The blocker rule (never park an unblocked issue behind a blocked one), the
never-replace-N-issues-with-one-tracker rule and its 2026-08-04 data loss, the
"does one person finish all of it in one sitting?" test, and the one-way mechanical
dependency → body edit verdict are all untouched.

## Applied under the new doctrine

Both merges were approved and applied before this PR:

- **#2050 → #1837** (absorb) — unaffected by the ruling; both `developer`.
- **#2032 → #2023** (merge) — this is the call the ruling reverses. Two
  `research-plan` plan-content rules that cannot share a paid eval run while they
  sit on separate cards. #2023 retitled to cover both; issue #2032 closed with the
  run arithmetic recorded.

Issue #2030 stays open: it touches `validator.ts`, `research-append.ts` and
`test_research_plan.py` and nothing inside the skill directory, so it consumes no
paid run and merging it would have saved none — the lane was never the real reason
to keep it separate.

## Verification

- Repo-wide grep of `.claude/`, `docs/`, `CLAUDE.md`, `DEVELOPMENT.md` for the
  retired premise (`across lanes`, `no single assignee`, `unassignable`, `spanning
  two lanes`): the only remaining hits are the new text explicitly retiring it. The
  one unrelated hit, `DEVELOPMENT.md:276`, is the `docs/TODOs.md` history.
- Both files are `.claude/skills/`, loaded on demand by the lead — not shipped
  plugin skills, so `CLAUDE.md`'s "no explanatory prose in a SKILL.md" token rule
  does not apply here.
- No CI job reads either file; nothing to run.

### DIFF
--- a/.claude/skills/audit-board/SKILL.md
+++ b/.claude/skills/audit-board/SKILL.md
@@ -13,15 +13,12 @@ allowed-tools:
 
 `/fill-ready` ranks the Backlog, `/review-ready` vets one shortlist,
 `/review-icebox` sweeps the frozen pool. All three look at issues one at a time.
-This skill is the only pass that looks at the pool **as a whole** — the merges,
-the rot, the clusters, and the handling that no single body can propose because
-no single body can see the others.
+This skill is the only pass that looks at the pool **as a whole** — merges, rot,
+clusters, and cross-cutting handling.
 
-**Run it weekly, and run it before `/fill-ready`.** That ordering is the point:
-filers search once and then file, because judging fit needs the whole pool and
-cannot be made from inside one PR. This pass is where duplicates get merged,
-premises get corrected and dead items get dropped. `/fill-ready` then ranks a
-deduped Backlog. Run it the other way round and it ranks duplicates.
+**Run it weekly, and run it before `/fill-ready`**, so that pass ranks a deduped
+Backlog. This is where duplicates get merged, premises get corrected and dead
+items get dropped.
 
 **You propose, then apply what is approved.** No branches, no PRs, no code edits.
 You have no `Edit` or `Write` tool on purpose.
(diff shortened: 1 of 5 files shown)


DallanQ:
Pushed two commits on top of this branch (05b23eb, 41e1a06). The lead asked for the history and the "why it's this way" to come out of the skills, then for the same pass across the rest of the board skills. Retitled the PR to match — the merge-doctrine change is still in here, it is just no longer the whole diff.

**Five files, −271 lines net** (`audit-board`, `merge-recent-issues`, `find-big-wins`, `fill-ready`, `triage-feedback`). Removed: dated lead rulings and their quotes, the incident behind each rule (~20 named issue/PR post-mortems), frozen point-in-time measurements, and every "this used to say X, and X is retired" passage — including the three this PR had added. Every instruction survives; section-number cross-references (§1, §4) now name the thing they point at.

Six real defects turned up on the way through, all fixed:

- `audit-board`'s better-handling pass asked "is a family of near-identical issues better as one worklist?" and answered "a tracking issue plus a checklist table costs less and loses nothing" — the exact anti-pattern its own merges section forbids, citing the same evidence. Now points at the missing generator/lint/convention.
- `merge-recent-issues` called `/audit-board` the twice-weekly pass; it is weekly.
- `merge-recent-issues` told itself not to restate `/audit-board`'s rules in its own words, then restated three sections of them. It now points and keeps only the two instructions that are its own.
- `find-big-wins` carried an "On the first run" section and a matching output-shape line. ADR-0010's ledger holds six rows, so the first run happened — both were dead instructions.
- `fill-ready` said the same thing twice in its verify section (two paragraphs on not repeating the read for the shortlist `/review-ready` gates).
- `fill-ready` froze a 2026-08-01 arrival/closure baseline inside a paragraph that itself said "recompute, never quote it".

No CI job reads any of these files; nothing to run.

Issue #1344, opened 2026-08-05
gps-mentor is instructed to use both wiki tools at five sites and calls neither — compliance gap, not an unused grant


DallanQ:
**Touches:** packages/engine/plugin/agents/gps-mentor.md

**Icebox — unblock condition: issue #1253 ("gps-mentor has no eval coverage — no
harness path can invoke an agent") must land first.** Nothing can measure this
until an eval path can invoke the agent. Do not start before then.

## The finding

@aghadiayeamayanvboernest's `make e2e-agent-tools` report (issue #1085, PR #1338)
found that `gps-mentor` declares `wiki_search` and `wiki_place_page` and calls
neither across the committed e2e runlogs. The other three agents are clean.

The report frames this as a declared-but-unused grant. **It is not.**
`packages/engine/plugin/agents/gps-mentor.md` instructs the agent to use those
tools at five sites:

| line | what it says |
|---|---|
| 430 | `wiki_place_page` (`section: "online_records"`) |
| 436 | "also check `wiki_place_page`" |
| 459 | `wiki_place_page` (`section: "research_tips"`) to identify … |
| 612 | **`wiki_place_page`** (`section: "online_records"` and …) |
| 623 | **`wiki_search`** — last-resort for finding published guidance |

So the agent is told to call them and does not. That is an
instruction-compliance failure, not dead weight — and the two conclusions point
opposite ways. Read as "unused grant", the fix is to drop the tools from
`tools:`. Read correctly, dropping them makes the agent permanently unable to do
what its body tells it to do.

This is exactly the ambiguity #1084 exists to resolve ("a report cannot tell 'the
grant did not bind' from 'the agent body never asks for it'") — except here the
body demonstrably *does* ask, which narrows it to two causes: the grant does not
bind, or the agent ignores the instruction.

## Check first, because it may be moot

If the Cowork zero-tools breakage turns out to also affect the e2e path, "never
called" is explained by the tool never binding, and there is nothing wrong with
the prompt at all. Rule that out before touching the body.

## Do not

Do not resolve this by adding more prose telling the agent to use the wiki tools.
`gps-mentor.md` is 34,426 bytes on main and PR #1249 proposes taking it to 42,855
(+24.5%). Per `docs/skill-lifecycle.md` §5, a prose edit does not compensate for a
tooling or eval defect, and this is one of those two until proven otherwise.

Emruthwill:
[...]

Issue #1253, opened 2026-08-04
gps-mentor is graded by nothing as the subject — the direct arm landed, but the gate and the snapshot still key on a skill directory


T-FEH:
> **High priority (2026-09-23):** lead: DallanQ — blocks the gps-mentor floor search (issue #2270) and the record-extraction conversion (issue #2818).

> **Re-reviewed 2026-09-21 (task-reviewer). Verdict `senior` — senior-only. The mechanism landed; the gate did not, and three claims below no longer hold.**
>
> 1. **What blocks a `gps-mentor` test today is the runnability gate, not the missing route.** The #2246 direct-agent arm landed as `d25e8560b`, but `eval/harness/harness/runnability.py:162` requires `packages/engine/plugin/skills/<test.skill>/` to exist — unconditionally, before any delegation branch. A minimal `gps-mentor` test carrying `input.delegation` validates against the schema and then aborts `not_runnable — skill not found: packages/engine/plugin/skills/gps-mentor` (measured 2026-09-21 by calling `harness.runnability.check_runnable` on an in-memory spec).
> 2. **Issue #2738 owns that gate change** — the same line, with an `agents_dir / f"{spec.skill}.md"` fallback — for the three thin routing skills. **Do not write a second one:** take #2738 first, or land the fallback here and name issue #2738 in the PR body *without* a closing keyword. **Do not satisfy the gate by creating `packages/engine/plugin/skills/gps-mentor/`**: `scripts/package-plugin.mjs:25` walks `skills` wholesale, so a stub ships into the plugin zip as a user-triggerable skill competing with the agent's own description, while `orchestrator.py:924` (`stage_skills=not spec.is_direct`) never stages it into the run — it would exist only to fool the gate. **And do not point `test.skill` at a neighbouring skill** (`research`, `proof-conclusion`): that grades gps-mentor under another skill's rubric, snapshot and eval slot.
[...]

PR #2055, opened 2026-08-31
eval reports: fix the Windows stdout crash, lint the class, and name the third reading of a never-called tool (#1344)

Emruthwill:
> [!NOTE]
> **Issue #1344 stays open.** The `gps-mentor` fix was split out of this PR (see the last commit),
> so nothing here resolves it. The stale close-link this PR had while the fix was still in it has
> been removed via `removeCloseIssueReferences`, and `closingIssuesReferences` is now empty --
> merging will not close the issue.

## Summary

Tier: **Normal**. Three commits, and the third is the actual fix for #1344.

- **The `gps-mentor` fix was split out of this PR** (see the final commit). Two reasons, neither about its correctness: #1344 still carries `icebox` and a live "do not start before #1253", which review asked be confirmed by the lead rather than merged over; and the paid `research` re-baseline it owed was superseded mid-flight by three newer annotated run logs landing on `main` (`16-55-13`, `17-38-18`, `18-35-01`). Rule 2 gates on the **latest** log, whose snapshot expects main's `gps-mentor.md` (`fe7eaf4e`) not the edited one (`7f60ee57`) — so the annotation would have been wasted and another run would race the same way. Its evidence is preserved on #1344, in ADR-0009, and in #2150. **Issue #1344 stays open.**
- **`make e2e-agent-tools` had never run to completion on Windows.** It prints `→` in a `Limits` footer that always prints, and no Makefile target sets `PYTHONIOENCODING`, so it died with `UnicodeEncodeError` at `print(format_report(...))` — for the whole Windows-based team, since the target shipped. Fixed, plus the same defect in 10 other runnable modules (including `make probe-agent-binding` and `eval/triggering/`), using the house pattern already in `e2e/author.py`.
- **A lint for that class**, in `test_encoding_lint.py` — the output half of the rule that file already enforces for input.
- **`agent_tool_usage_report.py` now names four readings** of a never-called-but-mentioned tool instead of two. The old two-reading action string is verbatim what #1344 was filed by paraphrasing. The third — an instruction in a branch the orchestrator no longer routes to — is why that issue's line table cited three sites that cannot fire. The fourth was added in review: a mention that *bounds* the tool (a last-resort or only-if note) where zero calls is the instruction working. It does not depend on the split-out agent edit: `wiki_search` at `gps-mentor.md:731` is marked "Last-resort", a bounded mention whose correct call count is zero, and the docstring example now cites only that.
- **ADR-0009** records the four claims refuted while measuring this, including the two tidy explanations that look right and are not.
- **`make test-all` could not pass on Windows either** (separate commit, separate subsystem). `apps/electron`'s `watcher.test.ts` asserted `stringContaining('sub/research.json')` against a string the real scan builds with `relative()` from `node:path` — so `sub\research.json` here. One assertion made the gate the PR template requires everyone to tick unpassable on the platform the genealogist team runs. Found by running that gate for the first commit.

## Review guidance

**Start here:** `eval/harness/tests/unit/test_encoding_lint.py`'s new `_needs_stdout_guard` and the 11 modules it covers. The claim to check is that the lint can actually fail: revert the 11 guards and it names exactly those 11 files, no more and no fewer — @Gennecis verified this independently in a scratch worktree. Second, the fourth reading in `agent_tool_usage_report.py`, which is the one change here that is a judgement call rather than a mechanical fix.

Second, `eval/harness/tests/unit/test_encoding_lint.py`'s new `_needs_stdout_guard` — specifically its **scope decision**. It fires on *runnable* (`__main__`) **and** *contains a cp1252-unencodable character*, deliberately over-approximating: the offending character may sit in a comment rather than a printed string. I judged proving reachability-to-`print` not worth it when the fix is three lines and harmless.

**Unsure about:** nothing outstanding. My one open question — whether `apps/server/app/sandbox_server.py` deserves an exemption, since it runs on Linux where cp1252 is never the default — was answered in review: six of the eleven patched modules are equally no-ops (the character sits in comments, not printed strings), so exempting one would be arbitrary and an exclusion list is a permanent maintenance item. Guard kept, no exemption.

## Plan

**Didn't change:** `wiki_search` at `gps-mentor.md:731` — "Last-resort for finding published guidance" is honest and correctly marked rare, so its zero call count is the instruction *working*, not failing. Sites `:458/:464/:487` likewise: they sit in the `### pre-exhaustiveness` rubric that `/research` stopped routing to on 2026-07-07 (commit `1bc84e58e`, PR #597), so they cannot fire, but they are correct if that gate ever returns. The wiki `tools:` grants stay — dropping them would break the `on-demand` → light-pre-exhaustiveness route, the exact wrong conclusion #1344 exists to prevent.

**A `research_append` precondition was built up and then rejected on its own evidence.** I specified it (an evaluation item citing Standard 14 must reference a `loc_`/collection id), counted it at 60–100 lines, and was about to ask for sign-off — then ran its would-fire rate over the corpus, which is free. **75% (9/12)**, because "Standard 14" is reasonably-exhaustive research *generally*: it caught correct findings about candidate pools and FAN follow-through that have no locality to cite. Narrowed to strings actually naming topical breadth it leaves **5** offending items corpus-wide, matched by regex against a free-text field with five different spellings. No basis for a blocking gate in the mandatory completion path. Withdrawn on the issue before it reached anyone.

**Acceptance check:** `make e2e-agent-tools` exits 0; on `main` it exits 1 with `UnicodeEncodeError: 'charmap' codec can't encode character '→'` at `agent_tool_usage_report.py:469`. Plus `test_runnable_modules_that_print_non_cp1252_reconfigure_stdout`, proven to fail by removing the guard again (it then names the file and prints `'→⇒'`), and the four-readings assertion, proven to fail by dropping one reading from the action string. Both re-verified after the review round.

**Deviated from the plan:** **Heavily — the plan's conclusion was refuted twice and the scope changed both times.** I first planned this as "#1344's premise is wrong; the wiki tools are unused because `/research` retired the gate that mandates them." Two rounds of `/critique-plan` plus an independent re-derivation killed it, and I verified all four blocking findings: all 91 gps-mentor invocations postdate the gate removal (2026-07-20 → 08-23), so the claim is unfalsifiable against this corpus; wiki calls per run *rose* after the retirement (1.00 → 5.25); `place_search`/`place_distance` have the identical structure and 44 calls; and the surviving site's condition does obtain under `proof-critique`. So #1344 was substantially right and I was wrong. The original centrepiece (a `body_mention_sections` report column) was cut for asserting location rather than reachability, and a planned "observed focus" line was cut for being gps-mentor-specific and printing "other 0" over 61 unknowns. Everything refuted is in ADR-0009 so it is not re-derived.

## Test plan

- [x] I ran `make test-all` and it passed — **but only after the electron fix; it was red on `main` for every Windows contributor.** All seven suites: ESLint clean, typecheck 5/5, test-js 4/4, server-pytest 214 passed / 1 skipped, engine-test 2744 passed / 1 skipped, eval-ui-test 168 passed, harness-test 2959 passed / 4 skipped. `gps-mentor.md` and its spec row are reverted to `main`, so `research`'s latest run log (`18-35-01`) is active with zero snapshot mismatches — verified locally — and its existing `.ann.json` satisfies rule 3.
      Two notes for whoever owns the gate: it takes **~6 minutes** here, not the "around 30s" the template's comment claims (harness-test alone is 4:25), and nothing detects that it is platform-broken.

- [x] **No paid eval run is owed.** With the agent edit split out, nothing under `packages/engine/plugin/` changes, so no skill snapshot is invalidated. I originally claimed this on the strength of `research` being in `RUNLOG_GATE_EXEMPT_SKILLS` — **that was wrong**: the exemption is keyed on directory existence and `main` had since added `eval/tests/unit/research/`. A run was genuinely owed, was performed ($3.97), and was then superseded; the split is what removes the obligation. No `eval-cosmetic-skip` label needed or wanted.

- [x] No skill `description` or DO-NOT clause changed, so no routing-pair negative tests are owed.

## Follow-on issues

**#2150** — `gps-mentor` ignores two unconditional instructions in its own body (`research.json` opened in 9 of 16 `Read` calls; `project_context` skipped in 56 of 82 captures). Labelled `genealogist` + `nothing-checks`, blocked on #1253. Filed at review's request so the findings stay visible in the `nothing-checks` register whenever issue #1344 is eventually resolved, rather than living only as comments on it. (Phrased to avoid a closing keyword — issue #1344 stays open, and this PR must not auto-close it.)

**Folded in / filed instead:** Two folded in, both step 1 (context loaded, fix mechanical): the Windows stdout crash and its class (11 modules), found while baselining the report this PR edits; and the electron `watcher.test.ts` separator bug, found while running `make test-all` for it. The electron one is a single self-contained commit (`9a968ae63`) in a subsystem this PR otherwise doesn't touch. I offered to split it out; **review said not to, and I agree** — it is one file and twelve lines, provably a no-op on Linux, and it unblocks the gate the PR template asks every contributor to tick. `watcher.test.ts:253-257` already carries the same fix with the same comment for `assertResearchProject`; this finishes the job on the one assertion that was missed.

Two findings measured and deliberately **not** filed, both recorded on #1344 and both out of its scope:

- **gps-mentor ignores two unconditional instructions.** 9 of 16 attributed `Read` calls open `research.json` against "Do not open `research.json`" (`:76`); only 26 of 82 tool-using captures call `project_context` against "Open every invocation with `project_context`" (`:82`). These are the same disease as #1344 with far more evidence behind them, and this PR's history is the argument that prose will not fix them.
- **`### on-demand` is wholly unexercised** by the e2e corpus — zero invocations, because every fixture is an autonomous task with no user review request.


packages/engine/plugin/agents/gps-mentor.md
```diff
@@ -717,10 +717,10 @@ not abstract.
   jurisdiction. "FamilySearch has 'Pennsylvania Probate Records,
   1683-1994' indexed and you haven't searched it" beats "consider
   probate."
-- **`wiki_place_page`** (`section: "online_records"` and
-  `section: "research_tips"`) — Use these to identify record
-  types and strategies the researcher hasn't tried. Quote specific
-  guidance.
+- **`wiki_place_page`** (`section: "online_records"` /
```
Gennecis: **The spec needs the same change.**

`docs/specs/gps-mentor-agent-spec.md` says in its own opening lines (3-6) that it is the source of truth for this file and that the implementation "must conform to what is written here." Its §14 tool table at lines **976-977** still carries the old rule:

```
| `wiki_place_page` (`section: "online_records"`) | When auditing topical breadth (pre-exhaustiveness rubric check 1). |
| `wiki_place_page` (`section: "research_tips"`) | When flagging repository diversity gaps or suggesting strategy improvements. |
```

That now contradicts the body. I ran the full engine packaging suite and it is green, so nothing in CI catches this — the next person checking the body against the spec will find the mismatch and may change the body back.

Please replace those two rows with one:

```
| `wiki_place_page` (`section: "online_records"` / `"research_tips"`) | Only when `project_context` returned no locality for the jurisdiction. Otherwise cite that locality's `collections` / `quirks` and its `loc_` id. |
```

Spec lines 315, 319 and 334 sit in the retired `pre-exhaustiveness` narrative and match what the body still says there, so leave those alone.

eval/harness/e2e/agent_tool_usage_report.py
```diff
@@ -383,7 +399,10 @@ def format_report(
             mentioned = set(d.mentioned_in_body)
             for tool in d.declared_never_used:
                 if tool in mentioned:
-                    action = "mentioned in body → clarify the body, or a binding gap (#1084)"
+                    action = (
+                        "mentioned in body → non-compliance, a binding gap "
+                        "(#1084), or an unrouted branch — check which"
```
Gennecis: **This PR creates a fourth reading, and the string does not carry it.**

After the `gps-mentor.md:720` change, `wiki_place_page` is still mentioned in the body, so `_body_mentions` still matches it and the next run of `make e2e-agent-tools` will print:

```
wiki_place_page   mentioned in body → non-compliance, a binding gap (#1084), or an unrouted branch — check which
```

None of the three is true. The body now mentions the tool in order to **bound** its use, so zero calls is the instruction working. That is the same reading the PR applies to `wiki_search` at `:731`, and the report has no way to express it.

Please change this string to:

```python
action = (
    "mentioned in body → non-compliance, a binding gap "
    "(#1084), an unrouted branch, or a bounded/last-resort "
    "mention where zero is correct — check which"
)
```

and make three matching updates so they stay in step:
- the `Limits` footer at line **434** (add the fourth reading to the "one of three" list, which becomes "one of four");
- the numbered list in the module docstring at lines **23-40** (add item 4);
- `tests/unit/test_e2e_agent_tool_usage_report.py:295-297` — add the fourth string to the loop so the new reading is asserted like the other three.

docs/adrs/ADR-0009-refuted-agent-design-claims.md
```diff
@@ -96,6 +99,29 @@ failure rather than into a vacuum.
...
+explanations; both are wrong.
+
+| claim | Why it was wrong |
+|---|---|
+| The wiki pair is unused because its mandatory sites sit in `gps-mentor.md`'s `### pre-exhaustiveness` rubric, and `/research` retired that gate | The premise is true — commit `1bc84e58e` (PR #597, 2026-07-07 11:48) deleted the rows that invoked the `pre-exhaustiveness` and `conclusion-readiness` gates, and both rubrics are still in the agent body. The **inference** is refuted three ways. (1) Wiki calls per run went **up** after the retirement: gate-era runs that wrote a pre-exhaustiveness verdict 1.00 (10 runs), gate-era runs that wrote none 1.10 (10), post-removal **5.25** (100). (2) The surviving mode-agnostic site (`gps-mentor.md`'s `## Tool usage guidance`) applies under `proof-critique`, and its condition demonstrably obtains — **17 of the 30** focus-attributed `proof-critique` verdicts raise untried record types or repository diversity ("Standard 14 — topical breadth"; "no probate, notarial, or passport records were consulted") with the tool still uncalled. (3) See the control row below. The retirement narrows *which* sites are live; it does not explain the zero |
```
Gennecis: I could not reproduce **"17 of the 30"** over the 159 committed run JSONs. Matching `topical breadth | repository diversity | Standard 14` against the 30 focus-attributed `proof-critique` writes gives **12**; adding `record type` gives **16**. I did not find a set of strings that gives 17.

Please either correct the figure, or add the matched strings to this row so the next reader can re-derive it — e.g. *"17 of the 30 … (matched on `topical breadth`, `repository diversity`, `Standard 14`, `record type`)"*. Everything else in this section reproduced exactly, so this one row is the only place a reader would get stuck.

eval/harness/tests/unit/test_encoding_lint.py
```diff
@@ -369,3 +369,110 @@ def test_pep695_file_is_parsed_and_scanned(tmp_path):
...
+        try:
+            ch.encode("cp1252")
+        except UnicodeEncodeError:
+            out.add(ch)
+    return out
+
+
+def _needs_stdout_guard(source: str) -> set[str]:
+    """The offending characters when *source* is a runnable module that prints
+    them without reconfiguring stdout; empty set when it is fine or in scope."""
```
Gennecis: Small wording fix: this reads `empty set when it is fine or in scope`, but the function returns an empty set when the module is **out of** scope. Please change `or in scope` to `or out of scope`.

apps/server/app/sandbox_server.py
```diff
@@ -391,6 +391,12 @@ async def handle(self, ws) -> None:
 
 
 async def main() -> None:
+    # The house pattern (`eval/harness/e2e/author.py`). A Windows console defaults
```
Gennecis, line 394: Answering your open question in the PR body: **keep this guard, no exemption.**

I checked where the offending characters actually sit in all 11 patched modules. Four of them print the character from code: `latency_report.py`, `skill_latency_report.py`, `eval/triggering/scripts/generate_report.py`, and `agent_tool_usage_report.py`. In the other **seven** — this file, `probe_agent_binding.py`, `corpus_report.py`, `stamp_findings_hash.py`, `try_record_extractor_thinking.py`, `check_runlogs.py` and `check_skill_frontmatter.py` — the character is only in comments or docstrings that nothing prints.

So this module is not a special case; six others are equally no-ops. Exempting one and not six would be arbitrary, and an exclusion list is a permanent item someone has to re-justify whenever a module moves. Three harmless lines with no exclusion list is the better trade. No change needed.

Gennecis (changes requested):
Nice work — I verified the main claims by running them, and they hold up:

- **The Windows crash is real and the fix works.** On `main`, `PYTHONIOENCODING=cp1252 python -m e2e.agent_tool_usage_report --since all` dies with `UnicodeEncodeError: 'charmap' codec can't encode character '\u2192'` at `agent_tool_usage_report.py:471`, exit 1. On this branch it prints the full report and exits 0.
- **The new lint can fail.** I reverted all 11 files to `main` in a scratch worktree and ran it — it failed and named exactly those 11 files, no more and no fewer.
- **The three-readings assertion can fail.** Removing `"or an unrouted branch"` from the action string makes the test fail.
- **The numbers reproduce exactly:** 91 gps-mentor invocations (84 captures + 7) over 68 runs; `place_search` 28 / `place_distance` 16 with the wiki pair at 0; 9 of 16 `Read` calls opening `research.json`; 26 of 82 captures calling `project_context`; 30 focus-attributed proof-critique writes.
- `project_context` does return each locality's `collections` and `quirks` (`project-context.ts:167-195`).
- Suites: server-pytest 214 passed / 1 skipped, engine packaging 616 passed, harness 3015 passed, electron watcher 32/32, frontmatter clean on 27 skills + 5 agents. All 12 CI checks green.

One extra measurement that supports the `gps-mentor` edit more strongly than the PR body does: across all **78** `localities` appends in the corpus, every one carries both `collections` and `quirks` — the locality entry is never empty. And `research_tips` is recorded `found: false` in **30 of 78**, so for 38% of localities the wiki page the old line asked for does not exist. The old instruction was not just redundant, it was often unsatisfiable.

Four things to change. Three are inline; the fourth is here.

**1. Update the spec** — see the inline comment on `gps-mentor.md:720`.

**2. Add a fourth reading to the report** — see the inline comment on `agent_tool_usage_report.py:404`.

**3. Name the trigger behind the "12 of 12" figure** (PR body, Review guidance, second bullet).

I reproduced "12 breadth verdicts, 11 runs, 1 citing the `loc_` id" exactly, but only when the match is `topical breadth | repository diversity | **Standard 14**`. Your own correction comment on #1344 says that trigger is too wide, and the corpus agrees — 5 of the 12 are candidate-pool and FAN items (`anna-findejsova-daughter`, `gertrudis-hoffmans-daughter`, `rozal-sarman-daughter`, `stribling-father-1821`, `victoriano-macatangay-parents`) that never owed a wiki call at all. Matching only `topical breadth | repository diversity` gives **7 items across 6 runs, 3 of which already cite a `loc_` or collection id**.

Your conclusion still holds — 7 of 7 also had a locality entry present — so this is a wording change, not a reason to drop the edit. Please change the bullet to read something like: *"12 items whose `standard` names Standard 14, 7 of them naming topical breadth or repository diversity specifically; all 12 occurred in a project that already had a `localities` entry."* Stating the trigger lets the next reader re-derive it.

**4. Two process points for the lead before merge.**

- #1344 still carries the `icebox` label and its "unblock condition: #1253 must land first — do not start before then." #1253 is still open. Your argument on the issue that this is the wrong unblock condition is well-evidenced and I agree with it, but that call is the lead's. Please get it confirmed rather than merging over a live "do not start".
- `Closes #1344` will hide the two compliance findings you measured (`research.json` opened in 9 of 16 `Read` calls; `project_context` skipped in 56 of 82 captures). They live only as comments on that issue, so closing it removes them from view for anyone reading `docs/architecture.md`'s "What nothing checks" register. Either leave #1344 open for them, or file one issue labelled `genealogist` + `nothing-checks` with `**Touches:** packages/engine/plugin/agents/gps-mentor.md`, blocked on #1253.

**Do not split the electron commit.** You offered to — please don't. `9a968ae63` is one file and twelve lines, provably a no-op on Linux, and it unblocks the gate the PR template asks every contributor to tick. `watcher.test.ts:253-257` already carries the same fix with the same comment for `assertResearchProject`; this finishes the job on the one assertion that was missed.

packages/engine/plugin/agents/gps-mentor.md
Emruthwill: Done in 35b54e2 — replaced both rows with your exact wording, and left 315/319/334 alone as you said.

You are right that nothing catches this. I had treated the spec as documentation of the body; its own opening lines say the reverse, that the body must conform to it. So the body edit alone would have left the next reader with a spec that contradicts it and a green suite either way — and the likeliest resolution of that contradiction is reverting the body, since the spec claims precedence.

eval/harness/e2e/agent_tool_usage_report.py
Emruthwill: Done in 35b54e2, all four sites — action string, `Limits` footer (now "one of four"), docstring item 4, and the test.

This is the best catch in the review. The PR would have shipped a report that flags its own fix as a defect: `wiki_place_page` is still mentioned in the body, so the next `make e2e-agent-tools` prints one of three readings, none of which is true, against the very line this PR corrected. That is the same failure mode as the original — a report whose vocabulary cannot express the real answer, so the reader picks the closest wrong one. Which is how #1344 got filed.

Proven to fail rather than assumed: removing `bounded/last-resort` from the action string fails the assertion with `AssertionError: the mentioned-in-body action omits 'bounded/last-resort'`.

docs/adrs/ADR-0009-refuted-agent-design-claims.md
Emruthwill: You could not reproduce it because it was not reproducible. Corrected in 35b54e2 by naming the match, which is what the row was missing.

Re-derived over the 159 committed run JSONs, against the 30 focus-attributed proof-critique writes:

| match | items | runs |
|---|---|---|
| `topical breadth\|repository divers` | 7 | 6 |
| `+ \|Standard 14` | 12 | 11 |
| `+ \|record types?` | 16 | 12 |

Your 12 and 16 are both exact. My 17 came from a wider match that also caught `probate\|notarial\|not searched\|never searched\|untried\|only .{0,25}FamilySearch` — never stated, so the figure was unverifiable by construction. The row now carries the strings and notes what the earlier draft matched.

Conclusion unchanged: all 12 occurred in a project that already had a `localities` entry.

eval/harness/tests/unit/test_encoding_lint.py
Emruthwill: Fixed in 35b54e2 — "fine or out of scope".

apps/server/app/sandbox_server.py
Emruthwill: Guard kept, no exemption, no change in 35b54e2.

Your reasoning settles it and I had the wrong frame: I was asking whether *this* module needs the guard, when the question is whether the *lint* should carry an exclusion list. Six of the eleven are equally no-ops, so the honest choice was six exemptions or none. Three harmless lines beats a list someone re-justifies whenever a module moves.

I have removed the open question from the PR body rather than leaving it for the next reviewer.

Emruthwill:
@Gennecis — all seven points addressed, but **the scope of this PR changed materially after your review**, so please re-look rather than clearing the old verdict. Summary:

## Addressed

| # | Request | Where |
|---|---|---|
| 2 | Fourth reading | action string, `Limits` ("one of four"), docstring item 4, and the test — all four sites you named |
| 3 | Name the trigger behind the figure | PR body + ADR-0009 |
| 4 | Correct or source "17 of the 30" | Now **12** naming Standard 14, **7** naming topical breadth / repository diversity, with the match strings given. Your 12 and 16 both reproduce exactly; my 17 came from a wider match (`probate\|notarial\|not searched\|untried`) I never stated. The old figure survives only as a labelled note explaining the correction |
| 5 | `or in scope` → `or out of scope` | `test_encoding_lint.py:416` |
| 6 | Keep the `sandbox_server` guard | Unchanged. Your reasoning settled it — six of the eleven are equally no-ops, so it was six exemptions or none |
| 7 | Don't split the electron commit | Still folded in |
| 4a | Get the lead's confirmation on `icebox` | Tag **restored** — I had removed it on the assignee's general go-ahead, not @DallanQ's, which was the overstep you were guarding against. Routed to him |

## Moot, not done

**Request 1 (sync the spec's §14 rows).** The `gps-mentor.md` edit is **reverted out of this PR**, so there is no body change for the spec to disagree with. Both are back at `main` and self-consistent.

## Why the fix was split out

Two reasons, neither about its correctness:

1. **#1344 still carries `icebox` and a live "do not start before #1253"** — your point, and it is @DallanQ's call, which is outstanding. The edit cannot merge today regardless of CI.
2. **The paid `research` re-baseline it owed was superseded mid-flight.** `research` lost its suiteless exemption when `main` gained `eval/tests/unit/research/`, so the edit invalidated that snapshot and owed a run. The run completed (`15-45-42`, $3.97) — and while it ran, three newer annotated logs landed on `main` (`16-55-13`, `17-38-18`, `18-35-01`). Rule 2 gates on the **latest**, and `18-35-01` expects main's `gps-mentor.md` (`fe7eaf4e`) not the edited one (`7f60ee57`). So the annotation would have been wasted and another run would race the same way.

With the edit out, `research`'s latest log is active with **zero snapshot mismatches** (verified locally) and its existing `.ann.json` satisfies rule 3 — no annotation owed, no further spend. **CI is green.**

## Note on request 4b

Beyond filing **#2150** for the two compliance findings, the stale close-link on this PR is **removed** (`removeCloseIssueReferences`; `closingIssuesReferences` is now empty). Editing the body was not enough — GitHub had recorded a user-link that survives body edits. Issue #1344 stays open.

## What is left here

Tooling only: the Windows stdout crash across 11 modules, its lint, the fourth reading, ADR-0009, and the electron `watcher.test.ts` fix. The fourth reading **does not depend on the reverted edit** — `wiki_search` at `gps-mentor.md:731` is marked "Last-resort", a bounded mention whose correct call count is zero, and the docstring example now cites only that.

The agent fix keeps its evidence on #1344, in ADR-0009 and in #2150, and lands in its own PR once @DallanQ rules.

Gennecis (approved):

EdmondOware (approved):
Approving. I checked the parts that are easy to assert and hard to verify, by running them rather than reading them.

The watcher fix is real on this platform: 32 of 32 pass on the branch, and 1 failed of 32 on a clean detached `origin/main` at `2cb17c05`. That test has been red for every Windows contributor. I hit it myself last night on an unrelated PR and reported it as unfixed upstream, which was true only because the fix is sitting here.

The new lint bites. I stripped the `reconfigure` guard from `corpus_report.py` and it named exactly that file and the character. Eleven modules gained the guard, the twelfth match being the lint itself. Harness suite 3050 passed, 3 skipped.

The split is clean in the way that matters for the gate: zero diff under `packages/engine/plugin/`, `gps-mentor.md` and its spec row reverted, so no snapshot is invalidated and no run is owed. `closingIssuesReferences` is empty and #2150 is open, so the two compliance findings stay visible and #1344 stays open. That is the closing-keyword trap handled properly rather than described.

Two things worth recording.

**`77fafc3c` landed fifty minutes after the approval, and its author is the approver.** `dismissStaleReviewsOnPush` is false here, so it would have merged on a review that predates it and I am the first person to read it. It is fine: comments only, and the correction is right. `gps-mentor.md:720-723` carries no only-if condition on `wiki_place_page`, and `:731` does mark `wiki_search` "Last-resort". Flagging the shape rather than the content, because a self-authored commit landing after a self-issued approval is a gap no check covers.

**This makes `make test-all` passable on Windows for the first time, and nothing detects that the gate is platform-broken.** Your test plan says so in a note. That is a `nothing-checks` item in the strict sense and it deserves a card rather than a line in a test plan, with `**Touches:** apps/electron`. It is the second Windows-only failure I have seen this week, and the team that runs Windows is the one that finds them last.

The refuted-plan section and the withdrawn `research_append` precondition are the best parts of this and I would not want them shortened. Measuring a gate's would-fire rate before asking anyone to review it, then killing it at 75% false fires, is ADR-0011's question asked properly.

mercyokum (approved):

Issue #3, opened 2026-04-15
Test wikipedia_search tool end to end


Gennecis:
Test the implemented wikipedia search tool end to end. The end goal: Tool should work in claude cowork. See docs/wikipedia-tool-testing-guide.md

Issue #1908, opened 2026-08-25
record_read's no-standardization rule rests on evidence that partly no longer reproduces


MMagnum:
> **Reviewed 2026-08-27 before junior handoff.** Premise holds, but the body invited a
> green-and-wrong one-line flip; rescoped to a phase-1 measurement with the doctrine half split
> off. Original text preserved at the bottom.

**Touches:** packages/engine/mcp-server/src/tools/record-read.ts (`:127-133`, `:166-173`), packages/engine/mcp-server/tests/tools/record-read.test.ts (`:271-287`), packages/engine/mcp-server/dev/, docs/specs/gedcomx-convert-spec.md (§7.4), docs/record-read-sidecar-scope.md (`:105-120`)

`record_read`'s live path calls `toSimplified`, not `toSimplifiedStandardized`
(`record-read.ts:133`), so it never fills `standard_place`. The justification at
`record-read.ts:127-133` cites two mis-resolutions observed **2026-07-08** ("Southampton, NY" →
"Southampton, England"; "Rochdale, England" → "Rochdale, South Africa"). Two resolver fixes landed
after that date and neither has been re-probed against this path:

- `4823ffdf` (2026-07-22) — phrase-quote the `name:` query (`place-api.ts:67-74`)
- `1af50fab` (2026-07-31) — derive a `contextName` from the input's second comma-segment and
  filter candidates by it (`place-resolver.ts:183-224`)

## What is already settled — do not re-derive it

- **The skip does not protect persisted state.** Both writer paths re-resolve the *same free-text
  string* through the *same* resolver when the caller supplies `place` without `standard_place`:
  `research-append.ts:1992-1998` and `tree-edit.ts:309-317` (both default-on, opt out via
  `resolveStandardPlace: false`). So whatever the resolver would get wrong on a `record_read`
  string, it gets wrong at write time anyway. The skip changes only what the model sees at read
  time.
- **`countryConsistency` already catches one of the two examples and cannot catch the other.**
  "Rochdale, England" vs "Rochdale, South Africa" → `contradiction` (South Africa is absent from
  `COUNTRY_ALIASES`, so `stdCountries` is empty; `place-resolver.ts:294-311`) — refused by
  `research_append`, cleared by `tree_edit`. "Southampton, New York" vs "Southampton, England" →
  `unverifiable`, because "New York" is not a country — written through with a warning. **The two
  cited failures are not equivalent and must not be summarised as one.**
[...]

PR #2058, opened 2026-08-31
record_read: probe script + doc corrections for the no-standardization rule (#1908 Phase 1a)

kofiatinka12:
## Start here
`packages/engine/mcp-server/dev/probe-record-read-places.ts` — the new probe that will re-measure whether the live `record_read` place-standardization skip is still warranted — plus the record/doc corrections. This is **Phase 1a** of #1908.

## What this does (and what it defers)
Issue #1908 Phase 1 is *"measure and correct the record"* with **no change to `record_read`'s runtime behavior**. This PR delivers the **correct-the-record** half now and the probe **script**; the **live measurement** is intentionally deferred (see "Deferred", below).

**1. New probe — `dev/probe-record-read-places.ts`.** New development probe; not part of production behavior. For record ARKs spanning US census, England parish, and Scandinavian church books / marriage index, it reports per place string: the raw `place.original`, whether recapi supplied `place.normalized`, the comma-segment count and `deriveContextName` value, what `resolveStandardPlace(original)` returns today, and the `countryConsistency` verdict — with a summary (ARK skip count, single-segment share, `normalized` supply, consistency breakdown). It reproduces `record_read`'s exact live recapi call and walks **both** person and relationship facts. ARKs are **fixture-derived** from committed `record-read-*.json`; the docblock flags that only the Richardson pair is CAPTURED LIVE while the other six are constructed / no-live-capture, so a run over them settles less than live-captured records.

**2. Corrected the "re-applies / needs place standardization" claim — and, this round, its inverse — across this PR’s own surfaces (see the acceptance check for the exact files).** The live path uses pure `toSimplified` (since PR #614); the sidecar returns the staged record as-is. The careful, measurement-proof framing now appears **consistently** in all sites: *the live path keeps whatever `standard_place` the record's own `normalized` value supplies and never falls back to the resolver* (it does **not** universally "leave `standard_place` unset" — whether recapi supplies a `normalized` value is exactly what the probe measures). Fixed:
- every internal comment in `record-read.ts` that stated the rule, plus the model-facing `resultsRef` schema description;
- `docs/specs/search-result-staging-spec.md`;
- the probe docblock (which had prejudged its own answer with "never fills `standard_place`").

**Provenance stated accurately.** The sidecar's `standard_place` is a **mixture** — FS-normalized where the search response supplied a `normalized` value, resolver-derived (via `record_search`'s `standardizePlaces`) where it did not. The reason to return it as-is: re-running the resolver changes nothing for the FS-normalized facts and merely re-rolls the same dice for the rest.

**3. Spec — `gedcomx-convert-spec.md` §7.4** names the exception with wording that survives the pending measurement (keeps the `normalized`-supplied `standard_place`; never falls back to the resolver), and its pointer now lands on a **true** doc: the three stale present-tense claims in `docs/record-read-sidecar-scope.md` (`:33`, `:40`, `:58` — "record_read uses toSimplifiedStandardized" / "closable by re-applying standardizePlaces" / sidecar "re-applies standardizePlaces, and returns it") are corrected to the shipped behavior, cross-referencing that doc's own Findings.

**4. Scope doc** — the 2026-07-08 observation is date-stamped, and its stale present-tense behavior claims are corrected (this round) / past-tensed (reviewer commit `5f9f072a`).

**5. Sidecar test assertion** — `expect(vi.mocked(resolveStandardPlace)).not.toHaveBeenCalled()` in the `sidecar mode` block pins "return staged as-is, do not re-standardize." Verified it fails under an artificial re-standardization, then restored.

## `record_read` runtime behavior is unchanged
`record_read`'s runtime logic is unchanged: live path is still `return toSimplified(body)`; `readFromSidecar` returns the staged record as-is. Non-runtime code changes only: the **advertised `resultsRef` schema-description string** (model-facing text), and `place-resolver.ts` exports `placeSegments` (one keyword) so the probe reuses the real segment splitter instead of hand-rolling one. `gedcomx-convert.ts` (incl. `toSimplified` / `toSimplifiedStandardized`) is untouched. Engine suite green (**2782 passed / 1 skipped**; typecheck clean).

## Deferred (blocked on #1877)
- The **live probe run and results table** (Phase 1 item 2's core) and the **actual re-probe date** (item 4) are **intentionally deferred until #1877 merges** — it rewrites `searchPlace`'s query sanitization, the exact call every resolution in the probe goes through, so pre-merge numbers would be stale. This PR targets `main`, **not** #1877's branch.
- The probe is **type-checked** (by `npm run typecheck`) but **not run live**; running it needs **FamilySearch auth** (`make e2e-login`) and is a human-read step.
- **No paid eval is required** — `record_read` is an MCP tool, not a skill snapshot; nothing here is in an `eval/runlogs` snapshot.

## Acceptance check
- Probe compiles; walks person + relationship facts; reports a skip count; ARKs fixture-derived with the constructed-vs-live caveat stated.
- A **newline-insensitive** sweep confirms no copy of the claim or its inverse remains **in the surfaces this PR touches**: `record-read.ts` (comments + schema string), `gedcomx-convert-spec.md`, `search-result-staging-spec.md`, `record-read-sidecar-scope.md`, and the probe. **Out of scope, deferred to a separate cross-cutting issue (#2092):** the same inverse claim survives in the **10 duplicated `places-guidance.md`** copies (shared across ~10 skills, pre-#614); correcting them together would flip ~10 skill snapshots and needs coordinated paid runs.
- §7.4's pointer lands on a doc with no stale present-tense claims.
- Sidecar re-standardization is pinned by a test; the live path by the existing test.
- `record_read` runtime logic unchanged; engine suite + typecheck green.

Refs #1908 (Phase 1a). Phase 2 (the behavior decision) remains a separate senior follow-on.


T-FEH (changes requested):
Reviewed on `6fd16f46` (your `a3bb77ff` plus a merge of main I pushed), plus one commit of my own, `5f9f072a`. Engine **2758 passed**, `npm run typecheck` clean, harness **2992 passed / 2 skipped**, gate exit 0 with **zero** warnings — your "no paid eval required" is exactly right, nothing here is a snapshot input.

The probe itself is good work and most of the record-correction is right. What holds this up is that the correction stopped short of the two copies that matter most, and one of the new sentences is itself inaccurate.

## What I verified and am happy with

"Comment-only" is the load-bearing claim in a PR touching a shipping tool, so I checked it mechanically rather than reading it — every changed line in `record-read.ts` is inside a `//`. Both corrected statements are true against the code: the live path is `return toSimplified(body)` at `:135`, and `record-search.ts:991` really does run `standardizePlaces`. The old comment was wrong in both directions.

The probe follows every convention (`fetchWithTimeout`, `BROWSER_USER_AGENT`, `getValidToken`), reports all four fields #1908 item 2 asks for, calls `deriveContextName` rather than merely naming it, and all 8 ARKs are genuinely fixture-derived across the three required record types — I resolved each one. Both commits blamed for making the 2026-07-08 evidence stale are real and do what the header says; I read `4823ffdf`'s diff and it does phrase-quote the `name:` value. Deferring the run while #1877 is open is right, and saying so in the probe's own header rather than only in the PR is better.

## [Blocking] The one copy the model actually reads was not corrected

`record-read.ts:40-41` — the `resultsRef` schema description, sent on every tool listing:

> "and correctly standardized places (more reliable than a live read, **whose place standardization can misfire**)"

That asserts the live read standardizes, which is the exact falsehood this PR exists to remove. Three internal comments were fixed and the model-facing string sixty lines above them was not. A repo-wide sweep finds one other survivor, and it is in the durable tier:

```
docs/specs/search-result-staging-spec.md:194   "— no live FS fetch — re-applying place
                                                standardization so it matches a live read"
```

That is verbatim the sentence deleted from `record-read.ts:70-75` in this diff. Correcting the record means these two as much as the comments.

## [Blocking] The new spec sentence over-claims, and the body says the opposite

`gedcomx-convert-spec.md`: "`standard_place` is therefore left unset rather than populated by the resolver."

`toSimplified` **does** fill `standard_place` — purely, no network — whenever the response carries a `normalized` value (`gedcomx-convert.ts:297-298`). So "left unset" holds only for facts recapi returns without one, and whether recapi ever supplies one is precisely what the probe was written to find out. The spec now states as settled the thing this PR says is unmeasured.

The body says it avoided this: *"Worded so it does not claim every recapi response lacks normalized places — that is what the probe measures."* As written, it does claim that. Something closer to "it keeps whatever `standard_place` the response's own `normalized` value supplies, and never falls back to the resolver" says the true thing and survives the measurement either way.

## [Blocking] The rule is now written in three places and pinned by none

The live half is properly pinned — I flipped `:135` to `toSimplifiedStandardized` and exactly one test caught it:

```
× does not re-standardize places on a live read       1 failed | 2757 passed
```

The sidecar half has nothing. I made `readFromSidecar` do precisely what all three new comments forbid:

```ts
const staged = match.gedcomx as SimplifiedGedcomX;
await standardizePlaces(collectFacts(staged));
return staged;
```

```
Test Files  115 passed (115)
      Tests  2758 passed (2758)
```

Fully green. The five sidecar tests assert person and relationship counts and that `mockFetch` was not called; none touches `resolveStandardPlace`, and the mock returns `null`, so a re-standardization is invisible. Documenting a rule three times while nothing enforces it is the shape #1908 item 1 was worried about — it says the live test "is the only thing pinning the current rule". A `expect(vi.mocked(resolveStandardPlace)).not.toHaveBeenCalled()` in the existing `sidecar mode` block closes it in four lines, and the branch is open.

## Minor

- `:73` says the sidecar's places come from "the search stage (`record_search` runs `standardizePlaces`)"; `:170` says they come "from FamilySearch". Those are different sources, and `standardizePlaces` fills gaps via the same resolver whose mis-resolutions justify the live skip — so the file currently argues the resolver's output is more trustworthy than the resolver's output. The acceptance check claims the three comments are mutually consistent; on provenance they are not.
- The probe's docblock cites `record-read.ts:127-133`; this PR's own first hunk pushed that block down a line.

## What I fixed rather than sent back

`5f9f072a` puts the scope doc's bold clause in the past tense — it still read "a live `record_read` **re-standardizes** it WRONGLY" for behaviour PR #614 removed on the very date the PR was stamping. Revert it if you would rather word it differently.

One correction to something a reviewer might otherwise assume: `dev/` **is** typechecked, by `npm run typecheck` via `tsconfig.typecheck.json`, which `pretest` invokes — so your "type-checked (clean)" claim is backed by a real check, not a manual one. The `no-bare-fetch` lint, though, walks only `src/`, so the probe's use of `fetchWithTimeout` is voluntary compliance rather than an enforced rule.

Requesting changes on the two uncorrected copies, the spec wording, and the missing sidecar assertion. The probe and the deferral I would approve as they stand.

kofiatinka12:
Thanks, this was an excellent review — every blocking point was right. Fixed in `08bc8faf`:

**1. The two uncorrected copies.** The model-facing `resultsRef` schema description (`record-read.ts:40-41`) no longer claims the live read standardizes — it now reads "standardized places from the search stage — a live read leaves `standard_place` unset." And `docs/specs/search-result-staging-spec.md` no longer says the sidecar re-applies standardization "to match a live read." Good catch that the schema string is the copy the model actually reads.

**2. §7.4 over-claim.** Adopted your wording verbatim: it "keeps whatever `standard_place` the response's own `normalized` value supplies, and never falls back to the resolver." The PR body's contradicting sentence is gone.

**3. The unpinned sidecar rule.** Added `expect(vi.mocked(resolveStandardPlace)).not.toHaveBeenCalled()` to the `sidecar mode` block. I verified it the way you did — patched `readFromSidecar` to `standardizePlaces(collectFacts(...))`, watched the assertion fail, then reverted.

**Minor.** Reconciled `:73`/`:170` provenance — both now say the sidecar's places come from the search stage (`record_search` runs `standardizePlaces`; the search response supplies FS-normalized places the recapi record lacks), so the file no longer argues the resolver's output beats the resolver's output. Fixed the probe docblock's `:127-133` → `:128-134`. Kept your `5f9f072a` past-tense fix.

Runtime behavior is unchanged (only the advertised schema string moved); engine suite 2757 passed, typecheck clean. No paid eval — confirmed nothing here is a snapshot input.

T-FEH (changes requested):
Second pass, on `38ab4726` (your `08bc8faf` plus a merge of main I pushed — main has since absorbed #2014, so its two new checks gate this branch for the first time and both pass, 8/8). Engine **2766 passed / 117 files**, `npm run typecheck` clean, harness **2992 passed / 2 skipped**, gate exit 0 with **zero** warnings.

Two of my items are properly closed. The other four turn out to be one problem with a common root: the *shape* of the false claim was fixed, and its *substance* was reproduced in new words.

## The sidecar rule is genuinely pinned now

This was the item that mattered. My round-1 mutation left the whole suite green at 2758/2758. Same mutation, this head:

```
clean    23 passed
mutant   1 failed | 22 passed
         expected "vi.fn()" to not be called at all, but has been called 1 times
         1st call: [ "Branch Township, Schuylkill, Pennsylvania" ]
```

Five lines on an existing test and it discriminates. Also confirmed the staged fixture's fact carries `place` without `standard_place`, so `standardizePlaces`' own filter does not neuter the assertion — the guard would have been hollow otherwise.

And the falsehood I blocked on is gone as a phrase:

```
$ grep -rniE "re-applying place standardization|standardization can misfire|live.*re-standardizes" docs/ packages/engine/
(no matches)
```

The probe's line cite is corrected to `:128-134`, which is exactly where the block sits.

## [Blocking] §7.4 got the precise wording; four other sites got the imprecise one

You wrote §7.4 carefully — "it keeps whatever `standard_place` the response's own `normalized` value supplies, and never falls back to the resolver" — which is exactly right against `gedcomx-convert.ts:297-298` and stays true whatever the probe finds. That wording did not travel:

```
record-read.ts:40   (model-facing schema)   "a live read leaves standard_place unset"
record-read.ts:175  (internal comment)      "leaving standard_place unset"
search-result-staging-spec.md:196           "leaves `standard_place` unset"
probe-record-read-places.ts:6               "so it never fills `standard_place`"
```

This holds only if recapi never supplies `place.normalized` — the proposition the probe exists to settle and which the body says is unmeasured. The repo already carries a counterexample: `eval/fixtures/mcp/record-read-gallo-1893-baptism.json` has `args: {"recordId": "~KHWD-N5R"}`, no `resultsRef`, so it models a **live** read, and its response carries three populated values (`Turany, Martin, Slovakia`; `Šútovo, Martin, Slovakia` ×2). A hand-written mock is not proof about live recapi — but it is this repo's own model of a live `record_read`, and it contradicts the sentence now shipping to the model. #2014's new shape check will not catch it: that compares top-level key names, and this is nested.

The probe docblock is the sharpest of the four — the instrument prejudges the answer it was built to find.

## [Blocking] The new provenance sentence does not compose

Reconciling my `:73`/`:170` item by asserting both provenances produced an argument that does not hold:

> the search stage (`record_search` runs `standardizePlaces`), and they are the more trustworthy value: the search response supplies FS-normalized places the recapi record response lacks.

`standardizePlaces` (`gedcomx-convert.ts:678`) filters to `facts.filter(f => f.place && !f.standard_place)` — precisely the facts FS did **not** normalize — and fills them with `resolveStandardPlace` (`:709`). It runs at `record-search.ts:991`, before staging at `:1055`.

So the sidecar's `standard_place` is a mixture: FS-normalized where the search response supplied one, resolver-derived where it did not. The trust clause is true only of the first group. For the second, the sidecar hands back exactly the artifact the live skip exists to prevent — a staged `"Southampton, NY"` with no FS value would have been resolved at search time by the same resolver, and is then returned "as-is, more trustworthy". This is the argument the whole rule rests on, so it is worth saying accurately: return as-is because re-running the resolver changes nothing for the first group and re-rolls the same dice for the second.

## [Blocking] §7.4's pointer lands on the contradiction

§7.4 now ends "See `docs/record-read-sidecar-scope.md` for the measurement and rationale." That doc still says, in the present tense:

```
:33   "`record_read` uses **`toSimplifiedStandardized`** = `toSimplified` + `standardizePlaces`"
:58   sidecar mode "re-applies `standardizePlaces`, and returns it"
:40   "Closable host-side by re-applying `standardizePlaces`"
```

`:58` is verbatim the sentence this PR deletes from the staging spec. Your acceptance check says no copy of the falsehood remains in "comments, schema string, staging spec" — accurate as scoped, but §7.4 makes this doc load-bearing, so it now has to be true too. Either correct those three lines or point §7.4 somewhere else. (I fixed `:108-109`'s tense last round, which is why the rest stands out.)

## Non-blocking

- The probe's 404 path prints `SKIP` to stderr while the table and summary go to stdout, and the docblock says to paste the table into #1908 — so a piped run yields percentages over an unknown denominator. A skip count in `=== summary ===` fixes it.
- "Real record ARKs": all eight are in the fixtures, but two fixtures describe themselves as constructed rather than captured (`record-read-ackerman-1860-m7qz-8kd`: "Stocked to stop an ABORT, not for realism"; `record-read-patrick-flynn-cflt-9k2`: "Mirrors the gedcomx already embedded in…"). Only the Richardson pair says "CAPTURED LIVE". Fixture-derived is not the same as real, and it bears on how much a probe run over these ARKs will settle.
- Body says "2757 passed / 1 skipped"; on the merged head it is 2766 / 0. Main moved.

Requesting changes. The test is exactly what I asked for and the phrase-level cleanup is done — what is left is making the replacement claim as careful as §7.4 already is, in the three other places it appears and in the doc §7.4 sends people to.

kofiatinka12:
Right on all counts — I'd fixed the *shape* of the claim and reproduced its *substance* inverted. Fixed in `623e3f7d`:

**§7.4's precise wording now travels to all four sites.** The schema string (`:40`), the sidecar comment, `search-result-staging-spec.md`, and the probe docblock all now say the live path *keeps whatever `standard_place` the record's own `normalized` value supplies and never falls back to the resolver* — no "leaves unset" / "never fills." The `gallo-1893` counterexample is exactly why; the probe docblock no longer prejudges the answer it was built to find.

**Provenance composes now.** The sidecar comment states the mixture — FS-normalized where the search response had a `normalized` value, resolver-derived where it didn't — and gives the real reason to return as-is: re-running the resolver changes nothing for the first group and re-rolls the same dice for the second.

**§7.4's pointer lands on a true doc.** Corrected the three stale present-tense lines in `record-read-sidecar-scope.md` (`:33` "uses toSimplifiedStandardized", `:40` "closable by re-applying standardizePlaces", `:58` "re-applies standardizePlaces, and returns it") to the shipped behavior, cross-referencing that doc's own Findings.

**Non-blocking:** added an ARK skip count to `=== summary ===` so a piped run has a known denominator; the docblock now says "fixture-derived" and flags that only the Richardson pair is CAPTURED LIVE (ackerman/patrick-flynn are constructed); PR body test count → 2766 / 0.

Runtime behavior unchanged (only the advertised schema string moved). Re-ran your break-test — the sidecar assertion still fails under an artificial re-standardization, then restored. No paid eval.

T-FEH (changes requested):
Third pass, on `623e3f7d`. Engine **2766 passed / 117 files**, both #2014 packaging checks 8/8, `tsc` and `npm run typecheck` clean, harness **2992 passed / 2 skipped**, gate exit 0 with zero warnings, CI green.

The four sites I named are all fixed and all accurate. But the sweep stopped at the four I named, and the primary statement of the rule was never opened — so the same correction is still missing from the block the PR's own probe cites.

## First, a correction to my own round-3 sweep

I ran this and reported "fully swept":

```
grep -rniE "leaves .?standard_place.? unset|leaving standard_place unset|…" docs/ packages/engine/
(no matches)
```

It returns nothing because the phrase wraps: `// … Leaving` ends one line and `// standard_place unset is correct` begins the next. A per-line grep cannot match across that break — the exact fragility CLAUDE.md names for the encoding lint, and I walked into it while checking someone else's sweep. Re-run newline-insensitively, there are four survivors, not zero.

## [Blocking] The primary statement of the rule is untouched

`record-read.ts:128-134` — the block the live path actually sits on:

> The recapi record response **carries no FS-normalized place** (only `original` + parsed County/City/State fields), so re-standardizing would … **Leaving standard_place unset is correct** — never fabricate a wrong one.

Two survivors in one comment, and it is the strongest form of the claim: not "usually unset" but "carries no FS-normalized place", full stop. The diff hunk headers on that file are `@@ -40`, `@@ -72`, `@@ -140`, `@@ -167` — 128 was never opened.

What makes this the one that matters: `probe-record-read-places.ts:9` cites *this block by line number* as the justification it re-measures. So the PR aims its new instrument at the one paragraph still asserting the instrument's answer.

Third survivor, and it carries a stamp: `docs/record-read-sidecar-scope.md:117-122`, inside the Findings section §7.4 now points readers to —

> the recapi record response carries **no** FS-normalized place (only `original` + parsed County/City/State fields — **verified**) … and leaves `standard_place` **unset**

Marked *verified*, for the proposition the body says is unmeasured. The diff past-tensed line 113 of that bullet and left the rest. (Fourth is `record-read.test.ts:285`, which reads as scoped to its own mock — a word change, not a blocker.)

The Acceptance check says "No copy of the … inverse over-claim remains." That is not true in three places.

## [Blocking] The mixture framing is right; the conclusion drawn from it is backwards

I endorsed this sentence in draft before checking what the filter actually implies, which was careless of me:

> re-running would change nothing for the FS-normalized facts and **merely re-roll the same resolver dice for the rest**.

`standardizePlaces` opens with `facts.filter(f => f.place && !f.standard_place)` (`gedcomx-convert.ts:679`), and sets `standard_place` on every group the resolver places (`:713-715`). So after the search-time pass, both the FS-normalized facts *and* the resolver-derived ones carry a value — and a re-run's filter excludes both. The resolver-derived half is precisely what a re-run would **not** re-roll. The only facts it reaches are those where FS supplied nothing and the resolver already returned null or hit the soft cap.

The real reason is stronger than the one written: re-running cannot change any value already present, so it could only spend round-trips retrying the few that failed — on a path whose whole purpose is avoiding a round-trip.

Same block, smaller: "produced by `record_search`'s `standardizePlaces` at search time" attributes the whole mixture to `standardizePlaces`. The FS-normalized half comes from `toSimplified`/`pickNormalizedPlace`; `standardizePlaces` only ever fills the other half.

## [Blocking] The scope doc contradicts itself ten lines apart

`:31` (untouched) says `record_search` stages "the **`toSimplified`** tool output". `:40-44` (new) says "the staged data already carries the search stage's `standard_place`". The doc never mentions that `record_search` also runs `standardizePlaces` — it does, at `record-search.ts:991`, before `stageSearchResults` at `:1055`, so the staged file genuinely carries resolver-derived values. As it stands a reader cannot tell where the staged `standard_place` came from, which is the question the whole page is about. The heading "Two deltas, one closable, one unverified" also no longer matches its list — this PR settles delta 1.

## What is genuinely closed

§7.4's wording reached the schema string, the staging spec and the probe docblock, and each is accurate against `gedcomx-convert.ts:293-298`. The probe line turned all the way around — it now names the open question rather than asserting the answer. `scope.md` :33/:42/:62 are corrected, with the old reading kept visible as superseded, which is more useful than deleting it. The skip count is on **stdout** with the table and names the skipped ids, so the denominator is honest. And the sidecar assertion is falsifiable — mutating `readFromSidecar` to re-standardize reds it on the resolver call.

## Requesting changes

Items 1 and 3 are mechanical. Item 2 needs a re-read of that filter before rewriting the sentence, not a reword. Everything else here is sound and every suite is green — but this is the third round in which the correction was right where it was applied and absent where nobody re-swept, and the surviving copy is the one the probe was built to test.

kofiatinka12:
You're right on all three, and right about the pattern — "fixed where named, missed where nobody re-swept." Fixed in `7cf70ce0`; this round I swept newline-insensitively (which is what hid the wrapped phrase from both of us) and corrected every survivor, not just a new named set.

**The primary block (`:128-134`) — the one the probe cites — is corrected.** No more "carries no FS-normalized place" / "leaving standard_place unset"; it now says `toSimplified` keeps whatever the record's own `normalized` value supplies and never resolver-fills, and names the probe as what measures how often that value is present. Same fix on the `sidecar-scope.md` Findings bullet that had it marked *verified*, and the test comment.

**B2 — I had the filter backwards, thank you.** `standardizePlaces` fills only facts lacking `standard_place`, so after the search-time pass both halves carry a value and a re-run's filter excludes both. The comment now gives the real reason: a re-run can't change any value already set — it could only retry the few that returned null or hit the soft cap, spending FS round-trips on the path built to avoid them. Also fixed the attribution (the FS-normalized half is `toSimplified`, not `standardizePlaces`).

**B3 — scope doc.** `:31` now states `record_search` runs `standardizePlaces` after `toSimplified` (so the staged file's `standard_place` is the FS-normalized + resolver-derived mixture), and the "Two deltas" heading marks delta 1 settled.

Comment/doc-only; runtime unchanged; the break-test still reds on the resolver call; suite green. (One self-inflicted catch: my first B3 draft cited `record-search.ts:991` and the doc-links lint rightly rejected the line number — now cited by symbol.)

T-FEH (changes requested):
Fourth pass, on `7cf70ce0`. Engine **2766 passed / 117 files**, #2014's two checks 8/8, `tsc` and engine `npm run typecheck` clean, harness **2992 passed / 2 skipped**, gate exit 0 with zero warnings, CI green. Comments-only again.

**All three round-3 items are genuinely closed**, and item 2 composes this time — I re-derived it from `standardizePlaces`' filter and `if (standard)` guard rather than reading it for difference. The probe-cited block is now the best statement of the rule in the PR: it names the open question *and* says "either way, we never resolver-fill it here", so the rule holds whatever the probe returns. The scope doc's heading matches its list again and the "**verified**" stamp is gone.

What blocks is that the body's scope claim is wrong in the one direction nobody swept.

## First, my own sweep was one-directional

I reported zero survivors last round and again this round. Both sweeps searched for the *over-claim's* wording — "leaves standard_place unset", "no FS-normalized place". The **inverse** claim uses none of those words, so neither sweep could have found it. Round 3's lesson was that my grep matched lines instead of text; this round's is that it matched one direction of a two-directional claim.

## [Blocking] The inverse claim survives in ten model-facing files

The body says the inverse was corrected "**everywhere**" and that a sweep "confirms **no copy** … **or its inverse over-claim** re[mains]". Ten copies ship in the plugin:

```
packages/engine/plugin/references/places-guidance.md         :76-78
packages/engine/plugin/skills/{init-project, conflict-resolution, tree-edit,
  search-external-sites, locality-guide, timeline, historical-context,
  record-extraction, research-plan}/references/places-guidance.md
```

each saying:

> If the place came from a `record_read` / `record_search` / `person_read` result, that fact **already carries a converter-resolved `standard_place` — copy it** (no tool call).

These are the only copies the model ever reads, and they assert exactly what this PR spent four rounds removing from the comments — for the live path, where it is false. Worse than a stale sentence: the `Otherwise` branch below is keyed on **provenance**, not on whether the field is present, so a live-read fact with no `standard_place` matches neither branch and the model is told to copy something that is not there.

It predates the PR (#274, 2026-06-05, before #614 made the live path pure `toSimplified`), so I would not have raised it against a body that said "the comments and specs". Against "everywhere", it is the finding. All ten must move together — there is a places-guidance drift lint.

## [Blocking] `7cf70ce0` introduces a fresh over-claim, contradicted by its own sibling comment

New at `record-read-sidecar-scope.md:33-34`:

> facts **each** carrying a `standard_place` (FS-normalized where the search response supplied a `normalized` value, **resolver-derived otherwise**)

Neither half of "each … otherwise" holds. `standardizePlaces` sets the field only under `if (standard)` (`gedcomx-convert.ts:713`), so a null resolve leaves it empty, and the soft cap slices the overflow away with its own log line — "`N` distinct place(s) over the soft cap … **left unstandardized**" (`:695-700`). The same commit says so 140 lines later in `record-read.ts:179-180`: "it could only retry the few that **resolved to null or hit the soft cap**". A PR whose subject is removing over-claims should not add one, and `:44` carries the same generalisation in softer form.

## [Important, before the probe is ever run] The instrument measures a looser condition than the behaviour

`probe-record-read-places.ts:135` counts a place as supplied on `Array.isArray(normalized) && normalized.length > 0`. The tool's actual gate is `pickNormalizedPlace` (`gedcomx-convert.ts:262-269`), which returns `undefined` unless some entry has a truthy `.value`. So an array of value-less entries reads `norm=yes` in the probe and yields no `standard_place` in the tool.

That is the probe's headline number — `recapi supplied place.normalized: N/total` — and the whole PR defers a behaviour decision to it. Cheap to align now; expensive to discover after the run is used to settle Phase 2.

## Smaller

- The docblock's constructed-fixture caveat names two of the six non-live ARKs. `anders (9XKT-M2P)` and `urna (NW44-PM2)` also say "mirrors the gedcomx already embedded in…", and `68Q9-K34P` / `birkeland (68Q3-5SGC)` claim no live capture at all — so only the Richardson pair is captured, and the current wording reads as though four others are.
- The docblock cites `record-read.ts:128-134`; the rewritten block runs 128-137. Third drift of that citation, each caused by editing the block it points at — citing it by its opening words would end the cycle.

## Requesting changes

The comment and spec work is finished and correct; the sidecar guard still reds under mutation (`1 failed | 22 passed`). What is left is that "everywhere" is not yet true of the ten files the model actually reads, and one new over-claim slipped in while removing the old ones.

kofiatinka12:
Fourth-round items handled, and you're right that both my sweeps were one-directional. Fixed in `e89003ca` (comments/probe/doc-only, no paid run):

- **Fresh over-claim (`:33-34`, `:44`):** corrected — "most facts carry a `standard_place`; a null resolve or soft-cap overflow leaves some empty," matching the sibling comment's "resolved to null or hit the soft cap."
- **Probe vs tool condition:** `hasNormalized` now requires a truthy `.value`, mirroring `pickNormalizedPlace`, so the headline number can't over-count value-less `normalized` arrays.
- **Fixture caveat:** now says only the Richardson pair is captured live; the other six are constructed / no-live-capture.
- **Cite drift:** the probe now cites the block by its opening words, not a line range — ends the cycle.
- **The 10 `places-guidance.md` copies:** you're right the body's "everywhere" was false. I've narrowed the body to the surfaces this PR actually sweeps, and filed the inverse-claim across the 10 duplicated copies as a separate cross-cutting issue, #2092 — it's pre-#614, shared across ~10 skills under a drift lint, so it needs coordinated paid runs rather than riding this developer PR.

Suite green; the sidecar guard still reds under mutation.

T-FEH (changes requested):
Fifth pass, on `45ad4ffe` (your `e89003ca` plus a merge of main I pushed). Engine **2766 passed / 117 files**, #2014's checks 8/8, `tsc` and engine `npm run typecheck` clean, harness **3021 passed / 2 skipped**, gate exit 0, sidecar guard still falsifiable (`1 failed | 22 passed`).

Four of five items are closed, and the fifth — the ten plugin files — is resolved the right way. One line survives, and it is inside a surface the acceptance check names as swept.

## The deferral is done properly

The body no longer says "everywhere"; it says "**in the surfaces this PR touches**" and lists them. Since those ten `places-guidance.md` copies predate this PR by three months, that is the honest scope. What makes it a resolution rather than a retreat is **#2092** — filed, linked, labelled `developer` + `cross-cutting` + `nothing-checks`, carrying the `Touches:` line for all ten paths and naming the part that matters most, that the `Otherwise` branch keys on provenance rather than presence. Ten coordinated skill-file edits behind a drift lint is genuinely a different reviewer's job.

I checked `record-extractor.md` separately, since it also says "correctly standardized places": it scopes that to `record_read({ recordId, resultsRef, projectPath })` for **sidecar-backed** records and then contrasts "(3) a **live** `record_read`". Accurate, not a survivor.

## The other three items

The fresh over-claim is fixed and names its mechanism — "facts **most of which** carry a `standard_place` … a null resolve or soft-cap overflow leaves some empty", which matches `if (standard)` at `:713` and the soft-cap slice at `:695-700`.

The probe predicate is now equivalent to `pickNormalizedPlace`, and I checked the semantics rather than the shape: the `lang === "en"` preference only selects *which* value, never whether one exists, so "some entry has a truthy `.value`" is the right boolean. Where the two could differ — a non-string truthy value the declared type forbids — the probe is stricter, so the headline figure can only under-count. Safe direction for a number Phase 2 will be decided on.

And the citation drift is ended rather than patched: "the `Use toSimplified, NOT toSimplifiedStandardized` block in record-read.ts", no coordinate to go stale.

## [Blocking] One line, in a surface the acceptance check calls clean

`docs/record-read-sidecar-scope.md:89-91` — five commits on this branch touched that file and no hunk touched these lines:

> Place standardization: the staged result already **has the correct** standardized place; the tool returns it as-is (**a live read's own standardization is *less* reliable** — see Findings).

Both halves are the claims this PR exists to remove. "Already has the correct" is the unhedged inverse, contradicting the hedge you added 56 lines above ("a null resolve or soft-cap overflow leaves some empty"). And "a live read's own standardization is less reliable" describes something that has not existed since #614 — the live path performs no resolver standardization at all, so it is not less reliable, it is absent. That is the last present-tense stale claim in the file.

It falsifies two acceptance bullets, not one:

```
"…no copy of the claim or its inverse remains in the surfaces this PR touches:
 … record-read-sidecar-scope.md …"          ← it remains, there
"§7.4's pointer lands on a doc with no stale present-tense claims."   ← it does not
```

and §7.4's pointer was added by this PR, so the doc is load-bearing now in a way it was not before.

## Why my sweeps kept missing these

This is the third round where a survivor slipped past a grep of mine, each time by wording: round 3 it wrapped across lines, round 4 it was the inverse phrasing, and this one says "already **has** the correct" rather than "already carries a converter-resolved". Matching strings was the wrong method. What found it is reading every sentence in that file that makes a claim about the live path or standardization — 19 of them — and judging each by meaning. That pass turns up exactly one survivor and no fourth, so I am confident this is the last of them.

- Item 1 still says the docblock names "**two others**" as constructed; it now names six.
- Item 2 says "the **three** internal comments + the schema description"; the diff changes five comment sites plus the string.

## Requesting changes

One bullet in one doc, and the two body counts with it. Everything else is finished — and the deferral on the plugin files is the cleanest use of a follow-on card I have seen on this repo: scoped, labelled, `Touches:`-lined, and linked from the PR that found it.

kofiatinka12:
Caught and fixed in `ece1380e` — that line (`:89-91`) was the last present-tense survivor, and both halves were wrong: "already has the correct" is the unhedged inverse (now "carries the search stage's standardized place for the facts the resolve reached", matching the hedge above it), and "a live read's own standardization is less reliable" describes something gone since #614 — the live path does no resolver standardization at all, so it's absent, not less reliable. And you're right that reading every claim-sentence by meaning is what found it where five string-greps didn't.

Body counts corrected too: item 1 now says the docblock names the **other six** ARKs as constructed / no-live-capture, and item 2 no longer undercounts the record-read.ts comment sites.

Doc-only; runtime unchanged; suite green; sidecar guard still reds under mutation.

T-FEH (approved):
Sixth pass, on `ece1380e`. Engine **2766 passed / 117 files**, #2014's checks 8/8, `tsc` and engine `npm run typecheck` clean, gate exit 0, sidecar guard still falsifiable (`1 failed | 22 passed`), CI green.

## The last stale claim is corrected, and corrected rather than softened

```
- the staged result already has the **correct** standardized
-  place; the tool returns it as-is (a live read's own standardization is *less*
-  reliable — see Findings).
+ the staged result carries the search stage's standardized
+  place (for the facts the resolve reached); the tool returns it as-is. A live read
+  performs **no** resolver standardization at all (pure `toSimplified` since #614),
+  so there is nothing to be less reliable — see Findings.
```

Two things I'd single out. The first half now carries the same hedge as `:46` — "for the facts the resolve reached" — so the doc is internally consistent about the mixture rather than hedged in one place and absolute in another. And the second half does not just delete the false comparison, it says **why** a comparison was the wrong shape: there is nothing to be less reliable, because the live path does none. That is the difference between removing a wrong sentence and fixing the thinking behind it.

## The meaning-based pass, which is what I should have been doing all along

I extracted every sentence in `record-read-sidecar-scope.md` that makes a claim about the live path, the sidecar path, or standardization — **19 of them** — and judged each against `record-read.ts`, `toSimplified`/`pickNormalizedPlace`, and `standardizePlaces`. No survivors.

The three that could read as unhedged all hold in context:

- ":45 the staged place is more reliable" is scoped to the alternative it rejects — a host-side `standardizePlaces` re-apply — and the sentence before it carries the "facts the resolve reached" hedge.
- ":113 the search result's is correct; a live `record_read` **used to** re-standardize it WRONGLY" sits under the heading "Findings (**verified live, 2026-07-08**)" and carries its own re-probe caveat. A dated observation, not a present-tense rule.
- ":115 The sidecar is therefore *more* reliable" is the same dated finding's conclusion.

`ece1380e` touches only this file, so the other four named surfaces are unchanged from the state I read and verified last round.

## Both body counts fixed, one of them permanently

Item 1 now says "the other **six** are constructed / no-live-capture". Item 2 replaced "the three internal comments" with "**every** internal comment in `record-read.ts` that stated the rule, plus the model-facing `resultsRef` schema description" — dropping the count rather than correcting it, the same durable move as replacing the line-number citation with the block's opening words. Neither can drift again.

## Two things to tidy, neither worth a seventh round

**`:122` — "The sidecar is therefore *more* reliable."** Read whole, that bullet is coherent: it states the 2026-07-08 finding in the past tense, draws the conclusion that held then, and closes with "**The live `record_read` path is also fixed:**". But `:92` — written by this very commit — says of that same comparison "there is nothing to be less reliable", and then sends the reader here with "see Findings". So the pointer's target still draws a comparison the pointer has just retired. Past-tensing the conclusion ("the sidecar was therefore the better value") settles it.

**The fixture caveat mis-sorts three of the six.** I read each fixture's own `description`:

```
"mirror an embedded gedcomx"   patrick-flynn, anders, urna
"stocked to stop an ABORT"     ackerman-1860
no capture note                68Q9-K34P, birkeland
```

The docblock puts ackerman with the mirrors and anders/urna with the no-note group. The headline — six of eight are not live captures — is right either way, which is why this is polish rather than a finding.

## Approving

Six rounds. The pattern throughout was a true statement being replaced by a slightly stronger one, and each round the strengthening landed somewhere the previous sweep had not looked — three of those escapes got past greps of mine, by line-wrapping, by inverse phrasing, and by a synonym. What ends it is not a better pattern: it is that the claims are now hedged where the code is conditional, dated where they are observations, and stated so they hold whatever the deferred probe returns. Reading the file's claims by meaning turns up nothing left.

Phase 1a is the right shape — the instrument and the corrected record land now, the measurement follows #1877, and the ten plugin copies are tracked on #2092.

chrisedeson (changes requested):
T-FEH's review covered the doc and comment accuracy thoroughly across 6 rounds. These are all in the probe script itself — measurement accuracy and code reuse.

No production code changes in this PR, so nothing here is blocking a merge. But the probe's numbers get pasted into #1908 to decide Phase 2, so the measurement gaps matter more than they would in a throwaway script.

## Measurement accuracy

**1. Segment count uses raw split, not placeSegments logic**

`segments: original.split(",").length` (line 175) counts raw comma tokens. But `deriveContextName` and `countryConsistency`, called on the same string, route through `placeSegments()` in place-resolver.ts which trims and filters empties. A place like `"Springfield,, Illinois"` prints `seg=3` while `deriveContextName` sees 2 real segments. The `single-segment` summary stat is skewed the same way.

Fix: trim and filter empties to match what `placeSegments` does, or export and call it.

**2. Dedup key ignores hasNormalized**

Dedup key is `${ark.id}::${original}` — text only. If the same record has two facts with identical `place.original` but one has `normalized` and the other doesn't (a real FS data inconsistency), only the first-encountered state is kept. The `recapi supplied place.normalized` headline number silently miscounts.

Fix: include `hasNormalized` in the key, or keep both and note the inconsistency.

**3. placesOf counts facts toSimplified discards**

`placesOf` walks raw `body.relationships[].facts` directly, but `simplifyRelationship` lifts the first ParentChild-subtype-matching fact out of `facts[]` into a bare `subtype` string and discards its place entirely. The probe counts a place `record_read` never returns. Low likelihood for these specific ARKs (census/parish records rarely put places on parent-nature facts), but the comment claims to measure "every place record_read hands back" and that's not exact.

**4. Six filter passes with no sum check**

The summary computes `byConsistency` via four independent `.filter()` calls. No assertion that ok + contradiction + unverifiable + noResolve === total. If `countryConsistency` ever gains a fifth return value, rows carrying it vanish from every bucket silently. A reduce keyed by `r.consistency` or a sum check after the filters would catch that.

## Code reuse

**5. fetchRecord reimplements record-read's live fetch**

`RECAPI_BASE`, the URL build, the headers — all duplicated from `record-read.ts`. The error handling is thinner too (one generic `!res.ok` throw instead of distinct 401/403/404/429 messages). The sibling script `dev/try-record-read.ts` already calls `recordReadTool` directly — same pattern works here and eliminates the RECAPI_BASE copy.

**6. placesOf reimplements collectFacts**

The probe's own comment names `collectFacts` from `gedcomx-convert.ts` as what it mirrors. `collectFacts` is exported. Using `recordReadTool` + `collectFacts(result)` would also fix #3 for free, since `toSimplified` already handles the ParentChild subtype lift.

## Smaller

**7. Both fetch and resolve loops are fully serial.** `mapWithConcurrency` is already exported from the same `place-resolver.ts` this file imports from. `resolveStandardPlace` memoizes results, so parallelizing is safe. Matters for re-runnability — a multi-minute serial run discourages the re-runs the docstring calls for.

**8. `Ark.note` is declared on all 8 entries but never printed.** The table shows only category and bare ark id. Either thread it into the output or remove it.

**9. Duplicated example comment in record-read.ts.** The Southampton/Rochdale example appears in full in both `recordReadTool`'s comment (lines 128-137) and `readFromSidecar`'s comment (lines 173-185). This PR expands both. When the probe's results update this evidence, one copy is easy to miss. State it once and reference.

kofiatinka12:
Thanks — all nine addressed in `2edae319`, one probe-focused change (no runtime, no paid eval).

**The refactor (#3/#5/#6):** the probe now calls the real `recordReadTool({ recordId })` and reads `collectFacts(result)` — no more `RECAPI_BASE`/fetch copy and no hand-rolled `placesOf`. It measures the actual `record_read` output, and `toSimplified`'s ParentChild subtype-lift is handled for free, so #3 is gone too.

**Measurement-semantics note (deliberate):** the normalization metric is now **"tool output carried `standard_place`"** rather than "raw `normalized` array present." That's what `pickNormalizedPlace` actually produces from recapi's `normalized` values — i.e. what `record_read` returns — so it's the right number for the Phase 2 decision, but it is a slightly tighter measure than the pre-refactor probe reported; flagging it so the headline count isn't read as the same quantity.

**The rest:**
- #1 — `placeSegments` is now exported and used for the segment count, so it matches what `deriveContextName`/`countryConsistency` see (`"Springfield,, Illinois"` → 2, not 3).
- #2 — dedup key includes `hasStandardPlace`, so same-text facts with different normalization state are both kept.
- #4 — the four `.filter()` buckets are now a keyed reduction with a sum-check that throws if a future `countryConsistency` value isn't tallied.
- #7 — ARK reads and resolves run through the exported `mapWithConcurrency`; `resolveStandardPlace` memoizes and the map preserves order, so it stays deterministic.
- #8 — `note` is now printed in each table row.
- #9 — the Southampton/Rochdale example lives once (the toSimplified block) and `readFromSidecar` references it. Comment-only.

Probe type-checks, engine suite green (2765 passed / 1 skipped), record_read regression + sidecar guard still pass. Production `record_read` behavior unchanged.

chrisedeson (approved):
All nine items from my review are fixed in `2edae319`. Verified each:

1. Segment count now uses the real `placeSegments` (imported, not hand-rolled).
2. Dedup key includes `hasStandardPlace`.
3. Probe calls `recordReadTool` + `collectFacts` — measures the tool's actual output, not a reconstruction.
4. Consistency buckets use a keyed reduce with a sum check that throws on an unaccounted value.
5. No more `RECAPI_BASE` copy or hand-rolled fetch.
6. `collectFacts` imported directly — the ParentChild subtype-lift gap is gone for free.
7. Both phases use `mapWithConcurrency`.
8. `Ark.note` is printed in the per-place output.
9. Southampton/Rochdale example consolidated to one site; the second references it.

Clean rewrite of the probe. T-FEH's doc/comment accuracy work holds, and the probe now measures what `record_read` actually returns rather than a parallel reconstruction of it.

johnmarkpeterbrown (commented):
Nice work, Solomon. I ran the full pass on `117076ed` and found nothing blocking. The probe measures the tool's real output rather than reconstructing it, the doc corrections all check out against the code, and the new sidecar assertion genuinely fails when I break it — that last one was the gap from the earlier rounds and it's properly closed now.

**One thing before this merges, and it isn't about your code.**

The only approval here from a senior team reads `ece1380e`. Your probe rewrite landed after that. Chris reviewed the rewrite, but he isn't on either senior team, so as it stands no senior has read the current probe — which is the main deliverable of this PR.

Pushing won't fix it on its own. This repo doesn't dismiss approvals on push, so the old one survives and nothing re-requests automatically. You need to ask:

```bash
gh pr edit 2058 --add-reviewer T-FEH
```

**Two small edits worth making in the same push.**

1. The PR body says two files were left alone, but one of them changed.

> You wrote: `gedcomx-convert.ts` and `place-resolver.ts` are untouched.
>
> Change it to: `gedcomx-convert.ts` is untouched. `place-resolver.ts` changes by one keyword — `placeSegments` is exported so the probe can reuse the real segment splitter instead of hand-rolling one.

Same paragraph, the engine count has drifted since you wrote it. It's 2782 passed / 1 skipped now, not 2766 / 0.

2. In `docs/record-read-sidecar-scope.md`, the Findings bullet still calls the search result's place flatly correct. Line 33 of that same file — the provenance line you added — says it's a mixture, resolver-derived wherever FamilySearch supplied no normalized value. That share came from the same resolver that produced the two mis-resolutions quoted just below it. The sentence is line-wrapped in the file:

> You wrote: The sidecar is therefore *more* reliable, and the sidecar tool returns the staged place **as-is** (no `standardizePlaces` re-run).
>
> Change it to: The sidecar is therefore more reliable only for its FS-normalized share — its resolver-derived share came from the same resolver — and the sidecar tool returns the staged place **as-is** (no `standardizePlaces` re-run).

T-FEH already flagged this one and waived it as not worth another round, so treat it as optional. I'd take it because the wording is already written and the branch is open.

Everything else I checked — the #2092 deferral, the #1877 blocker, the acceptance-check sweep — holds up.

--

kofiatinka12:
Thanks, John Mark. Good catch that no senior had read the rewrite — re-requesting T-FEH so a senior reviews the current probe, not the pre-rewrite commit.

Folded in both edits (`d9bd0abd`): the body no longer claims `place-resolver.ts` is untouched — it exports `placeSegments` now (one keyword, so the probe reuses the real segment splitter instead of hand-rolling one), `gedcomx-convert.ts` stays untouched, and the count is corrected to 2782 passed / 1 skipped. And I took T-FEH's waived wording on the Findings bullet: the sidecar is more reliable *only for its FS-normalized share*, since the resolver-derived share came from the same resolver that produced the two mis-resolutions quoted just below it.

johnmarkpeterbrown (approved):
Approving.

Amending one thing from my comment above: I told you to chase T-FEH for a senior read of the head. That's not needed. Every path here is `*.ts`, and that CODEOWNERS line names both senior teams as an OR — I'm on `senior-genealogists`, so this approval satisfies the code-owner rule on the current head. The gap I flagged was "no senior has read the current probe." One has now.

Both edits I listed are **non-blocking**. Take the PR-body one when convenient — it's just accuracy about `place-resolver.ts` and the engine count. The scope-doc wording is optional; T-FEH already saw it and waived it after six rounds, and I agree it doesn't hold anything up.

What I'm approving on: the probe reads through the real tool rather than reconstructing its output, every corrected claim holds against the code, the new sidecar assertion genuinely fails when the rule is broken, and `record_read`'s runtime logic is untouched apart from the model-facing schema string. CI is green on `117076ed`.

PR #2060, opened 2026-08-31
project-status: delete the pointers to a reference file that never existed (#1750)

### LINKED ISSUE #1750: project-status: in_progress plan items, the skill-naming contradiction, and the exhaustiveness guide's coverage lists

benter-070:
Closes #1750

Finding 4 — the last open item on that card. Findings 1–3 landed earlier in
`0447b62a`.

## The defect

`project-status/SKILL.md` named `references/output-formats.md` twice as the render
source for both of its summaries. That file **has never existed on any ref** —
`git log --all -- '*output-formats.md'` is empty, so it was never written, not
deleted. The agent was told to read something it could not open and improvised
instead, visible in `ut_project_status_006`, which narrated *"Now let me check the
output-formats reference"* immediately before doing so.

Since this is the skill's whole output contract, the failure was silent: the unit
harness records MCP calls, not failed file reads.

## The decision, and why this way

The card left the choice open — write the file, or delete the pointers. **Deleted.**

Nothing stated is lost. SKILL.md already lists what each summary must carry (12
elements for the detailed GPS-state block, 5 for the plain-language story), and
`rubric.md`'s "Both summaries present" dimension grades on presence and
distinctness, independently enumerating the detailed report's required content. The
missing file would not have added a contract — it would have added a *layout*.

Writing a literal skeleton would instead **decide what the skill outputs**: a new
behavioural constraint on output that currently passes, with no evidence in any run
log that inconsistent shape is a problem. The reachability lint's own exemption
said as much ("writing it decides what the skill outputs and owes a paid eval run").

The two pointers are replaced by one line each, pointing at the contract already
above them.

## Also in this PR

**Empties `MISSING_PENDING_ADJUDICATION`** in `skill-reference-reachability.test.ts`.
That list is shrink-only by design, and this is the sanctioned direction — the
pointer it exempted is gone.

**Corrects that file's header comment.** It said these adjudications are "issue
#1112 owns". That is wrong for the missing-pointer half: #1112's body states *"Not
in scope: the `project-status` pointer to a file that was never written. Issue
#1750 ruled on 2026-08-20 that it moves there."* Left as-is, the comment sends the
next reader to a senior-assigned card that explicitly disclaims this work. Worth
noting because it nearly misrouted this fix.

**This unblocks #1112.** Its lint cannot land until both content halves are
settled; this settles the missing half.

## Verification

- **Proved the lint still bites:** reintroduced a dangling pointer, watched
  `every references/ file a SKILL.md names is on disk` fail, restored.
- Full MCP suite: **120 files, 2832 passed, 1 skipped** (was 115/2746 when this PR
  was opened — refreshed after the main merge, per @EdmondOware). He measured 2827
  passed at `1065eaa9`; I get 2832 at that same clean commit, so the five-test gap
  is an environment difference between our machines, not anything this branch does.
  The review fixes are net-neutral on the count — one test replaced by one test.
- Eval run `v1_2026-08-31_09-15-36`: **11 pass of 11**, $1.91, annotated.
- **The symptom is gone, not just the pointer:** no test in the run reaches for
  `output-formats` anywhere in its output, including `_006`, the test the card
  cited as evidence.

## Three limits worth stating rather than implying

Both come from the card itself:

1. **The run is a no-regression signal only.** With `runs_per_test` pinned to 1 it
   cannot show the change improved anything, and 11/11 should not be read as proof
   that it did.
2. **Nothing in CI checks that a SKILL.md formula enumerates an enum's legal
   values** — relevant to Findings 1–3, already merged, and stated here so the
   green tick is not mistaken for coverage of that.
3. **The 11/11 was scored against a judge prompt that has since changed.** The run
   log carries `judge_prompt_hash` `03f306ff`; the current `eval/harness/judge/prompt.md`
   hashes to `9451d105`, so the scores would likely move on a re-run and 11/11 should
   not be read as current. Rule 2b warns on exactly this and is non-blocking by design
   (judge edits are a separate cadence), so this is a caveat rather than a re-run.
   Raised by @EdmondOware; verified here rather than taken on faith.


EdmondOware (changes requested):
Deleting rather than writing is the right call, and the reason you gave is what makes it reviewable: the rubric grades both summaries on content coverage and distinctness, "graded independently of how good either one is", so there was no layout contract to lose. Proving the lint still bites by reintroducing a dangling pointer is the right instinct on a change that removes an exemption. Worth noting the symptom was wider than the card suggests, it shows up across five run logs and nine distinct tests, and the new run has none of it.

One thing to fix, and it is the same class of defect this PR exists to remove.

1. `docs/architecture.md`, lines 356 to 361, the paragraph beginning "The same test pins the other direction", still says "One is exempt" and then describes `project-status` naming `output-formats.md` twice as the render source, and explains that writing it "decides what the skill outputs, which is a content call owing a paid eval run rather than a mechanical fix". After this merges none of that holds. Nothing is exempt, the pointers are gone, and the call was made and paid for. No lint catches it, and architecture.md is where the next reader goes looking.

On the header correction you asked about: it reads right and the quote is verbatim. The reason it nearly misrouted you sits one layer down. #1112's title still reads "and fix the pointer to a file that never existed" while its body disclaims exactly that. Your comment routes a reader who reads the body; the title still misroutes one who does not. Since this settles the half #1112 was waiting on, that clause is worth taking off its title in the same pass.

On the run: your framing is right and I would not soften it. Eleven of eleven under `runs_per_test: 1` is a no regression signal and nothing more, and saying so in the body rather than letting the tick imply otherwise is the correct way to ship it.

benter-070:
Both fixed in `46a58de8`.

**1. `docs/architecture.md:356`.** You're right, and it's the same defect this PR removes — a document asserting a state that isn't true, in the place the next reader goes to check. Rewritten: the half now says nothing is exempt and the list is empty, then records what the one entry was and that #1750 settled it by deleting the pointers, with the reason (SKILL.md already states the content, rubric.md grades coverage and distinctness, so there was no layout contract to lose). Costs no re-run — `docs/` is in no run-log snapshot, which I checked rather than assumed.

**2. #1112's title.** Retitled to drop "and fix the pointer to a file that never existed". Your diagnosis of why it nearly misrouted me is better than mine was: I corrected the lint comment, which routes a reader who opens the body, and left the title still promising the opposite to anyone who doesn't. Both now agree.

On the wider symptom — thank you for measuring it. Five run logs and nine distinct tests is a stronger claim than the card's single `ut_project_status_006` citation, and I'd only checked that the new run was clean rather than how far back it went. Worth having on the record here since the card will close with it.

And noted on the run framing; I'll leave it as written.

benter-070:
Both review points are addressed, plus one correction that came in after your review.

**1. `docs/architecture.md` (46a58de8, bbf4186d).** The "One is exempt" paragraph now
reads "**Nothing is exempt from this half** — its exemption list is empty, so a dangling
pointer fails outright", and the "content call owing a paid eval run" sentence is gone,
replaced by what actually happened: both pointers deleted, because `SKILL.md` already
states what each summary must carry and `rubric.md` grades that coverage. I dropped the
issue-number citation I first wrote there — it tripped `doc-links.test.ts`'s
issue-reference ratchet, so the paragraph states the fact rather than the ticket.

**2. #1112's title.** The "and fix the pointer to a file that never existed" clause is
off it. It now reads "Skill references: delete or wire up the 7 unreachable files, and
settle the search-records/citation validation contradiction", so the title and the body
route a reader the same way.

**3. New since your review — 68802ded, from @dallanq.** My header comment said an
unnamed reference was "unreachable in every environment". That is not true and the
correction matters for the guard's framing: both environments put every skill's
`references/` folder in front of a session — `workspace.py` copies each skill dir into
`.claude/skills/`, and the Cowork .zip ships the whole plugin — so a globbing model can
read any of them. The comment now says an unnamed file is *never loaded deliberately*
but is *not out of reach*, and cites the measured instance
(`convert-dates/v1_2026-09-01_11-26-50`, `ut_convert_dates_012` globbed `**/*`, read
three exempt files, and one changed its answer). Worth a look since it postdates your
pass.

**4. Judge-prompt caveat you raised on #2112.** It belongs to this PR's run log, so it
is now limit 3 in the body: the 11/11 was scored at `judge_prompt_hash` `03f306ff` and
the current prompt hashes to `9451d105`, so those scores are not current. Verified both
hashes rather than repeating the numbers.

**State:** main merged (153 commits, clean; `architecture.md` auto-merged and I checked
the correction survived). Main changed nothing under `skills/project-status/` or
`eval/tests/unit/project-status/`, so the run log stays active — rule 2 is satisfied
without a re-run. Full packaging suite 635 pass / 34 files, reachability lint and
`doc-links` included. CI green.

Re-requesting review. This one also unblocks #2112 — that branch was cut from here by
mistake, so six of its seven files are yours to review here; once this merges its diff
drops to the single guard file on its own.

EdmondOware (approved):
Both asks are done, and I checked the second one rather than assuming. `architecture.md` now reads "Nothing is exempt from this half", and #1112's title has lost the clause that disclaimed this work.

The correction you made beyond the ask is the best thing in the PR. The header comment claimed an unnamed file is "unreachable in every environment", and you refuted that with a measurement instead of softening the wording. I verified it end to end: exactly three of the seven exempted files were read by `ut_convert_dates_012`, and the run log carries the literal call, a `Read` on `.claude\skills\convert-dates\references\calendar-conflicts.md` out of the staged workspace, which is the mechanism your comment names. A test whose own comment asserted safety was wrong about its premise, and you fixed the claim. That is the re-measured-not-reworded rule applied to your own file.

I also proved the lint bites rather than taking the note for it. Reintroducing a dangling pointer fails `every references/ file a SKILL.md names is on disk`; restoring gives 23 passed. The rest verifies too: `git log --all -- '*output-formats.md'` is empty, the new run log contains the string `output-formats` zero times, 11 of 11 at $1.91, the annotation covers all five sampled tests with no rule 3 violations, and the runlog gate is clean.

**One thing you should know before this merges, and it is not about this PR's contents.** #2112 carries all six of these files. `project-status/SKILL.md` and the run log are byte identical; `architecture.md` and the reachability test are reduced there, 11 lines against 16 and 22 against 49. Critically, **#2112 does not carry the globbing finding at all**. If it lands first this measured correction is lost or arrives as a conflict. This PR should go first and #2112 should then drop the duplicates. I have said the same on #2112.

Small nit, not blocking. In the reachability comment, "…changed its answer (#1112). Eleven such files shipped for months, nine of them copies of one" runs onto one long line and joins the new paragraph to the old one mid-thought. A wrap fixes it.

Also worth refreshing: the body says the MCP suite is 115 files and 2746 pass. On `1065eaa9` I get 120 files, 2827 passed and 1 skipped.

chesworthrm:
Reviewed this against the merge with main. The change itself is right — deleting the pointers rather than inventing a layout is the correct call. I verified the lint still bites with the exemption list empty, that emptying it is forced rather than optional, and that all six required checks ran on this head. The globbing correction you made beyond the ask is the best part of the PR; I re-measured it independently and it holds.

Two things to fix before this merges. Both are cheap, and neither is in the diff's logic.

## 1. The annotation on `ut_project_status_004` records a failure that didn't happen

The run log for that test shows the skill never activated, made no tool calls, created no files, and that its only action was handing off to `proof-conclusion`. That is the pass condition for a routing negative — the outcome is decided by where it routed, not by the judge's dimension scores.

The judge read the empty output as silence and scored Correctness and Completeness 1 each. The harness flagged both at the time, and its advisory says that 1 is only right if the skill under test did the work itself. It didn't.

The annotation confirms both 1s. The comment then tells the next reader the fix is to make project-status print a decline message. That would send someone into `project-status/SKILL.md` to solve a judge problem — and project-status has no chance to print anything on a run it never enters.

In the eval UI, change both corrected scores to 3 and replace the comment. For Correctness:

> Override. The routing did happen — the run shows `skills_invoked: ["proof-conclusion"]`, `activated: false`, zero tool calls and zero files created. The judge graded the empty output of a skill that correctly never ran. The harness advisory says the 1 is right only if the skill under test carried out the task inline; it did not. project-status cannot print a decline on a run it never enters.

Same for Completeness, with the dimension name changed.

Worth knowing: this file is byte-identical in #2112, so if it isn't corrected here it lands twice.

@EdmondOware — you checked this file for rule 3 coverage and it is clean. The score itself is a calibration call in your lane, so I am flagging rather than deciding: does 3 look right to you?

## 2. `docs/architecture.md` still carries the claim this PR disproves

You corrected it in three places. Two more instances sit further down the same file, in the duplicated-references section. #2112 does not fix them either.

Line 464:

```diff
- The last two were 11 and 3 until the unreachable copies were deleted. Nine
+ The last two were 11 and 3 until the unnamed copies were deleted. Nine
```

Line 466:

```diff
- named by no `SKILL.md`, so no session could load them; three of the nine also
+ named by no `SKILL.md`, so nothing loaded them deliberately; three of the nine also
```

Those copies were on disk and globbable when that sentence was written, so it was wrong then for the same reason line 354 was.

## Two nits, not blocking

Line 23 of the reachability test runs to 133 characters and joins the new paragraph to the old one mid-sentence:

```diff
- // exemption list below, and one of them changed its answer (#1112). Eleven such files shipped for months, nine of them copies of one
+ // exemption list below, and one of them changed its answer (#1112).
+ //
+ // Eleven such files shipped for months, nine of them copies of one
```

And the body says the MCP suite is 115 files and 2746 pass. On this head I get 120 files and 2828.

Once the annotation is corrected I am happy to approve. This should still merge before #2112.

docs/architecture.md
```diff
@@ -351,7 +351,7 @@ for material too long to sit in the skill body.
 
 **A reference is loaded only if its own `SKILL.md` names it** — or if a
 reference the body names links on to it. Nothing lists a `references/` folder at
-runtime, so a file neither route reaches is unreachable: it costs no prompt
+runtime, so a file neither route reaches is never loaded deliberately: it costs no prompt
```
florencemashipei: The hedge landed on "unreachable" but not on what follows. "It costs no prompt tokens" is still stated absolutely, and that is the half this PR's own evidence disproves — `ut_convert_dates_012` globbed and read three exemption-list files, and one of them changed the answer. The bold **"A reference is loaded only if its own `SKILL.md` names it"** two lines up has the same problem.

CLAUDE.md sends readers here as the explanation of how references load, so this page shouldn't contradict the measurement being added in the same PR. The test file already has the right phrasing — carry the "in the normal path" framing across to both spots here.

packages/engine/mcp-server/tests/packaging/skill-reference-reachability.test.ts
```diff
@@ -7,22 +7,38 @@ import { fileURLToPath } from "node:url";
...
+// and drift cost of a live file.
+//
+// It is NOT out of reach, and this comment used to claim it was. Both
+// environments put every skill's references/ folder in front of a session:
+// `eval/harness/harness/workspace.py` copies each skill directory into the
+// workspace's `.claude/skills/`, and the Cowork .zip ships the whole plugin. So
+// a model that globs can read any of them. Measured, not theorised —
+// `eval/runlogs/unit/convert-dates/v1_2026-09-01_11-26-50.json`
+// (`ut_convert_dates_012`) globbed `**/*` and read three files that are on the
+// exemption list below, and one of them changed its answer (#1112). Eleven such files shipped for months, nine of them copies of one
```
florencemashipei: The pre-existing "Eleven such files shipped for months…" sentence got welded onto the end of the new evidence line. "Such files" now resolves to the three files `ut_convert_dates_012` read, not the eleven orphans it was originally describing — so the comment reads as though eleven files were in the measurement, or as though only three orphans ever shipped.

Start it on its own line. That also fixes the wrap: this line runs well past the ~79 columns the rest of the file keeps to.

packages/engine/mcp-server/tests/packaging/skill-reference-reachability.test.ts
```diff
@@ -73,12 +90,7 @@ const EXEMPT = new Set(UNREACHED_PENDING_ADJUDICATION.map((e) => e.path));
 // The mirror failure: a SKILL.md naming a references/ file that is not there.
 // The body tells the agent to read a file it cannot open, and the agent
 // improvises. Same shrink-only rule as the list above.
-const MISSING_PENDING_ADJUDICATION: Array<{ path: string; why: string }> = [
-  {
-    path: "project-status/references/output-formats.md",
-    why: "named twice as the render source for both summaries; never existed in this repo's history, so writing it decides what the skill outputs and owes a paid eval run",
-  },
-];
+const MISSING_PENDING_ADJUDICATION: Array<{ path: string; why: string }> = [];
```
florencemashipei, line 96: Emptying this list is right — that's the point of the PR. But it leaves the test at line 196, "every missing-reference exemption is still named and still missing", looping over an empty array. The list only shrinks by rule, so it stays empty, `stale` can never be populated, and that test has no reachable failure path.

CLAUDE.md's own line applies: a check that cannot fail reads as coverage and is worse than none. Either delete it, or turn it into an assertion that the list is empty with a comment saying the shrink-only rule is what keeps it that way.

florencemashipei (changes requested):
Reviewed the change itself and the evidence behind it, and both hold up. No `output-formats` pointer survives outside the test comment; `npx vitest run tests/packaging/` passes 634 tests on this branch including the 23 reachability ones; the eval snapshot hash for `project-status/SKILL.md` matches the file shipped here, so the 11/11 scored the actual change; `check_runlogs.py` against the real merge-base reports all rules satisfied with exactly the one Rule 2b judge-prompt warning the PR body discloses; and the removed runlog is the harness's automatic 5-candidate retention prune, not an unexplained deletion. Thank you for the trail — it made this reviewable.

Three things to fix before merge. None is a correctness bug and none touches a snapshotted file, so all three are free to change — no eval re-run.

1. `docs/architecture.md:354` — the qualifier landed on half the sentence.
2. `skill-reference-reachability.test.ts:23` — "Eleven such files" now has the wrong antecedent.
3. `skill-reference-reachability.test.ts:196` — the exemption test can no longer fail.

Details inline.

One more that I am explicitly **not** asking you to change here. `project-status/SKILL.md:60` — "Shape each one as the project's state warrants" reads as a licence for free-form output, but `rubric.md` grades structure, not just presence: pass wants a structured GPS-state report (question status, GPS elements, counts, conflicts, hypotheses, exhaustiveness, proof tier), and partial fires on "a thin paragraph missing most GPS structure". A run could satisfy the new instruction and land on partial. The 11/11 was scored against this exact file so the risk is theoretical, not observed — and `project-status` has an eval suite, so editing that line now buys a fresh paid run plus an annotation pass. Not worth it for a hypothetical. Worth carrying into whichever issue next owns a `project-status` run.

benter-070:
All five code/doc fixes are in (`f4a7e0d6`). Neither file is snapshotted, so no eval
re-run — thank you both for checking that before asking.

## @florencemashipei

**1. `architecture.md:354` — right, the qualifier covered half the sentence.** I had
hedged "unreachable" and left "it costs no prompt tokens" and the bold "loaded only
if its own `SKILL.md` names it" standing absolutely, which are exactly the halves
this PR's measurement disproves. Both are now scoped to the normal path, with the
globbing finding stated on the page CLAUDE.md sends readers to.

**2. The antecedent.** You are right that "such files" had come to mean the three
files `ut_convert_dates_012` read. Split onto its own paragraph with an explicit
antecedent — "eleven files named by no `SKILL.md`" — which fixes the wrap too.

**3. The test that could not fail — this was the one worth catching.** Emptying the
list left that loop walking an array the shrink-only rule keeps empty, so `stale`
could never populate. I took your second option: a direct assertion that the list is
empty, which fails the moment an entry is re-added, i.e. the regression the rule
actually forbids. **Proven rather than asserted:** adding a temporary entry fails it
with the intended message, removing it gives 25 passed. Comment says to restore the
staleness loop in the same commit if a genuinely un-fixable pointer ever has to be
exempted.

**On `project-status/SKILL.md:60` — agreed, and deliberately not touched here.** Your
reading is right: "shape each one as the project's state warrants" could satisfy the
instruction and still land on partial against a rubric that grades GPS structure. The
11/11 was scored against this exact file so it is theoretical, and the fix costs a
paid run plus annotation. I have written it into #1750 so it travels with whatever
next owns a `project-status` run rather than dying in this thread.

## @chesworthrm

**`architecture.md:464`/`:466` — fixed, both as you wrote them.** Good catch that I
corrected three instances and left two; they were in the duplicated-references
section further down, which I had not re-read.

**The `ut_project_status_004` annotation — I agree with your reading and have not
changed it.** Two reasons, both worth stating rather than sitting on. First, unit
`.ann.json` files may only be written through the eval CRUD UI; hand-editing is what
produces the deprecated correction shape that later crashes the `check-runlogs` gate.
So this is a UI pass, which I will do rather than script. Second, you asked
@EdmondOware whether 3 is the right score and that is a calibration call in his lane —
I would rather he answers before the override is recorded, since the comment you
drafted becomes the durable justification. Edmond, over to you; the moment you confirm
I will make both corrections in the UI with your wording.

You are right that it lands twice if it is not fixed here — #2112 carries the file
byte-identical.

## @EdmondOware

Thank you for the approval and for proving the lint bites rather than taking my note
for it.

**Body counts refreshed.** They said 115 files / 2746 pass. I get **120 files, 2832
passed, 1 skipped**. Worth one line since we disagree: you reported 2827 at
`1065eaa9`, and I ran that same clean commit to check — 2832 there too. So the
five-test gap is between our environments, not anything this branch does, and my edits
are net-neutral (one test replaced by one).

**Agreed on the merge order, and it is now load-bearing.** This PR should go first.
#2112 carries all six files but does **not** carry the globbing correction, and it now
does not carry any of the five fixes above either — so landing it first would lose
them or surface them as conflicts. I will drop the duplicates from #2112 once this
merges.

benter-070:
## Status — everything asked for is done; blocked on two people

@EdmondOware asked where this is, so putting it in one place for all three reviewers.

| Review | Items | State |
|---|---|---|
| @EdmondOware | 2 | done — **approved** 2026-09-03 09:23 |
| @florencemashipei | 3 | done in `f4a7e0d6` — **awaiting re-review** |
| @chesworthrm | 2 | `architecture.md` done; **annotation waiting on Edmond** |

**Branch is current.** Merged main just now (20 commits, `bff694ee`). Checked before
pushing rather than after: main touched 13 fixture files, none of them in this run
log's snapshot, so all 34 entries still match and **no paid re-run is needed**. CI
green, packaging suite 637 passed.

### What changed since the reviews

Florence's third point was the one worth catching. Emptying `MISSING_PENDING_ADJUDICATION`
left the staleness test at `:196` iterating an array the shrink-only rule keeps empty,
so `stale` could never populate and the test had **no reachable failure path** — coverage
that checks nothing, in the PR whose whole purpose is removing that class of defect.
Replaced with a direct emptiness assertion, which fails the moment an entry is re-added.
Proven by adding a temporary entry and watching it fail, then removing it.

Her other two — the `architecture.md` qualifier that had landed on only half the
sentence, and the "Eleven such files" antecedent — are fixed, along with Richard's two
further instances of the same claim further down the file that I had missed on the
first pass.

### The one thing that is not mine to finish

@EdmondOware — @chesworthrm asked **you** whether `3` is the right score for the
`ut_project_status_004` override. Your approval came in at 09:23 and his question at
10:13, so it may simply not have crossed your screen.

I am deliberately not recording it until you answer: it is a calibration call in your
lane, unit `.ann.json` files may only be written through the eval CRUD UI, and the
comment Richard drafted becomes the durable justification for the override. Say the
word and I will make both corrections in the UI with his wording.

Worth knowing it is not optional — #2112 carries that annotation **byte-identical**
(blob `ee480c97`), so if it is not corrected here it lands twice.

### Merge order

This PR first, then #2112. #2112 has none of the five fixes above and never carried the
globbing correction at all, so landing it first would lose them or surface them as
conflicts. I will drop the six duplicated files from #2112 once this merges.

florencemashipei (changes requested):
All three of my asks are done, and I checked each one rather than taking the note for it. Head `dfc6e142c`.

1. **`architecture.md`** — the qualifier now carries across both claims either side of it, not just the middle one, and the globbing finding is stated outright on the page CLAUDE.md sends readers to. Richard's two further instances at `:464`/`:466` got the same treatment.
2. **The test comment** — antecedent is explicit ("eleven files named by no `SKILL.md`"), and every line in that block is now inside 80 columns.
3. **The emptiness assertion** — proved it bites rather than reading it. Added a probe entry to `MISSING_PENDING_ADJUDICATION`; `the missing-reference exemption list stays empty` fails with the intended message. Removed it; 23 passed.

Re-verified independently while I was in there: no `output-formats` pointer survives anywhere under `packages/engine/plugin/` or `docs/`, and `diff_snapshot_vs_disk` reports **zero drift across all 34 snapshot entries** at this head — so the two main merges did not invalidate the run and no paid re-run is owed. Your call on that was right, and checking before pushing rather than after is the right order.

One number to fix in passing: the commit message and the status comment both say "removing it gives 25 passed". That file has 23 tests and the swap was net-neutral — I get 23 either way, as Edmond did. Nothing turns on it, but it is wrong in two places.

## Not approving yet — and this is a routing fix, not a new ask

The `ut_project_status_004` annotation is waiting on @EdmondOware, and it should not be. `.github/CODEOWNERS` names `/eval/runlogs/` **@PioneerAIAcademy/senior-genealogists first**, and its own header says the first-listed team "is the one whose judgment the path normally wants". `docs/task-lifecycle.md:226` puts "run-log annotation" under `genealogist`, not `developer`. Richard read the score as a calibration call in Edmond's lane and you held for it in good faith — but that is the same class of misroute this PR already caught in the reachability header comment. It is my call, so here it is.

**Ruling: override both to 3.** Richard's diagnosis holds and I re-measured it:

- `test_type: negative`, `grading_mode: routing`, `dimensions_gate_outcome: false`, `outcome: pass` — the dimension scores never gated this test.
- The run: `activated: false`, `skills_invoked: ["proof-conclusion"]`, `tool_calls: 0`, `files_created: 0`, `num_turns: 0`, and the single builtin call is `Skill(proof-conclusion, "Write the proof statement for Patrick Flynn's parentage.")`.
- Both advisories are `routing_negative_judge_fail`, and their own stated condition is *"if the skill under test carried out its own task inline, the 1 is right"*. It did not: zero tool calls, zero files, never activated.

**Read the base rate before treating this as routine.** The advisory says a human confirmed that 1 in **20 of 24** such cells. So this is a minority override, and that is precisely why the comment has to carry its evidence inline rather than assert a category — Richard's wording does exactly that. Use it as he wrote it.

**Why this is worth blocking on rather than deferring.** The current comment tells the next reader the fix is to make `project-status` print a decline message. That sends someone into `project-status/SKILL.md` — a snapshotted file — to fix a judge artifact, and editing it buys a paid run plus a fresh annotation pass for nothing. An annotation that prescribes a $10 no-op is worse than an uncorrected score, and #2112 carries the file byte-identical, so it lands twice uncorrected.

**Correcting it is free.** The `.ann.json` is not in the snapshot — all 34 entries sit under `eval/fixtures/`, `eval/tests/` and `packages/engine/` — so no re-run. Make both corrections in the CRUD UI with Richard's wording and I will approve on sight.

@EdmondOware — no answer needed from you on the score, and nothing here disagrees with your approval. The runlogs path is the genealogists' first, so I have taken it rather than leave the PR parked.

On merge order: agreed, and I have flagged it on #2112 rather than only here.

florencemashipei:
@benter-070 — the annotation call was mine, not Edmond's, so you are not waiting on anyone.

`.github/CODEOWNERS` names `/eval/runlogs/` **@PioneerAIAcademy/senior-genealogists first**, and `docs/task-lifecycle.md:226` puts "run-log annotation" in the `genealogist` lane. The ruling is in [my review above](https://github.com/PioneerAIAcademy/cowork-genealogy/pull/2060#pullrequestreview-5112703503): override `ut_project_status_004` to **3** on both `Correctness` and `Completeness`, with Richard's wording. You were right to hold it for the CRUD UI rather than script it — just not right about whose word you were holding for.

The text for both dimensions, from Richard's comment:

> Override. The routing did happen — the run shows `skills_invoked: ["proof-conclusion"]`, `activated: false`, zero tool calls and zero files created. The judge graded the empty output of a skill that correctly never ran. The harness advisory says the 1 is right only if the skill under test carried out the task inline; it did not. project-status cannot print a decline on a run it never enters.

One thing so you don't go hunting: Richard wrote "same for Completeness, with the dimension name changed", but that text carries no dimension name — it goes in **verbatim for both**. The only per-dimension difference is the annotation's own header line.

It is free — the `.ann.json` is not in the snapshot, so no re-run. Everything else you fixed checks out and I have resolved those threads. I approve as soon as this lands.

benter-070:
@florencemashipei — taking the ruling, and you are right about the count. Both are
worth writing down properly.

## The annotation

Accepted, and I have stopped waiting on @EdmondOware. I checked the routing lane
before acting on it rather than taking it on authority: `.github/CODEOWNERS` does
name `@PioneerAIAcademy/senior-genealogists` first on `/eval/runlogs/`, and
`docs/task-lifecycle.md:226` does put run-log annotation in the `genealogist` lane.
I misrouted it, and holding a free correction behind the wrong person's answer is
the more expensive half of that mistake.

I re-verified the evidence independently before recording a durable override. Every
figure in your ruling checks out — though not where I first looked, which is worth
noting for whoever reads this next: `activated`, `skills_invoked`, `tool_calls` and
`files_created` are **not** run-level keys, they live inside the `output` string.
Reading them off the run entry returns `None` for all four and looks like a
contradiction of your ruling. Parsed properly:

```
"activated": false, "skills_invoked": ["proof-conclusion"],
"tool_calls": [], "files_created": [],
"builtin_tool_calls": [{"tool": "Skill", "args": {"skill": "proof-conclusion", ...}}]
```

plus `test_type: negative`, `grading_mode: routing`, `dimensions_gate_outcome: false`,
`outcome: pass`, `num_turns: 0`, and the `routing_negative_judge_fail` advisory
carrying the 20-of-24 base rate you cite. Both dimensions go to **3** with Richard's
wording verbatim, in the CRUD UI. Confirmed free: the `.ann.json` is in none of the
run log's 34 snapshot entries.

Your point about the annotation prescribing a $10 no-op is the part I had not
weighed. An uncorrected score is a wrong number; an annotation that sends the next
person into a snapshotted file to fix a judge artifact is a wrong number that spends
money. Those are not the same thing and I was treating them as if they were.

## The count: 23 is right, and the cause is on my machine

I measured 25 at `dfc6e142` before conceding, so here is why we disagreed — it is
not a miscount on either side.

`origin/main` has **19** skills with a `references/` dir. My working tree had **21**,
and this branch changes no tracked file under any `references/`. The two extras were
`proof-conclusion/references` and `research-exhaustiveness/references`: **empty
directories, untracked, dated 26 Aug**. Git cannot track an empty directory, so they
appear in no diff and not even in `git status --untracked-files=all` — but this suite
generates one `it` per skill with a `references/` dir, so my local runs produced two
phantom passes. Deleted them; I now get **23**, matching you and Edmond.

So the number was wrong in the two places you name, and the reason it survived review
is that it was locally reproducible.

One thing worth keeping from it, given what this PR is about: **an empty
`references/` dir yields a silently-passing test** in this very suite. Not a defect I
would block on and not in this PR's scope, but it is the same family as the check
that could not fail — a test whose subject does not exist still reports green.

Correcting the annotation now.

benter-070:
@florencemashipei — the override is in (`4311788a`). `ut_project_status_004`,
`base`/`Correctness` and `base`/`Completeness`, both **1 → 3**, with Richard's
wording verbatim on each.

Verified rather than assumed, and one detail is worth leaving here for the next
person who checks this: the four fields the diagnosis rests on are **not
run-level keys**. They sit inside the `output` string, so reading `activated`,
`skills_invoked`, `tool_calls` and `files_created` off the run entry returns
`None` for all four and reads as a contradiction of your ruling. Parsed properly
it is exactly as you and Richard describe — `activated: false`,
`skills_invoked: ["proof-conclusion"]`, both arrays empty, one builtin
`Skill(proof-conclusion, …)` call, `num_turns: 0`.

Checks after the edit: `.ann.json` valid against `ann.schema.json`, rule 3 still
covers every dimension of all five sampled tests, and the run log is still active
— 34/34 snapshot entries match, including after merging main just now. CI green.

**One thing you should know about how it got there**, since annotation is your
lane and this is not a one-off. The CRUD UI mangled this correction three separate
ways in the course of one fix:

1. **Appended instead of replacing.** Each save stacked another copy of the
   auto-filled header plus the judge's original rationale, so Correctness reached
   1743 characters reading `Junior: 1` twice and `Junior: 3` once before arriving
   at the override.
2. **Silently did not save.** One round looked correct on screen; the file's mtime
   and hash were unchanged.
3. **Wrote to the wrong test.** A later save moved Richard's text onto
   `ut_project_status_006` — a 3/3 pass — where it asserted `activated: false` and
   zero tool calls about a run that produced a full GPS summary with 9 sources and
   13 assertions. Caught by diffing before commit and reverted; it is not in the
   committed file, and `_006` is back to 3/3 with its own rationale.

Nothing here is in the record — the committed annotation is clean, and I checked
that Richard's text appears on `_004` and nowhere else. But (3) is the one that
worries me: it is silent, it is plausible-looking, and it puts one test's evidence
onto another where nothing downstream would flag it. The same append behaviour hit
the init-project annotation on #2159 earlier today, so it is not specific to this
run log.

I will file it as a card rather than leave it in a PR thread. Flagging it to you
first since the annotation lane is yours and you may want it scoped differently —
or may already know the cause.

Branch is current with main (`#2159` landed while I was in here) and the packaging
suite is 635 passing. Over to you.

florencemashipei:
@benter-070 — the scores are right and the `_006` risk is genuinely clear, but the `Correctness` comment that got committed is the mangled one. Not approving yet; it is one field.

## What landed correctly

`4311788a5`, head `c19b355b0`. Both `_004` dimensions carry `corrected_score: 3`. Only the `.ann.json` changed, CI is green, and I re-ran the snapshot check myself at the new head — 34/34 entries, zero drift after the `#2159` merge.

**And the thing you were most worried about is clean.** I checked where Richard's text actually lives rather than taking it: `Override. The routing did happen` appears on `_004`/`Correctness` and `_004`/`Completeness` and **nowhere else**. `_006` is back to 3/3 on both base dimensions with its own rationales intact, 957 and 980 characters, the 9-sources/13-assertions one. Defect (3) is not in the record. You checked the right thing.

## What is in the record

Defect (1) is. Parsing the committed blob, `_004`/`Correctness` is **1743 characters** — the exact figure from your own 14:51 comment — and it reads:

| | |
|---|---|
| header 1 | `LLM: 1 → Junior: 1` |
| header 2 | `LLM: 1 → Junior: 3` |
| header 3 | `LLM: 1 → Junior: 1` |

with the judge's rationale repeated three times and Richard's override buried at the end after a second `(Pass - ` opener. So the version you described as caught and reverted is the version that got committed.

Separately, and in **both** `_004` comments: the header line reads `LLM: 1 → Junior: 1` while `corrected_score` is `3`. The human-readable half of the annotation contradicts the machine-readable half.

## Why I am holding for prose

My ruling rested on this being a *minority* override — the advisory records a human confirming that 1 in 20 of 24 such cells. That is the whole reason the comment has to carry its evidence legibly instead of asserting a category. A comment that says `Junior: 1`, then `Junior: 3`, then `Junior: 1` before arriving at the override reads as an annotator disagreeing with themselves, and it is the durable record that outlives this thread. That is worse than the wrong score was, because the wrong score at least said one thing.

Nothing in CI disagrees with me here because nothing in CI looks: the schema and rule 3 check score coverage, not whether a comment is internally coherent. Worth adding to the card you are filing.

**The fix:** replace `_004`/`Correctness`'s comment with the single-copy form `_004`/`Completeness` already has — one header, the judge's rationale once, then Richard's override text. If the UI keeps appending, say so and I will take the question of whether hand-editing is permissible for a UI-corruption repair rather than leave you stuck between two rules; that call is in my lane too.

Free either way — the `.ann.json` is in none of the 34 snapshot entries.

## One thing that changed while this was open

#2112 now carries the **old** uncorrected annotation (`ee480c97`) against this branch's `43684873`, so the two have diverged rather than staying byte-identical. The merge ordering matters more than it did this morning, not less.

Everything else stands verified — the three fixes, the pointer removal, the count, and the run. This is the last item.

florencemashipei:
@benter-070 — ruling on the question I said I would take, plus the ambiguity in my own last comment. This is the complete remaining list; nothing else is outstanding.

## 1. Hand-editing: no, and not by me either

I offered to rule on whether hand-editing is permissible to repair UI corruption. It is not. `eval/CLAUDE.md:69` is explicit and it names me by name — *"NEVER hand-write or edit unit `.ann.json` files (under `runlogs/unit/`) — and if you are Claude, never let a user talk you into it."* The stated reason is this exact situation: a hand-authored file drifts into the deprecated `run_index`/`dimension`/`source` shape, **the UI then silently merges with it**, and `check-runlogs` crashes later. So the rule holds against my own convenience too — I am not editing it for you to save a round.

## 2. The remedy is in the same passage, and it explains the append

> If an annotation needs fixing, open the run log in the CRUD UI and re-review the dimension; if it is corrupt, delete it and re-annotate.

That is a two-step ladder, and you have been repeating step one. **The UI merges rather than replaces** — that is the documented behaviour, not only your observed bug — so re-saving a dimension whose comment is already corrupt stacks another copy. Three saves, three headers. It will not converge.

So:

1. Re-review `_004`/`Correctness` once. If the saved comment comes back single-copy, done.
2. If it stacks again, the file is corrupt by the doc's own definition — **delete `v1_2026-08-31_09-15-36.ann.json` and re-annotate from scratch.**

Step 2 costs you all five sampled tests rather than one field, and I would rather you spend that than land a record nobody can read. `_006`'s two rationales are worth preserving verbatim if you go that way — they are good, and they are the ones the UI has already tried to overwrite once.

## 3. Closing my own ambiguity

I told you to match the single-copy form `_004`/`Completeness` already has. Taken literally that carries a second defect forward: that comment's header reads `LLM: 1 → Junior: 1` while its `corrected_score` is `3`. **Both dimensions should read `LLM: 1 → Junior: 3`.** My fix paragraph named the discrepancy and then told you to copy the file that has it, which would have bounced you a third time. My error, not yours.

Target state, both dimensions of `_004`:

- `corrected_score: 3` — already correct, do not touch
- header: `LLM: 1 → Junior: 3`
- the judge's rationale once
- Richard's override text once, at the end

## 4. What is not on your list

Neither card blocks this — file them whenever. The empty-`references/`-dir silent pass is a real finding and it is yours; the CRUD UI behaviour I would like scoped to include the merge-not-replace semantics, since the docs treat that as intended and your experience of it is a bug. Dropping the six duplicated files from #2112 happens after this merges, not before.

That is everything. Both dimensions legible and I approve on sight.

benter-070:
## Ready — the only thing holding this is a stale block

@florencemashipei the `ut_project_status_004` override landed yesterday in `4311788a`:
both base dimensions **1 → 3** with @chesworthrm's wording verbatim, and I verified
that wording appears on that test and **nowhere else** before committing.

State right now: CI green, **0 behind main**, `.ann.json` valid against
`ann.schema.json`, rule 3 still covering every dimension of all five sampled tests,
and the run log still active at 34/34 snapshot entries.

**I think your approve-on-sight was simply overtaken by the #2112 hold.** Your review
there is timestamped 16:09 UTC and my override landed 15:39 UTC — half an hour
earlier — and you cite the corrected blob `43684873` against this branch and
`ee480c97` against #2112, which is exactly what is on the two branches now. So the
correction is already reflected in your own record of it; only the
`CHANGES_REQUESTED` from yesterday lunchtime is still standing.

Nothing owed from me here that I can find. If something is still outstanding I would
rather hear it than sit on a green PR.

## On the ordering hold — agreed, and thank you for making it a mechanism

Doing it as a review rather than a comment is the right call for the reason you gave.
Both threads read correctly to anyone who has followed them; the exposure is the next
person to work the board, who sees an approval and no objection.

Your divergence table is what I would have had to reconstruct by hand, so for the
record I re-derived the two halves that matter:

- **The annotation has diverged** — `43684873` here vs `ee480c97` on #2112. This
  branch carries the corrected `_004`; that one still carries the version whose
  comment sends the next reader into `project-status/SKILL.md`, a snapshotted file, to
  fix a judge artifact. Landing #2112 first ships that instruction.
- **The two reduced files are the real loss** — #2112 has `architecture.md` at 7/-4
  and the reachability test at 12/-10 against 23/-11 and 50/-29 here, and it carries
  the globbing correction in no form at all. That correction is the measured one:
  `ut_convert_dates_012` globbed `**/*`, read three files off that test's own
  exemption list, and one changed its answer. A rebase conflict resolved by whoever
  holds it is the one edit nobody reviews.

**Once this merges I will drop the six duplicated files from #2112 the same day**, and
ping you to convert the hold into whatever the queue wants — an approval or a real
review of the guard.

@EdmondOware @chesworthrm — no action from either of you; recording it here so the
sequencing is visible in one place rather than across three threads.

florencemashipei (approved):
Approved. Merge ahead of #2112.

Everything substantive is verified and has been for a while: the pointers are gone, the three fixes from my first review hold, `architecture.md` carries the globbing correction, the emptiness assertion has a reachable failure path (I proved it — probe entry fails with the intended message, 23 pass without), and the `_004` scores are correctly overridden to 3 on both base dimensions. Snapshot clean at 34/34 at `c19b355b0`, CI green, no drift after the `#2159` merge.

**One thing is landing known-broken, and I am recording it rather than letting the green tick imply otherwise.**

`_004`/`Correctness`'s comment is the stacked version. At the current head:

```
_004 Correctness   corrected=3   len=1743   headers=[Junior: 1, Junior: 3, Junior: 1]   rationale x3
_004 Completeness  corrected=3   len= 781   headers=[Junior: 1]
```

So the durable justification for this override reads `Junior: 1`, then `Junior: 3`, then `Junior: 1` before arriving at Richard's text, with the judge's rationale repeated three times. Both comments' header lines also say `Junior: 1` against a `corrected_score` of `3`.

@benter-070 — your inference was the one thing that did not hold. My #2112 review cites `43684873` for **score** divergence between the branches; that same blob carries both the corrected scores and the stacked comment, so citing it was not a clean bill of health for the file. The scores have been right since `4311788a` and I said so at 16:01; the comments are what I was holding for. Not a failure of diligence on your part — you verified Richard's text was on `_004` and nowhere else, which was the more dangerous defect and is genuinely clean.

I am approving anyway because the machine-readable half is correct — `calibrate_judge` and rule 3 read `corrected_score`, and both say 3 — and because a third round through a UI that merges rather than replaces is not obviously going to produce a cleaner file than this one.

**Where the finding goes, so it is not stranded here.** This is a shipped instance of the CRUD UI append bug, not a one-off: it survived a reviewer asking for it twice, and nothing in CI looks at comment coherence — the schema and rule 3 check score coverage only. Put it in the card you are filing, with this measurement in the card rather than a link back to this thread, and include the merge-not-replace semantics: `eval/CLAUDE.md` treats merging as intended while your experience of it is corruption, and that gap is the actual defect. Worth a rule-3 companion check that a correction's comment contains exactly one header — cheap, and it would have caught this before review.

The `_004` comment itself is free to repair (the `.ann.json` is in none of the 34 snapshot entries), so whichever issue next owns a `project-status` run can fix it in passing. Also still open from my first review and already recorded on #1750: `project-status/SKILL.md:60` — "shape each one as the project's state warrants" against a rubric that grades GPS structure.

Once this merges, drop the six duplicated files from #2112 and ping me — I will convert the ordering hold there into a real review of the guard.

jud-sdev (approved):
Went through this on the current head, and it holds up. The core change is the right kind of fix: the old staleness loop iterated a list that this PR empties, so it walked nothing and read as coverage while checking nothing — the exact thing CLAUDE.md warns is worse than no check. Replacing it with a direct emptiness assertion gives it a real failure path, and emptying MISSING_EXEMPT re-arms the dangling-pointer guard so any new one fails outright. I traced both mutation paths by hand and they reach a genuine failure, not a vacuous pass.

I couldn't run vitest in my environment, so I'm leaning on CI (green at this head) plus Edmond's and Florence's independent mutation checks rather than my own run — flagging that honestly.

The annotation is correct where it counts: `ut_project_status_004` is overridden to 3 on both base dimensions, the shape is the current one, and the rationale is sound — it's a routing negative that correctly routed to proof-conclusion and never activated, so the judge graded the empty output of a skill that never ran. I saw the same stacked-comment artifact Florence recorded, and I agree with how she handled it: the machine-readable score is right, it's the CRUD UI's merge-not-replace behaviour rather than anything wrong here, and it belongs in the card plus a rule-3 companion check, not in a third UI round.

PR #2064, opened 2026-08-31
Detect searches that never reached the research log (#2056)

### LINKED ISSUE #2056: A search run outside the search skills logs nothing — the `log` row's only enforcement plane is a paid per-skill eval run
**Touches:** packages/engine/mcp-server/src/tools/record-search.ts, docs/specs/schemas/ownership.json, docs/specs/guardrail-enforcement-spec.md, eval/harness/validators/test_universal.py, eval/harness/e2e/

From alpha feedback #1985 (hosted web, 2026-08-27). The tester asked one research
question — Virginia census work on `218F-H7B` — and got a good answer in chat. The
project recorded none of it.

## What the bundle shows

- 11 `record_search`, 6 `record_read`, 2 `collections_search`.
- 4 of those searches staged results into the project and ranked **154 candidate
  rows** into `results/match-scores.jsonl`.
- 0 `research_log_append` calls. `research.json` came back with `log`, `sources`,
  `assertions`, `person_evidence`, `conflicts`, `hypotheses`, `timelines`,
  `proof_summaries` all `[]`.
- 1 `Skill` call in the session: `init-project`. The census question was answered
  free-hand on the main thread — no `search-records`, no `/research`.

So the nil results on the 1820–1840 censuses — the evidence that those query shapes
fail, which is what GPS Component 1 wants recorded — exist only in chat scrollback.

## Why nothing caught it

`search-records/SKILL.md:630` states the rule plainly: "A search without a log entry
is a search that didn't happen." That rule lives in a skill that was never invoked.

`ownership.json`'s `log` row declares `enforceableAt: ["unit"]`, and that manifest
describes the unit plane as running "only inside a paid per-skill eval run". So the
log's only enforcement plane is the per-skill eval of the five skills that own it. A
session that invokes none of them has no plane left. `test_universal.py` checks
append-only, ownership and write-then-validate; nothing asserts that a search
produced an entry.

The hook plane cannot take this one. `log`'s owner is `null` with five **skill**
callers, and a skill runs on the main thread with no `agent_id` — the hook's only
discriminator (`guard_project_files.py:296`) is subagent-vs-main-thread, so it cannot
tell a skill-driven search from a free-hand one.

## Proposed fix

Claim the `tool` plane for the `log` row — per the manifest, the only plane that is
caller-agnostic and binds in every environment. A staged search with no log entry is
decidable from the project documents alone (`results/.staging/` handles vs
`log[].results_ref`), which is the ADR-0011 test for a tool rule.

Recommended shape: a non-blocking obligation on the next `record_search` response,
following `RANKING_SKIPPED_NOTE`'s precedent in the same file (short, model-facing,
per-call cost accepted against a measured rate), plus an e2e detector so the rate is
visible. **The lead's call:** note vs. deny. A deny binds harder but would refuse a
legitimate broad survey search, and #1073 constrained `rankingSkipped` to args-only
with no staging read — a state-reading check is new ground in this file, not an
extension of that one.

`fulltext_search` and `external_links_search` stage the same way and get the same
fix from the shared helper. `image_search` does **not** stage, so this signal cannot
reach it — same obligation, no mechanism, tracked separately.

## Not covered by

- #1779 — log **fidelity**: an entry that claims a filter the search never sent.
  Different defect class; that one is about entries that exist.
- #1851, #2039 — a gate skipped *inside* the pipeline. This question never entered it.

johnmarkpeterbrown:
**Tier:** Normal. Closes #2056.

## What was broken

A research question answered without invoking one of the five search/extraction
skills wrote nothing to `research.json`, and no plane noticed. The rule is already
written — `search-records/SKILL.md:630`, "a search without a log entry is a search
that didn't happen" — but it is only read when that skill is invoked.

The alpha-feedback session behind #2056 answered a Virginia census question with 11
`record_search` calls, 6 `record_read` calls, **one** `Skill` invocation
(`init-project`) and **zero** `research_log_append` calls. 154 scored candidate rows
reached `results/match-scores.jsonl` while `log[]` stayed empty — the tooling
recorded its own work; the agent recorded none. The nil results on the 1820–1840
censuses, which are what a reasonably exhaustive search is most obliged to record,
died with the session.

Neither of the other planes could take it. The unit plane grades one skill's own run,
inside a paid eval. The hook plane discriminates subagent-vs-main-thread
(`guard_project_files.py:296`) and `log`'s callers are five **skills**, which run on
the main thread — the guard file says so itself: "a section owned by a SKILL cannot be
enforced through OWNED_SECTIONS".

## What this adds

Two advisory fields on **all three staging tools** — `record_search`,
`fulltext_search` and `external_links_search` — serialized before `results` (per that
spec's `#### Key order`) and withheld from the staged payload:

| Field | Fires when |
|---|---|
| `unloggedSearches` | this project holds staged responses that no log entry accounts for |
| `nilSearchNeedsLog` | `projectPath` was supplied and the search returned nothing |

**Nothing is refused.** No deny, no error, no changed exit path.

The count **pairs** each staged file against at most one log entry that recorded
results with no `results_ref` — same `tool`, `performed` at or after the file's
`retrieved` — and counts only unpaired files. It is read **before** this call stages
its own response, or it would fire on the first search of every session.

Two things worth knowing about why it is shaped that way:

- **Subtraction is wrong.** The unattached population also contains searches made
  with no `projectPath`, which stage nothing (`research-log-append.ts` says so, in the comment ending "turning a lossy log into no log at all"),
  so subtracting the whole set silently zeroes a real backlog. Measured over all 362
  committed `research.json` files: 126 of 1242 non-nil staging-capable entries sit in
  that population.
- **A nil search cannot be seen this way.** It stages nothing. Hence the separate
  args-only note, which is the only signal for the case the feedback actually hit.

`STAGING_CAPABLE_TOOLS` moves into `results-staging.ts`, imported by both the new
reader and `research-log-append.ts`; a `utils/` → `tools/` import would break the
no-util→tool rule and a second copy would drift.

The harness mirrors both notes and **calls the compiled counter** rather than
restating the pairing rule in Python. `_MOCK_ENRICHED_KEYS` gains both field names —
a nil search carries none of `staged`/`ranked`/`rankingSkipped`, so without that the
note is stripped from every run log and the measurement below has nothing to read.

## Reviewer: one thing to argue with

`ownership.json`'s `log` row now claims the `tool` plane, and **the plan critic
argued that overstates it** — the manifest defines that plane as "a writer-tool
precondition… it constrains the WRITE", while this adds an advisory note to two
*search* tools and constrains nothing. The packaging test only checks that the plane
name is declared, so nothing catches the mismatch mechanically.

The finding was accepted as correct and overridden by decision. The row's new `notes`
field states plainly that the claim covers **detection only**. If you would rather
the row stay `["unit"]` with this recorded as a shadow signal in
`guardrail-enforcement-spec.md`, that is a one-line change and I have no objection.

## Deviated from the plan

`docs/task-lifecycle.md` caps planning at two `/critique-plan` rounds and sends a
surviving blocking finding to the lead. Round 2 returned two, both verified in code,
and the decision was to fix in place rather than escalate — so this shipped on a
third planning round. Both were mechanical design errors with the fix named by the
critic, not scope ambiguity, but the override belongs on the record: each of my two
revisions introduced a *new* wrong mechanism, which is the pattern that rule exists
to catch.

## Reviewer commit

`f094271e` was pushed by @clack391 during the senior-developer review rather than
handed back, on the lead's guidance that a minor issue found in review can be fixed
in place. It coerces `results_available` once so the retained-none warning reads the
same value the field is persisted from, and adds the test that holds the mock's
nil-gate wiring. Reasoning, and the two overreaches caught in that fix before it
landed, are in the review comment below.

## Verification

Every acceptance check was watched to fail before it passed:

- Swapped the pairing for revision 2's subtraction → `still counts an unlogged staged
  file when the project also holds an older logged-without-sidecar entry` and `does
  not pair across tools` both fail. Restored → 20/20 pass.
- Keyed the nil note on `results` instead of the pre-filter set → `does not ask for a
  negative log entry when the host filter emptied results but links were staged`
  fails. Restored → 25/25 pass.
- Reworded the TS note → the Python drift test fails. Restored → passes.

```
cd packages/engine/mcp-server && npx vitest run     # 118 files, 2808 passed, 1 skipped
cd eval/harness && uv run pytest tests/unit -q      # 3049 passed, 2 skipped (Linux)
```

`tests/packaging/` (616) covers the manifest and the measured-figures ratchet; the
figures live in `ownership.json`'s `notes` and this body, neither of which that lint
scans, and are comma-free where it does.

## Folded in rather than deferred

**The `fulltext_search` arm.** It was scoped out on a sequencing argument that did
not survive checking: this branch already modifies `mock_mcp.py` and
`test_mock_mcp.py`, 2 of the 5 files #2001 touches, so that collision was already
incurred. Folding it adds conflicts on two more files — the fulltext spec (#2001,
#1978) and `fulltext-search.test.ts` (#1978) — both additive edits. Three staging
tools behaving inconsistently was the worse outcome.

**Rebase note for whoever merges second:** the conflicts are new table rows and new
test cases, not competing rewrites.

## Deferred, with the reason

- **#2067 — measure the unlogged-search rate across the corpus.** *Needs a different
  skill* (harness/Python, eval reviewer). Also sequencing: no committed run log carries
  these fields yet, so it would ship with nothing to read. It is what answers
  note-vs-deny, and this PR deliberately does not settle that.
- **#2068 — `image_search` has the same obligation and no `projectPath`.** Neither note
  can be conditioned without giving that tool project context: a new parameter, its
  schema entry, the manifest list, the spec and six fixtures. Different work, not a
  mirror. It also carries a design question worth answering first — whether
  `image_search` should stage like the other three, against its spec's deliberate "no
  sidecar" line.

Issues #2001 and #1978 stay open and are untouched by this PR.

---

## Round 2: senior-developer review applied (`clack391`)

- **The merge condition.** Three merges from main had silently reverted `bc61ae055`
  (#1826) in `docs/specs/fulltext-search-tool-spec.md` — the `FulltextSearchResponse`
  block carried `FulltextSearchResult[]` (an identifier with **0 hits** in `src/`), a
  nested `facets?` where the real type is flat, and no `staged?` / `stagingError?` at
  all. Rebuilt the block from `origin/main` and re-added only this PR's two fields; it
  now differs from main by exactly those two entries, and the whole file by nothing
  else.
- **`projectPath: null`** no longer reaches `join()`. All three callers gate on
  `!== undefined`, so an explicit null was a raw `TypeError` eight lines under a
  docstring promising it could not throw. Guarded once in the reader so a fourth
  caller inherits it, with a test on the shape one over from the covered case.
- **The mock's nil gate** now mirrors production on both halves via `_fixture_is_nil`,
  reading whichever total the fixture carries (`totalMatches` / `totalResults` /
  `totalForPlace`). It was bare `if not response.get("results")`, so the harness would
  have fired where production does not — which is what makes a zero firing count
  unreadable.
- **Two drift guards added**, both watched failing: `UNLOGGED_REFS_SHOWN` across the
  TS/Python copies, and `STAGING_SEARCH_TOOLS` against the `STAGING_CAPABLE_TOOLS`
  this PR consolidated.
- **The TTL premise was false on the nil path**, now stated correctly with a test
  pinning it: `pruneStale` runs inside `stageSearchResults` *after* its nil
  early-return, so a nil search sweeps nothing.
- Also taken: `ownership.json`'s notes and the staging-spec line now say three tools
  and handles rather than two and a count; the pre-review nil condition is gone from
  the spec row and both type comments; the no-op quote escape matches its seven
  siblings; the orphaned `STAGING_CAPABLE_TOOLS` comment is gone and four count-era
  names renamed; `results_available` is coerced the way `ops` already is.

**Answering @DallanQ's question** (whether `sourcesWithoutAssertionsWarning` firing
zero times means the condition never holds): it never holds. Across **249** committed
project states, **0** reach 3+ sources with zero assertions — the threshold the
warning requires — and only **1** has any sources at all with zero assertions. So
that zero is the good-news branch, not a nudge being ignored.

His measurement ask goes on issue #2067, widened to all four advisory fields, per
clack391's suggestion and Dallan's own stated preference.


packages/engine/mcp-server/src/utils/results-staging.ts
```diff
@@ -198,6 +211,126 @@ export async function finalizeStagedResults(args: {
...
+ * Emitted on a `projectPath`-carrying search that returned nothing. A nil result
+ * stages no file, so the note above can never see it — this is the only signal for
+ * the case that matters most, since a nil search is evidence that those query terms
+ * fail and is what a reasonably exhaustive search is required to record.
+ *
+ * Mirrored byte-for-byte in `eval/harness/harness/mock_mcp.py` — grep
+ * NIL_SEARCH_NEEDS_LOG_NOTE to find both copies when editing either.
+ */
+export const NIL_SEARCH_NEEDS_LOG_NOTE =
+  "Nothing returned, and nothing staged. A nil result is evidence: log it with " +
```
florencemashipei: **Blocking — this sentence contradicts the skill's own doctrine, in front of the one reader who has no access to the qualification.**

"A nil result is evidence" is the opposite of what `search-records/SKILL.md` says in three places:

- `:223` — "A nil is **not** evidence that some relative was absent"
- `:72` — "do NOT log negative as absence — the record is un-indexed, not absent"
- `:593` — "`outcome` is still **`negative`** — that field records what the search returned, **not what it means**… **Never let the narrative claim absence**"

In the skill, "log the nil, it's a finding" is safe because those paragraphs surround it. This note exists precisely to reach a session that invoked no skill — so it delivers the unqualified half to the reader with no qualification available. And the motivating case is itself the wrong kind of nil: the 1820–1840 census nils in #2056 are an *index-coverage* gap (FamilySearch's name index doesn't cover those years), which is `:593`'s exact scenario. The project also carries `evidence_type: negative` as a distinct enum, so "a nil result is evidence" is one short step from a negative `person_evidence` assertion written off an indexing artifact.

One clause fixes it, and it keeps every field the current text asks for:

```ts
export const NIL_SEARCH_NEEDS_LOG_NOTE =
  "Nothing returned, and nothing staged. A nil search is a finding and must be " +
  "recorded: log it with `research_log_append`, `outcome: \"negative\"` — which " +
  "records what the search returned, not that the record is absent — the exact " +
  "parameters used, and no `stagedResultsRef`.";
```

The `mock_mcp.py` mirror takes the same text. `test_unlogged_search_notes_have_not_drifted_from_the_typescript_source` parses the declaration out of this file and compares, so it needs no edit — two files, no test change.

packages/engine/mcp-server/src/tools/record-search.ts
```diff
@@ -1031,6 +1045,15 @@ export async function recordSearchTool(
     ...(input.projectPath !== undefined && !input.subjectId
       ? { rankingSkipped: RANKING_SKIPPED_NOTE }
       : {}),
+    // Both model-facing notes sit in the same pre-`results` slot as
+    // `rankingSkipped`, for the same reason: a field after `results` is the first
+    // thing a size bound drops (spec § Key order).
+    ...(unloggedStaged > 0
+      ? { unloggedSearches: UNLOGGED_SEARCHES_NOTE.replace("{n}", String(unloggedStaged)) }
+      : {}),
+    ...(input.projectPath !== undefined && results.length === 0
```
florencemashipei: **Blocking — this can fire on a search that wasn't nil, and the note then orders a negative log entry for a query that has matches.**

`results` here is the post-`mapEntry` set: `entries` mapped and null-filtered, and `mapEntry` returns null on a missing represented person (`:753`) or a missing `entry.id` (`:755`). So a page whose entries all fail mapping gives `results.length === 0` with `totalMatches` > 0. `stageSearchResults` also returns null on `results.length === 0`, so `staged` is null too and the note's "nothing staged" reads as corroboration.

Second path, which I can't settle from the code and you can: `offset` is validated only against the 4999 cap (`:206-220`), never against `totalMatches`. If FS answers an offset past the last page with an empty `entries` rather than an error, a paging walk that overruns emits this note with matches available.

The genealogical cost is asymmetric with the one this PR is fixing. A missing log entry loses evidence; a false `negative` entry *adds* a claim that a query shape produced nothing, and the log is append-only — `research_log_append` refuses to edit, so a wrong entry can only be corrected by a later append that a reader may never reach. Negative findings are also what later reasoning leans on hardest ("we already searched Virginia statewide, nothing there"), which is the reasoning failure #2056 documents.

`external_links_search` got exactly this care — keyed on the pre-filter set, with test case 13 to hold it. Suggest the same shape here: add `data.results === 0` (or `out.totalMatches === 0`) to the condition, plus a case with `{ results: 812, entries: [] }` asserting the note is absent. If the paging overrun genuinely can't happen, that's worth a sentence in the spec, because the mapping path stands on its own.

packages/engine/mcp-server/src/tools/fulltext-search.ts
```diff
@@ -207,12 +212,29 @@ export async function fulltextSearchTool(
...
     offset: data.index ?? input.offset ?? 0,
     hasMore: data.links?.next?.href != null,
+    // Both notes precede `results` — a field after the largest one is the first
+    // thing a size bound drops. Contract: record-search-tool-spec-v2.md.
+    ...(unloggedStaged > 0
+      ? { unloggedSearches: UNLOGGED_SEARCHES_NOTE.replace("{n}", String(unloggedStaged)) }
+      : {}),
+    // Safe on `results` here, unlike external_links_search: nothing narrows the
+    // inline copy before staging, so an empty `results` means nothing was staged.
+    ...(input.projectPath !== undefined && results.length === 0
```
florencemashipei: **Blocking, same defect as `record_search` — and the comment above it doesn't cover this case.**

"nothing narrows the inline copy before staging" is true of *host filtering*, which is what separates this tool from `external_links_search`. But `results` is still the post-`mapEntry` set (`:211-213`), and `mapEntry` returns null on a missing `entry.id` (`:94`). So an empty `results` here does not mean the search found nothing — it means nothing survived mapping. `totalResults` (`data.results`) is independent and can be non-zero.

Same fix and same reason as the `record_search` thread: add `data.results === 0` to the condition, and widen the comment to say what it actually rules out (host filtering) rather than "nothing narrows the inline copy", which reads as covering the mapping path too.

packages/engine/mcp-server/src/utils/results-staging.ts
```diff
@@ -198,6 +211,126 @@ export async function finalizeStagedResults(args: {
...
+/**
+ * Emitted when this project holds staged search responses with no `research.json`
+ * log entry behind them (issue #2056). `{n}` is substituted host-side.
+ *
+ * Mirrored byte-for-byte in `eval/harness/harness/mock_mcp.py` — grep
+ * UNLOGGED_SEARCHES_NOTE to find both copies when editing either.
+ */
+export const UNLOGGED_SEARCHES_NOTE =
+  "{n} earlier staged search response(s) in this project have no research.json log " +
+  "entry. Call `research_log_append` for each, passing the `staged.resultsRef` that " +
```
florencemashipei: **Not blocking — resolve this with a reply if you'd rather not take it. But the cheapest way for an agent to silence this note is a lossy log entry, and I think that's worth knowing before the rate gets measured in #2067.**

The note says "passing the `staged.resultsRef` that search returned". The backlog exists *because* the session lost track of it — #2056's own second consequence is the transcript being replaced by a summary that dropped the scope parameters. So at the moment this note fires, the ref is usually not recoverable from context. What the agent can do then:

1. **Log without the ref.** It must then supply `query` by hand — `research-log-append.ts:356-362` throws "`query` is required" when neither the caller nor a staged payload provides one, and the comment at `:315` puts hand-transcription of that object at a 20% failure rate. That entry is `results_available > 0` with no `results_ref`, so it takes the warn at `:336-350`… **and it pairs.** `countUnloggedStagedSearches` pairs a staged file against exactly that shape of entry (same tool, `performed` at or after `retrieved`), so the count drops to zero, the note stops firing, and the raw response is gone for good. The nag is satisfied by the one entry class this PR's own pairing rule was built to tolerate.
2. **Reconstruct the parameters from memory** — which is #1779's defect class, and the specific failure #2056 records: an agent reconstructing from a summary, conceding a county-scoping error it never made.

`countUnloggedStagedSearches` already opens every unpaired envelope, and the envelope's `payload` is the verbatim response, so `payload.query` and the filename are both in hand. Returning the unpaired handles (ref + tool + `retrieved`) instead of only a count makes the obligation satisfiable from disk — and with the ref, `:318` fills `query` from the staged payload verbatim, so the catch-up entry is fidelity-safe by construction rather than by the model's memory. That seems like the difference between a signal and a nag.

Separately, and genuinely minor: as written this reads as normal cadence, but `SKILL.md:577` forbids batched logging ("Log each retry as a separate `research_log_append` call immediately after it completes — do not batch log calls at the end"). A closing clause — "log each search as it completes, not in a batch at the end" — keeps the remediation from teaching the habit that produced the backlog.

florencemashipei (commented):
Senior-genealogist review. Filing this as a comment rather than a changes-request: `senior-queue.yml` describes senior review as the second review, and this PR hasn't had a first round, so I don't want to pre-empt that reviewer. It holds the merge regardless — `protect-main` has `required_review_thread_resolution: true`, verified on the ruleset.

The mechanism is right and the plane argument holds: the unit plane grades one skill's own run, and the session in #2056 invoked no skill, so the tool is the only place this signal can exist. What I'm pushing back on is entirely in the two model-facing strings and the condition that fires one of them — and that's not a nit, because those strings are read by an agent that has *by construction* not read `search-records/SKILL.md`. Everything the skill uses to qualify "log the nil" is absent at the moment the note lands.

Two threads I'd treat as blocking (the nil-note wording, and the condition that emits it), one shared with `fulltext_search`, one I'm explicitly not blocking on.

**What I checked and agree with:**

- The nil note's field list matches `search-records/SKILL.md:576` exactly — `outcome: "negative"`, the exact parameters, omit `stagedResultsRef`. No drift there.
- `planItemId: null` is explicitly supported "for an opportunistic search with no plan item", so the instruction is executable in a project whose plan never covered the search. That was my first worry and it's already handled.
- Withholding both notes from the sidecar is correct — the sidecar is the record of what the search returned, not of what the model was told.
- Advisory rather than deny is the right call at this stage. A deny would refuse a legitimate broad survey, and the rate isn't measured yet (#2067).

**On the plane question you asked reviewers to argue:** from the genealogy side, keep `["unit", "tool"]`. The row's own `failure` text — "the log is what makes a search reconstructible, including nil results" — is the doctrine at stake, and the new `notes` field says plainly that the claim is detection-only, so it doesn't mislead a reader. Whether the manifest's plane definition should stretch to a non-constraining note is a manifest judgment, not mine.

**On the third planning round:** `docs/task-lifecycle.md:107-110` settles that one ("Two rounds maximum… stop and go to the lead"), and you put the override on the record yourself. That's for the senior developer and Dallan to weigh — I'm not adjudicating it, only noting I read it.

florencemashipei:
@johnmarkpeterbrown — senior-genealogist pass is up above. Comment rather than changes-requested, since `senior-queue.yml` puts senior review second and round one hasn't happened; the threads hold the merge either way via `required_review_thread_resolution`.

Two I'd call blocking, both in the model-facing strings rather than the mechanism:

- **The nil note's "A nil result is evidence" clause** contradicts `search-records/SKILL.md:223`, `:72` and `:593`, and the 1820–1840 census nils behind #2056 are exactly `:593`'s index-coverage case. Replacement wording is in the thread, ready to paste; the drift test parses the TS declaration, so it's two files and no test edit.
- **`nilSearchNeedsLog` keys on the post-`mapEntry` set** in both `record_search` and `fulltext_search`, so an empty `results` with `totalMatches` > 0 orders an `outcome: "negative"` entry for a query that has hits. `external_links_search` already guards this via the pre-filter set.

**One question only you can answer:** does FS return an empty `entries` for an offset past the last page, or an error? The mapping path carries that finding on its own, so this only decides whether the condition needs `data.results === 0` or the spec needs a sentence saying the overrun can't happen.

Thread 4 (returning the unpaired handles instead of a bare count) is **not** blocking — resolve it with a reply if you'd rather not take it. Worth reading the first half though: an agent that can't recover the ref clears the note fastest by logging without one, and that entry *pairs*, so the count zeroes while the raw response is lost.

On the plane claim you flagged for reviewers: keep `["unit", "tool"]` — the `notes` field says detection-only, which is honest. Happy to approve once the two blocking threads are settled. Flagging that `protect-main` wants 2 approvals, so the `.ts`/`.py` still needs a senior developer regardless of me.

packages/engine/mcp-server/src/utils/results-staging.ts
johnmarkpeterbrown: Taken as written, in `ba73d18d` — your wording verbatim, in the TS and the `mock_mcp.py` mirror.

You're right and I don't think this was a wording preference. I wrote "a nil result is evidence" while `:593` says the opposite in the same breath as the field I'm telling the model to set: `negative` records what the search returned, not what it means. The part I hadn't joined up is that the 1820–1840 nils behind #2056 are that exact scenario — an index-coverage gap — so the note would have been pushing the wrong reading on the very case it was written for.

I also rewrote the docstring above the constant, which was selling the same idea in different words ("a nil search is evidence that those query terms fail"), and left your three citations in it so the next person editing this string sees why it's shaped this way.

Drift test needed no change, as you said — it parses the declaration out of the TS and compares. It passes.

packages/engine/mcp-server/src/tools/record-search.ts
johnmarkpeterbrown: Fixed in `ba73d18d`. The condition now also requires `(data.results ?? 0) === 0`, so it keys on the upstream total rather than the mapped set.

Confirmed the path you described before changing anything: `mapEntry` returns null at `:748` (no represented person) and `:750` (no `entry.id`), and `:987` filters those out, so an all-unmappable page gives `results.length === 0` with `totalMatches` non-zero. Added the test you suggested — `{ results: 812, entries: [] }`, asserting the note is absent — and I reverted the gate to watch it fail first. It does, and only that case.

**On the offset question:** I can't answer it either. `offset` is validated against the 4999 cap at `:212` and never against `totalMatches`, and there's no probe script or spec line in the repo recording what FS returns for an overrun. So rather than guess, I've written it into the spec as an explicitly unverified path, and noted that gating on `data.results` covers both readings — an empty `entries` with matches available doesn't fire the note whatever emptied it. If we ever want the real answer it needs a live probe, which I'd rather file than fake.

Worth saying plainly: this is the same defect class the plan critic caught in `external_links_search` before you saw the PR. I fixed it there, then reviewed the other two tools myself and declared them clean — I checked the filter path and never checked the mapping path.

packages/engine/mcp-server/src/tools/fulltext-search.ts
johnmarkpeterbrown: Fixed in `ba73d18d`, same gate: `(data.results ?? 0) === 0`. Test case 41 covers it, and it fails against the old condition.

Your point about the comment was the more useful half. "Nothing narrows the inline copy before staging" was true of host filtering and I let it stand in for the whole question — `mapEntry` at `:89` drops an entry with no `entry.id` here too, so the mapping path was never covered by that sentence. The comment now says what it actually rules out, and the response-interface entry in `fulltext-search-tool-spec.md` says the same, so the next reader doesn't have to re-derive it.

packages/engine/mcp-server/src/utils/results-staging.ts
johnmarkpeterbrown: Taken — this wasn't one I wanted to resolve with a reply.

The half I'd missed is that the lossy entry **pairs**. I knew the ref was often unrecoverable, but I'd assumed the failure mode was the agent ignoring the note. Logging without the ref is cheaper than ignoring it, and it produces exactly the entry shape my own pairing rule was written to tolerate — so the count zeroes, the note goes quiet, and the response is gone. The mechanism would have looked like it was working.

`ba73d18d`: `countUnloggedStagedSearches` is now `unloggedStagedSearches` and returns the unpaired handles — `ref`, `tool`, `retrieved`. The note names the refs (first five, then "and N more"), says the staged file carries the search's own query so the entry is filled host-side, and closes with logging each as you go rather than batching, per `SKILL.md:577`. A test asserts the returned ref is one `finalizeStagedResults` actually accepts, so the instruction is executable rather than just well-worded.

Two notes on the mirror, since they affect how this behaves in evals:

- The refs formatter is restated in `mock_mcp.py` — a five-item join, not the pairing rule, which still comes out of the compiled build.
- Renaming the export broke the mock's staging until I rebuilt, because `mock_mcp.py` imports from `build/` and its `except` swallows the ImportError as "nothing staged". Harmless once you know, but it fails silently, so it's worth knowing before the #2067 measurement runs against a stale build.

johnmarkpeterbrown:
@florencemashipei — all four addressed in `ba73d18d`, including the one you said I could resolve with a reply. Replies are on each thread; leaving them open for you to close, since three were blocking and you should see the code rather than take my word.

- **Nil-note wording** — your text verbatim, TS and mock mirror. I also rewrote the docstring above it, which was making the same claim in different words.
- **`nilSearchNeedsLog` on both tools** — now gated on `(data.results ?? 0) === 0`. A test each with `{ results: 812, entries: [] }`; I reverted each gate to watch its test fail before restoring.
- **Handles instead of a count** — `unloggedStagedSearches` returns `ref`/`tool`/`retrieved`, the note names the refs, and it now says log each as you go rather than batching.

**Your offset question:** I can't answer it. `offset` is checked against the 4999 cap only, never against `totalMatches`, and nothing in the repo records what FS returns past the last page. It's now in the spec as an explicitly unverified path, with the note that gating on `data.results` makes it moot for correctness. A real answer needs a live probe — happy to file one if you think it's worth having on its own.

The thing I'd flag from doing this: your threads 2 and 3 are the same defect the plan critic caught in `external_links_search` before you saw this. I fixed it there, then reviewed the other two tools myself and called them clean — I'd checked the filter path and never the mapping path. So that's twice this class got through me, and the reason it's caught now is that you went looking at the field the condition reads rather than the sentence I wrote about it.

Suites on the new head: 2806 engine (118 files), 3044 harness unit, all 13 checks green.

Still needs two approvals with at least one senior developer for the `.ts`/`.py`, so yours won't unblock merge on its own. The third-planning-round override is still open for whoever takes that half.

packages/engine/mcp-server/src/utils/results-staging.ts
florencemashipei: Verified in `ba73d18d` — the string matches in both copies, and the rewritten docstring keeps the citations, which is the part that will stop this regressing. Closing.

packages/engine/mcp-server/src/tools/record-search.ts
florencemashipei: Verified in `ba73d18d`. Gate reads `results.length === 0 && (data.results ?? 0) === 0`, and the new case pins it with `totalMatches` 812 and an unmappable entry. Recording the FS-paging question as unverified in the spec rather than guessing at it is the right disposal. Closing.

packages/engine/mcp-server/src/tools/fulltext-search.ts
florencemashipei: Verified in `ba73d18d` — same gate, same shape of test, and the comment now says what it actually rules out. Closing.

packages/engine/mcp-server/src/utils/results-staging.ts
florencemashipei: Verified in `ba73d18d`. Handles round-trip: the test asserting `unlogged.ref === handle.resultsRef` pins the property the note depends on, so an agent can paste the ref straight back. Closing.

florencemashipei (approved):
Approving on the genealogy side. All four threads verified against `ba73d18d` and resolved.

What I checked rather than took on trust:

- The nil note is the corrected wording in `results-staging.ts` and byte-identical in the `mock_mcp.py` mirror, and the rewritten docstring keeps the `SKILL.md:223/:72/:593` citations — that is what will stop the phrasing regressing.
- Both nil gates read `results.length === 0 && (data.results ?? 0) === 0`, each with a case pinning the mapping path. `external_links_search` kept its pre-filter gate through the rename.
- The returned handles round-trip: `ref` is `results/.staging/<uuid>.json`, which is what `stageSearchResults` emits and what `finalizeStagedResults` requires, so the ref in the note is directly usable as `stagedResultsRef` — and there is now a test asserting exactly that.
- CI green on this head, `vitest` and `pytest` included, so the drift test is covered.

**One non-blocking doc fix, deliberately not filed as a thread so it holds nothing up:** `docs/specs/record-search-tool-spec-v2.md:422` still reads "Present **only** when `projectPath` was supplied and the search returned nothing", which is the pre-`ba73d18d` condition. The prose you added at `:463` describes the upstream-total gate correctly, so the spec now disagrees with itself in one file — and that table row is both what a reader hits first and what `external-links-search-tool-spec.md:87` points at for the contract. Worth folding into whatever touches this next.

On the plane question you put to reviewers: keep `["unit", "tool"]`. The `notes` field says detection-only, which is an honest description of what the row now covers.

Note this is one of the two approvals `protect-main` requires, and the `.ts`/`.py` mechanism still wants a senior developer's eyes — my pass covered the model-facing wording and the genealogical semantics of what gets logged, not the pairing implementation.

DallanQ:
Not an objection to this PR — the reasoning that the other planes are closed is right, and the fields are cheap. One measurement it should carry, because this PR makes it four.

**The two advisory return-fields already shipped are invisible in the committed corpus.**

| field | shipped in | runs mentioning it, of 159 |
|---|---|---:|
| `browseBudget` (`image-transcribe.ts`) | accepted in ADR-0011's Alternatives table | **1** |
| `sourcesWithoutAssertionsWarning` (`research-append.ts`, rides `validation.warnings`) | issue #1478, closed completed 2026-08-17 | **0** |

**The instrument can see it, so this is a partial negative rather than a capture artifact.** Only 483 of 3,496 `research_append` calls still carry a `response_summary` — the rest are stripped past the capture window — but **310 of those 483 carry the `validation` block with its `warnings` array**, e.g. `{"ok":true,"section":"questions","op":"append","entryId":"q_001","filesWritten":["research.json"],"validation":{"valid":true,"warnings":[]}}`. Across every one of those, the persistence nudge never appears. Method: grep the literal across `eval/runlogs/e2e/*/run-*.json` matched on `run-\d{4}-\d{2}-\d{2}_\d{2}-\d{2}-\d{2}\.json$`, python3, `encoding="utf-8"`.

**Why it bears on this PR.** ADR-0011's layer map files a return-payload advisory under **Tool description**, whose own row says "strength unmeasured", and its Alternatives table accepted `browseBudget` with the words "ships as a field on a successful result, and knowingly does nothing if the agent ignores it." That was a deliberate, eyes-open bet. What has never happened is checking whether the bet paid. This PR doubles the number of such fields without that check existing, so if the shape does not work we will have four instances of it and still no way to know.

**The cheap ask, not a blocker:** make `unloggedSearches` / `nilSearchNeedsLog` *countable* in the same PR — a line in whichever corpus reader you prefer that reports how often each fired and whether the next call changed. That turns "we shipped an advisory" into "we know whether advisories work", which is worth more than either field individually and costs a few lines over committed logs. If you would rather keep this PR at its current scope, the same counter over all four fields is a small standalone follow-up and I am happy for it to go that way.

One thing to check while you are in `research-log-append.ts`: `sourcesWithoutAssertionsWarning` firing zero times may mean the condition genuinely never holds rather than that the agent ignores it. Those are different findings, and only the first is good news.

clack391 (changes requested):
Senior-developer pass on the `.ts`/`.py` mechanism, which is the half florencemashipei's approval explicitly did not cover. Requesting changes on **one** finding; everything else here is optional and I have marked it so.

@DallanQ, your comment landed while I was finishing. It does not change the verdict, but it did change three things in my write-up: it resolves the note-vs-deny routing item I was going to raise, it withdraws one of my findings, and it promotes another. Disposition of your ask is at the bottom.

## The one merge condition

**The three merges from main silently reverted `bc61ae055` (#1826).** `docs/specs/fulltext-search-tool-spec.md`'s `FulltextSearchResponse` block now:

* declares `results: FulltextSearchResult[]`, an identifier that exists in neither the spec (which declares `interface FulltextResult` at `:151`) nor anywhere in `packages/engine/mcp-server/src/` (0 hits)
* documents `facets?` as a nested `{collections?, places?, years?, recordTypes?}` object, where `src/types/fulltext-search.ts:105` declares a flat `FulltextFacet[]` and `fulltext-search.ts:252` assigns `mapFacets(...)`
* **drops `staged?` and `stagingError?` entirely**, while `:221` of the same file still describes the tool setting `staged: null` and `stagingError`

`origin/main` has all four correct; `bc61ae055` ("fulltext-search: reconcile the spec with the implementation", #1826) is the commit that made them so and is an ancestor of main. The branch's original base predates it, and the three merge resolutions kept the stale side.

Why I am blocking on a doc, given it is recoverable and has no runtime effect:

1. **It is certain.** Every other finding here has a trigger condition. This one takes effect on merge.
2. **It reverts landed work,** so the cost falls on someone else. #2001 is open right now doing exactly this reconciliation on this block, in the opposite direction. If this merges first, its author resolves a conflict against a corrupted block and #1826 has to be rediscovered by someone who does not know it was ever fixed. It also contradicts this PR's own rebase note, which tells that merger the conflicts are "new table rows and new test cases, not competing rewrites".
3. **Nothing can see it.** The full suite is green with the wrong block and green with the corrected one, and no test in `tests/packaging/` parses spec interface blocks. It is also the only unintended removal in the diff: I diffed all 21 changed files against `origin/main` and every other deletion is deliberate.
4. `staged` is the field this PR's own note instructs the agent to pass back as `stagedResultsRef`.

Restoring those six lines while keeping the two new fields is the whole condition.

## Fix before merge, but not on impact

**`projectPath: null` turns an already-successful search into a raw `TypeError`.** All three tools gate on `input.projectPath !== undefined` (`record-search.ts:1005`, `fulltext-search.ts:219`, `external-links-search.ts:193`), so an explicit null reaches `unloggedStagedSearches`, and `join(null, ...)` at `results-staging.ts:308` throws. That is eight lines below the docstring at `:292`: "Never throws, and returns an empty array on any failure."

Verified three ways: the compiled function throws on null; `recordSearchTool({surname:"Lincoln", projectPath:null})` throws with a mocked 200 and a valid response body; and with **only** the new call replaced by `[]`, the identical input returns normally. The sibling `stageSearchResults` a few lines later throws the same class of error but sits in a `try/catch` that degrades to `stagingError`, so this is a regression in robustness, not a pre-existing hole.

I first ranked this a blocker and then demoted it myself: 18,925 committed objects under `eval/runlogs`, `eval/fixtures` and `eval/tests` carry a `projectPath` key and **0** are null or empty, so I have no basis for calling it user-facing in practice. It stands on the contract and the regression only.

Worth naming why the 7 new `record_search` tests miss it: one of them is "says nothing about logging when no projectPath was supplied", which exercises `projectPath` **absent**. The crash is one shape over, present and null.

Suggested fix, which lands once instead of three times and covers any fourth caller:

```ts
): Promise<UnloggedStagedSearch[]> {
  if (typeof projectPath !== "string" || projectPath === "") return [];
  const stagingDir = join(projectPath, STAGING_SUBDIR);
```

## Notes, all optional

I applied all of these together with the two above, and both suites stayed green (engine 118 files / 2807, harness 3045 / 2 skipped), so none of them is a guess.

The two I would actually press:

1. **The mock's nil gate never got the upstream-total check.** `mock_mcp.py:653` is `if not response.get("results")`, with no `totalMatches`/`totalResults` test, while production now requires `(data.results ?? 0) === 0`. On a fixture shaped `{"totalMatches": 812, "results": []}` the harness emits `nilSearchNeedsLog` where production does not. This is the same defect flagged in threads 2 and 3, un-mirrored in the third copy, and the comment directly above it at `:634` enumerates the divergences from production and names only the `external_links_search` pre-filter one. Currently latent: 0 of 2,619 response-shaped objects under `eval/` have that shape. @DallanQ, this is why your closing paragraph promoted it for me. A mock that fires where production would not makes "the condition never holds" and "the agent ignored it" indistinguishable, in the plane that is supposed to tell them apart.
2. **`ownership.json`'s `log` row names two of the three tools.** `fulltext_search` occurs zero times in the new `notes`, though the PR body says "all three staging tools"; the text predates the fold-in. Same staleness in `search-result-staging-spec.md:191`, which says "the two search tools surface the **count**", wrong on both halves since the same bullet says eight lines later that it returns handles, not a count. These matter a little more than ordinary prose because they are the enforcement-plane manifest this PR exists to change. Nothing catches it: replacing both names with `totally_fake_tool`/`another_fake` leaves all 616 packaging tests green, since `ownership-manifest.test.ts` validates only `enforceableAt` against the plane list and merely declares `notes?: string`.

The rest, take or leave:

3. The pre-`ba73d18d` nil condition survives in three places: `record-search-tool-spec-v2.md:422` (which `:463` of the same file now contradicts, and which `external-links-search-tool-spec.md:87` names as the contract document), plus the `nilSearchNeedsLog` comments in `src/types/record-search.ts` and `src/types/fulltext-search.ts`. @florencemashipei raised the spec row in her approval and chose not to file it; I am folding it in rather than reversing her, since this PR is the "whatever touches this next" she pointed at, and it introduced the contradiction. The two type-comment copies are new.
4. The refs formatter is restated in Python with nothing guarding drift, which is the exact hazard the docstring gives as its reason for calling the pairing rule out of the compiled build. Mutating `UNLOGGED_REFS_SHOWN` 5 to 4 on the Python side alone leaves all 3045 harness tests green. The two agree today (I ran them differentially at n=0,1,4,5,6,7,12: identical at every point including the boundary), so this is drift exposure, not a live bug. The template is 30 lines above the new drift test in the same file: `test_ok_false_gate_set_has_not_drifted_from_the_typescript_source` parses a TS declaration and compares sets. One test would also cover the pre-existing `STAGING_SEARCH_TOOLS` (`mock_mcp.py:294`) against the `STAGING_CAPABLE_TOOLS` this PR consolidated for exactly the reason "a second copy would drift".
5. `mock_mcp.py:433` has `.replace("'", "\'")`, which is a no-op in Python, where the six other copies of that idiom in the same file (`:263`, `:378`, `:780`, `:783`, `:844`, `:909`) correctly use `"\\'"`. A repo path containing a quote would produce a `SyntaxError` in the generated import, swallowed by `except Exception: return []`. Not attacker reachable, since the interpolated value is a repo-internal path and `projectPath` goes in through `json.dumps`. One character.
6. The TTL skip's stated premise holds on the non-nil path and fails on the nil one. `pruneStale` is called from exactly one place (`results-staging.ts:94`), which sits **after** `if (results.length === 0) return null` at `:74`. Executed on a scratch project with a 30h-old unlogged staged file: after a nil search the file is still on disk and the reader reports `[]`; after a non-nil search it is gone. So on the path the second note exists for, a stale file is neither reported nor "about to be deleted". The test at `results-staging.test.ts:215` asserts only the skip, so the premise ships in its own title.
7. `research-log-append.ts:39` is a comment orphaned by the const's move and now documents nothing. Four names still say "count" after the rename to handles: `results-staging.test.ts:106` and `:223` ("returns 0 ..."), and `_unlogged_staged_handles`'s docstring, which says "Returns 0 on any failure" and returns a list.
8. The PR body's Verification block is stale, and it is what a merger reads: "115 files, 2778 passed, 1 skipped" against 118/2807/0, "2994 passed, 3 skipped" against 3045/2, "`tests/packaging/` (608)" against 616, and "17/17 pass" against 18/18 after `ba73d18d` added a test. Your later comment has current numbers; the body was not refreshed. Separately, the body cites `research-log-append.ts:337` for the no-`projectPath` sentence, which sits at `:334-335`; `:337` is the `if` guard.
9. Lowest value, genuinely optional: `entry.results_available = op.resultsAvailable` at `research-log-append.ts:285` has no coercion, `validator.ts` carries `results_available` only in field-name allow-lists with no type check, and the schema types it integer-or-null. A model sending `"5"` lands a string, the new filter's `typeof === "number"` skips the entry, and the staged file nags until the TTL. Confirmed by running the compiled `validateParsed` on a fixture with `results_available: "not-a-number"`: `valid: true`, 0 errors. It over-reports rather than hides, which is the benign direction, and the file already imports `coerceJsonArg`.

## Things I chased and dropped, so nobody spends the time again

* The sequential per-file loop in `unloggedStagedSearches`, where `pruneStale` in the same file uses `Promise.all`. Measured: 20 files 24ms, 50 files 51ms, 200 files 165ms, 2000 files 1825ms. Against the largest `research.json` in the whole committed corpus (365 KB, and the longest `log` anywhere is 55 entries) plus the 6 staged files of the motivating session, the whole call is **10.3ms**. Free at every real size.
* A symlink under `.staging/` double-counting. Needs a writer that is not the tool.
* `external-links-search-tool-spec.md`'s "(15 cases)" against 25 actual tests. The heading matches its own 15-row table and this PR correctly added 3 rows; table-versus-reality drift is pre-existing and already acknowledged in the sibling spec.
* The greedy oldest-first pairing. Attacked with clock skew, ties, `performed` equal to `retrieved`, cross-tool entries, missing and wrong-cased `tool`, non-ISO `retrieved`, and non-numeric `results_available`. I could not make it strand a pairing, and the failure paths degrade exactly as documented.

## @DallanQ, your ask

Verified your figures before relaying them: 159 e2e run logs match `run-\d{4}-\d{2}-\d{2}_\d{2}-\d{2}-\d{2}\.json$`, `browseBudget` appears in **1**, `sourcesWithoutAssertionsWarning` in **0**, and both new fields in 0, which incidentally confirms this PR's sequencing argument for deferring the detector. On the instrument-visibility half, **483** of the `research_append` calls retain a `response_summary` and **310** of those carry the `validation` block: both exact. My denominator for total `research_append` calls came out at 3,681 rather than 3,496, probably because I also matched namespaced spellings; it does not bear on your argument, which rests on the 310.

**Half your ask is already filed.** #2067's second bullet is your request nearly verbatim: "From the new response fields, `unloggedSearches` / `nilSearchNeedsLog` firings, and whether the agent acted on them." What #2067 does **not** cover is the half your data is actually about, the counter over the two already-shipped fields, which is what tests whether the advisory *shape* works rather than what this feature's rate is.

So rather than a fourth issue or folding it in against your stated preference, I would **widen #2067 to all four fields** and put your measurement on it. Your numbers also re-prioritise that issue, whose "Why it matters" already argues that `RANKING_SKIPPED_NOTE` was justified by a measured rate before shipping and this one was not. That also absorbs a finding I withdrew: the 362/1242/126 figure now sits in a permanent manifest with no re-derivation and no as-of date, and the same reader that counts firings re-derives it, so asking for a date clause separately would be a worse version of work you have already scoped.

Your second point, whether `sourcesWithoutAssertionsWarning` firing zero times means the condition never holds rather than being ignored, is a real open question and cheap to answer. @johnmarkpeterbrown will be in `research-log-append.ts` anyway for notes 7 and 9.

## Also outstanding, and not the author's to close

@florencemashipei, your paging-overrun question got an honest "I cannot settle this", the path is now in the spec as explicitly unverified, and the author offered to file a probe. That offer is still unanswered and no diff will remind anyone of it. A yes or a no from you closes it.

## What I checked that is right, because it is unusual

The corpus measurement is **exact** under three independent derivations, mine and two blind agents': 362 committed `research.json` files, 1242 non-nil staging-capable entries, 126 without a `results_ref`, 10.14%. All three of the named break-tests genuinely fail when reverted, which I re-ran rather than took on trust. The pairing rule is sound and I could not break it. Both notes are byte-identical across the TS and Python copies and the drift test really does fail on divergence. The withhold-from-sidecar guarantee is delivered at all three sites, including `external_links_search`, which needs no destructure because it stages a purpose-built `{ results: allLinks }`. The `encoding="utf-8"` rule holds on every new Python call, by the repo lint and by my own AST pass. And none of my 11 findings are in the mechanism this PR is about: they are in its prose, its mirror, and a merge resolution.

## One note on my own suite run

My first pass showed 8 failed test files plus 1 failed test on the engine side and 1 failure on the harness side. All environmental, none yours: the review worktree's `mcp-server/node_modules` held 0 packages so `open` did not resolve, and the gitignored engine `build/` was absent, which also silently skipped 6 harness tests. After symlinking the install and running `tsc`, both suites are green and your reported numbers reproduce. Flagging it because "the reviewer's suite was red" is worth exactly nothing here.

## Sequencing

`mergeStateStatus` is BEHIND; main is 3 commits ahead of the merge base and none of the 3 touches any of your 21 files, so the update-branch is a clean fast-merge. Given that merging main is what produced the blocker, I would do the merge first and fix the spec block after, then eyeball that one hunk in the resolution.

`dismiss-stale` is false on the live ruleset, so @florencemashipei's approval survives your push, and none of my findings touch the model-facing strings, so nothing needs a second genealogy read. Fix the spec block and I will do the second approval.

johnmarkpeterbrown:
@clack391 — all eleven addressed in `be92b054`, including the seven you marked optional. Verified each claim against the code before changing anything rather than taking the write-up on trust; every one held.

**The merge condition.** Rebuilt the `FulltextSearchResponse` block from `origin/main` rather than hand-restoring the six lines, then re-added only this PR's two fields. The block now differs from main by exactly those two entries, and — checked with line endings normalised, which is what hid it from me the first time — the whole file differs by nothing else. I'd confirmed your three sub-claims first: `FulltextSearchResult` has 0 hits under `src/`, the spec declares `FulltextResult` at `:151`, and the real `facets` at `src/types/fulltext-search.ts:105` is flat.

Your point 2 is the one I want to acknowledge properly. My own rebase note told whoever merges #2001 that the conflicts were "new table rows and new test cases, not competing rewrites" — which was false at the moment I wrote it, because the merge had already eaten that block. Someone would have resolved a conflict against corrupted content on my say-so.

**`projectPath: null`.** Guarded in the reader, as you suggested, so a fourth caller inherits it. Your note on why the seven new tests miss it is exact and I've pinned the shape one over.

**The mock's nil gate.** Mirrored on both halves — `_fixture_is_nil` reads whichever total the fixture carries (`totalMatches` / `totalResults` / `totalForPlace`), and a fixture declaring no total is treated as nil. I rewrote the divergence comment above it too: it now names the upstream-total half explicitly and leaves the `external_links_search` pre-filter case as the one remaining divergence, which is still real.

**Both drift guards are in and both were watched failing** — mutating `UNLOGGED_REFS_SHOWN` to 4 on the Python side and dropping `external_links_search` from `STAGING_SEARCH_TOOLS` each fails its own test and nothing else.

**The TTL premise.** Confirmed: `pruneStale` has exactly one call site, inside `stageSearchResults`, after the nil early-return. Reworded, and there's now a test that stages a 30h-old file, runs a nil search, asserts the file is still on disk, and asserts the reader still ignores it — so the premise is pinned rather than asserted in a title.

**Optional 3, 5, 7, 9** all taken as described. On 9 I used `coerceJsonArg`, which turns `"5"` into 5 and leaves a genuinely non-numeric value alone — I stopped short of a refusal, since that's a behaviour change with its own blast radius and this PR isn't the place to decide it.

**@DallanQ's question, answered.** It never holds. Across **249** committed project states, **0** reach the 3-sources-with-zero-assertions threshold `sourcesWithoutAssertionsWarning` requires, and only **1** has any sources at all with zero assertions. So the zero firings are the good-news branch. Agreed on widening #2067 to all four fields rather than a fourth issue; doing that now.

Numbers on the new head: engine 118 files / 2808 passed / 1 skipped, harness 3047 / 3 skipped, packaging 616. PR body refreshed — you were right that it was stale in four places and miscited `research-log-append.ts:337` for a sentence that lives at `:334-335`.

Two things I've left alone deliberately. The paging-overrun probe is still an open offer to @florencemashipei, not something I'll decide by filing. And I didn't touch the third-planning-round override, which is yours and Dallan's to weigh.

Worth stating plainly: none of your eleven findings were in the mechanism, and four of them were in prose I wrote *about* the mechanism after reviewing it myself. The spec-block revert in particular was invisible to the full suite in both directions, which is the category I'm least able to catch on my own.

johnmarkpeterbrown:
@clack391 — ready for another look. Head is `9e60b017`. Two commits since your review: `be92b054` applies all eleven findings, `9e60b017` is the main merge you recommended doing first.

**The merge condition.** Rebuilt the `FulltextSearchResponse` block from `origin/main` rather than hand-restoring the six lines, then re-added only this PR's two fields. `staged?`, `stagingError?`, `FulltextResult[]` and the flat `facets?` are all back. Checked twice — after `be92b054` and again after the merge — and the block differs from main by exactly the two new entries, with the whole file differing by nothing else.

That second check mattered: **incoming main touched `search-result-staging-spec.md`**, one of my files, so I merged before asking you back rather than leaving you reviewing a state about to change. #2001 is still open, so nothing re-touched the fulltext spec. The merge was clean, no conflicts, and I re-verified the block and re-ran both suites after it.

The line-ending detail is worth passing on: my original whole-file diff was CRLF-vs-LF against `git show`, so every line read as changed and I took "everything differs" for "nothing to see". `tr -d '\r'` on both sides is what makes that check actually work.

**Fixed as specified:**

- `projectPath: null` guarded once in the reader, so a fourth caller inherits it. Test on the shape one over from the covered case — your read of why the seven new tests missed it was exact.
- The mock's nil gate mirrors production on both halves through `_fixture_is_nil`, reading whichever total the fixture carries (`totalMatches` / `totalResults` / `totalForPlace`); a fixture declaring no total counts as nil. I rewrote the divergence comment above it to name this half explicitly and leave the `external_links_search` pre-filter case as the one that remains.
- Both drift guards in, both **watched failing**: `UNLOGGED_REFS_SHOWN` at 4 on the Python side, and `external_links_search` dropped from `STAGING_SEARCH_TOOLS`. Each fails its own test and nothing else.
- The TTL premise is now stated correctly, with a test that stages a 30h-old file, runs a nil search, asserts the file is still on disk and still ignored — so the premise is pinned rather than asserted in a title.
- Optional 2, 3, 5, 7, 9 all taken. On 9 I used `coerceJsonArg`, which turns `"5"` into 5 and leaves genuinely non-numeric values alone; I stopped short of a refusal, since that is a behaviour change with its own blast radius.

**@DallanQ's question, answered.** The condition never holds: across **249** committed project states, **0** reach the 3-sources-with-zero-assertions threshold, and only **1** has any sources at all with zero assertions. Good-news branch. Issue #2067 is widened to all four fields and retitled, and I made "distinguish never-fired from fired-and-ignored, per field" the deliverable — that measurement is what shows why a bare zero is unreadable.

**Numbers on this head:** engine 118 files / 2808 passed / 1 skipped, harness 3047 / 3 skipped, packaging 616. PR body refreshed, including the four stale figures and the `research-log-append.ts:337` miscite (the sentence is at `:334-335`).

Still open and not mine to close: the paging-overrun probe offer to @florencemashipei, and the third planning round for you and Dallan.

One note for the record. None of your eleven were in the mechanism, and four were in prose I wrote *about* the mechanism after reviewing it myself and calling it clean. The spec-block revert was invisible to the full suite in both directions, which is the class I can least be trusted to catch alone — so thanks for diffing all 21 files rather than reading the diff I described.

clack391:
@johnmarkpeterbrown: all eleven verified fixed, each read in the code rather than taken from your summary, and the merge condition is satisfied. Two things were left. Rather than send them back I fixed them and pushed `f094271e` to your branch, on the lead's guidance that a minor issue found in review can be fixed in place. I also corrected two spots in the PR body that my own commit would otherwise have made staler. Everything is below so you can check my work rather than inherit it.

First the part that matters most: I re-ran the removals sweep across all 21 files at your new head and the clobber did not recur, including on `search-result-staging-spec.md`, which incoming main touched and which was the live risk on this merge. Merging first and re-verifying the block after is why. The block itself I checked in the direction that could falsify your claim rather than confirm it: whole file against `origin/main`, carriage returns stripped both sides, which returns exactly the 12 lines of your two new fields and nothing else.

## 1. The coercion silenced the data-loss warning, and that one is mine

`:290` persisted `coerceJsonArg(op.resultsAvailable)` while the retained-none warning below it still gated on the raw `op.resultsAvailable`:

| `resultsAvailable` | persisted | retained-none warning |
|---|---|---|
| `42` | `42` | fires |
| `"42"` | `42` | **silent** |

Before your change a string persisted as a string, which `unloggedStagedSearches` skips, so the staged file stayed unpaired and `unloggedSearches` kept nagging. One signal was on. After, the entry pairs, that note goes quiet, and the warning is still silent. Strictly fewer signals for that input than before.

It is the second-order version of @florencemashipei's thread 4, "the cheapest way for an agent to silence this note is a lossy log entry, and it pairs". The coercion widened that from lossy-but-well-typed entries to mistyped ones.

**This is my fault, not yours.** You implemented my note as written. I enumerated `coerceJsonArg` against seventeen inputs in isolation and never traced the field to its other consumer. A blind agent found it, not me.

Reachability is zero: none of the 2,222 committed `resultsAvailable` arguments under `eval/` is a string. So it was never urgent, only cheap.

## 2. My first version of that fix overreached twice.

Worth your eyes precisely because a reviewer's own commit is the code nobody attacks. I diffed my fix against yours across fourteen inputs through the real tool, and four rows differed where only two should have:

| input | yours | my first attempt |
|---|---|---|
| `"42"` | no warning | warns (intended) |
| `"5.7"` | no warning | warns (intended) |
| `"null"` | sets `results_available: null` | **field absent** |
| `"1e400"` | no warning | **warns, message reads "Infinity"** |

I had moved the presence guard onto the coerced value and let a non-finite number through the gate. Both corrected before pushing: the guard reads the raw argument again, and the gate takes `Number.isFinite`. Re-ran the differential and exactly two rows now differ from your behaviour, both the intended fix. That table is the thing to re-check if you check nothing else.

## 3. The nil-gate fix was correct in the helper and unheld at its call site

Reverting only the call site to `if not response.get("results"):`, leaving `_fixture_is_nil` untouched and correct, left the entire harness suite green at 3048. The two `_fixture_is_nil` unit tests prove the predicate; nothing proved it was called. The reason is that the only fixture routed through dispatch is `record-search-patrick-flynn-no-results`, whose `totalMatches` is 0, which is the one case where the old and new conditions agree.

The commit adds a test for the wiring. The discriminating response is built in `tmp_path` rather than committed, since it exercises a branch instead of describing a real search. Proven through the cycle: passes on the correct code, fails when the call site is reverted, suite green with it in.

## 4. The divergence comment described behaviour the same commit removed

It said a host-filtered-to-zero fixture "would still over-emit here". Executed: `_fixture_is_nil({"totalForPlace": 10, "results": []})` returns `False`, so it no longer over-emits. Your fix closed that case and the comment still claimed it. It also missed the other direction: `allLinks` is the year-filtered set, as your own comment at `external-links-search.ts:177` says, while `totalForPlace` sits before the year filter, so a fixture whose year window emptied a non-zero total under-emits.

Restated structurally in the commit, because the enumerated version has now gone stale twice. You already did this correctly four hundred lines away: the new TTL test's comment gives the structural reason, which is why that one cannot rot.

Same class one layer up, also in the commit: `ownership.json` described the nil gate as reading an upstream total for all three tools, and `external_links_search` has no upstream total to read.

## 5. The citation, and why I did not give it a new line number

`:334-335` was **correct** at `b414e19f`. `be92b054`'s own six inserted comment lines pushed the sentence to `:339-340`, so the commit that fixed a stale citation invalidated its own fix in the same diff. My commit inserts seven more lines and moves it again to `:346-347`.

So I did not renumber it. The body now quotes the sentence instead, which cannot rot while this PR keeps editing that file. Worth carrying as a habit rather than a one-off.

## 6. Nothing needed from you

Your comment says the TTL test stages a 30h-old file; it backdates by 25 hours, which is correct for a 24h TTL. Only the prose overstates, and it is in a comment rather than the code, so I left it.

`replay.py` is a third consumer of `results_available`: `_LOG_FIELD_RENAMES` rebuilds the entry from arguments and does not coerce, by its own renames-only contract, so a replayed entry differs from a live one for a stringly-typed value. I executed that rather than assuming it, and **documented** it in the coercion comment rather than changing it, because the replay contract is a harness decision and not this PR's to make.

## What I checked and found right

The corpus measurement is exact under three independent derivations. Your 249 project states is the right denominator and my earlier 359 was the wider file set, so that one was mine too. Both drift guards fire on shapes you did not test, which I checked by adding a fourth tool to the Python set and removing one from the TypeScript side. Your round-2 commit was 11 files, +164/-32, scoped to the eleven findings with no creep. The claim audit came back 23 of 23 supported.

Your answer to @DallanQ holds. I built a transient-state reconstruction to test it, on the reasoning that the warning is evaluated per call and a final-state scan cannot see a state a run passes through. It produced one candidate and I withdrew it when my own prediction failed against that run's final `research.json`: my counter had treated `op: "update"` rows as appends. Recounted correctly, the final source total caps that call below the threshold. Your conclusion is right, and it now rests on two methods rather than one.

One for #2067 while you are in there: in that same run six source appends were issued and two landed, and the responses are stripped. So a run log's issued ops are not a proxy for document state, and the per-field counter you scoped will need applied state or captured responses rather than arguments.

## Suites and what I changed in the body

Engine 118 files / 2809 passed, harness 3049 passed / 2 skipped, on `f094271e`. Your own 2808/1 and 3047/3 were never wrong: `no-project.test.ts:243` is `it.skipIf(process.platform === "win32")`, so yours is a Windows run and mine Linux, with identical totals.

In the body I changed only what my commit made staler: the citation above, and the harness count. I added a short "Reviewer commit" section so the history is not silently mixed. Nothing else in your body was touched.

clack391 (approved):

PR #2084, opened 2026-08-31
Delete senior-queue and its two labels: CODEOWNERS already gives seniors the queue natively

DallanQ:
Deletes `senior-queue.yml` and the two labels it maintained,
`ready-for-senior-developer` and `ready-for-senior-genealogist`.

## Why

The workflow labelled a PR once CI was green, a peer had approved, and no review
thread was outstanding, picking the team from the first CODEOWNERS owner of the
changed paths. Seniors were meant to work from two saved searches on those labels.

**CODEOWNERS already requests review from both senior teams the moment a PR opens** —
verified on PR #2064 (`requested_teams=[senior-genealogists,senior-developers]`) and
PR #2014. So every senior already has that queue natively, on GitHub's own
review-requests page. The label added exactly one thing on top of routing GitHub does
for free: a readiness filter. And getting a senior to look at a ready PR is the
author's job.

Ruled on issue #1968, which was resolved as not planned rather than fixed.

## What was verified before deleting

- **Nothing in the repo reads the labels.** `ready-for-senior` appeared in exactly one
  tracked file — the workflow itself. No skill, no other workflow, no doc.
- **The queue was converging correctly.** At the time of the decision all six
  genuinely-ready open PRs carried a label, and every unlabelled one was correctly
  unlabelled (outstanding changes-requested, or no review at all). Zero false negatives.
- **`senior-queue-readiness.test.mjs` exists only for this workflow** and goes with it.
- **`codeowners-routing.test.mjs` was not only a routing test** — see the next section.

## What survives the workflow: two CODEOWNERS invariants

Review caught that two of that file's three checks are statements about who may
**merge**, not about which queue a PR lands in:

- every rule names BOTH senior teams (owners are an OR, so this is what lets a senior
  of either kind unblock any PR)
- no `*` default

**GitHub checks neither.** `codeowners/errors` reports unparseable patterns and
nonexistent teams; a rule that parses and names a real team is accepted whatever it
does to the merge gate. And `.github/CODEOWNERS` is claimed by no rule in itself — no
pattern matches an extensionless file at that path — so without a check, two junior
approvals could narrow who is allowed to merge with every check green.

Both are re-implemented in `.github/tests/codeowners-teams.test.mjs`, which lifts no
parser from any workflow: split the line, compare the owner list. **Proven to fail
before committing** — clean on the real file (19 rules), red on dropping
`senior-genealogists` from the `*.ts` line, red on adding a `*` default, clean again on
restore. The `.github/CODEOWNERS` trigger path is restored with it.

The third check, the path-to-team table, stays deleted: it genuinely needs the
workflow's parser, and any edit that drops a team now fails the first check anyway.

## What goes with it

- a `*/30` cron that was dropping most of its runs — a real 11h23m gap on 2026-08-28
  against an expected 30 minutes
- a `pull_request_target` trigger the file itself flagged as a standing pwn-request
  surface requiring permanent care
- the `readiness` job in `workflow-logic-tests.yml`
- 772 lines of workflow and test code, net of the new test

## What is deliberately NOT changed

`.github/CODEOWNERS` is untouched. It still requires a senior approval on the same
paths, and branch protection still enforces it — this removes a notification
convenience, not a merge gate.

The gap the deleted test explicitly could **not** cover is unchanged, and the new one
does not cover it either: a CODEOWNERS rule naming a nonexistent team is silently
dropped by GitHub, and the only way to see it is
`gh api /repos/{owner}/{repo}/codeowners/errors` after editing.

## Docs

`docs/task-lifecycle.md` loses the two sentences pointing at the workflow. It also
gains a caveat the deleted workflow header was the only place recording: the
`DELETE /pulls/{n}/requested_reviewers` call the doc recommends works for a team the
current file no longer claims, but against a team CODEOWNERS *does* still mandate,
under `require_code_owner_review`, it returns HTTP 200 and does nothing.

## Follow-up for whoever merges this

The two labels still exist on the repo and sit on 5 open PRs (#2033, #2027, #2000,
#1929, #1877 — it was 6 when this PR opened; #2014 has since merged). Delete them
**after** this merges, or the workflow still on `main` re-adds them on its next sweep:

```sh
gh label delete ready-for-senior-developer   --repo PioneerAIAcademy/cowork-genealogy --yes
gh label delete ready-for-senior-genealogist --repo PioneerAIAcademy/cowork-genealogy --yes
```

## Verification

- `node .github/tests/project-status-sync-routing.test.mjs` — pass
- `node .github/tests/codeowners-teams.test.mjs` — pass, and proven to fail on two
  mutations before being committed
- `node .github/tests/check-runlogs-post-gate-lints.test.mjs` — pass
- `python3 scripts/claude-hooks/test-gate-issue-create.py` — 12/12 pass
- every remaining `run:` step in `workflow-logic-tests.yml` points at a file that exists;
  the YAML anchor/alias pair is intact
- no tracked file references `senior-queue`, `ready-for-senior`, `codeowners-routing`,
  `queue label` or `both queues` — the second sweep matched on meaning, not just the two
  literal strings the first one used, which is how a stale line at
  `docs/task-lifecycle.md:284` survived the first pass

https://claude.ai/code/session_01LJMa4biet16dH62y8KZb9d


DallanQ:
Good catch on the test file — you're right and I was wrong about it, in a way I should have caught by reading the checks instead of the filename.

**Fixed, in `d42c54d8`.**

I re-derived it rather than taking it on faith: the file has three checks, and checks 2 and 3 in it are `r.teams` off the workflow's parser — but the *rule* check 2 asserts is parser-independent, which is the whole point. Confirmed the hole is real: `.github/CODEOWNERS` matches no pattern in itself (no rule covers an extensionless file at that path), so with the trigger path and the test both gone, two junior approvals could drop a senior team from a rule and go green.

Wrote `codeowners-teams.test.mjs` as you specced it and broke the repo to watch it fail before committing — clean on the real file (19 rules), red on dropping `senior-genealogists` from `*.ts`, red on adding a `*` default, clean again on restore. Trigger path restored, job added, header rewritten to cover both governance files again.

**Agreed on dropping check 3.** The path-to-team table needs the parser, and any edit that drops a team now fails check 1 anyway — so the table's useful half is covered without carrying a lifted parser.

**Stale line: fixed, and thank you.** Took your replacement verbatim. Worth naming why it survived: my sweep matched the literal strings `senior-queue` and `ready-for-senior`, and that line says "queue labels". I matched the string, not the meaning. Re-swept on `queue label` and `both queues` too — nothing else left.

**On the two negatives — I split them, which is a partial disagreement with "let both go".**

- `resolveReviewThread` refused to `GITHUB_TOKEN`: let it go, agreed. It was a "don't retry this" about a workflow that no longer exists, and nothing left in the repo can attempt it.
- The `DELETE /pulls/{n}/requested_reviewers` no-op: kept, as one clause at the doc's own recommendation. Your hedge is right that the two cases differ — the deleted header says GitHub refuses to remove a **CODEOWNERS-mandated** team request, and the doc's case is a team that no longer owns any file, so that advice probably works. But the doc now recommends a call with no note of the condition under which it silently returns 200 and does nothing, and "silent success reads as working" is exactly the trap. That's operational guidance rather than history, so it isn't the thing the no-history-in-team-docs rule is aimed at.

**Label count: 5 is right now, 6 was right when I wrote it** — #2014 merged in between, and it's the same commit that put this branch one behind. Body updated with the explicit list.

PR #2087, opened 2026-08-31
feat: wire given-name diminutive expansion into fulltext_search and image_transcribe

### LINKED ISSUE #607: fulltext_search + image_transcribe: wire given-name diminutive expansion (the table is issue #1868)

yinkid28:
## Summary

Closes #607. Wires the given-name diminutive variant table (landed in PR #1944, issue #1868) into `fulltext_search` and `image_transcribe` so that searching for a formal given name also matches its historical diminutives.

- **`fulltext_search`:** Recognized English given names in the `name` parameter are expanded into a Lucene OR group (e.g. `(Elizabeth OR Betty OR Betsy OR Bess ...) Martin`). The expansion is bidirectional — searching for "Betty" also includes "Elizabeth" and all sibling variants. A `nameExpansion` response field reports what was expanded and which variants appeared in results. Period-containing scribal abbreviations are excluded from the OR group to avoid Lucene parse errors.
- **`image_transcribe`:** `lookingFor` is expanded with all variant forms (including scribal abbreviations) in a natural-language `(also known as ...)` clause before building the VLM prompt. No response changes needed.
- **Bundling fix:** Moved `given-name-variants.json` from `src/data/` to `config/` so the `.mcpb` build bundles it — `tsc` doesn't copy JSON from `src/`, and `.mcpbignore` excludes `src/`. The `config/` directory is already staged by `scripts/build-mcpb.mjs`.

**Touches:** `packages/engine/mcp-server/src/tools/fulltext-search.ts`, `packages/engine/mcp-server/src/tools/image-transcribe.ts`, `packages/engine/mcp-server/src/types/fulltext-search.ts`, `packages/engine/mcp-server/src/utils/name-variants.ts` (new), `packages/engine/mcp-server/config/given-name-variants.json` (moved from `src/data/`), `docs/specs/fulltext-search-tool-spec.md`, `docs/specs/image-transcribe-tool-spec.md`

## Design decisions (all from issue #607 rulings)

- **Host-side, not prose.** Issue #1007 measured ~50 lines of triage doctrine being complied with in only 3 of 4 runs — prose does not reliably change behavior. Tool-side expansion cannot decay.
- **Expand and report which variants fired.** Returning variants as a nil-result hint was rejected because it only helps if the agent acts on it.
- **Optional group, not require-default off.** The OR group keeps `m.queryRequireDefault=on` intact; the group is required as a whole while any member matching satisfies it.

## Pre-merge gate

The Lucene OR group syntax `(term1 OR term2)` in `q.fullName` has not been tested against the live FamilySearch fulltext API (requires auth). If the API rejects it with a 400, the fallback is multi-query merge — the utility function signatures are unchanged, only the call site in `fulltextSearchTool` would change.

## Test plan

- [x] `npm run build` — TypeScript compiles cleanly
- [x] 98 new/affected tests pass (23 utility + 45 fulltext + 30 image-transcribe)
- [x] Full suite: 2759 pass, 1 pre-existing unrelated failure (windows-wrappers worktree detection)
- [ ] Probe live FamilySearch fulltext API with `name: "(Elizabeth OR Betty)"` to confirm OR group syntax is accepted
- [ ] Verify `.mcpb` build packages the variant table from `config/`


yinkid28:
Superseded by a clean branch — the original was branched off 1743-dead-stub-agent-smoke instead of main, pulling in unrelated commits. Reopening from a clean branch.

PR #2089, opened 2026-08-31
fulltext sidecar cards show real title, correct FamilySearch link

### LINKED ISSUE #1996: fulltext sidecar cards show the record type instead of the title, and their FamilySearch link is dead

taiwo-stack:
Closes #1996.

## #2049 merged into main first — what's actually left in this diff

While this PR was in review, #2049 ("Constrain the destination of links the
UI presents as FamilySearch", #1018) merged into `main` and independently
fixed the title/link bugs this PR originally targeted — with a much more
thorough, security-hardened approach (main-process link-channel enforcement,
resolvability gating instead of truthiness gating, a shared
`resolveFamilySearchTarget` policy with its own drift tests between the
Electron and web copies). It fully supersedes this PR's original
`PersonCard.tsx`/`SidecarResultCard.tsx` link fixes and helper module, which
have been dropped in favor of main's version.

What #2049 does **not** touch, and what this PR still adds:

- **The actual root cause of #1996**: both hand-maintained `FulltextSearchResult`
  mirrors (`packages/schema`, `eval/app`) were missing five fields the shipped
  tool has — `sourceUrl`, `collectionId`, `title`, `recordDate`, `dates`. Fixed
  both to match the shipped type field-for-field.
- The card now titles from `title` first (falling back to `recordType` then
  `collectionTitle`), instead of skipping straight to `recordType`.
- A merged/deduped `Dates` row combining `recordDate` with `dates` — a probate
  filing can carry a filing date plus an earlier death date in the document
  body, so the two are not redundant.
- The fulltext footer link now prefers `sourceUrl` (the upstream-supplied
  resolver URL) when it resolves, falling back to `id` — which itself can be
  a legacy full-URL shape for sidecars staged before #272. Layered on top of
  #2049's resolvability gating rather than replacing it.
- Fixed `patrick-flynn.ts`'s `ark` fixture, which stored the non-canonical
  full-URL form — the same fixture-shape trap #2049's own review caught
  elsewhere — and the `relationship-label.test.ts` assertion that had pinned
  that wrong shape as correct.

## Test plan

- `packages/viewer-ui` vitest: 327/327 passed, including #2049's
  `link-channel-routing` test confirming `SidecarResultCard.tsx` still routes
  all 3 sinks through the constrained channel after this merge.
- `make typecheck`: green (5 tasks).
- Every new/changed assertion mutation-tested: reverted the fallback logic,
  confirmed the affected tests failed on the real assertion; restored,
  confirmed green again.
- No eval-skill run needed — no `SKILL.md`, agent body, rubric, or unit test
  touched.


jud-sdev (approved):
Went through this carefully and ran it in a worktree. The two `FulltextSearchResult` mirrors now match the shipped `FulltextResult` field-for-field, the tool actually populates `title`/`sourceUrl`/`recordDate`/`dates` (fulltext-search.ts:115-125), and the card's title fallback plus the `sourceUrl` link with the bare-ARK fallback are correct — `id` is a canonical ARK, so the derived `https://www.familysearch.org/<ark>` URL is clean. SidecarResultCard tests pass 13/13 and the workspace typecheck is green. The new cases are real guards — each fails under the old card.

The replaced fixture is the important part: the old one gave `id` a full URL and no `title`, a shape the tool never emits, which is exactly why both bugs sailed through CI. Good catch.

One thing not blocking this, filed separately as issue #2091: while reviewing I found the same mirror-drift has already broken record_search tree matches — the tool emits `treeMatches` as `{ treePersonId, stars }` but the mirror's `TreeMatch` is `{ personId, personName, treeId, ark }`, so the card reads fields that are not there and those rows render blank. That is a different result type and a bigger fix (the engine does not emit those fields), so it is its own issue rather than something to grow this PR. It is also good evidence for a drift guard between the mirrors and the engine types, since that gap has now produced two bugs.

This still needs a senior-genealogist as well, since the `.ts`/`.tsx` paths route to both teams.

florencemashipei (changes requested):
Thanks for this — the mirror fix is solid. I checked both `FulltextSearchResult` mirrors against the shipped `FulltextResult` and they match field-for-field. Two things to change before I approve.

**1. The link fallback breaks on sidecars already on disk.**

`id` is a canonical ARK for anything the tool emits today, so the derived URL is right for new searches. But sidecars staged before #272 store `id` as a full resolver URL with no `sourceUrl`, and the viewer still opens those projects. Four of the scenario fixtures under `eval/fixtures/scenarios/` hit this. They render a URL with the host twice, and `openExternal(result.id)` used to open them correctly — so this is a regression.

You wrote:
```
              openExternal(result.sourceUrl || `https://www.familysearch.org/${result.id}`)
```
Change it to:
```
              openExternal(result.sourceUrl || familySearchUrl(result.id))
```

And add this above `function FulltextSearchBody(`:
```tsx
// Mirrors toArk()/arkToUrl() in packages/engine/mcp-server/src/utils/ark.ts.
// `id` reaches the viewer in three shapes: a full resolver URL (sidecars
// staged before #272, still on disk), a bare `ark:/61903/<n:n>:<id>`, and a
// type-prefixed `3:1:<id>`. Only the middle one can take the host prefix.
const FS_URL_PREFIX_RE = /^https?:\/\/(?:www\.)?familysearch\.org\//i
const BARE_PREFIXED_RE = /^\d:\d:[A-Za-z0-9.-]+$/

function familySearchUrl(id: string): string {
  const trimmed = id.trim()
  if (FS_URL_PREFIX_RE.test(trimmed)) return trimmed
  const ark = BARE_PREFIXED_RE.test(trimmed) ? `ark:/61903/${trimmed}` : trimmed
  return ark.startsWith('ark:/') ? `https://www.familysearch.org/${ark}` : trimmed
}
```

**2. The new fixture removed the case that would have caught it.**

The comment is right about the tool and wrong about the data. Worth keeping both shapes.

You wrote:
```
  // never actually emits, which is why the title/link bugs passed CI.
```
Change it to:
```
  // never emits today. Sidecars staged before #272 do carry that shape and
  // are still on disk, so it is covered separately below.
```

And add a case:
```tsx
  it('opens a legacy full-URL id as-is, without prefixing the host again', async () => {
    const { sourceUrl: _sourceUrl, ...legacy } = fulltextResult
    const spy = vi.fn()
    setOpenExternal(spy)
    render(
      <SidecarResultCard
        result={{ ...legacy, id: 'https://familysearch.org/ark:/61903/1:1:K5M9-P3WT' }}
        tool="fulltext_search"
        defaultExpanded={true}
      />
    )
    await userEvent.click(screen.getByRole('button', { name: /Open in FamilySearch/ }))
    expect(spy).toHaveBeenCalledWith('https://familysearch.org/ark:/61903/1:1:K5M9-P3WT')
  })
```

**3. Not blocking, your call.** Titling from `title` means `recordType` and `collectionTitle` no longer appear anywhere on a fulltext card — the sub-meta strip is gated to record_search only. The header used to show `recordType`. If that context is worth keeping, ungating that strip for fulltext is the smallest change.

Also needs a merge from main — it's behind, and one drift I initially thought was here turned out to be already fixed up there.

DallanQ:
Please widen this PR to the two remaining bare-ARK call sites before it merges. A tester just hit the same dead link from a different screen (feedback issue #2167), and since this PR closes issue #1996, those sites lose their home when it lands.

**1. `PersonCard.tsx:18` — `openExternal(person.ark)`.** `person.ark` is a bare ARK by specification, not by accident: `docs/specs/simplified-gedcomx-spec.md:131` defines it as "canonical form (e.g. `ark:/61903/4:1:KGS8-LY1`) ... with the resolver-URL prefix stripped." This is the link the tester reported:

> On the "Known Information" page, there are links next to each person that say "View on FamilySearch." Those are ARK links and don't seem to resolve to the person page in the family tree.

Note their second clause. A person ARK and the tree profile page are different destinations, so check where yours lands rather than assuming the fallback pattern you used for fulltext transfers unchanged.

**2. `SidecarResultCard.tsx:125` — `openExternal(tm.ark)`,** in the file you are already editing. `TreeMatch.ark` is an untyped optional string in `packages/schema/src/index.ts`; line 142 of the same file passes `result.arkUrl`, so the codebase distinguishes bare from URL by name. Verify which this is and fix it if it is bare.

**3. Add a `PersonCard` test.** There is no test file for it at all, and `packages/viewer-ui/src/lib/__fixtures__/patrick-flynn.ts:553` stores `ark` as a full URL (`https://familysearch.org/ark:/61903/4:1:KGS8-LY1`), which the spec says is not the canonical shape — the same fixture-shape trap you already called out in your test plan for the fulltext card. Fix the fixture to the bare form or the new test passes while production stays broken.

**4. Consider putting the conversion in one place.** `openExternal` validates nothing — `packages/viewer-ui/src/lib/external.ts:17-19` is `if (url) impl(url)` — so every future caller can pass a bare ARK and get a silently dead link. Three call sites needing the same conversion is the signal to lift it into a shared ark-to-URL helper (or guard inside `openExternal`) rather than inlining the fallback a third time. Your call which, but say which you chose and why in the PR body.

The other three call sites are fine: `SourcesSection.tsx:70` and `ResearchLogSection.tsx:214` take real URLs, and `Linkify.tsx:73` takes an extracted one.

**When this merges, email the tester who filed feedback issue #2167.** They are identified in that bundle's `_feedback/feedback.json` — the feedback endpoint files every submission under the lead's account, so the issue author is not the tester. Tell them which screens were affected.

packages/viewer-ui/src/components/shared/SidecarResultCard.tsx
```diff
@@ -185,12 +193,20 @@ function FulltextSearchBody({ result }: { result: FulltextSearchResult }): React
...
+        </div>
+      )}
       {result.id && (
         <div className={styles.footerLink}>
           <button
             type="button"
             className={styles.externalLink}
-            onClick={() => openExternal(result.id)}
+            onClick={() =>
+              openExternal(result.sourceUrl || `https://www.familysearch.org/${result.id}`)
```
florencemashipei: This breaks the link for sidecars that are already on disk.

`id` is a canonical ARK for anything the tool emits today, so the derived URL is correct for new searches. Sidecars staged before #272 store `id` as a full resolver URL with no `sourceUrl`, and the viewer still opens those projects. Four scenario fixtures under `eval/fixtures/scenarios/` hit it, producing a URL with the host twice:

`https://www.familysearch.org/https://familysearch.org/ark:/61903/1:1:K5M9-P3WT`

The previous `openExternal(result.id)` opened those correctly, so this is a regression.

Change it to:
```
              openExternal(result.sourceUrl || familySearchUrl(result.id))
```

The helper and a test case are in my review above. `arkToUrl` in `packages/engine/mcp-server/src/utils/ark.ts` already has both guards, so it is worth mirroring rather than reinventing.

florencemashipei:
Findings are in my review above, with the one that matters left as a thread on the link fallback.

For routing: this is labelled `ready-for-senior-developer`, so it sits in the developer queue rather than the genealogist one — every CODEOWNERS rule here names both senior teams as an OR, so @jud-sdev's approval already satisfies the code-owner requirement. Any senior developer can verify the fix and give the second approval. No need to wait on me.

taiwo-stack:
@florencemashipei heads up — main merged #2049 while this was in review, which superseded the PersonCard/SidecarResultCard code your review was against. Merged main in and rebuilt on top of it; ready for a fresh look.

packages/viewer-ui/src/components/shared/SidecarResultCard.tsx
florencemashipei: Superseding my comment above — #2049 landed the resolver and this is rebuilt on it, so the double-host case is genuinely gone. Dropping the local helper was the right call.

One thing left on the new code, and it is the same class of bug this PR closes.

The gate is resolvability-based:

```
  const fulltextLinkTarget =
    (result.sourceUrl && resolveFamilySearchTarget(result.sourceUrl)) ||
    (result.id && resolveFamilySearchTarget(result.id))
```

The click is truthiness-based:

```
            onClick={() => openFamilySearch(result.sourceUrl || result.id)}
```

When `sourceUrl` is present but does not resolve, and `id` does, the gate passes on `id` while the click hands the sink `sourceUrl`. Both transports resolve that to `null`, so the button renders and opens nothing.

`ARK_RE` is `$`-anchored, so any ARK URL carrying a page parameter fails to resolve — `https://www.familysearch.org/ark:/61903/3:1:S3HT-XKR9-NXF?i=353`, the shape `Linkify.test.tsx` already pins as occurring in project data. No committed fulltext fixture has it in `sourceUrl` today, so this is a latent defect rather than a live one.

Change it to:
```
            onClick={() => openFamilySearch(fulltextLinkTarget)}
```

`fulltextLinkTarget` is already the resolved URL and re-resolving it at the sink is idempotent, so the value you gate on becomes the value you open. Worth a test where `sourceUrl` is a familysearch.org URL with `?i=`, asserting the sink receives something resolvable — the current cases cover resolvable, absent, and empty-string, but not present-and-unresolvable.

florencemashipei:
@jud-sdev could you re-read this one when you get a chance?

Your approval is still recorded against `a5149dda`, and because the ruleset does not dismiss stale reviews on push it still counts toward the two required. The problem is that nothing it read survives: #2049 landed in between and taiwo rebuilt on top of it, so the `PersonCard`/`SidecarResultCard` link fixes and the local helper you reviewed are all gone in favour of main's `resolveFamilySearchTarget`. The current head is `3e7309e3`.

I have re-reviewed it and left one finding on the thread above — the link gate and the click handler use different values, so the button can render and open nothing. One-word fix.

Everything else checks out: both mirrors match the shipped `FulltextResult` field-for-field, all four of @DallanQ's widening items are covered, the merge with main is clean, and viewer-ui is 327/327.

taiwo-stack:
@jud-sdev fixed — pushed 52910f61. The gate and click handler now derive from the same raw-field selection (same resolvability check, but keeping the raw identifier per this component's existing pattern), plus a regression test for the exact gap (sourceUrl present but unresolvable, id that resolves). 328/328 green.

packages/viewer-ui/src/components/shared/SidecarResultCard.tsx
taiwo-stack: Same bug jud-sdev's re-review caught — fixed in 52910f61, pushed just before this comment landed. Went with raw-passthrough instead of `openFamilySearch(fulltextLinkTarget)`: passing the resolved value re-normalizes the legacy full-URL id case (adds `www.`), which breaks the existing "legacy full-URL id as-is" test — confirmed by trying it and watching that test fail. Computed which raw field wins with the same resolvability check instead, matching how `tm.ark`/`result.arkUrl` already work. Also added your `?i=` query-param case in 9ff5132c — realistic instance of the same fallback branch.

florencemashipei (commented):
Re-reviewed at `d94e7581`. My finding is fixed and my earlier changes-requested no longer applies.

Your approach was the right one and better than what I suggested — passing the resolved value re-normalizes the legacy full-URL id and adds `www.`, which breaks the "as-is" case. Computing the raw winner with the same resolvability check keeps the gate and the click on one field, and matches the other two sinks in the file. I checked it against every shape combination: both resolvable, `sourceUrl` with `?i=`, legacy full-URL id, empty `sourceUrl` with a bare id, both unresolvable, and a spoofed `familysearch.org.evil.com` host — the gate and the click agree in all six, and the spoof correctly falls through to the ARK. The `?i=` regression case pins the branch that was broken.

Also verified: merges clean with main, typecheck green, viewer-ui 329/329, and both mirrors still match the shipped `FulltextResult` field-for-field.

No objection from me. Leaving the second approval to the senior developer queue, since this is labelled `ready-for-senior-developer` and @jud-sdev's approval is still recorded against `a5149dda` — which predates the whole #2049 rebase, so none of the current code has actually been read by it. Worth a fresh look from that side before it merges. It is also 11 commits behind main again.

florencemashipei (approved):
Approving at `d94e7581`.

The gate/click fix is correct, and your approach was better than the one I suggested — passing the resolved value re-normalizes the legacy full-URL id and adds `www.`, which breaks the "as-is" case. Computing the raw winner with the same resolvability check keeps the gate and the handler on one field and matches the other two sinks in the file.

Verified across every shape combination: both resolvable, `sourceUrl` with `?i=`, legacy full-URL id, empty `sourceUrl` with a bare id, both unresolvable, and a spoofed `familysearch.org.evil.com` host. The gate and the click agree in all six, and the spoof correctly falls through to the ARK. The `?i=` regression case pins the branch that was broken.

Also checked: merges clean with main, typecheck green, viewer-ui 329/329 in 37 files, both `FulltextSearchResult` mirrors still match the shipped `FulltextResult` field-for-field, and all four of @DallanQ's widening items are covered — `PersonCard` and the tree-match sink route through the constrained channel via #2049, `PersonCard.test.tsx` exists, the non-canonical `ark` fixture is fixed along with the assertion that had pinned it, and the conversion lives in one place.

Two things before merge, neither blocking my approval: it is behind main and needs a fresh merge, and @jud-sdev's approval is still recorded against `a5149dda`, which predates the #2049 rebase — none of the current code has been read by it, so a fresh look from the developer side is worth having rather than letting that one carry the second slot.

And per @DallanQ's note above: the tester behind feedback issue #2167 should be emailed once this lands.

PR #2090, opened 2026-08-31
eval: conflict-resolution validators V6 (one conflict per turn) and V2 (word caps) (#1972)

francis-2008-happy:
## Summary

- Two of the seven validator requests from the conflict-resolution deep dive (#1652): **V6** (at most one conflict resolved per invocation, gating) and **V2** (word caps on `weighing_analysis` / `resolution_rationale`, reporting). **Issue #1972 stays open** — V1, V3, V4, V5 and V7 are untouched.
- Both live in `eval/harness/validators/test_conflict_resolution.py`, with a new direct test file. No skill body, rubric, fixture or run log is touched.

## Review guidance

**Start here:** the two band constants and the comment above them. The issue floats 240/300; **240 is wrong for `weighing_analysis`, whose corpus maximum is 251** — it would report 2 of the 13 writes over the skill's own 200 cap, and a green arm would read as "weighing lengths are fine". Set from each field's own distribution instead: weighing > 210 (reports 7 of 13), rationale > 300 (12 of 17).

**Unsure about:** whether V6 should gate at all. It is a clean count with no judgement, and `ut_conflict_resolution_002`'s own `judge_context` already states the rule — but a gating failure suppresses the LLM judge for that run, so a false positive costs a graded run. The corpus says 2 violations in 62 runs, both the known case. If a genealogist thinks a legitimate turn could touch two conflicts, this should be `report_*` instead and that is a one-word change.

## Plan

**Didn't change:** V7 (blocked — issue #1629 and PR #1931 both open, and #1972 is explicit that #1629 must land first or a V7 scan silently reads a stale run log), V3 (blocked on #1823 step 2), V5 (its own PR — engine `validator.ts` plus four fixture corrections, wants a senior on the fixture half), V1 and V4 (tier-2, lower confidence; V4's second half needs a lead ruling on `hypotheses[].status`).

The issue proposes "V6 + V7 as one small PR". That is no longer available, so V2 is the substitute: same lane, same file, same instrument, unblocked since #1922 merged.

**Acceptance check:** `cd eval/harness && uv run --frozen pytest tests/unit/test_conflict_resolution_validator.py` — 13 pass, including a **corpus replay** that runs both validators over the five committed `eval/runlogs/unit/conflict-resolution/v1_*.json` and pins the exact firing set: V6 flags `ut_conflict_resolution_002` in `v1_2026-08-18_15-37-43` and `v1_2026-08-18_19-42-11` and nothing else; V2 reports 7 weighing and 12 rationale writes.

Both mutation-checked:

| Mutation | Result |
|---|---|
| exempt all newly-created conflicts in V6 | `test_v6_creating_an_already_resolved_conflict_counts` dies |
| read `competing_assertion_ids` from the diff in V2 | the regression test **and** the corpus-replay count both die |

**Deviated from the plan:** two changes after adversarial review, both blocking findings I had wrong.

1. **The bands.** The first draft took 240/300 from the issue and justified it against the 460-word worst case — but 460 is a *rationale* figure and says nothing about weighing. Re-measured per field and reset.
2. **One helper became two.** The first draft shared a "modified conflicts" helper between V6 and V2 and sold it as drift protection. The rules genuinely need different populations: V6 asks how many conflicts a run *resolved*; V2 asks how long the prose it *wrote* is, which includes conflicts the run created. The shared version silently dropped a created conflict carrying a 460-word rationale, and gave V6 a bypass where a create arriving already `resolved` escaped the count.

## Test plan

- [x] `make harness-test` — 2988 passed. The 10 failures are pre-existing and environmental: 9 in `test_setup_feedback_case.py` need `unzip` (not installed here), 1 in `test_mock_mcp.py` needs a built engine. Both clean after `npm run build`; neither touches this diff.
- [x] Measurements re-derived against the post-#1973 corpus rather than quoted: 32/37 rationale over cap, 13/37 weighing over cap, 17 of 32 where the escape cannot apply, worst case 460 words at `v1_2026-08-19_15-24-31` / `ut_conflict_resolution_001` / `c_001`.
- [x] No paid eval run, verified both halves: `harness/snapshot.py::build_snapshot` does not cover `eval/harness/validators/**`, and `check_runlogs.py` derives touched skills from path regexes a validator-only change matches none of.

## The trap worth knowing about

**`competing_assertion_ids` must be read from `after_state["research_json"]`, never from a diff of changed fields.** A run almost never writes that field — it is absent from `changed_fields` on **0 of the 32** over-cap writes in the corpus — so a diff-based read makes the three-or-more-way escape never apply and roughly doubles the finding, 17 → 32. I hit this while re-deriving the numbers before writing any code. It is a code comment and a named regression test, not just a note here.

## Follow-on issues

None filed. Everything deferred above is already tracked on #1972 with its gate named.


yinkid28 (changes requested):
# Peer Review: 1972-conflict-resolution-v6-v2

**Base:** main | **Commits:** 2 | **Files changed:** 8

## Summary

Two pieces of work: (1) deleting `senior-queue.yml` and its tests, replaced by a simpler CODEOWNERS-teams test; (2) adding V6 (gating, one conflict per turn) and V2 (reporting, word caps) validators for conflict-resolution. The validator code is well-structured, the V6/V2 population split is deliberate and documented, the create-and-resolve bypass is correctly caught, and the `competing_assertion_ids` after-state-vs-diff trap is correctly handled with a named regression test. The senior-queue deletion is clean with no dangling references. Test coverage is good but has a few gaps at the boundaries.

## Warnings (should fix)

### eval/harness/validators/test_conflict_resolution.py

- **Line 411-413** [correctness]: Misleading comment. "it is absent from changed_fields on 0 of the 32 over-cap writes" literally means "present on all 32," but the intended meaning is the opposite — it NEVER appears in changed_fields. The same confusing phrasing also appears in the test file at lines 151-152. **Fix:** Reword to "it appears in changed_fields on 0 of the 32 over-cap writes" or "it is absent from changed_fields on all 32 over-cap writes."

### eval/harness/tests/unit/test_conflict_resolution_validator.py

- **Line 191-227** [test coverage]: `_replay` only processes `diff.conflicts.modified`, not `diff.conflicts.added`. If a future run log creates a new conflict, the replay would silently miss it — and V6's create-and-resolve bypass (the exact thing `test_v6_creating_an_already_resolved_conflict_counts` pins) would be invisible to the corpus replay. Not a current failure (the committed corpus has zero `added` entries), but the gap is precisely at the code path the validator was designed to catch. **Fix:** Either add `added` processing to `_replay`, or add a comment documenting the limitation so the next person adding a run log with created conflicts knows to extend it.

- [test coverage]: No test for V6 with zero conflicts modified. The most common real-world case (a run that skips conflict analysis entirely) has no explicit coverage. `len(touched) <= 1` at `len(touched) == 0` is untested. **Fix:** Add `test_v6_zero_conflicts_modified_passes` with identical before/after states.

- [test coverage]: No boundary tests for V2 at exactly the band thresholds. At exactly 210 words weighing and 300 words rationale, the `>` comparison means the validator should NOT fire — but an accidental change to `>=` would be undetected. **Fix:** Add `test_v2_weighing_at_exactly_the_band_is_not_observed` (210 words) and `test_v2_rationale_at_exactly_the_band_is_not_observed` (300 words).

## Nits (consider)

### eval/harness/tests/unit/test_conflict_resolution_validator.py

- No V2 skip test for `None` research_json. V6 has `test_v6_skips_without_research_json` but V2 doesn't, even though both have the same guard. A typo changing `or` to `and` would break V2 silently.

- No test exercising both weighing AND rationale over their caps on the same conflict. The validator loops over both fields independently and appends to `observations`, but an accidental `elif` would go undetected.

### docs/task-lifecycle.md

- **Line 366** [clarity]: Dense sentence — "the same call against a team CODEOWNERS *does* still mandate returns HTTP 200" requires multiple reads to parse. Consider: "the same call against a team still mandated by CODEOWNERS returns HTTP 200 and does nothing."

## What looks good

- **The V6/V2 population split** is well-reasoned. Two helpers that look similar but serve genuinely different populations, each with their own docstring explaining why.
- **The `competing_assertion_ids` after-state regression test** is a standout — the trap is documented in code, in the test, and in the PR body, and the test concretely exercises the exact case.
- **The corpus replay tests** pin exact firing sets against real data, catching shape slips between the validator's dict access patterns and the run logs' actual structure.
- **The senior-queue deletion is thorough.** All references removed or rewritten. 920 lines distilled to 43 that check the two surviving invariants.

## Verdict

**Needs minor work** — the misleading comment and the `_replay` gap are worth addressing before merge; the boundary tests are worth adding but not blocking.

francis-2008-happy:
All seven in (`bf51c4c8`) — the four warnings and all three nits. Each new test mutation-checked against the specific thing it guards.

## The two warnings that mattered

**The inverted comment.** You're right that "absent from changed_fields on 0 of the 32" reads as *present on all 32*. Fixed in the validator and in its twin in the test docstring, using your first phrasing.

**The `_replay` gap.** This was the sharpest finding, and it's worse than "would silently miss it": `if not modified: continue` skipped the whole run, so a created-then-resolved conflict would never have reached the validator at all — the exact bypass `test_v6_creating_an_already_resolved_conflict_counts` exists to pin.

I checked the corpus shape before writing anything: **0 `added` entries across 62 runs in 5 run logs**, but the key is present in the diff shape. And the two carry *different* shapes — `harness/diff.py:_diff_array` emits `added` as the whole after-state object, while `modified` is `{id, changed_fields}`. Treating them alike would have produced dead code, so `added` is appended rather than overlaid, guarded on id.

## The boundary tests, each proven to fire

| Mutation | Fails |
|---|---|
| `len(touched) <= 1` → `== 1` | `test_v6_zero_conflicts_modified_passes`, alone |
| weighing `>` → `>=` | `test_v2_weighing_at_exactly_the_band_is_not_observed` |
| rationale `>` → `>=` | `test_v2_rationale_at_exactly_the_band_is_not_observed` |

## The nits, and one that turned out to be sharper than filed

**Your `None` skip-test nit was right for a reason you didn't state.** I first wrote it the obvious way — `None` for both sides, mirroring V6's existing test — and then ran the `or`→`and` mutation you predicted. **The whole suite stayed green.** Both-None satisfies either operator, so that shape structurally cannot catch the typo.

Split per side, and it now fails on that mutation. I applied the same split to **V6's pre-existing twin**, which had the identical weakness and was equally unable to catch it.

**The both-fields-over-cap test found a bug in my own work.** It failed on first run reporting only `resolution_rationale`. Cause: I wrote `evidence_weighing`; the field is `weighing_analysis`. Which also meant my *weighing boundary test* had been passing **vacuously** against a field the validator never reads. Both fixed — recorded because "the test passed" was the wrong signal twice in one sitting.

`task-lifecycle.md:366` reworded to your version.

## Verification

`eval/harness` pytest **3051 passed / 2 skipped** on the merged head, plus 9 failures I confirmed are all `test_setup_feedback_case.py` failing on a missing `unzip` binary — environmental, unrelated to this diff, and reproducible on other branches here. The conflict-resolution file itself: **18 passed** (was 13).

Merged `main` (was 94 behind), 0 behind now.

Thank you for the review — the `_replay` gap and the `None`-both-sides nit were both cases where the test existed and proved less than it looked like it proved, which is the failure mode I keep having to be caught on.

yinkid28 (approved):
Re-review: all seven addressed.

  Verified each fix against the current head (26f7ddd1):
  #: 1
  Original finding: Misleading "absent from changed_fields on 0 of the 32" comment
  Status: Fixed — now reads "it appears in changed_fields on 0 of the 32" in the validator; test file uses its own clear docstring
  ────────────────────────────────────────
  #: 2
  Original finding: _replay gap: added entries silently skipped
  Status: Fixed — added processing with correct shape handling (whole after-state appended, not overlaid via changed_fields)
  ────────────────────────────────────────
  #: 3
  Original finding: No test for V6 with zero conflicts
  Status: test_v6_zero_conflicts_modified_passes added
  ────────────────────────────────────────
  #: 4
  Original finding: No boundary tests at exact band thresholds
  Status: test_v2_weighing_at_exactly_the_band_is_not_observed and test_v2_rationale_at_exactly_the_band_is_not_observed added
  ────────────────────────────────────────
  #: 5
  Original finding: No V2 skip test for None research_json
  Status: Added, and split per side — the both-None weakness that structurally can't distinguish or from and was caught and applied
  to
     V6's pre-existing twin too
  ────────────────────────────────────────
  #: 6
  Original finding: No test with both fields over cap on one conflict
  Status: test_v2_both_fields_over_cap_are_both_observed added — and it caught a misspelled field name that was passing vacuously
  ────────────────────────────────────────
  #: 7
  Original finding: Dense sentence in task-lifecycle.md:366
  Status: Reworded with bold for clarity
  Test count 13 -> 18. CI all green. No scope creep (still 3 files). The _replay fix and the split-None tests are both stronger than
  what I asked for.

clack391 (changes requested):
## Review: changes requested

The work here is good, and I want to be clear up front that the central problem is not a mistake you made. I re-derived every figure in the diff against the pre-merge commit `26f7ddd1` and reproduced your originals exactly (37 writes, 13/7/4/2, 17/12, 32/37). Your measurements were right when you took them. What broke is that #2149 merged on 2026-09-02 and replaced the corpus underneath them.

Verified as correct, by running rather than reading: both mutation-table claims reproduce exactly; the `after_state` read of `competing_assertion_ids` is right and genuinely regression-tested; `_ANALYSIS_FIELDS` is exactly the resolution-bearing subset of the schema's conflict object; V6 enforces `SKILL.md:438-449` verbatim; "no paid eval run" holds on both halves (I positive-controlled the `check_runlogs` regex so the clean result is not vacuous); `encoding="utf-8"` is satisfied; `harness-lint` is clean; there is no merge clobber; and merging will not auto-close #1972.

### Blocking

**1. CI `pytest` is red, on this PR's own acceptance check.** Reproduced locally: `2 failed, 16 passed`. #2149 (`e874661d7`) deleted `v1_2026-08-18_15-37-43`, one of the two logs the V6 test pins, and `v1_2026-08-18_13-28-10`, and added `v1_2026-09-01_18-38-25`. The corpus went from 5 logs to 4 and from 37 writes to 29.

Re-derived on the current corpus, independently and then confirmed by a second pass:

| | comment says | corpus now |
|---|---|---|
| logs / writes | 5 logs, 37 writes | 4 logs, 29 writes |
| weighing >200 / >210 / >220 / >240 | 13 / 7 / 4 / 2 | 14 / 7 / 3 / 1 |
| rationale <3-way >250 / >300 | 17 / 12 | 14 / 10 |
| V6 multi-conflict turns | 2 | 1 |
| weighing max | 251 | 251 (unchanged) |

Only `corpus MAXIMUM 251`, `>210: 7`, and the `<3 competing` denominator of 17 survive.

**2. The design that caused it will cause it again, in someone else's PR.** The replay tests pin literal counts and a literal filename against a corpus that `versioning.DEFAULT_KEEP_CANDIDATES = 5` rotates on every paid run. Measured from a green baseline: adding a non-violating run log breaks `test_v2_reports_the_measured_counts_on_the_corpus` unconditionally; adding a violating one breaks the V6 test too. So the next person to land a conflict-resolution run log inherits a red `pytest` in a file they never touched.

Worth noting this is a known shape here. #1972 itself says "re-derive any count below against the post-merge five-log corpus rather than quoting one", and `validators/test_search_records.py:205-224` documents the identical staleness and deliberately chose prose plus a re-derivation recipe over a pinned count.

I tested a fix rather than just asking for one: replace both literals with an assertion that the validator agrees with a second derivation read straight from raw `changed_fields`. It gives 18 passed, stays green when a run log is added, and still goes red when I inject a shape slip (`after.get("conflicts")` to `after.get("conflict_entries")`). That keeps the shape-slip guard you built it for and drops the rotation fragility. Happy to hand over the exact diff if useful.

**3. Eleven present-tense corpus figures are stale.** They are the stated justification for the live constants, so they need to move with the fix: the constant block at `:355-372`, the docstring at `:380`, the trap comment at `:411`, the "2 of 5 committed runs" at `:338`, and the module docstring plus `:213` in the test file. The band choice itself survives, and is if anything better supported now: 240 would report 1 of 14 on today's corpus.

**4. `_ANALYSIS_FIELDS`, which is what V6's rule consists of, has no per-field coverage.** I dropped each of the four field names in turn and the suite stayed at baseline every time. Every synthetic V6 test writes `status` and `resolution_rationale` together and none writes the other three at all, so four of the five are asserted by nothing. A repaired corpus replay would not catch it either, since the one corpus violation writes all five together. Someone could silently drop `preferred_assertion_id` from a gating rule and nothing would go red.

Tested fix: one parametrised test per field, a conflict whose only change is that field must count. With it in place, dropping any of the five now fails.

**5. The two "names both conflicts" assertions are vacuous.** I changed V6's message to name no ids at all and both tests stayed green. `"c_001" in str(e.value)` is satisfied by pytest's rewritten explanation, which embeds a repr of `touched` containing the id.

One warning, because I got this wrong first: asserting on `e.value.args[0]` instead does **not** fix it. I tried that, ran the mutation, and it survived. pytest collects `validators/test_conflict_resolution.py` under `python_files = ["test_*.py"]`, so the module is rewritten and the explanation is concatenated into `args[0]` as well.

What does work, tested: import the validator the way `harness/validator_runner.py:145-170` does, via `importlib.util.spec_from_file_location`, which bypasses the rewrite hook. Baseline is preserved, and under the mutation both naming tests now fail. It also removes the divergence outright, since the tests then grade the exact string the harness emits, and it makes your `splitlines` rewrite-noise filter unnecessary.

### One ordering constraint, please read before starting

You may be tempted by the reuse below, and it is a real improvement, but **do item 4 first**.

`harness/skill_invocation.py:265-270` already defines `CONFLICT_ANALYSIS_FIELDS`, and `:274` `_is_conflict_resolution_product` is substantially the same predicate as `_analysis_written`, with the `isinstance(dict)` guard yours lacks. Validators already import from `harness.*`, so the path exists.

But the existing constant has **four** fields, not five: it has no `status`, and #1972's V6 rule names `status` explicitly. I measured this: swapping it in makes two status-only resolutions pass, and the suite stays at baseline while it happens. The safe form is `(*CONFLICT_ANALYSIS_FIELDS, "status")`, which preserves behaviour exactly. With item 4's per-field tests in place, the unsafe form correctly fails, which is why the order matters.

### Your open question, answered

You asked whether V6 should gate at all. **Keep it gating. Do not make the one-word change.**

#1972's sequencing table rules it `gating test_*`, and `SKILL.md:438-449` states the rule verbatim in the skill's own voice: "do the full independence/weighing/resolution work on that one conflict only, leaving the others' fields untouched this turn". The measured false-positive cost is 1 firing in 51 corpus runs.

The one thing worth adding is a deliberate test for the `moot` shape. `SKILL.md:315` tells the skill to set `status: "moot"` when other evidence makes a conflict irrelevant, and `:447` tells it to leave the others' fields untouched that turn; in a single turn those can conflict. V6 currently sides with `:447`, and there are zero `moot` writes in the corpus to have surfaced it. Pin it with a test so the behaviour is chosen rather than inherited. Whether the doctrine itself should change is a genealogist's call and belongs in a comment on #1972, not in this PR.

### Smaller things

- `orchestrator.py:594` is now `:600` (the merge shifted it by six). `SKILL.md:145-152` is really 148-152, and `validator_runner.py:264-274` is 264-276.
- `report_resolution_word_caps` is the only one of the repo's nine `report_*` validators that takes no `test` parameter and has no negative-test guard. Low impact, since a correctly-declining run writes no conflict prose, but the other eight all have both.
- The PR body needs a pass. It still carries the inverted "absent from `changed_fields` on **0 of the 32**" that your `97eaaaf85` already fixed in the code; "13 pass" is 18 now; "#1629 and PR #1931 both open" is half stale, since #1931 merged on 2026-09-01 (the deferral still holds, #1629 is genuinely open); the `make harness-test` line is no longer accurate in either half; and the Summary enumerates the touched files without mentioning `docs/task-lifecycle.md`. Also worth knowing `make test-all` reaches `harness-test`, so it fails at this head too.

### Not asking you to change, recorded for the record

The "worst case 460 words" that the PR body cites is a case V2 deliberately exempts: `ut_conflict_resolution_001` / `c_001` is scenario `flynn-with-birthplace-conflict`, `conflict_type: "fact"`, `competing_assertion_ids: ["a_002", "a_009", "a_012"]`, so three, and the escape applies. That error did not start with you. The deep dive names it as the highest surviving instance of the single-assertion identity class, #1972 says "Cite that one", and the PR body inherited it. Its other named example checks out, so it is one misattribution rather than a broken finding. I will raise it on #1972 so it gets corrected at the source.

Two other things I looked at and am explicitly not raising: the create-with-analysis branch is a faithful reading of #1972, which offers "or a before/after diff of `conflicts[]`" and exempts creation because identification is not resolution, so a create arriving fully resolved is not identification; and the `report_` prefix being the only thing separating tier-2 from gating is a pre-existing gap across all nine `report_*` validators, not yours to close alone.

Nothing here needs a re-run: `eval/harness/validators/**` is not in the snapshot, and `check_runlogs`' path regexes match none of the three changed files. I confirmed both, and positive-controlled the regex against paths that should match.

francis-2008-happy:
All five blocking items in (`0ab3ec1e`), plus the `moot` test and the line-number corrections. Thank you for testing the fixes rather than only specifying them — and for the ordering constraint on item 4, which turned out to matter exactly as you said.

## 1 + 2 — the rotation, and the design that caused it

Both replay tests now assert the validator **agrees with an independent second derivation** read straight from raw `changed_fields`, rather than pinning filenames and counts. Verified both halves of your claim:

| | result |
|---|---|
| add a non-violating run log | 30 passed (was: broke unconditionally) |
| add a violating run log | 30 passed (was: broke the V6 test too) |
| shape slip `after.get("conflicts")` → `"conflict_entries"` | **13 failed** |

So the rotation fragility is gone and the shape-slip guard survives, which was the whole point of the pinned form.

The second derivations deliberately don't share the validator's helpers *or* its field constant — sharing the thing most likely to be wrong would make agreement vacuous. I restated the field names with a comment saying why.

**One thing worth recording against myself:** my first draft of the V2 derivation read `diff["modified"]` instead of `diff["conflicts"]["modified"]` and returned `0/0`. That is precisely the shape slip these tests exist to catch, committed in the test helper itself. It failed loudly, which is the system working, but it is a good argument for your "second derivation" framing over anything cleverer.

## 3 — figures re-derived, not copied

I re-derived rather than transcribing your table, and **my numbers match yours exactly**: 4 logs, 29 analysis writes, 51 runs; weighing >200/>210/>220/>240 = **14/7/3/1**, max **251** unchanged; rationale **14 of 17** over cap (>300: 10); **1** multi-conflict turn.

Your observation that the 240 argument got *stronger* is now in the comment — it would report 1 of 14, so a green arm reading as "weighing lengths are fine" is a sharper risk than before.

## 4 — per-field coverage, and why your ordering was right

You were right that four of the five fields were asserted by nothing. Dropping each in turn, before the fix: baseline every time.

Two parametrised tests now cover each field alone, positive and negative. Dropping any one field fails the parametrisation — verified for all five.

**Then** the reuse: `_ANALYSIS_FIELDS = (*CONFLICT_ANALYSIS_FIELDS, "status")`. And the unsafe form does exactly what you predicted — swapping in the bare constant now **fails the `status` per-field test** instead of silently letting two status-only resolutions pass. Doing the reuse first would have shipped that hole.

## 5 — the vacuous naming assertions

Confirmed and fixed the way you specified: loaded via `spec_from_file_location` the way `validator_runner` does, bypassing the rewrite hook. Under the name-no-ids mutation, **both naming tests now fail** (4 failed vs baseline 2 at the time).

Your warning about `e.value.args[0]` saved me a wrong turn — that was going to be my first attempt.

## The `moot` shape

Pinned, with a control so it can't pass on the validator firing at all. V6 siding with one-conflict-per-turn is now chosen rather than inherited. I've left the doctrine question for #1972 as you said, and haven't commented there yet.

## Smaller things

Line citations corrected and **each verified against the file** rather than taken on trust: `orchestrator.py` 594 → **600** (the judge-suppression check moved), `SKILL.md` 145 → **148-152**, `validator_runner.py` 264-274 → **264-276**.

**V6 stays gating**, per your answer. I'm not making the one-word change.

`report_resolution_word_caps` still takes no `test` parameter — noted, and I've left it, since you called it low impact and adding a negative-test guard to one of nine `report_*` validators seems worse than the consistent gap.

## PR body

Rewritten: the inverted "0 of the 32" is gone, counts corrected, `#1931` marked merged with the deferral intact, the `make harness-test` line fixed, and `docs/task-lifecycle.md` added to the Summary's file list.

## Verification

Harness suite **3152 passed**. The two `test_mock_mcp` failures I hit mid-merge were a **stale engine build**, not main breaking — they pass after `npm run build`, which is worth knowing since they look alarming on a fresh worktree.

Branch is level with `main`, and I reconciled a remote merge rather than force-pushing over it.

clack391 (commented):
## Re-review: round 1 is fixed, and three things remain

I reproduced every claim in your fix report rather than reading it, using shapes you did not use. All of it holds:

| | round 1 | now |
|---|---|---|
| add a non-violating run log | 1 failed | **30 passed** |
| add a violating run log | 2 failed | **30 passed** |
| shape slip `conflicts` to `conflict_entries` | red | **14 failed** (still caught) |
| drop each of the 4 fields from `CONFLICT_ANALYSIS_FIELDS` | baseline every time | **1 failed each** |
| bare `CONFLICT_ANALYSIS_FIELDS` (drops `status`) | passed silently | **2 failed** |
| V6 message names no ids | both tests green | **both fail** |

Line citations opened rather than trusted: `orchestrator.py:600`, `SKILL.md:148`/`:152`, `validator_runner.py:276`. All three correct. Sequencing the reuse after the per-field tests was the right call and it shows. The positive controls you added unprompted are stronger than what I asked for.

**Before the three items: most of what follows is my fault, not yours.** The agreement-instead-of-literals design was my prescription, handed over on the strength of a break-test I ran once. You implemented it faithfully. The failure modes below are in the design, and I should have given you the failing case and let you design the cure rather than shipping you a shape I had not attacked properly.

### 1. A permitted turn reddens the V6 corpus test

`test_conflict_resolution_validator.py:337-340`. The second derivation counts an `added` entry with `any(a.get(f) for f in _SECOND_DERIVATION_ANALYSIS_FIELDS)`. `"status"` is in that tuple and the creation template writes `"status": "unresolved"`, which is truthy. So every created conflict counts as carrying analysis, while the validator's `_analysis_written` correctly excludes `(None, "unresolved")`.

A run that resolves one conflict and identifies a new unresolved one, which V6's own message calls permitted, now disagrees:

```
AssertionError: validator fired on [('v1_2026-08-18_19-42-11.json', 'ut_conflict_resolution_002')]
but a second derivation from raw changed_fields says
[('v1_2026-08-18_19-42-11.json', ...), ('v1_2026-09-08_00-00-00.json', 'ut_synth')]
```

This is the fragility we were removing, narrowed rather than closed: from "any new run log" to "any new run log carrying an `added` conflict", which is the arm the `_replay` docstring says the replay exists to cover. The docstring at `:323-324` is also wrong about its own code: it says a created conflict counts "when it arrives carrying analysis", and it counts all of them.

### 2. The V2 derivation is blind to `added` while `_replay` replays them

`test_conflict_resolution_validator.py:371`. The loop is `for m in (cdiff.get("modified") or [])` with no `added` arm, but `_replay` appends `added` entries to the after state and `_conflicts_written` includes new entries. So V2 observes a created conflict's prose and the derivation does not. A synthesized log creating a conflict with a 400-word rationale:

```
AssertionError: validator observed 7 weighing / 11 rationale over-cap writes;
a second derivation from raw changed_fields says 7 / 10
```

Same mechanism as item 1, other validator. `:348` says it "Mirrors V2's escape", and it does mirror the escape, but not the population.

### 3. The before-state read has no coverage at all, on the gating validator

`validators/test_conflict_resolution.py:268-269`. Making `_conflicts_by_id` return `{}` unconditionally leaves the suite at **30 passed**. So does breaking only the before-side of the `conflicts` to `conflict_entries` slip. The after-side half fails 14 tests; the before-side half fails nothing.

It is not theoretical. I walked the fixtures: **32 of 41 committed scenario conflicts ship `status: "resolved"` with all four analysis fields populated**, and one of those scenarios, `flynn-competing-fathers`, is in this skill's own corpus. With a broken before-reader every one of those looks new, `_analysis_written` returns True, and V6 fires on a run that touched exactly one conflict. V6 is gating, so that fails the run and suppresses the judge at `orchestrator.py:600`.

The replay structurally cannot cover it: `_replay` skips runs with no conflicts diff, and every scenario it does reach starts with zero analysed conflicts. No synthetic test puts analysis in a before state either, since every `_states(...)` call uses template defaults.

### On the fixes

I have tested patches for items 1 and 3 and can post them if useful, but I would rather you write your own. Two of my five round-1 prescriptions were defective, and the design that produced items 1 and 2 was also mine, so my patches on this file have earned scepticism rather than trust. If you do take them, break-test them with shapes I did not use. That is exactly how item 1 got past me: my round-1 rotation test copied existing run logs, which carry `modified` entries and zero `added` ones, so I tested the shape the corpus had rather than the shape it lacked.

For item 2 I have no patch, only the reproduction.

### Deliberately not raised

I found several smaller things and am dropping them rather than spending your time: two stale figures in the test file, one of which this commit introduced; the fact that adding a field to `CONFLICT_ANALYSIS_FIELDS` is uncaught where removing one is caught; and the `moot` test taking the modify branch rather than the create branch. None changes behaviour today. Fold them in if you are touching those lines anyway, and otherwise ignore them.

One non-code thing worth a look: the PR body still reads as it did before your last push, so the six corrections your comment describes do not appear in it. Most likely drafted and not saved.

francis-2008-happy:
All three in (`e0d59e1c`). And thank you for the note about the design being yours — but the two that mattered were mine to catch and I didn't: I wrote both derivations, and I break-tested them against the corpus rather than against the shapes the corpus lacks. That's the same mistake in the same file twice.

## 1 — `"unresolved"` is truthy

Confirmed. The derivation used `any(a.get(f) for f in FIELDS)` and the creation template writes `"status": "unresolved"`, so every created conflict read as a resolution while `_analysis_written` correctly excludes `(None, "unresolved")`.

Now applies the **rule** rather than truthiness, in `_second_derivation_carries_analysis` — restated rather than imported, for the same reason the field tuple is restated: importing the thing most likely to be wrong makes agreement vacuous. The docstring no longer claims "when it arrives carrying analysis" while counting all of them.

Proven necessary: reverting to truthiness, against your exact shape (resolve one + identify a new unresolved one), reddens the V6 test alone.

## 2 — the population, not just the escape

Also confirmed. `_replay` appends `added` to the after state and `_conflicts_written` is deliberately wider than V6's, so the validator saw a created conflict's over-cap prose and my derivation didn't. Added the `added` arm and corrected the docstring, which claimed to mirror the escape and did — but not the population.

Proven necessary: reverting it, against a created conflict carrying a 400-word rationale, reddens the V2 test alone.

## 3 — the before-state read, and it's the one that worried me

Both your mutations reproduced at baseline: `_conflicts_by_id` returning `{}`, and the before-side-only slip. Neither failed anything.

**I re-derived the blast radius rather than quoting yours, and it lands on the same figures:** 32 of the 41 committed scenario conflicts ship `status: "resolved"` with all four analysis fields populated, and `flynn-competing-fathers` — one of them — is in this skill's own corpus. So a broken before-reader makes each look newly created-and-resolved, V6 fires on a run that touched exactly one conflict, and because V6 gates it fails the run and suppresses the judge at `orchestrator.py:600`. A silent false-fail on a gating validator is worse than the hole I was sent to close.

Two tests now, with a control so the passing one can't pass by V6 never firing on an analysed before state. Both your mutations now fail.

## Why none of the three could have been caught by the replay

Worth stating, because it is the lesson rather than the fix: the replay skips runs with no conflicts diff, every scenario it reaches starts with **zero** analysed conflicts, and the committed corpus carries **no `added` entries at all**. So the corpus cannot exercise the created-conflict arm or the analysed-before-state arm — and my round-1 rotation test, built by copying existing run logs, tested the shape the corpus had rather than the shape it lacked. That is precisely how item 1 got past both of us.

## Your three dropped items

Taking two, since I was in those lines anyway: the stale figures are corrected, and I have not touched the `moot` create-branch point or the add-a-field asymmetry — the latter feels like it wants a different mechanism than a per-field list, and I would rather not invent one here.

## Verification

Harness suite **3154 passed**; the 9 `test_setup_feedback_case` failures are the missing-`unzip` environmental ones. Branch level with `main` after reconciling two remote merges rather than force-pushing over them.

clack391 (approved):
## Approving

All three items from the last round are fixed, and I verified them behaviourally with my own reproductions rather than reading the report: the resolve-one-plus-identify-one shape is green, the created-conflict-with-a-400-word-rationale shape is green, and both before-state mutations now fail as they should. I also differential-tested `_second_derivation_carries_analysis` against the validator's `_analysis_written` across 371,293 input combinations and found zero disagreements, which is the thing that would have been easiest to get subtly wrong in a restatement.

Worth saying plainly what stood out across the three rounds, because it is unusual.

You reported your own errors against yourself twice, unprompted. The misspelled field name that made a boundary test pass vacuously, and then "the two that mattered were mine to catch and I didn't". Neither was visible to me. Volunteering them cost you something and made the review better.

You re-derived rather than transcribed. Every round you re-measured instead of copying my table, and landed on the same figures independently, including the 32-of-41 blast radius. That is the difference between a number a reviewer can rely on and one that is just being passed along.

You proved each fix necessary rather than asserting it. Reverting every change against the specific shape that exposes it, and stating which test dies, is what let me verify three rounds of work quickly instead of re-deriving your reasoning each time.

You caught a wrong turn of mine before taking it. The `e.value.args[0]` suggestion would not have worked, and you tested it rather than trusting it. The design behind two of the last round's findings was also mine, handed to you under-specified; you implemented it faithfully and then found the parts I had not thought through.

And the ordering constraint on the reuse was load-bearing. Doing the per-field tests first is what made the unsafe form fail loudly instead of shipping a silent hole, and you sequenced it correctly without being told twice.

Approving. Nice work.

Issue #1956, opened 2026-08-27
The judge never sees conflicts[], so any rubric rule that consults a conflict is graded on the skill's own testimony


florencemashipei:
The LLM judge is never shown `research.json`'s `conflicts[]`, so any rubric rule that asks it to consult a conflict is graded on whatever the skill under test asserted about that conflict. Found while reviewing #1954, which adds exactly such a rule and has a run log demonstrating the failure.

## The demonstration

`eval/runlogs/unit/search-external-sites/v1_2026-08-27_00-08-56.json`, `ut_search_external_sites_002`, scenario `mid-research-flynn` (which carries `c_001`: `disputed_attribute: "birthplace"`, `status: "resolved"`, `preferred_assertion_id: a_002` → Ireland; Pennsylvania rejected):

The skill emitted:

> "No conflicts are on file for Patrick's birth year or place, so I can encode `birth_year=1845` and `birth_place=Pennsylvania` safely."

and generated `https://www.myheritage.com/research?action=query&first=Patrick&last=Flynn&birth_year=1845&birth_place=Pennsylvania`.

The judge, grading against a rubric dimension that explicitly says *"check `conflicts[]` in the scenario's `research.json`"*, scored **URL generation 3** with:

> "No conflicts exist on file for Patrick's birth year or place, so encoding `birth_year=1845` and `birth_place=Pennsylvania` is correct and unsupported-claim-free."

All eight dimensions scored 3; the test passed. It was not in that run's review sample, so no annotator saw it.

The inverse appears on the sibling run `v1_2026-08-26_23-13-30`: the same test encoded `Ireland`, said so citing `c_001` and `preferred_assertion_id: a_002`, and the judge "confirmed" it. Both gradings are the judge repeating the skill back to itself.

## Why it happens

`render_prompt_parts` (`eval/harness/harness/judge.py:255-315`) fills `{rubric} {judge_context} {before_state} {scenario_readme} {user_message} {skills_invoked} {text_response} {file_changes_summary} {tool_calls} {validator_failures}`. None of them carries `conflicts[]`:

- **`_summarize_before_state`** (`orchestrator.py:1494`) renders **only `sources`**, from research.json and tree.gedcomx.json. It was written for the fabricated-citation misgrade and never extended past that section.
- **`_summarize_changes`** (`orchestrator.py:1363`) renders added/modified/deleted entries only. No skill in this family writes `conflicts`, so it never appears there.
[...]

PR #2093, opened 2026-09-01
judge: render every tool result and before-state conflicts[] to the judge (#1902)

### LINKED ISSUE #1902: The judge's rendered evidence is incomplete: tool responses truncate at 3 items, and before-state never renders conflicts[] — grounding graded on the skill's own testimony

Asimi1234:
## Summary

- The judge's rendered evidence was incomplete on two paths, so a **grounded** skill claim could be graded as fabricated. Half A: tool `response_summary` sampled every result list down to 3 items, so a correct citation of result 4+ (e.g. Alabama collection `1986396`) read as invented. Half B: the before-state block never rendered `conflicts[]`, so the judge could only repeat a skill's "no conflict on file" claim back to itself.
- Fix (one decision — *what evidence does the judge see?* — applied to both paths): render every tool result (`array_sample=None` at the single call site, bounded by `_TOOL_CALLS_MAX_CHARS`), and render before-state conflicts with each conflict's preferred/competing assertions **resolved to their values** so a place-name claim is checkable against the record.
- Closes #1902 (which absorbed #1956). Tier: Normal.

## Review guidance

**Start here:** `orchestrator.py` `_summarize_before_state_conflicts` — the design call is that a conflict's assertion *ids* are useless to the judge (it compares a URL place-name against a value), so each `preferred_assertion_id` / `competing_assertion_ids` entry is resolved through `assertions[]` to `{id, fact_type, value, structured_value, place, date}`. Check that resolution + the dangling-ref (`_unresolved`) fallback. The conflicts block goes through the same `{count, all_ids, detail}` shape as a sources block — complete ids never clipped, heavy detail trimmed under the shared `_BEFORE_STATE_MAX_CHARS` budget — but is rendered **before** the source blocks so conflict values win the budget over source citations. Half A is a one-line `array_sample=None`.

**Unsure about:** Nothing. `_RESPONSE_ARRAY_SAMPLE=3` is deliberately untouched (still the default, still exercised by the before-state per-source detail path); only the tool-response call site opts out.

## Plan

**Didn't change:**
- The `_RESPONSE_ARRAY_SAMPLE` default and the e2e run-log `_summarize_tool_response` (independent impl) — left alone on purpose.
- **The e2e judge was considered and is unaffected.** Its `conflicts_addressed` dimension grades *"does the narrative resolve them with reasoning?"* against `proof_summaries` — which it is shown — not the structured `conflicts[]`, so it has no analogous blind spot. Its tool-response summary passes small responses through verbatim, so the Half A truncation cannot reproduce there either. This is the unit judge only, by design.
- The `#1933` wording gaps (`moot`, identity `disputed_attribute: null`) are made *checkable* here but not fixed; the code tolerates them without crashing.

**Acceptance check:** `test_render_prompt_shows_fourth_collection_id` and `test_before_state_renders_conflict_values_end_to_end` both fail on `main` (the 4th collection id / the conflict values never reach the prompt) and pass on this branch — proven fails-first by stashing the source edits. `test_conflict_values_win_the_budget_over_source_detail` pins the ordering: I confirmed it goes red if the conflicts block renders after the sources instead of before.

**Deviated from the plan:** Nothing (drift-critic: no drift). Two refinements after the plan: the spec "one decision line" became a short paragraph (its `#1902` cite dropped to satisfy the doc-links issue-ref ratchet — state the fact, not the ticket), and conflicts render before sources in the budget order so their resolved values are not the first thing a tight budget drops.

## Test plan

- [x] I ran `make test-all` and it passed (3001 harness + all JS/packaging suites; "All checks passed").
- No skill run-log snapshot changed — `eval/harness/**` is outside `snapshot.py`'s snapshot, so **no paid eval run is needed** to verify this and none was run.
- No skill `description`/DO-NOT frontmatter changed.

This carries `nothing-checks`: no check distinguished "the judge saw the whole response" from "the judge saw three of four." The prove-fails-first tests added here are that missing guard.

## Follow-on issues

**Folded in / filed instead:** Nothing filed. #1933 (four `search-external-sites` wording gaps) and #1000 (restore `response` into committed run logs) remain open and out of scope — named in the issue, neither closed by this.


yinkid28 (commented):
# Peer Review: 1902-judge-evidence-completeness

**Base:** main | **Commits:** 3 | **Files changed:** 12
**PR:** #2093 — judge: render every tool result and before-state conflicts[] to the judge (#1902)

## Summary

Two fixes to the judge's evidence rendering, addressing a single question — *what evidence does the judge see?* Half A: the tool-response path now renders every result (`array_sample=None`) instead of sampling the first 3, fixing the Alabama `collections_search` misgrade where result 4+ was invisible. Half B: the before-state block now renders `conflicts[]` with resolved assertion values, so a "no conflict on file" claim is checkable. The PR also deletes the senior-queue workflow (shared with PR #2090) and replaces its tests with a simpler CODEOWNERS invariant check. Well-structured, with prove-fails-first tests for both halves and a budget-ordering test pinning that conflict values survive prompt-size pressure.

## Warnings (should fix)

### eval/harness/harness/orchestrator.py
- **Line 1898** [conventions]: The shared detail-rendering loop uses `"per-source detail (heavy fields truncated)"` and `"per-source detail omitted for prompt size"` as labels for ALL `labelled` entries — including conflicts. When the judge sees a conflicts block, it reads "per-source detail" which is misleading. The judge is comparing place-name values against conflict assertions, and seeing them labeled as "sources" could confuse grounding. The same mislabeling appears in the omission note at line 1913. **Fix:** Use a generic label like `"per-entry detail"` or derive the label from the block type.

### eval/harness/tests/unit/test_before_state.py
- **~line 190-362** [test coverage]: No test for a conflict with an empty `competing_assertion_ids` list (`[]`). The code handles it correctly via `c.get("competing_assertion_ids") or []`, but this is a valid shape (a newly-created conflict with only a preferred assertion). **Fix:** Add a small test case with `"competing_assertion_ids": []`.

- **~line 190-362** [test coverage]: No test exercising `_CONFLICT_HEAVY_FIELDS` string truncation. The `_FLYNN_CONFLICT` fixture includes `weighing_analysis` and `resolution_rationale` but these are short (100 chars). No test verifies that a value exceeding `_BEFORE_STATE_STRING_MAX` (4000 chars) gets truncated through the `_summarize_response(c[k], string_max=...)` path. **Fix:** Add a test with a heavy-field value > 4000 chars and assert truncation.

## Nits (consider)

### eval/harness/harness/orchestrator.py
- **Line 1795** [conventions]: The `_summarize_before_state` docstring still says "Render the source entries that existed BEFORE the skill ran" without mentioning conflicts.

- **Line 1826-1841** [conventions]: The 15-line inline comment in the `if conflicts:` block explaining the budget ordering could be shortened to ~5 lines — the same information appears in the function docstring and in the test.

### eval/harness/harness/judge.py
- **Line 397-403** [conventions]: The 6-line inline comment explaining `array_sample=None` is longer than typical for this file (1-2 lines). Could be shortened to 2 lines with an issue reference.

### eval/harness/tests/unit/test_before_state.py
- **Line 1** [conventions]: Module docstring only mentions sources and `_RESPONSE_ARRAY_SAMPLE`; doesn't mention the conflicts rendering tests added here.

## What looks good

- The prove-fails-first methodology is well-executed: `test_render_prompt_shows_fourth_collection_id` and `test_before_state_renders_conflict_values_end_to_end` both fail on `main` and pass on the branch, pinning the exact bugs.
- `_resolve_assertion` handles dangling refs gracefully (`_unresolved: True` fallback) with matching test coverage, avoiding crashes from real-world data where assertion IDs can be missing.
- The budget-ordering test (`test_conflict_values_win_the_budget_over_source_detail`) is well-designed — it monkeypatches the budget to create a squeeze where only one block's detail fits, discriminating the ordering rather than just testing "something got dropped."
- The `{count, all_ids, detail}` shape reuse between conflicts and sources is clean — conflicts flow through the existing shared rendering loop with no special-casing needed.

## Verdict

**Ready to merge** with minor suggestions. The "per-source detail" label (the one warning) is worth fixing but not blocking — it's a misleading label in rendered output, not a logic error. The test gaps are small. The core changes are correct, well-tested, and close a real grading blind spot.

yinkid28 (approved):
# Peer Review: 1902-judge-evidence-completeness

**Base:** main | **Commits:** 3 | **Files changed:** 12

## Summary

Two fixes to the judge's evidence rendering — Half A renders every tool result (`array_sample=None`) instead of sampling the first 3, fixing the Alabama misgrade. Half B renders before-state `conflicts[]` with resolved assertion values so "no conflict on file" claims are checkable. Well-structured with prove-fails-first tests for both halves.

## Minor suggestions (non-blocking)

### eval/harness/harness/orchestrator.py
- **Line 1898**: The shared detail loop labels everything `"per-source detail (heavy fields truncated)"` / `"per-source detail omitted for prompt size"` — including conflicts. The judge sees "per-source" on a conflict block, which is misleading. Same at line 1913. Consider `"per-entry detail"` or deriving the label from the block type.
- **Line 1795**: `_summarize_before_state` docstring still says "Render the source entries" without mentioning conflicts.

### eval/harness/tests/unit/test_before_state.py
- No test for a conflict with empty `competing_assertion_ids: []` (valid shape, code handles it, just untested).
- No test exercising `_CONFLICT_HEAVY_FIELDS` string truncation (fixture values are 100 chars; no test with a value exceeding `_BEFORE_STATE_STRING_MAX`).

## What looks good

- Prove-fails-first tests pin the exact bugs: `test_render_prompt_shows_fourth_collection_id` and `test_before_state_renders_conflict_values_end_to_end` both fail on `main`.
- `_resolve_assertion` handles dangling refs gracefully with `_unresolved: True` fallback.
- Budget-ordering test discriminates the ordering by creating a squeeze where only one block's detail fits.
- `{count, all_ids, detail}` shape reuse between conflicts and sources is clean.

promise-emmanuel (changes requested):
Requesting changes on one thing. Everything else here is good and I have verified
it — see the notes at the end.

The conflicts block is rendered to the judge under a header that calls it
"per-source detail", and when the budget drops a conflict the note reads
"per-source detail omitted for prompt size: c_001". This is text the judge reads
while deciding whether a conflict was on file, and it names a conflict as a source.

orchestrator.py, line 1908.

You wrote:      header = f"{label} — per-source detail (heavy fields truncated):\n"
Change it to:   header = f"{label} — per-entry detail (heavy fields truncated):\n"

orchestrator.py, line 1923.

You wrote:      f"\n\n[per-source detail omitted for prompt size: "
Change it to:   f"\n\n[per-entry detail omitted for prompt size: "

That rename breaks four assertions in test_before_state.py — lines 137, 160, 164
and 165 — plus line 189. Change "per-source" to "per-entry" in each. Nothing else
in the repo depends on the string.

While you are in the file, the docstring on _summarize_before_state still says it
renders "the source entries that existed BEFORE the skill ran". Add conflicts to it.

And the spec addition covers Half A only. unit-test-spec.md's judge-prompt slot list
has no {before_state} entry at all, so Half B lands in a block the spec does not
describe. The gap predates you; this is the PR that makes it visible, so it is the
cheapest place to close it.

For the record, what I checked and what held.

I re-proved the fails-first claims rather than taking them on trust. With main's
judge.py restored, test_render_prompt_shows_fourth_collection_id and
test_render_prompt_renders_every_result_not_a_sample both fail on the missing 4th
result. With main's orchestrator.py, test_before_state.py cannot import
_summarize_before_state_conflicts. I moved the conflicts block to render after the
sources and watched test_conflict_values_win_the_budget_over_source_detail go red
on "Ireland" — that test genuinely pins the ordering.

Harness suite in the worktree: 3050 passed, 9 skipped, 1 failed. The failure is
test_mock_mcp::test_successful_call_carries_no_is_error, caused by a fresh worktree
with no engine build. Environmental, not yours.

I rendered the real before-state block for flynn-identity-geographic: 7,952 chars,
three conflicts with values resolved, reads correctly.

One headroom note, no action needed. array_sample=None makes the single case the
size guard cannot bound reachable with fewer calls — the drop loop stops at
len(calls) > 1, so one call over 50,000 chars is accepted as overage. Largest
fixture in eval/fixtures/mcp is 23,921 bytes, so nothing triggers it today.

Asimi1234:
Thanks Promise — all four addressed in 98b8e1a5, and thanks for re-proving the fails-first claims independently.

- **The label.** `orchestrator.py` 1908 and 1923 now read `per-entry detail` / `per-entry detail omitted for prompt size`. Renamed `per-source` → `per-entry` in the five `test_before_state.py` assertions (137, 160, 164, 165, 189), and while in there fixed the now-stale internal comment at 1883 too. `test_before_state.py`: 18 passed.
- **Docstring.** `_summarize_before_state` now says it renders "the sources and conflicts that existed BEFORE the skill ran."
- **Spec.** Added a `{before_state}` entry to the judge-prompt slot list in `unit-test-spec.md` with a short paragraph on the block — complete `count`/`all_ids` first (never clipped), heavy detail trimmed under `_BEFORE_STATE_MAX_CHARS`, conflicts first so their resolved values win the budget, added-this-run material lives under `{file_changes_summary}`. No `#NNN` cites added, so the doc-links issue-ref ratchet stays put (`doc-links.test.ts`: 142 passed).

On the headroom note — agreed, no action: `array_sample=None` makes the one-call-over-50,000-chars overage reachable in principle, but the largest `eval/fixtures/mcp` fixture is 23,921 bytes, so nothing hits it today. Left it as you found it.

promise-emmanuel (approved):
Approving. Re-reviewed at 98b8e1a5.

The delta since my last pass is exactly the three fixes and nothing else — three
files, +22/-10, no scope creep.

The label rename is complete in the rendered output. I re-rendered the real
before-state block for flynn-identity-geographic and the conflicts header now reads
"per-entry detail (heavy fields truncated)". grep for "per-source" across
eval/harness/ leaves one hit, orchestrator.py:1679, which is inside
_summarize_before_state_sources' own docstring where the word is correct. Good catch
on the stale comment at 1883 — I had not asked for that one.

I checked every claim in the new spec paragraph against the code rather than reading
it: complete count/all_ids first and never clipped, detail trimmed under
_BEFORE_STATE_MAX_CHARS, conflicts first so their resolved values win the budget,
dropped entries named in the omission note, "(none)" meaning no prior sources or
conflicts, added-this-run material under {file_changes_summary}. All true. No #NNN
cites, so the doc-links ratchet is untouched.

Harness suite in the worktree at 98b8e1a5: 3050 passed, 9 skipped, 1 failed. Same
environmental failure as last round — test_mock_mcp::test_successful_call_carries_no_is_error,
no engine build in a fresh worktree. Unchanged and not yours. CI is green on this head.

One note for whoever merges, not a request. yinkid28's approval is at 33d21e56, not
at this head. Stale approvals are not dismissed on push in this repo, so it still
counts toward the two required. That is fine here — the only thing that landed after
it is the three fixes I asked for, and I have read and verified all three. Flagging
it so it is on the record rather than assumed.

Good to merge.

PR #2094, opened 2026-09-01
eval: unique temp path so concurrent saves can't corrupt .ann.json

### LINKED ISSUE #1964: Concurrent annotation saves corrupt .ann.json, and the run log page 500s with no way back
> **Reviewed 2026-08-27 before junior handoff.** The stated cause was refuted — the
> writer is already atomic and already validates; the real defect is a shared temp-file
> name one layer down. Decision recorded below.

### Recovery: explain, do not offer a delete button

`eval/app/app/results/[...id]/page.tsx` renders the 422 from the run-log route as a blocking
message that names the full path of the corrupt `.ann.json` and gives the `rm` line from
"Recovery" above. **No delete button, and do not wire `deleteAnnotation`** — it stays
uncalled. Chosen over a one-click discard button because an unrecoverable destructive action
on up to 35 written comments does not belong on the page where the annotator is already
confused; it follows the partial-judge guard, the spec's only existing precedent for this
shape.

Record that choice and the rejected button in one clause in
`docs/specs/eval-crud-ui-spec.md`, next to the partial-judge guard at :186.

Two annotation saves in quick succession can leave `<run>.ann.json` holding **two JSON
documents concatenated**, which is unparseable. The run log's page then 500s and there is no
route back through the UI -- the only fix is to delete the file and re-annotate from the
start.

Hit while annotating `check-warnings/v1_2026-08-27_09-42-49` on 2026-08-27. Backed up before
deletion; the file was 728 bytes and looked like this:

```json
{
  "run_log": "v1_2026-08-27_09-42-49.json",
  "annotator": "...",
  "corrections": [
    { "test_id": "ut_check_warnings_001", "dimension_name": "Actionability", ... }
  ]
}
                                    <-- first document ends here
      "test_id": "ut_check_warnings_001",
      "dimension_source": "base",
      "dimension_name": "Correctness",
      ...
    }
  ]
}
```

`JSON.parse` fails at position 332. The API surfaces it as:

    GET /api/runlogs/check-warnings/v1_2026-08-27_09-42-49 -> 500
    Error: Annotation file ...ann.json is not valid JSON: Unexpected non-whitespace
    character after JSON at position 332
      at readAnnotation (lib/fs/annotations.ts:88)

## Why this is worse than it looks

**It fails on the next read, not on the write.** The save that corrupts the file appears to
succeed. Nothing tells the annotator anything is wrong until they reload -- by which point
the page is a bare 500 with no indication that an annotation file is the cause, no
indication which file, and no way to recover from inside the UI.

**The work is unrecoverable through the product.** `readAnnotation` throws before
`validateAnnotation` runs, so every route that reads the run log is dead, including the one
that would let you re-save over the bad file. `eval/CLAUDE.md` already prescribes the
remedy -- "if it is corrupt, delete it and re-annotate" -- but nothing surfaces that to the
person it happens to, and deleting a file under `eval/runlogs/` is not a genealogist-facing
action.

**The blast radius scales with diligence.** We lost two cells because it happened early. The
`check-warnings` samples run to 35 cells with a written comment on each; the same race
thirty cells in loses all of them, silently, with the annotator having no idea why the page
broke. Rule 3 of `check-runlogs` then blocks the PR until the whole thing is redone.

## Cause — verified, and not what it looked like

**The writer is already atomic and already validates.** `writeAnnotation` calls
`validateAnnotation` (`eval/app/lib/fs/annotations.ts:96`) and then `atomicWriteJson`
(`:98`), which writes a temp file and `rename()`s over the target — both since the CRUD
UI's first commit (`48cae125`, 2026-05-18). Neither of the two fixes originally proposed
here is missing, and neither would have prevented this: both concurrent payloads were
schema-valid, and the file was spliced, not appended.

**The defect is the temp filename.** `eval/app/lib/fs/atomic.ts:42`:

```ts
const tmpPath = `${filePath}.tmp-${process.pid}-${Date.now()}`;
```

The comment two lines above (`:39`) says "Suffix with PID + counter"; there is no counter.
Two writes to the same target within one millisecond, in one process, get the **same** temp
path. Both `fs.writeFile` calls open it with `O_TRUNC` and write from offset 0, so the
shorter payload overwrites the head of the longer one and its tail survives. One `rename`
then wins and the other fails `ENOENT`.

Reproduced standalone against a verbatim copy of `atomicWriteText`'s temp-name and write
(measured 2026-08-27, node 22, `Promise.all` of an 84-byte and a 3975-byte write to one
path):

```
rename results: [ 'ok', 'ENOENT' ]
CORRUPT: Unexpected non-whitespace character after JSON at position 84
```

Same error family, same shape (short document complete, then the tail of the long one), and
the reported 728-byte file with its error at position 332 fits it exactly: a 332-byte
one-correction save landed over the head of a 728-byte two-correction save.

**A lock is therefore not the fix, and neither is validating harder.** Give each write a
unique temp path — a module-level monotonic counter appended to the existing suffix. This is
`atomic.ts`, not `annotations.ts`, so it also closes the same hole for the three other
writers listed under Touches.

**Most likely trigger (unverified):** the save debounce
(`eval/app/app/results/[...id]/page.tsx:1227-1250`) clears its own timer, so one tab cannot
schedule two saves at once — but two browser tabs open on the same run log each hold their
own timer. Worth confirming, but do not gate the fix on it; the collision is reachable
regardless.

## Second defect: the 500 has no guard, and the sibling route already has one

`eval/app/app/api/runlogs/[...id]/route.ts:20` calls `readAnnotation` with no `try`, so a
throw becomes a bare 500 and the run-log page is dead. The annotation route already does
this correctly — `eval/app/app/api/runlogs/annotation/[...id]/route.ts:12-21` catches and
returns a 422 with the message. The precedent for what the page should then show is
`docs/specs/eval-crud-ui-spec.md:186` (the partial-judge guard): refuse to open the
annotation view and explain, rather than degrade to `annotation: null` — degrading would
silently discard the annotator's prior corrections, which is the exact failure
`annotations.ts:78-80` says it was written to stop.

`deleteAnnotation` (`eval/app/lib/fs/annotations.ts:102`) is exported and has **zero call
sites** — no route, no button. Whether to wire it is the decision block below; do not decide
it in the PR.

## Sites this touches

- [ ] `eval/app/lib/fs/atomic.ts` — unique temp path (counter, per the existing comment).
      Shared by four writers, all of which get the fix: `lib/fs/annotations.ts:98`,
      `lib/fs/tests.ts:183`, `lib/identity.ts:60`, `lib/release.ts:83-84`.
- [ ] `eval/app/tests/fs/atomic.test.ts` — the guard. **Prove it fails first**
      (`CLAUDE.md` § "A new lint must be proven to fail"): write the concurrent test, watch
      it corrupt the file on unpatched `atomic.ts`, then fix.
- [ ] `eval/app/app/api/runlogs/[...id]/route.ts:20` — catch, and return the structured
      error instead of 500ing.
- [ ] `eval/app/app/results/[...id]/page.tsx` — render that error as a clear message naming
      the file, per the decision below.
- [ ] `docs/specs/eval-crud-ui-spec.md` — add the corrupt-annotation behaviour beside the
      partial-judge guard at :186, and record in one clause that a temp-name collision (not
      a missing lock) was the cause, so the lock idea is not re-derived.

**Instrument:** `make eval-ui-test` (vitest, `eval/app`), then `make test-all`. The
concurrency case is fully deterministic in vitest — nothing here rests on human judgment.
The `nothing-checks` label clears when that test lands.

**Split, if you want half of it moving now:** the `atomic.ts` fix plus its guard test is
independent of the decision below and is startable immediately. The recovery-UX half waits
on the answer.

## Recovery, for whoever hits this before it is fixed

    # the file name is in the 500's error text
    rm eval/runlogs/unit/<skill>/<run>.ann.json

then reload and re-annotate. Copy the file somewhere first if you want the partial content.

Found during the `check-warnings` deep dive (#1616, PR #1896), which separately surfaced
`dimensionAllowsNa` refusing to offer N/A for a dimension the judge scored -- filed with the
validator requests. This one is filed on its own because it destroys work rather than
mis-scoring it.

**Touches:** `eval/app/lib/fs/annotations.ts`, `eval/app/app/api/runlogs/annotation/route.ts`,
`eval/app/tests/unit/`

Asimi1234:
## Summary

- Concurrent `.ann.json` saves within one millisecond corrupted the file: `atomicWriteText` built its temp path as `.tmp-${pid}-${Date.now()}` with no counter (despite the comment claiming one), so two same-ms writes shared one temp path, both opened it `O_TRUNC` from offset 0, and the shorter payload's head overwrote the longer — leaving two spliced JSON documents. The run-log page then 500'd with no way back through the UI. Fix: a module-level counter in the temp suffix, centralized in `atomic.ts` so all four `atomicWrite*` callers get it.
- Recovery UX: the run-log route now catches a corrupt annotation and returns a structured `422` (mirroring the sibling annotation route) instead of a bare 500; the page renders a blocking message naming the corrupt file's full path plus the `rm` line that recovers it. **No delete button, `deleteAnnotation` stays uncalled** — per the lead ruling on #1964.
- Tier: **Normal**.

Closes #1964.

## Review guidance

**Start here:** `eval/app/tests/fs/atomic.test.ts` — the guard's load-bearing assertion is that `await Promise.all([long, short])` **resolves** (not the final-file-state checks). On unpatched code the losing write's rename `ENOENT`s while the on-disk file often stays valid JSON, so asserting only on final state is a can't-fail check; "both writes resolve" is the discriminator. Verified red→green by hand (reverted the counter → the new test failed with the exact `ENOENT`, restored → green).

**Unsure about:** Nothing. The collision is fully deterministic in vitest once `Date.now()` is pinned.

## Plan

**Didn't change:** the two-tabs save-debounce (`page.tsx` `persist`) — the issue names it as the likely trigger but says not to gate on it; the counter fixes the collision at the source regardless of trigger. Did not wire `deleteAnnotation` / no discard button (ruling). Did not add a lock or stronger validation — the issue proved both are non-fixes (both colliding payloads were schema-valid).

**Acceptance check:** `eval/app/tests/fs/atomic.test.ts` → "both concurrent writes resolve, leaving one intact payload" — fails on unpatched `atomic.ts`, passes with the counter. This clears the `nothing-checks` label.

## Test plan

- [x] I ran `make test-all` and it passed. (All suites green: JS 416, server pytest 214, MCP vitest 2766, eval-app vitest 169, harness pytest 3021 — final line `All checks passed`.)

No skill run-log snapshot changed, and no skill `description`/DO-NOT clause changed, so the eval-run and negative-reciprocity boxes do not apply.

## Follow-on issues

**Folded in / filed instead:** None filed. One note for the reviewer, not drift: the issue's bottom `**Touches:**` trailer is stale — it names `annotation/route.ts` and `tests/unit/`, neither of which this work edits. I followed the issue body's authoritative "Sites this touches" checklist (the five files in this diff), which the plan-critic confirmed both rounds.


Asimi1234:
Author update — ran a self-review pass and folded two non-blocking improvements into `6122a1b3`:

1. **Recovery `rm` line now renders for both corruption shapes.** `readAnnotation` throws two messages, but only the JSON-splice one contained the file path; the page's regex silently produced no `rm` line for the off-schema case. The route now returns the path as a structured `filePath` field (via the newly-exported `annPathForRunLog`), so the page drops the brittle regex and always shows the recovery line. Spec clause updated to say the guard covers both unparseable-JSON and off-schema.

2. **Added route coverage for Half B.** `tests/integration/runlog-route-corrupt-annotation.test.ts` asserts the route returns `422 { error, filePath }` for both a spliced file and an off-schema file. Verified it goes red if the try/catch is removed (both cases 500 through). Previously Half B had no automated test.

Full `make test-all` green on the new head (eval/app 171, harness 3021, MCP 2766, server 214). The atomic guard's red→green mutation still holds.

yinkid28 (approved):
Review: PR #2094, full pass, 0 commits behind main

  Does: Adds a per-write monotonic counter to atomicWriteText's temp-file suffix so two saves within one millisecond get distinct temp paths instead of splicing. Catches corrupt .ann.json at the run-log route (structured 422 with file path) so the page shows a recovery message instead of a bare 500.

  Intent: Matches. "Didn't change" the debounce or deleteAnnotation, confirmed, neither touched. Acceptance check ("both concurrent writes resolve") is the discriminating assertion, on unpatched code one rename ENOENTs while on-disk state alone can look valid either way.

  Checks: All required checks green on 7970421a: pytest, runlogs, e2e-fixtures, vitest, lockfile-drift, scan, eval CRUD UI, server-pytest.

  Still needs: senior-developers + senior-genealogists (CODEOWNERS *.ts/*.tsx rules).

  Verified: eval app vitest 169/169 tests passed. The new integration test file (runlog-route-corrupt-annotation.test.ts) fails standalone vitest because it imports next/server  environmental, passes in CI's "eval CRUD UI" step. All four atomicWrite* callers (annotations.ts, release.ts, identity.ts, tests.ts) go through atomicWriteText, the counter covers every write site.

Existing reviews: None, first review.

  No stop-rule conditions. No critical or informational findings. The counter resets on process restart, which is correct, PID already provides process-level uniqueness; the counter handles within-process same-millisecond collisions.

  Approving.

promise-emmanuel (changes requested):
One thing to finish before this merges.

The guard covers the run-log route, but two sibling routes still read the
annotation bare, so a corrupt `.ann.json` 500s them the same way the run-log
page used to. Same defect, same fix, and the branch is already open.

In `eval/app/app/api/runlogs/compare/route.ts`:

You wrote:
  const [recentAnn, previousAnn] = await Promise.all([
    readAnnotation(recentId),
    readAnnotation(previousId),
  ]);

Change it to:
  let recentAnn, previousAnn;
  try {
    [recentAnn, previousAnn] = await Promise.all([
      readAnnotation(recentId),
      readAnnotation(previousId),
    ]);
  } catch (err) {
    return NextResponse.json(
      { error: 'invalid_annotation', message: (err as Error).message },
      { status: 422 },
    );
  }

In `eval/app/app/api/runlogs/trend/route.ts`:

You wrote:
    const ann = (await readAnnotation(r.id)) as
      | { corrections: Array<{ test_id: string; dimension_source: string; dimension_name: string; corrected_score: ScoreOrNull }> }
      | null;

Change it to:
    let ann;
    try {
      ann = (await readAnnotation(r.id)) as
        | { corrections: Array<{ test_id: string; dimension_source: string; dimension_name: string; corrected_score: ScoreOrNull }> }
        | null;
    } catch (err) {
      return NextResponse.json(
        { error: 'invalid_annotation', message: (err as Error).message },
        { status: 422 },
      );
    }

Everything else is good. The counter fix is correct and I verified it against
the real `atomicWriteText`: without the counter two concurrent writes leave 367
bytes of spliced JSON and one rejects with ENOENT; with it, both finish and the
file holds one whole payload.

Asimi1234:
@promise-emmanuel Fixed in 32fffa88 — thanks, you were right that the guard stopped at the run-log route.

Both bare callers now catch and return the structured 422:
- `compare/route.ts` — wrapped the `Promise.all([readAnnotation, readAnnotation])`
- `trend/route.ts` — wrapped the per-run `readAnnotation` in the loop

I also added parity coverage the original diff was missing: `tests/integration/annotation-reader-routes-corrupt-annotation.test.ts`, one test per route feeding a spliced-JSON `.ann.json` (the temp-name-collision shape) and asserting `422` + `invalid_annotation`. Proven red→green by reverting the two route fixes and watching both tests fail with the uncaught throw, then restoring. Full eval-app suite 173/173, `tsc --noEmit` clean.

I kept the message-only 422 shape from your snippet (no separate `filePath` field like the run-log route) since these routes read multiple annotations and the thrown message already carries the offending path.

promise-emmanuel (approved):
Both route guards check out. I reverted each one and watched the new tests go
red with the uncaught throw, then restored them — they discriminate. The
parity test file you added wasn't something I asked for and it's the right
call.

I pushed one small follow-up rather than sending you round again (fc0119ef).

The compare and trend pages were discarding the 422 body and throwing a bare
status string, which is what their red Alert renders — so the message your
routes build never reached the annotator. They saw `compare → 422`: no file
name, no recovery line. Both pages now read body.message the way the run-log
page already did.

Checked it against a real dev server with a deliberately spliced .ann.json on
disk. Before: the alert reads `trend → 422`. After: it reads the full path of
the corrupt file. Lint, tsc and the 173-test suite are green.

One thing still open, not for this PR: for a parseable-but-off-schema
annotation, validateAnnotation's message carries no file path, so on compare
and trend the alert names no file even now. Only the run-log route passes
filePath separately. Worth an issue if it bothers anyone — the fix is to
attach the path inside readAnnotation, which is a shared-function change.

Approving.

Issue #2106, opened 2026-09-01
Three skills have zero invocations in 159 e2e runs — they are exactly the three the orchestrator never names


DallanQ:
**Touches:** packages/engine/plugin/skills/research/SKILL.md, eval/harness/ (a guard — location to be decided by whoever takes it)

Three shipped skills — `translation`, `convert-dates`, `historical-context` — have **zero invocations** across all 159 committed e2e runs. They are exactly the three skills the `/research` orchestrator never names.

This card is the shared root cause. The per-skill cards are **#2074** (translation, `needs-decision`), **#2105** (historical-context) and **#2107** (convert-dates); each of them points here first, and each may reduce to nothing once this lands.

## The measurement

Segmenting every run's `tool_calls` ledger on `Skill`/`SlashCommand` calls across `eval/runlogs/e2e/` (159 runs):

| Skill | Invocations | Named in `research/SKILL.md`? |
|---|---|---|
| question-selection | 248 | yes (×6) |
| search-records | 204 | yes (×2) |
| research-plan | 187 | yes (×7) |
| person-evidence | 154 | yes (×4) |
| record-extraction | 126 | yes (×3) |
| research-exhaustiveness | 116 | yes (×11) |
| proof-conclusion | 97 | yes (×20) |
| locality-guide | 73 | yes (×1) |
| check-warnings | 40 | — |
| search-external-sites | 27 | yes (×1) |
| search-full-text | 11 | — |
| conflict-resolution | 9 | yes (×3) |
| search-images | 8 | yes (×1) |
| init-project | 4 | yes (×2) |
| **translation** | **0** | **no** |
| **convert-dates** | **0** | **no** |
| **historical-context** | **0** | **no** |

The orchestrator names 15 skills. The three at zero are the three it does not name.

```sh
for s in translation convert-dates historical-context; do
  printf '%-22s %s\n' "$s" "$(grep -c "$s" packages/engine/plugin/skills/research/SKILL.md)"
done   # -> 0, 0, 0
```

## The demand is real

Measured over the 94 fixtures' final trees:

- **58 of 94 (62%)** carry at least one fact in a non-English-language jurisdiction — Norway, Germany, Mexico, Slovakia, Netherlands, France, Brazil, Denmark, Hungary, Sweden and others.
- **3 of 94** carry a pre-1752 dated fact, where Julian/Gregorian and Old Style/New Style corrections apply.
[...]

Issue #2081, opened 2026-08-31
Lift the autonomous plan freeze: edits written and pushed, blocked on two contended eval slots


DallanQ:
**Touches:** packages/engine/plugin/skills/search-records/SKILL.md, packages/engine/plugin/skills/research-plan/SKILL.md, packages/engine/plugin/skills/research/SKILL.md

The lead ruled on 2026-08-31 (issue #2077) that autonomous runs may revise a plan mid-execution, and that autonomous runs should simulate production runs as closely as possible. **The edits are written and up for review — PR #2085.** This card exists because they cannot land yet: both skills they touch are eval-slot contended, so CI on that PR is red by construction until the two paid runs are bought.

## What is on the branch

Three edits, each removing a mode split rather than adding an exception to one.

**`search-records`** — delete "An autonomous `/research` run has no ad-hoc searches — the orchestrator only dispatches you when a plan item is waiting." That sentence closed the skill's **self-initiated** route back to research-plan whenever no user was watching — the Step 1 row it takes when it *thinks of* a search rather than nilling on a planned one — so an autonomous run could never raise a plan problem it noticed, while an interactive one could. It is one route of three, not the only one: the Step 2 boundary-change bounce ("If the plan offers none and the nil persists, **bounce back to `research-plan`**") and the Step 6 fallback-miss suggestion are not user-gated and stay open. The drift guard beside it stays ("inventing a search nobody asked for is how a session drifts off its question") — the skill still may not invent a search, it must route through `research-plan`, which owns plan structure and applies the breadth standards. Only the dead end goes. The routed table row no longer presupposes that a user says yes.

**`search-records`** — "Do not re-query between searches; **the plan does not change while you are searching it**." The justifying clause becomes false once revision is reachable. The limit stands, scoped to the invocation.

**`research-plan`** — Re-invocation said to mark a plan `superseded` "only when the user is explicitly re-planning." With no user that is never true, so it contradicted Step 1a's own supersede condition ("the active plan has unfinished items but new information invalidates its assumptions") and **would have made the whole lift a no-op** — the loop would route to research-plan, which would recap the plan and change nothing. It now states Step 1a's condition.

[...]

PR #2112, opened 2026-09-01
eval: guard against a skill going dark in the e2e corpus (#2106)

benter-070:
Builds the guard half of #2106. **Does not touch `research/SKILL.md`** — see below.

> **Stacked on #2060 — review only `eval/harness/tests/unit/test_skill_reachability_e2e.py`.**
> This branch was cut from `1750-output-formats-pointers` instead of `main`, so six of
> the seven files in the diff (`project-status/SKILL.md`, `docs/architecture.md`, the
> emptied `MISSING_PENDING_ADJUDICATION`, and the two project-status run logs) are
> #2060's work and are reviewed there, where #1750's "say which option you took" is
> answered. Only the 143-line guard belongs to this PR. Once #2060 merges, merging main
> here drops those six from the diff on their own — no force-push, which this repo's
> ruleset blocks anyway. My mistake in cutting the branch; flagged in review and noted
> so I do not repeat it.

## What this adds

`eval/harness/tests/unit/test_skill_reachability_e2e.py`, a shrink-only ratchet on
skills that no e2e run invokes. It follows the pattern
`skill-reference-reachability.test.ts` already uses: a recorded population that may
only get smaller, where adding an entry means something regressed.

Measured over the 159 committed e2e runs: **12 of 27 shipped skills have zero
invocations.**

Nothing in either eval tier fails on that today. A unit suite grades a skill the
harness invokes *by construction* — it names the skill — and the e2e tier grades
the research outcome, not which skills reached it. So a skill can ship, be
maintained, cost prompt tokens in every router decision, and never run. That is
what happened, unnoticed, until someone counted.

## Four arms, each proven to fail before commit

Not asserted — each was broken deliberately and watched fail:

| Arm | Broken by | Failure |
|---|---|---|
| new skill goes dark | removing `timeline` from the list | names `['timeline']` |
| stale entry | adding `search-records` (204 invocations) | names it, says the list only shrinks |
| deleted skill listed | adding a fake name | names it |
| corpus unreadable | repointing `E2E_RUNLOGS` | "no skill invocations found" |

The fourth arm exists because without it the other three pass vacuously on an empty
scan — the failure mode this repo has been bitten by before.

## Two entries are expected, and both were checked rather than assumed

- **`project-status`** reports state to a human, and every e2e run is launched as
  `/research --autonomous` (`eval/harness/e2e/orchestrator.py:876`). There is no
  human mid-run to report to.
- **`forget-and-rederive`** is operator-driven and no fixture requests it. The only
  match across `eval/tests/e2e/*/fixture.json` is provenance prose in
  `cook-spouse-children`'s `notes` — that fixture's tree was pre-stripped to mirror
  a real forget-and-rederive bug (#1473), rather than asking the agent to invoke
  the skill. I nearly recorded the opposite after a loose grep.

Both are listed rather than exempted, so the count stays honest.

## Why the routing half is not here

@florencemashipei refuted it in review and I verified her claims rather than
re-running her count. Both hold:

- **Naming is not sufficient** — `hypothesis-tracking` is a routing-table row at
  `research/SKILL.md:149` and has **0** invocations; `project-status` is named at
  line 16 and has **0**.
- **Naming is not necessary** — `check-warnings` and `search-full-text` have
  `grep -c` = **0** mentions in `research/SKILL.md` and are invoked 40 and 11 times.

Verifying that turned up one counterexample to her own hypothesis, recorded on the
card: `historical-context` *is* named by `locality-guide` (73 invocations) and is
still dark. All three of those mentions are boundary markers — a DO-NOT clause, a
"belongs in", a "Redirect to" row — whereas the live referrals are imperative
delegations (`person-evidence:643` "**invoke** `check-warnings` on the affected
persons"). So what predicts invocation is being named as a step the parent
*performs*, not merely being named by a busy parent.

## The card's cost estimate is correct — an earlier draft of this body said otherwise

**#2106 is right that editing `research/SKILL.md` buys a paid run. Budget for it.**

An earlier revision of this body claimed the opposite. That claim was true when
written — at this branch's merge base (`8b92cf54`) `research` sat in
`RUNLOG_GATE_EXEMPT_SKILLS`, had no unit-test directory, and appeared in no
run-log snapshot — and `be0d0a0b` (#1494 phase 1a) has since landed. All three
legs now fail on current main, re-verified here rather than taken from review:

- `RUNLOG_GATE_EXEMPT_SKILLS` is `frozenset({"forget-and-rederive"})` alone
  (`check_runlogs.py:103`)
- `eval/tests/unit/research/` exists — 16 files
- `skills/research/SKILL.md` is in **5** committed unit run-log snapshots

Live proof rather than inference: #2162 edits `packages/engine/plugin/agents/gps-mentor.md`,
which all five of those snapshots contain, and its `runlogs` check fails today with
`skill research: latest full-skill run log v1_2026-09-01_18-35-01.json is NOT active`.

So the contention with #2081 and the other cards holding that file is real, **and
so is the cost.** Whoever sequences the routing half should plan a paid `research`
run into it. Correction caught by @EdmondOware in review; recorded here rather
than only in the thread, because the repo squash-merges and this body is what
`git log` keeps.

## Verification

- New guard: 4 pass; each arm separately proven to fail
- `test_encoding_lint.py`: 6 pass — the new file passes the repo-wide `encoding=` AST lint
- Re-measured after merging main (153 commits, up to `c98dc54c`): still **159 runs, 27
  skills, 12 dark** — the recorded list survives the update, as @EdmondOware also found
  independently. Counted with the guard's own `invocation_counts()` rather than a shell
  glob: a `grep -v ann` silently drops every fixture slug containing "ann" and
  undercounts the corpus to 148. Guard re-run on the merged tree: 4 pass.


EdmondOware (changes requested):
The measurement is right. I recounted it independently rather than reading your table, using your own glob and filters: 159 `run-*.json` files, 27 shipped skills, 12 dark, and my list matches yours name for name. Every supporting figure holds too. `check-warnings` 0 mentions in `research/SKILL.md` and 40 invocations, `search-full-text` 0 and 11, `hypothesis-tracking` 2 mentions and 0 invocations, `locality-guide` 73 invocations while the `historical-context` it names 3 times stays dark.

That last one is the best thing in the PR. You went looking to confirm Florence's hypothesis, found a counterexample to it, recorded it on the card rather than dropping it, and it produced the sentence the routing half will actually need: what predicts invocation is being named as a step the parent performs, not being named by a busy parent. Volunteering evidence against your own framing is the hard version of this.

The vacuous scan arm is the right instinct too. `test_the_corpus_is_actually_readable` is what stops the other three passing in the wrong direction on an empty glob, which is the exact way a check in this repo silently reads as coverage.

Two things to fix. Neither is code.

**1. The cost correction has gone stale, and it is the load bearing claim for the half you deferred.**

In fairness it was true when you wrote it. At your merge base `8b92cf54` the set really was `{"research", "forget-and-rederive"}` and there were zero research run logs, so this is main moving under you rather than a mistake. But `be0d0a0b` (#1494 phase 1a) has since landed and all three legs now fail:

- `RUNLOG_GATE_EXEMPT_SKILLS` is `frozenset({"forget-and-rederive"})` alone
- `eval/tests/unit/research/` exists
- `skills/research/SKILL.md` is in **5** committed unit run-log snapshots

So editing `research/SKILL.md` does now buy a paid run. Live proof rather than inference: #2162 is blocked on that gate right now for merely touching a file inside `research`'s snapshot. Since the repo squash merges, this paragraph becomes the permanent `git log` entry, and the next person sequences the routing work on a cost that no longer exists.

Updating the branch is cheap and safe: a dry run merge of main is clean, and I recounted the dark population on a tree with current main merged and it is still the same 12, so the recorded list survives.

**2. Three of the seven files are not in the body at all.**

`project-status/SKILL.md`, `docs/architecture.md` and the emptied `MISSING_PENDING_ADJUDICATION`, plus the 3265 line run log they required, are undocumented here. The body describes the guard and nothing else, and never mentions `output-formats.md` or #1750.

That matters more than usual because #1750 asks for exactly this: *"Decide and do one, then say which in the PR: write `references/output-formats.md` with the literal skeletons, or delete both pointers and carry the shapes inline in step 5."* What shipped is a third option. You deleted the pointers and, rather than carrying the shapes inline, wrote "Shape each one as the project's state warrants." That may well be the right call, but the card asked to be told, and #1750 stays open with no reference from here.

For what it is worth I checked your justification and it holds. `rubric.md` grades Completeness of summary, Accuracy and Actionability, and nothing in it grades layout, shape or format, so there was no layout contract to lose. And your run is 11 of 11 pass, including `ut_project_status_006`, which is the test #1750 cited as live evidence of the agent improvising. That is a strong result and it belongs in the body where a reviewer can see it.

One small thing while you are in there: that run was scored against judge prompt `03f306ff` and the current one is `9451d105`, so the scores would likely move on a re run. Not blocking, and the gate only warns, but worth a line so nobody reads 11 of 11 as current.

benter-070:
Both fixed. Thank you for re-counting independently rather than reading my table — and
for catching the cost claim, which was the one that would have done real damage.

**1. The cost correction is withdrawn.** You are right on all three legs; I re-verified
each on current main rather than taking them from the review:

- `RUNLOG_GATE_EXEMPT_SKILLS` is `frozenset({"forget-and-rederive"})` alone (`check_runlogs.py:103`)
- `eval/tests/unit/research/` exists — 16 files
- `skills/research/SKILL.md` is in 5 committed unit run-log snapshots

I also confirmed your #2162 example end to end rather than repeating it: it edits
`packages/engine/plugin/agents/gps-mentor.md`, all five of those snapshots contain that
file, and its `runlogs` check is red right now with `skill research: latest full-skill
run log v1_2026-09-01_18-35-01.json is NOT active`. The section now says the card's
estimate is correct and to budget a paid `research` run. I corrected the matching
comment on #2106 too, since that is where anyone sequencing the routing half will read
it.

Your point about squash-merge is what made me fix the body rather than just reply here.

**2. The three files are #2060's, not undocumented work.** I cut this branch from
`1750-output-formats-pointers` instead of `main`, so `project-status/SKILL.md`,
`docs/architecture.md`, the emptied `MISSING_PENDING_ADJUDICATION` and the 3265-line run
log are all #2060's commits riding along. Only the 143-line guard is mine here. Rather
than document another PR's work in this body, I have added a note at the top pointing
review of those six files at #2060 — where I have now answered #1750's "say which option
you took", added the 11/11 framing, and added your judge-prompt caveat as limit 3
(verified: run log `03f306ff`, current prompt `9451d105`).

I did not force-push to un-stack — the ruleset here blocks it, and once #2060 merges,
merging main drops those six from this diff on its own. #2060 is green and back with you.

**3. Branch updated as you suggested.** Main merged (153 commits). Guard re-runs 4 pass
and the population holds at 159 runs / 27 skills / 12 dark, matching your recount. One
thing worth recording: my first re-count said 148, because `grep -v 'ann'` drops every
fixture slug containing "ann". The guard's own `invocation_counts()` is authoritative
and I have noted that in the body so the next person does not repeat it.

EdmondOware:
Following up on my review above, having since reviewed #2060.

My note about the three undocumented files understated it. **#2112 carries all six of #2060's files**, not just three, and in a reduced form:

| file | #2060 | here |
|---|---|---|
| `project-status/SKILL.md` | | identical |
| the run log + annotation | | identical |
| `docs/architecture.md` | 16 lines, 3 hunks | 11 lines, 1 hunk |
| `skill-reference-reachability.test.ts` | 49 lines | 22 lines |

The gap that matters is not size. #2060 corrects a false claim in the reachability test's own header comment, that an unnamed reference is "unreachable in every environment". It is not: both environments stage every skill's `references/` folder into the session, and `ut_convert_dates_012` globbed `**/*` and read three of the seven exempted files, one of which changed its answer. I verified that, including the literal `Read` call in the run log. **None of that is here.**

So the ask changes from what I wrote. Rather than documenting these files, **drop them and rebase once #2060 lands.** #2060 is approved and ahead; this PR's own contribution is the dark-skill guard, which is good work and stands on its own.

The two points from my review still hold: the branch needs updating, and the cost correction about `research` and `RUNLOG_GATE_EXEMPT_SKILLS` is now stale.

EdmondOware (approved):
Read your 08:03 reply after posting mine, which is my error and it duplicated most of what you had already answered. Two corrections to it.

"Drop them and rebase" was wrong. Force-push is blocked repo-wide, as you say, and your mechanism is the right one: once #2060 merges, merging main takes those six out of this diff. Nothing to do.

The cost point you had already withdrawn before I repeated it, and you went further than I asked by fixing #2106 too, which is where it would actually have done the damage. The `grep -v 'ann'` note is a good catch to leave behind.

Approving on the guard alone, per your stacked note. It is unchanged from what I reviewed, 4 pass on the merged tree, and the population holds at 159 / 27 / 12.

One nit on #2060 while it is open: that section is now headed "Three limits" but still opens "Both come from the card itself", and limit 3 came from review rather than the card.

florencemashipei (commented):
Senior review of the guard only, per your stacked note. CI is green and I reproduced every figure from scratch on current main (`4d824de24`, one commit past your merge base) rather than reading the table: 159 run logs, 27 shipped skills, **12 dark**, list matching name for name. The supporting numbers hold too — `check-warnings` 0 mentions / 40 invocations, `search-full-text` 0 / 11, `hypothesis-tracking` at line 149 with 0, `historical-context` named 3 times by `locality-guide` (73 invocations) and still dark. `RUNLOG_GATE_EXEMPT_SKILLS` is `frozenset({"forget-and-rederive"})` and `eval/tests/unit/research/` has 16 files, so the corrected cost paragraph is right. The `forget-and-rederive` justification checks out: the only fixture match is `cook-spouse-children`'s `notes`, describing the #1473 provenance.

I broke each of the four arms and watched it fail rather than taking the table's word for it. All four fail as described, and the vacuous-scan arm does the job you built it for — with `E2E_RUNLOGS` repointed, `test_no_new_skill_has_gone_dark` also fails loudly naming 15 skills, so an empty glob cannot read as coverage.

I also checked the one way this measurement could be quietly wrong: a skill reached from inside a subagent rather than by the top-level agent. It is covered. `tool_calls` is flat and carries `agent_id`/`agent_type` on 3714 entries, including one `Skill` call (`question-selection`) made by a subagent, so your scan counts those. And no dark skill appears as a `subagent_type` either — those are `record-extractor`, `image-reader`, `gps-mentor`, `proof-conclusion`, `question-selection`. Nothing is being reached by a route the guard misses.

Two things I would like answered before this merges. Neither is a doubt about the measurement.

**1. The list cannot only shrink, and a brand-new skill is the proof.**

I created `packages/engine/plugin/skills/zz-brand-new-skill/` and ran the suite:

```
AssertionError: skill(s) with zero e2e invocations that were reachable before:
['zz-brand-new-skill']. Either the change that made them unreachable should be
reconsidered, or - if this is deliberate - add them to DARK_SKILLS_2026_09_01
with a reason.
```

Three things go wrong in that one message. It asserts the skill "was reachable before", which is false for a skill that has never existed. It instructs the reader to *add* an entry to a list your module docstring says at line 19 "may only **SHRINK**". And the constant is dated `2026_09_01`, so a skill added next month lands in a list named for a measurement it was never part of.

This is certain rather than hypothetical: `eval-harness-tests.yml`'s `PATTERNS` includes `^packages/engine/plugin/skills/`, so every new-skill PR runs this job and hits it. And the only clean way off the list is a paid e2e run that invokes the new skill, which means the guard quietly makes "ship a new skill" depend on "buy an e2e run" — a real cost you may well be happy to impose, but it should be a stated decision rather than a side effect. A separate `NEW_SINCE_BASELINE` set, or a rule keyed on the skill directory's age rather than on the dated list, would keep the shrink-only invariant honest.

**2. The failure message asks for a reason the data structure cannot hold.**

The message says "add them to `DARK_SKILLS_2026_09_01` with a reason", but the structure is a `frozenset` of bare strings with nowhere to put one. The block comment records reasons for 2 of the 12 entries; the other ten have none.

`skill-reference-reachability.test.ts`, the precedent this PR cites, does hold them — `Array<{ path, why }>` with a per-entry `why`, and `EXEMPT` derived from it, so an entry cannot be added without stating why. A `dict[str, str]` here would match that pattern and make the message truthful. It also matters more on this side than on the JS side: a `references/` file becomes reachable by being named in a SKILL.md, which is free, whereas a skill becomes non-dark only via a paid run, so entries here will sit longer and their reasons are what a reader a year from now will need.

Cleanups, none blocking:

**3. The corpus predicate already exists.** `e2e/runlog_selection.py` exports `E2E_RUNLOGS`, `is_result_json()` ("a committed `run-<ts>.json`, excluding `.ann.json` and `.final-*`") and `all_result_jsons()`. `e2e/inventory.py` already uses them to print the 159 count, and that module's whole premise is that a count should be defined by one named predicate so it cannot drift. I confirmed your inline glob and `all_result_jsons()` select an identical 159 paths, and that the import resolves from the harness venv. Same for the skills directory: `harness/orchestrator.py` has `DEFAULT_SKILLS` and `e2e/orchestrator.py:112` has `DEFAULT_PLUGIN_SKILLS`, which is the directory the orchestrator actually copies into each e2e run and so is the more exact meaning of "shipped". Reusing both would drop the `parents[3]` root derivation and two duplicated constants, and would mean a future change to the filename convention updates the guard along with everything else.

**4. `shipped_skills()` counts dotted directories.** `harness/orchestrator.py:498` does the same enumeration as `d.is_dir() and not d.name.startswith(".")`. There are no dotted entries under `skills/` today, so this is precautionary, but a stray cache directory would become a phantom dark skill.

**5. The `orchestrator.py:876` citation is off by eight lines.** The `/research --autonomous` launch is at `:884` — on main and on this branch, so this is not drift from merging main. `:876` lands in the middle of `_render_user_message`'s docstring. Worth fixing in both places: the guard's line 45 is where a future reader will follow it to decide whether `project-status` still belongs on the list, and the body is what `git log` keeps after the squash.

**6. The `SlashCommand` branch is unexercised.** There are zero `SlashCommand` entries in the corpus today, which is fine for defensive code. But the docstring only anticipates a `plugin:` prefix, and a real slash command carries arguments (`/research --autonomous <question>`), which would not normalize to a bare skill name. If such an entry ever appears it reads as *dark* rather than as invoked — a silent undercount in the direction that produces a false positive. One clause in the docstring, or a `.split()[0]`, closes it.

**7. Nit, take it or leave it.** `invocation_counts()` rescans the 66MB corpus three times, once per test. I measured 0.43s per warm scan, so an `lru_cache` buys about a second of an 81-second job. Mentioning it only so nobody else measures it.

One observation on the stacking, since it changes what a reviewer sees rather than just how the diff reads. #2112 carries #2060's commits only up to `bbf4186d9`; `68802ded6` is on #2060 and not here. So the `docs/architecture.md` and `skill-reference-reachability.test.ts` shown in this diff are a *stale* copy of #2060's work, with that PR's own review fix missing. Your note says the six files are reviewed on #2060, which is right, but anyone who reads them here reads a superseded version. Nothing to do on this branch — merging main after #2060 lands fixes it — but worth a line in the note so the next reader does not review the old text.

Guard is sound and the honesty of the measurement is not in question. Holding on 1 and 2.

florencemashipei (approved):
Approving on the guard, per your stacked note. The measurement is verified independently, all four arms fail as claimed, and the subagent route is covered — nothing reaches a listed skill by a path the scan misses.

My two items stand as follow-ups rather than blockers:

1. A brand-new skill trips `test_no_new_skill_has_gone_dark` with "was reachable before", and the only exit is a paid e2e run. That lands on whoever adds the next skill, not on this PR, so it wants an answer on the record — here or on #2106 — rather than a change here.
2. The message asks for a reason `frozenset` cannot hold. A `dict[str, str]` would match `skill-reference-reachability.test.ts` and can come with 1 or separately.

3 through 6 are cleanups, take them whenever suits.

One thing to hold to, since my approval is not dismissed on push (`dismiss_stale_reviews_on_push: false`): **please let #2060 merge first.** What I have approved is the guard. Merging this ahead of #2060 would land that PR's six files at their pre-`68802ded6` state, and I have changes requested open there. Once #2060 lands and you merge main, this diff becomes the one file and the approval fits it.

benter-070:
## Pending correction on an annotation this PR carries — do not merge ahead of #2060

Flagging here because the finding is on #2060 and nothing on this PR said so, and this
branch carries the file **byte-identical** (blob `ee480c97` on both branches — it rode
along with the other five files when I cut this branch from `1750-output-formats-pointers`
rather than `main`).

@chesworthrm found that `eval/runlogs/unit/project-status/v1_2026-08-31_09-15-36.ann.json`
records a failure that did not happen. On `ut_project_status_004` the run shows
`skills_invoked: ["proof-conclusion"]`, `activated: false`, zero tool calls and zero
files created — which is the **pass** condition for a routing negative, since the
outcome is decided by where it routed, not by the judge's dimension scores. The judge
read the empty output as silence and scored Correctness and Completeness 1 each, and
the annotation currently **confirms both 1s**:

| dimension | llm | corrected |
|---|---|---|
| Correctness | 1 | 1 |
| Completeness | 1 | 1 |

Its comment then tells the next reader to make `project-status` print a decline
message — which would send someone into `project-status/SKILL.md` to fix a judge
problem, on a run the skill never enters.

**Status.** Not corrected yet, deliberately. Unit `.ann.json` files may only be written
through the eval CRUD UI — hand-editing produces the deprecated correction shape that
later crashes the `check-runlogs` gate — and @chesworthrm asked @EdmondOware to confirm
`3` is the right score before the override is recorded, since his drafted comment
becomes the durable justification. Once Edmond confirms I will make both corrections in
the UI on #2060.

**What this means for merge order** — same conclusion Edmond reached from the other
direction. #2060 goes first. It now also carries five review fixes this branch does not
have, and the globbing correction this branch never had. If this PR merged first the
annotation lands twice uncorrected and those fixes arrive as conflicts. I will drop the
six duplicated files from here once #2060 is in.

florencemashipei:
## Please don't merge this ahead of #2060

Flagging as a senior genealogist on this queue, not as a review of the guard itself — I have not reviewed that yet.

This PR is approved and mergeable, and #2060 is not (I am still holding it on one annotation correction). That ordering is backwards for the six files the two share, and if this lands first the loss is silent.

What I verified by diffing the blobs at both heads rather than reading the file lists:

| file | here | on #2060 |
|---|---|---|
| `project-status/SKILL.md` | byte-identical | byte-identical |
| `v1_2026-08-31_09-15-36.json` | byte-identical | byte-identical |
| `v1_2026-08-31_09-15-36.ann.json` | byte-identical | byte-identical |
| `docs/architecture.md` | 7 / -4 | 23 / -11 |
| `skill-reference-reachability.test.ts` | 12 / -10 | 50 / -29 |

The two reduced files are the problem. #2060's versions carry a measured correction this branch does not have at all: the claim that an unnamed `references/` file is "unreachable in every environment" is **false**, and `ut_convert_dates_012` proves it — it globbed `**/*`, read three files off that test's own exemption list, and one of them changed its answer. #2060 also replaces a staleness loop that lost its only failure path when the exemption list was emptied. None of that is here. Landing this first either drops those corrections or turns them into a conflict resolved by whoever happens to be holding the rebase.

The `.ann.json` matters for a second reason. Its `ut_project_status_004` entry confirms two scores of 1 on a routing negative where the skill never activated, and its comment prescribes a fix inside `project-status/SKILL.md` — a snapshotted file — for what is a judge artifact. Acting on that costs a paid run for nothing. I have ruled on it on #2060 (per CODEOWNERS, `/eval/runlogs/` is the genealogists' path first). Since this branch carries the file byte-identical, correcting it there and not here means it ships twice, once wrong.

So: #2060 first, then drop the six duplicated files from this branch as @benter-070 has already offered to do. @EdmondOware and @chesworthrm both reached the same ordering independently.

I will pick this PR up on its own merits once the ordering is settled.

florencemashipei (changes requested):
## Ordering block — not a verdict on the guard

**I have not reviewed this PR on its merits.** This is a mechanical hold so it cannot be merged ahead of #2060, and I will clear it the moment #2060 lands. @EdmondOware's approval stands; nothing here disagrees with it.

I am doing this as a review rather than another comment because a comment is not a mechanism. This PR is approved and mergeable while #2060 is not, and everyone who has read both threads knows the ordering — the risk is the next person to work the board, who sees a green tick and no objection.

The paths that make it mine to hold: `.github/CODEOWNERS` names `@PioneerAIAcademy/senior-genealogists` first on both `/eval/runlogs/` and `/packages/engine/plugin/skills/`, and this PR touches both.

### What is at stake, verified by diffing the blobs at both heads

| file | here | on #2060 |
|---|---|---|
| `project-status/SKILL.md` | byte-identical | byte-identical |
| `v1_2026-08-31_09-15-36.json` | byte-identical | byte-identical |
| `v1_2026-08-31_09-15-36.ann.json` | `ee480c97` | **`43684873`** |
| `docs/architecture.md` | 7 / -4 | 23 / -11 |
| `skill-reference-reachability.test.ts` | 12 / -10 | 50 / -29 |

Two things changed since I first raised this.

**The annotation has diverged.** It is no longer byte-identical. #2060 now carries a corrected `ut_project_status_004` — both base dimensions overridden 1 → 3 on my ruling, because the judge scored a routing negative as silence on a run where the skill never activated. This branch still carries the uncorrected version, whose comment tells the next reader to fix the problem inside `project-status/SKILL.md` — a snapshotted file, so acting on it buys a paid eval run for a judge artifact. If this lands first that instruction ships.

**The two reduced files are the real loss.** #2060's `architecture.md` and reachability test carry a measured correction this branch does not have in any form: the claim that an unnamed `references/` file is "unreachable in every environment" is **false**, and `ut_convert_dates_012` proves it — it globbed `**/*`, read three files off that test's own exemption list, and one of them changed its answer. #2060 also replaces a staleness loop that lost its only failure path when the exemption list was emptied. Landing this first either drops both or turns them into a conflict resolved by whoever holds the rebase, and a conflict resolution is the one edit nobody reviews.

### To clear this

#2060 merges, then drop the six duplicated files from this branch as @benter-070 has already offered. Ping me and I will convert this to an approval or a real review of the guard, whichever the queue wants.

@chesworthrm and @EdmondOware both reached this ordering independently.

florencemashipei:
## #2060 has landed — `91fc340cc` on main. The hold's reason has changed, so here is where it stands.

@benter-070 The ordering hold has done its job and I am no longer holding this on sequencing. I am leaving `CHANGES_REQUESTED` in place because the branch is now **`CONFLICTING` / `DIRTY`** and 2 behind main, and the resolution is the one edit that could still lose the correction. Ping me when it is rebased and I will convert this into a real review of the guard — which I still have not read on its merits.

## The resolution is "take main's version", not "resolve the conflict"

Two of the six duplicated files are now **byte-identical to main** and drop cleanly:

- `packages/engine/plugin/skills/project-status/SKILL.md`
- `eval/runlogs/unit/project-status/v1_2026-08-31_09-15-36.json`

The other three are **not duplicates any more — this branch's versions are reverts.** I diffed `main → pr-2112` rather than reading the file list, and each one takes main backwards:

| file | what this branch's version does |
|---|---|
| `v1_2026-08-31_09-15-36.ann.json` | reverts `ut_project_status_004` to `corrected_score: 1` on both base dimensions — undoing the override |
| `docs/architecture.md` | restores **"a file neither route reaches is unreachable"** and **"no session could load them"**, deleting the globbing paragraph |
| `skill-reference-reachability.test.ts` | restores **"unreachable in every environment"** in the header comment and drops the `ut_convert_dates_012` measurement |

That last pair is the specific loss I blocked for, and it is worth being blunt about why: the claim being restored is **false**, and we have a run log proving it. So take main's side on all six and re-apply nothing. If the e2e guard genuinely needs its own line in `architecture.md` or that test, add it on top of main's current text rather than resolving into this branch's copy.

Your actual contribution here is untouched by any of that — `eval/harness/tests/unit/test_skill_reachability_e2e.py` is new, exists nowhere else, and is what I will be reviewing.

## Nothing else from me

The rest of the `main → pr-2112` diff is `#2159` staleness (`init-project` SKILL.md, its tests and rubric, its run log), not anything you did — it clears on the rebase.

Two follow-ups from #2060 are recorded there and are **not** yours to carry into this PR: the stacked `_004`/`Correctness` comment, which is free to repair since the `.ann.json` is in none of the run log's 34 snapshot entries, and `project-status/SKILL.md:60`, which is on #1750. The CRUD UI card and the empty-`references/`-dir finding are yours to file whenever, and neither blocks this.

florencemashipei:
Correcting my own last line: I said I would convert this into a real review of the guard once rebased. **I should not be that reviewer.**

`eval/harness/tests/unit/test_skill_reachability_e2e.py` matches only `*.py` in CODEOWNERS (line 31), which is **@PioneerAIAcademy/senior-developers first**. No `/eval/harness/` prefix rule covers it — only `/eval/harness/judge/` is genealogist-first — so nothing later matches and the extension rule stands. My seat on this PR was the duplicated `/eval/runlogs/` and `/packages/engine/plugin/skills/` files, and those disappear in the rebase.

So: ping me when it is rebased, I clear my hold, and that is the whole of my involvement. The guard itself is @EdmondOware's read. Don't wait on me for it — that would be me reviewing a Python harness test because I had already read the context, which is not the same as it being mine.

Everything in my previous comment stands: take main's side on all six files, re-apply nothing.

florencemashipei:
Two corrections to what I posted here today, both mine.

**1. "I have not reviewed this PR on its merits" was wrong.** I reviewed the guard on 3 Sep and approved it — 159 run logs and the 12 dark skills reproduced from scratch, all four arms broken and watched to fail, the subagent route checked for a path the scan might miss. That review stands, and so do its two follow-ups: the brand-new-skill message that says "was reachable before" and points at a shrink-only list, and the failure text asking for a reason a `frozenset` cannot hold. Both are still follow-ups, not blockers.

**2. Disregard my handing the guard to @EdmondOware.** CODEOWNERS does put `*.py` in the developers' lane first, which is what I was reading off — but that is beside the point when the review has already happened. @EdmondOware, nothing is waiting on you here.

## What the hold is actually for now

Only this: the diff still carries the pre-#2060 copies of `docs/architecture.md`, `skill-reference-reachability.test.ts` and the `_004` `.ann.json`, and against current main those are **reverts** rather than duplicates. My 3 Sep approval said the sequence was #2060 lands, you merge main, and then the approval fits the diff. Only the first of those has happened.

I am keeping `CHANGES_REQUESTED` until the second, and the reason is mechanical rather than a doubt about your work: `dismiss_stale_reviews_on_push: false` and `required_approvals: 2` mean this PR already holds two standing approvals, mine and @EdmondOware's, which my review is currently the only thing suppressing. Clear it early and the conflict resolution — the one edit nobody re-reads — would merge unreviewed.

**So: merge main, take main's side on all six files, and ping me.** The diff becomes `eval/harness/tests/unit/test_skill_reachability_e2e.py` alone, which is what I approved on 3 Sep, and I will restore that approval the same day.

benter-070:
@florencemashipei — done, same day as asked. **The diff is one file.**

```
$ git diff --stat origin/main...HEAD
 eval/harness/tests/unit/test_skill_reachability_e2e.py | 143 +++++++++++++++++++
 1 file changed, 143 insertions(+)
```

All six duplicated files are gone. Ready for the real review of the guard whenever
the queue wants it.

**One thing in the merge worth your eye**, since it is precisely the case you named
— "a conflict resolution is the one edit nobody reviews". Merging main conflicted on
exactly one file, `v1_2026-08-31_09-15-36.ann.json` (add/add). I resolved it by
taking main's version and then **verified the result is byte-identical to main's
blob** rather than trusting the resolution: `ut_project_status_004` carries the
corrected 3 on both base dimensions. Nothing of this branch's uncorrected `ee480c97`
survived.

Also refreshed a stale figure while I was in the body: the corpus is **161 runs**
now, not 159 — two landed since. Population holds at 27 shipped skills, 12 dark, and
the guard's 4 arms pass on the merged tree.

## On your correction — you are right, and it is the more useful half of the review

I cited blob `43684873` as evidence you had seen and accepted the annotation. Your
#2112 review cited it for **score** divergence; the same blob carries the corrected
scores *and* the stacked comment, so citing it was never a clean bill of health for
the file. I read a hash as covering the whole file when it covered one property of
it, and then built a conclusion about your intent on top of that.

The thing I would have needed to check is the one I did not: I verified Richard's
text was on `_004` and nowhere else, and stopped. I never re-counted the headers
after the final save, despite having counted them earlier in the day and flagged
them as "untidy, not wrong". That judgement was mine and it was wrong — a
justification that reads `Junior: 1` three times against a `corrected_score` of 3 is
not untidy, it is a record that argues against itself.

## The finding is in #2234, not stranded here

Filed before your review and now updated with your measurement inline rather than a
link, per your ask, plus the two things you named:

- **The semantics gap as the actual defect.** `eval/CLAUDE.md` documents the merge as
  *intended* — "which the UI then silently merges with" — as a defence against a
  malformed hand-authored file. Applied to a comment field on a well-formed file it
  is accumulation, not defence. A patch that makes the field replace-on-save without
  settling that intent will read as a regression to whoever wrote that paragraph.
- **The rule-3 companion check**: a correction's comment contains exactly one
  generated header. Cheap, a regex count on text the UI itself emits, and it would
  have failed this file before review. Recorded that it catches mode 1 only —
  nothing ties a comment's content to its `test_id`, so a comment written to the
  wrong test still needs the UI fixed rather than guarded.

Repair of the shipped instance is free (the `.ann.json` is in none of the 34 snapshot
entries), noted on the card so nobody re-runs `project-status` to fix a comment.

benter-070:
@florencemashipei — rebased, and I have taken main's side on all six as you asked.
Verifying it the way you would rather than reporting it:

```
$ git diff --name-only origin/main HEAD
eval/harness/tests/unit/test_skill_reachability_e2e.py
```

Two-dot against main, not three, so this is what actually differs rather than what
the merge base makes it look like. Per file:

| file | vs main |
|---|---|
| `project-status/SKILL.md` | byte-identical |
| `v1_2026-08-31_09-15-36.json` | byte-identical |
| `v1_2026-08-31_09-15-36.ann.json` | byte-identical (`_004` at **3** on both) |
| `docs/architecture.md` | byte-identical |
| `skill-reference-reachability.test.ts` | byte-identical |
| `v1_2026-08-11_19-00-27.json` | absent on both — #2060's deletion honoured |

Nothing re-applied. The one conflict was the `.ann.json` (add/add); I resolved it
by taking main's version and then compared the resulting blob to
`origin/main`'s rather than trusting the resolve, because you named that edit as
the one nobody re-reads. The reverting copies of `architecture.md` and the
reachability test are gone with it — the false "unreachable in every environment"
claim is not in this branch in any form.

## Your two 3 Sep follow-ups are fixed — `3207fec3`

Not blockers, but both were in the one message and it was wrong in the way you
said:

- **"that were reachable before"** is false for a skill added since the baseline.
  It was never in the recorded population, so nothing regressed.
- **"add them to `DARK_SKILLS_2026_09_01` with a reason"** cannot be done. The set
  is a `frozenset` of bare names and holds no reason — and adding an entry
  contradicts the shrink-only rule this module's own docstring states, which is
  precisely how a regression would become the new baseline.

The message now splits the two causes and gives each the opposite response it
needs: reconsider the change, or give the new skill e2e coverage / re-baseline
deliberately with the reason in the comment above the set, since the set cannot
carry one. Proven the same way as the original four arms — removing `timeline`
fires the new text, restoring it passes 4.

## One refreshed figure

The corpus is **161 runs** now, not 159 — two landed since the measurement. Still
27 shipped skills and **12 dark**, so the recorded set survives the update
unchanged. Body updated; the stacking note is replaced by what the diff actually
is now.

CI green, 0 behind main. Over to you whenever — and understood that your seat here
was the duplicated files rather than the guard, so if you would rather let your
3 Sep approval stand and simply clear the hold, that is entirely your call.

florencemashipei (approved):
Approved. This supersedes my ordering hold — a dismissal alone would have left my 3 Sep approval standing against a diff that has changed since, which is not the same thing.

Verified rather than read, at `3207fec3d`:

| check | result |
|---|---|
| diff vs main | one file, `test_skill_reachability_e2e.py` |
| the six shared files | byte-identical to main; `v1_2026-08-11_19-00-27.json` absent on both, #2060's prune honoured |
| `.ann.json` blob | `4368487` — identical to main's, `_004` at 3 on both base dimensions |
| corpus | 161 result jsons via `all_result_jsons()` on current main — your refreshed figure is right, mine was measured on a stale tree |
| guard | 4 arms pass in the harness venv |

**The conflict resolution holds.** That was the edit I named as the one nobody re-reads, so I checked the result rather than the method: the resolved `.ann.json` is main's blob exactly, and this branch carries the reverting copies of `docs/architecture.md` and `skill-reference-reachability.test.ts` in no form. The false "unreachable in every environment" claim is gone from the branch.

**Both 3 Sep follow-ups are fixed, and I proved the new message rather than taking the note for it.** Dropping `timeline` from the recorded set fires the new text naming both causes with opposite responses — including "Do NOT add the skill to `DARK_SKILLS_2026_09_01`: that set only shrinks", which is the instruction the old message got backwards. Restoring it passes 4. You did not convert the set to `dict[str, str]`, and that is fine: the message is now truthful about where a reason lives, which was the ask underneath item 2.

One mechanical thing before you merge, not a review item: `5f355a5fe` landed after your last comment, so GitHub has this as `BEHIND` — update the branch first. Nothing in this approval depends on it.

## On the two corrections

Yours was the harder one to make, and you made it properly — you named the inference (a blob hash covering one property read as covering the file) and then named the check you skipped, which is the part that generalises. "Untidy, not wrong" was the judgement that let it through, and re-reading it as a record that argues with itself is the right conclusion.

Mine to own too: I told this thread I had not reviewed the guard on its merits when I had, in depth, on 3 Sep — and then routed it to @EdmondOware on that basis. Reading CODEOWNERS for who *should* review something is no substitute for checking who already had.

#2234 covers the rest, with the measurement inline and both additions I asked for. The distinction you drew there is the sharp one: the companion check catches mode 1 only, because nothing ties a comment's content to its `test_id` — so a comment written to the wrong test still needs the UI fixed rather than guarded. That is worth keeping at the top of the card.

Nice work on the guard. It is a better test than the one I approved on 3 Sep.

PR #2114, opened 2026-09-01
Restore senior-queue, and converge it on CI completions instead of the cron alone

### LINKED ISSUE #1968: senior-queue never runs on approval, so a ready PR can sit unqueued indefinitely (3 found at once)
> **Reviewed 2026-08-27 before junior handoff. Closed as not planned 2026-08-31, reopened
> 2026-09-01.** The close was ruled on the premise that we do not need this mechanism at all —
> senior-queue was being deleted in PR #2084. That premise no longer holds: PR #2084 was
> reverted because the label is the routing signal seniors actually see, so the latency this
> card describes is live again. Decision below; **implemented in PR #2114**, which closes this.

## Decision (2026-09-01): `workflow_run` sweep, not a new identity

Add a `workflow_run` trigger on the repo's CI workflows and treat it as a sweep event, so the
existing all-open-PRs path re-converges the queue within minutes of any CI activity. Chosen over
granting the Projects App `Issues: write` (a deterministic approval trigger, but an org-admin
permission change on an identity holding a private key, for a path that fires ~1% as often as CI
does) and over a detect-only check (visibility without a fix).

Do not wire it per-PR off `context.payload.workflow_run.pull_requests` — that array is **empty
for fork-authored PRs**, which is exactly the population the forbidden approval trigger already
fails on. It must go through the `sweep` branch at `senior-queue.yml:337`.

Record the decision and the alternative it beat in the header ledger at
`.github/workflows/senior-queue.yml:109-121`, next to the `pull_request_review` prohibition it
sits beside.

### Verification cannot happen before merge, by construction

Option A was recorded with the caveat that base-repo `issues: write` on `workflow_run` is
**documented GitHub behavior but unverified in this repo**, and that the PR must prove it from a
run log. That proof is not obtainable pre-merge: `workflow_run` only fires from the workflow file
on the **default branch**, so no `workflow_run`-triggered run of `senior-queue` can exist until
this is on `main`.

So the check moves to just after merge: confirm a `senior-queue` run appears with
`event: workflow_run`, that it is green, and that it applied a label. If the token turns out not
to carry `issues: write` there, the run reddens on the `addLabels` 403 rather than failing
silently — the same shape that condemned `GITHUB_TOKEN` on the review trigger in PR #1564.

**Touches:** .github/workflows/senior-queue.yml

`senior-queue` labels a PR `ready-for-senior-*` once CI is green and a peer has
approved. It never runs on the event that most often makes a PR ready — the
approval itself — so a PR can sit fully ready and invisible to both senior
queues indefinitely.

## Triggers today

```yaml
on:
  pull_request_target:
    types: [opened, reopened, synchronize, ready_for_review]
  check_suite:
    types: [completed]        # documented DEAD — GITHUB_TOKEN events don't start runs
  schedule:
    - cron: "*/30 * * * *"
  workflow_dispatch:
```

There is no `pull_request_review`. The last `pull_request_target` type that fires
is `synchronize` — a push. So after the final push, an approval changes nothing:
the workflow does not run, and the label is not added.

The `schedule` is the only thing that covers "became ready without a push", and
the workflow's own comment concedes it is best-effort ("GitHub delays or drops
them under load").

## Observed

PR #1909, 2026-08-27. Peer-approved 11:51:38Z, all 11 checks green, zero
unresolved review threads — genuinely ready to merge pending one senior approval.
No label. It was not in either queue, and the reviewer had to ask for a senior by
hand, which is how it was noticed at all.

Scheduled runs of this workflow at that moment:

| Last scheduled run | Gap |
|---|---|
| 2026-08-27T03:26:56Z | **>10 hours** against a 30-minute cron |

A manual `workflow_dispatch` swept it correctly and added
`ready-for-senior-genealogist` — along with **#1963 and #1929**, both also ready
and unlabelled. Three PRs stuck behind the same gap, found only because someone
complained about one of them.

## Why this is `nothing-checks`

The mechanism looks like it works: the workflow is green, the label exists, the
queue searches return results. What is missing is any signal when a ready PR is
*not* in the queue. A dropped cron and a missing trigger both fail silently and
identically — as an empty queue, which is indistinguishable from "nothing is
ready". The reviewer asking by hand is currently the only detector.

## Constraints any remedy must satisfy

1. **No `pull_request_review` under `GITHUB_TOKEN`.** It has no `_target` variant, so a fork
   PR's run is read-only and the label write 403s *while the run goes green*. Removed
   2026-08-13, verified on PR #1564, and forbidden in bold at
   `.github/workflows/senior-queue.yml:109-121`. This is not hypothetical here: PR #1929, one of
   the three stuck PRs above, is itself a fork PR
   (`gh pr view 1929 --json isCrossRepository`), so this "fix" would have missed one of its own
   three instances.
2. **A remedy that returns before writing goes green.** This file's header lists three past
   changes that went green while doing nothing (`senior-queue.yml:66-78`). Anything landed here
   needs a live observation, not a green CI run — see "How to verify".
3. **The schedule cannot be the answer, and is worse than its cron says.** Measured 2026-08-27
   over `gh run list --repo PioneerAIAcademy/cowork-genealogy --workflow senior-queue.yml
   --event schedule --limit 200 --json createdAt` (window 2026-08-20T06:43Z → 2026-08-27T14:30Z):
   **200 runs delivered where a `*/30` cron predicts 352**; median gap 51 min, p90 76 min, max
   **664 min**. Re-run the command to refresh; GitHub's scheduled-run delivery varies with
   platform load, so this moves.

## Sites a trigger change touches

- [ ] `.github/workflows/senior-queue.yml` `on:` (line 123) — add the new trigger. Leave the
      dead `check_suite` block (line 126) in place *or* delete it in the same PR, but do not
      leave two comments disagreeing about what covers "CI turned green".
- [ ] `.github/workflows/senior-queue.yml:337-338` — `const sweep = eventName === 'schedule' ||
      'workflow_dispatch'`. **A new sweep-shaped event must be added here too**, or the script
      falls through to `context.payload.pull_request.number` and throws on a payload that has
      none.
- [ ] `.github/workflows/senior-queue.yml:155-163` — `concurrency`. The group expression
      resolves to the literal `'sweep'` for any event without `pull_request.number` or
      `check_suite.id`, and `cancel-in-progress: false` **queues** rather than drops. A
      high-frequency trigger therefore builds a serial backlog: 400 non-schedule workflow runs
      fired in 2h46m on 2026-08-27, against 23 open PRs each costing ~5 API calls per sweep.
      Give the new event its own group with `cancel-in-progress: true`, or restrict which
      workflows can trigger it. Say in the PR which you chose.
- [ ] `.github/workflows/senior-queue.yml:87-97` — the "THE SCHEDULE IS WHAT MAKES THE QUEUE
      CONVERGE" paragraph, and the "WHAT THIS COSTS: up to 30 minutes" line below it, both
      become false. Update them, and add the decision + the alternative it beat to the ledger at
      lines 109-121 — this file is the only place the senior queue is documented at all
      (`grep -rl "ready-for-senior" docs .github *.md` returns it and nothing else).
- [ ] `.github/tests/senior-queue-readiness.test.mjs` — it lifts the script body out of the YAML
      by regex (`scriptBody`, `grab`). Re-run it after any edit to the `script: |` block; it
      fails loudly if a function was renamed.
- [ ] `.github/workflows/workflow-logic-tests.yml` — if you add a new test file, it needs its own
      **job** here. The `paths:` glob already covers `.github/tests/**`, so a new test file will
      look wired up while never running — the same failure this file's `issue-gate` job comment
      records.

## How to verify

- `node .github/tests/senior-queue-readiness.test.mjs` — the only automated check, and it covers
  **only** the two pure decision functions. It cannot see the `on:` block, the token's
  permissions, or whether a label write landed.
- `gh workflow run senior-queue.yml -f dry_run=true` — exercises the readiness rules against
  every live open PR with no writes.
- **Live, and required:** after merge, find a run of `senior-queue` started by the new trigger
  (`gh run list --workflow senior-queue.yml --event <event>`), open its log, and confirm the
  `Issues: write` line in the token-permissions block — the same evidence shape used to condemn
  `pull_request_review` on PR #1564. Confirm it **on a fork-authored PR**; an internal-PR check
  would pass while the fork path is broken. Paste both in the PR body.
- This is a `nothing-checks` item and the fix does not remove the label: nothing reports a PR
  that is ready-and-unqueued. If your PR does not add such a check, say so explicitly rather
  than closing this issue.

DallanQ:
Restores `senior-queue.yml`, which PR #2084 deleted, and adds the `workflow_run`
sweep that issue #1968 asked for.

## Why the deletion was wrong

PR #2084 argued that CODEOWNERS already requests both senior teams when a PR
opens, so a senior has the same queue natively. The information is there — but the
comparison was wrong. It weighed the label against GitHub's review-request queue as
*two things a senior navigates to*, when the property that actually matters is that
**a label renders inline on the PR list everyone already looks at**. A review
request renders nowhere in that list; you have to filter for it. Replacing an
at-a-glance signal with a saved search is a worse product whatever the search
returns.

The two labels were never deleted, so nothing needs recreating — they are still live
on 5 open PRs. This restores the workflow that maintains them.

## Restored unchanged

- `.github/workflows/senior-queue.yml`
- `.github/tests/senior-queue-readiness.test.mjs`
- the `readiness` job and the `senior-queue.yml` trigger path in `workflow-logic-tests.yml`
- `docs/task-lifecycle.md`'s statement that a ready PR carries a `ready-for-senior-*`
  label — now saying *why* that matters: it is on the PR list itself, so a senior
  sees what is waiting without filtering for it

## Restored with one deliberate change

`.github/tests/codeowners-routing.test.mjs` comes back **without its second check**.
That check — every rule names both senior teams — is a statement about who may
**merge**, not which queue a PR lands in, so it must hold whether or not this
workflow exists. PR #2084 correctly moved it to `codeowners-teams.test.mjs`, which
reads CODEOWNERS directly rather than through the parser lifted out of the workflow,
and it stays there.

Re-adding it here would put two guards on one condition and make both untestable.
Proven, not assumed — with `senior-genealogists` dropped from the `*.ts` line:

```
codeowners-teams.test.mjs    FAIL   <- caught it
codeowners-routing.test.mjs  PASS   <- correct, no longer its job
```

## The `workflow_run` sweep (issue #1968)

Restoring the workflow restores its latency problem. `pull_request_review` cannot be
used here — GitHub has no `_target` variant, so a fork PR's review run gets a
read-only token and the label write 403s (proven on PR #1564) — which left the
`*/30` cron as the only thing observing either readiness event. Measured at 200 of an
expected 352 runs on 2026-08-27, with an 11h23m gap on 2026-08-28.

`workflow_run` fires in the **base-repo context** even for a fork PR's CI, so its
`GITHUB_TOKEN` keeps the `issues: write` the review trigger is denied, and it needs no
new identity. Triggered on the repo's eight CI workflows.

**Handled as a sweep, never per-PR.** `context.payload.workflow_run.pull_requests` is
empty for fork-authored PRs — the exact population the forbidden approval trigger
already fails on — so narrowing to it would rebuild that hole somewhere new. Sweeping
is also what makes the trigger worth having: any CI completion re-converges every open
PR, not just the one whose CI finished.

Chosen over granting the Projects App `Issues: write` (the only deterministic remedy,
but an org-admin permission change on a private-key-holding identity, widened to label
writes across every issue and PR, for a path firing ~1% as often as CI does), and over
a detect-only check (visibility without a fix). **The schedule stays as the floor** —
`workflow_run` converges in minutes while CI is busy and does nothing on a quiet
weekend.

**It costs nothing extra in Actions minutes.** `workflow_run` carries neither a
pull_request nor a check_suite payload, so it falls through to the existing
concurrency group's `sweep` key: a burst of CI completions collapses to one running
sweep plus one pending, and since every sweep evaluates every open PR, the survivor
does the work the cancelled ones would have.

### New guard, because this rots silently

`workflow_run` matches a workflow by its `name:`, not its filename, and a name that
matches nothing simply never fires — no error, no warning, no red run. The sweep just
stops converging and the queue falls back to the cron. Renaming a CI workflow is an
ordinary thing to do and nothing else in the repo connects the two files.

`senior-queue-readiness.test.mjs` now asserts every `workflow_run` target resolves to
a real workflow `name:`. **Proven to fail before committing:**

- red on renaming a target (`"Engine tests"` → `"Engine CI"`), naming the 17 known workflows
- red on removing the `workflow_run` trigger entirely
- green on restore, with the workflow byte-identical to the backup

## Verification

All six workflow-logic tests pass:

- `project-status-sync-routing` · `codeowners-teams` · `codeowners-routing` ·
  `senior-queue-readiness` · `check-runlogs-post-gate-lints` · `gate-issue-create`

Plus: every `workflow_run` target checked against the real workflow names (8/8); every
`run:` step in `workflow-logic-tests.yml` points at a file that exists; the YAML
anchor/alias pair is intact.

Closes #1968.

That issue was resolved as not planned on the premise that we do not need this
mechanism at all. The revert above voids that premise, so it has been reopened and
its `needs-decision` label removed — the decision is option A, which is what this
implements.

## The one thing that cannot be verified before merge

Option A was recorded with the caveat that base-repo `issues: write` on `workflow_run`
is **documented GitHub behavior but unverified in this repo**, and that the PR must
prove it from a run log. That proof is not obtainable here: `workflow_run` only fires
from the workflow file on the **default branch**, so no `workflow_run`-triggered run of
`senior-queue` can exist until this is on `main`. Nothing in this PR's CI can close
that gap, and no amount of local testing substitutes.

**So the check moves to just after merge:** confirm a `senior-queue` run appears with
`event: workflow_run`, that it is green, and that it applied a label. The failure mode
is loud rather than silent — if the token does not carry `issues: write` there, the run
reddens on the `addLabels` 403, which is the same shape that condemned `GITHUB_TOKEN`
on the review trigger in PR #1564. The cron remains the floor either way, so a failure
here degrades to today's behaviour rather than breaking the queue.

https://claude.ai/code/session_01LJMa4biet16dH62y8KZb9d

### DIFF
--- a/.github/tests/codeowners-routing.test.mjs
+++ b/.github/tests/codeowners-routing.test.mjs
@@ -1,0 +1,137 @@
+// Review routing truth table for .github/CODEOWNERS.
+//
+// WHY THIS EXISTS. senior-queue.yml resolves CODEOWNERS itself to decide which
+// senior team's queue a PR lands in, and its parser understands only a few
+// pattern shapes — anything else it drops with a warning nobody reads. A
+// pattern GitHub honours but that parser cannot read enforces the merge
+// correctly while routing the review to the wrong team, with no visible
+// symptom. This file resolves paths through that same parser, lifted out of the
+// workflow rather than copied, so the two cannot drift.
+//
+// WHAT IT CANNOT COVER. That the named teams exist. A rule naming a nonexistent
+// team is silently dropped by GitHub and this file cannot see it — check
+// `gh api /repos/{owner}/{repo}/codeowners/errors` after editing CODEOWNERS.
+//
+// Run: node .github/tests/codeowners-routing.test.mjs
+import fs from 'node:fs';
+
+const CODEOWNERS = '.github/CODEOWNERS';
+const WORKFLOW = '.github/workflows/senior-queue.yml';
+
+// Pull the `script: |` block scalar out of the raw YAML and dedent it.
+function scriptBody(path) {
+  const lines = fs.readFileSync(path, 'utf8').split('\n');
+  const start = lines.findIndex(l => /^\s*script: \|\s*$/.test(l));
+  if (start === -1) throw new Error(`${path}: no "script: |" block found`);
+  const indent = lines[start].match(/^\s*/)[0].length;
+  const out = [];
+  for (const line of lines.slice(start + 1)) {
+    if (line.trim() !== '' && line.match(/^\s*/)[0].length <= indent) break;
+    out.push(line.slice(indent + 2));
+  }
+  return out.join('\n');
+}
+
+const body = scriptBody(WORKFLOW);
+const grab = (re, label) => {
+  const m = body.match(re);
+  if (!m) {
+    throw new Error(
+      `could not extract ${label} from ${WORKFLOW} — it was renamed or reformatted. ` +
+      `Update this regex, and re-check that this file still tests the real parser.`);
+  }
+  return m[0];
+};
+
+// scriptBody() dedents to column 0, so a top-level `}` in column 0 ends a function.
+const loadSrc = grab(/async function loadCodeowners\(\) \{[\s\S]*?\n\}/, 'loadCodeowners');
+const ownersSrc = grab(/^function ownersOf\(file, rules\) \{[\s\S]*?\n\}/m, 'ownersOf');
+const queueSrc  = grab(/^function queueTeamOf\(file, rules\) \{[\s\S]*?\n\}/m, 'queueTeamOf');
+
+// loadCodeowners() reads the default branch over the API; here it reads disk.
+const stubGithub = `
+  const github = { rest: {
+    repos: {
+      get: async () => ({ data: { default_branch: 'main' } }),
+      getContent: async () => ({ data: {
+        content: Buffer.from(require('fs').readFileSync(${JSON.stringify(CODEOWNERS)}, 'utf8')).toString('base64'),
+      } }),
+    },
+  } };
+  const owner = 'o', repo = 'r';
+`;
+const AsyncFunction = Object.getPrototypeOf(async function () {}).constructor;
+const { loadCodeowners, ownersOf, queueTeamOf } = await new AsyncFunction('require',
+  `${stubGithub}${loadSrc}\n${ownersSrc}\n${queueSrc}\nreturn { loadCodeowners, ownersOf, queueTeamOf };`)(
+  (await import('node:module')).createRequire(import.meta.url));
+
+const { rules, unknown } = await loadCodeowners();
+
+let failures = 0;
+const fail = msg => { failures++; console.error(`FAIL  ${msg}`); };
+
+// 1. Every pattern must be one senior-queue.yml can resolve.
+if (unknown.length) {
+  fail(`${CODEOWNERS} patterns senior-queue.yml cannot match, so it routes those ` +
+       `paths to the wrong review queue: ${unknown.join(', ')}\n` +
+       `      Teach loadCodeowners() the shape rather than dropping the pattern.`);
+} else {
+  console.log(`ok    all ${rules.length} CODEOWNERS rules are shapes senior-queue.yml can resolve`);
+}
+
+const D = 'senior-developers', G = 'senior-genealogists';
+
+// 2. Every rule names BOTH senior teams — NOT checked here. That rule is about
+//    who may MERGE, not which queue a PR lands in, so it must hold whether or
+//    not this workflow exists. It lives in codeowners-teams.test.mjs, which
+//    reads the file directly instead of through the parser lifted below. Do not
+//    re-add it here: two guards on one condition make both untestable, since
+//    neither can be shown to fail on its own.
+
+// 3. Which QUEUE each path lands in, stated as a table so a reordering that
+//    inverts last-match-wins fails here rather than in review. This is the
+//    FIRST-listed team, which is what senior-queue.yml labels for — widening
+//    who may approve must not change who gets queued. `null` means no senior
+//    team is required, and no label; the two ordinary approvals still are.
+const TABLE = [
+  // Code and the infra that breaks silently.
+  ['packages/engine/mcp-server/src/index.ts', D],
+  ['apps/server/app/main.py', D],
+  ['apps/server/pyproject.toml', D],
+  ['.github/workflows/js-tests.yml', D],
+  ['scripts/windows/test-all.bat', D],
+  ['Makefile', D],
+  ['apps/electron/Makefile', D],
+  ['.gitattributes', D],
+  ['docs/specs/schemas/research.schema.json', D],
+  // Genealogist content, including where it lives in developer-looking files.
+  ['packages/engine/plugin/skills/citation/SKILL.md', G],
+  ['packages/engine/plugin/skills/timeline/scripts/build.py', G],
+  ['packages/engine/plugin/agents/gps-mentor.md', G],
+  ['eval/fixtures/scenarios/x/research.json', G],
+  ['eval/tests/e2e/x/fixture.json', G],
+  ['eval/runlogs/unit/citation/run.json', G],
+  // Documentation and incidental files need no senior. Losing these to a team
+  // is the regression this table is here to catch.
+  ['README.md', null],
+  ['docs/architecture.md', null],
+  ['docs/specs/research-schema-spec.md', null],
+  ['.claude/skills/fill-ready/SKILL.md', null],
+  ['apps/web/src/styles.css', null],
+  ['apps/web/public/favicon.svg', null],
+  ['.gitignore', null],
+  ['scripts/test.sh', null],
+];
+{
+  const before = failures;
+  for (const [file, want] of TABLE) {
+    const got = queueTeamOf(file, rules);
+    if (got !== want) {
+      fail(`${file} queues to ${got ?? 'no senior team'}, expected ${want ?? 'no senior team'}`);
+    }
+  }
+  if (failures === before) console.log(`ok    ${TABLE.length} queue-routing assertions`);
+}
+
+console.log(failures ? `\n${failures} check(s) failed` : '\nall checks passed');
+process.exit(failures ? 1 : 0);
--- a/.github/tests/senior-queue-readiness.test.mjs
+++ b/.github/tests/senior-queue-readiness.test.mjs
@@ -231,8 +231,12 @@ for (const { label, threads, approvals, want } of SUPERSEDED) {
     fail(`${WORKFLOW}: no \`workflow_run:\` trigger — the sweep that substitutes for the ` +
          `forbidden approval trigger is gone; the cron is the only thing left converging`);
   } else {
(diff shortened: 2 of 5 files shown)


DallanQ:
Both findings are right and both are fixed in `ca86623e`. The API one would have taken down every `GITHUB_TOKEN` workflow in the repo, not just this queue — thank you.

I re-derived rather than accepting, and every input you gave checks out independently:

- org plan is **`free`**, so the limit is 1,000 REST/hour, not the Enterprise Cloud 15,000
- **129 completions** across the eight targeted workflows in the hour `2026-09-01T11` — I measured the same hour and got the same number
- **42s** median sweep duration — same
- **~5 REST calls per non-draft PR**: `pulls.get`, `listReviews`, `checks.listForRef`, `getCombinedStatusForRef`, `listFiles`, plus one GraphQL for threads. 24 non-draft + 4 draft + overhead ≈ 127

**Where I diverged from your patch: the debounce interval alone doesn't fit the budget.**

Your ~780 REST/hour prices the `workflow_run` sweeps only. It omits two things that share the same 1,000: the cron's own 2 sweeps/hour (~254), and this workflow's per-PR `pull_request_target` runs. That's easy to miss from the data, because `senior-queue` shows **0 runs** in the hour you sampled — it was deleted from `main` at the time, so the baseline you measured is a repo without it. Add it back and 6 sweeps/hour at 130 calls puts sweeps alone at ~1,040 before anything else runs.

So I took the debounce **and** made the sweep cheap enough to justify it:

- **`DEBOUNCE_MIN = 15`**, your mechanism, with one change: skip cancelled runs when reading the last one. Concurrency cancels the *pending* run when a newer event arrives, and debouncing against one of those would skip on a sweep that never ran.
- **`workflow_run` sweeps evaluate only PRs updated in the last hour.** `updated_at` is already on the objects the sweep lists, so the filter costs nothing, and an approval — the event this trigger exists for — bumps it. The **cron sweep stays unfiltered** and remains the floor, so anything the filter misses is caught within 30 minutes: never worse than today.

That reframes the win honestly, and I corrected the header to say so: this was never really about beating the cron's 30-minute cadence. It is about not depending on GitHub's best-effort scheduler, which is what produced the 11h23m gap in the first place.

`actions: read` added exactly as you specified — and you were right about why: the explicit `permissions:` block makes every unlisted scope `none`.

**On the guard's blind spot — that one is embarrassing and the lesson is the useful part.** I wrote that guard, proved it fails on the mutation I had in mind (renaming a target), and shipped it calling it proven. I never tested the shape a *different* editor would write. Your unquoted case is exactly right, and it reported `ok all 8` while nine entries sat in the file. Widened to all three scalar styles and re-proven across the matrix:

```
Totally Fake Workflow            -> FAIL  <- caught
'Totally Fake Workflow'          -> FAIL  <- caught
"Totally Fake Workflow"          -> FAIL  <- caught
baseline restored                -> PASS
```

**On the non-blocking suggestion — declining, with a reason.** Adding `workflow-logic-tests` to the trigger list buys no coverage now that the sweep is repo-wide and debounced: any of the eight already re-converges every recently-updated PR, and a ninth trigger only adds `listWorkflowRuns` debounce calls that mostly return "skip". If a PR could complete CI with *none* of the eight firing that would change the answer, but the cron covers that case by construction.

Issue #2104, opened 2026-09-01
3 negative tests fail on triggering — the description already excludes the case


DallanQ:
**Touches:** packages/engine/plugin/skills/{translation,timeline,locality-guide}/SKILL.md (description frontmatter), eval/runlogs/unit/{translation,timeline,locality-guide}/ (new run log + .ann.json per skill)

> **Rewritten 2026-08-31 after `/review-ready`.** The earlier version of this card named 12 tests and said "no prose edits, cheap". Both were wrong. Only **3 fail in the newest committed run log**, 2 more are already green on an open PR, and 4 are stale one-off flaps. The cost framing was also wrong — see "Cost" below. Re-derive the count before starting; the command is at the bottom.

## What is actually failing

| test | `skills_invoked` | turns | judge min | what the main thread did instead |
|---|---|---|---|---|
| `ut_translation_003` | `[]` | 3 | 1 | called `wikipedia_search`, answered in 217 words |
| `ut_timeline_008` | `[]` | 11 | 1 | `project_context` + 6× `research_query`, 482-word identity analysis |
| `ut_locality_guide_003` | `[]` | 4 | 1 | called `wikipedia_search` **and saved a file** |

No skill loaded; the main thread did the neighbouring skill's work with real tool calls and real side effects. The judge agrees it was wrong (min dimension score 1).

## Do this first: work out which binding missed

`docs/architecture.md` §4, "Fix a skill that isn't triggering", says a triggering miss has **three** causes with different fixes. Do not skip to the description.

1. **Did the request go through `/research`?** Then the routing table in `research/SKILL.md` decided first, and no unit suite covers it — a live `make e2e-run` is the only instrument.
2. **Was the skill addressed directly?** Then it is the `description` (§3.2).
3. **Did it trigger and then do the wrong thing?** Not a triggering failure — classify with §3.6.

These three tests address the skill directly, so (2) is the likely branch — confirm per test rather than assuming.

## The description already carries the exclusion, and the test still fails

This is the load-bearing fact and the reason this is not a quick edit. **An earlier version of this card cited these same lines as proof "the bodies are already correct" — that was backwards. They are the `description` frontmatter, which is exactly the lever this card tells you to pull.**

[...]

PR #2131, opened 2026-09-01
convert-dates: don't assert a Gregorian reading for dates before the calendar existed

mercyokum:
**The issue:** `ut_convert_dates_004` was failing. The `convert-dates` skill's worked example taught the model to present a bare Gregorian date for 14 September 1582 (Madrid) — a date that comes before Spain actually adopted the Gregorian calendar (15 October 1582). The real `convert_calendar` tool correctly refuses to compute a day offset for a date like this (there's no Gregorian calendar yet to convert into), but the skill's own example showed a confident, unlabelled Gregorian answer for it anyway.

**The fix:**
- Added an explicit presentation rule, in its own paragraph, for when the tool rejects a conversion because the date precedes 15 October 1582 anywhere: show the date once, labelled Julian / Old Style, and never show a "Gregorian date" field at all (not even as "unchanged").
- Replaced the worked example with a date the tool actually accepts, away from any calendar-switch boundary — 10 June 1650, Julian, England (England didn't adopt Gregorian until 1752, so this date is genuinely Julian and the pre-1700 10-day offset applies correctly).
- Reworded "always preserve both forms" so it only applies when the tool actually returns a converted date, so it can't be read as contradicting the new rejection-case guidance.
- Added a deterministic validator, `test_no_gregorian_equivalent_gets_tool_rejection` (gated on the `no-gregorian-equivalent` tag already on `ut_convert_dates_004`'s test JSON): asserts `convert_calendar` was actually called and returned `ok: false`, so the mechanical half of this fix can't silently regress unnoticed by a lucky judge call. The narrative half (never showing a "Gregorian date" field) stays judge-graded, same as every other genealogical-presentation dimension in this rubric.

While verifying the fix, found and fixed one more small, unrelated gap: `ut_convert_dates_008` (a date that's already Gregorian, so no conversion applies at all) sometimes had the skill calling the tool with an empty correction request, which the tool correctly rejects. Added one sentence telling the model to skip the tool call entirely when no jurisdiction's calendar transition applies.

Verified with the full suite (`v1_2026-09-01_14-32-09`, 16 tests): `ut_convert_dates_004` now passes (Correctness and Genealogical presentation both moved 1→3 in the `gate-skill` comparison), all three named hold-outs still pass (`ut_convert_dates_002`, `_006`, `_009` — the source issue named a fourth item, `protestant-german-1700.json`, but that's the same test as `_009` cited by filename instead of id), and `ut_convert_dates_008` now passes too. 15/16 pass overall.

**Deferred issues found:**
- `ut_convert_dates_012` aborted on this run (an SDK/execution error, retried twice, still errored — not a judged failure). It is **not** tracked in #2104: that issue's own text explicitly declines to track it ("separate card if you want it tracked"). Its real history across the four committed run logs is fail → pass → pass → fail — a flap, not a persistent failure. Out of scope for this PR either way.
- `ut_convert_dates_011` came back red on the first full run, but it's judge noise, not a skill problem: the judge's own rationale contradicted its own test's `judge_context`, which explicitly allows the exact behaviour the skill produced (declining to treat a same-jurisdiction date conflict as a calendar issue, and reasoning about it directly instead of insisting on a hand-off). It passed clean on both a standalone re-run and the second full suite run. It was never annotated (it isn't in either run's `review_sample`) — not persistent enough to warrant a rubric edit.


jud-sdev (approved):
Went through this carefully and ran the new validator in a worktree. Clean, well-engineered fix.

The mechanical/narrative split is exactly right: the validator guards "convert_calendar was actually called and the tool rejected it," and the judge keeps grading the narration withholding a Gregorian field. I exercised the validator with synthetic inputs — it fails when the tool was never called, fails when a call comes back `ok: true`, passes on `ok: false`, and handles the server prefix. So it is a real, falsifiable guard, not one that passes vacuously. It also runs and passes on ut_004 in the committed run log and correctly skips the other 15.

The genealogy holds up: Gregorian floor of 15 October 1582, England Julian until 1752, the pre-1700 10-day offset, and 10 June 1650 + 10 = 20 June 1650. Swapping the worked example off the 1582 boundary is the right call, and scoping "preserve both forms" to when a conversion actually returns removes the contradiction with the new rejection rule.

The reworded prose did not break the neighbours — the committed run shows the named hold-outs still passing and ut_008 now passing. And the deferred-issues section is honest: ut_011 is judge noise against its own judge_context, and ut_012 is a flapping SDK abort that #2104 explicitly declines to track.

EdmondOware (approved):
Approved.

The correction is the right one and the carve-out is what makes it right. Refusing to print a Gregorian field for a date before 15 October 1582, while still allowing a proleptic alignment when it is labelled as proleptic, draws the line where it belongs: the objection was never to the arithmetic, it was to asserting a reading the calendar could not yet produce. The old example's 24 September 1582 was numerically fine and historically unsayable, and the new paragraph says why in one place rather than leaving it to be inferred.

The replacement example is better chosen than it needed to be. 10 June 1650 in England is genuinely Old Style, and June clears the 25 March year start, so the example teaches the day offset without pulling dual dating in beside it. It also lands on the distinction people most often get wrong: the 10 days that apply to a 1650 date, not the 11 days Britain used at adoption in 1752. The table's own note that the threshold is 1 March Julian rather than New Year already carried that, and the example now agrees with it.

Scoping "always preserve both forms" to the case where the tool actually returns a converted date closes the contradiction the new rule would otherwise have created.

Splitting the guard is the right call too. The rejection is mechanical and now deterministic; whether the narration withholds the Gregorian field is judgement and stays with the judge.

One thing to tighten. `test_no_gregorian_equivalent_gets_tool_rejection`'s docstring says a tagged test must call `convert_calendar` "with julianToGregorianDay", but `_calendar_calls` matches on the tool name alone, so what the test enforces is that some call happened and none returned `ok: true`. A rejection for an unrelated reason would satisfy it. It is correct on this test today, since any Julian to Gregorian request on that date draws the pre-adoption refusal. Either narrow the filter to calls requesting that correction, or assert the error names the 1582 boundary, so the check guards the reason and not just the refusal.

